PR #1801, opened 2018-04-03
Fix chain recovery by changing getCommonBlock - Closes #1795

### LINKED ISSUE #1795: Fail to recover chain when syncing
### Expected behavior
In case of common block failure, we should delete block from the blockchain.
### Actual behavior
In certain scenarios, when there is a fork in the network and (Step 1) the node tries to get on the right chain by calling loadCommonBlock, (Step 2)it sends the list of block ids of last five rounds, if peer has any of those blocks available, (Step 3) it responds with that block id and height(it responds with most recent block information), after this step, (Step 4) we call `loadBlockFromPeer` with our latest block, if peer has that block, (Step 5) it returns a 33 list of blocks with height greater than height of the block it received. if it doesn't have that block in the database, it returns genesis block and next 33 blocks(which is an odd behavior). and (Step 6) the genesisBlock fails on verification step. Afterwards, sync process ends, and we continue without trying to recover the chain.

I see two issues with our current approach to the common block:
1- While syncing with a peer of greater height, after receiving the response of getCommonBlock (Step 3), we should check if the peer contains our latest block, if it doesn't we should try to recover the chain. (Because that means that peer has a different set of blocks, and we should delete the block and try to follow the longest chain)
2- In case we don't find the block id (Step 5), we should return empty response, as opposed to returning genesisBlock.
### Steps to reproduce
Stated in the actual behavior.

SargeKhan:
### What was the problem?
We weren't deleting block when our last block did not exist for peer
### How did I fix it?
Updated getCommonBlock to return null if last block doesn't exist.
### How to test it?
Run unit tests and create a devnet of forked nodes.
### Review checklist

* The PR solves #1795 
* Linting passes
* Tests pass

### DIFF
--- a/db/repos/blocks.js
+++ b/db/repos/blocks.js
@@ -274,16 +274,15 @@ class BlocksRepository {
 	}
 
 	/**
-	 * Get multiple blocks to be transported to peers.
+	 * Get block to be transported to peers.
 	 *
-	 * @param {string} ids - Comma separated string of ids
+	 * @param {string} id - id of the block
 	 * @returns {Promise}
 	 * @todo Add description for the return value
 	 *
 	 */
-	getBlocksForTransport(ids) {
-		// TODO: Must use a result-specific method, not .query
-		return this.db.query(sql.getBlocksForTransport, [ids]);
+	getBlockForTransport(id) {
+		return this.db.oneOrNone(sql.getBlockForTransport, [id]);
 	}
 
 	/**
--- a/db/sql/blocks/get_block_for_transport.sql
+++ b/db/sql/blocks/get_block_for_transport.sql
@@ -19,8 +19,6 @@
   PARAMETERS: ?
 */
 
-SELECT max(height) AS height, id, "previousBlock", "timestamp"
+SELECT height, id, "previousBlock", "timestamp"
 FROM blocks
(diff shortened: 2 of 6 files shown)


db/repos/blocks.js
```diff
@@ -274,16 +274,15 @@ class BlocksRepository {
...
+	 * @param {string} id - id of the block
 	 * @returns {Promise}
 	 * @todo Add description for the return value
 	 *
 	 */
-	getBlocksForTransport(ids) {
-		// TODO: Must use a result-specific method, not .query
-		return this.db.query(sql.getBlocksForTransport, [ids]);
+	getBlockForTransport(id) {
+		return this.db.oneOrNone(sql.getBlockForTransport, [id]);
```
MaciejBaj: Then it doesn't need to be an array. Just pass `id`. I see the sql query already expects a single id (matching by `=` operator instead of `IN`.

SargeKhan: I am passing it in the form of an array because that's how pgpromise expects parameters to be passed for query; using positional array parameters.. which means that first element of the array will be the first query parameter and so on and so forth.

MaciejBaj: Fair enough :)

PR #1815, opened 2018-04-04
Put transaction in right queue of transaction pool - Closes #1782

### LINKED ISSUE #1782: Transactions are put in wrong queue when delete 2 blocks in a row
### Expected behavior
- Delete one block with transactions type `0` - transactions are put back in `queued` list
- Delete another one block with transactions type `0 `- transactions should be put back in `queued` list
### Actual behavior
- Delete one block with transactions type `0` - transactions are put back in `queued` list
- Delete another one block with transactions type `0` - transactions are put back in `multisignatures` queue instead of `queued` due to presence of `signatures` property that is an empty array.

It happens because after delete first block - second one is loaded from database and get through normalization functions.
### Steps to reproduce
Run functional system rounds tests.
`1.0.0`, probably also `0.9.12`

Tschakki:
### What was the problem?
When deleting 2 or more blocks in a row, the transaction were stored in the wrong queue of transaction pool.

### How did I fix it?
only store transaction in multisignature queue, if they are type 4 or the transaction is sent from a multisignature account.

### How to test it?
`mocha test/functional/system/rounds.js`

### Review checklist

* The PR solves #1782 
* Linting passes
* Tests pass


logic/transaction_pool.js
```diff
@@ -298,20 +298,29 @@ TransactionPool.prototype.getMergedTransactionList = function(reverse, limit) {
  * @param {Object} transaction - Transaction object
  */
 TransactionPool.prototype.addUnconfirmedTransaction = function(transaction) {
+	const that = self;
```
sdrpa: Can ```const that = self``` be removed? Is there any advantage using ```const that = self``` in this function instead of using self?

logic/transaction_pool.js
```diff
@@ -787,14 +795,61 @@ __private.isTransactionInUnconfirmedQueue = function(transaction) {
...
+			function getAccount(waterCb) {
+				if (transaction.senderPublicKey) {
+					hasSender = true;
+					modules.accounts.getAccount(
+						{ publicKey: transaction.senderPublicKey },
+						waterCb
+					);
+				} else {
+					hasSender = false;
+					library.bus.message(
```
4miners: `library.logger.debug/info.error` and we need better error description

logic/transaction_pool.js
```diff
@@ -787,14 +795,61 @@ __private.isTransactionInUnconfirmedQueue = function(transaction) {
...
+			function checkMulti(sender, waterCb) {
+				const isMultisignature =
+					hasSender &&
+					Array.isArray(sender.multisignatures) &&
+					sender.multisignatures.length;
+
+				setImmediate(waterCb, null, isMultisignature);
+			},
+		],
+		(err, isMulti) => {
```
4miners: please keep naming consistent `isMulti`/`isMultisignature`

logic/transaction_pool.js
```diff
@@ -787,14 +795,61 @@ __private.isTransactionInUnconfirmedQueue = function(transaction) {
...
 };
 
+/**
+ * Check if transaction is from multisig account.
+ *
+ * @private
+ * @param {Object} transaction - Transaction object
+ * @returns {Boolean}
+ */
+__private.isMultisigTransaction = function(transaction, cb) {
```
4miners: `isMultisignatureTransaction`

logic/transaction_pool.js
```diff
@@ -787,14 +795,61 @@ __private.isTransactionInUnconfirmedQueue = function(transaction) {
...
+					hasSender = false;
+					library.bus.message(
+						'error: sender publickey is missing',
+						transaction
+					);
+				}
+			},
+			function checkMulti(sender, waterCb) {
+				const isMultisignature =
+					hasSender &&
```
4miners: `hasSender` is not needed, as we have access to `transaction.senderPublicKey` here.

logic/transaction_pool.js
```diff
@@ -787,14 +795,61 @@ __private.isTransactionInUnconfirmedQueue = function(transaction) {
...
+	async.waterfall(
+		[
+			function getAccount(waterCb) {
+				if (transaction.senderPublicKey) {
+					hasSender = true;
+					modules.accounts.getAccount(
+						{ publicKey: transaction.senderPublicKey },
+						waterCb
+					);
+				} else {
```
4miners: Sweetie, you need to call `waterCb` in `else` block also, or Jenkins will be stuck for hours when hit that condition (like it happening currently)... 😢

logic/transaction_pool.js
```diff
@@ -596,30 +599,29 @@ TransactionPool.prototype.processUnconfirmedTransaction = function(
...
-		}
-		self.addBundledTransaction(transaction);
-	} else if (
-		transaction.type === transactionTypes.MULTI ||
-		Array.isArray(transaction.signatures)
-	) {
-		if (
-			self.countMultisignature() >= config.transactions.maxTransactionsPerQueue
+	__private.isMultisignatureTransaction(transaction, isMulti => {
+		if (transaction.bundled) {
```
nazarhussain: This way `isMultisignatureTransaction` will be called for every transaction, which is doing a DB read. While its necessary only if transaction is not bundled. So move this call after `transaction.bundled` condition to avoid DB calls without requirement.

logic/transaction_pool.js
```diff
@@ -787,14 +789,50 @@ __private.isTransactionInUnconfirmedQueue = function(transaction) {
...
+ */
+__private.isMultisignatureTransaction = function(transaction, cb) {
+	async.waterfall(
+		[
+			function getAccount(waterCb) {
+				if (transaction.senderPublicKey) {
+					modules.accounts.getAccount(
+						{ publicKey: transaction.senderPublicKey },
+						waterCb
+					);
```
nazarhussain: The definition says: 
> sent from a multisig account

But the logic here is just checking for a normal account, not the multisig account. If you really need a multisig account you can use `logic.account. getMultiSignature`

logic/transaction_pool.js
```diff
@@ -297,21 +297,24 @@ TransactionPool.prototype.getMergedTransactionList = function(reverse, limit) {
...
-	if (self.unconfirmed.index[transaction.id] === undefined) {
-		self.unconfirmed.transactions.push(transaction);
-		const index = self.unconfirmed.transactions.indexOf(transaction);
-		self.unconfirmed.index[transaction.id] = index;
-	}
+		if (self.unconfirmed.index[transaction.id] === undefined) {
+			self.unconfirmed.transactions.push(transaction);
+			const index = self.unconfirmed.transactions.indexOf(transaction);
+			self.unconfirmed.index[transaction.id] = index;
+		}
```
nazarhussain: This can be refactored later, we don't need to maintain two arrays for this purpose. Simply `self.unconfirmed.transactions` is good enough to maintain the unconfirmed transactions.

PR #1818, opened 2018-04-05
Block are not broadcasted properly - Closes #1817

### LINKED ISSUE #1817: Block are not broadcasted properly
### Expected behavior
Block should be broadcasted to peers with matching broadhash first, then to peers with unmatching broadhash. Additionally after change of last block (or after sync when syncing) we should update our headers and notify remote peers about that change.
### Actual behavior
Logic is different.
### Steps to reproduce
N/A
1.0.0-alpha.4

4miners:
### What was the problem?
Logic for broadcasting blocks, updating headers and notify peers about update was wrong.
### How did I fix it?
Adjust logic to expected behavior described in the issue.
### How to test it?
Manually, rounds functional system tests, blocks sub-modules unit tests.
### Review checklist

* The PR solves #1817
* Linting passes
* Tests pass

### DIFF
--- a/modules/blocks/chain.js
+++ b/modules/blocks/chain.js
@@ -736,36 +736,47 @@ Chain.prototype.deleteLastBlock = function(cb) {
 		return setImmediate(cb, 'Cannot delete genesis block');
 	}
 
-	async.waterfall(
-		[
-			waterCb => {
-				// Delete last block, replace last block with previous block, undo things
-				__private.popLastBlock(lastBlock, (err, newLastBlock) => {
+	let deletedBlockTransactions;
+
+	async.series(
+		{
+			popLastBlock(seriesCb) {
+				// Perform actual delete of last block
+				__private.popLastBlock(lastBlock, (err, previousBlock) => {
 					if (err) {
 						library.logger.error('Error deleting last block', lastBlock);
+					} else {
+						// Store actual lastBlock transactions in reverse order
+						deletedBlockTransactions = lastBlock.transactions.reverse();
+
+						// Set previous block as our new last block
+						lastBlock = modules.blocks.lastBlock.set(previousBlock);
 					}
-					return setImmediate(waterCb, err, newLastBlock);
+					return setImmediate(seriesCb, err);
 				});
 			},
-			(newLastBlock, waterCb) => {
-				modules.system.update(() => setImmediate(waterCb, null, newLastBlock));
+			updateSystemHeaders(seriesCb) {
+				// Update our own headers: broadhash and height
+				modules.system.update(seriesCb);
 			},
-			(newLastBlock, waterCb) => {
+			broadcastHeaders(seriesCb) {
+				// Notify all remote peers about our new headers
+				modules.transport.broadcastHeaders(seriesCb);
+			},
+			receiveTransactions(seriesCb) {
 				// Put transactions back into transaction pool
 				modules.transactions.receiveTransactions(
-					lastBlock.transactions.reverse(),
+					deletedBlockTransactions,
 					false,
 					err => {
 						if (err) {
 							library.logger.error('Error adding transactions', err);
 						}
-						// Replace last block with previous
-						lastBlock = modules.blocks.lastBlock.set(newLastBlock);
-						return setImmediate(waterCb, null, newLastBlock);
+						return setImmediate(seriesCb, null, lastBlock);
 					}
 				);
 			},
-		],
+		},
 		cb
 	);
 };
--- a/modules/blocks/chain.js
+++ b/modules/blocks/chain.js
@@ -722,6 +722,9 @@ __private.popLastBlock = function(oldLastBlock, cb) {
 
 /**
  * Deletes last block.
+ * - Apply the block to database if both verifications are ok
+ * - Update headers: broadhash and height
+ * - Put transactions from deleted block back into transaction pool
  *
  * @param  {function} cb - Callback function
  * @returns {function} cb - Callback function from params (through setImmediate)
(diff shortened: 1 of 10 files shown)


modules/blocks/chain.js
```diff
@@ -736,37 +739,49 @@ Chain.prototype.deleteLastBlock = function(cb) {
...
 					false,
 					err => {
 						if (err) {
 							library.logger.error('Error adding transactions', err);
 						}
-						// Replace last block with previous
-						lastBlock = modules.blocks.lastBlock.set(newLastBlock);
-						return setImmediate(waterCb, null, newLastBlock);
+						deletedBlockTransactions = null;
+						return setImmediate(seriesCb);
```
ManuGowda: when there is an error in `modules.transactions.receiveTransactions(` the error is not propagated `return setImmediate(seriesCb);`, IMO the error should be propagated `return setImmediate(seriesCb, err);` to have the consistancy of the `Chain.prototype.deleteLastBlock` function.

4miners: that is correct behavior, error cannot be propagated.

modules/blocks/chain.js
```diff
@@ -736,37 +739,49 @@ Chain.prototype.deleteLastBlock = function(cb) {
...
-			waterCb => {
-				// Delete last block, replace last block with previous block, undo things
-				__private.popLastBlock(lastBlock, (err, newLastBlock) => {
+	let deletedBlockTransactions;
+
+	async.series(
+		{
+			popLastBlock(seriesCb) {
+				// Perform actual delete of last block
+				__private.popLastBlock(lastBlock, (err, previousBlock) => {
```
nazarhussain: Follow the convention that we agreed upon. and use `popLastBlockErr` instead of `err`

modules/blocks/chain.js
```diff
@@ -736,37 +739,49 @@ Chain.prototype.deleteLastBlock = function(cb) {
...
+				// Notify all remote peers about our new headers
+				modules.transport.broadcastHeaders(seriesCb);
+			},
+			receiveTransactions(seriesCb) {
 				// Put transactions back into transaction pool
 				modules.transactions.receiveTransactions(
-					lastBlock.transactions.reverse(),
+					deletedBlockTransactions,
 					false,
 					err => {
```
nazarhussain: Same here  `receiveTransactionsErr`

modules/blocks/chain.js
```diff
@@ -736,37 +739,49 @@ Chain.prototype.deleteLastBlock = function(cb) {
...
 					}
-					return setImmediate(waterCb, err, newLastBlock);
+					return setImmediate(seriesCb, err);
 				});
 			},
-			(newLastBlock, waterCb) => {
-				modules.system.update(() => setImmediate(waterCb, null, newLastBlock));
+			updateSystemHeaders(seriesCb) {
+				// Update our own headers: broadhash and height
+				modules.system.update(seriesCb);
```
nazarhussain: I suggest that we should move call to `modules.transport.broadcastHeaders` inside `modules.system.update`, so when ever and where eve we update system headers, we should broadcast it to peers automatically.

modules/blocks/verify.js
```diff
@@ -801,10 +807,19 @@ Verify.prototype.processBlock = function(block, broadcast, saveBlock, cb) {
...
+			// Perform next two steps only when 'broadcast' flag is set, it can be:
+			// 'true' if block comes from generation or receiving process
+			// 'false' if block comes from chain synchronisation process
+			updateSystemHeaders(seriesCb) {
+				// Update our own headers: broadhash and height
+				broadcast ? modules.system.update(seriesCb) : seriesCb();
+			},
+			broadcastHeaders(seriesCb) {
+				// Notify all remote peers about our new headers
+				broadcast ? modules.transport.broadcastHeaders(seriesCb) : seriesCb();
```
nazarhussain: As stated above, if `broadcastHeaders` is moved into `system.update` then we don't need to add it every where.

modules/blocks/verify.js
```diff
@@ -787,6 +790,9 @@ Verify.prototype.processBlock = function(block, broadcast, saveBlock, cb) {
 			verifyBlock(seriesCb) {
 				__private.verifyBlock(block, seriesCb);
 			},
+			broadcastBlock(seriesCb) {
+				__private.broadcastBlock(block, broadcast, seriesCb);
+			},
```
nazarhussain: This broadcast can be done parallel to `processBlock` to increase the performance.

modules/transport.js
```diff
@@ -365,61 +365,82 @@ Transport.prototype.onUnconfirmedTransaction = function(
...
+		library.logger.debug(
+			'Transport->broadcastHeaders: Broadcasting headers to remote peers',
+			{ count: peers.length }
+		);
+
+		// Execute remote procedure updateMyself for every peer
+		async.each(
+			peers,
+			(peer, eachCb) => {
+				peer.rpc.updateMyself(library.logic.peers.me(), err => {
```
nazarhussain: Here as well, rename to `updateMyselfErr`

modules/transport.js
```diff
@@ -365,61 +365,82 @@ Transport.prototype.onUnconfirmedTransaction = function(
...
-		library.network.io.sockets.emit('blocks/change', block);
+	// Exit immediately when 'broadcast' flag is not set
+	if (!broadcast) return;
+
+	// Check if we are free to broadcast
+	if (__private.broadcaster.maxRelays(block)) {
+		return library.logger.debug(
+			'Transport->onBroadcastBlock: Aborted - max block relays exhausted'
+		);
+	} else if (modules.loader.syncing()) {
```
nazarhussain: We can avoid nested `else` here, as we are already returning from above `if`

PR #1825, opened 2018-04-06
Map of peer connections in worker process can go out of sync with map of peers in master process - Closes #1823

### LINKED ISSUE #1823: Map of peer connections in worker process can go out of sync with map of peers in master process
### Expected behavior

Within a node, peer mappings should always be consistent across worker and master processes.

### Actual behavior

We've identified at least one scenario which causes the peer (nonce/address/connectionId) mappings to become out of sync across the two processes and this causes peers to be removed from the peer list in such a way that they cannot be added back (rediscovered) later. We suspect that there are other hard-to-reproduce scenarios which might also cause this issue.

### Steps to reproduce

Connect two nodes together (A and B), kill the worker process on node B then allow some time for peer B to recover.
Using the HTTP GET /api/peers endpoint, get the list of peers from node A; the list should be (and remain) empty (which is incorrect; since they should have reconnected to each other). If you call the same endpoint on the node B (the one which crashed), you will see node A as expected (so the problem is only visible on the peer node).

1.0.0

jondubois:
### What was the problem?

There were two sources of truth when dealing with peer removal; sometimes peer removal was initiated by the master process, other times it was triggered by a command from the worker. This caused the peer nonce/address/connectionId mappings on the worker and master processes to be in inconsistent states which prevented peers from being added back to the peer list at a later time.

### How did I fix it?

Don't let the master process decide when to remove a peer; always wait for the instruction from the worker process. If the master process notices a failed connection on an outbound socket, it will just put the peer in a disconnected state and wait for instruction from the worker process.

### How to test it?

1. You need to prepare two nodes: A and B.
2. The config.json file of node B should contain node A as a seed node in `peers.list` *but not the other way round*; let node A be discovered by node B.
3. Launch both nodes and wait for them to discover each other.
4. Use the HTTP GET `/api/peers` endpoint to get the list of peer nodes of node A - Node B should be visible.
5. Kill the worker process of node B and wait a bit.
6. According to the `/api/peers` endpoint on node A, node B should still show up in the peers list (after the node B worker has respawned).
7. Shut down the entire node B (master process).
8. According to the `/api/peers` endpoint on node A, the peers list should now be empty.

Note that it's important that node B is not listed as a seed peer in node A's config file or else node B will never be removed from the list (there will be an error thrown because you cannot remove a 'frozen peer').

### Review checklist

* The PR solves #1823
* Linting passes
* Tests pass

### DIFF
--- a/app.js
+++ b/app.js
@@ -368,6 +368,14 @@ d.run(() => {
 						scope.logger.info('Socket Cluster ready for incoming connections');
 						cb();
 					});
+
+					scope.socketCluster.on('workerExit', workerInfo => {
+						var exitMessage = `Worker with pid ${workerInfo.pid} exited`;
+						if (workerInfo.signal) {
+							exitMessage += ` due to signal: '${workerInfo.signal}'`;
+						}
+						scope.logger.error(exitMessage);
+					});
 				},
 			],
 
--- a/helpers/peers_manager.js
+++ b/helpers/peers_manager.js
@@ -16,6 +16,7 @@
 
 const connect = require('../api/ws/rpc/connect');
 const disconnect = require('../api/ws/rpc/disconnect');
+const Peer = require('../logic/peer');
 
 /**
  * Description of the class.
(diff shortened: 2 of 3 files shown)


jondubois:
This PR is an improvement on https://github.com/LiskHQ/lisk/compare/1.0.0...disconnect_peers_not_remove - The problem with that PR is that we don't want to change the functionality of `PeersManager.prototype.remove` because it is still needed (it just had to be called in a different place).

jondubois:
Also, some test cases related to the sort order of peers were failing (sometimes) because the comparator function that we were using wasn't deterministic when dealing with undefined and null values.

Issue #1819, opened 2018-04-05
Update passphrase encryption to use authenticated encryption


willclarktech:
### Expected behavior

We should be using authenticated encryption for passphrases, as in lisk-elements.

https://github.com/LiskHQ/lisk-js/blob/development/src/cryptography/encrypt.js#L107-L124

### Actual behavior

We use AES-256-CBC, which is unauthenticated.

https://github.com/LiskHQ/lisk/blob/development/modules/delegates.js#L335

### Which version(s) does this affect? (Environment, OS, etc...)

1.0.0

webmaster128:
What is encrypted here? Don't we have the plaintext secret in the API call and the plaintext secret in the config file anyway?

willclarktech:
@webmaster128 Not with v1.0.0

willclarktech:
## Summary of meeting

### Statement of the problem

In 0.9.x delegates can manage forging in one of two ways:

1. By storing plaintext passphrases in `config.json` and setting `forging.force` to `true`.
1. By sending plaintext passphrases to their node to turn forging on/off for that account.

The first option here is intended only for testing purposes, and delegates should not be using it in production. The second option raises a couple of problems:

1. Some delegates may not be using secure connections when sending API requests.
1. Some delegates may be making requests to the API using a relatively insecure server (e.g. a cheap VPS used for management).

The first problem here should be solvable with good documentation/education. The second prompts reconsideration of our current recommendation that delegates do not store plaintext passphrases in their `config.json` in production. Supposing an attacker gains root access to the delegate's forging node, there's no way to prevent theft of the passphrase. Supposing the attacker instead gains only read access to the file system, for example via a stolen backup, a level of security can be achieved by encrypting the passphrase.

In the case of a management server intended to manage automatic failover between several nodes (switching forging to a new node in case of an outage), it cannot be expected that the delegate will be able to manually input a passphrase whenever necessary.

### Proposed solution

1. Encrypt passphrases in the `config.json` on the node.
1. Store the key required for decryption wherever it's needed (e.g. on a management server).
1. Send the decryption key in the clear (but ideally using a secure connection) to the API endpoint to toggle the forging status of a node.

This way, if the decryption key is compromised, the forging status can be toggled for a while, but the delegate can just reencrypt the passphrase under a different key to regain control. If read access to the node server is gained, then the passphrase is not compromised. Both the decryption key *and* the encrypted passphrase would have to be retrieved before the passphrase would be compromised.

[...]

PR #1829, opened 2018-04-09
Use authenticated passphrase encryption

willclarktech:
### What was the problem?

Core was still using unauthenticated encryption for storing passphrases.

### How did I fix it?

I switched to AES-256-GCM as used in lisk-elements already.

### How to test it?

Encrypt a passphrase under a password using Lisk Commander or Lisk Elements, store it in the `config.json` and try to use your node.

### Review checklist

* The PR solves #1819 
* Linting passes

Note: one test is failing and it's not obvious to me why. I'm hoping somebody who's more familiar with the test suite can help me fix this.

### Additional comments

I used this as a basis for the string to be stored: http://man7.org/linux/man-pages/man3/crypt.3.html#NOTES

Parameters not covered by that scheme are included after the ciphertext, prepended by the key. (These additional parameters can appear in any order.)


willclarktech:
@MaciejBaj @4miners Unit and functional tests passed locally.

4miners:
@willclarktech Please provide reason for that change, issue this PR is referring to doesn't explain anything. What are advantages and disadvantages of algorithm change?

Original idea behind introducing passphrase encryption was to mitigate two kind of security risks:
- intercept during sending passphrase over the network
- management server (one that is used to enable/disable forging on few nodes) get compromised

Passphrase encryption doesn't mitigate case when server used for forging get compromised - because passphrase can be captured from memory, or decryption key can be captured when coming from management server.

So what are the advantages of using proposed encryption other than mitigating brute-force (which still should be pretty hard with decryption key of proper length)?

I see some disadvantages:
- Decryption of all 101 passphrases will take ~97 seconds, which will make our testing very inconvenient (we need to load all delegates every test, so now we need to wait a lot).
- It's using a lot of CPU time for decryption.
- Add complexity for average delegate - we need to store in `config.json` more fields - `iv`, `salt`, `tag`, `version`, which in case of compromise forging server gives us nothing.

For me issue #1819 is invalid. Please clarify.

webmaster128:
@4miners This PR brings in two major changes that could (or should) be separated:

1. AES-CBC is replaced with AES-GCM and
2. the very cheap password -> AES-256 key operation from `crypto.createDecipher(algorithm, password[, options])` is replaced with a not-so-cheap pbkdf2 operation

One argument for doing change 1 is: you are guaranteed to get an error when the wrong decryption key is used. This is not a feature of plain AES-CBC. You can use Botan (patched version by me locally) for example to see that AES-CBC gives different plaintexts for different decryption keys. But no error:

```
# same key, okay
$ echo -n "1234567890123456" | ./botan encryption --mode=aes-256-cbc --iv=000102030405060708090A0B0C0D0E0F --key=603DEB1015CA71BE2B73AEF0857D77811F352C073B6108D72D9810A30914DFF4 | ./botan encryption --mode=aes-256-cbc --decrypt --iv=000102030405060708090A0B0C0D0E0F --key=603DEB1015CA71BE2B73AEF0857D77811F352C073B6108D72D9810A30914DFF4
1234567890123456

# different decryption key, different output
$ echo -n "1234567890123456" | ./botan encryption --mode=aes-256-cbc --iv=000102030405060708090A0B0C0D0E0F --key=603DEB1015CA71BE2B73AEF0857D77811F352C073B6108D72D9810A30914DFF4 | ./botan encryption --mode=aes-256-cbc --decrypt --iv=000102030405060708090A0B0C0D0E0F --key=603DEB1015CA71BE2B73AEF0857D77811F352C073B6108D72D9810A30914DFF5
]B&���K�K^�yE�
```

I think that out of plain luck, you are using a AES-CBC implementation from OpenSSL which adds a padding around AES-CBC which can detect wrong decryption keys as well. Switching to AES-GCM makes this feature transparent. It also improves interoperability with other cryptographic tool sets.

Change 2 is the extra step that takes ~800ms per run on my laptop (similar to the timings you mentioned).

If you know that your uses use high entropy passwords, keeping the current code is good. But the examples given are not high entropy. 6 English words from a let's say 20k wordlist is an okayish 80bit entropy. With 4 words or less you are basically doomed by any brute-force attach. This is because the operation `password -> EVP_BytesToKey -> 256bit AES key -> AES decryption` is very fast and can be done offline and distributed.

So if passwords should be used, change 2 is basically required (but better implemented using Argon2 than PBKDF2 in 2018). However, if you enforce the use of raw 256bit AES keys that come straight from a RNG, change 2 can be avoided.

---------------------

ps.: The current implementation of `decryptSecret` takes a password, not an AES key. Thus the naming should be checked in `test/data/genesis_delegates.json` as well as the API field "decryptionKey".

4miners:
@webmaster128 Thank you for detailed explanation. I didn't even check the proposed implementation as descriptions provided both in issue and PR are not sufficient in order to proceed with a review.

Encrypted secret can be brute-forced only when it's compromised ofc, those are stored on server used for forging. It's actually easier for an potential attacker to modify source code of running node or examine memory and intercept plain (decrypted) secret when he already have access to the machine.

However using brute-force "resistant" version makes sense when somehow only `config.json` get compromised (read-only access to file system? users commiting by mistake `config.json` to GitHub public repos?).

For me AES-256-GCM makes sense, but PBKDF2 is kind of too costly.

4miners:
@LiskHQ/cryptography What's your opinion on that?

webmaster128:
> However using brute-force "resistant" version makes sense when somehow only config.json get compromised (read-only access to file system? users commiting by mistake config.json to GitHub public repos?).

Why not make this a feature? People can check-in their server configuration including the encrypted passphrase in a repo that let's them spawn new servers very quickly (using Ansible or something similar).

-----------------------

But to be honest, I don't see that the entire encryption really helps here. You are splitting the plain passphrase into `decryptionKey` and `encryptedPassphrase`. `decryptionKey` can be compromised the same was as in 0.9.x and `encryptedPassphrase` is now stored on disk and potentially in configuration repos. This extra complexity could lead to people messing up much easier than when they only need to care for the passphrase.

If you already assume, that a compromised forging server leaks the passphrase (which I agree with), you can store the passphrase in a separate file that only the file owner can read and use an API that enables/disables forging without sending a secret. That way you don't need to have confidential information on the management server. So both of the points in https://github.com/LiskHQ/lisk/pull/1829#issuecomment-381153269 are covered

IkerAlus:
@webmaster128 Quick answer regarding 

> better implemented using Argon2 than PBKDF2 in 2018

 before addressing the whole issue:

Earlier this year, we looked into this issue and we were comparing  PBKDF2 vs Argon2 algorithms. We concluded that Argon2 is still a very young algorithm (for cryptography terms), there are not properly reviewed and widely used implementations of it and PBKDF2, with an adequate number of iterations, is secure enough for any current application.

willclarktech:
IIRC, there were properly reviewed implementations of Argon2, but we felt a bit more time in the wild would be needed to have full confidence. Also, PBKDF2 comes natively in Node.js, so we didn't need to add any additional dependencies (with all the associated costs).

modules/delegates.js
```diff
@@ -321,21 +321,99 @@ __private.forge = function(cb) {
...
+	const tagRegExp = /^tag=(.+)$/;
+
+	const getAdditionalParameter = (parameters, regExp) => {
+		const parameter = parameters.find(parameter => regExp.test(parameter));
+		const match = parameter ? parameter.match(regExp) : null;
+		return match && match[1];
+	};
+
+	// eslint-disable-next-line no-unused-vars
+	const [_, id, ...encryptedSecretWithoutID] = encryptedSecret.split(delimiter);
```
4miners: @SargeKhan Please give your input on that line.

4miners:
Please avoid to use rebase / alter the commits. 😢

modules/delegates.js
```diff
@@ -321,21 +321,100 @@ __private.forge = function(cb) {
...
  * @returns {string} decryptedSecret
  * @todo Add description for the params
  */
-__private.decryptSecret = function(encryptedSecret, key) {
-	const decipher = crypto.createDecipher('aes-256-cbc', key);
-	let decryptedSecret = decipher.update(encryptedSecret, 'hex', 'utf8');
-	decryptedSecret += decipher.final('utf8');
-	return decryptedSecret;
+__private.decryptSecret = function({ encryptedSecret }, password) {
+	// const { tag, salt, iv, iterations, cipherText } = parseEncryptedSecret(encryptedSecret);
```
SargeKhan: we need this comment now?

test/unit/modules/delegates.js
```diff
@@ -113,12 +113,133 @@ describe('delegates', () => {
...
+					// method is set to 1: MD5
+					encryptedSecret:
+						'$1$rounds=1$5d502e8156eeeaf2acf23d801351dda7$f856ab7731c046da5bf7b9f11fdf20fbc1870845633627b82f84151e432d4e4ca5f69a9c40f15bc2a749e28a66c2e11bbdc45049e0a1f53087fbdd542536c19059e294d231f7aaf30a0600a79ef78519$iv=4a500624cba9c3cb3b6e380556e197d6$tag=24a118a1abab32d256a5de013c20539e$version=1.0.0',
+				};
+
+				config.forging.secret = [accountDetails];
+
+				loadDelegates(err => {
+					expect(err).to.equal(
+						`Invalid encryptedSecret for publicKey: ${accountDetails.publicKey}`
```
SargeKhan: Shouldn't we receive [this error here?](https://github.com/LiskHQ/lisk/pull/1829/files#diff-9d5cd48f1ba4ec6b2770fd44edc90323R341)

test/unit/modules/delegates.js
```diff
@@ -113,12 +113,133 @@ describe('delegates', () => {
...
+					// rounds is removed
+					encryptedSecret:
+						'$5$5d502e8156eeeaf2acf23d801351dda7$f856ab7731c046da5bf7b9f11fdf20fbc1870845633627b82f84151e432d4e4ca5f69a9c40f15bc2a749e28a66c2e11bbdc45049e0a1f53087fbdd542536c19059e294d231f7aaf30a0600a79ef78519$iv=4a500624cba9c3cb3b6e380556e197d6$tag=24a118a1abab32d256a5de013c20539e$version=1.0.0',
+				};
+
+				config.forging.secret = [accountDetails];
+
+				loadDelegates(err => {
+					expect(err).to.equal(
+						`Invalid encryptedSecret for publicKey: ${accountDetails.publicKey}`
```
SargeKhan: We should return error that round is omitted

test/unit/modules/delegates.js
```diff
@@ -113,12 +113,133 @@ describe('delegates', () => {
...
+					// tag is 1 character different
+					encryptedSecret:
+						'$5$rounds=1$5d502e8156eeeaf2acf23d801351dda7$f856ab7731c046da5bf7b9f11fdf20fbc1870845633627b82f84151e432d4e4ca5f69a9c40f15bc2a749e28a66c2e11bbdc45049e0a1f53087fbdd542536c19059e294d231f7aaf30a0600a79ef78519$iv=4a500624cba9c3cb3b6e380556e197d6$tag=24a118a1abab32d256a5de013c20539f$version=1.0.0',
+				};
+
+				config.forging.secret = [accountDetails];
+
+				loadDelegates(err => {
+					expect(err).to.equal(
+						`Invalid encryptedSecret for publicKey: ${accountDetails.publicKey}`
```
SargeKhan: Should we return actual error here?

helpers/z_schema.js
```diff
@@ -152,6 +152,28 @@ var liskFormats = {
...
+		// - $HEX_CIPHERTEXT => encrypted secret
+		// - Optional additional parameters (order and number is unimportant):
+		//   - $iv=HEX_IV => IV used for encryption
+		//   - $tag=HEX_TAG => tag used to authenticate when decrypting
+		//   - $version=SEMVER_VERSION => version check for future proofing
+		// NOTE: Maximum lengths chosen here are arbitrary
+		return /^\$\d(\$rounds=\d{1,32})?\$[a-f0-9]{32}\$[a-f0-9]{2,1024}(\$((iv|tag)=[a-f0-9]{32}|version=\d{1,4}\.\d{1,4}\.\d{1,4})){0,3}$/i.test(
+			str
+		);
+	},
```
nazarhussain: @willclarktech If you can add some sample for the input of this method, will be helpful to understand this regular expression.

modules/delegates.js
```diff
@@ -492,16 +571,18 @@ __private.loadDelegates = function(cb) {
 			let secret;
 			try {
 				secret = __private.decryptSecret(
-					encryptedItem.encryptedSecret,
+					encryptedItem,
```
nazarhussain: That will be more readable code if you pass directly `encryptedItem.encryptedSecret` and accept a direct param instead  of hash in method.

modules/delegates.js
```diff
@@ -492,16 +571,18 @@ __private.loadDelegates = function(cb) {
...
 					library.config.forging.defaultKey
 				);
-			} catch (e) {
+			} catch (error) {
+				const errorMessage = error.message.endsWith('.')
+					? error.message
+					: `${error.message}.`;
```
nazarhussain: Why its necessary that error message must end with a `.`

modules/delegates.js
```diff
@@ -583,10 +669,7 @@ Delegates.prototype.toggleForgingStatus = function(publicKey, secretKey, cb) {
 
 	if (encryptedItem) {
 		try {
-			decryptedSecret = __private.decryptSecret(
-				encryptedItem.encryptedSecret,
-				secretKey
-			);
+			decryptedSecret = __private.decryptSecret(encryptedItem, secretKey);
```
nazarhussain: Same here, it looks more readable code if you pass explicitly `encryptedItem.encryptedSecret`

modules/delegates.js
willclarktech: > accept a direct param instead of hash in the method

I don't understand this suggestion.

test/data/config.json
```diff
@@ -77,607 +77,607 @@
 		"secret": [
 			{
 				"encryptedSecret":
-					"19e9f8eee5d9516234a10953669518bf23371d34e114713c8043a98378fd866834946380c4cc0e40f23305643206c1c5be496074f350d09d87735c42eae30c604cabea9862f4fe8f906313c67a7146d54fa844171e6cf925b7c953987082cdd6",
+					"$5$rounds=1$e33b33c7621fc33d361ff26d1b9611b4$aedfce0b845f7c962ea57a76f7bb7e48228273356fdab39ce71c9c385e105f7660101a7ffbee14ef5aa9dc8d03d6715e0828acae8d8625bbec9a167839d0f935445d14a874e18cbed8b01f7da74d990e$iv=51b98dbc815996de7821e56fcc985fbb$tag=75a9571b9018e59255f5ed8ac8f7db75$version=1.0.0",
```
nazarhussain: Why we used `$` as the key separator? Means we can use any `|` or `-`, is there any specific reason to use `$` symbol? Agree or not `$` is in general used for a special meaning for variables in web development in javascript. So we should rather use some other symbol as the key separator for our hashed/encrypted strings.

willclarktech: Yes, it conforms to the existing standard taken from Linux shadow files.

modules/delegates.js
nazarhussain: You are passing an object `encryptedItem ` and function definition is extracting the required attribute `encryptedSecret` from that object. Instead, I think if we pass a direct attribute `encryptedItem.encryptedSecret` to function call that will be more readable.

test/data/config.json
nazarhussain: @willclarktech I don't feel its a standard. Correct me if I am wrong. The convention was introduced in 1987, and ported to Linux in 1992. Later ported to different linux distros with the variety of changes. The modern linux distro like `Ubuntu` don't seems to follow that convention of `$` symbol. 

Here is some excerpt from Ubuntu shadow file. 

```
root:*:16176:0:99999:7:::
daemon:*:17596:0:99999:7:::
bin:*:17596:0:99999:7:::
sys:*:17596:0:99999:7:::
sync:*:17596:0:99999:7:::
```

It seems to use `:` as key separator. Correct me if I am missing something.

modules/delegates.js
willclarktech: Ah by `hash` you mean an object?

test/data/config.json
willclarktech: @nazarhussain Do you have an alternative proposal?

willclarktech: @webmaster128 Thanks for the info. `$pbkdf2-sha256` is in any case more explicit than `$5`.

willclarktech: For the record, the idea as discussed with @fchavant @Nazgolze @4miners and @SargeKhan, was not to adopt this as a rigid standard, but to take it as a starting point. For one thing we have hex strings here, whereas the original standard seems to assume base64.

webmaster128: Having a custom format is great because in this case there is no format widely adopted by the industry. If you don't mind, let me think out loud. I suggest:

* Having a version is a good thing
* Decouple the format version from the Lisk core version. The format may stay stable between Core 1.0.0 and 3.8 and suddenly change for both branches 3.9 and 4.0. What's the version then?
* Use an integer version. When using 1.0.0 or 2.3.4 you need to parse and handle all the possible versions in between. But there is no value in having multiple version positions. Either bump from version 1 to version 2, or don't.
* Hardcode the parameters `aes-256-gcm`, `pdkdf2-sha256` for version `1`. If you need to change them, bump the version
* Use binary concatenation for the encrypted data: `iv | ciphertext | tag`. `iv` and `tag` have a fixed length, so they can be cut out easily. This has the advantage that you only need 1 field for the entire ciphertext of aes-256-gcm and you do not have algorithm-specific parameters
* [Use a 96bit IV for aes-256-gcm](https://crypto.stackexchange.com/questions/41601/aes-gcm-recommended-iv-size-why-12-bytes)

Then only the fields `version`, `rounds`, `salt`, `ciphertext` are left, and can be encoded using whatever separator (maybe not `$`) or just json: `'{"version": 1, "rounds": 1, "salt": "aabbccdd", "ciphertext": "5c0dc4e86063b4a3814c7f08d9759a98b99ff7d2541eee5cd49ea290dd22dabfee18d23775c5a9f47f7ecbb50b15953c2abc"}'`

willclarktech:
Suspended pending a meeting to clarify the requirements.

test/data/config.json
nazarhussain: @willclarktech I checked your message late. I second @webmaster128 for having `custom format`.

modules/delegates.js
nazarhussain: Yes I mean object. Sorry being Ruby addictive thoughts, I tend to use `hash` for key-value paired objects.

modules/delegates.js
```diff
@@ -321,21 +321,97 @@ __private.forge = function(cb) {
 	);
 };
 
+const parseEncryptedSecret = encryptedSecret => {
```
4miners: Would be nice to add descriptions for those new functions.

helpers/z_schema.js
```diff
@@ -152,6 +152,30 @@ var liskFormats = {
...
+	 * @param {string} str
+	 * @returns {boolean}
+	 * @todo Add description for the function, the params and the return value
+	 */
+	encryptedSecret(str) {
+		// Explanation of regex structure:
+		// - 1 or more 'key=value' pairs delimited with '&'
+		// Examples:
+		// - encryptedSecret=abcd1234
+		// - encryptedSecret=abcd1234&rounds=10000&iv=ef012345
```
webmaster128: `rounds` -> `iterations`

willclarktech: Nice catch, thanks.

webmaster128:
I created a command line tool for decrypting and encrypting passphrases as suggested by this PR: https://github.com/prolina-foundation/lisk-passphrase-encrypt

Usage decrypt:

```
$ echo "iterations=1&salt=41cecfe964e5759c7c891ff8893e7f5b&cipherText=3a3f9f790fcfcb0be8a8f2f0d003feefcb7889d6a142725be1242bd3ccd0ac04e3b1d81f3459fb0af7b0fe714b3a6e21055c668835d6abfe56ea9b6c2ef05a57f8343fdae677015cba5266193fd3&iv=b1a00485d4f114888b3132e5&tag=cde6171f8a8a1a7b26e0ca1f32d734ff&version=1" | ./lisk_passphrase_encrypt --decrypt
Password:
<type password elephant tree paris dragon chair galaxy and press enter>
lounge label clap camera zebra junk analyst citizen arrest caution odor figure
```

Usage encrypt

```
$ echo "lounge label clap camera zebra junk analyst citizen arrest caution odor figure" | ./lisk_passphrase_encrypt --generate-password
Encrypting Lisk passphrase using password 'WK2m;J&L7dP7@DByvp82wr*UCJ*}J|Fsgg*669<='
iterations=1000000&salt=1eaf0e3d4e8a550bed46d21bf9ae3946&cipherText=a1394723dd7d563b48bbc3546f24e4a7ba84cd482d2875168a44eb99f62d5b3519ee48cab5a0ee16cfa867327775307afe417a422e6959e6b64a79177dc6e4e297ca61a413ece46bf06841d5a3ff&iv=52d0549f57fcf93b345fd99d&tag=455d1e16b669d41f5b14823225d8140e&version=1
```

Full usage

```
$ ./lisk_passphrase_encrypt --help
usage: lisk_passphrase_encrypt [-h] [--decrypt] [--generate-password]
                               [--iterations ITERATIONS] [--salt SALT]

Encrypt a Lisk passphrase for your node.

optional arguments:
  -h, --help            show this help message and exit
  --decrypt             decrypt (default: false, i.e. encrypt)

Encryption arguments:
  --generate-password   Auto generate an encryption password (default: ask the
                        user)
  --iterations ITERATIONS
                        pbkdf2 iterations (default: 1,000,000)
  --salt SALT           hex encoded salt for pbkdf2 (default: auto generate)
```

The tool is explicitly using distinct implementations (Python + Botan) to allow cross implementation testing.
It turns out that the example secrets can be decoded using my independently developed script – nice! If there is any questions or interest regarding the tool, feel free to open an issue at https://github.com/prolina-foundation/lisk-passphrase-encrypt to avoid pollution of this thread.

Only one thing is bothering me: you call the encryption password a "key". But it is not a key by any industry standard understanding of what a key is. A key is binary and of fixed length. What we use for encryption is a password.

willclarktech:
> Only one thing is bothering me: you call the encryption password a "key". But it is not a key by any industry standard understanding of what a key is. A key is binary and of fixed length. What we use for encryption is a password.

@webmaster128 Can you provide an example?

webmaster128:
@willclarktech sure

Usage of password:

* In https://nodejs.org/api/crypto.html#crypto_crypto_createcipher_algorithm_password_options the second argument `password` is internally converted to a key using "EVP_BytesToKey with the digest algorithm set to MD5, one iteration, and no salt"
* OpenSSL cli can do `echo 123 | openssl enc -aes-128-cbc -pass pass:somepassword` where the password is converted to a key internally
* libsodium has a [password to key conversion](https://download.libsodium.org/doc/password_hashing/the_argon2i_function.html), where the password is a string of arbitrary length and the result is a key of constant size.
* [pbkdf2 consumes a password as an input parameter](https://en.wikipedia.org/wiki/PBKDF2#Key_derivation_process)

Usage of key:

* In https://nodejs.org/api/crypto.html#crypto_crypto_createcipheriv_algorithm_key_iv_options the second argument `key` must be 128/192/256 bit for aes-128/192/256-cbc. Try `crypto.createCipheriv('aes-128-cbc', new Buffer("fd0f3c5267f321001b30bd75839bdf98", "hex"), new Buffer("fd0f3c5267f321001b30bd75839bdf98", "hex"))` vs `crypto.createCipheriv('aes-128-cbc', new Buffer("fd0f3c5267f321001b30bd75839bdf98aa", "hex"), new Buffer("fd0f3c5267f321001b30bd75839bdf98", "hex"))`
* OpenSSL cli can do `echo 123 | openssl enc -aes-128-cbc -K fd0f3c5267f321001b30bd75839bdf98 -iv fd0f3c5267f321001b30bd75839bdf98` but not `echo 123 | openssl enc -aes-128-cbc -K fd0f3c5267f321001b30bd75839bdf98aa -iv fd0f3c5267f321001b30bd75839bdf98`
* [libsodium requires a key parameter](https://download.libsodium.org/doc/secret-key_cryptography/aes-256-gcm.html) of size `crypto_aead_aes256gcm_KEYBYTES = 32`
* [pbkdf2 returns a key](https://en.wikipedia.org/wiki/PBKDF2#Key_derivation_process)

willclarktech:
@webmaster128 I meant code that forms part of this PR which you think should be changed prior to merging.

webmaster128:
Ah sorry, I just realized that there is some key -> password renaming already ongoing.

What is missing is at least

* the field name `key` in test/data/genesis_delegates.json
* the field name `key` in test/unit/logic/delegate.js
* the field name `decryptionKey` in https://app.swaggerhub.com/apis/LiskHQ/Lisk/1.0.26#/Node/updateForgingStatus

willclarktech:
@webmaster128 Ah, then I agree those things are poorly named, but I think it's beyond the scope of this PR, since it then involves a change to the API and/or migration issues concerning the `config.json`. Feel free to open up a separate issue!

Issue #1831, opened 2018-04-09
Strange logs related to P2P on node - (Needs investigation)


SargeKhan:
### Expected behavior
If node disconnects with the a peer, it should only log once that it has disconnected/exited. But the logs are displayed multiple times, which could mean that there is a possible memory leak or there are multiple socket connections to the same peer.
### Actual behavior
Logs are displayed repeatedly, and then node gets the error:
```
2018-04-06 23:43:17 158: [dbg] 2018-04-06 23:43:17 | [Outbound socket :: close] Peer connection to 37.139.29.38 failed with code 4001 and reason -
2018-04-06 23:43:17 158: [dbg] 2018-04-06 23:43:17 | [Outbound socket :: close] Peer connection to 167.99.251.165 failed with code 4001 and reason -
2018-04-06 23:43:17 159: [dbg] 2018-04-06 23:43:17 | [Outbound socket :: close] Peer connection to 167.99.165.231 failed with code 4001 and reason -
2018-04-06 23:43:17 160: [dbg] 2018-04-06 23:43:17 | [Outbound socket :: close] Peer connection to 206.189.48.87 failed with code 4001 and reason -
2018-04-06 23:43:17 160: [dbg] 2018-04-06 23:43:17 | [Outbound socket :: close] Peer connection to 94.237.41.95 failed with code 4001 and reason -
2018-04-06 23:43:17 161: [dbg] 2018-04-06 23:43:17 | [Outbound socket :: close] Peer connection to 94.237.41.95 failed with code 4001 and reason -
2018-04-06 23:43:17 162: [dbg] 2018-04-06 23:43:17 | [Outbound socket :: close] Peer connection to 94.237.41.95 failed with code 4001 and reason -
2018-04-06 23:43:17 162: [dbg] 2018-04-06 23:43:17 | [Outbound socket :: close] Peer connection to 94.237.41.95 failed with code 4001 and reason -
2018-04-06 23:43:17 163: [dbg] 2018-04-06 23:43:17 | [Outbound socket :: close] Peer connection to 94.237.41.95 failed with code 4001 and reason -
2018-04-06 23:43:17 164: [dbg] 2018-04-06 23:43:17 | [Outbound socket :: close] Peer connection to 167.99.165.232 failed with code 4001 and reason -
2018-04-06 23:43:17 164: [dbg] 2018-04-06 23:43:17 | [Outbound socket :: close] Peer connection to 167.99.165.231 failed with code 4001 and reason -
[...]

PR #1837, opened 2018-04-10
Sockets are not being cleaned up properly and are disconnecting too often - Closes #1836

### LINKED ISSUE #1836: Sockets are not being cleaned up properly and are disconnecting too often
### Expected behavior

- Each peer should only have one outbound socket at any given time.
- Sockets should not become disconnected so often.

### Actual behavior

- Old sockets are causing a memory leak because they are not being discarded correctly before a new socket is attached to the peer.
- Sockets are disconnected too often and send back a code of 1000 which is not useful.

### Steps to reproduce

Connect a node to the beta test net.
Run it with the `--inspect` flag like this to watch the SCSocket objects memory leak:

```sh
node --inspect app.js
```

jondubois:
### What was the problem?

- Memory leak related to outbound peer sockets.
- Sockets disconnecting too easily.

### How did we fix it?

- Destroy the old sockets when they become disconnected.
- Destroy the old socket when an old peer socket is replaced with a new one.
- Removed handler which force-disconnected a socket on error (regardless of the type of error).

### How to test it?

To test the memory leak fix, run the node with `node-inspector` or just:

```sh
node --inspect app.js
```

Take memory heap snapshots and check that the number of `SCSocket` objects doesn't grow over time.

To test the reduced rate of disconnection, we may not know for sure until we roll out to the beta network.

### Review checklist

* The PR solves #1836, #1831
* Linting passes
* Tests pass

### DIFF
--- a/api/ws/rpc/connect.js
+++ b/api/ws/rpc/connect.js
@@ -147,13 +147,11 @@ const connectSteps = {
 		});
 
 		socket.on('disconnect', () => {
-			logger.error(
+			logger.trace(
 				`[Outbound socket :: disconnect] Peer connection to ${
 					peer.ip
 				} disconnected`
 			);
-
-			logger.error(socket.clientId);
 		});
 
 		// When handshake process will fail - disconnect
--- a/api/ws/rpc/connect.js
+++ b/api/ws/rpc/connect.js
@@ -180,6 +180,7 @@ const connectSteps = {
 					peer.ip
 				} failed with code ${code} and reason - ${reason}`
 			);
+			socket.destroy();
 			onDisconnectCb();
 		});
 
--- a/api/ws/rpc/connect.js
+++ b/api/ws/rpc/connect.js
@@ -56,6 +56,13 @@ const connectSteps = {
 	},
 
 	addSocket: peer => {
+		if (peer.socket) {
+			// if a socket connect exisits
+			// before issuing a new connection
+			// destroy the exisiting connection
+			// to avoid too many socket connections
+			peer.socket.destroy();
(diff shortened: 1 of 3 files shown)


api/ws/rpc/connect.js
```diff
@@ -56,6 +56,15 @@ const connectSteps = {
 	},
 
 	addSocket: peer => {
+		if (peer.socket) {
```
shuse2: if socket exist, does it even need to go through all the process?

workers_controller.js
```diff
@@ -169,9 +169,6 @@ SCWorker.create({
 				scServer.on('connection', socket => {
 					scope.slaveWAMPServer.upgradeToWAMP(socket);
 					socket.on('disconnect', removePeerConnection.bind(null, socket));
-					socket.on('error', err => {
```
shuse2: is socket not connected when it receives error event?

api/ws/rpc/connect.js
ManuGowda: The socket can exist in a closed state and not destroyed, so every time the socket connection closes a new socket connection is issued from peers_manager with a different header(different height, broadhash) this will lead to multiple socket connection for same peer and these connections are not destroyed and leads to memory leaks.

workers_controller.js
jondubois: It is connected but we don't necessarily need to disconnect it for every error. Some errors like message timeouts will be emitted as an error on the socket but we don't need to disconnect the socket in these cases. The socket will already disconnect itself if there is a fatal error on it.

PR #1867, opened 2018-04-12
Cover stress tests for tx type 1, 4, 5 and 0 with data - Closes #1761

### LINKED ISSUE #1761: Cover stress tests for transaction types 1, 4, 5 and 0 with data
### Expected behavior
Create stress test for:
- type 0 + data (non empty asset.data field),
- type 1 (second signature),
- type 4 (multisignature + signatures to confirm), 
- type 5 (dapps registration). 
Tests should be included in integration tests layer.
Keep in mind ease of reuse during transaction creation in QA process. 
### Actual behavior
Stress test includes only type 0 and 2, 3 when #1758 is done.
### Steps to reproduce
Integration tests
1.0.0

Tschakki:
### What was the problem?
Stress tests for transaction types 1,4,5 and 0 with data were missing.

### How did I fix it?
Added the tests.

### How to test it?
`npm test -- mocha:slow:integration`

### Review checklist

* The PR solves #1761
* Linting passes
* Tests pass


test/integration/scenarios/stress/4.register_multisignature.js
```diff
@@ -0,0 +1,123 @@
...
+var getUnsignedTransaction = require('../../utils/http').getUnsignedTransaction;
+var getTransaction = require('../../utils/http').getTransaction;
+
+module.exports = function(params) {
+	describe('stress test for type 4 transactions @slow', () => {
+		var transactions = [];
+		var accounts = [];
+		var maximum = 1000;
+
+		function confirmUnsignedTransactionsOnAllNodes() {
```
diego-G: Why we don't sign any of the multisig upgrade transactions? That way we will stress the network is not consistent with how we are doing in the rest of cases. The file's name itself points to a complete registration which is not carried out.
Right now, we only perform the first phase of a multisig upgrade account. Then we only stress the queue `unsigned` and the fillPool batcher but we don't process any transaction.

Tschakki: Signing the multitransactions was not part of the stress test that we already defined for type 4 transaction in lisk-core-qa repo, that is why it isn't included here, too.

I have nothing against signing the transaction as part of the stress test, though. Just wanted to keep it consistent.

test/common/utils/random.js
```diff
@@ -75,6 +75,16 @@ random.applicationName = function() {
 	});
 };
 
+// Returns a random string of a given size
+random.dataField = function(bytes) {
+	var custom = 'ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789';
+
+	return randomstring.generate({
+		length: bytes,
+		charset: custom,
```
diego-G: Please, use `alphanumeric` charset. It matches the behaviour you look for.

diego-G:
I realise all files have in common function `confirmTransactionsOnAllNodes` and describe block `prepare accounts`. I suggest to extract them into a `common.js` to avoid duplicated code.

PR #1872, opened 2018-04-13
Tests negative verification result for manipulated data - Closes #1871

### LINKED ISSUE #1871: Add tests for manipulated block data
### Expected behavior

When a single field in a block is manipulated, verifySignature must return false. This should be covered by unit tests.

### Actual behavior

The two tests of verifySignature check that

1. broken blocks throw
2. verifySignature returns true for valid blocks

These tests will also pass for an implementation that statically returns `true` for every block with the correct data type.

### Steps to reproduce

See `test/unit/logic/block.js` in current development branch

* development

webmaster128:
### What was the problem?

Test coverage was not sufficient for verifySignature

### How did I fix it?

Added tests that ensure verifySignature returns `false` as soon as any of the fields is manipulated.

### How to test it?

Run `./node_modules/.bin/mocha ./test/unit/logic/block.js`

### Review checklist

* The PR solves #1871
* Linting passes
* Tests pass

### DIFF
--- a/test/unit/logic/block.js
+++ b/test/unit/logic/block.js
@@ -562,8 +562,77 @@ describe('block', () => {
 			}).to.throw();
 		});
 
-		it('should return verification response for a given block', () => {
+		it('should return true for a good block', () => {
 			return expect(block.verifySignature(blockData)).to.be.true;
+		});
+
+		it('should return false for a block with modified version', () => {
+			const blockCopy = Object.assign({}, blockData);
+			blockCopy.version += 1;
+			return expect(block.verifySignature(blockCopy)).to.be.false;
+		});
+
+		it('should return false for a block with modified timestamp', () => {
+			const blockCopy = Object.assign({}, blockData);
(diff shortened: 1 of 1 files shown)


test/unit/logic/block.js
```diff
@@ -562,9 +562,78 @@ describe('block', () => {
 			}).to.throw();
 		});
 
-		it('should return verification response for a given block', () => {
+		it('should return positive for a good block', () => {
 			return expect(block.verifySignature(blockData)).to.be.true;
 		});
+
+		it('should return negative for a block with modified version', () => {
```
4miners: Please change those descriptions, as `negative` is more like negative number. For example: `should fail for a block...` or `should return false for...`

webmaster128:
Rebased onto 1.0.0 and feedback from @4miners applied

webmaster128:
It's a pleasure! Thanks for being open to external contributions

PR #1874, opened 2018-04-13
Improve readability of verifySignature() - Closes #1873

### LINKED ISSUE #1873: Improve readability of verifySignature()
### Expected behavior

verifySignature() should be easy to review (such that the good hackers have fun reading the code)

### Actual behavior

* verifySignature() contains some variable names, that do not describe what they are doing
* verifySignature() contains a complicated copy mechanism where only a view of a Buffer is required
* getBytes calls a variable `size`, which is not a size. If it was a size, verifications would fail
* getBytes contains a list of integers which is hard to understand

### Steps to reproduce

Read https://github.com/LiskHQ/lisk/blob/a0e0c0a/logic/block.js#L188 and https://github.com/LiskHQ/lisk/blob/a0e0c0a/logic/block.js#L384

Current development branch

webmaster128:
### What was the problem?

verifySignature() was hard to review

### How did I fix it?

Refactor the code

### How to test it?

* Run unit tests
* Merge #1872 for extended coverage and run unit tests again

### Review checklist

* Resolves #1873
* Linting passes
* Tests pass

### DIFF
--- a/logic/block.js
+++ b/logic/block.js
@@ -382,7 +382,7 @@ Block.prototype.schema = {
  * @todo Add description for the function and the params
  */
 Block.prototype.getBytes = function(block) {
-	const size =
+	const capacity =
 		4 + // version (int)
 		4 + // timestamp (int)
 		8 + // previousBlock
@@ -398,7 +398,7 @@ Block.prototype.getBytes = function(block) {
 	let bytes;
 
 	try {
-		const byteBuffer = new ByteBuffer(size, true);
+		const byteBuffer = new ByteBuffer(capacity, true);
 		byteBuffer.writeInt(block.version);
 		byteBuffer.writeInt(block.timestamp);
 
--- a/logic/block.js
+++ b/logic/block.js
@@ -186,19 +186,19 @@ class Block {
 	 * @todo Add description for the params
 	 */
 	verifySignature(block) {
-		const remove = 64;
+		const signatureLength = 64;
 		let res;
 
 		try {
 			const data = this.getBytes(block);
-			const data2 = Buffer.alloc(data.length - remove);
+			const dataWithoutSignature = Buffer.alloc(data.length - signatureLength);
 
-			for (let i = 0; i < data2.length; i++) {
-				data2[i] = data[i];
+			for (let i = 0; i < dataWithoutSignature.length; i++) {
+				dataWithoutSignature[i] = data[i];
 			}
 			const hash = crypto
 				.createHash('sha256')
-				.update(data2)
+				.update(dataWithoutSignature)
 				.digest();
 			const blockSignatureBuffer = Buffer.from(block.blockSignature, 'hex');
 			const generatorPublicKeyBuffer = Buffer.from(
--- a/logic/block.js
+++ b/logic/block.js
@@ -382,7 +382,20 @@ Block.prototype.schema = {
  * @todo Add description for the function and the params
  */
 Block.prototype.getBytes = function(block) {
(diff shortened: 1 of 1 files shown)


logic/block.js
```diff
@@ -191,11 +191,7 @@ class Block {
 
 		try {
 			const data = this.getBytes(block);
-			const dataWithoutSignature = Buffer.alloc(data.length - signatureLength);
-
-			for (let i = 0; i < dataWithoutSignature.length; i++) {
-				dataWithoutSignature[i] = data[i];
-			}
+			const dataWithoutSignature = data.slice(0, -signatureLength);
```
diego-G: Please, we feel this commit changes critical part of the code without adding any readability. At this point of heavy development, we prefer to be cautious and ask you kindly to revert or delete the related commit.

logic/block.js
```diff
@@ -382,11 +378,23 @@ Block.prototype.schema = {
...
+		8 + // previousBlock
+		4 + // numberOfTransactions (int)
+		8 + // totalAmount (long)
+		8 + // totalFee (long)
+		8 + // reward (long)
+		4 + // payloadLength (int)
+		32 + // payloadHash
+		32 + // generatorPublicKey
+		64 + // blockSignature or unused
+		4; // unused
```
willclarktech: It would even be nice to store these in variables/constants.

webmaster128:
@MaciejBaj I can rebase this onto the 1.0.0 branch if you prefer to have it there. This would remove the 22 commits and make per-commit review easier.

logic/block.js
webmaster128: Agree. But I'd prefer to to do this as a follow up step. The idea here is to make things readable without risking any functional change and without requiring major refactoring decisions. There are more follow up step to be done after this. For example decide what to do with those 4 bytes of unused capacity.

diego-G: Besides, we should check figures apply against these variables before writing to the buffer.

logic/block.js
webmaster128: You are talking about https://github.com/LiskHQ/lisk/pull/1874/commits/950ddcbd73aacf1035200aca66accb6df00caad8, right?

webmaster128:
@diego-G which branch is "latest stable"?

diego-G:
@webmaster128 `1.0.0`

webmaster128:
Rebased and commit https://github.com/LiskHQ/lisk/commit/950ddcbd73aacf1035200aca66accb6df00caad8 deleted. cc @diego-G

logic/block.js
diego-G: @IkerAlus last 4 unused bytes might have any not clear use?

webmaster128: Those 4 bytes do not make it into the result of `getBytes`. See for example

```
const ByteBuffer = require('bytebuffer');
var b = new ByteBuffer(30, true);
b.writeLong(123);
b.flip();
b.toBuffer().length; // returns 8
```

willclarktech: This is a different list of numbers to what's in place currently. Can you explain the discrepancy?

webmaster128: Jap, see this commit https://github.com/LiskHQ/lisk/pull/1874/commits/1ba82f540981e47a594f76da1651ebc5e040095b. A 4 was associated with int, which was wrong and an unused extra 4 was moved to the end to get it out of the way.

webmaster128:
Let me rebase this into 1.0.0 to make sure the additional tests from #1872 are passing with this change as well.

@willclarktech did I answer you question from yesterday? Or is there anything unclear I can elaborate on?

webmaster128:
Rebased onto current 1.0.0. When tests pass, this is ready from my point of view.

PR #1877, opened 2018-04-13
Never reuse sockets and always destroy socket instead of just disconnecting - Closes #1876

### LINKED ISSUE #1876: Never reuse sockets and always destroy socket instead of just disconnecting and always unbind all event handlers - Potential memory leak
### Expected behavior

The peer socket should be completely destroyed when the peer disconnects or is removed. All event handlers should be removed completely.

### Actual behavior

It's possible under some scenarios that a socket could be reused and this seems to cause event handlers to leak on that socket.

### Steps to reproduce

Run a localhost node, connect to latest beta net.

1.0.0, latest beta net.

jondubois:
### What was the problem?

The socket should be completely destroyed when the peer disconnects or is removed. All event handlers should be removed completely. This appeared to have caused part of the memory leaks that we've encountered.

### How did I fix it?

Destroy the socket completely and never reuse it and unbind all events.

### How to test it?

Connect localhost node to latest betanet.5

### Review checklist

* The PR solves #1876
* Linting passes
* Tests pass


shuse2:
The message is coming from server side, and probably the node is too busy to establish the connection.
I think, therefore, the error message will not be related directly to this PR.

api/ws/rpc/connect.js
```diff
@@ -159,13 +160,7 @@ const connectSteps = {
 	},
 
 	registerSocketListeners: (peer, logger, onDisconnectCb = () => {}) => {
-		// Unregister all the events just in case
 		const socket = peer.socket;
-		socket.off('connect');
```
ManuGowda: I believe we need to remove the event listeners if we delete this code we might get other errors, later on, so removing event listeners don't cause any side effects.

ManuGowda:
@shuse2 thats correct, however, my node node should not crash due to the `Client connection establishment timed out` so we need to fix this issue, @jondubois has created an issue for it https://github.com/LiskHQ/lisk/issues/1875

api/ws/rpc/connect.js
jondubois: We don't need to remove the event listeners anymore because now that `multiplex` is false, it will always create a new socket object so the socket will never have any event handlers attached at this point in the code so calling `off` here is wasteful and it opens up the possibility that an 'error' event on the socket could be uncaught and crash the worker process.  It's possible that the `socket.off('error')` call here is responsible for this issue: https://github.com/LiskHQ/lisk/issues/1875

All event handlers are attached to the SCSocket itself so they will be all be garbage collected at the same time as the SCSocket itself is garbage collected (when it.

PR #1881, opened 2018-04-16
Inconsistent behavior between syncing and snapshotting - Closes #1879

### LINKED ISSUE #1879: Inconsistent behavior between syncing and snapshotting
### Expected behavior
Snapshotting and syncing from scratch will perform exactly the same operations while applying transactions during recreating blockchain. Missing `__private.checkTransaction` step of `blocks/verify` submodule will be performed while applying transaction during creating a snapshot.
### Actual behavior
Creating a snapshot and syncing with the rest of the network might result in a different result. `__private.checkTransaction` is missing while creating a snapshot.
### Steps to reproduce
Sync with the network and create a snapshot.
1.0.0 and 0.9.*

nazarhussain:
### What was the problem?
Snapshotting logic was using different logic compared to sync, which was causing inconsistent behavior while syncing with the network or restoring the snapshot. 

### How did I fix it?
Update the snapshotting logic to use the same code base as syncing with modified behavior for not saving blocks. 

### How to test it?
Take a snapshot and restore. 

### Review checklist

* The PR solves #1879
* Linting passes
* Tests pass

### DIFF
--- a/logic/transaction.js
+++ b/logic/transaction.js
@@ -407,15 +407,24 @@ Transaction.prototype.process = function (trs, sender, requester, cb) {
  * @param {transaction} trs
  * @param {account} sender
  * @param {account} requester
+ * @param {boolean} checkConfirmed - Check if trs already persisted in database
  * @param {function} cb
  * @return {setImmediateCallback} validation errors | trs
  */
-Transaction.prototype.verify = function (trs, sender, requester, cb) {
+Transaction.prototype.verify = function (trs, sender, requester, checkConfirmed, cb) {
 	var valid = false;
 	var err = null;
 
 	if (typeof requester === 'function') {
 		cb = requester;
+		requester = {};
+	}
+
+	if(checkConfirmed === undefined || checkConfirmed === null) {
+		checkConfirmed = true;
+	}
+	if(requester === undefined || requester === null) {
+		requester = {};
 	}
 
 	// Check sender
@@ -591,7 +600,11 @@ Transaction.prototype.verify = function (trs, sender, requester, cb) {
 			return setImmediate(cb, err);
 		} else {
 			// Check for already confirmed transaction
-			return self.checkConfirmed(trs, cb);
+			if(checkConfirmed) {
+				return self.checkConfirmed(trs, cb);
+			} else {
+				return setImmediate(cb);
+			}
 		}
 	});
 };
--- a/logic/transaction.js
+++ b/logic/transaction.js
@@ -591,14 +591,11 @@ Transaction.prototype.verify = function (trs, sender, requester, checkExists, cb
 	__private.types[trs.type].verify.call(this, trs, sender, function (err) {
 		if (err) {
 			return setImmediate(cb, err);
-		} else {
+		} else if (checkExists) {
 			// Check for already confirmed transaction
-			if (checkExists) {
-				return self.checkConfirmed(trs, cb);
-			} else {
-				return setImmediate(cb);
-			}
+			return self.checkConfirmed(trs, cb);
 		}
+		return setImmediate(cb);
 	});
 };
 
--- a/logic/transaction.js
+++ b/logic/transaction.js
@@ -416,7 +416,7 @@ Transaction.prototype.verify = function (trs, sender, requester, checkExists, cb
 	var err = null;
 
 	// Set default value of param if not provided
-	if(requester === undefined || requester === null) {
+	if (requester === undefined || requester === null) {
 		requester = {};
 	}
 
@@ -593,7 +593,7 @@ Transaction.prototype.verify = function (trs, sender, requester, checkExists, cb
 			return setImmediate(cb, err);
 		} else {
 			// Check for already confirmed transaction
-			if(checkExists) {
+			if (checkExists) {
 				return self.checkConfirmed(trs, cb);
 			} else {
 				return setImmediate(cb);
--- a/logic/transaction.js
+++ b/logic/transaction.js
@@ -407,22 +407,20 @@ Transaction.prototype.process = function (trs, sender, requester, cb) {
  * @param {transaction} trs
  * @param {account} sender
  * @param {account} requester
- * @param {boolean} checkConfirmed - Check if trs already persisted in database
+ * @param {boolean} checkPersistent - Check if trs already persisted in database
  * @param {function} cb
  * @return {setImmediateCallback} validation errors | trs
  */
-Transaction.prototype.verify = function (trs, sender, requester, checkConfirmed, cb) {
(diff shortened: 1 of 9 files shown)


modules/blocks/verify.js
```diff
@@ -82,10 +68,22 @@ __private.checkTransaction = function (block, transaction, cb) {
...
 		}
 	], function (err) {
-		return setImmediate(cb, err);
+
+		if(err && err.match(/Transaction is already confirmed/)) {
+			// Fork: Transaction already confirmed.
+			modules.delegates.fork(block, 2);
+			// Undo the offending transaction.
+			// DATABASE: write
+			modules.transactions.undoUnconfirmed(transaction, function (err2) {
```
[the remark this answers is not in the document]
nazarhussain: @SargeKhan As per flow of the code `undoUnconfirmed` will only be performed if `checkConfirmed` is true and we have some error during verify process, which also contains `checkConfirmed` step.

logic/transaction.js
```diff
@@ -407,15 +407,24 @@ Transaction.prototype.process = function (trs, sender, requester, cb) {
  * @param {transaction} trs
  * @param {account} sender
  * @param {account} requester
+ * @param {boolean} checkConfirmed - Check if trs already persisted in database
  * @param {function} cb
  * @return {setImmediateCallback} validation errors | trs
  */
-Transaction.prototype.verify = function (trs, sender, requester, cb) {
+Transaction.prototype.verify = function (trs, sender, requester, checkConfirmed, cb) {
```
4miners: Please refactor all calls of that function to always pass all parameters instead of later `if (typeof requester === 'function') {` thing.

modules/blocks/chain.js
```diff
@@ -358,6 +358,13 @@ Chain.prototype.applyBlock = function (block, broadcast, saveBlock, cb) {
 		},
 		// Apply transactions to unconfirmed mem_accounts fields.
 		applyUnconfirmed: function (seriesCb) {
+
+			// If block is not to be saved, then we don't need to process unconfirmed state
+			// that block must be restored from the database for snapshot process
+			if(!saveBlock) {
```
4miners: Please restore `applyUnconfirmed` for now, as we want snapshotting process to be as close as possible to sync process.

modules/blocks/verify.js
```diff
@@ -82,10 +68,22 @@ __private.checkTransaction = function (block, transaction, cb) {
 		function (sender, waterCb) {
 			// Check if transaction id valid against database state (mem_* tables).
 			// DATABASE: read only
-			library.logic.transaction.verify(transaction, sender, waterCb);
+			library.logic.transaction.verify(transaction, sender, null, checkConfirmed, waterCb);
 		}
 	], function (err) {
-		return setImmediate(cb, err);
+
+		if(err && err.match(/Transaction is already confirmed/)) {
```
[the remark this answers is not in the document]
4miners: I don't like `err.match` against error message.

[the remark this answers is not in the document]
nazarhussain: @4miners I hate such errors, but we are using string-based error messages around the whole application. We don't specify any error codes, so why have to deal with it like this.

logic/transaction.js
```diff
@@ -407,15 +407,22 @@ Transaction.prototype.process = function (trs, sender, requester, cb) {
...
  */
-Transaction.prototype.verify = function (trs, sender, requester, cb) {
+Transaction.prototype.verify = function (trs, sender, requester, checkPersistent, cb) {
 	var valid = false;
 	var err = null;
 
-	if (typeof requester === 'function') {
-		cb = requester;
+	// Set default value of param if not provided
+	if(checkPersistent === undefined || checkPersistent === null) {
```
4miners: Default to `true` is not good, we should call that function with `checkPersistent` either `true` or `false`, so that will be not needed.

logic/transaction.js
```diff
@@ -591,7 +598,11 @@ Transaction.prototype.verify = function (trs, sender, requester, cb) {
 			return setImmediate(cb, err);
 		} else {
 			// Check for already confirmed transaction
-			return self.checkConfirmed(trs, cb);
+			if(checkPersistent) {
+				return self.checkConfirmed(trs, cb);
```
4miners: please correct name of function `checkConfirmed` - it's misleading.

logic/transaction.js
```diff
@@ -407,15 +407,22 @@ Transaction.prototype.process = function (trs, sender, requester, cb) {
  * @param {transaction} trs
  * @param {account} sender
  * @param {account} requester
+ * @param {boolean} checkPersistent - Check if trs already persisted in database
```
4miners: maybe `checkExists`?

modules/blocks/process.js
```diff
@@ -139,18 +139,17 @@ Process.prototype.getCommonBlock = function (peer, height, cb) {
  * @method loadBlocksOffset
  * @param  {number}   limit Limit amount of blocks
  * @param  {number}   offset Offset to start at
- * @param  {boolean}  verify Indicator that block needs to be verified
  * @param  {Function} cb Callback function
  * @return {Function} cb Callback function from params (through setImmediate)
  * @return {Object}   cb.err Error if occurred
  * @return {Object}   cb.lastBlock Current last block
  */
-Process.prototype.loadBlocksOffset = function (limit, offset, verify, cb) {
```
4miners: We removed `verify` parameter, is it related with https://github.com/LiskHQ/lisk/issues/1868? If yes - please remove unused logic.

modules/blocks/verify.js
```diff
@@ -38,11 +38,15 @@ function Verify (logger, block, transaction, db) {
...
  * @param  {Object}   block Block object
  * @param  {Object}   transaction Transaction object
+ * @param  {boolean} checkPersistent - Check for confirmed transactions persisted state
  * @param  {Function} cb Callback function
  * @return {Function} cb Callback function from params (through setImmediate)
  * @return {Object}   cb.err Error if occurred
  */
-__private.checkTransaction = function (block, transaction, cb) {
+__private.checkTransaction = function (block, transaction, checkPersistent, cb) {
+	if(checkPersistent === undefined) {
```
4miners: Same here - please don't default to `true`.

modules/blocks/verify.js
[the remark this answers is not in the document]
4miners: I'm not sure if here is best place to do that.

logic/transaction.js
nazarhussain: Its used on more than one places in code with same name. Not seems right to rename it in the current scope.

modules/blocks/process.js
nazarhussain: We removed `verify` parameter because now its not optional. We will verify blocks during snapshotting in any case.

PR #1886, opened 2018-04-17
Terminate node on postgres connection error - Fixes #1779

### LINKED ISSUE #1779: Node does not die if Postgresql server stops
### Expected behavior

Node must die if it realises PSQL server has stopped.

### Actual behavior

Node continues alive after PSQL stops. Then it recovers properly when PSQL server restart.
Also the node's stills listening API calls as posting transaction. Fortunately, it produces 409 conflict error. 

### Steps to reproduce

`bash lisk.sh stop_db`

`v1.0.0-alpha1`

SargeKhan:
### What was the problem?
Application keeps running even when unable to connect to database
### How did I fix it?
Emit cleanup once application has connection error
### How to test it?
Kill postgres server and see if application exits.
### Review checklist

* The PR solves #1779 
* Linting passes
* Tests pass

### DIFF
--- a/db/index.js
+++ b/db/index.js
@@ -92,6 +92,15 @@ module.exports.connect = (config, logger) => {
 		info.display = false;
 	};
 
+	initOptions.error = (error, e) => {
+		logger.error(error);
+		// e.cn corresponds to an object, which exists only when there is a connection related error.
+		// https://vitaly-t.github.io/pg-promise/global.html#event:error
+		if (e.cn) {
+			process.emit('cleanup');
+		}
+	};
+
 	config.user = config.user || process.env.USER;
 
 	pgp.end();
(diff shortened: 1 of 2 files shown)


SargeKhan:
I don't know what kind of tests will be suitable for application startup. Do you have suggestions?

db/index.js
```diff
@@ -92,6 +92,15 @@ module.exports.connect = (config, logger) => {
 		info.display = false;
 	};
 
+	initOptions.error = (error, e) => {
+		logger.error(error);
+		// e.cn corresponds to an object, which exists only when there is a connection related error.
+		// https://vitaly-t.github.io/pg-promise/global.html#event:error
+		if (e.cn) {
```
MaciejBaj: @vitaly-t please confirm that the check is valid.

vitaly-t: This looks fine. You can always test it by tweaking the connection parameters :wink;

Issue #1839, opened 2018-04-10
Make webSocket task dependent on database


MaciejBaj:
### Expected behavior
If node is unable to connect to database on startup, it should not load websockets, and cleanly exit.
### Actual behavior
Even when unable to connect to db, we initialize socket cluster and make connections with peers.
### Steps to reproduce
Kill postgres server, and start the application.
### Which version(s) does this affect? (Environment, OS, etc...)
1.0.0

SargeKhan:
@MaciejBaj, close this issue?

SargeKhan:
Closed by #1890

PR #1890, opened 2018-04-17
Make webSocket task dependent on database - Closes # 1839

SargeKhan:
### What was the problem?
Even when unable to connect to db, we initialize socket cluster and make connections with peers.
### How did I fix it?
If node is unable to connect to database on startup, it will not load websockets, and cleanly exit.
### How to test it?
Kill postgres server, and start the application.
### Review checklist

* The PR solves #1839
* Linting passes
* Tests pass

### DIFF
--- a/app.js
+++ b/app.js
@@ -305,85 +305,6 @@ d.run(() => {
 				},
 			],
 
-			webSocket: [
-				'config',
-				'connect',
-				'logger',
-				'network',
-				/**
-				 * Description of the function.
-				 *
-				 * @func webSocket[4]
-				 * @memberof! app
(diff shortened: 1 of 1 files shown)


SargeKhan:
@jondubois @MaciejBaj, for readability purposes, different tasks in `async.auto` are ordered based on their expected execution. For example, if `logic` task is dependent on `db` task, then it is placed after `db` task. I've followed same approach, and now since `websocket` task was dependent on `db` I have moved it after `db` task (as a matter of fact, after all task which have no dependencies)
I've rebased, and now there are no extraneous commits.

Issue #1880, opened 2018-04-16
Introduce inert exceptions


MaciejBaj:
### Expected behavior
Inert exceptions will introduce a new way of applying exceptions. Transactions present in inert exceptions will be applied to blockchain and be present in `trs` table, but they will not produce any change in `mem_accounts` table. 
### Actual behavior
Evert transaction exception populate changes in `mem_accounts` table before being included in the blockchain. Recreating the same `mem_accounts` state as the rest of the network is impossible after fixing #1860. 
### Steps to reproduce
Create snapshot / Sync with the network which contains inert exceptions.
### Which version(s) does this affect? (Environment, OS, etc...)
0.9.*, 1.0.0

PR #1897, opened 2018-04-18
Add code changes for handling inert transactions - Subtask #1880

SargeKhan:
### What was the problem?
Invalid transactions after variable unshadowing were causing problems with syncing.
### How did I fix it?
Pass verification and application for invalid transactions already in the blockchain.
### How to test it?
N/A
### Review checklist

* The PR solves #1880 
* Linting passes
* Tests pass


4miners:
@SargeKhan Would be nice to have functional system tests for that, scenario:
- add inert transactions of affected types to execptions (delegate registration, vote)
- push a block with those transactions
- check if behavior is as expected (transactions are not applied to `mem_accounts` and `mem_round`)
- forge until end of a round
- check if behavior is as expected:
  - `vote` column is updated properly (not affected by those transactions)
  - fees (not including affected transactions) are split correctly to forging delegates (`rounds_fees` table on `0.9.x`, `rounds_rewards` table on `1.0.0`)

Can you write them?

SargeKhan:
@4miners I’ll write these tests, but since testnet snaphshotting etc. takes a couple of days, i would ask to merge open pr, since we have definitely confirmed that this fix solves our problem. I’ll create a new branch with only tests, and we can merge that branch later. what do you think?

PR #1909, opened 2018-04-20
Reviving destroyed sockets memory leak - Closes #1907

### LINKED ISSUE #1907: Reviving destroyed sockets memory leak
### Expected behavior

Outbound client sockets should not cause a memory leak.

### Actual behavior

When the socket is destroyed, it gets removed from the clients list but it is not always removed from the peer.socket property so it's still possible to send messages to it - So it's possible that a socket can become reactivated even though it's no longer attached to a peer.

### Steps to reproduce

Test on beta net with lots of connections and disconnections.

1.0.0

jondubois:
### What was the problem?

Some sockets were being revived after they were destroyed.
A better strategy is needed for handling socket disconnections so that it doesn't leak.

### How did I fix it?

Use SocketCluster's multiplexing feature so that sockets are reused instead of creating a new one each time we re-add a peer (which causes a memory leak).
This patch was tested on beta net and it appears to fix the issue of leaking socket handles.

### How to test it?

Test on beta net.

### Review checklist

* The PR solves #1907
* Linting passes
* Tests pass

### DIFF
--- a/api/ws/rpc/connect.js
+++ b/api/ws/rpc/connect.js
@@ -127,14 +127,20 @@ const connectSteps = {
 						data
 					);
 
-					peer.socket
-						.call(rpcProcedureName, data)
-						.then(res => {
-							setImmediate(rpcCallback, null, res);
-						})
-						.catch(err => {
-							setImmediate(rpcCallback, err);
-						});
+					if (peer.socket) {
+						peer.socket
+							.call(rpcProcedureName, data)
+							.then(res => {
+								setImmediate(rpcCallback, null, res);
+							})
+							.catch(err => {
+								setImmediate(rpcCallback, err);
+							});
+					} else {
+						logger.debug(
+							'Tried to call RPC function on outbound peer socket which no longer exists'
+						);
+					}
 				};
 				return peerExtendedWithRPC;
 			},
@@ -203,6 +209,14 @@ const connectSteps = {
 				} closed with code ${code} and reason - ${reason}`
 			);
 			socket.destroy();
+			if (socket == peer.socket) {
+				delete peer.socket;
+			} else {
+				// This condition should never happen but log it in order to catch regressions.
+				logger.error(
+					'The socket which closed did not match the current peer socket'
+				);
+			}
 			onDisconnectCb();
 		});
 
--- a/api/ws/rpc/connect.js
+++ b/api/ws/rpc/connect.js
@@ -15,7 +15,6 @@
 'use strict';
 
 const _ = require('lodash');
-const async = require('async');
 const scClient = require('socketcluster-client');
 const WAMPClient = require('wamp-socket-cluster/WAMPClient');
 const failureCodes = require('../../../api/ws/rpc/failure_codes');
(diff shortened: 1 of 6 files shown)


jondubois:
I want to make a few more changes.

api/ws/rpc/connect.js
```diff
@@ -199,8 +199,14 @@ const connectSteps = {
 					peer.ip
 				} failed with code ${code} and reason - ${reason}`
 			);
-			socket.destroy();
-			onDisconnectCb();
+
+			if (peer.socket && peer.socket.state === peer.socket.CLOSED) {
```
ManuGowda: is it `peer.socket.state === peer.socket.CLOSED` or `peer.socket.state === Peer.STATE.CLOSED`?

api/ws/rpc/connect.js
```diff
@@ -199,8 +199,14 @@ const connectSteps = {
...
+				peer.state = Peer.STATE.DISCONNECTED;
```
ManuGowda: why are we changing the state from `closed` to `disconnected`?

api/ws/rpc/connect.js
jondubois: @ManuGowda it's peer.socket.state not peer.state because we want to rely on the socket's state not the peer's state.

api/ws/rpc/connect.js
```diff
@@ -48,25 +50,17 @@ const connectSteps = {
 			connectTimeout: TIMEOUT,
 			ackTimeout: TIMEOUT,
 			pingTimeout: TIMEOUT,
-			connectAttempts: 1,
 			port: peer.wsPort,
 			hostname: peer.ip,
 			query: System.getHeaders(),
-			multiplex: false,
+			multiplex: true,
```
ManuGowda: If I understand correctly, we want to reuse a socket instead of creating a second socket to the same address, so we `multiplex` is set to true? if its the case we can remove this option, as by default the `multiplex` is true.

helpers/peers_manager.js
```diff
@@ -52,22 +51,15 @@ PeersManager.prototype.add = function(peer) {
...
-				lostPeer.state = Peer.STATE.DISCONNECTED;
-			}
-		});
+		connect(peer, this.logger);
+	} else {
+		// Reconnect existing socket
+		peer.socket.connect();
 	}
-	this.addressToNonceMap[peer.string] = peer.nonce;
 	if (peer.nonce) {
```
ManuGowda: we don't need a check for `peer.string`? I think consistent check again both `peer.string` and `peer.nonce` will ensure consistency.

api/ws/rpc/connect.js
jondubois: @ManuGowda In one case, we're comparing the socket state and in the other, we're comparing the peer state.

api/ws/rpc/connect.js
jondubois: @ManuGowda Yes it's the default but it doesn't hurt to be explicit to avoid confusion. Also if there is an update to socketcluster-client in the future which changes the default, it's better if we lock it down.

PR #1919, opened 2018-04-23
Fix logging of database events received from pgMonitor - Closes #1885

### LINKED ISSUE #1885: Fix logging of database events received from pgMonitor
### Expected behavior
Log levels for database events are decided by `db.logEvents` property, and application should log to file and console based on event types specified in the property. 
### Actual behavior
The application uses database logger with configuration based on application events (debug, trace, info) which have no one-to-one mapping with dblog events. 
### Steps to reproduce
Look at the db log file.
1.0.0

SargeKhan:
### What was the problem?
The application uses database logger with configuration based on application events (debug, trace, info) which have no one-to-one mapping with dblog events.
### How did I fix it?
Log all events on log level based on logEvents property.
### How to test it?
Check the logs
### Review checklist

* The PR solves #1885 
* Linting passes
* Tests pass

### DIFF
--- a/app.js
+++ b/app.js
@@ -167,9 +167,10 @@ if (
 ) {
 	dbLogger = logger;
 } else {
+	// since log levels for database monitor are different than node app, i.e. "query", "info", "error" etc, which is decided using "logEvents" property
 	dbLogger = new Logger({
-		echo: process.env.DB_LOG_LEVEL || appConfig.db.consoleLogLevel,
-		errorLevel: process.env.FILE_LOG_LEVEL || appConfig.db.fileLogLevel,
+		echo: process.env.DB_LOG_LEVEL || 'log',
+		errorLevel: process.env.FILE_LOG_LEVEL || 'log',
 		filename: appConfig.db.logFileName,
 	});
 }
@@ -543,9 +544,9 @@ d.run(() => {
 			db(cb) {
 				var db = require('./db');
 				db
-					.connect(config.db, dbLogger)
+					.connect(config.db, dbLogger, logger)
 					.then(db => cb(null, db))
-					.catch(err => cb(err));
+					.catch(cb);
 			},
 
 			/**
--- a/db/index.js
+++ b/db/index.js
@@ -75,11 +75,12 @@ module.exports.pgp = pgp;
  *
  * @function connect
  * @param {Object} config
- * @param {function} logger
+ * @param {Object} dbLogger - logger for database log file
+ * @param {Object} logger - logger for application
  * @returns {Promise}
  * @todo Add description for the params and the return value
  */
-module.exports.connect = (config, logger) => {
+module.exports.connect = (config, dbLogger, logger) => {
 	if (monitor.isAttached()) {
 		monitor.detach();
 	}
@@ -88,7 +89,7 @@ module.exports.connect = (config, logger) => {
 	monitor.setTheme('matrix');
 
 	monitor.log = (msg, info) => {
-		logger.log(info.event, info.text);
+		dbLogger.log(info.event, info.text);
 		info.display = false;
 	};
 


app.js
```diff
@@ -167,9 +167,10 @@ if (
 ) {
 	dbLogger = logger;
 } else {
+	// since log levels for database monitor are different than node app, i.e. "query", "info", "error" etc, which is decided using "logEvents" property
 	dbLogger = new Logger({
-		echo: process.env.DB_LOG_LEVEL || appConfig.db.consoleLogLevel,
-		errorLevel: process.env.FILE_LOG_LEVEL || appConfig.db.fileLogLevel,
+		echo: process.env.DB_LOG_LEVEL || 'log',
+		errorLevel: process.env.FILE_LOG_LEVEL || 'log',
```
nazarhussain: We should update the config `schema` if we don't want to use these values from config.

https://github.com/LiskHQ/lisk/blob/20d0bbab7bc5874b802344a1c31a8cdabea7dcd9/schema/config.js#L99-L104

Or we can use like 

```
process.env.DB_LOG_LEVEL || appConfig.db.consoleLogLevel || 'log'
```

db/index.js
```diff
@@ -75,11 +75,12 @@ module.exports.pgp = pgp;
...
  * @function connect
  * @param {Object} config
- * @param {function} logger
+ * @param {Object} dbLogger - logger for database log file
+ * @param {Object} logger - logger for application
  * @returns {Promise}
  * @todo Add description for the params and the return value
  */
-module.exports.connect = (config, logger) => {
+module.exports.connect = (config, dbLogger, logger) => {
```
nazarhussain: It seems you added new param but never used it anywhere.

app.js
```diff
@@ -543,9 +544,9 @@ d.run(() => {
 			db(cb) {
 				var db = require('./db');
 				db
-					.connect(config.db, dbLogger)
+					.connect(config.db, dbLogger, logger)
```
nazarhussain: No need to pass two logger. Just db specific logger is enough for `db` to log any thing.

db/index.js
SargeKhan: this new `dbLogger` is used for logging to database file. And the original `logger` is used [here](https://github.com/LiskHQ/lisk/pull/1919/files/f575473e78e9a180ff2a1363193064fc829f8020#diff-b13c80ec549dd76940c7f534e5d7f290R97)

app.js
SargeKhan: `dbLogger` is used for logging to database log file, where as `logger` is used for logging incase of an `error` which later mighht causes application cleanup. [here](https://github.com/LiskHQ/lisk/pull/1919/files/f575473e78e9a180ff2a1363193064fc829f8020#diff-b13c80ec549dd76940c7f534e5d7f290R97)

db/index.js
nazarhussain: @SargeKhan It seems illogical that we log some part of db layer to one file and some part other log file. 

https://github.com/LiskHQ/lisk/blob/20d0bbab7bc5874b802344a1c31a8cdabea7dcd9/db/index.js#L95-L97

We should update this line as well to log any db error same database log file.

SargeKhan:
@nazarhussain and I have a disagreement on the logging behavior. Would nice get more reviews, and follow the approach which is a majority of the team agrees on. :)

nazarhussain:
@SargeKhan The key factor is we should not log duplicate entires in log files. You are considering that since both are seperate files so there will be no duplicate entires in particular log file. But splitting log files is optional, and if we don't split you will have duplicate entries. 

For any particular period of time, if our application is splitting logs to seperate files based on components and layers, there must not be duplicate entries among any file. If we want something important to highlight, then we can create a shared log file, where each application component can log. e.g. errors.

jondubois:
I think the main log should is used for critical DB errors (which affect the application layer) but the dbLogger is a more detailed version which also includes non-critical stuff.

I think some overlap is probably OK?

nazarhussain:
@SargeKhan @jondubois @MaciejBaj That's the line of code we are discussing in this PR; 

https://github.com/LiskHQ/lisk/blob/20d0bbab7bc5874b802344a1c31a8cdabea7dcd9/db/index.js#L95-L101

So its straight forward `logger` here must refer to logger of db layer, not application layer. And if you want to log something in application layer then pass that error with the event. 

```
process.emit('cleanup', new LiskDbError('Connection Error'));
```
Then in application layer where we have `process.on('cleanup', ....)` handle that error and do what ever needed. 

If you think we don't have `LiskDbError `, yes we don't but we planned to create all custom error classes for application after `1.0.0`. So leave that part until then.

PR #1920, opened 2018-04-23
Provide API endpoint for transactions of an account - Closes #1892 

### LINKED ISSUE #1892: Provide API endpoint for transactions of an account
### Expected behavior
Based on the discussion between frontend team, full-stack team, and @MaciejBaj we had in the office today we came to the conclusion that there should be an endpoint to get all (incoming and outgoing) transactions of a given address.
Something like:

`/api/accounts/{address}/transactions?offset=100&limit=20&type=0`

Listing transactions of an account is a basic need of every banking application.

### Actual behavior

In Lisk Core 0.9.X it was possible to get transactions of account with
`/api/transactions?senderId={address}&recipientId={address}&offset=100&limit=20`

But in Lisk Core 1.0.0 this is no longer possible because it connects `senderId` and `recipientId` with `AND`, not `OR` as it used to in 0.9.X

To some extent, it can be emulated by doing a separate request for `senderId` and separate for `recipientId`, but that doesn't scale with higher `offset`. As soon as, there is a higher offset (like 1000) I would have to fetch all the transactions for both `senderId` and `recipientId` up to that offset, because I cannot know how many outgoing and how many incoming are before the offset.

`/api/transactions?senderId={address}&recipientId={address}&offset=1000&limit=20`

Lisk Core 1.0.0

Tschakki:
### What was the problem?
The new API had no endpoint to collect all the transactions of a specific address at once.

### How did I fix it?
Create new filter `senderIdOrRecipientId` for transactions.

### How to test it?
Make API call: `/api/transactions?senderIdOrRecipientId=<address>`

### Review checklist

* The PR solves #1892 
* Linting passes
* Tests pass

### DIFF
--- a/modules/transactions.js
+++ b/modules/transactions.js
@@ -103,7 +103,7 @@ __private.list = function(filter, cb) {
 	const where = [];
 	const allowedFieldsMap = {
 		senderIdOrRecipientId:
-			'"t_senderId" IN ${senderIdOrRecipientId} OR "t_recipientId" IN ${senderIdOrRecipientId}',
+			'"t_senderId" IN (${senderIdOrRecipientId}) OR "t_recipientId" IN (${senderIdOrRecipientId})',
 		id: '"t_id" = ${id}',
 		blockId: '"t_blockId" = ${blockId}',
 		fromHeight: '"b_height" >= ${fromHeight}',
--- a/modules/transactions.js
+++ b/modules/transactions.js
@@ -103,7 +103,7 @@ __private.list = function(filter, cb) {
 	const where = [];
 	const allowedFieldsMap = {
 		senderIdOrRecipientId:
-			'"t_senderId" IN (${senderIdOrRecipientId}) OR "t_recipientId" IN (${senderIdOrRecipientId})',
+			'"t_senderId" = ${senderIdOrRecipientId} OR "t_recipientId" = ${senderIdOrRecipientId}',
 		id: '"t_id" = ${id}',
 		blockId: '"t_blockId" = ${blockId}',
 		fromHeight: '"b_height" >= ${fromHeight}',
--- a/modules/transactions.js
+++ b/modules/transactions.js
@@ -103,7 +103,7 @@ __private.list = function(filter, cb) {
 	const where = [];
 	const allowedFieldsMap = {
 		senderIdOrRecipientId:
-			'"t_senderId" IN (${senderIdOrRecipientId:csv}) OR "t_recipientId" IN (${senderIdOrRecipientId:csv})',
+			'"t_senderId" IN ${senderIdOrRecipientId} OR "t_recipientId" IN ${senderIdOrRecipientId}',
 		id: '"t_id" = ${id}',
 		blockId: '"t_blockId" = ${blockId}',
 		fromHeight: '"b_height" >= ${fromHeight}',
--- a/modules/transactions.js
+++ b/modules/transactions.js
@@ -102,6 +102,8 @@ __private.list = function(filter, cb) {
 	const params = {};
 	const where = [];
 	const allowedFieldsMap = {
+		address:
+			't_senderId" IN (${senderId:csv}) OR t_recipientId" IN (${recipientId:csv})',
 		id: '"t_id" = ${id}',
 		blockId: '"t_blockId" = ${blockId}',
 		fromHeight: '"b_height" >= ${fromHeight}',
--- a/modules/transactions.js
+++ b/modules/transactions.js
@@ -103,7 +103,7 @@ __private.list = function(filter, cb) {
 	const where = [];
 	const allowedFieldsMap = {
 		senderIdOrRecipientId:
-			'"t_senderId" IN (${senderId:csv}) OR "t_recipientId" IN (${recipientId:csv})',
+			'"t_senderId" IN (${senderIdOrRecipientId:csv}) OR "t_recipientId" IN (${senderIdOrRecipientId:csv})',
 		id: '"t_id" = ${id}',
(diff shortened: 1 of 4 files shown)


modules/transactions.js
```diff
@@ -102,6 +102,8 @@ __private.list = function(filter, cb) {
 	const params = {};
 	const where = [];
 	const allowedFieldsMap = {
+		senderIdOrRecipientId:
+			't_senderId" IN (${senderId:csv}) OR t_recipientId" IN (${recipientId:csv})',
```
lsilvs: It seems to be missing double quote in the beginning of `t_senderId` and `t_recipientId`

schema/swagger.yml
```diff
@@ -1088,13 +1088,13 @@ parameters:
...
-  	name: senderIdOrRecipientId
-  	in: query
-  	description: Sender or Recipient address to query
-  	type: string
-  	format: address
-  	minLength: 1
-  	maxLength: 22
+    name: senderIdOrRecipientId
+    in: query
+    description: Public key to query
```
nazarhussain: @Tschakki Its the address not the `public key`. Seems you forget to update it.

modules/transactions.js
```diff
@@ -102,6 +102,8 @@ __private.list = function(filter, cb) {
 	const params = {};
 	const where = [];
 	const allowedFieldsMap = {
+		senderIdOrRecipientId:
+			'"t_senderId" IN (${senderId:csv}) OR "t_recipientId" IN (${recipientId:csv})',
```
lsilvs: `(${senderId:csv})` and `(${recipientId:csv})` should be `(${senderIdOrRecipientId:csv})`

Tschakki: haha yes, just found it out myself ^^

test/functional/http/get/transactions.js
```diff
@@ -383,6 +398,52 @@ describe('GET /api/transactions', () => {
...
 		});
 
+		describe('senderIdOrRecipientId', () => {
+			it('using invalid senderIdOrRecipientId should fail', () => {
+				return transactionsEndpoint
+					.makeRequest({ senderIdOrRecipientId: '' }, 400)
+					.then(res => {
+						expectSwaggerParamError(res, 'senderIdOrRecipientId');
+					});
+			});
```
nazarhussain: @Tschakki This case handles the case fo `empty` string. You should also add a case where you specify value but invalid value.

test/functional/http/get/transactions.js
```diff
@@ -383,6 +398,52 @@ describe('GET /api/transactions', () => {
...
+						expect(res.body.data).to.not.empty;
+						res.body.data.map(transaction => {
+							if (accountId === transaction.recipientId) {
+								recipient.push(transaction);
+							} else if (accountId === transaction.senderId) {
+								sender.push(transaction);
+							}
+						});
+						expect(sender).to.have.length(1);
+						expect(recipient).to.have.length(1);
```
nazarhussain: The transactions you posted are deterministic and you know there are only two such transactions. So having more clear expectations will be nice. 

```
expect(res.body.data).to.not.empty;
expect(res.body.data[0].recipientId).to.be.eql(accountId);
expect(res.body.data[1].senderId).to.be.eql(accountId);
```

Notice how clean three line expectations look like and easy to understand.

test/functional/http/get/transactions.js
```diff
@@ -383,6 +398,52 @@ describe('GET /api/transactions', () => {
...
+								accountFixtures.genesis.address,
+								accountFixtures.existingDelegate.address,
+							],
+						},
+						400
+					)
+					.then(res => {
+						expectSwaggerParamError(res, 'senderIdOrRecipientId');
+					});
+			});
```
nazarhussain: This definitely will fail, as the format of the param you specified in swagger is `string` not the array. So I don't think its a right test to be added.

modules/transactions.js
```diff
@@ -102,6 +102,8 @@ __private.list = function(filter, cb) {
 	const params = {};
 	const where = [];
 	const allowedFieldsMap = {
+		senderIdOrRecipientId:
+			'"t_senderId" IN (${senderIdOrRecipientId:csv}) OR "t_recipientId" IN (${senderIdOrRecipientId:csv})',
```
nazarhussain: One test case you should add is to check `:csv` formatter. Means send comma separated string value of this param. I believe it will fail because of `address` format we specified in swagger. So either you have to create new format for it or drop `:csv` here.

test/functional/http/get/transactions.js
Tschakki: It is true there are only these 2 transactions, but is are they returned always the same chronological order inside of `res.body.data`?
If not this expectations could fail randomly.

nazarhussain: Logically there should be one specific order whatever it would be. There can't be random order.

Tschakki: ok, will change it then!

modules/transactions.js
```diff
@@ -102,6 +102,8 @@ __private.list = function(filter, cb) {
 	const params = {};
 	const where = [];
 	const allowedFieldsMap = {
+		senderIdOrRecipientId:
+			'"t_senderId" IN (${senderIdOrRecipientId}) OR "t_recipientId" IN (${senderIdOrRecipientId})',
```
nazarhussain: @Tschakki That needs to be changed, now as you are doing checking for only one variable then you should do with the direct assignment. 

```
"t_senderId" = (${senderIdOrRecipientId}) OR "t_recipientId" = (${senderIdOrRecipientId})
```

PR #1926, opened 2018-04-24
Remove exceptions added in 0.9.13 for mainnet - Closes #1916

### LINKED ISSUE #1916: Remove exceptions added in 0.9.13 for mainnet
### Expected behavior
Exceptions should be there only if required. Looks like exceptions added in commit https://github.com/LiskHQ/lisk/commit/fdd64a3c856df3e3c6f6e1e2b4fe0bd2b349db86 are not needed, so we should remove them.
### Actual behavior
We have exceptions that are not needed.
### Steps to reproduce
- perform snapshotting without those exceptions
- perform full sync without those exceptions
0.9.13

4miners:
### What was the problem?
Exceptions should be there only if required. Looks like exceptions added in commit https://github.com/LiskHQ/lisk/commit/fdd64a3c856df3e3c6f6e1e2b4fe0bd2b349db86 are not needed, so we should remove them.
### How did I fix it?
Removed exceptions added in `0.9.13` for mainnet.
### How to test it?
Tested and confiormed manually.
### Review checklist

* The PR solves #1916
* Linting passes
* Tests pass


4miners:
Sync passed.

nazarhussain:
@4miners @MaciejBaj @SargeKhan  I believe we should debug the exact reason why we got those exceptions before and not this time. There may be the wrong logic to log the exceptions before, but we should look into it.

PR #1938, opened 2018-04-25
Update Jenkins Migration to Use Lisk Bridge script to test migration - Closes #1861

### LINKED ISSUE #1861: Update Jenkins Migration to Use Lisk Bridge script to test migration
### Expected behavior
We have a special script developed to migrate old network to latest 1.0.0. We call it lisk bridge. We should use that script in combination with Jenkins to test the migration more periodically and automatically. 

### Actual behavior
Currently we are just starting 1.0.0 instance on top of previous database and considering it successful. 

### Steps to reproduce

https://jenkins.lisk.io/job/lisk-migration/

1.0.0

nazarhussain:
### What was the problem?
Current Jenkins.migration job is not processing the migration in the correct way. Its just terminating one branch and starting the app with a different branch. 

### How did I fix it?
Update the jenkins job to test new migration process with `lisk_bridge.sh`

### How to test it?
Run this job https://jenkins.lisk.io/job/lisk-migration-test/

### Review checklist

* The PR solves #1861
* Linting passes
* Tests pass

### DIFF
--- a/Jenkinsfile.migration
+++ b/Jenkinsfile.migration
@@ -22,21 +22,35 @@ def generateSrcCodeRelease(directory) {
 	}
 	return readFile("/tmp/${directory}-version").trim()
 }
+
+def getBuildBranch(liskDirectory) {
+	if(fileExists("${env.WORKSPACE}/${liskDirectory}/genesisBlock.json")) {
+		return "master"
+	} else {
+		return "1.0.0"
+	}
+}
 pipeline {
 	agent { node { label "node-07" } }
 	stages {
 		stage("Prepare workspace") {
 			steps {
 				deleteDir()
-				dir("lisk-build") {
-					git url: "https://github.com/LiskHQ/lisk-build.git", branch: "master"
-					sh "bash toolchains/install_linux.sh"
-				}
 				dir(params.MIGRATE_FROM) {
-					git url: "https://github.com/LiskHQ/lisk.git", branch: params.MIGRATE_TO
+					git url: "https://github.com/LiskHQ/lisk.git", branch: params.MIGRATE_FROM
 				}
 				dir(params.MIGRATE_TO) {
 					git url: "https://github.com/LiskHQ/lisk.git", branch: params.MIGRATE_TO
+				}
+				script {
+					env.MIGRATE_FROM_BUILD_BRANCH = getBuildBranch(params.MIGRATE_FROM)
+					env.MIGRATE_TO_BUILD_BRANCH = getBuildBranch(params.MIGRATE_TO)
+				}
+				dir("lisk-build-${env.MIGRATE_FROM_BUILD_BRANCH}") {
+					git url: "https://github.com/LiskHQ/lisk-build.git", branch: "${env.MIGRATE_FROM_BUILD_BRANCH}"
+				}
+				dir("lisk-build-${env.MIGRATE_TO_BUILD_BRANCH}") {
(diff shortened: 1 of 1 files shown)


nazarhussain:
This PR is suspended because of these issues. 

https://github.com/LiskHQ/lisk-scripts/issues/38
https://github.com/LiskHQ/lisk-scripts/issues/46

nazarhussain:
@MaciejBaj @karmacoma We stuck here with another roadblocker. The `network` flag the `lisk-build` does not actually have any impact. What ever network you select during build have no impact, the release is compatible only with the network for the git branch its based on. 

So we can't test the migration between a master branch to any other branch which is not compatible with main net. I am thinking on any other way for it. But with that lot of trouble it seems nice to test the migration manually for now.

Jenkinsfile.migration
```diff
@@ -11,65 +11,120 @@
...
+}
+
+def waitForBlockHeight(targetBlockHeight, waitAttempts=0, previousHeight=0) {
+	def height = getBlockHeight()
+	echo "Height: ${height}"
+
+	if(previousHeight != height) {
+		previousHeight = height
+		waitAttempts = 0
+	} else if (waitAttempts < 3) {
```
MaciejBaj: I can expect this part might be really fragile. 

The safer solution would be to check the height reached the target. Throw the error if it's already too high, but ignore potential downtimes which might occur.

Jenkinsfile.migration
```diff
@@ -11,65 +11,120 @@
...
 
-				dropdb lisk_test || true
-				createdb lisk_test
-				cd "${params.MIGRATE_FROM}"
-				json_files_copy "${params.MIGRATE_FROM}"
-				npm install
-				cd "../${params.MIGRATE_TO}"
-				json_files_copy "${params.MIGRATE_TO}"
-				npm install
-				sudo service postgresql restart
```
MaciejBaj: Where and when is DB migration happening now?

Jenkinsfile.migration
```diff
@@ -11,65 +11,120 @@
...
+					"""
+				}
 			}
 		}
-		stage("Start new Lisk") {
+
+		stage("Verify migration") {
 			steps {
+				script {
+					waitForBlockHeight(env.TARGET_BLOCK_HEIGHT.toInteger() + 5)
```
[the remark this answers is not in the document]
nazarhussain: No reason, just want to see if blockchain is moving forward.

Jenkinsfile.migration
nazarhussain: I can't find any way for `timeout` in Jenkins, so added this alternative. If blockchain is stuck and not moving forward, then only check for target height will never pass. So this code will check every 10s for 3 times if the block chain is moving forward, and if not then it will fail the job. I don't see its fragile.

Jenkinsfile.migration
nazarhussain: Inside the core itself. When we start the node with `bash lisk.sh start` it will start the node and the DB migration will happen. In current structure `lisk_bridge.sh` triggers the `installLisk.sh upgrade` and that triggers the `lisk.sh start`

MaciejBaj:
@nazarhussain @Nazgolze please run the Jenkins job for both master -> development and testnet-master -> development migrations.

nazarhussain:
@Nazgolze main net migration passed successfully. https://jenkins.lisk.io/job/lisk-migration-test/155/console

After we merge the PR, we have to setup some jobs to run this Jenkins pipeline nightly with specific parameters as per above mentioned link.

PR #1950, opened 2018-04-26
Set api get forges and votes limit to 101 - Closes #1800

### LINKED ISSUE #1800: Can't get all 101 votes making just one call
### Expected behavior
I want to get the list of votes using `/votes?address=:address&limit=101`
As each address can have up to 101 votes, I reckon it makes sense to allow to get all 101 votes at once instead of limiting to 100.

### Actual behavior
It currently return an error 

> "Value 101 is greater than maximum 100"

Meaning I need to make two different calls in order to get all 101 votes.

### Steps to reproduce
Call `/votes?address=:address&limit=101` where :address is any valid address

Version 1.0.0-alpha.3

diego-G:
It provokes a bump in the API versioning to `1.0.26`

### Review checklist

* The PR solves #1800
* Linting passes
* Tests pass

### DIFF
--- a/schema/swagger.yml
+++ b/schema/swagger.yml
@@ -16,7 +16,7 @@ swagger: '2.0'
 info:
   description: Lisk API Documentation
   title: Lisk API Documentation
-  version: '1.0.25'
+  version: '1.0.26'
   contact:
     email: admin@lisk.io
   license:
--- a/schema/swagger.yml
+++ b/schema/swagger.yml
@@ -582,8 +582,15 @@ paths:
       produces:
         - application/json
       parameters:
-        - $ref: '#/parameters/limit'
         - $ref: '#/parameters/offset'
+        - name: limit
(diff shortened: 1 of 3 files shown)


schema/swagger.yml
```diff
@@ -16,7 +16,7 @@ swagger: '2.0'
 info:
   description: Lisk API Documentation
   title: Lisk API Documentation
-  version: '1.0.25'
+  version: '1.0.26'
```
nazarhussain: @diego-G Just to highlight there is another open PR pointing to exact same change. 

https://github.com/LiskHQ/lisk/pull/1940/files#diff-053c90a95ed06515a37c330053ee1a63R19

So when that will be merged you will have conflicts. I suggest to merge both changes under `1.0.26` and publish it to swagger hub. 

Everything else LGTM.

PR #1968, opened 2018-04-30
Nightly builds - Closes #1936

### LINKED ISSUE #1936: Nightly builds
### Expected behavior

We need a Jenkins job to do nightly builds and upload them, so we can do nighlty tests.
They will be released following the naming pattern `<branch name>-<YYYYMMDD>` and deployed to devnetX to run the tests.

For this purpose Jenkinsfile.nightly-build will be added.

### Actual behavior

Jenkinsfile.nightly is present but it only includes basic tests.

### Steps to reproduce

Link TBD

1.0.0

abelboldu:
### What was the problem?

Jenkins job for nightly builds is needed.

### How did I fix it?

Jenkinsfile to do a build and push it to DigitalOcean spaces, every day.

### Review checklist

* The PR solves #1936 

NOTE: Jenkinsfile assumes the branch is aligned with lisk-build project.

### DIFF
--- a/Jenkinsfile.nightly-build
+++ b/Jenkinsfile.nightly-build
@@ -21,60 +21,60 @@ def date = new Date()
 def timestamp = dateFormat.format(date)
 
 pipeline {
-	agent { 
-		node { 
-			label 'nightly-build'
-			// Forcing a short path workspace to prevent issues with arguments lentgh
-			customWorkspace "workspace/${URLDecoder.decode(JOB_NAME)}/${BUILD_NUMBER}"
-			}
-	}
-	stages {
-		stage('Generate release') {
-			steps {
-	    		sh """
-	    		npm install
-	    		node ./node_modules/.bin/grunt release
-	    		jq -r '.version' config.json > version
-	    		"""
-				script {
-					env.LISK_VERSION = readFile("version").trim()
-				}
-			}
-		}
-		stage('Prepare build environment') {
-			steps {
-				dir('lisk-build') {
-					git url: "https://github.com/LiskHQ/lisk-build.git", branch: "1.0.0"
-				}
-				sh "cp release/lisk-${env.LISK_VERSION}.tgz lisk-build/src"
-			}
-		}
-		stage('Build release') {
-			steps {
-				dir('lisk-build') {
-					sh """
-					bash build.sh -n dev -v ${env.LISK_VERSION}
-					mv release/lisk-${env.LISK_VERSION}-Linux-x86_64.tar.gz release/lisk-${env.LISK_VERSION}-${timestamp}-Linux-x86_64.tar.gz
-					sha256sum release/lisk-${env.LISK_VERSION}-${timestamp}-Linux-x86_64.tar.gz > release/lisk-${env.LISK_VERSION}-${timestamp}-Linux-x86_64.tar.gz.SHA256
-					cat release/lisk-${env.LISK_VERSION}-${timestamp}-Linux-x86_64.tar.gz.SHA256
-					"""
-				}
-			}
-		}
-		stage('Upload nightly release.') {
-			steps {
-				dir('lisk-build') {
-					sh """
-					s3cmd put release/lisk-${env.LISK_VERSION}-${timestamp}-Linux-x86_64.tar.gz s3:///lisk/nightly/lisk/lisk-${env.LISK_VERSION}-${timestamp}-Linux-x86_64.tar.gz
-					s3cmd put release/lisk-${env.LISK_VERSION}-${timestamp}-Linux-x86_64.tar.gz.SHA256 s3:///lisk/nightly/lisk/lisk-${env.LISK_VERSION}-${timestamp}-Linux-x86_64.tar.gz.SHA256
-					"""
-				}
-			}
-		}
-	}
-	post {
-		always {
-			cleanWs()
-		}
-	}
+    agent { 
+        node { 
+            label 'nightly-build'
+            // Forcing a short path workspace to prevent issues with arguments lentgh
+            customWorkspace "workspace/${URLDecoder.decode(JOB_NAME)}/${BUILD_NUMBER}"
+        }
+    }
+    stages {
+        stage('Generate release') {
+            steps {
+                sh """
+                npm install
+                node ./node_modules/.bin/grunt release
+                jq -r '.version' config.json > version
+                """
+                script {
+                    env.LISK_VERSION = readFile("version").trim()
+                }
+            }
+        }
+        stage('Prepare build environment') {
+            steps {
+                dir('lisk-build') {
+                    git url: "https://github.com/LiskHQ/lisk-build.git", branch: "1.0.0"
+                }
+                sh "cp release/lisk-${env.LISK_VERSION}.tgz lisk-build/src"
+            }
+        }
+        stage('Build release') {
+            steps {
+                dir('lisk-build') {
+                    sh """
+                    bash build.sh -n dev -v ${env.LISK_VERSION}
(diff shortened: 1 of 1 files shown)


Jenkinsfile.nightly-build
```diff
@@ -0,0 +1,84 @@
...
+def date = new Date()
+def timestamp = dateFormat.format(date)
+
+def generateSrcCodeRelease() {
+		sh """
+		npm install
+		node ./node_modules/.bin/grunt release
+		jq -r '.version' config.json > "/tmp/version"
+		"""
+	return readFile("/tmp/version").trim()
```
fchavant: Unsafe: if we really have to use a file, let's have it leave in `$WORKSPACE`

Jenkinsfile.nightly-build
```diff
@@ -0,0 +1,84 @@
...
+}
+
+pipeline {
+	agent { 
+		node { 
+			label 'node-07'
```
fchavant: Let's use a more descriptive label.

Jenkinsfile.nightly-build
```diff
@@ -0,0 +1,84 @@
...
+			customWorkspace "workspace/${URLDecoder.decode(JOB_NAME)}/${BUILD_NUMBER}"
```
fchavant: There should be a comment here explaining why this is done.

Jenkinsfile.nightly-build
```diff
@@ -0,0 +1,84 @@
...
+				}
+			}
+		}
+		stage('Prepare build environment') {
+			steps {
+				dir('lisk-build') {
+					git url: "https://github.com/LiskHQ/lisk-build.git", branch: "${env.BRANCH_NAME}"
+				}
+				sh """
+				cp release/lisk-${env.LISK_VERSION}.tgz lisk-build/src
```
fchavant: One-liner, no need to use the multiline format.

Jenkinsfile.nightly-build
```diff
@@ -0,0 +1,84 @@
...
+pipeline {
+	agent { 
+		node { 
+			label 'node-07'
+			customWorkspace "workspace/${URLDecoder.decode(JOB_NAME)}/${BUILD_NUMBER}"
+			}
+	}
+	stages {
+		stage('Generate release') {
+			steps {
```
fchavant: We should have the build steps (the whole `sh` step) here to make things clearer.

Jenkinsfile.nightly-build
```diff
@@ -0,0 +1,84 @@
...
+				"""
+			}
+		}
+		stage('Build release') {
+			steps {
+				dir('lisk-build') {
+					sh """
+					bash build.sh -n dev -v ${env.LISK_VERSION}
+					mv release/lisk-${env.LISK_VERSION}-Linux-x86_64.tar.gz release/lisk-${env.LISK_VERSION}-${timestamp}-Linux-x86_64.tar.gz
+					sha256sum release/lisk-${env.LISK_VERSION}-${timestamp}-Linux-x86_64.tar.gz > release/lisk-${env.LISK_VERSION}-${timestamp}-Linux-x86_64.tar.gz.SHA256
```
fchavant: Wouldn't make sense to `cat` the checksum for reference?

Jenkinsfile.nightly-build
```diff
@@ -0,0 +1,80 @@
...
+def date = new Date()
+def timestamp = dateFormat.format(date)
+
+pipeline {
+	agent { 
+		node { 
+			label 'nightly-build'
+			// Forcing a short path workspace to prevent issues with arguments lentgh
+			customWorkspace "workspace/${URLDecoder.decode(JOB_NAME)}/${BUILD_NUMBER}"
+			}
```
fchavant: The identation looks wrong here.

Jenkinsfile.nightly-build
```diff
@@ -0,0 +1,80 @@
...
+                    env.LISK_VERSION = readFile("version").trim()
+                }
+            }
+        }
+        stage('Prepare build environment') {
+            steps {
+                dir('lisk-build') {
+                    git url: "https://github.com/LiskHQ/lisk-build.git", branch: "1.0.0"
+                }
+                sh "cp release/lisk-${env.LISK_VERSION}.tgz lisk-build/src"
```
nazarhussain: This code blocks assumes that you are inside `lisk` core directory so you can copy release from a relative path `release/` but if that's true then cloning `lisk-build` inside core repo is not a good idea. Clone it outside main repo.

Jenkinsfile.nightly-build
```diff
@@ -0,0 +1,80 @@
...
+            }
+        }
+        stage('Build release') {
+            steps {
+                dir('lisk-build') {
+                    sh """
+                    bash build.sh -n dev -v ${env.LISK_VERSION}
+                    mv release/lisk-${env.LISK_VERSION}-Linux-x86_64.tar.gz release/lisk-${env.LISK_VERSION}-${timestamp}-Linux-x86_64.tar.gz
+                    sha256sum release/lisk-${env.LISK_VERSION}-${timestamp}-Linux-x86_64.tar.gz > release/lisk-${env.LISK_VERSION}-${timestamp}-Linux-x86_64.tar.gz.SHA256
```
nazarhussain: This is already been don in `lisk-build` so no need it here in jenkins. 

https://github.com/LiskHQ/lisk-build/blob/4efa4655639eee36aa8e8301671289844e385940/build.sh#L227-L229

Jenkinsfile.nightly-build
```diff
@@ -0,0 +1,80 @@
...
+                }
+                sh "cp release/lisk-${env.LISK_VERSION}.tgz lisk-build/src"
+            }
+        }
+        stage('Build release') {
+            steps {
+                dir('lisk-build') {
+                    sh """
+                    bash build.sh -n dev -v ${env.LISK_VERSION}
+                    mv release/lisk-${env.LISK_VERSION}-Linux-x86_64.tar.gz release/lisk-${env.LISK_VERSION}-${timestamp}-Linux-x86_64.tar.gz
```
nazarhussain: @Nazgolze Shouldn't we add this time stamp feature inside `lisk-build`?

Jenkinsfile.nightly-build
```diff
@@ -0,0 +1,80 @@
...
+                    sha256sum release/lisk-${env.LISK_VERSION}-${timestamp}-Linux-x86_64.tar.gz > release/lisk-${env.LISK_VERSION}-${timestamp}-Linux-x86_64.tar.gz.SHA256
+                    cat release/lisk-${env.LISK_VERSION}-${timestamp}-Linux-x86_64.tar.gz.SHA256
```
nazarhussain: What this `cat` is doing at this step?

Jenkinsfile.nightly-build
abelboldu: This part appends the timestamp, which is not the case in `lisk-build`

nazarhussain: That's what I am suggesting to add this to `lisk-build` to be used everywhere.

abelboldu:
> It seems to contain a lot of code that I am also having in Jenkins.migration. We need to finalize a way to move shared code to a central file for Jenkins and to reuse the same code. May be this would be helpful for us.

@nazarhussain  I agree, but it may need a shared Jenkinsfile or another repo shared across projects, so can we just keep it for now?

Jenkinsfile.nightly-build
abelboldu: @Nazgolze Is this something we could add to `lisk-build` ?

nazarhussain:
@fchavant No its not just few lines, see this commit for details https://github.com/LiskHQ/lisk/pull/1938/commits/41a94b24cc2d03273311e85bb3dbe3952293608c

@abelboldu We created ` lisk-jenkins` repo to contain shared code of different jenkins code. Please see above mentioned commit to see how to use it. 

Please update the PR accordingly, so we can merge soon.

PR #1971, opened 2018-05-02
Inconsistent response for /transactions endpoint - Closes #1884

### LINKED ISSUE #1884: Inconsistent response for /transactions endpoint
### Expected behavior
Given I make the following calls:

`/transactions?blockId=:blockId&sort=timestamp:desc&offset=0&limit=51` and
`/transactions?blockId=:blockId&sort=timestamp:desc&offset=50&limit=51`

I expect the last item returned by the first call to be the same as the first item returned by the second call.

### Actual behavior
It is not.

### Steps to reproduce
Using the following [test database](https://downloads.lisk.io/lisk-explorer/dev/blockchain_explorer.db.gz):

Compare the [first call](http://localhost:4000/api/transactions?blockId=6524861224470851795&sort=timestamp:desc&offset=0&limit=51) with the [second call](http://localhost:4000/api/transactions?blockId=6524861224470851795&sort=timestamp:desc&offset=50&limit=51)

v1.0.0-beta.5

Tschakki:
### What was the problem?
In case the first sorting param for transaction is not precise enough, pagination of the transactions endpoint was not consistent.

### How did I fix it?
Added a second sorting param, which determines the  order in that the transactions get returned, if the first sorting param is not precise enough.

### How to test it?
the last entry of this call:
`http://localhost:4000/api/transactions?blockId=6524861224470851795&sort=timestamp:desc&offset=0&limit=51`
and the first entry of this call:
`http://localhost:4000/api/transactions?blockId=6524861224470851795&sort=timestamp:desc&offset=50&limit=51`
should be always identical.

### Review checklist

* The PR solves #1884
* Linting passes
* Tests pass


Tschakki:
Why is `t_id` not the right candidate? It should be unique for each transaction, because it is based on the unique signature. Do you have a different suggestion?

And I think none of our sort params are deterministic, atm we allow `amount`, `fee`, `type` and `timestamp`, which is OK because we have use cases for it, but we need a second sorting param to make the response and its pagination consistent.

nazarhussain:
@Tschakki As explained above, first look into our existing functional tests and see why this was not identified there before. As we have tests for `limit`, `sort` and ` offset` and these are passing by just using one sort param. If user manually identify this an issue then our tests should also behave same, otherwise this could be a reason of wrong data in the reported issue. 


Yes `t_id` is unique, but not every unique attribute is a good for default sorting. Usually people expect the default sorting order, the order in which data inserted into database. So why this can cause a wrong expectation from user. Instead we have `rowId` column in `trs` table which is unique as well as follow data insertion order. To me `trs.rowId` is better candidate for this purpose. Rest @MaciejBaj can decide on it.

Tschakki:
Thanks @lsilvs !
I added the test case now, to validate that the usage of `offset` param in transactions endpoint does not paginate nicely, if an unprecise sort param is provided.

@MaciejBaj We already have a default sorting param `amount:asc`, which is specified in `swagger.yml`.
But for the case that one specifies a sorting parameter that is not precise enough, it wont use the default sorting parameter additionally.
So this is why we need a second sorting param, that is unique to every transaction.

We can use `rowId`, as I see that this is more fitting in this case.
But I guess we need to make a few more changes then to fix the issue, because `rowId` is not part of our transactions-schema, or API documentation.

db/repos/transactions/index.js
```diff
@@ -334,7 +334,9 @@ const Queries = {
 				? ` AND ${params.owner}`
 				: params.owner,
 			params.sortField
-				? `ORDER BY ${[params.sortField, params.sortMethod].join(' ')}`
+				? `ORDER BY ${[params.sortField, params.sortMethod].join(
+						' '
+					)}, "t_rowId" ASC`
```
nazarhussain: @Tschakki Just thought of a scenario, what if while calling `db.transactions.list()` someone don't pass the `params.sortField` then this query will be invalid. To make sure this always works, we have to set a default value for `params.sortField` and `params.sortField` in `db.transactions.list()`

nazarhussain: Discard this comment, as I noticed that `sortField` check is already in place. The indentation of `?` make me think there is not.

PR #1984, opened 2018-05-02
Network height sometimes off - Closes #1913

### LINKED ISSUE #1913: Network height sometimes off
### Expected behavior
`Network height` (in log and in /api/node/status) should be the same as `height` or at least pretty near if the node is in sync

### Actual behavior
`Network height` is mostly correct (the same as `height`) but every few seconds it switches to a different value, in my case 29902, when height is 28559 right now. I also saw 3xx once.

### Steps to reproduce
Watch `Network height` and `height` in the log and in /api/node/status for some time

1.0.0-beta6

ManuGowda:
### What was the problem?
Network height switches every few seconds to a different value.
### How did I fix it?
The group by `lodash` function was returning heights value as a string so the popular height was picked inconsistently.
### How to test it?
Run `mocha test/unit/modules/peers.js` to verify the `network height` among peers.
### Review checklist

* The PR solves #1913
* Linting passes
* Tests pass

### DIFF
--- a/test/fixtures/peers.js
+++ b/test/fixtures/peers.js
@@ -51,7 +51,7 @@ const Peer = stampit({
 	},
 	init({ broadhash, nonce, state }) {
 		this.dappid = null;
-		this.height = faker.random.number({ max: 9999, min: 1 });
+		this.height = _.sampleSize([50, 70, 90, 110], 1);
 		this.ip = faker.internet.ip();
 		this.os = faker.lorem.slug();
 		this.wsPort = `5${faker.random.number({ max: 999, min: 100 })}`;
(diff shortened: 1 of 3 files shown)


test/fixtures/peers.js
```diff
@@ -51,7 +51,7 @@ const Peer = stampit({
 	},
 	init({ broadhash, nonce, state }) {
 		this.dappid = null;
-		this.height = faker.random.number({ max: 9999, min: 1 });
+		this.height = _.sample([50, 70, 90, 110]);
```
SargeKhan: In our application, do we store height as a number or as a string?

MaciejBaj: swagger.yml:
```
      height:
        type: integer
        example: 123
        minimum: 1
```

ManuGowda: we keep height as an integer when we do the group by `height` using lodash it returns the string, so it's about changing that string to the integer. In the application, we always have the integer for both `height` and `networkHeight`.

MaciejBaj: For clarity, we should `parseInt` here then.

PR #1989, opened 2018-05-03
Rename encryptedSecret -> encryptedPassphrase - Closes #1965

### LINKED ISSUE #1965: Rename encryptedSecret -> encryptedPassphrase
### Expected behavior

We should be consistent when referring to passphrases.

### Actual behavior

In https://github.com/LiskHQ/lisk/pull/1829/files we use `passphrase` etc consistently except in a few places we use `encryptedSecret` instead of `encryptedPassphrase`.

### Steps to reproduce

n/a

1.0.0

4miners:
### What was the problem?
We should be consistent when referring to passphrases.
### How did I fix it?
Rename `encryptedSecret` to `encryptedPassphrase` and few other things for consistency.
### How to test it?
Run tests.
### Review checklist

* The PR solves #1965
* Linting passes
* Tests pass

### DIFF
--- a/helpers/z_schema.js
+++ b/helpers/z_schema.js
@@ -163,8 +163,8 @@ var liskFormats = {
 		// Explanation of regex structure:
 		// - 1 or more 'key=value' pairs delimited with '&'
 		// Examples:
-		// - encryptedPassphrase=abcd1234
-		// - encryptedPassphrase=abcd1234&iterations=10000&iv=ef012345
+		// - cipherText=abcd1234
+		// - cipherText=abcd1234&iterations=10000&iv=ef012345
 		// NOTE: Maximum lengths chosen here are arbitrary
 		const keyRegExp = /[a-zA-Z0-9]{2,15}/;
 		const valueRegExp = /[a-f0-9]{1,256}/;
--- a/helpers/z_schema.js
+++ b/helpers/z_schema.js
@@ -159,22 +159,22 @@ var liskFormats = {
 	 * @returns {boolean}
 	 * @todo Add description for the function, the params and the return value
 	 */
-	encryptedSecret(str) {
+	encryptedPassphrase(str) {
 		// Explanation of regex structure:
 		// - 1 or more 'key=value' pairs delimited with '&'
 		// Examples:
-		// - encryptedSecret=abcd1234
-		// - encryptedSecret=abcd1234&iterations=10000&iv=ef012345
+		// - encryptedPassphrase=abcd1234
+		// - encryptedPassphrase=abcd1234&iterations=10000&iv=ef012345
 		// NOTE: Maximum lengths chosen here are arbitrary
 		const keyRegExp = /[a-zA-Z0-9]{2,15}/;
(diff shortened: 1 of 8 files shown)


test/data/config.json
```diff
@@ -76,607 +76,607 @@
 		"defaultPassword": "elephant tree paris dragon chair galaxy",
 		"secret": [
```
willclarktech: Can we change this key? It's really confusing.

test/data/genesis_delegates.json
```diff
@@ -17,7 +17,7 @@
 				"9d3058175acab969f41ad9b86f7a2926c74258670fe56b37c429c01fca9f2f0f",
 			"username": "genesis_1",
 			"password": "elephant tree paris dragon chair galaxy",
```
willclarktech: Not directly a part of this issue, but we should be able to use a default password approach here too to avoid the redundancy of storing the default password a hundred times.

test/unit/modules/delegates.js
```diff
@@ -417,12 +417,12 @@ describe('delegates', () => {
 			});
 
 			it('should load secrets in encrypted format with the key', done => {
-				config.forging.secret = encryptedSecret;
+				config.forging.secret = encryptedPassphrase;
```
willclarktech: `encryptedPassphrase` is being used in this file to describe sometimes an encrypted passphrase, and sometimes a list of items with an encrypted passphrase property. Can we make a clearer distinction?

helpers/z_schema.js
```diff
@@ -159,22 +159,22 @@ var liskFormats = {
...
 		const valueRegExp = /[a-f0-9]{1,256}/;
 		const keyValueRegExp = new RegExp(
 			`${keyRegExp.source}=${valueRegExp.source}`
 		);
-		const encryptedSecretRegExp = new RegExp(
+		const encryptedPassphraseRegExp = new RegExp(
 			`^(${keyValueRegExp.source})(?:&(${keyValueRegExp.source})){0,10}$`
 		);
-		return encryptedSecretRegExp.test(str);
+		return encryptedPassphraseRegExp.test(str);
```
nazarhussain: The final regex here become,

```
^([a-zA-Z0-9]{2,15}=[a-f0-9]{1,256})(?:&([a-zA-Z0-9]{2,15}=[a-f0-9]{1,256})){0,10}$
```

I tried to match 
```
encryptedPassphrase=abcd1234&iterations=10000&iv=ef012345
var1=john&var2=robert&var3=carrell
```

And it didn't worked. Can you double check the reg expression.

willclarktech: Great catch. `{2,15}` was set relative to the length of `encryptedSecret`, but `encryptedPassphrase` is longer.

test/unit/modules/delegates.js
willclarktech: Follow up issue: https://github.com/LiskHQ/lisk/issues/1994

helpers/z_schema.js
4miners: `encryptedPassphrase` never appear as key, I updated function description.

Issue #1841, opened 2018-04-10
Simplify peer headers


MaciejBaj:
### Expected behavior

The query string passed to the socket should only contain fixed data that is relevant to the peer/socket.

### Actual behavior

Right now the peer node requires a lot of unnecessary params to be set such as `height`, `broadhash` and others - These params change every time the node reconnects and they don't actually help to identify the socket.

This means that:
- The protocol for connecting to a peer via WebSockets is more complicated since we have to lookup all these extra params.
- We cannot use SC's auto-reconnect feature because the query parameters keep changing so we need to keep throwing away old sockets in order to pass the latest query string.

### Steps to reproduce

Check the WebSocket requests; the HTTP upgrade request path is very long with lots of unnecessary query params.

### Which version(s) does this affect? (Environment, OS, etc...)

1.0.0

PR #1991, opened 2018-05-03
Simplify peer headers

jondubois:
### What was the problem?

- The WebSocket handshake query string contained dynamic fields such as 'broadhash' and 'height' which change over time and do not help to identify the connection.
- Because the broadhash and height keep changing every time a node tries to reconnect to a peer, the previous socket associated with that peer could not be identified for reuse.
- A regression was introduced in PR https://github.com/LiskHQ/lisk/pull/1909 which caused strange behaviour with rediscovering a previously discovered peer. 

### How did I fix it?

- Do not include broadhash and height in the query string.
- Loosened schema validation.
- Fixed the regression from PR https://github.com/LiskHQ/lisk/pull/1909

### How to test it?

- Run two nodes on localhost and check that they can connect to each other (including after restarts) and check that they can still sync blocks.
- Run the nodes on beta net.

### Review checklist

* The PR solves #1841
* Linting passes
* Tests pass

### DIFF
--- a/api/ws/rpc/connect.js
+++ b/api/ws/rpc/connect.js
@@ -42,6 +42,8 @@ const connect = (peer, logger) => {
 
 const connectSteps = {
 	addConnectionOptions: peer => {
+		const systemHeaders = System.getHeaders();
+
 		peer.connectionOptions = {
 			autoConnect: false, // Lazy connection establishment
 			autoReconnect: false,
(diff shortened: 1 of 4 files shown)


jondubois:
@MaciejBaj All the functional, unit and integration tests are passing on my local machine, not sure why they're failing on Jenkins.

SargeKhan:
I've fixed the issue with failing tests. Besides that, I am not very well-versed of the issues related with p2p layer, so @MaciejBaj can review the PR in detail.

PR #2001, opened 2018-05-04
Inconsistent peers data from /api/peers - Closes #1955

### LINKED ISSUE #1955: Inconsistent peers data from /api/peers
### Expected behavior
Data under `GET /api/peers` must be consistent and contains all required attributes. 

### Actual behavior
Found out that there is inconsistent data for peers under `GET /api/peers`. Some required attributes are not found there. 

```
{
    "meta": {
        "offset": 0,
        "limit": 10,
        "total": 10
    },
    "data": [
        {
            "ip": "127.0.0.1",
            "wsPort": 5003,
            "state": 2,
            "os": "darwin17.6.0",
            "version": "0.0.1",
            "broadhash": "e6c26c5dda5bf6f2f744cf0598fa7d82fba132ffb867b6ca9b2599cdcc1d1f39",
            "height": 7,
            "nonce": "k5PIxE39lMyrCesS"
        },
        {
            "ip": "127.0.0.1",
            "wsPort": 5007,
            "state": 2,
            "os": "darwin17.6.0",
            "version": "0.0.1",
            "broadhash": "e6c26c5dda5bf6f2f744cf0598fa7d82fba132ffb867b6ca9b2599cdcc1d1f39",
            "height": 7,
            "nonce": "l6q4LJt29Uj2QNzv",
            "httpPort": 4007
        },
        {
            "ip": "127.0.0.1",
            "wsPort": 5009,
            "state": 2,
            "os": "darwin17.6.0",
            "version": "0.0.1",
            "broadhash": "e6c26c5dda5bf6f2f744cf0598fa7d82fba132ffb867b6ca9b2599cdcc1d1f39",
            "height": 7,
            "nonce": "NqT8J8fEL2gjOjCR"
        },
        {
            "ip": "127.0.0.1",
            "wsPort": 35608,
            "state": 1,
            "os": "debian",
            "version": "0.0.1",
            "broadhash": "198f2b61a8eb95fbeed58b8216780b68f697f26b849acf00c8c93bb9b24f783d",
            "height": 1,
            "nonce": "xKuEybs16XIHFoHA",
            "httpPort": 35609
        },
```
Notice for some peers `httpPort` is missing. For some it exists. For the last peer its related to self connecitng peer issue https://github.com/LiskHQ/lisk/issues/1954 so not been addressed here. 

The possible reason would be the inconsistent behavior of adding peers from seeding, discovery, and handshake. 

### Steps to reproduce
1. Run the integration tests and when nodes are up 
2. Call `curl http://127.0.0.1:4000/api/peers` 
3. The list must contain 9 peers but you will find 10

1.0.0

ManuGowda:
### What was the problem?
`GET /api/peers` returns list of peers with missing `httpPort` in the response.
### How did I fix it?
When the peer is created for the first time using the peer seed list the peer is created with params `ip, wsPort` after handshake with the other peer the remaining peer properties are updated, since the `logic/peer/update` function checks the immutable properties before update the current peer, the `httpPort` is never updated if the peer was created from the seed list.
### How to test it?

1. Run the integration tests and when nodes are up
2. Call curl http://127.0.0.1:4000/api/peers
3. The response should have peer list with these properties `'ip','wsPort','state','os','version','broadhash','httpPort','height','nonce'`

### Review checklist

* The PR solves #1955
* Linting passes
* Tests pass

### DIFF
--- a/logic/peer.js
+++ b/logic/peer.js
@@ -114,17 +114,13 @@ class Peer {
 		// Accept only supported properties
 		_.each(this.properties, key => {
 			// Change value only when is defined
-			if (
-				peer[key] !== null &&
-				peer[key] !== undefined &&
-				!_.includes(this.required, key)
-			) {
+			if (peer[key] && !this.required.includes(key)) {
 				// Update optional httpPort and nonce
 				// for the first time when peer object
 				// has httpPort and nonce as undefined
-				if (!this[key] && _.includes(this.optional, key)) {
-					this[key] = peer[key];
-				} else if (!_.includes(this.optional, key)) {
+				const isOptional = this.optional.includes(key);
+				const isExists = this[key];
+				if (!isOptional || (!isExists && isOptional)) {
 					this[key] = peer[key];
 				}
 			}
--- a/logic/peer.js
+++ b/logic/peer.js
@@ -117,9 +117,16 @@ class Peer {
 			if (
 				peer[key] !== null &&
 				peer[key] !== undefined &&
-				!_.includes(this.immutable, key)
+				!_.includes(this.required, key)
 			) {
-				this[key] = peer[key];
+				// Update optional httpPort and nonce
+				// for the first time when peer object
+				// has httpPort and nonce as undefined
+				if (!this[key] && _.includes(this.optional, key)) {
+					this[key] = peer[key];
+				} else if (!_.includes(this.optional, key)) {
+					this[key] = peer[key];
+				}
 			}
 		});
 
@@ -175,7 +182,9 @@ Peer.prototype.properties = [
 	'httpPort',
 ];
 
-Peer.prototype.immutable = ['ip', 'wsPort', 'httpPort', 'string'];
+Peer.prototype.required = ['ip', 'wsPort', 'string'];
+
+Peer.prototype.optional = ['httpPort', 'nonce'];
 
 Peer.prototype.connectionProperties = ['rpc', 'socket', 'connectionOptions'];
 
--- a/test/integration/scenarios/network/peers.js
+++ b/test/integration/scenarios/network/peers.js
@@ -76,21 +76,49 @@ module.exports = function(params) {
 
 	describe('Peers', () => {
 		describe('mutual connections', () => {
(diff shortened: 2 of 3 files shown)


logic/peer.js
```diff
@@ -117,9 +117,16 @@ class Peer {
 			if (
 				peer[key] !== null &&
 				peer[key] !== undefined &&
-				!_.includes(this.immutable, key)
+				!_.includes(this.required, key)
```
nazarhussain: Please use `Array.includes` builtin function.

logic/peer.js
```diff
@@ -117,9 +117,16 @@ class Peer {
...
 			) {
-				this[key] = peer[key];
+				// Update optional httpPort and nonce
+				// for the first time when peer object
+				// has httpPort and nonce as undefined
+				if (!this[key] && _.includes(this.optional, key)) {
+					this[key] = peer[key];
+				} else if (!_.includes(this.optional, key)) {
+					this[key] = peer[key];
+				}
```
nazarhussain: Suggest to rewrite this logic in a way: 

```
const isOptional = this.optional.includes(key);
const isEmpty = !this[key];

if(!isOptional || (isOptional && isEmpty)) {
   his[key] = peer[key];
}
```
This will make the code more readable.

logic/peer.js
```diff
@@ -175,7 +182,9 @@ Peer.prototype.properties = [
 	'httpPort',
 ];
 
-Peer.prototype.immutable = ['ip', 'wsPort', 'httpPort', 'string'];
+Peer.prototype.required = ['ip', 'wsPort', 'string'];
+
+Peer.prototype.optional = ['httpPort', 'nonce'];
```
nazarhussain: Is it really required to make them a prototype property? If I understand right, this way both arrays will be associated/created for every instance of `Peer` object. 

Instead why not just a module level private member.

ManuGowda: @nazarhussain thats a good point, we can make them private members of the modules, since, we are accessing these properties through peer instance in tests we might need to keep them the way it is now.

PR #2013, opened 2018-05-07
If a peer object for a specific nonce already exists and has a socket, destroy the socket - Closes #2012

### LINKED ISSUE #2012: If a peer object for a specific nonce already exists and has a socket, destroy the socket
### Expected behavior

If two peer objects with the same nonce are added to the peer manager and the previous peer object's socket exists, then destroy the socket from the existing peer.

### Actual behavior

It's possible for two peer objects with the same nonce can get added to the peer manager and that causes the socket on the old peer object to leak.

### Steps to reproduce

Check nodes on betanet. 

Latest betanet.

jondubois:
### What was the problem?

It's possible that two different peer objects could be added to peer manager with the same nonce; in this case, it's possible that the previous peer object might still have an open socket which is no longer needed.

### How did I fix it?

When adding a new peer to the peerManager, if a peer already exists with that nonce, delete the socket object on the existing peer.

### How to test it?

Beta net.
This was tested on the latest beta net on a single node before and after the change (checking for outbound connections) and it appears to solve the issue with leaking active handles (outbound).

### Review checklist

* The PR solves #2012 
* Linting passes
* Tests pass

### DIFF
--- a/helpers/peers_manager.js
+++ b/helpers/peers_manager.js
@@ -51,6 +51,12 @@ PeersManager.prototype.add = function(peer) {
 		return false;
 	}
 
+	const existingPeer = this.peers[peer.string];
+
+	if (existingPeer && existingPeer.socket) {
+		existingPeer.socket.destroy();
+		delete existingPeer.socket;
+	}
 	this.peers[peer.string] = peer;
 
 	if (peer.socket && peer.socket.active) {


jondubois:
This PR fixes the socket leaks but it introduces a new major issue which causes the node to disconnect from peers too often (on every peer update) which is incorrect behaviour (this causes major issues with syncing). We need a more thorough solution.

jondubois:
Closing this PR but we should keep the issue open.

PR #2016, opened 2018-05-08
Review API documentation - Closes #1691

### LINKED ISSUE #1691: Review the API documentation and prepare the publishable format
We are in process of closure of `1.0.0`, so need to get ready to publish our API documentation to a publicly accessible location. For these: 

1. Need to review the documentation thoroughly to remove in inconsistency 
2. Add more conceptual details e.g. tag descriptions, endpoint descriptions 
3. Analyze & research the tool/process to convert specification to human readable format 
4. Develop mechanism to publish it to to docs.lisk.io automatically on release

Tschakki:
### What was the problem?
API documentation needs to be finalised.

### How did I fix it?
Updating `swagger.yml`, add missing information.
Add host variable to make swagger interactive.
Regrouped endpoints.
Swagger can be included in our official documentation later.

### Review checklist

* The PR solves #1691 
* Linting passes
* Tests pass


schema/swagger.yml
```diff
@@ -14,20 +14,24 @@
...
 
 # All paths relative to specified basePath
-# basePath: /v1
-
 basePath: /api
 
+# The host (name or ip) serving the API
+#host: 46.101.179.89:5000
+host: localhost:4000
+
```
nazarhussain: Don't make host and IP as static entry. First thing, port value can be different and configureable int he application, secondly instead of localhost one can restrict traffic only listen on local network IP.

schema/swagger.yml
```diff
@@ -42,12 +46,12 @@ tags:
...
-  description: Queue related API calls
 - name: Signatures
   description: Signatures related API calls
 - name: Transactions
   description: Transactions related API calls
+- name: Voters
+  description: Votes related API calls
 - name: Votes
   description: Votes related API calls
 - name: Private
```
nazarhussain: Suggest to change this `Private` to `Forging` or something related. Because this tag applies to endpoint restrict with `config.forging.access.whiteList`

schema/swagger.yml
```diff
@@ -66,9 +70,9 @@ paths:
       - Accounts
       summary: Requests account data
       operationId: getAccounts
-      description:
-        By passing in the appropriate parameters, you can search for
-        matching accounts in the system
+      description: |
+        By passing the appropriate parameters, you can search for matching accounts in the system.
+        Otherwise returns a list of accounts.
```
nazarhussain: I believe we need to rephrase such descriptions, e.g. `By passing the appropriate parameters`  is really not needed to mention. Its something obvious, instead we put more focus on adding more details to parameters documentation.

schema/swagger.yml
```diff
@@ -112,9 +116,9 @@ paths:
       - Accounts
       summary: Requests multisignature groups data
       operationId: getMultisignatureGroups
-      description:
-        By passing in the appropriate options, you can search for a
-        specified account in the system and get its multisignature groups
+      description: |
+        By passing an address of an account, you can search for the specified account in the system.
+        As response, you will get a list of the multisignature groups that this account is member of.
```
nazarhussain: This is an old documentation phrase `By passing an address of an account` when account  was specified as `address` parameter for the request. Instead focus on `multisignature groups `  description.

schema/swagger.yml
```diff
@@ -155,9 +159,9 @@ paths:
       - Accounts
       summary: Requests multisignature memberships data
       operationId: getMultisignatureMemberships
-      description:
-        By passing in the appropriate options, you can search for a
-        specified account in the system and get its multisignature memberships
+      description: |
+        By passing the address of a multisignature group, you can search for the specified account in the system.
+        As response, you will get all members of this particular multisignature group.
```
nazarhussain: Same here

schema/swagger.yml
```diff
@@ -2588,191 +2599,3 @@ definitions:
...
-      - block
-      - nonce
-    properties:
-      nonce:
-        type: string
-        example: sYHEDBKcScaAAAYg
-        minLength: 16
-        maxLength: 16
-      block:
-        $ref: '#/definitions/ReducedBlock'
```
nazarhussain: @Tschakki Why you removed this from specification? These schema's are necessary for the application.

Tschakki:
I think it looks kind of messed up, because I ordered the paths alphabetically,  by moving some of them around. Github doesn't seem to understand it's moved and not changed, bc the code is still too similar, it seems... I will try to make it clearer with changing it back to the old order, and redo this as a very last step.
Btw this PR is still in progress, I will request a review when ready :)

schema/swagger.yml
Tschakki: Alright, removed the host variable from `swagger.yml` for now.

schema/swagger.yml
Tschakki: I decided to remmove the `Private` tag completely, as the API calls are now grouped by their endpoints.

schema/swagger.yml
Tschakki: I simplified the descriptions now

schema/swagger.yml
```diff
@@ -14,18 +14,21 @@
...
 info:
-  description: Lisk API Documentation
   title: Lisk API Documentation
-  version: '1.0.26'
+  description: |
+    Welcome to Lisks API documentation!
+
+    Below you can see all possible API calls, ordered by endpoints.
+
+    By clicking the "Try it out"-button beside each API call, you can test out the API calls and their responses interactively.
```
nazarhussain: We need to update this section further: 

1. Since we don't have that featured enabled and planned for now, so we should remove this line. 
2. Add more details are here, which audience is target for this REST API
3. Small summary on what users can do with current available endpoints 
3. A clearly highlighted notice that current version of API have breaking changes with previous version. 

Out of many see these few examples 

http://api.population.io
https://api.carnivalmobile.com

How they explained the API documentation and provided a bit of detail information on usage of it. Like time formats used in API, some detail about authentication etc.

PR #2017, opened 2018-05-08
Add integration test to cover Peer disconnects - Closes #2002

### LINKED ISSUE #2002: Add integration test to cover Peer disconnects
### Expected behavior
Integration test (or functional test running at least 2 SocketCluster server instances on separate processes) will check the correct behaviour when Peer is disconnected. When Peer gets disconnected, its socket and entry in PeersManager should be removed. 
### Actual behavior
Scenario is missing.
1.0.0

ManuGowda:
### What was the problem?
Missing integration test for peer disconnects
### How did I fix it?
Added scenario for testing peer disconnects
### How to test it?
Run the integration
### Review checklist

* The PR solves #2002 
* Linting passes
* Tests pass


ManuGowda:
@MaciejBaj @jondubois I have modified the peer disconnect logic 
- To use peer(5000) from peer list to avoid an extra dummy client being created with dummy server, this way we always have peer from (5000-5009).
- After these changes I am still getting 220+ estabilshed connection to peer(5000-5009), I have also testes these chanes on `1.0.0-beta.7` branch as well and the results are the same.

PR #2026, opened 2018-05-12
Peers list should include some unmatched broadhash peers to broadcast to - Closes #2025

### LINKED ISSUE #2025: Peers list should include some unmatched broadhash peers to broadcast to
### Expected behavior

When broadcasting, we should broadcast not only to peers which match our own broadhash but also to peers which do not match our broadhash (because they might be behind).

### Actual behavior

Right now when doing a broadcast, a node will try to select 100 peers in total; if it finds fewer than 100 peers with a matching broadhash, then it will select remaining peers which do not match its broadhash.

If our node is able to find 100 peers which match its own broadhash, then it will not broadcast to any non-matching peers (after it selects 25 from those 100); this means that the remaining (non-matching) peers in the network may not receive the latest block from our node.

### Steps to reproduce

Run on latest betanet; some of the nodes fall behind the latest height because some blocks are not reaching them.

Betanet 7.4

jondubois:
### What was the problem?

When launching a network, there was a situation where a node would only broadcast to peers that had a matching broadhash and ignore the others which were slightly delayed; in the early stages of a network there is a short timeframe when the broadhash only exists for half of the nodes (since the broadhash only exits after the 5th block); this may be the cause of a fork in the early stages of a network (from the 6th block).

### How did I fix it?

Half of all peers selected to receive our broadcast will be ones that do not match our own broadhash.

### How to test it?

Betanet.

### Review checklist

* The PR solves #2025
* Linting passes
* Tests pass

### DIFF
--- a/logic/broadcaster.js
+++ b/logic/broadcaster.js
@@ -103,6 +103,9 @@ class Broadcaster {
 	 */
 	getPeers(params, cb) {
 		params.limit = params.limit || this.config.peerLimit;
+		// Leave some room in the broadcast list for peers with a
+		// non-matching broadhash in order to avoid forks.
+		params.matchingPeersRatio = constants.matchingPeersRatio;
 		params.broadhash = params.broadhash || null;
 		params.normalized = false;
 
--- a/modules/peers.js
+++ b/modules/peers.js
@@ -29,7 +29,7 @@ let self;
 const __private = {};
 let definitions;
 
-const peerDiscoveryFrequency = 30000;
+const peerDiscoveryFrequency = 10000;
 
 /**
  * Main peers methods. Initializes library with scope content.
--- a/modules/peers.js
+++ b/modules/peers.js
@@ -91,6 +91,29 @@ __private.countByFilter = function(filter, cb) {
 };
 
 /**
+ * Shuffles peers (using Fisher-Yates-Durstenfeld shuffle algorithm).
+ *
+ * @todo Add @param tags
+ * @todo Add @returns tag
+ * @todo Add description of the function
+ */
+__private.shuffle = function(array) {
+	let m = array.length;
+	let t;
+	let i;
+	// While there remain elements to shuffle
+	while (m) {
+		// Pick a remaining element
+		i = Math.floor(Math.random() * m--);
+		// And swap it with the current element
+		t = array[m];
+		array[m] = array[i];
+		array[i] = t;
+	}
+	return array;
+};
+
+/**
  * Gets randomly ordered list of peers by filter.
  *
  * @private
@@ -170,29 +193,6 @@ __private.getByFilter = function(filter, cb) {
 		};
 	};
 
-	/**
-	 * Shuffles peers (using Fisher-Yates-Durstenfeld shuffle algorithm).
-	 *
-	 * @todo Add @param tags
-	 * @todo Add @returns tag
-	 * @todo Add description of the function
-	 */
-	const shuffle = function(array) {
-		let m = array.length;
-		let t;
-		let i;
-		// While there remain elements to shuffle
-		while (m) {
-			// Pick a remaining element
-			i = Math.floor(Math.random() * m--);
-			// And swap it with the current element
-			t = array[m];
-			array[m] = array[i];
-			array[i] = t;
-		}
-		return array;
-	};
-
 	// Apply filters (by AND)
 	const normalized = filter.normalized === undefined ? true : filter.normalized;
 	let peers = library.logic.peers.list(normalized);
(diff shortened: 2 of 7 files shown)


logic/broadcaster.js
```diff
@@ -103,6 +103,9 @@ class Broadcaster {
 	 */
 	getPeers(params, cb) {
 		params.limit = params.limit || this.config.peerLimit;
+		// Leave some room in the broadcast list for peers with a
+		// non-matching broadhash in order to avoid forks.
```
nazarhussain: I believe stating here as non-matching broadhash is the cause of fork is not justified. Fork occurs because broadcast does not work and sync was triggered only on the set of nodes which were on low height in the network. 

As we discussed earlier, my thoughts are that we do broadcasting to only matching boradhash peers, as its a real time and costly event across the network. For syncing we should pull blocks from any peers with the highest height in the network irrespective of the broadhash.

modules/peers.js
```diff
@@ -29,7 +29,7 @@ let self;
 const __private = {};
 let definitions;
 
-const peerDiscoveryFrequency = 30000;
+const peerDiscoveryFrequency = 10000;
```
nazarhussain: Is there any thought process behind changing this value? Are you changing it for testing purposes on devnet or proposing that change for mainnet. 

When node startup, discovery should trigger and is triggering immediately afterwards 30s is a reasonable delay for this activity.

modules/peers.js
```diff
@@ -90,6 +90,29 @@ __private.countByFilter = function(filter, cb) {
...
+	while (m) {
+		// Pick a remaining element
+		i = Math.floor(Math.random() * m--);
+		// And swap it with the current element
+		t = array[m];
+		array[m] = array[i];
+		array[i] = t;
+	}
+	return array;
+};
```
nazarhussain: The ES6 equivalent can be
```
function shuffle(a) {
    for (let i = a.length - 1; i > 0; i--) {
        const j = Math.floor(Math.random() * (i + 1));
        [a[i], a[j]] = [a[j], a[i]];
    }
    return a;
}
```

We are using `lodash` in this module and it contains `shuffle` function with same Fisher-Yates algorithm, then why we need to create our new function.

modules/peers.js
```diff
@@ -644,7 +644,9 @@ Peers.prototype.acceptable = function(peers) {
  * @returns {setImmediateCallback} cb, err, peers
  */
 Peers.prototype.list = function(options, cb) {
+	const matchingPeersRatio = options.matchingPeersRatio || 1;
 	let limit = options.limit || constants.maxPeers;
+	const matchingPeersLimit = Math.ceil(limit * matchingPeersRatio);
 	const broadhash = options.broadhash || modules.system.getBroadhash();
 	const allowedStates = options.allowedStates || [Peer.STATE.CONNECTED];
 	const attempts =
```
nazarhussain: It would be nice to define constants for `ATTEMP_MATCHED` and `ATTEMPT_UNMATCHED`.

ManuGowda:
@jondubois the PR LGTM, however, it would be great if we can have tests which cover the scenario we are trying to fix.

modules/peers.js
shuse2: This was experimental.
Reason of experiment was
- It was 5s originally
- 30s for discovery is I think it was good, but we thought updating status of peers every 30s is not enough for our current logic

but now we found the reason of connect/disconnect problem, it might not be related.

modules/peers.js
shuse2: I thought about this too!
since we are using lodash, we should've probably just use shuffle.
However, I think the intention was not to change the original code.

4miners:
@shuse2 Idk about final solution, I see `What was the problem?` description in this PR and that description is invalid because there is no such problem.

jondubois:
@4miners the issue we observed appeared to be caused by blocks not being propagated to nodes that were not directly connected to the forging delegate and we thought it might be related to the broadhash info not being propagated.

It appears that removing the MAX_CALLS_ALLOWED limitation https://github.com/LiskHQ/wamp-socket-cluster/pull/36 fixed that issue.

I will make another branch from this one and revert the broadcast changes and see if the problem is still fixed.

modules/peers.js
jondubois: @nazarhussain @shuse2 Yes, this function was already there (I just moved it) and I didn't want to use a different algorithm and risk breaking something which works. This is a change we can do later.

jondubois:
I'm closing this PR (for now) in favour of https://github.com/LiskHQ/lisk/pull/2031 which doesn't have the controversial changes - We tested on the new beta net and it worked properly as well (no fork).

PR #2031, opened 2018-05-15
Block propagation failure/partition and node crash - Closes #2030

### LINKED ISSUE #2030: Block propagation failure/partition and node crash
### Expected behavior

Blocks should propagate across the network.
Also, if the socket is no longer attached to a peer object, then we should not call emit on that socket.

### Actual behavior

There was a fork in the latest betanet because some blocks were not reaching some nodes. We thought it was a broadhash broadcasting issue initially but it appears be related to this issue https://github.com/LiskHQ/wamp-socket-cluster/issues/35

### Steps to reproduce

Run on latest code on betanet with 250 nodes.

Beta 7.4

jondubois:
### What was the problem?

There was an issue with block propagation which was causing a network partition.

### How did I fix it?

The changes in this PR are built on top of https://github.com/LiskHQ/lisk/pull/2026 but I reverted the controversial changes related to broadcasting (the ones about always including some unmatched peers).

### How to test it?

This was tested on beta net 7.4 and has been successful so far :)

### Review checklist

* The PR solves #2030
* Linting passes
* Tests pass

### DIFF
--- a/modules/peers.js
+++ b/modules/peers.js
@@ -29,7 +29,7 @@ let self;
 const __private = {};
 let definitions;
 
-const peerDiscoveryFrequency = 30000;
+const peerDiscoveryFrequency = 10000;
 
 /**
  * Main peers methods. Initializes library with scope content.
--- a/modules/peers.js
+++ b/modules/peers.js
@@ -644,9 +644,7 @@ Peers.prototype.acceptable = function(peers) {
  * @returns {setImmediateCallback} cb, err, peers
  */
 Peers.prototype.list = function(options, cb) {
-	const matchingPeersRatio = options.matchingPeersRatio || 1;
 	let limit = options.limit || constants.maxPeers;
-	const matchingPeersLimit = Math.ceil(limit * matchingPeersRatio);
 	const broadhash = options.broadhash || modules.system.getBroadhash();
 	const allowedStates = options.allowedStates || [Peer.STATE.CONNECTED];
 	const attempts =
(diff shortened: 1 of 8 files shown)


modules/peers.js
```diff
@@ -29,7 +29,7 @@ let self;
 const __private = {};
 let definitions;
 
-const peerDiscoveryFrequency = 30000;
+const peerDiscoveryFrequency = 10000;
```
nazarhussain: So we discovered that its related to PRC max calls limit, then why we still decreasing this value. Twice a minute for a network discovery is a good value.

jondubois: Risk mitigation ;p but yeah it's definitely worth running another betanet test with 30s discovery.

jondubois: Note that it's not just about discovery though, it also fetches the latest status of peers based on that interval.

modules/peers.js
```diff
@@ -710,6 +710,7 @@ Peers.prototype.list = function(options, cb) {
 					// Unmatched broadhash
 					return randomList(peers, waterCb);
 				}
+				peers = __private.shuffle(peers);
```
4miners: Why we shuffle peers again here?

jondubois: @4miners Before, it would concatenate the shuffled list of matched broadhash peers with the shuffled list of unmatched peers - This means that the matched broadhash peers would always be at the top of the list and would always be in the top 20 when choosing which peers to broadcast to (because we use array slice). So it's like both halves of the final list were shuffled but the whole list wasn't actually shuffled.

It seems like the previous behaviour was unintentional but I can roll back if necessary.

In any case, we don't need both both. I didn't want to remove the first shuffle in the getByFilter function because it appears to be used in other parts of the code.

jondubois:
This PR was closed in favour of https://github.com/LiskHQ/lisk/pull/2043  - The new PR keeps some of the functionality from this PR but rolls back certain features and also adds new features on top.

Issue #1781, opened 2018-03-26
Inconsistency in mem_accounts after delete last block of round


4miners:
### Expected behavior
`mem_accounts` states after delete last block of round 1 should match states from snapshot taken before forge last block of round 1.
### Actual behavior
States not match:
- `virgin` column values for genesis delegates are set to `0` (previously `1`)
- `blockId` is equal to ID of deleted block - due to bug in refactoring queries/db-repos parameters are passed in wrong order, so we try to overwrite rows with block ID `0` with deleted block ID:
https://github.com/LiskHQ/lisk/blob/a027a995dfe7d58e3ed63688d81e47a8cd435d89/db/repos/rounds.js#L119-L121
https://github.com/LiskHQ/lisk/blob/a027a995dfe7d58e3ed63688d81e47a8cd435d89/logic/round.js#L172-L175
`0.9.12` - deleted block ID is overwritten with `0`:
https://github.com/LiskHQ/lisk/blob/7bf54a05815049f31931ffaba379b6d3b5febe79/sql/rounds.js#L20
https://github.com/LiskHQ/lisk/blob/7bf54a05815049f31931ffaba379b6d3b5febe79/logic/round.js#L124
### Steps to reproduce
Run functional system rounds tests.
### Which version(s) does this affect? (Environment, OS, etc...)
`1.0.0`, `0.9.12`

4miners:
Children are merged.

PR #2035, opened 2018-05-17
Application gets stuck when deleting last block of the same round twice - Closes #2032

### LINKED ISSUE #2032: Application gets stuck when deleting last block of the same round twice
### Expected behavior
Node should be able to delete the last block of the same round as many time as needed without corrupting snapshots
### Actual behavior
We perform snapshotting process at the second last block of the round here: https://github.com/LiskHQ/lisk/blob/development/modules/rounds.js#L228Is there any particular reason why we do this at the second last block? and not before at the last block of the round (before making any calculation for the next round?)
### Steps to reproduce:
```
describe('snapshotting when deleting last block of the round twice', () => {
	before(() => {
		// Forge 201 blocks to reach height 202 (genesis block is already tyhere)
		return Promise.mapSeries([...Array(201)], () => {
			return addTransactionsAndForgePromise(library, [], 0);
		}).then(() => {
			return deleteLastBlockPromise(library);
		}).then(() => {
			return addTransactionsAndForgePromise(library, [], 0);
		});
	});

	it('should be able to delete the block again', () => {
		return deleteLastBlockPromise(library);
	});
});
```
1.0.0

4miners:
### What was the problem?
Unable to delete last block of round twice in a row because round snapshot was not available.
### How did I fix it?
- Keep `mem_round_snapshot` and `mem_votes_snapshot` tables after round rollback, so those tables can be reused for same round.
- Check if there is snapshot available for current round before we restore it, to prevent usage of wrong snapshot (results with corrupted data) when going back more than one round of blocks.

It's not the best solution, but not so hard to implement and reliable. 
### How to test it?
Run test suite.
### Review checklist

* The PR solves #2032
* Linting passes
* Tests pass

### DIFF
--- a/db/sql/rounds/check_snapshot_availability.sql
+++ b/db/sql/rounds/check_snapshot_availability.sql
@@ -14,9 +14,22 @@
 
 
 /*
-  DESCRIPTION: Check round snapshot availability for a particular round.
+  DESCRIPTION: Check round snapshot availability for a particular round
+               Returns TRUE when snapshot available or table is empty, FALSE otherwise
 
   PARAMETERS: round - Round for we are checking availability
 */
 
-SELECT 1 AS available FROM mem_round_snapshot WHERE round = ${round} LIMIT 1
+SELECT (
+	CASE WHEN (
+		SELECT 1 AS available FROM mem_round_snapshot WHERE round = ${round} LIMIT 1
+	) = 1 THEN TRUE
+	ELSE (
+		CASE WHEN (
+			SELECT COUNT(*) FROM mem_round_snapshot
+		) = 0 THEN TRUE
+		ELSE FALSE
+		END
+	)
+	END
+) AS available
--- a/db/sql/rounds/check_snapshot_availability.sql
+++ b/db/sql/rounds/check_snapshot_availability.sql
@@ -1,0 +1,22 @@
+/*
+ * Copyright © 2018 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+
+/*
+  DESCRIPTION: Check round snapshot availability for a particular round.
+
+  PARAMETERS: round - Round for we are checking availability
+*/
+
+SELECT 1 FROM mem_round_snapshot WHERE round = ${round} LIMIT 1
--- a/db/sql/rounds/check_snapshot_availability.sql
+++ b/db/sql/rounds/check_snapshot_availability.sql
@@ -14,22 +14,9 @@
 
 
 /*
-  DESCRIPTION: Check round snapshot availability for a particular round
-               Returns TRUE when snapshot available or table is empty, FALSE otherwise
+  DESCRIPTION: Check round snapshot availability for a particular round.
 
   PARAMETERS: round - Round for we are checking availability
 */
 
-SELECT (
-	CASE WHEN (
-		SELECT 1 AS available FROM mem_round_snapshot WHERE round = ${round} LIMIT 1
-	) = 1 THEN TRUE
-	ELSE (
-		CASE WHEN (
-			SELECT COUNT(*) FROM mem_round_snapshot
-		) = 0 THEN TRUE
-		ELSE FALSE
-		END
-	)
-	END
-) AS available
+SELECT 1 AS available FROM mem_round_snapshot WHERE round = ${round} LIMIT 1
--- a/db/sql/rounds/check_snapshot_availability.sql
+++ b/db/sql/rounds/check_snapshot_availability.sql
@@ -19,4 +19,4 @@
   PARAMETERS: round - Round for we are checking availability
 */
 
-SELECT 1 FROM mem_round_snapshot WHERE round = ${round} LIMIT 1
+SELECT 1 AS available FROM mem_round_snapshot WHERE round = ${round} LIMIT 1
--- a/test/functional/system/rounds.js
+++ b/test/functional/system/rounds.js
@@ -1130,6 +1130,22 @@ describe('rounds', () => {
 		});
 	});
 
+	describe('deleting last block of round twice in a row', () => {
+		before(() => {
+			return deleteLastBlockPromise().then(() => {
+				return addTransactionsAndForgePromise(library, [], 0);
+			});
+		});
+
+		after(() => {
+			return addTransactionsAndForgePromise(library, [], 0);
+		});
+
+		it('should be able to delete last block of round again', () => {
+			return deleteLastBlockPromise();
+		});
+	});
+
 	describe('rollback more than 1 round of blocks', () => {
 		let lastBlock;
 
--- a/test/functional/system/rounds.js
+++ b/test/functional/system/rounds.js
@@ -1182,4 +1182,22 @@ describe('rounds', () => {
 			return expect(lastBlock.height).to.equal(101);
 		});
 	});
+
+	describe('deleting last block of round twice in a row - no transactions during round', () => {
+		before(() => {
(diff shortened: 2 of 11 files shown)


4miners:
Notice: This PR contains changes from https://github.com/LiskHQ/lisk/pull/2038 and https://github.com/LiskHQ/lisk/pull/2039, those should be merged prior to this one.

db/sql/rounds/check_snapshot_availability.sql
```diff
@@ -0,0 +1,35 @@
...
+  PARAMETERS: round - Round for we are checking availability
+*/
+
+SELECT (
+	CASE WHEN (
+		SELECT 1 AS available FROM mem_round_snapshot WHERE round = ${round} LIMIT 1
+	) = 1 THEN TRUE
+	ELSE (
+		CASE WHEN (
+			SELECT COUNT(*) FROM mem_round_snapshot
```
SargeKhan: This case of query is for when we rollback to first block?

SargeKhan: wouldn't it be nicer to just write to mem_round_snapshot and mem_votes_snapshot after first block?

test/functional/system/rounds.js
```diff
@@ -726,6 +726,33 @@ describe('rounds', () => {
...
+				return addTransactionsAndForgePromise(library, [], 0);
+			});
+
+			it('should be able to delete last block of round again', () => {
+				return deleteLastBlockPromise();
+			});
+
+			// FIXME: Unskip that test after https://github.com/LiskHQ/lisk/issues/1781 is closed
+			// eslint-disable-next-line
+			it.skip('mem_accounts table should be equal to one generated before last block of round deletion', () => {
```
SargeKhan: Issue #1781 is already closed.

test/functional/system/rounds.js
```diff
@@ -1155,4 +1182,22 @@ describe('rounds', () => {
 			return expect(lastBlock.height).to.equal(101);
 		});
 	});
+
+	describe('deleting last block of round twice in a row - no transactions during round', () => {
+		before(() => {
+			return Promise.mapSeries([...Array(202)], () => {
```
SargeKhan: This should be 201 since we already have genesisBlock in the database.

test/functional/system/rounds.js
4miners: On `beta.8` is not skipped, but in `beta.7` we have exactly same expectation that is still skipped.

db/sql/rounds/check_snapshot_availability.sql
SargeKhan: I think this query is overcomplicated. We are adding this case specifically for the first round. So, we can modify the function that calls this query to handle that edge case (for first round), IMO we shouldn't move this logic to the database layer.

test/functional/system/rounds.js
SargeKhan: :/. I don't like these tests. If in the future it fails, the person looking at the code will have to understand all the tests to see what is the state of the chain (how many blocks it contains) before he can see what the actual problem is. Besides, it makes our tests tightly coupled with each other, meaning even just changing the order will make tests fail.

db/sql/rounds/check_snapshot_availability.sql
4miners: It's not for the first round, it can happen in every round. There are 2 options:
- one query like this, that returns `true` or `false` if we can use snapshot
- 2 queries
-- one that give us info if snapshot table contains records for requested round
-- second that counts records in snapshot table
and perform condition in round logic:
```
	checkSnapshotAvailability() {
		return this.t.rounds
			.isRoundExistsInRoundSnapshot(this.scope.round)
			.then(isExists => {
				if (!isExists) {
					return this.t.rounds
						.countRoundSnapshot()
						.then(count => {
							if (count) {
								throw new Error(
									`Snapshot for round ${
										this.scope.round
										} not available, unable to perform round rollback`
								);
							}
						});
				}
			});
	}
```
For me first option is better because we have:
- less queries in db repo (don't need additional file + additional function + unit tests for it)
- less complicated logic and unit tests for checkSnapshotAvailability function

shuse2: As you said,
> It's not the best solution, but not so hard to implement and reliable.
this might not be final solution, so this might not be applicable,
- complicated SQL is harder to optimize the performance later
- being able / easy to test in application is better than harder to test on DB
- 1 complex query vs 2 simple query, I think 2 simple query would be better

test/functional/system/rounds.js
4miners: Yes, such a behavior is intended. Those tests are functional system, the longer chain is - the better reliability of those tests is.

db/sql/rounds/check_snapshot_availability.sql
4miners: - There are only 2 simple conditions in that query, is it really complicated?
- It's not hard to test that query on DB, tests for that are part of this PR
- Regarding performance - it should be faster than executing 2 queries, and I don't think that there is a room for further performance optimization later.

4miners: By `not the best solution` I meant there are some better that require a little brainstorming, for example:
- decouple `tick` to `beforeTick` and `afterTick` (so we can snapshot before last block of round, not after one block before last block of round)
- rewrite round 😉

test/functional/system/rounds.js
SargeKhan: The purpose of these tests is not to check how the application behaves for the longer chain, they are for checking what happens when we delete the last block of the round twice. But I am not asking to change these tests, just talking about the structure we follow sometime for writing tests, so we may have a consistent style of writing tests by different developers. :)

db/sql/rounds/check_snapshot_availability.sql
shuse2: I guess I'm just strongly against putting the logic in the database layer.
- 2 simple select can be optimize a lot easier than 1 sql with condition.
- it's always a lot harder to optimize database layer compare to application layer

Even though this 1 SQL would be performance wise slightly better, for consistency through out the code and future scalability, I think we should put it on application layer.

4miners: @shuse2 Ok, I will rewrite.

test/functional/system/rounds.js
4miners: @SargeKhan I can refactor those tests and decouple them to few separate files (file per scenario tested), if we all agree that they will be easier to maintain then. However after `beta.7` and `beta.8` get merged together, as those branches contain separate changes to those tests.

db/sql/rounds/check_snapshot_availability.sql
nazarhussain: You guys are thinking this query complicated? You had not seen others queries already. 

Complicated SQL is a relative term, completely depending on the nature of the application you are dealing with. If you are writing a simple e-commerce system, then yes any query more than 2/2 lines or with sub queries can be considered complicated. But if you are dealing with some analytical system, normally a query with 2-3 joins and sub queries considered simple ;) 

Depending on the nature of the application we are dealing, we have to gave some luxury to database layer to handle some complexity. Don't consider DB layer as abstraction to run one line queries.  Testing DB level complex queries is not difficult at all, we already have tests for more complicated db scenarios for this purpose.

shuse2: `SELECT COUNT(*) FROM mem_round_snapshot WHERE round = ?`
can't you just do this and handle the error properly?

test/functional/system/rounds.js
MaciejBaj: For simplicity of merging let's leave them in the same file for now.

db/sql/rounds/check_snapshot_availability.sql
MaciejBaj: I would suggest split the query to keep the logic in JS code where possible.

4miners: No, it will not work, we need to allow restoring snapshot in 2 cases:
- when we have some records with requested round (means we have data for that round, and can use it)
- when we don't have any records for requested round and table is empty (table is empty only when there was no transactions during entire round)

4miners:
@nazarhussain Done.

PR #2039, opened 2018-05-17
Undo on transactions called with wrong block - Closes #2037

### LINKED ISSUE #2037: Undo on transactions called with wrong block
### Expected behavior
Undo on transactions should be called with block which those transactions come from.
### Actual behavior
It's called with wrong block. It results with incorrect data written to database, for example changes can be credited to wrong round.
### Steps to reproduce
- forge block 101, round finished, `mem_round` empty
- forge block 102 with transaction
- delete block 102, `mem_round` contains changes for `round` 1, which is incorrect
1.0.0

4miners:
### What was the problem?
Undo on transactions should be called with block which those transactions come from.
### How did I fix it?
Pass proper block to undoStep from popLastBlock.
### How to test it?
There are no complete unit tests for that function. However described scenario is covered in functional system rounds tests as part of https://github.com/LiskHQ/lisk/pull/2035.
### Review checklist

* The PR solves #2037
* Linting passes
* Tests pass

### DIFF
--- a/modules/blocks/chain.js
+++ b/modules/blocks/chain.js
@@ -704,7 +704,7 @@ __private.popLastBlock = function(oldLastBlock, cb) {
 						oldLastBlock.transactions.reverse(),
 						transaction =>
 							__private
-								.undoStep(transaction, secondLastBlock, tx)
+								.undoStep(transaction, oldLastBlock, tx)
 								.then(() => __private.undoUnconfirmStep(transaction, tx))
 					);
 				})


modules/blocks/chain.js
```diff
@@ -704,7 +704,7 @@ __private.popLastBlock = function(oldLastBlock, cb) {
 						oldLastBlock.transactions.reverse(),
 						transaction =>
 							__private
-								.undoStep(transaction, secondLastBlock, tx)
+								.undoStep(transaction, oldLastBlock, tx)
```
nazarhussain: If changes can be credited to wrong round, then its a critical thing. How does this get through from round tests? should not we include some test cases for it?

4miners: @nazarhussain See description of this PR, I added some tests in https://github.com/LiskHQ/lisk/pull/2035, I can add more when I'll be refactoring functional system rounds tests.

PR #2043, opened 2018-05-17
Broadcast, FindGoodPeer and Broadcast header does not select intended peers - Closes #2042

### LINKED ISSUE #2042: Broadcast, FindGoodPeer and Broadcast header does not select intended peers
### Expected behavior
For `broadcast`, `findgoodpeer`, and `broadcastHeader`, it should send a information to the nodes who needs. (I think this is the all cases where broadhash is used)
We should not filter the candidate peers of those function by `modules.peers.list` but use randomly selected peers instead.

### Actual behavior
In current 1.0.0 protocol, it is more likely to have the same broadhash than the previous http protocol since nodes communicates more in real time.

Therefore, if there are 100 nodes that are on the same broadhash, it will broadcast the received information only to those nodes that already have the information (because they are on the same broadhash).

Also, if a node is in fork with 100 nodes network, there is no way to recover since when `sync` happens, `findGoodPeers` is called, but it will filter by the broadhash first. If there is enough member in the forked network, there is very low chance that node will recover from the fork.

1.0.0

jondubois:
### What was the problem?

There is a potential case where a node could be neither synching nor receiving broadcasts from its highest peers (e.g if the node's broadhash is only one or two blocks behind their highest peers; not enough to trigger sync); during this time, if the node receives a block from a delegate (whose broadhash is also the same number of blocks behind) then it's possible that they (the node, the delegate and other nodes in the same situation) could fork.

### How did I fix it?

This PR moves further in the direction of broadcasting to a random sample of peers instead of only selecting those that have a matching broadhash - The last beta stress test using this approach has been successful.

### How to test it?

Betanet.

### Review checklist

* The PR solves #2042
* Linting passes
* Tests pass

### DIFF
--- a/api/ws/rpc/connect.js
+++ b/api/ws/rpc/connect.js
@@ -168,9 +168,14 @@ const connectSteps = {
 					if (peer.socket) {
 						peer.socket.emit(eventProcedureName, data);
 					} else {
-						const eventNotExistError =
-							'Tried to emit event on outbound peer socket which no longer exists';
+						const eventNotExistError = `Tried to emit event on outbound peer socket '${
+							peerExtendedWithPublish.string
+						}' which no longer exists`;
 						logger.debug(eventNotExistError);
+						logger.trace(
+							'Peer does not have a socket',
+							peerExtendedWithPublish.object()
+						);
 					}
 				};
 				return peerExtendedWithPublish;
@@ -184,14 +189,16 @@ const connectSteps = {
 
 		socket.on('connect', () => {
 			logger.trace(
-				`[Outbound socket :: connect] Peer connection to ${peer.ip} established`
+				`[Outbound socket :: connect] Peer connection to ${
+					peer.string
+				} established`
 			);
 		});
 
 		socket.on('disconnect', () => {
 			logger.trace(
 				`[Outbound socket :: disconnect] Peer connection to ${
-					peer.ip
+					peer.string
 				} disconnected`
 			);
 		});
@@ -208,7 +215,9 @@ const connectSteps = {
 		// When error on transport layer occurs - disconnect
 		socket.on('error', err => {
 			logger.debug(
-				`[Outbound socket :: error] Peer error from ${peer.ip} - ${err.message}`
+				`[Outbound socket :: error] Peer error from ${peer.string} - ${
+					err.message
+				}`
 			);
 			socket.disconnect(
 				1000,
@@ -220,7 +229,7 @@ const connectSteps = {
 		socket.on('close', (code, reason) => {
 			logger.debug(
 				`[Outbound socket :: close] Peer connection to ${
-					peer.ip
+					peer.string
 				} closed with code ${code} and reason - ${reason}`
 			);
 
@@ -238,7 +247,7 @@ const connectSteps = {
 		socket.on('message', message => {
 			logger.trace(
 				`[Outbound socket :: message] Peer message from ${
-					peer.ip
+					peer.string
 				} received - ${message}`
 			);
 		});
--- a/api/ws/rpc/connect.js
+++ b/api/ws/rpc/connect.js
@@ -165,7 +165,13 @@ const connectSteps = {
 						`[Outbound socket :: emit] Peer event '${eventProcedureName}' called with data`,
 						data
 					);
-					peer.socket.emit(eventProcedureName, data);
+					if (peer.socket) {
+						peer.socket.emit(eventProcedureName, data);
+					} else {
+						const eventNotExistError =
+							'Tried to emit event on outbound peer socket which no longer exists';
+						logger.debug(eventNotExistError);
+					}
 				};
 				return peerExtendedWithPublish;
 			},
--- a/logic/broadcaster.js
+++ b/logic/broadcaster.js
@@ -130,8 +130,11 @@ class Broadcaster {
 					if (!params.peers) {
 						return self.getPeers(params, waterCb);
 					}
-					const peers = params.limit ? params.peers.slice(0, params.limit) : params.peers;
-					return setImmediate(waterCb, null, peers);
+					return setImmediate(
+						waterCb,
+						null,
+						params.peers.slice(0, params.limit)
+					);
 				},
 				function sendToPeer(peers, waterCb) {
 					library.logger.debug('Begin broadcast', options);
--- a/logic/broadcaster.js
+++ b/logic/broadcaster.js
@@ -103,15 +103,12 @@ class Broadcaster {
 	 */
 	getPeers(params, cb) {
 		params.limit = params.limit || this.config.peerLimit;
-
 		const peers = library.logic.peers.listRandomConnected(params);
-
 		library.logger.info(
 			['Broadhash consensus now', modules.peers.getLastConsensus(), '%'].join(
 				' '
 			)
 		);
-
 		return setImmediate(cb, null, peers);
 	}
 
@@ -125,7 +122,7 @@ class Broadcaster {
 	 * @todo Add description for the params
 	 */
 	broadcast(params, options, cb) {
-		params.limit = params.limit || this.config.peerLimit;
+		params.limit = params.limit || this.config.broadcastLimit;
 
 		async.waterfall(
 			[
@@ -133,14 +130,12 @@ class Broadcaster {
 					if (!params.peers) {
 						return self.getPeers(params, waterCb);
 					}
-					return setImmediate(waterCb, null, params.peers);
+					const peers = params.limit ? params.peers.slice(0, params.limit) : params.peers;
+					return setImmediate(waterCb, null, peers);
 				},
 				function sendToPeer(peers, waterCb) {
 					library.logger.debug('Begin broadcast', options);
-					peers = peers.slice(0, params.limit);
-					peers
-						.slice(0, self.config.peerLimit)
-						.map(peer => peer.rpc[options.api](options.data));
+					peers.forEach(peer => peer.rpc[options.api](options.data));
 					library.logger.debug('End broadcast');
 					return setImmediate(waterCb, null, peers);
 				},
--- a/logic/broadcaster.js
+++ b/logic/broadcaster.js
@@ -103,6 +103,9 @@ class Broadcaster {
 	 */
 	getPeers(params, cb) {
 		params.limit = params.limit || this.config.peerLimit;
+		// Leave some room in the broadcast list for peers with a
+		// non-matching broadhash in order to avoid forks.
+		params.matchingPeersRatio = constants.matchingPeersRatio;
 		params.broadhash = params.broadhash || null;
 		params.normalized = false;
 
--- a/logic/broadcaster.js
+++ b/logic/broadcaster.js
@@ -129,7 +129,6 @@ class Broadcaster {
 	 */
 	broadcast(params, options, cb) {
 		params.limit = params.limit || this.config.peerLimit;
-		params.broadhash = params.broadhash || null;
 
 		async.waterfall(
 			[
--- a/logic/broadcaster.js
+++ b/logic/broadcaster.js
@@ -105,15 +105,12 @@ class Broadcaster {
 		params.limit = params.limit || this.config.peerLimit;
 
 		const peers = library.logic.peers.listRandomConnected(params);
-		const originalLimit = params.limit;
 
-		if (originalLimit === constants.maxPeers) {
-			library.logger.info(
-				['Broadhash consensus now', modules.peers.getLastConsensus(), '%'].join(
-					' '
-				)
-			);
-		}
+		library.logger.info(
+			['Broadhash consensus now', modules.peers.getLastConsensus(), '%'].join(
+				' '
+			)
+		);
 
 		return setImmediate(cb, null, peers);
 	}
@@ -140,11 +137,7 @@ class Broadcaster {
 				},
 				function sendToPeer(peers, waterCb) {
 					library.logger.debug('Begin broadcast', options);
-
-					if (params.limit === self.config.peerLimit) {
-						peers = peers.slice(0, self.config.broadcastLimit);
-					}
-
+					peers = peers.slice(0, params.limit);
 					peers
 						.slice(0, self.config.peerLimit)
 						.map(peer => peer.rpc[options.api](options.data));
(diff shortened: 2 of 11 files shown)


api/ws/rpc/connect.js
```diff
@@ -165,7 +165,13 @@ const connectSteps = {
 						`[Outbound socket :: emit] Peer event '${eventProcedureName}' called with data`,
 						data
 					);
-					peer.socket.emit(eventProcedureName, data);
+					if (peer.socket) {
+						peer.socket.emit(eventProcedureName, data);
+					} else {
+						const eventNotExistError =
+							'Tried to emit event on outbound peer socket which no longer exists';
+						logger.debug(eventNotExistError);
```
4miners: Please log something more, like `peer.string` on `debug` level and maybe some more data (entire `peer` object?) on `trace` level.

logic/broadcaster.js
```diff
@@ -103,27 +103,19 @@ class Broadcaster {
 	 */
 	getPeers(params, cb) {
 		params.limit = params.limit || this.config.peerLimit;
-		params.broadhash = params.broadhash || null;
-		params.normalized = false;
 
+		const peers = library.logic.peers.listRandomConnected(params);
 		const originalLimit = params.limit;
```
4miners: It's not needed anymore, as `listRandomConnected` is not mutating params.

logic/broadcaster.js
```diff
@@ -103,27 +103,19 @@ class Broadcaster {
 	 */
 	getPeers(params, cb) {
 		params.limit = params.limit || this.config.peerLimit;
```
4miners: We're already using `self` in this module, so we should be consistent.

logic/broadcaster.js
```diff
@@ -137,7 +129,6 @@ class Broadcaster {
 	 */
 	broadcast(params, options, cb) {
 		params.limit = params.limit || this.config.peerLimit;
```
4miners: `self.config.peerLimit`

4miners: Regarding that default limit. Would be nice if we can get rid of the condition that is used below:
- when `params.limit` is equal to `self.config.peerLimit` then truncate `peers` to `self.config.broadcastLimit`, then truncate to `self.config.peerLimit`
- when above condition is not met just truncate to `self.config.peerLimit`

It's unclear and can result with unexpected behavior.
Also we're returnign `body` property from this function, for what?

logic/peers.js
```diff
@@ -254,6 +254,24 @@ Peers.prototype.list = function(normalize) {
...
+ * @param {number} [options.limit=null] - Maximum number of peers to get; defaults to all
+ * @returns {peer[]} List of peers
+ */
+Peers.prototype.listRandomConnected = function(options) {
+	options = options || {};
+	const peerList = Object.keys(self.peersManager.peers)
+		.map(key => self.create(self.peersManager.getByAddress(key)))
+		.filter(peer => peer.state === Peer.STATE.CONNECTED);
+	const shuffledPeerList = _.shuffle(peerList);
+	return options.limit == null
```
4miners: `return option.limit ?` or use strict operator `===` for comparison.

logic/peers.js
```diff
@@ -254,6 +254,24 @@ Peers.prototype.list = function(normalize) {
 	);
 };
 
+/**
+ * Returns a random list of connected peers.
+ *
+ * @param {Object} [options] - Optional
+ * @param {number} [options.limit=null] - Maximum number of peers to get; defaults to all
```
4miners: Why default to `null`?

logic/peers.js
```diff
@@ -254,6 +254,24 @@ Peers.prototype.list = function(normalize) {
...
+ * @returns {peer[]} List of peers
+ */
+Peers.prototype.listRandomConnected = function(options) {
+	options = options || {};
+	const peerList = Object.keys(self.peersManager.peers)
+		.map(key => self.create(self.peersManager.getByAddress(key)))
```
4miners: `self.create` is not needed, as every peer that comes from `self.peersManager.peers` is alredy a `Peer` instance.

logic/peers.js
```diff
@@ -254,6 +254,24 @@ Peers.prototype.list = function(normalize) {
...
+/**
+ * Returns a random list of connected peers.
+ *
+ * @param {Object} [options] - Optional
+ * @param {number} [options.limit=null] - Maximum number of peers to get; defaults to all
+ * @returns {peer[]} List of peers
+ */
+Peers.prototype.listRandomConnected = function(options) {
+	options = options || {};
+	const peerList = Object.keys(self.peersManager.peers)
```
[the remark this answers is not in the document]
4miners: Please refactor, calling `Object.keys` and then `self.peersManager.getByAddress(key)` to get that peer again is not the best way.

logic/peers.js
```diff
@@ -254,6 +254,24 @@ Peers.prototype.list = function(normalize) {
...
 };
 
+/**
+ * Returns a random list of connected peers.
+ *
+ * @param {Object} [options] - Optional
+ * @param {number} [options.limit=null] - Maximum number of peers to get; defaults to all
+ * @returns {peer[]} List of peers
+ */
+Peers.prototype.listRandomConnected = function(options) {
```
4miners: Why that function is in `logic/peers` and not in `modules/peers` as `list`?

modules/transport.js
```diff
@@ -376,40 +376,42 @@ Transport.prototype.onUnconfirmedTransaction = function(
...
+	// Grab a random list of connected peers.
+	const peers = library.logic.peers.listRandomConnected({
+		limit: constants.maxPeers,
+	});
 
-		library.logger.debug(
-			'Transport->broadcastHeaders: Broadcasting headers to remote peers',
-			{ count: peers.length }
-		);
+	if (!peers || peers.length === 0) {
```
4miners: `listRandomConnected` always returns array, isn't it?

modules/peers.js
```diff
@@ -225,7 +225,7 @@ __private.getByFilter = function(filter, cb) {
 		}
 	} else {
 		// Sort randomly by default
-		peers = shuffle(peers);
+		peers = __private.shuffle(peers);
```
nazarhussain: In `logic/peer` we used `_.shuffle`, so for consistency purpose we should use same shuffling method every where.

modules/peers.js
```diff
@@ -710,6 +710,7 @@ Peers.prototype.list = function(options, cb) {
 					// Unmatched broadhash
 					return randomList(peers, waterCb);
 				}
+				peers = __private.shuffle(peers);
```
nazarhussain: Same here

modules/transport.js
```diff
@@ -376,40 +376,42 @@ Transport.prototype.onUnconfirmedTransaction = function(
...
-	// Grab list of all connected peers
-	modules.peers.list({ normalized: false }, (err, peers) => {
-		if (!peers || peers.length === 0) {
-			library.logger.debug('Transport->broadcastHeaders: No peers found');
-			return setImmediate(cb);
-		}
+	// Grab a random list of connected peers.
+	const peers = library.logic.peers.listRandomConnected({
+		limit: constants.maxPeers,
+	});
```
nazarhussain: In the previous change where you moved `modules.peers.list` to `logic.peers.listRandomConnected`, you also called `findGoodPeers`, which provide you peers with maximum height in the network. 

Here we are making this change without `findGoodPeers`, so bad peers which are not up to network level will also be used. Is that really necessary? 

`broadcastHeaders` is used during `sync` and then `processBlock`, We can remove it from `sync` as it will eventually use `processBlock` in the end.

4miners: @nazarhussain We shouldn't remove `broadcastHeaders` from sync, as `processBlock` from sync is called with `broadcast = false` so it will not call `broadcastHeaders`.

logic/peers.js
jondubois: My thinking was that ideally modules shouldn't call methods of each other (since they already communicate via message passing) and that logic/ classes exist for the purpose of sharing overlapping functionality between modules.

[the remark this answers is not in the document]
jondubois: It seems that we all have slightly different interpretations about what module/logic/helper is responsible for what. It's hard to come up with a single clear design principle/reasoning/explanation is behind each different type.

logic/broadcaster.js
jondubois: WTF yes that's really bad logic ;p

logic/broadcaster.js
ManuGowda: @jondubois you can move the static function `getPeers` outside of the class and use to `self` in order to maintain consistency, otherwise, we have to refactor the `Broadcaster` constructor to use `this` instead of `self`.

logic/peers.js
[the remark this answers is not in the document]
jondubois: @shuse2 Our current node.js version doesn't support `Object.values(...)` but I'll make it not use `self.peersManager.getByAddress(key)`.

logic/broadcaster.js
```diff
@@ -103,27 +103,16 @@ class Broadcaster {
...
-			if (originalLimit === constants.maxPeers) {
-				library.logger.info(
-					[
-						'Broadhash consensus now',
-						modules.peers.getLastConsensus(),
-						'%',
-					].join(' ')
-				);
-			}
+		library.logger.info(
```
4miners: Removing `if (originalLimit === constants.maxPeers) {` condition will not cause spam with that log?

logic/broadcaster.js
```diff
@@ -137,7 +126,6 @@ class Broadcaster {
 	 */
 	broadcast(params, options, cb) {
 		params.limit = params.limit || this.config.peerLimit;
```
4miners: Are you sure that default souldn't be `self.config.broadcastLimit` instead of `this.config.peerLimit`?

logic/peers.js
```diff
@@ -254,6 +254,24 @@ Peers.prototype.list = function(normalize) {
...
+ * @param {number} [options.limit] - Maximum number of peers to get; defaults to all
+ * @returns {peer[]} List of peers
+ */
+Peers.prototype.listRandomConnected = function(options) {
+	options = options || {};
+	const peerList = Object.keys(self.peersManager.peers)
+		.map(key => self.peersManager.peers[key])
+		.filter(peer => peer.state === Peer.STATE.CONNECTED);
+	const shuffledPeerList = _.shuffle(peerList);
+	return options.limit === undefined
```
4miners: Why `=== undefined`? Why not just `options.limit` (`undefined`, `null`, `0`, etc.)?

logic/broadcaster.js
jondubois: Yes I think that would be better, also the logic below can be simplified it's slicing the array multiple times.

4miners: @jondubois Are you really sure that you want to change behavior? Notice that we are calling `broadcast` here:
https://github.com/LiskHQ/lisk/blob/912d22381a4c3c17447ac4c57b2f00dffbeaac11/modules/transport.js#L438-L442
So before the changes second truncate (to `self.config.broadcastLimit`) was performed, now - it's not. Also `self.config.broadcastLimit` is not used anymore.

logic/peers.js
jondubois: @4miners I wanted to account for both `null` and `undefined` but not `0` (that's why I used `x == null` initially).

I can change it to check only for truthiness though (treat 0 as the same as undefined) if that's consistent with the rest of the code.

logic/broadcaster.js
jondubois: @ManuGowda I prefer to keep it like this for this PR to keep the diff clean/simple and to avoid conflicts but we can cleanup later once betanet is stable.

logic/broadcaster.js
```diff
@@ -103,27 +103,13 @@ class Broadcaster {
...
-						modules.peers.getLastConsensus(),
-						'%',
-					].join(' ')
-				);
-			}
-
-			return setImmediate(cb, null, peers);
-		});
+		const peers = library.logic.peers.listRandomConnected(params);
+		library.logger.info(
```
4miners: Please check if that log will not be spammy after removing condition.

logic/broadcaster.js
```diff
@@ -136,34 +122,27 @@ class Broadcaster {
...
+		params.limit = params.limit || this.config.broadcastLimit;
 
 		async.waterfall(
 			[
 				function getPeers(waterCb) {
 					if (!params.peers) {
 						return self.getPeers(params, waterCb);
 					}
-					return setImmediate(waterCb, null, params.peers);
+					const peers = params.limit ? params.peers.slice(0, params.limit) : params.peers;
```
4miners: `params.limit` will be always present because there is default value.

modules/transport.js
```diff
@@ -437,7 +439,7 @@ Transport.prototype.onBroadcastBlock = function(block, broadcast) {
 	// Perform actual broadcast operation
 	__private.broadcaster.broadcast(
 		{
-			limit: constants.maxPeers,
+			limit: library.config.broadcasts.broadcastLimit,
```
4miners: You can skip that if there is default limit in broadcaster that is the same.

logic/broadcaster.js
jondubois: I checked on localhost, it seems OK, it's only once per broadcast (once for all peers), also it's info log level.

logic/broadcaster.js
```diff
@@ -136,34 +122,30 @@ class Broadcaster {
...
-						.slice(0, self.config.peerLimit)
-						.map(peer => peer.rpc[options.api](options.data));
+					peers.forEach(peer => peer.rpc[options.api](options.data));
 					library.logger.debug('End broadcast');
 					return setImmediate(waterCb, null, peers);
 				},
 			],
 			(err, peers) => {
 				if (cb) {
-					return setImmediate(cb, err, { body: null, peer: peers });
```
MaciejBaj: `body` wasn't used anywhere?

jondubois: No it wasn't. There was a comment earlier to remove it.

test/unit/modules/transport.js
```diff
@@ -1505,7 +1505,6 @@ describe('transport', () => {
 					return expect(
 						__private.broadcaster.broadcast.calledWith(
 							{
-								limit: library.config.broadcasts.broadcastLimit,
```
4miners: Please update description of that test (`it`).

PR #2045, opened 2018-05-18
Peer updates itself every-time during peers discovery and update - Closes #2029

### LINKED ISSUE #2029: Fix unreachable peers code of self-update
### Expected behavior
The intended use `me` functions was `library.logic.peers.me()`
### Actual behavior
In `__private.updatePeerStatus` we are using undefined function `self.me().nonce` where `me` function is undefined.
### Steps to reproduce
https://github.com/LiskHQ/lisk/blob/6951f0d307e223073d8e9839c5ffbceb0f51eccf/modules/peers.js#L277
1.0.0

ManuGowda:
### What was the problem?
- Bug in the way calling `self.me()`
- In the process of `peersDiscoveryAndUpdate` (every 10 sec) the peer updates itself
### How did I fix it?
- Updated calling `self.me()` to `library.logic.peers.me()`
- Before the update, the peer list will be filtered by calling `self.acceptable(peers)`
### How to test it?
Run integration test and enable `trace` level log there should be any update happening to self-peer during the discovery process.
### Review checklist

* The PR solves #2029 
* Linting passes
* Tests pass

### DIFF
--- a/modules/peers.js
+++ b/modules/peers.js
@@ -813,11 +813,15 @@ Peers.prototype.onPeersReady = function() {
 				updatePeers(seriesCb) {
 					let updated = 0;
 					const peers = library.logic.peers.list();
+					library.logger.trace('Total list of peers', { count: peers.length });
 
-					library.logger.trace('Updating peers', { count: peers.length });
+					const acceptablePeers = self.acceptable(peers);
+					library.logger.trace('Acceptable peers to update', {
+						count: acceptablePeers.length,
+					});
 
 					async.each(
-						peers,
+						acceptablePeers,
 						(peer, eachCb) => {
 							// If peer is not banned and not been updated during last 3 sec - ping
 							if (
--- a/modules/peers.js
+++ b/modules/peers.js
@@ -813,15 +813,11 @@ Peers.prototype.onPeersReady = function() {
 				updatePeers(seriesCb) {
 					let updated = 0;
 					const peers = library.logic.peers.list();
-					library.logger.trace('Total list of peers', { count: peers.length });
 
-					const acceptablePeers = self.acceptable(peers);
-					library.logger.trace('Acceptable peers to update', {
-						count: acceptablePeers.length,
-					});
+					library.logger.trace('Updating peers', { count: peers.length });
 
 					async.each(
-						acceptablePeers,
+						peers,
 						(peer, eachCb) => {
 							// If peer is not banned and not been updated during last 3 sec - ping
 							if (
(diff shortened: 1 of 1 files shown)


modules/peers.js
```diff
@@ -274,7 +274,7 @@ __private.updatePeerStatus = function(err, status, peer) {
 			// state so that the node doesn't keep trying to reconnect to itself.
 			peer.applyHeaders({
 				state: Peer.STATE.BANNED,
-				nonce: self.me().nonce,
+				nonce: library.logic.peers.me().nonce,
```
4miners: 👍

modules/peers.js
```diff
@@ -813,11 +813,15 @@ Peers.prototype.onPeersReady = function() {
 				updatePeers(seriesCb) {
 					let updated = 0;
 					const peers = library.logic.peers.list();
+					library.logger.trace('Total list of peers', { count: peers.length });
 
-					library.logger.trace('Updating peers', { count: peers.length });
+					const acceptablePeers = self.acceptable(peers);
```
4miners: We have function `pickPeers` that is calling `acceptable` and logging both total and accepted peers. Use that one.

shuse2: I'm not sure why we need to filter with acceptable here.
When we discover, it is already filtered with this function, and it should be already only acceptable in the list

4miners: Why we need that at all? `acceptable` is already called during discover process (`Peers.prototype.discover`), so our own peer shouldn't reach peers manager and then shouldn't be returned from `library.logic.peers.list`. How did it get there?

ManuGowda:
@shuse2 @4miners Thank you for the feedback, when I tested the `peersDiscoveryAndUpdate` last week I found the `library.logic.peers.list` returning itself, however, I tried to cross-check the behavior today and I see that is not happening and `library.logic.peers.list` is excluding itself, so I am reverting the changes, thank you for the feedback.

4miners:
@ManuGowda Please double-check that, why it stopped returning itself? Did we change anything related to it in other PR?

ManuGowda:
@4miners I double checked by putting the log in `modules/peers.js - peersDiscoveryAndUpdate - updatePeers` and I don't see itself connected. here is the logic what i am try to test inside `updatePeers` in `peersDiscoveryAndUpdate`.
```
peers.forEach(peer => {
    if(library.logic.peers.me().wsPort === peer.wsPort) {
       console.error('peer connected to itself');
    }
```

ManuGowda:
closing the PR as I created new PR against `1.0.0-beta.7.0` https://github.com/LiskHQ/lisk/pull/2050

PR #2048, opened 2018-05-22
Add config.json migration script from from old version 0.9.x - Closes #2046

### LINKED ISSUE #2046: Add config.json migration script from from old version 0.9.x
### Expected behavior

Need a script inside core to migrate the config.json file from 0.9.x version to current version. The script should be initiated manually by user if running from source. And we will trigger the script from lisk_bridge.sh script during core migration. 

### Actual behavior
Currently the migration script is in lisk-scripts repo and not updated. 

1.0.0

nazarhussain:
### What was the problem?
Config migration script is in lisk-scripts repo and not updated.

### How did I fix it?
Created a script inside core to migrate the config.json file from 0.9.x version to current version

### How to test it?
Run the script and match the config file from master and 1.0.0 branch. 

### Review checklist

* The PR solves #2046
* Linting passes
* Tests pass

### DIFF
--- a/scripts/update_config.js
+++ b/scripts/update_config.js
@@ -59,9 +59,6 @@ if (program.old) {
 		delete p.port;
 	});
 
-	// Secrets Migration
-	oldConfig.forging.delegates = [];
-
 	if (oldConfig.forging.secret && oldConfig.forging.secret.length) {
 		if (!program.password.trim()) {
 			const rl = readline.createInterface({
@@ -72,10 +69,8 @@ if (program.old) {
 				'We found some secrets in your config, if you want to migrate, please type in your password (enter to skip): ',
 				password => {
 					rl.close();
-					if (!password.trim()) {
-						return console.info('\nSkipping the secret migration.');
-					}
 					migrateSecrets(password);
+					copyTheConfigFile();
 				}
 			);
 			// To Patch the password support
@@ -86,9 +81,11 @@ if (program.old) {
 			};
 		} else {
 			migrateSecrets(program.password);
+			copyTheConfigFile();
 		}
 	} else {
-		delete oldConfig.forging.secret;
+		migrateSecrets('');
+		copyTheConfigFile();
 	}
 } else {
 	console.info('Previous config.json not found, exiting.');
@@ -96,6 +93,14 @@ if (program.old) {
 }
 
 function migrateSecrets(password) {
+	oldConfig.forging.delegates = [];
+
+	if (!password.trim()) {
+		console.info('\nSkipping the secret migration.');
+		delete oldConfig.forging.secret;
+		return;
+	}
+
 	console.info('\nMigrating your secrets...');
 	oldConfig.forging.defaultPassword = password;
 	oldConfig.forging.secret.map(secret => {
@@ -113,17 +118,20 @@ function migrateSecrets(password) {
 			).publicKey,
 		});
 	});
+
 	delete oldConfig.forging.secret;
-	console.info('Configuration migration completed.');
 }
 
-if (program.new) {
-	newConfig = JSON.parse(fs.readFileSync(program.new, 'utf8'));
-	newConfig = extend(true, {}, newConfig, oldConfig);
+function copyTheConfigFile() {
+	if (program.new) {
+		newConfig = JSON.parse(fs.readFileSync(program.new, 'utf8'));
+		newConfig = extend(true, {}, newConfig, oldConfig);
 
-	fs.writeFile(program.new, JSON.stringify(newConfig, null, '\t'), err => {
-		if (err) {
-			throw err;
-		}
-	});
+		fs.writeFile(program.new, JSON.stringify(newConfig, null, '\t'), err => {
+			if (err) {
+				throw err;
+			}
+		});
+	}
+	console.info('Configuration migration completed.');
 }
--- a/scripts/update_config.js
+++ b/scripts/update_config.js
@@ -143,8 +143,9 @@ function copyTheConfigFile() {
 		err => {
 			if (err) {
 				throw err;
+			} else {
+				console.info('Configuration migration completed.');
 			}
 		}
 	);
-	console.info('Configuration migration completed.');
 }
--- a/scripts/update_config.js
+++ b/scripts/update_config.js
@@ -14,6 +14,9 @@
  *
  * Usage Example:
  * 		node scripts/update_config.js ../lisk-backup/config.json ./config.json
+ *
+ * 	Reference:
+ * 		A user manual can be found on documentation site under /documentation/lisk-core/upgrade/upgrade-configurations
  */
 
 const fs = require('fs');
--- a/scripts/update_config.js
+++ b/scripts/update_config.js
@@ -19,77 +19,82 @@ const program = require('commander');
 const extend = require('extend');
 const lisk = require('lisk-js');
 
+let oldConfigPath;
+let newConfigPath;
+
 program
 	.version('0.1.1')
-	.option('-o, --old <path>', 'Old config.json')
-	.option('-n, --new <path>', 'New config.json', './config.json')
+	.arguments('<old_config_file> <new_config_file>')
+	.action((oldConfig, newConfig) => {
+		oldConfigPath = oldConfig;
+		newConfigPath = newConfig;
+	})
 	.option('-p, --password <string>', 'Password for secret encryption', '')
 	.parse(process.argv);
 
-let oldConfig;
-let newConfig;
+if (!oldConfigPath && !newConfigPath) {
+	console.error('error: no config file provided.');
+	process.exit(1);
+}
 
-if (program.old) {
-	oldConfig = JSON.parse(fs.readFileSync(program.old, 'utf8'));
+console.info('Starting configuration migration...');
+const oldConfig = JSON.parse(fs.readFileSync(oldConfigPath, 'utf8'));
+const newConfig = JSON.parse(fs.readFileSync(newConfigPath, 'utf8'));
 
-	// Values to keep from new config file
-	delete oldConfig.version;
-	delete oldConfig.minVersion;
+// Values to keep from new config file
+delete oldConfig.version;
+delete oldConfig.minVersion;
 
-	// Rename old port to new wsPort
-	oldConfig.httpPort = oldConfig.port;
-	oldConfig.wsPort = oldConfig.httpPort + 1;
-	delete oldConfig.port;
+// Rename old port to new wsPort
+oldConfig.httpPort = oldConfig.port;
+oldConfig.wsPort = oldConfig.httpPort + 1;
+delete oldConfig.port;
 
-	oldConfig.db.max = oldConfig.db.poolSize;
-	delete oldConfig.db.poolSize;
+oldConfig.db.max = oldConfig.db.poolSize;
+delete oldConfig.db.poolSize;
 
-	delete oldConfig.peers.options.limits;
+delete oldConfig.peers.options.limits;
 
-	oldConfig.transactions.maxTransactionsPerQueue =
-		oldConfig.transactions.maxTxsPerQueue;
-	delete oldConfig.transactions.maxTxsPerQueue;
+oldConfig.transactions.maxTransactionsPerQueue =
+	oldConfig.transactions.maxTxsPerQueue;
+delete oldConfig.transactions.maxTxsPerQueue;
 
-	delete oldConfig.loading.verifyOnLoading;
-	delete oldConfig.dapp;
+delete oldConfig.loading.verifyOnLoading;
+delete oldConfig.dapp;
 
-	// Peers migration
-	oldConfig.peers.list.map(p => {
-		p.wsPort = p.port;
-		delete p.port;
-	});
+// Peers migration
+oldConfig.peers.list.map(p => {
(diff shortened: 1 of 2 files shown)


scripts/update_config.js
```diff
@@ -0,0 +1,131 @@
...
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+const fs = require('fs');
+const readline = require('readline');
+const program = require('commander');
+const extend = require('extend');
+const lisk = require('lisk-js');
+
+/* eslint-disable no-console */
```
willclarktech: Why not use `console.info`, `console.error` etc instead?

scripts/update_config.js
```diff
@@ -0,0 +1,131 @@
...
+
+program
+	.version('0.1.1')
+	.option('-o, --old <path>', 'Old config.json')
```
willclarktech: Why make this an option if it's not optional?

nazarhussain: @willclarktech I don't find any other way to provide a required option in `commander` I ported the script from other repo, so most of boilerplate is the same. Do you know is there any option to specify required option in commander? 
https://www.npmjs.com/package/commander

willclarktech: Isn't that what the `.parse(process.argv)` bit is for?

nazarhussain:
@diego-G All our scripts works with assumption that user choose the right network himself. Even thought for testing I tried producing the issue you mentioned but I can't. What were version of old config? and new config file? you tried with.

diego-G:
Problem comes when we have different peers list. The script only overwrites or adds the ones existing in the old config file, keeping all the rest with wrong ports and provoking errors in the logs. We should clean the peerslist before migrating.

nazarhussain:
The script merges the `peers.list` from old config and new config and copy to new config. And I believe that's the right behavior.

scripts/update_config.js
```diff
@@ -0,0 +1,137 @@
...
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
```
MaciejBaj: The fast script manual would be very helpful here - examples.

scripts/update_config.js
```diff
@@ -0,0 +1,137 @@
...
+const fs = require('fs');
+const readline = require('readline');
+const program = require('commander');
+const extend = require('extend');
+const lisk = require('lisk-js');
+
+program
+	.version('0.1.1')
+	.option('-o, --old <path>', 'Old config.json')
```
willclarktech: I still think the path to the old config.json should be an arg, not an option. It's not optional. Maybe the new path would also make sense as an (optional) arg?

scripts/update_config.js
```diff
@@ -0,0 +1,137 @@
...
+	.option('-n, --new <path>', 'New config.json', './config.json')
+	.option('-p, --password <string>', 'Password for secret encryption', '')
```
willclarktech: Have we considered more secure ways to pass the password without a prompt? Do we even want to give users the option to do this?

scripts/update_config.js
```diff
@@ -0,0 +1,137 @@
...
+				secret
+			).publicKey,
+		});
+	});
+
+	delete oldConfig.forging.secret;
+}
+
+function copyTheConfigFile() {
+	if (program.new) {
```
willclarktech: What's the point of running this script without specifying a new path?

scripts/update_config.js
```diff
@@ -19,77 +19,82 @@ const program = require('commander');
...
+delete oldConfig.loading.verifyOnLoading;
+delete oldConfig.dapp;
 
-	// Peers migration
-	oldConfig.peers.list.map(p => {
-		p.wsPort = p.port;
-		delete p.port;
-	});
+// Peers migration
+oldConfig.peers.list.map(p => {
```
willclarktech: This should be a `forEach`.

scripts/update_config.js
```diff
@@ -19,77 +19,82 @@ const program = require('commander');
...
+			'We found some secrets in your config, if you want to migrate, please type in your password (enter to skip): ',
+			password => {
+				rl.close();
+				migrateSecrets(password);
+				copyTheConfigFile();
+			}
+		);
+		// To Patch the password support
+		rl.stdoutMuted = true;
+		rl._writeToOutput = function _writeToOutput(stringToWrite) {
```
willclarktech: This doesn't seem like a best practice.

scripts/update_config.js
```diff
@@ -123,15 +128,16 @@ function migrateSecrets(password) {
 }
 
 function copyTheConfigFile() {
-	if (program.new) {
-		newConfig = JSON.parse(fs.readFileSync(program.new, 'utf8'));
-		newConfig = extend(true, {}, newConfig, oldConfig);
+	const modifiedConfig = extend(true, {}, newConfig, oldConfig);
 
-		fs.writeFile(program.new, JSON.stringify(newConfig, null, '\t'), err => {
+	fs.writeFile(
```
willclarktech: Do you want `writeFileSync`?

scripts/update_config.js
```diff
@@ -0,0 +1,143 @@
...
+let newConfigPath;
+
+program
+	.version('0.1.1')
+	.arguments('<old_config_file> <new_config_file>')
+	.action((oldConfig, newConfig) => {
+		oldConfigPath = oldConfig;
+		newConfigPath = newConfig;
+	})
+	.option('-p, --password <string>', 'Password for secret encryption', '')
```
willclarktech: I think here is a good location to document that this is not recommended for real passwords.

scripts/update_config.js
```diff
@@ -0,0 +1,143 @@
...
+
+	if (!password.trim()) {
+		console.info('\nSkipping the secret migration.');
+		delete oldConfig.forging.secret;
+		return;
+	}
+
+	console.info('\nMigrating your secrets...');
+	oldConfig.forging.defaultPassword = password;
+	oldConfig.forging.secret.map(secret => {
```
willclarktech: Either you want a `forEach` here or you want to assign `oldConfig.forging.delegates` to the result of the `map` function, and return the object rather than pushing it.

scripts/update_config.js
```diff
@@ -19,77 +19,82 @@ const program = require('commander');
...
+let newConfigPath;
+
 program
 	.version('0.1.1')
-	.option('-o, --old <path>', 'Old config.json')
-	.option('-n, --new <path>', 'New config.json', './config.json')
+	.arguments('<old_config_file> <new_config_file>')
+	.action((oldConfig, newConfig) => {
+		oldConfigPath = oldConfig;
+		newConfigPath = newConfig;
```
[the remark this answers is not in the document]
nazarhussain: @willclarktech Not necessary... Remember this script will be initiated by another script also maintained by us. And this script is not that used quite often, once for migration. So we don't need to optimize it further.

scripts/update_config.js
```diff
@@ -19,77 +19,82 @@ const program = require('commander');
...
-			rl._writeToOutput = function _writeToOutput(stringToWrite) {
-				if (rl.stdoutMuted) rl.output.write('*');
-				else rl.output.write(stringToWrite);
-			};
-		} else {
-			migrateSecrets(program.password);
-			copyTheConfigFile();
-		}
+if (oldConfig.forging.secret && oldConfig.forging.secret.length) {
+	if (!program.password.trim()) {
```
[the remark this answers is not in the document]
nazarhussain: Its a best practice to trim user input irrespective of the platform. Sometimes carriage return characters are also been embedded to the input.

scripts/update_config.js
```diff
@@ -19,77 +19,82 @@ const program = require('commander');
...
 
+let oldConfigPath;
+let newConfigPath;
+
 program
 	.version('0.1.1')
-	.option('-o, --old <path>', 'Old config.json')
-	.option('-n, --new <path>', 'New config.json', './config.json')
+	.arguments('<old_config_file> <new_config_file>')
+	.action((oldConfig, newConfig) => {
```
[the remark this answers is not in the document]
nazarhussain: I don't see any odd thing even in this approach. Just assign appropriate values here and execute the reset of the functionality outside this method.

scripts/update_config.js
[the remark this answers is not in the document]
willclarktech: But what if somebody's password has spaces in it?

scripts/update_config.js
```diff
@@ -19,77 +19,82 @@ const program = require('commander');
...
+		oldConfigPath = oldConfig;
+		newConfigPath = newConfig;
+	})
 	.option('-p, --password <string>', 'Password for secret encryption', '')
 	.parse(process.argv);
 
-let oldConfig;
-let newConfig;
+if (!oldConfigPath && !newConfigPath) {
```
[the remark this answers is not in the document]
nazarhussain: There is no way I found in commander to make arguments required. If you can find let me know.

scripts/update_config.js
```diff
@@ -19,77 +19,82 @@ const program = require('commander');
...
+	console.error('error: no config file provided.');
+	process.exit(1);
+}
 
-if (program.old) {
-	oldConfig = JSON.parse(fs.readFileSync(program.old, 'utf8'));
+console.info('Starting configuration migration...');
+const oldConfig = JSON.parse(fs.readFileSync(oldConfigPath, 'utf8'));
+const newConfig = JSON.parse(fs.readFileSync(newConfigPath, 'utf8'));
```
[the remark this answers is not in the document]
nazarhussain: Yes that's what supposed to be.

scripts/update_config.js
nazarhussain: Yes already commented there, its a patch, because this functionality is not supported in current node version.

scripts/update_config.js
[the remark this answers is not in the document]
willclarktech: When I tried running this script, the timing of code inside this function compared to outside was not easy to work out.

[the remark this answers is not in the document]
nazarhussain: Can't understand...

scripts/update_config.js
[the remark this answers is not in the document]
nazarhussain: No you can't, you need to have both of the config files to migrate your configuration from old config file to new config file.

scripts/update_config.js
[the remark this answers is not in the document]
nazarhussain: Its not has spaces, trim only effects if someone starts or end with spaces. Which I don't see any probability.

scripts/update_config.js
```diff
@@ -0,0 +1,157 @@
...
+	.action((oldConf, newConf) => {
+		oldConfigPath = oldConf;
+		newConfigPath = newConf;
+
+		if (!oldConfigPath && !newConfigPath) {
+			console.error('error: no config file provided.');
+			process.exit(1);
+		}
+
+		migrateConfiguraiton();
```
willclarktech: Typo: `Configuration`

scripts/update_config.js
```diff
@@ -0,0 +1,157 @@
...
+		'-p, --password <string>',
+		"Password for secret encryption. This feature is only for testing purpose, don't use is it production.",
+		''
+	)
+	.arguments('<old_config_file> <new_config_file>')
+	.action((oldConf, newConf) => {
+		oldConfigPath = oldConf;
+		newConfigPath = newConf;
+
+		if (!oldConfigPath && !newConfigPath) {
```
willclarktech: This check doesn't seem to run. :( https://github.com/tj/commander.js/issues/729

scripts/update_config.js
```diff
@@ -0,0 +1,157 @@
...
+	oldConfig.forging.delegates = [];
+
+	if (!password.trim()) {
+		console.info('\nSkipping the secret migration.');
+		delete oldConfig.forging.secret;
+		return;
+	}
+
+	console.info('\nMigrating your secrets...');
+	oldConfig.forging.defaultPassword = password;
```
willclarktech: Should this definitely be stored here?

scripts/update_config.js
```diff
@@ -0,0 +1,157 @@
...
+	fs.writeFile(
+		newConfigPath,
+		JSON.stringify(modifiedConfig, null, '\t'),
+		err => {
+			if (err) {
+				throw err;
+			}
+		}
+	);
+	console.info('Configuration migration completed.');
```
willclarktech: This should be moved inside the callback, or `fs.writeFileSync` should be used.

scripts/update_config.js
nazarhussain: Updated it, but if you put this check outside the action an check the variables value which were set inside action, then what's was the reason of doing all refactoring before.

scripts/update_config.js
nazarhussain: Yes

nazarhussain:
The manual for this script cane be found on documentation under `/documentation/lisk-core/upgrade/upgrade-configurations`

MaciejBaj:
@nazarhussain thanks for preparing the manual. Please include it into the opening comment of `scripts/update-version.js`

PR #2050, opened 2018-05-22
Fix unreachable peers code of self-update - Closes #2029

### LINKED ISSUE #2029: Fix unreachable peers code of self-update

ManuGowda:
### What was the problem?
- Bug in the way calling `self.me()`
### How did I fix it?
- Updated calling `self.me()` to `library.logic.peers.me()`
### How to test it?
Run integration test and enable `trace` level log there should be any update happening to self-peer during the discovery process.
### Review checklist

* The PR solves #2029 
* Linting passes
* Tests pass

### DIFF
--- a/modules/peers.js
+++ b/modules/peers.js
@@ -272,7 +272,7 @@ __private.updatePeerStatus = function(err, status, peer) {
 			// state so that the node doesn't keep trying to reconnect to itself.
 			peer.applyHeaders({
 				state: Peer.STATE.BANNED,
-				nonce: self.me().nonce,
+				nonce: library.logic.peers.me().nonce,
 			});
 		} else {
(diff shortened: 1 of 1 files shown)


ManuGowda:
@4miners I think I will update the issue with the information as the peer is not updating itself as part of `peersDiscoveryAndUpdate`.

PR #2052, opened 2018-05-23
Count peers after applying filter instead of total - Closes #2020 

### LINKED ISSUE #2020: API peers response total count and count of peer in list doesn't match
### Expected behavior
API peers endpoint should always return all the peers as mentioned in the `total` count in `meta`
### Actual behavior
API peers return inconsistent response against the `total` count of peers.

### Steps to reproduce
1. Make a GET API call to endpoint `http://188.166.111.69:4000/api/peers?state=2&limit=100&offset=0`
2. Observer the meta `total` value in the response, it will be around 150 to 200(depending on the number of connected peers at that point in time
3) Continue getting the response from offset 100 to next 100 `http://188.166.111.69:4000/api/peers?state=2&limit=100&offset=100`
4) The response will have less than expected peers.
1.0.0

diego-G:
### What was the problem?

Endpoint returned always the total number of peers no matter which filter we were requesting.

### How did I fix it?

Instead of ~peers.getPeersCount~, we use now `peers.getCountByFilter` + removing offset and limit before using the filter. 

### How to test it?

`mocha tes/functional/http/get/peers.js`

### Review checklist

* The PR solves #2020
* Linting passes
* Tests pass

### DIFF
--- a/modules/peers.js
+++ b/modules/peers.js
@@ -77,19 +77,18 @@ class Peers {
 
 // Private methods
 /**
- * Returns peers length after getting them by filter.
+ * Returns peers length by filter but without offset and limit.
  *
  * @private
  * @param {Object} filter
- * @param {function} cb - Callback function
- * @returns {setImmediateCallback} cb, null, peers length
+ * @returns {int} count
  * @todo Add description for the params
  */
-__private.countByFilter = function(filter, cb) {
-	filter.normalized = false;
-	__private.getByFilter(filter, (err, peers) =>
-		setImmediate(cb, null, peers.length)
-	);
+__private.getCountByFilter = function(filter) {
+	delete filter.limit;
+	delete filter.offset;
+	var peers = __private.getByFilter(filter);
+	return peers.length;
 };
 
 /**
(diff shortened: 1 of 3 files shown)


modules/peers.js
```diff
@@ -938,6 +937,26 @@ Peers.prototype.shared = {
 	getPeersCount() {
```
ManuGowda: This is no longer used anywhere, so IMO we can remove this unused function.

modules/peers.js
```diff
@@ -938,6 +937,26 @@ Peers.prototype.shared = {
...
+	 * @param {string} parameters.os - OS of the peer
+	 * @param {string} parameters.version - Version the peer is running
+	 * @param {int} parameters.state - Peer State
+	 * @param {int} parameters.height - Current peer height
+	 * @param {string} parameters.broadhash - Peer broadhash
+	 * @returns {int} count
+	 * @todo Add description for the return value
+	 */
+	getPeersCountByFilter(parameters) {
+		parameters.normalized = true;
```
ManuGowda: @diego-G IMO we need to set the `normalized` flag to false to avoid unwanted peer properties in the peer list, as we are doing just the count of peers.

Issue #398, opened 2017-01-19
Support multiple networks with one installation


Isabello:
In our current implementation various files are required to be modified to change what network is able to be used.

These files are:
```
helpers/constants.js
helpers/exceptions.js
config.json
genesisBlock.json
```

The goal of this change is to simplify the use:

- the ease of running a node from source with any network from a single command, like `node app --mainnet`
- clarity of network - specific variables
- decreasing risk of making mistake while merging network specific files during releases

MaciejBaj:
After 1.0.0 the list of "network-dependant files" shapes as follows:
- [ ] config.json
- [ ] genesis_block.json
- [ ] helpers/exceptions.js
- [ ] helpers/constants.js
- [ ] scripts/update_config.js

Keeping them on one branch would decrease possible confusion and risk connected with working and releasing next versions on different networks:
- Alphanet (Devnet),
- Betanet,
- Testnet,
- Mainnet.

Each change to "network-dependant file" would have to added to:
- feature branch (like `1.0.0-beta.9.0`).

Isabello:
:)

2018-06-15 14:28 GMT+02:00 Maciej Baj <notifications@github.com>:

diego-G:
Please guys, don't forget `helpers/milestones.js` and `genesis_delegates.js` when applies.

PR #2056, opened 2018-05-24
Dockerfile with config.json templating - Closes #2053

### LINKED ISSUE #2053: Add Dockerfile to the lisk repository
Relates to liskhq/lisk-docker#84, liskhq/lisk-docker#85 and liskhq/lisk-docker#87

### Expected behavior

Docker images are built directly from within the `lisk` repository and `config.json` is configurable through environment variables.

### Actual behavior

For historical reasons docker images are built from `Dockerfile`(s) in a different repository and are only configurable to a limited extent.

### Steps to reproduce

Build and use docker images.

Version 0.9.x - 1.x; docker images.

ghost:
### What was the problem?

Docker images are were being built from Dockerfile(s) in a different repository and were only configurable to a limited extent.

### How did I fix it?

Added `Dockerfile` to this repository along with a template for `config.json`.

### How to test it?

- build docker image from top-level `Dockerfile`
- ensure `config.json` can be customized by setting environment variables, e.g. `LISK_DB_HOST`

### Review checklist

* The PR solves #2053
* Linting passes
* Tests pass


hendrikhofstadt (commented):
@fchavant I appreciate that you added my config proposal/template to the repo here 👍 Is there a reason you are still using the full ubuntu image instead of the much smaller alpine one ?

Also will you completely remove liskhq/lisk-docker and keep the scripts in this repo for better managebility (as we also discussed a submodule approach)

Dockerfile
```diff
@@ -0,0 +1,46 @@
...
+    else \
+        rm -f /tmp/confd; \
+	exit 1; \
+    fi
+
+ARG LISK_VERSION
+ARG LISK_MIN_VERSION=${LISK_VERSION}
+ENV LISK_VERSION ${LISK_VERSION}
+ENV LISK_MIN_VERSION ${LISK_MIN_VERSION}
+
```
Slamper: Is there a reason you are not adding the version specific seed peers here using 
```
ENV LISK_PEERS_LIST_1 "94.237.41.99:5001"

ENV LISK_PEERS_LIST_2 "209.50.52.217:5001"

ENV LISK_PEERS_LIST_3 "94.237.26.150:5001"

ENV LISK_PEERS_LIST_4 "83.136.249.102:5001"

ENV LISK_PEERS_LIST_5 "94.237.65.179:5001"
```
Like here: https://github.com/LiskHQ/lisk-docker/commit/c8b49ff761fd8980d08808c299ebc9e158bbbd8c#diff-ed9d43ac74c2d8e77f8497b190a5ab5aR23

For ease of release purposes the version specific nethash and seed peers should be set in this list, shouldn't they ?

hendrikhofstadt (changes requested):
Adding a peer list to the dockerfile and nethash should be required to create a straightforward build and release process.

fchavant:
@Slamper 

> Is there a reason you are still using the full ubuntu image instead of the much smaller alpine one ?

I am concerned the packages that build native code (e.g. sodium) might not work (properly) -- I just haven't had time to look into it.

> Also will you completely remove liskhq/lisk-docker and keep the scripts in this repo for better managebility (as we also discussed a submodule approach)

In a first step we'll probably keep the `docker-compose` based examples in the `lisk-docker` repository.

Dockerfile
fchavant: If we add see nodes here then this `Dockerfile` becomes specific to a network (betanet given the peers you listed)

docker_files/etc/confd/templates/config.json.tmpl
```diff
@@ -0,0 +1,142 @@
...
+			},
+			"cors": {
+				"origin": "{{getv "/lil/api/options/cors/origin" "*"}}",
+				"methods": ["GET", "POST", "PUT"]
+			}
+		}
+	},
+	"peers": {
+		"enabled": {{getv "/lisk/peers/enabled" "true"}},
+		"list": [
```
diego-G: Is it possible to insert another variable which it'd read a file having the peers list? We need to bear in mind the goal of having an network agnostic template.

diego-G (changes requested):
How about the Docker Cloud check? Is this going to be mandatory for every PR or is optional?

docker_files/etc/confd/templates/config.json.tmpl
fchavant: If such a file were to be introduced I suppose we could. Another option would be to add an `ARG` and set it at build time.

diego-G: As discussed, we will add a new file containing the list of peers when we tackle #398.

Dockerfile
Slamper: Ah I thought you might want to do this because there are other network specific files on the repo and to maintain compatibility with the old Docker images. But since the Docker image is using the "next" version they are breaking anyway and the proposed adding of an extra file for the peers list is a very good idea.

Slamper:
@fchavant
The PR really looks good.

Another thing about nethash and seed peers. I think it's great that this is decoupled from the repo, which is a good step towards #398. Will you then add the required config variables to the docker-compose example.

This could be a good chance to you an .env-file for docker-compose which is network specific so the docker-compose.yml is not bloated and we have more decoupling.

About the Alpine image. I agree that there should be some testing before trying it out.
I tested a full Alpine version using confd some time ago and it worked just perfectly and had a much smaller size (about 70MB).

I can also open a PR with the adapted dockerfile so you can save time on it and request changes if you think any are necessary.

hendrikhofstadt (approved):

Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+RUN groupadd --gid 1100 lisk && \
+    useradd --create-home --home-dir /home/lisk --shell /bin/bash --uid 1100 --gid 1100 lisk
+COPY . /home/lisk/lisk/
+RUN chown lisk:lisk --recursive /home/lisk/lisk
+
+USER lisk
+WORKDIR /home/lisk/lisk
+
+RUN npm install
+RUN ./node_modules/.bin/grunt release
```
Slamper: What's the purpose of doing grunt release. I see that it adds files that mark the version which can help identify the version but it also creates a packaged tarball and copies it to a release directory creating unnecessary bulk for the docker image.

If the build and revision files are necessary grunt build/revision should be enough and not create unnecessary files.

Slamper:
@fchavant If you want to test the alpine image, here is a version of my alpine Dockerfile that I adapted with your changes: https://gist.github.com/Slamper/f96105fd4a130b15fd301c4315956fcf

It also omits the chown and uses COPY --chown instead which makes the build much quicker.

Dockerfile
fchavant: @diego-G can you chime in on this?

fchavant:
@Slamper thanks; I'll have a look as soon as everything is done here. The reason I am not using the `--chown` flag is that it's not supported in docker cloud yet (it's running docker version `17.06.2-ee-6`)

docker_files/home/lisk/run.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+confd -backend env -onetime
+node app.js
```
nazarhussain: Why we are not using `lisk.sh` script here, if possible we should use it for consistency across distribution platforms.

nazarhussain: Normally I noticed `entrypoint.sh` as startup file in different docker images, should we not follow that naming convention.

Dockerfile
diego-G: @fchavant I agree we don't need grunt release in the docker image. Regarding the other comment, I prefer to keep `npm install` without `--production` flag. But It is true we should use `NODE_ENV` environment variable to signalise what kind of installation we want in the image. By default `npm` listens it, not installing devDependencies if it is set to `production`.

docker_files/home/lisk/run.sh
fchavant: We are not using `lisk.sh` here because 
- it is meant to manage an all-in-one installation which this is not and
- it is meant to be run interactively: it starts processes and then returns which is not what the PID 1 of a container should do

fchavant:
@diego-G the `ci/dockercloud` check will only run for tags, not PR (except this one)

diego-G (approved):

Dockerfile
```diff
@@ -0,0 +1,53 @@
+FROM node:6 AS builder
+
+ENV ENV_NODE=production
```
nazarhussain: Convention is to use `NODE_ENV`, many libraries use this convention to check environment variable for production use case. So better to rename it.

Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+
+RUN groupadd --gid 1100 lisk && \
```
nazarhussain: I can't find any reference for using `1100` gid specific to distro. Why we are hardcoding the group id. What if some dependent docker container use the same group id? 

Would it not be nice to add group only with name and then get its id afterwards. This may be helpful `getent group lisk`

Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+    useradd --create-home --home-dir /home/lisk --shell /bin/bash --uid 1100 --gid 1100 lisk
```
nazarhussain: Same applies to `uid` here.

Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+# As of May 2018 cloud.docker.com runs docker 17.06.1-ce
+# however version 17.12 is required to use the chown flag
+COPY . /home/lisk/lisk/
```
nazarhussain: We shouldn't just copy every thing to docker. First run `grunt release` and copy the compiled source code. You may not find any noticeable difference, but any future change in Gruntfile may break the docker. So better to use compile release after grunt.

Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+RUN chown lisk:lisk --recursive /home/lisk/lisk
+
+USER lisk
+WORKDIR /home/lisk/lisk
+
+RUN npm install
+
+
+FROM node:6
```
nazarhussain: Duplicate line.

Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+
+ENV CONFD_VERSION 0.16.0
+ENV CONFD_SHA256 255d2559f3824dd64df059bdc533fd6b697c070db603c76aaf8d1d5e6b0cc334
+ENV ENV_NODE=production
+
+RUN groupadd --gid 1100 lisk && \
+    useradd --create-home --home-dir /home/lisk --shell /bin/bash --uid 1100 --gid 1100 lisk
```
nazarhussain: Duplicate line.

Dockerfile
```diff
@@ -0,0 +1,53 @@
...
+WORKDIR /home/lisk/lisk
+
+RUN npm install
+
+
+FROM node:6
+
+ENV CONFD_VERSION 0.16.0
+ENV CONFD_SHA256 255d2559f3824dd64df059bdc533fd6b697c070db603c76aaf8d1d5e6b0cc334
+ENV ENV_NODE=production
```
nazarhussain: Duplicate line.

docker_files/etc/confd/templates/config.json.tmpl
```diff
@@ -0,0 +1,142 @@
+{
+	"wsPort": {{getv "/lisk/wsport" "8001"}},
+	"httpPort": {{getv "/lisk/httpport" "8000"}},
+	"address": "{{getv "/lisk/address" "0.0.0.0"}}",
+	"version": "{{getenv "LISK_VERSION"}}",
+	"minVersion": "{{getenv "LISK_MIN_VERSION"}}",
```
nazarhussain: If these **env variables** meant to run container with different min values, then I believe its not the right case. Once the docker container is built, its built for a specific version, which is bound to a specific minVersion. User should not be allowed to change those on runtime.

docker_files/etc/confd/templates/config.json.tmpl
```diff
@@ -0,0 +1,142 @@
...
+	"fileLogLevel": "{{getv "/lisk/fileloglevel" "info"}}",
+	"logFileName": "{{getv "/lisk/logfilename" "logs/lisk.log"}}",
+	"consoleLogLevel": "{{getv "/lisk/consoleloglevel" "error"}}",
+	"trustProxy": {{getv "/lisk/trustproxy" "false"}},
+	"topAccounts": {{getv "/lisk/topaccounts" "false"}},
+	"cacheEnabled": {{getv "/lisk/cacheenabled" "false"}},
+	"wsWorkers": {{getv "/lisk/wsworkers" "1"}},
```
nazarhussain: Why these variables are all small case? We follow either camel or snake can for readability.

docker_files/etc/confd/templates/config.json.tmpl
```diff
@@ -0,0 +1,142 @@
...
+		"poolIdleTimeout": {{getv "/lisk/db/poolidletimeout" "30000"}},
+		"reapIntervalMillis": {{getv "/lisk/db/reapintervalmillis" "1000"}},
+		"logEvents": ["error"],
+		"logFileName": "{{getv "/lisk/db/logfilename" "logs/lisk_db.log"}}"
+	},
+	"redis": {
+		"host": "{{getv "/lisk/redis/host" "127.0.0.1"}}",
+		"port": {{getv "/lisk/redis/port" "6380"}},
+		"db": {{getv "/lisk/redis/db" "0"}},
+		"password": {{getv "/lisk/redis/password" "null"}}
```
nazarhussain: I believe here you want to set `null` instead of `"null"`, consider `null` without quote means no password. But if you quote it will check for that password.

docker_files/etc/confd/templates/config.json.tmpl
nazarhussain: This is the right use case for configurable values. User may want load its own list of peers. Can't understand why its should be done separately. Issue #398 is totally different thing, to support multiple networks in one installation.

docker_files/home/lisk/wait-for-it.sh
```diff
@@ -0,0 +1,177 @@
+#!/usr/bin/env bash
```
nazarhussain: This file is not relevant to lisk docker image, so it should not be included here. Even we can't see any usage of this script file in our Dockerfile.

Its dependency is used by docker-composer under `lisk-docker` repo, so this script should be kept and managed there.

nazarhussain (changes requested):

docker_files/home/lisk/wait-for-it.sh
nazarhussain: I thought on the topic further. If you really want lisk container to wait for db port to be available before running. Then use this script inside `run.sh` not inside the docker-compose. This will make sure that our image will always have this behavior. And then it also makes sense to make this script file part of this repo.

Dockerfile
fchavant: The GID is not distribution specific. We chose `1100` arbitrarily.

> What if some dependent docker container use the same group id?
That would be no problem in a different container. I you mean a container that mounts a volume from this one then that'd allow having the "same" lisk user (same UID and GID).

Dockerfile
fchavant: This is the  build container, we must copy everything in order to be able to run `npm install` or `grunt release`. Regarding the latter I was told to remove it; @diego-G can you please confirm once more?

Dockerfile
fchavant: This is a different container (the final one, not the build container)

Dockerfile
fchavant: Not duplicated: different container, see above.

Dockerfile
fchavant: Ditto.

docker_files/home/lisk/wait-for-it.sh
fchavant: That is exactly what I plan on doing, like so:
https://github.com/LiskHQ/lisk-docker/blob/2.1.0/examples/betanet/docker-compose.yml#L15

docker_files/etc/confd/templates/config.json.tmpl
fchavant: Indeed, good catch.

fchavant: This will output `null` not `"null"`: notice there are not double quotes around the expression in double curly braces (compare to `host` for instance)

docker_files/etc/confd/templates/config.json.tmpl
fchavant: > Once the docker container is built, its built for a specific version, which is bound to a specific minVersion.
That is exactly what will happen, those environment variables will come from the docker image (see https://github.com/LiskHQ/lisk/blob/83de84142be1c7603941278be0d51f0a6c833abb/Dockerfile#L44)

docker_files/etc/confd/templates/config.json.tmpl
fchavant: `confd` will convert those depeding on the chosen backend: environment variables in our case. That means `/lisk/wsworker` will become `LISK_WSWORKER` and capitalization will get lost.

Dockerfile
nazarhussain: @fchavant @diego-G I am sure we need to run `grunt release` and copy only the archive files to container. 

@fchavant We can run `npm install` and `grunt release` on host machine, where we are actually building container. I think we don't need a seperate container just to run `grunt release`.

docker_files/home/lisk/wait-for-it.sh
nazarhussain: No you are running the script in docker-compose.

Please do it in `run.sh` and in compose start the container. The `run.sh` file should have usage of `wait-for-it.sh` to wait for db to appear online and then run `node app.js`

fchavant: If we were to call `wait-for-it.sh` in `run.sh` then
- it would not be optional any more and
- passing parameters (host and port of the db to wait for) would be akward.

I think it makes sense to keep it where optional and as it is right now (`wait-for-it.sh` waiting for the DB and then calling `run.sh`)

Dockerfile
fchavant: > I think we don't need a seperate container just to run grunt release.
It is a way to have a clean build environment w/o polluting the final containers with build time dependencies.

> I am sure we need to run grunt release and copy only the archive files to container.
What would be the advantage in using the tarball `grunt` creates? It's meant to be consumed by lisk-build's `build.sh` script which makes sense for all-in-one release but not docker.

nazarhussain: @fchavant Nothing will be polluted, if you run `grunt release` on host machine and copy the archive build file to container. It will not contain `node_modules` directory. So you will do a clean `npm install` in container.

docker_files/home/lisk/wait-for-it.sh
nazarhussain: If its an optional then don't copy it in container through docker file. 

That's clearly seems a utility script used in docker-compose, then keep it there, mount the utility scripts volume to container and run it. 

If something not required or not related to an image, don't copy it.

fchavant: By optional I mean one can use it but one does not have to.
To give this possibility the script needs to be in the image though (the lisk container must wait for the postgres container; having it somewhere else would not work.

nazarhussain: > the lisk container must wait for the postgres container
Then make it part of `run.sh`

Having a utility script in image, but not using it somewhere else (composer) does not make sense to me.

fchavant: If we make it part of `run.sh` the it becomes very much non optional. Also it would make passing parameters hard; see above.

> Having a utility script in image, but not using it somewhere else (composer) does not make sense to me.

We have a script in the image, it can be used in one's `docker-compose.yml` or by swarm or rancher or whatever. Or one may choose not to use it. That's why I call it "optional".

nazarhussain: @fchavant Ok please document it somewhere our docker image have this script and the usage of this script.

nazarhussain (approved):
Please document it somewhere our docker image have `waitf-for-it` script and its usage.

Issue #2061, opened 2018-05-25
Always select a non-lower peer to sync with


SargeKhan:
### Expected behavior
During sync process, a node should never try syncing with peers which have a lower height. Because we know that they will not have any new blocks (if our information about their status is not outdated)
### Actual behavior
We filter peers based on height in the beginning of the sync process, so if we loaded blocks in the first step, our peers might include peers with height less than our height. Which is against node's interest.  
### Steps to reproduce
Investigate logs: 
```
] 2018-05-22 21:21:35 | Starting sync
[dbg] 2018-05-22 21:21:35 | Establishing broadhash consensus before sync
[inf] 2018-05-22 21:21:35 | Broadhash consensus now 36 %
[...]

PR #2065, opened 2018-05-28
Always select a non-lower peer to sync with

MaciejBaj:
### What was the problem?
It is possible to sync with peers below our own blocks height. It could occur after successful blocks application. Our own height had increased, the peers list received from `findGoodPeers` function doesn't reflect the update though. It is possible to sync with Peer that is below our own height in one of the following 4 (the worst case scenario) sync rounds.

### How did I fix it?
At the beginning of every of 5 round of sync process, the lower peers are rejected.

### How to test it?
Check if sync process doesn't select the lower peers to sync with.
### Review checklist

* The PR solves #2061
* Linting passes
* Tests pass

### DIFF
--- a/modules/loader.js
+++ b/modules/loader.js
@@ -660,9 +660,8 @@ __private.loadBlocksFromNetwork = function(cb) {
 		async.whilst(
 			() => !loaded && errorCount < 5,
 			next => {
-				const peer =
-					network.peers[Math.floor(Math.random() * network.peers.length)];
 				let lastBlock = modules.blocks.lastBlock.get();
+				const peer = getGoodPeer(lastBlock.height, network.peers);
 
 				function loadBlocks() {
 					__private.blocksToSync = peer.height;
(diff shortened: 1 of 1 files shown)


modules/loader.js
```diff
@@ -660,9 +660,8 @@ __private.loadBlocksFromNetwork = function(cb) {
 		async.whilst(
 			() => !loaded && errorCount < 5,
 			next => {
-				const peer =
-					network.peers[Math.floor(Math.random() * network.peers.length)];
 				let lastBlock = modules.blocks.lastBlock.get();
+				const peer = getGoodPeer(lastBlock.height, network.peers);
```
shuse2: Why are we not using the same `findGoodPeers` function that we already have?

MaciejBaj: Reusing `findGoodPeers` adds the calculation overhead to that issue. For keeping the responsibility for selecting good peers I agree it makes sense. 

I will reuse it then.

PR #2070, opened 2018-05-29
Nightly builds changes - Closes #2027

### LINKED ISSUE #2027: Nightly builds issues
### Expected behavior

Clean and proper versioned nightly builds

### Actual behavior

- Nightly tarballs are being generated from the 1.0.0 branch, but with 0.9.8 versioning that comes from config.json, probably related with https://github.com/LiskHQ/lisk/issues/2019

- The libraries imported from https://github.com/LiskHQ/lisk-jenkins make use of `/tmp` which is not the standard way for jenkins, leaving all the workspaces behind.
 
- The imported library is hardcoded to `lisk-jenkins@nh-lisk-build` which is not a master or default branch, it should be merged in `lisk-jenkins` first.

- Build branch hardcoded to 1.0.0

### Steps to reproduce

See Jenkins logs of nightly builds

All lisk branches with `Jenkinsfile.nightly-build`

abelboldu:
### What was the problem?

The current nightly build pipeline uses `nh-lisk-build` branch for libs and it should use `master`.
Also, the bucket to push the tarballs has changed to `lisk-releases`.

### How did I fix it?

Changed the branch and the bucket. Also some indentation.

### How to test it?

Already tested the branch in jenkins job `lisk-nightly-build`

### Review checklist

* The PR solves #2027 
* Tests pass

### DIFF
--- a/Jenkinsfile.nightly-build
+++ b/Jenkinsfile.nightly-build
@@ -14,7 +14,11 @@
 
 import java.text.SimpleDateFormat
 
-properties([pipelineTriggers([cron('15 2 * * *')])])
+properties([pipelineTriggers([cron('15 2 * * *')]),
+    parameters([
+    string(defaultValue: "dev", description: 'Network', name: 'NETWORK'),
+   ])
+])
 
 def dateFormat = new SimpleDateFormat("yyyyMMdd")
 def date = new Date()
@@ -32,18 +36,18 @@ pipeline {
         skipDefaultCheckout true
     }
     stages {
-        stage("Prepare workspace") {
+    	stage("Prepare workspace") {
             steps {
-                library identifier: 'lisk-jenkins@ab-lisk-build', retriever: modernSCM(
-                [$class: 'GitSCMSource',
-                remote: 'https://github.com/LiskHQ/lisk-jenkins.git'])
+            	library identifier: 'lisk-jenkins@master', retriever: modernSCM(
+                    [$class: 'GitSCMSource',
+                    remote: 'https://github.com/LiskHQ/lisk-jenkins.git'])
             }
-        }
+    	}
         stage('Build release') {
             steps {
-                script {
-                    def release = liskBuild(branch: "${env.BRANCH_NAME}", network: "dev")
-                    env.VERSION =  release.version
+            	script {
+                    def release = liskBuild(branch: env.BRANCH_NAME, network: params.NETWORK)
+                    env.VERSION = release.version
                     env.RELEASE_FILE =  release.file
                     env.DESTINATION_FILE = "lisk-${env.VERSION}-${timestamp}-Linux-x86_64.tar.gz"
                 }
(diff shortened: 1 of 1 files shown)


Jenkinsfile.nightly-build
```diff
@@ -33,19 +33,19 @@ pipeline {
...
+            }
     	}
         stage('Build release') {
             steps {
             	script {
-					def release = liskBuild(branch: '1.0.0', network: 'dev')
-                 	env.VERSION =  release.version
-                 	env.RELEASE_FILE =  release.file
-                 	env.DESTINATION_FILE = "lisk-${env.VERSION}-${timestamp}-Linux-x86_64.tar.gz"
+                    def release = liskBuild(branch: "${env.BRANCH_NAME}", network: 'dev')
```
nazarhussain: @abelboldu Would it be `params.BRANCH_NAME` or `env.BRANCH_NAME`? Param is much better option so we can run this job with different parameters interactively.

Jenkinsfile.nightly-build
```diff
@@ -33,19 +33,19 @@ pipeline {
...
+            steps {
+            	library identifier: 'lisk-jenkins@master', retriever: modernSCM(
+                    [$class: 'GitSCMSource',
+                    remote: 'https://github.com/LiskHQ/lisk-jenkins.git'])
+            }
     	}
         stage('Build release') {
             steps {
             	script {
-					def release = liskBuild(branch: '1.0.0', network: 'dev')
```
nazarhussain: @abelboldu Why its build against the `dev` net? It should be `main` net.

Jenkinsfile.nightly-build
abelboldu: No, this is a nightly build,  it's automated and we cannot pass any parameter.

Jenkinsfile.nightly-build
abelboldu: I don't think it has any effect anyway, but we are building nightlies from the `devnet-development` branch, so I think the network should be `dev`

Jenkinsfile.nightly-build
nazarhussain: Yes its nightly build job, but still having it parameterized makes more sense. Considering running nightly builds on two different branches for `main` net and `test` net separately.

Jenkinsfile.nightly-build
nazarhussain: If the branch is parameterized or env variable then the network should also be passed in same approach. Instead of hardcoding it in the file.

Jenkinsfile.nightly-build
abelboldu: The approach is to run the job relative on the branch we are running. If we want to run on a different branch we can move the `Jenkinsfile` there

nazarhussain: @abelboldu Jenkins will not be aware that which branch belongs to which network. So we have to pass the network somehow to Jenkins job. In future that will be useful, when our build process can build any branch for any network.

Jenkinsfile.nightly-build
fchavant: 1/ `env.BRANCH_NAME` is the branch Jenkins is currently working on, there is no need for a parameter.
2/ I would argue nightly builds will always be `dev` builds (and not `test` or `main`)

Jenkinsfile.nightly-build
fchavant: `env.BRANCH_NAME` is the branch Jenkins is currently working on, there is no need for a parameter.

nazarhussain: I just want not to hardcode the network. Network is based on branches right now, so not necessary that `master` branch is built with `dev` net.

fchavant: The *library* is being pulled from the `master` branch of lisk-jenkins; this has nothing to do with the `master` branch of `LiskHQ/lisk`

abelboldu:
Don't merge, still testing

abelboldu:
Ok, ready to merge

PR #2079, opened 2018-05-31
Remove fee for data - Closes #2033

### LINKED ISSUE #2033: Don't charge a fee for optional data field
### Expected behavior
Optional data field should not cost anything extra, because almost nobody will use it if it does. The fee makes this feature almost useless. 

It also makes worse UX because we have to explain this to the users as they are not used to this from traditional banking. 

### Actual behavior
Optional data field requested in https://github.com/LiskHQ/lisk/issues/26 and added in https://github.com/LiskHQ/lisk/pull/470 adds 0.1 LSK fee on top of the 0.1 LSK fee of type 0 transaction.

I see that the fee is there because of storing additional data on the blockchain, but users already pay 0.1 LSK for the transaction which is IMO more than enough to prevent from spamming the blockchain. If I wanted to spam the blockchain I think that adding two transactions for 0.2 LSK will add more data than 1 transaction with 64 char data field for 0.2 LSK.

Lisk Core 1.0.0

diego-G:
### Review checklist

* The PR solves #2033
* Linting passes
* Tests pass


package.json
```diff
@@ -107,7 +107,7 @@
 		"jsdoc": "=3.4.3",
 		"jsdox": "=0.4.10",
 		"lint-staged": "=6.1.0",
-		"lisk-js": "1.0.0-beta.2",
+		"lisk-elements": "beta",
```
willclarktech: This should be pinned to `1.0.0-beta.4` to avoid unexpected breakages if we release another beta version.

package.json
```diff
@@ -107,7 +107,7 @@
 		"jsdoc": "=3.4.3",
 		"jsdox": "=0.4.10",
 		"lint-staged": "=6.1.0",
-		"lisk-js": "1.0.0-beta.2",
+		"lisk-elements": "1.0.0-beta.4",
```
SargeKhan: shouldn't there be a '=' before version?

willclarktech: Actually there shouldn't be an `=` before any of them. No prefix means "exact" and `=` doesn't isn't defined in the npm documentation.

willclarktech: See https://docs.npmjs.com/files/package.json#dependencies

PR #2087, opened 2018-06-02
Check int32 range for transaction timestamps - Closes #2088

### LINKED ISSUE #2088: Check INT_32 range for transaction timestamps
### Expected behavior
Check to reject transactions of timestamp above INT_32 max value (2147483648) or below INT_32 min value (-2147483648) needs to be added.

### Actual behavior
Transaction timestamp is defined in Postgres schema is defined as an INT_32 format.

When transaction timestamp is above 2147483648 or below -2147483648, a block that includes it cannot be saved in Database. Inserting the wrong transaction timestamp to database results with error:
```
ERROR:  integer out of range
STATEMENT:  INSERT INTO "trs"("id","blockId","type","timestamp","senderPublicKey","requesterPublicKey","senderId","recipientId","amount","fee","signature","signSignature","signatures") VALUES ('181175095785369468','5488578331239914243',0,-3704634000,'\x6dd24c92d91c0082f5be68f7350d87b7cdf105267543f1f61d3043a5c2d8a00b',null,'3402562013208542942L','3402562013208542942L',1,10000000,'\xcde9a3459b1f5590a9b6f32a5d9c2e85596be9d01ded14fa9d9c5f276a37e2562e40f459c727599323e1ee84435a63316b3a3a50fb3b5d687dc7ea1f1ad9e001',null,null)
ERROR:  insert or update on table "blocks" violates foreign key constraint "blocks_previousBlock_fkey"
DETAIL:  Key (previousBlock)=(5488578331239914243) is not present in table "blocks".
STATEMENT:  INSERT INTO "blocks"("id","version","timestamp","height","previousBlock","numberOfTransactions","totalAmount","totalFee","reward","payloadLength","payloadHash","generatorPublicKey","blockSignature") VALUES ('13977984917448353211',0,63803270,6144655,'5488578331239914243',0,0,0,400000000,0,'\xe3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855','\x6cb825715058d2e821aa4af75fbd0da52181910d9fda90fabe73cd533eeb6acb','\x0b534abcbe640cef4df6a579bbffe00eb62401b9090e6d7f1ef72e4dafaa31bdb8594b85007f7d8ed79d4166835334ca2adb0830a175d6ac39263fe5fcf8950f')
ERROR:  insert or update on table "blocks" violates foreign key constraint "blocks_previousBlock_fkey"
```
Transaction is reflected in `mem_accounts` though. 
The last block value also points into the wrong block:
https://github.com/LiskHQ/lisk/blob/master/modules/blocks.js#L84.

### Steps to reproduce
Send transction with timeout exceeding INT_32 max value.
*

webmaster128:
This should avoid the integer out of range error from postgres in

```
ERROR:  integer out of range
STATEMENT:  INSERT INTO "trs"("id","blockId","type","timestamp","senderPublicKey","requesterPublicKey","senderId","recipientId","amount","fee","signature","signSignature","signatures") VALUES ('181175095785369468','5488578331239914243',0,-3704634000,'\x6dd24c92d91c0082f5be68f7350d87b7cdf105267543f1f61d3043a5c2d8a00b',null,'3402562013208542942L','3402562013208542942L',1,10000000,'\xcde9a3459b1f5590a9b6f32a5d9c2e85596be9d01ded14fa9d9c5f276a37e2562e40f459c727599323e1ee84435a63316b3a3a50fb3b5d687dc7ea1f1ad9e001',null,null)
ERROR:  insert or update on table "blocks" violates foreign key constraint "blocks_previousBlock_fkey"
DETAIL:  Key (previousBlock)=(5488578331239914243) is not present in table "blocks".
STATEMENT:  INSERT INTO "blocks"("id","version","timestamp","height","previousBlock","numberOfTransactions","totalAmount","totalFee","reward","payloadLength","payloadHash","generatorPublicKey","blockSignature") VALUES ('13977984917448353211',0,63803270,6144655,'5488578331239914243',0,0,0,400000000,0,'\xe3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855','\x6cb825715058d2e821aa4af75fbd0da52181910d9fda90fabe73cd533eeb6acb','\x0b534abcbe640cef4df6a579bbffe00eb62401b9090e6d7f1ef72e4dafaa31bdb8594b85007f7d8ed79d4166835334ca2adb0830a175d6ac39263fe5fcf8950f')
ERROR:  insert or update on table "blocks" violates foreign key constraint "blocks_previousBlock_fkey"
```

-------

**Update** As you can see in the first INSERT INTO statement of the logs above, there is -3704634000 in the 4th column (timestamp). This value is a valid JavaScript number and within the [safe integer range](https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Number/MIN_SAFE_INTEGER), i.e. the number can be represented without rounding errors in JavaScript. However, the value should be stored in a Postgres database column of type [integer](https://www.postgresql.org/docs/9.6/static/datatype-numeric.html), which is signed and of size 4 bytes = 32 bit. The signed 32 range is -2147483648 to +2147483647. Since -3704634000 < -2147483648, the number cannot be stored in the database.

### DIFF
--- a/test/unit/logic/transaction.js
+++ b/test/unit/logic/transaction.js
@@ -679,7 +679,7 @@ describe('transaction', function () {
 			});
 		});
 
-		it('should return error on timestamp smaller than the int32 range', function (done) {
+		it('should return error on timestamp below the int32 range', function (done) {
 			var trs = _.cloneDeep(validTransaction);
 			trs.timestamp = -2147483648 - 1;
 			delete trs.signature;
@@ -690,7 +690,7 @@ describe('transaction', function () {
 			});
 		});
 
-		it('should return error on timestamp bigger than the int32 range', function (done) {
+		it('should return error on timestamp above the int32 range', function (done) {
 			var trs = _.cloneDeep(validTransaction);
 			trs.timestamp = 2147483647 + 1;
 			delete trs.signature;


webmaster128:
@vekexasia right, but there is existing data in the blockchain with timestamp < 0 and I am afraid to invalidate the existing blockchain when restricting this too much.

vekexasia:
I think an exception just for txid 10204034354762287439 should be made code side.

webmaster128:
@vekexasia I agree. But I don't want to introduce a new exception type as part of an emergency patch. Your suggestion could and should be done later on. For now, the blockchain is known to work with negative timestamps as well

Nazgolze:
@vekexasia you beat me to it :), I'm partial to that approach as well

Nazgolze:
@webmaster128 makes a good point

xunga:
Looks good to me...

vekexasia:
Actually I think the real fix for this would be moving this:
https://github.com/LiskHQ/lisk/blob/development/modules/blocks/chain.js#L432

to line 460 & to line 476.  (Moving later to a better implementation).

Atomic block processing implemented by @4miners should already keep the database clean. The problem here was that `node` lastBlock variable was being updated even if  the database transaction did fail.

webmaster128:
@vekexasia interesting. I am not familiar with this part of the code. But wouldn't that keep the transaction being valid and thus remaining in the transaction pool?

Gr33nDrag0n69:
feel like the 2000 year bug.
Timestamp should be an unsigned int32 and to be safe for lisk future (many many years) maybe it should be a double unsigned (didnt made the calculation)

vekexasia:
@webmaster128 the tx caused this to happen. So your fix is still more than welcome. But there are many reasons why this could happen again. (temp db disconnections, disk low on storage, other missing tx logic checks (or even blocks).

Since the db will always be clean thanks to atomic block processing - memory variables should reflect db state.

webmaster128:
> Timestamp should be an unsigned int32 and to be safe for lisk future (many many years)

Lisk shifted the epoch from 1970-01-01 to something in 2016 in order to be able to use 32 bit ints for many many years. It will be fine for the foreseeable future.

4miners:
Both approach are valid:
- just check for `integer` range is fine, as it's constant with database layer
- not allowing timestamp below `0` and add exception for existing transactions (one on mainnet) looks also fine

@Gr33nDrag0n69 We cannot do that, as PostgreSQL is not supporting unsigned types - https://www.postgresql.org/docs/9.6/static/datatype-numeric.html

Gr33nDrag0n69:
about just adding an exception, i think it's abad idea since if the tx was from a malicious user, nothing prevent someone to just send another bad tx

webmaster128:
> about just adding an exception, i think it's a bad idea since if the tx was from a malicious user, nothing prevent someone to just send another bad tx

The idea is to disallow timestamps < 0 entirely and add the existing transaction [10204034354762287439](https://explorer.lisk.io/tx/10204034354762287439) as an exception

Gr33nDrag0n69:
@4miners thanks for the info about postgres not supporting unsigned
@webmaster128 I would be curious to know how many years the Int32 will work ... Because between you and me, when lisk is bigger with many sidechains, i don't think it will be time to address that. I know it's not a priority for now but maybe if only good for a couple years it should be addressed at least before sdk is out ...

webmaster128:
@Gr33nDrag0n69 the datatype will work until [2038](https://en.wikipedia.org/wiki/Year_2038_problem) +46 = 2084

Gr33nDrag0n69:
Fine with me, I'll be dead by then

irebel1:
Lisk has 44 devs and one CTO who was the IT guy at a bike shop and no one picked up on this?

webmaster128:
Unittest added. Please have a look at that too. Thanks!

logic/transaction.js
```diff
@@ -583,6 +583,9 @@ Transaction.prototype.verify = function (trs, sender, requester, checkExists, cb
 	}
 
 	// Check timestamp
+	if (trs.timestamp < -2147483648 || trs.timestamp > 2147483647) {
```
MaciejBaj: Assign as constant

test/unit/logic/transaction.js
```diff
@@ -679,6 +679,17 @@ describe('transaction', function () {
 			});
 		});
 
+		it('should return error on timestamp out of int32 range', function (done) {
+			var trs = _.cloneDeep(validTransaction);
+			trs.timestamp = -2147483648 - 1;
+			delete trs.signature;
+			trs.signature = transaction.sign(senderKeypair, trs);
+			transaction.verify(trs, validSender, null, true, function (err) {
+				expect(err).to.include('Invalid transaction timestamp');
```
4miners: Please expect exact message, part of it it's not reliable.

test/unit/logic/transaction.js
```diff
@@ -679,6 +679,28 @@ describe('transaction', function () {
...
+			});
+		});
+
+		it('should return error on timestamp bigger than the int32 range', function (done) {
+			var trs = _.cloneDeep(validTransaction);
+			trs.timestamp = 2147483647 + 1;
+			delete trs.signature;
+			trs.signature = transaction.sign(senderKeypair, trs);
+			transaction.verify(trs, validSender, null, true, function (err) {
+				expect(err).to.include('Invalid transaction timestamp');
```
4miners: same here

test/unit/logic/transaction.js
webmaster128: done

test/unit/logic/transaction.js
```diff
@@ -679,6 +679,17 @@ describe('transaction', function () {
 			});
 		});
 
+		it('should return error on timestamp out of int32 range', function (done) {
```
MaciejBaj: below INT_32 range

MaciejBaj: Also the second test checking the positive range exceed, 2147483648 + 1;

webmaster128: > Also the second test checking the positive range exceed, 2147483648 + 1;

Already included :)

logic/transaction.js
webmaster128: Where should this constant be stored?

MaciejBaj: `var MAX_INT_32 = 2147483648` in the line 416 logic/transction.js

test/unit/logic/transaction.js
```diff
@@ -679,13 +679,35 @@ describe('transaction', function () {
...
+			trs.timestamp = -2147483648 - 1;
+			delete trs.signature;
+			trs.signature = transaction.sign(senderKeypair, trs);
+			transaction.verify(trs, validSender, null, true, function (err) {
+				expect(err).to.include('Invalid transaction timestamp. Timestamp is not in the int32 range');
+				done();
+			});
+		});
+
+		it('should return error on timestamp above the int32 range', function (done) {
```
MaciejBaj: Thanks

webmaster128:
Thanks everyone for all the reviews and feedback. In such a high pressure situation, no singe developer's eyes and brain are sufficient to do a patch like that alone.

xunga:
@MaciejBaj and @karmacoma is a reward/bounty planned for @webmaster128 ? 

The community expressed their gratitude to Simon for his immediate response in offering the hotfix solution and for his time spent on auditing the code in the past, but I believe that an incentive should also come from LiskHQ in order to promote and incentivise the spirit of white hat. What do you think on this?

nazarhussain:
@xunga Reward and Bounty both are totally different things. I also feel time spent by @webmaster128 and specially his immediate attention on serious issues should be acknowledged by some reward.

xunga:
@nazarhussain totally agree on the term :)

PR #2095, opened 2018-06-05
Blacklist peers from config - Closes #2010

### LINKED ISSUE #2010: Port the feature of blackList for P2P layer to web sockets
### Expected behaviour
We can blacklist any IP to connect on P2P layer. It is specified in configuration. 

https://github.com/LiskHQ/lisk/blob/20d0bbab7bc5874b802344a1c31a8cdabea7dcd9/config.json#L92-L94

Previously in `0.9.x` versions it was used in `/peer*` endpoints for http APIs. But its never been ported to websockets. 

### Actual behavior
Currently its been applied on HTTP APIs which is supposed not to be P2P layer communication. HTTP Layer already have a ip whitelist feature. 

`1.0.0`

diego-G:
### How did I fix it?

Peers existing in the config file are banned depending on two situations:

- new peers : we ban them during the handshake
- known peers: ban is performed on the functions `insertSeeds` `dbLoad` & `updatePeerStatus`

### How to test it?

`npm test -- mocha:untagged:integration`

### Review checklist

* The PR solves #2010
* Linting passes
* Tests pass


modules/peers.js
```diff
@@ -71,6 +71,8 @@ class Peers {
 		self.consensus = scope.config.forging.force ? 100 : 0;
 		self.broadhashConsensusCalculationInterval =
 			scope.config.peers.options.broadhashConsensusCalculationInterval;
+		self.blackListPeers = scope.config.peers.access.blackList;
```
nazarhussain: Either it should be `peersBlackList` or `blackListedPeers`

modules/system.js
```diff
@@ -56,6 +56,7 @@ class System {
 				nethash: scope.config.nethash,
 				minVersion: scope.config.minVersion,
 				nonce: scope.config.nonce,
+				blackListPeers: scope.config.blackListPeers,
```
nazarhussain: Why do we need this list in `system` module. Its responsibility of `peers` module and only kept there.

modules/system.js
```diff
@@ -245,6 +247,17 @@ System.prototype.nonceCompatible = function(nonce) {
...
+/**
+ * Check if the ip exists in the peer blacklist present in the config file.
+ *
+ * @param ip
+ * @returns {boolean}
+ * @todo Add description for the params and the return value
+ */
+System.prototype.isBlacklisted = function(ip) {
+	return __private.blackListPeers.indexOf(ip) > -1;
+};
```
nazarhussain: Same method available to `peers` module, so its duplicate. Please keep it only in `peers` module.

test/integration/scenarios/index.js
```diff
@@ -18,6 +18,7 @@ module.exports = {
 	network: {
 		peers: require('./network/peers'),
 		peerDisconnect: require('./network/peer_disconnect'),
+		peerBlacklist: require('./network/peer_blacklist'),
```
nazarhussain: To keep pragmatic in place. Please rename it to `peersBlackList`. as its a singular list containing plural peers.

test/integration/scenarios/network/peer_blacklist.js
```diff
@@ -0,0 +1,191 @@
...
+
+			it(`there should be ${expectedOutgoingConnections} established connections from 500[0-9] ports`, done => {
+				utils.getEstablishedConnections(
+					Array.from(wsPorts),
+					(err, numOfConnections) => {
+						if (err) {
+							return done(err);
+						}
+
+						if ((numOfConnections - 20) <= expectedOutgoingConnections) {
```
nazarhussain: Please add a comment line here stating why you subtracted 20 from number of connections.

app.js
```diff
@@ -502,6 +502,7 @@ d.run(() => {
 						nethash: scope.config.nethash,
 						port: scope.config.wsPort,
 						nonce: scope.config.nonce,
+						blackListPeers: scope.config.peers.access.blackList,
```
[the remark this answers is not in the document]
nazarhussain: Noticed that in worker you only check for the error code `BLACKLISTED_PEER` so no need to pass this list to worker.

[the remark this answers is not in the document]
jondubois: @nazarhussain That error code comes from `handshake.js` which relies on the `isBlacklisted(...)` function which needs the `blackListPeers` value. So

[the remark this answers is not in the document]
jondubois: About naming, for me the current name is OK. I think it makes it easier to search if the keyword 'blackList' is part of it.

modules/peers.js
```diff
@@ -336,6 +366,11 @@ __private.dbLoad = function(cb) {
...
 				rows,
 				(peer, eachCb) => {
+					// Ban peer if it is presented in the array of black listed peers
+					if (__private.isBlacklisted(peer.ip)) {
+						peer.state = Peer.STATE.BANNED;
+					}
+
 					peer = library.logic.peers.create(peer);
 					if (library.logic.peers.upsert(peer, true) !== true) {
 						return setImmediate(eachCb);
```
jondubois: Maybe we should change `peer.state > 0` below (on line 378 and also line 838) to `peer.state !== Peer.STATE.BANNED` to be more clear/explicit.

modules/peers.js
```diff
@@ -264,13 +277,21 @@ __private.updatePeerStatus = function(err, status, peer) {
 		);
 		library.logic.peers.remove(peer);
 	} else {
+		let state;
+		// Ban peer if it is presented in the array of black listed peers
+		if (__private.isBlacklisted(peer.ip)) {
+			state = Peer.STATE.BANNED;
```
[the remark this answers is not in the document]
jondubois: At the unit level (blackbox approach; ignoring all the logic above), I think that it's OK if `updatePeerStatus` can handle peers in any state (BANNED, CONNECTED, DISCONNECTED) but then we should be consistent. Right now if peer is banned then state will stay banned (good), if peer is connected then state will stay connected (good) but if peer is disconnected then we will change the state to connected (bad/inconsistent).

modules/system.js
jondubois: Yes we should have or or the other.

modules/peers.js
```diff
@@ -236,6 +238,17 @@ __private.getMatched = function(test, peers) {
...
 
+/**
+ * Check if the ip exists in the peer blacklist coming from config file.
+ *
+ * @param ip
+ * @returns {boolean}
+ * @todo Add description for the params and the return value
+ */
+__private.isBlacklisted = function(ip) {
+	return self.blackListPeers.indexOf(ip) > -1;
```
ManuGowda: since we are using latest javascript we can start using self.blackListPeers.includes(ip);

workers_controller.js
```diff
@@ -109,6 +109,10 @@ SCWorker.create({
 								// Because of WebSocket protocol handshake restrictions, we can't call next(err) here because the
 								// error will not be passed to the client. So we can attach the error to the request and disconnect later during the SC 'handshake' event.
 								req.failedHeadersValidationError = err;
+								// When ip is blacklisted, error is thrown as soon as possible.
+								if (err.code == failureCodes.BLACKLISTED_PEER) {
```
ManuGowda: let's keep strict check condition with `===` check

modules/system.js
```diff
@@ -245,6 +247,17 @@ System.prototype.nonceCompatible = function(nonce) {
...
 
+/**
+ * Check if the ip exists in the peer blacklist present in the config file.
+ *
+ * @param ip
+ * @returns {boolean}
+ * @todo Add description for the params and the return value
+ */
+System.prototype.isBlacklisted = function(ip) {
+	return __private.blackListPeers.indexOf(ip) > -1;
```
ManuGowda: Same as Above comment to use `includes`

modules/peers.js
diego-G: We should do this in a wider way. I see many other cases in the code.

modules/peers.js
diego-G: Done

workers_controller.js
shuse2: For consistency we should keep strict check

modules/peers.js
```diff
@@ -291,20 +312,29 @@ __private.insertSeeds = function(cb) {
...
-				if (!err) {
-					updated += 1;
-				} else {
-					library.logger.trace(`Ping peer failed: ${peer.string}`, err);
-				}
-				return setImmediate(eachCb);
-			});
+
+			// Continue if peer it is not blacklisted nor banned
+			if (peer.state != Peer.STATE.BANNED) {
```
shuse2: we should keep the strict checking.
there is no callback called when it's banned

modules/peers.js
```diff
@@ -335,6 +335,8 @@ __private.insertSeeds = function(cb) {
 					return setImmediate(eachCb);
 				});
 			}
+
+			return setImmediate(eachCb);
```
jondubois: If peer is not banned then the `eachCb` will be called twice (once inside the status callback and also here).

diego-G: Yes, it is in progress guys. Patience...

modules/peers.js
```diff
@@ -291,20 +312,31 @@ __private.insertSeeds = function(cb) {
...
-			peer.rpc.status((err, status) => {
-				__private.updatePeerStatus(err, status, peer);
-				if (!err) {
-					updated += 1;
-				} else {
-					library.logger.trace(`Ping peer failed: ${peer.string}`, err);
-				}
+
+			// Continue if peer it is not blacklisted nor banned
+			if (peer.state != Peer.STATE.BANNED) {
```
shuse2: strict checking...?

diego-G:
@shuse2 two things:
- strict checking should be enforced by Eslint. They are not. why?
- please, do a whole review in once. Giving comments like drops doesn't help the review process
Extra: feel free to add small commits when needed to speed up the process if there are no critical changes

shuse2:
Im pretty sure it's mentioned several times =p
https://github.com/LiskHQ/lisk/pull/2095#discussion_r193972212
https://github.com/LiskHQ/lisk/pull/2095#discussion_r193721839

Also, eslint is not enable because there will be too much changes now, but that doesn't mean we should add more.

Issue #2096, opened 2018-06-06
During syncing, reject blocks received through broadcasts


SargeKhan:
### Expected behavior
Syncing process happens within the balance sequence, so if a node is syncing, it should reject the blocks received from broadcasts immediately without adding them to the balance sequence. 
### Actual behavior
Blocks received from the network through broadcasts are added to the balance sequence. So, if the node is syncing, it can hit the memory limit because blocks received from the network keep piling up. The wrong logic is here: 
```

Process.prototype.onReceiveBlock = function(block) {
	let lastBlock;

	// Execute in sequence via sequence
	library.sequence.add(cb => {
		// When client is not loaded, is syncing or round is ticking
		// Do not receive new blocks as client is not ready
		if (
			!__private.loaded ||
			modules.loader.syncing() ||
			modules.rounds.ticking()
		) {
			library.logger.debug('Client not ready to receive block', block.id);
			return setImmediate(cb);
		}
```
The if condtions should be outside of balance sequence, so that reference to the received block is removed automatically.
### Which version(s) does this affect? (Environment, OS, etc...)
1.0.0

PR #2106, opened 2018-06-07
During syncing, reject blocks received through broadcasts

nazarhussain:
### What was the problem?
Blocks received from the network through broadcasts are added to the balance sequence. So, if the node is syncing, it can hit the memory limit because blocks received from the network keep piling up. 

### How did I fix it?
Reject the block outside the balance sequence. 

### How to test it?
`npx mocha test/unit/modules/blocks/process.js`

### Review checklist

* The PR solves #2096
* Linting passes
* Tests pass

### DIFF
--- a/modules/blocks/process.js
+++ b/modules/blocks/process.js
@@ -628,23 +628,26 @@ __private.validateBlockSlot = function(block, lastBlock, cb) {
  * @todo Add @returns tag
  */
 Process.prototype.onReceiveBlock = function(block) {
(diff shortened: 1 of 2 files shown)


nazarhussain:
@MaciejBaj This PR also introduce a behavioral change. Previously we were rejecting blocks while doing round tick. Based on discussion with @4miners this is not required now. So now we only reject the blocks if not ready or syncing.

PR #2110, opened 2018-06-08
Improve Lisk Timestamps explanation - Closes #2109

### LINKED ISSUE #2109: Improve explanation of timestamps in swagger API introduction
### Expected behaviour
Lisk introduction should give a short and precise description of the used timestamp formats in the Lisk Protocol.

### Actual behaviour
Explanation of the timestamps does not explain the difference between unix timestamps and Lisk Timestamps precisely enough

### Steps to reproduce
Open the Lisk Core API documentation in swagger

1.0.0

Tschakki:
### What was the problem?
Explanation of the timestamps does not explain the difference between unix timestamps and Lisk Timestamps precisely enough

### How did I fix it?
Removed the explanation of unit timestamps and added explanation of Lisk timestamps in relation to Lisk Epoch Time.

### How to test it?
Open the Lisk Core API documentation in swagger

### Review checklist

* The PR solves #2109
* Linting passes
* Tests pass

### DIFF
--- a/schema/swagger.yml
+++ b/schema/swagger.yml
@@ -57,7 +57,7 @@ info:
     ## List of Endpoints
     All possible API endpoints for Lisk Core are listed below.
     Click on an endpoint to show descriptions, details and examples.
-  version: '1.0.28'
+  version: '1.0.29'
   contact:
     email: admin@lisk.io
   license:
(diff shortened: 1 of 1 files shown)


nazarhussain:
@Tschakki The changes are good, but we need to plan how to publish it. As on swagger hub we can't change an already published spec. And for this change I don't think we should bump the version.

Tschakki:
Agreed. We can use this PR for gathering smaller changes to the `swagger.yml`, and update them all-together before publishing the new docs.

Issue #2069, opened 2018-05-28
Forward compatibility for single bidirectional connection


jondubois:
### Expected behavior

In terms of RPCs, our outbound peer connection should be able to behave both as a client (making RPCs to peers and receiving responses) and as a server (handling RPCs from peers and responding to them).

### Actual behavior

Currently, our node requires two bidirectional connections.

One connection is initiated by us (outbound) and the other is initiated by the peer (inbound).

The outbound connection is used by us to invoke RPCs on a peer and for us to receive the matching responses.

The inbound connection is used by us to receive RPCs from a peer and for us to send matching responses.

Sometime after we roll out 1.0, we would like to enforce a single bidirectional connection per peer (makes peer management easier, cheaper and more reliable and will help when scaling up across processes; it will also allow for some other interesting use cases).

The problem now is that each of our two connections (inbound vs outbound) only supports one behaviour or another (either making RPCs or serving them; but not both); therefore, our code is not currently in a position to support a single-bidirectional-connection (without breaking backwards compatibility in the future).

Ideally, we should allow our outbound connection to support serving RPCs as part of 1.0; that way when we roll out support for a single-bidirectional-connection, new nodes will work with the old nodes.

Based on initial investigation, it should be relatively easy to implement.

### Steps to reproduce

Check that a single connection cannot be used to both make and receive RPCs to/from a peer.

### Which version(s) does this affect? (Environment, OS, etc...)

1.0.0

PR #2111, opened 2018-06-08
Allow client socket to behave as an RPC server

jondubois:
### What was the problem?

Peers could not send RPCs on inbound connections.

### How did I fix it?

Allowed the client socket on the peer to act as an RPC server.

### How to test it?

Currently it supports all the RPCs except of `updateMyself` (because it is not currently designed to be invoked directly from the master process).

### Review checklist

* The PR solves #2069
* Linting passes
* Tests pass

### DIFF
--- a/api/ws/rpc/connect.js
+++ b/api/ws/rpc/connect.js
@@ -28,11 +28,13 @@ const wampClient = new WAMPClient(TIMEOUT); // Timeout failed requests after 1 s
 const socketConnections = {};
 
 const connect = (peer, logger) => {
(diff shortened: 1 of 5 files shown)


jondubois:
@shuse2 I'm away next week.
This feature supports all bidirectional RPCs except for `updateMyself`.
If we decide to go ahead and keep `updateMyself` then feel free to add extra commits in this PR to support it.

Feel free to build on top of this PR or raise a separate issue to deal with `updateMyself`.

Issue #1956, opened 2018-04-27
Make PUT /node/status/forging idempotent


webmaster128:
### Expected behavior

[PUT requests are idempotent](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PUT)

### Actual behavior

The current implementation of PUT /node/status/forging is a toggle function. The same request is sent to activate and deactivate forging. This is a problem because you do not know in advance, which action you are doing. It becomes of practical importance when you are thinking of network issues and request retries. Every sane HTTP client would retry a PUT request when it does not get a response. But when a request does not return data, you do not know if the action was performed on the server or not.

Requesting the forging state in advance is no reliable solution also because the state may have changed between the status request and the toggle request. 

### Steps to reproduce

See https://app.swaggerhub.com/apis/LiskHQ/Lisk/1.0.26#/Delegates/updateForgingStatus

### Which version(s) does this affect? (Environment, OS, etc...)

API version 1.0.26

----------------------------

Since PUT is a method to override a resource, there is a simple way to fix this. We already have a `forging` field in the response, which can be used for requests as well.

Enable forging:

```
PUT /node/status/forging
{
  "decryptionKey": "happy tree friends",
  "publicKey": "968ba2fa993ea9dc27ed740da0daf49eddd740dbd7cb1cb4fc5db3a20baf341b",
  "forging": true
}
```
Disable forging:

```
PUT /node/status/forging
{
  "decryptionKey": "happy tree friends",
  "publicKey": "968ba2fa993ea9dc27ed740da0daf49eddd740dbd7cb1cb4fc5db3a20baf341b",
  "forging": false
}
```

A request where `forging` is unset then must result in HTTP 400 Bad Request or maybe HTTP 422 Unprocessable Entity.

nazarhussain:
As per our request validation layer if `forging` is unset while required will return automatically to 400 status code.

webmaster128:
> As per our request validation layer if forging is unset while required will return automatically to 400 status code.

i.e. you just need to mark the field as required?

nazarhussain:
Yes

simonmorgenthaler:
I agree, this API call should be adapted. the suggestion from @webmaster128 looks good to me

PR #2119, opened 2018-06-12
Make PUT /node/status/forging idempotent

nazarhussain:
### What was the problem?
There was `PUT  /node/status/forging` which was not idempotent. And behaving like toggle endpoint. 

### How did I fix it?
Updated endpoint to `PATCH  /node/status/forging`

### How to test it?
Functional HTTP Tests for Node

### Review checklist

* The PR solves #1956
* Linting passes
* Tests pass


nazarhussain:
This PR is needed to be merged and rebased, then tests will pass on current branch.

modules/delegates.js
```diff
@@ -17,6 +17,7 @@
 const crypto = require('crypto');
 const _ = require('lodash');
 const async = require('async');
+const lisk = require('lisk-elements').default;
```
shuse2: I think for 1.0, we decided not to use `lisk-elements` in core, and when we are going to migrate, we should probably do it in other PR.

schema/swagger.yml
```diff
@@ -405,7 +405,7 @@ paths:
           description: Unexpected error
           schema:
             $ref: '#/definitions/UnexpectedError'
-    put:
+    patch:
```
shuse2: I think not necessary to change it to `PATCH` here.
Both `PUT` and `PATCH` make sense

modules/delegates.js
```diff
@@ -642,24 +532,22 @@ __private.loadDelegates = function(cb) {
 					if (!account) {
 						return setImmediate(
 							seriesCb,
-							[
-								'Account with public key:',
-								keypair.publicKey.toString('hex'),
-								'not found',
-							].join(' ')
+							['Account with public key:', keypair.publicKey, 'not found'].join(
```
4miners: use template strings

modules/delegates.js
```diff
@@ -718,45 +606,31 @@ Delegates.prototype.toggleForgingStatus = function(publicKey, password, cb) {
 		);
```
4miners: template strings (line above)

helpers/ed.js
```diff
@@ -52,6 +52,9 @@ ed.makeKeypair = function(hash) {
  * @todo Add description for the params and the return value
  */
 ed.sign = function(hash, privateKey) {
+	if (typeof privateKey === 'string') {
```
4miners: - So now instead of keeping `Buffer` instance in memory we will always create `Buffer` here every `sign` call? Not very efficient.
- Looks like now we're only passing strings here, condition is not needed. Also after removing the condition some tests should fail, because they still testing behavior with `Buffer`, so that condition is only for tests? 😏

modules/delegates.js
4miners: If we want to use `lisk-elements` we should add that to dependencies (we have only in dev dependencies).

modules/delegates.js
```diff
@@ -642,24 +532,22 @@ __private.loadDelegates = function(cb) {
...
-							].join(' ')
+							['Account with public key:', keypair.publicKey, 'not found'].join(
+								' '
+							)
 						);
 					}
 
 					if (account.isDelegate) {
-						__private.keypairs[keypair.publicKey.toString('hex')] = keypair;
+						__private.keypairs[keypair.publicKey] = keypair;
```
4miners: So `keypair.publicKey`mand `keypair.privateKey` are both strings now?

It will not fail here?
https://github.com/LiskHQ/lisk/blob/0daca9dd47f026a9344516a5e3c9b2b74c4fbe70/logic/block.js#L137

4miners:
I don't really like replacing `library.ed.makeKeypair` with `lisk.cryptography.getPrivateAndPublicKeyFromPassphrase`, whatever library is used there (`js-nacl`/`tweetnacl-js`) because `libsodium` is still faster.

helpers/ed.js
4miners: Also we should unit test that.

modules/delegates.js
nazarhussain: Can anyone explain the reason to write duplicate code in 1.0 and where did that decision made? 

Don't we have trust on the `lisk-elements`?

schema/swagger.yml
nazarhussain: Yes its best practice to use the patch here. By convention PUT accept full resource to update. If you need to update few attributes then better to use PATCH.

helpers/ed.js
nazarhussain: Condition is not only for tests. `sodium.crypto_sign_detached` only accepts buffer, so why have to convert it.

modules/delegates.js
nazarhussain: As mentioned in the PR already, the dependency change was already been part of another PR, so I didn't moved that commit here.

modules/delegates.js
nazarhussain: Yes both are hex strings now. I can look into occurrences to fix it.

modules/delegates.js
diego-G: We do but it wasn't planned for now. We will make a full migration of our cryptography towards `lisk-elements` in consequent releases.

helpers/ed.js
4miners: I mean it should be better to convert it to `Buffer` after creating the keypair, so we will keep `Buffer` instances in memory and just pass them here instead of converting them to `Buffer` every time we call `sign`.

modules/delegates.js
4miners: @diego-G Really? What about performance?

nazarhussain:
@4miners I am not convinced that there is some performance issue. Please look at this link https://gist.github.com/jonasfj/af453cc2c569312ac59f Point `lisk-elements` is also our library, its not a third party library. If we are not trusting it to use, then how would community will use it?

modules/delegates.js
diego-G: It is not the time to discuss that @4miners. When the moment arrives, we will measure if the performance affects us for real.

nazarhussain:
@SargeKhan 

1. We are changing those methods with few others also well tested on lisk-elements side. Its our library if we can't trust it then we should drop it. 

2. This is called "Continues Improvement" in software development life cycle. If there is something you can improve, do it right away. Planning every thing for future and doing everything in big chunk is not the modern software development practices. 

3. Already explained PUT is specifically used when you need to update complete resource, if you need to update few attributes then PATCH is used. See https://stackoverflow.com/questions/28459418/rest-api-put-vs-patch-with-real-life-examples If it would have been an in-house project then it does not matter we use PUT/PATCH. But since its a open source project and community is looking on our efforts we should have to align with practices used in industry.

modules/delegates.js
4miners: I'm just worried, because you wrote `We will make a full migration of our cryptography towards lisk-elements in consequent releases.` - that means the decision was already made and I don't even know about it.

diego-G: Using `lisk-elements` in Core not only in the test suites but as a regular dependency, it is not new at all. We are talking about this since ages... Obviously, security it is the main expertise of `lisk-elements`. Are you sure you weren't aware? 
Of course, there is nothing 100% settled until we create issues an assign them to milestone. There is room for discussions if you have any concern.

modules/delegates.js
```diff
@@ -632,7 +523,7 @@ __private.loadDelegates = function(cb) {
 
 			modules.accounts.getAccount(
 				{
-					publicKey: keypair.publicKey.toString('hex'),
+					publicKey: keypair.publicKey,
```
4miners: Shouldn't be `keypair.publicKey.toString('hex')`?

webmaster128:
> And behaving like toggle endpoint.

This is not what you want to have as a user.

4miners:
@diego-G I don't get your comment, `PUT` is also not `SAFE`. Using that endpoint our intention is to change state on the server, so why it should be `SAFE`?

diego-G:
@4miners is not my opinion. I've referred to RFC. Please, read carefully and you will find the answers. We should them read more frequently and we wouldn't have this kind of discussions.

4miners:
@diego-G Ahh, it was a direct quote from RFC, ok. I through you wanted to use `SAFE` method.

webmaster128:
> I'm not an expert regarding to PUT/PATCH, so I don't have opinion on that.

PUT replaces a resource, PATCH overrides some attributes of a resource. Since this updates the forging status for one delegate only, I think PATCH is slightly better.

PR #2127, opened 2018-06-14
Balances in mem_accounts off after snapshotting - Closes #2126

### LINKED ISSUE #2126: Balances in mem_accounts off after snapshotting
### Expected behavior

All balances in mem_accounts should be correct after snapshotting 

### Actual behavior

The delegate's balances are off by a small amount (e.g. 99009 Beddows), which is probably the missing distribution of the round fee.

Here are 3 examples from snapshots of height 34138, 77366 and 162206
https://gist.github.com/webmaster128/1cda2679885f331ac22c97691e12c437

This is a comparison between actual balance and expected balance from the last snapshot
[balances-diff.xlsx](https://github.com/LiskHQ/lisk/files/2097814/balances-diff.xlsx)

### Steps to reproduce

- Snapshot using beta8/beta9
- compare actual and expected balance in mem_accounts

- beta8/beta9

4miners:
### What was the problem?
Rewards for round 1 were applied twice during snapshotting.
### How did I fix it?
Forced sumRound to return `0` fees/rewards and genesis block generator when round is summed just after genesis block (dummy round 1 for electing delegates).
### How to test it?
- Run test suite.
- Run migration and check results manually.
### Review checklist

* The PR solves #2126
* Linting passes
* Tests pass

### DIFF
--- a/db/sql/migrations/updates/20180205000003_create_rounds_rewards_table.sql
+++ b/db/sql/migrations/updates/20180205000003_create_rounds_rewards_table.sql
@@ -39,7 +39,7 @@ CREATE TABLE "rounds_rewards"(
 );
 
 -- Insert record for dummy round 1 when genesis block is in database
-INSERT INTO rounds_rewards SELECT timestamp, 0, 0, 1, "generatorPublicKey" FROM blocks WHERE height = 1;
+INSERT INTO rounds_rewards (timestamp, fees, reward, round, "publicKey") SELECT timestamp, 0, 0, 1, "generatorPublicKey" FROM blocks WHERE height = 1;
 
 -- Compute all rewards for previous rounds and insert them to 'rounds_rewards'
 DO $$
--- a/db/sql/migrations/updates/20180205000003_create_rounds_rewards_table.sql
+++ b/db/sql/migrations/updates/20180205000003_create_rounds_rewards_table.sql
@@ -39,7 +39,7 @@ CREATE TABLE "rounds_rewards"(
 );
 
 -- Insert record for dummy round 1 when genesis block is in database
-INSERT INTO rounds_rewards SELECT timestamp, 0, 0, 1, "publicKey" FROM blocks WHERE height = 1;
+INSERT INTO rounds_rewards SELECT timestamp, 0, 0, 1, "generatorPublicKey" FROM blocks WHERE height = 1;
 
 -- Compute all rewards for previous rounds and insert them to 'rounds_rewards'
 DO $$
--- a/db/sql/migrations/updates/20180205000003_create_rounds_rewards_table.sql
+++ b/db/sql/migrations/updates/20180205000003_create_rounds_rewards_table.sql
@@ -38,6 +38,9 @@ CREATE TABLE "rounds_rewards"(
 	"publicKey" BYTEA  NOT NULL
 );
 
+-- Insert record for dummy round 1 when genesis block is in database
+INSERT INTO rounds_rewards SELECT timestamp, 0, 0, 1, "publicKey" FROM blocks WHERE height = 1;
+
 -- Compute all rewards for previous rounds and insert them to 'rounds_rewards'
 DO $$
 	DECLARE
@@ -66,11 +69,8 @@ DO $$
 					FROM blocks b
 					WHERE ceil(b.height / 101::float)::int = row.round
 				),
-				filtered_round AS (
-					SELECT * FROM round WHERE height > 1
-				),
 				-- Calculating total fees of round
-				fees AS (SELECT sum(fees) AS total, floor(sum(fees) / 101) AS single FROM filtered_round),
+				fees AS (SELECT sum(fees) AS total, floor(sum(fees) / 101) AS single FROM round),
 				-- Get last delegate and timestamp of round's last block
 				last AS (SELECT "publicKey", timestamp FROM filtered_round ORDER BY height DESC LIMIT 1)
 			INSERT INTO rounds_rewards
--- a/modules/rounds.js
+++ b/modules/rounds.js
@@ -387,6 +387,16 @@ __private.getOutsiders = function(scope, cb, tx) {
  * @todo Add description for the params and the return value
  */
 __private.sumRound = function(scope, cb, tx) {
+	// When we need to sum round just after genesis block (height: 1)
+	// - set data manually directly from genesis block object
+	if (scope.block.height === 1) {
+		library.logger.debug(`Summing round - ${scope.round} (genesis block)`);
+		scope.roundFees = scope.block.totalFee;
+		scope.roundRewards = [scope.block.reward];
+		scope.roundDelegates = [scope.block.generatorPublicKey];
+		return setImmediate(cb);
+	}
+
 	library.logger.debug('Summing round', scope.round);
 
 	(tx || library.db).rounds
--- a/modules/rounds.js
+++ b/modules/rounds.js
@@ -388,11 +388,11 @@ __private.getOutsiders = function(scope, cb, tx) {
  */
 __private.sumRound = function(scope, cb, tx) {
 	// When we need to sum round just after genesis block (height: 1)
-	// - set data manually directly from genesis block object
+	// - set data manually to 0, they will be distributed when actual round 1 is summed
 	if (scope.block.height === 1) {
 		library.logger.debug(`Summing round - ${scope.round} (genesis block)`);
-		scope.roundFees = scope.block.totalFee;
-		scope.roundRewards = [scope.block.reward];
+		scope.roundFees = 0;
+		scope.roundRewards = [0];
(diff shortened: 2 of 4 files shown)


test/unit/modules/rounds.js
```diff
@@ -369,29 +369,22 @@ describe('rounds', () => {
...
-
-			after(() => {
-				return stub.restore();
-			});
+		describe('when last block is genesis block', () => {
+			const scope = {
+				round: 1,
+				block: {
+					height: 1,
+					totalFee: 123,
```
MaciejBaj: `const validTotalFee = 123` and so on.

db/sql/migrations/updates/20180205000003_create_rounds_rewards_table.sql
```diff
@@ -38,6 +38,9 @@ CREATE TABLE "rounds_rewards"(
 	"publicKey" BYTEA  NOT NULL
 );
 
+-- Insert record for dummy round 1 when genesis block is in database
+INSERT INTO rounds_rewards SELECT timestamp, 0, 0, 1, "generatorPublicKey" FROM blocks WHERE height = 1;
+
```
nazarhussain: Please change the query to 

```sql
INSERT INTO rounds_rewards (timestamp, fees, reward, round, "publicKey") 
      SELECT timestamp, 0, 0, 1, "generatorPublicKey" FROM blocks WHERE height = 1;
```
That makes more readable to know which values goes to which columns.

test/functional/system/snapshotting.js
```diff
@@ -51,18 +54,32 @@ describe('snapshotting', () => {
 
 		before(() => {
 			// Forge 201 blocks to reach height 202 (genesis block is already tyhere)
```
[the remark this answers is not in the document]
4miners: It's valid, we're still forging 201 blocks here in `before` block, but it's just splitted into few phases.

modules/rounds.js
```diff
@@ -387,6 +387,16 @@ __private.getOutsiders = function(scope, cb, tx) {
  * @todo Add description for the params and the return value
  */
 __private.sumRound = function(scope, cb, tx) {
```
nazarhussain: The `__private.sumRound` is triggered from `tick`. There is a condition. 

```
// Establish if finishing round or not
	scope.finishRound =
		round !== nextRound || (block.height === 1 || block.height === 101);
```

I can't understand why this condition contains `|| block.height === 101`

test/functional/system/snapshotting.js
```diff
@@ -51,18 +54,32 @@ describe('snapshotting', () => {
...
-			return Promise.mapSeries([...Array(201)], () => {
+			return Promise.mapSeries([...Array(99)], () => {
```
nazarhussain: Its a bit confusing you are forging `99` blocks here and comment above states `201` blocks. After `genesisBlock` at height `1` and then rest of `99` blocks we will have 100 blocks, how you calculated those to 201?

nazarhussain: If we see steps just below the total blocks become 

```
1 + 99 + 101 + 101 = 302
```

modules/rounds.js
```diff
@@ -387,6 +387,16 @@ __private.getOutsiders = function(scope, cb, tx) {
...
+		library.logger.debug(`Summing round - ${scope.round} (genesis block)`);
+		scope.roundFees = 0;
+		scope.roundRewards = [0];
+		scope.roundDelegates = [scope.block.generatorPublicKey];
+		return setImmediate(cb);
+	}
+
 	library.logger.debug('Summing round', scope.round);
 
 	(tx || library.db).rounds
```
nazarhussain: Th eline below states:

```
.summedRound(scope.round, constants.activeDelegates)
```
To avoid any mistake we should always calculate `round` from the `height` of the block given in scope. Because its fully dependent value. 

With this approach there can be mistake in future to sum a `round` with wrong block height.

test/unit/modules/rounds.js
```diff
@@ -369,29 +369,22 @@ describe('rounds', () => {
...
-				stub = sinon.stub(db.rounds, 'summedRound').resolves(rows);
-				done();
-			});
-
-			after(() => {
-				return stub.restore();
-			});
+		describe('when last block is genesis block', () => {
+			const scope = {
+				round: 1,
```
nazarhussain: As stated above, `round` should not be passed as scope to sumRound, it should always calculate `round` from the `height` of the block.

test/unit/modules/rounds.js
```diff
@@ -402,37 +395,81 @@ describe('rounds', () => {
 			});
 
 			it('should set scope.roundFees correctly', () => {
```
nazarhussain: Better to set the test case explicitly to 

> should set scope.roundFees correctly to zero

test/unit/modules/rounds.js
```diff
@@ -402,37 +395,81 @@ describe('rounds', () => {
...
-				return expect(scope.roundFees).to.equal(100);
+				return expect(scope.roundFees).to.equal(0);
 			});
 
 			it('should set scope.roundRewards correctly', () => {
```
nazarhussain: Same as above

test/unit/modules/rounds.js
```diff
@@ -402,37 +395,81 @@ describe('rounds', () => {
...
+							delegates: ['delegate1', 'delegate2', 'delegate3'],
+						},
+					];
+					stub = sinon.stub(db.rounds, 'summedRound').resolves(rows);
+					done();
+				});
+
+				after(() => {
+					return stub.restore();
+				});
```
nazarhussain: Would be nice if convert these to `beforeEach` and `afterEach`, I notice difficulties with `before` and `after` maintaining tests in long run.

nazarhussain:
@4miners The logic in `slots` 

```
calcRound(height) {
		return Math.ceil(height / this.delegates);
	},
```

still don't reflect our logic that `genesisBlock` will be only block in round 1.

modules/rounds.js
4miners: Out of scope of this PR, I didn't touch that part.

test/functional/system/snapshotting.js
4miners: Yeah, you're and @MaciejBaj have right, I will update the comment.

4miners:
@nazarhussain No, in round 1 there will be still 101 blocks.

modules/rounds.js
MaciejBaj: @nazarhussain good refactoring suggestion. I would appreciate if you could open it as a separate issue.

test/unit/modules/rounds.js
MaciejBaj: As above, can be done as a separate issue.

test/unit/modules/rounds.js
MaciejBaj: Please do @4miners, the behaviour shouldn't change.

modules/rounds.js
nazarhussain: But you touch the logic of `genesisBlock` to be round 1. And I believe this code relates to the logic.

db/sql/migrations/updates/20180205000003_create_rounds_rewards_table.sql
4miners: Table structure is defined just 1 line above, but ok - I'll change that.

nazarhussain:
@4miners Look at this code logic

```
// Establish if finishing round or not
	scope.finishRound =
		round !== nextRound || (block.height === 1 || block.height === 101);
```

It clearly states that **round will be finished** on block height 1.

modules/rounds.js
4miners: No, I didn't changed the logic - round 1 was allways summed twice:
- just after genesis block (to elect delegates, update their vote weight, etc.)
- after 101 blocks (actual round 1 ended)
To be honest that condition probably do nothing, but I don't really want to touch that part there, as is out of scope of the issue.

4miners:
@nazarhussain Yes, round 1 is finished twice, it always was.

The difference was:
- during sync/forging round 1 summed just after genesis block contains one block
- during snapshotting round 1 summed just after genesis block contains 101 blocks (because we have them in database).

What I changed here is round 1 summed just after genesis block will be always forced to contain 1 block (with 0 fees, 0 rewards) regardless if we sync or snapshotting.

nazarhussain:
I don't like this behavior: 

> round 1 is finished twice and rest of the round once

Not now, but in future we should move genesis block to round 0, that can be called **genesis round** and only contains the genesis block. So rest of the logic stay aligned.

4miners:
@nazarhussain I agree, I also don't like that behavior.

4miners:
I resolved comments, @MaciejBaj @nazarhussain please check again.

Issue #1774, opened 2018-03-22
Refactor lisk-core constants to use const, upper case with underscores


ManuGowda:
### Expected behavior
- The constants should be defined using `const` 
- The constants names should have an upper case with underscores for separators.
### Actual behavior
- Lisk-core constants are using `vars`, this would lead to the mutability of the constants and leading to change in behaviour in the application. 
- The constants are defined using lowercase with camelCase for separators.

### Steps to reproduce
```
var constants = require('../helpers/constants.js');
constants.activeDelegates = 300;
Now the activeDelegates count will be changed to 300 from 101.
```
### Which version(s) does this affect? (Environment, OS, etc...)
1.0.0

4miners:
Reopen, because it was closed automatically by mistake.

MaciejBaj:
Consecutively in the whole application change var to const when possible, and var to let when it gets mutated.

PR #2149, opened 2018-06-21
Support multiple networks with one installation - Closes #398, #1774

### LINKED ISSUE #398: Support multiple networks with one installation

nazarhussain:
### What was the problem?
Application was tightly integrated with particular network. To change the network, we had to switch to relevant branch. That makes it difficult to manage multiple networks. 

### How did I fix it?
Created network specific configuration folder and a startup option to choose the network

### How to test it?
Run core and integration tests. 

### Review checklist

* The PR solves #398 , #1774
* Linting passes
* Tests pass


api/controllers/delegates.js
```diff
@@ -33,6 +33,7 @@ var modules;
  */
 function DelegatesController(scope) {
 	modules = scope.modules;
+	constants = scope.config.constants;
```
4miners: I don't really like that, can we be consistent? Why not use `global.constants` here?

config/betanet/constants.js
```diff
@@ -0,0 +1,118 @@
...
+		lifetime: {
+			minimum: 1,
+			maximum: 72,
+		},
+		keysgroup: {
+			minItems: 1,
+			maxItems: 15,
+		},
+	},
+	nethashes: [
```
4miners: Do we need that here? From logical point of view there should be hesh of `betanet` only. Same for other networks.

config/betanet/exceptions.js
```diff
@@ -0,0 +1,57 @@
...
+ * @property {string[]} senderPublicKey
+ * @property {string[]} signatures
+ * @property {string[]} multisignatures
+ * @property {string[]} votes
+ * @todo Add description for the namespace and the properties
+ */
+module.exports = {
+	blockRewards: [],
+	delegates: [],
+	genesisPublicKey: {
```
4miners: Same here, why we need all?

nazarhussain:
@4miners The PR is in-progress, don't review it right now. Once its finalized, I will definitely request you to do the first review.

4miners:
@nazarhussain Thanks, but what is the purpose of opening PR that is not ready? 😏

Slamper:
@nazarhussain now that you are already refactoring the constants you could already take care of #1774

nazarhussain:
@4miners To keep track of changes and because we have `in-progress` label for PRs. 🤓

nazarhussain:
The commit https://github.com/LiskHQ/lisk/pull/2149/commits/8c97b976532a8b21e4bab4ed48b45f2e61315301 will close #1774

nazarhussain:
Important changes to review: 

https://github.com/LiskHQ/lisk/pull/2149/files#diff-0364f57fbff2fabbe941ed20c328ef1a
https://github.com/LiskHQ/lisk/pull/2149/files#diff-10596d7c5748592d2e76555a83d88a1a
https://github.com/LiskHQ/lisk/pull/2149/files#diff-9f30204b3444de4b458d4395cfe3bdf7L495

nazarhussain:
These two issues should be done right after this PR: 

https://github.com/LiskHQ/lisk/issues/2155
https://github.com/LiskHQ/lisk/issues/2156

nazarhussain:
Created issues for relevant changes in other repos; 

https://github.com/LiskHQ/lisk-build/issues/292
https://github.com/LiskHQ/lisk-scripts/issues/102

app.js
```diff
@@ -111,6 +109,11 @@ var appConfig = AppConfig(require('./package.json'));
 
 // Define availability of top accounts endpoint
 process.env.TOP = appConfig.topAccounts;
+process.env.NETWORK = appConfig.network;
+
+// Global objects to be utilized under modules/helpers where scope is not accessible
+global.constants = appConfig.constants;
```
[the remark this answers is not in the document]
nazarhussain: Yes without global its preferred. But that requires refactoring complete `helpers` namespace. As any of helper is not designed to accept `scope`. So for now we should go with `global` and when we will modularize code this change can be handled there.

app.js
```diff
@@ -111,6 +109,11 @@ var appConfig = AppConfig(require('./package.json'));
 
 // Define availability of top accounts endpoint
 process.env.TOP = appConfig.topAccounts;
+process.env.NETWORK = appConfig.network;
```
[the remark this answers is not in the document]
nazarhussain: The convention is already used for env variables without `LISK_` prefix. So followed it.

config/betanet/config.json
```diff
@@ -0,0 +1,117 @@
+{
```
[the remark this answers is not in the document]
nazarhussain: As per discussion all networks should exists, if required we can delete those files. @MaciejBaj ?

helpers/config.js
```diff
@@ -38,6 +37,11 @@ function Config(packageJson) {
 	program
 		.version(packageJson.version)
 		.option('-c, --config <path>', 'config file path')
+		.option(
+			'-n, --network [network]',
+			'lisk network [dev|beta|main|test]. Defaults to "dev"',
```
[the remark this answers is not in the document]
nazarhussain: Yes I prefer to use conventions. Not required to match exact names with directory names. So convention I opted here is to have `<network_name>net` directories.

app.js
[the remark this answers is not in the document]
shuse2: In helper, for example peer_manager, there is a singleton.
I think if you follow that it would be pretty easy to create the config manager without so much change.

app.js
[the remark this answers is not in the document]
shuse2: yea i agree putting just `LISK_` would be inconsistent here.
but what is the reason of using environment variable here?
environment variable should be used as an input to the application

config/betanet/constants.js
4miners: Bump

helpers/config.js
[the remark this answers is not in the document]
4miners: Why not `devnet|betanet|mainnet|testnet`? We always used those names for our networks.

test/functional/http/get/node/node.js
```diff
@@ -57,7 +57,7 @@ describe('GET /node', () => {
 			return expect(constantsResponse)
 				.to.have.property('version')
 				.to.match(
-					/^([0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3})(-(alpha|beta|rc)\.[0-9]{1,3})?$/
+					/^([0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3})(-(alpha|beta|rc)\.[0-9]{1,3}(\.[0-9]{1,3})?)?$/
```
4miners: What is that change for?

test/integration/scenarios/stress/2.register_delegate.js
```diff
@@ -17,7 +17,7 @@
 var Promise = require('bluebird');
 var lisk = require('lisk-elements').default;
 var accountFixtures = require('../../../fixtures/accounts');
-var constants = require('../../../../helpers/constants');
+const constants = require('../../../../config/mainnet/constants');
```
4miners: Using `mainnet` constants here (and in many other places) for tests is not very consistent.

test/functional/http/get/node/node.js
nazarhussain: The test was failing due as we were using wrong regular expression. Was not identified before as in test config there was `0.0.0` version always.

test/integration/scenarios/stress/2.register_delegate.js
nazarhussain: It will be fixed by #2156

app.js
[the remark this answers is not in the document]
nazarhussain: @shuse2 I have to create this env variable to be used by child process from socket cluster.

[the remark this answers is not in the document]
shuse2: Wouldn't it be better if we actually read from env?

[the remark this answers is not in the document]
nazarhussain: Yes that will be good, but for that our configuration manager should support env variable support. If needed we can convert `commander` to have environment variable support for each option. So user can either pass variable or command option. But its out of scope for now.

logic/transaction.js
```diff
@@ -36,21 +36,19 @@ const __private = {};
...
  * @requires lodash
  * @requires helpers/bignum
- * @requires helpers/constants
- * @requires helpers/exceptions
  * @requires helpers/slots
  * @param {Database} db
  * @param {Object} ed
  * @param {ZSchema} schema
- * @param {Object} genesisblock
+ * @param {Object} genesisBlock
```
diego-G: If we do this change, we should do it across the whole code. There are a few places in the test suite where we still call it `genesisblock`.

modules/blocks/chain.js
```diff
@@ -251,7 +251,7 @@ Chain.prototype.applyGenesisBlock = function(block, cb) {
 		err => {
 			if (err) {
 				// If genesis block is invalid, kill the node...
-				process.exit(0);
+				process.emit('cleanup', err.message);
```
diego-G: Why is this change here? How is it related to the PR?

modules/blocks/verify.js
```diff
@@ -38,8 +38,6 @@ __private.lastNBlockIds = [];
  * @requires async
  * @requires crypto
  * @requires lodash
- * @requires helpers/constants
```
diego-G: This is not removed uniformly across the code. Please, scan the code and remove every result.

config/mainnet/exceptions.js
```diff
@@ -39,10 +39,6 @@
 module.exports = {
 	blockRewards: [],
 	delegates: [],
-	genesisPublicKey: {
```
diego-G: Why is this been removed? If it is correct, we should remove the description headers in every new exception file like https://github.com/LiskHQ/lisk/blob/dce611af6a5068b8443d6ec84021345baa0259f9/config/devnet/exceptions.js#L25

config/betanet/constants.js
diego-G: @nazarhussain could you answer to this comment.

logic/transaction.js
nazarhussain: I tried to fix it across the code base. Will see where it is missed.

modules/blocks/chain.js
nazarhussain: During some testing, identified this use case where lisk exit without showing error. So fix it right there instead of creating a 1 line PR.

config/mainnet/exceptions.js
nazarhussain: Yes this is removed because of this change https://github.com/LiskHQ/lisk/pull/2149/files#diff-9f30204b3444de4b458d4395cfe3bdf7R493

nazarhussain:
@4miners @shuse2 @diego-G You guys may have missed a change. I removed `nethash` attribute from every config file. As its the same as `genesisBlock.payloadHash`. Assigned that value runtime to `config.nethash`. 

Do you guys think this is good to go?

shuse2:
I think it's more accurate and good except the fact that we probably shouldn't injecting to the file input, but we can fix that later.
Also, there was an issue where nethash in the config and payloadhash doesn't match from genesis block. is it checked?

fchavant:
@diego-G if the configration file file now lives in `config/<network>/config.json` then at the very least https://github.com/LiskHQ/lisk/blob/dce611af6a5068b8443d6ec84021345baa0259f9/docker_files/etc/confd/conf.d/lisk.toml#L3 needs to be adapted.

nazarhussain:
@shuse2 Yes its checked and fixed.

README.md
```diff
@@ -215,9 +215,11 @@ pm2 stop lisk
 **NOTE:** The **port**, **address** and **config-path** can be overridden by providing the relevant command switch:
 
 ```
-pm2 start --name lisk app.js -- -p [port] -a [address] -c [config-path]
+pm2 start --name lisk app.js -- -p [port] -a [address] -c [config-path] -n [network]
 ```
 
+You can pass any of `devnet`, `testnet`, `betanet` or `mainnet` for the network option.
```
MaciejBaj: `devnet` is used by default.

config/betanet/constants.js
MaciejBaj: Also, nethashes are defined in exceptions files. It should be sorted out. It's not a scope of this task though.

config/testnet/constants.js
```diff
@@ -0,0 +1,118 @@
...
+			minimum: 1,
+			maximum: 72,
+		},
+		keysgroup: {
+			minItems: 1,
+			maxItems: 15,
+		},
+	},
+	nethashes: [
+		// Mainnet
```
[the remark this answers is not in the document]
4miners: We don't really need those nethashes at all, because we can get nethash from genesis block.

nazarhussain:
@MaciejBaj Had created issue to take care of your major feedback. https://github.com/LiskHQ/lisk/issues/2184

nazarhussain:
@MaciejBaj Related to db error the logic to stop the application is still there and working. 

```
6:32:55 error: database "lisk_dev" does not exist
         connection: {
.....
.....
.....
[inf] 2018-06-29 14:32:55 | Cleaning up...
```

During cleanup some error appears on line 
```
				scope.socketCluster.removeAllListeners('fail');
				scope.socketCluster.destroy();
```

```
Unhandled rejection TypeError: Cannot read property 'removeAllListeners' of undefined
    at process.once.error (/Users/nazar/Hub/LightCurve/Lisk/dev/lisk/app.js:771:24)
```

This error breaks the cleaning process and avoid shutting down the application.

nazarhussain:
@MaciejBaj Created this issue for the cleanup error https://github.com/LiskHQ/lisk/issues/2188

PR #2165, opened 2018-06-25
Send back better error when disconnecting because of invalid peer headers - Closes #2068

### LINKED ISSUE #2068: Send back better error when disconnecting because of invalid peer headers
### Expected behavior

If a peer tries to connect with missing required header fields, a node should send back a descriptive error message.

### Actual behavior

If a peer tries to connect to us with missing headers, we disconnect them with a generic 'BadConnectionError'.
This generic error is useful for rejecting malicious peers (as not to give away too much info) but it's not ideal in this case.

See https://github.com/LiskHQ/lisk/pull/2064#discussion-diff-191176538R132

### Steps to reproduce

Try to connect to a node with missing headers as part of the WebSocket connection query string.

1.0.0

jondubois:
### What was the problem?

When a peer tried to connect with invalid headers, the error messages were not ideal.
Note that a descriptive error message was sent back to the peer's close event handler before (and it was logged correctly by the peer) but it wasn't always formatted correctly.

### How did I fix it?

- Fixed the formatting related to failed handshake errors.
- Added test cases which shows how to handle the connection error.

### How to test it?

`npm run mocha test/functional/ws/transport/client.js`

### Review checklist

* The PR solves #2068
* Linting passes
* Tests pass

### DIFF
--- a/api/ws/workers/middlewares/handshake.js
+++ b/api/ws/workers/middlewares/handshake.js
@@ -72,11 +72,17 @@ var middleware = {
 		return function(headers, cb) {
 			z_schema.validate(headers, definitions.WSPeerHeaders, error => {
 				if (error) {
+					var errorDescription;
+					if (error[0].path && error[0].path.length) {
+						errorDescription = `${error[0].path}: ${error[0].message}`;
+					} else {
+						errorDescription = `${error[0].message}`;
+					}
 					return setImmediate(
 						cb,
 						{
 							code: failureCodes.INVALID_HEADERS,
-							description: `${error[0].path}: ${error[0].message}`,
+							description: errorDescription,
 						},
 						null
 					);
@@ -125,11 +131,11 @@ var middleware = {
 				// TODO : double check socket IP
 				if (config.blackListedPeers.includes(headers.ip)) {
 					return setImmediate(
-						cb, {
+						cb,
+						{
 							code: failureCodes.BLACKLISTED_PEER,
-							description: failureCodes.errorMessages[
-								failureCodes.BLACKLISTED_PEER
-							],
+							description:
+								failureCodes.errorMessages[failureCodes.BLACKLISTED_PEER],
 						},
 						peer
 					);
--- a/api/ws/workers/middlewares/handshake.js
+++ b/api/ws/workers/middlewares/handshake.js
@@ -72,11 +72,9 @@ var middleware = {
 		return function(headers, cb) {
 			z_schema.validate(headers, definitions.WSPeerHeaders, error => {
 				if (error) {
-					var errorDescription;
+					var errorDescription = error[0].message;
 					if (error[0].path && error[0].path.length) {
-						errorDescription = `${error[0].path}: ${error[0].message}`;
-					} else {
-						errorDescription = `${error[0].message}`;
+						errorDescription = `${error[0].path}: ${errorDescription}`;
 					}
 					return setImmediate(
 						cb,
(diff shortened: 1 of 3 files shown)


api/ws/workers/middlewares/handshake.js
```diff
@@ -72,11 +72,17 @@ var middleware = {
 		return function(headers, cb) {
 			z_schema.validate(headers, definitions.WSPeerHeaders, error => {
 				if (error) {
+					var errorDescription;
```
4miners: Why not `let`?

api/ws/workers/middlewares/handshake.js
```diff
@@ -72,11 +72,17 @@ var middleware = {
...
+					if (error[0].path && error[0].path.length) {
```
4miners: Maybe
```
errorDescription = `${error[0].message}`;
if (error[0].path && error[0].path.length) {
    errorDescription = `${error[0].path}: ${errorDescription}`;
}
```

test/functional/ws/transport/client.js
```diff
@@ -36,7 +39,27 @@ describe('RPC Client', () => {
...
+		validPeerStub.socket.removeAllListeners('close');
+		validPeerStub.socket.removeAllListeners('connect');
+	}
+
+	function captureConnectionResult() {
+		closeErrorCode = null;
+		closeErrorReason = null;
+		unbindConnectionHandlers();
+		validPeerStub.socket.on('close', (code, reason) => {
+			unbindConnectionHandlers();
```
4miners: Why we need that here if we're already calling it 2 lines above?

test/functional/ws/transport/client.js
```diff
@@ -36,7 +39,27 @@ describe('RPC Client', () => {
...
+			closeErrorCode = code;
+			closeErrorReason = reason;
+		});
+		validPeerStub.socket.on('connect', () => {
+			unbindConnectionHandlers();
```
4miners: same here

api/ws/workers/middlewares/handshake.js
ManuGowda: Can be rewritten to 
```
const { path, message } = error[0];
const errorDescription = path ? `${path}: ${message}` : `${message}`;
```

api/ws/workers/middlewares/handshake.js
jondubois: To keep consistency within the file module. This file uses `var` everywhere.

test/functional/ws/transport/client.js
jondubois: The second one happens asynchronously when the socket disconnects. If the user calls the function twice before the socket closes, then it will remove the previous event listeners.

api/ws/workers/middlewares/handshake.js
jondubois: That file doesn't use es6 yet, I thought that the priority is to keep consistency at the file level.

jondubois: I like the `var somevar;` declaration convention; we use it in other files as well but I can change it if you don't like it.

test/functional/ws/transport/client.js
jondubois: I thought about it a bit more. The second unbindConnectionHandlers() is not really needed since the socket itself will be discarded when the test suite finishes so I simplified that logic a bit.

jondubois:
@4miners @MaciejBaj Please review. @MaciejBaj I didn't cover all the possible failure codes because it would take me too long to do this (it's hard to reproduce some of them in the functional test).

MaciejBaj:
@jondubois please open the following issue - Revisit P2P error codes and tidy it's tidy failure codes tests.

PR #2166, opened 2018-06-25
Update logging to enable JSON output - Closes #2148

### LINKED ISSUE #2148: Improve logging system
### Description
Currently, logging does not have standard and outputted in text format.
It is hard to use for analysis of KPI.

### Expected
- console log in the text format for readability
- file log in the JSON format for extensible usage
- Standardised format and keys for better searchability 

1.0

shuse2:
### What was the problem?
There was no way to output JSON format, which was critical to analyze the log in QA.

```javascript
library.logger = new Logger({xxx}).child('loader');
library.logger.debug(
	['Broadhash consensus:', self.getLastConsensus(), '%'].join(' ')
);
```
Will produce

JSON
```javascript
{  
   "name":"logs/lisk.log",
   "hostname":"my.local",
   "pid":48673,
   "module":"peers",
   "level":20,
   "msg":"Broadhash consensus: 0 %",
   "time":"2018-06-29T09:48:13.966Z",
   "src":{  
      "file":"/lisk/modules/peers.js",
      "line":879,
      "func":"calculateConsensus"
   },
   "v":0
}
```

`src` will only shown when level is `debug` or `trace`

will produce key `durationMs`.
### How did I fix it?
Rewrite logging module using `bunyan`.

### How to test it?
start application with logging enabled

### Review checklist

* The PR solves #2148 
* Linting passes
* Tests pass


package.json
```diff
@@ -82,6 +82,7 @@
 		"sway": "=1.0.0",
 		"valid-url": "=1.0.9",
 		"wamp-socket-cluster": "=2.0.0-beta.4",
+		"winston": "=3.0.0",
```
nazarhussain: @shuse2 Based on discussion in issue, any specific reason you choose `winston`. Its the slowest logger in all benchmarks. 

I still believe `bunyan` is better option for us, as it give good set of features for CLI usage that is helpful for us in our project scope. e.g. filtering logs based on some conditions. See https://github.com/trentm/node-bunyan#cli-usage

If you choose `winston` for profiling, then I guess its not the right choice. Logging and profiling are both totally different things. We should consider these different. For logging our choice should be based on logging features. When we need profiler, we will search some tool/lib specialized in profiling.

logger.js
```diff
@@ -10,135 +10,142 @@
...
+	new winston.transports.File({
+		level,
+		filename,
+		format: format.combine(
+			format.timestamp(),
+			injectKey('module', module)(),
+			injectKey('hostname', os.hostname())(),
+			formatInfo(),
+			format.json()
+		),
```
nazarhussain: I am not sure about this structure, does that means every log entry will make 6 additional function calls? Or minimum 3 additional calls to `injectKeys` and `formatInfo`. I feel that's a performance degrade for the each log entry.

logger.js
```diff
@@ -10,135 +10,142 @@
...
+		level,
+		levels: customLevels.levels,
+		format: format.combine(
+			format.colorize(),
+			injectKey('module', module)(),
+			format.timestamp(),
+			formatInfo(),
+			format.printf(consoleFormat)
+		),
+	});
```
nazarhussain: I really like bunyan CLI approach here. It just take JSON object from `STDOUT` and format it outside the main process. This avoid performance overhead for just formatting logs to show in console.

logger.js
```diff
@@ -10,135 +10,142 @@
...
-	}
-
-	Object.keys(config.levels).forEach(name => {
-		function log(message, data) {
-			var log = {
-				level: name,
-				timestamp: strftime('%F %T', new Date()),
-			};
+		return accumulated;
+	}, {}));
```
nazarhussain: We have to revisit the whole logic here. In one single operations we may be logging 20-30 log entries. For each log entry:

- This function is looping over all keys in json object.
- Then for each key its looping over another array `logRootKeys` to validate `includes`.
- So if a log object contains 8 keys, then logging that object may need 56 loop iterations and 8 function calls. 
- Also initializing default value `.meta = {}` does not make sense here, instead should be done on log object initialization. 

Overall I believe this logging approach will create a lot of unwanted overhead to system performance.

logger.js
shuse2: It will go through all of the functions like middleware, but I think it will happen in any logger where we want to format the log before outputting it

logger.js
shuse2: >  This avoid performance overhead for just formatting logs to show in console.
console log shouldnt be turned on the server or at least with less level.

package.json
shuse2: Since console output is for mainly development purpose, I think it's more simpler and useful if it's just console log when it's turned on without external tool.

If we only consider performance, probably `pino`, but if we consider feature it would be `winston`.
I chose `winston` based on the fact that
- feature rich (formatting, profiling, container)
- 3.0 just released, so active development
- console logging by default

logger.js
shuse2: Yes main reason of this is more like a util function to format the meta data, so we can use it like
`logger.info('some message', { thiswill: 'be in meta'});`

if we want to write it like
`logger.info('some message', { meta: { thiswill: 'be in meta'}});` we don't need this process

logger.js
nazarhussain: Still its not the right way to do it. You can use `tail -f lisk.log | bunyan` and it will show you logs on screen in the way you normally see in console. So it does not matter if you enabled console or not, formatting logs for console outside main process is helpful.

logger.js
nazarhussain: If that's happening in all logging tools then there is a right reason to create our own logger. This is too much of the computation power to be used unnecessary for logging.

package.json
shuse2: Also, yes winston is slower than bunyan, but
```
benchBunyan*10000: 737.315ms
benchWinston*10000: 766.665ms
benchPino*10000: 306.935ms
```
i dont think it's huge difference to be considered

logger.js
shuse2: Even if we do, we will need similar operation, unless we have better way to inject globally into the logger.

shuse2:
This PR should moved to `2.0` since it's breaking change ='(

MaciejBaj:
@shuse2 also a lot of conflicts now.

app.js
```diff
@@ -144,12 +144,13 @@ var config = {
  *
  * @memberof! app
  */
-var logger = new Logger({
-	echo: process.env.LOG_LEVEL || appConfig.consoleLogLevel,
-	errorLevel: process.env.FILE_LOG_LEVEL || appConfig.fileLogLevel,
+var logger = createLogger({
+	level: process.env.FILE_LOG_LEVEL || appConfig.fileLogLevel,
 	filename: appConfig.logFileName,
 });
```
ManuGowda: can we add `createLogger` specific name parameter here like `name: core` `name: p2p` so on and so forth, and in the `bunyan.createLogger` functions instead of using filename for the name parameter u can use the custom name `name: core` `name: p2p` we passed while creating a logger, the log would look much clear and better
From:
```
{"name":"logs/lisk.log","hostname":"MacBook-Pro-6.local","pid":48427,"module":"delegates"}
```
To:
```
{"name":"core","hostname":"MacBook-Pro-6.local","pid":48427,"module":"delegates"}
```
```
{"name":"database","hostname":"MacBook-Pro-6.local","pid":48427,"module":"delegates"}
```

app.js
```diff
@@ -175,12 +175,12 @@ if (
 try {
 	lastCommit = git.getLastCommit();
 } catch (err) {
-	logger.debug('Cannot get last git commit', err.message);
+	appLogger.debug('Cannot get last git commit', err.message);
```
ManuGowda: this isn't convincing, we are catching an error and logging it as `debug`. I agree it was like that before, but good to change it now.

app.js
```diff
@@ -752,9 +762,9 @@ d.run(() => {
 			// Receives a 'cleanup' signal and cleans all modules
 			process.once('cleanup', error => {
 				if (error) {
-					scope.logger.fatal(error.toString());
+					appLogger.fatal(error.toString());
```
ManuGowda: I am not sure why we have specifics here, everywhere are logging `error` and only here I see converting to a string, not a huge concern, just matter of consistency.

logger.js
```diff
@@ -10,135 +10,45 @@
...
 
-	function snipFragileData(data) {
-		for (var key in data) {
-			if (key.search(/passphrase|password/i) > -1) {
-				data[key] = 'XXXXXXXXXX';
-			}
-		}
-		return data;
+const createDirIfNotExists = filename => {
+	if (!filename || filename === '') {
```
ManuGowda: if (!filename) handles all check `null, undefined and empty string` so we can remove `|| filename === ''`

test/integration/configs/config.node-4.json
```diff
@@ -0,0 +1,203 @@
+{
```
ManuGowda: these config files should not be tracked, these are generated automatically during the integration test run, please remove all the configs `config.node-0.json` to `config.node-9.json`

test/setup.js
```diff
@@ -56,8 +56,7 @@ if (process.env.LOG_DB_EVENTS === 'true') {
 	testContext.config.db.logEvents = ['error'];
 }
 
-testContext.consoleLogLevel =
-	process.env.LOG_LEVEL || testContext.consoleLogLevel;
+testContext.fileLogLevel = process.env.LOG_LEVEL || testContext.fileLogLevel;
```
ManuGowda: this should be `FILE_LOG_LEVEL`

test/unit/modules/system.js
```diff
@@ -26,12 +26,23 @@ describe('system', () => {
...
 	let self;
 	let loggerStub;
+	let childLoggerStub;
 	let dbStub;
 	let dummyConfig;
 
 	beforeEach(done => {
 		// Library
+		childLoggerStub = {
+			done: sinonSandbox.spy(),
```
ManuGowda: what is `done` log level? I saw this at 3 different places above as well.

shuse2: I guess forgotten from the last library

ManuGowda:
@shuse2 also can you please add ignore for this log file `test/integration/utils/integrationTestsLogger.logs`? after running integration test this pops up

app.js
shuse2: I think this `debug` level is intention. Especially for built package, i dont think there will be a git included.

logger.js
```diff
@@ -10,135 +10,45 @@
...
-						log.timestamp.grey,
-						'|',
-						log.message
-					);
-				}
-			}
-		}
-
-		exports[name] = log;
+const shouldShowSrc = level => level === 'debug' || level === 'trace';
```
nazarhussain: Not sure about `error` level but we should also show src on `fatal` level.

logger.js
```diff
@@ -10,135 +10,45 @@
...
-	child_process.execSync(`mkdir -p ${path.dirname(config.filename)}`);
-	var log_file = fs.createWriteStream(config.filename, { flags: 'a' });
-
-	exports.setLevel = function(errorLevel) {
-		config.errorLevel = errorLevel;
-	};
+const defaultLoggerConfig = {
+	filename: 'logs/lisk.log',
+	level: 'none',
+};
```
nazarhussain: As said earlier, there should be `fileLogLevel` and `consoleLogLevel` and handle it by multiple streams in bunyan.

modules/accounts.js
```diff
@@ -62,7 +62,7 @@ class Accounts {
 			transactionTypes.VOTE
 		] = library.logic.transaction.attachAssetType(
 			transactionTypes.VOTE,
-			new Vote(scope.logger, scope.schema)
+			new Vote(scope.logger.child({ module: 'accounts/vote' }), scope.schema)
```
nazarhussain: This should be `logic/vote`

modules/blocks.js
```diff
@@ -54,25 +54,25 @@ __private.isActive = false;
 class Blocks {
 	constructor(cb, scope) {
 		library = {
-			logger: scope.logger,
+			logger: scope.logger.child({ module: 'blocks' }),
```
nazarhussain: More nicer would be to have:

```
modules/blocks
modules/blocks/api
modules/blocks/verify 
```
And so one.

modules/blocks/process.js
```diff
@@ -487,10 +487,10 @@ Process.prototype.loadBlocksFromPeer = function(peer, cb) {
...
 				// Update last valid block
 				lastValidBlock = block;
-				library.logger.info(
-					['Block', block.id, 'loaded from:', peer.string].join(' '),
-					`height: ${block.height}`
-				);
+				library.logger.info('Block loaded', {
+					blockId: block.id,
+					peer: peer.string,
+				});
```
nazarhussain: Bunyan consider that first param must be `object` and additional message should be passed as `second argument`. How you override this behavior? 

https://github.com/trentm/node-bunyan#log-method-api

helpers/config.js
```diff
@@ -91,7 +91,7 @@ function Config(packageJson, parseCommandLineOptions = true) {
 		wsPort: +program.port || process.env.LISK_WS_PORT || null,
 		httpPort: +program.httpPort || process.env.LISK_HTTP_PORT || null,
 		address: program.address,
-		consoleLogLevel: program.log || process.env.LISK_CONSOLE_LOG_LEVEL,
+		fileLogLevel: program.log || process.env.FILE_LOG_LEVEL,
```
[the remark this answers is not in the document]
shuse2: I think this is the same point as @ManuGowda is making.
to make it readable, we still need to pipe with bunyan, i thought it wouldnt add much value
I have no problem to put it back

modules/blocks.js
shuse2: since we have `module` key in the log, wouldn't it be redundant?

logger.js
shuse2: i was thinking of cutting off by the log level. (debug or lower)
info or above is more for non-developer, so fatal should be already easily identifiable without src

MaciejBaj:
Changes are well-written and valid. However, the overall approach of logging system across different part of applications and use cases needs to be rethought. The PR and issue #2148 are unassigned from project 1.1.0 and will be included in the upcoming releases.

diego-G:
PR closed till final agreement.

Issue #1783, opened 2018-03-26
Incorrect missedBlocks update after delete last block of round


4miners:
### Scenario
Round rollback when forger of last block of round is replaced in last block of round.
- unvote deletage A, vote delegate B into 101 at last block of round
- delete last block of round

### Expected behavior
- delegate B should have `missedBlocks` equal to `0`
### Actual behavior
- delegate B have `missedBlocks` equal to `-1`
### Steps to reproduce
Run rounds functional system tests.
### Which version(s) does this affect? (Environment, OS, etc...)
`1.0.0`, `0.9.12` not tested

PR #2169, opened 2018-06-26
Round fixes - Closes #216, #1783

### LINKED ISSUE #216: Why is mem_round flushed twice while finishing a round
https://github.com/LiskHQ/lisk/blob/development/modules/round.js#L136
https://github.com/LiskHQ/lisk/blob/development/modules/round.js#L137

```
return this.updateVotes()
            .then(this.updateMissedBlocks)
            .then(this.flushRound)
            .then(this.applyRound)
            .then(this.updateVotes)
            .then(this.flushRound)
            .then(function () {
                private.ticking = false;
                return t;
            });
```

I think the later updateVotes and flushRound can be removed.

4miners:
### What was the problem?
Inconsistencies in round module:
- `missedBlocks` and `producedBlocks` skipped during round snapshot creation
- redundant calls during `land` and `backwardLand` in round logic
- `missedBlocks` and `producedBlocks` casted to big integer (database type is `INT`)
### How did I fix it?
- snapshot `missedBlocks` and `producedBlocks` during round snapshot and then restoring them when we restore the snapshot
- simplify logic in `land` (removed `flushRound` call)
- simplify logic in `backwardLand` (removed `updateMissedBlocks`, `flushRound`, `updateVotes` calls - because they will be overwritten anyway from snapshot)
- remove `BIGINT` cast on `missedBlocks` and `producedBlocks`
- adjust expectations in various tests
- unskip scenario in functional system rounds test (previously failing)
### How to test it?
Run test suite.
### Review checklist

* The PR solves #216, #1783
* Linting passes
* Tests pass

### DIFF
--- a/api/controllers/accounts.js
+++ b/api/controllers/accounts.js
@@ -71,8 +71,6 @@ AccountsController.getAccounts = function(context, next) {
 				delete account.delegate;
 			} else {
 				account.delegate.rank = parseInt(account.delegate.rank);
-				account.delegate.missedBlocks = account.delegate.missedBlocks;
-				account.delegate.producedBlocks = account.delegate.producedBlocks;
 			}
 			if (_.isNull(account.publicKey)) {
 				account.publicKey = '';
--- a/api/controllers/accounts.js
+++ b/api/controllers/accounts.js
@@ -71,10 +71,8 @@ AccountsController.getAccounts = function(context, next) {
 				delete account.delegate;
 			} else {
 				account.delegate.rank = parseInt(account.delegate.rank);
-				account.delegate.missedBlocks = parseInt(account.delegate.missedBlocks);
-				account.delegate.producedBlocks = parseInt(
-					account.delegate.producedBlocks
-				);
+				account.delegate.missedBlocks = account.delegate.missedBlocks;
+				account.delegate.producedBlocks = account.delegate.producedBlocks;
 			}
 			if (_.isNull(account.publicKey)) {
 				account.publicKey = '';
--- a/api/controllers/delegates.js
+++ b/api/controllers/delegates.js
@@ -77,8 +77,8 @@ DelegatesController.getDelegates = function(context, next) {
 			delete delegate.publicKey;
 			delete delegate.address;
 
-			delegate.missedBlocks = parseInt(delegate.missedBlocks);
-			delegate.producedBlocks = parseInt(delegate.producedBlocks);
+			delegate.missedBlocks = delegate.missedBlocks;
+			delegate.producedBlocks = delegate.producedBlocks;
 			delegate.rank = parseInt(delegate.rank);
 
 			return delegate;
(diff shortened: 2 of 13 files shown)


db/repos/accounts.js
```diff
@@ -40,8 +40,8 @@ const normalFields = [
 	{ name: 'fees', cast: 'bigint', def: '0', skip: ifNotExists },
 	{ name: 'rewards', cast: 'bigint', def: '0', skip: ifNotExists },
 	{ name: 'vote', cast: 'bigint', def: '0', skip: ifNotExists },
-	{ name: 'producedBlocks', cast: 'bigint', def: '0', skip: ifNotExists },
-	{ name: 'missedBlocks', cast: 'bigint', def: '0', skip: ifNotExists },
+	{ name: 'producedBlocks', def: 0, skip: ifNotExists },
```
SargeKhan: (Not related to PR but) aren't `def: 0` and `skip: ifNotExists` conflicting? Meaning if default is 0 then it will always exist and never be skipped?

nazarhussain: @SargeKhan `skip` is used by `update` and `sets` and ignored by `inserts` and `values`. So the `def` covers the use case of inserting records.

api/controllers/accounts.js
```diff
@@ -71,10 +71,8 @@ AccountsController.getAccounts = function(context, next) {
 				delete account.delegate;
 			} else {
 				account.delegate.rank = parseInt(account.delegate.rank);
-				account.delegate.missedBlocks = parseInt(account.delegate.missedBlocks);
-				account.delegate.producedBlocks = parseInt(
-					account.delegate.producedBlocks
-				);
+				account.delegate.missedBlocks = account.delegate.missedBlocks;
+				account.delegate.producedBlocks = account.delegate.producedBlocks;
```
nazarhussain: If you are not parsing, then you don't need these lines anymore.

api/controllers/delegates.js
```diff
@@ -77,8 +77,8 @@ DelegatesController.getDelegates = function(context, next) {
 			delete delegate.publicKey;
 			delete delegate.address;
 
-			delegate.missedBlocks = parseInt(delegate.missedBlocks);
-			delegate.producedBlocks = parseInt(delegate.producedBlocks);
+			delegate.missedBlocks = delegate.missedBlocks;
+			delegate.producedBlocks = delegate.producedBlocks;
```
nazarhussain: Same here. If you are not parsing, then no need these lines.

db/sql/rounds/restore_votes_snapshot.sql
```diff
@@ -20,6 +20,6 @@
 */
 
 UPDATE mem_accounts m
-SET vote = b.vote
+SET vote = b.vote, "missedBlocks" = b."missedBlocks", "producedBlocks" = b."producedBlocks"
 FROM mem_votes_snapshot b
```
nazarhussain: As you are refactoring, I suggest to use `FROM mem_votes_snapshot snapshot` and `UPDATE mem_accounts accounts`. That will make the query more readable like `SET accounts.vote = snapshot.vote`.

4miners:
@nazarhussain Resolved, please check.

PR #2176, opened 2018-06-28
Merge helpers/milestones.js to network specific exceptions - Closes #2155

### LINKED ISSUE #2155: Merge helpers/milestones.js to network specific exceptions
### Expected behavior
Values specified in `helpers/milestones.js` are sort of constants specifying block height for different networks. Since now we have network specific configurations in one place, so we should merge this helper into `config/exceptions.js` 

### Actual behavior
Seperate file available under `helpers/milestones.js`, which have to manage separately under different branches. 

1.0.0

nazarhussain:
### What was the problem?
There was inconsistent usage of a some static values, which can be considered as exceptions. 

### How did I fix it?
Moved `helpers/milestones.js` content to `config/<network>/exceptions.js`

### How to test it?
Run functional tests 

### Review checklist

* The PR solves #2155
* Linting passes
* Tests pass


nazarhussain:
@diego-G I considered that this configuration value to disable DAPP transactions best reside in exceptions rather than constants. 

Constants are the config values to drive the behavior of the system, where as exceptions are right candidate to apply some specific condition in operations of the system.

config/testnet/exceptions.js
```diff
@@ -59,4 +61,7 @@ module.exports = {
 		'12298100805070303137', // 3057955 - Delegate transaction
 	],
 	transactionFee: [],
+	disableTransactions: {
+		dappTransfer: 5594491, // Disable Dapp Transfer on this block height
```
4miners: Not sure about property name and comment here. We're not supporting dapp transfer transactions anymore, so from logical point of view - exceptions should `allow` those up to certain block height, not `disable` from certain height.

4miners:
@diego-G I cannot agree with you. Maybe milestone is not an exception, but allowing transactions of no longer supported type to be accepted is not a milestone, it's an exception.

nazarhussain:
After discussion with @MaciejBaj decided to keep it in exceptions and use a new naming convention.

nazarhussain:
@diego-G Point taken we will rename it to `disableDappTransfer` that will be appropriate for it. 

Related to exceptions, what we envision after discussion is that: 

> Exceptions in Lisk are checkpoints where we perform certain operations exceptionally (differently), from normal operations. 

So we thought of putting that block height to disable dapp transfer into exceptions.

config/testnet/exceptions.js
nazarhussain: Its the same thing :) 

Think in this direction. We allowed dapp transfer from start of the chain starting from that that particular block height we disabled those.

PR #2185, opened 2018-06-29
Use library cache and notification

abelboldu:
### What was the problem?

Jenkinsfile was using old legacy functions, with old Slack channels. 

### How did I fix it?

Importing shared library to use new functions, with better cache approach. Send notifications to the current CI channel.

### How to test it?

In Jenkins, lisk-core-integration job

### DIFF
--- a/Jenkinsfile.integration
+++ b/Jenkinsfile.integration
@@ -11,22 +11,8 @@
  *
  * Removal or modification of this copyright notice is prohibited.
  */
-def report() {
-	if (currentBuild.result == 'FAILURE') {
-		def prBranch = ''
-		if (env.CHANGE_BRANCH != null) {
-			prBranch = " (${env.CHANGE_BRANCH})"
-		}
-		slackSend color: 'danger', message: "Build #${env.BUILD_NUMBER} of <${env.BUILD_URL}|${env.JOB_NAME}>${prBranch} failed (<${env.BUILD_URL}/console|console>, <${env.BUILD_URL}/changes|changes>)", channel: '#lisk-core-jenkins'
-	}
-}
 
-def statusFail() {
-	currentBuild.result = 'FAILURE'
-	error('Stopping build, build failed')
-	report()
-	cleanUp()
-}
+@Library('lisk-jenkins') _
 
 def cleanUp() {
 	sh '''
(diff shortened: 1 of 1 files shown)


Jenkinsfile.integration
```diff
@@ -50,16 +36,16 @@ node('lisk-integration') {
...
-
-				sh '''
-					# Cycle postgres, install deps and prepare config
-					sudo service postgresql restart
-					rsync -axl -e "ssh -oUser=jenkins" master-01:/var/lib/jenkins/lisk/node_modules/ "$WORKSPACE/node_modules/" || true
-					npm install
-					cp test/data/config.json test/data/genesis_block.json .
-				'''
+				// Cycle postgres, restore chache install deps and prepare config
+				sh 'sudo service postgresql restart'
```
nazarhussain: I remember we avoid running any thing on Jenkins with sudo privileges. @Nazgolze did refactor to remove sudo from all scripts. Please check with him if its really good to use.

Jenkinsfile.integration
```diff
@@ -50,16 +36,16 @@ node('lisk-integration') {
...
+				cache_file = restoreCache("config.json")
```
nazarhussain: With recent changes into core there will be no `config.json` on root directory. Does that really matters here?

Jenkinsfile.integration
```diff
@@ -50,16 +36,16 @@ node('lisk-integration') {
...
+				sh 'npm install'
+				saveCache(cache_file, './node_modules', 7)
+				sh 'cp test/data/config.json test/data/genesis_block.json .'
```
nazarhussain: When this PR will be merged https://github.com/LiskHQ/lisk/pull/2181 these files will be no longer exists. So I suggest we hold this PR, until other one merged.

Jenkinsfile.integration
abelboldu: Yes I'm aware of that, but this is another issue we need to fix for this branch 1.1.0.  For what matters here I suggest to get it merged and work in the config changes in another PR.

Jenkinsfile.integration
Nazgolze: Restarting postgres is the one sudo command that the lisk user is allowed to do on the Jenkins machines

And since the job locks the resource and this one machine only does the integration job, I don't see any problem doing this here

PR #2187, opened 2018-07-01
Store amounts as strings - Closes #716

### LINKED ISSUE #716: Store amounts as strings
All LSK values like balances, fees, amounts in transactions, etc. should be stored as strings in JS and never casted/converted to numbers - in entire application. Calculations should be done via `BigNumber`, database layer is not affected (we use `BIGINT` there).

ManuGowda:
### What was the problem?
balances, fees, amounts in parts of the application were using integer, and strings rather than using bignumber for calculation and comparisons.
### How did I fix it?
balances, fees, amounts are stored as bignumber throughout the application and used bignum for calculations and for another purpose, only during saving to the database it will be converted to string.
### Review checklist

* The PR solves #716 
* Linting passes
* Tests pass


db/repos/blocks.js
```diff
@@ -312,7 +312,9 @@ class BlocksRepository {
 				'hex'
 			);
 			saveBlock.blockSignature = Buffer.from(block.blockSignature, 'hex');
-			saveBlock.reward = block.reward || 0;
+			saveBlock.reward = block.reward.toString() || '0';
```
4miners: Can you check if we really need to do `|| '0'` here?

db/repos/rounds.js
```diff
@@ -88,10 +88,11 @@ class RoundsRepository {
 	 * Update the votes of for a particular account.
 	 *
 	 * @param {string} address - Address of the account
-	 * @param {int} amount - Votes to update
+	 * @param {BigNumber} amount - Votes to update
 	 * @todo Add @returns tag
 	 */
 	updateVotes(address, amount) {
+		amount = amount ? amount.toString() : '0';
```
4miners: Why we need that? Is `amount` not required? If yes we should make it required.

helpers/z_schema.js
```diff
@@ -275,6 +276,33 @@ var liskFormats = {
...
+		}
+		return false;
+	},
+	/**
+	 * Transaction totalAmount, totalFee, reward.
+	 *
+	 * @param {Object} value
+	 * @returns {boolean}
+	 */
+	minAmount(value) {
```
4miners: Why we cannot use `amount` check for that?

logic/account.js
```diff
@@ -498,28 +498,30 @@ class Account {
 						break;
 
 					// [u_]balance, [u_]multimin, [u_]multilifetime, rate, fees, rank, rewards, votes, producedBlocks, missedBlocks
+					// eslint-disable-next-line no-case-declarations
 					case Number:
-						if (isNaN(updatedValue) || updatedValue === Infinity) {
-							throw `Encountered insane number: ${updatedValue}`;
+						const value = new Bignum(updatedValue);
```
4miners: Consistency! Globally `new bignum` or `new Bignum`, preferred last one.

logic/account.js
```diff
@@ -498,28 +498,30 @@ class Account {
...
 
 							// If updated value is negative number
-						} else if (updatedValue < 0) {
+						} else if (value.lessThan(0)) {
 							promises.push(
 								dbTx.accounts.decrement(
 									address,
 									updatedField,
-									Math.floor(Math.abs(updatedValue))
+									value.abs().toString()
```
4miners: Missing `.floor`, possible change of behavior.

logic/block.js
```diff
@@ -87,10 +87,12 @@ class Block {
 				return 1;
 			}
 			// Place depending on amount (lower first)
-			if (a.amount < b.amount) {
+			const prev = new bignum(a.amount);
```
4miners: Why `a.amount` and `b.amount` are not instances of `Bignum` already?

logic/block.js
```diff
@@ -414,9 +416,9 @@ Block.prototype.getBytes = function(block) {
 		}
 
 		byteBuffer.writeInt(block.numberOfTransactions);
-		byteBuffer.writeLong(block.totalAmount);
-		byteBuffer.writeLong(block.totalFee);
-		byteBuffer.writeLong(block.reward);
+		byteBuffer.writeLong(block.totalAmount.toString());
```
4miners: Previously numbers, now strings - it doesn't change the behavior of `byteBuffer.writeLong`?

logic/block.js
```diff
@@ -477,7 +479,7 @@ Block.prototype.getId = function(block) {
  * @todo Delete unused param
  */
 Block.prototype.calculateFee = function() {
-	return constants.fees.send;
+	return constants.fees.send.toString();
```
4miners: Why not return `Bignum` from here?

logic/dapp.js
```diff
@@ -65,7 +66,7 @@ DApp.prototype.bind = function() {};
  * @returns {number} Transaction fee
  */
 DApp.prototype.calculateFee = function() {
-	return constants.fees.dappRegistration;
+	return constants.fees.dappRegistration.toString();
```
4miners: same here, why not `Bignum`?

logic/delegate.js
```diff
@@ -64,7 +65,7 @@ Delegate.prototype.bind = function(accounts) {
  * @todo Delete unused transaction, sender parameters
  */
 Delegate.prototype.calculateFee = function() {
-	return constants.fees.delegate;
+	return constants.fees.delegate.toString();
```
4miners: same

logic/delegate.js
```diff
@@ -81,7 +82,9 @@ Delegate.prototype.verify = function(transaction, sender, cb, tx) {
 		return setImmediate(cb, 'Invalid recipient');
 	}
 
-	if (transaction.amount !== 0) {
+	const amount = new bignum(transaction.amount);
```
4miners: Why not already a `Bignum`?

logic/multisignature.js
```diff
@@ -82,7 +82,7 @@ Multisignature.prototype.calculateFee = function(transaction) {
 	return (
 		(transaction.asset.multisignature.keysgroup.length + 1) *
 		constants.fees.multisignature
-	);
+	).toString();
```
4miners: Perform that multiply using bignums

logic/signature.js
```diff
@@ -61,7 +62,7 @@ Signature.prototype.bind = function(accounts) {
  * @returns {number} Transaction fee
  */
 Signature.prototype.calculateFee = function() {
-	return constants.fees.secondSignature;
+	return constants.fees.secondSignature.toString();
```
4miners: Same here, why not return `Bignum`? Globally.

logic/transaction.js
```diff
@@ -629,7 +627,7 @@ class Transaction {
 				transaction,
 				sender
 			) || false;
-		if (!fee || transaction.fee !== fee) {
+		if (!fee || !transaction.fee.equals(fee)) {
```
4miners: What if `fee` is a bignum instance of 0?

logic/transaction.js
```diff
@@ -642,19 +640,19 @@ class Transaction {
...
 		// Check amount
+		let amount = transaction.amount;
 		if (
-			transaction.amount < 0 ||
-			transaction.amount > constants.totalAmount ||
-			String(transaction.amount).indexOf('.') >= 0 ||
-			transaction.amount.toString().indexOf('e') >= 0
+			amount.lessThan(0) ||
+			amount.greaterThan(constants.totalAmount) ||
+			amount.toString().indexOf('.') >= 0 ||
```
4miners: Bignum should provide better check for decimals.

logic/transaction.js
```diff
@@ -642,19 +640,19 @@ class Transaction {
...
+			amount.toString().indexOf('e') >= 0
```
4miners: same for that

logic/transaction.js
```diff
@@ -1138,11 +1128,11 @@ class Transaction {
 		}
 
 		if (transaction.amount) {
-			transaction.amount = parseInt(transaction.amount);
+			transaction.amount = new bignum(transaction.amount);
```
4miners: `transaction.amount = new bignum(transaction.amount || 0);` and remove conditions

logic/transaction.js
```diff
@@ -1138,11 +1128,11 @@ class Transaction {
...
 		}
 
 		if (transaction.fee) {
-			transaction.fee = parseInt(transaction.fee);
+			transaction.fee = new bignum(transaction.fee);
```
4miners: same

modules/blocks/chain.js
```diff
@@ -230,6 +231,14 @@ Chain.prototype.applyGenesisBlock = function(block, cb) {
 			// Apply transactions through setAccountAndGet, bypassing unconfirmed/confirmed states
 			// FIXME: Poor performance - every transaction cause SQL query to be executed
 			// WARNING: DB_WRITE
+			if (transaction.amount) {
+				transaction.amount = new bignum(transaction.amount);
```
4miners: same

modules/blocks/chain.js
```diff
@@ -539,6 +548,17 @@ Chain.prototype.applyBlock = function(block, saveBlock, cb) {
  * @param {boolean} broadcast - Indicator that block needs to be broadcasted
  */
 Chain.prototype.broadcastReducedBlock = function(reducedBlock, broadcast) {
+	if (reducedBlock.totalAmount) {
```
4miners: Why we need conditions here?

modules/blocks/process.js
```diff
@@ -628,6 +629,17 @@ __private.validateBlockSlot = function(block, lastBlock, cb) {
  * @todo Add @returns tag
  */
 Process.prototype.onReceiveBlock = function(block) {
+	if (block.totalAmount) {
```
4miners: Same here

modules/blocks/verify.js
```diff
@@ -573,22 +573,41 @@ Verify.prototype.deleteBlockProperties = function(block) {
 	}
 	if (reducedBlock.totalAmount === 0) {
```
4miners: That check will pass?

modules/blocks/verify.js
```diff
@@ -573,22 +573,41 @@ Verify.prototype.deleteBlockProperties = function(block) {
...
 		delete reducedBlock.totalAmount;
+	} else {
+		reducedBlock.totalAmount = new bignum(reducedBlock.totalAmount);
```
4miners: There is something wrong with logic here. Before if property was `0` we removed it, now we create bignum instance if it's not `0`. We should always create instance or never create.

modules/blocks/verify.js
```diff
@@ -573,22 +573,41 @@ Verify.prototype.deleteBlockProperties = function(block) {
...
 	}
 	if (reducedBlock.transactions && reducedBlock.transactions.length === 0) {
 		delete reducedBlock.transactions;
+	} else {
+		convertToBigNum(reducedBlock.transactions);
 	}
 	return reducedBlock;
 };
 
+const convertToBigNum = transactions =>
```
4miners: From where those transactions come from? It's verify module - we should already have those as bignums.

modules/node.js
```diff
@@ -148,7 +148,7 @@ Node.prototype.shared = {
 			nethash: library.config.nethash,
 			nonce: library.config.nonce,
 			milestone: blockReward.calcMilestone(height),
-			reward: blockReward.calcReward(height),
+			reward: blockReward.calcReward(height).toString(),
```
4miners: It's for API? Maybe not the best place to cast `.toString()`

modules/transactions.js
```diff
@@ -397,11 +397,26 @@ __private.getPooledTransactions = function(method, filters, cb) {
...
-		[sortAttribute.sortField],
-		[sortAttribute.sortMethod.toLowerCase()]
-	);
+
+	if (
+		sortAttribute.sortField === 'fee' ||
+		sortAttribute.sortField === 'amount'
+	) {
+		const sortFactor =
+			sortAttribute.sortMethod.toLowerCase() === 'desc' ? -1 : 1;
```
4miners: Add description what is going on here.

test/functional/http/post/0.transfer.js
```diff
@@ -124,7 +124,7 @@ describe('POST /api/transactions (type 0) transfer funds', () => {
 
 		it('using entire balance should fail', () => {
 			transaction = lisk.transaction.transfer({
-				amount: Math.floor(accountFixtures.genesis.balance),
+				amount: Math.floor(accountFixtures.genesis.balance) + 10000,
```
4miners: Where is bignum?

test/functional/http/post/0.transfer.js
```diff
@@ -133,9 +133,7 @@ describe('POST /api/transactions (type 0) transfer funds', () => {
 				transaction,
 				errorCodes.PROCESSING_ERROR
 			).then(res => {
-				expect(res.body.message).to.match(
-					/^Account does not have enough LSK: [0-9]+L balance: /
-				);
+				expect(res.body.message).to.match(/^Invalid transaction body -/);
```
4miners: Why error changed?

test/functional/system/blocks/verify/verify.js
```diff
@@ -148,10 +150,18 @@ function createBlock(
 		previousBlock: blocksModule.lastBlock.get(),
 		transactions,
 	});
+
 	// newBlock.id = blockLogic.getId(newBlock);
 	return newBlock;
 }
 
+const convertToBignum = transactions => {
```
4miners: I already saw that function somewhere, can we reuse? Or if not - please move that to some common test file, as for sure we will use that more in the future.

test/functional/system/blocks/verify/verify.js
```diff
@@ -1059,29 +1067,33 @@ describe('blocks/verify', () => {
...
-						]);
-						expect(block2.previousBlock).to.equal(genesisBlock.id);
-						done();
-					});
+							expect(block2.version).to.equal(0);
+							expect(block2.timestamp).to.equal(time);
+							expect(block2.numberOfTransactions).to.equal(1);
+							expect(block2.reward.equals(0));
+							expect(block2.totalFee.equals(0));
+							expect(block2.totalAmount.equals(10000000000000000));
```
4miners: compare with string, globally I think

test/functional/system/common.js
```diff
@@ -21,6 +21,14 @@ var slots = require('../../../helpers/slots');
 var application = require('../../common/application');
 var randomUtil = require('../../common/utils/random');
 var accountFixtures = require('../../fixtures/accounts');
+var bignum = require('../../../helpers/bignum.js');
+
+const convertToBigNum = transactions => {
```
4miners: Yeah, I saw that function already. ;)

test/functional/system/rounds.js
```diff
@@ -437,43 +437,7 @@ describe('rounds', () => {
...
-									} else {
-										// In every other case - fail the test
-										throw err;
-									}
-								}
-							},
-							[]
-						);
-					}
+					expect(tick.after.accounts).to.deep.equal(expected);
```
4miners: ❤️

test/functional/system/rounds.js
```diff
@@ -774,9 +738,13 @@ describe('rounds', () => {
 
 					lastBlock = library.modules.blocks.lastBlock.get();
 					// Delete one block more
-					return deleteLastBlockPromise().then(() => {
-						done();
-					});
+					return deleteLastBlockPromise()
```
4miners: Can we remove `done` and simply return the promise?

test/functional/system/snapshotting.js
```diff
@@ -100,10 +100,14 @@ describe('snapshotting', () => {
 			);
 
 			__private.snapshotFinished = function() {
-				getMemAccounts().then(_accounts => {
-					expect(_accounts).to.deep.equal(memAccountsBeforeSnapshot);
-					done();
-				});
+				getMemAccounts()
+					.then(_accounts => {
+						expect(_accounts).to.deep.equal(memAccountsBeforeSnapshot);
```
4miners: empty line?

test/functional/system/transactions/1.second_signature/1.1.second_signature_from_pool_peer.js
```diff
@@ -69,8 +70,8 @@ describe('system test (type 1) - second signature transactions from pool and pee
 					passphrase: signatureAccount.passphrase,
 					secondPassphrase: signatureAccount.secondPassphrase,
 				});
-				signatureTransaction.amount = parseInt(signatureTransaction.amount);
-				signatureTransaction.fee = parseInt(signatureTransaction.fee);
+				signatureTransaction.amount = new bignum(signatureTransaction.amount);
```
4miners: Why not use that convertToBignum function?

test/functional/system/transactions/1.second_signature/1.1.second_signature_from_pool_peer.js
```diff
@@ -135,8 +136,10 @@ describe('system test (type 1) - second signature transactions from pool and pee
 						secondPassphrase: randomUtil.password(),
 					});
 					signatureTransaction2.senderId = signatureAccount.address;
-					signatureTransaction2.amount = parseInt(signatureTransaction2.amount);
-					signatureTransaction2.fee = parseInt(signatureTransaction2.fee);
+					signatureTransaction2.amount = new bignum(
```
4miners: same here, same globally

test/setup.js
```diff
@@ -34,8 +37,13 @@ chai.use(chaiAsPromised);
 
 var testContext = {};
 
-testContext.config = require('../config/devnet/config.json');
-testContext.config.genesisBlock = require('../config/devnet/genesis_block.json');
+testContext.config = devNetConfig;
```
4miners: What is that?

test/unit/logic/block.js
```diff
@@ -664,7 +664,9 @@ describe('block', () => {
 
 	describe('calculateFee', () => {
 		it('should return the constant fee', () => {
-			return expect(block.calculateFee(blockData)).to.eql(constants.fees.send);
+			return expect(block.calculateFee(blockData)).to.eql(
+				constants.fees.send.toString()
```
4miners: Those in constants should be strings already.

test/unit/logic/block_reward.js
```diff
@@ -148,116 +148,116 @@ describe('BlockReward @slow', () => {
...
+			return expect(
+				blockReward.calcReward(13451520 * 10000).equals('100000000')
 			);
 		});
 
 		it('when height == (milestoneFour * 100000) should return 100000000', () => {
-			return expect(blockReward.calcReward(13451520 * 100000)).to.equal(
-				100000000
+			return expect(
+				blockReward.calcReward(13451520 * 100000).equals('100000000')
```
4miners: Perform multiply using bignum

test/unit/logic/transaction.js
```diff
@@ -471,7 +471,8 @@ describe('transaction', () => {
 				recipientId: transactionData.recipientId,
 			};
 			var transaction = lisk.transaction.transfer(transferObject);
-			transaction.fee = Number(transaction.fee);
+			transaction.amount = new bignum(transaction.amount);
```
4miners: convertToBignum?

db/repos/rounds.js
ManuGowda: @4miners yes amount is required, however, we write weird tests to support undefined amount values here https://github.com/LiskHQ/lisk/blob/133036a05e8eb09397586af0f4d2aadd4e1ebd8a/test/unit/db/repos/rounds.js#L290
previously this test was working, because amount was integer, now its bignum and i convert that to `toString` before we save. so let me know weather to remove this invalid test or to keep the exisisting code.

helpers/z_schema.js
ManuGowda: @4miners I don't see any JSON schema validation available for amount type http://json-schema.org/latest/json-schema-validation.html, please refer me to any information where it specifies we can use amount, I will be glad to use and update the same.

logic/block.js
ManuGowda: writeLong internally calls `writeInt64` and handles the Long value accordingly, here is the gist
```
if (typeof value === 'number')
                    value = Long.fromNumber(value);
                else if (typeof value === 'string')
                    value = Long.fromString(value);
```

modules/blocks/chain.js
ManuGowda: because every block might not have `totalAmount`, so checking it, modified the code to 
```
reducedBlock.totalAmount = reducedBlock.totalAmount.toString() || '0';
reducedBlock.totalFee = reducedBlock.totalFee.toString() || '0';
reducedBlock.reward = reducedBlock.reward.toString() || '0';
```

4miners: That function is only calling the event, why modify block object here before broadcast? That is for sure not the proper place for doing that (it should be when received reduced block or when calling object normalize, etc.).

helpers/z_schema.js
4miners: I mean we can use `amount` check instead of `minAmount` - the only difference is `amount` checking also for max amount, which should be fine.

logic/account.js
```diff
@@ -498,28 +498,33 @@ class Account {
...
 
 						// If updated value is positive number
-						if (Math.abs(updatedValue) === updatedValue && updatedValue !== 0) {
+						if (value.greaterThan(0)) {
 							promises.push(
 								dbTx.accounts.increment(
 									address,
 									updatedField,
-									Math.floor(updatedValue)
+									value.floor().toString()
```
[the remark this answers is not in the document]
ManuGowda: actually, it's not required here, as we are already validating if amount is a valid integer, however, I doesn't harm much to keep it as it, as we need to do a thorough check before we remove.

helpers/round_changes.js
```diff
@@ -39,13 +39,13 @@ function RoundChanges(scope) {
 	if (exceptions.rounds[scope.round]) {
 		// Apply rewards factor
 		this.roundRewards.forEach((reward, index) => {
-			this.roundRewards[index] = new bignum(reward.toPrecision(15))
+			this.roundRewards[index] = new Bignum(reward.toPrecision(15))
```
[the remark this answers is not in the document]
ManuGowda: The reason why we have precision here to validate if the rewards exceed the total amount, which is if the rewards is too high then the `new Bignum(reward.toPrecision(15))` would become an exponential notation rather than fixed point.
```
var aa = new bignum(500000000)
aa.toPrecision(15) ====> '500000000.000000'
```
in case of exceeding 
```
var aa = new bignum(5000000000000000)
aa.toPrecision(15) ====> '5.00000000000000e+15'
```
conclusion: The precesion needs to exisits.

modules/blocks/verify.js
```diff
@@ -563,24 +558,24 @@ Verify.prototype.addBlockProperties = function(block) {
  * @returns {Object} Block object reduced
  */
 Verify.prototype.deleteBlockProperties = function(block) {
-	const reducedBlock = JSON.parse(JSON.stringify(block));
+	const reducedBlock = Object.assign({}, block);
```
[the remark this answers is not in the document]
ManuGowda: its completely the wrong way to clone a object, if you do stringify and parse then it would change the object properties like, if its a bignum then it apply string and convert, so the intension is to clone not to modify the state, so I have changed it to the correct way

test/common/utils/random.js
```diff
@@ -132,4 +135,13 @@ random.password = function() {
 		.substring(7);
 };
 
+const convertToBignum = transactions => {
```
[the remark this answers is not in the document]
ManuGowda: I wanted to keep this function in root common utils, so there is no other files which makes sense place this function other than random.

modules/accounts.js
```diff
@@ -15,7 +15,7 @@
 'use strict';
 
 const crypto = require('crypto');
-const bignum = require('../helpers/bignum.js');
+const Bignum = require('../helpers/bignum.js');
```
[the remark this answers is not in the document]
ManuGowda: changed it to bignum.js everywhere

db/repos/rounds.js
```diff
@@ -88,11 +88,11 @@ class RoundsRepository {
 	 * Update the votes of for a particular account.
 	 *
 	 * @param {string} address - Address of the account
-	 * @param {int} amount - Votes to update
+	 * @param {BigNumber} amount - Votes to update
```
MaciejBaj: I think it will be more simple to test/maintain the functions if we would always pass `string` into it and convert it to BigNumber when needed. It would also simplify the possible external tools basing on JsDOC to create documentation/typing etc.

4miners: No, there it was decided to maintain `Bignumber` instances whatever possible and convert to `string` when needed (API, save to database).

ManuGowda: Yes, since we keeping bignum over the application layer, whenever its required like frontend or database it can be converted to the string, this way we are 100% sure we are operating with bignum all over the application.

MaciejBaj: Ok then :)

Issue #2168, opened 2018-06-26
Improve vote transaction verification/processing efficiency


SargeKhan:
### Expected behavior
For each step in verify, applyUnconfirmed, apply all votes should be verified in parallel.
### Actual behavior
For vote transaction, each of the steps i.e. verify, applyUnconfirmed, apply, we independently verify each vote (with database calls). So, for a vote transaction with 33 votes, it will make 33 database calls in series for verifying a vote. Total DB calls (only for verifying the vote, and not other necessary checks) for each of these steps is 33 * 3 = 99. And all of these database calls are run in series.
### Steps to reproduce
N/A
### Which version(s) does this affect? (Environment, OS, etc...)
1.0.0

PR #2189, opened 2018-07-02
Improve vote transaction verification performance

SargeKhan:
### What was the problem?
The vote transaction verification was slow because the dependent rows for validating a transaction were fetched from the database in series. 
### How did I fix it?
When verifying a transaction, the application now aggregates addresses of all accounts in memory first and then runs one query to fetch all accounts.
### How to test it?
Create alpha-net and run vote transactions stress tests. Compare performance with the older version. 
### Review checklist
Check behavior is exactly the same.
Check there are significant performance improvements.
* The PR solves #2168
* Linting passes
* Tests pass

### DIFF
--- a/modules/delegates.js
+++ b/modules/delegates.js
@@ -353,112 +353,146 @@ __private.decryptPassphrase = function(encryptedPassphrase, password) {
  * @returns {setImmediateCallback} cb, err
  * @todo Add description for the params
  */
-__private.checkDelegates = function(publicKey, votes, state, cb, tx) {
+__private.checkDelegates = function(senderPublicKey, votes, state, cb, tx) {
 	if (!Array.isArray(votes)) {
 		return setImmediate(cb, 'Votes must be an array');
 	}
 
-	modules.accounts.getAccount(
-		{ publicKey },
-		(err, account) => {
-			if (err) {
-				return setImmediate(cb, err);
+	// Converting publicKeys to addresses because module.accounts do not support fetching multiple accounts by publicKeys
+	const actionWithAddresses = votes.map(vote => {
+		const action = vote[0];
+		let voteAddress;
+
+		try {
+			if (action !== '+' && action !== '-') {
+				throw 'Invalid math operator';
 			}
+			const votePublicKey = vote.slice(1);
 
-			if (!account) {
-				return setImmediate(cb, 'Account not found');
-			}
+			voteAddress = modules.accounts.generateAddressByPublicKey(votePublicKey);
+		} catch (e) {
+			library.logger.error(e.stack);
+			return setImmediate(cb, e);
+		}
+		return {
+			action,
+			address: voteAddress,
+		};
+	});
 
-			const delegates =
-				state === 'confirmed' ? account.delegates : account.u_delegates;
-			const existingVotes = Array.isArray(delegates) ? delegates.length : 0;
-			let additions = 0;
-			let removals = 0;
+	async.waterfall(
+		[
+			function getVotedAccounts(waterfallCb) {
+				modules.accounts.getAccount(
+					{ publicKey: senderPublicKey },
+					(err, account) => {
+						if (err) {
+							return setImmediate(waterfallCb, err);
+						}
 
-			async.eachSeries(
-				votes,
-				(action, cb) => {
-					const math = action[0];
+						if (!account) {
+							return setImmediate(waterfallCb, 'Account not found');
+						}
 
-					if (math === '+') {
-						additions += 1;
-					} else if (math === '-') {
-						removals += 1;
-					} else {
-						return setImmediate(cb, 'Invalid math operator');
-					}
+						const delegates =
+							state === 'confirmed' ? account.delegates : account.u_delegates;
+						const votedAddresses = Array.isArray(delegates)
+							? delegates.map(delegate =>
+									modules.accounts.generateAddressByPublicKey(delegate)
+								)
+							: [];
 
-					const publicKey = action.slice(1);
+						return setImmediate(waterfallCb, null, votedAddresses);
+					},
+					tx
+				);
+			},
+			function validateVotes(existingVotedAddresses, waterfallCb) {
+				modules.accounts.getAccounts(
+					{
+						address: actionWithAddresses.map(({ address }) => address),
+						isDelegate: 1,
+					},
+					(err, votesAccounts) => {
+						if (err) {
+							return setImmediate(waterfallCb, err);
+						}
 
-					try {
-						Buffer.from(publicKey, 'hex');
-					} catch (e) {
-						library.logger.error(e.stack);
-						return setImmediate(cb, 'Invalid public key');
-					}
+						if (
+							!votesAccounts ||
+							votesAccounts.length < actionWithAddresses.length
+						) {
+							return setImmediate(waterfallCb, 'Delegate not found');
+						}
 
-					modules.accounts.getAccount(
-						{ publicKey, isDelegate: 1 },
-						(err, account) => {
-							if (err) {
-								return setImmediate(cb, err);
-							}
-
-							if (!account) {
-								return setImmediate(cb, 'Delegate not found');
-							}
-
-							if (
-								math === '+' &&
-								(delegates != null && delegates.indexOf(publicKey) !== -1)
-							) {
-								return setImmediate(
-									cb,
-									`Failed to add vote, delegate "${
-										account.username
-									}" already voted for`
-								);
-							}
-
-							if (
-								math === '-' &&
-								(delegates === null || delegates.indexOf(publicKey) === -1)
-							) {
-								return setImmediate(
-									cb,
-									`Failed to remove vote, delegate "${
-										account.username
-									}" was not voted for`
-								);
-							}
-
-							return setImmediate(cb);
-						},
-						tx
-					);
-				},
-				err => {
-					if (err) {
-						return setImmediate(cb, err);
-					}
-
-					const totalVotes = existingVotes + additions - removals;
-
-					if (totalVotes > constants.activeDelegates) {
-						const exceeded = totalVotes - constants.activeDelegates;
-
-						return setImmediate(
-							cb,
-							`Maximum number of ${
-								constants.activeDelegates
-							} votes exceeded (${exceeded} too many)`
+						const upvoteAccounts = votesAccounts.filter(voteAccount =>
+							actionWithAddresses.find(
+								actionWithAddress =>
+									actionWithAddress.action === '+' &&
+									actionWithAddress.address === voteAccount.address
+							)
 						);
-					}
-					return setImmediate(cb);
-				}
-			);
-		},
-		tx
+						const downvoteAccounts = votesAccounts.filter(voteAccount =>
+							actionWithAddresses.find(
+								actionWithAddress =>
+									actionWithAddress.action === '-' &&
+									actionWithAddress.address === voteAccount.address
+							)
+						);
+
+						const invalidUpvoteAccount = upvoteAccounts.find(
+							voteAccount =>
+								existingVotedAddresses.find(
+									existingVotedAddress =>
+										voteAccount.address === existingVotedAddress
+								) !== undefined
+						);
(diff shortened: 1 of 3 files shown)


modules/delegates.js
```diff
@@ -353,112 +353,146 @@ __private.decryptPassphrase = function(encryptedPassphrase, password) {
...
+
+						if (!account) {
+							return setImmediate(waterfallCb, 'Account not found');
+						}
+
+						const delegates =
+							state === 'confirmed' ? account.delegates : account.u_delegates;
+						const votedAddresses = Array.isArray(delegates)
+							? delegates.map(delegate =>
+									modules.accounts.generateAddressByPublicKey(delegate)
```
4miners: Maybe try...catch it?

modules/delegates.js
```diff
@@ -353,112 +353,146 @@ __private.decryptPassphrase = function(encryptedPassphrase, password) {
...
-					if (totalVotes > constants.activeDelegates) {
-						const exceeded = totalVotes - constants.activeDelegates;
+	async.waterfall(
+		[
+			function getVotedAccounts(waterfallCb) {
+				modules.accounts.getAccount(
+					{ publicKey: senderPublicKey },
+					(err, account) => {
+						if (err) {
+							return setImmediate(waterfallCb, err);
```
4miners: Check if there is proper message when logging.

modules/delegates.js
```diff
@@ -353,112 +353,146 @@ __private.decryptPassphrase = function(encryptedPassphrase, password) {
...
-				},
-				err => {
-					if (err) {
-						return setImmediate(cb, err);
-					}
-
-					const totalVotes = existingVotes + additions - removals;
+			voteAddress = modules.accounts.generateAddressByPublicKey(votePublicKey);
+		} catch (e) {
+			library.logger.error(e.stack);
```
4miners: also some additional message about source

modules/delegates.js
```diff
@@ -353,112 +353,146 @@ __private.decryptPassphrase = function(encryptedPassphrase, password) {
...
+					(err, votesAccounts) => {
+						if (err) {
+							return setImmediate(waterfallCb, err);
+						}
+
+						if (
+							!votesAccounts ||
+							votesAccounts.length < actionWithAddresses.length
+						) {
+							return setImmediate(waterfallCb, 'Delegate not found');
```
4miners: Please add additional log here with inconsistent data.

modules/delegates.js
```diff
@@ -353,112 +353,146 @@ __private.decryptPassphrase = function(encryptedPassphrase, password) {
...
+							)
+						);
 
-						return setImmediate(
-							cb,
-							`Maximum number of ${
-								constants.activeDelegates
-							} votes exceeded (${exceeded} too many)`
+						const invalidUpvoteAccount = upvoteAccounts.find(
+							voteAccount =>
```
4miners: review logic (performance)

4miners:
Instead of making alphanet tests those changes can be tested locally with higher accuracy.

modules/delegates.js
SargeKhan: We are getting this information from the database. So, I think it is not required here.

modules/delegates.js
SargeKhan: e.stack is very descriptive. For example, 
> x = new Error('database error')
> x.stack
'Error: database error\n    at repl:1:5\n    at sigintHandlersWrap (vm.js:22:35)\n    at sigintHandlersWrap (vm.js:73:12)\n    at ContextifyScript.Script.runInThisContext (vm.js:21:12)\n    at REPLServer.defaultEval (repl.js:340:29)\n    at bound (domain.js:280:14)\n    at REPLServer.runBound [as eval] (domain.js:293:12)\n    at REPLServer.<anonymous> (repl.js:539:10)\n    at emitOne (events.js:101:20)\n    at REPLServer.emit (events.js:188:7)'
It gives all the details about the error.

modules/delegates.js
SargeKhan: The log message is what is returned from accounts repo. And it's the same as previous implementation. I think we should not change it.

modules/delegates.js
SargeKhan: There can be one performance improvement. But I think it'll make the code more complicated. I think performance benefit is not so much.

modules/delegates.js
```diff
@@ -353,112 +353,160 @@ __private.decryptPassphrase = function(encryptedPassphrase, password) {
  * @returns {setImmediateCallback} cb, err
  * @todo Add description for the params
  */
-__private.checkDelegates = function(publicKey, votes, state, cb, tx) {
```
nazarhussain: `publicKey` is just identifier of account for which you want to check the votes. Renaming it to `senderPublicKey` make it confusing.

modules/delegates.js
nazarhussain: Yes I agree with @4miners we need to have a descriptive error message here, as we are doing everywhere else in application. May something like: 

> Invalid vote format found in transaction

modules/delegates.js
```diff
@@ -353,112 +353,160 @@ __private.decryptPassphrase = function(encryptedPassphrase, password) {
...
-				err => {
-					if (err) {
-						return setImmediate(cb, err);
-					}
-
-					const totalVotes = existingVotes + additions - removals;
+			voteAddress = modules.accounts.generateAddressByPublicKey(votePublicKey);
+		} catch (e) {
+			library.logger.error(e.stack);
+			return setImmediate(cb, e);
```
nazarhussain: Does this `return` will stop the consequent iterations of `votes.map` on line 363 ?

modules/delegates.js
```diff
@@ -353,112 +353,160 @@ __private.decryptPassphrase = function(encryptedPassphrase, password) {
...
 
-	modules.accounts.getAccount(
-		{ publicKey },
-		(err, account) => {
-			if (err) {
-				return setImmediate(cb, err);
-			}
+	// Converting publicKeys to addresses because module.accounts do not support fetching multiple accounts by publicKeys
+	// TODO: Use Hashmap to improve performance further.
+	const actionWithAddresses = votes.map(vote => {
```
nazarhussain: More readable name suggestion is `voteAddressesWithActions`

modules/delegates.js
```diff
@@ -353,112 +353,160 @@ __private.decryptPassphrase = function(encryptedPassphrase, password) {
...
+									actionWithAddresses,
+									votesAccounts,
+									(
+										{ address: addressFromTransaction },
+										{ addressFromAccount }
+									) => addressFromTransaction === addressFromAccount
+								)
+							);
+							return setImmediate(waterfallCb, 'Delegate not found');
+						}
```
nazarhussain: There are two different steps of validation: 

1. The votes (or public keys) appears valid delegates 
2. Then validate individual vote

So the flow should be implemented as: 

1. Parse all votes to strip the actions 
2. Convert all public keys to addresses 
3. Check if any of those are not valid delegate 
4. Fetch all existing votes from that account 
5. Validate individual vote 

I believe that's the more efficient and effective flow to validate votes. 

Currently the order is implemented as 1, 2, 4, 3, 5.

modules/delegates.js
SargeKhan: Every account has pubicKey, and previously we have variable shadowing for `publicKey`. It makes sense to have a more descriptive name. So we avoid making same mistakes again.

modules/delegates.js
SargeKhan: If there is an exception thrown from https://github.com/LiskHQ/lisk/pull/2189/files#diff-9d5cd48f1ba4ec6b2770fd44edc90323R369, we will have a message `Invalid math operator`. And also, `modules.accounts.generateAddressByPublicKey` throws `Invalid public key: ${publicKey}` in case of invalid publicKey. This is as descriptive as it gets.

modules/delegates.js
SargeKhan: Yes. Since javascript is single threaded and there are no asynchronous calls in this `map`. It will not continue any further. But I get your point that it makes the code look more imperative. Do you have any suggestions? I could have one `.each` for verifying and another `map` for converting votes to actionsWithAddresses but that seemed less performant.

modules/delegates.js
SargeKhan: Why do you say its more efficient? I agree that the flow seems more consistent in the order you described it. But 4 fetches 1 record, and 3 will fetch 33 records. So, it's better to perform step 4  because it requires less I/O.

modules/delegates.js
SargeKhan: Why?

SargeKhan:
@4miners, I created a custom system test for this feature in order to benchmark performance of validating votes. The test logs the time it takes for 25 vote transactions (with 33 votes each) to be processed in the unconfirmed state. The results of the test was as following:
With version 1.1.0, the time it took was: **2181.467ms**
With this PR (#2189), the time it took was: **563.820ms**
This result was consistent with multiple runs of the same test. 

Code is as follow:  

```
'use strict';

var async = require('async');
var lisk = require('lisk-elements').default;
var accountFixtures = require('../../../fixtures/accounts');
var randomUtil = require('../../../common/utils/random');
var constants = require('../../../../helpers/constants');
var localCommon = require('../common');
var genesisDelegates = require('../../../data/genesis_delegates.json').delegates;

describe('system test (type 3) - voting with performance test', () => {
	var library;
	localCommon.beforeBlock('system_3_3_votes', lib => {
		library = lib;
	});

	// performance checks
	describe('check performance of full vote transactions', () => {
		let votersAccounts;
		before('send money to voters accounts', done => {
			votersAccounts = _.range(25).map(() => randomUtil.account());

			const transferTransactions = votersAccounts.map(account => lisk.transaction.transfer({
				amount: 25000 * constants.normalizer,
				passphrase: accountFixtures.genesis.passphrase,
				recipientId: account.address,
			}));

			localCommon.addTransactionsAndForge(library, transferTransactions, done);
		});

		describe('create votes for accounts', () => {
			let voteTransactions;

			before('create vote transactions', done => {
				voteTransactions = votersAccounts.map(account => lisk.transaction.castVotes({
					passphrase: account.passphrase,
					votes: (_.uniq(_.sortBy(genesisDelegates, () => Math.floor(Math.random() * 101)))).map(delegate => delegate.publicKey).splice(0, 33),
					timeOffset: -10000,
				}));
				done();
			});

			it('measure time for forging a block', done => {
				console.time('forging block with 25 vote transactions');
				async.each(voteTransactions, (voteTransaction, eachCb) => {
					localCommon.addTransaction(library, voteTransaction, err => {
						eachCb(err);
					});
				}, err => {
					expect(err).to.not.exist;
					console.timeEnd('forging block with 25 vote transactions');
					done();
				});
			});
		});
	});
});
```

modules/delegates.js
nazarhussain: I suggest to move `try/catch` outside `map` and use the callback there.

modules/delegates.js
nazarhussain: Its a suggestion. To me it's more readable.

modules/delegates.js
nazarhussain: As if something failed in step 3, we don't need to go for step 4.

SargeKhan: It's a matter of preference. And I prefer the current approach. Let me know if you strongly feel about this. If so, we can discuss it further.

nazarhussain: I still feel that its not matter of the preference. Means its not a syntax or js feature usage. Its a business flow, if you feel that above steps are the right order, then we must go with those. 

If you feel above steps are not in correct order, then we can discuss further.

MaciejBaj: The well-informed decision about the order would have to rely on studies - is it more probable to receive a transaction that exceeds the votes limit (4. Fetch all existing votes from that account) or votes for accounts not being delegates (3. Check if any of those are not valid delegate). 
My intuition basing on the current situation says the exceeding the votes limit is more common situation. It has to be done first then to avoid unnecessary following steps. 
Then the order implemented by Usman is valid.

PR #2191, opened 2018-07-02
Remove "Broadhash consensus now" log messages - Closes #2263

### LINKED ISSUE #2263: Remove "Broadhash consensus now" log messages
### Expected behavior

See #2191 

### Actual behavior

See #2191 

### Steps to reproduce

See #2191 

Lisk 1.x

webmaster128:
### What was the problem?

The log message "Broadhash consensus now" is very misleading for users who consider it the primary source of broadhash consensus observation, since it is in the info log level.

There are multiple issues with the log command:
1. it is misplaces because getPeers is a helper that can be called from multiple places and not a lifecycle method. The log content is unrelated to the functionality of getPeers.
2. it is delayed, since the logged value is not calculated before the call. So it can be multiple seconds old
3. it only represents half (or less) of the consensus values available. Since getPeers is usually called in a block broadcast step, it is called about every 10 seconds. But broadhash consensus is recalculated every 5 seconds, which covers different positions within the 10 second slot instead of one.

### How did I fix it?

- delete the misplaces log command
- if broadhash consensus logging should remain in the app lifecycle every 10 seconds, create a new logging with the text "Last broadhash consensus was" instead of "Broadhash consensus now". This should be called no matter if broadcasting happens or not. However, having a log always at the same second of a slot can lead to problematic values (in second 0, I might be the only one that received a fresh block).
- if the best available data should be logged to info, upgrade log level in the [calculateConsensus job](https://github.com/LiskHQ/lisk/blob/v1.0.0-rc.1/modules/peers.js#L876) from debug to info.

I choose to implement the first point only since it avoids spam and let's users find the true broadhash consensus via debug logging.

### How to test it?

See this user logs:

```
[dbg] 2018-07-01 22:30:07 | Broadhash consensus: 100 %
[dbg] 2018-07-01 22:30:12 | Broadhash consensus: 100 %
[dbg] 2018-07-01 22:30:17 | Broadhash consensus: 100 %
[inf] 2018-07-01 22:30:20 | Broadhash consensus now 100 %
[dbg] 2018-07-01 22:30:22 | Broadhash consensus: 93 %
[dbg] 2018-07-01 22:30:27 | Broadhash consensus: 94 %
[dbg] 2018-07-01 22:30:32 | Broadhash consensus: 95 %
[dbg] 2018-07-01 22:30:37 | Broadhash consensus: 98 %
[inf] 2018-07-01 22:30:40 | Broadhash consensus now 98 %
[dbg] 2018-07-01 22:30:42 | Broadhash consensus: 89 %
[dbg] 2018-07-01 22:30:47 | Broadhash consensus: 100 %
[inf] 2018-07-01 22:30:50 | Broadhash consensus now 100 %
[dbg] 2018-07-01 22:30:52 | Broadhash consensus: 83 %
[dbg] 2018-07-01 22:30:57 | Broadhash consensus: 82 %
[inf] 2018-07-01 22:31:00 | Broadhash consensus now 82 %
[dbg] 2018-07-01 22:31:02 | Broadhash consensus: 92 %
[dbg] 2018-07-01 22:31:07 | Broadhash consensus: 93 %
[dbg] 2018-07-01 22:31:12 | Broadhash consensus: 96 %
[dbg] 2018-07-01 22:31:17 | Broadhash consensus: 98 %
```

### Review checklist

* The PR solves #2263
* Linting passes
* Tests pass

### DIFF
--- a/modules/loader.js
+++ b/modules/loader.js
@@ -760,7 +760,7 @@ __private.sync = function(cb) {
 
 	async.series(
 		{
-			getPeersBefore(seriesCb) {
+			getConsensusBefore(seriesCb) {
 				library.logger.debug(
 					`Establishing broadhash consensus before sync: ${modules.peers.getLastConsensus()} %`
 				);
@@ -777,7 +777,7 @@ __private.sync = function(cb) {
 				// Notify all remote peers about our new headers
 				modules.transport.broadcastHeaders(seriesCb);
 			},
-			getPeersAfter(seriesCb) {
+			calculateConsensusAfter(seriesCb) {
 				library.logger.debug(
 					`Establishing broadhash consensus after sync: ${modules.peers.calculateConsensus()} %`
 				);
--- a/modules/loader.js
+++ b/modules/loader.js
@@ -761,7 +761,9 @@ __private.sync = function(cb) {
 	async.series(
 		{
 			getPeersBefore(seriesCb) {
-				library.logger.debug(`Establishing broadhash consensus before sync: ${ modules.peers.calculateConsensus() } %`);
+				library.logger.debug(
+					`Establishing broadhash consensus before sync: ${modules.peers.getLastConsensus()} %`
+				);
 				return seriesCb();
 			},
 			loadBlocksFromNetwork(seriesCb) {
@@ -776,7 +778,9 @@ __private.sync = function(cb) {
 				modules.transport.broadcastHeaders(seriesCb);
 			},
 			getPeersAfter(seriesCb) {
-				library.logger.debug(`Establishing broadhash consensus after sync: ${ modules.peers.calculateConsensus() } %`);
+				library.logger.debug(
+					`Establishing broadhash consensus after sync: ${modules.peers.calculateConsensus()} %`
+				);
 				return seriesCb();
 			},
 		},
--- a/modules/loader.js
+++ b/modules/loader.js
@@ -761,11 +761,8 @@ __private.sync = function(cb) {
 	async.series(
 		{
 			getPeersBefore(seriesCb) {
-				library.logger.debug('Establishing broadhash consensus before sync');
-				return modules.transport.getPeers(
-					{ limit: constants.maxPeers },
-					seriesCb
-				);
+				library.logger.debug(`Establishing broadhash consensus before sync: ${ modules.peers.calculateConsensus() } %`);
+				return seriesCb();
 			},
 			loadBlocksFromNetwork(seriesCb) {
 				return __private.loadBlocksFromNetwork(seriesCb);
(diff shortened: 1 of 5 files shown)


webmaster128:
The log message [is coming from a time](https://github.com/LiskHQ/lisk/blame/ab94ea1c4673d66452467b99b57957aaff70ba21/logic/broadcaster.js#L63) where getPeers did update a consesus state, which is long gone in this part of the code.

4miners:
Yeah, that log is misleading and not needed at all, we should log broadhash consensus only when we're recalculating it, so here is fine:
https://github.com/LiskHQ/lisk/blob/9dcb55e41962282d044a96e74b416445abf8ddd8/modules/peers.js#L876-L878

diego-G:
@webmaster128 please, follow the guidelines to contribute. I'm missing the issue related to this PR. Could you open it? That way we can keep track, open discussion and tackle them quicker.

webmaster128:
@diego-G of course, yes

SargeKhan:
@4miners I think we can remove it. Good job @webmaster128 on understanding the issue. I think if we are removing the log messages from getPeers, we should upgrade the log level to info in the calculateConsensus job. Since this job is configurable, users can increase interval of the job if they feel it is bloating their logs.

webmaster128:
> @webmaster128 let us know if you are eager to proceed with the changes. Otherwise we would be happy to finish the work you started.

Thanks for taking over. I don't plan to do any further changes to the sync logging or broadhash calculation. I also don't care if consensus log is debug or info. But maybe it is better to pull out extra improvements and merge this as it is for the reasons I described in the original PR description.

modules/loader.js
```diff
@@ -761,11 +761,10 @@ __private.sync = function(cb) {
 	async.series(
 		{
 			getPeersBefore(seriesCb) {
-				library.logger.debug('Establishing broadhash consensus before sync');
-				return modules.transport.getPeers(
-					{ limit: constants.maxPeers },
-					seriesCb
+				library.logger.debug(
+					`Establishing broadhash consensus before sync: ${modules.peers.getLastConsensus()} %`
```
SargeKhan: Not gonna call `modules.transport.getPeers` anymore? If not, why call it `getPeersBefore`?

modules/loader.js
```diff
@@ -779,11 +778,10 @@ __private.sync = function(cb) {
 				modules.transport.broadcastHeaders(seriesCb);
 			},
 			getPeersAfter(seriesCb) {
-				library.logger.debug('Establishing broadhash consensus after sync');
-				return modules.transport.getPeers(
-					{ limit: constants.maxPeers },
-					seriesCb
+				library.logger.debug(
+					`Establishing broadhash consensus after sync: ${modules.peers.calculateConsensus()} %`
```
SargeKhan: same question here.

modules/loader.js
MaciejBaj: True, the loader's steps are now calculating consensus only, so if we decide to keep calculating consensus before and after sync it should be renamed 'getConsensusBefore' and 'calculateConsensusAfter'.

PR #2195, opened 2018-07-03
Remove passing the scope of transaction logic to tx types - Closes #750

### LINKED ISSUE #750: Remove the usage of this variables in all transaction types
We call functions of transactions types by assigning the value of `this` variable from main `/logic/transaction.js` file. For example, [In multisignature transaction objectNormalize function](https://github.com/LiskHQ/lisk/blob/development/logic/multisignature.js#L352), we have:
```
Multisignature.prototype.objectNormalize = function (trs) {
	var report = library.schema.validate(trs.asset.multisignature, Multisignature.prototype.schema);

	if (!report) {
		throw 'Failed to validate multisignature schema: ' + this.scope.schema.getLastErrors().map(function (err) {
			return err.message;
		}).join(', ');
	}

	return trs;
};
```
We use `library.schema.validate` for generating report, whereas we use `this.scope.schema.` to get the errors. We do not define `this.scope` variable in `multisignature.js` file, it works because we call `objectNormalize` of multisignature trs from `logic/transaction.js` file which has already defined `this.scope` variable. It works, but it is inconsistent and makes the code hard to understand.

Tschakki:
### What was the problem?
Passing the scope of `logic/transaction.js` to every transaction type when calling their specific function made the usage of `this` confusing.

### How did I fix it?
Removed usage of `this` from all transaction types, changed function calls in `logic/transaction.js` to call the functions without passing the scope.

### How to test it?
Run tests and see if they still pass, e.g.:
```npm test -- mocha:untagged:unit```

### Review checklist

* The PR solves #750 
* Linting passes
* Tests pass

### CODE REVIEW

**Comment by SargeKhan** [comment]
@Tschakki is there any particular reason why git diff shows changes in getBytes, undo, ready etc functions? And can you revert these changes? It makes reviewing pull request harder.

**Comment by Tschakki** [comment]
@SargeKhan yes there is the reason, that we use the eslint rule `class-methods-use-this`, and so every class method is enforced to use `this` or declare themselves as static function.

For better readability I put the function back in their place and added now eslint ignore tags.

For the future, it might make sense to make these particular function static, but not as part of this PR :)

File: logic/transaction.js [hunk 1]
```diff
@@ -1110,6 +1098,54 @@ class Transaction {
...
+	 * Calls `dbRead` based on transaction type (privateTypes) to add tr asset.
+	 *
+	 * @see {@link privateTypes}
+	 * @param {Object} raw
+	 * @throws {string} If unknown transaction type
+	 * @returns {null|transaction}
+	 * @todo Add description for the params
+	 */
+	/* eslint-disable class-methods-use-this */
+	dbRead(raw) {
```
SargeKhan: I think you missed this one. :)

logic/vote.js
```diff
@@ -43,68 +43,69 @@ let self;
...
+ * sender address with inverted votes as delegates.
+ *
+ * @param {transaction} transaction
+ * @param {block} block
+ * @param {account} sender
+ * @param {function} cb - Callback function
+ * @returns {SetImmediate} error
+ * @todo Add description for the params
+ */
+Vote.prototype.undo = function(transaction, block, sender, cb, tx) {
```
SargeKhan: And here.

logic/vote.js
```diff
@@ -43,68 +43,69 @@ let self;
...
+ * Calls Diff.reverse to change asset.votes signs and merges account to
+ * sender address with inverted votes as unconfirmed delegates.
+ *
+ * @param {transaction} transaction
+ * @param {account} sender
+ * @param {function} cb - Callback function
+ * @returns {SetImmediate} error
+ * @todo Add description for the params
+ */
+Vote.prototype.undoUnconfirmed = function(transaction, sender, cb, tx) {
```
SargeKhan: And here.

PR #2196, opened 2018-07-03
Clean the version and minVersion config from all config files - Closes #2184

### LINKED ISSUE #2184: Clean the version and minVersion config from all config files
### Expected behavior
`version` of the lisk itself is identifier for the software so should not be configureable. So instead `version` inside `package.json` is the right place to specify the version of the current software package. 

Also the `minVersion` should be set as `constant` value which is shipped along with the build of the system. It should not be configurable value. 

### Actual behavior
Currently `version` and `minVersion` are both set in `config.json`

1.0.x

nazarhussain:
### What was the problem?
`version` of the lisk itself is identifier for the software so should not be configureable. So instead `version` inside `package.json` is the right place to specify the version of the current software package. 

### How did I fix it?
Removed the `version` and `minVersion` from all configuration files and moved it to `package.json`

### How to test it?
Run core and integration tests. 

### Review checklist

* The PR solves #2184
* Linting passes
* Tests pass

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -47,8 +47,8 @@ RUN curl --silent --show-error --location --output /tmp/confd \
 	exit 1; \
     fi
 
-RUN LISK_VERSION=$( jq --raw-output .version /home/lisk/lisk/config.json ) && \
-    LISK_MIN_VERSION=$( jq --raw-output .minVersion /home/lisk/lisk/config.json ) && \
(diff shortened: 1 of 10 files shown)


nazarhussain:
Had created relevant issues: 

https://github.com/LiskHQ/lisk-scripts/issues/106
https://github.com/LiskHQ/lisk-build/issues/298

PR #2197, opened 2018-07-04
SSL breaks Websocket connection - Closes #2154

### LINKED ISSUE #2154: SSL breaks Websocket connection
### Expected behavior
With SSL enabled incoming websocket connections should work.

### Actual behavior
With SSL enabled incoming Websocket connections fail.

Since no incoming Websocket connections are being established the broadhash consensus can only be established every 30 seconds (when `peer.rpc.status` is called) as no `updateMyself` broadcasts can reach the node. See #2083

The problems are the following:

SocketCluster enables the secure flag when https is enabled. Therefore incoming connections need to use `wss `as protocol. Currently the peer headers don't contain a flag signaling whether the node has secure websocket connections enabled which is why other peers can't know whether they have to connect via `ws` or `wss` => The connection fails with secure peers.

Peer connections are established using the IP address and not hostnames. Therefore the certificate validation fails when a secure WebSocket connection is being established. If cert checking should be disabled the `rejectUnauthorized` flag would have to be set. However this would essentially make the secure connection pointless.

### Steps to reproduce

Launch a Lisk Core node with HTTPS enabled in the config.

### Proposed solution

Disable HTTPs for WebSocket connections as certificate validation is not really possible without hostnames/domains and implement a better encryption mechanism post-1.0.0.

Completely remove the SSL options from Lisk Core. It just adds unnecessary complexity and is not as performant as using a reverse proxy like nginx to do SSL Termination. I'd rather recommend to add a guide on how to setup the Lisk core in conjunction with nginx/etc.

1.0.0+

jondubois:
### What was the problem?

Turning on SSL/TLS broke the P2P layer because peers expect unencrypted `ws://` protocol (not `wss://`).

### How did I fix it?

Turning on SSL in the config file (for the node) will only enable it for for the HTTP API; it will not affect/break the WebSockets P2P API.

Despite this fix, note that the recommended approach for SSL/TLS termination is still to setup a proxy like nginx or HAProxy in front of the Lisk node as it offers more flexibility. See discussion https://github.com/LiskHQ/lisk/issues/2154

### How to test it?

Enable `ssl` in config.json and specify the path to a SSL/TLS certificate and private key file as options.
In the browser, navigate to any API endpoint on your node using the https:// prefix; e.g. https://localhost:4001/api/peers (Note that the browser may show a warning if using a self-signed certificate, but you can continue and it should work). Check that the node can still connect to peers over `ws://` and can sync blocks, etc...

### Review checklist

* The PR solves #2154
* Linting passes
* Tests pass

### DIFF
--- a/app.js
+++ b/app.js
@@ -270,9 +270,9 @@ d.run(() => {
 					var https;
 					var https_io;
 
-					if (scope.config.ssl.enabled) {
-						privateKey = fs.readFileSync(scope.config.ssl.options.key);
-						certificate = fs.readFileSync(scope.config.ssl.options.cert);
+					if (scope.config.api.ssl.enabled) {
+						privateKey = fs.readFileSync(scope.config.api.ssl.options.key);
+						certificate = fs.readFileSync(scope.config.api.ssl.options.cert);
 
 						https = require('https').createServer(
 							{
@@ -723,15 +723,15 @@ d.run(() => {
 							);
 
 							if (!err) {
-								if (scope.config.ssl.enabled) {
+								if (scope.config.api.ssl.enabled) {
 									scope.network.https.listen(
-										scope.config.ssl.options.port,
-										scope.config.ssl.options.address,
+										scope.config.api.ssl.options.port,
+										scope.config.api.ssl.options.address,
 										err => {
 											scope.logger.info(
 												`Lisk https started: ${
-													scope.config.ssl.options.address
-												}:${scope.config.ssl.options.port}`
+													scope.config.api.ssl.options.address
+												}:${scope.config.api.ssl.options.port}`
 											);
 
 											cb(err, scope.network);
--- a/app.js
+++ b/app.js
@@ -17,7 +17,6 @@
 var path = require('path');
 var fs = require('fs');
 var d = require('domain').create();
-var extend = require('extend');
 var SocketCluster = require('socketcluster');
 var async = require('async');
 var Logger = require('./logger.js');
@@ -36,7 +35,6 @@ var swaggerHelper = require('./helpers/swagger');
  * @namespace app
  * @requires async
  * @requires domain.create
- * @requires extend
  * @requires fs
  * @requires socketcluster
  * @requires logger.js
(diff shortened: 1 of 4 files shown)


app.js
```diff
@@ -487,18 +485,6 @@ d.run(() => {
 						logLevel: 0,
 					};
 
-					if (scope.config.ssl.enabled) {
```
nazarhussain: We should think of it thoroughly and decide for some certain foreseen future. If we see that SSL option will never be part of P2P layer soon, or if it will, it will be seperate implementation. Then we should move `ssl` configuration key to under `api`. Then it will be clear that this `ssl` is only for API, not for p2p layer.

jondubois: This is definitely something that we should do in the future but one of the major goals of this PR is to make the migration from 0.9 to 1.0 as smooth as possible from the DevOps/SysAdmin point of view. Maintaining backwards compatibility with the old config file helps in that regard.

nazarhussain: This PR is created against `1.1.0`. Does it really have to do something with 0.9.x release? 

And there is no point where DevOps/SysAdmin involved here. Core itself have the configuration migration script, you can modify it as the way you want. That script always run during upgrade.

jondubois: @nazarhussain I'm referring to application migration in a broader sense, not database migration. When I say 'DevOps/SysAdmin', I'm referring to both independent people who run a Lisk node operators and also to people who operate Lisk nodes on behalf of large companies (e.g. exchanges).

MaciejBaj: @nazarhussain agree - SSL will not be the part of the P2P layer for now. The current place of SSL config doesn't really fit and should be moved to "api" section.

jondubois: @nazarhussain I misunderstood what you meant about config migration earlier. I didn't realise we had such a script.

jondubois:
@nazarhussain I made the changes. Could you please review?

MaciejBaj:
@jondubois please clean the scripts/update_config.js script and include the SSL change.

jondubois:
@nazarhussain I tested the migration manually.

nazarhussain:
@jondubois Thanks, you added changes for your PR, but that file needed to be updated to migrate from 1.0.x to 1.1.0. I will add a commit shortly with the required changes.

jondubois:
@nazarhussain I thought I did that here: https://github.com/LiskHQ/lisk/pull/2197/commits/8758e3fd7425cc4f35bbf7d7b583f0c211586c00 but feel free to add whatever is necessary as a commit on top.

nazarhussain:
@jondubois There is a lot more to do, as we don't need the secret migration any more from `1.0.x` to `1.1.x`. 

I had created seperate issue for it and approving this PR. We will handle it separately. 

https://github.com/LiskHQ/lisk/issues/2202

nazarhussain:
@jondubois Didn't checked, but some unit tests are failing in this PR consecutively. Can you please check. 

```
18:17:44 >>   at Immediate.vote.checkConfirmedDelegates.err (/home/lisk/workspace/lisk-core_PR-2197-J2Z5Z7EJG64F7N77KPFT6UAN2UM4CDM4T6FF5HVO6WYPSHQQ36RQ/test/unit/logic/vote.js:416:20)
18:17:44 >>   at runCallback (timers.js:674:20)
18:17:44 >>   at tryOnImmediate (timers.js:645:5)
18:17:44 >>   at processImmediate [as _immediateCallback] (timers.js:617:5)
18:17:44 Test failed: test/unit/logic/vote.js
```

Issue #2188, opened 2018-07-02
Check all initialization steps in case of errors and cleanup


nazarhussain:
### Expected behavior
If some error occurred during initialization, then it should trigger the cleanup and that cleanup goes smooth. 

### Actual behavior

Currently if we get this error during initialization (or any db error) 

```
6:32:55 error: database "lisk_dev" does not exist
         connection: {
.....
.....
.....
[inf] 2018-06-29 14:32:55 | Cleaning up...
```
Application cleanup crashes: 

```
Unhandled rejection TypeError: Cannot read property 'removeAllListeners' of undefined
    at process.once.error (/Users/nazar/Hub/LightCurve/Lisk/dev/lisk/app.js:771:24)
```

Because fo these lines: 

```
				scope.socketCluster.removeAllListeners('fail');
				scope.socketCluster.destroy();
```

### Steps to reproduce
1. Change the db name in config which actually does not exist 
2. Start the app with `node app.js`

### Which version(s) does this affect? (Environment, OS, etc...)
1.1.0

PR #2204, opened 2018-07-06
Check if socket cluster exists before calling remove

ManuGowda:
### What was the problem?
The process clean up was failing due to socket cluster was not initialized as the database initialization failed.
### How did I fix it?
Since database initialization happens first and then the socket cluster, before removing and destroying the socket connection, checking of the socket cluster was initialized.
### How to test it?
1. Change the DB name in config which actually does not exist 
2. Start the app with `node app.js`
### Review checklist

* The PR solves #2188 
* Linting passes
* Tests pass

### DIFF
--- a/app.js
+++ b/app.js
@@ -773,8 +773,10 @@ d.run(() => {
 					scope.logger.fatal(error.toString());
 				}
 				scope.logger.info('Cleaning up...');
-				scope.socketCluster.removeAllListeners('fail');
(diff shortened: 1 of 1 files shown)


nazarhussain:
@jondubois As described in the issue for this particular use case the error is pretty explanatory. 

```
error: database "lisk_dev" does not exist
```

PR #2215, opened 2018-07-10
Log level of "Inserted new peer" and "Removed peer" should be symmetric - Closes #1992

### LINKED ISSUE #1992: Log level of "Inserted new peer" and "Removed peer" should be symmetric
### Expected behavior

Log level of "Inserted new peer" and "Removed peer" should be symmetric

I think the "Removed peer" logs are too spammy for a default info configuration and should be log level debug.

### Actual behavior

* "Inserted new peer" logs are created for log level debug
* "Removed peer" logs are created for log level info

### Steps to reproduce

Watch the logs in the default configuration (`"fileLogLevel": "info"`). As you can see, there are always peers removed (but never added according to the logs).

0.9.14a

jondubois:
### What was the problem?

Logging related to new peers being added/removed was inconsistent depending on the log level.

### How did I fix it?

- Made 'Removed peer' message consistent (same logLevel, and using the same peer.string argument) as the 'Inserted new peer' message.
- Changed it so that the full peer object is only logged when logLevel is `trace` (since this takes a lot of space on the console).

### How to test it?

Run the node on logLevel info; the 'Removed peer' message should no longer come up.
When the logLevel is 'debug' then both the 'Removed peer' and 'Inserted new peer' messages should come up. When logLevel is 'trace' then the full peer object should be displayed when the peer is removed.

### Review checklist

* The PR solves #1992
* Linting passes
* Tests pass


shuse2:
PR needs to target `1.2` branch.

PR #2217, opened 2018-07-10
Fix config usage in worker_controller - Closes #2216

### LINKED ISSUE #2216: Worker controller needs to adopted to use the new configs
### Expected behavior
New Config style introduced by #398, should be applied to `worker_controller.js`

### Actual behavior
`worker_controller` still directly reads the config.

1.1.0

shuse2:
### What was the problem?
`worker_controller.js` was using the outdated way of getting the config.

### How did I fix it?
Use the latest way of getting the config as the `app.js`

### How to test it?
Observe if the child process is successfully started

### Review checklist

* The PR solves #2216
* Linting passes
* Tests pass

### DIFF
--- a/workers_controller.js
+++ b/workers_controller.js
@@ -29,7 +29,7 @@ var Rules = require('./api/ws/workers/rules');
 var failureCodes = require('./api/ws/rpc/failure_codes');
 var Logger = require('./logger');
 var AppConfig = require('./helpers/config.js');
-var config = AppConfig(require('./package.json'));
+var config = AppConfig(require('./package.json'), false);
 
 /**
  * Instantiate the SocketCluster SCWorker instance with custom logic
--- a/workers_controller.js
+++ b/workers_controller.js
@@ -14,8 +14,6 @@
 
 'use strict';
 
-var path = require('path');
-var fs = require('fs');
 var async = require('async');
 var SCWorker = require('socketcluster/scworker');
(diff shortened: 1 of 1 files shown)


workers_controller.js
```diff
@@ -30,14 +28,8 @@ var PeersUpdateRules = require('./api/ws/workers/peers_update_rules');
...
-
-var config = fs.readFileSync(
-	path.resolve(
-		process.cwd(),
-		`./config/${process.env.LISK_NETWORK}/config.json`
-	),
-	'utf8'
-);
+var AppConfig = require('./helpers/config.js');
+var config = AppConfig(require('./package.json'));
```
nazarhussain: Pass an extra param for `false` which will make AppConfig not to parse the command line params here.

nazarhussain: There are only three config values, that worker is dependent on configuration. 

```
config.consoleLogLevel
config.fileLogLevel
config.logFileName
```

If those can be handled by 

https://github.com/LiskHQ/lisk/blob/cb12a1e9d4d9d1bf65a7fc5d9427ee37e95d3084/app.js#L484-L491

that will leave no dependency  of worker over configuration file.

shuse2: I think change the config to send it to child process rather than reading it directly is not that easy since the childProcessOptions are sent async through IPC after starting the process.
In the current state, we need the config file at the initalization

PR #2219, opened 2018-07-11
Expire transactions doesn't call undoUnconfirmed - Closes #1731

### LINKED ISSUE #1731: Expire transactions doesn't call undoUnconfirmed
### Expected behavior
The `TransactionPool.prototype.expireTransactions` needs to be wrapped inside a `balancesSequence` and after each `expireTransactions` call, whatever the transactions were expired needs to be called for `undoUnconfirmed` to ensure the transactions is removed from the database.
### Actual behavior
In Transaction pool when we call `__private.expireTransactions`, it removes the unconfirmed transactions from the in-memory queue, but never calls `undoUnconfirmed`.
### Steps to reproduce
N/A

ManuGowda:
### What was the problem?
`undoUnconfirmed` function was not called while expiring the transactions and removed the transactions only from memory.
### How did I fix it?
whenever the expire transaction function is called, the `__private.expireTransactions` is wrapped inside balance sequence and invoked `modules.transactions.undoUnconfirmed` to ensure the unconfirmed transaction to be undone.
### How to test it?
Run functional system test.
### Review checklist

* The PR solves #1731 
* Linting passes
* Tests pass


test/functional/system/transactions/expire_transactions.js
```diff
@@ -0,0 +1,123 @@
...
+	let memAccountsSnapshotBeforeUndo;
+
+	const { expiryInterval } = __testContext.config.transactions;
+	const unconfirmedTransactionTimeOut =
+		global.constants.unconfirmedTransactionTimeOut;
+	// override transaction expiryInterval and unconfirmedTransactionTimeOut
+	// to test undo unConfirmed expired transactions
+	before(done => {
+		__testContext.config.transactions.expiryInterval = 5000;
+		global.constants.unconfirmedTransactionTimeOut = 0;
```
SargeKhan: How will application behave in case of setting the value of `global.constants.unconfirmedTransactionTimeOut` to 0? Will it never expire transactions?

test/functional/system/transactions/expire_transactions.js
```diff
@@ -0,0 +1,123 @@
...
+			recipientId: randomUtil.account().address,
+			amount: randomUtil.number(100000000, 1000000000),
+			passphrase: accountsFixtures.genesis.passphrase,
+		});
+
+		localCommon.addTransactionToUnconfirmedQueue(library, transaction, () => {
+			// Wait for transaction to get expired and
+			// undo the unconfirmed transaction, so that mem accounts
+			// balance is updated
+			setTimeout(done, 10000);
```
SargeKhan: Would be nice to have one more test for checking whether the transaction was applied to the unconfirmed state in the database or not. After that test, (Or as part of a before block for a nested describe block) we can use the timeout to expire the transaction and confirm that unconfirmed state is reverted.

logic/transaction_pool.js
```diff
@@ -992,29 +993,46 @@ __private.transactionTimeOut = function(transaction) {
...
+					if (seconds > timeOut) {
+						modules.transactions.undoUnconfirmed(
+							transaction,
+							undoUnconfirmErr => {
+								if (undoUnconfirmErr) {
+									library.logger.error(
+										`Failed to undo unconfirmed transaction: ${transaction.id}`,
+										undoUnconfirmErr
+									);
+									return setImmediate(cb);
```
SargeKhan: In case of an error on undoUnconfirm. We should throw an exception which restarts the application. Restarting the application allows us to revert changes applied to the unconfirmed state.

logic/transaction_pool.js
```diff
@@ -993,46 +998,78 @@ __private.transactionTimeOut = function(transaction) {
...
+			}
+
+			if (__private.isExpired(transaction)) {
+				modules.transactions.undoUnconfirmed(transaction, undoUnconfirmErr => {
+					if (undoUnconfirmErr) {
+						library.logger.error(
+							`Failed to undo unconfirmed transaction: ${transaction.id}`,
+							undoUnconfirmErr
 						);
+						return setImmediate(cb);
```
SargeKhan: We shouldnt call parent `Cb` from here.

logic/transaction_pool.js
```diff
@@ -993,46 +998,78 @@ __private.transactionTimeOut = function(transaction) {
...
+					// Remove transaction from unconfirmed, queued and multisignature lists
+					self.removeUnconfirmedTransaction(transaction.id);
+					library.logger.info(
+						`Expired transaction: ${
+							transaction.id
+						} received at: ${transaction.receivedAt.toUTCString()}`
+					);
+				});
+			}
+			return setImmediate(eachSeriesCb);
```
SargeKhan: Node will reach this line before finishing `undoUnconfirmed` async task.

logic/transaction_pool.js
```diff
@@ -993,46 +998,78 @@ __private.transactionTimeOut = function(transaction) {
...
+ *
+ * @private
+ * @param {Object} transactions - transaction to be expired
+ * @returns {Boolean} true if transactions expired
+ */
+__private.isExpired = transaction => {
+	const timeNow = Math.floor(Date.now() / 1000);
+	const timeOut = __private.transactionTimeOut(transaction);
+	// transaction.receivedAt is instance of Date
+	const seconds = timeNow - Math.floor(transaction.receivedAt.getTime() / 1000);
```
SargeKhan: Is it an instance of Date in every scenario?

test/functional/system/transactions/expire_transactions.js
```diff
@@ -0,0 +1,173 @@
...
+				queries
+					.getAccounts()
+					.then(memAccountsSnapshotAfterUndo => {
+						const beforeAccountState = memAccountsSnapshotBeforeUndo.find(
+							account => account.address === address
+						);
+						const afterAccountState = memAccountsSnapshotAfterUndo.find(
+							account => account.address === address
+						);
+						expect(
```
SargeKhan: same problem with expectations related to bignum

test/functional/system/transactions/expire_transactions.js
```diff
@@ -0,0 +1,173 @@
...
+
+	it('should expire transaction from unconfirmed transaction list', done => {
+		const transaction = liskElements.transaction.transfer({
+			recipientId: randomUtil.account().address,
+			amount: randomUtil.number(100000000, 1000000000),
+			passphrase: accountsFixtures.genesis.passphrase,
+		});
+
+		localCommon.addTransactionToUnconfirmedQueue(library, transaction, () => {
+			// Wait for transaction to get expired and
```
SargeKhan: no expectations in this test?

test/functional/system/transactions/expire_transactions.js
```diff
@@ -0,0 +1,173 @@
...
+		}
+
+		// update mem account snapshot for comparision
+		memAccountsSnapshotBeforeUndo.push(multiSigAccount);
+
+		localCommon.addTransaction(library, transaction, () => {
+			// Wait for transaction to get expired and
+			// undo the unconfirmed transaction, so that mem accounts
+			// balance is updated
+			setTimeout(done, 1000);
```
SargeKhan: no expectations in the test?

test/functional/system/transactions/expire_transactions.js
ManuGowda: fixed.

test/functional/system/transactions/expire_transactions.js
ManuGowda: `localCommon.addTransactionToUnconfirmedQueue` doesn't return anything so there are no expectations for that, however, `validateMemAccountSnapshot()` has expectations, let me know if there is anything else that I can try.

logic/transaction_pool.js
```diff
@@ -703,27 +713,32 @@ TransactionPool.prototype.undoUnconfirmedList = function(cb, tx) {
...
-					self.getMultisignatureTransactionList(true),
-					seriesCb
-				);
-			},
-		],
+	library.balancesSequence.add(
+		balancesSequenceCb => {
+			async.waterfall(
+				[
+					function(seriesCb) {
```
SargeKhan: waterFall and seriesCb.

config/default/config.json
```diff
@@ -77,7 +77,10 @@
 		"relayLimit": 3
 	},
 	"transactions": {
-		"maxTransactionsPerQueue": 1000
+		"maxTransactionsPerQueue": 1000,
+		"expiryInterval": 30000,
+		"multiSignatureTimeoutMultiplier": 3600,
+		"signaturesTimeoutMultiplier": 8
```
nazarhussain: We have `unconfirmedTransactionTimeOut` and `blockReceiptTimeOut` in constants. So we need revisit the though tif these new timeout values should go to `config` or `constants`. 

General rule can be, If these values can be configured differently on each node and still node works smoothly with the network, then its a config. If each node have to have same value in network to operate then its a constants. 

@MaciejBaj can confirm where to put these values in `constants` or `config`?

logic/transaction_pool.js
```diff
@@ -703,27 +713,32 @@ TransactionPool.prototype.undoUnconfirmedList = function(cb, tx) {
...
+						__private.expireTransactions(
+							self.getMultisignatureTransactionList(true),
+							waterfallCb
+						);
+					},
+				],
+				balancesSequenceCb
+			);
+		},
 		() => setImmediate(cb)
```
nazarhussain: No error handling here. Even if you think there could't be any error, but still handle the errors appropriately.

logic/transaction_pool.js
```diff
@@ -975,13 +990,34 @@ __private.applyUnconfirmedList = function(transactions, cb, tx) {
...
-		return transaction.asset.multisignature.lifetime * 3600;
+		return (
+			transaction.asset.multisignature.lifetime *
+			library.config.transactions.multiSignatureTimeoutMultiplier
+		);
 	} else if (Array.isArray(transaction.signatures)) {
-		return constants.unconfirmedTransactionTimeOut * 8;
+		return (
+			constants.unconfirmedTransactionTimeOut *
+			library.config.transactions.signaturesTimeoutMultiplier
```
nazarhussain: Again its confusing here, why 8 times of actual transaction timeout? Please add some comments to explain it.

logic/transaction_pool.js
```diff
@@ -975,13 +990,34 @@ __private.applyUnconfirmedList = function(transactions, cb, tx) {
...
+ * @private
+ * @param {Object} transactions - transaction to be expired
+ * @returns {Boolean} true if transactions expired
+ */
+__private.isExpired = transaction => {
+	const timeNow = Math.floor(Date.now() / 1000);
+	const timeOut = __private.transactionTimeOut(transaction);
+	// transaction.receivedAt is instance of Date
+	const seconds = timeNow - Math.floor(transaction.receivedAt.getTime() / 1000);
+	return seconds > timeOut;
```
nazarhussain: Please rename `seconds` to `time Passed`, that will make it more readable `return timePassed > timeOut`

logic/transaction_pool.js
```diff
@@ -1020,4 +1050,46 @@ __private.expireTransactions = function(transactions, cb) {
...
+					// Remove transaction from unconfirmed, queued and multisignature lists
+					self.removeUnconfirmedTransaction(transaction.id);
+					library.logger.info(
+						`Expired transaction: ${
+							transaction.id
+						} received at: ${transaction.receivedAt.toUTCString()}`
+					);
+					return setImmediate(eachSeriesCb);
+				});
+			} else {
```
nazarhussain: Remove this `else` part. its a good practice to not use `else` with single `if` block.

logic/transaction_pool.js
```diff
@@ -1020,4 +1050,46 @@ __private.expireTransactions = function(transactions, cb) {
...
+			}
+
+			if (__private.isExpired(transaction)) {
+				modules.transactions.undoUnconfirmed(transaction, undoUnconfirmErr => {
+					if (undoUnconfirmErr) {
+						library.logger.error(
+							`Failed to undo unconfirmed transaction: ${transaction.id}`,
+							undoUnconfirmErr
+						);
+						return setImmediate(eachSeriesCb);
```
nazarhussain: Why not sending the `undoUnconfirmErr `

logic/transaction_pool.js
```diff
@@ -1020,4 +1050,46 @@ __private.expireTransactions = function(transactions, cb) {
...
+							transaction.id
+						} received at: ${transaction.receivedAt.toUTCString()}`
+					);
+					return setImmediate(eachSeriesCb);
+				});
+			} else {
+				return setImmediate(eachSeriesCb);
+			}
+		},
+		() => setImmediate(cb)
```
nazarhussain: No error handling on this block.

ManuGowda: the error is handled in specific functions and logged, the reason for not propagation the error in case of `expireTransactions` is that `nextExpiry` is a job and the error doesn't lead to any action, so we just log the error and do not return the error anywhere, so let me know if we want to change to behaviour to something else.

logic/transaction_pool.js
ManuGowda: explained the same in the below comment, the error is not propagated as `nextExpiry` is a job we don't take any action.

logic/transaction_pool.js
ManuGowda: Honestly, I am not sure why 8 times, if anybody knows why I am happy to update the same in the comments.

config/default/config.json
```diff
@@ -77,7 +77,10 @@
 		"relayLimit": 3
 	},
 	"transactions": {
-		"maxTransactionsPerQueue": 1000
+		"maxTransactionsPerQueue": 1000,
+		"expiryInterval": 30000,
+		"multiSignatureTimeoutMultiplier": 3600,
```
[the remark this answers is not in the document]
ManuGowda: the expression is `transaction.asset.multisignature.lifetime * library.config.transactions.multiSignatureTimeoutMultiplier` so clearly the the multisignature transaction timeout is dervied from the lifetime of the multisignatire transaction with the multiplier `1hr` so thats the reason i chosen the name `multiSignatureTimeoutMultiplier`, please suggest me any other name which is approprate for the scenario.

logic/transaction_pool.js
```diff
@@ -975,13 +990,34 @@ __private.applyUnconfirmedList = function(transactions, cb, tx) {
  */
 __private.transactionTimeOut = function(transaction) {
 	if (transaction.type === transactionTypes.MULTI) {
-		return transaction.asset.multisignature.lifetime * 3600;
+		return (
+			transaction.asset.multisignature.lifetime *
+			library.config.transactions.multiSignatureTimeoutMultiplier
```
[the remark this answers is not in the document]
ManuGowda: same here, I do not know why the specific numbers of `3600` and `8`, so @MaciejBaj should know why these number?

config/default/config.json
ManuGowda: Agree, if `3600` and `8` are constant values which is specific, then we can move it to `constants`, @MaciejBaj can you please give me the feedback?

nazarhussain:
@MaciejBaj Please provide feedback here. Why these multipliers are `3600` and `8`, and if these are consistent for network then why not constants?

logic/transaction_pool.js
```diff
@@ -703,28 +709,33 @@ TransactionPool.prototype.undoUnconfirmedList = function(cb, tx) {
  * @returns {SetImmediate} error, ids[]
  */
 TransactionPool.prototype.expireTransactions = function(cb) {
```
MaciejBaj: Is it added to the jobs queue?

PR #2222, opened 2018-07-12
Change default scheme to https - Closes #2221

### LINKED ISSUE #2221: Use HTTPS as default scheme in swagger spec
### Expected behavior
When using interactive mode in 1.0 API documentation, the requests should be successful with default values.

### Actual behavior
When using interactive mode of swagger-ui, the requests fail.
This is because the default scheme is HTTP and has to be changed manually to HTTPS.

### Steps to reproduce
Go to https://docs.liskdev.net/documentation/lisk-core/user-guide/api/1-0
Hit` try-it-out` button, hit `execute`.

1.0.0

Tschakki:
### What was the problem?
When using interactive mode of swagger-ui, the requests fail.
This is because the default scheme is HTTP and has to be changed manually to HTTPS.

### How did I fix it?
Changed order of schemes in `swagger.yml`

### How to test it?
run `node app.js`
explore `http://localhost:4000/api/spec` with http://petstore.swagger.io/
Link to Lisk Core API documentation: https://lisk.io/documentation/lisk-core/user-guide/api/1-0

### Review checklist

* The PR solves #2221 
* Linting passes
* Tests pass

### DIFF
--- a/schema/swagger.yml
+++ b/schema/swagger.yml
@@ -91,8 +91,8 @@ tags:
   description: Votes related API calls
 
 schemes:
-  - http
   - https
+  - http
 
 paths:
   /accounts:


nazarhussain:
@Tschakki Here is the relevant issue on Swagger UI, that may be helpful for us to fix https://github.com/swagger-api/swagger-ui/issues/3704

Tschakki:
@nazarhussain The way I see it, to change it on swagger-ui side we need to change code somewhere here: https://raw.githubusercontent.com/swagger-api/swagger-ui/2.x/dist/swagger-ui.js
I think it will be much easier to just fix the schemes order in swagger.yml.

nazarhussain:
@Tschakki Changing the scheme order will break the usage of swagger in development. That's for me is more important because many people will be impacted by that change.

PR #2224, opened 2018-07-12
SSL breaks Websocket connection - Closes #2154

### LINKED ISSUE #2154: SSL breaks Websocket connection

nazarhussain:
### What was the problem?
Turning on SSL/TLS broke the P2P layer because peers expect unencrypted `ws://` protocol (not `wss://`).

### How did I fix it?
Turning on SSL in the config file (for the node) will only enable it for for the HTTP API; it will not affect/break the WebSockets P2P API.

Despite this fix, note that the recommended approach for SSL/TLS termination is still to setup a proxy like nginx or HAProxy in front of the Lisk node as it offers more flexibility. See discussion https://github.com/LiskHQ/lisk/issues/2154

### How to test it?
Enable `ssl` in config.json and specify the path to a SSL/TLS certificate and private key file as options.
In the browser, navigate to any API endpoint on your node using the https:// prefix; e.g. https://localhost:4001/api/peers (Note that the browser may show a warning if using a self-signed certificate, but you can continue and it should work). Check that the node can still connect to peers over `ws://` and can sync blocks, etc...

### Review checklist

* The PR solves #2154
* Linting passes
* Tests pass

### DIFF
--- a/scripts/update_config.js
+++ b/scripts/update_config.js
@@ -51,6 +51,9 @@ console.info('Starting configuration migration...');
 const oldConfig = JSON.parse(fs.readFileSync(oldConfigPath, 'utf8'));
 const newConfig = JSON.parse(fs.readFileSync(newConfigPath, 'utf8'));
 
+newConfig.api.ssl = extend(true, {}, oldConfig.ssl);
+delete oldConfig.ssl;
+
 // Values to keep from new config file
 delete oldConfig.version;
 delete oldConfig.minVersion;
--- a/app.js
+++ b/app.js
@@ -277,9 +277,9 @@ d.run(() => {
 					var https;
 					var https_io;
 
-					if (scope.config.ssl.enabled) {
-						privateKey = fs.readFileSync(scope.config.ssl.options.key);
-						certificate = fs.readFileSync(scope.config.ssl.options.cert);
+					if (scope.config.api.ssl.enabled) {
+						privateKey = fs.readFileSync(scope.config.api.ssl.options.key);
+						certificate = fs.readFileSync(scope.config.api.ssl.options.cert);
 
 						https = require('https').createServer(
 							{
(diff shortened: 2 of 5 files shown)


scripts/update_config.js
```diff
@@ -79,6 +79,9 @@ oldConfig.peers.list = oldConfig.peers.list.map(p => {
 	return p;
 });
 
+// Move SSL to API only
+oldConfig.api.ssl = extend(true, {}, oldConfig.ssl);
```
jondubois: Should this be `newConfig.api.ssl = ...`? Also, shouldn't `oldConfig.ssl` be deleted?
It looks a bit different than how it was implemented in https://github.com/LiskHQ/lisk/pull/2197/files#diff-02f64463a1deb2818037febeeb0fa51d

scripts/update_config.js
```diff
@@ -51,6 +51,9 @@ console.info('Starting configuration migration...');
 const oldConfig = JSON.parse(fs.readFileSync(oldConfigPath, 'utf8'));
 const newConfig = JSON.parse(fs.readFileSync(newConfigPath, 'utf8'));
 
+newConfig.api.ssl = extend(true, {}, oldConfig.ssl);
```
MaciejBaj: newConfig.api.ssl = extend(true, newConfig.api.ssl, oldConfig.ssl); ?

nazarhussain: There is no change in the structure of the `ssl` objet itself in previous and current release. So this code will also work fine. 
```
newConfig.api.ssl = extend(true, {}, oldConfig.ssl);
```

PR #2228, opened 2018-07-16
NetworkHeight in /api/node/status is wrong - Closes #2179

### LINKED ISSUE #2179: networkHeight in /api/node/status is wrong
### Expected behavior
networkHeight should give the *correct* height of the network, meaning the height with the highest consensus or in other words with the most nodes behind it. I use this to verify if my nodes are on the "correct" height.

### Actual behavior
currently networkHeight seems to be the highest height seen in the network, even if it is only a single forked node on this height, which is wrong in my opinion.

### Steps to reproduce
directly after the fork on testnet at block 5594490, networkHeight was 5594683 for a long time, because 2 peers were on this height. See network monitor:
wrong_networkheight

1.0.0-rc1 (testnet)

ManuGowda:
### What was the problem?
Currently, NetworkHeight is picking the highest height seen in the network.
### How did I fix it?
The NetworkHeight should be the height of the most popular height among all the peers, instead of taking the highest height in the network, taking the most popular height.
### How to test it?
In the network, whichever group of peers are in a majority, that should be the most popular height and the NetworkHeight should reflect the same.
### Review checklist

* The PR solves #2179 
* Linting passes
* Tests pass

### DIFF
--- a/modules/peers.js
+++ b/modules/peers.js
@@ -750,17 +750,17 @@ Peers.prototype.networkHeight = function(options, cb) {
 			return setImmediate(cb, err, 0);
 		}
 		// count by number of peers at one height
-		const heights = _.countBy(peers, 'height');
-		const height = Object.keys(heights).reduce((curr, next) => {
-			if (heights[next] > heights[curr]) {
-				return next;
-			}
-			return curr;
-		});
-		const networkHeight = Number(height);
+		const mostPopularHeight = _(peers)
+			.countBy('height')
+			.map((count, height) => ({
+				height,
+				count,
+			}))
+			.maxBy('count');
+		const networkHeight = Number(mostPopularHeight.height);
 
 		library.logger.debug(`Network height is: ${networkHeight}`);
-		library.logger.trace(heights);
+		library.logger.trace(mostPopularHeight);
 
 		return setImmediate(cb, null, networkHeight);
 	});
--- a/modules/peers.js
+++ b/modules/peers.js
@@ -736,17 +736,31 @@ Peers.prototype.list = function(options, cb) {
 	);
 };
 
+/**
+ * Gets the height of maximum number of peers at one particular height.
+ *
+ * @param {Object} options
+ * @param {string} [options.normalized=false] - Return peers in normalized (json) form
+ * @param {function} cb - Callback function
+ * @returns {setImmediateCallback} cb, err, peers
+ */
 Peers.prototype.networkHeight = function(options, cb) {
 	self.list(options, (err, peers) => {
-		if (err) {
+		if (err || peers.length === 0) {
 			return setImmediate(cb, err, 0);
 		}
-		const peersGroupedByHeight = _.groupBy(peers, 'height');
-		const popularHeights = Object.keys(peersGroupedByHeight).map(Number);
-		const networkHeight = _.max(popularHeights);
+		// count by number of peers at one height
+		const heights = _.countBy(peers, 'height');
+		const height = Object.keys(heights).reduce((curr, next) => {
+			if (next === undefined || heights[next] > heights[curr]) {
+				return next;
+			}
+			return curr;
+		});
+		const networkHeight = Number(height);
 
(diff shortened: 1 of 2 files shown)


test/unit/modules/peers.js
```diff
@@ -379,20 +379,49 @@ describe('peers', () => {
...
-				peersLogicMock.list = sinonSandbox.stub().returns(randomPeers);
 				done();
 			});
 
 			const randomPeerNetworkHeight = randomPeers => {
-				const peersGroupedByHeight = _.groupBy(randomPeers, 'height');
-				const popularHeights = Object.keys(peersGroupedByHeight).map(Number);
-				return _.max(popularHeights);
+				const heights = _.countBy(randomPeers, 'height');
+				const height = Object.keys(heights).reduce((curr, next) => {
```
MaciejBaj: It's not a good test.
It doesn't check the functionality, but rewrites and possibly repeats the mistakes introduced in code. 

Please get rid of `randomPeerNetwrokHeight` functionality. Change the generic test `should return the height of maximum number of peers at one particular height` to `should return 19 for peers being on heights: [19, 19, 19, 20, 20]`

modules/peers.js
```diff
@@ -736,17 +736,31 @@ Peers.prototype.list = function(options, cb) {
...
-		if (err) {
+		if (err || peers.length === 0) {
 			return setImmediate(cb, err, 0);
 		}
-		const peersGroupedByHeight = _.groupBy(peers, 'height');
-		const popularHeights = Object.keys(peersGroupedByHeight).map(Number);
-		const networkHeight = _.max(popularHeights);
+		// count by number of peers at one height
+		const heights = _.countBy(peers, 'height');
+		const height = Object.keys(heights).reduce((curr, next) => {
```
MaciejBaj: How about simply `const networkHeight = _(peers).countBy('height').maxBy('height').value()`?

modules/peers.js
```diff
@@ -736,17 +736,31 @@ Peers.prototype.list = function(options, cb) {
...
+			if (next === undefined || heights[next] > heights[curr]) {
```
SargeKhan: Why do we check if `next === undefined` here?

test/unit/modules/peers.js
```diff
@@ -379,20 +379,49 @@ describe('peers', () => {
...
+					expect(err).to.be.null;
+					expect(networkHeight);
+					expect(networkHeight)
+						.to.be.an('number')
+						.and.to.deep.eql(myself[0].height);
+					done();
+				});
+			});
+
+			it('should return the height of maximum number of peers at one particular height', done => {
```
SargeKhan: It's better to have expectations of real values here as opposed to random numbers. So, it's clearer to verify whether it works or not.

modules/peers.js
ManuGowda: if there is only one height, then the curr will have the value, but the next will be undefined.

modules/peers.js
ManuGowda: `countBy` returns `{5: 3, 7: 1, 10: 1}` where `5,7,10` are the heights and `3,1,1` are the corresponding counts, so I dont think this will work`const networkHeight = _(peers).countBy('height').maxBy('height').value()`

ManuGowda:
@MaciejBaj after the discussion with @SargeKhan it's evident that we need further discussion about what is `networkHeight` when the peer list is empty, do we need to consider all the peers in the network to get the majority, I think it will be a good thing to discuss first and decide before we can do further reviews on this. I am going to mark this issues `suspended` until we come back with the correct approach.

test/unit/modules/peers.js
MaciejBaj: Bump

modules/peers.js
MaciejBaj: You are right @ManuGowda. How about:
```
const mostPopularGroup = _(peers).countBy('height').map((counts, height) => {
return {
  height: +height,
  counts
};
}).maxBy('counts');

const networkHeight = mostPopularGroup.height
```

PR #2233, opened 2018-07-16
Add support for block versions - Closes #2199

### LINKED ISSUE #2199: Update block version to 1
While 0.9.16 and 1.0. versions are not compatible with each other (peers cannot connect) - there is still possible to keep the 0.9.16 chain running. Because there is nothing at stake - probably most of the delegates would like to keep forging on that chain, and because of such a possibility - it's reasonable to assume that 0.9.16 chain will be the longer one. There can be some issues if somebody decide to migrate 0.9.16 chain later (after migration height) and announce it to rest of the peers that are already running 1.0.0 chain. To ensure smooth migration and prevent situation described above from happening - would be nice to bump block's property `version` (which is taken into account when generating the signature).

4miners:
### What was the problem?
Need support for bumping `block.version` property at particular heights.
### How did I fix it?
Add support for bumping `block.version` property.
### How to test it?
Run test suite.
### Review checklist

* The PR solves #2199
* Linting passes
* Tests pass


helpers/constants.js
```diff
@@ -110,7 +110,10 @@ var constants = {
 		offset: 1451520, // Start rewards at block (n)
 		distance: 3000000, // Distance between each milestone
 	},
-	// WARNING: When changing totalAmount you also need to change getBlockRewards(int) SQL function!
+	// Block height at which block version increases
+	blockVersions: [
+		0,
+	],
```
nazarhussain: This is confusing to me. If I follow the comment above, it says: 

> at block height 0 the block version will be increased 

I believe better would be to manage an object with **starting block height** and associated **block version**

```
blockVersions: {
   1: 0,
   6521962: 1
},
```

This is just a thought, target is to make this constant more readable. May be some other way is also possible.

helpers/block_version.js
```diff
@@ -0,0 +1,44 @@
...
+		if (constants.blockVersions[i] <= height) {
+			version = i;
+		}
+	}
+	return version;
+}
+
+module.exports = {
+	get,
+};
```
nazarhussain: Would not it be better to move it to `modules.blocks.utils` or under `logic/block_version.js` as already have `logic/block_reward.js`

test/functional/system/block_version.js
```diff
@@ -0,0 +1,88 @@
...
+describe('block_version', () => {
+	let library;
+	let Queries;
+	let addTransactionsAndForgePromise;
+
+	constants.blockVersions = [
+		0,
+		102, // Bump block version at height 102
+		203, // Bump block version at height 203
+	];
```
nazarhussain: That seems more readable to me:

```
constants.blockVersions = {
   1: 0,
   102: 1,
   203: 2
};
```

test/unit/helpers/block_version.js
```diff
@@ -0,0 +1,80 @@
...
+		});
+
+		it('should return 2 for height > 203', () => {
+			const height = 204;
+			const version = 2;
+
+			return expect(blockVersion.get(height)).to.equal(version);
+		});
+	});
+});
```
nazarhussain: 👍

helpers/constants.js
4miners: Added comments and changed initial height from 0 to 1.

4miners:
@nazarhussain Please check again.

helpers/constants.js
```diff
@@ -110,7 +110,10 @@ var constants = {
 		offset: 1451520, // Start rewards at block (n)
 		distance: 3000000, // Distance between each milestone
 	},
-	// WARNING: When changing totalAmount you also need to change getBlockRewards(int) SQL function!
+	// Every version starts on provided height
+	blockVersions: [
```
MaciejBaj: In `exceptions.js` there is a special `precedent` field design to deal with legacy issues like block versioning https://github.com/LiskHQ/lisk/blob/development/config/default/exceptions.js#L51.
Please move the constant to exceptions.

4miners: 1. That structure is not present in `rc.2` version.
2. I don't agree that block versions should be moved to exceptions. While the thing we already have there (disabling particular transactions) is a precedent (happened once and will not happen again), block version is not. That block version will be used in every future hard-fork and will become integral part of the application. It's why I think it should stay in constants, like block rewards milestones for example.

helpers/constants.js
```diff
@@ -110,7 +110,10 @@ var constants = {
 		offset: 1451520, // Start rewards at block (n)
 		distance: 3000000, // Distance between each milestone
 	},
-	// WARNING: When changing totalAmount you also need to change getBlockRewards(int) SQL function!
```
diego-G: Why we remove this?

helpers/constants.js
diego-G: Is this replacing or affecting somehow the exceptions to freeze Dapp transfers?
Could we use it to remove those exceptions?

nazarhussain: The feature `blockVersion` is similar to `rewards.milestones` where we specify some values for milestones on particular block height. 

```
	rewards: {
		milestones: [
			500000000, // Initial Reward
			400000000, // Milestone 1
			300000000, // Milestone 2
			200000000, // Milestone 3
			100000000, // Milestone 4
		],
```
Similar is for `blockVersions`
```
blockVersions: [
     1, // Initial block version to 0
      1334445, Block Version 1
]
```

So this options seems to be perfect to reside in constants.

MaciejBaj: https://github.com/LiskHQ/lisk/pull/2233#discussion_r202700076

In constants, we store the Lisk network parameters. The newly created network wouldn't contain any hard forks, nor exceptions. The same behaviour. The network is parametrized by 'constants' which supposed to be... constant, not changed by _"become integral part of the application"_.

https://github.com/LiskHQ/lisk/pull/2233#discussion_r202720478
Changing rewards in milestones is the network feature that decreases the inflation over time. 

Keeping on track of all of the block versions (hard forks) is the not the network parameter, nor the feature. 
It is a legacy change needed to be introduced to allow the network to continue. Same as all of the other exceptions. I insist to move it to exceptions under the newly created "precedent" key, that will be merged with dapp in/out disable while merging with 1.1.0 version.

4miners:
@MaciejBaj Done.

4miners:
What actually `exceptions` file already contains and where we use that? We have there following properties:
- `blockRewards` - allow invalid block reward to pass block verification process (testnet)
- `delegates` - allow transaction that contains delegate username that is already present in database to pass transactions verification process (not used)
- `genesisPublicKey` - allow transaction that spend from genesis accounts only in genesis block (both)
- `rounds` - it's a modifier that affects rewards distribution in round (mainnet)
- `senderPublicKey` - allow transaction with invalid `senderPublicKey` (doesn't match - `publicKey` of actual account that spends) to pass transactions verification process (both)
- `signatures` - allow transaction with invalid signature to pass transactions verify process (both)
- `multisignatures` - allow multisignature registraction transaction which requires minimal signatures that is equal or greater its keygroup (both)
- `votes` - allow invalid vote transaction (conditions can vary) to pass transactions verification process (both)
- `inertTransactions` - change the behavior of the application, skip `apply`/`undo` to `mem_accounts` table for particular transaction (testnet)
- `transactionFee` - allow transaction with invalid fee to pass transactions verification process (not used on testnet, doesn't exists on mainnet)
`precedent.disableDappTransfer` - makes transaction verification process fail for `InTransfer` and `OutTransfer` transaction types if block height is greater or equal provided one (from 1.1.0 version)

Those exceptions were created to **legitimize the invalid state** of blockchain (as blockchain is an immutable ledger) which arose as a result of various conditions, such as bugs, conscious changes in the code (increasing security or removing features). That allows new nodes to synchronize from scratch with the network and also allows snapshots creation. These two things would not be possible if we wouldn't allow for incorrect blockchain state through exceptions.

Property `version` of the block object can be used for many different things, for example it can:
- Allow us to break the compatibility and cut off blocks from legacy chain (hard-fork) by accepting only blocks with specified version from particular height.
- Instructing the application how it should behave according to that property. Some features can be supported in version `1` and not in `0` while still be backward compatible.

Implemented structure `blockVersions` is however something different than early mentioned exceptions, it defines the behavior of the application in **future event** which will occur after certain block height is hit by the moving blockchain. I agree with @nazarhussain that it's more similar to `rewards.milestones` structure - we can still change blocks rewards that are not yet reached without making the blockchain state invalid, same with `blockVersions`.

Going forward, `blockVersions` structure contains heights at which block `version` property will be increased, it starts from `0` at height `1`.
```
	blockVersions: [
		1, // version: 0
	],
```
That is **valid state** of the blockchain (all current block have `version = 0`). When we add another height there and after it will be reached by the chain - all further blocks will have `version = 1` which will be still a **valid state** of the blockchain.

By adding that structure to exceptions we will **legitimize valid state** which in my opinion just doesn't make sense. Please consider your decision @MaciejBaj.

nazarhussain:
**We had thorough discussion about the topic and here are key points:** 

1. Block version is the key indicator to identify any changes in the block structure (now or in future) or handle the hard forks. 
2. Block version maps directly with the protocol version that software is implementing 
3. So clearly block version is a protocol specific attribute so should not be added to network specific configurations.

**Here are the decision factors:** 

1. We are bumping block version to 1 in Core version  1.0.0 
2. So only Core 1.0.0 and upcoming release can process blocks with version 1
3. That leads to fact that all previous blocks will considered invalid by 1.0.0 which have version 0

**To handle the historical track of valid blocks:** 

1. We will add the an exception to specific network which we migrate to 1.0.0 
2. That exception will tell software up to which height blocks are valid with version 0
3. That height will be exactly the same of the migration point 


For reference @MaciejBaj @4miners

willclarktech:
@nazarhussain @4miners @MaciejBaj Can we have a record of how @4miners 's concerns were answered.

> By adding that structure to exceptions we will legitimize valid state which in my opinion just doesn't make sense.

This point in particular seems powerful to me and I don't see an answer to it in Nazar's comment.

Also

> So clearly block version is a protocol specific attribute so should not be added to network specific configurations.

The height at which the protocol changes is specific to each network. It shouldn't be a matter of *configuration* obviously, but an alternative to treating it as an exception is to treat it as a network-specific *constant*.

helpers/constants.js
```diff
@@ -110,7 +110,6 @@ var constants = {
...
```
nazarhussain: Why removed this comment? Is is irrelevant now?

modules/blocks/verify.js
```diff
@@ -221,7 +222,10 @@ __private.verifyAgainstLastNBlockIds = function(block, result) {
  * @returns {Array} result.errors - Array of validation errors
  */
 __private.verifyVersion = function(block, result) {
-	if (block.version > 0) {
+	if (
+		block.version !== blockVersion.currentBlockVersion &&
+		!blockVersion.inExceptions(block.version, block.height)
```
nazarhussain: Can we move this complete logic to `logic/block_version.isValid(height, version)` that will be more useful for future.

test/functional/system/block_version.js
```diff
@@ -0,0 +1,166 @@
...
+	describe('when there are proper exceptions for blocks versions', () => {
+		it('snapshotting should succeed', done => {
+			// Set current block version to 3
+			blockVersion.currentBlockVersion = 3;
+
+			// Set proper exceptions for blocks versions
+			exceptions.blockVersions = {
+				0: { start: 1, end: 101 },
+				1: { start: 102, end: 202 },
+				2: { start: 203, end: 303 },
```
nazarhussain: I like your idea/concept that in future we can have block versions with overriding heights.

helpers/constants.js
4miners: Yep, that SQL function was removed.

4miners:
@nazarhussain Thanks for the review, great idea. Resolved.

test/unit/modules/blocks/verify.js
```diff
@@ -455,14 +456,140 @@ describe('blocks/verify', () => {
 	describe('__private.verifyVersion', () => {
 		let verifyVersion;
 
-		it('should return error when block version > 0', () => {
-			verifyVersion = __private.verifyVersion({ version: 3 }, { errors: [] });
-			return expect(verifyVersion.errors[0]).to.equal('Invalid block version');
+		describe('when there are no exceptions for block versions', () => {
+			describe('when block height provided', () => {
```
SargeKhan: Does this mean that we can have block versions like 1, 3, 5 without including 2,4?

4miners: Yes.

test/functional/system/block_version.js
SargeKhan: Is that a good thing though? We say block version is linked with software version. I think if different nodes with same `expeptions.blockVersions` are setup, then they should behave in the same way no matter which software version they are using.

modules/blocks/verify.js
```diff
@@ -221,7 +222,7 @@ __private.verifyAgainstLastNBlockIds = function(block, result) {
  * @returns {Array} result.errors - Array of validation errors
  */
 __private.verifyVersion = function(block, result) {
-	if (block.version > 0) {
+	if (!blockVersion.isValid(block.version, block.height)) {
```
MaciejBaj: What's the path for Genesis Blocks with version 0?

4miners: Genesis blocks are not going through `processBlock` (`modules/blocks/utils`) which calls `verifyBlock`, so the `version` is never checked for them.

logic/block_version.js
```diff
@@ -0,0 +1,69 @@
...
+ * @returns {boolean}
+ */
+function isValid(version, height) {
+	// Check if there is an exception for particular block version
+	const heightsRange = exceptions.blockVersions[version];
+
+	const isCurrentVersion = version === this.currentBlockVersion;
+
+	// If there is no exception - check against current block version
+	if (!heightsRange && isCurrentVersion) {
```
MaciejBaj: So to summarize the condition: 

If there are no heights range defined the block is always valid.

To add clarity I think the more suited sentence would be:

If there are no **past** heights ranges defined the block is always valid.

How about renaming the constant to `pastHeightsRanges`?

4miners: You can specify any height in range, even in past - there are not checks for that.

PR #2234, opened 2018-07-16
Make docker images for 1.1 configurable - Closes #2180

### LINKED ISSUE #2180: Fix docker container workflow in 1.1
### Expected behavior

Docker image can be used to run Lisk on devnet/alphanet/betanet/testnet/mainnet.

### Actual behavior

- template must be removed, will be done by the application in a backwards compatible way LiskHQ/lisk#2225
- a `-n` switch must now be passed to the application LiskHQ/lisk#2149

### Steps to reproduce

Start Lisk in Docker.

Docker with Lisk 1.1

ghost:
### What was the problem?

The configuration mechanism changed in 1.1; docker images were not configurable any more.

### How did I fix it?

Work around:
- make all default values in the configuration template `null`
- remove all `null` values from the templated `config.json` (contains only explicitely set values)
- pass this custom `config.json` to the application using the `-c` switch

A more elegant solution (that does not duplicate the configuration management logic) will be implement in 1.2.

### How to test it?

Use docker image setting environment variables (e.g. `LISK_CONSOLELOGLEVEL=debug`) and ensure they're working.

### Review checklist

* The PR solves #2180 
* Linting passes
* Tests pass

### DIFF
--- a/docker_files/etc/confd/templates/config.json.tmpl
+++ b/docker_files/etc/confd/templates/config.json.tmpl
@@ -1,120 +1,85 @@
 {
-	"wsPort": {{getv "/lisk/wsport" "8001"}},
-	"httpPort": {{getv "/lisk/httpport" "8000"}},
-	"address": "{{getv "/lisk/address" "0.0.0.0"}}",
-	"version": "__LISK_VERSION__REPLACE_ME__",
-	"minVersion": "__LISK_MIN_VERSION__REPLACE_ME__",
-	"fileLogLevel": "{{getv "/lisk/fileloglevel" "info"}}",
-	"logFileName": "{{getv "/lisk/logfilename" "logs/lisk.log"}}",
-	"consoleLogLevel": "{{getv "/lisk/consoleloglevel" "error"}}",
-	"trustProxy": {{getv "/lisk/trustproxy" "false"}},
-	"topAccounts": {{getv "/lisk/topaccounts" "false"}},
-	"cacheEnabled": {{getv "/lisk/cacheenabled" "false"}},
-	"wsWorkers": {{getv "/lisk/wsworkers" "1"}},
+	"wsPort": {{getv "/lisk/wsport" "null"}},
+	"httpPort": {{getv "/lisk/httpport" "null"}},
+	"address": "{{getv "/lisk/address" "null"}}",
+	"fileLogLevel": "{{getv "/lisk/fileloglevel" "null"}}",
+	"logFileName": "{{getv "/lisk/logfilename" "null"}}",
+	"consoleLogLevel": "{{getv "/lisk/consoleloglevel" "null"}}",
+	"trustProxy": {{getv "/lisk/trustproxy" "null"}},
+	"topAccounts": {{getv "/lisk/topaccounts" "null"}},
+	"cacheEnabled": {{getv "/lisk/cacheenabled" "null"}},
+	"wsWorkers": {{getv "/lisk/wsworkers" "null"}},
 	"db": {
-		"host": "{{getv "/lisk/db/host" "localhost"}}",
-		"port": {{getv "/lisk/db/port" "5432"}},
-		"database": "{{getv "/lisk/db/database" "lisk_main"}}",
-		"user": "{{getv "/lisk/db/user" "lisk"}}",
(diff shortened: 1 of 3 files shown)


docker_files/home/lisk/run.sh
```diff
@@ -1,4 +1,25 @@
...
+        with_entries(select( .value != null and .value != {} and .value != [] and .value != "null" ))
+    elif type == "array" then
+        map(select( . != null and . != {} and .!= [] ))
+    else
+        .
+    end
+)' config.json >config.json_
+mv config.json_ config.json
+
+node app.js -c /home/lisk/lisk/config.json $@
```
nazarhussain: I am not sure about how docker run the entry point. But if `$@` means that it will have `-n` option default to `main net` then its good to go.

nazarhussain (approved):

docker_files/home/lisk/run.sh
abelboldu: Yes, it's main net by default, as  defined here: `CMD ["-n", "mainnet"]`

docker_files/home/lisk/run.sh
```diff
@@ -1,4 +1,25 @@
 #!/bin/bash
 
 confd -backend env -onetime
-node app.js
+jq '
+def walk(f):
```
abelboldu: What is this function for? Reduce the template to the used keys only?
Is it required?

fchavant: @abelboldu correct, this is the implementation of:
> remove all null values from the templated config.json (contains only explicitely set values)
it is a necessary workaround to make the template and the new configuration system coexist in 1.1 and will disappear in 1.2

docker_files/home/lisk/run.sh
fchavant: @nazarhussain default is `-n mainnet` (see [our Dockerfile](https://github.com/LiskHQ/lisk/pull/2234/files#diff-3254677a7917c6c01f55212f86c57fbfR54) as well as the [Dockerfile reference](https://docs.docker.com/v17.09/engine/reference/builder/#cmd); this can be overriden at runtime.
When using `run.sh` bash will pass all paramters (`$@`) to the application and so `-n testnet` can be passed as parameter, for example. Note that there is no explicit default when using `run.sh`, something like the following `docker-compose.yml` snipped must be used:
```
[...]
entrypoint: ["/home/lisk/wait-for-it.sh", "db:5432", "--", "/home/lisk/run.sh"]
command: ["-n", "testnet"]
[...]
```

The examples in LiskHQ/lisk-docker will be updated accordingly.

docker_files/etc/confd/templates/config.json.tmpl
```diff
@@ -1,120 +1,85 @@
...
+				"port": {{getv "/lisk/ssl/options/port" "null"}},
+				"address": "{{getv "/lisk/ssl/options/address" "null"}}",
+				"key": "{{getv "/lisk/ssl/options/key" "null"}}",
+				"cert": "{{getv "/lisk/ssl/options/cert" "null"}}"
 			}
 		}
 	},
 	"peers": {
-		"enabled": {{getv "/lisk/peers/enabled" "true"}},
-		"list": [
```
diego-G: I would prefer to have the opportunity to choose my peers. I don't see the point to force Docker users to use Lisk Official peers.

docker_files/home/lisk/run.sh
diego-G: Please, bear in mind the idea is to have one single `docker-compose` file. I even foresee we can remove the `lisk-docker` repo and include in `lisk-core` that file. It will be easy to maintain and more standard. 

Perhaps we can add some docker env files to show some examples.

In the command line we should add a comment to highlight we pick the network in there.

diego-G (changes requested):

docker_files/home/lisk/run.sh
fchavant: @diego-G I agree, let's do that in a separate issue/PR though.

diego-G: Please, open the issue to not to forget it and create a reference to this PR.

diego-G (approved):

abelboldu (approved):

PR #2235, opened 2018-07-17
Update to Node.js version 8.11.3 - Closes #2135

### LINKED ISSUE #2135: Update to Node.js version 8.12.0
### Expected behavior

We should upgrade Node.js to version `v8.12.0(LTS)` and align the other products [lisk-build](https://github.com/LiskHQ/lisk-build) and [lisk-commander](https://github.com/LiskHQ/lisky).

The rationale behind upgrading to version `v8.12.0(LTS)` are:
- NPM 5 supports freezing the dependency with `package-lock.json`, this will allow us to lock down the dependencies versions.
- Node.js `6.x` is under `Maintenance LTS` and we want to move to Node.js `8.x` which the current ` Active LTS`

Following is summary of features comes with upgrading from `6.x` to `8.x`:
- NPM 5 (Introduction of `package-lock.json` and faster package install)
- Use of Async and Await functions
- Static Error Codes
- N-API (Experimental)
- Object REST AND SPREAD operator usage
- Experimental HTTP/2 (faster, simpler, and more robust) for more [info](https://developers.google.com/web/fundamentals/performance/http2/)
### Actual behavior

Currently, `lisk-core` supports node.js version `v6.14.1` and npm `3.10.10`.

1.0.0

ManuGowda:
### What was the problem?

- Upgrading node.js version from 6.14.3 to 8.11.3
- Hex string to Buffer backward compatibility
### How did I fix it?
- Upgraded node.js version from 6.14.3 to 8.11.3 to support `node >= 6.14.3 <=8.11.3`
- Added support for backward compatibility for hex string to buffer
### How to test it?
- Follow `README.md` documentation and upgrade node.js to a specific version
- Run all the tests (Unit, Functional, and Integration)
- Start the node
### Review checklist

* The PR solves #2135 
* Linting passes
* Tests pass


logic/block.js
```diff
@@ -16,6 +16,7 @@
 
 const crypto = require('crypto');
 const ByteBuffer = require('bytebuffer');
+const ed = require('../helpers/ed.js');
```
nazarhussain: `ed` is already part of the scope, don't need to require it here.

logic/block.js
```diff
@@ -199,10 +200,11 @@ class Block {
...
 				.digest();
-			const blockSignatureBuffer = Buffer.from(block.blockSignature, 'hex');
-			const generatorPublicKeyBuffer = Buffer.from(
-				block.generatorPublicKey,
-				'hex'
+			const blockSignatureBuffer = this.scope.ed.hexToBuffer(
+				block.blockSignature
+			);
+			const generatorPublicKeyBuffer = this.scope.ed.hexToBuffer(
+				block.generatorPublicKey
```
nazarhussain: So you are using `ed` from scope, not the one declared at top level. So better not to require it.

test/unit/helpers/ed.js
```diff
@@ -176,4 +176,35 @@ describe('ed', () => {
...
+			return expect(() => ed.hexToBuffer(hex)).to.throw(
+				'Argument must be a valid hex string.'
+			);
+		});
+
+		it('should return buffer for a valid hex string', () => {
+			const hex =
+				'c094ebee7ec0c50ebee32918655e089f6e1a604b83bcaa760293c61e0f18ab6f';
+			return expect(ed.hexToBuffer(hex)).to.be.an.instanceof(Buffer);
+		});
```
nazarhussain: Please add a test case with capital alphabets. I know we downcast the keys in upper layer, but this particular unit of code also support it as logically these are valid hex strings.

test/unit/logic/block.js
```diff
@@ -551,7 +551,8 @@ describe('block', () => {
 			const bytes1 = block.getBytes(blockData);
 			const blockDataCopy = Object.assign({}, blockData);
 			blockDataCopy.height = 100;
-			blockDataCopy.generatorPublicKey = 'this is some key';
```
nazarhussain: 🤒

helpers/ed.js
```diff
@@ -79,4 +79,26 @@ ed.verify = function(hash, signature, publicKey) {
...
+ * @param {String} hex
+ * @returns {Buffer}
+ * @todo Add description for the params
+ */
+ed.hexToBuffer = hex => {
+	if (typeof hex !== 'string') {
+		throw new TypeError('Argument must be a string.');
+	}
+	// Regex to match valid hex string with even length
+	const hexRegex = /^([0-9a-f]{2})+$/i;
```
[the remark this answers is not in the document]
ManuGowda: yes it is still valid, `/^([0-9a-f]{2})+$/i` this regex is ignoring case with flag `i`

helpers/ed.js
```diff
@@ -79,4 +79,26 @@ ed.verify = function(hash, signature, publicKey) {
...
+	const matchedHex = (hex.match(hexRegex) || [])[0];
```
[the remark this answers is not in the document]
ManuGowda: not necessary as the regex ignore the case

package.json
```diff
@@ -16,8 +16,8 @@
 		"url": "https://github.com/LiskHQ/lisk/issues"
 	},
 	"engines": {
-		"node": "=6.14.1",
-		"npm": "=3.10.10"
+		"node": ">=6.14.1 <=8.11.3",
```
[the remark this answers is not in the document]
ManuGowda: yes, for now, we are gonna continue supporting old version, that's the conformation I got from Maciej.

Issue #2227, opened 2018-07-15
Set default db username during configuration migration


nazarhussain:
### Expected behavior
There must be a db user name set in the configuration. If not then default username is `lisk` to be used everywhere. 

### Actual behavior
Currently during configuration migration the sometimes db.username is set to blank that cause trouble during the migration process.

### Steps to reproduce
Migrate an instance from `0.9.16` to `1.0.0`

### Which version(s) does this affect? (Environment, OS, etc...)
1.0.0-rc.2

PR #2240, opened 2018-07-19
In update_config.js, set "lisk" user for database if its empty - Closes #2178

### LINKED ISSUE #2178: Set "lisk" user for database if its empty during migration
### Expected behavior
There must be a username for database set at the end of migration process. If not set it tries to load `USER` env variable. Which is not available when lisk scripts are started from the some non-interactive shell. 

### Actual behavior
Version `0.9.x` had behavior to use `lisk` username if there is no username set in `config.json`. This behavior is changed in `1.0.x`. Not sure the which cause this behavioral change, `pg-promise` may be looked into it. 

### Steps to fix
Update the `update_config.js` script to set user to `lisk` if there is no user after the final configuration is generated. 

1.0.x

Nazgolze:
### What was the problem?
When migrating config.json, if the previous config has db.user set to '', that gets migrated when it should default to 'lisk'
### How did I fix it?
I added a final check at the end to see of modifiedConfig.db.user is set. If not, we set it to 'lisk'.
### How to test it?
I tested this with a default 0.9.16a config and a default release 1.0.0-rc.1 config with the scenarios listed below.
* Without this change:
  * if db.user is set to '' in the old config and 'lisk' in the new one, db.user in new config is set to ''
  * if db.user is set to 'arbitrary' in old config and 'lisk' in the new one, 'arbitary' is preserved
* With this change:
  * if db.user is set to '' in the old config and 'lisk' in the new one, db.user in new config is set to 'lisk'
  * if db.user is set to 'arbitrary' in old config and 'lisk' in the new one, 'arbitary' is preserved

### DIFF
--- a/scripts/update_config.js
+++ b/scripts/update_config.js
@@ -151,6 +151,9 @@ function migrateSecrets(password) {
 
 function copyTheConfigFile() {
 	const modifiedConfig = extend(true, {}, newConfig, oldConfig);
+	if (!modifiedConfig.db.user) {
+		modifiedConfig.db.user = 'lisk';
+	}
 
 	fs.writeFile(
 		newConfigPath,


Nazgolze:
Ideally, this would be for 1.0, 1.1 and 1.2

Nazgolze:
Already closed with #2227 / #2229

PR #2242, opened 2018-07-20
Expectations are missing assertions - Closes #2236

### LINKED ISSUE #2236: Expectations are missing assertions
### Expected behaviour
In tests, expectations should always consist of assertions that can be answered with `true` or `false`.

### Actual behavior
In some tests, the assertion is missing. This is because of the use of `.equals()`, which resolves directly to a boolean.

### Steps to reproduce
See in tests, e.g.: https://github.com/LiskHQ/lisk/blob/1.2.0/test/unit/logic/transfer.js#L146

1.2.0

ManuGowda:
### What was the problem?
A missing expectation for the amount, fee and reward fields
### How did I fix it?
Added expectations
### How to test it?
Run tests
### Review checklist

* The PR solves #2236 
* Linting passes
* Tests pass

### DIFF
--- a/test/functional/system/blocks/verify/verify.js
+++ b/test/functional/system/blocks/verify/verify.js
@@ -915,9 +915,9 @@ describe('blocks/verify', () => {
 					expect(block1.version).to.equal(0);
 					expect(block1.timestamp).to.equal(time);
 					expect(block1.numberOfTransactions).to.equal(0);
-					expect(block1.reward.equals('0'));
-					expect(block1.totalFee.equals('0'));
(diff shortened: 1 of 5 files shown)


ManuGowda:
@SargeKhan, to be honest, I don't see any difference, `.to.be.true === .true` and `.to.equal(true) === .equal(true)`. it is just a syntactical sugar difference and my expectation in this test is an x amount should be equal to y amount and that should be true. if you still think using `.to.equal(true)`make it absolutely correct from using `.to.be.true` then sure I will make the change.

PR #2247, opened 2018-07-25
Return asset data for type 6 transactions - Closes #2194

### LINKED ISSUE #2194: No asset data is returned for transactions type 6 (inTransfer)
### Expected behavior
I expected to see `asset` object data when requesting `/api/transactions?type=6`

### Actual behavior
Empty object is returned

### Steps to reproduce
`curl testnet.lisk.io:7000/api/transactions?type=6`

Testnet

Tschakki:
### What was the problem?
No asset data was returned when querying type 6 transactions (in-transfer).

### How did I fix it?
The SQL query was not case-sensitive, added quotes around `in_dappId` to make the query successful.
Without quotes, it would always search for lower case.

### How to test it?
`npm run mocha -- test/unit/db/repos/transactions/index.js`

### Review checklist

* The PR solves #2194 
* Linting passes
* Tests pass

### DIFF
--- a/test/fixtures/accounts.js
+++ b/test/fixtures/accounts.js
@@ -25,7 +25,7 @@ accounts.existingDelegate = {
 	publicKey: 'addb0e15a44b0fdc6ff291be28d8c98f5551d0cd9218d749e30ddb87c6e31ca9',
 	passphrase:
 		'actress route auction pudding shiver crater forum liquid blouse imitate seven front',
-	balance: '0',
+	balance: '1000000000',
 	delegateName: 'genesis_100',
 };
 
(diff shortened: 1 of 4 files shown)


test/fixtures/accounts.js
```diff
@@ -25,7 +25,7 @@ accounts.existingDelegate = {
 	publicKey: 'addb0e15a44b0fdc6ff291be28d8c98f5551d0cd9218d749e30ddb87c6e31ca9',
 	passphrase:
 		'actress route auction pudding shiver crater forum liquid blouse imitate seven front',
-	balance: '0',
+	balance: '1000000000',
```
diego-G: Is this intended?

Tschakki: this doesn't seem necessary, removed it!

test/functional/http/get/transactions.js
```diff
@@ -865,5 +913,46 @@ describe('GET /api/transactions', () => {
 				});
 			});
 		});
+		/* eslint-disable mocha/no-skipped-tests */
```
MaciejBaj: Can we add a comment clarifying why this exact test is skipped and when will be unskipped?

Tschakki: good idea!
I added the comment now

PR #2251, opened 2018-07-25
Validate chain on startup of the application - Closes #2249

### LINKED ISSUE #2249: Validate chain on startup of the application
### Expected behavior
As we introduced block version in `RC2`, so any block forged with `RC1` above from the expected height should be removed on the startup of the application. 

### Actual behavior
Currently if any node is migrated to latest version with some delay and it contains few blocks in database which have previous version still contains in database and can't sync. 

### Steps to reproduce
- Set block version in `logic/block_version` to `0`
- Start the app forge 10  blocks stop the app
- Increase block version in `logic/block_version` to `1`
- Add exception for `blockVersions` in `helpers/exceptions` `{0: { start: 0, end: 5}}`
- Now start the app again 
- The last 5 blocks should be removed as these are invalid 
- But these will persist in the database 

1.0.0-rc.2

nazarhussain:
### What was the problem?
Currently if any node is migrated to latest version with some delay and it contains few blocks in database which have previous version still contains in database and can't sync.

### How did I fix it?
Check the block validation on startup and delete the blocks. 

### How to test it?
Run `functional/system/loader/validate_own_chain.js` 

### Review checklist

* The PR solves #2249
* Linting passes
* Tests pass

### DIFF
--- a/helpers/slots.js
+++ b/helpers/slots.js
@@ -161,4 +161,24 @@ module.exports = {
 	calcRound(height) {
 		return Math.ceil(height / this.delegates);
 	},
+
+	/**
+	 * Calculate starting height of the round
+	 *
+	 * @param round
+	 * @return {number}
+	 */
+	calcRoundStartHeight(round) {
+		return (round < 1 ? 0 : round - 1) * this.delegates + 1;
+	},
+
+	/**
+	 * Calculating end height of the round
+	 *
+	 * @param round
+	 * @return {number}
+	 */
+	calcRoundEndHeight(round) {
+		return (round < 1 ? 1 : round) * this.delegates;
+	},
 };
--- a/modules/blocks/utils.js
+++ b/modules/blocks/utils.js
@@ -256,6 +256,19 @@ Utils.prototype.getIdSequence = function(height, cb) {
 		});
 };
 
+Utils.prototype.loadBlockByHeight = function(height, cb, tx) {
+	(tx || library.db).blocks
+		.loadBlocksOffset(height, height + 1)
+		.then(rows => {
+			const blocks = self.readDbRows(rows);
+			return setImmediate(cb, null, blocks[0]);
+		})
+		.catch(err => {
+			library.logger.error(err.stack);
+			return setImmediate(cb, 'Blocks#loadBlockByHeight error');
+		});
+};
+
 /**
  * Generates a list of full blocks for another node upon sync request from that node, see: modules.transport.internal.blocks.
  *
--- a/modules/blocks/utils.js
+++ b/modules/blocks/utils.js
@@ -256,6 +256,16 @@ Utils.prototype.getIdSequence = function(height, cb) {
 		});
 };
 
+/**
+ * Load full block with a particular height
+ *
+ * @param {number} height - Block height
+ * @param {function} cb - Callback function
+ * @param {object} tx - Database transaction object
+ * @returns {function} cb - Callback function from params (through setImmediate)
+ * @returns {Object} cb.err - Error if occurred
+ * @returns {Object} cb.block - Block with requested height
+ */
 Utils.prototype.loadBlockByHeight = function(height, cb, tx) {
 	(tx || library.db).blocks
 		.loadBlocksOffset(height, height + 1)
--- a/modules/loader.js
+++ b/modules/loader.js
@@ -542,20 +542,18 @@ __private.loadBlockChain = function() {
 							return reload(blocksCount, 'No delegates found');
 						}
 
-						__private.validateOwnChain(validateOwnChainError => {
-							if (validateOwnChainError) {
-								throw validateOwnChainError;
+						modules.blocks.utils.loadLastBlock((err, block) => {
+							if (err) {
+								return reload(blocksCount, err || 'Failed to load last block');
 							}
 
-							modules.blocks.utils.loadLastBlock((err, block) => {
-								if (err) {
-									return reload(
-										blocksCount,
-										err || 'Failed to load last block'
-									);
+							__private.lastBlock = block;
+
+							__private.validateOwnChain(validateOwnChainError => {
+								if (validateOwnChainError) {
+									throw validateOwnChainError;
 								}
 
-								__private.lastBlock = block;
 								library.logger.info('Blockchain ready');
 								library.bus.message('blockchainReady');
 							});
@@ -570,6 +568,38 @@ __private.loadBlockChain = function() {
 };
 
 /**
+ * Validate given block
+ *
+ * @param {object} block
+ * @param {function} cb
+ * @returns {setImmediateCallback} cb, err
+ */
+__private.validateBlock = (blockToVerify, cb) => {
+	const lastBlock = modules.blocks.lastBlock.get();
+
+	modules.blocks.utils.loadBlockByHeight(
+		blockToVerify.height - 1,
+		(lastBlockError, secondLastBlockToVerify) => {
+			if (lastBlockError) {
+				return setImmediate(cb, lastBlockError);
+			}
+
+			// Set the block temporarily for block verification
+			modules.blocks.lastBlock.set(secondLastBlockToVerify);
+			const result = modules.blocks.verify.verifyBlock(blockToVerify);
+
+			// Revert last block changes
+			modules.blocks.lastBlock.set(lastBlock);
+
+			if (result.verified) {
+				return setImmediate(cb, null);
+			}
+			return setImmediate(cb, result.errors);
+		}
+	);
+};
+
+/**
  * Validate own block chain before startup
  *
  * @private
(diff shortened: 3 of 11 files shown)


helpers/slots.js
```diff
@@ -161,4 +161,24 @@ module.exports = {
...
 	},
+
+	/**
+	 * Calculate starting height of the round
+	 *
+	 * @param round
+	 * @return {number}
+	 */
+	calcRoundStartHeight(round) {
+		return (round < 1 ? 0 : round - 1) * this.delegates + 1;
```
4miners: Why such a condition? Round cannot be `0`, same for height.

helpers/slots.js
```diff
@@ -161,4 +161,24 @@ module.exports = {
...
+	},
+
+	/**
+	 * Calculating end height of the round
+	 *
+	 * @param round
+	 * @return {number}
+	 */
+	calcRoundEndHeight(round) {
+		return (round < 1 ? 1 : round) * this.delegates;
```
4miners: Same here.

modules/blocks/utils.js
```diff
@@ -256,6 +256,19 @@ Utils.prototype.getIdSequence = function(height, cb) {
 		});
 };
 
+Utils.prototype.loadBlockByHeight = function(height, cb, tx) {
```
4miners: Would be nice to add description for that function.

modules/loader.js
```diff
@@ -559,6 +567,131 @@ __private.loadBlockChain = function() {
...
+ * @returns {setImmediateCallback} cb, err
+ */
+__private.validateOwnChain = cb => {
+	const currentBlock = modules.blocks.lastBlock.get();
+	const currentHeight = currentBlock.height;
+	const currentRound = slots.calcRound(currentHeight);
+
+	// Starting height should be end height of second last round
+	let startHeight;
+	if (currentRound === 1) {
```
4miners: Sorry, I don't understand that logic here, can you clarify?
If current round is `1` - then start from height `2`.
If current round is `2` - then we pass `0` as `round`, but `round` cannot be `0`.

modules/loader.js
```diff
@@ -559,6 +567,131 @@ __private.loadBlockChain = function() {
...
+							return setImmediate(seriesCb, null);
+						});
+					}
+				);
+			},
+
+			// Delete all invalid blocks in that range
+			seriesCb => {
+				async.doDuring(
+					doWhilstCb => {
```
4miners: doDuringCb

helpers/slots.js
nazarhussain: That's why its protecting that part, if someone passed that round as 0 or negative.

helpers/slots.js
nazarhussain: Same here.

modules/loader.js
```diff
@@ -559,6 +567,131 @@ __private.loadBlockChain = function() {
...
+ * @param {object} block
+ * @param {function} cb
+ * @returns {setImmediateCallback} cb, err
+ */
+__private.validateBlock = (blockToVerify, cb) => {
+	const lastBlock = modules.blocks.lastBlock.get();
+
+	modules.blocks.utils.loadBlockByHeight(
+		blockToVerify.height - 1,
+		(lastBlockError, secondLastBlockToVerify) => {
```
SargeKhan: `lastBlockError` and `secondLastBlockToVerify`? Shouldn't it be `secondLastBlockError`?

modules/loader.js
```diff
@@ -559,6 +567,131 @@ __private.loadBlockChain = function() {
...
+			if (lastBlockError) {
+				return setImmediate(cb, lastBlockError);
+			}
+
+			// Set the block temporarily for block verification
+			modules.blocks.lastBlock.set(secondLastBlockToVerify);
```
SargeKhan: Does `verifyBlock` not work if last block is not set here?

helpers/slots.js
4miners: Ok, it's fine.

modules/loader.js
4miners: Yep, it will fail.

modules/loader.js
```diff
@@ -559,6 +567,131 @@ __private.loadBlockChain = function() {
...
+					`Validating current block with height ${currentHeight.height}`
+				);
+
+				__private.validateBlock(currentBlock, validateBlockErr => {
+					if (validateBlockErr) {
+						// Current block is invalid move to next step in async.series
+						return setImmediate(seriesCb, null);
+					}
+					// Current block is valid don't need to do any thing exit async.series
+					return setImmediate(seriesCb, true);
```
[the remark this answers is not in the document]
SargeKhan: And passing `error` value as true is not a very elegant way to exit from `async.series`.

modules/loader.js
```diff
@@ -559,6 +567,131 @@ __private.loadBlockChain = function() {
...
+				modules.blocks.utils.loadBlockByHeight(
+					startHeight,
+					(lastBlockError, startBlock) => {
+						__private.validateBlock(startBlock, validateBlockErr => {
+							if (validateBlockErr) {
+								library.logger.error(
+									`There are more than ${currentHeight -
+										startHeight} invalid blocks. Can't delete those to recover the chain.`
+								);
+								return process.emit(
```
SargeKhan: Do we need to emit cleanup event here? Can't we pass error back to `loadBlockChain` function and from there decide whether node wants to exit the application or throw the error even further

modules/loader.js
```diff
@@ -559,6 +567,131 @@ __private.loadBlockChain = function() {
...
+			// Delete all invalid blocks in that range
+			seriesCb => {
+				async.doDuring(
+					doWhilstCb => {
+						modules.blocks.chain.deleteLastBlock(doWhilstCb);
+					},
+					(deleteLastBlockStatus, testCallBb) => {
+						__private.validateBlock(
+							modules.blocks.lastBlock.get(),
+							validateError => setImmediate(testCallBb, null, !!validateError)
```
SargeKhan: testCallBb?

SargeKhan: And why not `setImmediate(testCallBb, validateError)`?

modules/loader.js
```diff
@@ -559,6 +567,131 @@ __private.loadBlockChain = function() {
...
+						);
+					},
+					doDuringErr => setImmediate(seriesCb, doDuringErr)
+				);
+			},
+		],
+		() => {
```
SargeKhan: No error handling in case some block fails verification? Or the `true` value we are passing as error in first series function?

test/functional/system/loader/validate_own_chain_first_round.js
```diff
@@ -0,0 +1,120 @@
...
+			);
+		}
+	);
+
+	describe('forge 101 blocks with version = 0', () => {
+		before(() => {
+			// Set current block version to 0
+			blockVersion.currentBlockVersion = 0;
+
+			// 9 blocks with 1 genesis block
```
SargeKhan: outdated comment?

test/functional/system/loader/validate_own_chain_first_round.js
```diff
@@ -0,0 +1,120 @@
...
+								expect(row.version).to.be.equal(0);
+							}
+						});
+					});
+				});
+
+				it('last 5 blocks should have version = 1', () => {
+					return Queries.getAllBlocks().then(rows => {
+						_.each(rows, row => {
+							if (row.height > 5) {
```
SargeKhan: nicer if rows.filter({height} => height > 5).forEach({version} => expect(version).to.be.equal(1)))

test/functional/system/loader/validate_own_chain_more_than_two_rounds.js
```diff
@@ -0,0 +1,98 @@
...
+
+				__private.loadBlockChain();
+			});
+
+			it('system should stop with error', () => {
+				expect(library.logger.error).to.be.calledWith(
+					"There are more than 202 invalid blocks. Can't delete those to recover the chain."
+				);
+				return expect(process.emit).to.be.calledWithExactly(
+					'cleanup',
```
SargeKhan: I am not sure if this test will fail on jenkins if there was an error. Because the process will exit before mocha compiles the failed and passed tests. So, the result will not be reported. I think we need to stub process.exit.

modules/loader.js
SargeKhan: That is bad. It should be fixed.

nazarhussain: `verifyBlock` function is developed with logic to verify only blocks when last block is one step behind that block. Yes it should be updated but out of scope for this PR.

modules/loader.js
nazarhussain: As per `async.doDuring` docs, test function need to send truthy as object param not the error param.

modules/loader.js
nazarhussain: We reach to this callback by explicitly on need. Only way to do is to send an error in previous steps. So we explicitly ignored the errors here.

modules/loader.js
```diff
@@ -546,9 +546,17 @@ __private.loadBlockChain = function() {
...
 								return reload(blocksCount, err || 'Failed to load last block');
 							}
+
 							__private.lastBlock = block;
-							library.logger.info('Blockchain ready');
-							library.bus.message('blockchainReady');
+
+							__private.validateOwnChain(validateOwnChainError => {
+								if (validateOwnChainError) {
+									throw validateOwnChainError;
```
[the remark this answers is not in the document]
nazarhussain: Yes it exit.

nazarhussain:
@SargeKhan Async.series was not working in the scenario when we reload the chain which is valid.

nazarhussain:
@4miners @SargeKhan Do you guys think of any other scenario that we should cover in functional tests for this change?

PR #2253, opened 2018-07-25
Ban peers when validation of their blocks fails during sync - Closes #2252

### LINKED ISSUE #2252: Ban peers when validation of their blocks fails during sync
### Expected behavior
The need to ban peers had been revealed to be necessary to be made after introducing the block versioning as a part of `1.0.0-rc.2`. 

Peers should be banned temporarily if they are sending malformed blocks.

### Actual behavior
Peers are not banned after sending invalid blocks (also with a wrong version). When they are staying in the majority, they can cause the poor consensus while being picked as good peers to sync with.
### Steps to reproduce
Synchronize invalid blocks from malicious peers. 

1.0.0-rc.2

MaciejBaj:
### What was the problem?
Peers were not banned during the sync process.
### How did I fix it?
Peers are banned temporarily if validation of their blocks fails in the sync process.
### How to test it?
Sync with peers distributing invalid peers. Check if it was banned.
### Review checklist

* The PR solves #2252
* Linting passes
* Tests pass

### DIFF
--- a/app.js
+++ b/app.js
@@ -604,7 +604,7 @@ d.run(() => {
 							peers: [
 								'logger',
 								function(scope, cb) {
-									new Peers(scope.logger, cb);
+									new Peers(scope.logger, scope.config, cb);
 								},
 							],
 						},
--- a/app.js
+++ b/app.js
@@ -431,7 +431,9 @@ d.run(() => {
 				db
 					.connect(config.db, dbLogger)
 					.then(db => cb(null, db))
-					.catch(cb);
+					.catch(err => {
+						console.error(err);
+					});
 			},
 
 			/**
@@ -540,6 +542,9 @@ d.run(() => {
 						{
 							bus(cb) {
 								cb(null, scope.bus);
+							},
+							config(cb) {
+								cb(null, scope.config);
 							},
 							db(cb) {
 								cb(null, scope.db);
--- a/app.js
+++ b/app.js
@@ -603,6 +603,7 @@ d.run(() => {
 							],
 							peers: [
 								'logger',
+								'config',
 								function(scope, cb) {
 									new Peers(scope.logger, scope.config, cb);
 								},
--- a/helpers/ban_manager.js
+++ b/helpers/ban_manager.js
@@ -47,7 +47,7 @@ BanManager.BAN_TIMEOUT = 2000;
  * Hardcoded ban length - 2 minutes also needs justification.
  */
 BanManager.prototype.banTemporarily = function(peer, onBanFinished) {
-	// Don't ban temporarily (ban forever) a peer hardcoded in config.json by a user.
+	// Peer hardcoded in config.json by a user should stays always banned.
 	if (this.config.peers.access.blackList.includes(peer.ip)) {
 		return;
 	}
--- a/helpers/ban_manager.js
+++ b/helpers/ban_manager.js
@@ -29,12 +29,6 @@ function BanManager(logger, config) {
 }
 
 /**
- * Ban every peer for 2 minutes.
- * @type {number}
- */
-BanManager.BAN_TIMEOUT = 2000;
-
-/**
  * Temporarily bans a peer for 2 minutes.
  * Calls externally provided unbanPeer.
  *
@@ -52,6 +46,7 @@ BanManager.prototype.banTemporarily = function(peer, onBanFinished) {
 		return;
 	}
 
+	const BAN_TIMEOUT = 120000; // Ban a peer for 2 min.
 	const alreadyBannedPeer = this.bannedPeers[peer.string];
 	if (alreadyBannedPeer) {
 		clearTimeout(alreadyBannedPeer.banTimeoutId);
@@ -62,7 +57,7 @@ BanManager.prototype.banTemporarily = function(peer, onBanFinished) {
 		banTimeoutId: setTimeout(() => {
 			onBanFinished(peer);
 			delete this.bannedPeers[peer.string];
-		}),
+		}, BAN_TIMEOUT),
 	};
 };
 
--- a/helpers/ban_manager.js
+++ b/helpers/ban_manager.js
@@ -1,0 +1,71 @@
+/*
+ * Copyright © 2018 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+'use strict';
+
+/**
+ * Temporarily peers banning mechanism.
+ *
+ * @class
+ * @memberof helpers
+ * @param {Logger} logger
+ * @param {Object} config - app configuration
+ * @param {function} unbanPeer - app configuration
+ */
+function BanManager(logger, config, unbanPeer) {
+	this.bannedPeers = {};
+	this.config = config;
+	this.logger = logger;
+	this.unbanPeer = unbanPeer;
+}
+
+/**
+ * Ban every peer for 2 minutes.
+ * @type {number}
+ */
+BanManager.BAN_TIMEOUT = 2000;
+
+/**
+ * Temporarily bans a peer for 2 minutes.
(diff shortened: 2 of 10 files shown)


helpers/ban_manager.js
```diff
@@ -0,0 +1,71 @@
...
+'use strict';
+
+/**
+ * Temporarily peers banning mechanism.
+ *
+ * @class
+ * @memberof helpers
+ * @param {Logger} logger
+ * @param {Object} config - app configuration
+ * @param {function} unbanPeer - app configuration
```
4miners: Description is not accurate.

helpers/ban_manager.js
```diff
@@ -0,0 +1,71 @@
...
+ */
+function BanManager(logger, config, unbanPeer) {
```
4miners: Why we pass a function here? That seems a little weird to me, can we go with different approach?

helpers/ban_manager.js
```diff
@@ -0,0 +1,71 @@
...
+ * @param {Peer} peer
+ *
+ * ToDo: Introduce well-designed and reasonable peer banning mechanism.
+ *
+ * For now no penalties for banning a peer multiple times are given.
+ * Hardcoded ban length - 2 minutes also needs justification.
+ */
+BanManager.prototype.ban = function(peer) {
+
+	// Don't ban temporarily (ban forever) a peer hardcoded in config.json by a user.
```
4miners: I don't understand that comment, can you clarify?

helpers/ban_manager.js
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	const alreadyBannedPeer = this.bannedPeers[peer.ip];
+	if (alreadyBannedPeer) {
+		clearTimeout(alreadyBannedPeer.banTimeoutId);
+	}
+
+	this.bannedPeers[peer.ip] = {
+		peer: peer,
+		banTimeoutId: setTimeout(() => {
```
4miners: I don't like that. Maybe consider going with different approach instead, like have one loop that iterates every X second and check if ban should be released.

logic/peers.js
```diff
@@ -108,6 +110,42 @@ Peers.prototype.get = function(peer) {
...
+ *
+ */
+Peers.prototype.ban = function(peer) {
+	peer.state = Peer.STATE.BANNED;
+	// Banning peer can only fail with ON_MASTER.UPDATE.INVALID_PEER error.
+	// Happens when we cannot obtain the proper address of a given peer.
+	// In such a case peer will be removed.
+	if (self.upsert(peer) === false) {
+		self.remove(peer);
+		library.logger.info('Attempt to ban a peer failed and resulted with removal');
```
4miners: That should be logged before `self.remove(peer)`, because that function is logging also.

4miners: Consider to log something like:
```
	library.logger.debug('Failed to ban peer', {
		err: 'INVALID_PEER',
		peer: peer.object(),
	});
```

logic/peers.js
```diff
@@ -108,6 +110,42 @@ Peers.prototype.get = function(peer) {
...
+ * @param {Peer} peer
+ */
+Peers.prototype.unban = function(peer) {
+	peer.state = Peer.STATE.DISCONNECTED;
+	// Banning peer can only fail with ON_MASTER.UPDATE.INVALID_PEER error.
+	// Happens when we cannot obtain the proper address of a given peer.
+	// In such a case peer will be removed.
+	if (self.upsert(peer) === false) {
+		self.remove(peer);
+		library.logger.info('Attempt to unban a peer failed and resulted with removal');
```
4miners: Same as above.

logic/peers.js
```diff
@@ -108,6 +110,42 @@ Peers.prototype.get = function(peer) {
...
+	}
+	library.logger.info(`Peer ${peer.string} unbanned successfully and moved to disconnected`);
```
4miners: Same as above.

logic/peers.js
```diff
@@ -108,6 +110,42 @@ Peers.prototype.get = function(peer) {
...
+	peer.state = Peer.STATE.BANNED;
+	// Banning peer can only fail with ON_MASTER.UPDATE.INVALID_PEER error.
+	// Happens when we cannot obtain the proper address of a given peer.
+	// In such a case peer will be removed.
+	if (self.upsert(peer) === false) {
+		self.remove(peer);
+		library.logger.info('Attempt to ban a peer failed and resulted with removal');
+	}
+	self.banManager.ban(peer);
+	library.logger.info(`Peer ${peer.string} banned successfully`);
```
4miners: Maybe worth to be consistent with remove:
```
library.logger.info('Banned peer', peer.string);
library.logger.debug('Banned peer', { peer: peer.object() });
```

app.js
```diff
@@ -604,7 +604,7 @@ d.run(() => {
 							peers: [
 								'logger',
```
4miners: We should add `config` here.

MaciejBaj: Ofc, thanks

helpers/ban_manager.js
MaciejBaj: BanManger bans temporarily. Adding a peer in config.json to blackList bans a peer always.

helpers/ban_manager.js
MaciejBaj: I am not finding any flaws in the current approach. I prefer to leave it like this, instead of iterating over all banned timeouts and introducing new interval constant of checks.

modules/loader.js
```diff
@@ -676,6 +676,9 @@ __private.loadBlocksFromNetwork = function(cb) {
 								library.logger.error(
 									`Failed to load blocks from: ${peer.string}`
 								);
+								// Ban a peer if block validation fails
+								// Invalid peers won't get chosen in the next sync attempt
+								library.logic.peers.ban(peer);
```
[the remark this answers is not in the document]
MaciejBaj: Good point, when there is a network error peer should be removed before:
```
	function getFromPeer(seriesCb) {
		peer.rpc.blocks(
			{ lastBlockId: lastValidBlock.id, peer: library.logic.peers.me() },
			(err, res) => {
				err = err || res.error;
				if (err) {
					modules.peers.remove(peer);
					return setImmediate(seriesCb, err);
				}
				return setImmediate(seriesCb, null, res.blocks);
			}
		);
	}
```
Then banning won't be possible. But the check should be more strict.

logic/peers.js
```diff
@@ -267,7 +327,8 @@ Peers.prototype.listRandomConnected = function(options) {
 		.map(key => self.peersManager.peers[key])
 		.filter(peer => peer.state === Peer.STATE.CONNECTED);
 	const shuffledPeerList = _.shuffle(peerList);
-	return options.limit ? shuffledPeerList.slice(0, options.limit)
+	return options.limit
+		? shuffledPeerList.slice(0, options.limit)
```
SargeKhan: options.limit is set to a list of peers?

MaciejBaj: It's prittier's change, not mine. But yes, you can pass a limit to peers.list if that's your question.

modules/blocks/process.js
```diff
@@ -475,7 +475,14 @@ Process.prototype.loadBlocksFromPeer = function(peer, cb) {
 					return setImmediate(eachSeriesCb);
 				}
 				// ...then process block
-				return processBlock(block, eachSeriesCb);
+				return processBlock(block, err => {
+					// Ban a peer if block validation fails
+					// Invalid peers won't get chosen in the next sync attempt
+					if (err) {
+						library.logic.peers.ban(peer);
```
SargeKhan: Is there a reason why we remove peer on here: https://github.com/LiskHQ/lisk/pull/2253/files#diff-60e1e043f6f3fdde809b81641313508bR447 but only ban peer here?

test/unit/helpers/ban_manager.js
```diff
@@ -0,0 +1,153 @@
...
+					it('should not call onBanFinished', () => {
+						return expect(onBanFinishedSpy).not.to.be.called;
+					});
+				});
+				describe('when 2 min have passed', () => {
+					beforeEach(done => {
+						clock.tick(twoMinTimeout);
+						return done();
+					});
+					it('should call onBanFinished', () => {
```
4miners: Redundant, next one already checking that.

MaciejBaj:
@4miners we probably can have an integration test sharing the invalid blocks to the other peers, but it's much simpler to cover in QA. I will add a scenario.

PR #2260, opened 2018-07-27
Improve process exit - Closes #1788

### LINKED ISSUE #1788: Improve process exit
### Expected behavior

When the application encounters a critical error that requires the process shut down, it must invoke an internal code block that will process the error and invoke `process.exit(value)` in the end with correct exit code.

This will let the hosting process know which re-spawning strategy needs to be executed:

* Immediate process restart
* Delayed process re-start
* Restart with a back-up configuration
* Machine restart
* Etc,...combinations

### Actual behavior

Use of `process.exit` is all over the place, and the hosting process does not know what is going on, and what should be done when the process dies.

### Steps to reproduce

Any step that will hit one of those `process.exit` calls, of which there are plenty.

All.

jondubois:
### What was the problem?

The 'cleanup' event caused the master process to always exit with code 1.
Sometimes the 'cleanup' event is sometimes triggered intentionally; in those cases, we want to shut down the process with code 0.

### How did I fix it?

If an error is passed to the 'cleanup' event handler, then the exit code will default to 1. Otherwise it will default to 0. Alternatively, a custom error code can be provided explicitly.

This PR narrows down the scope of #1788 to keep things simple - Now we just have 2 distinct error codes. If we end up wanting to add more exit codes in the future (and want to handle them in special ways), then we can raise separate issues for those.

### How to test it?

Test manually.
If you shut down the Lisk node from the command line (I.e. `Ctrl + C`), the exit code (e.g. `$?`) should be 0 (intentional exit). If you trigger a 'cleanup' event and pass an Error as second argument, then the exit code should be 1. If you trigger a 'cleanup' event and pass a custom integer as third argument, then the exit code of the process should be that integer.

### Review checklist

* The PR solves #1788
* Linting passes
* Tests pass

### DIFF
--- a/app.js
+++ b/app.js
@@ -798,7 +798,7 @@ d.run(() => {
 
 			if (err) {
 				logger.fatal(err);
-				process.emit('cleanup');
+				process.emit('cleanup', err);
 			} else {
 				scope.logger.info('Modules ready and launched');
 			}
--- a/app.js
+++ b/app.js
@@ -806,8 +806,12 @@ d.run(() => {
 	);
 });
 
+// TODO: This should be the only place in the master process where
+// 'uncaughtException' is handled. Right now, one of our dependencies (js-nacl;
+// which is a dependency of lisk-elements) adds its own handler which interferes
+// with our own process exit logic.
 process.on('uncaughtException', err => {
 	// Handle error safely
 	logger.fatal('System error', { message: err.message, stack: err.stack });
-	process.emit('cleanup');
+	process.emit('cleanup', err);
 });
--- a/app.js
+++ b/app.js
@@ -181,7 +181,7 @@ try {
 // Domain error handler
 d.on('error', err => {
 	logger.fatal('Domain master', { message: err.message, stack: err.stack });
-	process.exit(0);
+	process.emit('cleanup', err);
 });
 
 logger.info(`Starting lisk with "${appConfig.network}" genesis block.`);
--- a/app.js
+++ b/app.js
@@ -750,9 +750,14 @@ d.run(() => {
 		},
 		(err, scope) => {
 			// Receives a 'cleanup' signal and cleans all modules
-			process.once('cleanup', error => {
+			process.once('cleanup', (error, code) => {
 				if (error) {
 					logger.fatal(error.toString());
+					if (code === undefined) {
+						code = 1;
+					}
+				} else if (code === undefined || code === null) {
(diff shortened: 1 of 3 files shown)


jondubois:
@SargeKhan there are still some dependencies which call `process.exit()` directly. I'll see if I can change this behaviour easily; maybe throw an error instead and let the main error handler deal with the process exit.

jondubois:
There is one issue remaining related to process exit when dealing with an `uncaughtException` in the master process. One of our dependencies (`js-nacl`) adds its own `uncaughtException` handler to our master process which essentially re-throws the error (which instantly crashes the process). This causes our master process to exit before we could finish our cleanup operation (which is asynchronous).

jondubois:
The remaining issue caused by `js-nacl` (described in the previous comment) can be dealt with by upgrading `lisk-elements` to the latest version since it no longer uses `js-nacl`.

I've opened a separate issue for it here: https://github.com/LiskHQ/lisk/issues/2266

app.js
```diff
@@ -750,9 +750,14 @@ d.run(() => {
...
 		(err, scope) => {
 			// Receives a 'cleanup' signal and cleans all modules
-			process.once('cleanup', error => {
+			process.once('cleanup', (error, code) => {
 				if (error) {
 					logger.fatal(error.toString());
+					if (code === undefined) {
+						code = 1;
+					}
+				} else if (code === undefined || code === null) {
```
MaciejBaj: Do we exit from the application intentionally only after finished `snapshot` process? We should add a comment here.

jondubois: OK, I added a comment.

nazarhussain:
@jondubois There was an edge case you were diagnosing, that `domain error` handler was not catching the exceptions of `js-nacl`, is it resolved?

jondubois:
@nazarhussain The problem doesn't happen in RC2. I couldn't see any difference in the code which would cause the difference in behaviour. In any case this PR is for 1.2 so we should upgrade lisk-element and that will also take care of the issue.

nazarhussain:
The purpose of this PR was to do successful cleanup with proper error code while exiting. So it seems the purpose is not achieved unless we upgrade the lisk-element, so we must create an issue and reference to this PR, before merging it.

MaciejBaj:
Lisk Elements is already merged in version 1.0.0 - https://github.com/LiskHQ/lisk/pull/2284

PR #2265, opened 2018-07-30
Add sorting to block block loading - Relates #2249

nazarhussain:
### What was the problem?
There were no sorting while loading block for verifying 

### How did I fix it?
Add the sorting logic which were used in the load last block. 

### Review checklist

* The PR relates #2249
* Linting passes
* Tests pass

### DIFF
--- a/modules/blocks/utils.js
+++ b/modules/blocks/utils.js
@@ -270,8 +270,25 @@ Utils.prototype.loadBlockByHeight = function(height, cb, tx) {
 	(tx || library.db).blocks
 		.loadBlocksOffset(height, height + 1)
 		.then(rows => {
-			const blocks = self.readDbRows(rows);
-			return setImmediate(cb, null, blocks[0]);
+			// Normalize block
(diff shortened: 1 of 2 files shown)


modules/blocks/utils.js
```diff
@@ -270,8 +270,25 @@ Utils.prototype.loadBlockByHeight = function(height, cb, tx) {
 	(tx || library.db).blocks
 		.loadBlocksOffset(height, height + 1)
 		.then(rows => {
-			const blocks = self.readDbRows(rows);
-			return setImmediate(cb, null, blocks[0]);
+			// Normalize block
+			const block = modules.blocks.utils.readDbRows(rows)[0];
```
4miners: why not `self`?

nazarhussain: I just copied the same code from other method, where it was used.

MaciejBaj:
As agreed with @nazarhussain and @4miners the PR will be closed for now in favour of extending the block to verify logs - #2267.

PR #2275, opened 2018-08-01
Duplicated signatures in transactions - Closes #2254

### LINKED ISSUE #2254: Multisign issue with repeated signatures in tx
I broadcasted 840 tx this evening on testnet, to cosign them later. I expected all nodes receive same data

I noted that on the server where I cosigned the tx, reading unsigned tx (/api/node/transactions/unsigned?offset=0&limit=100) it has regularly 2 signatures (I cosigned them with 2 priv keys on a total of 3)
result is this:
`{"amount":"10","recipientId":"15797001169110345751L","senderPublicKey":"15b50402822a1ecc3d8f17ccae407858078a21ca940cfa919ec42ed50c4e612f","timestamp":68438429,"type":0,"fee":"10000000","asset":{},"signature":"050673ba60e30ee803a5b2d776afb78e6d0a22ce63c6cae1e4f23019600776e0fa15bc1ebe57d3f5f232d6fa83261abd73ecc51b3013f4f657dca288da42370d","id":"11993133628867109456","signatures":["7ddca64be39468c82c9e89a30c3756ff35eafce8b62c1ff1a6fb593e17c71bf985bfba73c88375eec15be23108466ac78a59602cc40bd09d1ccd725c514fb801","b688dadb9bb9e2b3ab1c50c5ea0e3fd5cc4f8906f1cfd407dea844c43364658aca59f95b4b04bad5a6f51c67420763b99ce371bd75bc86863cf010d6f188500c"],"senderId":"17803179600206889337L","relays":1,"receivedAt":"2018-07-25T19:40:29.202Z","ready":true,"recipientPublicKey":""}`

Instead on a secondary node, who receives the tx directly broadcasted from the network, I noted the same tx, but with a lot of repeated signatures like this:

`{"amount":"10","recipientId":"15797001169110345751L","senderPublicKey":"15b50402822a1ecc3d8f17ccae407858078a21ca940cfa919ec42ed50c4e612f","timestamp":68438429,"type":0,"fee":"10000000","asset":{},"signature":"050673ba60e30ee803a5b2d776afb78e6d0a22ce63c6cae1e4f23019600776e0fa15bc1ebe57d3f5f232d6fa83261abd73ecc51b3013f4f657dca288da42370d","id":"11993133628867109456","signatures":["b688dadb9bb9e2b3ab1c50c5ea0e3fd5cc4f8906f1cfd407dea844c43364658aca59f95b4b04bad5a6f51c67420763b99ce371bd75bc86863cf010d6f188500c","b688dadb9bb9e2b3ab1c50c5ea0e3fd5cc4f8906f1cfd407dea844c43364658aca59f95b4b04bad5a6f51c67420763b99ce371bd75bc86863cf010d6f188500c","b688dadb9bb9e2b3ab1c50c5ea0e3fd5cc4f8906f1cfd407dea844c43364658aca59f95b4b04bad5a6f51c67420763b99ce371bd75bc86863cf010d6f188500c","b688dadb9bb9e2b3ab1c50c5ea0e3fd5cc4f8906f1cfd407dea844c43364658aca59f95b4b04bad5a6f51c67420763b99ce371bd75bc86863cf010d6f188500c","b688dadb9bb9e2b3ab1c50c5ea0e3fd5cc4f8906f1cfd407dea844c43364658aca59f95b4b04bad5a6f51c67420763b99ce371bd75bc86863cf010d6f188500c","b688dadb9bb9e2b3ab1c50c5ea0e3fd5cc4f8906f1cfd407dea844c43364658aca59f95b4b04bad5a6f51c67420763b99ce371bd75bc86863cf010d6f188500c","b688dadb9bb9e2b3ab1c50c5ea0e3fd5cc4f8906f1cfd407dea844c43364658aca59f95b4b04bad5a6f51c67420763b99ce371bd75bc86863cf010d6f188500c","b688dadb9bb9e2b3ab1c50c5ea0e3fd5cc4f8906f1cfd407dea844c43364658aca59f95b4b04bad5a6f51c67420763b99ce371bd75bc86863cf010d6f188500c","b688dadb9bb9e2b3ab1c50c5ea0e3fd5cc4f8906f1cfd407dea844c43364658aca59f95b4b04bad5a6f51c67420763b99ce371bd75bc86863cf010d6f188500c","b688dadb9bb9e2b3ab1c50c5ea0e3fd5cc4f8906f1cfd407dea844c43364658aca59f95b4b04bad5a6f51c67420763b99ce371bd75bc86863cf010d6f188500c"],"senderId":"17803179600206889337L","relays":3,"receivedAt":"2018-07-25T19:40:39.572Z","ready":true,"recipientPublicKey":""}`

I initially broadcasted 840 tx on testnet, and then I cosigned them using 2 accounts

Issue affects testnet core 1.0.0-rc.1

4miners:
### What was the problem?
Duplicated signatures in transactions (send from multisignature, multisignature registration). More details in https://github.com/LiskHQ/lisk/issues/2254#issuecomment-408406183
### How did I fix it?
Refactored `processSignature` function in `multisignatures` module, execute all the logic through `balanceSequence`
### How to test it?
Run test suite.
### Review checklist

* The PR solves #2254
* Linting passes
* Tests pass


4miners:
@nazarhussain, @SargeKhan Please review. I'll write unit tests for new functions after resolving all feedback from the review.

modules/multisignatures.js
```diff
@@ -80,158 +80,221 @@ function Multisignatures(cb, scope) {
...
-				return setImmediate(cb, 'Failed to verify signature');
-			}
+		// If there are no signatures yet - initialise with empty array
+		transaction.signatures = transaction.signatures || [];
 
-			if (!verify) {
-				return setImmediate(cb, 'Failed to verify signature');
-			}
+		// Check if received signature already exists in transaction
+		if (transaction.signatures.indexOf(signature.signature) !== -1) {
```
ManuGowda: use of includes here as well

test/functional/system/transactions/4.multisignature/duplicate_signatures.js
```diff
@@ -0,0 +1,391 @@
...
+				passphrase: accounts.multisignature.passphrase,
+				keysgroup: [
+					accounts.multisignatureMembers[0].publicKey,
+					accounts.multisignatureMembers[1].publicKey,
+				],
+				lifetime: 4,
+				minimum: 2,
+			});
+
+			// Create signatures (strings)
```
ManuGowda: object instedof string

modules/multisignatures.js
```diff
@@ -80,158 +80,221 @@ function Multisignatures(cb, scope) {
...
-		library.balancesSequence.add(cb => {
-			const multisignatureTransaction = modules.transactions.getMultisignatureTransaction(
-				transaction.transactionId
-			);
-
-			if (!multisignatureTransaction) {
-				return setImmediate(cb, 'Transaction not found');
+__private.isValidSignature = (signature, members, transaction) => {
+	let isValid = false;
+
```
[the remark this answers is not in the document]
4miners: I cannot do that because I log the entire `signature` object later.

modules/multisignatures.js
```diff
@@ -80,158 +80,221 @@ function Multisignatures(cb, scope) {
...
+		// If there are no signatures yet - initialise with empty array
+		transaction.signatures = transaction.signatures || [];
 
-			if (!verify) {
-				return setImmediate(cb, 'Failed to verify signature');
-			}
+		// Check if received signature already exists in transaction
+		if (transaction.signatures.includes(signature.signature)) {
+			const message = 'Unable to process signature, signature already exists';
+			library.logger.error(message, { signature, transaction });
```
SargeKhan: Is signature and transactions logged here as an actual object or 'Object [Object]'

modules/multisignatures.js
```diff
@@ -80,158 +80,221 @@ function Multisignatures(cb, scope) {
...
+	try {
+		// If publicKey is provided we can perform direct verify
+		if (signature.publicKey) {
+			// Check if publicKey is present as member of multisignature account in transaction
+			if (!members.includes(signature.publicKey)) {
+				library.logger.error(
+					'Unable to process signature, signer not in keysgroup.',
+					{ signature, members, transaction }
+				);
+				return isValid;
```
nazarhussain: This is confusing to return `isValid` here, always have to look up what value was assigned to it. It would be more clear just return `false` from this check. The only return statement in the end is more appropriate.

modules/multisignatures.js
```diff
@@ -80,158 +80,221 @@ function Multisignatures(cb, scope) {
...
-
-	function done(cb) {
-		library.balancesSequence.add(cb => {
-			const multisignatureTransaction = modules.transactions.getMultisignatureTransaction(
-				transaction.transactionId
-			);
-
-			if (!multisignatureTransaction) {
-				return setImmediate(cb, 'Transaction not found');
+__private.isValidSignature = (signature, members, transaction) => {
```
nazarhussain: Suggest to rename `members` to `memberPublicKeys`

modules/multisignatures.js
```diff
@@ -80,158 +80,221 @@ function Multisignatures(cb, scope) {
...
+			);
+		} else {
+			// publicKey is not there - we need to iterate through all members of multisignature account in transaction
+			isValid = members.find(member =>
+				// Try to verify the signature
+				library.logic.transaction.verifySignature(
+					transaction,
+					member,
+					signature.signature
+				)
```
nazarhussain: This code block will assign a `publicKey` to `isValid`, while its expected to be boolean value. If later on use someone did a strict match with `true/false` then code will break.

modules/multisignatures.js
4miners: Not sure, will check it, thanks.

4miners:
@SargeKhan Thanks for feedback regarding the tests, I will refactor them. What do you mean by dividing them? Split to several files?

modules/multisignatures.js
```diff
@@ -80,158 +80,221 @@ function Multisignatures(cb, scope) {
...
+ */
+__private.processSignatureForMultisignatureAccountCreation = (
+	signature,
+	transaction,
+	cb
+) => {
+	// Normalize members of multisignature account from transaction
+	const members = transaction.asset.multisignature.keysgroup.map(member =>
+		member.substring(1) // Remove first character, which is '+'
+	);
```
nazarhussain: Why this normalization is not done on the right place 

https://github.com/LiskHQ/lisk/blob/4b1c95af9345fac6f56d031c9b2bf9a9b6d8525f/logic/transaction.js#L1099-L1102

modules/multisignatures.js
```diff
@@ -80,158 +80,221 @@ function Multisignatures(cb, scope) {
...
+
+	// Check if signature is valid
+	if (!__private.isValidSignature(signature, members, transaction)) {
+		return setImmediate(cb, 'Unable to process signature, verification failed');
```
nazarhussain: Send an error object instead to capture the trace `new Error('Unable to process signature, verification failed')`

modules/multisignatures.js
```diff
@@ -80,158 +80,221 @@ function Multisignatures(cb, scope) {
...
 	}
 
-	if (multisignatureTransaction.type === transactionTypes.MULTI) {
-		multisignatureTransaction.signatures =
-			multisignatureTransaction.signatures || [];
+	// Add signature to transaction
+	transaction.signatures.push(signature.signature);
+	// Check if transaction is ready to be processed
+	transaction.ready = Multisignature.prototype.ready(
```
nazarhussain: Please add `scope.logic.multisignature` to `library` and use its object instead of directly accessing a method through `prototype`. Calling a prototype method directly is only used while implementing object specific behavior to override something in object.

modules/multisignatures.js
```diff
@@ -80,158 +80,221 @@ function Multisignatures(cb, scope) {
...
+	cb
+) => {
+	// Get sender account of correscponding transaction
+	modules.accounts.getAccount(
+		{ address: transaction.senderId },
+		(err, sender) => {
+			if (err || !sender) {
+				const message = 'Unable to process signature, account not found';
+				library.logger.error(message, { signature, transaction, error: err });
+				return setImmediate(cb, message);
```
nazarhussain: Please return an error object instead `new Error(message)`

modules/multisignatures.js
```diff
@@ -80,158 +80,221 @@ function Multisignatures(cb, scope) {
...
+			// Emit events
+			library.network.io.sockets.emit(
+				'multisignatures/signature/change',
+				transaction
+			);
+			library.bus.message('signature', signature, true);
+			return setImmediate(cb);
+		}
+	);
+};
```
nazarhussain: `__private.processSignatureFromMultisignatureAccount` and `__private.processSignatureForMultisignatureAccountCreation` have a lot of similar code, can we refactor those two to share the same logic.

modules/multisignatures.js
```diff
@@ -80,158 +80,221 @@ function Multisignatures(cb, scope) {
...
+			const message = 'Unable to process signature, signature already exists';
+			library.logger.error(message, { signature, transaction });
+			return setImmediate(balanceSequenceCb, message);
+		}
 
-			library.network.io.sockets.emit(
-				'multisignatures/signature/change',
-				multisignatureTransaction
+		// Process signature for multisignature account creation transaction
+		if (transaction.type === transactionTypes.MULTI) {
```
nazarhussain: I believe this check should be in start of function and return if its not a multi-sig transaction.

modules/multisignatures.js
4miners: It's not a real normalization, I'm just removing `+` from `publicKey` to be able to use same function for validating later.

modules/multisignatures.js
4miners: I don't understand, can you clarity? We're doing the check for multisignature transaction here - https://github.com/LiskHQ/lisk/pull/2275/files#diff-7d87322fa8c7f1c80db1ecbc41cc0bb5R265

4miners:
Comments resolved, please review again. 🤗

modules/multisignatures.js
```diff
@@ -67,171 +74,254 @@ function Multisignatures(cb, scope) {
...
-					key,
-					transaction.signature
-				);
-			}
-		} catch (e) {
-			library.logger.error(e.stack);
-			return setImmediate(cb, 'Failed to verify signature');
-		}
+	// Set empty object as sender, as we don't need sender in case of multisignature registration
+	const sender = {};
```
SargeKhan: Are we sure that we don't need to get sender account? What would happen if there multiple multisig registration transactions from the same account?

modules/multisignatures.js
```diff
@@ -67,171 +74,254 @@ function Multisignatures(cb, scope) {
...
+			if (err || !sender) {
+				const message = 'Unable to process signature, account not found';
+				library.logger.error(message, { signature, transaction, error: err });
+				return setImmediate(cb, new Error(message));
 			}
 
-			let verify = false;
-			const multisignatures = account.multisignatures;
+			// Assign members of multisignature account from transaction
+			const membersPublicKeys = sender.multisignatures;
```
SargeKhan: don't need to account for u_multisingatures?

modules/multisignatures.js
SargeKhan: Did you check?

modules/multisignatures.js
4miners: It's not needed. In that case here `sender` have to be already a multisignature account, so need to contains `multisignatures` property (confirmed state). In case of collecting signatures for multisignature account registration transaction - we're grabbing members directly from transaction.

modules/multisignatures.js
4miners: I checked logic in `Multisignature.prototype.ready` and passing `sender` in that case is not needed because it will never contains `multisignatures` property (account is not yet multisignature account). There is a non-explicit check for exactly that thing - when transaction is multisignature registration then compare things from `transaction` object itself instead of `sender`.

modules/multisignatures.js
4miners: Yes, it's fine:
```
[ERR] 2018-08-02 23:28:25 | Unable to process signature, signature already exists - {"signature":{"transactionId":"6242316680238493239","publicKey":"7782833dde90339776320cd765d1acbacb46e2cda060bc8b0b065a8f0579abe1","signature":"8ceae235015aca221ce888395b6770972f7eca39485f75674fa72682b9b751604c395ffd0afb9e27862fb85e2fdab1a179d8264769e8d6e77f55d391dd613c0f"},"transaction":{"amount":"0","senderPublicKey":"7ddf8cdcf374caa99c7a12051458534c0753916762bc0c47c75019e4997385e4","timestamp":69143266,"type":4,"fee":"1500000000","asset":{"multisignature":{"min":2,"lifetime":1,"keysgroup":["+7782833dde90339776320cd765d1acbacb46e2cda060bc8b0b065a8f0579abe1","+d88800d9290b595a1b692b29c91759e49c64d41b2fbae58b745f84297a92834c"]}},"signature":"e22ff1f54f7f89b6276aace8cee8c4098833f88a5c3e4610cc5051b124e0564e9fc3a4a5e8b5521cdd73cd320cb866a304dd7a54950d21bd727188df06e43202","id":"6242316680238493239","senderId":"16150081938385139566L","relays":1,"receivedAt":"2018-08-02T23:28:24.853Z","signatures":["8ceae235015aca221ce888395b6770972f7eca39485f75674fa72682b9b751604c395ffd0afb9e27862fb85e2fdab1a179d8264769e8d6e77f55d391dd613c0f"],"ready":false}}
[ERR] 2018-08-02 23:28:25 | Unable to process signature, signer not in keysgroup. - {"signature":{"transactionId":"6242316680238493239","publicKey":"3f0f1aab87667899844ae92c4a80b2a795396e3cef9d49f766bef1222946ea7d","signature":"7e84bb8dd2fee2ae9cf254ae679e136839a9f47e40ca760c6fcc769b57c7c4844973bb70a85a987bf8c2b990f43268e2735e28bfe5a156c055afd475b85a340a"},"membersPublicKeys":["7782833dde90339776320cd765d1acbacb46e2cda060bc8b0b065a8f0579abe1","d88800d9290b595a1b692b29c91759e49c64d41b2fbae58b745f84297a92834c"],"transaction":{"amount":"0","senderPublicKey":"7ddf8cdcf374caa99c7a12051458534c0753916762bc0c47c75019e4997385e4","timestamp":69143266,"type":4,"fee":"1500000000","asset":{"multisignature":{"min":2,"lifetime":1,"keysgroup":["+7782833dde90339776320cd765d1acbacb46e2cda060bc8b0b065a8f0579abe1","+d88800d9290b595a1b692b29c91759e49c64d41b2fbae58b745f84297a92834c"]}},"signature":"e22ff1f54f7f89b6276aace8cee8c4098833f88a5c3e4610cc5051b124e0564e9fc3a4a5e8b5521cdd73cd320cb866a304dd7a54950d21bd727188df06e43202","id":"6242316680238493239","senderId":"16150081938385139566L","relays":1,"receivedAt":"2018-08-02T23:28:24.853Z","signatures":["8ceae235015aca221ce888395b6770972f7eca39485f75674fa72682b9b751604c395ffd0afb9e27862fb85e2fdab1a179d8264769e8d6e77f55d391dd613c0f"],"ready":false}}
```

4miners:
@SargeKhan I don't like `beforeEach` because it will be executed before every `it` statement, which is not the desired behavior. There are particular steps there than needs to be executed in order. I can move entire process to `beforeEach` but it's not very reliable. What do you think?

SargeKhan:
That's where nested describe blocks come into play.
```
-> describe()
---> beforeEach() (executed for all its)
---> it()
---> describe()
-----> beforeEach() (executed only for tests within this describe block)
-----> it()
---> describe()
-----> it() (only top level beforeEach will be executed)
```
But I think your solution is good. And we don't need to change those tests

MaciejBaj:
Waiting for unit tests for processSignature function.

PR #2287, opened 2018-08-06
Equalise to bytea type all the parameters related to publicKey - Closes #1694

### LINKED ISSUE #1694: Equalise to bytea type all the parameters related to publicKey
### Expected behavior

All the references to publicKey presented in the database should be stored as the same data type `bytea`.

### Actual behavior

Some of them are as `bytea` , some others are as `varchar(64)`.

`1.1.0` ?

SargeKhan:
### What was the problem?
Public Key was stored in an inconsistent manner within database. Sometimes in text and sometimes in bytea.
### How did I fix it?
Converted all text type columns to bytea for storing publicKeys.
### How to test it?
Run tests
### Review checklist

* The PR solves #1694 
* Linting passes
* Tests pass


SargeKhan:
For me, the benefits of making this change are as follow:
1- Ensure consistency in how hex strings are stored in the database. Note that public key as well as transaction and block signatures are hex strings. Signatures everywhere were stored as binary (as per my research).
2- Since public keys and signatures are hex strings, storing them in bytea is the most performant solution because the size of storing public key in text or VARCHAR format will be bigger.
3- Storing public key and signatures in bytea format ensures that we never store an invalid string in the database. So, the application will fail early. I've fixed some tests which used invalid hex strings as public keys, but there was no error received for them. (It's not the strongest point, but I think it's still valuable)

SargeKhan:
@MaciejBaj please give input on this PR.

MaciejBaj:
@nazarhussain 
> We can segregate our data flow in three buckets, outside app (API), inside app (js code), in store (database)

I agree that it would be perfect to have a consistency across all 3 buckets - API, Code, DB. And for increased readability I would also prefer to change it to `hex string`.

However, currently we have no consistency in the 3rd bucket itself - DB. This issue introduces it by converting publicKeys format from `hex string` to `bytea` in mem_accounts and dependent tables, leaving `trs` table untouched. The other way around - conversion publicKey in trs to `hex string` could be beneficial when it comes to performance and simplicity - `mem_accounts`. The measures needs to be performed to establish it.

4miners:
Closed in favor of https://github.com/LiskHQ/lisk/issues/2314.

PR #2297, opened 2018-08-08
Remove redundant tests out of integration test suite - Closes #2274

### LINKED ISSUE #2274: Remove redundant tests out of integration test suite
### Expected behavior
Remove out of the integration test suite:
- stress tests, 
- sync only,
- broadcast only.

They are tested in QA process for each release candidate.
### Actual behavior
The mentioned steps are duplicated in QA and integration tests.

`>=1.0.0`

ManuGowda:
### What was the problem?
The stress is duplicated and the same stress test is being tested on `lisk-core-qa`, also the broadcast and sync only mechanism needs to be moved to `lisk-core-qa`. 
### How did I fix it?
Deleted all the unwanted stress test and related configuration changed from `network`
### Review checklist

* The PR solves #2274 
* Linting passes
* Tests pass


nazarhussain:
@ManuGowda I am ok with the change but not convinced about the direction. If there is some automated testes which been handled in QA process, we should move those to Core test suits not the vice versa.

MaciejBaj:
@nazarhussain that's the final goal we've agreed on. Those tests would be a part of automated test suite after the network tests would be fully automated to run on any (local or remote machines).

PR #2298, opened 2018-08-08
Add epochTime to node/status endpoint - Closes #1733

### LINKED ISSUE #1733: Add current time to node/status endpoint
### Expected behavior

node/status endpoint should also return current time known by the node

### Actual behavior

node/status endpoint does not return current time

### Description

We have experienced timestamp in the future problems with clients that might be a few seconds in the future in comparison with the node they talk to. It would be great to have a comparison of timestamps in node/status so that clients are aware of any differences.

yatki:
### Review checklist

* The PR resolves #1733 
* Linting passes
* Tests pass

### DIFF
--- a/modules/node.js
+++ b/modules/node.js
@@ -171,7 +171,7 @@ Node.prototype.shared = {
 				broadhash: modules.system.getBroadhash(),
 				consensus: modules.peers.getLastConsensus() || null,
 				currentTime: Date.now(),
-				epochTime: slots.getTime(),
+				secondsSinceEpoch: slots.getTime(),
 				height: modules.blocks.lastBlock.get().height,
 				loaded: modules.loader.loaded(),
 				networkHeight,
--- a/modules/node.js
+++ b/modules/node.js
@@ -16,6 +16,7 @@
 
 const _ = require('lodash');
 const BlockReward = require('../logic/block_reward.js');
+const slots = require('../helpers/slots.js');
 
 const constants = global.constants;
 
@@ -170,6 +171,7 @@ Node.prototype.shared = {
 				broadhash: modules.system.getBroadhash(),
 				consensus: modules.peers.getLastConsensus() || null,
 				currentTime: Date.now(),
+				epochTime: slots.getTime(),
 				height: modules.blocks.lastBlock.get().height,
 				loaded: modules.loader.loaded(),
 				networkHeight,
--- a/schema/swagger.yml
+++ b/schema/swagger.yml
@@ -2334,7 +2334,7 @@ definitions:
       currentTime:
         type: integer
         example: 1533558858128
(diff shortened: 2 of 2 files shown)


modules/node.js
```diff
@@ -170,6 +171,7 @@ Node.prototype.shared = {
 				broadhash: modules.system.getBroadhash(),
 				consensus: modules.peers.getLastConsensus() || null,
 				currentTime: Date.now(),
```
MaciejBaj: So, currentTime can be used to sync a clock, epochTime to estimate transaction timestamp.

yatki:
@4miners regarding your point on tests: the swagger definition validates the properties. @nazarhussain told me that the properties will be automatically tested which is the case. 

For other remarks, I wasn't aware of them. It's better if @MaciejBaj comments on them?

schema/swagger.yml
```diff
@@ -2334,7 +2334,11 @@ definitions:
       currentTime:
         type: integer
         example: 1533558858128
-        description: Current time of the node (Unix Time Stamp).
+        description: Current time of the node in miliseconds (Unix Time Stamp).
```
diego-G: Typo: `Timestamp`

4miners:
@yatki I mean unit tests:
https://github.com/LiskHQ/lisk/blob/b96348878e919e1c101b62a3dac187077d755709/test/unit/modules/node.js#L314

MaciejBaj:
@4miners so out of node status after this PR you can receive both epochTime and currentTime. 
Yes, if you are creating a transaction and want to assign some timestamp you could:
1. GET node/status/epochTime
2. GET node/status/currentTime
3. Do the math transaction.timestamp = `(currentTime - epochTime) / 1000`
The endpoint node/status/epochTime does all 3 steps for you.

> epochTime is network-specific, doesn't fit into /node/status endpoint which is node-specific

it's not network specific anymore when node gets out of time sync (`ntp` not performed). The purpose of the currentTime / epochTime would be to validate if a node you're sending a transaction to is not out of sync.

MaciejBaj:
@yatki would be nice if you could fill the skeletons of corresponding `getStatus` unit tests.

schema/swagger.yml
```diff
@@ -2334,7 +2334,11 @@ definitions:
       currentTime:
         type: integer
         example: 1533558858128
-        description: Current time of the node (Unix Time Stamp).
+        description: Current time of the node in miliseconds (Unix Timestamp).
+      epochTime:
```
nazarhussain: It may have some usability on third party. But to avoid any confusion we should rename this attribute to `secondsSinceEpoch`, because when ever we talk about `epochTime` it always referred to a constant value specified to a network.

yatki:
@nazarhussain @diego-G can you take a look again.

PR #2299, opened 2018-08-08
Refactor checkConfirmed function to behave as expected - Closes #1822

### LINKED ISSUE #1822: Naming of "checkConfirmed" is confusing
### Expected behavior

When reading the name "checkConfirmed", I expect that `confirmed` is the positive state and `unconfirmed` is the error state.

### Actual behavior

The method checkConfirmed returns an error for confirmed transactions. 

### Steps to reproduce

See code https://github.com/LiskHQ/lisk/blob/0f8a54d/test/unit/logic/transaction.js and https://github.com/LiskHQ/lisk/blob/0f8a54d/modules/blocks/verify.js

Current development branch

---------------

There is already a related comment bug in "// Check if transaction is already in database, otherwise fork 2" which should be "// If transaction is already in database, then fork 2".

I suggest renaming the method to `checkUnconfirmed`.

yatki:
### What was the problem?
checkConfirmed function wasn't returning intuitively expected results.
### How did I fix it?
refactored the checkConfirmed function and also the places that it was being used.

### Review checklist

* The PR resolves #1822 
* Linting passes
* Tests pass

### DIFF
--- a/logic/transaction.js
+++ b/logic/transaction.js
@@ -285,6 +285,9 @@ class Transaction {
 	 * @todo Add description for the params
 	 */
 	checkConfirmed(transaction, cb) {
+		if (!transaction || !transaction.id) {
+			return setImmediate(cb, 'Invalid transaction id', false);
+		}
 		this.countById(transaction, (err, count) => {
 			if (err) {
 				return setImmediate(cb, err, false);
@@ -691,7 +694,11 @@ class Transaction {
 
 		if (checkExists) {
 			this.checkConfirmed(transaction, (checkConfirmedErr, isConfirmed) => {
-				if (checkConfirmedErr || isConfirmed) {
+				if (checkConfirmedErr) {
+					return setImmediate(cb, checkConfirmedErr);
+				}
+
+				if (isConfirmed) {
 					return setImmediate(
 						cb,
 						`Transaction is already confirmed: ${transaction.id}`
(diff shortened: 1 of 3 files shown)


logic/transaction.js
```diff
@@ -693,9 +690,12 @@ class Transaction {
 		};
 
 		if (checkExists) {
-			this.checkConfirmed(transaction, checkConfirmedErr => {
-				if (checkConfirmedErr) {
-					return setImmediate(cb, checkConfirmedErr);
+			this.checkConfirmed(transaction, (checkConfirmedErr, isConfirmed) => {
+				if (checkConfirmedErr || isConfirmed) {
```
SargeKhan: We should return different error messages to callback based on whether `checkConfirmredErr` had an error value or `isConfirmed` was set to true.

test/unit/logic/transaction.js
```diff
@@ -376,20 +376,24 @@ describe('transaction', () => {
...
 
 			transactionLogic.checkConfirmed(transaction, err => {
-				expect(err).to.not.exist;
+				expect(err).to.be.a('null');
 				done();
 			});
 		});
 
-		it('should return error for transaction which is already confirmed', done => {
+		it('should return true for transaction which is already confirmed', done => {
```
SargeKhan: Add another case for when there is a `Transaction#countById error` (after making the changes described in the first comment.

yatki:
@4miners We discussed it with @MaciejBaj and decided to refactor the function to have proper behaviour. I'd prefer to keep the changes since it returns confirmed status as a parameter and separated from err message.

PR #2300, opened 2018-08-08
Remove broadcast flag & always update system headers - Closes #2261

### LINKED ISSUE #2261: RPC status call gives wrong height when syncing
### Expected behavior

The RPC status call returns the same height as the HTTP during sync

### Actual behavior

The RPC status call returns a fixed height and broadhash for a very ling time, even if the node's height increases during sync.

In this example, I called the RPC status every 2 seconds on my node between 20:12:01 and 20:36:40 UTC. The returned status increases only a handful of times and I get the same height for a very long period of time

| height | number of status responses |
| ----| ----|
|  5030757 | 57 |
| 5038050 | 149 |
| 5039766 | 51 |
| 5041515 | 405 |
| 5044650 | 78 |

* [Series of status responses](https://gist.github.com/webmaster128/bbf730f669fea48d4411d612c494d290)
* [Logs of the syncing node](https://gist.github.com/webmaster128/9e5aa466be0cbbe6459f952355c98bdf)

### Steps to reproduce

Repeatedly run the RPC status call on a given node that is syncing

Testnet, 1.0.0-rc.1

yatki:
### Review checklist

* The PR resolves #2261 
* Linting passes
* Tests pass

### DIFF
--- a/modules/blocks/verify.js
+++ b/modules/blocks/verify.js
@@ -821,7 +821,7 @@ Verify.prototype.processBlock = function(block, broadcast, saveBlock, cb) {
 			// 'false' if block comes from chain synchronisation process
 			updateSystemHeaders(seriesCb) {
 				// Update our own headers: broadhash and height
-				broadcast ? modules.system.update(seriesCb) : seriesCb();
+				modules.system.update(seriesCb);
 			},
 			broadcastHeaders(seriesCb) {
 				// Notify all remote peers about our new headers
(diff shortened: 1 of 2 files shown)


4miners:
After the changes we're always calling `modules.system.update` from inside `Verify.prototype.processBlock`. `System.prototype.update` is calling `System.prototype.getBroadhash` which is performing database query `SELECT * FROM blocks_list ORDER BY b_height DESC LIMIT 5 OFFSET 0` and calculate SHA-256. While both the query (0.5ms) and hash operation are fast - they are executed for every block. Those operations are not needed for snapshotting and will decrease the performance:
6000000  blocks * 0.5 ms = 3000000 ms = 3000 s = 50 minutes.
So change introduced in this PR will make snapshotting process longer by at least 50 minutes (only because of query). Please consider not executing those in when snapshotting.

PR #2308, opened 2018-08-09
Fix application hanging during clean up when processing block - Closes #2271

### LINKED ISSUE #2271: Application halts clean up when processing block
### Expected behavior
The app should clean up the all the modules gracefully.
### Actual behavior
Cleaning up of block processing gets stuck, due to which the application never restarts and continues logging message: `Waiting for block processing to finish...`
### Steps to reproduce
Unclear. Occurred once in snapshotting process.
1.X

SargeKhan:
### What was the problem?
The application halts cleanup process if the block is being processed. Problem is described in detail in the comments of #2271 
### How did I fix it?
Removed usage of t.batch from round tick function.
### How to test it?
I've added tests, but they don't reproduce the exact problem encountered in the issue. The edge case can be reproduced as described [here.](https://github.com/LiskHQ/lisk/issues/2271#issuecomment-411771099)
### Review checklist

* The PR solves #2271 
* Linting passes
* Tests pass


SargeKhan:
I've implemented a solution to the problem as I saw fit. But the discussion is still continuing on issue #2271. So, the decision on merging this PR should be made afterward.

nazarhussain:
@SargeKhan Can you please try out the the workaround @vitaly-t suggested on the issue https://github.com/LiskHQ/lisk/issues/2271#issuecomment-411876661

SargeKhan:
@nazarhussain, @vitaly-t's suggestion also doesn't work.

4miners:
I fixed this PR as @SargeKhan asked me to do.

Regarding the solution:
- I'm not really sure what are the implications on performance by dropping `t.batch`
- we're using `t.batch` in 6 other places in our codebase, which should be unified

vitaly-t:
@SargeKhan That is a radical idea :smile: I will be getting into the dev tasks this week, and will be looking into the issue more closely 😉

4miners:
Suspended, until we confirm if how significant is the performance drop of the solution.

PR #2310, opened 2018-08-12
Reset forging.delegates on migration of old secrets only - Closes #2311

### LINKED ISSUE #2311: Do not reset forging.delegates section on update if no old secrets to migrate
### Expected behavior
The update script should not touch a possibly existing forging.delegates section, if there are no old secrets to migrate.

### Actual behavior
The update script resets the forging.delegates section regardless of migrating old secrets or not.
It's not possible to prepare the forging.delegates key manually before update to core 1.0.0 (assuming forging.secret is empty)

Lisk Core 1.0.0

TheGoldenEye:
### What was the problem?
The update script resets the forging.delegates section regardless of migrating old secrets or not. 
It's not possible to prepare the forging.delegates key manually before update to core 1.0.0 (assuming forging.secret is empty)

### How did I fix it?
Reset forging.delegates key only, if there are old secrets to migrate.

This PR fixes #2311

### DIFF
--- a/scripts/update_config.js
+++ b/scripts/update_config.js
@@ -135,7 +135,6 @@ if (oldConfig.forging.secret && oldConfig.forging.secret.length) {
 }
 
 function migrateSecrets(password) {
-	oldConfig.forging.delegates = [];
 	password = password.trim();
 	if (!password) {
 		console.info('\nSkipping the secret migration.');
@@ -153,6 +152,7 @@ function migrateSecrets(password) {
 	}
 
 	console.info('\nMigrating your secrets...');
+	oldConfig.forging.delegates = [];
 	oldConfig.forging.secret.forEach(secret => {
 		console.info('.......');
 		oldConfig.forging.delegates.push({
--- a/scripts/update_config.js
+++ b/scripts/update_config.js
@@ -129,8 +129,11 @@ if (oldConfig.forging.secret && oldConfig.forging.secret.length) {
 		migrateSecrets(program.password);
 		copyTheConfigFile();
 	}
-} else {
-	migrateSecrets('');
+} else {		// migration not necessary, initialize key, if not yet exist
+	if (!oldConfig.forging.delegates) {
+		oldConfig.forging.delegates = [];
+	}
+
 	copyTheConfigFile();
 }
 
--- a/scripts/update_config.js
+++ b/scripts/update_config.js
@@ -135,6 +135,7 @@ if (oldConfig.forging.secret && oldConfig.forging.secret.length) {
 }
 
 function migrateSecrets(password) {
+	oldConfig.forging.delegates = [];
 	password = password.trim();
 	if (!password) {
 		console.info('\nSkipping the secret migration.');
@@ -152,7 +153,6 @@ function migrateSecrets(password) {
 	}
 
 	console.info('\nMigrating your secrets...');
-	oldConfig.forging.delegates = [];
 	oldConfig.forging.secret.forEach(secret => {
 		console.info('.......');
 		oldConfig.forging.delegates.push({


TheGoldenEye:
My intention is to get a smooth upgrade to core 1.0.0. Especially for us delegates it is important to enable an automated update on many nodes. Adding the encrypted passphrase manually after the migration is not a solution. I think you remember the instable testnet network after the migration to 1.0-rc1.
I agree with you, that the key 'forging.delegates' must be initialized, if an old secret exists in the config, But in the other case, I think the advantages outweigh to tolerate the existence of this key once with this migration.

nazarhussain:
@TheGoldenEye There is couple of ways to achieve the migration; 

1. Do the migration with `lisk_bridge.sh` and wait for the announced height and type-in the encryption password. 
2. Set the password to `LISK_MASTER_PASSWORD` variable run the `lisk_bridge.sh` in background 
3. Convert secrets to encrypted passphrase before hand, during migration skip the secrets migration. After migration add the encrypted passphrase and restart the node. 

Here is the observations on these options: 

- Option 1, is the recommended one and mostly tested by community during the testnet migration to rc1. 
- Option 2, is a workaround for automated migration but NOT recommended by us. As it would show up your master password in the ps list, so not secure. 
- Option 3, is also recommended and tested

With that said, I believe we on certain phases of release that we should proceed with options which are already tested by us and community.

TheGoldenEye:
@nazarhussain 

- The  options 1 and 2 are not meaningful because a delegate usually does not store the secret in the config file (or maybe: should not store) because of security reasons. Additionally most delegates use multiple nodes (forging and backup nodes) so forging-switching from outside is necessary. Maybe option 1 is mostly tested in testnet, but this is not a usecase because testnet delegates do not care about the security.
- Option 3 sounds good but is not automated. That's why I suggested that a possibly existing encrypted passphrase is not deleted during the migration. This allows a smooth migration without downtime.

btw,, the proposed PR consists of one moved line only, the effects could be predictable ;)

MaciejBaj:
@TheGoldenEye solutions 1. and 2. suggested by @nazarhussain are not storing delegates secrets anywhere, it is just a decryption password that is being provided. Those 2 ways are not recommended due to security reasons, anyone with access to delegates machines could decrypt his secrets. 
If a delegate is aware of that security risk but still chooses option 2., the decryption password can be changed anytime after a migration. A migration proceeds in a fully automated way using option 2..

The solution proposed by you gives delegates an option to include a newly introduced delegates section to an old config beforehand, leaving the responsibility of creating a proper structure up to a delegate. The only difference between Option 3: 
> Convert secrets to encrypted passphrase before hand, during migration skip the secrets migration. After migration add the encrypted passphrase and restart the node.

is that after a migration node doesn't have to be restarted. A delegate needs to enable his forging via API PUT call anyway. Nazar mentioned 3 options, there is more. A delegate willing to have his config.json prepared on his own might simply migrate installing Lisk Core v1.0.0 from builds or sources (https://lisk.io/documentation/lisk-core/upgrade) and enable forging via API PUT call just after migration happens.

I feel sceptcal about accepting this PR. Releasing version `1.0.0-rc.3` with a change untested during test migration (as its not a breaking change) puts the most important migration of 1.0.0 to Mainnet into unnecessary risk, while not introducing a functionality that wouldn't be present now.

TheGoldenEye:
I've changed the PR to make it a little bit clearer.
If **no old secret exist** in the config.json, no migration is necessary.
In this case, a possibly existing 'forging.delegates' key is retained, otherwise it is initialized.
I hope you can accept this PR for the core 1.0.0 release, even if he is untested.

MaciejBaj:
@TheGoldenEye from the mentioned reasons we won't accept the PR. I hope you understand my concerns.

PR #2313, opened 2018-08-14
Disable updating system headers while snapshotting - Closes #2261

### LINKED ISSUE #2261: RPC status call gives wrong height when syncing

yatki:
### What was the problem?
As @4miners stated here https://github.com/LiskHQ/lisk/pull/2300#issuecomment-412379504: 

> After the change we're always calling modules.system.update from inside Verify.prototype.processBlock. System.prototype.update is calling System.prototype.getBroadhash which is performing database query SELECT * FROM blocks_list ORDER BY b_height DESC LIMIT 5 OFFSET 0 and calculate SHA-256. While both the query (0.5ms) and hash operation are fast - they are executed for every block. Those operations are not needed for snapshotting and will decrease the performance:
6000000 blocks * 0.5 ms = 3000000 ms = 3000 s = 50 minutes.
So change introduced in this PR will make snapshotting process longer by at least 50 minutes (only because of the query). Please consider not executing those when snapshotting.

### How did I fix it?
I disabled updating system headers while snapshotting.

### Review checklist

* The PR resolves #2261 
* Linting passes
* Tests pass

### DIFF
--- a/modules/blocks/verify.js
+++ b/modules/blocks/verify.js
@@ -45,7 +45,7 @@ __private.lastNBlockIds = [];
  * @todo Add description for the class
  */
 class Verify {
-	constructor(logger, block, transaction, db) {
+	constructor(logger, block, transaction, db, config) {
 		library = {
 			logger,
 			db,
@@ -53,6 +53,7 @@ class Verify {
 				block,
 				transaction,
 			},
+			config,
 		};
 		self = this;
 		__private.blockReward = new BlockReward();
@@ -821,7 +822,9 @@ Verify.prototype.processBlock = function(block, broadcast, saveBlock, cb) {
 			// 'false' if block comes from chain synchronisation process
 			updateSystemHeaders(seriesCb) {
 				// Update our own headers: broadhash and height
-				modules.system.update(seriesCb);
+				!library.config.loading.snapshotRound
+					? modules.system.update(seriesCb)
+					: seriesCb();
 			},
 			broadcastHeaders(seriesCb) {
 				// Notify all remote peers about our new headers
--- a/modules/blocks/verify.js
+++ b/modules/blocks/verify.js
@@ -53,7 +53,11 @@ class Verify {
 				block,
 				transaction,
 			},
-			config,
+			config: {
+				loading: {
+					snapshotRound: config.loading.snapshotRound,
+				},
+			},
 		};
 		self = this;
 		__private.blockReward = new BlockReward();
--- a/test/unit/modules/blocks/verify.js
+++ b/test/unit/modules/blocks/verify.js
@@ -1937,8 +1937,8 @@ describe('blocks/verify', () => {
 			done();
 		});
 
-		describe('system headers', () => {
-			it('should be updated if snapshotting was not activated', done => {
+		describe('system update', () => {
+			it('should be called if snapshotting was not activated', done => {
 				blocksVerifyModule.processBlock(
 					dummyBlock,
 					broadcast,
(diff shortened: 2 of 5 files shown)


modules/blocks/verify.js
```diff
@@ -45,14 +45,15 @@ __private.lastNBlockIds = [];
...
-	constructor(logger, block, transaction, db) {
+	constructor(logger, block, transaction, db, config) {
 		library = {
 			logger,
 			db,
 			logic: {
 				block,
 				transaction,
 			},
+			config,
```
4miners:
```
config: {
	loading: {
		snapshotRound: config.loading.snapshotRound,
	},
},
```

modules/blocks/verify.js
```diff
@@ -821,7 +822,9 @@ Verify.prototype.processBlock = function(block, broadcast, saveBlock, cb) {
 			// 'false' if block comes from chain synchronisation process
 			updateSystemHeaders(seriesCb) {
 				// Update our own headers: broadhash and height
-				modules.system.update(seriesCb);
+				!library.config.loading.snapshotRound
```
4miners: Maybe the opposite?
`library.config.loading.snapshotRound ? seriesCb() : modules.system.update(seriesCb);`

test/unit/modules/blocks/verify.js
```diff
@@ -1931,6 +1937,44 @@ describe('blocks/verify', () => {
 			done();
 		});
 
+		describe('system headers', () => {
```
4miners: This file is for unit tests of `modules/blocks/verify.js`, you shouldn't add describe block for `system headers`, but add corresponding paths in `processBlock` tests.

test/unit/modules/blocks/verify.js
```diff
@@ -1942,7 +1986,6 @@ describe('blocks/verify', () => {
 						saveBlock,
 						err => {
 							expect(err).to.be.null;
-							expect(modules.system.update.calledOnce).to.be.true;
```
4miners: We should still test those.

modules/blocks/verify.js
yatki: I considered that too. But I'd like to provide the desired functionality in the first part and I'd prefer to keep the same structure with rest of the short-hand if usages.

test/unit/modules/blocks/verify.js
yatki: It's consistent with the current suite structure.

modules/blocks/verify.js
yatki: @4miners with this usage we will override the config object and loose the other properties?

4miners: Yep

test/unit/modules/blocks/verify.js
yatki: We already cover that case here https://github.com/LiskHQ/lisk/blob/0206f72e80fad817b48e7fe8dc420d54b7af72cb/test/unit/modules/blocks/verify.js#L1941

PR #2315, opened 2018-08-14
Align README with docs - Closes #2295

### LINKED ISSUE #2295: Align README with docs on lisk.io
### Expected behavior
The instructions in `README.md`  of the current release to set up a node, and the instructions on docs.lisk.io should be identical.
In future, we should update README and docs parallel, if changes in both documents are necessary.

### Actual behavior
Current README.md is out of date in some places, e.g. supported OS versions and required prerequisites.

1.0

Tschakki:
### What was the problem?
Documentation in README was not fully up to date with documentation on the lisk docs page.

### How did I fix it?
Compared README and docs pages and updated content.

### Review checklist

* The PR solves #2295 
* Linting passes
* Tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,88 +20,100 @@ The next section details the prerequisites to install Lisk Core from source usin
 
 ### System Install
 
-* Create a new user
+#### Create a new user
 
-  * Ubuntu 14|16 / Debian:
+* Ubuntu:
 
-  ```
-  sudo adduser lisk
-  ```
+```
+sudo adduser lisk
+```
 
-  Note: The lisk user itself does not need any sudo rights to run Lisk Core.
+Note: The lisk user itself does not need any sudo rights to run Lisk Core.
 
-* Tool chain components -- Used for compiling dependencies
+#### Tool chain components
 
-  * Ubuntu 14|16 / Debian:
+Used for compiling dependencies.
 
-    ```
-    sudo apt-get update
-    sudo apt-get install -y python build-essential curl automake autoconf libtool
-    ```
+* Ubuntu:
 
-  * MacOS 10.12-10.13 (Sierra/High Sierra):
+```
+sudo apt-get update
+sudo apt-get install -y python build-essential curl automake autoconf libtool ntp
+```
 
-    Make sure that you have both [XCode](https://developer.apple.com/xcode/) and [Homebrew](https://brew.sh/) installed on your machine.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -12,98 +12,131 @@ Lisk is a next generation crypto-currency and decentralized application platform
...
+
+## Index
+
+* [Pre-Installation](#pre-installation)
+  * [Create lisk user](#create-the-lisk-user)
+  * [Tool Chain Components](#tool-chain-components)
+  * [Git](#git)
+  * [Node.JS](#nodejs)
+  * [PostgreSQL](#postgresql)
+  * [Redis](#redis)
```
MaciejBaj: Redis is optional. Can we mark it with "*" and add an explanation?

README.md
```diff
@@ -12,98 +12,131 @@ Lisk is a next generation crypto-currency and decentralized application platform
...
+curl -sL https://deb.nodesource.com/setup_6.x | sudo -E bash -
+sudo apt-get install -y nodejs
+```
 
-* Node.js serves as the underlying engine for code execution.
+* MacOS 10.12-10.13 (Sierra/High Sierra):
 
-  Install System wide via package manager:
+```
+brew install node@6.14.1
```
ManuGowda: wouldn' this be different from `curl -sL https://deb.nodesource.com/setup_6.x | sudo -E bash -`
the `setup_6.x` would install LTS version of 6 and `brew install node@6.14.1` would install `6.14.1` version of node, so in my opinion both should install the same version of the node we support.

README.md
```diff
@@ -12,98 +12,131 @@ Lisk is a next generation crypto-currency and decentralized application platform
...
 
-    ```
-    brew install node@6.14.1
-    ```
+2. Install nvm following these [instructions](https://github.com/creationix/nvm#installation)
+3. Install the correct version of Node.js using nvm:
 
-* _(Recommended)_ Install using a version manager such as nvm.
+```shell
+nvm install 6.14.1
```
ManuGowda: same here let's make the version compatible across all types of installations.

README.md
Tschakki: I added now a new section `Check correct version`, which is referring to `package.json` to make sure people have the right version installed before continuing.

The `6.14.1` in README can be seen more as example, as we are not too strict about the node version to use. If we want to me more strict, we need to reflect that changes in https://github.com/LiskHQ/lisk/blob/development/package.json#L19as well

PR #2322, opened 2018-08-16
Use DNS name for seed nodes in config.json - Closes #2320

### LINKED ISSUE #2320: Use DNS name instead of IP addresses for seed nodes in config.json
### Expected behavior

`config.json` uses DNS name for seed nodes.
Rationale: DNS names give us more flexibility and are not bound to a particular service provider.

### Actual behavior

IP addresses are being used.

### Steps to reproduce

All versions of lisk core up until now.

fchavant:
### What was the problem?

`config.json` used DNS name for seed nodes.

### How did I fix it?

IP addresses were replaced with DNS names.

### How to test it?

Ensure that `config/{main,test}net/config.json` contains DNS names under `.peers.list[].ip`, make sure those resolve (except for betanet which does not exist at the moment) and that there are no duplicates.

### Review checklist

* The PR solves #2320 
* Linting passes
* Tests pass

### DIFF
--- a/config/betanet/config.json
+++ b/config/betanet/config.json
@@ -9,23 +9,23 @@
 	"peers": {
 		"list": [
 			{
-				"ip": "94.237.41.99",
+				"ip": "betanet-seed-01.lisk.io",
 				"wsPort": 5001
 			},
 			{
-				"ip": "209.50.52.217",
+				"ip": "betanet-seed-02.lisk.io",
 				"wsPort": 5001
 			},
 			{
-				"ip": "94.237.26.150",
+				"ip": "betanet-seed-03.lisk.io",
 				"wsPort": 5001
 			},
 			{
-				"ip": "83.136.249.102",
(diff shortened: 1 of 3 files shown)


fchavant:
@diego-G alphanet only has `127.0.0.1` so it wouldn't make sense; betanet is down at the moment but I guess we could still decide on DNS names and use those. How about `betanet-seed-*.lisk.io`?

willclarktech:
@fchavant I thought devnet was exclusively `127.0.0.1` but alphanet might include other IPs?

fchavant:
@willclarktech I am not absolutely sure what alphanet is (meant to be); I was not involved in its inception. However if you look at it's default configuration there is only `127.0.0.1` at the moment: https://github.com/LiskHQ/lisk/blob/08f92ce21f77257fea870499d6b27dafd5001b16/config/alphanet/config.json#L18

diego-G:
Close in favour of PR #2325

PR #2325, opened 2018-08-20
Use DNS name instead of IP addresses for seed nodes in config.json - Closes #2320

### LINKED ISSUE #2320: Use DNS name instead of IP addresses for seed nodes in config.json

jondubois:
### What was the problem?

`config.json` did not support domain names for seed nodes.

### How did I fix it?

When the node launches, it preprocesses the peer list from the config file and resolves any domain names into IP addresses before creating the relevant peer objects.

This approach was chosen because the node currently relies on IP addresses in order to uniquely identify peers. To allow a peer to have both an IP address AND a domain name would have complicated things; it's easier to resolve all domains during node startup.

I cherry-picked the commits of @fchavant from https://github.com/LiskHQ/lisk/pull/2322 so we can probably close that other PR.

### How to test it?

- In the peer list in the relevant configuration file, change the ip addresses from `127.0.0.1` to the string `localhost`; then run integration tests.
- Run the node locally using `localhost` instead of `127.0.0.1` as the IP address of a peer in the config file; the node should run without errors.

### Review checklist

* The PR solves #2320
* Linting passes
* Tests pass

### DIFF
--- a/app.js
+++ b/app.js
@@ -17,6 +17,8 @@
 var path = require('path');
 var fs = require('fs');
 var d = require('domain').create();
+var dns = require('dns');
+var net = require('net');
 var SocketCluster = require('socketcluster');
 var async = require('async');
 var Logger = require('./logger.js');
@@ -188,7 +190,7 @@ d.run(() => {
 	async.auto(
 		{
 			/**
-			 * Attempts to determine nethash from genesis block.
+			 * Prepare the config object.
 			 *
 			 * @func config
 			 * @memberof! app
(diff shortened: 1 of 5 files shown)


app.js
```diff
@@ -199,7 +201,37 @@ d.run(() => {
 				if (!appConfig.nethash) {
 					throw Error('Failed to assign nethash from genesis block');
 				}
-				cb(null, appConfig);
+
+				// In case domain names are used, resolve those to IP addresses.
```
fchavant: How often are those names resolved? Once on startup or periodically?

jondubois: @fchavant Only on startup; it's basically a preprocessing step done to the config file before it's used by the node. The internal logic only deals knows about IP addresses.

As part of the discovery process, peers still need to report their IPv4 addresses (we don't want peers to report themselves by domain name). If we allow peers to report anything (domain name, IPv4 or IPv6 addresses), our life will get a lot more difficult. Potentially, we could stop relying on the ip address as an identifier, but that change would be a lot more significant and I'm not sure if that would be a good idea.

fchavant: OK, that's fine for me.

MaciejBaj:
@jondubois network tests are failing if you could have a look.

jondubois:
@MaciejBaj It's passing now.

PR #2327, opened 2018-08-22
Unit tests for node/getStatus function - Closes #2312

### LINKED ISSUE #2312: Write unit tests for modules/node/getStatus function
### Expected behavior
We should have unit tests covering modules/node/getStatus function as suggested here: https://github.com/LiskHQ/lisk/pull/2298#issuecomment-412440740

### Actual behavior
We don't have it right now.

### Description
I decided to write unit tests after finishing #1723

yatki:
### Review checklist

* The PR resolves #2312
* Linting passes
* Tests pass

### DIFF
--- a/modules/node.js
+++ b/modules/node.js
@@ -169,7 +169,7 @@ Node.prototype.shared = {
 		modules.peers.networkHeight({ normalized: false }, (err, networkHeight) => {
 			setImmediate(cb, null, {
 				broadhash: modules.system.getBroadhash(),
-				consensus: modules.peers.getLastConsensus() || null,
+				consensus: modules.peers.getLastConsensus(),
 				currentTime: Date.now(),
 				secondsSinceEpoch: slots.getTime(),
 				height: modules.blocks.lastBlock.get().height,
--- a/test/unit/modules/node.js
+++ b/test/unit/modules/node.js
@@ -443,6 +443,11 @@ describe('node', () => {
 							});
 						});
 					});
+
+					afterEach(done => {
+						sinonSandbox.restore();
+						done();
+					});
 				});
 			});
 		});
--- a/test/unit/modules/node.js
+++ b/test/unit/modules/node.js
@@ -418,21 +418,29 @@ describe('node', () => {
 						done();
 					});
 
-					it('should call modules.peers.networkHeight and call callback with result containing networkHeight = network.height', done => {
-						node_module.shared.getStatus(null, (err, status) => {
+					it('should call modules.peers.networkHeight', done => {
+						node_module.shared.getStatus(null, () => {
 							expect(library.modules.peers.networkHeight).to.have.been.called;
+							done();
+						});
+					});
+
+					it('should call callback with result containing networkHeight = network.height', done => {
+						node_module.shared.getStatus(null, (err, status) => {
 							expect(status.networkHeight).to.equal(123);
 							done();
 						});
 					});
 
-					it('should call callback with result containing networkHeight = null if modules.peers.networkHeight returns networkHeight as null', done => {
-						library.modules.peers.networkHeight = sinonSandbox
-							.stub()
-							.callsFake((filters, cb) => cb(null, null));
-						node_module.shared.getStatus(null, (err, status) => {
-							expect(status.networkHeight).to.equal(null);
-							done();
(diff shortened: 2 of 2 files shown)


test/unit/modules/node.js
```diff
@@ -313,40 +321,120 @@ describe('node', () => {
...
+					beforeEach(done => {
+						stubs.networkHeight = sinonSandbox
+							.stub()
+							.callsFake((filters, cb) => cb(null, 123));
+						const modules = library.rewiredModules.node.__get__('modules');
+						modules.peers.networkHeight = stubs.networkHeight;
+						done();
+					});
+
+					it('should call modules.peers.networkHeight and call callback with result containing networkHeight = network.height', done => {
```
MaciejBaj: Looks like 2 separate tests for me - "call callback" and "call modules.peers.networkHeight"

test/unit/modules/node.js
```diff
@@ -313,40 +321,120 @@ describe('node', () => {
...
+						node_module.shared.getStatus(null, (err, status) => {
+							expect(library.modules.peers.networkHeight).to.have.been.called;
+							expect(status.networkHeight).to.equal(123);
+							done();
+						});
+					});
+
+					it('should call callback with result containing networkHeight = null if modules.peers.networkHeight returns networkHeight as null', done => {
```
MaciejBaj: "if modules.peers.networkHeight returns networkHeight as null" condition can be extracted into describe block

test/unit/modules/node.js
```diff
@@ -313,40 +321,120 @@ describe('node', () => {
...
+					});
 				});
 
-				describe('when modules.loader.getNetwork succeeds and returns network', () => {
-					it(
-						'should call callback with result containing networkHeight = network.height'
-					);
+				describe('modules.loader.getNetwork', () => {
+					beforeEach(done => {
+						stubs.networkHeight = sinonSandbox
```
SargeKhan: Does this stub get restored after test?

modules/node.js
```diff
@@ -169,7 +169,7 @@ Node.prototype.shared = {
 		modules.peers.networkHeight({ normalized: false }, (err, networkHeight) => {
 			setImmediate(cb, null, {
 				broadhash: modules.system.getBroadhash(),
-				consensus: modules.peers.getLastConsensus() || null,
+				consensus: modules.peers.getLastConsensus(),
```
SargeKhan: Did you find the reason as to why `|| null` was there earlier?

yatki: @SargeKhan No, actually consensus property is used here https://github.com/LiskHQ/lisk/blob/2312-unit-tests-for-node-getStatus/api/controllers/node.js#L101 in a weird way because if it's null we are converting it back to 0. So `|| null` was totally useless in my opinion. If you think it's better to fix it in a separate issue I can create one for that.

SargeKhan: I was just asking. I think `modules.peers.getLastConsensus()` might return `undefined`. And it makes sense to set the default values in the controller as opposed to modules.

test/unit/modules/node.js
yatki: @SargeKhan Not at the moment. I just made sure in beforeEach hooks they are being reassigned. Actually i was hoping to get a direction on this. This discussion https://github.com/LiskHQ/lisk/issues/1417 suggests to use a _global_hooks.js file but it doesn't exist yet and I followed the structure in `test/unit/modules/multisignatures.js` since it's written recently by @4miners. 

So please feel free to guide me here.

modules/node.js
yatki: I could find only to places that we assign it's value. here: https://github.com/LiskHQ/lisk/blob/2312-unit-tests-for-node-getStatus/modules/peers.js#L70 and here https://github.com/LiskHQ/lisk/blob/2312-unit-tests-for-node-getStatus/modules/peers.js#L475. Seems like it always returns an integer?

SargeKhan: You are right! We don't need that check. I thought maybe on application start `getLastConsensus` may have undefined value.

test/unit/modules/node.js
SargeKhan: I think it's always better to restore spy/stubs in `afterEach` if they were created in a `beforeEach` block.

Issue #2212, opened 2018-07-09
checkIpInList should not manipulate inputs


webmaster128:
### Expected behavior

checkIpInList should not manipulate inputs

### Actual behavior

checkIpInList adds a key ti the input list that serves as a cache. This may be very unexpected for the caller and creates a bug when the input object is updated.

The whole _subNets idea from 2 years ago is probably premature optimization.

### Steps to reproduce

Run `node` in a Lisk installation directory

```
> const checkIpInList = require('./helpers/check_ip_in_list');
> const a = ["127.0.0.1", "111.90.213.195", "89.12.157.122"]
> a
[ '127.0.0.1', '111.90.213.195', '89.12.157.122' ]
> checkIpInList(a, "127.0.0.1")
true
> a.shift()
> checkIpInList(a, "127.0.0.1")
true
```

As you can see in the last command, `checkIpInList(a, "127.0.0.1")` returns true where `a` is `['111.90.213.195', '89.12.157.122' ]`.

### Which version(s) does this affect? (Environment, OS, etc...)

Current development

PR #2344, opened 2018-08-28
Refactor checkIpInList function - Closes #2343, #2212

### LINKED ISSUE #2343: Refactor checkIpInList function
### Expected behavior
Function should accept a list of IPs and an IP as parameters and return true if the list contains the IP and false if not.

### Actual behaviour
The function expects a third parameter `returnListIsEmpty `, which doesn't make sense in the context.

1.2.0

Tschakki:
### What was the problem?
Useless third parameter `returnListIsEmpty ` and conversion of IPs created unnecessary complexity.

### How did I fix it?
Removed `returnListIsEmpty `.
Removed creation of a separate subnet array.

### How to test it?
Run
```
npm run mocha -- test/unit/helpers/http_api.js
npm run mocha -- test/unit/helpers/check_ip_in_list.js
```

### Review checklist

* The PR solves #2343, #2212 
* Linting passes
* Tests pass


helpers/check_ip_in_list.js
```diff
@@ -26,51 +26,22 @@ var ip = require('ip');
  * @param {boolean} returnListIsEmpty - The return value, if list is empty
```
SargeKhan: updage the docs here.

helpers/check_ip_in_list.js
```diff
@@ -26,51 +26,22 @@ var ip = require('ip');
...
  * @returns {boolean} True if ip is in the list, false otherwise
  */
-function CheckIpInList(list, addr, returnListIsEmpty) {
+function CheckIpInList(list, addr) {
 	var i;
```
SargeKhan: more descriptive variable naming here?

helpers/check_ip_in_list.js
```diff
@@ -26,51 +26,22 @@ var ip = require('ip');
...
-		return returnListIsEmpty;
-	}
-
-	// Check subnets
-	for (i = 0, n = list._subNets.length; i < n; i++) {
-		if (list._subNets[i].contains(addr)) {
+	for (i = 0; i < list.length; i++) {
+		if (!ip.isV4Format(list[i])) {
+			var message = `Entry with wrong format in list of IPs: ${list[i]}`;
+			console.error('CheckIpInList:', message);
```
SargeKhan: We should use logger.info/logger.error here. And should not use console.error.

helpers/check_ip_in_list.js
```diff
@@ -26,51 +26,22 @@ var ip = require('ip');
...
-			}
-			try {
-				var subnet = ip.cidrSubnet(entry);
-				list._subNets.push(subnet);
-			} catch (err) {
-				console.error('CheckIpInList:', err.toString());
-			}
-		}
+	if (!ip.isV4Format(addr)) {
+		throw new TypeError('Provided address is not in IPv4 format');
```
SargeKhan: We weren't throwing from the function before, why do we do it now?

SargeKhan: And what will hapen incase its an ipv6 addr?

helpers/check_ip_in_list.js
```diff
@@ -26,51 +26,22 @@ var ip = require('ip');
...
-
 	if (!_.isArray(list) || list.length === 0) {
-		return returnListIsEmpty;
+		return false;
 	}
-
-	if (!list._subNets) {
-		// First call, create subnet list
-		list._subNets = [];
-		for (i = list.length - 1; i >= 0; i--) {
```
SargeKhan: I think this piece of code is still needed.

SargeKhan: We should just stop using the global `list._subNets` variable.

helpers/check_ip_in_list.js
```diff
@@ -26,51 +26,22 @@ var ip = require('ip');
...
-	}
-
-	// Check subnets
-	for (i = 0, n = list._subNets.length; i < n; i++) {
-		if (list._subNets[i].contains(addr)) {
+	for (i = 0; i < list.length; i++) {
+		if (!ip.isV4Format(list[i])) {
+			var message = `Entry with wrong format in list of IPs: ${list[i]}`;
+			console.error('CheckIpInList:', message);
+		} else if (ip.isEqual(list[i], addr)) {
```
SargeKhan: `ip.isEqual` !== `ip.contains`. Refer to npm ip module and `cidrSubnet` for details.

helpers/check_ip_in_list.js
```diff
@@ -26,51 +26,22 @@ var ip = require('ip');
...
 	}
-
-	if (list._subNets.length === 0) {
-		return returnListIsEmpty;
-	}
-
-	// Check subnets
-	for (i = 0, n = list._subNets.length; i < n; i++) {
-		if (list._subNets[i].contains(addr)) {
+	for (i = 0; i < list.length; i++) {
```
SargeKhan: I don't like this loop structure. I think it's better to divide it in two loops as it was previously.
1- Which checks formats of list entries, and creates a list of cidrSubnets.
2- Which checks that IP address is part of any subnet of the list.

test/unit/helpers/check_ip_in_list.js
```diff
@@ -41,27 +41,15 @@ describe('checkIpInList', () => {
...
 		describe('when validList is not an array', () => {
 			before(done => {
 				validReturnListIsEmpty = true;
 				validList = null;
 				done();
 			});
 
-			it('should return validReturnListIsEmpty', done => {
-				expect(checkIpInListResult).to.eq(validReturnListIsEmpty);
+			it('should return false', done => {
```
SargeKhan: We should update the beforeEach here: https://github.com/LiskHQ/lisk/pull/2344/files#diff-3ea1e3ed6b5ed11246f4d2b735f21bd8R39 and remove `validReturnListIsEmpty` usages.

TheGoldenEye:
1. right, the parameter "returnListIsEmpty" is obsolete now. It comes from times there was no explicit config key for an open api and an empty whitelist was a flag for it
2. the new code will not work with ipv6 addresses (if needed at some point)
3. the new code can not handle cidr subnets in the whitelist

helpers/check_ip_in_list.js
Tschakki: dunno, it is a simple iterator. Do you have suggestions?

SargeKhan: Then we should just declare it in the for loop like: `for(let i =0; i < list.length; i++)`

PR #2346, opened 2018-08-29
Add sort support to api/voters endpoint - Closes #2330

### LINKED ISSUE #2330: bad sorting when using balance:asc
### Expected behavior
When retrieving voters for a delegate i'd expect the sort field to be applied properly so that low balance accounts are returned first and page 2 contains voters that have a higher balance than all entries in page 1 (when using balance:asc).

### Actual behavior
Instead voters are queried without any apparent ordering and sorting is applied after the query has run

### Steps to reproduce
please take a look at:  https://github.com/LiskHQ/lisk/blob/development/modules/voters.js#L245-L252

the following 2 queries are just to see the problem:
/api/voters?sort=balance:asc&address=17670127987160191762L&limit=100&offset=0

/api/voters?sort=balance:asc&address=17670127987160191762L&limit=100&offset=100

core 1

yatki:
### Review checklist

* The PR resolves #2330 
* Linting passes
* Tests pass

### DIFF
--- a/db/sql/voters/get_voters.sql
+++ b/db/sql/voters/get_voters.sql
@@ -19,8 +19,8 @@
   PARAMETERS: ?
 */
 
-SELECT address, balance, encode("publicKey", 'hex') AS publicKey FROM mem_accounts2delegates t1
+SELECT t2.address, t2.balance, encode(t2."publicKey", 'hex') AS "publicKey" FROM mem_accounts2delegates t1
 INNER JOIN mem_accounts t2 ON t1."accountId" = t2.address
 WHERE t1."dependentId" = ${publicKey}
-ORDER BY ${sortField:raw} ${sortMethod:raw}
+ORDER BY "${sortField:raw}" ${sortMethod:raw}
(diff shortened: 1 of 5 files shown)


yatki:
@nazarhussain Do you think we should also apply the same functionality to api/votes endpoint in this PR? Should be pretty straightforward. Or I can create another issue for that.

db/sql/voters/get_voters.sql
```diff
@@ -19,7 +19,8 @@
   PARAMETERS: ?
 */
 
-SELECT "accountId"
-FROM mem_accounts2delegates
-WHERE "dependentId" = ${publicKey}
+SELECT t2.address, t2.balance, encode(t2."publicKey", 'hex') AS "publicKey" FROM mem_accounts2delegates t1
+INNER JOIN mem_accounts t2 ON t1."accountId" = t2.address
```
nazarhussain: For better readability, please rename `t2` to `accounts` and `t1` to `delegates`

yatki: @nazarhussain done.

PR #2359, opened 2018-08-30
Poor performance of GET /api/node/status - Closes #2348

### LINKED ISSUE #2348: Poor performance of GET /api/node/status
### Expected behavior
Performance of API `GET /api/node/status` should be satisfactory (under 1 second). 

### Actual behavior
We run some benchmarks with following spec:  

```
CPU: 1vCPU (Digital Ocean) 
RAM: 1 GB Memory
Disk: 25 GB Disk
Connections: 10 
Time: 30s
```
And got the following results;

```
Fastest: 3.17
Slowest: 5.34
Average: 4.36
```

### Steps to reproduce
Query endpoints mentioned above.

1.0.1 (after indexes)

nazarhussain:
### What was the problem?
The `count` on all data was applied which contains millions of transactions records. That was causing the slow response. 

### How did I fix it?
Used cached count until we save any transaction. 

### How to test it?
```
npx mocha test/functional/http/get/node.js
```

### Review checklist

* The PR solves #2348
* Linting passes
* Tests pass

### DIFF
--- a/modules/cache.js
+++ b/modules/cache.js
@@ -328,7 +328,7 @@ Cache.prototype.onTransactionsSaved = function(transactions, cb) {
 
 	async.parallel(
 		[
-			function deleteDelegateCache(parallelCb) {
+			async.reflect(reflectCb => {
 				const pattern = '/api/delegates*';
 
 				const delegateTransaction = transactions.find(
@@ -337,7 +337,7 @@ Cache.prototype.onTransactionsSaved = function(transactions, cb) {
 				);
 
 				if (!delegateTransaction) {
-					return setImmediate(parallelCb, null);
+					return setImmediate(reflectCb, null);
 				}
 
 				self.removeByPattern(pattern, removeByPatternErr => {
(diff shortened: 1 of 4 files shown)


modules/cache.js
```diff
@@ -354,7 +354,20 @@ Cache.prototype.onTransactionsSaved = function(transactions, cb) {
 			return cb(err);
 		});
 	} else {
```
4miners: `transactionsSaved` event is emitted during block saving process with `block.transactions`, doesn't matter if the block contains transactions or not. Then in this function, we are assigning to `delegateTransaction` only if we can find transactions with type `DELEGATE`. So this `else` statement will be not reached if the block contains a transaction of this type. We should rely only on `transactions` to invalidate cache for `transactionCount`.

modules/transactions.js
```diff
@@ -818,23 +820,86 @@ Transactions.prototype.shared = {
...
+
+				function getTransactionCountFromDb(cachedCount, waterCb) {
+					if (cachedCount) {
+						return setImmediate(waterCb, null, cachedCount, null);
+					}
+
+					library.db.transactions
+						.count()
+						.then(transactionsCount =>
+							setImmediate(waterCb, null, cachedCount, transactionsCount)
```
4miners: Why we're passing `cachedCount` instead of `null` explicitly?

modules/transactions.js
```diff
@@ -818,23 +820,86 @@ Transactions.prototype.shared = {
...
+						);
+				},
+
+				function getTransactionCountToCache(cachedCount, dbCount, waterCb) {
```
4miners: In the step above we use `transactionsCount`, here `dbCount`.  Should be nice to keep naming consistent.

modules/transactions.js
```diff
@@ -818,23 +820,86 @@ Transactions.prototype.shared = {
...
+					});
+				},
+			],
+			(err, result) => {
+				if (err) {
+					library.logger.error('Error in getTransactionsCount', err, result);
+					return setImmediate(cb, 'Failed to count transactions');
+				}
+
+				result.total =
```
4miners: Inconsistent naming, before we stored `dbCount` as `total` in `transactionCount` cache.

PR #2360, opened 2018-08-30
Poor performance of GET /api/delegates/[address]/forging_statistics - Closes #2352

### LINKED ISSUE #2352: Poor performance of GET /api/delegates/[address]/forging_statistics
### Expected behavior
Performance of API `GET /api/delegates/[address]/forging_statistics` should be satisfactory (under 1 second). 

### Actual behavior
We run some benchmarks with following spec:  

```
CPU: 1vCPU (Digital Ocean) 
RAM: 1 GB Memory
Disk: 25 GB Disk
Connections: 10 
Time: 30s
```
And got the following results;

```
Fastest: 8.23
Slowest: 8.91
Average: 9.74
```

### Steps to reproduce
Query endpoints mentioned above.

1.0.1 (after indexes)

nazarhussain:
### What was the problem?
Performance of the endpoint `/api/delegates/[address]/forging_statistics` was slow because of aggregating all data when no time filter is provided. 

### How did I fix it?
When no time filter is provided we get data from mem_accounts instead of aggregating. 

### How to test it?

```
npx mocha test/functional/http/get/delegates.js
```

### Review checklist

* The PR solves #2352
* Linting passes
* Tests pass

### DIFF
--- a/modules/delegates.js
+++ b/modules/delegates.js
@@ -24,6 +24,7 @@ const BlockReward = require('../logic/block_reward.js');
 const jobsQueue = require('../helpers/jobs_queue.js');
 const Delegate = require('../logic/delegate.js');
 const slots = require('../helpers/slots.js');
+const Bignum = require('../helpers/bignum.js');
 const transactionTypes = require('../helpers/transaction_types.js');
 
 // Private fields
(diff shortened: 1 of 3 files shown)


modules/delegates.js
```diff
@@ -1077,6 +1078,58 @@ Delegates.prototype.shared = {
...
+			modules.delegates.getDelegates(
+				{ address: filters.address },
+				['rewards', 'fees', 'producedBlocks'],
+				(err, data) => {
+					if (err) {
+						return setImmediate(cb, err);
+					}
+
+					if (!data) {
+						return setImmediate(cb, 'Account not found');
```
4miners: In which cases `data` will be empty? Maybe there is an account, but not a delegate?

modules/delegates.js
```diff
@@ -1077,6 +1078,58 @@ Delegates.prototype.shared = {
...
+				processStatistics(reward);
+			});
+		}
+
+		const processStatistics = reward => {
+			reward.forged = new Bignum(reward.fees)
+				.plus(new Bignum(reward.rewards))
+				.toString();
+
+			return setImmediate(cb, null, reward);
```
4miners: Calling a `getForgingStatistics`'s callback from `processStatistics` function is not very nice. Also, it will be hard to write proper unit tests for it. Can we maybe get rid of this function or make it synchronous? 🤔

PR #2371, opened 2018-09-03
Refactor Lisk core constants - Closes #1774

### LINKED ISSUE #1774: Refactor lisk-core constants to use const, upper case with underscores

yatki:
### Review checklist

* The PR resolves #1774
* Linting passes
* Tests pass


test/integration/rounds.js
```diff
@@ -25,6 +26,7 @@ const randomUtil = require('../common/utils/random');
 const queriesHelper = require('../common/integration/sql/queriesHelper.js');
 const localCommon = require('./common');
 
+global.constants = _.cloneDeep(global.constants);
```
ManuGowda: can you move this logic to `test/setup.js`, in there we are already doing `global.constants = testContext.config.constants;`,  you can modify the logic in setup to cater the constants behavior for the tests.

test/integration/snapshotting.js
```diff
@@ -15,12 +15,14 @@
...
 
 const Promise = require('bluebird');
+const _ = require('lodash');
 const elements = require('lisk-elements').default;
 const randomUtil = require('../common/utils/random');
 const accountsFixtures = require('../fixtures/accounts');
 const queriesHelper = require('../common/integration/sql/queriesHelper.js');
 const localCommon = require('./common');
 
+global.constants = _.cloneDeep(global.constants);
```
ManuGowda: once you move this logic to setup then we don't need this, also you can remove the lodash require. In future, if you want to use lodash, its exposed globally for the test, you can directly use `_`.

test/integration/transactions/expire_transactions.js
```diff
@@ -33,15 +34,16 @@ describe('expire transactions', () => {
 	let library;
 	let queries;
 
+	global.constants = _.cloneDeep(global.constants);
```
ManuGowda: same here.

test/unit/logic/block_reward.js
```diff
@@ -14,13 +14,15 @@
 
 'use strict';
 
+const _ = require('lodash');
```
ManuGowda: remove the lodash require here.

helpers/deep_freeze_object.js
```diff
@@ -0,0 +1,18 @@
+'use strict';
+
+module.exports = function deepFreeze(o) {
+	Object.freeze(o);
+
+	Object.getOwnPropertyNames(o).forEach(prop => {
+		if (
+			Object.prototype.hasOwnProperty.call(o, prop) &&
+			o[prop] !== null &&
+			(typeof o[prop] === 'object' || typeof o[prop] === 'function') &&
```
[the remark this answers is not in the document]
yatki: I copied the logic from here: https://stackoverflow.com/a/34776962/1417536

I think they just wanted to follow what official documentation says: ` Note that strings, numbers, and booleans are always immutable and that Functions and Arrays are objects.`.  See: [Freezing Arrays](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze#Freezing_Arrays)

I think it doesn't bring any value, since the property which holds the function as value will be already frozen and there is no need to freeze a function. I looked into `new Fn()` usage but `typeof new Fn()` is an object. 

hasOwnProperty check is completely useless though. I'll remove that.

config/default/constants.js
```diff
@@ -20,97 +20,108 @@
...
+ * @property {number} BLOCK_RECEIPT_TIMEOUT - Seconds to check if the block is fresh or not.
+ * @property {Date} EPOCH_TIME	- Timestamp indicating the start of lisk core.
+ * @property {Object} FEES - Object representing amount of fees for different types of transactions.
+ * @property {number} FEES.SEND	- Fee for sending a transaction.
+ * @property {number} FEES.VOTE - Fee for voting a delegate.
+ * @property {number} FEES.SECOND_SIGNATURE	- Fee for creating a secondSignature.
+ * @property {number} FEES.DELEGATE - Fee for registering as a delegate.
+ * @property {number} FEES.MULTISIGNATURE - Fee for multisignature transaction.
+ * @property {number} FEES.DAPP_REGISTRATION	- Fee for registering as a dapp.
+ * @property {number} FEES.DAPP_WITHDRAWAL	- Fee for registering as a dapp.
```
yatki: @ManuGowda can you provide descriptions for two constants above, bitte?

config/default/constants.js
```diff
@@ -20,97 +20,108 @@
...
+ * @property {number} MAX_TRANSACTIONS_PER_BLOCK -	Maximum Number of transactions allowed per block.
+ * @property {number} MAX_VOTES_PER_TRANSACTION - The maximum number of votes allowed in transaction type(3) votes.
+ * @property {number} MIN_BROADHASH_CONSENSUS - Minimum broadhash consensus(%) among connected {MAX_PEERS} peers.
+ * @property {number} MULTISIG_CONSTRAINTS - Description needed
+ * @property {number} MULTISIG_CONSTRAINTS.MIN.MINIMUM - Description needed
+ * @property {number} MULTISIG_CONSTRAINTS.MIN.MAXIMUM - Description needed
+ * @property {number} MULTISIG_CONSTRAINTS.LIFETIME.MINIMUM - Description needed
+ * @property {number} MULTISIG_CONSTRAINTS.LIFETIME.MAXIMUM - Description needed
+ * @property {number} MULTISIG_CONSTRAINTS.KEYSGROUP.MIN_ITEMS - Description needed
+ * @property {number} MULTISIG_CONSTRAINTS.KEYSGROUP.MAX_ITEMS - Description needed
```
yatki: @ManuGowda can you provide descriptions for MULTISIG_CONSTRAINTS constants as well?

config/default/constants.js
ManuGowda: `DAPP_WITHDRAWAL` I am not sure what should be the description and we dont use that constant in the application except for the get constants function, maybe @4miners can help to explain here.

config/default/constants.js
ManuGowda: @diego-G Please evaluate if the descriptions are ok and reaffirm @yatki so that he can use the following description.

MULTISIG_CONSTRAINTS.LIFETIME.MINIMUM - Minimum time frame in which the Multisignature transaction will exist in the memory before the transaction is confirmed

MULTISIG_CONSTRAINTS.LIFETIME.MAXIMUM - Maximum time frame in which the Multisignature transaction will exist in the memory before the transaction is confirmed

MULTISIG_CONSTRAINTS.KEYSGROUP.MIN_ITEMS - Minimum number of keys required to sign the Multisignature transaction in the keys group

MULTISIG_CONSTRAINTS.KEYSGROUP.MAX_ITEMS - Maximum number of keys required to sign the Multisignature transaction in the keys group

api/controllers/delegates.js
```diff
@@ -147,7 +147,7 @@ DelegatesController.getForgingStatistics = function(context, next) {
 				count: reward.count,
 			},
 			meta: {
-				fromTimestamp: filters.start || constants.epochTime.getTime(),
+				fromTimestamp: filters.start || constants.EPOCH_TIME.getTime(),
```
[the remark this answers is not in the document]
yatki: @ManuGowda I totally agree on using deconstruction. But there is a consistent usage of constants in the whole project right now with constants.CONTANT_NAME and I would prefer to do this with another issue. This PR is already pretty big.

I will create a separate issue for this now.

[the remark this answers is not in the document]
yatki: Created: https://github.com/LiskHQ/lisk/issues/2381

modules/node.js
```diff
@@ -144,8 +144,17 @@ Node.prototype.shared = {
 		return setImmediate(cb, null, {
 			build: library.build,
 			commit: library.lastCommit,
-			epoch: constants.epochTime,
-			fees: constants.fees,
+			epoch: constants.EPOCH_TIME,
+			fees: {
```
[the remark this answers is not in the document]
yatki: As far as I'm concerned, getContants method is only used in node/constants controller, so should be possible and pretty straightforward. But I would prefer to do it in a separate issue.

yatki:
I will fix comments as well.

PR #2372, opened 2018-09-03
Remove duplicate asset rows from trs tables - Closes #2054

### LINKED ISSUE #2054: Multiple rows for one transaction ID in transaction asset tables in snapshots
### Expected behavior

1. Since the table `votes` contains asset data for transactions, every transaction ID should be contained exactly once for every type 3 transaction
2. A [UNIQUE](https://www.postgresql.org/docs/9.4/static/ddl-constraints.html) constraint on the `transactionId` column should enforce this on a database level and help Postgres to optimize things

### Actual behavior

In the Testnet snapshots of [gr33ndragon](https://testnet-snapshot.lisknode.io/), [Bioly](https://testnet-snapshot.mylisk.com/) and [Lisk HQ](http://snapshots.lisk.io/?prefix=lisk/test/), there are multiple rows for one transaction ID in the votes table.

These are the > 3000 transaction IDs that are contained more than once in lisk_test_backup-5278664.gz from Lisk HQ: https://gist.github.com/webmaster128/2710ae3d3ffeb774e079c5231b451d7c

### Steps to reproduce

Run `SELECT "transactionId" FROM votes GROUP BY "transactionId" HAVING count("transactionId") > 1`

* 0.9.x
* Testnet (Mainnet not tested yet)

----

Even if this does no damage at the moment (I don't know), it's a matter of luck. As soon as you start JOINing votes to transactions, you run into big trouble.

SargeKhan:
### What was the problem?
Due to some previous bug in block save. Votes, Dapps, and Multisignature tables had duplicate entries and the tables also did not have unique constraint.
### How did I fix it?
Removed duplicate rows, and added a PRIMARY KEY constraint.
### How to test it?
Load testnet [with snapshot which contains duplicate entries](http://snapshots.lisk.io.s3-eu-west-1.amazonaws.com/lisk/test/lisk_test_backup-4555600.gz), and start the application. The duplicate entries should be removed by added migrations.
### Review checklist

* The PR resolves #2054 
* Linting passes
* Tests pass

### DIFF
--- a/db/sql/migrations/updates/20180829161959_remove_duplicate_rows_and_add_primary_key_in_dapps.sql
+++ b/db/sql/migrations/updates/20180829161959_remove_duplicate_rows_and_add_primary_key_in_dapps.sql
@@ -1,0 +1,24 @@
+/*
+ * Copyright © 2018 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+BEGIN;
+	DELETE FROM "dapps" "d0"
+		USING "dapps" "d1"
+	WHERE "d0"."ctid" < "d1"."ctid"
+		AND "d0"."transactionId" = "d1"."transactionId";
+
+	ALTER TABLE "dapps"
+		ADD PRIMARY KEY ("transactionId");
+
+COMMIT;
--- a/db/sql/migrations/updates/20180829161959_remove_duplicate_rows_and_add_unique_constraint_in_dapps.sql
+++ b/db/sql/migrations/updates/20180829161959_remove_duplicate_rows_and_add_unique_constraint_in_dapps.sql
@@ -17,6 +17,9 @@ DELETE FROM "dapps" "d0"
 WHERE "d0"."ctid" < "d1"."ctid"
 	AND "d0"."transactionId" = "d1"."transactionId";
 
+-- Drop existing index on dapps table --
+DROP INDEX "dapps_trs_id";
+
 -- Add unique constraint on dapps table --
 ALTER TABLE "dapps"
 	ADD UNIQUE ("transactionId");
--- a/db/sql/migrations/updates/20180829161959_remove_duplicate_rows_and_add_unique_constraint_in_dapps.sql
+++ b/db/sql/migrations/updates/20180829161959_remove_duplicate_rows_and_add_unique_constraint_in_dapps.sql
@@ -12,12 +12,11 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
-BEGIN;
-	DELETE FROM "multisignatures" "m0"
-		USING "multisignatures" "m1"
-	WHERE "m0"."ctid" < "m1"."ctid"
-		AND "m0"."transactionId" = "m1"."transactionId";
+DELETE FROM "dapps" "d0"
+	USING "dapps" "d1"
+WHERE "d0"."ctid" < "d1"."ctid"
+	AND "d0"."transactionId" = "d1"."transactionId";
 
-	ALTER TABLE "multisignatures"
-		ADD PRIMARY KEY ("transactionId");
-COMMIT;
+-- Add unique constraint on dapps table --
+ALTER TABLE "dapps"
+	ADD UNIQUE ("transactionId");
--- a/db/sql/migrations/updates/20180829162051_remove_duplicate_rows_and_add_unique_constraint_in_votes.sql
+++ b/db/sql/migrations/updates/20180829162051_remove_duplicate_rows_and_add_unique_constraint_in_votes.sql
@@ -12,13 +12,12 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
-BEGIN;
-	DELETE FROM "dapps" "d0"
-		USING "dapps" "d1"
-	WHERE "d0"."ctid" < "d1"."ctid"
-		AND "d0"."transactionId" = "d1"."transactionId";
+DELETE FROM "votes" "v0"
+	USING "votes" "v1"
+WHERE "v0"."ctid" < "v1"."ctid"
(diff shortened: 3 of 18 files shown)


db/sql/migrations/updates/20180829161959_remove_duplicate_rows_and_add_primary_key_in_dapps.sql
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+BEGIN;
+	DELETE FROM "dapps" "d0"
+		USING "dapps" "d1"
+	WHERE "d0"."ctid" < "d1"."ctid"
+		AND "d0"."transactionId" = "d1"."transactionId";
+
+	ALTER TABLE "dapps"
+		ADD PRIMARY KEY ("transactionId");
```
nazarhussain: I will never prefer to have foreign key to be a primary key as well. It's against database ACID property of database. 

Yes people have mixed opinion on the topic in general (not related to PostgreSQL), you can look for it as well.
https://www.quora.com/Can-I-have-the-same-field-as-Primary-Key-and-Foreign-Key-in-MySQL
https://stackoverflow.com/questions/10982992/is-it-fine-to-have-foreign-key-as-primary-key


So prefer to use stuff which have common ground. The actual problem was we had an index on that column 

https://github.com/LiskHQ/lisk/blob/8f7de3963eaa48a6e2d8b081de37921b85911a53/db/sql/migrations/updates/20160723182900_create_schema.sql#L167
We just need to make that index unique, and keep the foreign key constraint as it is.

nazarhussain:
@4miners You have better insight on PostgreSQL related stuff, so had requested your thoughts on this PR as well.

db/sql/migrations/updates/20180829161959_remove_duplicate_rows_and_add_primary_key_in_dapps.sql
SargeKhan: I agree with what you said. My thought processes was to ensure consistency, because `transfer` and `signatures` table had `transactionId` as primary key. But looking further, i found that other asset tables like `delegates` and `outtransfer` and `intransfer` use unique index. I'll make the index unique.

4miners: I would prefer to use a unique constraint.
https://www.postgresql.org/docs/9.6/static/ddl-constraints.html

`
Note: The preferred way to add a unique constraint to a table is ALTER TABLE ... ADD CONSTRAINT. The use of indexes to enforce unique constraints could be considered an implementation detail that should not be accessed directly. One should, however, be aware that there's no need to manually create indexes on unique columns; doing so would just duplicate the automatically-created index.
`

db/sql/migrations/updates/20180829161959_remove_duplicate_rows_and_add_primary_key_in_dapps.sql
```diff
@@ -0,0 +1,24 @@
...
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+BEGIN;
+	DELETE FROM "dapps" "d0"
+		USING "dapps" "d1"
+	WHERE "d0"."ctid" < "d1"."ctid"
+		AND "d0"."transactionId" = "d1"."transactionId";
```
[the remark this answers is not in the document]
nazarhussain: As discussed because we don't have unique identifier in these tables, so keep the current implementation.

db/sql/migrations/updates/20180829161959_remove_duplicate_rows_and_add_unique_constraint_in_dapps.sql
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+DELETE FROM "dapps" "d0"
+	USING "dapps" "d1"
+WHERE "d0"."ctid" < "d1"."ctid"
+	AND "d0"."transactionId" = "d1"."transactionId";
+
+-- Add unique constraint on dapps table --
+ALTER TABLE "dapps"
+	ADD UNIQUE ("transactionId");
```
4miners: Add `DROP INDEX dapps_trs_id;`, as we don't need this index anymore.

db/sql/migrations/updates/20180829162051_remove_duplicate_rows_and_add_unique_constraint_in_votes.sql
```diff
@@ -0,0 +1,23 @@
...
+
+DELETE FROM "votes" "v0"
+	USING "votes" "v1"
+WHERE "v0"."ctid" < "v1"."ctid"
+	AND "v0"."transactionId" = "v1"."transactionId";
+
+-- Add unique constraint on votes table --
+ALTER TABLE "votes"
+	ADD UNIQUE ("transactionId");
+
```
4miners: `DROP INDEX votes_trs_id;`

db/sql/migrations/updates/20180829162052_remove_duplicate_rows_and_add_unique_constraint_in_transfer.sql
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+DELETE FROM "transfer" "tf0"
+	USING "transfer" "tf1"
+WHERE "tf0"."ctid" < "tf1"."ctid"
+	AND "tf0"."transactionId" = "tf1"."transactionId";
+
+-- Add unique constraint on transfer table --
+ALTER TABLE "transfer"
+	ADD UNIQUE ("transactionId");
```
4miners: `DROP INDEX transfer_trs_id;`

db/sql/migrations/updates/20180830113940_remove_duplicate_rows_and_add_unique_constraint_in_multisignatures.sql
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+DELETE FROM "multisignatures" "m0"
+	USING "multisignatures" "m1"
+WHERE "m0"."ctid" < "m1"."ctid"
+	AND "m0"."transactionId" = "m1"."transactionId";
+	
+-- Add unique constraint on multisignatures table --
+ALTER TABLE "multisignatures"
+	ADD UNIQUE ("transactionId");
```
4miners: `DROP INDEX multisignatures_trs_id;`

PR #2375, opened 2018-09-05
Better regular expression for verifying vote in verify vote - Close #627

### LINKED ISSUE #627: Better regular expression for vote in logic/vote -> verifyVote function
In verifyVote function, we check the format of vote using `if (!/[-+]{1}[0-9a-z]{64}/.test(vote)) ` condition, we should update regular expression to only pass for hexadecimal strings.

SargeKhan:
### What was the problem?
Vote regular expression was passing on non-hex vote string.
### How did I fix it?
Updated the regular expression to fail on non-hex vote string.
### How to test it?
Run the unit tests
### Review checklist

* The PR resolves #627 
* Linting passes
* Tests pass

### DIFF
--- a/logic/vote.js
+++ b/logic/vote.js
@@ -227,12 +227,8 @@ Vote.prototype.verifyVote = function(vote, cb) {
 		return setImmediate(cb, 'Invalid vote type');
 	}
 
-	if (!/[-+]{1}[0-9a-z]{64}/.test(vote)) {
+	if (!/^[-+]{1}[0-9a-f]{64}$/.test(vote)) {
 		return setImmediate(cb, 'Invalid vote format');
-	}
-
-	if (vote.length !== 65) {
-		return setImmediate(cb, 'Invalid vote length');
 	}
 
 	return setImmediate(cb);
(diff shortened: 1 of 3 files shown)


logic/vote.js
```diff
@@ -227,14 +227,10 @@ Vote.prototype.verifyVote = function(vote, cb) {
 		return setImmediate(cb, 'Invalid vote type');
 	}
 
-	if (!/[-+]{1}[0-9a-z]{64}/.test(vote)) {
+	if (!/^[-+]{1}[0-9a-f]{64}$/.test(vote)) {
```
willclarktech: `{1}` is redundant here

SargeKhan: I think it makes the regex more descriptive. What do you think?

test/unit/logic/vote.js
```diff
@@ -378,7 +378,7 @@ describe('vote', () => {
 	});
 
 	describe('verifyVote', () => {
-		it('should throw if vote is of invalid length', done => {
+		it('should return error if vote is of invalid length', done => {
```
ManuGowda: The description and expected error doesn't match, it should be `should return error if vote is of invalid format`

SargeKhan: the test description doesn't say what the error will be. The test description specifies what the kind of input we pass to the function.

PR #2376, opened 2018-09-05
Add sort support for username field on api/voters - Closes #2330

### LINKED ISSUE #2330: bad sorting when using balance:asc

yatki:
### What was the problem?
The fix we provided with #2346 was not backwards compatible. We had to keep username sort support.
### How did I fix it?
I added sort support for username field as well.
### Review checklist

* The PR resolves #2330 
* Linting passes
* Tests pass

### DIFF
--- a/db/sql/voters/get_voters.sql
+++ b/db/sql/voters/get_voters.sql
@@ -19,7 +19,7 @@
   PARAMETERS: ?
 */
 
-SELECT accounts.address, accounts.balance, accounts.username, encode(accounts."publicKey", 'hex') AS "publicKey" FROM mem_accounts2delegates delegates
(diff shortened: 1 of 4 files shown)


4miners:
@ManuGowda We bumped it already in this release, so should be fine.

PR #2379, opened 2018-09-05
Expire transaction test is failing randomly - Closes #2377

### LINKED ISSUE #2377: Expire transaction test is failing randomly
### Expected behavior
Expire transaction test should consistently pass.
### Actual behavior
Expire transaction tests are failing on Jenkins and in local at times, as the transaction is getting expired 
### Steps to reproduce
Run the test for 100 times and it will fail randomly
`for i in {1..100}; do ./node_modules/.bin/mocha test/integration/transactions/expire_transactions.js; done`

ManuGowda:
### What was the problem?
Expire transaction test was failing randomly
### How did I fix it?
`addTransactionsAndForge` was not called with `0` timeout, so the transaction was not forged and the test was failing.
### How to test it?
on the local machine run the test 100 times to ensure it never fails
`for i in {1..100}; do ./node_modules/.bin/mocha test/integration/transactions/expire_transactions.js; done`
### Review checklist

* The PR resolves #2377 
* Linting passes
* Tests pass

### DIFF
--- a/test/integration/transactions/expire_transactions.js
+++ b/test/integration/transactions/expire_transactions.js
@@ -207,7 +207,7 @@ describe('expire transactions', () => {
 				memAccountBefore = account;
 				// Transfer balance to multi-signature account
 				// so that multi-signature account can be registered
-				return addTransactionsAndForgePromise(library, [transaction]);
+				return addTransactionsAndForgePromise(library, [transaction], 0);
 			});
 		});
 


test/integration/transactions/expire_transactions.js
```diff
@@ -207,7 +207,7 @@ describe('expire transactions', () => {
 				memAccountBefore = account;
 				// Transfer balance to multi-signature account
 				// so that multi-signature account can be registered
-				return addTransactionsAndForgePromise(library, [transaction]);
+				return addTransactionsAndForgePromise(library, [transaction], 0);
```
4miners: 💯

PR #2380, opened 2018-09-06
Disable totalAmount verification on Transaction - Closes #2319

### LINKED ISSUE #2319: Do not verify transactions amount against of "totalAmount"
### Expected behavior
Transactions amounts should not be verified against `constants.totalAmount`, which is equal to 100 mln LSK. For today the total Mainnet supply is ~124 mln LSK and the supply still raises due to planned inflation. It is possible for one account to accumulate > 100 mln LSK, especially on test networks.
### Actual behavior
Transactions amount is checked if it doesn't exceed 100 mln LSK:
https://github.com/LiskHQ/lisk/blob/development/logic/transaction.js#L639

### Steps to reproduce
Send transaction with an amount that exceeds `constants.totalAmount`

*

yatki:
### What was the problem?
We were validating the transaction amount agains totalAmount constant which is equal to 100 mln LSK. But It is possible for one account to accumulate > 100 mln LSK, especially on test networks.
### How did I fix it?
I disabled the conditional check for totalAmount in transaction logic. Since we already have a check against user balance, we didn't need that.
### Review checklist

* The PR resolves #2319 
* Linting passes
* Tests pass

### DIFF
--- a/test/unit/logic/transaction.js
+++ b/test/unit/logic/transaction.js
@@ -751,7 +751,7 @@ describe('transaction', () => {
 			});
 		});
 
-		it('should return error when transaction amount is invalid', done => {
+		it('should not return "Invalid transaction amount" error when transaction amount is bigger than total amount of Lisk tokens', done => {
 			// lisk-elements cannot create a transaction with higher amount than max amount. So, this test needs to use hardcoded transaction object
 			var transaction = {
 				type: 0,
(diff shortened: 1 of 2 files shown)


SargeKhan:
@yatki can you check what would happen in case a transaction is created with the total supply. One possible way to do it would be to send all the money genesis account has on devnet to an account. Also, can you check what is the maximum value Bignumber supports? And how the application will behave in case a transaction exceeds that value?

test/unit/logic/transaction.js
```diff
@@ -751,7 +751,7 @@ describe('transaction', () => {
 			});
 		});
 
-		it('should return error when transaction amount is invalid', done => {
+		it('should not return error when transaction amount is bigger than total amount of Lisk tokens', done => {
```
willclarktech: This is misleading since the test expects that there is an error.

yatki: @willclarktech improved it.

PR #2382, opened 2018-09-07
Use asynchronous childProcess commands in Network tests - Closes #2334

### LINKED ISSUE #2334: Use asynchronous childProcess commands in Network tests
### Expected behavior
Make `restartNode`, `stopNode`, `startNode` asynchronous. Switch `childProcess.execSync` to `childProcess.exec` and utilize the output to verify if the task had been successful or not. Refactor the function usages by removing waiting for actions to succeed in timeout.

`>1.0.0`

jondubois:
### What was the problem?

- Functions used by network tests to spawn nodes were synchronous.
- Timeouts were used to estimate when nodes would be ready (before running test cases).

### How did I fix it?

- Made test cases asynchronous.
- Wait for nodes to be ready before executing tests.

### How to test it?

`grunt mocha:default:network`

### Review checklist

* The PR resolves #2334
* Linting passes
* Tests pass


test/network/scenarios/p2p/peer_disconnect.js
```diff
@@ -116,23 +124,33 @@ module.exports = function(
...
-						done();
-					}, 10000);
+					console.info('Wait for nodes to be stopped');
+					Promise.all(peersPromises)
+						.then(() => {
+							done();
+						})
+						.catch(err => {
+							done(err.message);
+						});
```
nazarhussain: This would be much compact option.

```js
it('stop all the nodes in the network except node_0', () => {
    return Promise.map(new Array(TOTAL_PEERS), (value, index) => {
        return common.stopNode(`node_${index + 1}`); 
    });
});
```

test/network/scenarios/p2p/peer_disconnect.js
```diff
@@ -116,23 +124,33 @@ module.exports = function(
...
-						done();
-					}, 10000);
+					console.info('Wait for nodes to be started');
+					Promise.all(peersPromises)
+						.then(() => {
+							done();
+						})
+						.catch(err => {
+							done(err.message);
+						});
```
nazarhussain: Same here.

test/network/scenarios/p2p/peer_disconnect.js
```diff
@@ -54,10 +54,14 @@ module.exports = function(
...
-						done();
-					}, 2000);
+					common
+						.stopNode('node_1')
+						.then(() => {
+							done();
+						})
+						.catch(err => {
+							done(err.message);
+						});
```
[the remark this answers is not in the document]
jondubois: The suggested approach looks cleaner but it's more brittle; if the `stopNode()` function changes in the future, it could break stuff in an unexpected way. I think it's better to be explicit.

test/network/scenarios/common.js
```diff
@@ -17,6 +17,9 @@
 const childProcess = require('child_process');
 const utils = require('../utils');
 
+const NODE_READY_REGEX = /Finished sync/;
```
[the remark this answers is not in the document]
jondubois: If I only wait for 'Blockchain ready', it breaks the test.

test/network/scenarios/common.js
```diff
@@ -78,13 +81,53 @@ module.exports = {
...
+				});
+				pm2LogProcess.stdout.on('data', data => {
+					const dataString = data.toString();
+					if (NODE_READY_REGEX.test(dataString)) {
+						clearTimeout(nodeReadyTimeout);
+						pm2LogProcess.stdout.removeAllListeners('error');
+						pm2LogProcess.stdout.removeAllListeners('data');
+						resolve();
+					}
+				});
```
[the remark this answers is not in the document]
jondubois: That feature would be perfect but I can't get it to work for some reason.

test/network/scenarios/common.js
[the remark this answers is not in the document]
nazarhussain: The name of regular expression you are using is `NODE_READY_REGEX ` https://github.com/LiskHQ/lisk/blob/ed88a37008d92b82260d86db956ec543c5c71eca/test/network/scenarios/common.js#L20

Which clearly is not the case, you are not waiting for the node to be ready, you are waiting for node to be synced at least once. If that's the case please use proper naming for the variables.

test/network/scenarios/p2p/peer_disconnect.js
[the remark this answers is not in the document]
nazarhussain: If `stopNode` change in the future we should should its usage. So why not use the cleaner approach which fits in current circumstances.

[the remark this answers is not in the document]
jondubois: A function should be a blackbox as much as possible. The `stopNode` function shouldn't have to worry about how its output is being used by dependent functions (`done` in this case). That's why the extra fat-arrow functions are important in this case. It's also more readable because looking at the test you can see exactly what goes into the `done` function without having to dig into the code.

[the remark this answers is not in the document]
nazarhussain: There is nothing to do with internal functionality of the function. 


Yes function should be blackbox but also follow the documented interface. If its documented that the function will return a Promise then it should always return promise. And all relevant code should be written in that perspective that it is returning the Promise. 

And once function return type is change, its reference usage must be updated accordingly.

test/network/scenarios/common.js
```diff
@@ -17,6 +17,9 @@
 const childProcess = require('child_process');
 const utils = require('../utils');
 
+const NODE_READY_REGEX = /Finished sync/;
+const NODE_READY_TIMEOUT = 20000;
```
ManuGowda: the disadvantage of having the timeout as a way to check if the application starts within this time frame is that it will be indeterministic if the performance of the application degrades then the tests will fail and if the performance improves then still it has to wait 20 seconds, as I said before if we subscribe to `blockchainReady` event then regardless of the time the tests will be consistent.

test/network/scenarios/common.js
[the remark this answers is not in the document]
jondubois: >> Which clearly is not the case

It's not so black and white. 'Node ready' can mean a lot of things depending on whether we're looking at it from the perspective of the node or from the perspective of the test.

I'll rename to `NODE_FINISHED_SYNC_REGEX` to avoid any confusion.

test/network/scenarios/p2p/peer_disconnect.js
nazarhussain: Just remove the `+1` from the index.

jondubois: But then it stops `node_0` which we don't want.

test/network/scenarios/common.js
jondubois: The timeout specifies the upper bound; there needs to be a timeout in case the node is taking way too long to launch; then it should fail the test with an error. It should be a large value though. I can make it higher. 20s is a bit short.

test/network/scenarios/p2p/peer_disconnect.js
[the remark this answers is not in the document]
jondubois: I'll do it since most people seem to disagree with me :(

test/network/scenarios/common.js
```diff
@@ -78,13 +81,53 @@ module.exports = {
...
+				const nodeReadyTimeout = setTimeout(() => {
+					pm2LogProcess.stdout.removeAllListeners('data');
+					pm2LogProcess.removeAllListeners('error');
+					reject(new Error(`Failed to start node ${nodeName} before timeout`));
+				}, NODE_READY_TIMEOUT);
+
+				let pm2LogProcess = childProcess.spawn('pm2', ['logs', nodeName]);
+				pm2LogProcess.once('error', err => {
+					clearTimeout(nodeReadyTimeout);
+					pm2LogProcess.stdout.removeAllListeners('data');
```
SargeKhan: why not remove error listener here as well?

jondubois: It uses the `once('error', ...)` handler so if the handler executes once then it will already have removed itself implicitly. I can change it to `on('error', ...)` and then explicitly remove the 'error' event handler if you prefer.

jondubois: Also note that I moved up the `let pm2LogProcess = ...` declaration since that's more clear. Also I made it `const`.

test/network/scenarios/common.js
```diff
@@ -17,6 +17,9 @@
 const childProcess = require('child_process');
 const utils = require('../utils');
 
+const NODE_FINISHED_SYNC_REGEX = /Finished sync/;
```
SargeKhan: Why are we using `Finished sync` as a regular expression?

jondubois: I like doing `regex.test(...)`  instead of `string.indexOf(...) !== ...` it doesn't matter for me.

test/network/scenarios/common.js
jondubois: I made it 40 seconds.

jondubois: I agree that it's not the best but we're running the node with PM2 so we can't listen to the node's events directly. I investigated using the HTTP API to check the sync status of the node but that would add a lot of complexity because of how we're running the nodes, it's difficult to associate them with a URL.

test/network/scenarios/common.js
jondubois: I added a comment in the code.

test/network/scenarios/p2p/peer_disconnect.js
```diff
@@ -54,10 +54,10 @@ module.exports = function(
...
 			describe('when a node is stopped', () => {
 				before(done => {
-					common.stopNode('node_1');
-					setTimeout(() => {
-						done();
-					}, 2000);
+					common
+						.stopNode('node_1')
+						.then(done)
+						.catch(done);
```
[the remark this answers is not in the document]
jondubois: We were running the network tests with the mocha `--bail` flag which meant that our `after()` hook was skipped and so the test suite would not cleanup after itself.

This is an open issue with mocha: https://github.com/mochajs/mocha/issues/3398

jondubois:
The purpose of the `--bail` flag in mocha is to terminate the test suite as soon as any one of the test cases fails; this feature saves time and it would have been nice to keep it but none of the alternative solutions I investigated could be used.

- Using `process.on('exit', cleanupPm2Processes)` doesn't work because we need to launch the `pm2 kill` command in a separate process; by the time the 'exit' event triggers, the main process is already shutting down and because child processes are killed automatically along with the parent process, it doesn't give enough time for the `pm2 kill` command to execute so it doesn't work.
- Using `process.on('SIGTERM', cleanupPm2Processes)` also doesn't work; same reason as above.

The best way to detect if the tests have finished is to use Mocha's `after` hook but currently this is not possible if we use the `--bail` flag.

PR #2387, opened 2018-09-12
Use two different DNS domains for seed nodes - Closes #2347

### LINKED ISSUE #2347: Use two different DNS domains for seed nodes in config.json
### Expected behavior

Use two DNS domains from different TLDs for seed nodes, for increased reliability.
This is a followup of #2320 

### Actual behavior

Only one DNS domain is being used at the moment.

### Steps to reproduce

Core >= 1.2.0

ghost:
### What was the problem?

Host names for seeds nodes were all under the same domain name (SPOF) and might lead to decreased reliability (e.g. TLD outage)

### How did I fix it?

Use two DNS domains from different TLDs.

### How to test it?

Make sure hostnames (FQDN) for all peers under `config/*/config.json` resolve properly to an IP.

### Review checklist

* The PR resolves #2347 
* Linting passes
* Tests pass

### DIFF
--- a/config/betanet/config.json
+++ b/config/betanet/config.json
@@ -13,7 +13,7 @@
 				"wsPort": 5001
 			},
 			{
-				"ip": "betanet-seed-02.lisk.io",
+				"ip": "betanet-seed-02.lisk-nodes.net",
 				"wsPort": 5001
 			},
 			{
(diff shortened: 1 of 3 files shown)


diego-G (changes requested):
Why we don't do the same for Alphanet ? https://github.com/LiskHQ/lisk/blob/b92095a9d7aee774b56e4c0ec0eebab207f4e953/config/alphanet/config.json#L18

fchavant:
@diego-G because alphanet defaults to using localhost (`127.0.0.1`) as only peer. No need for a DNS name here.

diego-G (approved):

MaciejBaj (approved):

jondubois (approved):

Issue #1780, opened 2018-03-23
Develop a strategy and related toolset to benchmark database layer


nazarhussain:
We have to define a strategy/guideline/steps to benchmark current database layer. Most of our database interactions already been abstracted to individual SQL files, so we need to develop some toolset that helps us to monitor the performance of those SQLs. 

e.g. 

1. We define a threshold time for execution of any SQL, e.g. 100ms, if any SQL takes more time that should be identified 
2. During some certain set of operations, we should be aware that which SQL/QueryFile is executed most of the time. 

The outcome of that task would be: 

1. Some document to defining that strategy 
2. Code base / toolset to implement that strategy 
3. Measurable outcome from that toolset / code base for future reference 

Note: 

1. Here we are not discussing the database profiling on operating system level, we want similar on application level 
2. This task is initial step to further extend in stress testing for database layer 
3. This task is initial step for identification of bad performance queries for optimization process.

vitaly-t:
This one could be an entry point:

https://github.com/LiskHQ/lisk/blob/4ebaa55f97da632c2a72a86716715ef3239a062e/db/index.js#L40

That's why I added it there in the first place.

nazarhussain:
@vitaly-t 👍 I know this is the entry point we would be using. But I don't want to add boilerplate code to actual code base. So why I used word `toolset` here, I foresee some lib/tool outside our codebase to encapsulate every thing related to profiling/benchmarking, and hooked up only in development mode or on demand to production mode.

vitaly-t:
For that there's a huge variety of tools to analyze performance of PostgreSQL. Just pick one.

vitaly-t:
Interesting points to consider when analyzing how the database layer behaves underneath...

You can hook into the connection pool like this, and see what is going on:

```js
db.$pool.on('acquire', function () {
    // this.totalCount - number of connections currently consumed from the pool.
    // It will never exceed the maximum pool size set in the connection.

   // this.idleCount - number of idle connections in the pool

  // this.waitingCount - number of queries/tasks currently waiting for the next available connection
});
```

**Tips:**

* Big `waitingCount` will always be a bad sign, meaning the system is too busy in terms of the connectivity to be efficient.
* `totalCount` sitting at the maximum means the pool size is insufficient for the way it is used by the app.

MaciejBaj:
Should be done no sooner than #1623. Hopefully, the benchmarking of API can cover DB queries analysis because of following the REST standards.

[...]

PR #2396, opened 2018-09-14
API performance benchmarking via newrelic - Closes #1623, #1780

### LINKED ISSUE #1623: API performance benchmarking
Presently, the database API is indifferent as to the priority of calls being made against it, i.e. high-priority calls are mixed with low-priority calls.

By adding support for API priority, we will improve the database API availability, and make database access scale in a way that's weighted and predictable.

---

The first step is to identify priority within the database API - which ones require high availability, and which ones do not, according to:

* Business requirements for each API call, as for importance and frequency
* Relative percentage of time consumed by each API call, versus the frequency of calls

Identifying just 2 levels should be sufficient at first - as low-priority and high-priority ones.

Then we can introduce support for this within the code as explained below.

---

At the moment we create a single `db` object, setting its pool size as this: 

```
min: 10, max: 95
```

This in turn allocates a single connection pool `db.$pool` within the object. What we want is to have 2 connection pools instead, one for low-priority calls, and one for high-priority calls.

In order to achieve this, we create two `db` objects instead, with connection pools configured like this:

* High-priority `dbHigh` object: `min: 10, max: 80`
* Low-priority `dbLow` object: `min: 5, max: 15`

(this is just an approximation) Note that in reality, the value of `min` is utterly unimportant, only `max` is important.

So, our `./db` module would expose two `db` objects, and for every use of the database API, the app will use the `db` object that corresponds to the priority of that API call within the system.

---

Next step after this will be: Use a separate user/password account for each `db` object, as [PostgreSQL supports configuring load priority](https://wiki.postgresql.org/wiki/Priorities) based on the login account, down to memory and CPU usage.

nazarhussain:
### What was the problem?

There were no tool or strategy available to benchmark/measure the performance of different aspects of application. 

### How did I fix it?

Integrated newrelic with custom code instrumentation to monitor the application performance. 

### How to test it?

Set `NEW_RELIC_LICENSE_KEY` variable and start the app normally. Then check the numbers in your application under newrelic account.

### Review checklist

* The PR resolves #1623, #1780
* Linting passes
* Tests pass

### DIFF
--- a/app.js
+++ b/app.js
@@ -14,6 +14,10 @@
 
 'use strict';
 
+if (process.env.NEW_RELIC_LICENSE_KEY) {
+	require('./helpers/newrelic_lisk');
+}
+
 var path = require('path');
 var fs = require('fs');
 var d = require('domain').create();
--- a/app.js
+++ b/app.js
@@ -14,6 +14,18 @@
 
 'use strict';
 
+// eslint-disable-next-line import/order
+const AppConfig = require('./helpers/config.js');
+
+/**
+ * Default list of configuration options. Can be overridden by CLI.
+ *
+ * @memberof! app
+ * @default 'config.json'
+ */
+// eslint-disable-next-line import/order
+const appConfig = AppConfig(require('./package.json'));
+
 if (process.env.NEW_RELIC_LICENSE_KEY) {
 	require('./helpers/newrelic_lisk');
 }
@@ -28,7 +40,6 @@ var async = require('async');
 var Logger = require('./logger.js');
 var wsRPC = require('./api/ws/rpc/ws_rpc').wsRPC;
 var wsTransport = require('./api/ws/transport');
-var AppConfig = require('./helpers/config.js');
 var git = require('./helpers/git.js');
 var Sequence = require('./helpers/sequence.js');
 var httpApi = require('./helpers/http_api.js');
@@ -101,14 +112,6 @@ if (typeof gc !== 'undefined') {
 		gc(); // eslint-disable-line no-undef
 	}, 60000);
 }
-
-/**
- * Default list of configuration options. Can be overridden by CLI.
- *
- * @memberof! app
- * @default 'config.json'
- */
-var appConfig = AppConfig(require('./package.json'));
 
 // Global objects to be utilized under modules/helpers where scope is not accessible
 global.constants = appConfig.constants;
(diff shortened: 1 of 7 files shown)


MaciejBaj:
NewRelic instrumentation code will be placed in a separate repository under LiskHQ.

app.js
```diff
@@ -14,6 +14,22 @@
...
+/**
+ * Default list of configuration options. Can be overridden by CLI.
+ *
+ * @memberof! app
+ * @default 'config.json'
+ */
+// eslint-disable-next-line import/order
+const appConfig = AppConfig(require('./package.json'));
+
+if (process.env.NEW_RELIC_LICENSE_KEY) {
```
SargeKhan: it'd be nicer to move this condition to `./helpers/newrelic_list` file and only initialize new relic if the environment variable is set. There is no need to have a conditional import here.

package-lock.json
```diff
@@ -4,17 +4,44 @@
...
+			"optional": true,
+			"requires": {
+				"nan": "2.10.0"
+			}
+		},
 		"@sinonjs/formatio": {
 			"version": "2.0.0",
 			"resolved":
-				"https://registry.npmjs.org/@sinonjs/formatio/-/formatio-2.0.0.tgz",
+				"http://registry.npmjs.org/@sinonjs/formatio/-/formatio-2.0.0.tgz",
```
SargeKhan: do we know why its changged to http?

workers_controller.js
```diff
@@ -14,6 +14,10 @@
 
 'use strict';
 
+if (process.env.NEW_RELIC_LICENSE_KEY) {
```
SargeKhan: We can have this condition within the `./helpers/newrelic_list` file as well.

app.js
nazarhussain: In that case we have to do conditional `require('newrelic')` or have a file level conditional return. Because their pattern is as soon you require the file it initialized automatically. Initialization of newrelic means a big layer of wrappers on almost all native libraries. That I want to completely avoid. 

So I feel its the best way to keep this condition in that situation.

package-lock.json
nazarhussain: Don't really know how that happens. May be new `npm` did this.

workers_controller.js
nazarhussain: Explained earlier as well..

PR #2401, opened 2018-09-18
Use proper name formatting for all SQL queries - Closes #2383 

### LINKED ISSUE #2383: Use proper name formatting for all SQL queries
### Expected behavior

There is a formatting method available in `Pg-Promise` which helps to inject some `column` or `table` names into query. 

http://vitaly-t.github.io/pg-promise/formatting.html#.name

We need to make sure its used properly everywhere. 

### Actual behavior
Some where noticed that we missed to use it. 

https://github.com/LiskHQ/lisk/blob/beced0e590bdff2f433f982a6af6faf7c1a4d50e/db/sql/voters/get_voters.sql#L25

So instead it should be 

```
ORDER BY ${sortField:name} ${sortMethod:raw} 
```

1.1.x

nazarhussain:
### What was the problem?
We were using `:raw` format specifier in SQL on some places which can be avoided. 

### How did I fix it?
Looked through each SQL file and optimized the usage of `:raw` format specifier. 

### How to test it?
Run all test suite. 

### Review checklist

* The PR resolves #2383
* Linting passes
* Tests pass


test/unit/db/repos/blocks.js
```diff
@@ -576,7 +576,7 @@ describe('db', () => {
 		});
 
 		describe('getCommonBlock()', () => {
-			it('should call a function with SQL parameters', function*() {
+			it('should call a normal SQL file with parameters if previousBlock is not provided', function*() {
```
MaciejBaj: to follow the used patterns:
``
describe('when previousBlock = undefined')
   it('should call a "getCommonBlock" SQL file with passed parameters')
describe('when previousBlock = 12345')
   it('should call a "getCommonBlockByPreviousBlock" SQL file with passed parameters'
``

db/sql/blocks/get_common_block_by_previous_block.sql
```diff
@@ -0,0 +1,27 @@
...
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+
+/*
+  DESCRIPTION: ?
+
+  PARAMETERS: ?
```
MaciejBaj: Can we specify Query parameters and their types?

db/sql/blocks/get_common_block_by_previous_block.sql
```diff
@@ -0,0 +1,27 @@
...
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+
+/*
+  DESCRIPTION: ?
```
MaciejBaj: Description may be removed. The title is descriptive enough.

db/sql/blocks/get_common_block_by_previous_block.sql
nazarhussain: There is no standard like jsdoc, its just a comment we can specify what ever we want.

db/sql/blocks/get_common_block_by_previous_block.sql
nazarhussain: Its consistent across all files. Should we remove from all files?

nazarhussain: @MaciejBaj Should we do it in this PR? or we update comments in SQL files for all files in other PR?

db/repos/blocks.js
```diff
@@ -193,13 +193,12 @@ class BlocksRepository {
...
 	 */
 	getCommonBlock(params) {
-		const query = values => {
-			values.comparePreviousBlock = values.previousBlock
-				? this.pgp.as.format('AND "previousBlock" = ${previousBlock}', values)
-				: '';
-			return sql.getCommonBlock;
-		};
-		return this.db.any(query, params);
+		return this.db.any(
```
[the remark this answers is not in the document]
4miners: I think it was originally for genesis block case (because genesis block is the only one that doesn't have `previousBlock` id), but later we added a condition for genesis block here:
https://github.com/LiskHQ/lisk/blob/059935f5befd1529dbb2d69f4383dd384d167e6a/modules/blocks/process.js#L298-L304

PR #2402, opened 2018-09-18
Add support for running standalone tests - Closes #955

### LINKED ISSUE #955: Fix unstable cache tests
### Expected behavior
Application should cache response for `GET /api/delegates` endpoint. There is a test case for this functionality [here](https://github.com/LiskHQ/lisk/blob/development/test/functional/http/get/delegates.js#L26).

### Actual behavior
It sometimes does not cache response in Redis, or the Redis deletes the key. (And the test fails) 
`/api/delegates/` endpoint caches responses, but only sometimes.

### Steps to reproduce:
Run tests on Jenkins.

yatki:
### What was the problem?
Since we were running all the tests in parallel redis checks in cache.js and blocks.js were conflicting. Tests in cache.js have to be run in syncronous way.

### How did I fix it?
I improved the parallel_test.js file to support running standalone tests.

### Review checklist

* The PR resolves #955 
* Linting passes
* Tests pass


test/common/parallel_tests.js
```diff
@@ -57,101 +77,168 @@ function parallelTests(tag, suite, section) {
...
+			let i = 0, limit = Math.min(allFilesLength, maxParallelism);
+			i < limit;
+			i += 1
+		) {
+			next();
+		}
+	});
+};
+
+const runStandaloneTests = (suiteFolder, mochaArguments) => {
```
nazarhussain: To me `runSequentialTests` makes more sense to explain the behavior. Point is we can run any test (specified by any tag) sequential .

test/mocha.opts
```diff
@@ -2,3 +2,4 @@
 --exit
 --colors
 --require ./test/setup.js
+--recursive
```
nazarhussain: 👍

test/common/parallel_tests.js
```diff
@@ -57,101 +77,168 @@ function parallelTests(tag, suite, section) {
...
+			}
+		}
+	});
 }
 
-parallelTests(process.argv[2], process.argv[3], process.argv[4]);
+(async () => {
+	await executeTests(process.argv[2], process.argv[3], process.argv[4]);
+	if (process.argv[2] !== 'standalone') {
+		await executeTests('standalone', process.argv[3], process.argv[4]);
```
SargeKhan: In case of extensive, we will run `standalone` tests twice?

.eslintrc.json
```diff
@@ -53,7 +53,7 @@
 			}
 		],
 		"strict": "off",
-		"space-before-function-paren": ["error", "never"],
+		"space-before-function-paren": "off",
```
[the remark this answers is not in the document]
yatki: @nazarhussain It's conflicting with prettier. We already have lots of overrides in eslintrc.json file I would say let's keep it like this for the moment. I believe since we moved to node 8, we can refactor lots of eslint rules in near future.

test/common/parallel_tests.js
```diff
@@ -57,101 +77,168 @@ function parallelTests(tag, suite, section) {
...
-				return;
+				console.info('All standalone tests finished successfully.');
+				return resolve();
 			}
 
-			console.info('Test failed:', test);
-			cleanupRunningTests();
-			process.exit(code);
+			console.error('Standalone tests failed:', suiteFolder);
+			child.kill('SIGTERM');
```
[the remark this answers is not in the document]
yatki: good point. I actually conserved that line from original code. Didn't notice that. I checked the node documentation. I think it's safe to remove it.

test/common/parallel_tests.js
yatki: We shouldn't.

```js
case 'extensive':
			mochaArguments.push('--', '--grep', '@unstable|@standalone', '--invert');
			break;
```

We are skipping standalone tests with `--invert` in extensive tests too.

Gruntfile.js
```diff
@@ -51,13 +51,20 @@ module.exports = function(grunt) {
 					if (suite === 'network') {
 						var filter = '';
 						if (tagFilter === 'default') {
-							filter = "--grep '@slow|@unstable' --invert";
+							filter = "--grep '@slow|@unstable|@standalone' --invert";
```
nazarhussain: Will this not be `sequential` instead of `standalone`?

PR #2409, opened 2018-09-20
Refactor constants usage to use deconstructors - Closes #2381

### LINKED ISSUE #2381: Prefer deconstruction when using Lisk Core constants
### Expected behavior
We should deconstruct constants object on top of modules like below:

```js
const {EPOC_TIME} = global.constants;
```
### Actual behavior
We are trying to access the constants via constants variable.

```js
const constants = global.constants;

// ...

console.log(constants.EPOCH_TIME)
```

See Discussion here: https://github.com/LiskHQ/lisk/pull/2371#discussion_r215214501
1.3.0+

yatki:
### Review checklist

* The PR resolves #2381 
* Linting passes
* Tests pass

### DIFF
--- a/helpers/config.js
+++ b/helpers/config.js
@@ -286,9 +286,8 @@ function cleanDeep(
 function validateForce({ constants, forging, nethash }) {
 	if (forging.force) {
 		const { NETHASHES } = constants;
-		var index = NETHASHES.indexOf(nethash);
 
-		if (index !== -1) {
+		if (NETHASHES.indexOf(nethash) !== -1) {
 			console.info('Forced forging disabled for nethash', nethash);
 			forging.force = false;
 		}
--- a/helpers/config.js
+++ b/helpers/config.js
@@ -283,13 +283,14 @@ function cleanDeep(
  * @param {Object} configData
  * @todo Add description for the params
  */
-function validateForce(configData) {
-	if (configData.forging.force) {
-		var index = configData.constants.NETHASHES.indexOf(configData.nethash);
+function validateForce({ constants, forging, nethash }) {
+	if (forging.force) {
+		const { NETHASHES } = constants;
+		var index = NETHASHES.indexOf(nethash);
 
 		if (index !== -1) {
-			console.info('Forced forging disabled for nethash', configData.nethash);
-			configData.forging.force = false;
+			console.info('Forced forging disabled for nethash', nethash);
+			forging.force = false;
 		}
 	}
 }
--- a/helpers/slots.js
+++ b/helpers/slots.js
@@ -38,10 +38,7 @@ function getEpochTime(time) {
 		time = Date.now();
 	}
 
-	var d = EPOCH_TIME;
-	var t = d.getTime();
-
-	return Math.floor((time - t) / 1000);
+	return Math.floor((time - EPOCH_TIME.getTime()) / 1000);
 }
 
 module.exports = {
@@ -71,8 +68,7 @@ module.exports = {
 			epochTime = this.getTime();
 		}
 
-		var d = EPOCH_TIME;
-		var t = Math.floor(d.getTime() / 1000) * 1000;
+		var t = Math.floor(EPOCH_TIME.getTime() / 1000) * 1000;
 
 		return t + epochTime * 1000;
 	},
--- a/helpers/slots.js
+++ b/helpers/slots.js
@@ -14,7 +14,7 @@
 
 'use strict';
 
-const constants = global.constants;
+const { EPOCH_TIME, ACTIVE_DELEGATES } = global.constants;
 
 /**
  * Description of the module.
@@ -38,7 +38,7 @@ function getEpochTime(time) {
 		time = Date.now();
 	}
 
-	var d = constants.EPOCH_TIME;
+	var d = EPOCH_TIME;
 	var t = d.getTime();
 
 	return Math.floor((time - t) / 1000);
@@ -46,7 +46,7 @@ function getEpochTime(time) {
 
 module.exports = {
 	interval: 10,
-	delegates: constants.ACTIVE_DELEGATES,
+	delegates: ACTIVE_DELEGATES,
 
 	/**
 	 * Description of the function.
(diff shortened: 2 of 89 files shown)


logic/account.js
```diff
@@ -306,7 +306,7 @@ class Account {
 			}
 		});
 
-		let limit = constants.ACTIVE_DELEGATES;
+		let limit = ACTIVE_DELEGATES;
```
ManuGowda: I don't see any usage of limit in this file, can you remove it if its not being used.

yatki: @ManuGowda it is being used [here](https://github.com/LiskHQ/lisk/blob/2381-prefer-deconstruction/logic/account.js#L344)

helpers/slots.js
```diff
@@ -38,15 +38,12 @@ function getEpochTime(time) {
...
-	var t = d.getTime();
-
-	return Math.floor((time - t) / 1000);
+	return Math.floor((time - EPOCH_TIME.getTime()) / 1000);
 }
 
 module.exports = {
 	interval: 10,
-	delegates: constants.ACTIVE_DELEGATES,
+	delegates: ACTIVE_DELEGATES,
```
4miners: We shouldn't export `ACTIVE_DELEGATES` as `delegates` here and use `ACTIVE_DELEGATES` in corresponding modules that uses `slots.delegates`, can we fix this?

helpers/config.js
```diff
@@ -283,13 +283,14 @@ function cleanDeep(
  * @param {Object} configData
  * @todo Add description for the params
  */
-function validateForce(configData) {
-	if (configData.forging.force) {
-		var index = configData.constants.NETHASHES.indexOf(configData.nethash);
+function validateForce({ constants, forging, nethash }) {
+	if (forging.force) {
+		const { NETHASHES } = constants;
+		var index = NETHASHES.indexOf(nethash);
```
4miners: Please use more meaningful variable name.

helpers/slots.js
```diff
@@ -71,8 +68,7 @@ module.exports = {
 			epochTime = this.getTime();
 		}
 
-		var d = constants.EPOCH_TIME;
-		var t = Math.floor(d.getTime() / 1000) * 1000;
+		var t = Math.floor(EPOCH_TIME.getTime() / 1000) * 1000;
```
4miners: Same here, would be nice to rename this variable.

logic/block_reward.js
```diff
@@ -33,13 +33,13 @@ const __private = {};
 class BlockReward {
 	constructor() {
 		// Array of milestones
-		this.milestones = constants.REWARDS.MILESTONES;
+		this.milestones = REWARDS.MILESTONES;
 
 		// Distance between each milestone
-		this.distance = Math.floor(constants.REWARDS.DISTANCE);
+		this.distance = Math.floor(REWARDS.DISTANCE);
```
4miners: Why we need to call `Math.floor` on this?

logic/block_reward.js
```diff
@@ -33,13 +33,13 @@ const __private = {};
...
 
 		// Start rewards at block (n)
-		this.rewardOffset = Math.floor(constants.REWARDS.OFFSET);
+		this.rewardOffset = Math.floor(REWARDS.OFFSET);
```
4miners: Same here.

logic/block_reward.js
```diff
@@ -33,13 +33,13 @@ const __private = {};
 class BlockReward {
 	constructor() {
 		// Array of milestones
-		this.milestones = constants.REWARDS.MILESTONES;
+		this.milestones = REWARDS.MILESTONES;
```
4miners: Why do we need to assign those instead of using directly?

helpers/config.js
yatki: I removed the variable since we didn't need it.

helpers/slots.js
yatki: I removed the variable here as well

logic/block_reward.js
yatki: @4miners I can't comment on this. Is there any chance that rewards constants can be float?

logic/block_reward.js
yatki: Removed: `this.milestones`.

logic/block_reward.js
yatki: @ManuGowda can you comment on this ?

4miners: They cannot.

yatki: @4miners then having `this.distance` variable doesn't make sense either.

4miners: Yep, I was referencing to all of them.

logic/account.js
ManuGowda: ok, can change let to const its not being reassigned.

logic/block_reward.js
ManuGowda: you can directly use `REWARDS.DISTANCE` directly. I don't know the intention behind `floor` logic here, but as Mariusz said rewards will not be a float.

yatki: @4miners Let's handle this in a separate issue. see: https://github.com/LiskHQ/lisk/issues/2412

logic/block_reward.js
yatki: same above.

helpers/slots.js
yatki: this requires a big refactor. Let's do it in a separate issue. See: https://github.com/LiskHQ/lisk/issues/2413

yatki: @4miners see my comment above.

4miners: Ahh, ok, sorry. 🤗

logic/account.js
ManuGowda: @yatki did you address the limit to `const`?

yatki: @ManuGowda it is being reassigned [here](https://github.com/LiskHQ/lisk/blob/2381-prefer-deconstruction/logic/account.js#L319)

PR #2418, opened 2018-09-24
Update package jsons - Closes #2411

### LINKED ISSUE #2411: Re-use package.lock.json to point to public sodium-native repository
### Expected behavior
As the package-lock.json was introduced together with the migration to npm v5, we don't have to point to the internal LiskHQ/sodium-native version.
### Actual behavior
LiskHQ forked `sodium-native` repository and froze its dependencies, which doesn't have to happen since package-lock.json is present.
`>=1.1.0`

shuse2:
### What was the problem?
We are having `package-lock.json` from 1.2.0, so we don't need to reference our fork version of `sodium-native` anymore.

### How did I fix it?
Change to npm version of sodium native in the `package.json` and update `package-lock.json`.

### Review checklist

* The PR resolves #2411
* Linting passes
* Tests pass

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -419,6 +419,15 @@
 			"resolved": "https://registry.npmjs.org/aws4/-/aws4-1.7.0.tgz",
 			"integrity": "sha1-1NDpudv8p3vwjusKikcVUP454ok="
 		},
+		"axios": {
+			"version": "0.18.0",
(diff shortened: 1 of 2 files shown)


shuse2:
After discussion, for cryptographic packages, we are not using `npm` package.

PR #2431, opened 2018-10-01
Wait for consensus to be achieved in Peers Network tests - Closes #2332

### LINKED ISSUE #2332: Wait for consensus to be achieved in Peers Network tests
### Expected behavior
Make the forging test strict, relying on waiting for a block only after the consensus has been achieved.
### Actual behaviour
Right now the expectation is to wait for 30 sec timeout:
> // Expect some blocks to be forged after 30 seconds
`>1.0.0`

jondubois:
### What was the problem?

- Network test suite was using timeouts (in multiple places) instead of checking node statuses before running assertions.
- The logic to control individual nodes in the network was separate from the network launch logic. Related configuration objects (e.g. pm2Config objects and node configuration objects) were kept in different parts of the code; this made it difficult to perform operations on individual nodes (like restarting a specific node, waiting for a specific node to be ready or waiting for a specific node to forge some blocks).

### How did I fix it?

- Attached the original node configuration object to the pm2Config object so that we can interact with the same node via either pm2 or through its HTTP and WS API.
- Removed common.js file and moved its logic into a new network.js object; this new network object contains all the functionality necessary for managing our test network; it exposes methods to target and control individual nodes within the network.
- Async methods in the new network object use Promises instead of callbacks.
- The main index.js file which bootstraps the network tests now creates the main Network instance and then passes it to individual test scripts so that they can run their own scenarios against it.
- Skipped 2 test cases related to networkHeight; there appears to be a significant delay in updating the networkHeight for some nodes (even though the actual height of all nodes in the network match and all nodes are mutually connected). We should raise a separate issue to investigate why the networkHeight is taking so long to update.

### How to test it?

```grunt mocha:default:network```

### Review checklist

* The PR resolves #2332 
* Linting passes
* Tests pass

### CODE REVIEW

File: test/common/utils/wait_for.js [hunk 1]
```diff
@@ -76,8 +80,8 @@ function blockchainReady(cb, retries, timeout, baseUrl) {
 	})();
 }
 
-function nodeStatus(cb) {
-	var request = popsicle.get(`${__testContext.baseUrl}/api/node/status`);
+function nodeStatus(cb, baseUrl) {
```
nazarhussain: We already done this mistake previously. I would be nice to not repeat it here. The `cb` should be last argument of the function as the best practices for NodeJS.

nazarhussain: Please make this change in every where introduced a parameter after the callback.

test/network/index.js
```diff
@@ -26,78 +27,95 @@ const NUMBER_OF_TRANSACTIONS = process.env.NUMBER_OF_TRANSACTIONS || 1000;
...
+		failedToLaunchNetworkError = err;
+		initTestSuite();
+		run();
 	});
 
-	after(done => {
-		setup.exit(() => {
-			done(testFailedError);
+function initTestSuite() {
+	describe(`Start a network of ${TOTAL_PEERS} nodes with address "127.0.0.1", WS ports 500[0-9] and HTTP ports 400[0-9] using separate databases`, () => {
```
nazarhussain: I don't think its good practice to have tests in this pattern. 

Why can't you create `describe` at root level and put `network.launchNetwork({ enableForging: true })` in a `before` block?

test/network/index.js
```diff
@@ -26,78 +27,95 @@ const NUMBER_OF_TRANSACTIONS = process.env.NUMBER_OF_TRANSACTIONS || 1000;
...
 
-			if (numOfConnections === TOTAL_PEERS) {
-				done();
-			} else {
-				done(
-					`There are ${numOfConnections} listening connections on web socket ports.`
-				);
-			}
+		after(() => {
+			return network.killNetwork();
```
nazarhussain: Its interesting. You added `network.killNetwork();` in `after` block but `launchNetwork` is not in the `before` block. 

I can't understand the difference here. You can add `killNetwork` in the top level promise chain if you follow same pattern, which I wouldn't prefer.

nazarhussain:
The issue described is 

> Make the forging test strict, relying on waiting for a block only after the consensus has been achieved.

I feel the changes in that PR is much more than targeting the issue. Shouldn't we do this refactoring in small chunks?

test/common/utils/wait_for.js
jondubois: I agree that having the callback as the last argument is the ideal approach. It was done like this to maintain backwards-compatibility. Note that we are already using this incorrect pattern for the `blockchainReady` function so at least it's consistent with the rest of the `wait_for` module.

The `wait_for` module is used everywhere in all our tests (functional, integration, ...); refactoring the `wait_for` module and all its usages would be out of scope for this issue. We should raise a separate issue for it.

jondubois: Agreed, as a separate PR/issue. Also we should refactor the `blockchainReady` function which has the same problem.

jondubois:
@nazarhussain This PR is large but these two issues were tightly related.

I considered just patching the immediate issue without this big refactoring but it would have required too many hacks.

I made a change to `network/index` so that it launches the network in a `before` block instead of delaying the entire suite (as you suggested). This change means that each sub-test needs to individually check when the network is ready before they run,

jondubois:
@nazarhussain I made a change based on some earlier feedback you gave about the way the test suites were bootstrapped. I also left some replies on some of your other comments.

test/common/utils/wait_for.js
```diff
@@ -26,7 +26,7 @@ var apiHelpers = require('../helpers/api');
  * @param {number} [timeout=200] timeout
  * @param {string} [baseUrl='http://localhost:5000'] timeout
  */
-function blockchainReady(cb, retries, timeout, baseUrl) {
+function blockchainReady(cb, retries, timeout, baseUrl, doNotLogRetries) {
```
MaciejBaj: Thank you for "doNotLogRetries"!

test/network/scenarios/p2p/peers.js
```diff
@@ -72,53 +75,45 @@ module.exports = function(configurations) {
...
 
-			describe('network status after 30 seconds', () => {
-				let getNodesStatusError;
+			describe('network status after 3 blocks', () => {
+				let getAllNodesStatusError;
 				let networkHeight;
 				let networkAverageHeight;
-				let peersCount;
-				let peerStatusList;
+				// TODO: Uncomment when networkHeight issue has been fixed.
```
MaciejBaj: Could you please insert the issue reference here and everywhere where "networkHeight" issue is mentioned?

jondubois:
I've raised the issue about the `wait_for` module https://github.com/LiskHQ/lisk/issues/2449

PR #2432, opened 2018-10-01
Wrong amount check in verify function of logic/In_transfer and logic/out_transfer - Closes #2430

### LINKED ISSUE #2430: Wrong amount check in verify function of logic/In_transfer and logic/out_transfer
### Expected behavior
InTransfer and OutTransfer verify function should check if `amount.equals(0)`
## Actual behavior
InTransfer verify function checks if the transaction amount is not equals zero
https://github.com/LiskHQ/lisk/blob/ca8a69fdaa4d69ec729721a3b49b04172a337dff/logic/in_transfer.js#L96
OutTransfer verify function checks if the transaction amount is not equals zero
https://github.com/LiskHQ/lisk/blob/ca8a69fdaa4d69ec729721a3b49b04172a337dff/logic/out_transfer.js#L94 
1.1.0

nazarhussain:
### What was the problem?
InTransfer/OutTransfer transactions were using wrong logic to check the amount. 

### How did I fix it?
Reverted the logic to check the existence of amount attribute. 

### How to test it?
Run all functional tests. 

### Review checklist

* The PR resolves #2430
* Linting passes
* Tests pass


logic/out_transfer.js
```diff
@@ -89,9 +89,7 @@ OutTransfer.prototype.verify = function(transaction, sender, cb) {
 		return setImmediate(cb, 'Invalid recipient');
 	}
 
-	const amount = new Bignum(transaction.amount);
-
-	if (!amount.equals(0)) {
+	if (!transaction.amount) {
```
ManuGowda: @nazarhussain I think this will not work, `if (!transaction.amount) {` will always result in `false` as `new BigNum(0)` will be `true`. The condition should be `if (transaction.amount.equals(0)`. Also can you ensure once that `transaction.amount` is bignumber?

ManuGowda: reference for the implementation here https://runkit.com/5baa0981ae64c20012b8228b/5bb234611992d60012821384

nazarhussain: I believe `transaction.amount` at this context is not a big number. So just checking if `transaction` have an `amount` attribute is good enough. Its exactly the old logic, I just reverted it back.

ManuGowda: I think it's Bignumber, in logic transaction object normalize the amount will be converted to Bignumber, so before it goes to any transaction verify step it will be converted to Bignumber, here is the logic
https://github.com/LiskHQ/lisk/blob/ca8a69fdaa4d69ec729721a3b49b04172a337dff/logic/transaction.js#L1126

nazarhussain: If that's the case then why we were casting it to `bignum` in the `verify` method.

logic/in_transfer.js
```diff
@@ -92,8 +92,7 @@ InTransfer.prototype.verify = function(transaction, sender, cb, tx) {
 	}
 
 	const amount = new Bignum(transaction.amount);
```
4miners: We don't need to create `Bignum` instance here, as it's already a `Bignum`.

logic/in_transfer.js
```diff
@@ -92,8 +92,7 @@ InTransfer.prototype.verify = function(transaction, sender, cb, tx) {
...
-
-	if (!amount.equals(0)) {
+	if (amount.lessThanOrEqualTo(0)) {
```
4miners: We should only check for `0`, as negative value is checked here:
https://github.com/LiskHQ/lisk/blob/4ae8553042b93d3e7ce41c11a7d910fef24ba644/logic/transaction.js#L643-L649

nazarhussain: Its consistent  across other transaction types. 

https://github.com/LiskHQ/lisk/blob/4ae8553042b93d3e7ce41c11a7d910fef24ba644/logic/transfer.js#L85-L87

4miners: Do you think we should follow this "consistency"?

nazarhussain: @4miners Yes at least for `1.1.0`. We try to avoid more code change for this release. We can fix this logic across all transactions separately in next release.

4miners: Ok, I agree, but can you open an issue for this then?

ManuGowda: @nazarhussain what you eventually mean to say is that we will create a separate issue and fix the `consistency` in next release? if that is the case I have one concern, I remember Usman mentioning about not creating issues which are not worthy to be an issue to tackle separately, so if you guys agree to create a separate issue I am fine with it.

nazarhussain: I am also advocate of this approach to fix small stuff in between issues, but I was just worried about the time of release. But as discussed we will do a complete QA cycle afterwards, so I will make the relevant changes.

logic/vote.js
```diff
@@ -140,6 +140,11 @@ Vote.prototype.calculateFee = function() {
  * @todo Add description for the params
  */
 Vote.prototype.verify = function(transaction, sender, cb, tx) {
+	const amount = new Bignum(transaction.amount);
+	if (amount.greaterThan(0)) {
```
4miners: @nazarhussain In testnet we have a transaction that will fail this check, please make sure if it's in exceptions:
```
lisk_test=# select id, type, amount, "blockId" from trs where amount > 0 AND type > 0 AND type < 6;
          id          | type |  amount   |       blockId       
----------------------+------+-----------+---------------------
 15449731671927352923 |    3 | 100000000 | 2528992701257192568
(1 row)
```

logic/vote.js
```diff
@@ -140,61 +140,75 @@ Vote.prototype.calculateFee = function() {
...
-				return setImmediate(cb, err);
+					},
+					waterCb
+				);
+			},
+		],
+		waterErr => {
+			if (waterErr) {
+				if (exceptions.votes.includes(transaction.id)) {
+					library.logger.error(waterErr);
```
4miners: Please consider changing the log level to `warn`

logic/vote.js
```diff
@@ -140,61 +140,75 @@ Vote.prototype.calculateFee = function() {
...
-					if (err) {
-						return setImmediate(
-							eachSeriesCb,
-							[
-								'Invalid vote at index',
-								transaction.asset.votes.indexOf(vote),
-								'-',
-								err,
-							].join(' ')
+	async.waterfall(
```
4miners: What happened with the check for `amount`? 🤔

PR #2442, opened 2018-10-05
Add fix for vote collision scenario - Closes #2434

### LINKED ISSUE #2434: Change of vote processing logic ignore address collision scenario
### Expected behavior
The votes processing should be done against the `publicKeys` to avoid any address collision. Like the way it was implemented in the `1.0.0`

### Actual behavior
With some changes introduced in PR https://github.com/LiskHQ/lisk/pull/2189, we change the vote processing logic to be used on `address`. Which identifies that we have few address collisions on both networks. 

So votes casted for different public key result into same address (in collision case), and cause error `already casted votes. 

Few sample are to verify the issue on testnet.
```
b26dd40ba33e4785e49ddc4f106c0493ed00695817235c778f487aea5866400a, ce33db918b059a6e99c402963b42cf51c695068007ef01d8c383bb8a41270263
```

Few sample are to verify the issue on mainnet.
```
ce3d5baae39cce28d1fa8b41ab1ebac69473a1ad38d5caf9184d6c58c081447c, aaf8c168739eb5aa7e923bbdbe9c5a7a07c77a487629df8fe1da608df5f96cec
```

### Steps to reproduce
Sync form the network with `1.1.0`. 

`1.1.0`

SargeKhan:
### What was the problem?
Votes processing was done against addresses which are known to have collisions.
### How did I fix it?
Changed vote processing from using addresses to public keys.
### How to test it?
Run the tests.
### Review checklist

* The PR resolves #2434 
* Linting passes
* Tests pass

### DIFF
--- a/modules/delegates.js
+++ b/modules/delegates.js
@@ -360,25 +360,21 @@ __private.checkDelegates = function(senderPublicKey, votes, state, cb, tx) {
 		return setImmediate(cb, 'Votes must be an array');
 	}
 
-	let voteAddressesWithActions;
+	let votesWithAction;
 
 	try {
-		// Converting publicKeys to addresses because module.accounts do not support fetching multiple accounts by publicKeys
 		// TODO: Use Hashmap to improve performance further.
-		voteAddressesWithActions = votes.map(vote => {
+		votesWithAction = votes.map(vote => {
 			const action = vote[0];
 
 			if (action !== '+' && action !== '-') {
 				throw 'Invalid math operator';
 			}
 			const votePublicKey = vote.slice(1);
-			const voteAddress = modules.accounts.generateAddressByPublicKey(
-				votePublicKey
-			);
 
 			return {
 				action,
-				address: voteAddress,
+				publicKey: votePublicKey,
 			};
 		});
 	} catch (e) {
@@ -388,8 +384,8 @@ __private.checkDelegates = function(senderPublicKey, votes, state, cb, tx) {
 
 	async.waterfall(
 		[
-			// get all the addresses (converted from public keys) the sender has voted for. Confirmed or unconfirmed based on state parameter.
-			function getExistingVotedAddresses(waterfallCb) {
+			// get all  public keys of delegates sender has voted for. Confirmed or unconfirmed based on state parameter.
+			function getExistingVotedPublicKeys(waterfallCb) {
 				modules.accounts.getAccount(
 					{ publicKey: senderPublicKey },
 					(err, account) => {
@@ -403,22 +399,20 @@ __private.checkDelegates = function(senderPublicKey, votes, state, cb, tx) {
 
 						const delegates =
 							state === 'confirmed' ? account.delegates : account.u_delegates;
-						const votedAddresses = Array.isArray(delegates)
-							? delegates.map(delegate =>
-									modules.accounts.generateAddressByPublicKey(delegate)
-								)
+						const existingVotedPublicKeys = Array.isArray(delegates)
+							? delegates
 							: [];
 
-						return setImmediate(waterfallCb, null, votedAddresses);
+						return setImmediate(waterfallCb, null, existingVotedPublicKeys);
 					},
 					tx
 				);
 			},
 			// Validate votes in the transaction by checking that sender is not voting for an account already, and also that sender is not unvoting an account it did not vote before.
-			function validateVotes(existingVotedAddresses, waterfallCb) {
+			function validateVotes(existingVotedPublicKeys, waterfallCb) {
 				modules.accounts.getAccounts(
 					{
-						address: voteAddressesWithActions.map(({ address }) => address),
+						publicKey: votesWithAction.map(({ publicKey }) => publicKey),
 						isDelegate: 1,
 						sort: 'address:desc',
 					},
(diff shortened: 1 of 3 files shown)


modules/delegates.js
```diff
@@ -403,22 +399,20 @@ __private.checkDelegates = function(senderPublicKey, votes, state, cb, tx) {
 
 						const delegates =
 							state === 'confirmed' ? account.delegates : account.u_delegates;
-						const votedAddresses = Array.isArray(delegates)
-							? delegates.map(delegate =>
-									modules.accounts.generateAddressByPublicKey(delegate)
-								)
+						const existingVotedPublicKeys = Array.isArray(delegates)
+							? delegates
 							: [];
```
nazarhussain: We can simplify it: 

```js
const existingVotedPublicKeys = (state === 'confirmed' ? account.delegates : account.u_delegates) || [];
```

modules/delegates.js
```diff
@@ -429,69 +423,68 @@ __private.checkDelegates = function(senderPublicKey, votes, state, cb, tx) {
...
 						const upvoteAccounts = votesAccounts.filter(voteAccount =>
-							voteAddressesWithActions.find(
-								actionWithAddress =>
-									actionWithAddress.action === '+' &&
-									actionWithAddress.address === voteAccount.address
+							votesWithAction.find(
+								({ action, publicKey }) =>
+									action === '+' && publicKey === voteAccount.publicKey
 							)
 						);
```
nazarhussain: This will be much simpler. 

```js
const upVotePublicKeys  = votesWithAction.filter(({ action }) => action === '+').map(({publicKey}) => publicKey);
```

modules/delegates.js
```diff
@@ -429,69 +423,68 @@ __private.checkDelegates = function(senderPublicKey, votes, state, cb, tx) {
...
 						const downvoteAccounts = votesAccounts.filter(voteAccount =>
-							voteAddressesWithActions.find(
-								actionWithAddress =>
-									actionWithAddress.action === '-' &&
-									actionWithAddress.address === voteAccount.address
+							votesWithAction.find(
+								({ action, publicKey }) =>
+									action === '-' && publicKey === voteAccount.publicKey
 							)
 						);
```
nazarhussain: Same applies here. 


```js
const downVotePublicKeys  = votesWithAction.filter(({ action }) => action === '-').map(({publicKey}) => publicKey);
```

modules/delegates.js
```diff
@@ -429,69 +423,68 @@ __private.checkDelegates = function(senderPublicKey, votes, state, cb, tx) {
...
-								existingVotedAddresses.find(
-									existingVotedAddress =>
-										voteAccount.address === existingVotedAddress
-								) !== undefined
+						const invalidUpvoteAccounts = _.intersectionWith(
+							upvoteAccounts,
+							existingVotedPublicKeys,
+							({ publicKey: upvoteAccountPublicKey }, existingVotedPublicKey) =>
+								upvoteAccountPublicKey === existingVotedPublicKey
 						);
```
nazarhussain: Can be simplified as,

```js
const invalidUpvoteAccounts = _.intersect(upvotePublicKeys, existingVotedPublicKeys);
```

modules/delegates.js
```diff
@@ -429,69 +423,68 @@ __private.checkDelegates = function(senderPublicKey, votes, state, cb, tx) {
...
-									voteAccount.address !== existingVotedAddress
-							)
+						const invalidDownvoteAccounts = _.differenceWith(
+							downvoteAccounts,
+							existingVotedPublicKeys,
+							(
+								{ publicKey: downvoteAccountPubicKey },
+								existingVotedPublicKey
+							) => downvoteAccountPubicKey === existingVotedPublicKey
 						);
```
nazarhussain: Can be simplified as,

```js
const invalidDownVotes = _.difference(downVotePublicKeys, existingVotedPublicKeys);
```

test/functional/system/transactions/3.votes_collision.js
```diff
@@ -0,0 +1,157 @@
...
+		passphrase:
+			'merry field slogan sibling convince gold coffee town fold glad mix page',
+		publicKey:
+			'ce33db918b059a6e99c402963b42cf51c695068007ef01d8c383bb8a41270263',
+		collisionPassphrase:
+			'annual youth lift quote off olive uncle town chief poverty extend series',
+		collisionPublicKey:
+			'b26dd40ba33e4785e49ddc4f106c0493ed00695817235c778f487aea5866400a',
+		username: 'xyz',
+	};
```
nazarhussain: 👍

modules/delegates.js
SargeKhan: We need upvoteAccounts, not upvotePublicKeys because we need to return an error with the username of the account if the account has already voted for the delegate.

modules/delegates.js
SargeKhan: same as above

modules/delegates.js
SargeKhan: Doing this because I don't want to traverse the list again in case of invalidUpvoteAccount and invalidDownvoteAccount checks.

modules/delegates.js
SargeKhan: We need the username of the account.

PR #2456, opened 2018-10-09
Remove transactionFee exceptions from Betanet config -  Closes #2443

### LINKED ISSUE #2443: Remove transaction fee exception from betanet config
### Expected behavior
We don't need `transactionFee` exception for `betanet`

### Actual behavior
We have `transactionFee` exception for `betanet`
https://github.com/LiskHQ/lisk/blob/5d5af525eec7f2ee687ca921a6a2d1457c471da9/config/betanet/exceptions.js#L18-L24

`>= 1.0.0`

diego-G:
### Review checklist

* The PR resolves #2443
* Linting passes
* Tests pass


diego-G:
@ManuGowda I don't think we are forced to. It's true that it's not currently used but if we find a new case like it happened before, we will have the code ready for it. I let @MaciejBaj to decide.

4miners:
I don't like the approach of keeping unused code in the codebase. If we need it in the future then we can revert the commit.

ManuGowda:
> @ManuGowda I don't think we are forced to. It's true that it's not currently used but if we find a new case like it happened before, we will have the code ready for it. I let @MaciejBaj to decide.

In that case I think we should remove the exception from default config and remove the logic associated with it. I agree with @4miners we should not keep unwanted code.

diego-G:
@4miners  @ManuGowda Then the same happens with `exceptions.delegates` https://github.com/LiskHQ/lisk/blob/2e65e80462dcc82a4fd1de1669f98960c1dcd5ef/config/default/exceptions.js#L40 . It exists in the default config and the control is present in the logic but there is no single exception. Probably it happened the same: at some point it existed and that's why can be useful. 

@MaciejBaj please let me know what you think is better if you want to tackle none, one or both cases. I've checked and there are no more.

MaciejBaj:
We should avoid of keeping the non-used code. Please remove the logic responsible for transactionFee exceptions.

And nice catch @diego-G, the same applies to exceptions.delegates. 

We should do the same: remove the empty exceptions array and logic responsible for handling it. Could you open a separate issue for that?

MaciejBaj:
Closing in favour of https://github.com/LiskHQ/lisk/pull/2463

PR #2460, opened 2018-10-10
Update script adds null value to api whitelist array - Closes #2459

### LINKED ISSUE #2459: Update script adds null value to api whitelist array
### Expected behavior
config should have correct values insted of `null` when its not expected.
### Actual behavior
A `null` value is added to `whiteList` array as the ip `127.0.0.1` already exists in default config, the custom config contains `null` value due to the way update script merges the config.
```
"whiteList": [
        null,
	"62.23.239.233"
]
```
### Steps to reproduce
Upgrade node from `1.0.0-rc.x` to `1.1.0-rc.0`
1.1.0

nazarhussain:
### What was the problem?
The update script was extracting only the difference between two config files. So if there is change of any item in array type config attribute, it was extracting the attribute only at that index and setting every other index value to null. 

### How did I fix it?

We simply preserved what is available in old config file.

### Review checklist

* The PR resolves #2459
* Linting passes
* Tests pass

### DIFF
--- a/README.md
+++ b/README.md
@@ -324,8 +324,8 @@ pm2 stop lisk
    * Network specific configuration file
    * Custom configuration file (if specified by user)
    * Command line configurations, specified as command `flags` or `env` variables
-8. If you override any config option of type array, it will completely override. e.g. If you specify one peer at `peers.list` in your custom config file, it will replaces every default peer for the network.
-9. For development purposes use `devnet` as network option, others network are specific to public lisk networks.
+8. Any config option of array type gets completely overridden. If you specify one peer at `peers.list` in your custom config file, it will replace every default peer for the network.
+9. For development use `devnet` as the network option.
 
 ### Command Line Options
 
--- a/README.md
+++ b/README.md
@@ -324,7 +324,7 @@ pm2 stop lisk
    * Network specific configuration file
    * Custom configuration file (if specified by user)
    * Command line configurations, specified as command `flags` or `env` variables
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -324,7 +324,8 @@ pm2 stop lisk
    * Network specific configuration file
    * Custom configuration file (if specified by user)
    * Command line configurations, specified as command `flags` or `env` variables
-8. For development purposes use `devnet` as network option, others network are specific to public lisk networks.
+8. Please remember the fact that if you override an array type config value, it will completely override. e.g. If you specify one peer at `peers.list` in your custom config file, it will replaces every default peer for the network.
```
fchavant: "array type config value" might be best explained by "any configuration option of type array"
Or maybe we could just state that arrays do not get merged at all.

scripts/update_config.js
```diff
@@ -22,7 +22,7 @@
 const fs = require('fs');
 const path = require('path');
 const program = require('commander');
-const merge = require('lodash/merge');
+const _ = require('lodash');
```
ManuGowda: you can still do `const { merge, set, get } = require('lodash');` as you were intending to do `const merge = require('lodash/merge');` before.

nazarhussain: But I don't want to have `set`, `get` as seperate function. Looks less readable than `_.set(`

ManuGowda: :) so what was the intention behind `const merge = require('lodash/merge');`?

nazarhussain: Yes that time we need only `merge` from lodash. But now as we need more functions from it, we can use the library directly.

PR #2461, opened 2018-10-10
Upgrade bignumber.js library to 7.2.1 - Closes #2427

### LINKED ISSUE #2427: Upgrade bignumber.js dependency
### Expected behavior
We should use the newest version of `bignumber.js` library - `7.2.1`. This requires some refactor across the application because methods naming changed in `6.0.0` version of this library:
- `equals` -> `isEqualTo`/`eq`
- `floor` -> `integerValue(Bignum.ROUND_FLOOR)`
- `lessThan` -> `isLessThan`/`lt`
- `lessThanOrEqualTo` -> `isLessThanOrEqualTo`/`lte`
- `greaterThan` -> `isGreaterThan`/`gt`
- `greaterThanOrEqualTo` -> `isGreaterThanOrEqualTo`/`gte`

Optional, depends if we decide to go with short aliases or not:
- For `dividedBy` we can use an alias `div`.

Also would be nice to review usage of `toPrecision(15)` in `helpers->round_changes`, as is probably not needed at all and only add unnecessary complexity to the logic.

Another thing to review while performing this upgrade is custom methods that we're using in `helpers->bignum`. We have there `toBuffer` and `fromBuffer` which maybe are not required and can be replaced with some native methods.
### Actual behavior
We're currently using version `4.0.2` of the `bignumber.js` library, which is an old version.
### Steps to reproduce
N/A
`>= 1.0.0`

diego-G:
### How did I fix it?

An update across the application has been performed to match the new methods names of the newest version of the library. The decision was to follow with the full names instead of using alias. Here are all the changes implemented:

- `equals` -> `isEqualTo`
- `floor()` -> `integerValue(Bignum.ROUND_FLOOR)`
- `floor(2)` -> `decimalPlaces(2)`
- `round` -> `toFixed`
- `lessThan` -> `isLessThan`
- `lessThanOrEqualTo` -> `isLessThanOrEqualTo`
- `greaterThan` -> `isGreaterThan`
- `greaterThanOrEqualTo` -> `isGreaterThanOrEqualTo`
- `mul` -> `multipliedBy`
- `times` -> `multipliedBy`

### Review checklist

* The PR resolves #2427
* Linting passes
* Tests pass


logic/account.js
```diff
@@ -404,9 +404,10 @@ class Account {
 		const totalSupplyBignum = new Bignum(totalSupply);
 		const approvalBignum = votersBalanceBignum
 			.dividedBy(totalSupplyBignum)
-			.times(100)
-			.round(2);
-		return !approvalBignum.isNaN() ? approvalBignum.toNumber() : 0;
+			.multipliedBy(100)
+			.toFixed(2);
```
4miners: `.decimalPlaces(2)` instead of `.toFixed(2)` is better here because is not converting to `string`

logic/account.js
```diff
@@ -404,9 +404,10 @@ class Account {
...
+
+		return !isNaN(approvalBignum) ? Number(approvalBignum) : 0;
```
4miners: Then we can go with previous logic here, but I suggest changing `!approvalBignum.isNaN()` to `approvalBignum.isFinite()`

logic/account.js
```diff
@@ -422,9 +423,10 @@ class Account {
 		const missedBlocksBignum = new Bignum(missedBlocks || 0);
 		const percent = producedBlocksBignum
 			.dividedBy(producedBlocksBignum.plus(missedBlocksBignum))
-			.times(100)
-			.round(2);
-		return !percent.isNaN() ? percent.toNumber() : 0;
+			.multipliedBy(100)
+			.toFixed(2);
+
+		return !isNaN(percent) ? Number(percent) : 0;
```
4miners: Same here.

helpers/round_changes.js
```diff
@@ -64,18 +64,20 @@ function RoundChanges(scope) {
...
 		.dividedBy(slots.delegates)
-		.floor();
+		.integerValue(Bignum.ROUND_FLOOR);
 	var feesRemaining = new Bignum(this.roundFees.toPrecision(15)).minus(
-		fees.times(slots.delegates)
+		fees.multipliedBy(slots.delegates)
 	);
 	var rewards =
-		new Bignum(this.roundRewards[index].toPrecision(15)).floor() || 0;
+		new Bignum(this.roundRewards[index].toPrecision(15)).integerValue(
```
4miners: In the issue, there is a statement about reviewing the usage of `.toPrecision(15)` and getting rid of it if we can. What is the decision regarding it?

logic/account.js
```diff
@@ -508,24 +510,24 @@ class Account {
...
 
 						// If updated value is positive number
-						if (value.greaterThan(0)) {
+						if (value.isGreaterThan(0)) {
 							promises.push(
 								dbTx.accounts.increment(
 									address,
 									updatedField,
-									value.floor().toString()
+									value.toFixed().toString()
```
4miners: Possible change of behavior here, it should be `value.integerValue(Bignum.ROUND_FLOOR).toFixed()` or even better `value.toFixed(0, Bignum.ROUND_FLOOR)`. Are you sure that we always have integer here? If yes - we don't need `.toFixed()` here (or `toString()`, depends on preference).

logic/account.js
```diff
@@ -508,24 +510,24 @@ class Account {
...
-						} else if (value.lessThan(0)) {
+						} else if (value.isLessThan(0)) {
 							promises.push(
 								dbTx.accounts.decrement(
 									address,
 									updatedField,
 									value
 										.abs()
-										.floor()
+										.toFixed()
```
4miners: Same here.

MaciejBaj:
@diego-G I merged `1.1.0` branch into development, and few files are conflicting. Please have a look at: 
``
Conflicting files
logic/dapp.js
logic/delegate.js
logic/in_transfer.js
logic/out_transfer.js
logic/signature.js
``

logic/dapp.js
```diff
@@ -85,7 +85,7 @@ DApp.prototype.verify = function(transaction, sender, cb, tx) {
 	}
 
 	const amount = new Bignum(transaction.amount);
-	if (!amount.isEqualTo(0)) {
+	if (amount.isGreaterThan(0)) {
```
shuse2: why is it changing the logic here?

logic/delegate.js
```diff
@@ -82,7 +82,8 @@ Delegate.prototype.verify = function(transaction, sender, cb, tx) {
 		return setImmediate(cb, 'Invalid recipient');
 	}
 
-	if (!transaction.amount.isEqualTo(0)) {
+	const amount = new Bignum(transaction.amount);
+	if (amount.isGreaterThan(0)) {
```
shuse2: same as above

logic/in_transfer.js
```diff
@@ -92,8 +92,7 @@ InTransfer.prototype.verify = function(transaction, sender, cb, tx) {
 	}
 
 	const amount = new Bignum(transaction.amount);
-
-	if (!amount.isEqualTo(0)) {
+	if (amount.isLessThanOrEqualTo(0)) {
```
shuse2: same as above

logic/out_transfer.js
```diff
@@ -90,8 +90,7 @@ OutTransfer.prototype.verify = function(transaction, sender, cb) {
 	}
 
 	const amount = new Bignum(transaction.amount);
-
-	if (!amount.isEqualTo(0)) {
+	if (amount.isLessThanOrEqualTo(0)) {
```
shuse2: same as above

logic/signature.js
```diff
@@ -81,7 +81,7 @@ Signature.prototype.verify = function(transaction, sender, cb) {
 	}
 
 	const amount = new Bignum(transaction.amount);
-	if (!amount.isEqualTo(0)) {
+	if (amount.isGreaterThan(0)) {
```
shuse2: same as above

logic/dapp.js
diego-G: This change comes from development. I've just solved the conflicts after the merge following the existing pattern. The commit was introduced in this PR https://github.com/LiskHQ/lisk/pull/2432
I was checking with @ManuGowda one of the reviewers, if there was an important change of the logic overlooked. He confirmed me @nazarhussain was aware of this possibility and everything is alright.

test/unit/logic/block_reward.js
```diff
@@ -121,37 +121,49 @@ describe('BlockReward @slow', () => {
 
 		it('when height == (milestoneFour * 2) should return 4', () => {
 			return expect(
```
4miners: Those tests in this file are not good. `blockReward.calcMilestone` is accepting `height` as parameter, not a `Bignum` instance. Also, expectations are wrong, returning false-positives (`expect(someting);`).

test/unit/logic/block_reward.js
```diff
@@ -162,129 +174,133 @@ describe('BlockReward @slow', () => {
 		});
 
 		it('when height == 0 should return 0', () => {
-			return expect(blockReward.calcReward(0).equals(0));
+			return expect(blockReward.calcReward(0).isEqualTo(0));
```
4miners: Similar here, should be `expect(something).to.be.true;`, there are many in this file.

test/unit/logic/dapp.js
```diff
@@ -112,7 +112,7 @@ describe('dapp', () => {
 		describe('calculateFee', () => {
 			it('should return FEES.DAPP_REGISTRATION', () => {
 				return expect(
-					dapp.calculateFee(transaction).equals(FEES.DAPP_REGISTRATION)
+					dapp.calculateFee(transaction).isEqualTo(FEES.DAPP_REGISTRATION)
```
4miners: Same here, evaluates to `expect(true/false);`.

test/unit/logic/block_reward.js
MaciejBaj: Please @diego-G open a separate issue for fixing the unit tests with the `equals` pointless expectation checks. Please specify which functions need to be covered.

ManuGowda:
@diego-G @MaciejBaj @4miners the false positive test has been fixed as part this [commit](https://github.com/LiskHQ/lisk/pull/2474/commits/30a5aa3ff0c4644fb18f7400c9c05c21d70f6c5c), however, the specific issue has not been created, I will create the issue and reference it the issue.
@diego-G Once `1.1.0` is merged with development everything will be fixed related to `block_reward`.

ManuGowda:
@diego-G, as discussed with @4miners still a part of the issue, needs to be fixed related to height being `bignumber` and other comments form @4miners.

helpers/round_changes.js
ManuGowda: please create the separate issue and consider fixing it in all the places because it had implications, here is an example https://runkit.com/1manugowda/bignumber-toprecision-issue.

logic/account.js
```diff
@@ -508,25 +510,18 @@ class Account {
...
 						// If updated value is positive number
-						if (value.greaterThan(0)) {
+						if (value.isGreaterThan(0)) {
 							promises.push(
-								dbTx.accounts.increment(
-									address,
-									updatedField,
-									value.floor().toString()
-								)
+								dbTx.accounts.increment(address, updatedField, value.toString())
```
ManuGowda: why did we remove `floor` here? are you sure that it will not have any implications? a good suggestion from my side for this PR, in general, is to try to sync the `testnet` from scratch, because in the previous encounter related to bignum changes we faced lot of issues, so it will be great if you can sync using this branch.

logic/account.js
```diff
@@ -508,25 +510,18 @@ class Account {
...
+						} else if (value.isLessThan(0)) {
 							promises.push(
 								dbTx.accounts.decrement(
 									address,
 									updatedField,
-									value
-										.abs()
-										.floor()
-										.toString()
+									value.abs().toString()
```
ManuGowda: same here, I am not sure about the implications for removing `floor`

test/unit/logic/account.js
```diff
@@ -661,7 +661,7 @@ describe('account', () => {
 
 		it('should throw error when a numeric field receives non numeric value', done => {
 			account.merge(validAccount.address, { balance: 'Not a Number' }, err => {
-				expect(err).to.equal('Account#merge error');
+				expect(err).to.equal('Encountered insane number: NaN');
```
ManuGowda: any reason why this error message changed?

logic/account.js
diego-G: @4miners was in the loop. We confirmed that `floor()` wasn't adding anything in those cases.

test/unit/logic/account.js
diego-G: Now the error is caught earlier than before.

ManuGowda: ok, in that case, shouldn't we cover the `Account#merge error` scenario with a different test?

logic/account.js
ManuGowda: ok, if you guys have a clue that it will work then cool, my only suggestion was a quick check to verify if these big number changes work, if you are confident then 👍

test/unit/logic/account.js
diego-G: It is cover. I find at least 4 test cases.

test/unit/logic/block.js
```diff
@@ -667,7 +667,7 @@ describe('block', () => {
 
 	describe('calculateFee', () => {
 		it('should return the constant fee', () => {
-			return expect(block.calculateFee(blockData).equals(FEES.SEND));
+			return expect(block.calculateFee(blockData).isEqualTo(FEES.SEND));
```
4miners: false-positive here

test/unit/logic/in_transfer.js
```diff
@@ -205,7 +205,7 @@ describe('inTransfer', () => {
 
 	describe('calculateFee', () => {
 		it('should return FEES.SEND', () => {
-			return expect(inTransfer.calculateFee(trs).equals(FEES.SEND));
+			return expect(inTransfer.calculateFee(trs).isEqualTo(FEES.SEND));
```
4miners: also here

test/unit/logic/transaction.js
```diff
@@ -1109,7 +1109,7 @@ describe('transaction', () => {
 										accountAfter.balance.toString()
 									);
 
-									expect(balanceAfter.equals(balanceBefore));
+									expect(balanceAfter.isEqualTo(balanceBefore));
```
4miners: and here

4miners:
Thanks @diego-G, really nice work on this one! :D

diego-G:
@MaciejBaj I've opened another issue to research on how to fix the false positives problem once for all. Here you go the link https://github.com/LiskHQ/lisk/issues/2480

PR #2467, opened 2018-10-12
Rollback version patching mechanism - Closes #2466

### LINKED ISSUE #2466: Rollback version patching mechanism
### Expected behavior

Our approach to versioning should be compatible with older nodes.

Our last patch for this issue https://github.com/LiskHQ/lisk/issues/2389 did not solve the problem completely (peers were still getting our node's unpatched version number via other mechanisms); moving further in that direction by patching this issue everywhere in the code would add too much complexity. The simplest way forward is to revert our previous patch and to simplify our versioning approach by dropping the `-rc.x` suffix from the `version` property in package.json for testnet deployments.

### Actual behavior

Our current versioning scheme deals with release candidates by adding a `-rc.x` suffix after the version number in package.json; this versioning scheme is backwards-compatible but not forward-compatible with old `1.0.0-rc.x` nodes. This issue only affects testnet.

### Steps to reproduce

Run a `1.1.0-rc.0` node and connect it to testnet nodes `1.0.0-rc.x` - It should disconnect often with `code 1000`.

Removing the `-rc.0` suffix from the `version` property in `package.json` fixes the issue.

`1.1.0-rc.0` and its forward compatibility with `1.0.0-rc.x`

jondubois:
### What was the problem?

Our approach to versioning should be compatible with older nodes.
A previous incomplete patch needed to be rolled back. We need to change our approach to versioning to be compatible with testnet nodes.

### How did I fix it?

- Reverted commits in https://github.com/LiskHQ/lisk/issues/2389 and https://github.com/LiskHQ/lisk/pull/2444/commits
- Changed the version number in package.json to `1.1.0`
- Fixed an issue with integration tests which was causing it to count the same node monitoring sockets multiple times after a node restarted.

### How to test it?

Run the node and connect to testnet, peer disconnects with `code 1000` should not (or rarely) happen during the sync phase.

### Review checklist

* The PR resolves #2466
* Linting passes
* Tests pass


jondubois:
@webmaster128 That's correct. This PR reverts a patch which did not work as intended. There were two options; push forward with the hack (making more changes to the code) until everything appears to work, or roll back what we had and find an alternative solution.

It was decided that the best solution for this issue did not require a code change.
There should be an announcement very soon.

webmaster128:
@jondubois thanks for the info. I am happy this is expeted behaviour and well known. Looking forward to see 1.1 being deployed.

MaciejBaj:
@webmaster128 I summarized the problems here https://github.com/LiskHQ/lisk/issues/2389#issuecomment-430239260.

PR #2470, opened 2018-10-15
Add filter data field to get transactions endpoint - Closes #2404

### LINKED ISSUE #2404: Allow to search for transactions by data field
### Expected behavior
It should be possible to search for 0-type transactions by the optional data field.

`/api/transactions/?senderAddress=12498L&data=message` would return transactions with sender address `12498L` that contain word `message` in data field

In Lisk Hub, we would use this feature to allow users to look up their transactions by their "reference message".

### Actual behavior
It is not possible to filter by data field
https://app.swaggerhub.com/apis/LiskHQ/Lisk/1.0.20#/Transactions/getTransactions

diego-G:
### What was the problem?

There was no way to get a transaction from its attached additional data field.

### How did I fix it?

Adding a subquery filter on GET transactions endpoint.

### How to test it?

`mocha test/functional/http/get/transactions.js`

### Review checklist

* The PR resolves #2404 
* Linting passes
* Tests pass


modules/transactions.js
```diff
@@ -121,6 +121,9 @@ __private.list = function(filter, cb) {
 		maxAmount: '"t_amount" <= ${maxAmount}',
 		type: '"t_type" = ${type}',
 		minConfirmations: 'confirmations >= ${minConfirmations}',
+		data: `ENCODE((SELECT data FROM transfer WHERE transfer."transactionId" = trs_list."t_id"), 'escape') LIKE '${
+			filter.data
+		}'`,
```
nazarhussain: After research, I suggest for now we go with exact match instead of regexpr: 

1. This must be going to impact the performance 
2. The `encode` with `escape` just escape the special characters so will change the actual content 
   - If user added a transaction with `\u0000 hey how :)`, escaping it will turn it to `\000 hey how :)`
   - If user search with key word `00` the query will match and return the transaction 
3. Since the actual content is different then escaped one so match will not result accurate 

So until we have this `data` field as binary, we should only provide the exact match. Once we find way to convert it to utf8 strings, then we may provide the pattern matching.

modules/transactions.js
```diff
@@ -106,7 +106,7 @@ __private.list = function(filter, cb) {
 	if (filter.data) {
 		// PostgreSQL does not support zero byte character in utf8 strings
 		// so we have to specify text in hex representation
```
shuse2: This comment is not true anymore. it's Buffer

4miners: @shuse2 Fixed.

PR #2472, opened 2018-10-15
Add from/to timestamp filter to get blocks endpoint - Closes #2391

### LINKED ISSUE #2391: Add fromTimestamp and toTimestamp to Blocks API endpoint
### Expected behavior
Filter block results by fromTimestamp and toTimestamp.

### Actual behavior
Currently there is no way to retrieve data from a specific range of blocks using timestamps. 

This would be a great feature for (forging) delegates who'd like to make detailed reports of their (daily, weekly, monthly or yearly) forged blocks.

Example: Get all forged blocks for delegate tembo from 2018/07/01 to 2018/08/01:
`https://node01.lisk.io/api/blocks?limit=100&fromTimestamp=66286800&toTimestamp=68965200&offset=0&generatorPublicKey=326bff18531703385d4037e5585b001e732c4a68afb8f82efe2b46c27dcf05aa&sort=height%3Aasc`

`>1.0`

diego-G:
### How to test it?

`mocha test/functional/http/get/blocks.js`

### Review checklist

* The PR resolves #2391
* Linting passes
* Tests pass


modules/blocks/api.js
```diff
@@ -106,6 +106,16 @@ __private.list = function(filter, cb) {
 		params.height = filter.height;
 	}
 
+	if (filter.fromTimestamp >= 0) {
```
nazarhussain: In swagger we say `0` is the valid value. Should not we simplify it on API level to disallow zero as param value? 

https://github.com/LiskHQ/lisk/blob/5acfc7bdd902965124263d448be124edf481b6ae/schema/swagger.yml#L1169-L1174

modules/blocks/api.js
```diff
@@ -106,6 +106,16 @@ __private.list = function(filter, cb) {
...
+		where.push('"b_timestamp" >= ${fromTimestamp}');
+		params.fromTimestamp = filter.fromTimestamp;
+	}
+
+	if (filter.toTimestamp >= 0) {
```
nazarhussain: Same here.

modules/blocks/api.js
diego-G: The idea is to keep `0` as an allowed value. However, there was a mistake in the expectation of `toTimestamp` that I've caught thanks to this comment. @nazarhussain could you check it again? I hope now it's easier to understand.

PR #2475, opened 2018-10-16
Logs are created dynamically based on network name - Closes #2452

### LINKED ISSUE #2452: Create log file dynamically based on network
### Expected behavior
Logfile name should be created dynamically based on the network if I create a network `testnet-sync` then the log file should be `lisk-testnet-sync.log` or we can create separate log directory specific to network like `logs/testnet-sync/lisk.log`, either of the approaches should be good.

This way if a user is running multiple node from same code base at same place just by having custom network configuration then the logs will be created dynamically based on the network rather than changing the log file name in the custom network config file.

### Actual behavior
Currently, when we start a network with parameter `testnet-sync`, we need to specify explicitly the log file name for custom log file name.

pablitovicente:
### What was the problem?

Logs were not unique to each network.

### How did I fix it?

Based on PR feedback I added the full path to each network. When creating a custom network the paths need to be set on its config file.

### How to test it?

- Run  `node app.js` and it should create `logs/devnet` (default network name)
- Create a config, set the log paths for both properties logFile and db.logFile and run `node app.js -n <your_network_name>` and it should create the logs in `logs/<your_network_name> ` 

### Review checklist

* The PR resolves #2452
* Linting passes
* Tests pass

### DIFF
--- a/app.js
+++ b/app.js
@@ -159,7 +159,6 @@ var logger = new Logger({
 	echo: process.env.LOG_LEVEL || appConfig.consoleLogLevel,
 	errorLevel: process.env.FILE_LOG_LEVEL || appConfig.fileLogLevel,
 	filename: appConfig.logFileName,
-	network: appConfig.network,
 });
 
 /**
@@ -180,7 +179,6 @@ if (
 		echo: process.env.DB_LOG_LEVEL || 'log',
 		errorLevel: process.env.FILE_LOG_LEVEL || 'log',
 		filename: appConfig.db.logFileName,
-		network: appConfig.network,
 	});
 }
 
--- a/app.js
+++ b/app.js
@@ -159,6 +159,7 @@ var logger = new Logger({
 	echo: process.env.LOG_LEVEL || appConfig.consoleLogLevel,
 	errorLevel: process.env.FILE_LOG_LEVEL || appConfig.fileLogLevel,
 	filename: appConfig.logFileName,
+	network: appConfig.network,
 });
 
 /**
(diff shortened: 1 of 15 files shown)


app.js
```diff
@@ -159,6 +159,7 @@ var logger = new Logger({
 	echo: process.env.LOG_LEVEL || appConfig.consoleLogLevel,
 	errorLevel: process.env.FILE_LOG_LEVEL || appConfig.fileLogLevel,
 	filename: appConfig.logFileName,
+	network: appConfig.network,
```
nazarhussain: Suggest not to add network specific dependency to logger. It's against of the separation of concern rule. Logger just need to log, for it it needs a file path, log level and any additional config attribute for logging. 

The logic to choose the log file path based on network is responsibility of one level higher.

logger.js
```diff
@@ -51,8 +51,16 @@ module.exports = function(config) {
 
 	config.errorLevel = config.errorLevel || 'log';
 
-	child_process.execSync(`mkdir -p ${path.dirname(config.filename)}`);
-	var log_file = fs.createWriteStream(config.filename, { flags: 'a' });
+	const logsSubDirectory = config.network || '';
+	const logsDirectory = path.join(
+		path.dirname(config.filename),
+		logsSubDirectory
```
nazarhussain: Here it makes things confusing. The log file path is a node configuration option. User set it to some path, and you changed that path runtime, make the user wondering where his log file is. 

So right way to do is to just add a custom configuration override for the network with different log file paths.

nazarhussain:
@Nazgolze @fchavant This change requires your attention to relevant change in archiving, logrotate or graylogs. 

@MaciejBaj We should highlight this change promptly, as user can get confused finding their logs after this change.

ManuGowda:
@nazarhussain the DevOps team is informed already and as part of the dev updates we should cover this and same can be communicated in the chat or wherever necessary.

PR #2488, opened 2018-10-19
Redundant logic in getBlockSlotData - Closes #2422

### LINKED ISSUE #2422: Redundant logic in getBlockSlotData
### Expected behavior
Logic in `modules.delegates->__private.getBlockSlotData` should be something like following:
- Call `generateDelegatesList`.
- Return `keypair` when delegate for current slot exists in `__private.keypairs` (forging enabled) or `null` if not.

Then in `__private.forge`:
- Remove check described below.
- Remove `currentBlockData === null` condition form:
https://github.com/LiskHQ/lisk/blob/7922d346e115406a0004aa729e2ecfd18c5d43e4/modules/delegates.js#L258
- Change check described below to check for `currentBlockData === null`.
- `time` can be set directly when calling `modules.blocks.process.generateBlock`.

### Actual behavior
Function `modules.delegates->__private.getBlockSlotData` is using following logic:
- Call `generateDelegatesList`.
- Iterate 101 times (through slots) until it finds delegate public key in `__private.keypairs`.
- Returns slot time together with corresponding keypair or `null` if it cannot found delegate on the list (forging enabled, but delegate is not the active one).

This function is called form `__private.forge` in the same module, but later we have a check here:
https://github.com/LiskHQ/lisk/blob/7922d346e115406a0004aa729e2ecfd18c5d43e4/modules/delegates.js#L263-L265 This check is `true` most of the times because delegate can forge only once during the round (in perfect conditions).

This logic is redundant and unnecessary complex.
### Steps to reproduce
N/A
`>= 1.0.0`

lsilvs:
### What was the problem?
The function `modules.delegates->__private.getBlockSlotData` was unnecessarily iterating 101 times and most of the times returning data that was never used.

### How did I fix it?
I've refactored the `getBlockSlotData` in order to only return the delegate keypair when it's its slot otherwise return `null`.

The function was also changed to just return the keypair as the slot time can be calculated outside it.

To improve readability, I've updated the function name to match what it does and also some variables names and log messages.

### How to test it?
Running the tests

### Review checklist

* The PR resolves #2422
* Linting passes
* Tests pass


modules/delegates.js
```diff
@@ -247,23 +239,16 @@ __private.forge = function(cb) {
...
-				return setImmediate(cb);
-			}
-
-			if (
-				slots.getSlotNumber(currentBlockData.time) !== slots.getSlotNumber()
-			) {
-				library.logger.debug('Delegate slot', slots.getSlotNumber());
+		(err, delegateKeypair) => {
+			if (err || delegateKeypair === null) {
+				library.logger.warn('Waiting for delegate slot', err);
```
4miners: When there is `err` present here it comes from the SQL query, it should be logged separately with `error` level.

4miners: When there is `err` present here it comes from the SQL query, it should be logged separately with `error` level.

4miners: Also `Waiting for delegate slot` should not be a `warn`, just `debug`, as this is completely normal behavior, nothing to worry about.

modules/delegates.js
```diff
@@ -247,23 +238,24 @@ __private.forge = function(cb) {
...
+				library.logger.error('Failed to get delegate keypair', err);
 				return setImmediate(cb);
 			}
 
-			if (
-				slots.getSlotNumber(currentBlockData.time) !== slots.getSlotNumber()
-			) {
-				library.logger.debug('Delegate slot', slots.getSlotNumber());
+			if (delegateKeypair === null) {
+				library.logger.debug('Waiting for delegate slot');
```
4miners:
```suggestion
				library.logger.debug('Waiting for delegate slot', { currentSlot: slots.getSlotNumber() });
```

modules/delegates.js
```diff
@@ -247,23 +238,24 @@ __private.forge = function(cb) {
...
+	__private.getDelegateKeypairForCurrentSlot(
 		currentSlot,
-		lastBlock.height + 1,
-		(err, currentBlockData) => {
-			if (err || currentBlockData === null) {
-				library.logger.warn('Skipping delegate slot', err);
+		round,
+		(err, delegateKeypair) => {
+			if (err) {
+				library.logger.error('Failed to get delegate keypair', err);
```
4miners:
```suggestion
				library.logger.error('Skipping delegate slot', err);
```

modules/delegates.js
```diff
@@ -181,35 +181,26 @@ __private.validateBlockSlot = function(block, source, cb) {
...
  * @param {number} slot
- * @param {number} height
+ * @param {number} round
  * @param {function} cb - Callback function
  * @returns {setImmediateCallback} cb, err, {time, keypair}
  * @todo Add description for the params
  */
-__private.getBlockSlotData = function(slot, height, cb) {
-	const round = slots.calcRound(height);
+__private.getDelegateKeypairForCurrentSlot = function(currentSlot, round, cb) {
```
nazarhussain: Just a thought as we are refactoring this function a lot, should we convert it to a Promise.

test/unit/modules/delegates.js
```diff
@@ -489,6 +489,95 @@ describe('delegates', () => {
...
+				done();
+			});
+
+			it('should be ok for slot 35 of first round', done => {
+				const currentSlot = 35;
+				const round = 1;
+
+				getDelegateKeypairForCurrentSlot(currentSlot, round, (err, keyPair) => {
+					expect(err).to.be.null;
+					expect(keyPair).to.have.keys('publicKey', 'privateKey');
```
nazarhussain: We are not expecting is this function is returning correct key pair or not.

test/unit/modules/delegates.js
```diff
@@ -489,6 +489,95 @@ describe('delegates', () => {
...
+				__private = library.rewiredModules.delegates.__get__('__private');
+
+				__private.keypairs = {};
+				config.forging.force = true;
+				config.forging.delegates = delegates;
+				loadDelegates();
+				done();
+			});
+
+			it('should be ok for slot 35 of first round', done => {
```
nazarhussain: Expectation statements must be more elaborated to be maintainable. Why slot 35 should be ok? Either we add some comments to explain or add more elaborated statement. e.g. 

```
describe('Having list of 3 delegates', () => {
    it('should return 3rd delegate on 35 slot of first round');
    it('should return 2nd delegate on 73 slot of second round');
    it('should return 1st delegate on 41 slot of third round');
    it('should return null for slot 1 of fourth round');
});
```

test/unit/modules/delegates.js
```diff
@@ -489,6 +489,95 @@ describe('delegates', () => {
...
+				const round = 3;
+
+				getDelegateKeypairForCurrentSlot(currentSlot, round, (err, keyPair) => {
+					expect(err).to.be.null;
+					expect(keyPair).to.have.keys('publicKey', 'privateKey');
+					done();
+				});
+			});
+
+			it('should return null for slot 1 of fourth round', done => {
```
nazarhussain: Why this should return null? As per my understanding `(101 * 4 + 1) mod 3` should return delegate in the list on `0` index. Am I missing something?

lsilvs: The `index` here is not for the 3 delegates array set in config. It is for the shuffled delegates list returned by `generateDelegateList`.
The correct calculation is `1 mod 101` returning the `1` index of the shuffled `activeDelegates` list for the 4th round.
Basically what this test says is that the slot 1 of the 4th round is not a slot of any 3 delegates set for the test.

nazarhussain: Good. Please elaborate the statement and add some comments. So later some one read the tests would understand the logic.

modules/delegates.js
```diff
@@ -247,23 +238,26 @@ __private.forge = function(cb) {
 	const lastBlock = modules.blocks.lastBlock.get();
 
 	if (currentSlot === slots.getSlotNumber(lastBlock.timestamp)) {
-		library.logger.debug('Waiting for next delegate slot');
+		library.logger.debug('Block already forged for the current slot');
```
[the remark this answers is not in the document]
lsilvs: I get your point but in this case the message `Waiting for next delegate slot` is not accurate.

I'm in favour of changing it and document it in the release notes.

@MaciejBaj What's your call?

modules/delegates.js
```diff
@@ -247,23 +238,26 @@ __private.forge = function(cb) {
...
+	__private.getDelegateKeypairForCurrentSlot(
 		currentSlot,
-		lastBlock.height + 1,
-		(err, currentBlockData) => {
-			if (err || currentBlockData === null) {
-				library.logger.warn('Skipping delegate slot', err);
+		round,
+		(err, delegateKeypair) => {
+			if (err) {
+				library.logger.error('Skipping delegate slot', err);
```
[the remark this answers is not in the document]
lsilvs: As @4miners pointed out, this is an explicit error and should be logged as such.
I'm in favour of changing it and document it in the release notes.
@MaciejBaj What's your call?

modules/delegates.js
lsilvs: As Promises are not heavily used across our application, I reckon it should be a separated task to the whole application.

test/unit/modules/delegates.js
lsilvs: Done

lsilvs:
@4miners I was discussing this with others yesterday and I agree with you. I've noticed that many of our unit tests are not actually unit tests and I just followed what was implemented before. I will make it unit then.

modules/delegates.js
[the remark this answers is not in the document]
MaciejBaj: @lsilvs, the `Waiting for next delegate slot` is not really accurate in this step of the `forge` function. It also covers the same log message in the next step:
```
library.logger.debug('Waiting for delegate slot', {
			) {						currentSlot: slots.getSlotNumber(),
				library.logger.debug('Delegate slot', slots.getSlotNumber());					});
```
It's a `debug` level message, so it should not be qualified as breaking. 

`Block already forged for the current slot` might suggest a delegate already forged this particular block, whereas the check only covers if a correct block was applied for a given slot.

More accurate would be: `Block already applied for the current slot, skipping forging attempt.`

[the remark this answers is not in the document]
4miners: @MaciejBaj I like `Block already forged for the current slot` log message more, the one you proposed is more confusing because we are not skipping forging attempt here as it's not our slot.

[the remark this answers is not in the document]
MaciejBaj: Ok please leave the "Block already forged..."

test/unit/modules/delegates.js
```diff
@@ -489,6 +490,172 @@ describe('delegates', () => {
...
+						publicKey: new Buffer.from(
+							'3ff32442bb6da7d60c1b7752b24e6467813c9b698e0f278d48c43580da972135',
+							'hex'
+						),
+						privateKey: new Buffer.from(
+							'd4b9f54978f789c1a291f948d32f19039d54962b5b4c14d314c1521f2975b9be3ff32442bb6da7d60c1b7752b24e6467813c9b698e0f278d48c43580da972135',
+							'hex'
+						),
+					},
+				};
```
nazarhussain: I feel non-confident of mentioning private keys. In our config.json we have delegates in the form; 

```
{
   "encryptedPassphrase":
          "iterations=1&salt=476d4299531718af8c88156aab0bb7d6&cipherText=663dde611776d87029ec188dc616d96d813ecabcef62ed0ad05ffe30528f5462c8d499db943ba2ded55c3b7c506815d8db1c2d4c35121e1d27e740dc41f6c405ce8ab8e3120b23f546d8b35823a30639&iv=1a83940b72adc57ec060a648&tag=b5b1e6c6e225c428a4473735bc8f1fc9&version=1",
    "publicKey":
          "9d3058175acab969f41ad9b86f7a2926c74258670fe56b37c429c01fca9f2f0f"
},
```

I saw you used same public keys, so I suggest to load their encrypted pass phrase from config runtime and generate the private keys, don't hardcode private keys.

test/unit/modules/delegates.js
```diff
@@ -489,6 +490,172 @@ describe('delegates', () => {
...
+
+				__private.getDelegateKeypairForCurrentSlot(
+					currentSlot,
+					round,
+					(err, keyPair) => {
+						expect(err).to.be.null;
+						expect(keyPair).to.have.keys('publicKey', 'privateKey');
+						expect(keyPair.publicKey.toString('hex')).to.equal(
+							'9d3058175acab969f41ad9b86f7a2926c74258670fe56b37c429c01fca9f2f0f'
```
nazarhussain: Better would be store those keys explicitly to variable e.g. `genesis_1_publicKey` and use those in expectations. This will make expectations more readable.

test/unit/modules/delegates.js
```diff
@@ -489,6 +491,173 @@ describe('delegates', () => {
...
+
+				done();
+			});
+
+			it('should return genesis_1 keypair for slot N where (N % 101 === 35) in the first round', done => {
+				// For round 1, delegates genesis_1, genesis_2 and genesis_3 should forge for slots 35, 53 and 16 respectively.
+				const currentSlot = 35;
+				const round = 1;
+
+				delegates.generateDelegateList = (round, source, cb) => {
```
4miners: This function is async, we're not waiting for it so there is no guarantee that it will finish before the next one. Globally in those tests.

lsilvs: I'm overriding the `generateDelegateList` function and not executing it.
It will be executed during `getDelegateKeypairForCurrentSlot` call.

4miners: Aaaa, ok. 😄

PR #2501, opened 2018-10-25
"Error writing cache count for transactions - Cache Disabled" when /api/node/status is called - Closes 2482

### LINKED ISSUE #2482: "Error writing cache count for transactions - Cache Disabled" when /api/node/status is called
### Expected behavior
No errors in the log when /api/node/status is called

### Actual behavior
When I call the /api/node/status endpoint, I get the following error in the log:
`Error writing cache count for transactions - Cache Disabled`

### Steps to reproduce
monitor the log with `tail -f logs/lisk.log`
then call
`curl -s localhost:7000/api/node/status`

1.1.0.rc0 on testnet

pablitovicente:
### What was the problem?

No errors in the log when /api/node/status is called

### How did I fix it?

Changed log level from error to warning

### How to test it?

run testnet then `curl -X GET http://localhost:7000/api/node/status`

### Review checklist

* The PR resolves #2482
* Linting passes
* Tests pass

### DIFF
--- a/modules/transactions.js
+++ b/modules/transactions.js
@@ -873,7 +873,7 @@ Transactions.prototype.shared = {
 						},
 						err => {
 							if (err) {
-								library.logger.error(
+								library.logger.warn(
 									'Error writing cache count for transactions',
 									err
 								);
--- a/modules/transactions.js
+++ b/modules/transactions.js
@@ -874,7 +874,7 @@ Transactions.prototype.shared = {
 						err => {
 							if (err) {
 								library.logger.warn(
-									'Error writing cache count for transactions',
+									'Not writing cache count for transactions',
 									err
 								);
 							}
--- a/modules/transactions.js
+++ b/modules/transactions.js
@@ -873,10 +873,7 @@ Transactions.prototype.shared = {
 						},
 						err => {
 							if (err) {
-								library.logger.warn(
-									'Unable to write cache count for transactions',
-									err
-								);
+								library.logger.warn("Transaction count wasn't cached", err);
 							}
 
 							return setImmediate(waterCb, null, dbCount);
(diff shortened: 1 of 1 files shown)


pablitovicente:
@4miners I did talk with @MaciejBaj about just what you mention and we thought that might be a bigger change as to how currently cache works and might need to be done in a separate issue. Thoughts @MaciejBaj ?

MaciejBaj:
@4miners @pablitovicente the check if the cache is enabled is just a few lines below:

```
Cache.prototype.setJsonForKey = function(key, value, cb) {
	cb =
		cb ||
		function() {
			logger.debug('Cache - Value set for key');
		};

	logger.debug(
		['Cache - Set value for key:', key, '| Status:', self.isConnected()].join(
			' '
		)
	);
	if (!self.isConnected()) {
		return cb(errorCacheDisabled);
	}

	// Redis calls toString on objects, which converts it to object [object] so calling stringify before saving
	client.set(key, JSON.stringify(value), cb);
};
```
Also one debug level message is there already. So the check is already there. Potentially they might be some other errors while caching the transactions count so I would rather keep the message in a tone that "Transactions count wasn't cached - ${cause}"
```
function updateConfirmedCountToCache(cachedCount, dbCount, waterCb) {
					if (cachedCount) {
						// Cache already persisted, no need to set cache again
						return setImmediate(waterCb, null, cachedCount);
					}

					modules.cache.setJsonForKey(
						modules.cache.KEYS.transactionCount,
						{
							confirmed: dbCount,
						},
						err => {
							if (err) {
								library.logger.error(
									'Error writing cache count for transactions',
									err
								);
							}

							return setImmediate(waterCb, null, dbCount);
						}
					);
				},
```

PR #2505, opened 2018-10-26
setAccountAndGet function returns sender as null leading to fatal exception - Closes #2494

### LINKED ISSUE #2494: setAccountAndGet function returns sender as null leading to fatal exception
### Expected behavior
`setAccountAndGet` function should always return sender.
### Actual behavior
Log detail from mainnet node
```
[FTL] 2018-10-23 12:00:52 | Domain master - {"message":"Cannot read property 'multisignatures' of null","stack":"TypeError: Cannot read property 'multisignatures' of null\n    at getRequester (/home/lisk/lisk-main/logic/transaction_pool.js:837:26)\n    at nextTask (/home/lisk/lisk-main/node_modules/async/dist/async.js:5273:14)\n    at Immediate.next (/home/lisk/lisk-main/node_modules/async/dist/async.js:5280:9)\n    at Immediate.<anonymous> (/home/lisk/lisk-main/node_modules/async/dist/async.js:906:16)\n    at Immediate.args.(anonymous function) (/home/lisk/lisk-main/lib/node_modules/pm2/node_modules/event-loop-inspector/index.js:138:29)\n    at runCallback (timers.js:676:20)\n    at tryOnImmediate (timers.js:645:5)\n    at processImmediate [as _immediateCallback] (timers.js:617:5)"}
[FTL] 2018-10-23 12:00:52 | TypeError: Cannot read property 'multisignatures' of null
[inf] 2018-10-23 12:00:52 | Cleaning up...
[inf] 2018-10-23 12:00:53 | Rebuilding blockchain, current block height: 1
```
Code reference:
https://github.com/LiskHQ/lisk/blob/d4736c1f05a8c87e2e25d6cb273c6ac07361e4f5/logic/transaction_pool.js#L844-L858
### Steps to reproduce
1) Run `1.1.0` on mainnet
2) Use `installLisk.sh` to upgrade to `1.1.1`
3) The issue is not consistantly reproducible, however keep trying it you might hit the jackpot 😄

lsilvs:
### What was the problem?
In some edge case scenario `setAccountAndGet` was returning `null` without any error.

### How did I fix it?
Forcing `setAccountAndGet` to use database transaction to perform both `logic.account.set` and `logic.account.get` operations.
This way we guarantee the queries are executed one after another and avoid getting before setting.

### How to test it?
As the issue is not reproducible, our test suit should be used.

### Review checklist

* The PR resolves #2494
* Linting passes
* Tests pass


db/repos/accounts.js
```diff
@@ -211,14 +212,17 @@ class AccountsRepository {
...
-		});
+		const task = t =>
+			t.accounts.list(conditionObject, ['address']).then(result => {
+				if (result.length) {
+					return t.accounts.update(result[0].address, updateData);
+				}
+				return t.accounts.insert(data);
+			});
+
+		return this.inTransaction
```
nazarhussain: Please add a comment on above this line saying

> To avoid nested transactions a.k.a. save points

modules/accounts.js
```diff
@@ -156,17 +158,36 @@ Accounts.prototype.setAccountAndGet = function(data, cb, tx) {
...
-		},
-		tx
-	);
+	const task = t =>
+		new Promise((resolve, reject) => {
+			library.logic.account.set(
+				address,
+				data,
+				err => {
+					if (err) {
```
nazarhussain: Please follow convention for specifying error object names, to avoid any error shadowing.

modules/accounts.js
```diff
@@ -156,17 +158,36 @@ Accounts.prototype.setAccountAndGet = function(data, cb, tx) {
...
+						library.logger.error('Set account failed', err);
+						return reject(err);
+					}
+					return library.logic.account.get(
+						{ address },
+						(err, user) => {
+							if (err) {
```
nazarhussain: Same here.

modules/accounts.js
```diff
@@ -156,17 +158,36 @@ Accounts.prototype.setAccountAndGet = function(data, cb, tx) {
...
+				address,
+				data,
+				setAccountErr => {
+					if (setAccountErr) {
+						library.logger.error('Set account failed', setAccountErr);
+						return reject(setAccountErr);
+					}
+					return library.logic.account.get(
+						{ address },
+						(getAccountErr, user) => {
```
4miners: What is `user`? We're using this term nowhere in our application.

modules/accounts.js
```diff
@@ -156,17 +158,40 @@ Accounts.prototype.setAccountAndGet = function(data, cb, tx) {
...
-		tx
-	);
+	const task = t =>
+		new Promise((resolve, reject) => {
+			library.logic.account.set(
+				address,
+				data,
+				setAccountErr => {
+					if (setAccountErr) {
+						library.logger.error('Set account failed', setAccountErr);
```
nazarhussain: We should mention some detail of the account on which error occurred. That will be helpful for troubleshooting. 

> Set account failed for "<address>"

modules/accounts.js
```diff
@@ -156,17 +158,40 @@ Accounts.prototype.setAccountAndGet = function(data, cb, tx) {
...
+						return reject(setAccountErr);
+					}
+					return library.logic.account.get(
+						{ address },
+						(getAccountErr, account) => {
+							if (getAccountErr) {
+								library.logger.error('Get account failed', getAccountErr);
```
nazarhussain: Same here.

modules/accounts.js
```diff
@@ -156,17 +158,40 @@ Accounts.prototype.setAccountAndGet = function(data, cb, tx) {
...
+								return reject(getAccountErr);
+							}
+							return resolve(account);
+						},
+						t
+					);
+				},
+				t
+			);
+		});
```
nazarhussain: Is there any possibility to optimize the code by use of `async.waterfall`?

test/unit/db/repos/accounts.js
```diff
@@ -549,6 +549,41 @@ describe('db', () => {
...
+			it('should execute all queries in one database transaction', done => {
+				const account = accountFixtures.Account();
+
+				db.$config.options.query = function(event) {
+					if (!(event.ctx && event.ctx.isTX && event.ctx.txLevel === 0)) {
+						return done(
+							`Some query executed outside transaction context: ${event.query}`,
+							event
+						);
+					}
```
nazarhussain: Can we use expectations here as well. 

```
expect(event.ctx).to.no.null;
expect(event.ctx.isTx).to.be.true;
expect(event.ctx.txLevel).to.be.eql(0);
done();
```

Above will work but not sure how can we send custom error message in case of failure.

test/unit/db/repos/accounts.js
```diff
@@ -549,6 +549,41 @@ describe('db', () => {
 						});
 				});
 			});
+
+			it('should execute all queries in one database transaction', done => {
+				const account = accountFixtures.Account();
+
+				db.$config.options.query = function(event) {
+					if (!(event.ctx && event.ctx.isTX && event.ctx.txLevel === 0)) {
```
nazarhussain: Also add a check for transaction label `Accounts:setAccountAndGet`

test/unit/db/repos/accounts.js
```diff
@@ -549,6 +549,41 @@ describe('db', () => {
...
+
+						delete db.$config.options.connect;
+						delete db.$config.options.disconnect;
+						delete db.$config.options.query;
+
+						done();
+					})
+					.catch(err => {
+						done(err);
+					});
```
nazarhussain: You didn't tested here either `modules.accounts.setAccountAndGet` use one database transaction. Please add a test for that as well.

test/unit/db/repos/accounts.js
lsilvs: In this case the label will be `db:accounts:upsert`

test/unit/modules/accounts.js
```diff
@@ -147,6 +150,64 @@ describe('accounts', () => {
...
+
+			accounts.setAccountAndGet(account, (error, data) => {
+				expect(error).to.be.null;
+				expect(data.publicKey).to.be.eql(account.publicKey);
+				expect(data.address).to.exist;
+				done();
+			});
+		});
+
+		it('should be ok with pre existing database transaction', done => {
```
MaciejBaj: Please be more specific in tests expectations: "should be ok" => "should return no error and address"

PR #2508, opened 2018-10-26
Fix more false positives on our test suites - Closes #2480

### LINKED ISSUE #2480: Enforce Chai expects to have actual assertions
### Expected behavior

Find an Eslint rule or a new npm module as https://www.npmjs.com/package/mocha-eslint to enforce Chai expects to have assertions..

### Actual behavior

Some tests are giving false positives due to missing assertions. This fact slows down our development and create hidden bugs.

For reference:
https://github.com/LiskHQ/lisk/pull/2461/commits/e2fb91fdd2a1d78f933e803f6491f9e3dc3b4b78

`+1.1`

diego-G:
### What was the problem?

Not accurate expectations in unit tests could produce false positives and hide bugs. 

### How did I fix it?

I've added https://github.com/Turbo87/eslint-plugin-chai-expect as devDependency activating two of its rules: `missing-assertion` + `no-inner-compare`

### How to test it?

`npm run lint`
`npm test mocha:untagged:unit`

### Review checklist

* The PR resolves #2480 
* Linting passes
* Tests pass


test/unit/api/ws/rpc/disconnect.js
```diff
@@ -44,7 +44,7 @@ describe('disconnect', () => {
 		it('should call peer.socket.disconnect', () => {
 			disconnect(validPeer);
 			// The socket should be deleted.
-			expect(validPeer.socket, undefined);
+			expect(validPeer.socket.destroy).calledOnce;
```
shuse2: if the change should be equal, it should be
`expect(validPeer.socket).to.be.undefined`

test/unit/helpers/http_api.js
```diff
@@ -169,8 +169,9 @@ describe('httpApi', () => {
 				done();
 			});
 
-			it('should call logger.log with string "GET req/url from 127.0.0.1"', done => {
-				expect(loggerMock.log.calledWith('GET req/url from 127.0.0.1'));
+			it('should call logger.log with string "GET api/url from 127.0.0.1"', done => {
+				expect(loggerMock.log.calledWith('GET api/url from 127.0.0.1')).to.be
+					.true;
```
shuse2: `calledWith` doesn't have to be checked with true

test/unit/logic/delegate.js
```diff
@@ -907,13 +907,11 @@ describe('delegate', () => {
...
 					accountsMock.setAccountAndGet.calledWith({
 						address: sender.address,
-						u_isDelegate: 1,
 						isDelegate: 0,
 						vote: 0,
 						username: null,
-						u_username: transaction.asset.delegate.username,
 					})
-				);
+				).to.be.true;
```
shuse2: this should be 
`expect(accountsMock.setAccountAndGet).to.be.calledWith({})`

test/unit/logic/delegate.js
```diff
@@ -926,13 +924,11 @@ describe('delegate', () => {
...
```
shuse2: same as above.
`expect(accountsMock.setAccountAndGet).to.be.calledWith()`

test/unit/logic/delegate.js
```diff
@@ -945,11 +941,9 @@ describe('delegate', () => {
 					accountsMock.setAccountAndGet.calledWith({
```
shuse2: same as above

test/unit/logic/transaction.js
```diff
@@ -198,7 +198,7 @@ describe('transaction', () => {
 			inValidTransaction.senderPublicKey =
 				'c094ebee7ec0c50ebee32918655e089f6e1a604b83bcaa760293c61e0f18ab6fx';
 			try {
-				expect(transactionLogic.sign(keyPair, inValidTransaction));
+				expect(transactionLogic.sign(keyPair, inValidTransaction)).to.be.error;
```
shuse2: if it returns error it's
`expect(transactionLogic.sign(keyPair, inValidTransaction)).to.be.an('error')`
and if it throws then
`transactionLogic.sign.bind(null, keyPair, inValidTransaction).to.throw(Error, 'message')`

test/unit/logic/transaction_pool.js
```diff
@@ -911,15 +911,13 @@ describe('transactionPool', () => {
...
-					);
-					expect(res).to.deep.eql(transaction);
-					done();
-				})
-			);
+			transactionPool.receiveTransactions(transaction, {}, (err, res) => {
+				expect(err).to.eql(
+					'Transaction is already processed: 10431411423423423L'
+				);
+				expect(res).to.deep.eql(transaction);
```
shuse2: `expect(res).to.deep.equal(transaction);`

test/unit/logic/transaction_pool.js
```diff
@@ -929,13 +927,11 @@ describe('transactionPool', () => {
...
-			expect(
-				transactionPool.receiveTransactions(transaction, {}, (err, res) => {
-					expect(err).to.be.null;
-					expect(res).to.deep.eql(transaction);
-					done();
-				})
-			);
+			transactionPool.receiveTransactions(transaction, {}, (err, res) => {
+				expect(err).to.be.null;
+				expect(res).to.deep.eql(transaction);
```
shuse2: `expect(res).to.deep.equal(transaction);`

test/unit/logic/transaction_pool.js
diego-G: We don't have standard for this.

test/unit/logic/transaction_pool.js
diego-G: We don't have standard for this.

test/unit/api/ws/rpc/disconnect.js
diego-G: It doesn't work this way. I tried before opening the PR. Do you have any other idea? Actually I had some problems trying to understand this one.

test/unit/logic/transaction_pool.js
shuse2: By reading the documentation, it seems like `deep.eql` doesn't exist

test/unit/helpers/http_api.js
diego-G: Without `to.be.true` it is a false positive. Why do you say it doesn't have to be checked? Can you share the info about it?

test/unit/api/ws/rpc/disconnect.js
shuse2: by looking at the code, i guess original test was ambiguous i guess.
i think what you wrote `expect(validPeer.socket.destroy).calledOnce;` is ok, but prefer `expect(validPeer.socket.destroy).to.be.calledWithExactlyWith(1000, 'Intentionally disconnected from peer because of disconnect call');`

test/unit/logic/delegate.js
diego-G: Your suggestion gives the following error `TypeError: true is not a spy or a call to a spy!`

test/unit/logic/delegate.js
diego-G: Your suggestion gives the following error `TypeError: true is not a spy or a call to a spy!`

test/unit/logic/delegate.js
diego-G: same as above

test/unit/logic/transaction.js
diego-G: `to.throw` doesn't work but I take your suggestion to remove the expect from here.

test/unit/logic/transaction_pool.js
diego-G: I think it is an alias. I'm changing all the matches to `deep.equal` to make it standard. There are less results compare to `deep.equal`.

test/unit/helpers/http_api.js
shuse2: ah sorry what i mean is
`expect(loggerMock.log).to.be.calledWith('GET api/url from 127.0.0.1');`

test/unit/logic/delegate.js
```diff
@@ -907,13 +907,11 @@ describe('delegate', () => {
 				expect(
 					accountsMock.setAccountAndGet.calledWith({
 						address: sender.address,
-						u_isDelegate: 1,
```
4miners: Why we removed those?

test/unit/logic/transaction.js
```diff
@@ -194,15 +194,12 @@ describe('transaction', () => {
 
 		it('should throw an error Argument must be a valid hex string.', done => {
 			const inValidTransaction = Object.assign({}, validTransaction);
-			let err = null;
 			inValidTransaction.senderPublicKey =
 				'c094ebee7ec0c50ebee32918655e089f6e1a604b83bcaa760293c61e0f18ab6fx';
 			try {
-				expect(transactionLogic.sign(keyPair, inValidTransaction));
+				transactionLogic.sign(keyPair, inValidTransaction);
```
4miners: `invalidTransaction`, can we fix the name?

test/unit/logic/transaction.js
```diff
@@ -194,15 +194,12 @@ describe('transaction', () => {
...
 			} catch (e) {
-				err = e.message;
-			} finally {
-				expect(err).to.equal('Argument must be a valid hex string.');
+				expect(e.message).to.equal('Argument must be a valid hex string.');
```
4miners: Change of behavior, previously we were always expecting error, now we only expect error if it goes into `catch` block, is this correct?

4miners: It can lead to false-positive. Previously `expect` was in `finally` block, so it was always executed. Now the expectation will be checked only if there will be an error caught in `catch` block. If there is no error - the test will pass.
Maybe we should use something like `expect(transactionLogic.sign(keyPair, inValidTransaction)).to.throw(...);` and remove `try...catch` block?

test/unit/logic/delegate.js
4miners: I'm not sure, can you double check if those keys are needed?

test/integration/blocks/verify/verify.js
```diff
@@ -922,7 +922,7 @@ describe('blocks/verify', () => {
 					expect(block1.totalFee.isEqualTo('0')).to.be.true;
 					expect(block1.totalAmount.isEqualTo('0')).to.be.true;
 					expect(block1.payloadLength).to.equal(0);
-					expect(block1.transactions).to.deep.eql([]);
+					expect(block1.transactions).to.deep.equal([]);
```
ManuGowda: @diego-G what is the reason behind changing the `to.deep.eql` to `to.deep.equal` apart from explicit readability?
if there is any other reason I would be glad to know. And if we are changing `to.deep.eql` to `to.deep.equal` why not change it in other places where it is `required` like `to.eql` to `to.equal` consistently?

diego-G: It was @shuse2's suggestion. I agreed because the majority of the cases are as `to.deep.equal`. We should define these rules in the developer's guide and apply them all at once.

ManuGowda: Ok, we can create an issue to change the same across the test suite. I am gonna approve the PR for now.

diego-G: https://github.com/LiskHQ/lisk/issues/2532

PR #2511, opened 2018-10-29
Fix public key size generation for tests - Closes #2510

### LINKED ISSUE #2510: Account fixture generates public keys 32 chars long
### Expected behavior
In our test suite, Account fixture should generate public keys 64 chars long.

### Actual behavior
It generates public keys 32 chars long.

### Steps to reproduce
https://github.com/LiskHQ/lisk/blob/development/test/fixtures/accounts.js#L119

1.1

lsilvs:
### What was the problem?
In our test suite, Account fixture was generate public keys 32 chars long.

### How did I fix it?
Changed it to 64 chars instead

### How to test it?
Run the tests

### Review checklist

* The PR resolves #2510 
* Linting passes
* Tests pass

### DIFF
--- a/test/fixtures/accounts.js
+++ b/test/fixtures/accounts.js
@@ -119,7 +119,7 @@ const Account = stampit({
 		this.publicKey =
 			publicKey ||
 			randomstring
-				.generate({ charset: '0123456789ABCDE', length: 64 })
+				.generate({ charset: '0123456789ABCDEF', length: 64 })
 				.toLowerCase();
 
 		this.vote = randomstring.generate({ charset: '123456789', length: 5 });
--- a/test/fixtures/accounts.js
+++ b/test/fixtures/accounts.js
@@ -119,7 +119,7 @@ const Account = stampit({
 		this.publicKey =
 			publicKey ||
 			randomstring
-				.generate({ charset: '0123456789ABCDE', length: 32 })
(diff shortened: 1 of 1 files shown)


test/fixtures/accounts.js
```diff
@@ -119,7 +119,7 @@ const Account = stampit({
 		this.publicKey =
 			publicKey ||
 			randomstring
-				.generate({ charset: '0123456789ABCDE', length: 32 })
+				.generate({ charset: '0123456789ABCDE', length: 64 })
```
nazarhussain: For reference and explanation. This function generate a random hexadecimal string of 32 length. While two hex characters combined represent one byte, resulting a public a key of 16 bytes. While the protocol definition says it would be 256 bits means 32 bytes.

nazarhussain: @lsilvs I just noticed that we are not including `F` in the charset, hence not using the full character set to represent the random key. Can you please add that as well.

nazarhussain: And to be on safe side, we are including `lisk-elements` in tests. If we can use it to generate the valid public key for lisk-protocol that will be a safest way to do it.

lsilvs: I reckon this is out of the scope of this PR.

lsilvs: Done

PR #2512, opened 2018-10-30
Multi-signature transaction process after long delay - Closes #2489

### LINKED ISSUE #2489: Multi-signature transaction process after long delay
### Expected behavior
When a 1000 multi-signature transactions sent to process it should get confirmed within `>=40 and <=60` blocks.
### Actual behavior
Currently on alphanet `1.2.0-alpha.1` when we send 1000 multi-signature transactions as part of stress test a last 100/200 multi-signature transactions are processed after a long delay, the first 800 were processed with full blocks and the rest 200 takes a long time to process meaning several 100 blocks.

Reference:
Here is the end of processing muti-signature transaction http://alphanet-seed-01.liskdev.net:6040/block/770161081904199981
After a while it processed 25 transactions http://alphanet-seed-01.liskdev.net:6040/block/17369599963050370292
some more time later it processed again 25 http://alphanet-seed-01.liskdev.net:6040/block/17369599963050370292
and there are still 140 transactions in unsigned queue.
### Steps to reproduce
http://alphanet-seed-01.liskdev.net:6040/networkMonitor
login to any of the machine on `alphanet` the `debug` logs are enabled.

MaciejBaj:
### What was the problem?
Signatures were not being broadcasted due to changed behaviour of `checkConfirmed` function introduced in 1.2.0. The function is checking now if the property "id" exists on a passed parameter: 
```
checkConfirmed(transaction, cb) {
		if (!transaction || !transaction.id) {
			return setImmediate(cb, 'Invalid transaction id', false);
		}
```
Broadcaster though was calling the `checkConfirmed` function with both transactions and signatures (which don't have "id" property, but "transactionId" of the corresponding transaction to sign (https://github.com/LiskHQ/lisk/compare/1.2.0...2489-multisignature_broadcaster_fix?expand=1#diff-9e4ef2f035528df11706d74e5683125dL228)
### How did I fix it?
To decide, wheter to broadcast a signature or not, I am passing the `transactionId` to verify if a corresponding transaction exists in a transaction pool or is not being confirmed: https://github.com/LiskHQ/lisk/compare/1.2.0...2489-multisignature_broadcaster_fix?expand=1#diff-9e4ef2f035528df11706d74e5683125dR258

### How to test it?
Run Multisignature stress tests. Confirm that signatures are being propagated across the nodes.
### Review checklist

* The PR resolves #2489
* Linting passes
* Tests pass

### DIFF
--- a/logic/broadcaster.js
+++ b/logic/broadcaster.js
@@ -225,10 +225,15 @@ __private.filterQueue = function(cb) {
 			if (broadcast.options.immediate) {
 				return setImmediate(filterCb, null, false);
 			} else if (broadcast.options.data) {
-				const transaction =
-					broadcast.options.data.transaction ||
-					broadcast.options.data.signature;
-				return __private.filterTransaction(transaction, filterCb);
+				let transactionId;
+				if (broadcast.options.data.transaction) {
+					// Look for a transaction of a given "id" when broadcasting transactions
+					transactionId = broadcast.options.data.transaction.id;
+				} else if (broadcast.options.data.signature) {
+					// Look for a corresponding "transactionId" of a given signature when broadcasting signatures
+					transactionId = broadcast.options.data.signature.transactionId;
+				}
+				return __private.filterTransaction(transactionId, filterCb);
 			}
 			return setImmediate(filterCb, null, true);
 		},
@@ -242,21 +247,21 @@ __private.filterQueue = function(cb) {
 };
 
 /**
- * Checks if transaction is in pool or confirm it.
+ * Checks if transaction of a given id is in pool or confirmed in database "trs" table.
  *
  * @private
- * @param {transaction} transaction
+ * @param {string} transactionId
  * @param {function} cb
  * @returns {SetImmediate} null, boolean
  * @todo Add description for the params
  */
-__private.filterTransaction = function(transaction, cb) {
-	if (transaction !== undefined) {
-		if (modules.transactions.transactionInPool(transaction.id)) {
+__private.filterTransaction = function(transactionId, cb) {
+	if (transactionId !== undefined) {
+		if (modules.transactions.transactionInPool(transactionId)) {
 			return setImmediate(cb, null, true);
 		}
 		return library.logic.transaction.checkConfirmed(
-			transaction,
+			{ id: transactionId },
 			(err, isConfirmed) => setImmediate(cb, null, !err && !isConfirmed)
 		);
 	}
--- a/logic/broadcaster.js
+++ b/logic/broadcaster.js
@@ -249,7 +249,7 @@ __private.filterQueue = function(cb) {
 					// err = null, isConfirmed = true => return false
 					// In case transaction doesn't exists in "trs" table:
 					// err = null, isConfirmed = false => return true
-					(err, isConfirmed) => !err && !isConfirmed
+					(err, isConfirmed) => filterCb(null, !err && !isConfirmed)
 				);
 			}
 			return setImmediate(filterCb, null, true);
--- a/logic/broadcaster.js
+++ b/logic/broadcaster.js
@@ -233,7 +233,24 @@ __private.filterQueue = function(cb) {
 					// Look for a corresponding "transactionId" of a given signature when broadcasting signatures
 					transactionId = broadcast.options.data.signature.transactionId;
 				}
-				return __private.filterTransaction(transactionId, filterCb);
+				if (!transactionId) {
+					return setImmediate(filterCb, null, false);
+				}
+				// Broadcast if transaction is in transaction pool
+				if (modules.transactions.transactionInPool(transactionId)) {
+					return setImmediate(filterCb, null, true);
+				}
+				// Don't broadcast if transaction is already confirmed
+				return library.logic.transaction.checkConfirmed(
+					{ id: transactionId },
+					// In case of SQL error:
+					// err = true, isConfirmed = false => return false
+					// In case transaction exists in "trs" table:
+					// err = null, isConfirmed = true => return false
+					// In case transaction doesn't exists in "trs" table:
+					// err = null, isConfirmed = false => return true
+					(err, isConfirmed) => !err && !isConfirmed
+				);
 			}
 			return setImmediate(filterCb, null, true);
 		},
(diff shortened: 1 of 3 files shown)


logic/broadcaster.js
```diff
@@ -242,21 +247,21 @@ __private.filterQueue = function(cb) {
...
- * @param {transaction} transaction
+ * @param {string} transactionId
  * @param {function} cb
  * @returns {SetImmediate} null, boolean
  * @todo Add description for the params
  */
-__private.filterTransaction = function(transaction, cb) {
-	if (transaction !== undefined) {
-		if (modules.transactions.transactionInPool(transaction.id)) {
+__private.filterTransaction = function(transactionId, cb) {
```
ManuGowda: The function name doesn't do what it says, it's not filtering the transaction rather it just checks if the transaction exists in transaction pool or transaction exists in the database, so it would be good to rename it to something meaningful like `isTransactionExistsInPoolOrDatabase`

logic/broadcaster.js
```diff
@@ -242,21 +247,21 @@ __private.filterQueue = function(cb) {
...
+	if (transactionId !== undefined) {
```
ManuGowda: no need for explicit check, we can use `if (transactionId) {` this will ensure both `null` and `undefined`.

logic/broadcaster.js
```diff
@@ -242,21 +247,21 @@ __private.filterQueue = function(cb) {
...
+		if (modules.transactions.transactionInPool(transactionId)) {
 			return setImmediate(cb, null, true);
 		}
 		return library.logic.transaction.checkConfirmed(
-			transaction,
+			{ id: transactionId },
 			(err, isConfirmed) => setImmediate(cb, null, !err && !isConfirmed)
```
ManuGowda: I have a concern here if the transaction is confirmed, then `library.logic.transaction.checkConfirmed` returns `null, true` and following part of logic `setImmediate(cb, null, !err && !isConfirmed)` negates and returns `null, false` which should return `null, true` if the transaction exists in database, correct?

MaciejBaj: The logic works fine.

logic/broadcaster.js
```diff
@@ -242,21 +247,21 @@ __private.filterQueue = function(cb) {
...
-	if (transaction !== undefined) {
-		if (modules.transactions.transactionInPool(transaction.id)) {
+__private.filterTransaction = function(transactionId, cb) {
+	if (transactionId !== undefined) {
+		if (modules.transactions.transactionInPool(transactionId)) {
 			return setImmediate(cb, null, true);
 		}
 		return library.logic.transaction.checkConfirmed(
-			transaction,
+			{ id: transactionId },
```
4miners: So `checkConfirmed` is expecting a transaction object, now from here we are sending an object with just `id` property. This creates inconsistency and can be an issue in the future. Consider changing this or modify documentation of `checkConfirmed` to explicitly say it's not a transaction object.

logic/broadcaster.js
```diff
@@ -225,10 +225,15 @@ __private.filterQueue = function(cb) {
...
-				return __private.filterTransaction(transaction, filterCb);
+				let transactionId;
+				if (broadcast.options.data.transaction) {
+					// Look for a transaction of a given "id" when broadcasting transactions
+					transactionId = broadcast.options.data.transaction.id;
+				} else if (broadcast.options.data.signature) {
+					// Look for a corresponding "transactionId" of a given signature when broadcasting signatures
+					transactionId = broadcast.options.data.signature.transactionId;
+				}
+				return __private.filterTransaction(transactionId, filterCb);
```
nazarhussain: What would be the `else` case? Why even call the `__private.filterTransaction` if there is no `transactionId`.

logic/broadcaster.js
nazarhussain: Agree to Manu to rename this function to something `isTransactionExistsInPoolOrDatabase`

logic/broadcaster.js
nazarhussain: Can we at least add a comment line explaining that logic used here. For future reference.

logic/broadcaster.js
MaciejBaj: Not really in the scope of the bug fix I hope, but I agree, I will rename on 2 corresponding functions.

MaciejBaj:
@diego-G I did not change the code of logic/transaction.js. `broadcaster.js` doesn't have a unit test coverage.

MaciejBaj:
@diego-G my bad - there are broadcaster tests. I will fix them.

logic/transaction.js
```diff
@@ -279,7 +279,8 @@ class Transaction {
 	/**
 	 * Returns true if a transaction was confirmed.
 	 *
-	 * @param {transaction} transaction
+	 * @param {Object} transaction
+	 * @param {string} transaction.id - only transaction id is necessary here
```
diego-G: There should be a single parameter.

test/unit/logic/broadcaster.js
```diff
@@ -331,79 +316,285 @@ describe('Broadcaster', () => {
...
+								immediate: false,
+							},
+						}
+					)
+				);
 				done();
 			});
+
+			describe('when all of them exist in transaction pool', () => {
+				beforeEach(done => {
```
diego-G: Do we really need to aisle this line inside a different beforeEach statement? Why?

logic/broadcaster.js
```diff
@@ -225,10 +225,32 @@ __private.filterQueue = function(cb) {
...
+				let transactionId;
+				if (broadcast.options.data.transaction) {
+					// Look for a transaction of a given "id" when broadcasting transactions
+					transactionId = broadcast.options.data.transaction.id;
+				} else if (broadcast.options.data.signature) {
+					// Look for a corresponding "transactionId" of a given signature when broadcasting signatures
+					transactionId = broadcast.options.data.signature.transactionId;
+				}
+				if (!transactionId) {
+					return setImmediate(filterCb, null, false);
```
diego-G: Can we write a comment of why/when we might arrive here? Is it possible to have anything else in the queues rather than transactions or signatures?

logic/transaction.js
MaciejBaj: Don't want to interfere in the other code changes. Mind that I didn't change any single line of that function.

test/unit/logic/broadcaster.js
MaciejBaj: The describe description corresponds with a context that is being enforced by `beforeEach` statements. Like here: 
```
describe('having many transaction and signatures broadcasts in queue')
   beforeEach is setting many broadcasts to queue
   describe('when all of them exist in transaction pool')
      beforeEach is making transaction pool to return true always
```
The same context for "having many transactions and signatures broadcasts in queue'" might be extended by additional contexts like "when some of them are confirmed". I didn't go so far.

logic/broadcaster.js
MaciejBaj: I don't think it's necessary in that case. The line will be reached when during broadcast we set: `broadcast.options.data.transaction.id` or `broadcast.options.data.transaction.id` to null/undefined/0 or we will try to filter queue with different keys than `transaction` or `signature`.

logic/transaction.js
diego-G: But this code change comes from a commit of yours https://github.com/LiskHQ/lisk/pull/2512/commits/9826bf8c483cd473fcc19b3f7b3b8e8df7afd0b1

nazarhussain:
@MaciejBaj We should run a special test in QA cycle to explicitly test this change.

PR #2513, opened 2018-10-31
Allow fuzzy search on data filter when getting transactions - Closes #2404

### LINKED ISSUE #2404: Allow to search for transactions by data field

diego-G:
### What was the problem?

Query was returning only exact matches.

### How did I fix it?

Changing the subquery to support `LIKE` matches and removing null characters from the string to avoid errors. 

### How to test it?

`mocha test/functional/http/get/transactions.js`

### Review checklist

* The PR resolves #2404 
* Linting passes
* Tests pass


modules/transactions.js
```diff
@@ -103,8 +103,9 @@ __private.list = function(filter, cb) {
 	const params = {};
 	const where = [];
 
+	// Removing null characters from data filter to avoid errors in pg-promise library
 	if (filter.data) {
-		params.additionalData = Buffer.from(filter.data, 'utf8');
+		filter.data = filter.data.replace('\0', '');
```
vitaly-t: This will remove only a single occurrence of the symbol.

modules/transactions.js
```diff
@@ -103,8 +103,9 @@ __private.list = function(filter, cb) {
 	const params = {};
 	const where = [];
 
+	// Removing null characters from data filter to avoid errors in pg-promise library
 	if (filter.data) {
-		params.additionalData = Buffer.from(filter.data, 'utf8');
+		filter.data = filter.data.replace(/\0/g, '');
```
4miners: Now, when we don't convert the field to text in the query, can we pass `Buffer` instance here?

modules/transactions.js
```diff
@@ -103,8 +103,9 @@ __private.list = function(filter, cb) {
 	const params = {};
 	const where = [];
 
+	// Removing null characters from data filter to avoid errors in pg-promise library
 	if (filter.data) {
-		params.additionalData = Buffer.from(filter.data, 'utf8');
+		filter.data = Buffer.from(filter.data);
```
4miners: We don't need `'utf8'` here?

diego-G: By default it's 'utf8' but I add it for clarity.

schema/swagger.yml
```diff
@@ -938,6 +937,13 @@ paths:
             - timestamp:asc
             - timestamp:desc
           default: amount:asc
+        - name: data
+          in: query
+          description: Fuzzy additional data field to query
+          type: string
+          format: additionalData
+          minLength: 1
+          maxLength: 64
```
nazarhussain:
```suggestion
          maxLength: 64  # Its an additional check, actual check is 64 bytes
```

PR #2515, opened 2018-10-31
Stop receiving blocks during snapshotting - Closes #2419

### LINKED ISSUE #2419: Node receives blocks during snapshotting
### Expected behavior
The node should not receive blocks/transaction during snapshotting process.
### Actual behavior
Node receives blocks during snapshotting
```
[inf] 2018-09-25 08:23:19 | Verify->verifyBlock succeeded for block 8323462238787390927 at height 1616.
[inf] 2018-09-25 08:23:20 | Rebuilding accounts states, current round: 17, height: 1617
[WRN] 2018-09-25 08:23:20 | Discarded block that does not match with current chain: 13500374753101519740 height: 6298816 round: 62365 slot: 7375460 generator: 6a8d02899c66dfa2423b125f44d360be6da0669cedadde32e63e629cb2e3195c
[inf] 2018-09-25 08:23:20 | Verify->verifyBlock succeeded for block 8197888260741386974 at height 1617.
```
### Steps to reproduce
Run snapshotting on node which is running under devnet/testnet/mainnet (`1.1.0`)

diego-G:
### What was the problem?

Socket Cluster was initiated in snapshotting mode but it is not used at all.

### How did I fix it?

If snapshot flag is required several config sections are turned off.  

### How to test it?

`node app -n testnet -s 20`

### Review checklist

* The PR resolves #2419
* Linting passes
* Tests pass


helpers/config.js
```diff
@@ -164,6 +164,15 @@ function Config(packageJson, parseCommandLineOptions = true) {
...
 		validateForce(appConfig);
 
+		// Disable api, peers, broadcasts, syncing and ws workers when snapshotting mode required
+		if (program.snapshot) {
+			appConfig.api.enabled = false;
+			appConfig.peers.enabled = false;
+			appConfig.peers.list = [];
+			appConfig.broadcasts.active = false;
+			appConfig.syncing.active = false;
+			appConfig.wsWorkers = 0;
```
nazarhussain: I really don't like setting wsWokers here. Or overriding the peers list. 

Also please put some logging here for user that you are overriding or disabling these values for snapshotting.

app.js
```diff
@@ -745,6 +757,9 @@ d.run(() => {
 				 * @param {function} cb - Callback function
 				 */
 				function(scope, cb) {
+					if (parseInt(appConfig.wsWorkers) < 1) {
+						return cb();
+					}
```
nazarhussain: I would suggest to implement this logic in other way. 

For now we have following steps in `async.auto`

- config
- logger
- build
- lastCommit
- geensisBlock 
- schema
- network 
- sequence 
- balanceSequence
- swagger
- ed
- bus
- db
- cache
- webSocket 
- logic
- modules 
- ready 
- listenWebSocket
- listenHttp 

So during snapshotting what we really don't need is not to run steps `listenWebSocket` and `listenHttp`. 

If we dig deeper, we don't even want to initialize web server or socket cluster, but that's a way ahead of the thought. 

I would suggest just change in those two steps `listenWebSocket` and `listenHttp` and make that conditional  based on `snapshotting` not any other config attribute.

helpers/config.js
diego-G: All these parameters affect the snapshot mode. They were present but not used till now.

Regarding the logging suggestion, the user requests a snapshot and it's been notified via https://github.com/LiskHQ/lisk/blob/d4736c1f05a8c87e2e25d6cb273c6ac07361e4f5/modules/loader.js#L766 .  i don't think we need to add anything else than a code comment.

app.js
diego-G: I tried your approach and the application keeps trying to update peers and releasing broadcast queue. My approach is suitable for another purposes.

Anyways, I've turned off `listenHttp` as well if snapshot is requested or api was disabled.

app.js
```diff
@@ -506,6 +512,12 @@ d.run(() => {
 				 * @todo Add description for the function and its params
 				 */
 				function(scope, cb) {
+					if (parseInt(appConfig.wsWorkers) < 1) {
```
MaciejBaj: That's a wrong assumption. The number of workers should stay independent. I would prefer to re-use the config.peers.enabled for that purpose

app.js
```diff
@@ -506,6 +512,12 @@ d.run(() => {
 				 * @todo Add description for the function and its params
 				 */
 				function(scope, cb) {
+					if (!appConfig.peers.enabled) {
+						scope.logger.info(
```
4miners: Is `scope.logger` available? Should it be just `logger`?

diego-G: It is available.

helpers/config.js
```diff
@@ -164,6 +164,14 @@ function Config(packageJson, parseCommandLineOptions = true) {
 
 		validateForce(appConfig);
 
+		// Disable api, peers, broadcasts, syncing and ws workers when snapshotting mode required
+		if (program.snapshot) {
```
MaciejBaj: 👍

app.js
```diff
@@ -506,6 +512,12 @@ d.run(() => {
...
+							'Socket Cluster was NOT intiatied due to peers layer is disabled.'
```
MaciejBaj: Skipping P2P server initialization due to the config settings - "peers.enabled" is set to false.
```suggestion
							'Socket Cluster was NOT intiatied due to peers layer is disabled.'
```

helpers/config.js
```diff
@@ -171,6 +171,14 @@ function Config(packageJson, parseCommandLineOptions = true) {
...
+			appConfig.api.enabled = false;
+			appConfig.peers.enabled = false;
+			appConfig.peers.list = [];
+			appConfig.broadcasts.active = false;
+			appConfig.syncing.active = false;
+		}
```
nazarhussain: I don't like overriding config attributes without telling user in the logs that some of his config attributes are override. Consider the fact user is using config some pm2 config file and actually not aware of it the flags in there. He can't see his API working, he came and check the config file and will see its there and set to true. Then he will get confused why API is not working.

PR #2519, opened 2018-11-01
Balance change after new block is delayed by several seconds - Closes #2403

### LINKED ISSUE #2403: Balance change after new block is delayed by several seconds
### Steps to reproduce

1. Connect to socket.io of Lisk Core, e.g. `wss://testnet.lisk.io/socket.io/?EIO=3&transport=websocket`
2. Send a transaction from account A to account B
3. Wait for the `'blocks/change'` socket event
4. Make RESTfull API Request to `/api/accounts?address=<address_of_A_or_B>`

### Expected behavior
`/api/accounts?address=<address_of_A_or_B>` should contain balance updated by transaction done in step 2.

### Actual behavior
`/api/accounts?address=<address_of_A_or_B>` contains balance unchanged

The balance will be changed only several seconds later.

Which causes issues in Lisk Hub like this https://github.com/LiskHQ/lisk-hub/issues/1188

`/api/transactions` has the same delay as `/api/accounts`

Another thing I noticed is that if the transaction in Step 2. is the first outgoing transaction of account A, then response in Step 4 already contains the publicKey, but still not the updated balance.

Lisk Core 1.0.0

lsilvs:
### What was the problem?
New blocks were been broadcasted before cache for blocks and transactions been cleared.

### How did I fix it?
Changed the `broadcastBlock` operation to be executed after `applyBlock`

### How to test it?
Run the tests.
Follow the steps described in the issue #2403 

### Review checklist

* The PR resolves #2403 
* Linting passes
* Tests pass


SargeKhan:
This PR is changing the behavior of broadcasting of blocks. By which I mean that earlier we were broadcasting blocks even if the node did not find the block to be valid whereas after this change we node will only broadcast blocks which it considers valid. This change may affect how the network will behave during fork recovery.

lsilvs:
Good point @SargeKhan 
I'll come up with something else then.

lsilvs:
As discussed with @4miners and @nazarhussain , just adding `onNewBlock` inside Block module is not enough as we can't guarantee that the cache will be cleared before emitting `blocks/change` message.

The solution was to refactor `onNewBlock` inside Cache module to a regular method and call it from `onNewBlock` inside Block module making sure we emit `blocks/change` message after the cache function is finished.

modules/blocks.js
```diff
@@ -228,12 +230,32 @@ Blocks.prototype.isCleaning = {
...
 
+/**
+ * Emit socket notification `blocks/change`.
+ *
+ * @param {Block} block
+ * @todo Add description for the params
+ * @todo Add @returns tag
+ */
+Blocks.prototype.onNewBlock = function(block) {
+	modules.cache.clearBlockRelatedCache(err => {
```
nazarhussain: There is already two methods `removeByPattern` and `deleteJsonForKey` in general. I would rather suggest to develop a generic method `clearCacheFor(scope)`, scope can be `blocks`, `transactions` or in future any thing. This way we can make that single method useable on different cases.

modules/cache.js
```diff
@@ -404,6 +402,9 @@ Cache.prototype.onSyncFinished = function() {
 
 Cache.prototype.KEYS = {
 	transactionCount: 'transactionCount',
+	blocksApi: '/api/blocks*',
+	transactionsApi: '/api/transactions*',
+	delegatesApi: '/api/delegates*',
```
4miners: Where this is used?

lsilvs: You mean the `delegatesApi`?
If so, I meant to use it in `onFinishRound` and `onTransactionsSaved` to keep consistency but it seems I've forgotten.
Good catch! I will added it.

4miners: Yep, thanks.

MaciejBaj:
It might solve the #955 as well.

PR #2523, opened 2018-11-02
Upgrade to Lisk Core 1.2.0 changes whitelisted local ip in config to 'null' - Closes #2520

### LINKED ISSUE #2520: Upgrade to Lisk Core 1.2.0 changes whitelisted local ip in config to 'null'
### Expected behavior
Local ip `127.0.0.1` under whitelisted sections should stay intact after upgrading to 1.2.0 using the installLisk.sh script.

### Actual behavior
Local ip `127.0.0.1` gets changed to `null` in both whitelisted sections of the `config.json` using the installLisk.sh script. This results in a situation where you can't enable forging and/or use the API.

### Steps to reproduce
Perform Core upgrade to 1.2.0 using installLisk.sh

cd ~
rm -f installLisk.sh
wget https://downloads.lisk.io/lisk/test/installLisk.sh
bash installLisk.sh upgrade -r test

Ubuntu 16.04 LTS. Upgrade done from 1.1.1-rc.1.

nazarhussain:
### What was the problem?

During update process, if there is no change in the array element, it didn't applied the original value in new custom config. Causing to have a `null` value at particular index. 

### How did I fix it?

Update the config to preserve original value during the migration of array elements. 

### How to test it?

Run the upgrade scripts. 

### Review checklist

* The PR resolves #2520
* Linting passes
* Tests pass

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -6,6 +6,7 @@ pids/*
 release/
 scripts/**/*
 !scripts/update_config.js
+!scripts/generate_config.js
 tasks/
 test/
 Grunt*
--- a/scripts/update_config.js
+++ b/scripts/update_config.js
@@ -287,9 +287,11 @@ history.migrate(
 			// if there is a change in array element we want to preserve it as well to original value
 			const changeInArrayElement = d.kind === 'A';
 
-			if (changeInArrayElement || changeInDeepObject) {
-				const path = _.clone(d.path);
+			const path = _.clone(d.path);
 
+			if (changeInArrayElement) {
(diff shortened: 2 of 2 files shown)


fchavant:
The `null` value does not appear any more but `"public": false,` now appeats out of nowhere.

```
--- a/config.json 2018-11-02 09:55:41.097067938 +0000
+++ b/config.json   2018-11-02 09:56:55.411844813 +0000
@@ -1,8 +1,9 @@
 {
        "api": {
                "access": {
+                       "public": false,
                        "whiteList": [
-                               null,
+                               "127.0.0.1",
                                "127.0.0.2"
                        ]
                }
```

PR #2530, opened 2018-11-06
Add ready and signatures parameters as required to all unsigned transactions - Closes #1695

### LINKED ISSUE #1695: Add ready and signatures parameters as required to all unsigned transactions
### Expected behavior
No matter the type or the state of a transaction, if it comes to the transaction pool, `ready` and `signatures` parameters must be presented. Default values should be:
- `ready` : `false`
- `signatures` : `[]` 

### Actual behavior
There are three different cases:

```
{
    type: 4,
    amount: '0',
    fee: '1500000000',
    senderPublicKey: 'f58f864c826134f46a66849761f0c73fb56005909a70cf188d0638be473d27c2',
    timestamp: 56320321,
    asset: [Object],
    signature: '6c05e390363331725fc7409596c5c09272a5cfd25cdf8f891b38b283012c1a1f7e817a80585eec070e89c9414684ccbd9578f83944a5ee55d891b20803c54b0f',
    id: '16882190972607543218',
    senderId: '15884693486361216910L',
    relays: 1,
    receivedAt: 2018-03-07T13:32:13.689Z,
    recipientId: '',
    recipientPublicKey: '',
    signatures: undefined 
},
{
    type: 4,
    amount: '0',
    fee: '1500000000',
    senderPublicKey: '31b1a91072f54346112c4d58487fdcb91434eaa87a1ad057fc3ff97483fdc017',
    timestamp: 56320321,
    asset: [Object],
    signature: '03b5a8efd16be3764674c768eecb7f21a5c1cbf8d97a1455accd87609bf87874d3a9b8333e5ae093ee88cb7d398dee77b3524a52d124b914377cbc3cf079a307',
    id: '3531016181021663084',
    senderId: '4048527204793426020L',
    relays: 1,
    receivedAt: 2018-03-07T13:32:21.937Z,
    ready: false,
    recipientId: '',
    recipientPublicKey: '',
    signatures: [Object] 
},
{ 
    type: 0,
    amount: '284635893',
    fee: '10000000',
    recipientId: '1243293569494342985L',
    senderPublicKey: '965192948134f2993a022aecee29e299d7e2622c6905c210e5dd35fe16259351',
    timestamp: 56320392,
    asset: {},
    signature: '2d90771ba331ad50dbf2f9434c2080943e0795b3d2357b1dc71c4c070c2549f1c48d5e21c08f3aca69b7650ec73294f69d3d91d9d340eabefbdfc1f8649ef008',
    signSignature: '598d011d1d4139fdfafe15720a518e9f829d0b52c4e7f1d62a45f0f5f675a0720de2ab604d273e1ad7ccf18814ba0af669b11c354c9e10a1da59e67f4514fb03',
    id: '1773445611140144002',
    senderId: '17052822151617774790L',
    relays: 1,
    receivedAt: 2018-03-07T13:33:12.181Z,
    recipientPublicKey: '',
    signatures: [] 
}
```

### Steps to reproduce
```
mocha test/functional/http/post/4.multisignature.js
mocha test/functional/http/get/node/transactions_unsigned.js 
curl -X GET 'http://localhost:4000/api/node/transactions/unsigned`
```

`+1.0.0`

pablitovicente:
### What was the problem?

Multi-signature transactions sent with invalid payloads.

### How did I fix it?

By applying logic in the `objectNormalize` method for multi-signature's logic.

### How to test it?

Check tests in `4.multisignature.advance.js` for obtaining sample transactions.

### Review checklist

* The PR resolves #1695
* Linting passes
* Tests pass


logic/multisignature.js
```diff
@@ -520,6 +520,18 @@ Multisignature.prototype.schema = {
  * @returns {transaction} Validated transaction
  */
 Multisignature.prototype.objectNormalize = function(transaction) {
+	if (
```
4miners: There is only one place when we set this `ready` flag:
https://github.com/LiskHQ/lisk/blob/4e4a344e0313fa21dd58fa14b56e96dd1e78f3db/modules/multisignatures.js#L172
It's calling this function:
https://github.com/LiskHQ/lisk/blob/4e4a344e0313fa21dd58fa14b56e96dd1e78f3db/logic/multisignature.js#L584-L599
The logic there is different. Also, there should be a single place where we set this `ready` flag, `objectNormalize` is not the proper place. But according to the issue we need to set default values only for `ready` and `signatures`. So what just not add here:
```
transaction.ready = false;
transaction.signatures = Array.isArray(transaction.signatures) ? transaction.signatures : [];
```

pablitovicente: You can not always set `transaction.ready = false` because that might cause offline transactions to get stuck in the transaction pool. If you send a multi-signature transaction with all signatures present and set the ready property to false it will never be removed from the pool.

pablitovicente: The logic you suggest to use from https://github.com/LiskHQ/lisk/blob/4e4a344e0313fa21dd58fa14b56e96dd1e78f3db/logic/multisignature.js#L584-L599 expects a transaction that does not presents some edge cases as the one mentioned above - though I did thought of using that function done in the objectNormalize makes sure the transaction pool always gets good data -

4miners: So then the logic needs to be fixed, `transaction.ready` flag should be never set by the user. Maybe weould be nice to set this flag to `true` here? https://github.com/LiskHQ/lisk/blob/4e4a344e0313fa21dd58fa14b56e96dd1e78f3db/logic/multisignature.js#L168

pablitovicente: It needs to be set for the case of offline transactions, that is the case when a user sends a transaction which includes the signatures. I think we could iterate over the transaction pool logic in a different issue. 

This method you suggest also looks like a good place for implementing this check but I think this PR is related to sanitation or recovery rather than verification of the final transaction.

4miners: I still think that the approach proposed in this PR is not right, `objectNormalize` is a function for making transactions objects consistent, it should not make any decisions.

pablitovicente: You've got a good point with regards to `objectNormalize` will figure out an alternative to the current solution.

pablitovicente:
@diego-G I'm preparing a new commit as I was requested changes before to remove changes to objectNormalize ready is only used by type 4 transactions and we need to accept it as we have one case that is valid for it to be and we in fact need it to be there (case when all signatures are present).

pablitovicente:
@4miners @pablitovicente @nazarhussain after exchanging more ideas with Diego I will probably be making some more changes so hold the review for now please.

pablitovicente:
@diego-G @4miners @nazarhussain could you please review again?

pablitovicente:
I did not modify signatures[] as it's managed by processVerifyTransaction()

logic/transaction.js
```diff
@@ -1209,6 +1212,18 @@ class Transaction {
 
 		return transaction;
 	}
+
```
diego-G: Do we really need this method here? I reckon it is redundant. It is just calling ready...

test/functional/http/post/4.multisignature.advanced.js
```diff
@@ -69,7 +70,7 @@ describe('POST /api/transactions (type 4) register multisignature', () => {
 	describe('transactions processing', () => {
 		describe('signatures property', () => {
 			describe('correctly offline signed transaction', () => {
-				it('using ready should be ok but never confirmed', () => {
+				it('using ready false should be corrected and confirmed', () => {
```
diego-G: Improve description stating if the transaction contains all the signatures or it is just partially signed.

test/functional/http/post/4.multisignature.advanced.js
```diff
@@ -341,6 +316,47 @@ describe('POST /api/transactions (type 4) register multisignature', () => {
 						badTransactions.push(scenario.multiSigTransaction);
 					});
 				});
+
+				it('using ready false should be ok and be corrected and processed', () => {
```
diego-G: Description is not clear.

logic/transaction.js
pablitovicente: The ready method only returns a boolean this method uses the result of that logic to set the correct value. We need it.

diego-G: I still believe if we don't introduce any additional logic to the new function, there is no need to add it. Please, remove it but keep the unit test for `ready` function.

pablitovicente: Good point, for now that method was sanitising only the ready method so it make sense to just use ready() directly.

test/functional/http/post/4.multisignature.advanced.js
pablitovicente: Done

pablitovicente:
@diego-G moved the tests to each section and pushed to correct expected queues please check.

test/functional/http/post/4.multisignature.advanced.js
```diff
@@ -33,11 +33,15 @@ describe('POST /api/transactions (type 4) register multisignature', () => {
 		offline_signed_empty_signatures: new Scenarios.Multisig(),
 		offline_signed_without_ready: new Scenarios.Multisig(),
 		offline_signed_with_ready_false: new Scenarios.Multisig(),
+		unsigned_with_ready_true: new Scenarios.Multisig(),
 		offline_signed_with_ready_true: new Scenarios.Multisig(),
 		duplicated_signature: new Scenarios.Multisig(),
 		extra_signature: new Scenarios.Multisig(),
 		unknown_signature: new Scenarios.Multisig(),
 		requesterPublicKey: new Scenarios.Multisig(),
+		all_signatures_ready_false: new Scenarios.Multisig(),
```
diego-G: These cases are in conflict with the ones at the beginning starting by `offline_signed*`.

test/functional/http/post/4.multisignature.advanced.js
```diff
@@ -341,6 +308,69 @@ describe('POST /api/transactions (type 4) register multisignature', () => {
...
+						res => {
+							expect(res.body.data.message).to.be.equal(
+								'Transaction(s) accepted'
+							);
+							pendingMultisignatures.push(scenario.multiSigTransaction);
+						}
+					);
+				});
+
+				it('Some signatures present, ready set to true, should be corrected and be pending', () => {
```
diego-G:
```suggestion
				it('Some signatures present, ready set to true, should be corrected and remain pending', () => {
```

test/functional/http/post/4.multisignature.advanced.js
```diff
@@ -341,6 +308,69 @@ describe('POST /api/transactions (type 4) register multisignature', () => {
...
+						res => {
+							expect(res.body.data.message).to.be.equal(
+								'Transaction(s) accepted'
+							);
+							goodTransactions.push(scenario.multiSigTransaction);
+						}
+					);
+				});
+
+				it('No signatures present, ready set to true, should be corrected and be be pending', () => {
```
diego-G:
```suggestion
				it('No signatures present, ready set to true, should be corrected and remain pending', () => {
```

test/unit/logic/test_data/multisignature.js
```diff
@@ -89,6 +90,97 @@ var validTransaction = {
 	},
 };
 
+var invalidAllSignaturesReadyFalse = {
```
diego-G: Is it truly invalid? If it is, please add a comment on the invalid parameter.

test/unit/logic/test_data/multisignature.js
```diff
@@ -89,6 +90,97 @@ var validTransaction = {
...
+			keysgroup: [
+				'+bd6d0388dcc0b07ab2035689c60a78d3ebb27901c5a5ed9a07262eab1a2e9bd2',
+				'+addb0e15a44b0fdc6ff291be28d8c98f5551d0cd9218d749e30ddb87c6e31ca9',
+			],
+		},
+	},
+	ready: false,
+};
+
+var invalidSomeSignaturesReadyTrue = {
```
diego-G: Same with the rest.

test/unit/logic/test_data/multisignature.js
diego-G: Ok, I get it. Ty

test/unit/logic/test_data/multisignature.js
pablitovicente: @diego-G it's invalid in the sense that if some signatures, but not all, are present and valid ready should be false. So if a transaction is sent with some valid signatures and ready: true it is an invalid payload but the code will fix it now.

pablitovicente: Sending an update for this one as the mock in did had all signatures (code follows the same path as for `invalidAllSignaturesReadyFalse ` so it's only the mock that's misleading)

test/functional/http/post/4.multisignature.advanced.js
pablitovicente: They do differ in the addition of the ready property for testing the cases when it is present in the transaction.

pablitovicente:
@diego-G 

- Description made similar 
- Added the setting of signatures to an empty array when it is not present (should it be only for accounts from multi-signature? Issue does not clarify it)
- "Transaction type 4 partially signed with ready true and ready false. It has to remain in the queue showing a signatures array not empty." there are already test for partially signed signatures ('using some signatures, setting ready to true, should be corrected and remain pending') 
- Working on tests cases for regular transfer from multi-signature account

logic/transaction.js
```diff
@@ -525,6 +529,12 @@ class Transaction {
 			}
 		}
 
+		// Sanitize signatures property if none were found at this point
+		transaction.signatures =
```
MaciejBaj: Array.
```suggestion
		transaction.signatures = Array.isArray(transaction.signatures) ? transaction.signatures || [];
```

pablitovicente: @MaciejBaj prettier made the change

MaciejBaj: this.ready is doing the same check. Can we rely on it?

```
Multisignature.prototype.ready = function(transaction, sender) {
	if (!Array.isArray(transaction.signatures)) {
		return false;
	}

```

pablitovicente: @MaciejBaj will check though my understanding was that for all transactions it should be present as per Diego's comments; @diego-G can you confirm this?

MaciejBaj: I would prefer to move `transaction.signatures` assignment to both transfer.js and multignature.js `objectNormalize` methods. If you sanitize the signatures outside of the normalize functions it introduces inconsistency.

diego-G: Yes, I confirm. Just bear in mind you mutate signatures too late. It has to be normalise/sanitise before ready. I would suggest to perhaps moving it to objectNormalise in transactions. What do you think @4miners ?

logic/transaction.js
```diff
@@ -426,6 +426,9 @@ class Transaction {
 			return setImmediate(cb, 'Multisig request is not allowed');
 		}
 
+		// Sanitize ready property
+		transaction.ready = this.ready(transaction, sender);
```
MaciejBaj: Please try to find a better place to call this.ready

logic/transaction.js
pablitovicente: @MaciejBaj @diego-G moved to objectNormalized in transactions.

MaciejBaj: thanks!

test/common/scenarios.js
```diff
@@ -57,6 +58,25 @@ function Multisig(options) {
 		passphrase: this.account.passphrase,
 		secondPassphrase: this.account.secondPassphrase,
 	});
+
+	this.transfer = (
```
diego-G: This helper has the same function than -> https://github.com/LiskHQ/lisk/blob/7a269664f65abeb2f5fffc595cbc3f75af59d3b7/test/common/scenarios.js#L52

test/common/scenarios.js
```diff
@@ -57,6 +58,25 @@ function Multisig(options) {
...
+		amount = 100000000000
+	) => {
+		return lisk.transaction.transfer({
+			amount,
+			passphrase,
+			recipientId,
+		});
+	};
+
+	this.signTransaction = (signers, transaction) => {
```
diego-G: I don't see the point of introducing this helper because it obfuscates the process. But if you really want to do so, please use it uniformly across the file and ideally across the whole suite...

test/common/scenarios.js
pablitovicente: It's almost the same but this one accepts a recipientId instead of using the instance property. But I think based on your other comment I could move this logic to the test so the process of what's going on each test case is more obvious. What do you thing about moving it directly to each test case by using lisk-elements directly?

pablitovicente: @diego-G thanks for the comments about scenarios they were indeed unneeded. Changes done!

test/functional/http/post/4.multisignature.advanced.js
diego-G: @pablitovicente `offline_signed_with_ready_false` & `all_signatures_ready_false` are pretty much the same. Please, follow the notation to avoid misunderstandings.
`no_signatures_ready_true` is equal to `unsigned_with_ready_true`
`some_signatures_ready_true `should be o `offline_partly_signed_with_ready_true` or `partly_offline_signed_with_ready_true`

pablitovicente: Improved the names in two cases and removed one that was unused.

logic/transaction.js
```diff
@@ -1179,6 +1183,12 @@ class Transaction {
 				.join(', ')}`;
 		}
 
+		// Sanitize signatures property
+		transaction.signatures =
```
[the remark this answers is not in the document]
pablitovicente: @4miners Good Catch! Changed for setting it to null for having the property always present but not altering the checks for signatures type being array (this is also what is done on save so it's coherent in that sense)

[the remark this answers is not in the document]
pablitovicente: Ended up removing this change as it will not validate against the schemas I think the issue of the signatures property needs a deeper discussion. @4miners @diego-G

diego-G:
@pablitovicente IMHO we should keep the behaviour introduced in 383f99d. The idea is to obtain always uniformed objects from unisgned queue.
I would place it inside `verifyTransaction()` as we have done with the ready parameter due to the existing access to the sender.

PR #2537, opened 2018-11-08
Wrong error message "API access denied" - Closes #2514

### LINKED ISSUE #2514: Wrong error message  "API access denied"
### Expected behavior
If the API is not open for access, any reqests should be answered with a response like that:
```
{
message: "API access denied",
errors: [...]
}
```

### Actual behavior
The response in still in old 0.9.x style
```
{
success: false,
error: "API access denied"
}
```

### Steps to reproduce
Try to access a closed API

Lisk core 1.1.1

pablitovicente:
### What was the problem?

applyAPIAccessRules() middleware was using error messages that do not match swagger ones.

### How did I fix it?

By changing the messages in the middleware to use the same schema as swagger ones. 

### How to test it?

Two cases for testing: 

1. Edit `config/default/config.json` and set `api.enabled` to `false` start the node and make a request you should get the new error message.
2. Edit  `config/devnet/config.json` and set `api.access` to `{"public": "false", "whiteList": ["127.2.2.1"]}` (you have to use an IP address that is not associated to the host making the request!) start the node and make a request you should get the new error message.

### Review checklist

* The PR resolves #2514
* Linting passes
* Tests pass

### DIFF
--- a/helpers/http_api.js
+++ b/helpers/http_api.js
@@ -138,22 +138,18 @@ var middleware = {
 	 */
 	applyAPIAccessRules(config, req, res, next) {
 		if (!config.api.enabled) {
-			res
-				.status(apiCodes.INTERNAL_SERVER_ERROR)
-				.send({
-					message: 'API access disabled',
-					errors: ['API is not enabled in this node.'],
-				});
+			res.status(apiCodes.FORBIDDEN).send({
+				message: 'API access disabled',
+				errors: ['API is not enabled in this node.'],
+			});
 		} else if (
 			!config.api.access.public &&
 			!checkIpInList(config.api.access.whiteList, req.ip)
 		) {
-			res
-				.status(apiCodes.FORBIDDEN)
-				.send({
-					message: 'API access denied',
-					errors: ['API access blocked.'],
-				});
+			res.status(apiCodes.FORBIDDEN).send({
+				message: 'API access denied',
+				errors: ['API access blocked.'],
+			});
 		} else {
 			next();
 		}
--- a/helpers/http_api.js
+++ b/helpers/http_api.js
@@ -152,7 +152,7 @@ var middleware = {
 				.status(apiCodes.FORBIDDEN)
 				.send({
 					message: 'API access denied',
-					errors: ['API is not public in this node'],
+					errors: ['API access blocked.'],
 				});
 		} else {
 			next();
--- a/helpers/http_api.js
+++ b/helpers/http_api.js
@@ -140,14 +140,20 @@ var middleware = {
 		if (!config.api.enabled) {
 			res
 				.status(apiCodes.INTERNAL_SERVER_ERROR)
-				.send({ success: false, error: 'API access disabled' });
+				.send({
+					message: 'API access disabled',
+					errors: ['API is not enabled in this node.'],
+				});
 		} else if (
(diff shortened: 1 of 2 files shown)


helpers/http_api.js
```diff
@@ -140,14 +140,20 @@ var middleware = {
...
 		} else if (
 			!config.api.access.public &&
 			!checkIpInList(config.api.access.whiteList, req.ip)
 		) {
 			res
 				.status(apiCodes.FORBIDDEN)
-				.send({ success: false, error: 'API access denied' });
+				.send({
+					message: 'API access denied',
+					errors: ['API is not public in this node'],
```
MaciejBaj:
```suggestion
					errors: ['API access blocked.'],
```

test/unit/helpers/http_api.js
```diff
@@ -310,8 +310,8 @@ describe('httpApi', () => {
 				expect(validNextSpy).to.be.not.called;
 				expect(resMock.status).to.be.calledWith(apiCodes.FORBIDDEN);
 				expect(resMock.send).to.be.calledWith({
-					success: false,
-					error: 'API access denied',
+					message: 'API access denied',
+					errors: ['API is not public in this node'],
```
MaciejBaj:
```suggestion
					errors: ['API access blocked.'],
```

helpers/http_api.js
```diff
@@ -140,14 +140,20 @@ var middleware = {
 		if (!config.api.enabled) {
 			res
 				.status(apiCodes.INTERNAL_SERVER_ERROR)
-				.send({ success: false, error: 'API access disabled' });
+				.send({
+					message: 'API access disabled',
```
ManuGowda: @MaciejBaj @nazarhussain @4miners I feel this is a breaking change in API response, as the end user until now was expecting `status` and `error` message, now we change `error` -> `message` and new field `errors`?
whats your opinion?

helpers/http_api.js
```diff
@@ -140,14 +140,20 @@ var middleware = {
 		if (!config.api.enabled) {
 			res
 				.status(apiCodes.INTERNAL_SERVER_ERROR)
```
ManuGowda: status `INTERNAL_SERVER_ERROR`? shouldn't this be `FORBIDDEN`?

pablitovicente: Good catch, I payed attention to the object change and not the status being returned - test is not new - but this will also be a breaking change as on your previous note wouldn't it? Depending on the answer to the other comment we can change this.

helpers/http_api.js
lsilvs: Our [API specification](https://github.com/LiskHQ/lisk/blob/development/schema/swagger.yml#L2434) says that the response for `AccessDeniedError` should be an object and must have `message` property. That being said, I reckon we are fixing a bug and not changing the expected API behaviour here.

helpers/http_api.js
lsilvs: I have the same opinion here. Not breaking changing. Just bug fix.
All `AccessDeniedError` should be 403 (`FORBIDDEN`)

pablitovicente: @lsilvs @ManuGowda I updated the status code and made an small fix in the test to for the code to be validated correctly.

PR #2543, opened 2018-11-13
Maintain delegate list in memory - Closes #2421

### LINKED ISSUE #2421: Maintain delegates list in memory
### Expected behavior
Maintain a copy of shuffled delegates list in the node's memory and reuse it during the same round instead of calling `generateDelegateList` multiple times.
### Actual behavior
Function `generateDelegateList` is responsible for generating a shuffled list of delegates for a particular round, it performs following operations:
- Grab a list of 101 delegates ordered by `vote ASC, "publicKey" DESC` from `mem_accounts` table
- Shuffle the list using round number as seed, it performs SHA-256 operation 21 times.

Currently `generateDelegateList` is called in a many places:
- `modules.blocks.process->onReceiveBlock`, when the block is in fork 1 or 5 states.
- `modules.blocks.verify->processBlock`, for every block.
- `modules.delegates->__private.forge`, when there is forging enabled - every second under until block appear in the current slot window.
- `modules.delegates->getForgers`, every time when the corresponding API endpoint is called.
- `modules.rounds->__private.getOutsiders`, every time when a round is finished.

As you can see above, this function is used heavy across the application, which results in an unnecessary decrease of the overall performance.
### Steps to reproduce
N/A
`>= 1.0.0`

yatki:
### What was the problem?
We would like to Maintain a copy of shuffled delegates list in the node's memory and reuse it during the same round instead of calling generateDelegateList multiple times. However, we were executing database queries which impacts the performance heavily.
### How did I fix it?
I'm caching delegate list in the memory.
### Review checklist

* The PR resolves #2421 
* Linting passes
* Tests pass


yatki:
@4miners can you take a look, if you confirm I'll add the unit tests.

modules/delegates.js
```diff
@@ -97,6 +98,32 @@ class Delegates {
...
+/**
+ * Caches delegate list for last 2 rounds.
+ *
+ * @private
+ * @param {number} roundNo - Round Number
+ * @param {array} list - Delegate list
+ * @returns {Void}
+ */
+__private.updateDelegateListCache = function(roundNo, list) {
+	library.logger.debug('Updating delegate list cache.');
```
lsilvs: Add round number to the `debug` message

```suggestion
	library.logger.debug('Updating delegate list cache for round', roundNo);
```

modules/rounds.js
```diff
@@ -119,6 +119,10 @@ Rounds.prototype.backwardTick = function(block, previousBlock, done, tx) {
 		(prevRound === round && nextRound !== round) ||
 		(block.height === 1 || block.height === 101);
 
+	if (scope.finishRound) {
+		modules.delegates.clearDelegateListCache();
+	}
```
lsilvs: Can you add a comment explaining why do we need it here, pls?

modules/delegates.js
```diff
@@ -687,17 +680,24 @@ Delegates.prototype.updateForgingStatus = function(
...
  */
-Delegates.prototype.generateDelegateList = function(round, source, cb, tx) {
-	// Set default function for getting delegates
-	source = source || __private.getKeysSortByVote;
+Delegates.prototype.generateDelegateList = function(round, cb, tx) {
+	const lastBlockRound = slots.calcRound(modules.blocks.lastBlock.get().height);
+	const source =
+		lastBlockRound > round
+			? 'getDelegatesFromPreviousRound'
+			: 'getKeysSortByVote';
```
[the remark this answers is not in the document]
lsilvs: After talking to @4miners and analysing the code further I reckon this new check is correct but it changes the behaviour of [`validateBlockSlot`](https://github.com/LiskHQ/lisk/blob/development/modules/blocks/process.js#L610) and this is beyond the scope of this PR.

Another issue should be created for this imo.

MaciejBaj:
@yatki I am waiting for a test which will cover lastBlock and newBlock in situations of:
- forkFive
- forkOne
- regular (newBlock.height = lastBlock.height + 1)

yatki:
@lsilvs @diego-G I reverted my changes now the PR only includes it's scope. I also created https://github.com/LiskHQ/lisk/issues/2574.

modules/delegates.js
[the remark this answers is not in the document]
yatki: @lsilvs Created https://github.com/LiskHQ/lisk/issues/2574

lsilvs:
@4miners A possible solution for this scenario would be to only cache rounds which there is no zero vote delegate present?

4miners:
@lsilvs The problem with this approach is that `generateDelegateList` doesn't know about `vote` column, as it gets only public keys. Also, the issue is present only in `testnet` network, so maybe exceptions are the way to go?

lsilvs:
@4miners I personally think we should avoid exceptions as much as we can especially in this case where the exception would be added to serve the implementation and not the protocol.

Maybe a minor tweak in the `getKeysSortByVote` will do the trick?

```
__private.getKeysSortByVote = function(cb, tx) {
	modules.accounts.getAccounts(
		{
			isDelegate: 1,
			sort: { vote: -1, publicKey: 1 },
			limit: ACTIVE_DELEGATES,
		},
		['publicKey', 'vote'],
		(err, rows) => {
			if (err) {
				return setImmediate(cb, err);
			}
			const isCacheable = rows.reduce((acc, curr) => acc && (curr.vote != 0), true);
			return setImmediate(cb, null, rows.map(el => el.publicKey), isCacheable);
		},
		tx
	);
};
```

yatki:
@4miners @lsilvs can you please check the latest commit.

modules/delegates.js
```diff
@@ -97,6 +99,47 @@ class Delegates {
...
 
+/**
+ * Caches delegate list for last 2 rounds.
+ *
+ * @private
+ * @param {number} round - Round Number
+ * @param {array} delegatesList - Delegate list
+ * @returns {Void}
+ */
+__private.updateDelegateListCache = function(round, delegatesList) {
```
lsilvs: Unit test missing for `updateDelegateListCache` function

modules/delegates.js
```diff
@@ -721,6 +769,13 @@ Delegates.prototype.generateDelegateList = function(round, source, cb, tx) {
 				.digest();
 		}
 
+		// If the round is not an exception, cache the round.
+		if (
+			!Array.isArray(exceptions.ignoreDelegateListCacheForRounds) ||
+			!exceptions.ignoreDelegateListCacheForRounds.includes(round)
+		) {
```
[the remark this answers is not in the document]
yatki: @lsilvs about your second comment, I think cacheEnabled is more related to redis cache, right? this one is memory cache. I'm not sure if we should control two of them with the same key. @4miners what do you think ?

config/testnet/exceptions.js
```diff
@@ -83,4 +83,14 @@ module.exports = {
 			'2eb06bf528d60231a6b93a4d03b02200c938692e8a92d51d4dbaf94087b2e1261a904eb00cba4a0ed7e9d7e6a996666d4cfe3b7011a64252a8a286b8111b4701',
 		],
 	},
+	/**
+	 * In modules.delegates.js we are using generateDelegateList
```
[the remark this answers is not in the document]
yatki: @lsilvs shouldn't we refactor the comments as well in that case? I would prefer to keep this, since it makes more clear the purpose of the property.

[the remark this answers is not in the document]
lsilvs: Ideally, yes. I foresee that not happening though.

But it's up to you. It was just a suggestion.

modules/delegates.js
yatki: @lsilvs Created https://github.com/LiskHQ/lisk/issues/2593

modules/delegates.js
```diff
@@ -769,11 +769,12 @@ Delegates.prototype.generateDelegateList = function(round, source, cb, tx) {
 				.digest();
 		}
 
-		// If the round is not an exception, cache the round.
-		if (
+		const isRoundInExceptions =
 			!Array.isArray(exceptions.ignoreDelegateListCacheForRounds) ||
-			!exceptions.ignoreDelegateListCacheForRounds.includes(round)
-		) {
+			!exceptions.ignoreDelegateListCacheForRounds.includes(round);
```
lsilvs: This will not work. If you want to name the variable `isRoundInExceptions`, you either do:

```
const isRoundInExceptions =
        Array.isArray(exceptions.ignoreDelegateListCacheForRounds) &&
        exceptions.ignoreDelegateListCacheForRounds.includes(round);
```
or
```
const isRoundInExceptions =
        !(!Array.isArray(exceptions.ignoreDelegateListCacheForRounds) ||
        !exceptions.ignoreDelegateListCacheForRounds.includes(round));
```

I initially suggested the first option because I find negated OR statement confusing.

config/testnet/exceptions.js
```diff
@@ -83,4 +83,14 @@ module.exports = {
...
+	/**
+	 * In modules/delegates.js we are using generateDelegateList
+	 * to get the list of forgers for the round. However, we are
+	 * also caching this list to reduce calls to the database.
+	 * In the rounds below, using cache, creates forks.
+	 * See: https://github.com/LiskHQ/lisk/pull/2543#pullrequestreview-178505587
+	 *
+	 * So we are using the exception key below to skip caching for the rounds provided in the array.
+	 * */
+	ignoreDelegateListCacheForRounds: [80, 82],
```
4miners: You need to add empty array to `default/config,.json`

modules/delegates.js
```diff
@@ -721,6 +769,14 @@ Delegates.prototype.generateDelegateList = function(round, source, cb, tx) {
 				.digest();
 		}
 
+		const isRoundInExceptions =
+			!Array.isArray(exceptions.ignoreDelegateListCacheForRounds) ||
```
4miners: Then you don't need to check for `Array.isArray` here.

modules/delegates.js
yatki: @lsilvs I added ignoreDelegateListCacheForRounds key to default/exceptions.js and got rid of Array.isArray check. That's why I got rid of that variable completely.

modules/delegates.js
```diff
@@ -721,6 +769,10 @@ Delegates.prototype.generateDelegateList = function(round, source, cb, tx) {
 				.digest();
 		}
 
+		// If the round is not an exception, cache the round.
+		if (exceptions.ignoreDelegateListCacheForRounds.includes(round)) {
```
lsilvs:
```suggestion
		if (!exceptions.ignoreDelegateListCacheForRounds.includes(round)) {
```

modules/rounds.js
```diff
@@ -170,7 +170,21 @@ Rounds.prototype.backwardTick = function(block, previousBlock, done, tx) {
 		err => {
 			// Stop round ticking
 			__private.ticking = false;
-			return done(err);
+			if (err) {
+				return done(err);
+			}
```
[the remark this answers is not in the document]
yatki: @lsilvs I am not so sure if the block has been deleted because one of the previous steps could fail as well. For me it's confusing because `round.backwardLand()` consists of many chained methods and it's not possible to determine which one failed. So like you addressed keeping cache can cause a problem. 

So maybe instead of clearing cache in every error i can put another call before line 166 and clear the cache only in that catch block? What do you think?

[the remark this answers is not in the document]
lsilvs: I will leave this one to @MaciejBaj and @4miners as it's not clear for me as well.
But I think we should have tests covering all these scenarios in this PR so that we can ensure it's working as expected.

modules/delegates.js
```diff
@@ -97,6 +99,47 @@ class Delegates {
 	}
 }
 
+/**
+ * Caches delegate list for last 2 rounds.
+ *
+ * @private
+ * @param {number} round - Round Number
+ * @param {array} delegatesList - Delegate list
+ * @returns {Void}
```
MaciejBaj: We aren't using the `{Void}` convention.

modules/delegates.js
```diff
@@ -97,6 +99,47 @@ class Delegates {
...
+		.reduce((acc, current) => {
+			acc[current] = __private.delegatesListCache[current];
+			return acc;
+		}, {});
+};
+
+/**
+ * Invalidates the cached delegate list.
+ *
+ * @returns {Void}
```
MaciejBaj: https://github.com/LiskHQ/lisk/pull/2543/files#r237438479

modules/rounds.js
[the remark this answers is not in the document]
MaciejBaj: Yep, a block is already deleted ("popped"). If backwardTick fails, it would mean that a previous round was unsuccessfully restored, and because the steps were in the same database transaction, the block will be restored back. Cached delegates may be reused, so it's fine like this.

modules/delegates.js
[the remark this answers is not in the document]
lsilvs: @MaciejBaj Can you give your input here as well, pls?

modules/delegates.js
MaciejBaj: Please remove that lines.

modules/delegates.js
MaciejBaj: Please remove the 
```suggestion
```

yatki:
@4miners https://github.com/LiskHQ/lisk/issues/2593 there is already an issue for unit tests. I'll write them when we decide how to go with private functions.

modules/delegates.js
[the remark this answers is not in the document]
MaciejBaj: @yatki @lsilvs as it's an obvious performance improvement, I don't think we should give a possibility not to use caching of delegates list do node maintainers. 
So the current implementation is correct.

PR #2545, opened 2018-11-13
Move LiskHQ/lisk-build into LiskHQ/lisk and introduce basic Makefile - Closes #2400

### LINKED ISSUE #2400: Makefile
Create Makefile that supplants lisk-build

ghost:
### What was the problem?
The build script was separated from the code altough it's tightly coupled with its dependencies.

### How did I fix it?
Moved `build.sh` from LiskHQ/lisk-build into the `build` subdirectory.

### How to test it?
Follow the instructions in `build/README.md`; it boils down to using Ubuntu 16.04, installing the dependencies and running `make LISK_NET=<test|main>`.

### Review checklist

* The PR resolves #2400 
* Linting passes
* Tests pass


build/README.md
```diff
@@ -0,0 +1,19 @@
...
+The following packages must be installed to build releases with lisk-build:
+
+```
+sudo apt-get install autoconf build-essential git jq libtool moreutils python2.7 tcl8.5 wget zlib1g-dev
+```
+
+### How-To
+
+```
+make LISK_NET=main
```
nazarhussain: We use variable `LISK_NETWORK` all around the app, please use same here as well. Also use full network names.

nazarhussain: I think better shell syntax is to put ENV variables before start of command. 

```
LISK_NETWORK=mainnet  make
```

build/build.sh
```diff
@@ -0,0 +1,192 @@
...
+. "$(pwd)/config.sh"
+
+if [[ "$CLEAN" == true ]]; then
+	echo "Cleaning build."
+	rm -rf "src/$BUILD_NAME"
+fi
+
+################################################################################
+
+# Requirements for Ubuntu 14.04.5 LTS base install:
```
nazarhussain: In readme its specified `16.04 LTS`

build/build.sh
```diff
@@ -0,0 +1,192 @@
...
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program.  If not, see <http://www.gnu.org/licenses/>.
+######################################################################
```
nazarhussain: We use different license header in Lisk Core, please copy same from any other file.

build/build.sh
```diff
@@ -0,0 +1,192 @@
...
+if [ ! -f "$LISK_FILE" ]; then
+	echo "$LISK_FILE needs to be copied to src/"
+	exit 2
+fi
+
+echo
+echo "Downloading jq..."
+echo "--------------------------------------------------------------------------"
+[[ -f "$JQ_FILE" ]] || wget -nv "$JQ_BIN_URL" --output-document="$JQ_FILE"
+echo "$JQ_SHA256SUM  $JQ_FILE" |sha256sum -c
```
nazarhussain: `jq` is a dependency of the build process. Can we not just use the available jq binary and don't download new copy.

build/config.sh
```diff
@@ -0,0 +1,74 @@
...
+LISK_NETWORK="$LISK_NETWORK"
+LISK_URL="https://downloads.lisk.io/lisk/$LISK_NETWORK/$VERSION/$LISK_FILE"
+
+LISK_SCRIPTS_VERSION="0.4.0"
+LISK_SCRIPTS_SHA256SUM="03ddb4d651bba12906c77a6f1c3a6931c0e227bcd20cf44d409c0cd59c8d7b12"
+LISK_SCRIPTS_DIR="lisk-scripts-$LISK_SCRIPTS_VERSION"
+LISK_SCRIPTS_FILE="$LISK_SCRIPTS_DIR.tar.gz"
+LISK_SCRIPTS_URL="https://github.com/LiskHQ/lisk-scripts/archive/v$LISK_SCRIPTS_VERSION.tar.gz"
+
+NODE_VERSION="8.12.0"
```
nazarhussain: Is it possible to extract this version from `package.json` key `engines.node`?

build/build.sh
```diff
@@ -0,0 +1,192 @@
...
+	echo "Creating etc/snapshot.json..."
+	echo "--------------------------------------------------------------------------"
+	cp config.json etc/snapshot.json
+	"./bin/$JQ_FILE" '.httpPort=9000' etc/snapshot.json |sponge etc/snapshot.json
+	"./bin/$JQ_FILE" '.wsPort=9001' etc/snapshot.json |sponge etc/snapshot.json
+	"./bin/$JQ_FILE" '.logFileName="logs/lisk_snapshot.log"' etc/snapshot.json |sponge etc/snapshot.json
+	"./bin/$JQ_FILE" '.fileLogLevel="info"' etc/snapshot.json |sponge etc/snapshot.json
+	"./bin/$JQ_FILE" '.db.database="lisk_snapshot"' etc/snapshot.json |sponge etc/snapshot.json
+	"./bin/$JQ_FILE" '.peers.list=[]' etc/snapshot.json |sponge etc/snapshot.json
+	"./bin/$JQ_FILE" '.loading.loadPerIteration=101' etc/snapshot.json |sponge etc/snapshot.json
```
nazarhussain: I don't think hardcode the configuration logic in build process is a good idea. We should create a seperate json file for with these values and then copy/merge it here.

build/build.sh
```diff
@@ -0,0 +1,192 @@
...
+	# extract nodejs
+	tar xf "$NODE_FILE" --strip-components=1 --directory="$BUILD_NAME"
+
+	pushd "$BUILD_NAME"
+
+	echo "Setting LISK_NETWORK in env.sh..."
+	echo "--------------------------------------------------------------------------"
+	echo "export LISK_NETWORK=${LISK_NETWORK}net" >>env.sh
+
+	echo "Create default custom config.json..."
```
nazarhussain: > Create default custom config.json

default or custom?

build/build.sh
```diff
@@ -0,0 +1,192 @@
...
+	. "$(pwd)/env.sh"
+	set -u
+	npm install --production
+
+	echo "Installing pm2 and lisk-commander..."
+	echo "--------------------------------------------------------------------------"
+	npm install --production --global "pm2@$PM2_VERSION"
+	npm install --production --global "lisk-commander@$LISK_COMMANDER_VERSION"
+
+	date "+%H:%M:%S %d/%m/%Y" >build
```
nazarhussain: The file you referred in `app.js` is `.build`, I think `.` is missing here.

build/Makefile
```diff
@@ -0,0 +1,22 @@
...
+.PHONY: check
+check:
+ifndef LISK_NET
+	$(error The LISK_NET environment variable must be set.)
+endif
+
+src/lisk-$(lisk_version).tgz: ../REVISION
+	cd ..; tar czf build/src/lisk-$(lisk_version).tgz \
+	    --exclude-vcs --exclude=build --exclude-ignore=build/build.ignore \
+	    --transform='s#^./#package/#' .
```
nazarhussain: It seems you completely removed the behavior of `npm pack` with the file `build.ignore`. I feel it's not good practice. You should use `.npmignore` file as mostly developers will not aware of `build.ignore` file. For developer point of view they will always modify/look `.npmignore` to see which files will be shipped in release.

build/build.ignore
```diff
@@ -0,0 +1,8 @@
+.[a-z]*
+docker
+Dockerfile
+docs
+Gruntfile.js
+Jenkinsfile*
+tasks
+test
```
nazarhussain: This file actually not behave same like `.npmignore`. Please see the list there. But better not to introduce new file and use only `.npmignore`.

nazarhussain (changes requested):

build/README.md
fchavant: `LISK_NETWORK` would contain `mainnet` while `LISK_NET` is set to `main`; this is because `build.sh` requires `main` or `test` to be passed as parameter.

fchavant: That is how you pass a variable to make.

build/config.sh
fchavant: I was thinking about taking the version in `.nvmrc` actually: https://github.com/LiskHQ/lisk-build/issues/351

build/build.sh
fchavant: The default version (empty: `{}`) of the custom configuration file.

build/Makefile
fchavant: While we do not use the same tool (`tar` instead of `npm pack`) or ignore list the behaviour is the same: we create the same file in the same format . It is true that we ignore a few more files though.
The reason we do not use the existing `.npmignore` is because its format is slightly different (and not documented?) -- do we need it at all if we never call `npm pack`?

Nazgolze (dismissed):
lgtm

MaciejBaj (dismissed):

Nazgolze (changes requested):
I missed this before - there's no logs directory in the resulting tarball. After manually adding it in though, the build works

Nazgolze (dismissed):

build/Makefile
nazarhussain: @fchavant That's not true exactly, Just check this out https://docs.npmjs.com/misc/developers

> .npmignore files follow the same pattern rules as .gitignore files:
> 
> Blank lines or lines starting with # are ignored.
> Standard glob patterns work.
> You can end patterns with a forward slash / to specify a directory.
> You can negate a pattern by starting it with an exclamation point !.
> .
> .
> Additionally, everything in node_modules is ignored, except for bundled dependencies. npm automatically handles this for you, so don’t bother adding node_modules to .npmignore.
> 

So `.npmignore` is very well documented and actually follow the exact rules of `.gitingore`. 

Now just consider a scenario that you ask any Core developer (any NodeJS developer):

- Which files will be shipped in production? 
- A developer created a file, but he don't want it to be shipped in production. 

Everyone will go and look for `.npmignore` file. So removing a file which have so much significance in technology domain is not right approach. 

The problem during the build is that you don't want to use `npm pack`, so you come up with a workaround for it, but by not removing `.npmignore` file.

nazarhussain (dismissed):

build/Makefile
fchavant: So what is your suggestion?

nazarhussain: Two possibilities. 

1. Use `npm pack` to we can use its power and also benefit if it introduce new feature in future. 
2 Use custom bash script to strip down the files using `.npmignore` and then pass it to tar.

shuse2: Now it's often mainly use 
https://docs.npmjs.com/files/package.json#files

MaciejBaj:
We decided that:
- `.npmignore` is used only for build
- we should use .npmignore to create builds instead of .buildignore
- .buildignore can be removed as it's not used

nazarhussain (approved):

build/config.sh
```diff
@@ -0,0 +1,69 @@
...
+LISK_NETWORK="$LISK_NETWORK"
+LISK_URL="https://downloads.lisk.io/lisk/$LISK_NETWORK/$VERSION/$LISK_FILE"
+
+LISK_SCRIPTS_VERSION="0.4.0"
+LISK_SCRIPTS_SHA256SUM="03ddb4d651bba12906c77a6f1c3a6931c0e227bcd20cf44d409c0cd59c8d7b12"
+LISK_SCRIPTS_DIR="lisk-scripts-$LISK_SCRIPTS_VERSION"
+LISK_SCRIPTS_FILE="$LISK_SCRIPTS_DIR.tar.gz"
+LISK_SCRIPTS_URL="https://github.com/LiskHQ/lisk-scripts/archive/v$LISK_SCRIPTS_VERSION.tar.gz"
+
+NODE_VERSION="8.13.0"
```
MaciejBaj: I will create a PR to update the node version to 8.13.0 for the sources as well.

MaciejBaj (approved):

PR #2547, opened 2018-11-13
Refactor eslint-disable comments - Closes #2447

### LINKED ISSUE #2447: Remove "eslint-disable" statements from on top of files
### Expected behavior
We shouldn't disable eslint rules generally and especially in the whole file. If it's really necessary, It is possible to disable certain rules for certain code blocks. See: [#disabling-rules-with-inline-comments](https://eslint.org/docs/user-guide/configuring#disabling-rules-with-inline-comments)

We should refactor the codebase and implement more convenient solutions.
### Actual behavior
We are disabling eslint rules on top test files. See: [test/unit/modules/accounts.js#L1](https://github.com/LiskHQ/lisk/blob/development/test/unit/modules/accounts.js#L1)
1.3.0+

yatki:
### What was the problem?
We were using `eslint-disable` comments at the top of some files and were not enabling again which can increase the chance of having inconsistent code style.
### How did I fix it?
I wrapped the code blocks with eslint-disable comments and enabled the rules at the end of the code block.
### Review checklist

* The PR resolves #2447 
* Linting passes
* Tests pass

### DIFF
--- a/test/unit/logic/round.js
+++ b/test/unit/logic/round.js
@@ -1,4 +1,3 @@
-/* eslint-disable mocha/no-skipped-tests */
 /*
  * Copyright © 2018 Lisk Foundation
  *
@@ -103,6 +102,7 @@ describe('rounds', () => {
 				done();
 			});
 
+			/* eslint-disable mocha/no-skipped-tests */
 			describe.skip('library', () => {
 				it('should throw', done => {
 					var property = 'library';
@@ -147,6 +147,7 @@ describe('rounds', () => {
 					done();
 				});
 			});
+			/* eslint-enable mocha/no-skipped-tests */
 
 			describe('round', () => {
 				it('should throw', done => {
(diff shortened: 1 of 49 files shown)


test/unit/logic/round.js
```diff
@@ -752,6 +754,7 @@ describe('rounds', () => {
 				});
 			});
 		});
+		/* eslint-disable mocha/no-skipped-tests */
```
pablitovicente: Shouldn't this one be `eslint-enable` to reactivate it ?

test/unit/modules/voters.js
```diff
@@ -222,3 +222,5 @@ describe('voters', () => {
 		it('should assign loaded = true');
 	});
 });
+
```
diego-G: Remove unneeded blank line.

test/unit/logic/round.js
yatki: Good catch, thanks!

PR #2552, opened 2018-11-14
Received post block broadcast request in unexpected format - Closes #2453

### LINKED ISSUE #2453: Received post block broadcast request in unexpected format
### Expected behavior
blocks should always have all the required parameters including `nonce`
### Actual behavior
Inconsistant behaviour while received post block broadcast request with missing `nonce` property
`
[dbg] 2018-10-09 07:26:00 | Received post block broadcast request in unexpected format - {"err":[{"code":"OBJECT_MISSING_REQUIRED_PROPERTY","params":["nonce"],"message":"Missing required property: nonce","path":[]}],"module":"transport","query":{"block":{"version":1,"reward":"500000000","payloadHash":"e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855","timestamp":74960760,"previousBlock":"17113026271781611947","generatorPublicKey":"01389197bbaf1afb0acd47bbfeabb34aca80fb372a8f694a1c0716b3398db746","blockSignature":"fa4e8a84c9ad01c35066945728cc1066087479308a430178eb4f8bf6997a9eaf3f45563535ac6a8b74a7ac016f7c93b98892bb345fb6d8f937aa03ca11c61a08","height":6084,"id":"8720819914036341200","relays":3}}}
`
### Steps to reproduce
login to `alphanet5-seed-01.liskdev.net` and grep for `Received post block broadcast request in unexpected format` you should be able to get all the information required for debugging.
`1.1.0`

lsilvs:
### What was the problem?
Node was receiving `postBlock` call from a peer not listed in the `connectionsTable`.

### How did I fix it?
As the issue is not reproducible I've improved the code by not undoing the `connectionsTable` operation in case `slaveToMasterSender.send` fails.
The reason is because when PeersUpdateRules.prototype.{insert|remove} is invoked it means the socket has already connected/disconnected hence the `connectionsTable` should not be reverted even when `slaveToMasterSender.send` fails.

### How to test it?
Run network tests

### Review checklist

* The PR resolves #2453 
* Linting passes
* Tests pass

### DIFF
--- a/api/ws/workers/peers_update_rules.js
+++ b/api/ws/workers/peers_update_rules.js
@@ -67,7 +67,6 @@ PeersUpdateRules.prototype.insert = function(peer, connectionId, cb) {
 			peer,
 			err => {
 				if (err) {
-					connectionsTable.remove(peer.nonce);
 					if (!err.code) {
 						err = new PeerUpdateError(
 							failureCodes.ON_MASTER.UPDATE.TRANSPORT,
@@ -104,7 +103,6 @@ PeersUpdateRules.prototype.remove = function(peer, connectionId, cb) {
 			peer,
 			err => {
 				if (err && !err.code) {
-					connectionsTable.add(peer.nonce, connectionId);
 					err = new PeerUpdateError(
 						failureCodes.ON_MASTER.UPDATE.TRANSPORT,
 						failureCodes.errorMessages[failureCodes.ON_MASTER.UPDATE.TRANSPORT],
--- a/test/unit/api/ws/workers/peers_update_rules.js
+++ b/test/unit/api/ws/workers/peers_update_rules.js
@@ -165,26 +165,34 @@ describe('PeersUpdateRules', () => {
 			});
 		});
 
-		it('should remove added entries from connectionsTable after receiving an error without code from server', done => {
+		it('should NOT remove added entries from connectionsTable after receiving an error without code from server', done => {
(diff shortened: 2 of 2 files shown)


api/ws/workers/peers_update_rules.js
```diff
@@ -67,7 +67,6 @@ PeersUpdateRules.prototype.insert = function(peer, connectionId, cb) {
 			peer,
 			err => {
 				if (err) {
-					connectionsTable.remove(peer.nonce);
```
MaciejBaj: Shouldn't we move also the `connectionsTable.add`, after the operation succeeds on the master process, to line https://github.com/LiskHQ/lisk/blob/6feb9a42e13884453b4de3fc64760acdf3af7ee5/api/ws/workers/peers_update_rules.js#L80?

api/ws/workers/peers_update_rules.js
```diff
@@ -67,7 +67,6 @@ PeersUpdateRules.prototype.insert = function(peer, connectionId, cb) {
...
 					if (!err.code) {
```
MaciejBaj: Shouldn't we move also the `connectionsTable.remove`, after the operation succeeds on the master process, to line https://github.com/LiskHQ/lisk/blob/6feb9a42e13884453b4de3fc64760acdf3af7ee5/api/ws/workers/peers_update_rules.js#L99 ?

PR #2554, opened 2018-11-16
Update Development Dependencies not requiring code changes - Closes #2499

### LINKED ISSUE #2499: Update Development Dependencies not requiring code changes
### Expected behavior

Development dependencies with newer minor/patch versions or non breaking changes should be upgraded to their latest versions.

### Actual behavior

Some development dependencies are outdated and can be upgraded without requiring code changes.

### Steps to reproduce

run `npm outdated -l` and check development dependencies (devDependencies)

1.2.0

pablitovicente:
### What was the problem?

Some development dependencies were out of date and the version jump didn't introduce breaking changes.

### How did I fix it?

Check change logs of devDependencies and upgraded the ones with no breaking changes.

### How to test it?

- Run test suit
- Check build passes

### Review checklist

* The PR resolves #2499
* Linting passes
* Tests pass


package.json
```diff
@@ -96,7 +93,7 @@
 		"chai": "=4.0.2",
```
nazarhussain: Can we update this as well?

package.json
```diff
@@ -96,7 +93,7 @@
...
 		"chai-as-promised": "=7.1.1",
 		"chai-bignumber": "=2.0.0",
```
nazarhussain: Can you double check if we are really using this one? I guess on very few places we use it. If we update those tests then we don't need this dependency.

pablitovicente: Did not found usages so I removed it.

Issue #2384, opened 2018-09-10
Improve performance of blockchain synchronisation process


karek314:
I was looking at 1.0.0 synchronisation process and I can propose few improvements to the process.

1. While syncing, usually block verification process halts for 2-6 seconds when new batch of blocks is acquired from network and even more if there is timeout issue 6-15s. Considering that there is over 7m blocks on mainnet right now, it's causing significant delay with reaching latest height. Why not introduce async networking task to accumulate blocks way ahead of verification process? Amount enough to prevent verification process from halting due refetching blocks from network.

2. For some reason synchronisation process halts when incorrect block occurred on latest height (?)
<pre>
[inf] 2018-09-08 22:28:23 | Verify->verifyBlock succeeded for block 7417561181081319467 at height 4931452.
[inf] 2018-09-08 22:28:23 | Block 7417561181081319467 loaded from: 89.40.125.40:8001 - height: 4931452
[inf] 2018-09-08 22:28:23 | Verify->verifyBlock succeeded for block 13371346198298708762 at height 4931453.
[inf] 2018-09-08 22:28:23 | Releasing enqueued broadcasts
[inf] 2018-09-08 22:28:23 | Queue empty
[inf] 2018-09-08 22:28:23 | Block 13371346198298708762 loaded from: 89.40.125.40:8001 - height: 4931453
[inf] 2018-09-08 22:28:23 | Verify->verifyBlock succeeded for block 6374672892190167342 at height 4931454.
[inf] 2018-09-08 22:28:24 | Block 6374672892190167342 loaded from: 89.40.125.40:8001 - height: 4931454
[inf] 2018-09-08 22:28:24 | Verify->verifyBlock succeeded for block 1406692377958485661 at height 4931455.
[inf] 2018-09-08 22:28:24 | Block 1406692377958485661 loaded from: 89.40.125.40:8001 - height: 4931455
[inf] 2018-09-08 22:28:24 | Looking for common block with: 47.93.19.241:8001
[inf] 2018-09-08 22:28:24 | Found common block: 1406692377958485661 with: 47.93.19.241:8001
[inf] 2018-09-08 22:28:24 | Loading blocks from: 47.93.19.241:8001
[inf] 2018-09-08 22:28:24 | Removed peer - 46.28.204.68:8001
[inf] 2018-09-08 22:28:24 | Removed peer - 18.188.59.171:8001
[inf] 2018-09-08 22:28:24 | Removed peer - 13.78.12.96:8001
[...]

PR #2555, opened 2018-11-16
Clean up loadBlocksFromNetwork function from loader module - Closes #2292 & #2384

### LINKED ISSUE #2292: Clean up loadBlocksFromNetwork function in modules.loader
### Expected behavior
The flow of the function should be more understandable and code should follow proper naming conventions.
### Actual behavior
The function adds a lot of obscurity in the understanding of the functionality, it jumps control flow in multiple ways and is missing useful comments. 
### Steps to reproduce
Look at [function implementation](https://github.com/LiskHQ/lisk/blob/427a300c329b83ffe4f48c6919d3a0dab867e8e0/modules/loader.js#L651)
All versions.

diego-G:
### What was the problem?

The flow of the function was not easy to understand obscuring errors.

### How did I fix it?
- Fixing described bug in the first two points from #2384 https://github.com/LiskHQ/lisk/commit/e7e435533429afb0e900431827e89587138779cf
- Deprecating `loader.getNetwork()` by `loader.getRandomPeerFromNetwork()`https://github.com/LiskHQ/lisk/commit/881fba2a8db8c6a994eabde57ad402d39a4ad625
- Using `async.waterfall` instead of callback 2 callback

### Review checklist

* The PR resolves #2292 & #2384
* Linting passes
* Tests pass

### DIFF
--- a/modules/blocks/process.js
+++ b/modules/blocks/process.js
@@ -261,6 +261,11 @@ __private.receiveForkFive = function(block, lastBlock, cb) {
 Process.prototype.getCommonBlock = function(peer, height, cb) {
 	let comparisonFailed = false;
 
+	// Allowing blocks to be loaded from genesis
+	if (height === 1) {
+		return setImmediate(cb);
+	}
+
 	async.waterfall(
 		[
 			function(waterCb) {
--- a/modules/blocks/process.js
+++ b/modules/blocks/process.js
@@ -261,11 +261,6 @@ __private.receiveForkFive = function(block, lastBlock, cb) {
 Process.prototype.getCommonBlock = function(peer, height, cb) {
 	let comparisonFailed = false;
 
-	// Allowing blocks to be loaded from genesis
-	if (height === 1) {
-		return setImmediate(cb);
-	}
-
 	async.waterfall(
 		[
 			function(waterCb) {
--- a/modules/loader.js
+++ b/modules/loader.js
@@ -191,9 +191,9 @@ __private.loadSignatures = function(cb) {
 	async.waterfall(
 		[
 			function(waterCb) {
-				self.getRandomPeerFromNetwork((err, peer) => {
-					if (err) {
-						return setImmediate(waterCb, err);
+				self.getRandomPeerFromNetwork((getRandomPeerFromNetworkErr, peer) => {
+					if (getRandomPeerFromNetworkErr) {
+						return setImmediate(waterCb, getRandomPeerFromNetworkErr);
 					}
 					return setImmediate(waterCb, null, peer);
 				});
@@ -253,9 +253,9 @@ __private.loadTransactions = function(cb) {
 	async.waterfall(
 		[
 			function(waterCb) {
-				self.getRandomPeerFromNetwork((err, peer) => {
-					if (err) {
-						return setImmediate(waterCb, err);
+				self.getRandomPeerFromNetwork((getRandomPeerFromNetworkErr, peer) => {
+					if (getRandomPeerFromNetworkErr) {
+						return setImmediate(waterCb, getRandomPeerFromNetworkErr);
 					}
 					return setImmediate(waterCb, null, peer);
 				});
@@ -853,22 +853,24 @@ __private.loadBlocksFromNetwork = function(cb) {
 			async.waterfall(
 				[
 					waterfallCb => {
-						self.getRandomPeerFromNetwork((err, peer) => {
-							if (err) {
-								failedAttemptsToLoad += 1;
-								library.logger.error(
-									'Failed to get random peer from network',
-									err.toString()
+						self.getRandomPeerFromNetwork(
+							(getRandomPeerFromNetworkErr, peer) => {
+								if (getRandomPeerFromNetworkErr) {
+									failedAttemptsToLoad += 1;
+									library.logger.error(
+										'Failed to get random peer from network',
+										getRandomPeerFromNetworkErr.toString()
+									);
+									return whilstCb();
+								}
+								__private.blocksToSync = peer.height;
+								const lastBlock = modules.blocks.lastBlock.get();
+								library.logger.info(
+									`Looking for common block with: ${peer.string}`
 								);
-								return whilstCb();
+								return waterfallCb(null, peer, lastBlock);
 							}
-							__private.blocksToSync = peer.height;
-							const lastBlock = modules.blocks.lastBlock.get();
-							library.logger.info(
-								`Looking for common block with: ${peer.string}`
-							);
-							return waterfallCb(null, peer, lastBlock);
-						});
+						);
 					},
 					(peer, lastBlock, waterfallCb) => {
 						// No need to call getCommonBlock if height is Genesis block
@@ -879,9 +881,9 @@ __private.loadBlocksFromNetwork = function(cb) {
 						modules.blocks.process.getCommonBlock(
 							peer,
 							lastBlock.height,
-							(err, commonBlock) => {
-								if (err) {
-									library.logger.error(err.toString());
+							(getCommonBlockErr, commonBlock) => {
+								if (getCommonBlockErr) {
+									library.logger.error(getCommonBlockErr.toString());
 									failedAttemptsToLoad += 1;
 									return whilstCb();
 								}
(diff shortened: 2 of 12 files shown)


modules/blocks/process.js
```diff
@@ -261,6 +261,11 @@ __private.receiveForkFive = function(block, lastBlock, cb) {
 Process.prototype.getCommonBlock = function(peer, height, cb) {
 	let comparisonFailed = false;
 
+	// Allowing blocks to be loaded from genesis
```
4miners: Why we need this actually?

modules/loader.js
```diff
@@ -846,86 +842,92 @@ __private.snapshotFinished = err => {
  * @todo Add description for the params
  */
 __private.loadBlocksFromNetwork = function(cb) {
-	let errorCount = 0;
+	// Number of failed attempts to load from peers.
+	let tries = 0;
```
4miners: Can we reset this when an attempt is successful? Also, the variable name can be more descriptive.

modules/loader.js
```diff
@@ -846,86 +842,92 @@ __private.snapshotFinished = err => {
...
+								`Looking for common block with: ${peer.string}`
+							);
+							waterfallCb(null, peer, lastBlock);
+						});
+					},
+					(peer, lastBlock, waterfallCb) => {
+						modules.blocks.process.getCommonBlock(
+							peer,
+							lastBlock.height,
+							(err, commonBlock) => {
```
4miners: What is `commonBlock`? This contains a block or just boolean? If boolean, then maybe change the name to something like `commonBlockFound`

modules/loader.js
```diff
@@ -846,86 +842,92 @@ __private.snapshotFinished = err => {
...
 									library.logger.error(err.toString());
+									tries += 1;
+									return whilstCb();
 								}
-								library.logger.error(
-									`Failed to find common block with: ${peer.string}`
-								);
-								errorCount += 1;
-								return next();
+								if (!commonBlock && lastBlock.height != 1) {
```
4miners: When `lastBlock.height` is `1` we will not log anything, it this behavior correct?

modules/loader.js
```diff
@@ -846,86 +842,92 @@ __private.snapshotFinished = err => {
...
+									return whilstCb();
+								}
+								loaded = lastValidBlock.id === lastBlock.id;
+								waterfallCb();
+							}
+						);
+					},
+				],
+				err => {
+					if (err) {
```
4miners: This will never happen because we always call `waterfallCb` without error.

modules/loader.js
```diff
@@ -846,86 +842,92 @@ __private.snapshotFinished = err => {
...
+					}
+					return whilstCb();
 				}
-			});
+			);
 		},
 		err => {
 			if (err) {
-				library.logger.error('Failed to load blocks from network', err);
+				library.logger.error(
```
4miners: Same here, yep?

modules/blocks/process.js
diego-G: I move it back to the former method.

modules/loader.js
diego-G: It is a common block.

modules/loader.js
```diff
@@ -846,90 +842,87 @@ __private.snapshotFinished = err => {
...
+								return whilstCb();
 							}
-							loaded = lastValidBlock.id === lastBlock.id;
-							lastBlock = null;
-							lastValidBlock = null;
-							next();
+							__private.blocksToSync = peer.height;
+							const lastBlock = modules.blocks.lastBlock.get();
+							library.logger.info(
+								`Looking for common block with: ${peer.string}`
```
nazarhussain: `Looking from common block from: peer`

modules/loader.js
```diff
@@ -846,90 +842,87 @@ __private.snapshotFinished = err => {
...
-					__private.blocksToSync = peer.height;
-
-					modules.blocks.process.loadBlocksFromPeer(
-						peer,
-						(err, lastValidBlock) => {
+		() => !loaded && failedAttemptsToLoad < 5,
+		whilstCb => {
+			async.waterfall(
+				[
+					waterfallCb => {
```
nazarhussain: Suggest tp use function names to make it more readable as the steps.

modules/loader.js
```diff
@@ -846,90 +842,87 @@ __private.snapshotFinished = err => {
...
-								errorCount += 1;
+								return whilstCb();
 							}
-							loaded = lastValidBlock.id === lastBlock.id;
-							lastBlock = null;
-							lastValidBlock = null;
-							next();
+							__private.blocksToSync = peer.height;
+							const lastBlock = modules.blocks.lastBlock.get();
+							library.logger.info(
```
nazarhussain: This log message must go to next step.

modules/loader.js
```diff
@@ -846,90 +842,87 @@ __private.snapshotFinished = err => {
...
+						modules.blocks.process.loadBlocksFromPeer(
+							peer,
+							(err, lastValidBlock) => {
+								if (err) {
+									failedAttemptsToLoad += 1;
+									library.logger.error(
+										`Failed to load blocks from: ${peer.string}`,
+										err.toString()
+									);
+									return whilstCb();
```
nazarhussain: This step is accepting `waterfallCb` and calling `whilstCb`, we should avoid this.

modules/loader.js
diego-G: I don't think it is correct the way you propose. Please, try again or resolve this conversation.

modules/loader.js
diego-G: How do you want to make it? The waterfall is inside the whilst. We need to stop the cycle from being propagated when we received an error.

modules/loader.js
diego-G: I've scanned the code for examples and I realise there is a mix of all sort of cases. I'll open an issue to make it uniform.

modules/loader.js
```diff
@@ -191,12 +191,10 @@ __private.loadSignatures = function(cb) {
 	async.waterfall(
 		[
 			function(waterCb) {
-				self.getNetwork((err, network) => {
+				self.getRandomPeerFromNetwork((err, peer) => {
```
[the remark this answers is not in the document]
diego-G: The entire code is written without following this convention. I'll open an issue to make it uniform.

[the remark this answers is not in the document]
nazarhussain: While you are refactoring the code and writing new one, its preferable to follow the right conventions. At least the functions you are refactoring should be updated right away in same PR.

modules/loader.js
nazarhussain: Same as above, the code you are refactoring, should be fixed right in same PR.

modules/loader.js
4miners: We can call `waterfallCb` with some error or just `true`, it will break the execution then.

modules/loader.js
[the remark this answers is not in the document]
diego-G: Can you point me to the **right conventions**? Do we have such guidelines?

modules/loader.js
diego-G: Same as above but even more clear https://github.com/LiskHQ/lisk/commit/24e363b6892e98147cefd7e8dae36cf1dce4ded6

diego-G: I mean it is a **recent** commit of yours using the wrong approach you just comment about. The criteria is clearly not clear.
Please, keep your review in the scope of the issue.

nazarhussain: The scope of any review includes to check if the proper conventions are practices are followed or not.

diego-G: Please, point me to the proper convention of waterfall.

nazarhussain: The error object convention already refactored, just use the function names in the async steps.

diego-G:
@nazarhussain I don't think we need more people reviewing the PR. I've added another commit to address the point 1. What I feel it is that we need a proper refactoring across the entire code to tackle these kind of concerns plus having any kind of prettifier/formatter. This will speed up our review process.

4miners:
The issue https://github.com/LiskHQ/lisk/issues/2384 is covered by this PR only partially. I think we shouldn't close it yet.

modules/loader.js
```diff
@@ -1054,13 +1041,13 @@ Loader.prototype.findGoodPeers = function(peers) {
 
 // Public methods
 /**
- * Gets a list of "good" peers from network.
```
MaciejBaj: Please extend the comment and explain what does it mean to be a good peer.

diego-G: A good peer is what `findGoodPeers()` decides. I haven't modified that function. It performs a complex calculation that cannot be explained easily:
- It filters unreachable peers and those ones below the node's height itself.
- It orders peers by descending height
- It creates a histogram of peer's heights
- It removes outliers

I'd open another PR if we need to review that process.

diego-G:
@4miners #2384 contains a bug in the first two bullets and an improvement in the third bullet that should be studied in a deeper way. If we think it is worthy, we should open an independent issue for it.

PR #2562, opened 2018-11-21
Apply eslint-rule eqeqeq - Related to #1597

### LINKED ISSUE #1597: Enable pending eslint rules
### Expected behavior
Turned off `eslint` rules should be fixed to ensure code quality and to be able to use latest ECMA 6 features.
### Actual behavior
Bunch of `eslint` rules were turned off during the `eslint` integration to `1.0.0`, the mocha test specific eslint rules were fixed here  #1485. In root there are bunch of eslint rules still disabled, following are list of root eslint rules with number of errors.

## Manual fix
- [ ] "no-unused-expressions": 1252
- [ ] "camelcase": 813
- [ ] "no-param-reassign": 470
- [ ] "prefer-destructuring": 296
- [ ] "no-underscore-dangle": 291
- [ ] "no-plusplus": 154
- [ ] "max-len": 153
- [ ] "radix": 76
- [ ] "no-use-before-define": 76
- [ ] "no-new": 26

## Warnings
- [ ] "func-names": 1016

## DONE
- [x] "array-callback-return": #2563 
- [x] "callback-return": #2561 
- [x] "consistent-return": #2565 
- [x] "eqeqeq": #2562 
- ~"global-require": 69~ to keep until `app.js` is refactored
- [x] "guard-for-in": 3 #2648
- ~"generator-star-spacing": 2~ overwritten with Prettier
- [x] "new-cap": #2566
- [X] "no-bitwise": 13 #2590
- [X] "no-nested-ternary": 8 #2589
- [x] "no-prototype-builtins": 6 #2646
- [X] "no-restricted-globals": 9 #2649
- [x] "no-restricted-syntax": 7 #2643 
- [x] "no-shadow": 321 #2583
- ~"no-template-curly-in-string": 58~ to keep until https://github.com/LiskHQ/lisk/milestone/35 is accomplished
- ~"no-throw-literal": 29~ to keep
- [x] "prefer-rest-params": 5 #2640
- ~"strict": 286 #2585~
- [x] "vars-on-top": , 0 #2649
- ~"wrap-iife"~ overwritten with Prettier

## Auto fix using `npm run lint:fix`
- ~"space-before-function-paren": 2180~ https://github.com/prettier/prettier/issues/3847
- [X] "no-var": 4519 #2603

### Steps to reproduce
In root `.eslintrc.json` under `rules` section turn on the rules and run `npm run lint` to see the errors/warnings

diego-G:
### Review checklist

* The PR is related to #1597
* Linting passes
* Tests pass

### DIFF
--- a/api/ws/rpc/connect.js
+++ b/api/ws/rpc/connect.js
@@ -45,20 +45,19 @@ const connectSteps = {
 	addConnectionOptions: peer => {
 		const systemHeaders = System.getHeaders();
 		const queryParams = {};
-
-		if (systemHeaders.version != null) {
+		if (systemHeaders.version) {
 			queryParams.version = systemHeaders.version;
 		}
-		if (systemHeaders.wsPort != null) {
+		if (systemHeaders.wsPort) {
 			queryParams.wsPort = systemHeaders.wsPort;
 		}
-		if (systemHeaders.httpPort != null) {
+		if (systemHeaders.httpPort) {
 			queryParams.httpPort = systemHeaders.httpPort;
 		}
-		if (systemHeaders.nethash != null) {
+		if (systemHeaders.nethash) {
 			queryParams.nethash = systemHeaders.nethash;
 		}
-		if (systemHeaders.nonce != null) {
+		if (systemHeaders.nonce) {
 			queryParams.nonce = systemHeaders.nonce;
 		}
 		peer.connectionOptions = {
--- a/api/ws/workers/middlewares/handshake.js
+++ b/api/ws/workers/middlewares/handshake.js
@@ -157,7 +157,7 @@ var extractHeaders = function(request) {
 	headers.ip = request.remoteAddress.split(':').pop();
 	headers.httpPort = +headers.httpPort;
 	headers.wsPort = +headers.wsPort;
-	if (headers.height != null) {
+	if (headers.height) {
 		headers.height = +headers.height;
 	}
 	return headers;
--- a/helpers/http_api.js
+++ b/helpers/http_api.js
@@ -182,7 +182,7 @@ var middleware = {
 
 				if (
 					isNaN(value) ||
-					parseInt(value) !== Number(value) ||
+					parseInt(value).toString() !== String(value) ||
 					isNaN(parseInt(value, radix))
 				) {
 					return value;
--- a/helpers/http_api.js
+++ b/helpers/http_api.js
@@ -182,7 +182,7 @@ var middleware = {
 
 				if (
 					isNaN(value) ||
-					parseInt(value) != value ||
+					parseInt(value) !== Number(value) ||
 					isNaN(parseInt(value, radix))
 				) {
 					return value;
--- a/helpers/z_schema.js
+++ b/helpers/z_schema.js
@@ -208,7 +208,7 @@ var liskFormats = {
 	 * @todo Add description for the function, the params and the return value
 	 */
 	queryList(obj) {
-		if (!_.isObject(obj) || _.isArray(obj)) {
+		if (!_.isObject(obj) || Array.isArray(obj)) {
 			return false;
 		}
 
@@ -223,7 +223,7 @@ var liskFormats = {
 	 * @todo Add description for the function, the params and the return value
 	 */
 	delegatesList(obj) {
-		if (!_.isObject(obj) || _.isArray(obj)) {
+		if (!_.isObject(obj) || Array.isArray(obj)) {
 			return false;
 		}
 
--- a/helpers/z_schema.js
+++ b/helpers/z_schema.js
@@ -208,7 +208,7 @@ var liskFormats = {
 	 * @todo Add description for the function, the params and the return value
 	 */
 	queryList(obj) {
-		if (!obj || typeof obj !== 'object' || _.isArray(obj)) {
+		if (!_.isObject(obj) || _.isArray(obj)) {
 			return false;
 		}
 
@@ -223,7 +223,7 @@ var liskFormats = {
 	 * @todo Add description for the function, the params and the return value
 	 */
 	delegatesList(obj) {
-		if (!obj || typeof obj !== 'object' || _.isArray(obj)) {
+		if (!_.isObject(obj) || _.isArray(obj)) {
 			return false;
 		}
 
--- a/helpers/z_schema.js
+++ b/helpers/z_schema.js
@@ -241,7 +241,7 @@ var liskFormats = {
 	parsedInt(value) {
 		if (
 			isNaN(value) ||
-			parseInt(value) !== Number(value) ||
+			parseInt(value).toString() !== String(value) ||
 			isNaN(parseInt(value, 10))
 		) {
 			return false;
--- a/helpers/z_schema.js
+++ b/helpers/z_schema.js
@@ -208,7 +208,7 @@ var liskFormats = {
 	 * @todo Add description for the function, the params and the return value
 	 */
 	queryList(obj) {
-		if (obj == null || typeof obj !== 'object' || _.isArray(obj)) {
+		if (!obj || typeof obj !== 'object' || _.isArray(obj)) {
 			return false;
 		}
 
@@ -223,7 +223,7 @@ var liskFormats = {
 	 * @todo Add description for the function, the params and the return value
 	 */
 	delegatesList(obj) {
-		if (obj == null || typeof obj !== 'object' || _.isArray(obj)) {
+		if (!obj || typeof obj !== 'object' || _.isArray(obj)) {
 			return false;
 		}
 
@@ -241,7 +241,7 @@ var liskFormats = {
 	parsedInt(value) {
 		if (
 			isNaN(value) ||
-			parseInt(value) != value ||
+			parseInt(value) !== Number(value) ||
 			isNaN(parseInt(value, 10))
 		) {
 			return false;
@@ -339,7 +339,7 @@ Object.keys(liskFormats).forEach(formatName => {
 });
 
 // Assigned as custom attribute to be used later
-// since z_schema.getRegisteredFormats() only resturns keys not the methods
+// since z_schema.getRegisteredFormats() only returns keys not the methods
 z_schema.formatsCache = liskFormats;
 
 // Exports
--- a/logic/peer.js
+++ b/logic/peer.js
@@ -79,7 +79,7 @@ class Peer {
 
 		peer.wsPort = this.parseInt(peer.wsPort, 0);
 
-		if (peer.httpPort != null) {
+		if (peer.httpPort) {
 			peer.httpPort = this.parseInt(peer.httpPort, 0);
 		}
 		peer.state = this.parseInt(peer.state, Peer.STATE.DISCONNECTED);
--- a/modules/dapps.js
+++ b/modules/dapps.js
@@ -141,7 +141,7 @@ __private.list = function(filter, cb) {
 	if (filter.category) {
 		const category = dappCategories[filter.category];
 
-		if (category != null) {
+		if (category !== null) {
 			where.push('"category" = ${category}');
 			params.category = category;
 		} else {
--- a/modules/loader.js
+++ b/modules/loader.js
@@ -1000,7 +1000,7 @@ Loader.prototype.findGoodPeers = function(peers) {
 	peers = peers.filter(
 		item =>
 			// Remove unreachable peers or heights below last block height
-			item != null && item.height >= lastBlockHeight
+			item !== null && item.height >= lastBlockHeight
 	);
 
 	library.logger.trace('Good peers - filtered', { count: peers.length });
--- a/modules/loader.js
+++ b/modules/loader.js
@@ -1000,7 +1000,7 @@ Loader.prototype.findGoodPeers = function(peers) {
 	peers = peers.filter(
 		item =>
 			// Remove unreachable peers or heights below last block height
-			item !== null && item.height >= lastBlockHeight
+			item && item.height >= lastBlockHeight
 	);
 
 	library.logger.trace('Good peers - filtered', { count: peers.length });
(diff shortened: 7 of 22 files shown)


modules/dapps.js
```diff
@@ -141,7 +141,7 @@ __private.list = function(filter, cb) {
 	if (filter.category) {
 		const category = dappCategories[filter.category];
 
-		if (category != null) {
+		if (category !== null) {
```
ManuGowda: `if (category)`?

diego-G: No, because if category is `0` which is different than `null` it won't pass the control.

ManuGowda: ahh ok 👍

api/ws/rpc/connect.js
```diff
@@ -45,20 +45,19 @@ const connectSteps = {
 	addConnectionOptions: peer => {
 		const systemHeaders = System.getHeaders();
 		const queryParams = {};
-
-		if (systemHeaders.version != null) {
```
MaciejBaj: The logic is changed here, previously for every undefined and null a header wasn't assigned, now it will not be assigned for every falsy value ("", NaN, 0, false in addition).

api/ws/rpc/connect.js
```diff
@@ -45,20 +45,19 @@ const connectSteps = {
...
-		if (systemHeaders.httpPort != null) {
+		if (systemHeaders.httpPort !== null) {
 			queryParams.httpPort = systemHeaders.httpPort;
 		}
-		if (systemHeaders.nethash != null) {
+		if (systemHeaders.nethash !== null) {
 			queryParams.nethash = systemHeaders.nethash;
 		}
-		if (systemHeaders.nonce != null) {
+		if (systemHeaders.nonce !== null) {
```
yatki: ❤️

MaciejBaj: So the equivalent to the previous logic would be:
```
if (systemHeaders.httpPort !== null || systemHeaders.httpPort !== undefined) {
```

api/ws/workers/middlewares/handshake.js
```diff
@@ -157,7 +157,11 @@ var extractHeaders = function(request) {
 	headers.ip = request.remoteAddress.split(':').pop();
 	headers.httpPort = +headers.httpPort;
 	headers.wsPort = +headers.wsPort;
-	if (headers.height != null) {
+
+	if (
+		(headers.height && headers.height !== null) ||
```
MaciejBaj: So if `headers.height` is true, it wasn't null or undefined so the following checks are redundant. 

Should be `headers.height !== undefined || headers.height !== null`

api/ws/rpc/connect.js
```diff
@@ -45,20 +45,22 @@ const connectSteps = {
 	addConnectionOptions: peer => {
 		const systemHeaders = System.getHeaders();
 		const queryParams = {};
-
-		if (systemHeaders.version != null) {
+		if (systemHeaders.version !== null && systemHeaders.version !== undefined) {
```
ManuGowda: why not just `if(systemHeaders.version)` ?
reference: https://runkit.com/1manugowda/5bf7fc40fbd88200127181b8

api/ws/rpc/connect.js
```diff
@@ -45,20 +45,22 @@ const connectSteps = {
...
 			queryParams.version = systemHeaders.version;
 		}
-		if (systemHeaders.wsPort != null) {
+		if (systemHeaders.wsPort !== null && systemHeaders.wsPort !== undefined) {
```
ManuGowda: same here and wherever we have the check for `null` and `undefined` explicitly in if in below code.

api/ws/rpc/connect.js
diego-G: https://github.com/LiskHQ/lisk/pull/2562#discussion_r235380896

api/ws/workers/middlewares/handshake.js
```diff
@@ -157,7 +157,8 @@ var extractHeaders = function(request) {
 	headers.ip = request.remoteAddress.split(':').pop();
 	headers.httpPort = +headers.httpPort;
 	headers.wsPort = +headers.wsPort;
-	if (headers.height != null) {
+
+	if (headers.height !== null && headers.height !== undefined) {
```
MaciejBaj: 👍

helpers/http_api.js
```diff
@@ -182,7 +182,7 @@ var middleware = {
 
 				if (
 					isNaN(value) ||
-					parseInt(value) != value ||
+					parseInt(value) !== Number(value) ||
```
MaciejBaj: That's tricky, so we want to check here if the passed number is not a Number - then return just its value, but parse to Int if it's a number (mind that it's impossible now to pass decimal numbers through API - they get converted to Integers because of `parseInt`). 

I would rather use:
```
if (_.isNumber(value)) { 
   return +value; //Changes integer strings to integers, decimal numbers in strings to decimal numbers
}
return value;
```
@diego-G @nazarhussain please tell me what do you think.

api/ws/rpc/connect.js
MaciejBaj: Your suggestion @ManuGowda returns false for the following: `"", false, 0, undefined, null`, whereas the code below only for `undefined, null`.

api/ws/rpc/connect.js
MaciejBaj: Your suggestion @ManuGowda returns false for the following: `"", false, 0, undefined, null`, whereas the code below only for `undefined, null`.

helpers/z_schema.js
```diff
@@ -208,7 +208,7 @@ var liskFormats = {
 	 * @todo Add description for the function, the params and the return value
 	 */
 	queryList(obj) {
-		if (obj == null || typeof obj !== 'object' || _.isArray(obj)) {
+		if (!obj || typeof obj !== 'object' || _.isArray(obj)) {
```
MaciejBaj: Same as explained above, changing obj == null to !obj is not correct.

Here the condition is checking if a value is not an object or an array, we can do the same by:
`if(!_.isObject(obj) || _.isArray(obj))` or even simpler by `if (_.isArray(obj))`. Explanation:
```
typeof null
"object"
typeof undefined
"undefined"
typeof {}
"object"
typeof [] 
"object"

_.isObject(null)
false
_.isObject(undefined)
false
_.isObject({})
true

_.isArray([])
true
_.isArray(null)
false
_.isArray(undefined)
false
_.isArray({})
false
``

File: helpers/http_api.js (hunk 8 above)
Comment (diego-G) [reply helpers/http_api.js]: I got in trouble with this one also because the custom schema validator `parsedInt` is never used. Please @nazarhussain shed some light when you can.

File: helpers/z_schema.js [hunk 10]
```diff
@@ -223,7 +223,7 @@ var liskFormats = {
 	 * @todo Add description for the function, the params and the return value
 	 */
 	delegatesList(obj) {
-		if (obj == null || typeof obj !== 'object' || _.isArray(obj)) {
+		if (!obj || typeof obj !== 'object' || _.isArray(obj)) {
```
MaciejBaj: Same here (https://github.com/LiskHQ/lisk/pull/2562/files#r235954385)

helpers/z_schema.js
```diff
@@ -241,7 +241,7 @@ var liskFormats = {
 	parsedInt(value) {
 		if (
 			isNaN(value) ||
-			parseInt(value) != value ||
+			parseInt(value) !== Number(value) ||
```
MaciejBaj: That's incorrect (https://github.com/LiskHQ/lisk/pull/2562/files#r235949316).

modules/loader.js
```diff
@@ -995,7 +995,7 @@ Loader.prototype.findGoodPeers = function(peers) {
 	peers = peers.filter(
 		item =>
 			// Remove unreachable peers or heights below last block height
-			item != null && item.height >= lastBlockHeight
+			item !== null && item.height >= lastBlockHeight
```
MaciejBaj: That's incorrect - the previous check was checking the item against both of `null` and `undefined` values as explained in https://github.com/LiskHQ/lisk/pull/2562/files#r235950284, so that `item.height` doesn't result with `Uncaught ReferenceError: item is not defined`. 
In this case, we can replace this check with:
`item && item.height >= lastBlockHeight`, as we know that the height of the block cannot be equal to 0.

test/common/helpers/api.js
```diff
@@ -82,7 +82,7 @@ var http = {
 function paramsHelper(url, params) {
 	if (
 		typeof params !== 'undefined' &&
```
MaciejBaj: There is no difference between `typeof params !== 'undefined'` and `params !== undefined`

helpers/z_schema.js
yatki: @MaciejBaj `typeof obj !== 'object' || Array.isArray(obj)` I'd prefer to use native solutions instead of depending on lodash. What do you think?

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray

api/ws/rpc/connect.js
diego-G: I want to keep the behaviour while making the statement easier to understand. Please, raise an issue if you think we need to rewrite/clean that function.

helpers/http_api.js
diego-G: @yatki that statement doesn't work for every of these cases https://github.com/LiskHQ/lisk/blob/23fea4774d4baf53d4b904fbe77b49f2cc007680/test/unit/helpers/z_schema.js#L405

nazarhussain: So here we are checking if the value of some attributes are not (integers or integer formatted strings). Because by default with express every query param is treated as strings. 

So original values were with `radix=10`

```
isNaN(value) || parseInt(value) != value || isNaN(parseInt(value, radix))

isNaN(value) - Value of param is not a number
parseInt(value) != value - Parsed integer value not same as original 
isNaN(parseInt(value, radix)) - Parsing to integer with 10 base is not possible 
```

If any of above condition meet we want to preserve the value in the same format. e.g. If a `balance` query param came as string value of `123454` it will be converted to integer equivalent value. But if `balance` query param came as string value of `1234.54` it will left out as it is because `parseInt(value) != value` condition matches. 

And this was the exact reason we used `!=` instead of `!==`, so we can match strings directly with integers. So would be nice to actually do double conversion and match against same number. 

```
parseInt(value).toString() !== String(value)
```

nazarhussain:
```suggestion
					parseInt(value).toString() !== String(value) ||
```

helpers/z_schema.js
MaciejBaj: I couldn't agree more @yatki

logic/peer.js
```diff
@@ -79,7 +79,7 @@ class Peer {
 
 		peer.wsPort = this.parseInt(peer.wsPort, 0);
 
-		if (peer.httpPort != null) {
+		if (peer.httpPort !== null && peer.httpPort !== undefined) {
```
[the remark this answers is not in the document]
MaciejBaj: Yep, ok let's change it here and in the following cases to the simple form as suggested.
Both of ws and http port needs to be greater or equal 1:
```
	wsPort: {
				type: 'integer',
				minimum: 1,
				maximum: 65535,
			},
```

helpers/z_schema.js
```diff
@@ -208,7 +208,7 @@ var liskFormats = {
 	 * @todo Add description for the function, the params and the return value
 	 */
 	queryList(obj) {
-		if (obj == null || typeof obj !== 'object' || _.isArray(obj)) {
+		if (!_.isObject(obj) || _.isArray(obj)) {
```
yatki: @diego-G can you replace _.isArrays to Array.isArray ?

lodash function does the following check:
```javascript
function isArrayLike(value) {
      return value != null && isLength(value.length) && !isFunction(value);
    }
```
That's why I prefer to use native is array function.

yatki: @MaciejBaj @diego-G just fyi guys, `obj === Object(obj)` this check is behaving as you expected. See: https://stackoverflow.com/a/14706877/1417536 

underscore.js was using this solution.

yatki: I'm also ok to create another issue for these usages If you agree to change these.

helpers/http_api.js
MaciejBaj: So, as query parameters, we only want to accept integers. No digits allowed as every value should come in Beddows. 

`parseInt(value).toString() !== String(value)` keeps the behaviour the same.

helpers/z_schema.js
diego-G: Yes, please. I prefer to open another issue for that.

helpers/z_schema.js
MaciejBaj: As mentioned above please change the whole condition to:
```
parseInt(value).toString() !== String(value)
```

helpers/z_schema.js
```diff
@@ -223,7 +223,7 @@ var liskFormats = {
 	 * @todo Add description for the function, the params and the return value
 	 */
 	delegatesList(obj) {
-		if (obj == null || typeof obj !== 'object' || _.isArray(obj)) {
+		if (!_.isObject(obj) || _.isArray(obj)) {
```
MaciejBaj:
```suggestion
		if (!_.isObject(obj) || Array.isArray(obj)) {
```

test/common/helpers/api.js
```diff
@@ -82,8 +82,8 @@ var http = {
 
 function paramsHelper(url, params) {
 	if (
-		typeof params !== 'undefined' &&
-		params != null &&
+		params !== 'undefined' &&
```
MaciejBaj:
```suggestion
		typeof params !== 'undefined' &&
```

PR #2565, opened 2018-11-22
Apply eslint-rule consistent-return - Related to #1597

diego-G:
### Review checklist

* The PR is related to #1597
* Linting passes
* Tests pass

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -19,6 +19,7 @@ docs/jsdoc/
 
 # Logs
 logs/*.log
+logs/**/*.log
 npm-debug.log
 tmux-client-*.log
 test/network/logs/
@@ -40,3 +41,4 @@ test/.coverage-unit/
 
 # Local config file useful for development
 config.local.json
+lisk.json
--- a/modules/peers.js
+++ b/modules/peers.js
@@ -324,7 +324,7 @@ __private.insertSeeds = function(cb) {
 
 			// Continue if peer it is not blacklisted nor banned
 			if (peer.state !== Peer.STATE.BANNED) {
-				peer.rpc.status((err, status) => {
+				return peer.rpc.status((err, status) => {
 					__private.updatePeerStatus(err, status, peer);
 					if (!err) {
 						updated += 1;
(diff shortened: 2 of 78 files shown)


.gitignore
```diff
@@ -40,3 +41,4 @@ test/.coverage-unit/
 
 # Local config file useful for development
 config.local.json
+lisk.json
```
nazarhussain: Who introduced this file? and why we need to add it to gitignore?

.gitignore
```diff
@@ -19,6 +19,7 @@ docs/jsdoc/
 
 # Logs
 logs/*.log
+logs/**/*.log
```
nazarhussain: The rule `logs/*` will cover both of above lines. And we don't need anything else in log directory.

.gitignore
diego-G: It is a file usually created by users who run the application on `pm2`. It shouldn't be a problem.

nazarhussain: Its not a strict convention to have pm2 file with name of `pm2`, and if some person run lisk with `pm2` why he would ever commit his files? This will limit us if in future we want to introduce a file with this name for any purpose.

So until any file is already used in our workflow, we should not add it to our version control. e.g. I always use `config.local.json` file and I always run `node app -c config.local.json`, but its never intended that I add that file in version control.

nazarhussain: I just tried `pm2 start app.js --name lisk`, so that file is not created automatically by `pm2.

MaciejBaj:
@diego-g please resolve the conflicts: `app.js`

modules/peers.js
```diff
@@ -324,7 +324,7 @@ __private.insertSeeds = function(cb) {
 
 			// Continue if peer it is not blacklisted nor banned
 			if (peer.state !== Peer.STATE.BANNED) {
-				peer.rpc.status((err, status) => {
+				return peer.rpc.status((err, status) => {
 					__private.updatePeerStatus(err, status, peer);
```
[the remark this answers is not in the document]
diego-G: Unrelated to the issue. I resolve the conversation.

PR #2581, opened 2018-11-26
Unify indexOf & includes usage in exception handling - Closes #2436

### LINKED ISSUE #2436: Unify indexOf & includes usage in exception handling
### Expected behavior
Exceptions handling must be consistent across the application and specially the transaction types. e.g.  Consistently use of exception in `verify` and `apply*` methods.  

### Actual behavior

Currently handling of exceptions is done on random places and in different patterns. 

https://github.com/LiskHQ/lisk/blob/4ae8553042b93d3e7ce41c11a7d910fef24ba644/logic/multisignature.js#L140-L142

https://github.com/LiskHQ/lisk/blob/4ae8553042b93d3e7ce41c11a7d910fef24ba644/logic/vote.js#L251-L254

1.1.0

yatki:
### What was the problem?
Exception handling has been spread throughout the application and not handled from a central place. Also includes and indexOf changes were being used inconsistently.
### How did I fix it?
After discussing with @nazarhussain, I decided that we can't combine all exception handling into a central logic due to complexity of the current architecture. I only replaced inconsistent indexOf usages and used `includes` in all checks.
### Review checklist

* The PR resolves #2436 
* Linting passes
* Tests pass

### DIFF
--- a/logic/multisignature.js
+++ b/logic/multisignature.js
@@ -136,7 +136,7 @@ Multisignature.prototype.verify = function(transaction, sender, cb) {
 		const err =
 			'Invalid multisignature min. Must be less than or equal to keysgroup size';
 
-		if (exceptions.multisignatures.indexOf(transaction.id) > -1) {
+		if (exceptions.multisignatures.includes(transaction.id)) {
 			library.logger.debug(err);
 			library.logger.debug(JSON.stringify(transaction));
 		} else {
--- a/logic/transaction.js
+++ b/logic/transaction.js
@@ -473,7 +473,7 @@ class Transaction {
 				sender.publicKey,
 			].join(' ');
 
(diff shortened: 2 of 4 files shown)


4miners:
Why `indexOf` over `includes`?

yatki:
@4miners I checked the performance difference between `indexOf` and `includes`. Seems like indexOf is more performant on chrome and since we are using node I thought we should stick with indexOf.

See:  https://stackoverflow.com/questions/47659972/array-indexof-vs-includes-perfomance-depeding-on-browser-and-needle-position

yatki:
@4miners after further investigation, I found out with latest version of V8 engine `includes` is almost as fast as `indexOf`. So I'm switching to use `includes`.

PR #2582, opened 2018-11-26
Refactor generateDelegateList & validateBlockSlot functions - Closes #2574, 2784

### LINKED ISSUE #2574: Refactor generateDelegateList & validateBlockSlot functions
### Expected behavior
1) generateDelegateList function does not need a source argument. The logic can be moved inside it as below:

```javascript
Delegates.prototype.generateDelegateList = function(round, cb, tx) {
	// Set default function for getting delegates		const lastBlockRound = slots.calcRound(modules.blocks.lastBlock.get().height);
	source = source || __private.getKeysSortByVote;		const source =
		lastBlockRound > round
			? 'getDelegatesFromPreviousRound'
			: 'getKeysSortByVote';
 	if (__private.delegatesListCache[round]) {
		library.logger.debug('Using delegate list from the cache for round', round);
		return setImmediate(cb, null, __private.delegatesListCache[round]);
	}
 	source((err, truncDelegateList) => {		__private[source]((err, truncDelegateList) => {
```

2) validateBlockSlot functions in most cases just passing it's own arguments to `modules.delegates.validateBlockslot` function. We can call `modules.delegates.validateBlockslot` directly and we can get rid of redundant wrappers.

Implementation is already done in this branch: https://github.com/LiskHQ/lisk/tree/maintain-delegate-list-backup.

3) We should also rename generateDelegateList to getForgersList to make it easier to understand what it does.
### Actual behavior
- generateDelegateList function has a redundant source argument.
- `modules.delegates.validateBlockslot` has redundant wrappers.
- generateDelegateList function is a misleading name for that method. I prefer getForgersList

### LINKED ISSUE #2784: Fix getting delegate list logic in validateBlockSlot function
### Expected behavior
When fork 5 happens if the last block of the node was also last block of the round, we should use the `getDelegatesFromPreviousRound` function to get delegates from the snapshot.

### Actual behavior

The logic we have is wrong and even it's an edge case, it's a bug at the moment.

https://github.com/LiskHQ/lisk/blob/e81cb2af687b2e3a4f3bd8e159d44c4750e42166/modules/blocks/process.js#L618-L645

yatki:
### What was the problem?
1) generateDelegateList function does not need a source argument. 
2) modules.delegates.validateBlockslot method had redundant wrappers.
### How did I fix it?
1) I removed source argument from generateDelegateList function and refactored unit tests. 
2) I removed redundant wrappers.
### Review checklist

* The PR resolves #2574, resolves #2784
* Linting passes
* Tests pass


modules/delegates.js
```diff
@@ -687,17 +652,19 @@ Delegates.prototype.updateForgingStatus = function(
...
  *
  * @param {number} round
- * @param {function} source - Source function for get delegates
  * @param {function} cb - Callback function
  * @param {Object} tx - Database transaction/task object
  * @returns {setImmediateCallback} cb, err, truncated delegate list
  * @todo Add description for the params
  */
-Delegates.prototype.generateDelegateList = function(round, source, cb, tx) {
+Delegates.prototype.getForgersList = function(round, cb, tx) {
```
4miners: I don't like the name, the function is sometimes (when cache not available) really generating the list and not just getting.

modules/delegates.js
```diff
@@ -687,17 +652,19 @@ Delegates.prototype.updateForgingStatus = function(
...
 	// Set default function for getting delegates
-	source = source || __private.getKeysSortByVote;
-
-	source((err, truncDelegateList) => {
+	const lastBlockRound = slots.calcRound(modules.blocks.lastBlock.get().height);
+	const source =
```
4miners: So it's not possible to make this function to decide which delegates list its need to return based only on `round` parameter. Because when we call it from inside of `receiveForkFive` - we need to validate against previous round when `lastBlock.height % ACTIVE_DELEGATES === 0` is `true`, but when we call it from a different source we need to return list for current round always. However, if we pass `height` instead of `round` as a parameter we can do that. 🤗

modules/delegates.js
```diff
@@ -687,17 +652,19 @@ Delegates.prototype.updateForgingStatus = function(
...
+		lastBlockRound > round
```
4miners: As I described in previous comments this condition `lastBlockRound > round` will never be `true` because it's always `lastBlockRound <= round`.

yatki:
@MaciejBaj fork 5 and fork 1 cases already covered by integration tests, that's why I added a unit test for the change I introduced!

modules/delegates.js
```diff
@@ -183,51 +183,19 @@ __private.getDelegatesFromPreviousRound = function(cb, tx) {
...
-	});
-};
-
 /**
  * Gets the assigned delegate to current slot and returns its keypair if present.
  *
  * @private
  * @param {number} slot
- * @param {number} round
+ * @param {block} block
```
lsilvs: This param expects block height, right?
If so, please rename it to `blockHeight` and type number instead.

modules/delegates.js
```diff
@@ -720,23 +688,27 @@ Delegates.prototype.updateForgingStatus = function(
...
 	if (__private.delegatesListCache[round]) {
-		library.logger.debug('Using delegate list from the cache for round', round);
 		return setImmediate(cb, null, __private.delegatesListCache[round]);
 	}
 
-	return source((err, truncDelegateList) => {
+	const lastBlockHeight = modules.blocks.lastBlock.get().height;
+	// Set default function for getting delegates
+	const source =
+		blockHeight === lastBlockHeight && lastBlockHeight % ACTIVE_DELEGATES === 0
```
lsilvs: I understand that this condition is correct and the old one (`lastBlock.height % ACTIVE_DELEGATES === 0 || roundLastBlock < roundNextBlock`) is wrong but I would open a separate issue to fix this as it changes behaviour and not only refactoring.

modules/delegates.js
```diff
@@ -720,23 +688,27 @@ Delegates.prototype.updateForgingStatus = function(
...
+			? 'getDelegatesFromPreviousRound'
+			: 'getKeysSortByVote';
```
lsilvs: I would rename this method to `getDelegatesForCurrentRound`

modules/delegates.js
```diff
@@ -720,23 +688,27 @@ Delegates.prototype.updateForgingStatus = function(
...
+ * @param {number} blockHeight
  * @param {function} cb - Callback function
  * @param {Object} tx - Database transaction/task object
  * @returns {setImmediateCallback} cb, err, truncated delegate list
  * @todo Add description for the params
  */
-Delegates.prototype.generateDelegateList = function(round, source, cb, tx) {
-	// Set default function for getting delegates
-	source = source || __private.getKeysSortByVote;
+Delegates.prototype.generateDelegateList = function(blockHeight, cb, tx) {
```
lsilvs: You haven't renamed this method to `getForgersList` as the PR description states.

yatki: Yes you are right, we decided to keep it this way after discussing it with @4miners. I updated pr description and I commented on the original issue.

modules/delegates.js
yatki: Hmm, you are right but it's really hard to decouple this change from the refactoring. Maybe it would be better to update the original issue and PR description ?

modules/delegates.js
yatki: Good catch, thanks.

modules/delegates.js
yatki: @4miners could you please comment on this ?

modules/delegates.js
yatki: @lsilvs created the issue: https://github.com/LiskHQ/lisk/issues/2784, and updated the PR.

PR #2589, opened 2018-11-27
Apply eslint-rule no-nested-ternary - Related to #1597

diego-G:
### Review checklist

* The PR is related to #1597
* Linting passes
* Tests pass


api/ws/rpc/connect.js
```diff
@@ -125,10 +125,9 @@ const connectSteps = {
 			(peerExtendedWithRPC, localHandler, rpcProcedureName) => {
 				peerExtendedWithRPC.rpc[rpcProcedureName] = (data, rpcCallback) => {
 					// Provide default parameters if called with non standard parameter, callback
+					const auxRpcCallback = typeof data === 'function' ? data : () => {};
 					rpcCallback =
-						typeof rpcCallback === 'function'
-							? rpcCallback
-							: typeof data === 'function' ? data : () => {};
+						typeof rpcCallback === 'function' ? rpcCallback : auxRpcCallback;
 					data = data && typeof data !== 'function' ? data : {};
```
yatki: I think we can simplify the logic between lines 128-131 by:
```suggestion
					if (typeof data === 'function') {
						rpcCallback = data;
						data = {};
					}
```

I'm not even sure if we should set a default value like `() => {}` for rpcCallback. I think if not a valid callback provided we should throw an error since we are using setImmediate to return the result. But if it's necessary we can add the following condition on top of the above condition:

```suggestion
					if (typeof rpcCallback !== 'function') {
						rpcCallback = () => {};
					}
```

yatki: I can understand if this can be considered as not scope of this issue. But since PR is relatively small maybe we can refactor these.

modules/peers.js
```diff
@@ -685,13 +686,13 @@ Peers.prototype.list = function(options, cb) {
 				peersList = peersList.filter(peer => {
```
MaciejBaj:
```suggestion
					// Skip banned peers by default
				peersList = peersList.filter(peer => allowedStates.indexOf(peer.state) !== -1);
				// Filter peers by broadhash if present
				if (broadhash) {
					if (attempt === 0) {
						// Look for peers matching (equal to) broadhash with the first attempt
						peersList = peersList.filter(peer => peer.broadhash === broadhash);
					} else if (attempt === 1) {
						// Look for peers unmatching (not equal to) broadhash with the second attempt
						peersList = peersList.filter(peer => peer.broadhash !== broadhash);
					}
				}
```

MaciejBaj: I think looks more readable like this

api/ws/rpc/connect.js
```diff
@@ -122,14 +122,18 @@ const connectSteps = {
...
-					data = data && typeof data !== 'function' ? data : {};
+					if (!data || typeof data === 'function') {
+						data = {};
+					}
+					if (typeof rpcCallback !== 'function') {
+						if (typeof data !== 'function') {
+							rpcCallback = {};
+						}
+						rpcCallback = data;
+					}
```
yatki: I think data assignment at line 129 is a breaking change and this implementation can be simplified like this:
```suggestion
	if (typeof rpcCallback !== function) {
	  rpcCallback = () => {};
	}
	
	if (typeof data === 'function' ) {
	   rpcCallback = data;
	   data = {};
	}
```

test/integration/rounds.js
```diff
@@ -218,22 +218,26 @@ describe('rounds', () => {
...
-								aPK === null ? -1 : 1;
+						: auxBufferComp;
 
+				const auxAvoteLt = aVote.gt(bVote) ? 1 : bufferComp;
 				// Compare vote weights first:
 				// if first is less than second - return -1,
 				// if first is greather than second - return 1,
 				// if both are equal - compare public keys
-				return aVote.lt(bVote) ? -1 : aVote.gt(bVote) ? 1 : bufferComp;
+				return aVote.lt(bVote) ? -1 : auxAvoteLt;
```
yatki:
```suggestion
				.sort((a, b) => {
				const aVote = new Bignum(accounts[a].vote);
				const bVote = new Bignum(accounts[b].vote);
				const aPK = accounts[a].publicKey;
				const bPK = accounts[b].publicKey;

				// Compare vote weights first:
				// if first is less than second - return -1,
				// if first is greather than second - return 1,
				// if both are equal - compare public keys

				if (aVote.lt(bVote)) {
					return -1;
				}

				if (aVote.gt(bVote)) {
					return 1;
				}

				// If both are buffers
				// Return result of the compare
				if (aPK && bPK) {
					return Buffer.compare(bPK, aPK);
				}

				// If both are null - return 0
				if (aPK === null && bPK === null) {
					return 0;
				}

				// If first is null - return -1,
				if (aPK === null) {
					return -1;
				}

				// if not return 1;
				return 1;
			})
```

api/ws/rpc/connect.js
```diff
@@ -121,14 +121,19 @@ const connectSteps = {
 		// Register RPC methods on peer
 		peer = _.reduce(
 			wsServer.endpoints.rpc,
-			(peerExtendedWithRPC, localHandler, rpcProcedureName) => {
+			(peerExtendedWithRPC, _localHandler, rpcProcedureName) => {
```
MaciejBaj: Why is `localHandler` changed to `_ localHandler`? I don't see any changes that would re-use that variable here. Please don't use underscores.

ManuGowda: when a variable is not used or if its an optional parameter, in most programming(that I know of) use `_` as convention and even `eslint` says the same https://eslint.org/docs/rules/no-unused-vars#argsignorepattern
so what @diego-G has done is correct.

diego-G: https://eslint.org/docs/rules/no-unused-vars#argsignorepattern

shuse2: I think when the argument is not used, `_` is often use but putting the underscore in front is different thing. In this case, it should've been `(peerExtendedWithRPC, _, rpcProcedureName) => {`
However, if you use lodash it conflicts.

FYI, in elements, we decided to use `_` prefix for private member in class, but in othercases we never use `_` prefix.
If it's non-used argument, we use (input1, _, input3)

ManuGowda: @shuse2 if we are following `eslint` conventions, shouldn't we stick to the conventions? here https://eslint.org/docs/rules/no-unused-vars#argsignorepattern they clearly say use `_` in front of the variable, however, I agree in some other languages only `_` is used when the variable is not used, but as `eslint` suggest we can stick to `_` with variable name convention rather than our own.

shuse2: I think this rule is only suggesting you can setup the pattern to ignore, and this is usable if you setup like `/*eslint no-unused-vars: ["error", { "argsIgnorePattern": "^_" }]*/`.
Also, I think by default, if second parameter is not used but third parameter is used, then it shouldn't complain

diego-G: What it is clear is that having a tag to know if a var is used or not is something suggested by eslint itself. IMHO having just an underscore is meaningless. Eslint suggests by default just adding an underscore at the beginning. I don't see the big deal. It is just a convention we can adopt or not. I've done it because the IDE suggests me to do it and I've found it useful.

ManuGowda: either only `_` or `_` prefixed with name, both patterns are fine with me, it's only a choice now to pick which one.

diego-G: @MaciejBaj in any case, this is not related to this PR. If it really bothers you I can remove it from here and from any other PR, and I will open another proper PR just for the case.

sdrpa:
> Why is `localHandler` changed to `_ localHandler`? I don't see any changes that would re-use that variable here. Please don't use underscores.

I also find confusing using underscore in front of a variable name, it makes me think it's private member variable. In this case, taking in account @shuse2 reply regarding lodash conflict, maybe the best option would be to use ```import reduce from 'lodash/reduce'``` to avoid _.reduce but use reduce directly (since reduce seems to be the only lodash function used in the file) and instead of (peerExtendedWithRPC, _localHandler, rpcProcedureName) use (peerExtendedWithRPC, _, rpcProcedureName).  Anyway, you're probably end up using es6 reduce instead of lodash version soon.

api/ws/rpc/connect.js
MaciejBaj: @diego-G @ManuGowda my question https://github.com/LiskHQ/lisk/pull/2589/files#r237434813 still remains not answered. I still don't know why you decided to change this very one `localHandle` variable name in this PR.

> when a variable is not used or if its an optional parameter, in most programming(that I know of) use `_` as convention and even `eslint` says the same https://eslint.org/docs/rules/no-unused-vars#argsignorepattern
> so what @diego-G has done is correct.

@ManuGowda you are referring to Eslint rule which wasn't accepted to implement as part of  https://github.com/LiskHQ/lisk/issues/1597, following the airbnb standards.





> What it is clear is that having a tag to know if a var is used or not is something suggested by eslint itself. IMHO having just an underscore is meaningless. Eslint suggests by default just adding an underscore at the beginning. I don't see the big deal. It is just a convention we can adopt or not. I've done it because the IDE suggests me to do it and I've found it useful.

Part of the issue #1597 you're solving now is applying the rule: ` "no-underscore-dangle": 161`, which makes the declaration `let _localHandler = "LH"` invalid, and `let _ = "LH"` valid. The reference - https://eslint.org/docs/rules/no-underscore-dangle.

To reiterate:
- Don't implement the rule `/*eslint no-unused-vars: ["error", { "argsIgnorePattern": "^_" }]*/`, which conflicts with rule "no-underscore-dangle", which agreed as part of #1597.
- Don't include random changes of other eslint rules in this (`Apply eslint-rule no-nested-ternary`) PR.

PR #2590, opened 2018-11-27
Apply eslint-rule no-bitwise - Related to #1597

diego-G:
As exception, the rule is disabled on bignumber helper.

### Review checklist

* The PR is related to #1597
* Linting passes
* Tests pass

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -24,7 +24,6 @@
 		"guard-for-in": "off",
 		"generator-star-spacing": ["error", { "before": false, "after": false }],
 		"max-len": "off",
-		"no-bitwise": "off",
 		"no-nested-ternary": "off",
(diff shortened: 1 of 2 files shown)


diego-G:
@MaciejBaj I was thinking pretty much the same but later I thought what we are doing in that file is creating a customised low-level function that it's not recommended. Actually, we had plans to remove that file and use another library like Elements did. What do you think @shuse2 ? 

Another fact is if @ManuGowda and @willclarktech introduced the action point once, I believe there was a strong reasoning behind.

shuse2:
I think In general, doing bitwise operation in node isn't good idea.
There could be a special case (such as BigNum), but in that case we can intentionally disable it.
If there are too much code relates to that, we can discuss it at that point.

PR #2596, opened 2018-11-28
Fix get last commit helper when git is not installed - Close #2588

### LINKED ISSUE #2588: getLastCommit fails if git is not installed
### Expected behavior
The `getLastCommit()` function from `helpers/git.js` should work even if the `git` command is not installed (it should read the content of the `REVISION` file).
IMO it'd make sense to try to read the file first and fallback to git.

### Actual behavior
If `git` cannot be called then `spawn.stderr` (and `spawn.sdout`) is `null` and a `TypeError` is raised; the exception is not logged, `REVISION` is not read and the API returns `""` (empty string).

### Steps to reproduce
Run core on a machine where the `git` is not available; call the API (`/api/node/contants`) and look for `.data.commit`.

`>=1.3.0`, likely older versions as well.

diego-G:
### How to test it?

```
mocha test/unit/helpers/git.js
mocha test/functional/http/get/node/node.js
```

### Review checklist

* The PR resolves #2588 
* Linting passes
* Tests pass

### DIFF
--- a/helpers/git.js
+++ b/helpers/git.js
@@ -37,7 +37,7 @@ function getLastCommit() {
 	try {
 		spawn = childProcess.spawnSync('git', ['rev-parse', 'HEAD']);
 		// If there is git tool available and current directory is a git owned directory
-		if (!spawn.stderr) {
+		if (!spawn.stderr.toString().trim()) {
 			return spawn.stdout.toString().trim();
 		}
 	} catch (error) {
--- a/helpers/git.js
+++ b/helpers/git.js
@@ -14,8 +14,8 @@
 
 'use strict';
 
-var childProcess = require('child_process');
-var fs = require('fs');
+const childProcess = require('child_process');
+const fs = require('fs');
 
 /**
  * Helper module for parsing git commit information.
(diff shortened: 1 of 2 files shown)


helpers/git.js
```diff
@@ -32,12 +32,19 @@ var fs = require('fs');
...
-	var err = spawn.stderr.toString().trim();
+	let spawn;
 
-	// If there is git tool available and current directory is a git owned directory
-	if (!err) {
-		return spawn.stdout.toString().trim();
+	try {
+		spawn = childProcess.spawnSync('git', ['rev-parse', 'HEAD']);
+		// If there is git tool available and current directory is a git owned directory
+		if (!spawn.stderr) {
```
nazarhussain: As per documentation `stderr` will always be returned as string or Buffer. 
https://nodejs.org/api/child_process.html#child_process_child_process_spawnsync_command_args_options . In case of empty string this logic may work, but in case of empty buffer it will fail. So better o use the same logic which was used earlier `!pawn.stderr.toString().trim()`

helpers/git.js
```diff
@@ -32,12 +32,19 @@ var fs = require('fs');
...
-	var err = spawn.stderr.toString().trim();
+	let spawn;
 
-	// If there is git tool available and current directory is a git owned directory
-	if (!err) {
-		return spawn.stdout.toString().trim();
+	try {
+		spawn = childProcess.spawnSync('git', ['rev-parse', 'HEAD']);
+		// If there is git tool available and current directory is a git owned directory
+		if (!spawn.stderr.toString().trim()) {
```
fchavant: This still raises a `TypeError: Cannot read property 'toString' of null`. The issue persists.

nazarhussain: Yes that problem really exits. Easy to test is to run in node console. 

```
cpr.spawnSync('agit', ['rev-parse', 'HEAD']).stderr`
```

nazarhussain: Yes that problem really exits. Easy to test is to run in node console. 

```
node 
cpr.spawnSync('agit', ['rev-parse', 'HEAD']).stderr`
```

PR #2603, opened 2018-11-30
Apply eslint-rule no-var- Related to #1597

diego-G:
### Review checklist

* The PR is related to #1597
* Linting passes
* Tests pass


api/controllers/node.js
```diff
@@ -152,9 +152,9 @@ NodeController.updateForgingStatus = function(context, next) {
 		return next(new Error('Access Denied'));
 	}
 
-	var publicKey = context.request.swagger.params.data.value.publicKey;
-	var password = context.request.swagger.params.data.value.password;
-	var forging = context.request.swagger.params.data.value.forging;
+	const publicKey = context.request.swagger.params.data.value.publicKey;
+	const password = context.request.swagger.params.data.value.password;
+	const forging = context.request.swagger.params.data.value.forging;
```
yatki: @diego-G @MaciejBaj I don't like to suggest stuff outside of the scope of the PR but changing 3 lines above to: 
```suggestion
const { publicKey, password, forging } = context.request.swagger.params.data.value;
```
would be nicer.

yatki: Let's ignore the comment above. I think we will fix those when we implement `prefer-destructuring`

PR #2607, opened 2018-12-04
Do not process bundled transactions while node is syncing - Closes #2398

### LINKED ISSUE #2398: During the syncing process, do not process bundled transactions
### Expected behavior
During the syncing process, the application is trying to catch up to the chain, so it should finish the syncing process before processing bundled transactions because processing transactions is an expensive task and it has less priority.
### Actual behavior
During the syncing process, application switches between loading blocks from peers and verifying bundled transactions through the `nextBundle` job. 
### Steps to reproduce
N/A
1.0.0

diego-G:
### How to test it?

` mocha test/unit/logic/transaction_pool.js `

### Review checklist

* The PR resolves #2398 
* Linting passes
* Tests pass


logic/transaction_pool.js
```diff
@@ -513,7 +513,12 @@ TransactionPool.prototype.reindexQueues = function() {
 TransactionPool.prototype.processBundled = function(cb) {
 	const bundled = self.getBundledTransactionList(true, self.bundleLimit);
 
-	library.balancesSequence.add(
+	if (modules.loader.syncing()) {
```
4miners: Why here and not in transport module? We should not allow receiving instead of processing, because the node will accumulate a lot of transactions during sync.

diego-G: We were discussing about that finally rejecting it because when the node enters into sync process due to a fork, it has to be able to receive and broadcast transactions. At least, this is how it works now. Otherwise, it will produce a behavioural change.

logic/round.js
```diff
@@ -222,7 +222,7 @@ class Round {
 						}
 					});
 				}
-				return Promise.resolve();
```
MaciejBaj: ?

test/network/scenarios/p2p/peers.js
```diff
@@ -82,7 +82,7 @@ module.exports = function(configurations, network) {
 						// to update across all nodes.
 						setTimeout(() => {
 							resolve();
-						}, 5000);
+						}, 10000);
```
MaciejBaj: ?

logic/round.js
diego-G: @yatki strongly suggested to return null in this cases instead of resolving promises. Test became unstable since we introduce this.

test/network/scenarios/p2p/peers.js
diego-G: Network test suite is failing due to the security time to wait for some actions is not enough. I've seen the same problem in propagation scenario.

PR #2612, opened 2018-12-05
Add unit tests for updateDelegateListCache & clearLastDelegateListCache methods - Closes #2593

### LINKED ISSUE #2593: Add unit tests for updateDelegateListCache & clearLastDelegateListCache methods
### Expected behavior
We need unit tests for updateDelegateListCache & clearLastDelegateListCache methods in delegates module.

After further discussion with @lsilvs, to reduce complexity of the cache structure we also decided to clear the whole cache instead of clearing it partially. This change will cause max 2 extra queries but it will also increase the maintainability.

### Actual behavior
- We don't have it at the moment. 
- We clear the cache partially.

yatki:
### What was the problem?
We didn't have unit tests for updateDelegateListCache & clearLastDelegateListCache methods in delegates module.

### How did I fix it?
I added unit tests

### Review checklist

* The PR resolves #2593 
* Linting passes
* Tests pass

### DIFF
--- a/test/unit/modules/delegates.js
+++ b/test/unit/modules/delegates.js
@@ -756,6 +756,32 @@ describe('delegates', () => {
 						initialSate['1']
 					);
 				});
+
+				it('should not mutate the __private.delegateListCache object when there is only 1 cached round.', () => {
+					// Arrange
+					const initialSate = {
+						1: ['j', 'k', 'l'],
+					};
+					__private.delegatesListCache = { ...initialSate };
+
+					// Act
+					library.modules.delegates.clearLastDelegateListCache();
+
+					// Assert
+					return expect(__private.delegatesListCache).to.deep.equal(
+						initialSate
+					);
+				});
+
+				it('should not mutate the __private.delegateListCache object when there is no cached round.', () => {
+					__private.delegatesListCache = {};
+
+					// Act
+					library.modules.delegates.clearLastDelegateListCache();
+
+					// Assert
+					return expect(__private.delegatesListCache).to.deep.equal({});
+				});
 			});
 		});
 	});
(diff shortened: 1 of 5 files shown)


test/unit/modules/delegates.js
```diff
@@ -658,6 +658,96 @@ describe('delegates', () => {
 				);
 			});
 		});
+
+		describe('__private.delegatesListCache operations', () => {
+			let __private;
+			beforeEach(done => {
+				__private = library.rewiredModules.delegates.__get__('__private');
+				done();
```
4miners: We are not waiting for any async operations here, you can get rid of `done();` and just `return;`

yatki: That conflicts with eslint rules. See:
```
Unexpected synchronous test.
Unnecessary return statement.
```

Also I don't have a strong opinion on this but I'd prefer to keep it this way, since it's intuitively easier to understand and read. I mean when you look at the code it's easier to catch `done()` statements with eye than return statements. I didn't use done in tests to keep it consistent but the with architecture I'd like to use `done()`more often.

test/unit/modules/delegates.js
```diff
@@ -756,6 +756,32 @@ describe('delegates', () => {
 						initialSate['1']
 					);
 				});
+
+				it('should not mutate the __private.delegateListCache object when there is only 1 cached round.', () => {
```
lsilvs: I thought it should clear the list if there is just one cached round.

yatki:
After further discussion with @lsilvs, to reduce complexity of the cache structure we decided to clear the whole cache instead of clearing it partially. This change will cause max 2 extra queries but it will also increase the maintainability.

test/unit/modules/delegates.js
yatki: See: https://github.com/LiskHQ/lisk/pull/2612#issuecomment-444804969

Issue #2614, opened 2018-12-06
Create Account entity with CRUD operations and unit tests 


nazarhussain:
Create an `Account` entity exposing following operations. 

- `get`
- `getOne`
- `create`
- `update`
- `updateOne`
- `isPersisted`

MaciejBaj:
https://github.com/LiskHQ/lisk/pull/2636

PR #2636, opened 2018-12-07
Create basic structure for entities and storage initialization - Closes #2613, #2614

### LINKED ISSUE #2613: Create basic structure for entities and storage initialization
Need to create `storage` component initial structure, with bootstrapping logic. It should mimic the current behavior of `db/index.js` to make sure nothing breaks afterwards and it can be replaced easily.

nazarhussain:
### What was the problem?
No entity model namespace and structure was present. 

### How did I fix it?
Introduced the storage initialization process with basic entity structure. 

### How to test it?
Run all unit tests for storage `npx mocha test/unit/stroage/*`

### Review checklist

* The PR resolves #2613, #2614 
* Linting passes
* Tests pass


storage/storage.js
```diff
@@ -0,0 +1,81 @@
...
+		Storage.instance = this;
+
+		Storage.instance.BaseEntity = BaseEntity;
+		Storage.instance.BaseEntity.adapter = null;
+	}
+
+	/**
+	 * @return Promise
+	 */
+	bootstrap() {
```
pablitovicente: I think we might need to wrap this with a Promise

storage/adapters/pgp_adapter.js
```diff
@@ -0,0 +1,198 @@
...
+
+		this.pgp = pgpLib(this.pgpOptions);
+		this.db = null;
+	}
+
+	/**
+	 * @return {Promise}
+	 */
+	connect() {
+		const self = this;
```
pablitovicente: This might need to be wrapped with a promise

storage/README.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Filters
+
+Filters are convention to provide a flexible but powerful structure to fetch data from persistence layer. When you register a field in entity you have to specify its filter type. Based on that filter type entity register few filter names available for you to be used as json objects. Each filter suffix an additional value to field name e.g. `_in`, `_lt`. It depends on filter type which suffixes are appended. A filter without any suffix to field name refers to `_eql` equals.
+
+Following is the list of available suffixes based on filter types
+
+| Filter Type | Filter Suffixes | Description |
+| ----------- | --------------- | ----------- |
+| BOOLEAN     | \_eql           |             |
```
diego-G: Descriptions here needed.

storage/entities/base_entity.js
```diff
@@ -0,0 +1,262 @@
...
+	 * Get one object from persistence layer
+	 *
+	 * @param {string | Object} filters - Multiple filters or just primary key
+	 * @param {Object} options - Extended options
+	 * @param {string} options.fieldSet - Fieldset to get
+	 * @param {Object} tx - transaction object
+	 *
+	 * @return {Promise}
+	 */
+	// eslint-disable-next-line class-methods-use-this,no-unused-vars
```
diego-G: I'm not so sure about disabling these rules in that many places. @nazarhussain could you explain both cases? 

`class-methods-use-this` -> Why not using `static` as written in the https://eslint.org/docs/rules/class-methods-use-this#enforce-that-class-methods-utilize-this-class-methods-use-this ?
`no-unused-vars` -> This rule belongs to the recommended Eslint list. I don't like to disable it at all.

test/unit/storage/adapters/pgp_adapter.js
```diff
@@ -0,0 +1,140 @@
...
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+'use strict';
+
+const rewire = require('rewire');
```
shuse2: prefer not to use rewire, and i dont think we need rewire.

storage/entities/base_entity.js
shuse2: @diego-G I think this is ok, as this class is more like a template.
In the extended version of this class, it would those eslint will be removed

lsilvs: @diego-G This class is acting as an interface to be implemented by entities.

`class-methods-use-this` -> the method is not a static method and its implementation will contain `this` (thee the Account entity)

`no-unused-vars` -> as there is no implementation here the vars are used as implementation references.

test/unit/storage/adapters/pgp_adapter.js
```diff
@@ -0,0 +1,140 @@
...
+};
+
+const validOptions = {
+	inTest: true,
+	sqlDirectory: '/sql/directory',
+	logger: loggerStub,
+};
+
+describe('PgpAdapter', () => {
+	afterEach(() => {
```
shuse2: maybe we can use the same pattern from `elements` we have set global_hook for `sandbox.restore`

storage/adapters/pgp_adapter.js
```diff
@@ -0,0 +1,198 @@
...
+	 *
+	 * @param {Object} options
+	 * @param {Boolean} options.inTest
+	 * @param {string} options.logger
+	 * @param {string} options.sqlDirectory
+	 */
+	constructor(options) {
+		super({
+			engineName: 'pgp',
+			inTest: options.inTest,
```
shuse2: what is the purpose of this `inTest`?

storage/entities/base_entity.js
diego-G: I've seen it in many other places not only in this file. It has to be a real exception plus perhaps a comment explaining why. Please, read the rules again and try to use it as less as possible.

test/unit/storage/adapters/pgp_adapter.js
```diff
@@ -0,0 +1,140 @@
...
+	});
+
+	it('should be a constructable function', () => {
+		expect(PgpAdapter.prototype).to.be.not.null;
+		return expect(PgpAdapter.prototype.constructor.name).to.be.eql(
+			'PgpAdapter'
+		);
+	});
+
+	it('should be be inherited by BaseAdapter', () => {
```
pablitovicente: I think the intention here was for the text to be `should extend BaseAdapter`

storage/adapters/pgp_adapter.js
shuse2: why do you need self here? Prefer using bind to hold context

storage/utils/inputSerializers.js
```diff
@@ -0,0 +1,24 @@
...
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+'use strict';
+
+/* eslint-disable no-unused-vars */
+
+module.exports = {
+	default: (value, mode, alias, fieldName) => `$\{${alias}}`,
```
shuse2: no where else is using `default` export, prefer to use `default` consistently

test/unit/storage/entities/account.js
```diff
@@ -0,0 +1,155 @@
...
+
+	describe('mergeFilters()', () => {
+		it('should accept filters as single object');
+		it('should accept filters as array of objects');
+		it(
+			'should merge provided filter with default filters by preserving default filters values '
+		);
+	});
+
+	it('should assign a prototype property FIELD_SET_SIMPLE');
```
shuse2: if it's outside of describe, it should be on the top too

test/unit/storage/utils/field.js
```diff
@@ -0,0 +1,107 @@
...
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+'use strict';
+
+const rewire = require('rewire');
+
+const Field = rewire('../../../../storage/utils/field');
```
shuse2: also, here prefer not to use `rewire`, I think it would be better to just use stub

storage/entities/base_entity.js
lsilvs: If it's not an interface I agree it should not be used. Can you point out the other ones you've noticed, pls?

storage/adapters/pgp_adapter.js
```diff
@@ -0,0 +1,198 @@
...
+			error: () => {
+				this.emit(this.EVENT_ERROR);
+			},
+			disconnect: () => {
+				this.emit(this.EVENT_DISCONNECT);
+			},
+		};
+
+		this.pgp = pgpLib(this.pgpOptions);
+		this.db = null;
```
shuse2: Consistent usage of `null` vs `undefined`.
After we move to TypeScript, we started to use this rule
https://palantir.github.io/tslint/rules/no-null-keyword/

storage/utils/inputSerializers.js
lsilvs: @shuse2 `default` here is just the name (label) of the serializer type.

shuse2: yea i know, but essentially if we use `babel` style, this is how they will produce it.
maybe avoid `default`?

lsilvs: Done

test/unit/storage/adapters/pgp_adapter.js
lsilvs: This would require refactoring all the old tests which is out of the scope of this PR imo.

storage/adapters/pgp_adapter.js
```diff
@@ -88,18 +86,18 @@ class PgpAdapter extends BaseAdapter {
 		};
 		monitor.attach(
 			Object.assign(this.pgpOptions, monitorOptions),
```
shuse2: Prefer Spread Opeerator.
```
{ ...this.pgpOptions, ...monitorOptions }
```

storage/adapters/pgp_adapter.js
lsilvs: It's only used for pg-promisse`noLocking` flag for now.
Not sure if it can be useful for other adapters though.

storage/adapters/pgp_adapter.js
```diff
@@ -0,0 +1,198 @@
...
+
+		let connectionObject = null;
+
+		return self.db
+			.connect()
+			.then(co => {
+				connectionObject = co;
+				return Promise.resolve(true);
+			})
+			.finally(() => {
```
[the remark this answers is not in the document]
pablitovicente: Well finally always run and rejections are happening. Ignore this comment for now.

storage/adapters/pgp_adapter.js
MaciejBaj: Maybe it's better to pass the separate Postgres option object, which would overwrite the default Postgres flags.

storage/adapters/pgp_adapter.js
```diff
@@ -0,0 +1,196 @@
...
+
+		this.pgp.end();
+		this.db = this.pgp(this.options);
+
+		// As of the nature of pg-promise the connection is acquired either a query is started to execute.
+		// So to actually verify the connection works fine
+		// based on the provided options, we need to test it by acquiring
+		// the connection a manually
+
+		// return Promise.resolve(true);
```
MaciejBaj: to remove?

storage/adapters/pgp_adapter.js
lsilvs: I also have some suggestions to make regarding the adapter that I would like to discuss but I reckon we should wait til @nazarhussain is back to discuss those points.

My suggestion is to merge without changing this and create an issue after the discussion if need be.

storage/adapters/pgp_adapter.js
```diff
@@ -0,0 +1,194 @@
...
+		// based on the provided options, we need to test it by acquiring
+		// the connection a manually
+
+		let connectionObject = null;
+
+		return this.db
+			.connect()
+			.then(co => {
+				connectionObject = co;
+				return Promise.resolve(true);
```
shuse2: within `then` I think u dont need to use `Promise.resolve`. Any particular reason?

storage/adapters/pgp_adapter.js
```diff
@@ -0,0 +1,194 @@
...
+			})
+			.finally(() => {
+				if (connectionObject) {
+					connectionObject.done();
+				}
+				return Promise.resolve(true);
```
shuse2: same as above

storage/adapters/pgp_adapter.js
```diff
@@ -0,0 +1,194 @@
...
+
+		const options = {
+			minify: true, // Minifies the SQL
+		};
+
+		const qf = new QueryFile(fullPath, options);
+
+		if (qf.error) {
+			console.error(qf.error); // Something is wrong with our query file
+			process.exit(1); // Exit the process with fatal error
```
shuse2: I think deciding on `process.exit` at this low level is too dangerous.
Prefer just to throw an error

storage/storage.js
```diff
@@ -0,0 +1,81 @@
...
+		const adapter = new PgpAdapter(
+			Object.assign({}, this.options, {
+				inTest: process.env.NODE_ENV === 'test',
+				sqlDirectory: path.join(path.dirname(__filename), './sql'),
+				logger: this.logger,
+			})
+		);
+
+		return adapter.connect().then(status => {
+			if (status) {
```
shuse2: prefer early return

test/unit/storage/utils/filters.js
```diff
@@ -0,0 +1,134 @@
...
+'use strict';
+
+const filterTypes = require('../../../../storage/utils/filter_types');
+const inputSerializers = require('../../../../storage/utils/inputSerializers');
+const filters = require('../../../../storage/utils/filters');
+
+const customSerializer = sinonSandbox.stub().returns('custom serialized value');
+
+describe('filters', () => {
+	beforeEach(done => {
```
shuse2: No need for done here, instead u can do
```
beforeEach(async () => {
  xxx
});
```

storage/adapters/base_adapter.js
```diff
@@ -0,0 +1,83 @@
...
+ */
+
+'use strict';
+
+/* eslint-disable class-methods-use-this */
+
+const { EventEmitter } = require('events');
+const { ImplementationPendingError } = require('../errors');
+
+const EVENT_CONNECT = Symbol('ADAPTER_EVENT_CONNECT');
```
shuse2: why `Symbol` here? simple string might be more useful?

storage/storage.js
lsilvs: What do you mean? It's always returning status here.

shuse2: ah nvm, my bad ='(

storage/schema/account.js
```diff
@@ -0,0 +1,28 @@
+/*
```
pablitovicente: I can not find usages of this file maybe it's better to remove it now and re-introduce it when planed functionality is implemented? (Although we might want to merge this initial version in the feature branch and later on iterate over this details) thoughts?

storage/adapters/base_adapter.js
lsilvs: Can you give an example where string would be more useful?

shuse2: maybe intended, but user need to use this const to listen to event, where in string, u can just define the same string to listen to the event.
I feel like it doesn't need to be that strict

test/unit/storage/utils/filters.js
lsilvs: I noticed that Elements is using `return Promise.resolve();`
We should be consistent on where to use `done()`, `return` and `async`
Let's discuss this and make the changes in another issue.

storage/adapters/pgp_adapter.js
lsilvs: I tested returning just `true` and it works.

I'll take note and discuss with @nazarhussain when he is back

storage/adapters/pgp_adapter.js
lsilvs: same as above

storage/adapters/base_adapter.js
lsilvs: Even if it's a string, the best practice is to use the const instead of just defining the same string.
I personally like enforce using the same const.

storage/schema/account.js
lsilvs: As this PR is not the final PR for the feature, let's leave the file here as reference.

storage/README.md
```diff
@@ -0,0 +1,70 @@
...
+| Filter Type | Filter Suffixes | Description                                                |
+| ----------- | --------------- | ---------------------------------------------------------- |
+| BOOLEAN     | \_eql           | returns entries that matches the value                     |
+|             | \_ne            | returns entries that does not matche the value             |
+| TEXT        | \_eql           | returns entries that matches the value                     |
+|             | \_ne            | returns entries that does not matche the value             |
+|             | \_in            | returns entries that matches any values in a list          |
+|             | \_like          | returns entries that matches a pattern                     |
+| NUMBER      | \_eq            | returns entries that matches the value                     |
+|             | \_ne            | returns entries that does not matche the value             |
```
MaciejBaj:
```suggestion
|             | \_ne            | returns entries that do not match the value             |
```

storage/README.md
```diff
@@ -0,0 +1,70 @@
...
+|             | \_gt            | returns entries that is greater than the value             |
```
MaciejBaj:
```suggestion
|             | \_gt            | returns entries greater than the value             |
```

storage/README.md
```diff
@@ -0,0 +1,70 @@
...
+|             | \_gte           | returns entries that is greater then or equal to the value |
```
MaciejBaj:
```suggestion
|             | \_gte           | returns entries greater than or equal to the value |
```

storage/README.md
```diff
@@ -0,0 +1,70 @@
...
+|             | \_lt            | returns entries that is less than the value                |
```
MaciejBaj:
```suggestion
|             | \_lt            | returns entries less than the value                |
```

storage/README.md
```diff
@@ -0,0 +1,70 @@
...
+|             | \_lte           | returns entries that is less than or equal to the value    |
```
MaciejBaj:
```suggestion
|             | \_lte           | returns entries less than or equal to the value    |
```

storage/README.md
```diff
@@ -0,0 +1,70 @@
...
+|             | \_in            | returns entries that matches any values in a list          |
```
MaciejBaj:
```suggestion
|             | \_in            | returns entries that match any of values from the list         |
```

storage/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+You can call `<Entity>.getFilters()` to see a list of available filters for any entity.
+
+### Filter Combinator
+
+If filters are provided as JSON object those will always joined with `AND` combinator. e.g. Specifying filters as `{name: 'Alpha', description_like: 'Bravo'}` will result in fetching all results which have name equal to `Alpha` and `Bravo`. Specifying filters as an array of objects: `[{name: 'Alpha'}, {description_like: 'Bravo'}]` will result in joining objects with `OR` combinator and result in fetching data which name equal either `Alpha` or `Bravo`.
```
MaciejBaj:
```suggestion
If filters are provided as JSON objects, they will always be joined with an `AND` combinator. For instance, specifying filters as `{name: 'Alpha', description_like: 'Bravo'}` results in fetching all results which have a name equal to `Alpha` and description matching `Bravo`. Specifying filters as an array of objects, e.g. `[{name: 'Alpha'}, {description_like: 'Bravo'}]`, will result in joining objects with `OR` combinator, i.e. fetching data which name equal to `Alpha` or description like `Bravo`.
 You can register a `CUSTOM` filter, by defining your own key and a function which will return a custom condition.
```

storage/README.md
```diff
@@ -0,0 +1,70 @@
...
+
+## Conventions
+
+Following conventions must be followed strictly
+
+* If getter functions are dynamic they must accept a `fitlers` object as first parameter.
```
MaciejBaj:
```suggestion
* Dynamic getter functions must accept `filters` object as the first parameter.
```

storage/README.md
```diff
@@ -0,0 +1,70 @@
...
+* Every entity should set some field sets to describe collection of fields.
```
MaciejBaj:
```suggestion
* Each entity should set fieldsets to describe the collection of fields.
```

storage/README.md
```diff
@@ -0,0 +1,70 @@
...
+* Entities should always return full set of attributes related to a field set and never implement field selection logic.
```
MaciejBaj:
```suggestion
* Entities should always return the full set of attributes related to a fieldset and don't implement field selection logic.
```

storage/README.md
```diff
@@ -0,0 +1,70 @@
...
+* Every interface in the entity must return `Promise` and declared as `async`.
```
MaciejBaj:
```suggestion
* Every interface in the entity must return `Promise` and be declared as `async`.
```

storage/README.md
```diff
@@ -0,0 +1,70 @@
...
+* Timestamp always refer to Epoch timestamp
```
MaciejBaj:
```suggestion
* Timestamp types always refer to the network Epoch timestamp.
```

storage/README.md
```diff
@@ -0,0 +1,70 @@
...
+* In case of Unix the attribute name must be prefixed with Unix
```
MaciejBaj:
```suggestion
```

storage/README.md
```diff
@@ -0,0 +1,70 @@
...
+* All values related to amount should be exposed as strings
+* Height should always be integer because of Number.MAX_SAFE_INTEGER is too high
+* Round number should be exposed as integer
+* All `BYTEA` fields with 'hex' encoding should be exposed as strings
```
MaciejBaj:
```suggestion
* All `BYTEA` fields in 'hex' encoding should be exposed as strings.
```

storage/README.md
```diff
@@ -0,0 +1,70 @@
...
+* All other `BYTEA` fields should be exposed as Buffer
```
MaciejBaj:
```suggestion
```

storage/adapters/pgp_adapter.js
```diff
@@ -0,0 +1,194 @@
...
+
+'use strict';
+
+const path = require('path');
+const Promise = require('bluebird');
+const monitor = require('pg-monitor');
+const pgpLib = require('pg-promise');
+const QueryFile = require('pg-promise').QueryFile;
+const BaseAdapter = require('./base_adapter');
+
```
MaciejBaj:
```suggestion

// ToDo: Shall be converted to `Enum` type after TS migration.
```

storage/README.md
```diff
@@ -0,0 +1,70 @@
...
+|             | \_like          | returns entries that matches a pattern                     |
+| NUMBER      | \_eq            | returns entries that matches the value                     |
+|             | \_ne            | returns entries that does not matche the value             |
+|             | \_gt            | returns entries that is greater than the value             |
+|             | \_gte           | returns entries that is greater then or equal to the value |
+|             | \_lt            | returns entries that is less than the value                |
+|             | \_lte           | returns entries that is less than or equal to the value    |
+|             | \_in            | returns entries that matches any values in a list          |
+
+You can call `<Entity>.getFilters()` to see a list of available filters for any entity.
```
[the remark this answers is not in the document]
lsilvs: What is the change here?

PR #2637, opened 2018-12-07
Endpoint /api/delegates/forgers fail with offset > 100 - Closes #2651

### LINKED ISSUE #2651: Endpoint /api/delegates/forgers fail with offset > 100
### Expected behavior
This case should be handled properly, for example by not allowing passing `offset` greater than `100`.
### Actual behavior
When endpoint `/api/delegates/forgers` is called with `offset` greater than `100` request fail with the message `An unexpected error occurred while handling this request` and internally with:
```
[log] 2018-12-10 17:22:20 | GET /api/delegates/forgers?offset=101 from 127.0.0.1
[dbg] 2018-12-10 17:22:20 | Using delegate list from the cache.
[ERR] 2018-12-10 17:22:20 | error: syntax error at or near ")"
    at Connection.parseE (/lisk/node_modules/pg/lib/connection.js:545:11)
    at Connection.parseMessage (/lisk/node_modules/pg/lib/connection.js:370:19)
    at Socket.<anonymous> (/lisk/node_modules/pg/lib/connection.js:113:22)
    at emitOne (events.js:116:13)
    at Socket.emit (events.js:211:7)
    at Socket.wrapped [as emit] (/lisk/node_modules/newrelic/lib/transaction/tracer/index.js:181:22)
    at addChunk (_stream_readable.js:263:12)
    at readableAddChunk (_stream_readable.js:250:11)
    at Socket.Readable.push (_stream_readable.js:208:10)
    at TCP.onread (net.js:601:20)
    at TCP.wrapped [as onread] (/lisk/node_modules/newrelic/lib/transaction/tracer/index.js:181:22)
18:22:20 error: syntax error at or near ")"
         query: SELECT encode("publicKey", 'hex') AS "publicKey", username, address FROM mem_accounts WHERE "isDelegate" = 1 AND encode("publicKey", 'hex') IN ()
```
### Steps to reproduce
Query endpoint `/api/delegates/forgers` with `offset=101`.
`1.x.x`

Bit-Shifts:
-Added offset parameter with min and max values for getForgers endpoint in the Delegates group

### What was the problem?

The problem is you can submit and numerical value for offset parameter for getForger endpoint and if the value is above 100 that means there will be no forgers to returned based on the logic and the fact there is only currently 101 forging delegates. This causes a internal server error because the syntax of the query  being sent to postgres is not correct as the array of delegate public keys is empty.

### How did I fix it?

I fixed this by adding validation to the swagger.yml on the offset parameter for this endpoint to not accept values greater then 100.
 
### How to test it?

I tested it by making the change and running a node with my changes and called to endpoint in question and seen the invalid parameter handled correctly.

### Review checklist

* The PR resolves #2651
* Linting passes
* Tests pass


Bit-Shifts:
@4miners I added a test for this case.

PR #2643, opened 2018-12-10
Apply eslint-rule no-restricted-syntax - Related to #1597

diego-G:
### Review checklist

* The PR related to #1597
* Linting passes
* Tests pass

### DIFF
--- a/logger.js
+++ b/logger.js
@@ -59,11 +59,11 @@ module.exports = function(config) {
 	};
 
 	function snipFragileData(data) {
-		for (const key in data) {
+		Object.keys(data).forEach(key => {
 			if (key.search(/passphrase|password/i) > -1) {
 				data[key] = 'XXXXXXXXXX';
 			}
-		}
+		});
 		return data;
 	}
 
--- a/logic/dapp.js
+++ b/logic/dapp.js
@@ -101,12 +101,11 @@ DApp.prototype.verify = function(transaction, sender, cb, tx) {
 	}
 
 	let foundCategory = false;
-	for (const i of Object.keys(dappCategories)) {
-		if (dappCategories[i] === transaction.asset.dapp.category) {
+	Object.keys(dappCategories).forEach(key => {
+		if (dappCategories[key] === transaction.asset.dapp.category) {
 			foundCategory = true;
-			break;
 		}
-	}
+	});
 
 	if (!foundCategory) {
 		return setImmediate(cb, 'Application category not found');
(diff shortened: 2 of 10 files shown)


diego-G:
@ManuGowda https://github.com/airbnb/javascript/blob/17e0454672f784772c5f5ab052523a93229d9adf/packages/eslint-config-airbnb-base/rules/style.js#L334

4miners:
I agree with @ManuGowda, we should allow `for...in` and `for...of` loops.

ManuGowda:
@diego-G I understand the violation of miss using `for..in` and `for..of` in a bad way, if we really want to restrict it I think we should give a better guideline for all the team members and keep everyone in the same loop, because in my opinion, enumerable iteration and collection iteration is something we use a lot in general, so my intention here is to get everyone on the same page, that's all.

logger.js
```diff
@@ -59,11 +59,11 @@ module.exports = function(config) {
 	};
 
 	function snipFragileData(data) {
-		for (const key in data) {
+		Object.keys(data).forEach(key => {
```
ManuGowda: why not `return Object.keys(data).map(key => {`?

diego-G:
@4miners and @ManuGowda the whole point is to avoid style discussions like this one. AirBnB states pretty clear to disallow `for in` and `for of` in favour of creating arrays and iterating over arrays with forEach or map. 

https://github.com/airbnb/javascript#iterators--nope

```
Use map() / every() / filter() / find() / findIndex() / reduce() / some() / ... to iterate over arrays, and Object.keys() / Object.values() / Object.entries() to produce arrays so you can iterate over objects.
```

Besides, the usage of `for in` and `for of` in our code is low as you can realise in the size of this PR.

logger.js
```diff
@@ -59,12 +59,12 @@ module.exports = function(config) {
...
-		for (const key in data) {
+		return Object.keys(data).map(key => {
 			if (key.search(/passphrase|password/i) > -1) {
 				data[key] = 'XXXXXXXXXX';
 			}
-		}
-		return data;
+			return data[key];
+		});
 	}
```
lsilvs: This new function is not equivalent to the old one. It returns an array rather than an object.
In order to make it equivalent you have to either return the mutated object or reconstruct it using `reduce`.
As the motivation to not use iterators is to have pure functions, I would suggest using `reduce`

```
function snipFragileData(data) {
	return Object.keys(data).reduce((acc, key) => {
		acc[key] = (key.search(/passphrase|password/i) > -1) ? 'XXXXXXXXXX' : data[key];
		return acc;
	}, {});
}
```

logic/dapp.js
```diff
@@ -101,12 +101,11 @@ DApp.prototype.verify = function(transaction, sender, cb, tx) {
...
 	let foundCategory = false;
-	for (const i of Object.keys(dappCategories)) {
-		if (dappCategories[i] === transaction.asset.dapp.category) {
+	Object.keys(dappCategories).forEach(key => {
+		if (dappCategories[key] === transaction.asset.dapp.category) {
 			foundCategory = true;
-			break;
 		}
-	}
+	});
```
lsilvs: Prefer using `find()` as it returns the first element in an array that pass a test and breaks the loop.

```
const foundCategory = !!Object.values(dappCategories).find(value => value === transaction.asset.dapp.category);
```

diego-G: We need to iterate over every category to check if all of them are valid.

logger.js
diego-G: Because it will return an array instead of an object.

logic/dapp.js
lsilvs: The original code does `break` when a category is found

lsilvs: The original code does `break` when a category is found

PR #2648, opened 2018-12-10
Apply eslint-rule guard-for-in - Related to #1597

diego-G:
### Review checklist

* The PR is related to #1597
* Linting passes
* Tests pass


modules/blocks/verify.js
```diff
@@ -319,27 +319,29 @@ __private.verifyPayload = function(block, result) {
...
 	for (const i in block.transactions) {
-		const transaction = block.transactions[i];
-		let bytes;
-
-		try {
-			bytes = library.logic.transaction.getBytes(transaction);
-		} catch (e) {
-			result.errors.push(e.toString());
-		}
+		if (Object.prototype.hasOwnProperty.call(block.transactions, i)) {
```
MaciejBaj: `block.transactions` is an array, I don't think this check is necessary here and introduces confusion in the critical function - `blocks.verify`.

MaciejBaj: Overall this rule looks odd to me. We should either introduce a rule of not-using `for ... in` at all, or keep this flag off.

diego-G: This code will change when introducing `no-restricted-syntax` https://github.com/LiskHQ/lisk/pull/2643 . Perhaps it'll make more sense after it.

MaciejBaj:
Please solve the conflicts:
```
modules/blocks/verify.js
modules/loader.js
modules/node.js
```

diego-G:
Rule doesn't trigger any error. It will be removed on the PR #2649 instead.

PR #2659, opened 2018-12-12
Clean network test suite - Closes #2438

### LINKED ISSUE #2438: networkHeight takes too long to update in some cases
### Expected behavior

The `networkHeight` needs to update faster and be consistent across all nodes.

### Actual behavior

Recent improvements to the network tests have uncovered that the networkHeight is not updating fast enough in some cases. For some nodes, `networkHeight` was found to be 2 blocks behind even though all nodes are on thencorrect `height` and they are all interconnected.

### Steps to reproduce

Run latest network tests `grunt mocha:default:network` after https://github.com/LiskHQ/lisk/pull/2431 has been merged.

v1.2.0

diego-G:
### What was the problem?

Firstly, the structure and syntax were too complex to easily debug. All test runs over the same network in a sequential way. The test that is always executed in first place, `p2p/peer_disconnect`, was introducing a case too complex from which the network couldn't recover properly producing failures in the rest of tests. This reasoning explains why we must limit the scenarios complexity that we introduced in this test suite. 

Besides, the `scClient` configuration which was being used in our network test suite was not deterministic and slow. The client was taken too long to reconnect to a manually restarted peer. Also It was including a randomness parameter necessary in production environments but not in test where we need deterministic behaviours.

### How did I fix it?

- Removing the out-of-scope scenario https://github.com/LiskHQ/lisk/commit/f9a28b59f284800c1c96e17fd0546855af45347b
- Clean out weird syntax
- Clarify tests structure
- Reduce `scClient` parameters to reconnect and acknowledge https://github.com/LiskHQ/lisk/pull/2659/commits/511a1e673eec5b5942b26d5015b586902c29e93f 
- Remove `scClient` random amount of time when reconnecting to a peer https://github.com/LiskHQ/lisk/pull/2659/commits/432731c575616e84462af7afdd9524efd74db5d7

### How to test it?
`NODE_ENV=test npm test -- mocha:default:network`

### Review checklist

* The PR resolves #2438
* Linting passes
* Tests pass


test/network/setup/config.js
```diff
@@ -36,7 +36,7 @@ const SYNC_MODES = {
 
 const SYNC_MODE_DEFAULT_ARGS = {
 	RANDOM: {
-		probability: 0.5, // (0 - 1)
+		probability: 0, // (0 - 1)
```
MaciejBaj: It's not RANDOM then :)

diego-G: Actually, I should remove that commit. It doesn't change anything. I misunderstood the parameter.

PR #2661, opened 2018-12-12
Create Block entity with CRUD operations and unit tests - Closes #2615

### LINKED ISSUE #2615: Create Block entity with CRUD operations and unit tests
Create an `Block` entity exposing following operations. 

- `get`
- `getOne`
- `create`
- `update`
- `updateOne`
- `isPersisted`

lsilvs:
### What was the problem?
Block entity was not implemented

### How did I fix it?
I've implemented the Block entity based on the Account entity.
I've also done some refactoring as suggestion to the initial implementation.
The tests skeleton was copied from Account and some were implemented to promote initial discussions.

### How to test it?
`npm run mocha -- test/unit/storage/`

### Review checklist

* The PR resolves #2615
* Linting passes
* Tests pass


test/unit/storage/entities/block.js
```diff
@@ -0,0 +1,669 @@
...
+			]);
+
+			done();
+		});
+
+		it('should setup specific filters', done => {
+			const adapter = { loadSQLFile: sinonSandbox.stub() };
+			const block = new Block(adapter);
+
+			expect(block.filters).to.have.all.keys([
```
pablitovicente: What if we move this to a mocks/ directory? In pears I'm doing the same test and the test would be cleaner if required()

test/unit/storage/entities/block.js
```diff
@@ -0,0 +1,669 @@
...
+			expect(Block.prototype.updateOne).to.throw(NonSupportedOperationError);
+			done();
+		});
+	});
+
+	describe('isPersisted()', () => {
+		let block;
+
+		before(done => {
+			const adapter = {
```
pablitovicente: What if we have only one adapter stub with all it's stubbed properties and gets reused for all tests and restored on afterEach() ?

test/unit/storage/entities/block.js
```diff
@@ -0,0 +1,669 @@
...
+			expect(block.mergeFilters.calledWith(filters)).to.be.ok;
+			done();
+		});
+
+		it('should call parseFilters with proper params', done => {
+			const filters = [{ height: 101 }, { timestamp_gte: 1234567890 }];
+			block.mergeFilters = sinonSandbox.stub().returns(filters);
+			block.parseFilters = sinonSandbox.stub();
+			block.isPersisted(filters);
+			expect(block.parseFilters.calledWith(filters)).to.be.ok;
```
pablitovicente: wouldn't  `.to.be.true` read better to `.to.be.ok` ?

storage/entities/block.js
```diff
@@ -14,8 +14,343 @@
 
 'use strict';
 
+const _ = require('lodash');
+const { stringToByte } = require('../utils/inputSerializers');
+const { NonSupportedOperationError } = require('../errors');
+const ft = require('../utils/filter_types');
```
pablitovicente: I would use as name `filterType` instead of `ft`

test/unit/storage/entities/block.js
```diff
@@ -0,0 +1,669 @@
...
+} = require('../../../../storage/errors');
+
+describe('Block', () => {
+	it('should be a constructable function', done => {
+		expect(Block.prototype.constructor).to.be.not.null;
+		expect(Block.prototype.constructor.name).to.be.eql('Block');
+		done();
+	});
+
+	it('should be be inherited by BaseEntity', done => {
```
pablitovicente: Should be `should extend BaseEntity`

test/unit/storage/entities/block.js
lsilvs: Some tests use different `adapter.executeFile` stub because it requires different `resolves()` scenarios

storage/entities/block.js
```diff
@@ -14,8 +14,343 @@
...
 const BaseEntity = require('./base_entity');
 
-class Block extends BaseEntity {}
+/**
+ * Basic Block
```
MaciejBaj:
```suggestion
 * Block fields
```

storage/entities/block.js
```diff
@@ -14,8 +14,343 @@
...
+ * @property {string} [version_in]
+ */
+
+/**
+ * @typedef {string} fieldSets.Block
+ * @enum
+ * @value 'FIELD_SET_SIMPLE'
+ */
+
+const FIELD_SET_SIMPLE = Symbol('FIELD_SET_SIMPLE');
```
MaciejBaj: Let's write the new "Convention" about using Symbols in DB Entities, post it into README.md and post the link to it here as a comment.

storage/entities/block.js
```diff
@@ -14,8 +14,343 @@
...
+			fieldName: 'previousBlock',
+		});
+		this.addField('timestamp', 'number', { filter: ft.NUMBER });
+		this.addField('totalAmount', 'string', { filter: ft.NUMBER });
+		this.addField('totalFee', 'string', { filter: ft.NUMBER });
+		this.addField('reward', 'string', { filter: ft.NUMBER });
+		this.addField('version', 'number', { filter: ft.NUMBER });
+
+		this.SQLs = {
+			selectSimple: this.adapter.loadSQLFile('blocks/get_simple.sql'),
```
MaciejBaj: Rename from `get_simple` to `get`.

storage/entities/block.js
```diff
@@ -14,8 +14,343 @@
...
+			create: this.adapter.loadSQLFile('blocks/create.sql'),
+			isPersisted: this.adapter.loadSQLFile('blocks/is_persisted.sql'),
+		};
+	}
+
+	// eslint-disable-next-line class-methods-use-this
+	get FIELD_SET_SIMPLE() {
```
MaciejBaj: We will discuss this.

storage/entities/block.js
```diff
@@ -14,8 +14,343 @@
...
+	 *
+	 * @param {filters.Block|filters.Block[]} [filters = {}]
+	 * @param {Object} [options = {}] - Options to filter data
+	 * @param {Number} [options.limit=10] - Number of records to fetch
+	 * @param {Number} [options.offset=0] - Offset to start the records
+	 * @param {fieldSets.Block} [options.fieldSet='FIELD_SET_SIMPLE'] - Fieldset to choose
+	 * @param {Object} tx - Database transaction object
+	 * @return {Promise.<BasicBlock[], NonSupportedFilterTypeError|NonSupportedOptionError>}
+	 */
+	get(filters = {}, options = {}, tx) {
```
MaciejBaj: Consider extracting the common part with `getOne` into a separate function.

storage/entities/base_entity.js
```diff
@@ -226,6 +236,74 @@ class BaseEntity {
...
+	 * @return {true, NonSupportedFilterTypeError>}
+	 */
+	validateOptions(options = {}, required = false) {
+		if (required && (!options || !Object.keys(options).length)) {
+			throw new NonSupportedOptionError(
+				'Options are required for this operation.'
+			);
+		}
+
+		let invalidOptions = [];
```
MaciejBaj: I am sure you will come up with sth more elegant :)

test/unit/storage/entities/block.js
lsilvs: Agreed.
@yatki is working on a proposal for how to structure the tests.
Can you give your thoughts here, @yatki ?

storage/entities/block.js
lsilvs: Will be discussed and added in another PR

storage/entities/block.js
lsilvs: Will be discussed and added in another PR

storage/entities/base_entity.js
```diff
@@ -226,6 +236,74 @@ class BaseEntity {
 		);
 	}
 
+	/**
+	 * Validate allowed filters
+	 * @param {Array.<Object>|Object} filters
+	 * @param {Boolean} required
+	 * @return {true, NonSupportedFilterTypeError>}
+	 */
+	validateFilters(filters = {}, required = false) {
```
shuse2: isn't checking `required`, responsibility of caller?
When you cal function, u are passing filter and saying this is required, but caller is supplying so it sounds redundant.
also, checking `!filters` doesn't make sense because it has default value.
Same for validateOptions.

lsilvs: I've added this `required` flag because in `isPersisted` we want to make sure you're passing at least one filter otherwise the method doesn't make sense. Maybe the name of the param is not clear. I'll update it.

The `!filters` was for the case where `filters` is empty string but I've just realised `Object.keys(filters)` will cater for this scenario as well. I'll remove the `!filter` check ;)

lsilvs: Actually the `!filters` check if required.
[This test](https://github.com/LiskHQ/lisk/blob/2615-create-block-entity/test/unit/storage/entities/block.js#L550) doesn't pass if I remove it.

```
AssertionError: expected [Function] to throw 'NonSupportedFilterTypeError' but 'TypeError: Cannot convert undefined or null to object' was thrown
```

PR #2662, opened 2018-12-13
Create npm-shrinkwrap.json, use 'npm ci' in build - Closes #2656

### LINKED ISSUE #2656: Use npm ci for builds
### Expected behavior
Use `npm ci` for builds, see [documentation](https://docs.npmjs.com/cli/ci) and [announcement](https://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable).

### Actual behavior
`npm install` is being used in `build/build.sh`.

### Steps to reproduce
Run build (see `build/README.md`)

`>1.4.0`

ghost:
### What was the problem?
`npm install` was being used for builds meaning dependencies of dependencies could change between builds.

### How did I fix it?
Use `npm ci` instead.

### How to test it?
Run build; ensure node_modules tree is reproducible (and matches locked packages)

### Review checklist

* The PR resolves #2656 
* Linting passes
* Tests pass


fchavant:
@MaciejBaj we're still using `npm install` for [pm2 and lisk-commander](https://github.com/LiskHQ/lisk/pull/2662/commits/bfbad26f7887012ced70df2677923920a1f76bf1#diff-fec83b7e693a8e104fa9713533e1c054R145). We already have pm2 in `package{,-lock}.json` but it's a devDependency. Can we move pm2 to `dependencies`? And add lisk-commander too?

MaciejBaj (dismissed):

MaciejBaj:
@fchavant how about https://docs.npmjs.com/files/package.json#bundleddependencies?

fchavant:
@MaciejBaj it's worth a try

diego-G (commented):
@fchavant @MaciejBaj shall we somehow mention this in the Readme ?

fchavant:
@diego-G we should, but we still need a solution for pm2 and commander

Nazgolze (dismissed):

diego-G (dismissed):
PM2 is now installed as dependency. Shall we then remove this instruction? https://github.com/LiskHQ/lisk/blob/d69980d74bd60b9c203321d89f28c7b4a5fa37b6/README.md#pm2-recommended

Nazgolze:
@diego-G good catch :+1:

Nazgolze (dismissed):

diego-G (dismissed):
@fchavant I've noticed `npm ci` doesn't check the existing engines from `package.json`. This allows to install with a Node version we don't support.
 
I'm not sure either to use it regularly for development where you don't need to build every time. I would leave `npm i` in the Readme otherwise we need to find a solution to respect the engines.

fchavant:
@diego-G 
The `engines` field "specif[ies] which versions of npm are capable of properly *installing* your program" ([source](https://docs.npmjs.com/files/package.json#engines); emphasis mine) which might not be the same as the version that can build the application. Still we use [enforce](https://github.com/LiskHQ/lisk/blob/development/build/Makefile#L2) the version used at build time. Anyone followint the [README](https://github.com/LiskHQ/lisk/blob/development/README.md#nodejs)  will be using the right version of Node.js also.

MaciejBaj (dismissed):

README.md
```diff
@@ -279,7 +279,7 @@ Clone the Lisk Core repository using Git and initialize the modules.
 git clone https://github.com/LiskHQ/lisk.git
 cd lisk
 git checkout master
-npm install
+npm ci
```
diego-G: Due to the explanation from the comment https://github.com/LiskHQ/lisk/pull/2662#issuecomment-448969174 , we should have `npm install` here and clarify that `npm ci` is meant to be used when we are building the app not when installing. We must mention the finding in a comment to prevent users to fall in the same issue:
- `npm ci` doesn't apply the restrictions `engines` set in the `package.json`.

fchavant: It seems we have to choose between having the right packages (from `package-json.lock`) or enforcing the `engines`; it is unfortunate we cannot have both.
I would argue it is easier to document which version of npm to use (and to follow that instruction) that to ensure the right packages are installed.

MaciejBaj (approved):

diego-G (approved):

README.md
MaciejBaj: `npm ci` in the manual might look a bit odd, but it serves the purpose of having the exact same code including the one in dependencies across all of the same Lisk Core versions running from Sources, Binaries and Docker. I found this [article](https://blog.npmjs.org/post/171556855892/introducing-npm-ci-for-faster-more-reliable) convincing.


There is a high probability that the `npm ci` would fail for other versions of Node.js due to the `engine` settings in package.json of other dependencies anyway (even if Lisk Elements or Lisk Commander).

PR #2663, opened 2018-12-13
Application no longer crashing on dns lookup failure - Closes #2608

### LINKED ISSUE #2608: Application crashes if any DNS name cannot be resolved
### Expected behavior
DNS names from differents domains/TLDs are used for seeds nodes in `config/*/config.json` to improve reliability. If one or a few of them cannot be resolved (DNS entry missing, resolver returning an error, TLD issue, etc.) the application still functions (especially if the `peers` table is populated).

### Actual behavior
Application crashes if even a single DNS name cannot be resolved. A message similar to the following is logged:
`2018-12-05 10:07:13 224: Failed to resolve peer domain name testnet-seed-05.will-not-resolve.io to an IP address`

### Steps to reproduce
- apply the following patch:
```
--- a/config.json       2018-12-05 10:20:24.029725152 +0000
+++ b/config.json       2018-12-05 10:20:13.049767544 +0000
@@ -23,7 +23,7 @@
                                "wsPort": 7001
                        },
                        {
-                               "ip": "testnet-seed-05.lisk.io",
+                               "ip": "testnet-seed-05.will-not-resolve.io",
                                "wsPort": 7001
                        }
                ]
```
- try to start the application with `LISK_NETWORK=testnet`

Core `>=1.2.0`

limiaspasdaniel:
### What was the problem?
The application was crashing on a dns lookup failure for the peers list.
### How did I fix it?
Removed the code that was throwing an error on a dns lookup failure that was responsible of crashing the application. Now if a dns lookup fails the error is logged but it no longer halts the execution.
### How to test it?
Manually: modify one or more seed peer addresses under config.json to a fake/nonworking one. The application should report the dns lookup failure but should continue working as expected.
### Review checklist

* The PR resolves #2608 
* Linting passes
* Tests pass

### DIFF
--- a/app.js
+++ b/app.js
@@ -236,7 +236,7 @@ d.run(() => {
 										peer.ip
 									} to an IP address`
 								);
-								return callback(err, peer);
+								return callback(null, peer);
 							}
 							return callback(null, Object.assign({}, peer, { ip: address }));
 						});
@@ -244,9 +244,7 @@ d.run(() => {
 				);
 
 				return async.parallel(peerDomainLookupTasks, (_, results) => {
-					appConfig.peers.list = results
-						.filter(result => result.hasOwnProperty('value'))
-						.map(result => result.value);
+					appConfig.peers.list = results.map(result => result.value);
 					return cb(null, appConfig);
 				});
 			},
--- a/app.js
+++ b/app.js
@@ -224,8 +224,8 @@ d.run(() => {
 				}
 
 				// In case domain names are used, resolve those to IP addresses.
-				const peerDomainLookupTasks = appConfig.peers.list.map(
-					peer => callback => {
+				const peerDomainLookupTasks = appConfig.peers.list.map(peer =>
+					async.reflect(callback => {
 						if (net.isIPv4(peer.ip)) {
 							return setImmediate(() => callback(null, peer));
 						}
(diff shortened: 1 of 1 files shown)


app.js
```diff
@@ -240,14 +240,13 @@ d.run(() => {
 							}
```
lsilvs: Should we not remove the peers which dns lookup fail?

`return callback(err);` should do that.

lsilvs: Github didn't allow me to select the correct line but the line I'm referring to is app.js:239

limiaspasdaniel: [async.reflect](https://caolan.github.io/async/reflect.js.html) collects all the task execution results into one single object `results` passed through a the callback on line 246 with either an error attribute or a value property. So the `results` object contains all the task execution results and then it's filter and mapped so the peers list only contains resolved dns' IPs

lsilvs: Nice! I thought it would have both `error` and `value` in this case. All good then.

app.js
```diff
@@ -240,14 +240,13 @@ d.run(() => {
...
+					})
 				);
 
-				return async.parallel(peerDomainLookupTasks, (err, results) => {
-					if (err) {
-						return cb(err, appConfig);
-					}
-					appConfig.peers.list = results;
+				return async.parallel(peerDomainLookupTasks, (_, results) => {
+					appConfig.peers.list = results
```
MaciejBaj: We definitely don't want to filter out the array of seed peers!

limiaspasdaniel: Cooopy

limiaspasdaniel:
@MaciejBaj commit reverted

MaciejBaj:
@limiaspasdaniel I am afraid Linter doesn't like your commit anymore since the rule https://eslint.org/docs/rules/no-prototype-builtins was applied after merging #2646.

PR #2664, opened 2018-12-13
Invalid error message on second signature registration - Closes #2606

### LINKED ISSUE #2606: Invalid error message while registering second signature for the second time
### Expected behavior
When a user tries to enable second signature for an account with second signature enabled, the application should respons with meaningful error message like `The account is already enabled with second signature` or `Second signautre already exists` etc.
### Actual behavior
I have an account and I have registered the second signature on the account when I try to add the second signature the already enabled second signature account I get an error message `Missing sender second signature` which is confusing for the end user.
### Steps to reproduce
1) Create an account
2) Enable second signature on the account - should be successful
3) Enable second signature on the account again - results in `Missing sender second signature`

limiaspasdaniel:
### What was the problem?
I have an account and I have registered the second signature on the account when I try to add the second signature the already enabled second signature account I get an error message

>  Missing sender second signature

 which is confusing for the end user.

### How did I fix it?
Added new logic to check whether the user is attempting to register a new second signature but its account already has one enabled.

Modified matching test cases to comply with new logic.

Added new unit test for `verify` function under `logic/transaction.js`

### How to test it?
(should return error when transaction is type 1 and sender already has a second passphrase)
`npm run mocha -- test/unit/logic/transaction.js`
`npm run mocha --test/integration/transactions/1.second_signature`

### Review checklist

* The PR resolves #2606 
* Linting passes
* Tests pass

### DIFF
--- a/logic/transaction.js
+++ b/logic/transaction.js
@@ -19,6 +19,7 @@ const ByteBuffer = require('bytebuffer');
 const _ = require('lodash');
 const Bignum = require('../helpers/bignum.js');
 const slots = require('../helpers/slots.js');
+const transactionTypes = require('../helpers/transaction_types.js');
 
 const exceptions = global.exceptions;
 const POSTGRESQL_BIGINT_MAX_VALUE = '9223372036854775807';
@@ -423,6 +424,12 @@ class Transaction {
 		// Reject if transaction has requester public key
 		if (transaction.requesterPublicKey) {
 			return setImmediate(cb, 'Multisig request is not allowed');
+		}
+
+		// Check if sender account has second signature enabled.
+		// Abort if registering again.
+		if (transaction.type === transactionTypes.SIGNATURE && sender.secondSignature) {
+			return setImmediate(cb, 'This account is already enabled with second signature');
 		}
 
 		// Check for missing sender second signature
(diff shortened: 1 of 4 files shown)


test/unit/logic/transaction.js
```diff
@@ -583,6 +583,24 @@ describe('transaction', () => {
 			});
 		});
 
+		it('should return error when transaction is type 1 and sender already has a second passphrase', done => {
```
SargeKhan:
```suggestion
		it('should return error when transaction is type 1 and sender already has second signature enabled', done => {
```

logic/transaction.js
```diff
@@ -425,6 +426,12 @@ class Transaction {
 			return setImmediate(cb, 'Multisig request is not allowed');
 		}
 
+		// Check if sender account has second signature enabled.
+		// Abort if registering again.
+		if (transaction.type === transactionTypes.SIGNATURE && sender.secondSignature) {
+			return setImmediate(cb, 'This account is already enabled with second signature');
```
SargeKhan: Maybe it's better to have a similar error message wording as the rest of errors.
```suggestion
			return setImmediate(cb, 'Sender already has second signature enabled');
```

limiaspasdaniel: Makes sense, updated :)

PR #2671, opened 2018-12-17
Change network test approach to 9 forging nodes plus 1 non forging - Close #2670

### LINKED ISSUE #2670: Restore stability of peers blacklists and disconnections in Network Tests
### Expected behavior

The small network of 10 nodes utilised in network test suite should be stable enough to not to affect the test being executed towards it. Instead of having 10 nodes producing blocks, the network will have 9 nodes producing blocks plus another not producing any. This node will become the number 9 instead of the node number 0.

### Actual behavior

There are two network tests, disconnect and blacklist peers, which are introducing collateral problems on the rest. This is because the chosen node is supposed to produce 10/11 blocks per round since we enable forging on all nodes at the beginning. Those two tests restart node 0 creating random issues if during the off-time the delegate supposed to forge belongs to node 0. 

### Steps to reproduce

Several executions `npm test -- mocha:default:network`

`1.0+`

diego-G:
### How to test it?

`npm test -- mocha:default:network`

### Review checklist

* The PR resolves #2670
* Linting passes
* Tests pass

### DIFF
--- a/test/common/ws/server_process.js
+++ b/test/common/ws/server_process.js
@@ -36,7 +36,6 @@ function WSServer(headers) {
 		workers: 1,
 		port: this.headers.wsPort,
 		wsEngine: 'ws',
-		appName: `LiskTestServer-${randomstring.generate(8)}`,
 		secretKey: 'liskSecretKey',
 		headers,
 		perMessageDeflate: false,
--- a/test/network/setup/config.js
+++ b/test/network/setup/config.js
@@ -73,7 +73,7 @@ const config = {
 
 		// Configuring nodes to forge with force or without
 		const delegatesMaxLength = Math.ceil(
-			devConfig.forging.delegates.length / configurations.length
+			devConfig.forging.delegates.length / (configurations.length - 1)
 		);
 		const delegates = _.clone(devConfig.forging.delegates);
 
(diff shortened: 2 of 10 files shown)


test/common/ws/server_process.js
```diff
@@ -40,9 +40,9 @@ function WSServer(headers) {
 		secretKey: 'liskSecretKey',
 		headers,
 		perMessageDeflate: false,
-		pingInterval: 5000,
-		pingTimeout: 60000,
-		processTermTimeout: 10000,
+		pingInterval: 500,
+		pingTimeout: 600,
```
jondubois: I don't think we want the socket to ever timeout during the tests so we should make the `pingTimeout` higher. I think `1100` would be a better value here; that means that it won't timeout/disconnect the socket unless it misses two full pings.

jondubois: Actually maybe a `10000` is better for `pingTimeout`; if the testsuite is running on a weak machine, it could lag so it's good to give extra margin of safety.

diego-G: I've reverted the three parameters as they were before.

test/network/setup/config.js
```diff
@@ -73,7 +73,7 @@ const config = {
 
 		// Configuring nodes to forge with force or without
 		const delegatesMaxLength = Math.ceil(
-			devConfig.forging.delegates.length / configurations.length
+			devConfig.forging.delegates.length / (configurations.length - 1)
```
MaciejBaj: That's wrong, just for the purpose of Peer Disconnect and Blacklists we shouldn't change the network setup for all of the remaining tests.

diego-G:
Close until stability is found again.

PR #2673, opened 2018-12-18
Refactor data persistence layer to improve quality and usability - Closes #2669

### LINKED ISSUE #2669: Refactor data persistence layer to improve quality and usability
Need to improve/refactor these points; 

- Remove any `Promise.resolve` from `then` blocks, and instead return value. 
- Always use `async` methods in tests, and only use `done` syntax when explicitly required. 
- Add usage of `Symbol` to conventions list
- Remove schema file added mistakenly 
- Replace `fieldSet` convention by boolean flag `extended` to make it easy to use 
- Update the sql file names according to above point for extended fields 
- Reuse the common logic for `get` and `getOne` 
- Rename `overrideDefaultOptions` to `extendDefaultOptions`
- Use Object spread instead of `Object.assign`
- Refactor `pgp` specific result specifiers to optimize the implementation
- Export all methods as `async` from entities which return Promise
- Initialize the storage in app.js add keep it under scope

nazarhussain:
### What was the problem?
Some improvements were required in basic structure of storage. 

### How did I fix it?
Based on points mentioned in the issue, updated the code. 

### Review checklist

* The PR resolves #2669
* Linting passes
* Tests pass


app.js
```diff
@@ -50,6 +50,7 @@ const Sequence = require('./helpers/sequence.js');
 const httpApi = require('./helpers/http_api.js');
 // eslint-disable-next-line import/order
 const swaggerHelper = require('./helpers/swagger');
+const createStorage = require('./storage');
```
lsilvs: We are doing the `require` inside each method for the other ones.
Maybe use the same approach here?

storage/entities/account.js
```diff
@@ -187,16 +187,6 @@ const readOnlyFields = ['address'];
  * @property {string} [votes_in]'
```
lsilvs: Just taking the opportunity to remove this :)

```suggestion
  * @property {string} [votes_in]
```

storage/entities/account.js
```diff
@@ -295,62 +283,28 @@ class Account extends BaseEntity {
...
-		return this.adapter.executeFile(
-			{
-				[Account.prototype.FIELD_SET_SIMPLE]: this.SQLs.selectSimple,
-				[Account.prototype.FIELD_SET_FULL]: this.SQLs.selectFull,
-			}[parsedOptions.fieldSet],
-			params,
-			{ expectedResult: 1 },
-			tx
-		);
+		return this._getResults(filters, options, tx, 1);
```
lsilvs: Prefer using variable names for function params to improve readability

storage/adapters/pgp_adapter.js
```diff
@@ -111,13 +105,12 @@ class PgpAdapter extends BaseAdapter {
...
 			.then(co => {
 				connectionObject = co;
-				return Promise.resolve(true);
+				return true;
 			})
 			.finally(() => {
 				if (connectionObject) {
 					connectionObject.done();
 				}
-				return Promise.resolve(true);
```
pablitovicente: is connectionObject.done() equivalent to this removed line?

storage/entities/account.js
```diff
@@ -483,8 +423,26 @@ class Account extends BaseEntity {
...
+		const parsedOptions = _.defaults(
+			{},
+			_.pick(options, ['limit', 'offset', 'extended']),
+			_.pick(this.defaultOptions, ['limit', 'offset', 'extended'])
+		);
+
+		const params = {
+			...{ limit: parsedOptions.limit, offset: parsedOptions.offset },
+			...{ parsedFilters },
+		};
```
lsilvs:
```
const params = {
	limit: parsedOptions.limit,
	offset: parsedOptions.offset,
	parsedFilters,
};
```

storage/adapters/pgp_adapter.js
```diff
@@ -190,6 +183,24 @@ class PgpAdapter extends BaseAdapter {
 	parseQueryComponent(query, params) {
 		return this.pgp.as.format(query, params);
 	}
+
+	_getExecutionHandler(tx, expectedResultCount) {
+		const count = Number(expectedResultCount);
+		const context = tx || this.db;
```
pablitovicente: the function is called _getExecutionHandler but returns a context.[none|one|any] could the names of function and variables be matched? maybe change `context` to `dbHandler` or `_getExecutionHandler` to `_getDBContext` or similar?

storage/storage.js
```diff
@@ -42,13 +42,14 @@ class Storage {
 	 * @return Promise
 	 */
 	bootstrap() {
-		const adapter = new PgpAdapter(
-			Object.assign({}, this.options, {
+		const adapter = new PgpAdapter({
+			...this.options,
+			...{
```
pablitovicente: Here you can directly set the keys without using the spread operator

test/unit/storage/entities/account.js
```diff
@@ -34,16 +32,14 @@ describe('Account', () => {
...
-		it(
-			'should call adapter.executeFile with proper param for FIELD_SET_SIMPLE'
-		);
-		it('should call adapter.executeFile with proper param for FIELD_SET_FULL');
+		it('should call adapter.executeFile with proper param for extended=false');
+		it('should call adapter.executeFile with proper param for extended=true');
 		it('should accept "tx" as last parameter and pass to adapter.executeFile');
 		it(
-			'should resolve with one object matching specification of type definition for FIELD_SET_SIMPLE'
+			'should resolve with one object matching specification of type definition of simple object'
```
pablitovicente: This is more of an integration test as in unit we're not using the DB then we'll be testing a mock is correct.

test/unit/storage/entities/account.js
```diff
@@ -34,16 +32,14 @@ describe('Account', () => {
...
 		);
 		it(
-			'should resolve with one object matching specification of type definition for FIELD_SET_FULL'
+			'should resolve with one object matching specification of type definition of full object'
```
pablitovicente: This is more of an integration test as in unit we're not using the DB then we'll be testing a mock is correct.

test/unit/storage/entities/account.js
```diff
@@ -63,16 +59,14 @@ describe('Account', () => {
...
-		it(
-			'should call adapter.executeFile with proper param for FIELD_SET_SIMPLE'
-		);
-		it('should call adapter.executeFile with proper param for FIELD_SET_FULL');
+		it('should call adapter.executeFile with proper param for extended=false');
+		it('should call adapter.executeFile with proper param for extended=true');
 		it('should accept "tx" as last parameter and pass to adapter.executeFile');
 		it(
-			'should resolve with array of objects matching specification of type definition for FIELD_SET_SIMPLE'
+			'should resolve with one object matching specification of type definition of simple object'
```
pablitovicente: This is more of an integration test as in unit we're not using the DB then we'll be testing a mock is correct.

test/unit/storage/entities/account.js
```diff
@@ -63,16 +59,14 @@ describe('Account', () => {
...
 		);
 		it(
-			'should resolve with array of objects matching specification of type definition for FIELD_SET_FULL'
+			'should resolve with one object matching specification of type definition of full object'
```
pablitovicente: This is more of an integration test as in unit we're not using the DB then we'll be testing a mock is correct.

storage/entities/base_entity.js
```diff
@@ -321,7 +310,7 @@ class BaseEntity {
 		let filtersObject = filters;
 
 		if (Array.isArray(filters)) {
-			filtersObject = Object.assign({}, ...filters);
+			filtersObject = { ...filters };
```
lsilvs: This won't work. You're deconstructing an array and creating an object of objects.

Maybe the following will do the trick?
```
filters.reduce((acc, curr) =>
	({ ...acc, ...curr }),
	{}
);
```

storage/entities/block.js
```diff
@@ -340,23 +275,43 @@ class Block extends BaseEntity {
...
+		const parsedOptions = _.defaults(
+			{},
+			_.pick(options, ['limit', 'offset']),
+			_.pick(this.defaultOptions, ['limit', 'offset'])
+		);
+
+		const params = {
+			...{ limit: parsedOptions.limit, offset: parsedOptions.offset },
+			...{ parsedFilters },
+		};
```
lsilvs:
```
const params = {
	limit: parsedOptions.limit,
	offset: parsedOptions.offset,
	parsedFilters,
};
```

storage/storage.js
```diff
@@ -42,13 +42,14 @@ class Storage {
...
 				inTest: process.env.NODE_ENV === 'test',
 				sqlDirectory: path.join(path.dirname(__filename), './sql'),
 				logger: this.logger,
-			})
-		);
+			},
+		});
```
lsilvs:
```
const adapter = new PgpAdapter({
	...this.options,
	inTest: process.env.NODE_ENV === 'test',
	sqlDirectory: path.join(path.dirname(__filename), './sql'),
	logger: this.logger,
})
```

test/unit/storage/entities/account.js
```diff
@@ -34,16 +32,14 @@ describe('Account', () => {
 		it('should throw error for in-valid filters');
 		it('should accept only valid options');
 		it('should throw error for in-valid options');
-		it(
-			'should call adapter.executeFile with proper param for FIELD_SET_SIMPLE'
-		);
-		it('should call adapter.executeFile with proper param for FIELD_SET_FULL');
+		it('should call adapter.executeFile with proper param for extended=false');
+		it('should call adapter.executeFile with proper param for extended=true');
```
pablitovicente: This tests now need to be changed as now `get()` and `getOne()` both call `_getResults()` so I think the tests for `get()` and `getOne()` need to only check they're calling `_getResults()` with the valid parameters and then the heavy load on testing is for the implementation of that method. This change in tests should be implemented for all existing tests.

storage/entities/account.js
```diff
@@ -463,11 +401,13 @@ class Account extends BaseEntity {
 	 * Check if the record exists with following conditions
 	 *
 	 * @param {filters.Account} filters
-	 * @param {Object} [options]
```
pablitovicente: This should be udpated to `_options` to match the method's argument name.

storage/adapters/pgp_adapter.js
nazarhussain: Yes i read the documentation and it says return value from `finally` has no impact on resolved value. http://bluebirdjs.com/docs/api/finally.html

storage/storage.js
nazarhussain: Need to merge this with `this.options` so why have to use spread here.

test/unit/storage/entities/account.js
nazarhussain: We will not be mocking for unit tests on DB level execution. We already have detail discussion on this topic. For DB execution, we will execute the queries directly on PostgreSQL and check the response. That's how we already doing and will use the same pattern.

nazarhussain: @lsilvs and I had detail discussion on this topic. He may add his thoughts here.

test/unit/storage/entities/account.js
nazarhussain: Same as above.

test/unit/storage/entities/account.js
nazarhussain: Same as above.

test/unit/storage/entities/account.js
nazarhussain: Same as above.

test/unit/storage/entities/account.js
pablitovicente: Ohh as this were unit tests I thought we will not be hitting the DB at all. Shouldn't then move this tests to integration then?

app.js
nazarhussain: That was a bad approach, don't know why it was used.

test/unit/storage/entities/account.js
lsilvs: I definitely think we should have tests that execute the query and check the response. I don't mind how we call them but I do think they should be separated from these tests here.

pablitovicente: After discussing it was agreed for this tests to connect to the DB for executing queries in the cases that mocking is not enough to test a method so I'm resolving this.

test/unit/storage/entities/account.js
pablitovicente: Same answer as above.

test/unit/storage/entities/account.js
pablitovicente: Same answer as above.

app.js
lsilvs: I agree but imo we should either follow the approach or fix it all together.

test/unit/storage/entities/account.js
pablitovicente: Same resolution as above.

storage/entities/block.js
```diff
@@ -262,30 +263,19 @@ class Block extends BaseEntity {
...
-		);
-
-		const params = {
-			...{ limit: parsedOptions.limit, offset: parsedOptions.offset },
-			...{ parsedFilters },
-		};
 
 		return this.adapter
-			.executeFile(this.SQLs.isPersisted, { params }, {}, tx)
+			.executeFile(this.SQLs.isPersisted, { parsedFilters }, {}, tx)
```
lsilvs: Doesn't pg-promise expect `params` attribute here?

lsilvs: My bad. It looks ok ;)

test/unit/storage/entities/account.js
pablitovicente: As per talk this will be dealt with in different issue/PR

storage/README.md
```diff
@@ -61,3 +61,4 @@ Following conventions must be followed strictly
 * Timestamp types always refer to the network Epoch timestamp. In the case of Unix timestamp, the attribute name should be prefixed with `unix_`.
 * All values related to the amounts should be exposed as strings, due to the limited JavaScript integers precision (find out more while reading about Number. MAX_SAFE_INTEGER). The block height and round number can be safely stored as integers and should be exposed as integers.
 * All `BYTEA` fields in 'hex' encoding should be exposed as strings. Any other encoding should be exposed as Buffer.
+* When required to use constants, use `Symbol` and then export it from interface, so later everyone explicitly use those defined constants instead of strings.
```
matchabay:
```suggestion
* Use `Symbol` to define constants within database entities and export them, so the other modules can use them explicitly instead of strings.
```

MaciejBaj:
```suggestion
* Use `Symbol` to define constants within database entities and export them, so the other modules can use them explicitly instead of strings.
```

PR #2681, opened 2018-12-20
Implement count method for entities - Closes #2676

### LINKED ISSUE #2676: Implement count method for entities
### Expected behavior
We are using `count` for different entities in our APIs, to make sure everything works the same way, new storage entities must expose an interface which use filters based counting for records. 

### Actual behavior
Currently no support of counting records for any entity.

lsilvs:
### What was the problem?
There was no support of counting records for any entity.

### How did I fix it?
I've implemented count method for Accounts and Blocks and also updated unit test for blocks.

### How to test it?
`npm run mocha -- test/unit/storage/`
and
`storage.entities.Account.count({ isDelegate: 1 });`
`storage.entities.Block.count({ height_gt: 101 });`

### Review checklist

* The PR resolves #2676
* Linting passes
* Tests pass

### DIFF
--- a/storage/entities/account.js
+++ b/storage/entities/account.js
@@ -284,6 +284,7 @@ class Account extends BaseEntity {
 		this.SQLs = {
 			selectSimple: this.adapter.loadSQLFile('accounts/get.sql'),
 			selectFull: this.adapter.loadSQLFile('accounts/get_extended.sql'),
+			count: this.adapter.loadSQLFile('accounts/count.sql'),
 			create: this.adapter.loadSQLFile('accounts/create.sql'),
 			update: this.adapter.loadSQLFile('accounts/update.sql'),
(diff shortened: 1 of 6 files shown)


storage/entities/account.js
```diff
@@ -322,6 +323,24 @@ class Account extends BaseEntity {
...
+	count(filters = {}) {
+		this.validateFilters(filters);
+
+		const mergedFilters = this.mergeFilters(filters);
+		const parsedFilters = this.parseFilters(mergedFilters);
+		const expectedResultCount = 1;
+
+		return this.adapter
+			.executeFile(this.SQLs.count, { parsedFilters }, { expectedResultCount })
+			.then(result => +result.count);
```
nazarhussain: There is a way to do same in Pg-Promise, but to make the implementation generic we can go with this approach.

PR #2682, opened 2018-12-20
Add unit tests for signatures module - Closes #2445

### LINKED ISSUE #2445: Add unit test coverage for signatures module
### Expected behavior
Signatures module must be covered with unit tests.
### Actual behavior
We just have suite skeleton at the moment. See: [test/unit/modules/signatures.js](https://github.com/LiskHQ/lisk/blob/development/test/unit/modules/signatures.js)
1.3.0+

limiaspasdaniel:
### What was the problem?
Unit tests for **signatures** module weren't implemented
### How did I fix it?
Implemented unit tests for **signatures** module
### How to test it?
`npm run mocha -- test/unit/signatures`

### Review checklist

* The PR resolves #2445 
* Linting passes
* Tests pass

### DIFF
--- a/test/unit/modules/signatures.js
+++ b/test/unit/modules/signatures.js
@@ -19,7 +19,6 @@ const transactionTypes = require('../../../helpers/transaction_types');
 const ApiError = require('../../../helpers/api_error');
 
 /* eslint-disable mocha/no-pending-tests */
-/* eslint-disable no-unused-vars */
 describe('signatures', () => {
 	let library;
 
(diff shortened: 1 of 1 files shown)


limiaspasdaniel:
@yatki perhaps I'm wrong but the [postSignatures](https://github.com/LiskHQ/lisk/blob/0e0281ffe103b6e3192618cd8f81d3f616fa26d5/modules/signatures.js#L145) function on the signatures module is not being used anywhere and after discussing with @jondubois it was probably meant for some other case we don't use now. Also, inside the function implementation:
```
postSignatures(signatures, cb) {
  return modules.transport.shared.postSignatures(
    { signatures },
    (err, res) => {
	__private.processPostResult(err, res, cb);
    }
  );
}
```
the `modules.transport.shared.postSignatures` function definition isn't accepting a callback as expected by the [unit tests](https://github.com/LiskHQ/lisk/blob/0e0281ffe103b6e3192618cd8f81d3f616fa26d5/test/unit/modules/signatures.js#L55), it's accepting [ just an argument called query](https://github.com/LiskHQ/lisk/blob/0e0281ffe103b6e3192618cd8f81d3f616fa26d5/modules/transport.js#L711).

The unit tests skeleton then isn't coherent with the current implementation.

What I did was instead, add unit testing coverage for the [postSignature](https://github.com/LiskHQ/lisk/blob/0e0281ffe103b6e3192618cd8f81d3f616fa26d5/modules/signatures.js#L132) function which is indeed being used.

Any thoughts on this? @MaciejBaj @jondubois @yatki

yatki:
@limiaspasdaniel you are right. There is a new relic definition for `postSignatures` tho but I think this is just for mapping purpose. @nazarhussain can you take a look and confirm. 

You are also right on your comment regarding `modules.transport.shared.postSignatures`. I don't see any issue with updating test skeletons.

test/unit/modules/signatures.js
```diff
@@ -14,66 +14,226 @@
 
 'use strict';
 
+const application = require('../../common/application');
+const transactionTypes = require('../../../helpers/transaction_types');
+const ApiError = require('../../../helpers/api_error');
+
 /* eslint-disable mocha/no-pending-tests */
+/* eslint-disable no-unused-vars */
```
MaciejBaj: Why is it important to keep?

limiaspasdaniel: I added it due to ESLint being picky when not using the last arguments on callback definitions. For example [here](https://github.com/LiskHQ/lisk/blob/00735eb98e22e4b4d8b007036f30d150f6750fcd/test/unit/modules/signatures.js#L160) and [here](https://github.com/LiskHQ/lisk/blob/00735eb98e22e4b4d8b007036f30d150f6750fcd/test/unit/modules/signatures.js#L195). Also it wouldn't let me type "_" as the name of the arguments to dismiss those errors.

yatki: If possible, please do not disable rules on file level. Disable the relevant lines instead with eslint-disable-next-line

test/unit/modules/signatures.js
```diff
@@ -14,66 +14,226 @@
...
+					(error, response) => {
+						done();
+					}
 				);
+			});
 
-				it('should call callback with ApiError containing code = 500');
+			it('should call modules.transport.shared.postSignature with req.body', done => {
+				expect(postSignatureSpy).to.be.calledWith({ signature: req.body });
+				sinonSandbox.restore();
```
yatki: please restore the sandbox in after hook. And restoring `sinonSandbox` here is not ok since you didn't use `beforeEach` but `before`

test/unit/modules/signatures.js
```diff
@@ -14,66 +14,226 @@
...
+				body: {
+					signature: 'aSignature',
+					transactionId: 'aTransactionId',
+					publicKey: 'aPublicKey',
+				},
+			};
+
+			let postSignatureSpy;
+
+			before(done => {
```
yatki: beforeEach would be better.

test/unit/modules/signatures.js
```diff
@@ -14,66 +14,226 @@
...
+					postSignatureStub = sinonSandbox.stub(
+						library.modules.transport.shared,
+						'postSignature'
+					);
+					done();
+				});
+
+				after(done => {
+					sinonSandbox.restore();
+					done();
```
yatki: sandbox.restore is synchronous. No need for `done`.

test/unit/modules/signatures.js
```diff
@@ -14,66 +14,226 @@
...
+	});
+
 	describe('isLoaded', () => {
-		it('should return true if modules exists');
+		it('should return true if modules exists', done => {
+			library.rewiredModules.signatures.__set__('modules', {
+				accounts: library.modules.accounts,
+				transactions: library.modules.transactions,
+				transport: library.modules.transport,
+			});
```
yatki: I would avoid setting rewiredModules in test scope. Also this will mutate the library object for upcoming tests which can cause having non-atomic tests.

test/unit/modules/signatures.js
```diff
@@ -14,66 +14,226 @@
...
+			expect(library.modules.signatures.isLoaded()).to.be.true;
+			done();
+		});
 
-		it('should return true if modules does not exist');
+		it('should return true if modules does not exist', done => {
+			const revert = library.rewiredModules.signatures.__set__(
+				'modules',
+				undefined
+			);
```
yatki: same above.

test/unit/modules/signatures.js
```diff
@@ -14,66 +14,226 @@
...
+			expect(library.modules.signatures.isLoaded()).to.be.false;
+			revert();
```
yatki: please do this kind of operations in the test hooks.

test/unit/modules/signatures.js
```diff
@@ -14,66 +14,226 @@
...
+			done();
+		});
 	});
 
 	describe('onBind', () => {
+		let privateModules;
+		let signatureLogicSpy;
+
+		before(done => {
```
yatki: prefer beforeEach

test/unit/modules/signatures.js
```diff
@@ -14,66 +14,226 @@
...
+				done();
+			});
 
-			it('should assign transport');
+			it('should assign transport', done => {
+				expect(privateModules).to.have.property(
+					'transport',
+					library.modules.transport
+				);
+				done();
```
yatki: The three tests above are not async.. So please avoid using `done`. Prefer sync function when it's not async.

test/unit/modules/signatures.js
```diff
@@ -14,66 +14,226 @@
...
+			});
 		});
 
 		describe('assetTypes', () => {
-			it('should call bind on signature logic with scope.accounts');
+			it('should call bind on signature logic with scope.accounts', done => {
+				expect(signatureLogicSpy).to.be.calledWith(library.modules.accounts);
+				done();
```
yatki: same above.

test/unit/modules/signatures.js
```diff
@@ -14,66 +14,226 @@
...
-				describe('when modules.transport.shared.postSignatures fails with result', () => {
-					it('should call callback with ApiError');
+				it('should call callback with ApiError', done => {
+					postSignatureStub.yields(null, { success: false });
 
-					it(
-						'should call callback with ApiError containing message = "Blockchain is loading"'
+					library.modules.signatures.shared.postSignature(
+						req.body,
+						(error, result) => {
```
yatki: you don't have to define the result argument since it's the second argument and not necessary here.

test/unit/modules/signatures.js
```diff
@@ -14,66 +14,226 @@
...
 				});
+
+				it('should call callback with error = null', done => {
+					expect(postSignatureError).to.be.null;
+					done();
+				});
+
+				it('should call callback with result containing status = "Signature Accepted"', done => {
+					expect(postSignatureResult.status).to.equal('Signature Accepted');
+					done();
```
yatki: Please apply my comments above to this file as well.

test/unit/modules/signatures.js
limiaspasdaniel: Fixed

test/unit/modules/signatures.js
limiaspasdaniel: Done

test/unit/modules/signatures.js
limiaspasdaniel: Done

test/unit/modules/signatures.js
limiaspasdaniel: Done

test/unit/modules/signatures.js
limiaspasdaniel: Done

test/unit/modules/signatures.js
limiaspasdaniel: Done

test/unit/modules/signatures.js
limiaspasdaniel: Done

test/unit/modules/signatures.js
limiaspasdaniel: Done

test/unit/modules/signatures.js
limiaspasdaniel: Done

test/unit/modules/signatures.js
limiaspasdaniel: Done

limiaspasdaniel:
@yatki applied your suggested changes.

test/unit/modules/signatures.js
```diff
@@ -14,66 +14,200 @@
...
+				],
+				'bind'
+			);
+			return library.modules.signatures.onBind(library.modules);
+		});
+
+		after(() => {
+			return signatureLogicSpy.restore();
+		});
+
```
yatki: I think moving `signatureLogicSpy` definition  `assetTypes` suite would be better since they were only being used there ?

test/unit/modules/signatures.js
yatki: please don't forget to use // Arrange, Act, Assert comments when logic gets a bit complicated inside the test.

limiaspasdaniel: Addressed it

test/unit/modules/signatures.js
limiaspasdaniel: Makes total sense. Done

limiaspasdaniel:
@yatki done, thanks :)

MaciejBaj:
Good job @limiaspasdaniel and @yatki on the review 👍

PR #2683, opened 2018-12-20
Create methods must accept arrays as well - Closes #2677

### LINKED ISSUE #2677: Create methods must accept arrays as well
### Expected behavior
While creating objects specially for transactions and peers we can provide bulk amount of data in form of array. The `create` interface must be check if provided data is array, then use multiple insertion in query. 

### Actual behavior
Currently it supports only one object for creating records.

nazarhussain:
### What was the problem?
The `storage.entities.Account.create` was accepting only one object. While our implementation requires to create multiple accounts as well. 

### How did I fix it?
Added the logic to make sure it accepts both arrays and objects. 

### Review checklist

* The PR resolves #2677
* Linting passes
* Tests pass

### DIFF
--- a/storage/entities/account.js
+++ b/storage/entities/account.js
@@ -354,7 +354,7 @@ class Account extends BaseEntity {
 	 * @return {*}
 	 */
 	create(data, _options, tx) {
- 		assert(data, 'Must provide data to create account');
+		assert(data, 'Must provide data to create account');
 		assert(
 			typeof data === 'object' || Array.isArray(data),
 			'Data must be an object or array of objects'
--- a/storage/entities/account.js
+++ b/storage/entities/account.js
@@ -340,14 +340,18 @@ class Account extends BaseEntity {
 		}
 
 		values = values.map(v => _.defaults(v, defaultCreateValues));
-		const createSet = this.getValuesSet(values);
-		const attributes = Object.keys(values[0])
+
+		// We assume that all accounts have same attributes
+		// and pick first account attributes as template
+		const attributes = Object.keys(values[0]);
+		const createSet = this.getValuesSet(values, attributes);
+		const fields = attributes
 			.map(k => `"${this.fields[k].fieldName}"`)
 			.join(',');
 
 		return this.adapter.executeFile(
 			this.SQLs.create,
-			{ createSet, attributes },
+			{ createSet, fields },
 			{ expectedResultCount: 0 },
 			tx
 		);
--- a/storage/entities/account.js
+++ b/storage/entities/account.js
@@ -325,22 +325,30 @@ class Account extends BaseEntity {
 	/**
 	 * Create account object
 	 *
-	 * @param {Object} data
-	 * @param {Object} [options]
-	 * @param {Object} tx - Transaction object
+	 * @param {Object|Array.<Object>} data
+	 * @param {Object} [_options]
+	 * @param {Object} [tx] - Transaction object
 	 * @return {*}
 	 */
 	create(data, _options, tx) {
-		const objectData = _.defaults(data, defaultCreateValues);
-		const createSet = this.getValuesSet(objectData);
-		const attributes = Object.keys(data)
+		let values;
+
+		if (Array.isArray(data)) {
+			values = data;
+		} else if (typeof data === 'object') {
+			values = [data];
+		}
+
+		values = values.map(v => _.defaults(v, defaultCreateValues));
+		const createSet = this.getValuesSet(values);
+		const attributes = Object.keys(values[0])
 			.map(k => `"${this.fields[k].fieldName}"`)
 			.join(',');
 
 		return this.adapter.executeFile(
 			this.SQLs.create,
 			{ createSet, attributes },
-			{},
+			{ expectedResultCount: 0 },
 			tx
 		);
 	}
--- a/storage/entities/account.js
+++ b/storage/entities/account.js
@@ -14,6 +14,7 @@
 
 'use strict';
 
+const assert = require('assert');
 const _ = require('lodash');
 const { stringToByte, booleanToInt } = require('../utils/inputSerializers');
 const ft = require('../utils/filter_types');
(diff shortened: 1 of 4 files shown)


storage/entities/account.js
```diff
@@ -325,22 +325,30 @@ class Account extends BaseEntity {
...
+
+		if (Array.isArray(data)) {
+			values = data;
+		} else if (typeof data === 'object') {
+			values = [data];
+		}
+
+		values = values.map(v => _.defaults(v, defaultCreateValues));
+		const createSet = this.getValuesSet(values);
+		const attributes = Object.keys(values[0])
```
lsilvs: I find it risky => `Object.keys(values[0])`
There is no validation for `data` meaning you can't guarantee it will only contain the `defaultCreateValues`. Also, if the `data` attributes have different order in different objects in the array, it will error.

I would add a validation for `data` and then order its keys to match `defaultCreateValues`.

nazarhussain: I don't think order of keys will have any impact. And wince we are merging all keys to default values, so all objects must have same keys.

lsilvs: You're trying to construct this, right?
```
INSERT INTO table (id, name)
VALUES  ('1', 'name1'),
        ('2', 'name2'),
        ('3','name3')
```
If so, the order matter.

Try `storage.entities.Account.create([{ address: '123L', username: 'primeiro' }, { username: 'segundo', address: '321L' }])` and you gonna see what I'm talking about.

pablitovicente: I do agree with @lsilvs here it could cause psql rejections in the best case or invalid data in the worst case (if types of data match)

storage/entities/account.js
```diff
@@ -325,22 +325,34 @@ class Account extends BaseEntity {
...
+	 * @param {Object|Array.<Object>} data
+	 * @param {Object} [_options]
+	 * @param {Object} [tx] - Transaction object
 	 * @return {*}
 	 */
 	create(data, _options, tx) {
-		const objectData = _.defaults(data, defaultCreateValues);
-		const createSet = this.getValuesSet(objectData);
-		const attributes = Object.keys(data)
+		let values;
```
pablitovicente: I would extract this to a different method so it can be reused for updates as well.

storage/entities/account.js
```diff
@@ -325,22 +325,34 @@ class Account extends BaseEntity {
...
+			values = data;
+		} else if (typeof data === 'object') {
+			values = [data];
+		}
+
+		values = values.map(v => _.defaults(v, defaultCreateValues));
+
+		// We assume that all accounts have same attributes
+		// and pick first account attributes as template
+		const attributes = Object.keys(values[0]);
```
lsilvs: Why not use `this.fields` or `defaultCreateValues` as attributes?
It's clearer than `values[0]` and avoid invalid fields if data contain any.

storage/entities/base_entity.js
```diff
@@ -175,13 +175,12 @@ class BaseEntity {
...
 	}
 
-	getValuesSet(data) {
-		return `(${this.adapter.parseQueryComponent(
-			Object.keys(data)
-				.map(key => this.fields[key].serializeValue(data[key], 'insert'))
-				.join(','),
-			data
-		)})`;
+	getValuesSet(data, attributes = []) {
```
lsilvs: You can't set `attribute` default value as empty array. If you do, `_getValueSetForObject` will map over the empt array and will never use `Object.keys(data)`

storage/entities/base_entity.js
```diff
@@ -332,6 +331,15 @@ class BaseEntity {
 		}
 		return { ...filters, ...this.defaultFilters };
 	}
+
+	_getValueSetForObject(data, attributes = []) {
```
lsilvs: Same as above

storage/entities/account.js
```diff
@@ -325,22 +325,34 @@ class Account extends BaseEntity {
...
 	 */
 	create(data, _options, tx) {
-		const objectData = _.defaults(data, defaultCreateValues);
-		const createSet = this.getValuesSet(objectData);
-		const attributes = Object.keys(data)
+		let values;
+
+		if (Array.isArray(data)) {
+			values = data;
+		} else if (typeof data === 'object') {
```
MaciejBaj: Doesn't work for `null`.

storage/entities/account.js
```diff
@@ -325,22 +325,34 @@ class Account extends BaseEntity {
...
+			values = [data];
+		}
```
lsilvs: Even though `values` is an auxiliary variable it will mutate the value in `data`.
To fix that you need something like:
```
if (Array.isArray(data)) {
	values = data.map(item => ({ ...item }));
} else if (typeof data === 'object') {
	values = [{ ...data }];
}
```

storage/entities/account.js
nazarhussain: Who will pass `null` to create account? 🤕

storage/entities/account.js
nazarhussain: We are not re-assigning or changing `data`, then how it would be mutated? can't understand.

storage/entities/account.js
lsilvs: This should be checked in the validation. If not implemented in this PR, it should have a separated issue for it.

storage/entities/account.js
nazarhussain: I think, for now we can go with, where we find any obvious case for code reusability, we will refactor.

storage/entities/account.js
lsilvs: In JS arrays and objects are pointers meaning when you do `values = data` you are assigning the `data` pointer to `value`.

storage/entities/account.js
```diff
@@ -325,22 +326,40 @@ class Account extends BaseEntity {
...
+	 * @param {Object|Array.<Object>} data
+	 * @param {Object} [_options]
+	 * @param {Object} [tx] - Transaction object
 	 * @return {*}
 	 */
 	create(data, _options, tx) {
-		const objectData = _.defaults(data, defaultCreateValues);
-		const createSet = this.getValuesSet(objectData);
-		const attributes = Object.keys(data)
+		assert(data, 'Must provide data to create account.');
```
MaciejBaj: We have never used a
```suggestion
 		assert(data, 'Must provide data to create account');
```

PR #2697, opened 2019-01-03
Remove clock from Peers - Closes #2687

### LINKED ISSUE #2687: Remove clock from Peers
### Expected behavior

Clock should not exist in Peer entity

### Actual behavior

Clock exists in Peer entity

### Steps to reproduce

- Check Peer entity
- This change needs checks and adjustments to old tests and code scanning to make sure it's safe to delete

pablitovicente:
### What was the problem?

Clocks should be removed from Peers as it was not being used by the application

### How did I fix it?

- Removed the clocks column from peers
- Removed references to clock from code
- Removed references to clock from peers

### How to test it?

- Run the application `node app.js` and the migration should be run and no clock column should exist in the peers table
- Build should be green
- Peers should be persisted correctly when stoping the applicaton

### Review checklist

* The PR resolves #2687
* Linting passes
* Tests pass

### DIFF
--- a/test/unit/logic/transaction.js
+++ b/test/unit/logic/transaction.js
@@ -415,7 +415,7 @@ describe('transaction', () => {
 		it('should throw an error with no param', () => {
 			return expect(
 				transactionLogic.checkConfirmed.bind(transactionLogic)
-			).to.throw('Callback must be a function');
+			).to.throw('"callback" argument must be a function');
 		});
 
 		it('should return an error with no transaction', done => {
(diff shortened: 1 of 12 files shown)


test/unit/logic/transaction.js
```diff
@@ -415,7 +415,7 @@ describe('transaction', () => {
 		it('should throw an error with no param', () => {
 			return expect(
 				transactionLogic.checkConfirmed.bind(transactionLogic)
-			).to.throw('Callback must be a function');
+			).to.throw('"callback" argument must be a function');
```
lsilvs: Assuming you're running node v10, this shouldn't be changed.

pablitovicente: Good catch was running node 8.14. Done. Resolving.

PR #2710, opened 2019-01-07
Remove db.accounts repository - Closes #2709

### LINKED ISSUE #2709: Remove db.accounts repository
Remove `db.accounts` and move custom interfaces to `storage.entities.Account`

nazarhussain:
### What was the problem?
Some interfaces in `db.accounts` having duplicate behavior what can be achieved by `storage.entities.Account` CRUD operations. Reset are some custom interfaces.

### How did I fix it?

Tried to utilized mostly the CRUD and then exported custom interfaces to storage entity. 

### How to test it?
Run all unit and functional tests. 

### Review checklist

* The PR resolves #2709
* Linting passes
* Tests pass


MaciejBaj:
Ok, please log is an issue then.

storage/entities/account.js
```diff
@@ -462,6 +497,200 @@ class Account extends BaseEntity {
...
+	/**
+	 * Increment a field value in mem_accounts.
+	 *
+	 * @param {filters.Account} [filters] - Filters to match the objects
+	 * @param {string} field - Name of the field to increment
+	 * @param {Number|string} value - Value to be incremented
+	 * @param {Object} [tx] - Transaction object
+	 * @return {*}
+	 */
+	increment(filters, field, value, tx) {
```
lsilvs: I would name it `incrementField`

storage/entities/account.js
```diff
@@ -462,6 +497,200 @@ class Account extends BaseEntity {
...
+	/**
+	 * Decrement a field value in mem_accounts.
+	 *
+	 * @param {filters.Account} [filters] - Filters to match the objects
+	 * @param {string} field - Name of the field to increment
+	 * @param {Number|string} value - Value to be incremented
+	 * @param {Object} [tx] - Transaction object
+	 * @return {*}
+	 */
+	decrement(filters, field, value, tx) {
```
lsilvs: I would name it `decrementField`

storage/entities/account.js
```diff
@@ -462,6 +497,200 @@ class Account extends BaseEntity {
...
+			SQL,
+			params,
+			{ expectedResultCount: 0 },
+			tx
+		);
+	}
+
+	_updateField(filters, field, value, mode = '+', tx) {
+		const atLeastOneRequired = true;
+		const validFieldName = Object.keys(this.fields).includes(field);
```
lsilvs: Check also if the field type is number

storage/entities/account.js
```diff
@@ -462,6 +497,200 @@ class Account extends BaseEntity {
...
+
+		return this.adapter.executeFile(
+			SQL,
+			params,
+			{ expectedResultCount: 0 },
+			tx
+		);
+	}
+
+	_updateField(filters, field, value, mode = '+', tx) {
```
lsilvs: Default value for mode isn't required as this is an internal method and we have the control of its use

```suggestion
	_updateField(filters, field, value, mode, tx) {
```

storage/entities/account.js
```diff
@@ -462,6 +497,200 @@ class Account extends BaseEntity {
...
+		const parsedFilters = this.parseFilters(mergedFilters);
+		const filedName = this.fields[field].fieldName;
+
+		const params = {
+			parsedFilters,
+			field: filedName,
+			value,
+		};
+
+		const SQL = mode === '+' ? this.SQLs.increment : this.SQLs.decrement;
```
lsilvs: Prefer using lowercase var names

storage/entities/account.js
```diff
@@ -462,6 +497,200 @@ class Account extends BaseEntity {
...
+		return this.adapter.executeFile(
+			SQL,
+			params,
+			{ expectedResultCount: 0 },
+			tx
+		);
+	}
+
+	_updateField(filters, field, value, mode = '+', tx) {
+		const atLeastOneRequired = true;
```
lsilvs: I don't think it should be required.

storage/entities/account.js
```diff
@@ -462,6 +497,200 @@ class Account extends BaseEntity {
...
+			'-',
+			tx
+		);
+	}
+
+	_updateDependentRecord(
+		dependencyName,
+		address,
+		dependentPublicKey,
+		mode = '+',
```
lsilvs: I would suggest using `insert` and `delete` rather than +/-

storage/entities/account.js
```diff
@@ -462,6 +497,200 @@ class Account extends BaseEntity {
...
+		assert(
+			Object.keys(dependentFieldsTableMap).includes(dependencyName),
+			`Invalid dependency name "${dependencyName}" provided.`
+		);
+		const params = {
+			tableName: dependentFieldsTableMap[dependencyName],
+			accountId: address,
+			dependentId: dependentPublicKey,
+		};
+		const SQL =
```
lsilvs: Prefer using lowercase var names

modules/loader.js
```diff
@@ -532,38 +532,32 @@ __private.loadBlockChain = function() {
...
-					const promises = [
-						t.accounts.updateMemAccounts(),
-						t.accounts.getDelegates(),
-					];
-					return t.batch(promises);
-				}
+				await library.storage.entities.Account.resetUnconfirmedState();
+				const delegatesPublicKeys = await library.storage.entities.Account.get(
+					{ isDelegate: true },
+					{ limit: null, offset: null }
```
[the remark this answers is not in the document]
nazarhussain: Yes want to get all delegates, so have to set the offset to null for it.

storage/entities/account.js
```diff
@@ -299,6 +312,19 @@ class Account extends BaseEntity {
 			update: this.adapter.loadSQLFile('accounts/update.sql'),
 			updateOne: this.adapter.loadSQLFile('accounts/update_one.sql'),
 			isPersisted: this.adapter.loadSQLFile('accounts/is_persisted.sql'),
+			delete: this.adapter.loadSQLFile('accounts/delete.sql'),
+			resetUnconfirmedState: this.adapter.loadSQLFile(
+				'accounts/reset_unconfirmed_state.sql'
+			),
+			resetMemTables: this.adapter.loadSQLFile('accounts/reset_mem_tables.sql'),
```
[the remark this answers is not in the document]
nazarhussain: Logically the table `rounds_rewards` should be renamed to `mem_rounds_rewards` as its also a temporarily table can be regenerated from the blockchain data. All of those tables and query are intended for the same purpose resetting accounts data. So seems perfect candidate as `resetMemTables` in `Account` entity.

storage/entities/account.js
```diff
@@ -404,7 +430,16 @@ class Account extends BaseEntity {
 	 * @return {*}
 	 */
 	update(filters, data, _options, tx) {
+		const atLeastOneRequired = true;
```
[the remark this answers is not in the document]
nazarhussain: Yes logically it should. But due to nature of the application, I just want to make sure we don't update the account without any scope. That's why I added this check here to have at least one filter to update accounts.

storage/entities/account.js
```diff
@@ -404,7 +430,16 @@ class Account extends BaseEntity {
...
+
+		this.validateFilters(filters, atLeastOneRequired);
+
 		const objectData = _.omit(data, readOnlyFields);
+
+		if (_.isEmpty(objectData)) {
+			return Promise.resolve();
```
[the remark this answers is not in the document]
nazarhussain: Its intended to keep the existing behavior of the application. 

On many places application is giving data `objectData = {address: '1234L'}`, for which `address` can't be updated. So the `objectData` will get empty. If we throw error in that case, it will stop the promise chain to update accounts in the block processing. Then we have to change the business logic of the block processing to not stop the promise chain in this particular case.

I want to avoid any kind of change in business logic, to have confidence of compatibility with previous versions.

storage/entities/account.js
```diff
@@ -462,6 +497,200 @@ class Account extends BaseEntity {
...
+	/**
+	 * Update data based on filters or insert data if no matching record found
+	 *
+	 * @param {filters.Account} filters - Filters to match the object
+	 * @param {Object} data - Object data to be inserted
+	 * @param {Object} [updateData] - If provided will be used for update, otherwise default data will be updated
+	 * @param {Object} [tx] - DB transaction object
+	 * @returns {Promise.<boolean, Error>}
+	 */
+	upsert(filters, data, updateData = {}, tx = null) {
```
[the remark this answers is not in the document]
nazarhussain: Yes it can be, but again it will be change in application logic, that I don't want to introduce at the moment.

storage/entities/account.js
nazarhussain: Yes don't want to update fields fo all accounts.

modules/loader.js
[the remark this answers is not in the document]
lsilvs: I had a similar situation for blocks and using only `limit: null` was enough.
`offset` is zero by default so I don't see the need of sending as null.

storage/entities/account.js
nazarhussain: Its just not for numbers, values defined as Bignumebrs are strings. So lets keep this check for later implementation of json-schema based validation. For now its already handled in `logic/account.js`

test/integration/rounds.js
```diff
@@ -58,6 +58,15 @@ describe('rounds', () => {
...
 			const accounts = {};
 			_.map(rows, acc => {
+				acc.nameExist = acc.nameexist;
+				acc.u_nameExist = acc.u_nameexist;
+				acc.multiLifetime = acc.multilifetime;
+				acc.u_multiLifetime = acc.u_multilifetime;
+				delete acc.nameexist;
+				delete acc.u_nameexist;
+				delete acc.multilifetime;
+				delete acc.u_multilifetime;
```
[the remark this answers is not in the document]
nazarhussain: Yes I would like to have seperate fixtures, but this change is not related to fixtures. Its just fixing up result from the `raw query`. Since `Queries.getAccounts()` on many places as well, so can't actually update the query itself.

storage/entities/account.js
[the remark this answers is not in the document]
lsilvs: But having in mind we want the db layer to be extensible and used by other applications I reckon it shouldn't be a constraint here but in the application level.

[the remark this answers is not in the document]
nazarhussain: For now have no plan to expose the UPDATE/INSERT/DELETE to others. ;)

MaciejBaj:
Ok, I guess all of the remainings in `db/repos` directory will be cleaned as part of https://github.com/LiskHQ/lisk/issues/2632.

PR #2724, opened 2019-01-09
Remove db.transactions repository - Closes #2628

### LINKED ISSUE #2628: Remove db.transactions repository
Remove `db.transactions` and move custom interfaces to `storage.entities.Transaction`

pablitovicente:
### What was the problem?

Transactions db repository still in use and can be replaced with Transactions from Storage

### How did I fix it?

By replacing usages with storage

### How to test it?

Build should be green

### Review checklist

* The PR resolves #2628
* Linting passes
* Tests pass


storage/entities/transaction.js
```diff
@@ -342,6 +346,8 @@ class Transaction extends BaseEntity {
 			t.signatures = t.signatures ? t.signatures.join() : null;
 			t.amount = t.amount.toString();
 			t.fee = t.fee.toString();
+			// Set default create values
+			_.defaults(t, defaultCreateValues);
```
nazarhussain:
```suggestion
			t.recipientId = t.recipientId || null;
```

storage/entities/transaction.js
```diff
@@ -374,7 +380,17 @@ class Transaction extends BaseEntity {
...
 
-			const groupedTransactions = _.groupBy(transactions, 'type');
+			// Not all transactions of type 0 might have an `asset` so remove the ones without `asset` before storing related data
+			const needSubtransactions = transactions.filter(
+				transaction =>
+					transaction.type !== 0 ||
+					(transaction.type === 0 &&
+						transaction.asset &&
+						transaction.asset.data)
+			);
```
nazarhussain: If any filtering required baed on transaction type, it should be done under `switch` below.

storage/entities/transaction.js
```diff
@@ -402,12 +418,10 @@ class Transaction extends BaseEntity {
...
-				values = transactions
-					.filter(transaction => transaction.asset && transaction.asset.data)
-					.map(transaction => ({
-						transactionId: transaction.id,
-						data: Buffer.from(transaction.asset.data, 'utf8'),
-					}));
+				values = transactions.map(transaction => ({
+					transactionId: transaction.id,
+					data: Buffer.from(transaction.asset.data, 'utf8'),
+				}));
```
nazarhussain: Something like this. 

```
values = [];
transactions.forEach(transaction => {
 if(any condition)
values.push({})
});
```

Idea is the transaction type based any logic should be scoped on one place. Later we may extract it to create seperate transaction type classes, or different function for each type.

test/unit/db/repos/dapps.js
```diff
@@ -79,7 +79,8 @@ describe('db', () => {
 				return expect(db.one.firstCall.args[2]).to.be.a('function');
 			});
 
-			it('should resolve to an integer count with number of transactions', function*() {
+			// eslint-disable-next-line mocha/no-skipped-tests
+			it.skip('should resolve to an integer count with number of transactions', function*() {
 				const dapps = yield seeder.seedDapps(db, 1);
 				const result = yield db.dapps.countByTransactionId(dapps[0].id);
```
nazarhussain: Update the use of `db.dapps.countByTransactionId` with `Transaction.isPersisted({type: 5, id: <id>})`.

nazarhussain: Replace all `db.dapps` related usage everywhere in code with `TransactionEntity` calls. Because `dapps`  is just a transaction type.

storage/entities/transaction.js
pablitovicente: Implemented the change directly as it also involved changing other lines but same change as suggested. Resolving.

storage/entities/transaction.js
pablitovicente: Updated based on suggestion.

storage/entities/transaction.js
pablitovicente: Updated based on suggestion.

test/unit/db/repos/dapps.js
pablitovicente: Done. Resolving.

pablitovicente: Done. Resolving.

api/controllers/dapps.js
```diff
@@ -43,25 +44,50 @@ DappsController.getDapps = function(context, next) {
...
-			}
+	// Remove options with null values
+	options = _.pickBy(options, v => !(v === undefined || v === null));
+
+	// We don't want to change the API so we fix the sort field name here
+	options.sort = options.sort.replace('name', 'dapp_name');
 
+	filters =
+		Object.keys(filters).length > 0
+			? Object.keys(filters).map(aFilter => ({
```
MaciejBaj: Can you simplify?

lsilvs: I reckon this should work:

```
const filters = [];

if (params.transactionId.value) {
	filters.push({
		id: params.transactionId.value,
		type: transactionTypes.DAPP,
	});
}

if (params.name.value) {
	filters.push({
		dapp_name: params.name.value,
		type: transactionTypes.DAPP,
	});
}
```

pablitovicente: @lsilvs yes that will do the same as the current code would you guys prefer it this way? @MaciejBaj

modules/dapps.js
```diff
@@ -281,8 +282,13 @@ DApps.prototype.shared = {
...
 shared.getGenesis = function(req, cb, tx) {
-	(tx || library.db).dapps
-		.getGenesis(req.dappid)
+	library.storage.entities.Transaction.get(
+		{
+			id: req.dappid,
+		},
+		{},
+		tx
+	)
```
lsilvs: Maybe it's a good idea to use `limit: 1` and `sort: rowId:asc`

Also, in the line 301 it's using `authorId`. Replace it by `senderId`, pls.

modules/dapps.js
```diff
@@ -81,21 +82,21 @@ class DApps {
...
-			new DApp(scope.db, scope.logger, scope.schema, scope.network)
+			new DApp(scope.logger, scope.schema, scope.network, scope.storage)
 		);
 
 		__private.assetTypes[
 			transactionTypes.IN_TRANSFER
 		] = library.logic.transaction.attachAssetType(
 			transactionTypes.IN_TRANSFER,
-			new InTransfer(scope.db, scope.schema)
+			new InTransfer(scope.db, scope.schema, scope.storage)
```
lsilvs: Why is `InTransfer` still using `scope.db`?

api/controllers/dapps.js
lsilvs: I find it easier to understand.

pablitovicente: Done. Resolving.

storage/entities/transaction.js
```diff
@@ -633,6 +645,12 @@ class Transaction extends BaseEntity {
 			transaction.signatures = transaction.signatures.filter(Boolean);
 		}
 
+		// console.log(transaction);
+		// console.log('#'.repeat(180));
+		// console.log('asset.dapp.type', transaction.asset);
+		// console.log('#'.repeat(180));
+		// console.log('row.asset.dapp.type', row.asset);
+		// console.log('#'.repeat(180));
```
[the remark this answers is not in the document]
pablitovicente: Removed. Resolving.

api/controllers/dapps.js
```diff
@@ -43,25 +44,50 @@ DappsController.getDapps = function(context, next) {
...
+				description: aDapp.asset.dapp.description,
+				tags: aDapp.asset.dapp.tags,
+				link: aDapp.asset.dapp.link,
+				type: aDapp.asset.dapp.type,
+				category: aDapp.asset.dapp.category,
+				icon: aDapp.asset.dapp.icon,
+				transactionId: aDapp.id,
+			}));
+
+			data = _.map(dapps, dapp => {
```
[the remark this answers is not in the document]
pablitovicente: Done. Resolving.

modules/dapps.js
pablitovicente: Good catch. Not using it but need to remove the parameter and check usages in the rest of the code.

modules/dapps.js
pablitovicente: Done and Done. Resolving

modules/dapps.js
pablitovicente: Updated. Resolving.

storage/entities/transaction.js
```diff
@@ -653,6 +671,24 @@ class Transaction extends BaseEntity {
...
+	 * @param {string} params.name
+	 * @param {string} params.link
+	 * @returns {Promise}
+	 * @todo This method should be replaced with calls to storage get() and correct filters
+	 */
+	getExistingDapps(filters = {}, tx) {
+		const sql = `SELECT name, link FROM dapps
+		WHERE (name = '${filters.name}' OR link = '${
+			filters.link
+		}') AND "transactionId" != '${filters.transactionId}'`;
```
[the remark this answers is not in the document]
pablitovicente: Fixed the parameter. Let's refactor to `get()` next time we iterate over storage filters so we can be more concise with them, sound good to you?

api/controllers/dapps.js
```diff
@@ -42,26 +43,55 @@ function DappsController(scope) {
...
-		try {
-			if (err) {
-				return next(err);
-			}
+	// We don't want to change the API so we fix the sort field name here
+	options.sort = options.sort.replace('name', 'dapp_name');
 
+	const filters = [];
+
+	if (params.transactionId.value) {
```
MaciejBaj: Much better 🍾

modules/dapps.js
```diff
@@ -286,7 +286,7 @@ shared.getGenesis = function(req, cb, tx) {
 		{
 			id: req.dappid,
 		},
-		{},
+		{ sort: 'rowId:asc', limit: 1 },
```
lsilvs: I don't think this sorting param will work.
The name of the field is `t_rowId` and I don't think we should use `t_rowId` here as well.

I've just noticed that transactions entity is already using `t_rowId:asc` as default.
The sort param can be removed.

pablitovicente: Done. Resolving. @lsilvs please check PR again.

logic/dapp.js
```diff
@@ -43,12 +43,12 @@ __private.unconfirmedAscii = {};
  * @todo Add description for the params
  */
 class DApp {
-	constructor(db, logger, schema, network) {
+	constructor(logger, schema, network, storage) {
```
nazarhussain: Why not just replace the `db` with `storage`. Why you have to change the order of params here?

logic/dapp.js
```diff
@@ -186,12 +186,14 @@ DApp.prototype.verify = function(transaction, sender, cb, tx) {
...
-	return (tx || library.db).dapps
-		.getExisting({
+	return library.storage.entities.Transaction.getExistingDapps(
+		{
 			name: transaction.asset.dapp.name,
 			link: transaction.asset.dapp.link || null,
-			transactionId: transaction.id,
-		})
+			transactionId_ne: transaction.id,
+		},
```
nazarhussain: You don't need this method it can be replaced by: 

```
library.storage.entities.Transaction.isPersisted([{name: ransaction.asset.dapp.name, transactionId_ne: transactio.id}, {link: ransaction.asset.dapp.link, transactionId_ne: transactio.id}])
```

nazarhussain: BTW you are passing `transactionId_ne` here and using `transactionId` in the function.

storage/entities/transaction.js
```diff
@@ -653,6 +664,24 @@ class Transaction extends BaseEntity {
...
+	 * @returns {Promise}
+	 * @todo This method should be replaced with calls to storage get() and correct filters
+	 */
+	getExistingDapps(filters = {}, tx) {
+		const sql = `SELECT name, link FROM dapps
+		WHERE (name = '${filters.name}' OR link = '${
+			filters.link
+		}') AND "transactionId" != '${filters.transactionId_ne}'`;
+		return this.adapter.execute(sql, {}, {}, tx);
+	}
```
nazarhussain: You don't need this as mentioned in above comment. This logic can be replaced by `isPersisted` with composite filters.

storage/entities/transaction.js
```diff
@@ -653,6 +664,24 @@ class Transaction extends BaseEntity {
...
 		return filters;
 	}
+
+	/**
+	 * Checks if a dapp exists.
+	 *
+	 * @param {Object} params
+	 * @param {string} params.transactionId
+	 * @param {string} params.name
+	 * @param {string} params.link
```
nazarhussain: The objet is `filters` not the `params`.

storage/entities/transaction.js
```diff
@@ -653,6 +664,24 @@ class Transaction extends BaseEntity {
...
+	 * @returns {Promise}
+	 * @todo This method should be replaced with calls to storage get() and correct filters
+	 */
+	getExistingDapps(filters = {}, tx) {
+		const sql = `SELECT name, link FROM dapps
+		WHERE (name = '${filters.name}' OR link = '${
+			filters.link
+		}') AND "transactionId" != '${filters.transactionId_ne}'`;
+		return this.adapter.execute(sql, {}, {}, tx);
```
nazarhussain: This will cause SQL injection because you are not sanitizing the parameters and directly concatenating in the query. The right way to do this is: 

```
const sql = 'SELECT name, link, FROM dapps WHERE name = ${name} OR link = ${link} AND transactionId != ${transactionId}'
return this.adapter.execute(sql, filters, {}, tx);
```

logic/in_transfer.js
```diff
@@ -31,15 +32,15 @@ let shared;
...
  * @requires helpers/milestones
  * @requires helpers/slots
- * @param {Database} db
  * @param {ZSchema} schema
+ * @param {Storage} storage
  * @todo Add description for the params
  */
 class InTransfer {
-	constructor(db, schema) {
+	constructor(schema, storage) {
```
nazarhussain: Can't understand why you change the order of parameters and not just replaced `db` with `storage`.

logic/out_transfer.js
```diff
@@ -31,15 +32,15 @@ __private.unconfirmedOutTansfers = {};
...
  * @see Parent: {@link logic}
- * @param {Database} db
  * @param {ZSchema} schema
  * @param {Object} logger
+ * @param {Storage} storage
  * @todo Add description for the params
  */
 class OutTransfer {
-	constructor(db, schema, logger) {
+	constructor(schema, logger, storage) {
```
nazarhussain: Same here.

logic/transaction.js
```diff
@@ -47,7 +47,7 @@ const __private = {};
  * @todo Add description for the params
  */
 class Transaction {
-	constructor(db, ed, schema, genesisBlock, account, logger, cb) {
+	constructor(db, storage, ed, schema, genesisBlock, account, logger, cb) {
```
nazarhussain: Why the `db` is left out here? Is it still in use?

modules/dapps.js
```diff
@@ -81,21 +82,21 @@ class DApps {
 			transactionTypes.DAPP
 		] = library.logic.transaction.attachAssetType(
 			transactionTypes.DAPP,
-			new DApp(scope.db, scope.logger, scope.schema, scope.network)
+			new DApp(scope.logger, scope.schema, scope.network, scope.storage)
```
nazarhussain: Due to change in order of parameters, it seems in git diff as you changed all parameters, not just one.

logic/dapp.js
pablitovicente: Db was removed after whole implementation. Will re-order and fix tests.

logic/dapp.js
pablitovicente: It uses a complex where query that storage only supports by building the filter by repeating filters. I would say for the sake of getting it merged to keep it as is and once we iterate over the filters for storage we remove this method. What you think?

storage/entities/transaction.js
pablitovicente: OK as this seems that needs to be replaced I will just replace it but it's a similar case for the controller.

api/controllers/dapps.js
```diff
@@ -42,26 +43,55 @@ function DappsController(scope) {
...
+			type: transactionTypes.DAPP,
+		});
+	}
+
+	if (filters.length === 0) {
+		filters.push({ type: transactionTypes.DAPP });
+	}
+
+	storage.entities.Transaction.get(filters, options)
+		.then(data => {
```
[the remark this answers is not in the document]
pablitovicente: Done.

logic/dapp.js
pablitovicente: Done.

logic/in_transfer.js
pablitovicente: Same as before replaced storage and then removed. Will change.

logic/out_transfer.js
pablitovicente: Same as before replaced storage and then removed. Will change.

logic/dapp.js
pablitovicente: It's using the returned values in the logic I don't think it's a direct replacement with isPersisted that will return true|false

pablitovicente: This have been already fixed by same comment by Lucas. but will replace whole method so resolving.

logic/transaction.js
pablitovicente: Not in use. Removed and updated related modules.

storage/entities/transaction.js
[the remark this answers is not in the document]
pablitovicente: Ended up replacing with call to `get()` resolving.

logic/dapp.js
pablitovicente: Changed to `get()` with filters as the rows being used for calling cb with different errors. We could refactor the whole method in the future but I think that part would be extending the scope of this PR.

storage/entities/transaction.js
pablitovicente: Removed method and logic is now using `get()`

storage/entities/transaction.js
pablitovicente: Method was removed

PR #2725, opened 2019-01-09
Restore stability of peers blacklists and disconnections in Network Tests - Closes #2670

### LINKED ISSUE #2670: Restore stability of peers blacklists and disconnections in Network Tests

diego-G:
### How to test it?

`npm test -- mocha:default:network`

### Review checklist

* The PR resolves #2670
* Linting passes
* Tests pass

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -23,8 +23,8 @@ module.exports = function(grunt) {
 				cmd(tagFilter, suite, section) {
 					if (suite === 'network') {
 						let filter = '';
-						if (tagFilter === 'network') {
-							filter = '--grep @network';
+						if (tagFilter === 'p2p') {
+							filter = '--grep @p2p';
 						} else if (tagFilter === 'propagation') {
 							filter = '--grep @propagation';
 						}
--- a/test/network/index.js
+++ b/test/network/index.js
@@ -51,7 +51,7 @@ describe(`Start a network of ${TOTAL_PEERS} nodes with address "127.0.0.1", WS p
 		return network.killNetwork();
 	});
 
-	describe('launching test scenarios', () => {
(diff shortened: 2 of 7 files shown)


diego-G:
@MaciejBaj every existing test is meaningful. Remove them or skip them is not the solution. The idea is to reduce the strictness on those problematic tests but, as less as possible while keeping a good balance in between test execution timing and stability. As you can realise the [Jenkins trend build](https://jenkins.lisk.io/job/lisk-core-network/view/change-requests/job/PR-2725/buildTimeTrend), I have found a good stability. It is reasonable and stable to be merged.

I want to state again this test suite is very useful for developers. It just needs more love. We should plan to seriously improve it for the future. Several points to perform are:
- Execute every test independently from each other : i.e, reinitiated the network completely after every scenario
- Execute scenarios in parallel to speed up the process : right now they are being executed in sequence over the same network. We might create several networks, for instance 1 per well-defined scenario. 
- Review the design from scratch and find more and better use cases
- When a test failure occurs, gathering more data from the network status to reproduce the case and debug it

PR #2730, opened 2019-01-10
Refactor `modules.blocks[.shared]` namespace - Closes #2625

### LINKED ISSUE #2625: Refactor `modules.blocks[.shared]` namespace
Need to refactor all `modules.blocks[.shared]` interfaces and try to replace with existing CRUD interfaces of `Block` entity.

lsilvs:
### What was the problem?
Blocks API was using blocks modules to retrieve blocks data.

### How did I fix it?
I update Blocks API to access the storage module directly removing blocks module dependency

### How to test it?
Regular test suite

### Review checklist

* The PR resolves #2625
* Linting passes
* Tests pass


api/controllers/blocks.js
```diff
@@ -87,4 +105,104 @@ BlocksController.getBlocks = function(context, next) {
...
+	) {
+		return setImmediate(
+			cb,
+			new ApiError('Invalid sort field', apiCodes.INTERNAL_SERVER_ERROR)
+		);
+	}
+
+	return library.storage.entities.Block.get(filters, options)
+		.then(rows => {
+			const blocks = [];
```
MaciejBaj:
```suggestion
			// FIXME: Can have poor performance because it performs SHA256 hash calculation for each block
			return setImmediate(cb, null, rows.map(library.logic.block.storageRead));
```

api/controllers/blocks.js
```diff
@@ -56,7 +74,7 @@ BlocksController.getBlocks = function(context, next) {
 	// Remove filters with null values
 	filters = _.pickBy(filters, v => !(v === undefined || v === null));
 
-	return modules.blocks.shared.getBlocks(_.clone(filters), (err, data) => {
+	return _list(_.clone(filters), (err, data) => {
```
MaciejBaj: That's the only place the function is used, it's easier to keep the logic of returning blocks in one place, so my suggestion is not to extract the _list logic into a separate function.

api/controllers/blocks.js
```diff
@@ -31,11 +32,22 @@ let library;
  */
 function BlocksController(scope) {
 	library = {
-		db: scope.db,
 		storage: scope.storage,
 		logic: scope.logic,
 		logger: scope.logger,
 	};
+
```
diego-G: I see that it is present in `db.blocks` yet. Is there any plan to remove it or move it to storage?
Why cannot we get it from anywhere else like storage for instance?

api/controllers/blocks.js
```diff
@@ -87,4 +105,104 @@ BlocksController.getBlocks = function(context, next) {
...
+ * @param {number} filter.reward - Block reward
+ * @param {number} filter.limit - Limit of blocks to retrieve, default: 100, max: 100
+ * @param {number} filter.offset - Offset from where to start
+ * @param {string} filter.sort - Sort order, default: height:desc
+ * @param {function} cb - Callback function
+ * @returns {function} cb - Callback function from params (through setImmediate)
+ * @returns {Object} cb.err - Error if occurred
+ * @returns {Object} cb.data - List of normalized blocks
+ */
+function _list(filter, cb) {
```
diego-G: Why do you name it with underscore at the beginning? 

In any case, I'd move to the top right after `BlocksController` function.

lsilvs: I used the underscore because it is a "private" method.
But I'm refactoring this as requested by @MaciejBaj already ;)

api/controllers/blocks.js
lsilvs: This will require a big change in the unit tests. I will keep as it is for now.

api/controllers/blocks.js
```diff
@@ -87,4 +105,104 @@ BlocksController.getBlocks = function(context, next) {
...
+		totalAmount: filter.totalAmount,
+		totalFee: filter.totalFee,
+		reward: filter.reward,
+	};
+
+	Object.keys(filters).forEach(key => {
+		if (!filters[key]) {
+			delete filters[key];
+		}
+	});
```
nazarhussain: This will be much simpler: 

```
filters = _.pickBy(filters, v => !(v === undefined || v === null));
```

api/controllers/blocks.js
```diff
@@ -87,4 +105,104 @@ BlocksController.getBlocks = function(context, next) {
...
+
+	if (!filter.limit) {
+		options.limit = 100;
+	} else {
+		options.limit = Math.abs(filter.limit);
+	}
```
nazarhussain:
```
options.limit = filter.limit ? Math.abs(filter.limit) : 100;
```

api/controllers/blocks.js
```diff
@@ -87,4 +105,104 @@ BlocksController.getBlocks = function(context, next) {
...
+
+	if (!filter.offset) {
+		options.offset = 0;
+	} else {
+		options.offset = Math.abs(filter.offset);
+	}
```
nazarhussain:
```
options.offset = filter.offset ? Math.abs(filter.offset) : 0;
```

api/controllers/blocks.js
MaciejBaj: I think it's a matter of taste, but if you do that @lsilvs you could also move that line up, where the `filters` object initialization is.

api/controllers/blocks.js
lsilvs: Done. I've used `_.omitBy` to avoid negated OR confusion.

PR #2742, opened 2019-01-15
Remove db.delegates repository - Closes #2735

### LINKED ISSUE #2735: Remove db.delegates repository
Remove `db.delegates` and move custom interfaces to `storage.entities.Account`

nazarhussain:
### What was the problem?
Need to remove `db.delegates` namespace. 

### How did I fix it?
Moved custom interfaces with `storage.Account` and removed the unused code. 

### How to test it?
Run all tests.

### Review checklist

* The PR resolves #2735
* Linting passes
* Tests pass


test/unit/storage/entities/account.js
```diff
@@ -1113,4 +1117,136 @@ describe('Account', () => {
...
+		it('should return zero if there are delegates but no duplicates', async () => {
+			const result = await AccountEntity.countDuplicatedDelegates();
+
+			expect(result).to.be.eql(0);
+		});
+
+		it('should return integer value of duplicate delegates', async () => {
+			const block = seeder.getLastBlock();
+
+			const trs1 = new transactionsFixtures.Transaction({
```
MaciejBaj: It's redundancy here - `trs1` and `trs2` are getting inserted, but their effect is not checked - `countDuplicatedDelegates()` returns 2 only because of the direct insert of 2 duplicated delegate names in the next line:
```
	await adapter.execute(
					'INSERT INTO delegates ("transactionId", "username") VALUES ($1, $2)',
					[delegate.transactionId, username]
				);
```
Insert of `trs1` and `trs2` can be skipped or they both should be inserted with duplicated "username" passed in transaction.asset.username. 

Btw constraint on `delegates` table which should prevent of inserting duplicated username is preventing only from inserting the same delegate with the same transaction id - 

`ALTER TABLE delegates ADD CONSTRAINT delegates_unique UNIQUE ("username", "transactionId");`

Do you think a separate constraint should be added to prevent duplicated delegates cases? Then `countDuplicatedDelegates()` would not be needed.

nazarhussain: To add a duplicated delegate, we need a valid `transactionId`, so first we process tow valid transactions, and then based on their id we added a duplicate name entry. That's exactly what data sate we need to test the SQL written  for `countDuplicateDelegates`. And that was actually migrated from existing test case. 

Yes the constraint can be added to avoid duplicate. But I am not sure about historical data, I suspect there was some issue in data so why this check was added. 

In any case, would not suggest to make this behavioral change int he `modules/loader` at this moment.

nazarhussain: Why I didn't processed `trs1` and trs2` to be duplicate of each other, to check if our query can work perfectly to find multiple duplicate count.

MaciejBaj: I have opened the issue for that https://github.com/LiskHQ/lisk/issues/2746.

MaciejBaj:
@nazarhussain please solve the conflicts.

PR #2751, opened 2019-01-17
Use consistent informative versioning scheme - Closes #2594

### LINKED ISSUE #2594: Use a consistent and informative versioning scheme
### Expected behavior
Follow the [implementation guidelines of LIP](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0007.md#implementation):

- Rely on `protocolVersion` field to determine compatibility between two nodes.
- Keep the backwards compatibility by preserving the check relying on `minVersion`.
- Start from `protocolVersion` equal `1.0`.

### Actual behavior
In the handshake process, the "minVersion" field is currently used to determine compatibility between two nodes.

`*`

limiaspasdaniel:
### What was the problem?
There was no separation of concerns between protocol changes and implementation changes in the current versioning scheme.

For further information, please refer to [LIP0007](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0007.md).
### How did I fix it?
By implementing **LIP007**. 

Now nodes are aware of the protocol version other peers are running on and therefore they perform compatibility checks against this version.

### How to test it?

- Run unit tests `npm test -- mocha:extensive:unit`
- Run functional tests `npm test --mocha:extensive:functional`
- Run network tests `npm test -- mocha:default:network`

### Review checklist

* The PR resolves #2594 
* Linting passes
* Tests pass


api/ws/rpc/connect.js
```diff
@@ -48,6 +48,9 @@ const connectSteps = {
 		if (systemHeaders.version) {
 			queryParams.version = systemHeaders.version;
 		}
+		if (systemHeaders.protocolVersion) {
```
diego-G: Add unit test.

db/sql/migrations/updates/20190111111557_add_protocolVersion_column_to_peers.sql
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+
+/*
+  DESCRIPTION: Add protocolVersion Column to Peers.
+
+  PARAMETERS: None
+*/
+
+ALTER TABLE "peers" ADD COLUMN IF NOT EXISTS "protocolVersion" VARCHAR(20);
```
diego-G: Do we need default value?

helpers/z_schema.js
```diff
@@ -288,6 +288,14 @@ const liskFormats = {
 			)
 		);
 	},
+	/**
+	 * Validation for the protocol version format.
+	 * @param str
+	 * @returns {boolean}
+	 */
+	protocolVersion(str) {
```
diego-G: Add unit test.

modules/peers.js
```diff
@@ -270,31 +271,51 @@ __private.updatePeerStatus = function(err, status, peer) {
...
-				peer.string
-			}, version: ${status.version}`
-		);
-		library.logic.peers.remove(peer);
 	} else {
-		let state;
-		// Ban peer if it is presented in the array of black listed peers
-		if (__private.isBlacklisted(peer.ip)) {
-			state = Peer.STATE.BANNED;
+		let compatible = false;
```
diego-G: No changes in related unit test?

modules/system.js
```diff
@@ -215,6 +217,15 @@ System.prototype.getMinVersion = function() {
 	return __private.minVersion;
 };
 
+/**
+ * Gets private variable `protocolVersion`
+ *
+ * @returns {string}
+ */
+System.prototype.getProtocolVersion = function() {
```
diego-G: Add unit test.

modules/system.js
```diff
@@ -237,6 +248,19 @@ System.prototype.versionCompatible = function(version) {
...
 };
 
+/**
+ * Checks protocol version compatibility from input param against
+ * private values.
+ *
+ * @param protocolVersion
+ * @returns {boolean}
+ */
+System.prototype.protocolVersionCompatible = function(protocolVersion) {
```
diego-G: Add unit test.

test/functional/http/get/peers.js
```diff
@@ -69,6 +69,11 @@ describe('GET /peers', () => {
 			invalid: ['9999.999.999a'],
 			checkResponse: true,
 		},
+		protocolVersion: {
+			valid: ['999.999'],
+			invalid: ['9999.9999', '-1.-1', 'a.b'],
```
diego-G: What about null, undefined or empty string?

test/network/scenarios/p2p/peer_protocol_version.js
```diff
@@ -0,0 +1,89 @@
...
+				await fs_writeFile(
+					`${__dirname}/../../pm2.network.json`,
+					JSON.stringify(pm2NetworkConfig, null, 4)
+				);
+
+				await network.reloadNode('node_0', true, true);
+				await network.waitForBlocksOnNode('node_1', 4);
+			});
+
+			after(async () => {
```
diego-G: Please, add two more test to verify everything comes to normality after correcting back the protocol version just modified. You can see an example on the blacklisting scenario.

I would also use the node number 9 instead of number 0 because it's the only one non forging node. Otherwise you need to enable forging when you reload the node in the after block.

limiaspasdaniel: Okay, will do.
Didn't know about the special case with node number 0, thanks for pointing it out.

test/functional/http/get/peers.js
limiaspasdaniel: So as far as I've seen this  validation is also performed on handshake time for inbound connections to check for valid WS headers, not only on the HTTP API. So we should allow null, undefined or empty string on `protocolVersion` to keep compatibility with older nodes. Therefore null, undefined or empty string values shouldn't be invalid. If one of those values is received on an updated node, it will be aware of it and therefore will continue to check for compatibility with the `version`.

db/sql/migrations/updates/20190111111557_add_protocolVersion_column_to_peers.sql
limiaspasdaniel: Outdated and compatible nodes won't have protocolVersion so it will be empty for them. Updated and compatible nodes will have their own, so I think no need for default value. What do you think?
Also @MaciejBaj

db/sql/migrations/updates/20190111111557_add_protocolVersion_column_to_peers.sql
```diff
@@ -0,0 +1,22 @@
+/*
```
lsilvs: Note that the migrations folder changed since https://github.com/LiskHQ/lisk/pull/2745

Please, move this file to `storage/sql/migrations/updates` and make sure your migration name has a date greater than the last one in this folder.

api/ws/rpc/connect.js
limiaspasdaniel: Done

helpers/z_schema.js
limiaspasdaniel: Done

modules/peers.js
limiaspasdaniel: Done

modules/system.js
limiaspasdaniel: Done

modules/system.js
limiaspasdaniel: Done

test/network/scenarios/p2p/peer_protocol_version.js
limiaspasdaniel: Implemented another test case to check whether a node with correct protocol version joins back the network with no issues. Please @diego-G let me know what you think.

db/sql/migrations/updates/20190111111557_add_protocolVersion_column_to_peers.sql
limiaspasdaniel: Done, thanks :)

modules/system.js
```diff
@@ -236,6 +247,19 @@ System.prototype.versionCompatible = function(version) {
...
+	const hard = parseInt(protocolVersion[0]);
+	const privateHard = parseInt(__private.protocolVersion[0]);
+	return privateHard === hard && hard >= 1;
```
MaciejBaj:
```suggestion
	return __private.protocolVersion[0] === protocolVersion[0] && parseInt(protocolVersion[0]) >= 1;
```

limiaspasdaniel: Gonna go for myHard and peerHard :D

limiaspasdaniel: Changed. Resolving. @MaciejBaj

PR #2758, opened 2019-01-21
Update the loading of SQL files to avoid warning messages - Closes #2737

### LINKED ISSUE #2737: Update the loading of SQL files to avoid warning messages
Currently the SQL files loading in entity constructors are causing warning messages if we extend any entity. Because due to nature of `pgp` library we can't load same SQL file twice. Need to fix this problem by creating `static` interface for the `pgp_adapter` to load the files outside the constructor.

lsilvs:
### What was the problem?
Storage entities was loading SQL files in the constructor and if the same entity was instantiated with the same adapter, `pg-promise` would raise a warning message:

```
WARNING: Creating a duplicate QueryFile object for the same file
```

### How did I fix it?
I've implemented a method called `loadSQLFiles` in `BaseEntity` that receives the entity label and SQL files array and only try to load the file if it hasn't been loaded yet.

### How to test it?
Start the server and check the console logs.
The above warning message should not be present anymore.

### Review checklist

* The PR resolves #2737
* Linting passes
* Tests pass

### DIFF
--- a/test/unit/storage/entities/base_entity.js
+++ b/test/unit/storage/entities/base_entity.js
@@ -1,0 +1,132 @@
+/* eslint-disable mocha/no-pending-tests */
+/*
+ * Copyright © 2018 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+'use strict';
+
+const { BaseEntity } = require('../../../../storage/entities');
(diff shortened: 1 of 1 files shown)


storage/entities/base_entity.js
```diff
@@ -204,6 +204,24 @@ class BaseEntity {
...
+	loadSQLFiles(entityLabel, sqlFiles) {
+		this.adapter.SQLs = this.adapter.SQLs || {};
+		this.adapter.SQLs[entityLabel] = this.adapter.SQLs[entityLabel] || {};
+		Object.keys(sqlFiles).forEach(key => {
+			if (!this.adapter.SQLs[entityLabel][key]) {
+				this.adapter.SQLs[entityLabel][key] = this.adapter.loadSQLFile(sqlFiles[key]);
+			}
+		});
+		return this.adapter.SQLs[entityLabel];
+	}
```
nazarhussain: With the following change, this function will become integral component for the storage layer. Please add some unit tests for it.

PR #2785, opened 2019-01-24
Implement pending tests of DB Entities - Closes #2688

### LINKED ISSUE #2688: Implement pending tests of DB Entities
### Expected behavior

The skipped tests of Database Entities (`test/unit/storage/entities`) should be implemented.

### Actual behavior

The following tests of DB Entities have been recognised as less important, therefore skipped for now:
-  `test/unit/storage/entities/peer.js` (8 out of 50 tests are pending)

`>1.5.0`

pablitovicente:
### What was the problem?

Some tests where not implemented for some entities

### How did I fix it?

- By implementing the missing tests 

Following tests are going to be worked in separate issues:
- all should skip if any invalid attribute is provided
- all should return matching result for provided filter
- should not change any of the provided parameter
- Pending Migration entity tests
- Pending Account  entity tests: "should sync rank attribute of all delegates based on their vote value and public key" and "should not throw error if there is no delegate available"

### How to test it?

- Build should be green

### Review checklist

* The PR resolves #2688
* Linting passes
* Tests pass

### DIFF
--- a/test/unit/storage/entities/account.js
+++ b/test/unit/storage/entities/account.js
@@ -17,7 +17,10 @@
 
 const randomstring = require('randomstring');
 const { BaseEntity, Account } = require('../../../../storage/entities');
-const { NonSupportedFilterTypeError, NonSupportedOptionError } = require('../../../../storage/errors');
+const {
+	NonSupportedFilterTypeError,
+	NonSupportedOptionError,
+} = require('../../../../storage/errors');
 const storageSandbox = require('../../../common/storage_sandbox');
 const seeder = require('../../../common/storage_seed');
 const accountFixtures = require('../../../fixtures').accounts;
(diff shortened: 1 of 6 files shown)


test/unit/storage/entities/account.js
```diff
@@ -460,9 +795,69 @@ describe('Account', () => {
...
 
+	describe('_getResults()', () => {
+		const accounts = [
+			new accountFixtures.Account(),
+			new accountFixtures.Account(),
+			new accountFixtures.Account(),
+		];
+
+		before(async () => {
+			await AccountEntity.create(accounts);
```
nazarhussain: Is it possible that we only use `beforeEach` in every test? Specially when we are changing data and test relies on that data.

pablitovicente: Done.

PR #2786, opened 2019-01-25
Update cache module to be modified as component - Closes #2579 & #2802

### LINKED ISSUE #2579: Update cache component to be modified as component pattern
In new lisk core architecture the `components` are exposed as one single method which actions `json` options and return the object. So we need to update the `modules/cache.js` to be refactored this way.

diego-G:
### How to test it?

```
mocha test/unit/components/cache.js

mocha test/unit/modules/blocks.js
mocha test/unit/modules/transactions.js
mocha test/unit/modules/rounds.js
mocha test/unit/modules/blocks/chain.js

mocha test/functional/http/get/cache.js
```

### Review checklist

* The PR resolves #2579 & #2802
* Linting passes
* Tests pass


modules/transactions.js
```diff
@@ -649,17 +657,25 @@ Transactions.prototype.shared = {
 		async.waterfall(
 			[
 				function getConfirmedCountFromCache(waterCb) {
-					modules.cache.getJsonForKey(
-						modules.cache.KEYS.transactionCount,
-						(err, data) => {
-							if (err) {
-								// If some issue in cache we will fallback to database
-								return setImmediate(waterCb, null, null);
+					if (library.config.cacheEnabled) {
```
lsilvs: I would suggest to do `if (components.cache)` and initialise `components.cache` inside `app.js` if `library.config.cacheEnabled`

config/default/constants.js
```diff
@@ -123,4 +123,12 @@ module.exports = {
 	TOTAL_AMOUNT: '10000000000000000',
 	UNCONFIRMED_TRANSACTION_TIMEOUT: 10800, // 1080 blocks
 	EXPIRY_INTERVAL: 30000,
+	CACHE: {
```
lsilvs: I reckon this constants should be part of the cache component.
I suggest to create a folder for each component inside `components` and move this constants there. Also it could be exported along with the component.

Something like:

```
/components
    cache/
        constants.js
        cache.js
        index.js
    index.js
```

components/index.js
```diff
@@ -0,0 +1,26 @@
...
+const Cache = require('./cache');
+
+module.exports = function createCacheComponent(options, logger) {
+	// delete password key if it's value is null
+	const cacheConfigParam = Object.assign({}, options);
+	if (cacheConfigParam.password === null) {
+		delete cacheConfigParam.password;
+	}
+	return new Cache(cacheConfigParam, logger);
+};
```
lsilvs: This file should just export the components itself so you could use like:

`const { cache } = require('./components')`

components/cache.js
```diff
@@ -0,0 +1,427 @@
...
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+'use strict';
+
+const async = require('async');
+const redis = require('redis');
+const transactionTypes = require('../helpers/transaction_types');
+const { CACHE } = require('../config/default/constants');
```
lsilvs: As mentioned before, I reckon it should be:

`const CONSTANTS = require('./constants');`

components/cache/index.js
```diff
@@ -0,0 +1,427 @@
...
+		this.options = options;
+		this.logger = logger;
+		this.cacheReady = false;
+	}
+
+	connect(cb) {
+		this.client = redis.createClient(this.options);
+		this.client.once('error', this._onConnectionError.bind(this, () => cb()));
+		this.client.once('ready', this._onReady.bind(this, () => cb()));
+	}
```
nazarhussain: As we are refactoring, we are trying to avoid callback patterns. So suggested to return Promise from this function.

components/cache/index.js
```diff
@@ -0,0 +1,427 @@
...
+					[
+						'Cache - Keys with pattern:',
+						pattern,
+						'cleared from cache on new Round',
+					].join(' ')
+				);
+			}
+			return cb(err);
+		});
+	}
```
nazarhussain: This logic should be moved away from the shared component. Its strictly related to chain module and should come under round namespace in it.

components/cache/index.js
```diff
@@ -0,0 +1,427 @@
...
+								);
+							}
+							return setImmediate(reflectCb, deleteJsonForKeyErr);
+						}
+					);
+				}),
+			],
+			() => setImmediate(cb, null) // Don't propagate cache error to continue normal operations
+		);
+	}
```
nazarhussain: Same as above.

components/cache/index.js
```diff
@@ -0,0 +1,427 @@
...
+	onSyncStarted() {
+		this.cacheReady = false;
+	}
+
+	/**
+	 * Enables changes in cache after syncing finished.
+	 */
+	onSyncFinished() {
+		this.cacheReady = true;
+	}
```
nazarhussain: In principal no events should be part of the component. Consider component to be just a shared library, so activities on events should move to the their respective namespaces. 

Consider this in relation, what if we use this cache component in the `http-api`, there are no such events there, it just need a wrapper to get and set the cache. And cache keys to be there, so both modules can reply on same keys.

components/cache/constants.js
```diff
@@ -0,0 +1,38 @@
...
+ * @memberof config
+ * @see Parent: {@link component/cache}
+ * @property {number} CACHE.KEYS.transactionCount
+ * @property {number} CACHE.KEYS.blocksApi
+ * @property {number} CACHE.KEYS.transactionsApi
+ * @property {number} CACHE.KEYS.delegatesApi
+ * @todo Add description for the namespace and the properties.
+ */
+module.exports = {
+	CACHE: {
```
shuse2: To be discussed, but i think it would be better to make it
```
const CACHE_KEYS_TRANSACTION_COUNT = 'transactionCount';
const CACHE_KEYS_API_BLOCKS = '/api/blocks/*';
```

test/common/components_loader.js
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+'use strict';
+
+const createCache = require('../../components');
+const Logger = require('../../logger');
+
+const componentsLoader = new function() {
```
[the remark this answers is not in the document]
diego-G: File removed.

modules/blocks.js
```diff
@@ -240,14 +241,15 @@ Blocks.prototype.onBind = function(scope) {
...
-	].map(pattern => callback => modules.cache.clearCacheFor(pattern, callback));
+	if (components.cache) {
+		const tasks = [CACHE.KEYS.blocksApi, CACHE.KEYS.transactionsApi].map(
+			pattern => callback => components.cache.removeByPattern(pattern, callback)
+		);
 
-	async.parallel(async.reflectAll(tasks), () =>
-		library.network.io.sockets.emit('blocks/change', block)
-	);
+		async.parallel(async.reflectAll(tasks), () =>
```
pablitovicente: It seem previously this was always being called now it seems it's tied to `components.cache` to be truthy else is not going to be called. If `components.cache` is always true then we don't need the if or maybe this change should keep the `async.parallel...` call outside the if scope?

diego-G: Good catch.

modules/loader.js
```diff
@@ -934,7 +935,9 @@ __private.loadBlocksFromNetwork = function(cb) {
  */
 __private.sync = function(cb) {
 	library.logger.info('Starting sync');
-	library.bus.message('syncStarted');
```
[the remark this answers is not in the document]
diego-G: It has been replaced by the setter `setReady`.

components/cache/index.js
```diff
@@ -0,0 +1,226 @@
...
+	constructor(options, logger) {
+		this.options = options;
+		this.logger = logger;
+		this.cacheReady = false;
+	}
+
+	async bootstrap() {
+		this.client = redis.createClient(this.options);
+		this.client.once('error', this._onConnectionError.bind(this));
+		this.client.once('ready', this._onReady.bind(this));
```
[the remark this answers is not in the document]
diego-G: It depends on how we want to react to any cache failover. Right now if the cache server shuts down, our application handle the error without stopping. It also recovers the connection if the cache server restarts.

However there is an error when set the CACHE_ENABLED to true. Our application finishes with an FTL error.

helpers/newrelic_lisk.js
```diff
@@ -27,6 +27,17 @@ newrelicLisk.instrumentBackgroundJobs();
 
 // callBackMethods array only support one level of nesting
 const modulesToInstrument = {
+	'./componentes/cache.js': {
+		identifier: 'components.cache',
+		callbackMethods: [
+			'onNewBlock',
+			'onFinishRound',
```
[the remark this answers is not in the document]
diego-G: A TOFIX line was added to review if New Relic is operational.

components/cache/index.js
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	/**
+	 * Enables or disables cache client
+	 *
+	 * @param {boolean} availability
+	 */
+	setReady(availability) {
+		this.client.ready = availability;
+	}
```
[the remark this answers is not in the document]
diego-G: Done.

modules/blocks.js
```diff
@@ -239,15 +239,36 @@ Blocks.prototype.onBind = function(scope) {
...
+			['Cache - onNewBlock', '| Status:', components.cache.isReady()].join(' ')
+		);
+		const keys = [
+			CACHE_CONSTANTS.KEYS_BLOCKS,
+			CACHE_CONSTANTS.KEYS_TRANSACTIONS,
+		];
+		try {
+			// eslint-disable-next-line no-restricted-syntax
+			for await (const key of keys) {
+				await components.cache.removeByPattern(key);
```
lsilvs: This can be done in parallel by using `await Promise.all`

app.js
```diff
@@ -487,31 +489,21 @@ d.run(() => {
...
 				}
-				const redisConnector = new RedisConnector(
-					config.cacheEnabled,
-					cacheConfigParam,
-					logger
-				);
-				redisConnector.connect((redisConnectError, redisClient) =>
-					cb(null, { cacheEnabled: config.cacheEnabled, client: redisClient })
-				);
+				const cache = createCacheComponent(config.cache, logger);
```
MaciejBaj: Why there is there a need for a separate factory`createCacheComponent` and the `bootstrap` method? They will be called always together, meaning bootstrap can be merged into the factory.

components/cache/index.js
```diff
@@ -0,0 +1,32 @@
...
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+'use strict';
+
+const CACHE_CONSTANTS = require('./constants');
+const Cache = require('./cache');
+
+function createCacheComponent(options, logger) {
```
shuse2: should this be in `cache.js`?

components/cache/index.js
```diff
@@ -0,0 +1,32 @@
...
+	// delete password key if it's value is null
+	const cacheConfigParam = Object.assign({}, options);
+	if (cacheConfigParam.password === null) {
+		delete cacheConfigParam.password;
+	}
+	return new Cache(cacheConfigParam, logger);
+}
+
+module.exports = {
+	CACHE_CONSTANTS,
```
shuse2: shouldn't this be destrutured like
```
...CACHE_CONSTANTS,
```

components/cache/index.js
```diff
@@ -0,0 +1,32 @@
...
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+'use strict';
+
+const CACHE_CONSTANTS = require('./constants');
```
shuse2:
```suggestion
const constants = require('./constants');
```

components/cache/index.js
```diff
@@ -0,0 +1,32 @@
...
+const Cache = require('./cache');
+
+function createCacheComponent(options, logger) {
+	// delete password key if it's value is null
+	const cacheConfigParam = Object.assign({}, options);
```
shuse2:
```suggestion
	const cacheConfigParam = { ...options };
```

components/cache/index.js
diego-G: @nazarhussain please, answer this.

components/cache/index.js
diego-G: It doesn't work that way.

components/cache/index.js
diego-G: I prefer to keep `CACHE_CONSTANTS`.

app.js
```diff
@@ -487,31 +489,21 @@ d.run(() => {
...
+				return cache.bootstrap().then(err => {
```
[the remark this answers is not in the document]
diego-G: Done.

app.js
nazarhussain: `factorycreateCacheComponent` is a synchronous call to create the object. `bootstrap` is to establish the physical connection with redis server. Its intended to have it this way, to bootstrap later in the stack and create object in start of the file.

test/unit/modules/blocks/utils.js
```diff
@@ -350,24 +350,26 @@ describe('blocks/utils', () => {
...
-						.returns({ id: '9314232245035524467', height: 1 }),
-				},
-				utils: {
-					readDbRows: blocksUtilsModule.readDbRows,
-					readStorageRows: blocksUtilsModule.readStorageRows,
+		bindingsStub = {
+			modules: {
+				blocks: {
+					lastBlock: {
+						get: sinonSandbox
```
pablitovicente: Something missing here maybe? like `sinonSandbox.stub()` ?

Issue #2763, opened 2019-01-23
Migrate Lisk Modular repository files to framework directory and create lisk-chain module


nazarhussain:
Need to move all framework related files to `lisk-modular` repository the `lisk` repository under `/framework` directory. 

After, move all code of lisk into a module `lisk-chain` and load it with the controller with minimum changes possible (https://github.com/LiskHQ/lisk/issues/2761).

PR #2796, opened 2019-01-30
Resolve the conflicting responsibilities between api/controller and module of Node  - Closes #2783

### LINKED ISSUE #2783: Resolve the conflicting responsibilities between api/controller and module of Node
### Expected behavior
There should be no dependency between `api/controllers/node`and `modules/node`.
Detach the dependencies by moving all of the API logic from `modules/node` to api/controllers/node`. 

As the result, the `modules/node` should be removed, as it's not required by any other module.

All of the node module dependencies needs to be replaced by the direct calls to chain module when implemented. Until chain module is not implemented as an independent module, the `ToDo` placeholders should be left when required. After all, the depencencies should be replaced by modules/component/entities as explained in the comments:
```
Node.prototype.onBind = function(scope) {
	modules = {
		blocks: scope.blocks, //should be replaced by storage.entity.block
		loader: scope.loader, //should be replaced by System Component (to verify)
		peers: scope.peers, //should be replaced by P2P Component/Module
		system: scope.system, //should be replaced by System Component
		delegates: scope.delegates, //should be replaced by Chain Module
	};
	loaded = true;
};
```

### Actual behavior
There is a conflict of responsibilities between the logic realised by `modules/node` and `api/controllers/node`which requires the dependency to the module while executing:

- `NodeController.getConstants`
- `NodeController.getStatus`
- `NodeController.getStatus`
- `NodeController.getForgingStatus`
- `NodeController.updateForgingStatus`

Inside each of this function there is a call to `modules.node.shared` property.

`*`

limiaspasdaniel:
### What was the problem?
There is a conflict of responsibilities between the logic realised by modules/node and api/controllers/node which requires the dependency to the module while executing:

- NodeController.getConstants
- NodeController.getStatus
- NodeController.getStatus
- NodeController.getForgingStatus
- NodeController.updateForgingStatus
- Inside each of this function there is a call to modules.node.shared property.

### How did I fix it?

- Moved all the logic from the node module to the node api controller.
- Removed Node module, NewRelic references and its unit tests.
- Added unit tests to private methods on node api controller.
- Updated existing functional tests.

### How to test it?
Run unit tests and functional tests.
### Review checklist

* The PR resolves #2783 
* Linting passes
* Tests pass


api/controllers/node.js
```diff
@@ -49,34 +60,39 @@ function NodeController(scope) {
...
+				dappWithdrawal: FEES.DAPP_WITHDRAWAL.toString(),
+				dappDeposit: FEES.DAPP_DEPOSIT.toString(),
+			},
+			nethash: library.config.nethash,
+			nonce: library.config.nonce,
+			milestone: blockReward.calcMilestone(height).toString(),
+			reward: blockReward.calcReward(height).toString(),
+			supply: blockReward.calcSupply(height).toString(),
+			version: library.config.version,
+			protocolVersion: library.config.protocolVersion,
```
nazarhussain: I suspect this attribute was not already there.  If you want to introduce it now, then must add to `swagger` specification as well.

api/controllers/node.js
```diff
@@ -86,35 +102,28 @@ NodeController.getConstants = function(context, next) {
...
-			);
-		} catch (error) {
-			return next(error);
-		}
-	});
+NodeController.getStatus = async (context, next) => {
+	try {
+		const networkHeight = await promisify(library.modules.peers.networkHeight)({
+			normalized: false,
+		});
```
nazarhussain: Please promisify the method in constructor and use it during every call.

api/controllers/node.js
```diff
@@ -86,35 +102,28 @@ NodeController.getConstants = function(context, next) {
...
+			consensus: library.modules.peers.getLastConsensus() || 0,
+			currentTime: Date.now(),
+			secondsSinceEpoch: slots.getTime(),
+			height: library.modules.blocks.lastBlock.get().height,
+			loaded: library.modules.loader.loaded(),
+			networkHeight: networkHeight || 0,
+			syncing: library.modules.loader.syncing(),
+		};
+
+		data.transactions = await promisify(library.modules.transactions.shared.getTransactionsCount)();
```
nazarhussain: Same here.

api/controllers/node.js
```diff
@@ -246,4 +247,58 @@ NodeController.getPooledTransactions = function(context, next) {
...
+ * Toggle the forging status of a delegate.
+ * @param {string} publicKey - Public key of a delegate
+ * @param {string} password - Password used to decrypt encrypted passphrase
+ * @param {boolean} forging - Forging status of a delegate to update
+ * @returns {Promise<object>}
+ * @todo Add description for the return value
+ * @private
+ */
+async function _updateForgingStatus(publicKey, password, forging) {
+	return promisify(library.modules.delegates.updateForgingStatus)(
```
nazarhussain: Same here. `promisify` returns a new function, just keep the reference of it in constructor and use it other places.

api/controllers/node.js
limiaspasdaniel: @nazarhussain it was introduced on LIP007 which is already merged on development branch. [here](https://github.com/LiskHQ/lisk/blob/2651b07d0023f8ff7665823bf1c810e2b56b2575/schema/swagger.yml#L2318)

api/controllers/node.js
limiaspasdaniel: done

api/controllers/node.js
limiaspasdaniel: done

api/controllers/node.js
limiaspasdaniel: done

api/controllers/node.js
```diff
@@ -86,35 +110,28 @@ NodeController.getConstants = function(context, next) {
...
+		const networkHeight = await getNetworkHeight({
+			normalized: false,
+		});
+
+		const data = {
+			broadhash: library.modules.system.getBroadhash(),
+			consensus: library.modules.peers.getLastConsensus() || 0,
+			currentTime: Date.now(),
+			secondsSinceEpoch: slots.getTime(),
+			height: library.modules.blocks.lastBlock.get().height,
```
MaciejBaj: Either `library.modules.system.getHeight()` or `storage.blocks.getOne({...})` to remove dependency on `blocks`.

api/controllers/node.js
```diff
@@ -86,35 +110,28 @@ NodeController.getConstants = function(context, next) {
...
-			return next(error);
-		}
-	});
+NodeController.getStatus = async (context, next) => {
+	try {
+		const networkHeight = await getNetworkHeight({
+			normalized: false,
+		});
+
+		const data = {
```
MaciejBaj: We should leave the `ToDo` here to replace all of the `library.moduels` calls after the chain module is extracted #2763

limiaspasdaniel: Done

api/controllers/node.js
limiaspasdaniel: Done

api/controllers/node.js
```diff
@@ -86,35 +110,35 @@ NodeController.getConstants = function(context, next) {
...
+			normalized: false,
+		});
+
+		const [lastBlock] = await library.storage.entities.Block.get(
+			{},
+			{ sort: 'height:desc', limit: 1 }
+		);
+		const { height } = lastBlock;
+
+		// TODO: Replace all library.modules calls after chain module extraction is done.
```
MaciejBaj:
```suggestion
		// TODO: Replace all library.modules calls after chain module extraction is done as part of https://github.com/LiskHQ/lisk/issues/2763.
```

api/controllers/node.js
```diff
@@ -246,4 +262,54 @@ NodeController.getPooledTransactions = function(context, next) {
...
+/**
+ * Get the forging status of a delegate.
+ *
+ * @param {string} publicKey - Public key of delegate
+ * @returns {Promise<object>}
+ * @private
+ */
+async function _getForgingStatus(publicKey) {
+	const keyPairs = library.modules.delegates.getForgersKeyPairs();
+	const internalForgers = library.config.forging.delegates;
```
ManuGowda: why the new name `internalForgers`? it can be `forgingDelegates`?

api/controllers/node.js
```diff
@@ -246,4 +262,54 @@ NodeController.getPooledTransactions = function(context, next) {
...
+	Object.keys(keyPairs).forEach(key => {
+		forgersPublicKeys[keyPairs[key].publicKey.toString('hex')] = true;
+	});
+
+	const fullList = internalForgers.map(forger => ({
+		forging: !!forgersPublicKeys[forger.publicKey],
+		publicKey: forger.publicKey,
+	}));
+
+	if (publicKey && _.find(fullList, { publicKey })) {
```
ManuGowda: the order of checking can be better here.
```
if (publicKey && !_.find(fullList, { publicKey })) {
		return [];
	}

if (_.find(fullList, { publicKey })) {
         return [
			{
				publicKey,
				forging: !!forgersPublicKeys[publicKey],
			},
		];
	}

api/controllers/node.js
limiaspasdaniel: Updated

PR #2811, opened 2019-01-31
Update logger to be modified as component - Closes #2578

### LINKED ISSUE #2578: Update logger component to be modified as component pattern
In new lisk core architecture the `components` are exposed as one single method which actions `json` options and return the object. So we need to update the `./logger.js` to be refactored this way.

diego-G:
### Review checklist

* The PR resolves #2578
* Linting passes
* Tests pass


framework/src/components/logger.js
```diff
@@ -0,0 +1,152 @@
...
+		Object.keys(data).forEach(key => {
+			if (key.search(/passphrase|password/i) > -1) {
+				data[key] = 'XXXXXXXXXX';
+			}
+		});
+		return data;
+	}
+}
+
+function createLoggerComponent(config = {}) {
```
nazarhussain: @diego-G Please move the logger class to `logger.js` and keep the factory method in `index.js`

diego-G: Done

PR #2836, opened 2019-02-06
Account.get() does not work when multiple filters are provided with same key - Closes #2822

### LINKED ISSUE #2822: Account.get() does not work when multiple filters are provided with same key
### Expected behavior
If we run the command: `Account.get([address: '1L}, {address: '2L'}])`, it should return accounts with addresses 1L and 2L if both accounts exist in the database.
### Actual behavior
Running the command: `Account.get([address: '1L}, {address: '2L'}])` returns only 1 account even when both accounts exist in the database.
1.5.0

SargeKhan:
### What was the problem?
The parseFilter function defined in BaseEntity class were not able to handle filters as input when it's an array with the same property name.
### How did I fix it?
Fixed the bug by refactoring the code.
### How to test it?
Run unit tests.
### Review checklist

* The PR resolves #2822 
* Linting passes
* Tests pass


SargeKhan:
@nazarhussain, in transactions element, we don't want to use SQL related filters like `in` or `like`. Therefore, we have gone with the second approach.

PR #2842, opened 2019-02-07
Snapshotting failed on testnet and mainnet for 1.5.0-alpha.2 

### LINKED ISSUE #2831: Snapshotting failed on testnet and mainnet for 1.5.0-alpha.2
### Expected behavior
Snapshotting should successed without error, if there is any error during the begining of snapshot, it should assert and stop snapshotting.
### Actual behavior
```
[inf] 2019-02-04 10:34:59 | Starting lisk with "testnet" genesis block.
[inf] 2019-02-04 10:35:00 | Skipping P2P server initialization due to the config settings - "peers.enabled" is set to false.
[inf] 2019-02-04 10:35:00 | Broadcasting data disabled by user through config.json
[inf] 2019-02-04 10:35:01 | Lisk started: 0.0.0.0:9000
[inf] 2019-02-04 10:35:01 | Modules ready and launched
[inf] 2019-02-04 10:35:02 | Blocks 7422538
[inf] 2019-02-04 10:35:02 | Genesis block matched with database
[inf] 2019-02-04 10:35:02 | Snapshot mode enabled
[inf] 2019-02-04 10:35:02 | Snapshotting to end of round: NaN, height: NaN
```
1.5.0

nazarhussain:
### What was the problem?
The behavior of `isNaN` and `Number.isNaN` is different. Previously we were using `NaN` and updated it to `Number.isNaN`.

### How did I fix it?

Used `parseInt` before passing parameter to `Number.isNaN`

### How to test it?
Do snapshotting for the full rounds. 

### Review checklist

* The PR resolves #2831
* Linting passes
* Tests pass

### DIFF
--- a/modules/loader.js
+++ b/modules/loader.js
@@ -769,7 +769,7 @@ __private.createSnapshot = height => {
 
 	const snapshotRound = library.config.loading.snapshotRound;
 	const totalRounds = Math.floor(height / ACTIVE_DELEGATES);
-	const targetRound = Number.isNaN(snapshotRound)
+	const targetRound = Number.isNaN(parseInt(snapshotRound))
 		? totalRounds
 		: Math.min(totalRounds, snapshotRound);
 	const targetHeight = targetRound * ACTIVE_DELEGATES;


modules/loader.js
```diff
@@ -769,7 +769,7 @@ __private.createSnapshot = height => {
 
 	const snapshotRound = library.config.loading.snapshotRound;
 	const totalRounds = Math.floor(height / ACTIVE_DELEGATES);
-	const targetRound = Number.isNaN(snapshotRound)
+	const targetRound = Number.isNaN(parseInt(snapshotRound))
```
4miners: Maybe `typeof snapshotRound === 'number'`?

nazarhussain: Ideal solution would to be to use Bignumber library, so we should refactor in seperate scope.

PR #2853, opened 2019-02-08
Add transaction proxy class - Closes #2819

### LINKED ISSUE #2819: Create Proxy class for initializing class instance of transactions defined in lisk-elements.
### Expected behavior
Multiple transactions are defined in lisk-elements, it is required to create a proxy function which initializes the relevant class instance (i.e, TransferTransaction, DelegateTransaction etc) based on the `type` property of the transaction.

Helpful link for implementation:
https://stackoverflow.com/questions/34655616/create-an-instance-of-a-class-in-es6-with-a-dynamic-name

### Actual behavior
The functionality is missing

1.6

mitsujutsu:
### What was the problem?

No mapping from type to specific transaction class in elements.

### How did I fix it?

Added proxy class which maps type to elements class and initializes. 

### How to test it?

Awaiting npm lisk-elements update before implementing tests.

### Review checklist

* The PR resolves #2819 
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/chain/helpers/init_transaction.js
+++ b/framework/src/modules/chain/helpers/init_transaction.js
@@ -12,6 +12,9 @@
  * Removal or modification of this copyright notice is prohibited.
  *
  */
+
+'use strict';
+
 const {
 	TransferTransaction,
 	SecondSignatureTransaction,
@@ -23,8 +26,7 @@ const {
 	OutTransferTransaction,
 } = require('@liskhq/lisk-transactions');
 
-class Transaction {
-	constructor(rawTx) {
+module.exports = rawTx => {
 		const map = new Map([
 			[0, TransferTransaction],
 			[1, SecondSignatureTransaction],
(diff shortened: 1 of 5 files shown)


shuse2:
Maybe consider to be a factory function?
class instantiating a different class from constructor seems bit weird

framework/src/modules/chain/helpers/init_transaction.js
```diff
@@ -0,0 +1,48 @@
...
+	]);
+
+	const TransactionClass = map.get(rawTx.type);
+
+	if (!TransactionClass) {
+		throw new Error('Transaction type not found.');
+	}
+
+	return new TransactionClass(rawTx);
+};
```
nazarhussain: This logic should be moved to `modules/transactions` rather than as a helper.

shuse2: this class will be deleted at the end so it should be helper

PR #2857, opened 2019-02-11
Update the temp directory structure suitable for deployment and check the PID file before startup - Closes #2849

### LINKED ISSUE #2849: Update the temp directory structure suitable for deployment and check the PID file before startup
Based on the discussion in STM, we need have a temp structure in place to avoid any issues in upcoming inter-process communication. 

That's the directory structure for temp files we need to follow onward. 

```
/project_root/tmp/lisk/{app_name}/temp/pids
/project_root/tmp/lisk/{app_name}/temp/sockets/bus.sock
/project_root/tmp/lisk/{app_name}/temp/*
```

To make it easy for user to create multiple app name, that option must accept a function as well to generate the app name.

nazarhussain:
### What was the problem?
You can run multiple instances of the application, that uses same directories for storing pid and socket files. That can cause stability issues in future when modules will be communicating to each other over RPC.

### How did I fix it?
Have seperate temp directories for individual application. Its user responsibility to provide unique name. If an application already running with same with same name, we throw error on startup. 

### Review checklist

* The PR resolves #2849
* Linting passes
* Tests pass


framework/src/controller/config/dirs.js
```diff
@@ -0,0 +1,10 @@
+const rootDir = process.cwd();
+
+const systemDirs = appLabel => ({
+	root: rootDir,
+	temp: `${rootDir}/tmp/${appLabel}/`,
+	sockets: `${rootDir}/tmp/${appLabel}/sockets`,
```
fchavant: Will the application create those directories if they do not exist?

framework/src/controller/application.js
```diff
@@ -84,6 +84,10 @@ module.exports = class Application {
 			};
 		}
 
+		if (typeof label === 'function') {
```
shuse2: if the `label` is a function, above `${label} doesn't make sense.
Also, prefer to be consistent in type

framework/src/controller/config/dirs.js
nazarhussain: Yes

framework/src/controller/application.js
nazarhussain: You didn't noticed the line after that condition. Label can be a string or a function that generate this string. 

That's a common pattern used along many other NodeJS project including PgPromise.

nazarhussain:
@MaciejBaj @fchavant We also need to check if we want to keep `logs` directory on top level or also move it to `tmp/logs`?

framework/src/controller/application.js
```diff
@@ -78,6 +78,10 @@ module.exports = class Application {
 		constants = {},
 		config = { components: { logger: null } }
 	) {
+		if (typeof label === 'function') {
+			label = label.call();
+		}
+
 		if (!config.components.logger) {
 			config.components.logger = {
 				filename: `~/.lisk/${label}/lisk.log`,
```
shuse2: Also, need to align the folder naming and contents with `lisk-commander` as from `v1.0` we are using `lisk` command with default folder `.lisk`

nazarhussain: This change will not happen to core as we decide to keep the project root. The code you mentioned above will be fixed in other PR, which specifies default values for each component configuration.

nazarhussain:
@fchavant Yes, we are keeping `logs` directory on rot level.

PR #2865, opened 2019-02-11
Improve the UX/DX for creating a module for framework - Closes #2863

### LINKED ISSUE #2863: Improve the UX/DX for creating a module for framework
### Expected behavior
With the json based module specification, it's easy to export, but its difficult to introduce some common behavior to all modules. 

Also its not aligned with other framework structs we are providing e.g. To create a new transaction we will specify something like 

```
class MyTransaction extends BaseTransaction {}
```

We want to align module framework struct with similar way and introduce a `BaseModule` class. 

### Actual behavior

Currently we are expecting any module to export a `json` object in following structure. 
```
{
   alias: '',
   defaults: {},
   load: async () {},
   unload: async () {},
}
```

nazarhussain:
### What was the problem?
Currently we are expecting any module to export a `json` object in following structure, which was not aligned with other framework structs. 

### How did I fix it?

Introduced a new `BaseModule` class. 

### Review checklist

* The PR resolves #2863
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/controller/application.js
+++ b/framework/src/controller/application.js
@@ -121,19 +121,19 @@ module.exports = class Application {
 	/**
 	 * Register module with the application
 	 *
-	 * @param {Object} moduleSpec - Module specification
+	 * @param {Object} moduleKlass - Module specification
 	 *  @see {@link '../modules/README.md'}
-	 * @param {Object} [config] - Modules configuration object. Provided config will override `moduleSpec.defaults` to generate final configuration used for the module
-	 * @param {string} [alias] - Will use this alias or fallback to `moduleSpec.alias`
+	 * @param {Object} [options] - Modules configuration object. Provided options will override `moduleKlass.defaults` to generate final configuration used for the module
+	 * @param {string} [alias] - Will use this alias or fallback to `moduleKlass.alias`
 	 */
-	registerModule(moduleSpec, config = {}, alias = undefined) {
-		assert(moduleSpec, 'ModuleSpec is required');
+	registerModule(moduleKlass, options = {}, alias = undefined) {
+		assert(moduleKlass, 'ModuleSpec is required');
 		assert(
-			typeof config === 'object',
-			'Module config must be provided or set to empty object.'
+			typeof options === 'object',
+			'Module options must be provided or set to empty object.'
 		);
-		assert(alias || moduleSpec.alias, 'Module alias must be provided.');
-		const moduleAlias = alias || moduleSpec.alias;
+		assert(alias || moduleKlass.alias, 'Module alias must be provided.');
+		const moduleAlias = alias || moduleKlass.alias;
 		assert(
 			!Object.keys(this.getModules()).includes(moduleAlias),
 			`A module with alias "${moduleAlias}" already registered.`
@@ -141,8 +141,8 @@ module.exports = class Application {
 
 		const modules = this.getModules();
 		modules[moduleAlias] = {
-			spec: moduleSpec,
-			config: config || {},
+			klass: moduleKlass,
+			options: options || {},
 		};
 		__private.modules.set(this, modules);
 	}
@@ -151,15 +151,15 @@ module.exports = class Application {
 	 * Override the module's configuration
 	 *
 	 * @param {string} alias - Alias of module used during registration
-	 * @param {Object} config - Override configurations, these will override existing configurations.
+	 * @param {Object} options - Override configurations, these will override existing configurations.
 	 */
-	overrideModuleConfig(alias, config) {
+	overrideModuleOptions(alias, options) {
 		const modules = this.getModules();
 		assert(
 			Object.keys(modules).includes(alias),
 			`No module ${alias} is registered`
 		);
-		modules[alias].config = Object.assign({}, modules[alias].config, config);
+		modules[alias].options = Object.assign({}, modules[alias].options, options);
 		__private.modules.set(this, modules);
 	}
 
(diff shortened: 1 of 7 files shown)


MaciejBaj:
Please resolve the conflicts.

framework/src/controller/application.js
```diff
@@ -111,28 +111,28 @@ module.exports = class Application {
 	/**
 	 * Register module with the application
 	 *
-	 * @param {Object} moduleSpec - Module specification
+	 * @param {Object} moduleKlass - Module specification
```
yatki: Naming looks klassy but Why not using `moduleClass` :D ?

framework/src/controller/controller.js
```diff
@@ -93,54 +93,65 @@ module.exports = class Controller {
...
 
-	async _loadInMemoryModule(alias) {
-		const module = this.modules[alias].spec;
-		const moduleConfig = this.modules[alias].config;
+	async _loadInMemoryModule(alias, Klass, options) {
+		const module = new Klass(options);
 		validateModuleSpec(module);
 
+		const moduleAlias = module.constructor.alias;
+		const moduleInfo = module.constructor.info;
```
yatki: We could deconstruct the object here.

```suggestion
		const {name, version} = module.constructor.info;
```

framework/src/controller/controller.js
```diff
@@ -93,54 +93,65 @@ module.exports = class Controller {
...
 			})`
 		);
 	}
 
 	async unloadModules(modules = null) {
-		return Promise.mapSeries(modules || Object.keys(this.modules), async m => {
-			await this.modules[m].spec.unload();
-			delete this.modules[m];
-		});
+		return Promise.mapSeries(
```
yatki: I think we should avoid using bluebird methods since it's just making harder to migrate using native Promises.

framework/src/controller/controller.js
```diff
@@ -93,54 +93,65 @@ module.exports = class Controller {
 		);
 	}
 
-	async _loadModules() {
-		return Promise.each(Object.keys(this.modules), m =>
-			this._loadInMemoryModule(m)
+	async _loadModules(modules) {
+		return Promise.each(Object.keys(modules), alias =>
```
yatki: I think we should avoid using bluebird methods since it's just making harder to migrate using native Promises.

framework/src/modules/README.md
```diff
@@ -1,32 +1,42 @@
...
+			name: '',
+			};
 	},
 
 	/**
 	 * Supported configurations for the module with default values
 	 */
 
-	defaults: {},
+	get defaults() { return {}; },
```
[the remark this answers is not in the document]
shuse2: I think if we only have `getter` it will throw an error if we do `* obj.defults = 'something'`
but it is very confusing to use `defaults` with getter because in babel/es rule you can export `default`. so,
```
export default class X {
  default: 'x',
}
```
and you can do
`import X from './above.js'`

framework/src/controller/application.js
nazarhussain: Nothing special, but nothing wrong either. I saw this convention is widely used.

framework/src/modules/README.md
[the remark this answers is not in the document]
nazarhussain: In this project unless we have babel or typescript we can not use `export default`. Event though its not confusing to if we have `const options = module.defaults`  and `export default`. Both have totally different reasons and syntax.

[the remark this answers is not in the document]
nazarhussain: @yatki getters are used if you don't want to a setter ;) 

Yes you can ask to freeze an object that will not be modified, but then it can not be modified by internal logic as well. So getters are a valid use case and in my experience its widely adopted language struct overall, for JS its newly announced language feature so will take some time for users to adopt.

framework/src/controller/controller.js
nazarhussain: Any native alternative to execute promises in sequence?

framework/src/errors.js
```diff
@@ -25,7 +25,22 @@ class SchemaValidationError extends FrameworkError {
...
+ * Error occurred when you some required function have no implementation
+ * @extends ImplementationMissingError
+ * @namespace Framework.errors
+ */
+class ImplementationMissingError extends FrameworkError {
+	/**
+	 * Create a schema validation error object
+	 */
+	constructor() {
+		super('Schema validation error');
```
lsilvs: The message was copied from `SchemaValidationError`

framework/src/controller/controller.js
```diff
@@ -160,54 +155,62 @@ module.exports = class Controller {
...
+		this.modules[moduleAlias] = module;
 		this.logger.info(
-			`Module ready with alias: ${module.alias}(${module.info.name}:${
-				module.info.version
-			})`
+			`Module ready with alias: ${moduleAlias}(${name}:${version})`
 		);
 	}
 
 	async unloadModules(modules = null) {
```
lsilvs: Prefer using `undefined` over `null`

framework/src/controller/controller.js
```diff
@@ -160,54 +155,62 @@ module.exports = class Controller {
...
-		return Promise.mapSeries(modules || Object.keys(this.modules), async m => {
-			await this.modules[m].spec.unload();
-			delete this.modules[m];
-		});
+		return Promise.all(
+			(modules || Object.keys(this.modules)).map(async alias => {
```
lsilvs: I don't think this is behaving as you expect.
Do you want to run the `unload`'s in parallel?

PR #2871, opened 2019-02-13
Implement StateStore for fetching accounts and transactions and saving updated accounts in the database - Closes #2820

### LINKED ISSUE #2820: Implement StateStore for fetching accounts and transactions and saving updated accounts in the database
### Expected behavior
In order to verify/apply transactions in memory, we need to implement StateStore. StateStore manages accounts and transactions in memory. Furthermore, it allows the functionality to update accounts in memory and save the updated state in the database on completion of processing of block. 
### Actual behavior
StateStore currently is not implemented.

1.6

SargeKhan:
### What was the problem?
We need to implement StateStore for transactions and accounts as part of transaction processing efficiency LIP
### How did I fix it?
Added functionality for StateStore
### How to test it?
Run tests
### Review checklist

* The PR resolves #2820
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/chain/logic/state_store/account_store.js
+++ b/framework/src/modules/chain/logic/state_store/account_store.js
@@ -66,17 +66,18 @@ class AccountStore {
 	}
 
 	getOrDefault(primaryValue) {
-		let element = this.data.find(
+		const element = this.data.find(
 			item => item[this.primaryKey] === primaryValue
 		);
-		if (!element) {
-			element = {
-				...defaultAccount,
-				[this.primaryKey]: primaryValue,
-			};
-			this.data.push(element);
+		if (element) {
+			return element;
 		}
-		return element;
+		const defaultElement = {
+			...defaultAccount,
+			[this.primaryKey]: primaryValue,
+		};
+		this.data.push(defaultElement);
+		return defaultElement;
 	}
 
 	find(fn) {
--- a/framework/src/modules/chain/logic/state_store/account_store.js
+++ b/framework/src/modules/chain/logic/state_store/account_store.js
@@ -1,0 +1,127 @@
+const _ = require('lodash');
+
+const defaultAccount = {
+	publicKey: null,
+	secondPublicKey: null,
+	secondSignature: false,
+	u_secondSignature: false,
+	username: null,
+	u_username: null,
+	isDelegate: false,
+	u_isDelegate: false,
+	balance: '0',
+	u_balance: '0',
+	missedBlocks: 0,
+	producedBlocks: 0,
+	rank: null,
+	fees: '0',
+	rewards: '0',
+	vote: '0',
+	nameExist: false,
+	u_nameExist: false,
+	multiMin: 0,
+	u_multiMin: 0,
+	multiLifetime: 0,
+	u_multiLifetime: 0,
+};
+
+class AccountStore {
+	constructor(accountEntity, { mutate } = { mutate: true }) {
+		this.account = accountEntity;
+		this.data = [];
+		this.updatedKeys = {};
+		this.primaryKey = 'address';
+		this.name = 'Account';
+		this.mutate = mutate;
+	}
+
+	async cache(filter, tx) {
+		const result = await this.account.get(filter, {}, tx);
+		this.data = _.uniqBy([...this.data, ...result], this.primaryKey);
+		return result;
+	}
+
+	get(primaryValue) {
+		const element = this.data.find(
+			item => item[this.primaryKey] === primaryValue
+		);
+		if (!element) {
+			throw new Error(
+				`${this.name} with ${this.primaryKey} = ${primaryValue} does not exist`
+			);
+		}
+		return element;
+	}
+
+	getOrDefault(primaryValue) {
+		let element = this.data.find(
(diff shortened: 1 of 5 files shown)


framework/src/modules/chain/logic/state_store/account_store.js
```diff
@@ -0,0 +1,136 @@
...
+		if (!element) {
+			throw new Error(
+				`${this.name} with ${this.primaryKey} = ${primaryValue} does not exist`
+			);
+		}
+		return element;
+	}
+
+	getOrDefault(primaryValue) {
+		let element = this.data.find(
```
shuse2: prefer to use `const`

framework/src/modules/chain/logic/state_store/account_store.js
```diff
@@ -0,0 +1,136 @@
...
+		this.data = [];
+		this.updatedKeys = {};
+		this.primaryKey = 'address';
+		this.name = 'Account';
+		this.mutate = mutate;
+		this.originalData = [];
+		this.originalUpdatedKeys = {};
+	}
+
+	async cache(filter, tx) {
```
shuse2: we don't have access to `tx` in the transaction module ='(

framework/src/modules/chain/logic/state_store/transaction_store.js
```diff
@@ -0,0 +1,61 @@
...
+
+class TransactionStore {
+	constructor(transactionEntity) {
+		this.transaction = transactionEntity;
+		this.data = [];
+		this.primaryKey = 'id';
+		this.name = 'Transaction';
+	}
+
+	async cache(filter, tx) {
```
shuse2: same as above

framework/src/modules/chain/logic/state_store/index.js
```diff
@@ -0,0 +1,35 @@
...
+const TransactionStore = require('./transaction_store');
+
+class StateStoreManager {
+	constructor(storage, cb) {
+		this.entities = {
+			Account: storage.entities.Account,
+			Transaction: storage.entities.Transaction,
+		};
+
+		this.sandbox = {};
```
shuse2: is this `sandbox` right terminology? maybe just `stateStore`?

framework/src/modules/chain/logic/state_store/account_store.js
SargeKhan: We initalize `element` variable if it doesn't exist. That's why I am using const.

shuse2: maybe something like
```
const element = this.data.find(xxx);
if (element)
  return element;

const defaultElement = { ...defaultAccount, xxx};
this.data.push(defaultElement)
return defaultElement;
```

framework/src/modules/chain/logic/state_store/account_store.js
```diff
@@ -0,0 +1,137 @@
...
+	u_nameExist: false,
+	multiMin: 0,
+	u_multiMin: 0,
+	multiLifetime: 0,
+	u_multiLifetime: 0,
+};
+
+class AccountStore {
+	constructor(accountEntity, { mutate } = { mutate: true }) {
+		this.account = accountEntity;
```
pablitovicente: wouldn't it be better to name this property `accountEntity` so when reading other methods it's very clear that the operation it's an storage access?

framework/src/modules/chain/logic/state_store/index.js
```diff
@@ -0,0 +1,35 @@
...
+	createStore(options) {
+		this.stateStore = {
+			account: new AccountStore(this.entities.Account, options),
+			transaction: new TransactionStore(this.entities.Transaction, options),
+		};
+		return this.stateStore;
+	}
+
+	createSnapshot() {
+		this.sandbox.account.createSnapshot();
```
pablitovicente: Need to update to this.stateStore maybe?

framework/src/modules/chain/logic/state_store/index.js
```diff
@@ -0,0 +1,35 @@
...
+		this.sandbox.transaction.createSnapshot();
```
pablitovicente: Need to update to this.stateStore maybe?

framework/src/modules/chain/logic/state_store/index.js
```diff
@@ -0,0 +1,35 @@
...
+	}
+
+	restoreSnapshot() {
+		this.sandbox.account.restoreSnapshot();
+		this.sandbox.transaction.restoreSnapshot();
```
pablitovicente: Need to update to this.stateStore maybe?

framework/src/modules/chain/logic/state_store/account_store.js
SargeKhan: There is no other property named `account` in `AccountStore`. So, I think it is fine. If you think it is really confusing then i can change it. :)

pablitovicente: No it's ok you have a good point :)

SargeKhan:
@shuse2, the PR for `account.get` was not merged in the feature branch. So, I just cherry-picked the commits.

Issue #2858, opened 2019-02-11
Istanbul throws "Transformation error" exceptions


yatki:
### Expected behavior

We should use nyc since istanbul is not being maintained anymore.

### Actual behavior

We are currently using istanbul, which causes the exceptions below:

![image](https://user-images.githubusercontent.com/26697004/52564922-a3d6fa80-2e05-11e9-9782-5aaf03fd24d0.png)

PR #2887, opened 2019-02-14
Refactor the chain module initialization to simplify and maintainable - Closes #2866

### LINKED ISSUE #2866: Refactor the chain module initialization to simplify and maintainable
### Expected behavior
The steps to initialize chain module must be very clear and readable, that helps to test and maintain the key business logic of the application. 

### Actual behavior

Currenlty there is a nested structure of `async.auto` causing ambiguity in code understanding and readability. 

https://github.com/LiskHQ/lisk/blob/25354edd4c46b9d4c114c3a823f03db822167e81/framework/src/modules/chain/chain.js#L138
https://github.com/LiskHQ/lisk/blob/25354edd4c46b9d4c114c3a823f03db822167e81/framework/src/modules/chain/chain.js#L516

1.6

nazarhussain:
### What was the problem?

Chain module initialization was pretty old pattern migrated from old app.js. Using a lot of `async.auto` which was making it confusing to read and maintain the code. 

### How did I fix it?
Refactored the complete logic by splitting it to different steps and also using async/await.

### Review checklist

* The PR resolves #2866
* Linting passes
* Tests pass


framework/src/modules/chain/init_steps/init_logic_structs.js
```diff
@@ -0,0 +1,51 @@
...
+		});
+	});
+
+	return {
+		account: accountLogic,
+		transaction: transactionLogic,
+		block: blockLogic,
+		peers: peersLogic,
+	};
+};
```
limiaspasdaniel: Just one question, because I am not really aware of its purpose. Why is `block_reward.js` which is under `logic` folder being required across different files and not being inserted on `scope` object?

framework/test/common/application.js
```diff
@@ -15,91 +15,100 @@
...
-		__init(options.scope, cb);
-	}
+	__init(options.sandbox, options.scope)
+		.then(scope => cb(null, scope))
+		.catch(err => cb(err));
 }
 
 // Init whole application inside tests
-function __init(initScope, done) {
+async function __init(sandbox, initScope) {
```
limiaspasdaniel: Are we using double `_` or just one? I've been seeing single `_` for private methods everywhere but `__` for private variables at the same time.

framework/test/common/application.js
```diff
@@ -108,429 +117,125 @@ function __init(initScope, done) {
...
+					}
+
+					const keypairs = scope.rewiredModules.delegates.__get__(
+						'__private.keypairs'
 					);
-					__testContext.debug('initApplication: Loading delegates...');
 
-					const loadDelegates = scope.rewiredModules.delegates.__get__(
-						'__private.loadDelegates'
+					const delegates_cnt = Object.keys(keypairs).length;
```
limiaspasdaniel: Can be this renamed to `delegatesCount` instead?

framework/test/unit/modules/chain/unit/modules/peers.js
```diff
@@ -79,21 +80,30 @@ describe('peers', () => {
...
-				},
+		// scope.nonce = NONCE;
+		//
+		// scope.components.storage = {
+		// 	entities: {
+		// 		Peer: {
+		// 			get: sinonSandbox.stub().resolves([prefixedPeer]),
+		// 		},
+		// 	},
+		// };
```
limiaspasdaniel: Do we need to keep this commented code here?

package.json
```diff
@@ -64,6 +64,7 @@
 		"lodash": "4.17.11",
 		"method-override": "3.0.0",
 		"newrelic": "5.0.0",
+		"p-event": "=2.3.1",
```
limiaspasdaniel: I think`=` can be removed.

framework/src/modules/chain/chain.js
```diff
@@ -114,673 +89,89 @@ module.exports = class Chain {
...
-					 */
-					function(scope, cb) {
-						httpApi.bootstrapSwagger(
-							scope.network.app,
-							scope.config,
-							scope.logger,
-							scope,
-							cb
-						);
+			const self = this;
```
shuse2: prefer to remove `self`

framework/src/modules/chain/init_steps/create_bus.js
```diff
@@ -0,0 +1,37 @@
+const changeCase = require('change-case');
+
+module.exports = async () => {
+	const Bus = function() {
```
shuse2: Prefer to consistently use `es6` class syntax.
Bus should be defined outside as class and used

framework/src/modules/chain/init_steps/create_http_server.js
```diff
@@ -0,0 +1,110 @@
...
+	new Promise((resolve, reject) => {
+		server.listen({ host, port }, err => (err ? reject(err) : resolve()));
+	});
+
+module.exports = async ({ config, components: { logger } }) => {
+	const app = express();
+
+	if (config.coverage) {
+		// eslint-disable-next-line import/no-extraneous-dependencies
+		const im = require('istanbul-middleware');
```
shuse2: are we still using this? (probably out of scope of this PR)

framework/src/modules/chain/init_steps/create_http_server.js
```diff
@@ -0,0 +1,110 @@
...
+	let privateKey;
+	let certificate;
+	let https;
+	let https_io;
+
+	if (config.api.ssl && config.api.ssl.enabled) {
+		privateKey = fs.readFileSync(config.api.ssl.options.key);
+		certificate = fs.readFileSync(config.api.ssl.options.cert);
+
+		https = require('https').createServer(
```
shuse2: prefer not to use dynamic import

framework/src/modules/chain/init_steps/create_http_server.js
```diff
@@ -0,0 +1,110 @@
...
+
+	if (config.trustProxy) {
+		app.enable('trust proxy');
+	}
+
+	const server = http.createServer(app);
+	const io = socket(server);
+
+	let privateKey;
+	let certificate;
```
shuse2: privateKey and certificate can be `const inside

framework/src/modules/chain/init_steps/create_http_server.js
```diff
@@ -0,0 +1,110 @@
...
+	const network = {
+		express,
+		app,
+		server,
+		io,
+		https,
+		https_io,
+	};
+
+	network.listen = async () => {
```
shuse2: Prefer to define this first, and then return `network`

framework/src/modules/chain/init_steps/create_socket_cluster.js
```diff
@@ -0,0 +1,82 @@
...
+	config,
+	network,
+	modules: { transport },
+	components: { logger },
+}) => {
+	if (!config.peers.enabled) {
+		logger.info(
+			'Skipping P2P server initialization due to the config settings - "peers.enabled" is set to false.'
+		);
+		return Promise.resolve();
```
shuse2: async automatically return `Promise`

framework/src/modules/chain/init_steps/init_modules.js
```diff
@@ -0,0 +1,48 @@
+const util = require('util');
+const async = require('async');
+
+const promisifyParallel = util.promisify(async.parallel);
```
shuse2: maybe we can omit using `async` here?
I don't think it has much benefit parallelizing the initialization

framework/src/modules/chain/init_steps/init_logic_structs.js
nazarhussain: Only creator knows...

framework/test/common/application.js
nazarhussain: As of now we are using double underscore notation in core. So we should stick wit it.

framework/src/modules/chain/chain.js
```diff
@@ -114,673 +89,89 @@ module.exports = class Chain {
...
-				}
-			}
-		);
+			await initSteps.bootstrapStorage(
+				scope,
+				global.constants.ACTIVE_DELEGATES
+			);
+			await initSteps.bootstrapCache(scope);
+
+			scope.bus = await initSteps.createBus();
```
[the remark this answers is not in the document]
nazarhussain: It may or may not, not sure at the moment.

framework/src/modules/chain/init_steps/create_bus.js
nazarhussain: Yes it should, but in this iteration I just used the existing structure.

nazarhussain: I believe in future with the help of channels we will not be using this bus any more :) so for now leave it as it is.

framework/src/modules/chain/init_steps/create_http_server.js
nazarhussain: That can be quested in seperate PR, where @yatki is moving to `nyc`, may be after that it can be removed.

yatki: @nazarhussain I agree that we should get rid of it. However, I didn't fix it with #2858. we need another issue because we need to modify the way we instantiate the application in functional tests.

framework/src/modules/chain/init_steps/attach_swagger.js
```diff
@@ -0,0 +1,12 @@
...
+
+const promisifiedBootstrapSwagger = util.promisify(httpApi.bootstrapSwagger);
+
+module.exports = async scope =>
+	promisifiedBootstrapSwagger(
+		scope.network.app,
+		scope.config,
+		scope.components.logger,
+		scope
+	);
```
yatki: I don't know if it is scope of this PR but I really do not like passing this `scope` variable around the modules. We already had a lot of issues with passing and object and reaching its nested properties. It makes mocking really painful, since I have to create a mock object like below for testing this module:

```js
const scope = {
 network : {
  app: sinon.spy(),
 },
config: {..},
components: {
  logger: sinon.stub()
 },
};
```

Also I really don't understand why we are passing child properties as first 3 arguments to `promisifiedBootstrapSwagger`, since we are again already passing the scope object as the 4th argument to this function.

Again, I can understand if in this PR focus wasn't improving the code. But just wanted to raise my concern on this. If possible, please refactor this in this PR or I can create an issue to refactor them later.

framework/src/modules/chain/init_steps/create_http_server.js
shuse2: @yatki Please create the related issue

framework/src/modules/chain/chain.js
nazarhussain: It is required because of reference in the sequence callback.

framework/src/modules/chain/init_steps/init_modules.js
```diff
@@ -19,29 +16,39 @@ const modulesList = {
...
-				// eslint-disable-next-line import/no-dynamic-require
-				const DynamicModule = require(modulesList[name]);
-				return new DynamicModule(configModulesCb, scope);
-			});
-		};
+			})
+	);
+
+	(await Promise.all(modulePromises)).forEach((module, index) => {
+		modules[moduleNames[index]] = module;
```
shuse2: instead of declearing at top, we should map it to the new variable.
```
const resolvedModules = await Promise.all(modulePromises);
const modules = resolvedModules.reduce((prev, module, index) => ({ ...prev, [index]: module }), {});
```

framework/src/modules/chain/chain.js
```diff
@@ -107,8 +107,8 @@ module.exports = class Chain {
 				lastCommit,
 				ed,
 				build: versionBuild,
-				config: this.options.config,
-				genesisBlock: { block: this.options.config.genesisBlock },
+				config: self.options.config,
```
shuse2: I mean remove `self` and use this instead 😕

nazarhussain: As commented earlier, we have to use self to pass reference to sequence. Otherwise have to bind this to sequence events.

framework/src/modules/chain/init_steps/attach_swagger.js
MaciejBaj: **Scope**
Would be nice not to have it at all and pass the dependencies directly into logic and modules. But it demands many change to the exisitng structure which is already multiplied in the test application loader. Not the priority for now, but something to carefully refactor later good planning. There is a second aspect to that - modyifying the variables and subscribing to those changes. Right now there is a system component that works de facto like a state container - we should come up with a pattern we could re-use (like React `setState`). We will discuss it further but not on this PR.

framework/src/modules/chain/init_steps/create_socket_cluster.js
```diff
@@ -0,0 +1,82 @@
...
+
+	socketCluster.on('workerExit', workerInfo => {
+		let exitMessage = `Worker with pid ${workerInfo.pid} exited`;
+		if (workerInfo.signal) {
+			exitMessage += ` due to signal: '${workerInfo.signal}'`;
+		}
+		logger.error(exitMessage);
+	});
+
+	return promisifyEvent(socketCluster, 'ready').then(() => {
```
MaciejBaj: Anything else we could use here not to introduce the new dependency to `p-event`? We introduced several with 1.6 anyway.

nazarhussain: I also wish not to, but in one review it was suggested to promisify event. I could not find any native utility method for it. Suggest either we create our own `helper` :) for it or don't promisify events.

shuse2: maybe we can do
```
await new Promise(resolve => {
  socketCluster.on('ready', resolve);
});
logger.info()
...
```
@nazarhussain  please consult with @jondubois

framework/src/modules/chain/init_steps/create_http_server.js
yatki: @shuse2 @nazarhussain created https://github.com/LiskHQ/lisk/issues/2919

framework/src/modules/chain/init_steps/create_socket_cluster.js
nazarhussain: @shuse2 @MaciejBaj I am well aware of using Promise constructor ;) But if you look at this issue https://github.com/LiskHQ/lisk-modular/issues/51 it was suggested to not use Promise constructor instead use some library to promisify events.

nazarhussain: After the discussion with @MaciejBaj we decided not to promisify event handlers. We will stick to the NodeJS event handler approach everywhere.

shuse2: I agree that it's better `event` and `Promise` is different

PR #2888, opened 2019-02-14
Validate genesis block - Closes #2884

### LINKED ISSUE #2884: Create controller/schemas/genesisBlock.js to validate genesis block
### Expected behavior
The genesis block structure passed to the application must be validated for a valid structure. 

### Actual behavior
Currently we are not validating the genesis block 

1.6.0

michielmulders:
### What was the problem?
No validation present for Genesis Block configuration when initializing the application.

### How did I fix it?
JSON Schema validation added.

### How to test it?
Start application or use an online schema validator.
Add the schema in it and test against all different genesis_block.json definitions.

### Review checklist

* The PR resolves #2884 
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/controller/schema/application.js
+++ b/framework/src/controller/schema/application.js
@@ -8,13 +8,29 @@ module.exports = {
 	genesisBlock: {
 		$id: '#genesisBlock',
 		type: 'object',
-		required: ['xxx'],
+		required: [
+			'version',
+			'totalAmount',
+			'totalFee',
+			'reward',
+			'payloadHash',
+			'timestamp',
+			'numberOfTransactions',
+			'payloadLength',
+			'previousBlock',
+			'generatorPublicKey',
+			'transactions',
+			'height',
+			'blockSignature',
+			'id',
+		],
 		properties: {
 			version: {
 				type: 'integer',
 			},
 			totalAmount: {
-				type: 'integer',
+				type: 'string',
+				pattern: '^[0-9]+$',
 			},
 			totalFee: {
 				type: 'integer',
@@ -87,7 +103,8 @@ module.exports = {
 				maximum: 7,
 			},
 			amount: {
-				type: 'integer',
+				type: 'string',
+				pattern: '^[0-9]+$',
 			},
 			fee: {
 				type: 'integer',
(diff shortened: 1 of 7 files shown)


framework/src/controller/schema/application.js
```diff
@@ -6,8 +6,144 @@ module.exports = {
...
+
+	transactions: {
+		id: 'transactions',
+		type: 'object',
+		required: ['type', 'timestamp', 'senderPublicKey', 'signature'],
+		properties: {
+			type: {
+				type: 'integer',
+				minimum: 0,
+				maximum: 7,
```
nazarhussain: In genesis block we can't have all transactions.

michielmulders: Pushed new commit: enum that allows only type `0`, `2`, and `3`. (The types that occur in the `genesis_block.json`.

nazarhussain:
@MaciejBaj We need to keep this in mind https://github.com/LiskHQ/lisk/pull/2888#discussion_r259390801 when we allow generation of genesis block for custom applications.

framework/src/modules/chain/helpers/z_schema.js
```diff
@@ -393,6 +392,23 @@ const liskFormats = {
 			);
 		},
 	},
+	/**
+	 * Returns true if value is lower than or equal to `TOTAL_AMOUNT`.
+	 *
+	 * @param {Object} value
+	 * @returns {boolean}
+	 */
+	maxAmount: {
```
MaciejBaj: `maxAmount` would be `TOTAL_AMOUNT`. The validation is checking if it's a correct amount - positive and lower than `TOTAL_AMOUNT`. 

I suggest to rename it to `amount`.

michielmulders: Not possible to define both `string` and `number` for a custom schema.
I changed it to `numAmount` therefore as `amount` is already been taken by the amount validator for strings.

MaciejBaj:
Also, we should address the issue of a different type for storing the amounts in genesis block with the goal to remove `numAmount` type from z-schema and change all of the amount occurrences across the application to strings. Can you open an issue for it @michielmulders?

PR #2890, opened 2019-02-15
Update system as a component - Closes #2760

### LINKED ISSUE #2760: Create a component system which will replace current modules/system to always have latest system status.
Please create a component named `system` that follow the guidelines specified as: 

https://github.com/LiskHQ/lisk-modular/blob/master/components/README.md

It should replace the current `modules/system.js` and don't implement any breaking changes. 

- The component should accept following as config parameters. 

```
version
minVersion
height
nethash
broadhash
protocolVersion
nonce
```

- [x] Additionally it should accept an instance of `storage` as config param. That should be used to fetch blocks and calculate the `broadhash`. 
- [x] The static attributes e.g.  `os` should not be configureable rather initialized in the component 
- [x] The component expose `update` method which don't accept any parameters, rather when called it get latest block height from storage and also calculate the broadhash. 
- ~~Module specific information `wsPort` and `httpPort` should not be part of the system component.~~
- [x] Update occurrences of `modules/system` with this component 
- [x] Remove the all references of `modules/system` and remove the file.

diego-G:
### Review checklist

* The PR resolves #2760
* Linting passes
* Tests pass


framework/src/components/system/system.js
```diff
@@ -0,0 +1,155 @@
...
+		return nonce && this.headers.nonce !== nonce;
+	}
+
+	/**
+	 * Updates private broadhash and height values.
+	 *
+	 * @param {function} cb - Callback function
+	 * @returns {setImmediateCallback} cb, err
+	 */
+	update(cb) {
```
jondubois: It looks like this function supports both callback `cb` and promises.

Is it possible to use promises only and change dependent parts of the code?

framework/src/modules/chain/chain.js
```diff
@@ -52,7 +53,6 @@ const config = {
 		loader: './modules/loader.js',
 		multisignatures: './modules/multisignatures.js',
 		peers: './modules/peers.js',
-		system: './modules/system.js',
 		signatures: './modules/signatures.js',
 		transactions: './modules/transactions.js',
 		transport: './modules/transport.js',
```
jondubois: This is not directly relevant to this PR, but I find the property name `modules` confusing here in the context of lisk-modular.

It seems that we should make a clear distinction between modules and components within our code. Is this something that is planned to be changed later?

framework/src/modules/chain/modules/loader.js
```diff
@@ -1151,7 +1151,7 @@ Loader.prototype.onPeersReady = function() {
 };
 
 /**
- * Assigns needed modules from scope to private modules constiable.
+ * Assigns needed modules and componenets from scope to private constants.
```
jondubois: Spelling error 'componenets'.

framework/test/unit/modules/chain/unit/api/ws/middleware/emit.js
```diff
@@ -24,45 +23,44 @@ describe('emitMiddleware', () => {
...
-					connectionsTable.getNonce = sinon.stub();
-					validReq.event = 'postBlock';
-					done();
-				});
-
-				afterEach(done => {
-					connectionsTable.getNonce.reset();
-					validReq.data = {};
-					done();
+		const receiveDataEvents = [
```
jondubois: Can this declaration be moved to near the top of the `describe` block, it looks a bit confusing here.

framework/src/components/system/system.js
```diff
@@ -0,0 +1,155 @@
...
+ * @requires os
+ * @requires semver
+ * @param {Object} options - System options
+ * @param {Object} logger
+ */
+class System {
+	constructor(config, logger, storage) {
+		this.logger = logger;
+		this.storage = storage;
+		this.headers = {
```
shuse2: any particular reason to make it nested?

framework/src/components/system/system.js
diego-G: @jondubois It only supports callback. It implies a cascade rewrite of functions that gets completely out of the issue's scope. If you think it's the time to make it, I can create another issue to convert that function to async.

framework/src/modules/chain/chain.js
diego-G: Please, raise this concern to @nazarhussain.

framework/src/components/system/system.js
```diff
@@ -0,0 +1,155 @@
...
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+'use strict';
+
+const os = require('os');
+const crypto = require('crypto');
+const semver = require('semver');
+
+let self;
```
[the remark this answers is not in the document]
diego-G: Removed. I forgot to clean this after removing the former async series on `update()` function.

framework/src/components/system/system.js
diego-G: To keep old behaviour.

framework/src/components/system/index.js
```diff
@@ -0,0 +1,25 @@
...
+
+const System = require('./system');
+
+function createSystemComponent(config, logger, storage) {
+	return new System(config, logger, storage);
+}
+
+module.exports = {
+	createSystemComponent,
+};
```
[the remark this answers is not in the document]
diego-G: We've done this with the rest of components. You can have a look to `cache`, `logger` or `storage`. This way is more convenient if we need to export constants as in `cache` for example.

framework/src/components/system/system.js
shuse2: Let's remove the return if `async` or `Promise` is not supported

diego-G:
@shuse2 FYI https://github.com/LiskHQ/lisk/issues/2904

framework/src/components/system/system.js
```diff
@@ -0,0 +1,151 @@
...
+		this.headers = {
+			os: os.platform() + os.release(),
+			version: config.version,
+			wsPort: config.wsPort,
+			httpPort: config.httpPort,
+			minVersion: config.minVersion,
+			protocolVersion: config.protocolVersion,
+			height: 1,
+			nethash: config.nethash,
+			broadhash: config.nethash,
```
nazarhussain: is this some logical error or we really assigned `broadhash = nethash` ?

framework/src/components/system/system.js
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	/**
+	 * Updates private broadhash and height values.
+	 *
+	 * @param {function} cb - Callback function
+	 * @returns {setImmediateCallback} cb, err
+	 */
+	update(cb) {
+		return this.storage.entities.Block.get(
```
nazarhussain: This function is returning promise as well as accepting a callback. Either we should refactor to not use callback or we should not return the Promise.

framework/src/components/system/system.js
diego-G: Yes, It was like that.

framework/src/components/system/system.js
diego-G: Already discussed #2904

framework/src/modules/chain/modules/loader.js
```diff
@@ -1178,14 +1178,14 @@ Loader.prototype.onBind = function(scope) {
 };
 
 /**
- * Sets private constiable loaded to true.
```
MaciejBaj: :(

framework/test/common/application.js
```diff
@@ -127,6 +127,12 @@ function __init(initScope, done) {
 		async.auto(
 			{
 				config(cb) {
+					__testContext.config.syncing.active = false;
+					__testContext.config.broadcasts.active = false;
```
MaciejBaj: For the future reference - can you explain why exactly are you changing it?

diego-G: Before the dummy node we initiate for our integration & unit suites, wasn't able to connect to another node independent from the tests. Therefore tests were not affected. With the current PR, the dummy node is able to connect to another, fact that changes the expectations. For example the node is able to launch synchronisation process and receive extra unexpected blocks.

PR #2891, opened 2019-02-15
Replace dependant trs tables by json field in trs table - Closes #2818

### LINKED ISSUE #2818: Replace dependant trs tables by json field in trs table
### Expected behavior
I expect to any custom transaction to be able to store its asset data without creating a migrations.

As a result we can get rid of `full_blocks_list` view (which is expensive) and its dependant tables:
```
delegates
votes
signatures
multisignatures
dapps
intransfer
outtransfer
transfer
```

### Actual behavior
Currently, if I want to introduce a new transaction type with assets I have to write a migration to add new table.
By introducing a json field in trs table all related asset can be saved there and different transactions can have different asset structures.

1.5

lsilvs:
### What was the problem?
Each transaction type had one dependant table to save its asset values making it expensive to join all tables together and making it harder to create new custom transactions.

### How did I fix it?
I've created a `jsonb` column in `trs` table called `asset` in order to save transactions specific asset in the same place.

The Transfer Transaction (type 0) was migrated from `transfer` table to a column called `transferData` in `trs` table due to limitation of Postgres with utf-8 null byte character (at present we have one transfer transaction with null byte in mainnet and testnet).

### How to test it?
Run test suite.

### Review checklist

* The PR resolves #2818
* Linting passes
* Tests pass


storage/sql/migrations/updates/20190204170819_migrate_trs_dependant_tables_to_trs_asset_field.sql
```diff
@@ -0,0 +1,99 @@
...
+
+
+/*
+  DESCRIPTION: Migrate all trs dependant tables into trs asset field
+
+  PARAMETERS: None
+*/
+
+-- Migrate transfer table into trs transferData field
+UPDATE trs
```
pablitovicente: Maybe disable index temporarily for `trs` table due to the big number of updates being run here could help with speed?

storage/utils/filters.js
```diff
@@ -30,50 +30,56 @@ function filterGenerator(
 	const getValue = filterAlias =>
 		serializer.call(null, null, 'select', filterAlias, fieldName);
 
+	// Allow to use table name as prefix (example: table.field)
+	fieldName = fieldName
+		.split('.')
+		.map(f => `"${f}"`)
```
pablitovicente: could this variable name be improved?

storage/utils/filters.js
```diff
@@ -30,50 +30,56 @@ function filterGenerator(
 	const getValue = filterAlias =>
 		serializer.call(null, null, 'select', filterAlias, fieldName);
 
+	// Allow to use table name as prefix (example: table.field)
+	fieldName = fieldName
```
shuse2: instead of modifying better to create new `const`?

storage/sql/migrations/updates/20190204170819_migrate_trs_dependant_tables_to_trs_asset_field.sql
```diff
@@ -0,0 +1,99 @@
...
+SET "transferData" = (
+        SELECT t."data"
+        FROM transfer as t
+        WHERE t."transactionId" = trs.id
+    )
+where id IN (SELECT "transactionId" FROM transfer);
```
shuse2: `where` => `WHERE` to be consistent?

storage/entities/transaction.js
```diff
@@ -180,96 +159,104 @@ class Transaction extends BaseEntity {
...
-		);
-		this.addField('signatures', 'string', { fieldName: 't_signatures' });
+
+		this.addField('signature', 'string', {}, stringToByte);
+
+		this.addField('signSignature', 'string', {}, stringToByte);
+
+		this.addField('signatures', 'string');
+
+		this.addField('asset', 'string');
```
shuse2: can we add support to do like `asset.dapp.name`? or maybe in next version?

storage/entities/transaction.js
```diff
@@ -684,6 +492,39 @@ class Transaction extends BaseEntity {
 
 		return filters;
 	}
+
+	static _sanitizeCreateData(data) {
+		const transactions = Array.isArray(data)
+			? _.cloneDeep(data)
+			: [_.cloneDeep(data)];
+
+		transactions.forEach(transaction => {
```
shuse2: instead of `forEach` map would be better?

storage/entities/transaction.js
```diff
@@ -684,6 +492,39 @@ class Transaction extends BaseEntity {
 
 		return filters;
 	}
+
+	static _sanitizeCreateData(data) {
```
shuse2: I would've preferred normal function instead of `static` but I see another `static` function above ='(

storage/sql/migrations/updates/20190214154800_remove_trs_dependant_tables.sql
```diff
@@ -14,17 +14,10 @@
...
-WITH duplicates AS
-  (
-    SELECT count(*)
-    FROM delegates
-    GROUP BY "transactionId"
-    HAVING count(*) > 1
-  )
-SELECT count(*)
-FROM duplicates
+DROP TABLE "transfer", "signatures", "delegates", "votes", "multisignatures", "dapps", "intransfer", "outtransfer" CASCADE;
```
MaciejBaj: It should be merged with the migration: `20190204170819_migrate_trs_dependant_tables_to_trs_asset_field.sql`.

Where is the migration of constant `duplicatedDelegates` then?

storage/entities/transaction.js
lsilvs: I reckon `forEach` is clearer when mutating the object

storage/entities/transaction.js
lsilvs: Out of scope of this PR imo.

storage/entities/transaction.js
shuse2: I guess the point is not to mutate and return the new object =)

lsilvs: But it's already cloning the object in the line before. So it's not mutating the passed param.

storage/utils/filters.js
```diff
@@ -30,50 +30,56 @@ function filterGenerator(
...
+		.split('.')
+		.map(f => `"${f}"`)
+		.join('.');
+
 	switch (filterType) {
 		case filterTypes.BOOLEAN:
-			filters[alias] = `"${fieldName}" = ${getValue(alias)}`;
```
MaciejBaj: Why are you removing `""` from here? I see you're still using different cases on `trs` column names (like "transferData" etc.), and that's the way to validate the case sensitivity.

storage/sql/migrations/updates/20190214154800_remove_trs_dependant_tables.sql
lsilvs: `duplicatedDelegates` was removed because it calculated duplicated `transactionId` in the `delegate` table. As `delegate` table was removed and `id` is uniq in `trs` table, there is no need for that check anymore.

Should I combine `20190214154800_remove_trs_dependant_tables.sql` into `20190204170819_migrate_trs_dependant_tables_to_trs_asset_field.sql`?

storage/utils/filters.js
lsilvs: The `""` are being added on the line 36.

storage/sql/migrations/updates/20190204170311_add_asset_field_as_json_to_trs_table.sql
```diff
@@ -12,4 +12,13 @@
...
+
+/*
+  DESCRIPTION: Add asset Column to trs.
+
+  PARAMETERS: None
+*/
+
+ALTER TABLE "trs" ADD COLUMN IF NOT EXISTS "asset" jsonb;
+
+CREATE INDEX IF NOT EXISTS "trs_asset" ON "trs" USING gin ("asset");
```
nazarhussain: `GIN` index is used explicitly for full text search. Are we going to use full-text search in `asset`? If not then I believe we can still use `btree`.

nazarhussain: Please communicate to `devops` team, maintaining `GIN` index can be quicker if we increase this configuration option in our PotgreSQL builds. 

https://www.postgresql.org/docs/9.1/runtime-config-resource.html#GUC-MAINTENANCE-WORK-MEM

storage/sql/migrations/updates/20190204170400_create_transfer_data_field_in_trs_table.sql
```diff
@@ -12,4 +12,11 @@
...
 
-INSERT INTO "votes" ("transactionId", "votes") VALUES ${values:raw};
+
+/*
+  DESCRIPTION: Add transfer data column to trs.
+
+  PARAMETERS: None
+*/
+
+ALTER TABLE "trs" ADD COLUMN IF NOT EXISTS "transferData" BYTEA;
```
nazarhussain: Migrations must be fully atomic, all the changes related to one particular thing must be in one migration file. We run single migration file in SQL transaction, so if something failed it should failed as a group. 

In this case we should do this steps in single migration file.  

1. We should create a new column "transferData"
2. Move old transfer data to new column 
3. Remove transfer table 

So as of now if that migration file run, we will be done fully what was required.

storage/sql/migrations/updates/20190204170819_migrate_trs_dependant_tables_to_trs_asset_field.sql
```diff
@@ -0,0 +1,99 @@
...
+*/
+
+-- Migrate transfer table into trs transferData field
+UPDATE trs
+SET "transferData" = (
+        SELECT t."data"
+        FROM transfer as t
+        WHERE t."transactionId" = trs.id
+    )
+WHERE id IN (SELECT "transactionId" FROM transfer);
```
nazarhussain: This part must be done in other migration.

storage/sql/migrations/updates/20190204170819_migrate_trs_dependant_tables_to_trs_asset_field.sql
nazarhussain: Please move 

> ALTER TABLE "trs" ADD COLUMN IF NOT EXISTS "asset" jsonb; 

to this migration and at the end of migration create the index on asset.

storage/sql/migrations/updates/20190204170819_migrate_trs_dependant_tables_to_trs_asset_field.sql
```diff
@@ -0,0 +1,99 @@
...
+
+
+-- Migrate signatures table into trs asset field
+UPDATE trs
+SET asset = (
+        SELECT concat('{"signature":{"publicKey":"',encode(s."publicKey",'hex'::text),'"}}')::json
```
nazarhussain: You try to use https://www.postgresql.org/docs/9.6/functions-string.html#FUNCTIONS-STRING-FORMAT to simplify the JSON string generation.

storage/sql/migrations/updates/20190214154800_remove_trs_dependant_tables.sql
nazarhussain: In such cases you can use `git rm` to mark some file is removed. Then git will not try to link up two files as rename.

storage/sql/transactions/count.sql
```diff
@@ -13,8 +13,8 @@
  */
 
 SELECT count(*)
-FROM
-	full_blocks_list
-
-WHERE "t_rowId" IS NOT NULL  ${parsedFilters:raw}
+FROM trs
+	LEFT JOIN blocks AS b ON trs."blockId"::text = b.id::text
```
nazarhussain: We don't need `::text` here.

storage/sql/transactions/count.sql
```diff
@@ -13,8 +13,8 @@
...
+	LEFT JOIN mem_accounts AS m ON trs."recipientId"::text = m.address::text
```
nazarhussain: We don't need `::text` here.

storage/sql/transactions/get.sql
```diff
@@ -13,32 +13,32 @@
...
-	LEFT JOIN blocks b ON (((fbl."b_id")::text = (b.id)::text)))
+	trs."id" AS "id",
+	b."height" AS "height",
+	trs."blockId" AS "blockId",
+	trs."type" AS "type",
+	trs."timestamp" AS "timestamp",
+	encode(trs."senderPublicKey", 'hex'::text) AS "senderPublicKey",
+	encode(m."publicKey", 'hex'::text) AS "recipientPublicKey",
+	upper(trs."senderId"::text) AS "senderId",
+	upper(trs."recipientId"::text) AS "recipientId",
```
nazarhussain: We were not using `upper` int he view before, why you are introducing it?

storage/sql/transactions/get.sql
```diff
@@ -13,32 +13,32 @@
...
+	regexp_split_to_array(trs."signatures", ',') as "signatures",
+	(( SELECT blocks.height + 1
+		FROM blocks
+		ORDER BY blocks.height DESC
+		LIMIT 1)) - b.height AS "confirmations"
 
-WHERE "t_rowId" IS NOT NULL  ${parsedFilters:raw}
+FROM trs
+	LEFT JOIN blocks b ON trs."blockId"::text = b.id::text
+	LEFT JOIN mem_accounts m ON trs."recipientId"::text = m.address::text
```
nazarhussain: You can remove `::text` everywhere in this file.

storage/sql/transactions/get_extended.sql
```diff
@@ -13,47 +13,33 @@
...
-	LEFT JOIN blocks b ON (((fbl."b_id")::text = (b.id)::text)))
+	trs."id" AS "id",
+	b."height" AS "height",
+	trs."blockId" AS "blockId",
+	trs."type" AS "type",
+	trs."timestamp" AS "timestamp",
+	encode(trs."senderPublicKey", 'hex'::text) AS "senderPublicKey",
+	encode(m."publicKey", 'hex'::text) AS "recipientPublicKey",
+	upper(trs."senderId"::text) AS "senderId",
+	upper(trs."recipientId"::text) AS "recipientId",
```
nazarhussain: Same as above.

storage/sql/migrations/updates/20190204170819_migrate_trs_dependant_tables_to_trs_asset_field.sql
lsilvs: I want `asset` field to be indexed. If I disable indexing before updating should I reindex after the migration?

storage/entities/transaction.js
```diff
@@ -180,96 +159,104 @@ class Transaction extends BaseEntity {
...
+
+		this.addField('signSignature', 'string', {}, stringToByte);
+
+		this.addField('signatures', 'string');
+
+		this.addField('asset', 'string');
 
 		this.addFilter('data_like', filterTypes.CUSTOM, {
-			condition: '"tf_data" LIKE ${data_like}',
+			condition: '"transferData" LIKE ${data_like}',
```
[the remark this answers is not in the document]
lsilvs: It's being converted in `_sanitizeFilters` method

storage/entities/transaction.js
```diff
@@ -592,15 +424,18 @@ class Transaction extends BaseEntity {
...
-			parsedOptions.sort = ['t_rowId:asc'];
+			parsedOptions.sort = ['rowId:asc'];
 		}
 
-		const parsedSort = this.parseSort(parsedOptions.sort);
+		let parsedSort = this.parseSort(parsedOptions.sort);
+
+		// TODO: improve this logic
+		parsedSort = parsedSort.replace('"rowId"', 'trs."rowId"');
+		parsedSort = parsedSort.replace('"dapp_name"', "asset->'dapp'->>'name'");
```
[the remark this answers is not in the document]
lsilvs: The first one to disambiguate `rowId` field and the second one to because there is no `dapp_name` field anymore.

storage/sql/migrations/updates/20190204170311_add_asset_field_as_json_to_trs_table.sql
lsilvs: I used `GIN` as per postgres documentation: `GIN indexes can be used to efficiently search for keys or key/value pairs occurring within a large number of jsonb documents`

REF.: https://www.postgresql.org/docs/10/datatype-json.html

storage/sql/transactions/get.sql
lsilvs: We were using it in `trs_list` view. Btw, it looks like we can get rid of `trs_list` view as well. What do you think?

storage/entities/transaction.js
```diff
@@ -616,57 +451,30 @@ class Transaction extends BaseEntity {
 				{ expectedResultCount },
 				tx
 			)
-			.then(data => {
+			.then(resp => {
+				const parseResponse = transaction => {
+					if (parsedOptions.extended) {
+						transaction.asset = transaction.asset ? transaction.asset : {};
+						if (transaction.transferData) {
```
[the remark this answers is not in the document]
lsilvs: I would still have to convert it using `toString('utf8')` and delete the key

storage/sql/transactions/get_extended.sql
lsilvs: Same answer: https://github.com/LiskHQ/lisk/pull/2891#discussion_r257741336

storage/sql/transactions/get.sql
nazarhussain: Yes if not in use drop that view.

PR #2892, opened 2019-02-15
Update checkTransactions function and use StateStore and lisk-element's transactions

### LINKED ISSUE #2821: Update checkTransactions function and use StateStore and lisk-element's transactions
### Expected behavior
Use StateStore and transactions defined in lisk-elements for verifying transactions
### Actual behavior
We are missing this functionality right now.

SargeKhan:
### What was the problem?
checkTransactions function was using logic/transactions to verify transactions.
### How did I fix it?
Refactored the function to use lisk-transactions module along with state store.
### How to test it?
Run tests.
### Review checklist

* The PR resolves #2821 
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/chain/modules/blocks/verify.js
+++ b/framework/src/modules/chain/modules/blocks/verify.js
@@ -16,6 +16,7 @@
 
 const crypto = require('crypto');
 const async = require('async');
+const TransactionStatus = require('@liskhq/lisk-transactions').Status;
 const BlockReward = require('../../logic/block_reward.js');
 const slots = require('../../helpers/slots.js');
 const blockVersion = require('../../logic/block_version.js');
@@ -71,11 +72,11 @@ class Verify {
 }
 
 /**
- * Check transaction - perform transaction validation when processing block.
+ * Check transactions - perform transactions validation when processing block.
  * FIXME: Some checks are probably redundant, see: logic.transactionPool
  *
  * @private
- * @func checkTransaction
+ * @func checkTransactions
  * @param {Object} block - Block object
  * @param {Object} transaction - Transaction object
  * @param  {boolean} checkExists - Check if transaction already exists in database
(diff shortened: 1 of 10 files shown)


framework/src/modules/chain/modules/blocks/verify.js
```diff
@@ -16,6 +16,7 @@
 
 const crypto = require('crypto');
 const async = require('async');
+const TransactionStatus = require('@liskhq/lisk-transactions').Status;
```
shuse2: prefer below?
```
const { Status: TransactionStatus } = require('@liskhq/lisk-transactions');
```

framework/src/modules/chain/modules/blocks/verify.js
```diff
@@ -71,71 +72,50 @@ class Verify {
...
+			);
+			throw `Transaction is already confirmed: ${persistedTransactions[0].id}`;
 		}
+	}
+
+	const transactionResponses = await modules.processTransactions.verifyTransactions(
+		transactions
 	);
+
+	const unverifiableTransactionResponse = transactionResponses.find(
```
shuse2: shouldn't this be filter? also, `unverifiedTransactionResponse`?

SargeKhan: This is not a filter because I am keeping the same functionality as the existing implementation.

framework/src/modules/chain/modules/blocks/verify.js
```diff
@@ -71,71 +72,50 @@ class Verify {
...
-				// Undo the offending transaction.
-				// DATABASE: write
-				return modules.transactions.undoUnconfirmed(
-					transaction,
-					undoUnconfirmedErr => {
-						modules.transactions.removeUnconfirmedTransaction(transaction.id);
-						return setImmediate(cb, undoUnconfirmedErr || waterCbErr);
-					}
-				);
+__private.checkTransactions = async (transactions, checkExists) => {
```
mitsujutsu: should we rename to `verifyTransactions` as that seems to be the main function (also the description above)

SargeKhan: It is verifying transactions as well as checking whether they are confirmed or not. That is the reason why I am reluctant about changing the function naming.

shuse2: I think id exists on confirmed state is a step in verify phase, but let's address in another issue

PR #2893, opened 2019-02-15
Validate system constants - Closes #2889

### LINKED ISSUE #2889: Create controller/schemas/constants.js to validate system wide constants
### Expected behavior
The constants passed to the application must be validated first.

### Actual behavior
Currently, we are not validating these constants.

1.6.0

michielmulders:
### What was the problem?
No validation present for system-wide constants when instantiating the app.

### How did I fix it?
Add JSON Schema validation.

### How to test it?
Run `node index -n devnet` or use any other network, inside of `/src` folder.

### Review checklist

* The PR resolves #2889 
* Linting passes
* Tests pass


framework/src/controller/schema/application.js
```diff
@@ -13,6 +13,211 @@ module.exports = {
 	constants: {
 		id: '#/app/constants',
```
nazarhussain: This is huge and major application element, so create it as `schema/constants.js` file.

framework/src/controller/schema/application.js
```diff
@@ -13,6 +13,211 @@ module.exports = {
...
+				required: ['MIN_LENGTH', 'MAX_LENGTH'],
+				properties: {
+					MIN_LENGTH: {
+						type: 'integer',
+					},
+					MAX_LENGTH: {
+						type: 'integer',
+					},
+				},
+			},
```
nazarhussain: Please extract this nested objects and keep them on top level and then reference it. This way its better readable.

framework/src/controller/schema/application.js
```diff
@@ -13,6 +13,211 @@ module.exports = {
...
+							MIN_ITEMS: {
+								type: 'integer',
+							},
+							MAX_ITEMS: {
+								type: 'integer',
+							},
+						},
+					},
+				},
+			},
```
nazarhussain: Same here for `MULTISIG_CONSTRAINTS`, extract it as top level object and refer it to constants.

framework/src/controller/schema/application.js
```diff
@@ -13,6 +13,211 @@ module.exports = {
...
+			'NORMALIZER',
+			'REWARDS',
+			'TOTAL_AMOUNT',
+			'UNCONFIRMED_TRANSACTION_TIMEOUT',
+			'EXPIRY_INTERVAL',
+		],
+		properties: {
+			ACTIVE_DELEGATES: {
+				type: 'integer',
+				min: 0,
```
nazarhussain: I think this must not be zero in any case. There would be a minimum value for it @MaciejBaj @4miners Please confirm. 

I am not sure if that's the one or `MIN_BROADHASH_CONSENSUS`, but any of these two must be minimum 51.

framework/src/controller/schema/application.js
```diff
@@ -13,6 +13,211 @@ module.exports = {
...
+			},
+			BLOCK_SLOT_WINDOW: {
+				type: 'integer',
```
nazarhussain: We had tested all of our application as per `5` second slow window. I think we should set it as a `min` limit. Unless we successfully test our blockchain with lower slot window we should not allow users to do it. 

@MaciejBaj Please confirm.

framework/src/controller/schema/application.js
```diff
@@ -13,6 +13,211 @@ module.exports = {
...
+			},
+			ADDITIONAL_DATA: {
+				type: 'object',
+				required: ['MIN_LENGTH', 'MAX_LENGTH'],
+				properties: {
+					MIN_LENGTH: {
+						type: 'integer',
```
nazarhussain: Min length `min` must be set to `1`

framework/src/controller/schema/application.js
```diff
@@ -13,6 +13,211 @@ module.exports = {
...
+					MAX_LENGTH: {
+						type: 'integer',
+					},
+				},
+			},
+			BLOCK_RECEIPT_TIMEOUT: {
+				type: 'integer',
+			},
+			EPOCH_TIME: {
+				type: 'object',
```
nazarhussain: We should make it a string and add use format called `timestamp`

framework/src/controller/schema/application.js
```diff
@@ -13,6 +13,211 @@ module.exports = {
...
+				pattern: '^[0-9]+$',
+			},
+			DAPP_WITHDRAWAL: {
+				type: 'string',
+				pattern: '^[0-9]+$',
+			},
+			DAPP_DEPOSIT: {
+				type: 'string',
+				pattern: '^[0-9]+$',
+			},
```
nazarhussain: Please add a string format `bignumber` and then set format everywhere in schema. Don't use redundant patterns.

framework/src/controller/schema/application.js
```diff
@@ -13,6 +13,211 @@ module.exports = {
...
+			EPOCH_TIME: {
+				type: 'object',
+			},
+			FEES: {
+				type: 'object',
+				$ref: '#/fees',
+			},
+			MAX_PAYLOAD_LENGTH: {
+				type: 'integer',
```
nazarhussain: There must be a minimum value for it. Please refer to https://lisk.io/documentation/lisk-protocol/transactions to calculate what the minimum a single transactions can have.

framework/src/controller/schema/application.js
```diff
@@ -13,6 +13,211 @@ module.exports = {
...
+			},
+			MAX_PEERS: {
+				type: 'integer',
+			},
+			MAX_SHARED_TRANSACTIONS: {
+				type: 'integer',
+			},
+			MAX_TRANSACTIONS_PER_BLOCK: {
+				type: 'integer',
```
nazarhussain: Set `min` to `1`, Block with zero transactions does not make sense.

framework/src/controller/schema/application.js
```diff
@@ -13,6 +13,211 @@ module.exports = {
...
+			},
+			MAX_VOTES_PER_TRANSACTION: {
+				type: 'integer',
```
nazarhussain: Here as well set min to `1`

framework/src/controller/schema/application.js
```diff
@@ -13,6 +13,211 @@ module.exports = {
...
+			MULTISIG_CONSTRAINTS: {
+				type: 'object',
+				required: ['MIN', 'LIFETIME', 'KEYSGROUP'],
+				properties: {
+					MIN: {
+						type: 'object',
+						required: ['MINIMUM', 'MAXIMUM'],
+						properties: {
+							MINIMUM: {
+								type: 'integer',
```
nazarhussain: Minimum can not be zero or negative.

framework/src/controller/schema/application.js
```diff
@@ -13,6 +13,211 @@ module.exports = {
...
+					},
+					KEYSGROUP: {
+						type: 'object',
+						required: ['MIN_ITEMS', 'MAX_ITEMS'],
+						properties: {
+							MIN_ITEMS: {
+								type: 'integer',
+							},
+							MAX_ITEMS: {
+								type: 'integer',
```
nazarhussain: As of now `MULTISIG_CONSTRAINTS.MIN.MAXIMUM` can't be greater than `MULTISIG_CONSTRAINTS.KEYSGROUP.MAX_ITEMS`. Check if there is away to refer this constraint in the json-schema.

framework/src/controller/schema/application.js
```diff
@@ -13,6 +13,211 @@ module.exports = {
...
+							},
+						},
+					},
+				},
+			},
+			NETHASHES: {
+				type: 'array',
+				items: {
+					type: 'string',
+					pattern: '^[a-fA-F0-9]{64}$',
```
nazarhussain: We don't allow capital letters in any of hex strings. Please refer to `helpers/z_schema.formats.hex` for it. And wherever possible use the already defined formats.

framework/src/controller/schema/application.js
```diff
@@ -13,6 +13,211 @@ module.exports = {
...
+				},
+			},
+			NORMALIZER: {
+				type: 'string',
+				pattern: '^[0-9]+$',
```
nazarhussain: Again its a big number amount. Use existing format or define new one.

framework/src/controller/schema/application.js
```diff
@@ -13,6 +13,211 @@ module.exports = {
...
+			},
+			TOTAL_AMOUNT: {
+				type: 'string',
+				pattern: '^[0-9]+$',
+			},
+			UNCONFIRMED_TRANSACTION_TIMEOUT: {
+				type: 'integer',
+			},
+			EXPIRY_INTERVAL: {
+				type: 'integer',
```
nazarhussain: Must not be negative or zero.

framework/src/controller/schema/application.js
```diff
@@ -13,6 +13,211 @@ module.exports = {
...
+						type: 'integer',
+					},
+				},
+			},
+			TOTAL_AMOUNT: {
+				type: 'string',
+				pattern: '^[0-9]+$',
+			},
+			UNCONFIRMED_TRANSACTION_TIMEOUT: {
+				type: 'integer',
```
nazarhussain: Must not be negative or zero.

framework/src/controller/schema/application.js
michielmulders: I should use `format: 'date'` here, but that only works when the Date object is passed as a string. What do you suggest? @nazarhussain

framework/src/controller/schema/application.js
MaciejBaj: There are the conditions which definitely will not work with ACTIVE_DELEGATES equal to 0, like:
```
// Check if we are one block before the last block of a round, if yes - perform round snapshot
				if ((block.height + 1) % ACTIVE_DELEGATES === 0) {
```

The value here should be min 1 and should always be odd. That's what is missing currently. 

While setting new parameters range each of required constants what should be inspected before is if the change of the parameter is affecting the application behaviour. 

@michielmulders please perform the experiments and see if the application changes the behaviour as expected after change of the most significant parameters:
1. ACTIVE_DELEGATES
2. BLOCK_SLOT_WINDOW
3. BLOCK_RECEIPT_TIMEOUT
4. EPOCH_TIME
5. MAX_PAYLOAD_LENGTH
6. MAX_TRANSACTIONS_PER_BLOCK
7. NORMALIZER

framework/src/controller/schema/application.js
michielmulders: Can reference value within the same schema with $data property, but not outside in a higher scope.

michielmulders: Decided to not verify this for now as it is subject to change.

framework/src/controller/schema/application.js
michielmulders: Can use `$data` to reference another property.
However, this is only possible within the same scope, whereas `MAX_ITEMS` is defined somewhere in another scheme higher up.

framework/src/controller/schema/application.js
michielmulders: Cannot convert string to number.
Found replacement regex that checks if a number does not start with a zero or dash for negative numbers.
`^[1-9][0-9]*$`

framework/src/controller/schema/application.js
michielmulders: JSON Schema does not support detection for odd/even numbers.
For strings, we can use this regex that detects odd numbers: `^\d*[13579]$`

framework/src/controller/schema/constants.js
```diff
@@ -0,0 +1,265 @@
...
+				type: 'object',
+				schema: {
+					$ref: '#keysgroupConstraints',
+				},
+			},
+		},
+	},
+
+	minConstraints: {
+		$id: '#minConstraints',
```
nazarhussain:
```suggestion
		$id: '#/constants/multisig/min,
```

framework/src/controller/schema/constants.js
```diff
@@ -0,0 +1,265 @@
...
+				// e.g. max: `{ $data: '1/path-to-MAX_ITEMS'}`
+			},
+			MAXIMUM: {
+				type: 'integer',
+			},
+		},
+	},
+
+	lifetimeConstraints: {
+		$id: '#lifetimeConstraints',
```
nazarhussain:
```suggestion
		$id: '#/constants/multisig/lifetime',
```

framework/src/controller/schema/application.js
nazarhussain: As discussed create a seperate issue for this change and we will tackle it separately.

nazarhussain:
@michielmulders Please look into tests, seems they started failing now.

shuse2:
we require a lot of specific format for validation.
Why don't we create validation folder (instead of putting in `helper`), and add custom format there to be used in the schema validation along with schema definition?
ie)
```
/controller
  /validation
    validator.js
    format.js
    schema_application.js (or schema/application.js)
    schema_config.js
```
```

File: framework/src/controller/helpers/validator.js [hunk 20]
```diff
@@ -1,7 +1,12 @@
 const Ajv = require('ajv');
 const { SchemaValidationError } = require('../../errors');
 
+const zSchema = require('../../modules/chain/helpers/z_schema');
```
shuse2: we use zSchema and Ajv...?

michielmulders: @shuse2 For now, we are using both yes, as z-schema allows to define custom validators and we can load those as a format with Ajv. In the future, this will be modified to fit one, this has been discussed and decided to do it this way. Please approve if the code looks good now.

framework/src/controller/schema/constants.js
```diff
@@ -0,0 +1,280 @@
...
+	minConstraints: {
+		$id: '#/constants/multisig/min',
+		type: 'object',
+		required: ['MINIMUM', 'MAXIMUM'],
+		properties: {
+			MINIMUM: {
+				type: 'integer',
+				min: 1,
+				// Max value is lower than or equal to MULTISIG_CONSTRAINTS.KEYSGROUP.MAX_ITEMS
+				// Problem: Cannot reference value outside of this object
```
MaciejBaj:
```suggestion
```

framework/src/controller/schema/constants.js
```diff
@@ -0,0 +1,280 @@
...
+				// e.g. max: `{ $data: '1/path-to-MAX_ITEMS'}`
```
MaciejBaj:
```suggestion
```

framework/src/controller/helpers/validator.js
```diff
@@ -1,7 +1,12 @@
...
+
 const validator = new Ajv({ allErrors: true, schemaId: 'auto' });
+Object.keys(zSchema.formatsCache).forEach(zSchemaType => {
```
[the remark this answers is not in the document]
nazarhussain: Just the clarification, these types are not related to z-schema library. These are general functions matcher functions and most of validation libraries support this syntax.

framework/src/controller/schema/constants.js
```diff
@@ -0,0 +1,265 @@
...
+			'EXPIRY_INTERVAL',
+		],
+		properties: {
+			ACTIVE_DELEGATES: {
+				type: 'integer',
+				min: 51,
+			},
+			BLOCK_SLOT_WINDOW: {
+				type: 'integer',
+				min: 5,
```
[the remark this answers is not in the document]
michielmulders: Tested this, can be 1 without problems

framework/src/controller/schema/constants.js
```diff
@@ -0,0 +1,265 @@
...
+			'NORMALIZER',
+			'REWARDS',
+			'TOTAL_AMOUNT',
+			'UNCONFIRMED_TRANSACTION_TIMEOUT',
+			'EXPIRY_INTERVAL',
+		],
+		properties: {
+			ACTIVE_DELEGATES: {
+				type: 'integer',
+				min: 51,
```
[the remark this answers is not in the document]
michielmulders: This is already addressed, it is set to 1 and odd.

PR #2894, opened 2019-02-15
API Performance degrade - Closes #2817

### LINKED ISSUE #2817: API Performance degrade
### Expected behavior
API Performance should be tolerant`(100-600ms)`.
### Actual behavior
Following API Performance have been degraded
`GET dapps`
`POST signatures`
`GET voters`
`GET accounts`
### Steps to reproduce
https://rpm.newrelic.com/accounts/2081667/applications/130457745/transactions?tw%5Bend%5D=1549281180&tw%5Bstart%5D=1549259580
1.5.0

lsilvs:
### What was the problem?
API was using `extended` version of accounts

### How did I fix it?
I moved `productivity` to the non-extended version and removed the `extended` from API calls

### How to test it?
Run test suite

### Review checklist

* The PR resolves #2817 
* Linting passes
* Tests pass

### DIFF
--- a/storage/sql/accounts/get.sql
+++ b/storage/sql/accounts/get.sql
@@ -28,7 +28,14 @@ SELECT
 	"rank",
 	"fees",
 	"rewards",
-	"vote"
+	"vote",
+	"u_balance",
+	case
+	when
+		"producedBlocks" + "missedBlocks" = 0 then 0
+	else
+		(("producedBlocks"::float / ("producedBlocks" + "missedBlocks")) * 100.0)::integer
+	end AS productivity
 FROM
 	mem_accounts
 
--- a/storage/sql/accounts/get.sql
+++ b/storage/sql/accounts/get.sql
@@ -29,6 +29,12 @@ SELECT
 	"fees",
 	"rewards",
 	"vote",
+	"u_username",
+	"u_isDelegate"::int::boolean,
+	"u_secondSignature"::int::boolean,
+	"u_nameexist"::int::boolean as "u_nameExist",
+	"u_multimin" as "u_multiMin",
+	"u_multilifetime" as "u_multiLifetime",
 	"u_balance",
 	case
 	when
--- a/test/unit/storage/entities/account.js
+++ b/test/unit/storage/entities/account.js
@@ -180,6 +180,13 @@ describe('Account', () => {
 			'fees',
 			'rewards',
 			'vote',
+			'u_username',
+			'u_isDelegate',
+			'u_secondSignature',
+			'u_nameExist',
+			'u_multiMin',
+			'u_multiLifetime',
+			'u_balance',
 			'productivity',
 		];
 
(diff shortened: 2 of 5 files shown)


test/unit/storage/entities/account.js
```diff
@@ -180,6 +180,14 @@ describe('Account', () => {
...
 			'rewards',
 			'vote',
+			'u_username',
+			'u_isDelegate',
+			'u_secondSignature',
+			'u_nameExist',
+			'u_multiMin',
+			'u_multiLifetime',
+			'u_balance',
+			'productivity',
```
MaciejBaj: returning productivity for every account?

storage/sql/accounts/get.sql
```diff
@@ -28,7 +28,20 @@ SELECT
...
+	"u_secondSignature"::int::boolean,
+	"u_nameexist"::int::boolean as "u_nameExist",
+	"u_multimin" as "u_multiMin",
+	"u_multilifetime" as "u_multiLifetime",
+	"u_balance",
+	case
+	when
+		"producedBlocks" + "missedBlocks" = 0 then 0
+	else
+		(("producedBlocks"::float / ("producedBlocks" + "missedBlocks")) * 100.0)::integer
```
MaciejBaj: Then the logic of logic/accounts.js, `calculateProductivity` function is redundant.

storage/sql/accounts/get.sql
```diff
@@ -28,7 +28,20 @@ SELECT
...
+	"u_username",
+	"u_isDelegate"::int::boolean,
+	"u_secondSignature"::int::boolean,
+	"u_nameexist"::int::boolean as "u_nameExist",
+	"u_multimin" as "u_multiMin",
+	"u_multilifetime" as "u_multiLifetime",
+	"u_balance",
+	case
+	when
+		"producedBlocks" + "missedBlocks" = 0 then 0
```
MaciejBaj:
```suggestion
		"isDelegate" = 0 then 0
```

test/unit/storage/entities/account.js
lsilvs: It's used for delegate accounts. It was already present in `extended` accounts but as the calculation is not costly I decide to move to `basic` accounts.

storage/sql/accounts/get.sql
lsilvs: `calculateProductivity` was removed as part of 1.5 development

storage/sql/accounts/get.sql
lsilvs: If the a delegate account has `producedBlocks` and `missedBlocks` equal to zero, the `else` statement will fail due to division by zero.

nazarhussain: I tend to believe on our application logic that produced blocks and missed blocks will only be incremented for delegate accounts.

lsilvs: But you can still have a delegate account with `producedBlocks` and `missedBlocks` equal to zero

PR #2900, opened 2019-02-18
Fix intermittent test failure for database unit tests - Closes #2834

### LINKED ISSUE #2834: Fix intermittent test failure for database unit tests
### Expected behavior
Mem_round_snapshot shouldn't be present.
### Actual behavior
An error is being thrown when running database unit tests stating that `mem_round_snapshot` relation already exists. 
Please [check](https://jenkins.lisk.io/blue/organizations/jenkins/lisk-core/detail/PR-2804/5/pipeline)
```
1) Round
       checkSnapshotAvailability()
         should use the correct SQL file with one parameter:
     error: relation "mem_round_snapshot" already exists
```
### Steps to reproduce
Run unit tests
1.5

yatki:
### What was the problem?
We were randomly getting the error below:

```
1) Round
       checkSnapshotAvailability()
         should use the correct SQL file with one parameter:
     error: relation "mem_round_snapshot" already exists
```
### How did I fix it?
I tagged this test suite with `@sequential` tag to solve the conflicts.

### Review checklist

* The PR resolves #2834 
* Linting passes
* Tests pass

### DIFF
--- a/framework/test/unit/components/storage/entities/round.js
+++ b/framework/test/unit/components/storage/entities/round.js
@@ -20,7 +20,7 @@ const {
 	BaseEntity,
 	Round,
 } = require('../../../../../src/components/storage/entities');
-const storageSandbox = require('../../../../common/storage_sandbox');
+const { StorageSandbox } = require('../../../../common/storage_sandbox');
 const seeder = require('../../../../common/storage_seed');
 const accountsFixtures = require('../../../../fixtures').accounts;
 const roundsFixtures = require('../../../../fixtures').rounds;
@@ -32,9 +32,9 @@ const {
 const checkTableExists = (adapter, tableName) =>
 	adapter
 		.execute(
-			`SELECT EXISTS 
+			`SELECT EXISTS
 					(
-						SELECT 1 
+						SELECT 1
 						FROM pg_tables
 						WHERE schemaname = 'public'
 						AND tablename = '${tableName}'
@@ -42,7 +42,7 @@ const checkTableExists = (adapter, tableName) =>
 		)
 		.then(result => result[0].exists);
 
-describe('Round', () => {
+describe('Round @sequential', () => {
 	let adapter;
 	let storage;
 	let RoundEntity;
@@ -105,10 +105,7 @@ describe('Round', () => {
 	const validFilter = { round: validRound.round };
 
 	before(async () => {
-		storage = new storageSandbox.StorageSandbox(
-			__testContext.config.db,
-			'lisk_test_round'
-		);
+		storage = new StorageSandbox(__testContext.config.db, 'lisk_test_round');
 		await storage.bootstrap();
 
 		adapter = storage.adapter;


framework/test/unit/components/storage/entities/round.js
```diff
@@ -678,7 +678,7 @@ describe('Round', () => {
 	});
 
 	describe('performRoundSnapshot()', () => {
-		afterEach(() => RoundEntity.clearRoundSnapshot());
+		afterEach(async () => RoundEntity.clearRoundSnapshot());
```
diego-G: There are more similar `afterEach` and also `beforeEach` than these that you modify. Why not adding `async` to all of them?

yatki: I just focused on solving this problem. Please feel free to create an issue to make sure we are handling all Promises correctly.

MaciejBaj:
@diego-G `storage/entities/round.js` is the only file where `RoundEntity.clearRoundSnapshot` is used. It does solve the "intermittent test failure for database unit tests" issue. 

If you see similar problems for other async functions being called directly in afterEach and beforeEach which are causing a problem please open a corresponding issue.

framework/test/unit/components/storage/entities/round.js
```diff
@@ -224,11 +224,11 @@ describe('Round', () => {
 
 	describe('_getResults()', () => {
 		beforeEach(async () => {
-			await storage.entities.Round.create(validRound);
+			return storage.entities.Round.create(validRound);
```
nazarhussain: How is it effecting any change? If both have different behavior then we have to look carefully on other places as well.

framework/test/unit/components/storage/entities/round.js
```diff
@@ -117,9 +117,9 @@ describe('Round', () => {
 		SQLs = RoundEntity.SQLs;
 	});
 
-	beforeEach(() => seeder.seed(storage));
+	beforeEach(async () => seeder.seed(storage));
 
-	afterEach(() => {
+	afterEach(async () => {
```
nazarhussain: If that's the problem then we are doing same on many other places as well. But personally I don't feel this is the issue. Mocha treat a function which return Promise as builtin feature.

framework/test/unit/components/storage/entities/round.js
```diff
@@ -972,7 +972,7 @@ describe('Round', () => {
 	});
 
 	describe('performVotesSnapshot()', () => {
-		afterEach(() => RoundEntity.clearVotesSnapshot());
+		afterEach(async () => RoundEntity.clearVotesSnapshot());
```
nazarhussain: `RoundEntity.clearVotesSnapshot()` returnes a Promise, and mocha support to handle promise returned from function. Why we had to explicitly declare it as `async`?

nazarhussain: That applies to every other change in this file.

yatki:
@MaciejBaj @diego-G https://github.com/LiskHQ/lisk/issues/2901 created this issue for following up this problem.

framework/test/unit/components/storage/entities/round.js
yatki: This is just for convenience. When you return a Promise in an async function it's gonna be chained and awaited as well.

framework/test/unit/components/storage/entities/round.js
yatki: @nazarhussain You are right here. I'll revert these changes.

framework/test/unit/components/storage/entities/round.js
```diff
@@ -271,11 +268,11 @@ describe('Round', () => {
 
 	describe('isPersisted()', () => {
 		beforeEach(async () => {
-			await RoundEntity.create(validRound);
+			return RoundEntity.create(validRound);
 		});
 
 		afterEach(async () => {
-			await RoundEntity.delete(validFilter);
+			return RoundEntity.delete(validFilter);
```
nazarhussain: Some how this change makes me feel that issue was with the use of `await` rather just the `@sequential` tag. 

Please revert that change and we are good to go with it.

PR #2906, opened 2019-02-18
Remove tables and columns related to unconfirmed state from the database - Closes #2827

### LINKED ISSUE #2827: Remove tables and columns related to unconfirmed state from the database
### Expected behavior
Transactions will be applied to the unconfirmed state in memory. Therefore, the unconfirmed tables and columns are not needed in the database and should be removed.
### Actual behavior
The transactions are applied to the unconfirmed state and the tables and columns for managing unconfirmed state exist in the database.
1.6.0

pablitovicente:
### What was the problem?

Unconfirmed state will be handled in memory with the new transactions implementation so we needed to remove unconfirmed state columns, tables, triggers, etc. 

### How did I fix it?

- By removing the tables and adjusting the code.
- I added the tag [UNCONFIRMED_STATE_REMOVAL] to all skipped tests so we can revisit them once we are done implementing the new transactions.

### How to test it?

Build should be green.

### Review checklist

* The PR resolves #2827
* Linting passes
* Tests pass


framework/src/components/storage/sql/migrations/updates/20190215094703_remove_uncofirmed_state.sql
```diff
@@ -0,0 +1,35 @@
+/*
+ * Copyright © 2018 Lisk Foundation
```
shuse2: probably filename for this needs to be updated later according to 1.6

framework/src/modules/chain/logic/delegate.js
```diff
@@ -247,8 +247,8 @@ Delegate.prototype.checkConfirmed = function(transaction, cb, tx) {
 Delegate.prototype.checkUnconfirmed = function(transaction, cb, tx) {
```
shuse2: probably comment this as well to `[UNCONFIRMED_STATE_REMOVAL]`?

framework/src/modules/chain/logic/delegate.js
```diff
@@ -359,8 +358,8 @@ Delegate.prototype.applyUnconfirmed = function(transaction, sender, cb, tx) {
 Delegate.prototype.undoUnconfirmed = function(transaction, sender, cb, tx) {
```
shuse2: as as above

framework/src/modules/chain/logic/transaction.js
```diff
@@ -1003,12 +1003,8 @@ class Transaction {
 	 * @returns {SetImmediate} error
 	 * @todo Add description for the params
 	 */
-	applyUnconfirmed(transaction, sender, requester, cb, tx) {
+	applyUnconfirmed(transaction, sender, requester, cb) {
```
shuse2: same as above

framework/src/modules/chain/logic/transaction.js
```diff
@@ -1032,34 +1028,39 @@ class Transaction {
...
-					tx
-				);
-			},
-			tx
+		this.scope.logger.error(
+			'[UNCONFIRMED_STATE_REMOVAL] Skipping acount merge when trying to modify u_balance'
 		);
+		return setImmediate(cb);
+
+		// return this.scope.account.merge(
```
shuse2: why keep this as comment?

framework/src/components/storage/sql/migrations/updates/20190215094703_remove_uncofirmed_state.sql
pablitovicente: You mean due to the new migrations added recently right?

shuse2: yes!

pablitovicente: We will need to deal with this once we start merges of ongoing PRs I think. What you think?

framework/src/modules/chain/api/controllers/accounts.js
```diff
@@ -45,11 +45,10 @@ function accountFormatter(totalSupply, account) {
 		'address',
 		'publicKey',
 		'balance',
-		'u_balance',
 		'secondPublicKey',
 	]);
-	object.unconfirmedBalance = object.u_balance;
```
SargeKhan: The API will expose the unconfirmedBalance property because this feature is developed in a minor release. Therefore, the unconfirmedBalance should equal the confirmed balance (which will usually be the case when the account does not have any transactions in the pool)

framework/src/modules/chain/api/controllers/accounts.js
```diff
@@ -136,7 +135,7 @@ async function multiSigAccountFormatter(account) {
 	]);
 	result.min = account.multiMin;
 	result.lifetime = account.multiLifetime;
-	result.unconfirmedBalance = account.u_balance;
```
SargeKhan: same here.

framework/src/modules/chain/api/controllers/accounts.js
pablitovicente: Fixed.

framework/src/modules/chain/api/controllers/accounts.js
pablitovicente: Fixed.

framework/src/components/storage/sql/migrations/updates/20190215094703_remove_uncofirmed_state.sql
pablitovicente: Resolving this one. I will create an issue for resolving migration naming that should be worked on before 1.7 release.

PR #2909, opened 2019-02-19
Convert to async update function system component - Closes #2904

### LINKED ISSUE #2904: Convert to async update function from system component
### Expected behavior

Components are classes using no callbacks. 

### Actual behavior

System component has a function that is receiving a callback as parameter.

`1.6`

diego-G:
### Review checklist

* The PR resolves #2904
* Linting passes
* Tests pass


framework/src/components/system/system.js
```diff
@@ -115,36 +115,34 @@ class System {
...
+				.createHash('sha256')
+				.update(seed, 'utf8')
+				.digest()
+				.toString('hex');
+			this.headers.broadhash = newBroadhash;
+			this.logger.debug('System headers', this.headers);
+			return null;
+		} catch (err) {
+			this.logger.error(err.stack);
+			return err;
```
limiaspasdaniel: Shouldn't this be `throw new Error(err)`?

framework/src/modules/chain/modules/blocks/chain.js
```diff
@@ -836,7 +836,7 @@ Chain.prototype.deleteLastBlock = function(cb) {
 			},
 			updateSystemHeaders(seriesCb) {
 				// Update our own headers: broadhash and height
-				components.system.update(seriesCb);
+				return components.system.update().then(() => setImmediate(seriesCb));
```
limiaspasdaniel: I think this can be written as follows:
```
await components.system.update();
return setImmediate(seriesCb));
```

framework/src/modules/chain/modules/blocks/verify.js
```diff
@@ -835,9 +835,10 @@ Verify.prototype.processBlock = function(block, broadcast, saveBlock, cb) {
 			// 'false' if block comes from chain synchronisation process
 			updateSystemHeaders(seriesCb) {
 				// Update our own headers: broadhash and height
-				!library.config.loading.snapshotRound
-					? components.system.update(seriesCb)
-					: seriesCb();
+				if (!library.config.loading.snapshotRound) {
+					return components.system.update().then(() => setImmediate(seriesCb));
+				}
+				return setImmediate(seriesCb);
```
limiaspasdaniel: This could be simplified this way:
```
if (!library.config.loading.snapshotRound) {
  await components.system.update();
}
return setImmediate(seriesCb);
```

framework/src/modules/chain/modules/loader.js
```diff
@@ -955,7 +955,7 @@ __private.sync = function(cb) {
 			},
 			updateSystemHeaders(seriesCb) {
 				// Update our own headers: broadhash and height
-				components.system.update(seriesCb);
+				return components.system.update().then(() => setImmediate(seriesCb));
```
limiaspasdaniel: Same as above here.

framework/src/modules/chain/modules/blocks/chain.js
diego-G: We cannot use `await` because the function is not `async`.

framework/src/modules/chain/modules/blocks/verify.js
diego-G: Same as above.

framework/src/components/system/system.js
diego-G: We don't have a standard way to throw errors. We just follow the rule of handling them locally.

lsilvs: The code was returning `setImmediate(cb, err)` here before so I reckon you should throw an error here and catch it when using in order to call `setImmediate(seriesCb, err)`

framework/src/components/system/system.js
```diff
@@ -115,36 +115,34 @@ class System {
...
+			this.headers.height = blocks[0].height;
+			const seed = blocks.map(row => row.id).join('');
+			const newBroadhash = crypto
+				.createHash('sha256')
+				.update(seed, 'utf8')
+				.digest()
+				.toString('hex');
+			this.headers.broadhash = newBroadhash;
+			this.logger.debug('System headers', this.headers);
+			return null;
```
lsilvs: As we will be using it as `const result = await systemComponent.update()`, consider returning `this.headers` or `true` here.

framework/src/components/system/system.js
diego-G: Done.

framework/src/components/system/system.js
```diff
@@ -114,37 +114,35 @@ class System {
 	/**
 	 * Updates private broadhash and height values.
 	 *
-	 * @param {function} cb - Callback function
-	 * @returns {setImmediateCallback} cb, err
+	 * @returns {Promise.<boolean, Error>}
```
shuse2: it should be `Promise<boolean | Error>`, but I think error should be handled in each places (it seems like anyway it's doing so)
So prefer to have Promise<void> here

framework/src/modules/chain/modules/blocks/verify.js
```diff
@@ -835,9 +835,13 @@ Verify.prototype.processBlock = function(block, broadcast, saveBlock, cb) {
...
 			updateSystemHeaders(seriesCb) {
 				// Update our own headers: broadhash and height
-				!library.config.loading.snapshotRound
-					? components.system.update(seriesCb)
-					: seriesCb();
+				if (!library.config.loading.snapshotRound) {
+					return components.system
+						.update()
+						.then(() => setImmediate(seriesCb))
+						.catch(seriesCb);
```
shuse2: why then and catch has different syntax (one is using then setImmediate, and one not

framework/src/modules/chain/modules/blocks/chain.js
```diff
@@ -836,7 +836,10 @@ Chain.prototype.deleteLastBlock = function(cb) {
 			},
 			updateSystemHeaders(seriesCb) {
 				// Update our own headers: broadhash and height
-				components.system.update(seriesCb);
+				return components.system
+					.update()
+					.then(() => setImmediate(seriesCb))
+					.catch(seriesCb);
```
MaciejBaj: What is the rationale behind the change of `update` to async? I see the code only gets more complicated. The `system.update` is only used in the surroundings of `async` library for now.

framework/src/components/system/system.js
diego-G: I've followed the same pattern suggested by @lsilvs . It is been used in storage entities.

framework/src/modules/chain/modules/blocks/verify.js
diego-G: I'm gonna remove `setInmediate`

framework/src/modules/chain/modules/blocks/chain.js
diego-G: The idea is to remove callbacks all over around while we refactor.

MaciejBaj: I would feel more comfortable with leaving the system.update working with callbacks.

I am not convinced about the way how it's implemented now:

1. system.update is an **async** function - `async update()`
2. in both of the calls in "chain" and "loader" it's called like a **promise** - `return components.system.update().then(() => seriesCb()).catch(seriesCb);`
3. both of the functions "chain.deleteLastBlock" and "loader.sync" are using **callbacks** and async library.

Previously there were callback used consistently.

diego-G: @nazarhussain, @shuse2 and @jondubois raised the point during another PR https://github.com/LiskHQ/lisk/pull/2890. Tackling it in a separate issue was the safe strategy to avoid any behavioural change.
The idea is to not to leave any callback behind. I've followed the same pattern in another components.

nazarhussain: @MaciejBaj Problem is not the usage, problem is the function definition. A function which is returning a promise and accepting a callback is a code smell, so if we get chance we should remove such straight forward issues. 

- Want callback then don't return Promise 
- Want a promise, then don't accept a callback 

That's what @diego-G is trying to resolve here, using a consistent approach.

MaciejBaj:
Please fix the tests failing on Jenkins.

PR #2925, opened 2019-02-20
Refactor verify in generation of new blocks - Closes #2823

### LINKED ISSUE #2823: Refactor verifying transactions and use the new StateStore and transactions library
### Expected behavior
In the current implementation, transactions are verified by calling the `verify` function for each transaction defined in the logic folder.  
### Actual behavior
In this issue, the verifying of the transactions should be refactored and the application should use StateStore and verify methods defined for each type in the lisk-element transactions library.

1.6

mitsujutsu:
### What was the problem?

As part of integration of lisk-elements `transactions` library into core, we replace verification process. 
Most verifications were replaced in [PR #2892]( https://github.com/LiskHQ/lisk/pull/2892). 

Case remained where transactions in block were verified before block creation. 

### How did I fix it?

Refactored `generateBlock` in `process` module

### How to test it?

Run unit tests for `process` module

### Review checklist

* The PR resolves #2823 
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/chain/modules/blocks/process.js
+++ b/framework/src/modules/chain/modules/blocks/process.js
@@ -267,30 +267,25 @@ Process.prototype.getCommonBlock = function(peer, height, cb) {
 				const ids = res.ids;
 				// Perform request to supplied remote peer
 				peer = library.logic.peers.create(peer);
-				peer.rpc.blocksCommon(
-					{
-						ids,
-					},
-					(err, blocksCommonRes) => {
-						if (err) {
-							modules.peers.remove(peer);
-							return setImmediate(waterCb, err);
-						}
-
-						if (!blocksCommonRes.common) {
-							// FIXME: Need better checking here, is base on 'common' property enough?
-							comparisonFailed = true;
-							return setImmediate(
-								waterCb,
-								`Chain comparison failed with peer: ${
-									peer.string
-								} using ids: ${ids}`
-							);
-						}
-
-						return setImmediate(waterCb, null, blocksCommonRes.common);
+				peer.rpc.blocksCommon({ ids }, (err, blocksCommonRes) => {
+					if (err) {
+						modules.peers.remove(peer);
+						return setImmediate(waterCb, err);
 					}
-				);
+
+					if (!blocksCommonRes.common) {
+						// FIXME: Need better checking here, is base on 'common' property enough?
+						comparisonFailed = true;
+						return setImmediate(
+							waterCb,
+							`Chain comparison failed with peer: ${
+								peer.string
+							} using ids: ${ids}`
+						);
+					}
+
+					return setImmediate(waterCb, null, blocksCommonRes.common);
+				});
 			},
 			function(common, waterCb) {
 				// Check if we received genesis block - before response validation, as genesis block have previousBlock = null
@@ -450,10 +445,7 @@ Process.prototype.loadBlocksFromPeer = function(peer, cb) {
 
 	function getFromPeer(seriesCb) {
 		peer.rpc.blocks(
-			{
-				lastBlockId: lastValidBlock.id,
-				peer: library.logic.peers.me(),
-			},
+			{ lastBlockId: lastValidBlock.id, peer: library.logic.peers.me() },
 			(err, res) => {
 				err = err || res.error;
 				if (err) {
--- a/framework/src/modules/chain/modules/blocks/process.js
+++ b/framework/src/modules/chain/modules/blocks/process.js
@@ -16,9 +16,15 @@
 
 const _ = require('lodash');
 const async = require('async');
+const {
+	Status: TransactionStatus,
+} = require('@liskhq/lisk-transactions');
 const slots = require('../../helpers/slots.js');
 
-const { MAX_TRANSACTIONS_PER_BLOCK, ACTIVE_DELEGATES } = global.constants;
+const {
+	MAX_TRANSACTIONS_PER_BLOCK,
+	ACTIVE_DELEGATES,
+} = global.constants;
 
 const __private = {};
 let modules;
@@ -84,7 +90,7 @@ class Process {
  * @param {Object} block - Full normalized block
  * @param {function} cb - Callback function
  */
-__private.receiveBlock = function(block, cb) {
+__private.receiveBlock = function (block, cb) {
 	library.logger.info(
 		`Received new block id: ${block.id} height: ${
 			block.height
@@ -107,7 +113,7 @@ __private.receiveBlock = function(block, cb) {
  * @param {Object} block - Received block
  * @param {function} cb - Callback function
  */
-__private.receiveForkOne = function(block, lastBlock, cb) {
+__private.receiveForkOne = function (block, lastBlock, cb) {
 	let tmp_block = _.clone(block);
 
 	// Fork: Consecutive height but different previous block id
@@ -123,7 +129,7 @@ __private.receiveForkOne = function(block, lastBlock, cb) {
 	}
 	return async.series(
 		[
-			function(seriesCb) {
+			function (seriesCb) {
 				try {
 					tmp_block = library.logic.block.objectNormalize(tmp_block);
 				} catch (err) {
@@ -132,11 +138,11 @@ __private.receiveForkOne = function(block, lastBlock, cb) {
 				return setImmediate(seriesCb);
 			},
 			// Check valid slot
-			function(seriesCb) {
+			function (seriesCb) {
 				__private.validateBlockSlot(tmp_block, lastBlock, seriesCb);
 			},
 			// Check received block before any deletion
-			function(seriesCb) {
+			function (seriesCb) {
 				const check = modules.blocks.verify.verifyReceipt(tmp_block);
 
 				if (!check.verified) {
@@ -171,7 +177,7 @@ __private.receiveForkOne = function(block, lastBlock, cb) {
  * @param {Object} block - Received block
  * @param {function} cb - Callback function
  */
-__private.receiveForkFive = function(block, lastBlock, cb) {
+__private.receiveForkFive = function (block, lastBlock, cb) {
 	let tmpBlock = _.clone(block);
 
 	// Fork: Same height and previous block id, but different block id
@@ -195,7 +201,7 @@ __private.receiveForkFive = function(block, lastBlock, cb) {
 	}
 	return async.series(
 		[
-			function(seriesCb) {
+			function (seriesCb) {
 				try {
 					tmpBlock = library.logic.block.objectNormalize(tmpBlock);
 				} catch (err) {
@@ -204,11 +210,11 @@ __private.receiveForkFive = function(block, lastBlock, cb) {
 				return setImmediate(seriesCb);
 			},
 			// Check valid slot
-			function(seriesCb) {
+			function (seriesCb) {
 				__private.validateBlockSlot(tmpBlock, lastBlock, seriesCb);
 			},
 			// Check received block before any deletion
-			function(seriesCb) {
+			function (seriesCb) {
 				const check = modules.blocks.verify.verifyReceipt(tmpBlock);
 
 				if (!check.verified) {
@@ -222,12 +228,12 @@ __private.receiveForkFive = function(block, lastBlock, cb) {
 				return setImmediate(seriesCb);
 			},
 			// Delete last block
-			function(seriesCb) {
+			function (seriesCb) {
 				library.logger.info('Last block loses due to fork 5');
 				modules.blocks.chain.deleteLastBlock(seriesCb);
 			},
 			// Process received block
-			function(seriesCb) {
+			function (seriesCb) {
 				return __private.receiveBlock(block, seriesCb);
 			},
 		],
@@ -251,22 +257,24 @@ __private.receiveForkFive = function(block, lastBlock, cb) {
  * @returns {Object} cb.err - Error if occurred
  * @returns {Object} cb.res - Result object
  */
-Process.prototype.getCommonBlock = function(peer, height, cb) {
+Process.prototype.getCommonBlock = function (peer, height, cb) {
 	let comparisonFailed = false;
 
 	return async.waterfall(
 		[
-			function(waterCb) {
+			function (waterCb) {
 				// Get IDs sequence (comma separated list)
 				modules.blocks.utils.getIdSequence(height, (err, res) =>
 					setImmediate(waterCb, err, res)
 				);
 			},
-			function(res, waterCb) {
+			function (res, waterCb) {
 				const ids = res.ids;
 				// Perform request to supplied remote peer
 				peer = library.logic.peers.create(peer);
-				peer.rpc.blocksCommon({ ids }, (err, blocksCommonRes) => {
+				peer.rpc.blocksCommon({
+					ids,
+				}, (err, blocksCommonRes) => {
 					if (err) {
 						modules.peers.remove(peer);
 						return setImmediate(waterCb, err);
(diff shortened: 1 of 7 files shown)


framework/src/modules/chain/modules/blocks/process.js
```diff
@@ -537,68 +547,40 @@ Process.prototype.loadBlocksFromPeer = function(peer, cb) {
...
-		(transaction, eachSeriesCb) => {
-			modules.accounts.getAccount(
-				{ publicKey: transaction.senderPublicKey },
-				(err, sender) => {
-					if (err || !sender) {
-						return setImmediate(eachSeriesCb, 'Sender not found');
-					}
+	let readyTransactions;
+	try {
+		const transactionIds = await modules.processTransactions.verifyTransactions(
```
SargeKhan: why do we need to first get transactionIds? Why don't we simply do this: 
```
readyTransactions = await modules.processTransactions.verifyTransactions(
		transactions,				transactions
).filter(response => response.status === TransactionStatus.OK);
```

framework/src/modules/chain/modules/blocks/process.js
```diff
@@ -537,68 +547,40 @@ Process.prototype.loadBlocksFromPeer = function(peer, cb) {
  * @returns {function} cb - Callback function from params (through setImmediate)
  * @returns {Object} cb.err - Error message if error occurred
  */
-Process.prototype.generateBlock = function(keypair, timestamp, cb) {
+Process.prototype.generateBlock = async function (keypair, timestamp, cb) {
```
SargeKhan: This function is not an async function. Therefore, we shouldn't make it one. I would suggest that when calling `modules.processTransactions.verifyTransactions`, we use `.then` and `.catch` blocks which call the callback functions.

framework/test/unit/modules/chain/unit/modules/blocks/process.js
```diff
@@ -1729,132 +1730,76 @@ describe('blocks/process', async () => {
...
+							modules.blocks.process.processTransactions.verifyTransactions.throws()
 						);
 
 						it('should call a callback with error', done => {
 							blocksProcessModule.generateBlock(
 								{ publicKey: '123abc', privateKey: 'aaa' },
 								41287231,
 								err => {
-									expect(err).to.equal('Sender not found');
+									expect(err).to.equal('Transaction type not found.');
```
SargeKhan: It's strange that we are getting this error message. Do you know why?

framework/src/modules/chain/modules/blocks/process.js
mitsujutsu: `readyTransactions` needs to be whole transactions, whereas `verifyTransactions` only returns transaction responses. the ids are used in the next line to filter the original transactions using the ids.

framework/src/modules/chain/modules/blocks/process.js
shuse2: @SargeKhan I think we cannot use `await` in non async function?

framework/test/unit/modules/chain/unit/helpers/init_transaction.js
```diff
@@ -28,209 +28,198 @@ const initTransaction = require('../../../../../../src/modules/chain/helpers/ini
 
 describe('init transaction', async () => {
 	it('should initialize a transfer transaction', async () => {
```
shuse2: this `it` seems un-aligned with other test

framework/test/unit/modules/chain/unit/modules/blocks/process.js
```diff
@@ -1729,132 +1730,76 @@ describe('blocks/process', async () => {
...
-													expect(
-														modules.blocks.verify.processBlock.args[0][0]
-															.transactions.length
-													).to.equal(0);
-													done();
-												}
-											);
-										});
-									});
+						it('should generate block with transactions', done => {
```
shuse2: why expectation changes here?

framework/src/modules/chain/modules/blocks/process.js
pablitovicente: @SargeKhan why not possible to turn it into an async/await function?

framework/test/unit/modules/chain/unit/modules/blocks/process.js
mitsujutsu: this expectation falls under 'when suceeds', whereas the original was under 'when fails', so it makes sense to generate block with transactions.

framework/test/unit/modules/chain/unit/helpers/init_transaction.js
mitsujutsu: how so?

framework/src/modules/chain/modules/blocks/process.js
```diff
@@ -537,68 +547,40 @@ Process.prototype.loadBlocksFromPeer = function(peer, cb) {
...
-	const ready = [];
-
-	async.eachSeries(
-		transactions,
-		(transaction, eachSeriesCb) => {
-			modules.accounts.getAccount(
-				{ publicKey: transaction.senderPublicKey },
-				(err, sender) => {
-					if (err || !sender) {
-						return setImmediate(eachSeriesCb, 'Sender not found');
```
pablitovicente: Why is this check gone? Who is validating this now?

framework/test/unit/modules/chain/unit/helpers/init_transaction.js
shuse2: spacing seems in correct? maybe only in github

framework/src/modules/chain/modules/blocks/process.js
mitsujutsu: after discussion with @shuse2 decision to use then to maintain callback usage, and to do refactoring into async at once in a later PR.

framework/src/modules/chain/modules/blocks/process.js
mitsujutsu: `verifyTransactions` will call `transaction.apply` which calls `stateStore.getOrDefault` which creates a default account in case it does not exist. therefore this check is no longer required.

framework/test/unit/modules/chain/unit/modules/blocks/process.js
```diff
@@ -1711,80 +1714,104 @@ describe('blocks/process', async () => {
...
 
-						it('should call a callback with error', done => {
+						it('should call a callback with error', async () => {
 							blocksProcessModule.generateBlock(
 								{ publicKey: '123abc', privateKey: 'aaa' },
 								41287231,
 								err => {
-									expect(err).to.equal('Transaction type not found.');
-									done();
+									expect(err).to.not.be.null;
```
pablitovicente: maybe it reads better either as `to.be.not.null` or as `not.to.be.null`

framework/src/modules/chain/modules/blocks/process.js
```diff
@@ -588,7 +571,7 @@ Process.prototype.generateBlock = function(keypair, timestamp, cb) {
 					keypair,
 					timestamp,
 					previousBlock: modules.blocks.lastBlock.get(),
-					transactions: ready,
+					transactions: readyTransactions,
 				});
 			} catch (e) {
```
SargeKhan: We dont need to have a try..catch block here..

mitsujutsu: are you sure? original implementation also has in try catch.

mitsujutsu: ah I see now because of the next catch

framework/test/unit/modules/chain/unit/modules/blocks/process.js
```diff
@@ -1729,132 +1730,76 @@ describe('blocks/process', async () => {
...
-									});
+						it('should generate block with transactions', done => {
+							blocksProcessModule.generateBlock(
+								{ publicKey: '123abc', privateKey: 'aaa' },
+								41287231,
+								err => {
+									expect(err).to.be.null;
+									expect(
+										modules.blocks.verify.processBlock.args[0][0]
+											.transactions.length
```
pablitovicente: Could it be possible to also test here that the included transactions are the correct ones?

framework/test/unit/modules/chain/unit/modules/blocks/process.js
```diff
@@ -1661,200 +1666,182 @@ describe('blocks/process', async () => {
...
 	describe('generateBlock', async () => {
-		describe('modules.transactions.getUnconfirmedTransactionList', async () => {
-			describe('when query returns empty array', async () => {
-				beforeEach(() => {
+		describe('modules.transactions.getUnconfirmedTransactionList', () => {
+			describe('when query returns empty array', () => {
+				beforeEach(async () => {
 					modules.transactions.getUnconfirmedTransactionList.returns([]);
-					return modules.blocks.verify.processBlock.callsArgWith(
+					modules.processTransactions.verifyTransactions.returns(
```
shuse2: `resolves`?

framework/test/unit/modules/chain/unit/modules/blocks/process.js
```diff
@@ -1661,200 +1666,182 @@ describe('blocks/process', async () => {
...
 			});
 
-			describe('when query returns transactions', async () => {
-				beforeEach(() => {
+			describe('when query returns transactions', () => {
+				beforeEach(async () => {
 					modules.transactions.getUnconfirmedTransactionList.returns([
-						{ id: 1, type: 0 },
-						{ id: 2, type: 1 },
+						{ id: 1, type: 0, senderPublicKey: '1L' },
```
shuse2: this is `senderId` =0
and why do we need to add new property here?

framework/test/unit/modules/chain/unit/modules/blocks/process.js
```diff
@@ -1661,200 +1666,182 @@ describe('blocks/process', async () => {
...
-										beforeEach(() =>
-											library.logic.transaction.verify.callsArgWith(
-												4,
-												'transaction.verify-ERR',
-												null
-											)
-										);
+					describe('when transactions verification succeeds', async () => {
+						beforeEach(async () => {
+							modules.processTransactions.verifyTransactions.returns(
```
shuse2: same as above `resolves`

framework/test/unit/modules/chain/unit/modules/blocks/process.js
```diff
@@ -1661,200 +1666,182 @@ describe('blocks/process', async () => {
...
+					);
+					modules.blocks.verify.processBlock.callsArgWith(
 						3,
 						null,
 						modules.blocks.verify.processBlock.args
 					);
 				});
 
-				it('should generate block without transactions', done => {
+				it('should generate block without transactions', async () => {
```
SargeKhan: We are not returning a promise from this function, and the assertions in this test are defined in the callback. So, I think this test case finishes before the assertions are reached. Are you sure this is a valid test?

framework/test/unit/modules/chain/unit/modules/blocks/process.js
```diff
@@ -1661,200 +1666,182 @@ describe('blocks/process', async () => {
...
```
SargeKhan: Same for this test.

framework/test/unit/modules/chain/unit/modules/blocks/process.js
```diff
@@ -1661,200 +1666,182 @@ describe('blocks/process', async () => {
...
 
-						it('should call a callback with error', done => {
+						it('should call a callback with error', async () => {
 							blocksProcessModule.generateBlock(
 								{ publicKey: '123abc', privateKey: 'aaa' },
 								41287231,
 								err => {
-									expect(err).to.equal('Sender not found');
-									done();
+									expect(err).not.to.be.null;
```
SargeKhan: What is the error here? I think we should have an assertion.

framework/test/unit/modules/chain/unit/modules/blocks/process.js
mitsujutsu: Yes it is reaching the assertions properly (changing expected values fails test). But I can use `done` for better clarity.

framework/test/unit/modules/chain/unit/modules/blocks/process.js
mitsujutsu: I added `senderPublicKey` so it's easy to differentiate from a `transaction response` object which looks very similar. Really all that's actually needed for this test is `id`. I can change to `senderId` or just have `id` also.

framework/test/unit/modules/chain/unit/modules/blocks/process.js
```diff
@@ -1749,7 +1749,7 @@ describe('blocks/process', async () => {
 								{ publicKey: '123abc', privateKey: 'aaa' },
```
shuse2: for `Promise.reject` above, it should be `rejects`, too

framework/src/modules/chain/modules/blocks/process.js
```diff
@@ -266,25 +267,30 @@ Process.prototype.getCommonBlock = function(peer, height, cb) {
...
 				// Perform request to supplied remote peer
 				peer = library.logic.peers.create(peer);
-				peer.rpc.blocksCommon({ ids }, (err, blocksCommonRes) => {
-					if (err) {
-						modules.peers.remove(peer);
-						return setImmediate(waterCb, err);
-					}
+				peer.rpc.blocksCommon(
+					{
+						ids,
```
MaciejBaj: I think your priettier is outdated.

framework/src/modules/chain/modules/blocks/process.js
```diff
@@ -444,7 +450,10 @@ Process.prototype.loadBlocksFromPeer = function(peer, cb) {
 
 	function getFromPeer(seriesCb) {
 		peer.rpc.blocks(
-			{ lastBlockId: lastValidBlock.id, peer: library.logic.peers.me() },
```
MaciejBaj: Same, outdated prettier change

framework/src/modules/chain/modules/blocks/process.js
```diff
@@ -539,66 +548,36 @@ Process.prototype.loadBlocksFromPeer = function(peer, cb) {
...
-					}
-					return setImmediate(eachSeriesCb);
-				}
+	const transactions =
+		modules.transactions.getUnconfirmedTransactionList(
+			false,
+			MAX_TRANSACTIONS_PER_BLOCK
+		) || [];
+
+	modules.processTransactions
```
MaciejBaj: Is it correct?

framework/src/modules/chain/modules/blocks/process.js
```diff
@@ -539,66 +548,36 @@ Process.prototype.loadBlocksFromPeer = function(peer, cb) {
...
+		.verifyTransactions(transactions)
+		.then(responses => {
+			const readyTransactions = transactions.filter(tx =>
+				responses
+					.filter(response => response.status === TransactionStatus.OK)
+					.map(r => r.id)
```
MaciejBaj: Please be consistent. Use `transactions` keyword instead of `tx` and response instead of `r`.

framework/src/modules/chain/modules/blocks/process.js
mitsujutsu: yes we are using ProcessTransaction module to keep elements functionality in one place as part of integration process.

framework/test/unit/modules/chain/unit/modules/blocks/process.js
```diff
@@ -1716,145 +1723,121 @@ describe('blocks/process', async () => {
...
-											).to.equal(0);
-											done();
-										}
-									);
-								});
-							});
-
-							describe('when returns true', async () => {
-								beforeEach(() => library.logic.transaction.ready.returns(true));
+						it('should generate block without transactions', async () => {
```
SargeKhan: I don't understand how the function can be async and wait for the callback. Can you please double check and maybe change this async function to the one which uses `done` callback.

PR #2931, opened 2019-02-21
Support running single test files with npx mocha - Closes #2930

### LINKED ISSUE #2930: Support running single test files with npx mocha
### Expected behavior
We should be able to run mocha tests with `npx mocha <file>` command.
### Actual behavior

We are not able to do that at the moment.

yatki:
### What was the problem?
We were not able to run mocha tests with `npx mocha <file>` command.
### How did I fix it?
I created a symlink for mocha.opts file under `test/mocha.opts` so mocha can read the options by default.
### How to test it?
Run:

`npx mocha framework/test/mocha/unit/components/cache/cache.js`

### Review checklist

* The PR resolves #2930 
* Linting passes
* Tests pass

### DIFF
--- a/test/mocha.opts
+++ b/test/mocha.opts
@@ -1,0 +1,1 @@
+../framework/test/mocha/mocha.opts


test/mocha.opts
```diff
@@ -0,0 +1 @@
+../framework/test/mocha/mocha.opts
```
nazarhussain: Does it work with IDEs e.g. Webstorm? What to do when root test directory will have its own test files?

yatki: Please checkout the branch and try it with your IDE. I don't have it installed. Also, I think IDE related configuration shouldn't be application configuration's concern.

When the root project have its own tests, they would conflict with the current framework tests anyway. I think we can come up with a solution then.

nazarhussain: I tried, its working with Webstorm.

nazarhussain: @yatki The point is not far to introduce root level tests for the application implementation. So start thinking of the solution to this problem.

PR #2941, opened 2019-02-22
Update Account entity to insert/delete entries in dependent tables for memberPublicKeys and votedDelegatePublicKeys properties - Closes #2938

### LINKED ISSUE #2938: Update Account entity to insert/delete entries in dependent tables for memberPublicKeys and votedDelegatePublicKeys properties
### Expected behavior
The account entity should internally handle the logic for saving/deleting data from `mem_account2_delegates` and `mem_account2_multisignatures` tables.  
### Actual behavior
When an account is updated for `memberPublicKeys` and `votedDelegatePublicKeys`, the account entity simply ignores these properties. These properties are updated only using `account.merge` function defined in the logic/account file. 
1.5

pablitovicente:
### What was the problem?

- Dependent records for mem_accounts were being created from within account.merge()

### How did I fix it?

- By making the account entity update those dependent records.

### How to test it?

- Build should be green 

### Review checklist

* The PR resolves #2938
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/components/storage/entities/account.js
+++ b/framework/src/components/storage/entities/account.js
@@ -439,10 +439,6 @@ class Account extends BaseEntity {
 
 		const objectData = _.omit(data, readOnlyFields);
 
-		if (_.isEmpty(objectData)) {
-			return Promise.resolve();
-		}
-
 		const mergedFilters = this.mergeFilters(filters);
 		const parsedFilters = this.parseFilters(mergedFilters);
 		const updateSet = this.getUpdateSet(objectData);
@@ -456,47 +452,48 @@ class Account extends BaseEntity {
 		const membersPublicKeys = data.membersPublicKeys || [];
 		const votedDelegatesPublicKeys = data.votedDelegatesPublicKeys || [];
 
+		const dependentRecordPromises = [];
+
 		if (membersPublicKeys.length > 0) {
-			return this.adapter
-				.executeFile(this.SQLs.update, params, {}, tx)
-				.then(() =>
-					this.updateDependentRecords(
-						'membersPublicKeys',
-						data.address,
-						data.membersPublicKeys,
-						tx
-					)
-				);
+			dependentRecordPromises.push(
+				this.updateDependentRecords(
+					'membersPublicKeys',
+					data.address,
+					data.membersPublicKeys,
+					tx
+				)
+			);
 		}
 
 		if (votedDelegatesPublicKeys.length > 0) {
-			return this.adapter
-				.executeFile(this.SQLs.update, params, {}, tx)
-				.then(() =>
-					this.updateDependentRecords(
-						'votedDelegatesPublicKeys',
-						data.address,
-						data.votedDelegatesPublicKeys,
-						tx
-					)
-				);
+			dependentRecordPromises.push(
+				this.updateDependentRecords(
+					'votedDelegatesPublicKeys',
+					data.address,
+					data.votedDelegatesPublicKeys,
+					tx
+				)
+			);
 		}
 
 		// Account remove all votes
 		if (votedDelegatesPublicKeys.length === 0) {
-			return this.adapter
-				.executeFile(this.SQLs.update, params, {}, tx)
-				.then(() =>
-					this.adapter.executeFile(
-						this.SQLs.deleteVotes,
-						{ accountId: data.address },
-						{},
-						tx
-					)
-				);
+			dependentRecordPromises.push(
+				this.adapter.executeFile(
+					this.SQLs.deleteVotes,
+					{ accountId: data.address },
+					{},
+					tx
+				)
+			);
 		}
 
-		return this.adapter.executeFile(this.SQLs.update, params, {}, tx);
+		return Promise.all(dependentRecordPromises).then(() => {
+			if (_.isEmpty(objectData)) {
+				return Promise.resolve();
+			}
+			return this.adapter.executeFile(this.SQLs.update, params, {}, tx);
+		});
 	}
 
 	/**
--- a/framework/src/components/storage/entities/account.js
+++ b/framework/src/components/storage/entities/account.js
@@ -736,43 +736,70 @@ class Account extends BaseEntity {
 			`Invalid dependency name "${dependencyName}" provided.`
 		);
 
-		const paramsForDelete = {
-			tableName: dependentFieldsTableMap[dependencyName],
-			accountId: address,
-			dependentIds: dependentPublicKeys,
-		};
-
 		const sqlForDelete = this.SQLs.deleteDependentRecords;
-
-		const valuesForInsert = dependentPublicKeys.map(dependentId => ({
-			accountId: address,
-			dependentId,
-		}));
 
 		const tableName = dependentFieldsTableMap[dependencyName];
 		const sqlForInsert = this.SQLs.createDependentRecords;
 
-		const createSet = this.getValuesSet(
-			valuesForInsert,
-			['accountId', 'dependentId'],
-			{ useRawObject: true }
-		);
-
 		return this.adapter
-			.executeFile(
-				sqlForDelete,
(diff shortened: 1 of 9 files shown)


framework/src/components/storage/entities/account.js
```diff
@@ -450,6 +453,29 @@ class Account extends BaseEntity {
 			updateSet,
 		};
 
+		const membersPublicKeys = data.membersPublicKeys || [];
+		const votedDelegatesPublicKeys = data.votedDelegatesPublicKeys || [];
+
+		if (membersPublicKeys.length > 0) {
+			this.adapter.executeFile(this.SQLs.update, params, {}, tx);
```
shuse2: isn't this `async`?

framework/src/components/storage/entities/account.js
```diff
@@ -450,6 +453,29 @@ class Account extends BaseEntity {
...
+			return this.updateDependentRecords(
+				'membersPublicKeys',
+				data.address,
+				data.membersPublicKeys,
+				tx
+			);
+		}
+
+		if (votedDelegatesPublicKeys.length > 0) {
+			this.adapter.executeFile(this.SQLs.update, params, {}, tx);
```
shuse2: same as above

framework/src/components/storage/entities/account.js
```diff
@@ -429,6 +431,7 @@ class Account extends BaseEntity {
 	 * @param {Object} tx - Transaction object
 	 * @return {*}
 	 */
+	// eslint-disable-next-line consistent-return
```
shuse2: which part is this complaining?

framework/src/modules/chain/logic/account.js
```diff
@@ -406,6 +406,7 @@ class Account {
 				switch (fieldType) {
 					// blockId
 					case String:
+					case Array:
```
shuse2: let's move the comment too

framework/src/components/storage/entities/account.js
pablitovicente: Old comment will remove.

framework/src/components/storage/entities/account.js
```diff
@@ -450,7 +449,51 @@ class Account extends BaseEntity {
 			updateSet,
 		};
 
-		return this.adapter.executeFile(this.SQLs.update, params, {}, tx);
+		const membersPublicKeys = data.membersPublicKeys || [];
```
SargeKhan: If it doesn't exist, then don't initialize it with an empty array. It will be better if we add another condition in if statement like: 
`if (membersPublicKey && membersPublicKey.length > 0)`
I am requesting this because we call `update` function with `data` property including only the fields which are modified. Therefore, if `data.membersPublicKeys` doesn't exist, it is possible that account is registered multisignature account, but we are not updating the multisignature related infomration for that account.

framework/src/components/storage/entities/account.js
```diff
@@ -677,6 +720,88 @@ class Account extends BaseEntity {
...
+				[address]
+			)
+			.then(dependentRecordsForAddress => {
+				const oldDependentPublicKeys = dependentRecordsForAddress.map(
+					dependentRecord => dependentRecord.dependentId
+				);
+				const toRemove = oldDependentPublicKeys.filter(
+					aPK => !dependentPublicKeys.includes(aPK)
+				);
+				const toInsert = dependentPublicKeys.filter(
```
SargeKhan: Maybe we can articulate the variable names in this function better, and organize `const` variable declerations.

framework/src/components/storage/entities/account.js
pablitovicente: Done

framework/src/components/storage/entities/account.js
pablitovicente: Done

framework/src/components/storage/entities/account.js
```diff
@@ -450,7 +449,54 @@ class Account extends BaseEntity {
...
+				this.adapter.executeFile(
+					this.SQLs.deleteVotes,
+					{ accountId: data.address },
+					{},
+					tx
+				)
+			);
+		}
+
+		return Promise.all(dependentRecordPromises).then(() => {
```
diego-G: Why don't we convert the method to `async` and use `awaits`?

framework/src/components/storage/entities/account.js
```diff
@@ -677,6 +723,85 @@ class Account extends BaseEntity {
...
+		const sqlForDelete = this.SQLs.deleteDependentRecords;
+		const sqlForInsert = this.SQLs.createDependentRecords;
+		const tableName = dependentFieldsTableMap[dependencyName];
+
+		return this.adapter
+			.execute(
+				`SELECT "dependentId" FROM ${tableName} WHERE "accountId" = $1`,
+				[address]
+			)
+			.then(dependentRecordsForAddress => {
```
diego-G: The same here. This then looks not so syntactic sugar. With await it would be much better.

framework/src/components/storage/sql/accounts/create_dependent_records.sql
```diff
@@ -0,0 +1,22 @@
...
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+
+/*
+  DESCRIPTION: ?
```
diego-G: please add description and parameter

framework/src/components/storage/sql/accounts/delete_dependent_records.sql
```diff
@@ -0,0 +1,23 @@
...
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+
+/*
+  DESCRIPTION: ?
```
diego-G: Add description and parameters

framework/src/components/storage/sql/accounts/delete_votes.sql
```diff
@@ -0,0 +1 @@
+DELETE FROM mem_accounts2delegates WHERE "accountId" = ${accountId};
```
diego-G: Add copyright and doc headers.

framework/test/unit/components/storage/entities/account.js
```diff
@@ -2107,4 +2107,211 @@ describe('Account', async () => {
 				));
 		});
 	});
+
+	describe('updateDependentRecords', () => {
```
diego-G: I don't see any changes to adjust `update()` to the latest.

framework/src/components/storage/entities/account.js
michielmulders: I agree with Diego here. It is not very readable as the `return` is on the 9th line but the code goes on a whole lot further. With `await` you can split the code further up in an extra function which you can await to reduce the size of the initial function.

Separate function for: `return Promise.all(dependentRecordPromises);`

pablitovicente: @diego-G converted. 
@michielmulders code changed. I could probably agree with more refactors but they're outside the scope of this PR which we need merged ASAP. If you feel more refactors could apply please feel free to report an issue for approaching it for all entities.

framework/src/components/storage/entities/account.js
```diff
@@ -429,28 +432,66 @@ class Account extends BaseEntity {
...
+			const updateSet = this.getUpdateSet(objectData);
+
+			const params = {
+				...objectData,
+				parsedFilters,
+				updateSet,
+			};
+
+			if (_.isEmpty(objectData)) {
+				return Promise.resolve();
```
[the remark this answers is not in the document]
pablitovicente: Done

PR #2958, opened 2019-02-25
Move rounds related logic from account.merge function to rounds module - Closes #2942

### LINKED ISSUE #2942: Move rounds related logic from account.merge function to rounds module.
### Expected behavior
The account.merge function should only update data relevant to accounts. The entries added to rounds tables should be handled in the rounds module.
### Actual behavior
The account.merge function calls the rounds module and adds records in rounds table.

1.7.0

SargeKhan:
### What was the problem?
The account.merge function should only update data relevant to accounts. The entries added to rounds tables should be handled in the rounds module.
### How did I fix it?
Moved database queries related to rounds to the rounds module.
### How to test it?
Run unit tests.
### Review checklist

* The PR resolves #2942 
* Linting passes
* Tests pass


framework/src/modules/chain/logic/account.js
```diff
@@ -446,16 +445,6 @@ class Account {
 							);
 						}
 
-						if (updatedField === 'balance') {
```
shuse2: can we already remove this? I thought `reward` part still goes through this path

SargeKhan: You are right. :( I will add this logic again. But this means that as long as we don't remove the transactions implemented in the core, we will have a different delegates list.

framework/test/unit/modules/chain/unit/modules/rounds.js
```diff
@@ -237,6 +246,282 @@ describe('rounds', async () => {
 		});
 	});
 
+	describe('__private.updateRecipientsRoundInformationWithAmountForTransactions', () => {
```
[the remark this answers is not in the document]
SargeKhan: We don't have this convention in core.

framework/test/unit/modules/chain/unit/modules/rounds.js
```diff
@@ -237,6 +246,282 @@ describe('rounds', async () => {
...
+				'__private.updateSendersRoundInformationWithAmountForTransactions'
+			);
+			createRoundInformationWithAmountStub = sinon.stub(
+				Rounds.prototype,
+				'createRoundInformationWithAmount'
+			);
+		});
+
+		it('should call createRoundInformationWithAmount with negative value when forwardTick is set to true', async () => {
+			updateSendersRoundInformationWithAmountForTransactions(
```
shuse2: should we `await` here?
same for below

framework/src/modules/chain/modules/rounds.js
```diff
@@ -62,6 +64,128 @@ class Rounds {
...
+			transactions,
+			forwardTick,
+			tx
+		),
+	].reduce((accPromises, currPromises) => accPromises.concat(currPromises), []);
+	return Promise.all(promises)
+		.then(() => setImmediate(cb))
+		.catch(err => setImmediate(cb, err));
+};
+
```
nazarhussain: Add js-doc to new created functions.

framework/src/modules/chain/modules/rounds.js
```diff
@@ -62,6 +64,128 @@ class Rounds {
 	}
 }
 
+__private.updateSendersRoundInformationWithAmountForTransactions = function(
```
nazarhussain: `updateSendersRoundInformationWithAmount` is good enough to explain. We may ned up function names with line character limit.

framework/src/modules/chain/modules/rounds.js
```diff
@@ -62,6 +64,128 @@ class Rounds {
...
+__private.updateRecipientsRoundInformationWithAmountForTransactions = function(
+	round,
+	transactions,
+	forwardTick,
+	tx
+) {
+	// For IN_TRANSFER TRANSACTION, we first need to get the authorId to get the recipient
+	return Promise.all(
+		transactions
+			.filter(transaction => transaction.type === transactionTypes.IN_TRANSFER)
```
nazarhussain: `filter( ({ transaction: type }) => type === transactionTypes.IN_TRANSFER)`

framework/src/modules/chain/modules/rounds.js
```diff
@@ -62,6 +64,128 @@ class Rounds {
...
+			.map(transaction =>
+				// Check the owner of the dapp for recipient account
+				library.storage.entities.Transaction.getOne(
+					{
+						id: transaction.asset.inTransfer.dappId,
```
nazarhussain: You can first map to get the ids of the transactions and then use `id_in` to avoid running multiple queries against active transaction connection.

framework/src/modules/chain/modules/rounds.js
```diff
@@ -62,6 +64,128 @@ class Rounds {
...
+	transactions,
+	forwardTick,
+	tx
+) {
+	return transactions
+		.filter(transaction => transaction.type === transactionTypes.VOTE)
+		.map(transaction =>
+			(forwardTick
+				? transaction.asset.votes
+				: Diff.reverse(transaction.asset.votes)
```
nazarhussain: We should use builtin functions for such basic behaviours. Having an helper for just for `Array.reverse` is not a good approach.

framework/src/modules/chain/modules/rounds.js
```diff
@@ -62,6 +64,128 @@ class Rounds {
...
+			).map(vote => {
+				// Fetch first character
+				const mode = vote[0];
+				const dependentId = vote.slice(1);
```
nazarhussain: This was the old logic to get `dependentId`, its not safe to replace with simply `slice`.

```
let mode = updatedValueItem[0];
let dependentId = '';

if (mode === '-' || mode === '+') {
	dependentId = updatedValueItem.slice(1);
} else {
	dependentId = updatedValueItem;
	mode = '+';
}
```

framework/src/modules/chain/modules/rounds.js
```diff
@@ -62,6 +64,128 @@ class Rounds {
...
+			forwardTick,
+			tx
+		),
+		__private.updateRoundInformationWithDelegatesForTransactions(
+			round,
+			transactions,
+			forwardTick,
+			tx
+		),
+	].reduce((accPromises, currPromises) => accPromises.concat(currPromises), []);
```
nazarhussain: Why we even need to use `reduce` here?

framework/src/modules/chain/modules/rounds.js
```diff
@@ -247,12 +379,20 @@ Rounds.prototype.tick = function(block, done, tx) {
...
 
+	// Start round ticking
+	__private.ticking = true;
 	async.series(
 		[
+			function updateRoundInformation(cb) {
+				__private.updateRoundInformationForTransactions(
+					round,
+					block.transactions,
+					true,
```
nazarhussain: In recent code, we are using a constant approach rather than passing static values. e.g. 

```
const isForwardTick = true; 
```

And then pass `isForwardTick` to function, that means code more readable.

framework/src/modules/chain/modules/rounds.js
shuse2: this `reverse` is changing the sign of the `vote`, not reversing the transaction

framework/src/modules/chain/modules/rounds.js
shuse2: At this point, every transaction has to be valid in all the case. Therefore, it should be safe to do this.
Also, it's kinda scary to update the `mode` if it's not + or -....

framework/src/modules/chain/modules/rounds.js
shuse2: each function returns Array<Promise>, so this reduce is same as `flatmap`

SargeKhan:
I merged this PR because it already had 2 reviews and I was open for quite some time. But thank you for your reviews. I will create a separate pull request addressing these comments.

PR #2967, opened 2019-02-25
Account productivity not showing decimal places - Closes #2965

### LINKED ISSUE #2965: Account productivity not showing decimal places
### Expected behavior
Expect to get account productivity with 2 decimal places when requesting `/api/delegates`  and `/api/accounts`

### Actual behavior
It rounds up and returns as string.

### Steps to reproduce
GET `/api/delegates`

1.5

lsilvs:
### What was the problem?
Account productivity not showing decimal places and returning as string

### How did I fix it?
Fixed the SQL responsible to calculate the productivity

### How to test it?
GET `/api/delegates` - productivity should be number with 2 decimal places

### Review checklist

* The PR resolves #2965
* Linting passes
* Tests pass

### DIFF
--- a/storage/sql/accounts/get.sql
+++ b/storage/sql/accounts/get.sql
@@ -40,7 +40,7 @@ SELECT
 	when
 		"producedBlocks" + "missedBlocks" = 0 then 0
 	else
-		(("producedBlocks"::float / ("producedBlocks" + "missedBlocks")) * 100.0)::integer
+		ROUND((("producedBlocks"::float / ("producedBlocks" + "missedBlocks")) * 100.0)::numeric, 2)::float
 	end AS productivity
 FROM
 	mem_accounts
--- a/test/unit/storage/entities/account.js
+++ b/test/unit/storage/entities/account.js
@@ -560,15 +560,15 @@ describe('Account', () => {
 			});
 
 			it('should fetch "productivity" with correct query', async () => {
-				const producedBlocks = 5;
+				const producedBlocks = 50;
 				const missedBlocks = 3;
 				const validAccount = new accountFixtures.Account({
 					producedBlocks,
 					missedBlocks,
 				});
 				await AccountEntity.create(validAccount);
-				const productivity = parseInt(
-					producedBlocks / (producedBlocks + missedBlocks) * 100.0
+				const productivity = parseFloat(
+					(producedBlocks / (producedBlocks + missedBlocks) * 100.0).toFixed(2)
 				);
 
 				const account = await AccountEntity.getOne(
--- a/test/unit/storage/entities/account.js
+++ b/test/unit/storage/entities/account.js
@@ -559,17 +559,33 @@ describe('Account', () => {
 				);
 			});
 
-			it('should fetch "productivity" with correct query', async () => {
+			it('should fetch "productivity" with two decimal places when value is not integer', async () => {
 				const producedBlocks = 50;
-				const missedBlocks = 3;
+				const missedBlocks = 25;
 				const validAccount = new accountFixtures.Account({
 					producedBlocks,
 					missedBlocks,
 				});
 				await AccountEntity.create(validAccount);
-				const productivity = parseFloat(
-					(producedBlocks / (producedBlocks + missedBlocks) * 100.0).toFixed(2)
+				const productivity = 66.67;
+
+				const account = await AccountEntity.getOne(
+					{ address: validAccount.address },
+					{ extended: true }
 				);
+
+				expect(account.productivity).to.be.eql(productivity);
+			});
+
+			it('should fetch "productivity" with no decimal places when value is integer', async () => {
+				const producedBlocks = 75;
+				const missedBlocks = 25;
+				const validAccount = new accountFixtures.Account({
+					producedBlocks,
+					missedBlocks,
+				});
+				await AccountEntity.create(validAccount);
+				const productivity = 75;
 
 				const account = await AccountEntity.getOne(
 					{ address: validAccount.address },
--- a/storage/sql/accounts/get_extended.sql
+++ b/storage/sql/accounts/get_extended.sql
@@ -40,7 +40,7 @@ SELECT
     when
     	"producedBlocks" + "missedBlocks" = 0 then 0
     else
-		(("producedBlocks"::float / ("producedBlocks" + "missedBlocks")) * 100.0)::integer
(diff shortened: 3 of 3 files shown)


test/unit/storage/entities/account.js
```diff
@@ -560,15 +560,15 @@ describe('Account', () => {
...
 				const missedBlocks = 3;
 				const validAccount = new accountFixtures.Account({
 					producedBlocks,
 					missedBlocks,
 				});
 				await AccountEntity.create(validAccount);
-				const productivity = parseInt(
-					producedBlocks / (producedBlocks + missedBlocks) * 100.0
+				const productivity = parseFloat(
+					(producedBlocks / (producedBlocks + missedBlocks) * 100.0).toFixed(2)
```
nazarhussain: Personally I prefer to hardcode expected values.

4miners: What's the point of `100.0` vs `100`?

lsilvs: Good point. I will check how it was being returned before 1.5

lsilvs: Just checked and it used to return `100` without decimals places. I will add more tests for that case.

storage/sql/accounts/get.sql
```diff
@@ -40,7 +40,7 @@ SELECT
 	when
 		"producedBlocks" + "missedBlocks" = 0 then 0
 	else
-		(("producedBlocks"::float / ("producedBlocks" + "missedBlocks")) * 100.0)::integer
+		ROUND((("producedBlocks"::float / ("producedBlocks" + "missedBlocks")) * 100.0)::numeric, 2)::float
```
4miners: It can be simplified:
````
ROUND((("producedBlocks"::numeric / ("producedBlocks" + "missedBlocks")) * 100), 2)::float
````
```
lisk_test=# select ROUND(((75::numeric / (75 + 25)) * 100), 2)::float;
 round 
-------
    75
(1 row)
```
But maybe we should be consistent here and always return decimal part?
```
WHEN "producedBlocks" + "missedBlocks" = 0 THEN 0.00 ELSE
ROUND((("producedBlocks"::numeric / ("producedBlocks" + "missedBlocks")) * 100), 2)
```
```
lisk_test=# select ROUND(((75::numeric / (75 + 25)) * 100), 2);
 round 
-------
 75.00
(1 row)
```
However, I don't really understand why we moved this logic into PostgreSQL, it should stay in the application layer.

Comment (4miners) [reply storage/sql/accounts/get.sql]: Btw, what's the point of those double casts like `"isDelegate"::int::boolean`?

Comment (nazarhussain) [reply storage/sql/accounts/get.sql]: @4miners I believe because small int can't be casted directly to boolean.

Comment (nazarhussain) [reply storage/sql/accounts/get.sql]: I agree to always return the decimal part.

Comment (MaciejBaj) [reply storage/sql/accounts/get.sql]: > But maybe we should be consistent here and always return decimal part?

I see a reason for introducing the consistency in returning decimal part always, but we shouldn't change the format in a minor release. 

The most straightforward way is to restore the exact behaviour of the v1.0 API.

Comment (4miners) [reply storage/sql/accounts/get.sql]: @MaciejBaj It's your decision, however, would be nice if we can finally decide what **is** a breaking change and what **is not** and then stick to that decision. I can't understand how adding a decimal part to a field that already can have decimal part can be a breaking change and require a major release.

Comment (4miners) [reply storage/sql/accounts/get.sql]: Also, what with `ROUND((("producedBlocks"::numeric / ("producedBlocks" + "missedBlocks")) * 100), 2)::float` then?

Comment (MaciejBaj) [reply storage/sql/accounts/get.sql]: I see no reason in changing the `productivity` format to 'number with two decimal places' also for the integers vs returning `float` numbers like before without having a clear reason nor priority or claim from the API users to do so.

> would be nice if we can finally decide what **is** a breaking change and what **is not** and then stick to that decision

That's pretty easy to follow and was already decided since we're going with Swagger from Lisk Core 1.0: https://swagger.io/docs/specification/data-models/data-types/.

> I can't understand how adding a decimal part to a field that already can have decimal part can be a breaking change and require a major release.

Let me explain you.

For numbers, there is two types: `number` and `integer`. Moreover, each number and integer has its format: 
```
type	format	Description
number	–	Any numbers.
number	float	Floating-point numbers.
number	double	Floating-point numbers with double precision.
integer	–	Integer numbers.
integer	int32	Signed 32-bit integers (commonly used integer type).
integer	int64	Signed 64-bit integers (long type).

source: https://swagger.io/docs/specification/data-models/data-types/
```
Now refering to our [swagger docs](https://app.swaggerhub.com/apis/LiskHQ/Lisk/1.0.30#/Accounts/getAccounts):

```
productivity:
        type: number
        example: 96.41
``` 
`productivity` has a type `number` but there is no `format` specified, meaning that it can be any number. Therefore, discussed change of `productivity` from the `float` format introduced in v1.0 (but not reflected in specs) to 'number with two decimal places` will **not be breaking**, despite of changing the numbers format. 

I would prefer for us to focus on implementing the features we want to implement and not to waste the energy on looking for more improvements in places where there is no demand for them.

4miners: > Therefore, discussed change of `productivity` from the `float` format introduced in v1.0 (but not reflected in specs) to 'number with two decimal places` will **not be breaking**, despite of changing the numbers format.

> I see a reason for introducing the consistency in returning decimal part always, but we shouldn't change the format in a minor release.

So what is the actual reason of not doing such a change in minor realase if this is **not breaking** change?



> I would prefer for us to focus on implementing the features we want to implement and not to waste the energy on looking for more improvements in places where there is no demand for them.

By not doing those simple improvments when refactoring, etc. we are creating more and more technical debt. How we will deal with this? It will not be fixed for months because there are more important things to do. For me approach like `it works, so leave it as is` is simply wrong. I was requested for doing a review on this PR, so I did and already "wasted" my energy on it and put there my suggestions. Applying them basically cost no time and the outcome is simplier to understand SQL and consistent API response, for me worth the effort.

MaciejBaj: The answer to your question is included in my previous comment. If you have some further questions in regards to breaking/non-breaking changes to the API it's best to clarify them in an offline discussion and keep the scope of this issue focused on fixing the bug.

If you are convinced to the idea of changing `productivity` format from `float` to 'number with two digital places' please open a separate issue for this.

PR #2969, opened 2019-02-26
API voters votes field value changed from count of voters to vote weight - Closes #2968

### LINKED ISSUE #2968: API voters votes field value changed from count of voters to vote weight
### Expected behavior
The `votes` field should have the count of voters
https://node01.lisk.io/api/voters?username=korben3&offset=0&limit=10&sort=publicKey%3Aasc
Even though the changes are valid, as the `votes` should reflect the vote weight, the breaking changes in a minor release is not allowed, so it needs to be the same as before, which is the count of voters.
### Actual behavior
`"votes"` field shows the vote weight of the delegate
https://testnet.lisk.io/api/voters?username=korben3&offset=0&limit=10&sort=publicKey%3Aasc
### Steps to reproduce
https://testnet.lisk.io/api/voters?username=korben3&offset=0&limit=10&sort=publicKey%3Aasc
1..5.0

lsilvs:
### What was the problem?
Wrong value for votes in `/api/voters`

### How did I fix it?
I used `storage.entities.Account.count` to get the total accounts that voted for queried delegate

### How to test it?
GET `/api/voters?username=genesis_1&limit=1`

### Review checklist

* The PR resolves #2968
* Linting passes
* Tests pass

### DIFF
--- a/api/controllers/voters.js
+++ b/api/controllers/voters.js
@@ -125,7 +125,11 @@ VotersController.getVoters = async function(context, next) {
 			_.pick(voter, ['address', 'publicKey', 'balance'])
 		);
 
-		data.votes = voters.length;
+		const votersCount = await storage.entities.Account.count({
+			votedDelegatesPublicKeys_in: [delegate.publicKey],
+		});
+
+		data.votes = votersCount;
 
 		return next(null, {
 			data,
--- a/api/controllers/voters.js
+++ b/api/controllers/voters.js
@@ -111,7 +111,6 @@ VotersController.getVoters = async function(context, next) {
 			'address',
 			'balance',
 		]);
-		data.votes = parseInt(delegate.vote);
 
 		// TODO: Make sure we return empty string in case of null username
 		// This can be avoided when we fix the `isDelegate` inconsistency mentioned above.
@@ -125,6 +124,8 @@ VotersController.getVoters = async function(context, next) {
 		data.voters = _.map(voters, voter =>
 			_.pick(voter, ['address', 'publicKey', 'balance'])
 		);
+
+		data.votes = voters.length;
 
 		return next(null, {
 			data,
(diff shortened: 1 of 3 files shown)


api/controllers/voters.js
```diff
@@ -126,6 +125,8 @@ VotersController.getVoters = async function(context, next) {
 			_.pick(voter, ['address', 'publicKey', 'balance'])
 		);
 
+		data.votes = voters.length;
```
4miners: It should be a total number of voters, not just length of the currently returned set. For example, if we specify `limit=1` but delegate have a total of `8000` voters it should still return `8000` here. Will it work like this?

lsilvs: oh! Got it. It won't. Gonna fix it.

lsilvs: @4miners  That's why the test was testing `to.be.least`, right?

4miners: @4miners Looks like the test was not reliable. We should also update the description of this field in `swagger.yml` to match the current behavior.

lsilvs: Ok. I will update swagger as well.

lsilvs:
@4miners Not sure how to improve this test as the number of votes will change each time the test runs. Any idea on how to make it more reliable?

4miners:
@lsilvs Maybe you can keep calling API with increasing offset until it returns no results and then you will have the count?

test/functional/http/get/voters.js
```diff
@@ -237,6 +237,32 @@ describe('GET /api/voters', () => {
...
+				let votesCount = 0;
+
+				// The following loop is a workaround to get votes count as the number of votes will change each time the test runs
+				// REF.: https://github.com/LiskHQ/lisk/pull/2969
+				do {
+					// eslint-disable-next-line no-await-in-loop
+					const result = await votersEndpoint.makeRequest(
+						{
+							username: validVotedDelegate.delegateName,
+							limit: 1,
```
4miners: We don't really need to pull only `1` vote per iteration, we can use max limit here.

PR #2972, opened 2019-02-26
API Accounts approval results changed from 1.4.1 to 1.5.0-rc.1 - Closes #2971

### LINKED ISSUE #2971: API Accounts approval results changed from 1.4.1 to 1.5.0-rc.1
### Expected behavior
Delegate approval should remain intact with `1.4.1` calculation.
### Actual behavior
`-approval` is the changes from `1.5.0-rc.1` and `+approval` from `1.4.1`. `productivity` issue is covered under https://github.com/LiskHQ/lisk/issues/2965
```
{
    data: [
...
        delegate: {
-         approval: 0,
+         approval: 14.8,
          missedBlocks: 7,
          producedBlocks: 20504,
-         productivity: 100,
+         productivity: 99.97,
          rank: 68,
...
    }
  }
```
### Steps to reproduce
http://167.99.47.154:7000/api/accounts?address=10431374778810601967L
http://104.248.83.98:7000/api/accounts?address=10431374778810601967L
1.5.0

lsilvs:
### What was the problem?
`api/accounts` returning wrong value for `approval` when account is a delegate

### How did I fix it?
Used the correct attribute when calculating the approval value

### How to test it?
`/api/accounts?address=10431374778810601967L`

### Review checklist

* The PR resolves #2971
* Linting passes
* Tests pass

### DIFF
--- a/api/controllers/accounts.js
+++ b/api/controllers/accounts.js
@@ -65,7 +65,10 @@ function accountFormatter(totalSupply, account) {
 		object.delegate.rank = parseInt(object.delegate.rank);
 
 		// Computed fields
-		object.delegate.approval = calculateApproval(object.vote, totalSupply);
+		object.delegate.approval = calculateApproval(
+			object.delegate.vote,
+			totalSupply
+		);
 	}
 
 	object.publicKey = object.publicKey || '';
--- a/test/functional/http/get/accounts/accounts.js
+++ b/test/functional/http/get/accounts/accounts.js
@@ -28,6 +28,7 @@ const expectSwaggerParamError = apiHelpers.expectSwaggerParamError;
 describe('GET /accounts', () => {
 	const account = randomUtil.account();
 	const accountsEndpoint = new SwaggerEndpoint('GET /accounts');
+	const constantsEndPoint = new SwaggerEndpoint('GET /node/constants 200');
 
 	describe('?', () => {
 		describe('address', () => {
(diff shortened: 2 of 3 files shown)


api/controllers/accounts.js
```diff
@@ -65,7 +65,10 @@ function accountFormatter(totalSupply, account) {
 		object.delegate.rank = parseInt(object.delegate.rank);
 
 		// Computed fields
-		object.delegate.approval = calculateApproval(object.vote, totalSupply);
+		object.delegate.approval = calculateApproval(
```
ManuGowda: use of variable `object` is not ok here, `Object` is a reserved keyword and also the variable can be named better here, since its related to `account` can we name it better? like `formatAccount` or something more meaningful.

test/functional/http/get/accounts/accounts.js
```diff
@@ -393,6 +394,26 @@ describe('GET /accounts', () => {
...
+					{ address: accountFixtures.existingDelegate.address },
+					200
+				),
+			];
+
+			const [
+				{ body: { data: constansts } },
+				{ body: { data: [{ delegate }] } },
+			] = await Promise.all(promises);
+			const calculatedApproval = parseFloat(
```
ManuGowda: we shouldn't be creating `calculatedApproval` function here, rather we should make use of the `calculatedApproval` function from the module and pass the object and expect the same behavior. If in the case in future someone change the logic, the test `calculatedApproval` needs to be updated as well.

api/controllers/accounts.js
MaciejBaj: @lsilvs why is it suddenly split into 4 lines:

from:
```
object.delegate.approval = calculateApproval(object.vote, totalSupply);
```
to
```
object.delegate.approval = calculateApproval(
			object.delegate.vote,
			totalSupply
		);
```

@mitsujutsu in his PRs has a similar behaviour

lsilvs: @ManuGowda I agree with your comment but your request is out of scope of this PR.

lsilvs: @MaciejBaj I believe it was split because adding `delegate` property increased the length of the line requiring this format.

ManuGowda: @LucasIsasmendi can you create an issue to at least fix it in the later stage, I agree it's not in the scope, but its something we need to fix later.

lsilvs: @ManuGowda  Can you open the issue as it's your request?

You tagged the wrong Lucas haha.

ManuGowda: 😄 ok will create one.

PR #2977, opened 2019-02-26
Support ENV vars and console parameters in application-level - Closes #2976

### LINKED ISSUE #2976: Support ENV vars and command line arguments in application-level
### Expected behavior
With recent changes the configuration is distributed to each module and components. So as the environment variables should move along to individual configuration.

### Actual behavior
 Currently we have a central place for all the definition of environment variables (old helpers/config.js)

1.6.0

nazarhussain:
### What was the problem?
Env variables definition was on central file, not compatible with new approach of modular configurations. 

### How did I fix it?
By making env variable and command line arguments support to be part of the schema itself. 

### How to test it?
From README file check the list of variables and try to set those individually. 

### Review checklist

* The PR resolves #2976
* Linting passes
* Tests pass


nazarhussain:
@MaciejBaj This PR is ready for review and following points to be kept in mind: 

1. `PROTOCOL_VERSION` - Added only for tests no config is schema attached to it
2. `config.coverage` - Removed as I can't find nay reference usage 
3. `--inspect-workers, --inspect-brokers` - Removed as I can't find any reference usage and its not bounded to any configuration as well
4. `config.api.mode = 'READONLY'` Setting this value during snapshotting is removed, because I also can't find any reference usage for `api.mode`
5. Reset of the options to disable syncing and broadcasting is moved chain module rather the configuration 
```
appConfig.network.enabled = false;
appConfig.network.list = [];
appConfig.broadcasts.active = false;
appConfig.syncing.active = false;
```


#### Things to do 

1. Now using `LISK_PEERS`, ` --peers` is required to specify the port along with ips. In old implementation the ports were required in config but was optional in the env variable and command line option. If you wan to keep the old behaviour then better is to adjust the schema so the port stay optional for `modules.chain.network.list` and then fill in ports with `modules.chain.network.wsPort` if not exists. 
2. Need to write some replacement logic for old `helpers/config.js` for `update_config.js`, it was only used for validation in that file. The script will work even without validation and application will validate it during the startup. Need to think on it further as configuration compilation logic is now distributed between modules so would not suggest to use it in the network level implementation script.

framework/src/controller/helpers/validator/keywords/index.js
```diff
@@ -0,0 +1,4 @@
+const env = require('./env');
+const arg = require('./arg');
+
+module.exports = { env, arg };
```
shuse2: ❤️

package.json
```diff
@@ -89,6 +89,7 @@
 		"tempy": "0.2.1",
 		"valid-url": "1.0.9",
 		"wamp-socket-cluster": "2.0.0-beta.4",
+		"yargs": "=13.2.1",
```
shuse2: no need for `=`

framework/src/controller/helpers/validator/keywords/arg.js
```diff
@@ -0,0 +1,84 @@
+const debug = require('debug')('lisk:validator:arg');
+const commandLineArguments = require('yargs').argv;
+const _ = require('lodash');
+const formatters = require('./formatters');
+
```
shuse2: prefer to put `validator` at the root level

src/index.js
```diff
@@ -1,32 +1,69 @@
...
+	},
+};
 
 try {
+	const { NETWORK, CUSTOM_CONFIG_FILE } = validator.validateWithDefaults(
+		appSchema,
+		{}
+	);
+
+	/* eslint-disable import/no-dynamic-require */
```
shuse2: prefer not to use `dynamic-require`.
Please use `fs` instead

framework/src/controller/helpers/validator/keywords/formatters.js
```diff
@@ -0,0 +1,27 @@
...
+				return {
+					ip,
+					wsPort: wsPort || 5000,
+				};
+			});
+		}
+		return [];
+	},
+
+	stringToDelegateList: value => {
```
shuse2: i think this is not framework logic

framework/src/controller/helpers/validator/keywords/env.js
```diff
@@ -0,0 +1,63 @@
...
+
+	if (typeof schema === 'string') {
+		envVariableName = schema;
+		variableFormatter = null;
+	} else {
+		envVariableName = schema[0];
+		variableFormatter = formatters[schema[1]] || null;
+	}
+
+	return function(data, dataPath, object, key) {
```
shuse2: prefer to use consistently anonymous function

framework/src/controller/helpers/validator/keywords/env.js
```diff
@@ -0,0 +1,63 @@
...
+			maxItems: 2,
+		},
+	],
+};
+
+const compile = (schema, parentSchema) => {
+	debug('compile: schema: %j', schema);
+	debug('compile: parent schema: %j', parentSchema);
+
+	let envVariableName;
```
shuse2: prefer to use `const`

framework/src/controller/helpers/validator/keywords/arg.js
```diff
@@ -0,0 +1,84 @@
...
+		return [option, alias];
+	}
+
+	return [];
+};
+
+const compile = (schema, parentSchema) => {
+	debug('compile: schema: %j', schema);
+	debug('compile: parent schema: %j', parentSchema);
+	let argNames;
```
shuse2: prefer to use const

framework/src/controller/helpers/validator/keywords/arg.js
```diff
@@ -0,0 +1,84 @@
...
+
+	if (typeof schema === 'string') {
+		argNames = extractArgsNames(schema);
+		argsFormatter = null;
+	} else if (Array.isArray(schema)) {
+		argNames = extractArgsNames(schema[0]);
+		argsFormatter = formatters[schema[1]] || null;
+	}
+
+	return function(data, dataPath, object, key) {
```
shuse2: prefer to consistently use anonymous function

src/index.js
```diff
@@ -39,5 +76,5 @@ try {
 			process.exit();
 		});
 } catch (e) {
-	console.error(e);
+	console.error(e.errors[1]);
```
shuse2: where does it actually throw and why it's index 1?

framework/src/controller/helpers/validator/keywords/arg.js
```diff
@@ -0,0 +1,84 @@
...
+		let argValue;
+
+		argNames.forEach(argName => {
+			if (!argValue) {
+				argValue = commandLineArguments[argName] || undefined;
```
shuse2: why we cannot use native function here?

framework/src/controller/helpers/validator/index.js
```diff
@@ -15,6 +16,8 @@ const validatorWithDefaults = new Ajv({
 	useDefaults: true,
 	$data: true,
 });
+validatorWithDefaults.addKeyword('env', envKeyword);
+validatorWithDefaults.addKeyword('arg', argKeyword);
```
shuse2: why do we need these why do we need these keywords?

framework/src/modules/http_api/defaults/config.js
```diff
@@ -26,10 +30,12 @@ const DefaultConfig = {
 				public: {
 					type: 'boolean',
 					default: false,
+					env: 'LISK_API_PUBLIC',
 				},
 				whiteList: {
 					type: 'array',
 					default: ['127.0.0.1'],
+					env: ['LISK_API_WHITELIST', 'stringToIpPortSet'],
```
shuse2: i think env and formatter should be separate property.
ie)
env: 'LISK_API_WHITELIST',
parser: 'stringToIPPortSet',
Also default doesn't have port setup

framework/src/controller/helpers/validator/keywords/formatters.js
```diff
@@ -0,0 +1,27 @@
+module.exports = {
```
shuse2: formatting is not validation imo

framework/src/controller/helpers/validator/index.js
MaciejBaj: I will add an explanation in JSDoc:
```
/**
 * Adds a possibility to validate the app parameters for Lisk Framework provided through:
 * 1. Environment variables (env) like "npm start LISK_NETWORK=test"
 * 2. Command line arguments (arg) like "npm start --network test"
 * These keywords extend Ajv Validator and follow the format:
 * @type {{compile, errors, modifying, valid, metaSchema}}
 */
```

framework/src/controller/helpers/validator/keywords/formatters.js
nazarhussain: Chain module uses `LISK_FORGING_DELEGATES` env variable which can only be specified as string. So this logic is responsible to convert string to delegate list.

framework/src/modules/http_api/defaults/config.js
nazarhussain: Passing parser as separate key to parent schema applies it to json config as well. If you want we can make `env` key it be an object and then specify parser.

framework/src/controller/helpers/validator/keywords/formatters.js
nazarhussain: Yes but its a step covert values to right format before validation.

nazarhussain:
@shuse2 @MaciejBaj @fchavant Please keep in mind this PR will introduce a breaking change to configuration. 

https://github.com/LiskHQ/lisk/pull/2977#issuecomment-468978151

Now using `LISK_PEERS`, `--peers` is required to specify the port along with ips. In old implementation the ports were required in config but was optional in the env variable and command line option. If you wan to keep the old behaviour then better is to adjust the schema so the port stay optional for modules.chain.network.list and then fill in ports with `modules.chain.network.wsPort` if not exists.

framework/src/controller/application.js
```diff
@@ -86,36 +87,49 @@ class Application {
 		constants = {},
 		config = { components: { logger: null }, modules: {} }
 	) {
-		if (typeof label === 'function') {
-			label = label.call();
+		let appConfig;
+
+		// If user passes multiple config objects merge them in left-right order
+		if (Array.isArray(config)) {
```
shuse2: Constructor config is initializing as object and treating as array is bit strange.
Please use consistent type

framework/src/controller/helpers/validator/keywords/formatters.js
shuse2: I think default and formatting are still not concern of validator.
I will create a discussion issue to gather opinion and if necessary we will change it later.
Also, I will create an issue to use @liskhq/lisk-validator

nazarhussain:
@shuse2 Apparently these seems different things, but we are using JSON-Schema specification which is very rich in feature set. It not just contains `default` but also `description` and `examples` keywords as well. 

https://json-schema.org/understanding-json-schema/reference/generic.html

So any tool like `Ajv` can't be called only validators rather than libraries implementing the specification. 

The topic you raise here is are we fully using a specification standard or subset of it. To me its much easier to communicate and establish the concept that we are using **JSON-Schema 7.0** specification and we use all of its features provided.

shuse2:
@nazarhussain Good point on the latest JSON schema spec.
For this PR, i think it's fine to include default/formatter, but let's discuss this with @MaciejBaj 

I think it's fine to have it in the file, but using the validator to modify the value 🤔

nazarhussain:
So we have separate validator method `validateWithDefaults` for this purpose, standard method `validate` does not modify the values and don't use defaults to validate.

framework/src/controller/application.js
nazarhussain: It was intended to pass multiple custom config files. Remember for application network config or custom config are both custom configuration. 

This particular part will be removed with the PR of https://github.com/LiskHQ/lisk/issues/3171 where we have config compilation logic separated and application will accept unified compile configuration.

scripts/update_config.js
```diff
@@ -393,7 +394,10 @@ history.migrate(
 		// Set the env variables to validate against correct network and config file
 		process.env.LISK_NETWORK = network;
 		process.env.LISK_CONFIG_FILE = tempFilePath;
-		new AppConfig(packageJSON, false);
+
+		// TODO: Add a replacement lgoic for config helper
```
limiaspasdaniel: `logic`

package.json
```diff
@@ -105,6 +105,7 @@
 		"tempy": "0.2.1",
 		"valid-url": "1.0.9",
 		"wamp-socket-cluster": "2.0.0-beta.4",
+		"yargs": "13.2.2",
```
MaciejBaj: Only yargs should be in this PR - ` feature/improve_overall_configuration` is not up to date.

framework/src/controller/helpers/validator/keywords/arg/index.js
```diff
@@ -0,0 +1,48 @@
...
+					formatter: formatters[schema.formatter] || null,
+				};
+
+	return function(data, dataPath, object, key) {
+		let argValue;
+		const commandLineArguments = yargs.argv;
+
+		argVariable.names.forEach(argName => {
+			if (!argValue) {
+				argValue = commandLineArguments[_.camelCase(argName)] || undefined;
```
MaciejBaj:
```suggestion
				// Remove "-" or "--" from command line argument names
				argValue = commandLineArguments[_.camelCase(argName)] || undefined;
```

package.json
nazarhussain: The only addition is this PR is `yargs`, rest were formatted by `npm install`

shuse2: I see only `yargs` included in this PR, so im resolving this

PR #2980, opened 2019-02-26
Remove numAmount z_schema and replace numbers genesis by strings - Closes #2957

### LINKED ISSUE #2957: Change all amount occurrences to strings across application
### Expected behavior
All amounts should be expressed as strings across the application.

### Actual behavior
`genesis_block.json` holds properties that represent a number (`amount`) both as an integer and string type. For consistency and more straightforward validation, it makes sense to convert all occurrences of integers to strings for system/application constants. 

This would help us to remove `numAmount` type from z-schema and change all of the amount occurrences across the application to strings.

### Steps to reproduce
Look at `genesis_block.json`.

/

michielmulders:
### What was the problem?
A custom schema had been defined in `z_schema` to apply validation for certain numbers in the genesis file like `fee`, `totalFee`, and `reward`. For the sake of better validation and consistency, this should be converted to strings.

### How did I fix it?
Modified genesis files
Modified JSON schema
Modified affected files

### How to test it?
Start any network.

### Review checklist

* The PR resolves #2957 
* Linting passes
* Tests pass


nazarhussain:
@michielmulders This change seems good, but few unit tests are failing not related to this change. Can you debug what's wrong with those tests.

michielmulders:
@nazarhussain Found the problem, tests should be passing now.

framework/src/components/storage/entities/block.js
```diff
@@ -197,8 +197,8 @@ class Block extends BaseEntity {
 		});
 		this.addField('timestamp', 'number', { filter: filterType.NUMBER });
 		this.addField('totalAmount', 'string', { filter: filterType.NUMBER });
-		this.addField('totalFee', 'string', { filter: filterType.NUMBER });
-		this.addField('reward', 'string', { filter: filterType.NUMBER });
+		this.addField('totalFee', 'number', { filter: filterType.NUMBER });
+		this.addField('reward', 'number', { filter: filterType.NUMBER });
```
nazarhussain: So you just converted `totalFee` to string, why the change in genesis block schema caused to change the storage layer. Storage layer were already treating those as strings and it should not change.

framework/src/components/storage/entities/transaction.js
```diff
@@ -243,7 +243,7 @@ class Transaction extends BaseEntity {
 			fieldName: 't_amount',
 		});
 		this.addField('fee', 'string', {
-			filter: filterTypes.NUMBER,
+			filter: filterTypes.TEXT,
```
nazarhussain: It should be consistent for every field. e.g. `reward` and `fee` fields both have filter type `NUMBER`, why we are changing this for one and not for other? 

This could cause different filter list on the entity level. e.g. With this change there will be no `fee_gt` and similar filters. The application may be using, so we should avoid that change. 

If its required we should come up with some better explanation for it.

PR #2983, opened 2019-02-27
Exclude sign signature if it is empty - Closes #2973

### LINKED ISSUE #2973: API Transactions exclude field signSignature if its empty
### Expected behavior
`signSignature` field should be removed if it's empty and only be shown when it has value.

### Actual behavior
`signSignature` field shows empty string if not present, this can lead to confusion, so we should not show `signSignature` if its empty.
### Steps to reproduce
`1.5.0-rc.1`: http://192.154.224.145:7000/api/transactions/?sort=timestamp:desc&type=2&limit=100&offset=0&fromTimestamp=0
`< 1.5.0-rc.1`: http://lindsay.lisk.plus:8000/api/transactions/?sort=timestamp:desc&type=2&limit=100&offset=0&fromTimestamp=0
1.5.0

diego-G:
### Review checklist

* The PR resolves #2973
* Linting passes
* Tests pass

### DIFF
--- a/api/controllers/transactions.js
+++ b/api/controllers/transactions.js
@@ -44,7 +44,7 @@ function transactionFormatter(transaction) {
 	result.senderId = result.senderId || '';
 	result.recipientId = result.recipientId || '';
 	result.recipientPublicKey = result.recipientPublicKey || '';
-	result.signSignature = result.signSignature || '';
+	result.signSignature = result.signSignature || undefined;
 	result.signatures = result.signatures || [];
 
 	return result;
--- a/test/functional/http/get/transactions.js
+++ b/test/functional/http/get/transactions.js
@@ -36,6 +36,7 @@ describe('GET /api/transactions', () => {
 	const account = randomUtil.account();
 	const account2 = randomUtil.account();
 	const account3 = accountFixtures.existingDelegate;
+	const accountSecondPass = randomUtil.account();
 	const minAmount = 20 * NORMALIZER; // 20 LSK
 	const maxAmount = 100 * NORMALIZER; // 100 LSK
 	const transaction1 = lisk.transaction.transfer({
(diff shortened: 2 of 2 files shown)


api/controllers/transactions.js
```diff
@@ -44,7 +44,7 @@ function transactionFormatter(transaction) {
 	result.senderId = result.senderId || '';
 	result.recipientId = result.recipientId || '';
 	result.recipientPublicKey = result.recipientPublicKey || '';
-	result.signSignature = result.signSignature || '';
+	result.signSignature = result.signSignature || undefined;
```
MaciejBaj:
```suggestion
	if (!result.signSignature) {
		delete result.signSignature;
	}
```
Np?

diego-G: The suggestion has the same effect than the solution. Swagger takes care of not showing it when it is undefined.

diego-G:
yes @ManuGowda, we recovered the former behaviour.

Regarding the tests, Swagger schema validator does the job.

ManuGowda:
how does swagger schema validates of the `signSignature` is present or not? so swagger will ensure the behavior? I do not want to keep this as blocker, I know we want hurry and rush for PR to be merged, if we identify and there is a fix for the bug without the coverage of test I am not sure about it, @MaciejBaj whats your call? not for this PR for any future PR.

MaciejBaj:
If Swagger would do the job, we wouldn't see this problem. 

@diego-G please extend the second signature tests to validate if the `signSignature` field is there only when the second signature is present.

MaciejBaj:
@ManuGowda covering the bug with proper test scenario is the behaviour we definitely want to keep, there is no excuse in being in rush in this case.

ManuGowda:
@michielmulders are you taking care of the tests here?

michielmulders:
Yes I am!

test/functional/http/get/transactions.js
```diff
@@ -1040,6 +1049,54 @@ describe('GET /api/transactions', () => {
...
+
+				await sendTransactionPromise(transaction, 200);
+				await waitFor.confirmations([transaction.id]);
+
+				// Act
+				const {
+					body: { data: transactions },
+				} = await transactionsEndpoint.makeRequest({ id: transaction.id }, 200);
+				// Assert
+				expect(transactions[0].signSignature).to.not.be.empty;
```
ManuGowda: this looks good, what would be nice to have is comparing `signSignature` against the expecting `signSignature`
something like `expect(transactions[0].signSignature).to.eq(signSignatureFuc(required params))`. but for now i am goona approve, you can use this pattern for future, this way the expectation is spot on.

michielmulders: Ok, thanks!

PR #2998, opened 2019-02-27
Add validation for delegate asset - Closes #2979

### LINKED ISSUE #2979: API Transaction Type Delegate Creation Missing Asset Fields
### Expected behavior
Transaction delegate registration asset delegate should have `username`, `publicKey` and `address`
### Actual behavior
`asset->delegate` contains only `username` in `1.5.0-rc.1`, and in previous versions the asset also had `publicKey` and `address`
this is the thing that breaks explorer
### Steps to reproduce
`1.5.0-rc.1`: http://192.154.224.145:7000/api/transactions/?sort=timestamp:desc&type=2&limit=100&offset=0&fromTimestamp=0
`< 1.5.0-rc.1`:  http://lindsay.lisk.plus:8000/api/transactions/?sort=timestamp:desc&type=2&limit=100&offset=0&fromTimestamp=0
1.5.0

michielmulders:
### What was the problem?
When registering a delegate via type 2 transaction, we only pass the `username` field.
The API response should contain an asset field with the following properties:
- `username`
- `address`
- `publicKey`

However, only `username` is present which breaks the explorer.

### How did I fix it?
Modify returned transaction via `transactionFormatter` to return `publicKey` and `address` at `api/controllers/transactions.js`.

Test title: `assets for type 2 transactions should contain key username, publicKey and address`.

### How to test it?
Start network: `node app.js`
Run test: `npm run mocha -- test/functional/http/get/transactions.js `

### Review checklist

* The PR resolves #2979 
* Linting passes
* Tests pass

### DIFF
--- a/api/controllers/transactions.js
+++ b/api/controllers/transactions.js
@@ -46,6 +46,10 @@ function transactionFormatter(transaction) {
 	result.recipientPublicKey = result.recipientPublicKey || '';
 	result.signSignature = result.signSignature || '';
 	result.signatures = result.signatures || [];
+	if (transaction.type === 2) {
+		result.asset.delegate.publicKey = result.senderPublicKey;
+		result.asset.delegate.address = result.senderId;
+	}
 
 	return result;
 }
--- a/api/controllers/transactions.js
+++ b/api/controllers/transactions.js
@@ -17,6 +17,7 @@
 const _ = require('lodash');
 const swaggerHelper = require('../../helpers/swagger');
 const ApiError = require('../../helpers/api_error');
+const transactionTypes = require('../../helpers/transaction_types');
 
 // Private Fields
 let modules;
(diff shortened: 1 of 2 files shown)


api/controllers/transactions.js
```diff
@@ -46,6 +46,10 @@ function transactionFormatter(transaction) {
 	result.recipientPublicKey = result.recipientPublicKey || '';
 	result.signSignature = result.signSignature || '';
 	result.signatures = result.signatures || [];
+	if (transaction.type === 2) {
```
ManuGowda: can we use the constants like the one below `transactionTypes.DELEGATE` in the code that way we dont have to relay on `2`.

michielmulders: Updated code addressing this comment +1

PR #3014, opened 2019-03-04
Upgrade lisk-elements - Closes #3007

### LINKED ISSUE #3007: Update Lisk-elements to use latest version
### Expected behavior
lisk-elements 1.0.0 is outdated, it should use the latest version 2.0.0 with specific packages.
@liskhq/lisk-transactions@2.0.0
@liskhq/lisk-cryptography@2.0.0

Also, if we update the package-lock.json, it will fail to install.

### Actual behavior
it's using lisk-elements@1.0.0

shuse2:
### What was the problem?
`lisk-elements` version was outdated.

### How did I fix it?
upgrade lisk-elements to use `@liskhq/lisk-transactions` and `@liskhq/lisk-cryptography`.
Main usage is test and delegate decrypting password.

### How to test it?
`npm t`

### Review checklist

* The PR resolves #3007 
* Linting passes
* Tests pass


framework/test/mocha/unit/modules/chain/logic/multisignature.js
```diff
@@ -36,6 +40,30 @@ const rawValidTransaction = testData.rawValidTransaction;
 const multiSigAccount1 = testData.multiSigAccount1;
 const multiSigAccount2 = testData.multiSigAccount2;
 
+const createInvalidRegisterMultisignatureTransaction = ({
```
diego-G: Please, move this function to `common/utils/elements`. It's pretty handy to reuse this in future scenarios.

shuse2: This function needs to be removed/changed when we introduce 3.0, and if we're gonna clean up we need to change a lot of stuff not only multiSignature, so i would say leave it as it is for now

diego-G: It's not about cleaning, it's about placing common methods related to the same library into the same file. The same method is being defined twice in different test suites. For reference: https://github.com/LiskHQ/lisk/blob/9e1b61e9a2fc361c0f88771b4f132bc340177543/framework/test/mocha/functional/http/post/4.multisignature.js#L38

PR #3015, opened 2019-03-04
Refactor undoing transactions and use the new StateStore and transactions library - Closes #3006

### LINKED ISSUE #3006: Refactor undoing transactions and use the new StateStore and transactions library
### Expected behavior
In the current implementation, transactions are reverted by calling the `undo` function for each transaction defined in the logic folder.

### Actual behavior
In this issue, the undoing of the transactions should be refactored and the application should use StateStore and undo methods defined for each type in the lisk-element transactions library.

1.6

pablitovicente:
### What was the problem?

Transaction undoing was not using state store and elements

### How did I fix it?

- Added `undoTransactions()` to process_transactions
- Rewrote `modules/blocks/chain.js` `undoConfirmedStep()` to use said method

### Review checklist

* The PR resolves #3006
* Linting passes
* Tests pass


framework/src/modules/chain/modules/process_transactions.js
```diff
@@ -60,6 +60,37 @@ class ProcessTransactions {
...
+				amount: transaction.amount.toString(),
+				recipientId: transaction.recipientId || '',
+			}))
+			.map(initTransaction);
+
+		await Promise.all(transactionInstances.map(t => t.prepare(stateStore)));
+
+		const transactionResponses = transactionInstances.map(transaction => {
+			const transactionResponse = transaction.undo(stateStore);
+			stateStore.transaction.add(transaction);
```
SargeKhan: I think in the undoConfirmedStep, all transactions should be added in the beginning and a particular transaction should be removed for which undo method is called.

framework/src/modules/chain/modules/blocks/chain.js
```diff
@@ -712,7 +696,7 @@ __private.popLastBlock = function(oldLastBlock, cb) {
 					oldLastBlock.transactions.reverse(),
```
SargeKhan: You are calling the undo method for transactions repeatedly. It should be only called once, while this mapSeries will repeat the undo call for all transactions in the block based on the number of transactions in the block.

framework/src/modules/chain/modules/blocks/chain.js
```diff
@@ -567,48 +567,32 @@ __private.loadSecondLastBlockStep = function(secondLastBlockId, tx) {
...
-						}
-						return setImmediate(resolve);
-					},
-					tx
-				);
-			},
-			tx
-		);
-	});
+__private.undoConfirmedStep = async function(blockToBeUndone, tx) {
```
SargeKhan: Just use the variable name: `block`. It's unnecessary to name the variable `blockToBeUndone` based on my opinion.

framework/src/modules/chain/modules/blocks/chain.js
pablitovicente: Yes I did the initial push with the basics working I'm about to commit the call to the undo only :)

framework/src/modules/chain/modules/blocks/chain.js
pablitovicente: Done

framework/src/modules/chain/modules/process_transactions.js
pablitovicente: Changed the code so it does not add transactions to the store as the accounts are the ones that are mutated on this operation and transactions are deleted by the FK `trs_blockId_fkey` on the `trs` table.

shuse2: was previously deleted automatically by FK or explicitly calling delete?

pablitovicente: Yes it was automatically deleted by the FK already.

framework/src/modules/chain/modules/blocks/chain.js
```diff
@@ -708,13 +692,7 @@ __private.popLastBlock = function(oldLastBlock, cb) {
 			.loadSecondLastBlockStep(oldLastBlock.previousBlock, tx)
 			.then(res => {
 				secondLastBlock = res;
-				return Promise.mapSeries(
-					oldLastBlock.transactions.reverse(),
-					transaction =>
-						__private
-							.undoConfirmedStep(transaction, oldLastBlock, tx)
-							.then(() => __private.undoUnconfirmStep(transaction, tx))
```
SargeKhan: Don't think we should remove undoUnconfirmStep just yet.

framework/test/mocha/integration/blocks/chain/delete_last_block_db_trs.js
```diff
@@ -151,16 +153,16 @@ describe('system test (blocks) - chain/popLastBlock', () => {
 					library.logic.account.merge = merge;
 					done();
 				});
-
```
SargeKhan: Why are these tests failing?

pablitovicente: The skipped ones are failing as they're expecting `account.merge()` error messages that are not the same returned by process transaction

framework/src/modules/chain/modules/blocks/chain.js
pablitovicente: Restored

framework/src/modules/chain/modules/blocks/chain.js
```diff
@@ -707,12 +691,13 @@ __private.popLastBlock = function(oldLastBlock, cb) {
 		__private
 			.loadSecondLastBlockStep(oldLastBlock.previousBlock, tx)
 			.then(res => {
+				secondLastBlock = res;
```
lsilvs: Duplicated line

pablitovicente: Fixed & Pushed.

MaciejBaj:
Please solve the conflicts.

framework/src/modules/chain/modules/blocks/chain.js
```diff
@@ -567,48 +567,32 @@ __private.loadSecondLastBlockStep = function(secondLastBlockId, tx) {
...
-						return setImmediate(resolve);
-					},
-					tx
-				);
-			},
-			tx
-		);
-	});
+__private.undoConfirmedStep = async function(block, tx) {
+	if (block.transactions.length <= 0) {
```
MaciejBaj: Why `<=` 0?

framework/src/modules/chain/modules/blocks/chain.js
```diff
@@ -567,48 +567,32 @@ __private.loadSecondLastBlockStep = function(secondLastBlockId, tx) {
...
+
+	const {
+		stateStore,
+		transactionResponses,
+	} = await modules.processTransactions.undoTransactions(
+		block.transactions,
+		tx
+	);
+
+	const unappliedTransactionResponse = transactionResponses.find(
```
MaciejBaj: Why is the error throwing logic here (in `__private.undoConfirmedStep`) instead of in `processTransactions.undoTransactions`?

SargeKhan: Because `processTransactions.undoTransactions` is going to be used in the transaction pool as well. And in transaction pool, we don't want the transaction responses to effectively decide which transactions to remove from the queue and which transactions to move to a different queue.

framework/src/modules/chain/modules/blocks/chain.js
pablitovicente: The same logic is being used for applying so I followed the same logic. Are you suggesting using `< 1` ?

framework/src/modules/chain/modules/blocks/chain.js
pablitovicente: I think that's the way process transactions was designed `applyTransactions` doesn't throw either and is left for the user to handle the error based on the state of `transactionResponses`

framework/src/modules/chain/modules/blocks/chain.js
MaciejBaj: More `=== 0`

framework/src/modules/chain/modules/blocks/chain.js
MaciejBaj: Ok.

framework/src/modules/chain/modules/blocks/chain.js
pablitovicente: Changed to `=== 0` after verifying that blocks are being pulled from the DB with the extended flag set to true, that means that transactions will always be an array (that might have length of zero).

PR #3018, opened 2019-03-04
Remove storageRead call from http_api to chain - Closes #2989

### LINKED ISSUE #2989: Remove usage of storageRead function on Chain actions
### Expected behavior
`storageRead` was intended to make the non-breaking changes for network layer. Any kind of formatting you need for blocks for the APIs, should be done in API module itself. So directly use `storage.entities.Block.get`, and format the response how the API is required.

Action can be removed and API module can directly perform the same logic itself instead of using the channel to call this action.

Discussed here:
https://github.com/LiskHQ/lisk/pull/2982/files/4667c5aa2129a7b604f2bb45220e62d5c0bee508#r260461947
### Actual behavior
`storageRead` is being used by Chain actions that are called by API module.

michielmulders:
### What was the problem?
`storageRead` was intended to make the non-breaking changes for the network layer. Any kind of formatting you need for blocks for the APIs, should be done in API module itself. So directly use `storage.entities.Block.get`, and format the response how the API is required.

### How did I fix it?
Remove `channel.invoke` call to `storageRead` in `http_api` module and move the API specific formatting logic for a block to the `http_api` module.

### How to test it?
Run tests.

### Comments
We can move the functionality of `parseBlockFromDatabase` into a `helper` object later.

### Review checklist

* The PR resolves #2989 
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/http_api/controllers/blocks.js
+++ b/framework/src/modules/http_api/controllers/blocks.js
@@ -15,9 +15,12 @@
 'use strict';
 
 const _ = require('lodash');
+const crypto = require('crypto');
 const apiCodes = require('../helpers/api_codes.js');
 const ApiError = require('../helpers/api_error.js');
+const Bignum = require('../helpers/bignum.js');
 
+const __private = {};
 let library;
 let sortFields;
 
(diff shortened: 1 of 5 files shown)


framework/src/modules/http_api/controllers/blocks.js
```diff
@@ -108,6 +111,67 @@ BlocksController.getBlocks = function(context, next) {
...
+	}
+
+	const block = {
+		id: raw.id,
+		version: parseInt(raw.version),
+		timestamp: parseInt(raw.timestamp),
+		height: parseInt(raw.height),
+		previousBlock: raw.previousBlockId,
+		numberOfTransactions: parseInt(raw.numberOfTransactions),
+		totalAmount: new Bignum(raw.totalAmount),
```
shuse2: if this is return as a JSON, this should be `string`, and same for other Bignum instances

framework/src/modules/http_api/controllers/blocks.js
```diff
@@ -108,6 +111,67 @@ BlocksController.getBlocks = function(context, next) {
...
+		generatorId: __private.getAddressByPublicKey(raw.generatorPublicKey),
+		blockSignature: raw.blockSignature,
+		confirmations: parseInt(raw.confirmations),
+	};
+
+	if (raw.transactions) {
+		block.transactions = raw.transactions;
+	}
+
+	block.totalForged = block.totalFee.plus(block.reward).toString();
```
shuse2: why not do this when creating the `block` above?

framework/src/modules/http_api/controllers/blocks.js
```diff
@@ -108,6 +111,67 @@ BlocksController.getBlocks = function(context, next) {
...
 
+/**
+ * Gets address by public.
+ *
+ * @private
+ * @param {publicKey} publicKey Public key
+ * @returns {address} address
+ * @todo Replace by Lisk Elements once integrated.
+ */
+__private.getAddressByPublicKey = function(publicKey) {
```
shuse2: avoid using `__private` in the new files. we can just create a function.
Also, lisk-cryptography has this function. can it be used in this version? @MaciejBaj

framework/src/modules/http_api/controllers/blocks.js
limiaspasdaniel: Perhaps we could add unit tests for this method.

framework/src/modules/http_api/controllers/blocks.js
michielmulders: Let's leave the function as is now, a TODO is present to address this later (use Lisk Elements here).

framework/src/modules/http_api/controllers/blocks.js
```diff
@@ -180,13 +242,7 @@ function _list(params, cb) {
 		library.storage.entities.Block.get(filters, options)
 			// FIXME: Can have poor performance because it performs SHA256 hash calculation for each block
 			.then(async rows =>
-				setImmediate(
-					cb,
-					null,
-					await Promise.all(
-						rows.map(row => library.channel.invoke('chain:storageRead', [row]))
```
MaciejBaj: Can `chain.storageRead` be removed then?

PR #3019, opened 2019-03-04
Integrate @liskhq/lisk-p2p module in chain module - Closes #2874

### LINKED ISSUE #2874: Create new network module
### Description

Using the channel concept defined in the architecture, network module should be created as standalone module.

Network module should integrate the @liskhq/lisk-p2p module and define `action` and `event` which can be used from chain module.

jondubois:
### What was the problem?

Lisk core needed to have the new Network module integrated and running so that other modules can start using it.

### How did I fix it?

- Created the Network module inside Lisk core.
- Passed configs to P2P library from the appropriate section of the Lisk config.
- Exposed actions from the Network module.
- Made the Network module publish events to the channel.
- Added support for invoking actions on modules from remote peers.

### How to test it?

Run the test suite.

### Review checklist

* The PR resolves #2874
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/components/system/index.js
+++ b/framework/src/components/system/index.js
@@ -16,8 +16,8 @@
 
 const System = require('./system');
 
-function createSystemComponent(config, logger, storage, onUpdateHandler) {
-	return new System(config, logger, storage, onUpdateHandler);
+function createSystemComponent(config, logger, storage) {
+	return new System(config, logger, storage);
 }
 
 module.exports = {
--- a/framework/src/components/system/index.js
+++ b/framework/src/components/system/index.js
@@ -16,8 +16,8 @@
 
 const System = require('./system');
 
-function createSystemComponent(config, logger, storage) {
-	return new System(config, logger, storage);
+function createSystemComponent(config, logger, storage, onUpdateHandler) {
+	return new System(config, logger, storage, onUpdateHandler);
 }
 
 module.exports = {
--- a/framework/src/components/system/system.js
+++ b/framework/src/components/system/system.js
@@ -17,6 +17,7 @@
 const os = require('os');
 const crypto = require('crypto');
 const semver = require('semver');
+const { EventEmitter } = require('events');
 
 /**
  * Main system methods. Initializes library with scope content and headers:
@@ -38,8 +39,9 @@ const semver = require('semver');
  * @param {Object} options - System options
  * @param {Object} logger
  */
-class System {
-	constructor(config, logger, storage, onUpdateHandler) {
+class System extends EventEmitter {
+	constructor(config, logger, storage) {
+		super();
 		this.logger = logger;
 		this.storage = storage;
 		this.headers = {
@@ -54,7 +56,6 @@ class System {
 			broadhash: config.nethash,
 			nonce: config.nonce,
 		};
-		this.onUpdateHandler = onUpdateHandler;
 	}
 
 	/**
@@ -139,9 +140,7 @@ class System {
 				.toString('hex');
 			this.headers.broadhash = newBroadhash;
 			this.logger.debug('System headers', this.headers);
-			if (this.onUpdateHandler) {
-				this.onUpdateHandler(this.headers);
-			}
+			this.emit('update', this.headers);
 			return true;
 		} catch (err) {
 			this.logger.error(err.stack);
--- a/framework/src/components/system/system.js
+++ b/framework/src/components/system/system.js
@@ -39,7 +39,7 @@ const semver = require('semver');
  * @param {Object} logger
  */
 class System {
-	constructor(config, logger, storage) {
+	constructor(config, logger, storage, onUpdateHandler) {
 		this.logger = logger;
 		this.storage = storage;
 		this.headers = {
@@ -54,6 +54,7 @@ class System {
 			broadhash: config.nethash,
 			nonce: config.nonce,
 		};
+		this.onUpdateHandler = onUpdateHandler;
 	}
 
 	/**
@@ -138,6 +139,9 @@ class System {
 				.toString('hex');
 			this.headers.broadhash = newBroadhash;
 			this.logger.debug('System headers', this.headers);
+			if (this.onUpdateHandler) {
+				this.onUpdateHandler(this.headers);
+			}
 			return true;
 		} catch (err) {
 			this.logger.error(err.stack);
--- a/framework/src/controller/schema/application.js
+++ b/framework/src/controller/schema/application.js
@@ -181,6 +181,15 @@ module.exports = {
 			},
 			modules: {
 				type: 'object',
+				properties: {
+					chain: {
+						type: 'object',
+					},
+					network: {
+						type: 'object',
+					},
+				},
+				additionalProperties: true,
 			},
 		},
 		additionalProperties: false,
--- a/framework/src/controller/schema/application.js
+++ b/framework/src/controller/schema/application.js
@@ -179,6 +179,9 @@ module.exports = {
 					},
 				},
 			},
+			modules: {
+				type: 'object',
+			},
 		},
 		additionalProperties: false,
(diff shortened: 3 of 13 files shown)


framework/src/controller/schema/application.js
```diff
@@ -179,6 +179,9 @@ module.exports = {
 					},
 				},
 			},
+			modules: {
+				type: 'object',
```
shuse2: should it be more specific?

framework/src/modules/network/defaults/index.js
```diff
@@ -0,0 +1,10 @@
+module.exports = {
+  blacklistedPeers: [],
```
ishantiw: should we also add `ackTimeout` as well?

framework/src/modules/network/index.js
```diff
@@ -0,0 +1,60 @@
...
+		return defaults;
+	}
+
+	get events() {
+		return [];
+	}
+
+	get actions() {
+		return {
+			request: async action => this.network.actions.request(action),
```
ishantiw: getNodeInfo need to be added as actions too? Maybe needed to get the height and also, applyNodeInfo in order to update the height? or maybe applyNodeInfo is should not be updated but network should listen to new block event to update its height?

framework/test/mocha/network/setup/config.js
```diff
@@ -56,6 +56,18 @@ const config = {
 			devConfigCopy.wsPort = 5000 + index;
 			devConfigCopy.httpPort = 4000 + index;
 			devConfigCopy.logFileName = `../logs/lisk_node_${index}.log`;
+			devConfigCopy.modules = {
+				network: {
+					blacklistedPeers: [],
+					connectTimeout: 5000,
```
ishantiw: ackTimeout here as well?

framework/src/modules/network/index.js
jondubois: I didn't expose it on purpose because P2PNodeInfo is currently the same as system peer headers and it's better to keep it in one place.

jondubois:
Note that the logic https://github.com/LiskHQ/lisk/pull/3019/files#diff-36244c4f244db393dce9137fe0db6fa1R52 currently allows any remote peer to call any action on any of our modules.

This needs to be addressed because some actions are only meant to be used by other modules on the same node. We need some kind of whitelisting to allow modules to choose what actions they want to expose to the network.

framework/src/modules/network/defaults/index.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+	blacklistedPeers: [],
+	connectTimeout: 5000,
+	ackTimeout: 5000,
```
shuse2: I think default `ackTimeout` should be `10000`, which is the case right now

framework/src/modules/network/network.js
```diff
@@ -0,0 +1,101 @@
...
+			} catch (error) {
+				request.error(error); // Send an error back to the peer.
+			}
+		});
+
+		this.p2p.on(EVENT_MESSAGE_RECEIVED, async packet => {
+			this.channel.publish(`network:${packet.event}`, packet.data);
+		});
+
+		try {
```
shuse2: no error event handling?

shuse2: this can be in other issue if needed =)

jondubois: Ah yep, I added event handling and logging.

framework/src/components/system/index.js
```diff
@@ -16,8 +16,8 @@
 
 const System = require('./system');
 
-function createSystemComponent(config, logger, storage) {
-	return new System(config, logger, storage);
+function createSystemComponent(config, logger, storage, onUpdateHandler) {
```
diego-G: Is this finally intended?

framework/src/components/system/system.js
```diff
@@ -138,6 +140,7 @@ class System {
 				.toString('hex');
 			this.headers.broadhash = newBroadhash;
 			this.logger.debug('System headers', this.headers);
+			this.emit('update', this.headers);
```
diego-G: The current solution seems weird to me. We emit update from the component, listen it on chain and republish again to the channel... I would prefer to publish to the channel directly in the chain module after calling the update() function. That way we avoid the modification of the system component.

framework/src/modules/network/index.js
```diff
@@ -0,0 +1,63 @@
...
+	get defaults() {
+		return defaults;
+	}
+
+	get events() {
+		return ['ready'];
+	}
+
+	get actions() {
+		return {
```
diego-G: Is this the standard to follow? Why not defining the actions in the constructor?

framework/src/modules/network/index.js
```diff
@@ -0,0 +1,63 @@
...
+			name: 'lisk-core-network',
+		};
+	}
+
+	get defaults() {
+		return defaults;
+	}
+
+	get events() {
+		return ['ready'];
```
diego-G: Why not defining the events in the constructor?

framework/src/modules/network/network.js
```diff
@@ -0,0 +1,166 @@
...
+				}),
+			getNetworkStatus: () => this.p2p.getNetworkStatus(),
+			applyPenalty: action => this.p2p.applyPenalty(action.params),
+		};
+	}
+
+	/* eslint-disable-next-line class-methods-use-this */
+	async cleanup() {
+		// TODO: Unsubscribe 'chain:system:updateNodeInfo' from channel.
+		await this.p2p.stop();
```
diego-G: `return` instead of await and remove the eslint disable next line.

framework/src/components/system/system.js
jondubois: I considered this approach but the update function was called in multiple places and some of those were too deep in the logic (it wouldn't make sense to pass a channel object there). Also, it's better if all channel actions/events are invoked/published from the top level of the module; it's more readable that way.

framework/src/modules/network/index.js
jondubois: That's how it's implemented in the chain module so I followed the pattern for consistency.

framework/src/modules/network/index.js
jondubois: The controller gets the list of events which the module supports using `moduleInstance.events`. You cannot return from the constructor so events could not be exposed to the controller that way.

framework/src/components/system/system.js
diego-G: I see the point. But in that case, I would prefer to publish to the channel straight from the component since this it's not restricted in the LIP.

diego-G: Please, @jondubois check if what you propose in this PR is aligned to https://github.com/LiskHQ/lisk/issues/3036.

jondubois: Not allowing components to publish directly to a channel is an important principle; it's better if components don't know about framework-level concepts.

This problem will be addressed properly by https://github.com/LiskHQ/lisk/issues/3036

diego-G: If that's right, we should update the LIP because it's stating the opposite. 
https://github.com/LiskHQ/lips/blob/master/proposals/lip-0005.md#components
```Component can use channels if required for implementation behavior. ```

jondubois: Yes I think this should be updated. It's related to the discussion we had last week about components managing state. The conclusion was that it would be better if components would not be responsible for managing/maintaining shared state because it breaks the `one source of truth` principle; they should behave more like utilities or clients.

framework/src/modules/network/index.js
limiaspasdaniel: Yes, actions need to be actually defined here, if they're defined in the constructor then they won't be registered and won't be able to be called.

Issue #3029, opened 2019-03-06
Refactor channel.invoke to not use callbacks as argument


michielmulders:
### Expected behavior
Not send a callback function as second/third parameter with the `channel.invoke` call.
Use promisify instead to add a callback and refactor the receiving part to `try-catch` with `await`.

### Actual behavior
Current: https://github.com/LiskHQ/lisk/blob/development/framework/src/modules/http_api/controllers/signatures.js#L46

### Which version(s) does this affect? (Environment, OS, etc...)
1.6.0

MaciejBaj:
Closed by https://github.com/LiskHQ/lisk/pull/3032

PR #3020, opened 2019-03-04
Remove Swagger from chain - Closes #2996

limiaspasdaniel:
### What was the problem?
Swagger was still being used in Chain module
### How did I fix it?
Removed usage of Swagger from Chain module and converted definitions to `Json Schema`. These definitions are validated with ZSchema for now without using `$refs`, until we get rid of it.
### How to test it?
Run all tests
### Review checklist
* The PR resolves #2996
* All new code is covered with unit tests
* All new code was formatted with Prettier
* Linting passes
* Tests pass
* Commit messages follow the [commit guidelines](CONTRIBUTING.md#git-commit-messages)
* Documentation has been added/updated

PR #3032, opened 2019-03-07
Add test coverage for moved api error codes - Closes #3021 #3029

### LINKED ISSUE #3021: Add test coverage for moved api error codes
### Expected behavior
Move the tests from `chain` to `http_api`

### Actual behavior
Still in `chain`, ref point: `framework/test/mocha/unit/modules/chain/modules/signatures.js:138`

1.6.0

michielmulders:
### What was the problem?
No tests created for the moved logic of `postSignature` and `postTransaction`.
In short, moved error logic from chain module to http_api module so we don't need to import API specific api_error codes into chain.

On the way doing this, we found callbacks being passed together with `invoke.channel` which is not allowed. All callbacks in `invoke.channel` functions have been modified to not pass it (using `promisify` down the chain).

### How did I fix it?
Added test coverage.
Refactor `invoke.channel` by removing callbacks as argument and callbacks in `chain.js`.
Modified tests accordingly.

### How to test it?
Run tests.

### Review checklist

* The PR resolves #3021 #3029 
* Linting passes
* Tests pass


framework/src/modules/chain/chain.js
```diff
@@ -217,16 +217,12 @@ module.exports = class Chain {
 					action.params[2]
 				),
 			getPeers: async action =>
-				this.scope.modules.peers.shared.getPeers(
-					action.params[0],
-					action.params[1]
-				),
+				promisify(this.scope.modules.peers.shared.getPeers)(action.params[0]),
```
lsilvs: Maybe not relevant to this PR but as you tackled it already, the functions `generateDelegateList`, `getUnProcessedTransactions`, `getUnconfirmedTransactions` and `getMultisignatureTransactions` also need to be promisify'd.

michielmulders: Ah good one, let's include this in the PR as we tackled half the problem here yet.

framework/src/modules/http_api/controllers/peers.js
```diff
@@ -60,33 +63,34 @@ PeersController.getPeers = async function(context, next) {
...
+		return next(null, {
+			data: filteredData,
+			meta: {
+				offset: filters.offset,
+				limit: filters.limit,
+				count: peersCount,
+			},
+		});
+	} catch (err) {
+		error = new ApiError(err, apiCodes.PROCESSING_ERROR);
```
limiaspasdaniel: Why is this a `PROCESSING_ERROR`?

framework/src/modules/http_api/controllers/signatures.js
```diff
@@ -73,12 +69,9 @@ SignaturesController.postSignature = async function(context, next) {
...
 	} catch (err) {
 		error = new ApiError(err, apiCodes.PROCESSING_ERROR);
-		console.log('inside');
-
-		context.statusCode = error.code;
-		delete error.code;
 	}
 
+	context.statusCode = error.code;
 	return next(error);
```
limiaspasdaniel: why is line 79 https://github.com/LiskHQ/lisk/pull/3032/commits/f26c3ebfeecc68f9e0008027cbc5b0fe791ccd72#diff-85d54af64d1362a64fb0e78e4c17ce65L79 removed?

limiaspasdaniel:
To be merged after #3020

framework/src/modules/http_api/controllers/signatures.js
michielmulders: Did it to have consistent behavior with transactions controller and write better tests using the `api_codes`. No breaking change.

framework/src/modules/http_api/controllers/peers.js
michielmulders: Reset back to: next(err) as this is breaking change, thanks!

PR #3034, opened 2019-03-07
Update bus to support RPC connections and relevant channel should be introduced - Closes #3013

### LINKED ISSUE #3013: Update bus to support RPC connections and relevant channel should be introduced
### Expected behavior
Bus should support RPC connections via sockets and channel which connects to it should be implemented

Use similar approach as proposed in https://github.com/LiskHQ/lisk-modular

lsilvs:
### What was the problem?
Bus communication was only possible by using event emitters but for child process sockets communication is required.

### How did I fix it?
It was implemented support to sub/pub and rpc connections via sockets to Bus and a new channel called `ChildProcessChannel` was introduced to communicate to Bus using sockets.

### How to test it?
Set `useSocketChannel` to true when registering the module
https://github.com/LiskHQ/lisk/blob/c3ecb6d83f462138c47f02a1f70f7594c5609c57/framework/src/controller/application.js#L129

### Review checklist

* The PR resolves #3013
* Linting passes
* Tests pass


framework/src/controller/controller.js
```diff
@@ -116,23 +119,24 @@ class Controller {
 	 * @async
 	 */
 	async _setupBus() {
-		this.bus = new Bus(this, {
+		this.bus = new Bus({
```
shuse2: ❤️

framework/src/modules/http_api/controllers/node.js
```diff
@@ -287,6 +287,12 @@ async function _getForgingStatus(publicKey) {
 	const forgersPublicKeys = {};
 
 	Object.keys(keyPairs).forEach(key => {
+		// Convert publicKey to buffer when received as object (ie.: { type: 'Buffer', data: [] })
```
jondubois: @lsilvs What can cause it to be received as an object with `type` of `Buffer`?
I would be better to always deal with strings if possible.

lsilvs: @jondubois Our RPC implementation convert the data before sending it back by doing `Buffer.from(JSON.stringify(message))`.
The `JSON.stringify` convert buffers to `{ type: 'Buffer', data: [] }`. That's why I had to convert it back here.

Ideally we should always transport strings but that is the current behaviour of `chain:getForgersKeyPairs` action.
This action will be removed as part of https://github.com/LiskHQ/lisk/issues/3033 anyway

jondubois: My view is that the complexity of converting to a Buffer is not worth it to begin with. The socket (Unix socket, TCP socket or whatever) will convert strings to binary anyway before sending them and it will probably do it at a lower level in a more efficient way.

This is similar to something we did in the past with block serialisation and tests showed that there was no benefit in that case so we removed it.

See https://github.com/LiskHQ/lisk/issues/1621 and https://github.com/LiskHQ/lisk/pull/1683

lsilvs: This conversion is done by https://github.com/tj/axon through https://github.com/tj/node-amp-message. Unless we decide not to use Axon, I would keep this behaviour.

jondubois: The solution is that we should avoid sending Buffers through the axon socket to begin with because that complicates the communication protocol between modules but I guess if it is the current behaviour and too time consuming to fix, we can do it later (as the TODO comment indicates).

framework/src/controller/channels/child_process_channel.js
```diff
@@ -0,0 +1,137 @@
...
+ * Channel responsible to communicate with bus for modules running in child process
+ *
+ * @class
+ * @memberof framework.controller.channels
+ * @requires module.Event
+ * @requires module.Action
+ * @requires channels/base_channel
+ * @type {module.ChildProcessChannel}
+ */
+class ChildProcessChannel extends BaseChannel {
```
limiaspasdaniel: What do you think of `CrossProcessChannel`? For me, this way seems more informative, describing that the channel works across different processes.

framework/src/controller/controller.js
```diff
@@ -147,15 +151,16 @@ class Controller {
...
-
 		// eslint-disable-next-line no-restricted-syntax
 		for (const alias of Object.keys(modules)) {
-			// eslint-disable-next-line no-await-in-loop
-			await this._loadInMemoryModule(
-				alias,
-				modules[alias].klass,
-				modules[alias].options
-			);
+			const { klass, options } = modules[alias];
```
limiaspasdaniel: What does `klass` mean here?

framework/src/controller/controller.js
```diff
@@ -147,15 +151,16 @@ class Controller {
...
+			if (options.useSocketChannel) {
+				// eslint-disable-next-line no-await-in-loop
+				await this._loadModuleWithSocketChannel(alias, klass, options);
```
limiaspasdaniel: I think `_loadModuleAsAChildProcess` or `_loadModuleAsASepareteProcess` would be clearer of what this method is doing (if I am not mistaken). Let me know what you think on this :)

package.json
```diff
@@ -44,6 +44,8 @@
 	"dependencies": {
 		"ajv": "6.7.0",
 		"async": "2.6.1",
+		"axon": "^2.0.3",
```
limiaspasdaniel: Are we using the `^` notation?

framework/src/controller/controller.js
lsilvs: This will be renamed in https://github.com/LiskHQ/lisk/issues/3059
For now there is no child process and this method is loading `httpApi` module using the `ChildProcessChannel` but not as a child process itself.

package.json
lsilvs: Good catch. It was copied from `lisk-modular` but shouldn't be used. I will remove it ;)

framework/src/controller/controller.js
lsilvs: `klass` is commonly used to name a variable that holds a class object and here it holds the module class the will be instantiated when loading the modules

package.json
```diff
@@ -44,6 +44,8 @@
...
+		"axon-rpc": "^0.0.3",
```
MaciejBaj: ``
```suggestion
		"axon-rpc": "0.0.3",
```

package.json
MaciejBaj:
```suggestion
		"axon": "2.0.3",
```

framework/src/controller/channels/child_process_channel.js
lsilvs: I agree, but let's leave as it is for now and change it when we have a better distinction of both `ChildProcessChannel` and `InMemoryChannel`.

Issue #3041, opened 2019-03-11
Update lisk scripts to work along new structure


MaciejBaj:
### Expected behavior
Lisk Scripts point to the updated entry point (from "app.js" to "src/index.js")
### Actual behavior
Lisk Scripts point to "app.js".
### Which version(s) does this affect? (Environment, OS, etc...)
`1.6.0`

PR #3044, opened 2019-03-11
Move "packaged" lisk-scripts - Closes #3042

### LINKED ISSUE #3042: Move "packaged" lisk-scripts into lisk repository
### Expected behavior
[packaged](https://github.com/LiskHQ/lisk-scripts/tree/development/packaged) scripts and configurations files live in LiskHQ/lisk and can be updated when there are changes in core (e.g. https://github.com/LiskHQ/lisk/issues/2848)

### Actual behavior
those scripts and configuration files live in a different repository and get pulled by the build process.

`1.6`

ghost:
### What was the problem?
Packaged scripts and configurations files lived in LiskHQ/lisk-scripts and could not be updated at the same time as the lisk core changed.

### How did I fix it?
Move "packaged" lisk-scripts (management scripts and configuration files) to `build/target/`; those get used instead of being downloaded during the build.

### How to test it?
Create build from this branch, ensure the resulting tarball does not significantly differ from one created from the `development` branch (binaries are not reproducible)

### Review checklist

* The PR resolves #3042
* Linting passes
* Tests pass

### DIFF
--- a/build/target/etc/pm2-lisk.json
+++ b/build/target/etc/pm2-lisk.json
@@ -1,0 +1,18 @@
+{
+        "apps": [{
+                "name": "lisk.app",
+                "script": "app.js",
+                "cwd": ".",
+                "pid_file": "./pids/lisk.app.pid",
+                "out_file": "./logs/lisk.app.log",
+                "error_file": "./logs/lisk.app.err",
+                "log_date_format": "YYYY-MM-DD HH:mm:ss SSS",
+                "watch": false,
+                "kill_timeout" : 10000,
+                "max_memory_restart": "1024M",
+                "node_args": "--max_old_space_size=1024",
+                "args": "-c config.json",
+                "min_uptime": 20000,
+                "max_restarts": 10
+        }]
+}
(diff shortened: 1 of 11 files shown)


build/target/etc/pm2-lisk.json
```diff
@@ -0,0 +1,18 @@
+{
+        "apps": [{
+                "name": "lisk.app",
+                "script": "app.js",
```
shuse2: probably this should be changed to `src/index.js`

shuse2 (changes requested):

build/target/etc/pm2-lisk.json
fchavant: Correct but this will be done in #3041

shuse2 (approved):

build/target/etc/redis.conf
```diff
@@ -0,0 +1,1316 @@
+# Redis configuration file example.
```
psychopenguin: Is it needed to have the full redis sample config?

psychopenguin (changes requested):

build/target/etc/redis.conf
fchavant: That's how we had it in LiskHQ/lisk-scripts (default config with a few changes); the advantage of having the full file is that it can be easily `diff`ed against a newer default configuration.

psychopenguin (approved):

MaciejBaj (approved):

PR #3049, opened 2019-03-11
Refactor transactions pool in lisk-core, and use transactions pool library - Closes #2829

### LINKED ISSUE #2829: Refactor transactions pool in lisk-core, and use transactions pool library
### Expected behavior
Transaction pool library implemented in lisk-elements should be incorporated in lisk-core.
### Actual behavior
Lisk-core does not use transaction pool library implemented in lisk-elements.

1.6.0

SargeKhan:
### What was the problem?
We needed to remove the old implementation of transaction pool and instead use the transaction pool library implemented in lisk-elements.
### How did I fix it?
I exposed similar API of the old transaction pool and internally used the transaction pool implemented in lisk-elements.
### How to test it?
Run unit tests.
### Review checklist

* The PR resolves #2829 
* Linting passes
* Tests pass


framework/src/modules/chain/logic/transaction_pool.js
```diff
@@ -14,25 +14,46 @@
...
-let self;
 const {
 	EXPIRY_INTERVAL,
 	MAX_TRANSACTIONS_PER_BLOCK,
 	MAX_SHARED_TRANSACTIONS,
-	UNCONFIRMED_TRANSACTION_TIMEOUT,
 } = global.constants;
-const __private = {};
+
+const wrapAddTransactionResponseInCb = (
```
pablitovicente: Why callbacks instead of async/await or promises ?

framework/src/modules/chain/modules/transactions.js
```diff
@@ -253,6 +250,17 @@ Transactions.prototype.transactionInPool = function(id) {
...
 
+/**
+ * Fills pool.
+ *
+ * @param {function} cb - Callback function
+ * @returns {function} Calls transactionPool.fillPool
+ * @todo Add description for the params
+ */
+Transactions.prototype.fillPool = function(cb) {
+	return __private.transactionPool.fillPool(cb);
```
shuse2: probably transaction pool shouldn't be referenced in `transactions` module.
Maybe we can fix in some other issue?

framework/src/modules/chain/modules/transport.js
```diff
@@ -216,15 +216,16 @@ __private.receiveTransactions = function(
...
  */
 __private.receiveTransaction = function(
-	transaction,
+	transactionJSON,
 	nonce,
 	extraLogMessage,
 	cb
 ) {
-	const id = transaction ? transaction.id : 'null';
+	const id = transactionJSON ? transactionJSON.id : 'null';
```
shuse2: should it still insert `null`?

framework/src/modules/chain/logic/transaction_pool.js
```diff
@@ -42,1058 +63,220 @@ const __private = {};
...
-		limit
-	);
-};
+	/**
+	 * Bounds input parameters to private and sets up the liskhq-transaction-pool module.
+	 *
+	 * @param {Transactions} processTransactions - Transactions processing module instance
+	 */
+	bind(processTransactions) {
+		modules = {
```
pablitovicente: Wouldn't it be better to avoid using a module variable and use a class property instead? That way it's simple to test/mock

framework/src/modules/chain/modules/transport.js
SargeKhan: It was working before and I didn't feel the need to change the functionality here.

framework/src/modules/chain/modules/transactions.js
SargeKhan: Why do you think it shouldn't be referenced here?

framework/src/modules/chain/logic/transaction_pool.js
SargeKhan: The wrapping functions by themselves return either synchronous response or return a promise. I am wrapping the return of those functions in callback because I wanted to keep the same interface as the old transaction pool.

framework/src/modules/chain/logic/transaction_pool.js
```diff
@@ -14,25 +14,46 @@
...
+	addTransactionResponse,
+	cb,
+	transaction,
+	err
+) => {
+	if (err) {
+		return cb(err);
+	}
+	if (addTransactionResponse.isFull) {
+		return cb('Transaction pool is full');
```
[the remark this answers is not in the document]
SargeKhan: Yes, I am returning the same error string as we were returning in the old transaction pool.

framework/src/modules/chain/logic/transaction_pool.js
SargeKhan: Sure! I was refactoring code and kept modules variable as defined in the old implementation. I will change it.

framework/src/modules/chain/logic/transaction_pool.js
```diff
@@ -42,1058 +63,220 @@ const __private = {};
...
 
-/**
- * Counts the number of transactions in the queued index.
- *
- * @returns {number} Of transactions in the queued index
- */
-TransactionPool.prototype.countQueued = function() {
-	return Object.keys(self.queued.index).length;
-};
+		for (let i = 0; i < transactions.length; i++) {
```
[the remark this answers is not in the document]
SargeKhan: This function was implemented in the old transaction pool. I guess here, in this loop, we are creating a new array for transactions. Perhaps this functionality was added to ensure that transactions returned aren't modified.

framework/src/modules/chain/logic/transaction_pool.js
```diff
@@ -42,1058 +63,220 @@ const __private = {};
...
- * @todo Compare / standardize the returns-description
- */
-TransactionPool.prototype.processBundled = function(cb) {
-	if (modules.loader.syncing()) {
-		// There is no need to process bundled transacions if node is syncing
-		return setImmediate(cb);
+	fillPool(cb) {
+		return this.pool
+			.processVerifiedTransactions()
+			.then(() => cb())
```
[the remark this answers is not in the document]
SargeKhan: No, we can not do it because `processVerifiedTransactions` returns a response. if we change the code to: `.then(cb)` then the `cb`'s first parameter - which is usually error - will have transactionsResponses and we do not want that.

framework/src/modules/chain/modules/transactions.js
shuse2: nvm, i think the direction of dependency is correct 👍

framework/src/modules/chain/logic/transaction_pool.js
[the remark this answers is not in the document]
shuse2: probably it's better to do
`const clonedTransaction = [...transactions];`

PR #3057, opened 2019-03-12
Functional tests for validating asset field API - Closes #3053

### LINKED ISSUE #3053: Improve functional test coverage for asset field API response
### Expected behavior
No `asserts` present to test the structure of asset fields if it holds the required properties.

### Actual behavior
Add test coverage.

1.6.0

michielmulders:
### What was the problem?
No checks for properties of asset field for each type returned by `API` when querying `/transactions` endpoint.

`http://localhost:4000/api/transactions?type=N`

### How did I fix it?
Add tests.

### How to test it?
Run tests for file.

### Review checklist

* The PR resolves /
* Linting passes
* Tests pass

### DIFF
--- a/framework/test/mocha/functional/http/get/transactions.js
+++ b/framework/test/mocha/functional/http/get/transactions.js
@@ -19,6 +19,7 @@ const Promise = require('bluebird');
 const {
 	transfer,
 	registerSecondPassphrase,
+	castVotes,
 } = require('@liskhq/lisk-transactions');
 const accountFixtures = require('../../../fixtures/accounts');
 const randomUtil = require('../../../common/utils/random');
@@ -70,6 +71,10 @@ describe('GET /api/transactions', () => {
 		passphrase: accountFixtures.genesis.passphrase,
 		recipientId: accountSecondPass.address,
 		data: 'tx 5',
+	});
+	const transactionType3 = castVotes({
+		passphrase: account2.passphrase,
+		votes: [`${accountFixtures.existingDelegate.publicKey}`],
 	});
 	const transactionType5 = {
 		amount: '0',
@@ -130,6 +135,7 @@ describe('GET /api/transactions', () => {
 			'8249786301f5cc7184b0681563dc5c5856568ff967bec22b778f773b0a86532b13d1ede9234f581e62388ada2d1e1366adaa03151a9e6508fb7c3a3e59425109',
 		id: '18307756018345914129',
 	};
+
 	// Crediting accounts'
 	before(() => {
 		const promises = [];
@@ -145,11 +151,15 @@ describe('GET /api/transactions', () => {
 			return waitFor
 				.confirmations(_.map(transactionList, 'id'))
 				.then(() => {
-					return apiHelpers.sendTransactionPromise(transaction3);
+					return Promise.all([
+						apiHelpers.sendTransactionPromise(transaction3),
+						apiHelpers.sendTransactionPromise(transactionType3),
+					]);
 				})
 				.then(() => {
 					transactionList.push(transaction3);
-					return waitFor.confirmations([transaction3.id]);
+					transactionList.push(transactionType3);
+					return waitFor.confirmations([transaction3.id, transactionType3.id]);
 				});
 		});
 	});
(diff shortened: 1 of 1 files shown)


framework/test/mocha/functional/http/get/transactions.js
```diff
@@ -357,6 +357,129 @@ describe('GET /api/transactions', () => {
...
+
+			describe('asset field', () => {
+				it('using type 0 should return asset field with correct properties', async () => {
+					return transactionsEndpoint
+						.makeRequest({ type: transactionTypes.SEND }, 200)
+						.then(res => {
+							expect(res.body.data).to.not.empty;
+
+							res.body.data.map(transaction => {
+								expect(transaction).to.have.property('asset');
```
yatki: This check is redundant because `Object.keys(transaction.asset)` would throw exception if asset is undefined.

framework/test/mocha/functional/http/get/transactions.js
```diff
@@ -357,6 +357,129 @@ describe('GET /api/transactions', () => {
...
+				});
+
+				it('using type 1 should return asset field with correct properties', async () => {
+					return transactionsEndpoint
+						.makeRequest({ type: transactionTypes.SIGNATURE }, 200)
+						.then(res => {
+							expect(res.body.data).to.not.empty;
+
+							res.body.data.map(transaction => {
+								expect(transaction).to.have.property('asset');
```
yatki: This check is redundant. The next assertions are validating the asset property.

framework/test/mocha/functional/http/get/transactions.js
```diff
@@ -357,6 +357,129 @@ describe('GET /api/transactions', () => {
...
+
+				it('using type 2 should return asset field with correct properties', async () => {
+					return transactionsEndpoint
+						.makeRequest({ type: transactionTypes.DELEGATE }, 200)
+						.then(res => {
+							expect(res.body.data).to.not.empty;
+
+							res.body.data.map(transaction => {
+								expect(transaction).to.have.property('asset');
+								expect(transaction.asset).to.have.property('delegate');
```
yatki: These 2 checks are redundant as well.

framework/test/mocha/functional/http/get/transactions.js
```diff
@@ -357,6 +357,129 @@ describe('GET /api/transactions', () => {
...
+
+				it('using type 3 should return asset field with correct properties', async () => {
+					return transactionsEndpoint
+						.makeRequest({ type: transactionTypes.VOTE }, 200)
+						.then(res => {
+							expect(res.body.data).to.not.empty;
+
+							res.body.data.map(transaction => {
+								expect(transaction).to.have.property('asset');
+								expect(transaction.asset).to.have.property('votes');
```
yatki: These 2 checks above are redundant as well.

framework/test/mocha/functional/http/get/transactions.js
```diff
@@ -357,6 +357,129 @@ describe('GET /api/transactions', () => {
...
+
+				it('using type 4 should return asset field with correct properties', async () => {
+					return transactionsEndpoint
+						.makeRequest({ type: transactionTypes.MULTI }, 200)
+						.then(res => {
+							expect(res.body.data).to.not.empty;
+
+							res.body.data.map(transaction => {
+								expect(transaction).to.have.property('asset');
+								expect(transaction.asset).to.have.property('multisignature');
```
yatki: These 2 checks above are redundant.

framework/test/mocha/functional/http/get/transactions.js
```diff
@@ -357,6 +357,129 @@ describe('GET /api/transactions', () => {
...
+								expect(transaction.asset.multisignature).to.have.property(
+									'min'
+								);
+								expect(transaction.asset.multisignature).to.have.property(
+									'lifetime'
+								);
+								expect(transaction.asset.multisignature).to.have.property(
+									'keysgroup'
+								);
+								expect(transaction.asset.multisignature.keysgroup).to.be.an(
```
yatki: This assertion makes the previous one redundant.

framework/test/mocha/functional/http/get/transactions.js
```diff
@@ -357,6 +357,129 @@ describe('GET /api/transactions', () => {
...
+
+				it('using type 5 should return asset field with correct properties', async () => {
+					return transactionsEndpoint
+						.makeRequest({ type: transactionTypes.DAPP }, 200)
+						.then(res => {
+							expect(res.body.data).to.not.empty;
+
+							res.body.data.map(transaction => {
+								expect(transaction).to.have.property('asset');
+								expect(transaction.asset).to.have.property('dapp');
```
yatki: same as above.

framework/test/mocha/functional/http/get/transactions.js
```diff
@@ -357,6 +357,111 @@ describe('GET /api/transactions', () => {
...
+					return transactionsEndpoint
+						.makeRequest({ type: transactionTypes.MULTI }, 200)
+						.then(res => {
+							expect(res.body.data).to.not.empty;
+
+							res.body.data.map(transaction => {
+								expect(Object.keys(transaction.asset).length).to.equal(1);
+								expect(transaction.asset.multisignature).to.have.property(
+									'min'
+								);
```
MaciejBaj: `.to.be.within(2,15)`

framework/test/mocha/functional/http/get/transactions.js
```diff
@@ -357,6 +357,111 @@ describe('GET /api/transactions', () => {
...
+								expect(transaction.asset.multisignature).to.have.property(
+									'lifetime'
+								);
```
MaciejBaj: `.to.be.within(1,72)`

framework/test/mocha/functional/http/get/transactions.js
```diff
@@ -357,6 +357,111 @@ describe('GET /api/transactions', () => {
...
+
+				it('using type 3 should return asset field with correct properties', async () => {
+					return transactionsEndpoint
+						.makeRequest({ type: transactionTypes.VOTE }, 200)
+						.then(res => {
+							expect(res.body.data).to.not.empty;
+
+							res.body.data.map(transaction => {
+								expect(Object.keys(transaction.asset).length).to.equal(1);
+								return expect(transaction.asset.votes).to.be.an('array');
```
MaciejBaj: Please validate the array length - up to 33 votes

michielmulders: @MaciejBaj this does not work as the transaction from `Genesis_block.json` holds a tx with 101 votes although limit is 33. What about this?

michielmulders:
Test fails because votes array for `type 3` transaction should be between `1` and `33` (edges inlcuded). However, the `genesis_block.json` holds a transaction with `101` votes so this causes the failed test.

framework/test/mocha/functional/http/get/transactions.js
michielmulders: Filter out Genesis transaction by account.

framework/test/mocha/functional/http/get/transactions.js
```diff
@@ -356,6 +357,120 @@ describe('GET /api/transactions', () => {
...
+					return transactionsEndpoint
+						.makeRequest({ type: TRANSACTION_TYPES.SIGNATURE }, 200)
+						.then(res => {
+							expect(res.body.data).to.not.empty;
+
+							res.body.data.map(transaction => {
+								expect(transaction.asset).to.have.property('signature');
+								expect(transaction.asset.signature).to.have.property(
+									'publicKey'
+								);
```
yatki: These two assertions are redundant.

framework/test/mocha/functional/http/get/transactions.js
```diff
@@ -356,6 +357,120 @@ describe('GET /api/transactions', () => {
 						});
 					});
 			});
+
+			describe('asset field', () => {
+				it('using type 0 should return asset field with correct properties', async () => {
```
yatki: I don't think we need async statement since we already returning promise right away. The same usage reoccurs in the tests below.

yatki: Also, I'd prefer to keep async and instead of doing assertions in the then callback, I'd simply await the request.

framework/test/mocha/functional/http/get/transactions.js
```diff
@@ -356,6 +357,120 @@ describe('GET /api/transactions', () => {
...
+					return transactionsEndpoint
+						.makeRequest({ type: TRANSACTION_TYPES.VOTE }, 200)
+						.then(res => {
+							expect(res.body.data).to.not.empty;
+							// Skip Genesis vote transaction - exception as it contains 101 votes
+							const transactionsType3 = res.body.data.filter(
+								transaction =>
+									transaction.recipientId !== '16313739661670634666L'
+							);
+							transactionsType3.map(transaction => {
```
lsilvs: Add an expectation like `expect(transactionsType3.length).to.greaterThan(0)` to make sure we have data to test the next line.

yatki: @lsilvs expect(res.body.data).to.not.empty; is kind of ensuring that tho @michielmulders I personally prefer to check length as well because empty matcher can be used for strings as well.

lsilvs: @yatki  But after `expect(res.body.data).to.not.empty` the `res.body.data` is being filtered meaning it can result in an empty array.

framework/test/mocha/functional/http/get/transactions.js
```diff
@@ -146,11 +151,15 @@ describe('GET /api/transactions', () => {
 			return waitFor
 				.confirmations(_.map(transactionList, 'id'))
 				.then(() => {
-					return apiHelpers.sendTransactionPromise(transaction3);
+					const promises2 = [];
+					promises2.push(apiHelpers.sendTransactionPromise(transaction3));
+					promises2.push(apiHelpers.sendTransactionPromise(transactionType3));
+					return Promise.all(promises2);
```
yatki: I think having promises2 array doesn't bring that much value. Maybe it's better to move the calls inside Promise.all as an array ?

michielmulders: Yes, you are right, looks weird now, thanks for fast review!

PR #3069, opened 2019-03-13
Refactor channel.invoke to return only valid data - Closes #3068

### LINKED ISSUE #3068: Refactor channel.invoke to return only valid data
### Expected behavior
`channel.invoke` should not return `function`, `date` or `undefined`.

When the channel communication is done via sockets it only transport data as string, buffer or json and the mentioned values are not valid json values, hence they are excluded from the response.

Additionally, by design the invoke action should always execute an action and return the data and not the function itself.

### Actual behavior
`chain:getSlotsHelper` returns functions

### Steps to reproduce
`const slots = await channel.invoke('chain:getSlotsHelper');`

`development`

lsilvs:
### What was the problem?
`channel.invoke` returning functions

### How did I fix it?
I've created one action per function

### How to test it?
Regular test suite

### Review checklist

* The PR resolves #3068
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/chain/chain.js
+++ b/framework/src/modules/chain/chain.js
@@ -239,13 +239,13 @@ module.exports = class Chain {
 				),
 			getSlotTime: async action =>
 				action.params
-					? this.slots.getTime(action.params[0])
+					? this.slots.getTime(action.params.time)
 					: this.slots.getTime(),
 			getSlotNumber: async action =>
 				action.params
-					? this.slots.getSlotNumber(action.params[0])
+					? this.slots.getSlotNumber(action.params.epochTime)
 					: this.slots.getSlotNumber(),
-			calcSlotRound: async action => this.slots.calcRound(action.params[0]),
+			calcSlotRound: async action => this.slots.calcRound(action.params.height),
 		};
 	}
 
--- a/framework/src/modules/chain/chain.js
+++ b/framework/src/modules/chain/chain.js
@@ -237,7 +237,15 @@ module.exports = class Chain {
 					action.params[0],
 					action.params[1]
 				),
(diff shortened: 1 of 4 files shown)


framework/src/modules/chain/chain.js
```diff
@@ -237,7 +237,15 @@ module.exports = class Chain {
...
 					action.params[1]
 				),
-			getSlotsHelper: async () => this.slots,
+			getSlotTime: async action =>
+				action.params
+					? this.slots.getTime(action.params[0])
+					: this.slots.getTime(),
+			getSlotNumber: async action =>
+				action.params
+					? this.slots.getSlotNumber(action.params[0])
```
michielmulders: Would you like to use named parameters? This has just been merged into dev. Ref: https://github.com/LiskHQ/lisk/pull/3065

`getSlotNumber(action.params.num)`

framework/src/modules/http_api/controllers/delegates.js
```diff
@@ -184,13 +184,15 @@ async function _getForgers(filters) {
...
-	const currentSlot = slots.getSlotNumber();
+	const lastBlockSlot = await channel.invoke('chain:getSlotNumber', [
+		lastBlock.timestamp,
+	]);
+	const currentSlot = await channel.invoke('chain:getSlotNumber');
 	const forgerKeys = [];
-	const round = slots.calcRound(lastBlock.height + 1);
+
+	const round = await channel.invoke('chain:calcSlotRound', [
+		lastBlock.height + 1,
```
michielmulders: Suggest to use named param here as well, ref: https://github.com/LiskHQ/lisk/pull/3065

`channel.invoke('action', { height: lastBlock.height + 1 });`

PR #3076, opened 2019-03-14
Convert system component to a general application state - Closes #3036

### LINKED ISSUE #3036: Change system component to application state
### Overview
Replace the current system component, maintained by all of the modules independently by one global application state, which changes on can be subscribed on by all of the modules and changes to can be submitted after the proper authorization. 

### Rationale
Current system component definition based on LIP:

> **System**
This component will provide a central registry of up-to-date system information. Especially network height, nonce, broadhash, nethash and network specific constants. This component will use channels and events to make all instances of the component stay in sync in different modules.

There is a concern about keeping all of the system components created by each module independently in sync, especially by modules working on separate processes. For instance, when the flag “syncing” is not set at the same time in network and chain modules it is possible that a node will attempt to forge in chain module while syncing in network module at the same time which leads to the node staying on its own fork.

One source of truth about the application knowledge would introduce the possibility of reloading modules in runtime after the height (stored in applicationState) changes. It's one of the ways which is being discussed now internally to support the blockchain data compatibility with legacy protocols. 

Having the shared global state means that the additional information published by modules can be also stored in the applicationState and used by other modules in a flexible way.

### Specification
Variables currently stored by the system component should be included in the applicationState. 

The applicationState is global for the whole system and serves as one source of truth for all of the registered modules within the application.

The applicationState implemented by Lisk Framework controller is loosely inspired by [Redux](https://redux.js.org/).

#### Read
Every change to the applicationState triggers the `applicationState:updated` event, which can be subscribed on by any module. A module reacts then to the changes to only the module-relevant parts of the applicationStore which are updated by following the reduce pattern in `applicationState:updated` event handlers.

#### Write
Changes to the application state can be proposed by any module by calling `applicationState.dispatch` method. 

*Security* aspect cannot be overlooked and the write access of modifying particular variables critical to the blockchain sanity will be restricted to the **application controller** and **chain module** only. 

The following variables can only be assigned once after the application starts and only by **application controller**:
- os
- nonce
- version
- minVersion
- protocolVersion
- nethash

The following variables can only be modified by the **chain module**:
- height
- broadhash

All of the other modules (also the custom ones) can loosely add and modify the custom keys in the applicationState. 

`1.6.0`

diego-G:
### Review checklist

* The PR resolves #3036 
* Linting passes
* Tests pass


framework/src/controller/applicationState.js
```diff
@@ -94,56 +87,39 @@ class System {
...
-	}
-
-	/**
-	 * Updates private broadhash and height values.
+	 * Updates broadhash and height values.
 	 *
 	 * @returns {Promise.<boolean, Error>}
 	 */
-	async update() {
+	async update(blocks) {
```
shuse2: application itself cannot have access to database

framework/src/controller/application.js
```diff
@@ -119,6 +124,11 @@ class Application {
 		__private.modules.set(this, {});
 		__private.transactions.set(this, {});
 
+		this.applicationState = new ApplicationState(
```
shuse2: considering multi-process application, ApplicationState need to emit/subscribe to event, and not passing instance to other modules. Please talk to @jondubois and @lsilvs regards to this

framework/src/controller/applicationState.js
diego-G: I don't understand this comment. I've removed the storage from the new applicationState 83ebc756281cd0921587de9c208db7361f4348d1

framework/src/controller/controller.js
```diff
@@ -42,13 +42,15 @@ class Controller {
 	 * @param {string} appLabel - Application label
 	 * @param {Object} componentConfig - Configurations for components
 	 * @param {component.Logger} logger - Logger component responsible for writing all logs to output
+	 * @param {applicationState} applicationState - General application state
 	 */
-	constructor(appLabel, componentConfig, logger) {
+	constructor(appLabel, componentConfig, logger, applicationState) {
```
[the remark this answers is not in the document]
diego-G: What would be the difference?

framework/src/controller/applicationState.js
```diff
@@ -1,3 +1,4 @@
+/* eslint-disable class-methods-use-this */
```
MaciejBaj: Rename to application_state

framework/src/controller/applicationState.js
```diff
@@ -35,38 +36,29 @@ const semver = require('semver');
...
  * @param {Object} logger
  */
-class System {
-	constructor(config, logger, storage) {
+class ApplicationState {
+	constructor(initialState, logger, channel) {
 		this.logger = logger;
-		this.storage = storage;
-		this.headers = {
+		this.state = {
```
MaciejBaj: needs to state private - const state to avoid overwriting it by other modules.

framework/src/controller/applicationState.js
```diff
@@ -35,38 +36,29 @@ const semver = require('semver');
...
-	 * @param {string} nethash
-	 * @returns {boolean}
-	 */
-	networkCompatible(nethash) {
-		if (!nethash) {
-			return false;
-		}
-		return this.headers.nethash === nethash;
+	getState() {
+		return this.state;
```
MaciejBaj: return state;

framework/src/controller/applicationState.js
```diff
@@ -35,38 +36,29 @@ const semver = require('semver');
...
-	 *
-	 * @param {string} nethash
-	 * @returns {boolean}
-	 */
-	networkCompatible(nethash) {
-		if (!nethash) {
-			return false;
-		}
-		return this.headers.nethash === nethash;
+	getState() {
```
MaciejBaj: `setState` is missing. 

`setState` can only update the state if a module has a sufficient authorization i.e as described in issue #3036 
> The following variables can only be assigned once after the application starts and only by application controller:
os
nonce
version
minVersion
protocolVersion
nethash
The following variables can only be modified by the chain module:
height
broadhash

MaciejBaj: As discussed in the offline meeting, to do this, the corresponding channel authorization needs to be implemented, for application to be able to verify a channel of which module is invoking the `setState`.

diego-G: The state is initiated in the constructor passing the config variables as argument (`initialState`). What we need to moderate is which module is able to call the update action.

framework/src/controller/channels/event_emitter.js
```diff
@@ -92,7 +92,7 @@ class EventEmitterChannel extends BaseChannel {
 	 * @param {array} params - Params associated with the action
 	 * @return {Promise<string>} Data returned by bus.
 	 */
-	async invoke(actionName, params) {
+	invoke(actionName, params) {
```
shuse2: `invoke` has to be async?

framework/src/controller/application_state.js
```diff
@@ -0,0 +1,98 @@
...
+	setChannel(channel) {
+		this.channel = channel;
+	}
+
+	/**
+	 * Updates broadhash and height values.
+	 *
+	 * @returns {Promise.<boolean, Error>}
+	 */
+	async update(blocks) {
```
shuse2: this whole thing should be in `submodules/blocks/chain.js`.
instead we need a general way to update the application state

framework/src/controller/application.js
```diff
@@ -250,11 +254,15 @@ class Application {
...
-		this.controller = new Controller(
-			this.label,
-			this.config.components,
-			this.logger
-		);
+		this.controller = new Controller({
+			label: this.label,
+			config: {
+				components: this.config.components,
+				initialState: this.config.initialState,
```
shuse2: since all the information is in the config, can we not use the initState input, and create initial state internally?

framework/src/controller/controller.js
```diff
@@ -123,15 +137,21 @@ class Controller {
...
+				updateApplicationState: action =>
+					this.applicationState.update(action.params),
 			},
 			this.bus,
 			{ skipInternalEvents: true }
 		);
+
 		await this.channel.registerToBus();
 
+		await this.applicationState.setChannel(this.channel);
```
shuse2: should it have access to channel? maybe to `bus` instead?

framework/src/controller/channels/event_emitter.js
diego-G: Ideally it has, but we cannot force to use it that way with the current status of our code. This is a key point to go forward with this issue. I had a talk with @lsilvs about it. Please, raise any concern in any case where removing the mandatory async use of this function creates a problem.

framework/src/controller/application_state.js
diego-G: This will be to get back to the starting point where system was a chain submodule. Therefore I don't agree.

framework/src/controller/controller.js
diego-G: Why bus? I think channel is the correct.

framework/src/controller/application_state.js
diego-G: https://github.com/LiskHQ/lisk/issues/3141

framework/src/controller/channels/event_emitter.js
diego-G: Reverted.

framework/src/controller/application_state.js
```diff
@@ -0,0 +1,99 @@
...
+		if (!blocks) {
+			throw new TypeError('Argument blocks should be an array.');
+		}
+		try {
+			if (blocks.length <= 1) {
+				this.state.broadhash = this.state.nethash;
+				return true;
+			}
+			this.state.height = blocks[0].height;
```
nazarhussain: This assumes that blocks array will always be sorted by height, which may not be true and cause to logical error. I suggest find highest height first.

framework/src/controller/application_state.js
```diff
@@ -0,0 +1,99 @@
...
+ * @param {Object} logger
+ */
+class ApplicationState {
+	constructor(initialState, logger) {
+		this.logger = logger;
+		this.state = {
+			os: os.platform() + os.release(),
+			version: initialState.version,
+			wsPort: initialState.wsPort,
+			httpPort: initialState.httpPort,
```
nazarhussain: As we are already refactoring this whole concept, then we should reconsider this approach. `wsPort` and `httpPort` have no relation to application state and they must not be mentioned here.

framework/src/controller/application_state.js
```diff
@@ -0,0 +1,99 @@
...
+ * - nonce
+ *
+ * @class
+ * @requires crypto
+ * @requires os
+ * @param {Object} initialState - Initial state of the application
+ * @param {Object} logger
+ */
+class ApplicationState {
+	constructor(initialState, logger) {
```
nazarhussain: Use object destructure format to specify the state parameters you are expecting. 

```
constructor({version, wsPort, ...}, logger) {
```

framework/src/controller/application_state.js
```diff
@@ -0,0 +1,99 @@
...
+			this.state.height = blocks[0].height;
+			const seed = blocks.map(row => row.id).join('');
+			const newBroadhash = crypto
+				.createHash('sha256')
+				.update(seed, 'utf8')
+				.digest()
+				.toString('hex');
+			this.state.broadhash = newBroadhash;
+			this.logger.debug('Application state', this.state);
+			this.channel.publish('lisk:applicationStateUpdated', this.state);
```
nazarhussain: Reconsider the naming pattern for events and avoid using such long names. We have `:` separator pattern that is useful while subscribing to events. So I suggest you to go with `lisk:state:updated` so in future if there is any other event related to application state, must also be grouped in same pattern.

framework/src/controller/bus.js
```diff
@@ -170,6 +170,36 @@ class Bus extends EventEmitter2 {
 		});
 	}
 
+	// TO REMOVE
```
nazarhussain: Why this code was added and when to be removed? Please refer the issues here.

framework/src/controller/channels/in_memory_channel.js
```diff
@@ -113,6 +113,29 @@ class InMemoryChannel extends BaseChannel {
 
 		return this.bus.invoke(action.serialize());
 	}
+
+	// TO REMOVE
```
nazarhussain: Same here?

framework/src/controller/controller.js
```diff
@@ -126,15 +140,20 @@ class Controller {
 
 		this.channel = new InMemoryChannel(
 			'lisk',
-			['ready'],
+			['ready', 'applicationStateUpdated'],
```
nazarhussain: Refactored this name as suggested earlier.

nazarhussain:
@diego-G As per discussion, please align the usage of application state to be more event driven approach rather than invoking every action every time. Action are good to have get initial state once and then rely on events to get the updates. The use of current approach in http-api can blow up the IPC layer if we have a lot of traffic and IPC is not meant to be designed for it.

framework/src/controller/bus.js
jondubois: We should start the comment with `TODO: ` so that it is easier to search for it. Also, we should specify when it should be removed; which is as part of this issue: https://github.com/LiskHQ/lisk/issues/2883 (you can put the link in the comment) because this is a hack and we should make sure that nobody on our team will be using this function.

framework/src/controller/application_state.js
jondubois: Yes this should be changed but we can do it as a new issue like you suggest.

framework/src/controller/application_state.js
diego-G: We will reconsider this in the P2P epic. Thanks for the reminder.

framework/src/controller/application_state.js
diego-G: I'll work this out in https://github.com/LiskHQ/lisk/issues/3141

framework/src/controller/application_state.js
```diff
@@ -0,0 +1,110 @@
...
+			height: 1,
+			nethash,
+			broadhash: nethash,
+			nonce,
+		};
+	}
+
+	getState() {
+		const state = _.cloneDeep(this.state);
+		return state;
```
nazarhussain:
```suggestion
		return  _.cloneDeep(this.state);
```

framework/src/controller/application_state.js
```diff
@@ -0,0 +1,110 @@
...
+			version,
+			wsPort,
+			httpPort,
+			minVersion,
+			protocolVersion,
+			height: 1,
+			nethash,
+			broadhash: nethash,
+			nonce,
+		};
```
nazarhussain: When you assign `state` this `this` then its accessible and modifiable without calling the `update`. So it would be nice to use the `WeakMap` approach that we used actually have private data members. 

Then you can simply have, 

```js
get state() {
    return _.cloneDeep(stateWeakMap.get(this));
}
```

So now one can't miss to change the state without performing full update logic.

framework/src/modules/http_api/http_api.js
```diff
@@ -63,7 +67,13 @@ module.exports = class HttpApi {
 			},
 			channel: this.channel,
 			config: this.options.config,
+			applicationState,
 		};
+
+		this.channel.subscribe('lisk:state:updated', event => {
+			Object.assign(this.scope.applicationState, event.data);
```
shuse2: prefer not to use `Object.assign`

framework/src/modules/chain/chain.js
```diff
@@ -158,6 +163,10 @@ module.exports = class Chain {
 			// Ready to bind modules
 			scope.logic.peers.bindModules(scope.modules);
 
+			this.channel.subscribe('lisk:state:updated', event => {
+				Object.assign(scope.applicationState, event.data);
```
shuse2: prefer not to use `Object.assign`

shuse2: Also, for applicationState, it should update the whole object

framework/src/modules/http_api/http_api.js
diego-G: We decide to keep `Object.assign` for this specific use case.

framework/src/modules/chain/chain.js
diego-G: We decide to keep `Object.assign` for this specific use case.

framework/test/mocha/unit/controller/application_state.js
```diff
@@ -0,0 +1,154 @@
+/*
+ * Copyright © 2018 Lisk Foundation
+ *
```
shuse2: shouldn't this be written in Jest?

framework/src/controller/application_state.js
```diff
@@ -0,0 +1,117 @@
...
+	set channel(channel) {
+		this.stateChannel = channel;
+	}
+
+	/**
+	 * Updates broadhash and height values.
+	 *
+	 * @returns {Promise.<boolean, Error>}
+	 */
+	async update(blocks) {
```
shuse2: `update` should take in state and update the state

framework/src/controller/application_state.js
```diff
@@ -0,0 +1,117 @@
...
+		if (!blocks) {
+			throw new TypeError('Argument blocks should be an array.');
+		}
+		try {
+			const newState = this.state;
```
shuse2: this is not doing anything

framework/src/controller/controller.js
```diff
@@ -106,6 +108,18 @@ class Controller {
...
 	}
 
+	/**
+	 * Initiate application state
+	 *
+	 * @async
+	 */
+	async _initState() {
+		this.applicationState = new ApplicationState({
+			initialState: this.config.initialState,
```
shuse2: do we need to change the config?
I think it's better to keep it as it is and compute the initial state from existing config?
Also `initialState` doesn't mean much for the user, so original `system` was better for the name?

framework/test/mocha/unit/controller/application_state.js
diego-G: I can try to migrate it.

framework/src/controller/application_state.js
diego-G: There is already an issue for this https://github.com/LiskHQ/lisk/issues/3141

framework/src/controller/controller.js
diego-G: I don't get the idea. We pass the part of the configuration used to initiate the state. That's why the name is as it is.

framework/test/mocha/unit/controller/application_state.js
diego-G: I guess I can migrate it alongside I move the logic out of update function in #3141, What do you think?

framework/test/mocha/unit/modules/chain/submodules/blocks/verify.js
```diff
@@ -74,12 +74,24 @@ describe('blocks/verify', () => {
...
 		};
 
+		channelMock = {
+			invoke: sinonSandbox
+				.stub()
+				.withArgs('lisk:updateApplicationState')
+				.returns(true),
+			once: sinonSandbox
+				.stub()
+				.withArgs('lisk:state:udpated')
```
shuse2:
```suggestion
				.withArgs('lisk:state:updated')
```

framework/test/mocha/unit/modules/chain/submodules/blocks/chain.js
```diff
@@ -115,14 +116,26 @@ describe('blocks/chain', () => {
...
 		};
 
+		channelMock = {
+			invoke: sinonSandbox
+				.stub()
+				.withArgs('lisk:updateApplicationState')
+				.returns(true),
+			once: sinonSandbox
+				.stub()
+				.withArgs('lisk:state:udpated')
```
shuse2:
```suggestion
				.withArgs('lisk:state:updated')
```

framework/src/controller/application_state.js
diego-G: It is calling `get state()` which creates a new copy.

framework/src/controller/controller.js
shuse2: i think the main point is `initialState` is not really clear to user.
Exposed variable should be descriptive what exactly is

shuse2:
@MaciejBaj I think we discussed that authentication of the message will be tackled in different issue, because the scope is a lot bigger and it's not straight forward

PR #3077, opened 2019-03-14
Wait for all sockets to be created on Bus and ChildProcessChannel - Closes #3073 

### LINKED ISSUE #3073: Wait for all sockets to be bound in Bus, ChildProcessChannel
### Expected behavior
Bus and ChildProcessChannel should be completely set up only when all its sockets are bound to a unix address.
### Actual behavior
Bus and ChildProcessChannel are not properly waiting for all sockets to be bound before completing the setup process.
`3013-update-bus-support-rpc`

limiaspasdaniel:
### What was the problem?
Bus and ChildProcessChannel were not waiting for all sockets to be bound or connected.
### How did I fix it?
Using a combination of Promise.race, Promise.all and setTimeout to wait for certain events that indicate a successful connection, bind or failure so the application doesn't continue until the
sockets are correctly initialized.
### How to test it?
Run all tests
### Review checklist

* The PR resolves #3073 
* Linting passes
* Tests pass


framework/src/controller/channels/child_process_channel.js
```diff
@@ -50,18 +50,20 @@ class ChildProcessChannel extends BaseChannel {
...
+
+		return Promise.race([
+			this._waitForAllToFinnish(),
+			this._rejectIfAtLeastOneFails(),
+			new Promise((resolve, reject) => {
+				timeout = setTimeout(async () => {
+					await this.cleanup();
+					// TODO: Review if logger.error might be useful.
+					reject(new Error('ChildProcessChannel setup timeout'));
+				}, 2000); // TODO: Get from constants
```
limiaspasdaniel: @lsilvs let's have a chat about what to do here :)

framework/src/modules/chain/chain.js
```diff
@@ -275,15 +275,19 @@ module.exports = class Chain {
...
+			});
 		}
 
 		// Run cleanup operation on each module before shutting down the node;
 		// this includes operations like snapshotting database tables.
 		await Promise.all(
-			modules.map(module => {
-				if (typeof module.cleanup === 'function') {
-					return promisify(module.cleanup)();
+			Object.keys(modules).map(key => {
```
limiaspasdaniel: The reason for this change is because modules is an `Object`, not an `Array`

framework/src/controller/bus.js
```diff
@@ -62,11 +63,21 @@ class Bus extends EventEmitter2 {
...
 		});
 
-		return new Promise((resolve, reject) => {
-			// TODO: wait for all sockets to be created
-			this.rpcSocket.once('bind', resolve);
-			this.rpcSocket.once('error', reject);
-			this.rpcSocket.bind(socketsPath.rpc);
+		let timeout;
+
+		return Promise.race([
```
lsilvs: I would try to not have promise declaration inside the race here to make the code more readable.

Suggestion:
```
const milliseconds = 2000;
return Promise.race([
	this._resolveWhenAllSucceed(),
	this._rejectWhenFirstFaill(),
	this._rejectWhenTimeout(milliseconds),
]);
```

We don't need to call `this.cleanup()` inside the timeout because the whole method will be reject and handled in the application level meaning we don't need the `finally` as well.
Note that I slightly changed the function names to better reflect its behaviour.

framework/src/controller/bus.js
```diff
@@ -216,6 +227,47 @@ class Bus extends EventEmitter2 {
...
+				this.rpcSocket.once('bind', () => {
+					resolve();
+				});
+			}),
+		]);
+	}
+
+	async _rejectIfAtLeastOneFails() {
+		return Promise.race([
+			new Promise((resolve, reject) => {
```
lsilvs:
```suggestion
			new Promise((_, reject) => {
```

framework/src/controller/bus.js
```diff
@@ -216,6 +227,47 @@ class Bus extends EventEmitter2 {
...
+				this.subSocket.sock.on('error', () => {
+					reject();
+				});
+			}),
+			new Promise((resolve, reject) => {
```
lsilvs:
```suggestion
			new Promise((_, reject) => {
```

framework/src/controller/bus.js
```diff
@@ -216,6 +227,47 @@ class Bus extends EventEmitter2 {
...
+				this.pubSocket.sock.on('error', () => {
+					reject();
+				});
+			}),
+			new Promise((resolve, reject) => {
```
lsilvs:
```suggestion
			new Promise((_, reject) => {
```

framework/src/controller/channels/child_process_channel.js
```diff
@@ -50,18 +50,20 @@ class ChildProcessChannel extends BaseChannel {
...
-				}
-			);
+		let timeout;
+
+		return Promise.race([
+			this._waitForAllToFinnish(),
+			this._rejectIfAtLeastOneFails(),
+			new Promise((resolve, reject) => {
+				timeout = setTimeout(async () => {
+					await this.cleanup();
```
lsilvs: As mentioned above, this is not required.

framework/src/controller/channels/child_process_channel.js
```diff
@@ -50,18 +50,20 @@ class ChildProcessChannel extends BaseChannel {
...
+					// TODO: Review if logger.error might be useful.
+					reject(new Error('ChildProcessChannel setup timeout'));
+				}, 2000); // TODO: Get from constants
+			}),
+		]).finally(() => {
+			clearTimeout(timeout);
```
lsilvs: If we remove the `this.cleanup()` from the timeout, the `finally` can be removed as well.

framework/src/controller/channels/child_process_channel.js
```diff
@@ -132,6 +134,61 @@ class ChildProcessChannel extends BaseChannel {
 			this.rpcSocket.close();
 		}
 	}
+
+	async _waitForAllToFinnish() {
```
lsilvs:
```suggestion
	async _resolveWhenAllSucceed() {
```

framework/src/controller/channels/child_process_channel.js
```diff
@@ -132,6 +134,61 @@ class ChildProcessChannel extends BaseChannel {
...
+							if (err) reject(err);
+							resolve(result);
+						}
+					);
+				});
+			}),
+		]);
+	}
+
+	async _rejectIfAtLeastOneFails() {
```
lsilvs:
```suggestion
	async _rejectWhenFirstFaill() {
```

src/index.js
```diff
@@ -42,6 +42,7 @@ try {
 		.run()
 		.then(() => app.logger.log('App started...'))
 		.catch(err => {
+			// TODO: Error objects are not being logged correctly.
```
lsilvs: Can you open an issue for this, pls?

framework/src/controller/channels/child_process_channel.js
lsilvs: Ideally it should be passed in `config`. I also want to create a config like `enableSocketChannel`. We can include this timeout together with the other one.

framework/src/controller/bus.js
```diff
@@ -216,6 +222,81 @@ class Bus extends EventEmitter2 {
...
+	 * Reject if time out
+	 *
+	 * @param {number} timeInMillis
+	 * @returns {Promise}
+	 * @private
+	 */
+	// eslint-disable-next-line class-methods-use-this
+	async _rejectWhenTimeout(timeInMillis) {
+		return new Promise((_, reject) => {
+			setTimeout(async () => {
```
lsilvs:
```suggestion
			setTimeout(() => {
```

framework/src/controller/bus.js
```diff
@@ -216,6 +222,81 @@ class Bus extends EventEmitter2 {
...
+				// TODO: Review if logger.error might be useful.
```
lsilvs: Only rejecting with message is enough. This comment can be removed imo.

framework/src/controller/channels/child_process_channel.js
```diff
@@ -83,7 +76,6 @@ class ChildProcessChannel extends BaseChannel {
 		if (event.module === this.moduleAlias) {
 			this.localBus.once(eventName, cb);
 		} else {
-			// TODO: make it `once` instead of `on`
```
lsilvs: This comment shouldn't be removed

framework/src/modules/chain/submodules/delegates.js
```diff
@@ -657,7 +657,7 @@ Delegates.prototype.updateForgingStatus = function(
 				password
 			);
 		} catch (e) {
-			return setImmediate(cb, 'Invalid password and public key combination');
+			throw 'Invalid password and public key combination';
```
lsilvs:
```suggestion
			throw new Error('Invalid password and public key combination');
```

framework/src/modules/chain/submodules/delegates.js
```diff
@@ -667,39 +667,32 @@ Delegates.prototype.updateForgingStatus = function(
 				.digest()
 		);
 	} else {
-		return setImmediate(cb, `Delegate with publicKey: ${publicKey} not found`);
+		throw `Delegate with publicKey: ${publicKey} not found`;
```
lsilvs: Same as above

framework/src/modules/chain/submodules/delegates.js
```diff
@@ -667,39 +667,32 @@ Delegates.prototype.updateForgingStatus = function(
...
 	}
 
 	if (keypair.publicKey.toString('hex') !== publicKey) {
-		return setImmediate(cb, 'Invalid password and public key combination');
+		throw 'Invalid password and public key combination';
```
lsilvs: Same as above

framework/src/modules/chain/submodules/delegates.js
```diff
@@ -667,39 +667,32 @@ Delegates.prototype.updateForgingStatus = function(
...
+			library.logger.info(`Forging disabled on account: ${account.address}`);
 		}
-	);
+
+		return {
+			publicKey,
+			forging,
+		};
+	}
+	throw 'Delegate not found';
```
lsilvs: Same as above

framework/src/controller/channels/child_process_channel.js
```diff
@@ -50,19 +50,12 @@ class ChildProcessChannel extends BaseChannel {
...
-				{ type: 'ipcSocket', rpcSocketPath: this.rpcSocketPath },
-				(err, result) => {
-					if (err) return reject(err);
-					return resolve(result);
-				}
-			);
-		});
+		return Promise.race([
+			this._resolveWhenAllSucceed(),
+			this._rejectWhenFirstFails(),
```
jondubois: Same as comment above about naming.

framework/src/controller/bus.js
```diff
@@ -62,12 +63,12 @@ class Bus extends EventEmitter2 {
...
-			this.rpcSocket.once('bind', resolve);
-			this.rpcSocket.once('error', reject);
-			this.rpcSocket.bind(socketsPath.rpc);
-		});
+		return Promise.race([
+			this._resolveWhenAllSucceed(),
+			this._rejectWhenFirstFails(),
+			// Timeout is needed here in case "bind" events never arrive
+			this._rejectWhenTimeout(2000), // TODO: Get value from config constant
+		]);
```
jondubois: Where are the event handlers from `_resolveWhenAllSucceed`, `_rejectWhenFirstFails`, and `_rejectWhenTimeout` being cleaned up?

Maybe instead of returning the `Promise.race(...)`, you can `await Promise.race(...)` then on the next line, you can unbind all events. You could add a single private method to unbind all events.

Note that this still applies even if you use `once` instead of `on` because then the handler will stay bound to the socket and could still be called unexpectedly at some point in the future.

limiaspasdaniel: I think only event handlers from `_resolveWhenAllSucceed` should be removed. The other handlers from `_rejectWhenFirstFails()` will be indirectly removed as if any of those events are fired the application stops it's execution (intentionally). @lsilvs what do you think?

framework/src/controller/channels/child_process_channel.js
limiaspasdaniel: Done

framework/src/controller/bus.js
```diff
@@ -62,12 +63,12 @@ class Bus extends EventEmitter2 {
...
 		});
 
-		return new Promise((resolve, reject) => {
-			// TODO: wait for all sockets to be created
-			this.rpcSocket.once('bind', resolve);
-			this.rpcSocket.once('error', reject);
-			this.rpcSocket.bind(socketsPath.rpc);
-		});
+		return Promise.race([
+			this._resolveWhenAllSucceed(),
```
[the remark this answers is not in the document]
limiaspasdaniel: I like it better, good one. I've done it.

framework/src/controller/bus.js
limiaspasdaniel: After discussion with @jondubois I now remove all the listener inside a `finally` clause.

PR #3079, opened 2019-03-14
Remove js extension filename from requires - Closes #3071

### LINKED ISSUE #3071: Remove js extension filename from requires
### Expected behavior

Requires doesn't contain file extension `.js`.

### Actual behavior

https://github.com/LiskHQ/lisk/blob/381c91af4726174ae35d49f7f8ad79b5342d8b7e/framework/src/modules/chain/submodules/dapps.js#L18

`+1.6`

diego-G:
### Review checklist

* The PR resolves #3071
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/chain/scripts/console.js
+++ b/framework/src/modules/chain/scripts/console.js
@@ -19,7 +19,7 @@ const repl = require('repl');
 const fs = require('fs');
 const Promise = require('bluebird');
 require('../../../../test/mocha/setup');
-const application = require('../../../../test/mocha/common/application.js');
+const application = require('../../../../test/mocha/common/application');
 
 // Created this before in global scope as its dependency of test/node.js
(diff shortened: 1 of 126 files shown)


framework/src/modules/chain/scripts/console.js
```diff
@@ -61,7 +61,7 @@ application.init(
 		const helpersFolder = './helpers/';
 		fs.readdirSync(helpersFolder).forEach(file => {
 			const filePath = path.resolve(helpersFolder, file);
-			const fileName = path.basename(filePath, '.js');
+			const fileName = path.basename(filePath, '');
```
yatki: I think we shouldn't remove .js from here fileName was being used as a key in the helpers object. Can cause a side effect.

But if you don't see a problem, we can remove fileName variable and useFilePath as a key.

PR #3083, opened 2019-03-14
Fix protocol version parsing with double digits - Closes #3081

### LINKED ISSUE #3081: Fix protocol version parsing with double digits
### Expected behavior
Protocol version should be parsing correctly the hard part.
When 1.0 -> 1.
When 10.0 -> 10
### Actual behavior
The hard part of protocolVersion ([1].0) of incoming peers is not being correctly parsed, only taking into account the first digit. 
This fails if it is > 9, right know having:
When 10.0 -> 1.

limiaspasdaniel:
### What was the problem?
The protocol version was being incorrectly parsed.
Using values with more than one digit on the hard
part caused the validation to only take in account
the first one.
### How did I fix it?
Now the hard part and the soft part of the protocol
version are correctly parsed by being delimited by
a perdiod `.`
### How to test it?
Run network tests and unit tests.
### Review checklist
* The PR resolves #3081
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/components/system/system.js
+++ b/framework/src/components/system/system.js
@@ -93,9 +93,11 @@ class System {
 		if (!protocolVersion) {
 			return false;
 		}
-		const peerHard = parseInt(protocolVersion.split('.')[0]);
-		const myHard = parseInt(this.headers.protocolVersion.split('.')[0]);
-		return myHard === peerHard && peerHard >= 1;
+		const peerHardForks = parseInt(protocolVersion.split('.')[0]);
+		const systemHardForks = parseInt(
+			this.headers.protocolVersion.split('.')[0]
+		);
+		return systemHardForks === peerHardForks && peerHardForks >= 1;
 	}
 
 	/**
--- a/framework/src/components/system/system.js
+++ b/framework/src/components/system/system.js
@@ -93,8 +93,8 @@ class System {
 		if (!protocolVersion) {
 			return false;
 		}
-		const peerHard = parseInt(protocolVersion[0]);
-		const myHard = parseInt(this.headers.protocolVersion[0]);
+		const peerHard = parseInt(protocolVersion.split('.')[0]);
+		const myHard = parseInt(this.headers.protocolVersion.split('.')[0]);
 		return myHard === peerHard && peerHard >= 1;
 	}
 
--- a/framework/test/mocha/unit/components/system/system.js
+++ b/framework/test/mocha/unit/components/system/system.js
@@ -160,37 +160,83 @@ describe('components: system', () => {
 	});
 
 	describe('protocolVersionCompatible', () => {
-		describe('when protocol version is exactly equal to system protocol version', () => {
-			it('should return true', async () =>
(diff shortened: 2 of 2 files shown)


framework/test/mocha/unit/components/system/system.js
```diff
@@ -160,37 +160,83 @@ describe('components: system', () => {
...
-		describe('when the hard part of protocol is not exactly equal than the one of the system protocol version', () => {
-			it("should return false if it's greater or lesser", async () =>
-				expect(systemComponent.protocolVersionCompatible('2.0')).to.be.false);
-			it("should return false if it's lesser", async () =>
-				expect(systemComponent.protocolVersionCompatible('0.0')).to.be.false);
-		});
-		describe('when the hard part of protocol is equal to  the one of the system protocol version', () => {
-			it('should return true', async () =>
-				expect(systemComponent.protocolVersionCompatible('1.5')).to.be.true);
+		describe('when the hard for part has one digit', () => {
```
ishantiw: minor typo `when the hard fork...`*

framework/src/components/system/system.js
```diff
@@ -93,8 +93,8 @@ class System {
 		if (!protocolVersion) {
 			return false;
 		}
-		const peerHard = parseInt(protocolVersion[0]);
-		const myHard = parseInt(this.headers.protocolVersion[0]);
+		const peerHard = parseInt(protocolVersion.split('.')[0]);
```
ishantiw: I think the name `peerHard` is not so clear, maybe we can have some name like `peerHardForks` or `numPeerHForks`. Whatever you feel could be better than `peerHard` is fine.

framework/src/components/system/system.js
```diff
@@ -93,8 +93,8 @@ class System {
...
+		const myHard = parseInt(this.headers.protocolVersion.split('.')[0]);
```
ishantiw: Same here, if we can use a better name than `myHard`

framework/test/mocha/unit/components/system/system.js
```diff
@@ -160,37 +160,83 @@ describe('components: system', () => {
...
+			beforeEach(async () => {
+				systemComponent.headers.protocolVersion = '1.0';
+			});
+
+			describe('when protocol version is exactly equal to system protocol version', () => {
+				it('should return true', async () =>
+					expect(systemComponent.protocolVersionCompatible('1.0')).to.be.true);
+			});
+			describe('when the hard part of protocol is not exactly equal than the one of the system protocol version', () => {
```
ishantiw: Could be simplified to => `when the hard part of the peer protocol is not exactly equal to the system protocol version`

framework/src/components/system/system.js
ishantiw: Also, if `protocolVersion` is `undefined` then this line will throw an error, or we already make sure that this is never `undefined`?

limiaspasdaniel: if the protocol version is undefined it's being checked for in line 93

PR #3103, opened 2019-03-18
Load http_api module as a child process - Closes #3059

### LINKED ISSUE #3059: Load http_api module as a child process
### Expected behavior
As highlighted in the [LIP](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0005.md#modules), implement the possibility of loading modules as child processes. 

Update module loading logic to load modules child process and establish RPC connection to the controller and load http_api module in a child process.
### Actual behavior
A module can be loaded and run on the same process as the controller.

`1.6.0`

lsilvs:
### What was the problem?
Not possible to run modules in child process

### How did I fix it?
I've used nodejs `child_process` to fork new processes when the module is set to be loaded as child process.
The new `ChildProcessChannel` is used and communicate with the parent process through sockets.

### How to test it?
In `devnet` httpApi is loaded as child process by default. To change the configuration for other networks, set `ipc.enabled: true` in the config file.

### Review checklist

* The PR resolves #3059
* Linting passes
* Tests pass


framework/src/controller/controller.js
```diff
@@ -203,23 +199,38 @@ class Controller {
...
-		channel.publish(`${moduleAlias}:loading:finished`);
+		const parameters = [
+			modulePath,
+			JSON.stringify({ ...options, socketsPath: this.socketsPath }),
+		];
+
+		const child = child_process.fork(program, parameters);
+
+		child.on('exit', () => {
+			this.logger.error(`Module ${moduleAlias}(${name}:${version}) exited`);
```
pablitovicente: It would be nice here to have the child have an error handler that sends the error that cause the exit back to its parent so we have more information on the exit reason.

yatki:
@lsilvs we need to have an environment variable to configure how we load modules. So we can execute functional tests with child process modules.

framework/src/controller/bus.js
```diff
@@ -113,12 +118,12 @@ class Bus extends EventEmitter2 {
...
 		let channel;
-		if (options.type === 'inMemory') {
-			channel = options.channel;
-		} else {
+		if (options.rpcSocketPath) {
 			const rpcSocket = axon.socket('req');
 			rpcSocket.connect(options.rpcSocketPath);
 			channel = new RPCClient(rpcSocket);
+		} else {
+			channel = options.channel;
```
yatki: I'd prefer to move this line to line 120 to remove else block.

framework/src/controller/bus.js
```diff
@@ -178,17 +183,46 @@ class Bus extends EventEmitter2 {
...
 
-		super.emit(eventName, eventValue); // Communicate throw event emitter
-		this.pubSocket.emit(eventName, eventValue); // Communicate throw unix socket
+		// Communicate through event emitter
+		super.emit(eventName, eventValue);
+
+		// Communicate through unix socket
+		if (this.config.ipc.enabled) {
+			this.pubSocket.emit(eventName, eventValue);
+		}
```
yatki: if ipc enabled, we are publishing 2 events here. What was the reason for that?

framework/src/controller/channels/child_process_channel.js
```diff
@@ -29,33 +29,38 @@ class ChildProcessChannel extends BaseChannel {
...
+			this.rpcServer = new RPCServer(this.rpcSocket);
 
-		this.rpcSocket.bind(this.rpcSocketPath);
+			this.rpcServer.expose('invoke', (action, cb) => {
+				this.invoke(action)
+					.then(data => cb(null, data))
+					.catch(error => cb(error));
+			});
+
+			this.rpcSocket.bind(this.rpcSocketPath);
```
yatki: maybe we can move this line after 49 to keep the structure consistent?

framework/src/controller/application.js
```diff
@@ -119,14 +119,20 @@ class Application {
 		__private.modules.set(this, {});
 		__private.transactions.set(this, {});
 
+		// TODO: move this configuration to module especific config file
+		const childProcessModules = process.env.CHILD_PROCESS_MODULES
```
shuse2: add `LISK_`?

framework/src/controller/bus.js
```diff
@@ -38,40 +40,43 @@ class Bus extends EventEmitter2 {
...
-		this.rpcServer = new RPCServer(this.rpcSocket);
-		this.rpcSocket.bind(socketsPath.rpc);
-
-		this.rpcServer.expose(
-			'registerChannel',
-			(moduleAlias, events, actions, options, cb) => {
-				this.registerChannel(moduleAlias, events, actions, options)
-					.then(() => cb(null))
+	async setup() {
+		if (this.config.ipc.enabled) {
```
shuse2: Prefer early return
```
if (!this.config.ipc.enabled) {
  return false;
}
```

framework/src/controller/bus.js
```diff
@@ -178,17 +183,46 @@ class Bus extends EventEmitter2 {
...
 	subscribe(eventName, cb) {
 		if (!this.getEvents().includes(eventName)) {
-			throw new Error(`Event ${eventName} is not registered to bus.`);
+			this.logger.info(
+				`Event ${eventName} was subscribed but not registered to the bus yet.`
+			);
+		}
+
+		// Communicate through event emitter
+		super.on(eventName, cb);
```
shuse2: `this.on`?

framework/src/controller/bus.js
```diff
@@ -178,17 +183,46 @@ class Bus extends EventEmitter2 {
...
-		this.subSocket.on(eventName, cb); // Communicate throw unix socket
+	once(eventName, cb) {
+		if (!this.getEvents().includes(eventName)) {
+			this.logger.info(
+				`Event ${eventName} was subscribed but not registered to the bus yet.`
+			);
+		}
+
+		// Communicate through event emitter
+		super.once(eventName, cb);
```
shuse2: `this.once`?

framework/src/controller/bus.js
```diff
@@ -178,17 +183,46 @@ class Bus extends EventEmitter2 {
 			throw new Error(`Event ${eventName} is not registered to bus.`);
 		}
 
-		super.emit(eventName, eventValue); // Communicate throw event emitter
-		this.pubSocket.emit(eventName, eventValue); // Communicate throw unix socket
+		// Communicate through event emitter
+		super.emit(eventName, eventValue);
```
shuse2: `this.emit`?

framework/src/controller/channels/child_process_channel.js
```diff
@@ -142,44 +150,59 @@ class ChildProcessChannel extends BaseChannel {
...
-						}
-					);
-				});
-			}),
-		]);
+		const promises = [];
+
+		if (this.pubSocket) {
+			promises.push(
+				new Promise(resolve => {
```
yatki: for the sake of readability would be nice to have a function like below to promisify once events: 
```js
const promisifyOnce = (emitter, event, reject = false) => new Promise((resolve, reject) => {
  emitter.once(event, (...data) => {
    if (reject) {
      return reject(...data);
    }
    return resolve(...data);
  });
});

// Usage:

const subSocketConnectPromise = promisifyOnce(this.subSocket.sock, 'connect');
const subSocketConnectErrPromise = promisifyOnce(this.subSocket.sock, 'error', true);
```

framework/src/controller/channels/child_process_channel.js
```diff
@@ -142,44 +150,59 @@ class ChildProcessChannel extends BaseChannel {
...
+				new Promise((resolve, reject) => {
+					this.busRpcSocket.once('connect', () => {
+						this.busRpcClient.call(
+							'registerChannel',
+							this.moduleAlias,
+							this.eventsList.map(event => event.name),
+							this.actionsList.map(action => action.name),
+							{ type: 'ipcSocket', rpcSocketPath: this.rpcSocketPath },
+							(err, result) => {
+								if (err) reject(err);
```
shuse2:
```
if (err) {
  reject(err);
}
```

framework/src/controller/child_process_loader.js
```diff
@@ -0,0 +1,29 @@
+// Parameters passed by `child_process.fork(_, parameters)`
+const modulePath = process.argv[2];
+const { moduleOptions, config } = JSON.parse(process.argv[3]);
+
+// eslint-disable-next-line import/no-dynamic-require
+const Klass = require(modulePath);
+const { ChildProcessChannel } = require('./channels');
```
shuse2: maybe static import first?

framework/src/controller/controller.js
```diff
@@ -152,9 +150,18 @@ class Controller {
...
+				if (this.config.ipc.enabled) {
+					// eslint-disable-next-line no-await-in-loop
+					await this._loadChildProcessModule(alias, klass, options);
+				} else {
+					this.logger.warn(
+						`IPC is disabled. ${alias} will be loaded in-memory.`
+					);
+					// eslint-disable-next-line no-await-in-loop
+					await this._loadInMemoryModule(alias, klass, options);
+				}
```
yatki: maybe if else statements can be simplified with using `continue;` ?

framework/src/controller/controller.js
```diff
@@ -203,23 +213,42 @@ class Controller {
...
-			`Module ready with alias: ${moduleAlias}(${name}:${version})`
-		);
+		const child = child_process.fork(program, parameters);
+
+		child.on('exit', (code, signal) => {
+			this.logger.error(
+				`Module ${moduleAlias}(${name}:${version}) exited with code: ${code} and signal: ${signal}`
+			);
+			// TODO: Reload child instead of exiting the parent process
+			process.exit(1);
```
yatki: Is this going to be fixed in another PR?

framework/test/jest/specs/unit/controller/channels/in_memory_channel.spec.js
```diff
@@ -75,21 +73,21 @@ describe('InMemoryChannel Channel', () => {
 	describe('#constructor', () => {
 		it('should create the instance with given arguments.', () => {
 			// Assert
-			expect(inMemoryChannel.bus).toBe(params.bus);
+			expect(inMemoryChannel.bus).toBe(undefined);
```
yatki: this can be replaced with `toHaveProperty` matcher.

framework/src/controller/bus.js
yatki: I kind of prefer to use `super` here because it's instantly clear that we are using a parent class method. But I don't have strong opinion on this.

framework/src/controller/bus.js
lsilvs: First emit is for in-memory modules and the second one for child process

framework/src/controller/controller.js
lsilvs: We have a eslint rule `no-continue`. That's why I didn't use it.
We shouldn't be using `for of` with `await` anyway.
Next iteration we can improve this.

framework/src/controller/controller.js
lsilvs: https://github.com/LiskHQ/lisk/issues/3142

framework/test/jest/specs/unit/controller/channels/in_memory_channel.spec.js
lsilvs: Can you give an example, pls? I've changed to `undefined` because we not setting the bus in the constructor anymore.

framework/src/controller/bus.js
lsilvs: Here we can't use `this` because `this.once` is its own method and we get `Maximum call stack size exceeded`

framework/src/controller/channels/child_process_channel.js
lsilvs: This code was part of another PR and I prefer keeping it as it is in this PR.

framework/src/controller/bus.js
shuse2: ah my bad, this will be recursive... =p

framework/src/controller/controller.js
shuse2: > We shouldn't be using for of with await anyway.
Not related to this but what's the reason for this? i think this is the only way though

lsilvs: afaik `for await...of` was created for this

framework/test/jest/specs/unit/controller/channels/in_memory_channel.spec.js
yatki: Ah in that case please move this test from constructor to registerToBus suite.

Also just fyi I meant `expect(inMemoryChannel).toHaveProperty('bus')` could be used if the property is declared but was not assigned yet.

lsilvs:
@diego-G Can you investigate your local environment better, pls?
It works perfectly on Jenkins and all tests are green.

PR #3105, opened 2019-03-19
Validate filter getter methods - Closes #3005

### LINKED ISSUE #3005: Validate filters for getter methods in Account and Transaction entities
### Expected behavior
I expect the storage entities Account and Transaction to validate the passed filters and throw and error if a filter is not valid.
Also, unit tests should be added to cover this scenario.

### Actual behavior
Currently it doesn't and the database query crashes with an unclear message

### Steps to reproduce
Run `storage.entities.Transaction.get({ invalid_filter: true })`

1.5

michielmulders:
### What was the problem?
For transactions and accounts, no validation for allowed filters and options present.

### How did I fix it?
Add validation and tests.

### How to test it?
Run test suite.

### Review checklist

* The PR resolves #3005 
* Linting passes
* Tests pass


framework/src/components/storage/entities/account.js
```diff
@@ -396,6 +397,9 @@ class Account extends BaseEntity {
 	}
 
 	_getResults(filters, options, tx, expectedResultCount = undefined) {
+		this.validateFilters(filters);
```
diego-G: Are there more cases like this in this entity or any other?

framework/src/components/storage/entities/account.js
```diff
@@ -276,6 +276,7 @@ class Account extends BaseEntity {
 
 		const defaultSort = { sort: 'balance:asc' };
 		this.extendDefaultOptions(defaultSort);
+		this.sortingFields.push('productivity');
```
diego-G: Please, add a test for this new sorting field.

framework/src/modules/http_api/controllers/accounts.js
```diff
@@ -84,6 +84,12 @@ function accountFormatter(totalSupply, account) {
  * @todo Add description for the function and the params
  */
 AccountsController.getAccounts = async function(context, next) {
+	const invalidParams = swaggerHelper.invalidParams(context.request);
```
diego-G: Could you increase the functional test coverage of every controller where you are addressing this?

framework/src/components/storage/entities/account.js
michielmulders: Good comment! No, these were the only ones where it was missing for.

framework/src/components/storage/entities/account.js
michielmulders: Tests are here: https://github.com/LiskHQ/lisk/blob/development/framework/test/mocha/functional/http/get/delegates.js#L456

framework/src/components/storage/entities/transaction.js
```diff
@@ -144,6 +144,10 @@ class Transaction extends BaseEntity {
 			fieldName: 'trs.rowId',
 		});
 
+		this.addField('dapp_name', 'string', {
```
lsilvs: Is this necessary? If so, please combine with `addFilter` in https://github.com/LiskHQ/lisk/blob/790e9e9ba6ac378c224c0522a56fd1e1165862e4/framework/src/components/storage/entities/transaction.js#L237

framework/test/mocha/functional/http/get/blocks.js
```diff
@@ -87,7 +88,7 @@ describe('GET /blocks', () => {
 			});
 
 			it('using invalid height = -1 should fail with error', async () => {
-				return blocksEndpoint.makeRequest({ height: 0 }, 400).then(res => {
+				return blocksEndpoint.makeRequest({ height: -1 }, 400).then(res => {
```
pablitovicente: Why this change in the height?

michielmulders: As we are testing -1 and the value was 0.
Two same tests above each other, probably typo?

framework/test/mocha/functional/http/get/voters.js
```diff
@@ -46,57 +46,45 @@ describe('GET /api/voters', () => {
 		expect(res.body.data.voters).to.be.empty;
 	}
 
+	function expectInvalidParamsResponse(res) {
```
pablitovicente: Although I understand the reasoning behind this change as we're going to move to Jest eventually all tests need to be atomic so I would revert this change so each tests is independent. Even though we could potentially be sure that the assertions will be the same; in the future we might need different ones so I would keep them on each tests instead of using a helper.

framework/test/mocha/functional/http/get/votes.js
```diff
@@ -52,57 +52,45 @@ describe('GET /api/votes', () => {
 		);
 	}
 
+	function expectInvalidParamsResponse(res) {
```
pablitovicente: Same as previous comment about helpers in tests

framework/test/mocha/functional/http/get/blocks.js
michielmulders: @pablitovicente Reference: https://github.com/LiskHQ/lisk/blob/development/framework/test/mocha/functional/http/get/blocks.js#L83

framework/test/mocha/functional/http/get/voters.js
michielmulders: Oh dear 👍

framework/test/mocha/functional/http/get/blocks.js
pablitovicente: Ohh then maybe add for zero and correct expectation (maybe split in new issue this part?)

michielmulders: Resolved!

framework/src/modules/http_api/controllers/accounts.js
michielmulders: @diego-G Added a lot of tests!

framework/test/mocha/functional/http/get/dapps.js
```diff
@@ -319,20 +386,20 @@ describe('GET /dapps', () => {
 		});
 
 		describe('unknown=', () => {
-			it('using empty string should return all results', async () => {
-				return dappsEndpoint.makeRequest({ unknown: '' }, 200).then(res => {
-					expect(res.body.data).to.have.length.at.least(registeredDappsAmount);
-				});
+			it('using empty string should return UNKNOWN_PARAM error', async () => {
```
pablitovicente: I think we may want to move this change to 1.7 instead of 1.6 I understand the fix and reasoning behind it but it's a behaviour change.  (Same for below)

@MaciejBaj @diego-G thoughts?

michielmulders: If I have to revert the behavior so it returns all results when you pass an invalid filter to the endpoint, then I'll have to remove the `validateFilters` and `validateOptions` functions for this `/dapps` endpoint. This means there is no validation for all other fields as well (old bug ->). Some sort option like `id:wrong` would throw `unexpected error`.

@lsilvs thoughts?

lsilvs: This behaviour was a bug and this test shouldn't cover this bug imo.
It changes the behaviour but fixing it (which is fine for me)

pablitovicente: Resolving as there's agreement that the change is acceptable.

PR #3116, opened 2019-03-20
Fix relative path -c option CLI

### LINKED ISSUE #3115: CLI option -c requires relative path
### Expected behavior
`-c` should accept a path starting from the root of Lisk Core.

### Actual behavior
`-c ../../../../config/default/config.json`
Starts from chain module for some reason.

### Steps to reproduce
Run `npm start -c path`

1.6.0

michielmulders:
### What was the problem?
The path that needed to be passed to `-c` option for starting Lisk Core from CLI required relative path like this `../../../../config/custom/config.json`.

### How did I fix it?
Fix the path to start from root.
You can use `-c` like this now `npm start -- -c config/cust/config.json`
### Review checklist

* The PR resolves #3115 
* Linting passes
* Tests pass


framework/src/modules/chain/helpers/config.js
```diff
@@ -104,7 +104,12 @@ function Config(packageJson, parseCommandLineOptions = true) {
 
 	let customConfig = {};
 	if (program.config || process.env.LISK_CONFIG_FILE) {
-		customConfig = loadJSONFile(program.config || process.env.LISK_CONFIG_FILE);
+		const customConfigPath = path.join(
+			__dirname,
+			'../../../../../',
```
fchavant: This looks more like a workaround more than a fix of the actual issue.

framework/src/modules/chain/helpers/config.js
```diff
@@ -104,7 +104,12 @@ function Config(packageJson, parseCommandLineOptions = true) {
...
+			program.config || process.env.LISK_CONFIG_FILE
```
fchavant: What happens if I pass an absolute path to `-c`?

michielmulders: Do we want that behavior? Until now, the path was relative from root of project. @fchavant

fchavant: OK that's fine then.

michielmulders: Decided to keep the relative path for now.

framework/src/modules/chain/helpers/config.js
michielmulders: Patch it for now.
Other paths in folder are constructed in the same way.

fchavant:
Noticed a bug while testing this PR: https://github.com/LiskHQ/lisk/issues/3123

PR #3126, opened 2019-03-20
Remove insecure chain actions - Closes #3033

### LINKED ISSUE #3033: Remove insecure and not used chain actions
### Expected behavior
Expect for chain module to expose only relevant actions

### Actual behavior
It exposes `getForgersKeyPairs` which should never be accessed from other modules.
Plus, in order to avoid braking changes in the future, we should only expose relevant actions and just add new ones when justified.

### Steps to reproduce
`apiChannel.invoke('chain:getForgersKeyPairs')`

1.6

michielmulders:
### What was the problem?
`getForgersKeyPairs` returns priv and pub key -> insecure chain action.

### How did I fix it?
Added logic to filter out priv key and return a same structured object with only pub keys.
Name change for action to `getForgersPublicKeys`.
+ Checked other actions, they are all in use.

### How to test it?
Request forging status of a node:
`http://localhost:4000/api/node/status/forging?publicKey=3ea481498521e9fb1201b2295d0e9afa826ac6a3ef51de2f00365f915ac7ac06`

### Review checklist

* The PR resolves #3033 
* Linting passes
* Tests pass


MaciejBaj:
@lsilvs great idea, please open the issue for that.

PR #3130, opened 2019-03-20
Process incoming signatures using statestore and elements - Closes #2825

### LINKED ISSUE #2825: Refactor verifying multi-signatures for transactions and use the new StateStore and transactions library
### Expected behavior
In the current implementation, multi signatures are verified by calling the `verifySignature` function.

### Actual behavior
In this issue, the verification of the multi signatures should be refactored and the application should use StateStore and verifySignature methods defined in the lisk-element transactions library.

1.6

mitsujutsu:
### What was the problem?

Incoming signatures were being processed using old code. 

### How did I fix it?

`processSignature` function in `processTransactions` module which uses StateStore and Transactions library from elements now handles incoming signature objects.

### How to test it?

Run unit test for transport 

### Review checklist

* The PR resolves #2825 
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/chain/submodules/multisignatures.js
+++ b/framework/src/modules/chain/submodules/multisignatures.js
@@ -14,6 +14,7 @@
 
 'use strict';
 
+const { Status: TransactionStatus } = require('@liskhq/lisk-transactions');
 const async = require('async');
 
 // Private fields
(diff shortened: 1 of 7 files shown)


framework/src/modules/chain/submodules/transport.js
```diff
@@ -154,24 +155,33 @@ __private.receiveSignatures = function(signatures = []) {
...
 		}
-
-		return modules.multisignatures.processSignature(
-			query,
-			processSignatureErr => {
-				if (processSignatureErr) {
+		return modules.processTransactions
+			.processSignature(signature)
+			.then(transactionResponse => {
+				if (transactionResponse.status === TransactionStatus.FAIL) {
```
shuse2: Just in case we should check `transactionResopnse.errors.length` too

framework/src/modules/chain/submodules/process_transactions.js
```diff
@@ -137,6 +140,56 @@ class ProcessTransactions {
...
+	}
 }
 
+// Events
+/**
+ * Assigns modules to private constant `modules`.
+ *
+ * @param {modules} scope - Loaded modules
+ */
+ProcessTransactions.prototype.onBind = function(scope) {
```
SargeKhan: `ProcessTransactions` has no dependencies on other modules and it doesn't change the state of the application or the transaction pool. I like it this way and I think we should stick to it.
I think we can follow the existing code flow structure in which the transaction is received in the `transport` module, which internally calls `multisignature` module where transaction signatures are processed. 
Therefore, we should define the logic for fetching the state for the transaction and adding the signature in processSignatrure function but the transaction should be fetched from the transaction pool in the `multisignature` module.

framework/src/modules/chain/submodules/process_transactions.js
```diff
@@ -137,6 +140,56 @@ class ProcessTransactions {
...
+			library.logger.error(message, { signature });
+			return {
+				id: signature.transactionId,
+				status: TransactionStatus.FAIL,
+				errors: [new Error(message)],
+			};
+		}
+
+		// Add multisignature to transaction and process
+		const transactionResponse = transaction.addMultisignature(
```
SargeKhan: Calling this function without transaction.prepare works?

framework/src/modules/chain/submodules/multisignatures.js
```diff
@@ -276,53 +277,49 @@ __private.processSignatureFromMultisignatureAccount = (
...
 
-		// Process signature for send from multisignature account transaction
-		return __private.processSignatureFromMultisignatureAccount(
-			signature,
-			transaction,
-			balanceSequenceCb
-		);
-	}, cb);
+			return setImmediate(cb);
+		});
```
shuse2: need to catch and call the callback

shuse2:
Is the base branch correct? Probably need to rebase to the latest feature branch?

mitsujutsu:
> Is the base branch correct? Probably need to rebase to the latest feature branch?

Yes before integration test was not merged in, but now that it is merged into feature I changed it back.

framework/src/modules/chain/submodules/multisignatures.js
```diff
@@ -50,6 +51,68 @@ function Multisignatures(cb, scope) {
...
+			) {
+				const message = `Error processing signature: ${
+					transactionResponse.errors[0].message
+				}`;
+				library.logger.error(message, { signature });
+				return setImmediate(cb, new Error(message));
+			}
+			// Emit events
+			library.channel.publish(
+				'chain:multisignatures:signature:change',
```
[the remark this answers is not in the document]
mitsujutsu: This event was being emitted from `validateSignature` after signature was validated and pushed into transaction. See commit: https://github.com/LiskHQ/lisk/commit/d51f0b07e4dca7dce9fc8596fdfae19bcf0adec2

Issue #2828, opened 2019-02-06
Remove usages of applying transaction to unconfirmed state from the application


SargeKhan:
### Expected behavior
Applying transactions to the unconfirmed state in the database is a step which should be removed and replaced with applying transactions in memory using state store.
### Actual behavior
In lisk-core, the application calls applyUnconfirmed step for transactions in multiple places which is not required.

pablitovicente:
This was re-applied in https://github.com/LiskHQ/lisk/pull/3144

PR #3144, opened 2019-03-21
Remove implementation of all transaction types and Remove unconfirmed state from database - Closes #2835, #2828

### LINKED ISSUE #2835: Remove implementation of all transaction types from logic folder
### Expected behavior
The application should use transactions defined in lisk-elements throughout the application.
### Actual behavior
Various transaction types are defined under the logic folder in the application.

SargeKhan:
### What was the problem?
This PR solves two issues which were tightly coupled. Firstly, the unconfirmed state was maintained in the database and secondly, the unused transaction definition in logic folder were not removed. 
### How did I fix it?
First issue it addresses by removing the unconfirmed state for accounts which were maintained in the database. It solves the second issue of removing all the transaction types defined in Core, and uses transactions defined in the lisk-transactions library.
### How to test it?
Run integration tests for blocks
### Review checklist

* The PR resolves #2835 & #2828
* Linting passes
* Tests pass


framework/src/modules/chain/submodules/accounts.js
```diff
@@ -32,15 +29,20 @@ __private.assetTypes = {};
...
  * Main accounts methods. Initializes library with scope content and generates a Vote instance.
- * Calls logic.transaction.attachAssetType().
  *
  * @class
  * @memberof modules
  * @see Parent: {@link modules}
  * @requires crypto
  * @requires helpers/bignum
  * @requires logic/block_reward
+<<<<<<< HEAD
```
pablitovicente: Merge conflict inside comment better removed

PR #3167, opened 2019-03-22
Event & Action name validation regex definitions are different in Base class - Closes #3008

### LINKED ISSUE #3008: Event & Action name validation regex definitions are different in Base class
### Expected behavior

Regex validation should be the same throughout the whole framework.

### Actual behavior

Event name cannot start with a number as its stated in Event class but `isValidEventName` method in Base Class has different regex definition.

https://github.com/LiskHQ/lisk/blob/cd67812d1548b3f383707d64274bc9f001517896/framework/src/controller/channels/base.js#L97

https://github.com/LiskHQ/lisk/blob/cd67812d1548b3f383707d64274bc9f001517896/framework/src/controller/event.js#L3

Also `isValidActionName` method in Base Class has different regex definition than Action class.

MaciejBaj:
### What was the problem?
Event & Action name validation regex definitions are different in Base class.
### How did I fix it?
Fixed the `module:event` regexp and reuse the same in base and event.

### Review checklist

* The PR resolves #3008
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/controller/channels/base/constants.js
+++ b/framework/src/controller/channels/base/constants.js
@@ -4,7 +4,7 @@ const INTERNAL_EVENTS = Object.freeze([
 	'loading:finished',
 ]);
 
-const eventWithModuleNameReg = /[A-Za-z]+[A-Za-z0-9]*:[A-Za-z]+[A-Za-z0-9]*/;
+const eventWithModuleNameReg = /^([^\d][\w]+)((?::[^\d][\w]+)+)$/;
 
 module.exports = {
 	eventWithModuleNameReg,
--- a/framework/src/controller/channels/base/constants.js
+++ b/framework/src/controller/channels/base/constants.js
@@ -4,6 +4,9 @@ const INTERNAL_EVENTS = Object.freeze([
 	'loading:finished',
 ]);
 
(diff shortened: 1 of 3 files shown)


MaciejBaj:
Fixing tests 😢

michielmulders:
@MaciejBaj seems to be something wrong with retrieving the constants from the node as the tests already timeout before running any of them.

framework/src/controller/channels/base/constants.js
```diff
@@ -4,6 +4,9 @@ const INTERNAL_EVENTS = Object.freeze([
 	'loading:finished',
 ]);
 
+const eventWithModuleNameReg = /[A-Za-z]+[A-Za-z0-9]*:[A-Za-z]+[A-Za-z0-9]*/;
```
shuse2: no nesting more than 2 is allowed?

michielmulders: @MaciejBaj other regex is used: `/^([a-zA-Z][a-zA-Z0-9]*)((?::[a-zA-Z][a-zA-Z0-9]*)+)$/`
If you use this one, app starts again.

MaciejBaj: True, fixed both - nesting multiple actions is now possible: https://regex101.com/r/etJocJ/3

PR #3168, opened 2019-03-22
Prevent application from crashing if child process crashes - Closes #3142

### LINKED ISSUE #3142: Fix plugin loading in child process and prevent application from crash if child process crashes
### Expected behavior

Any plugin should be able to load up in child process. And if a plugin crashes the parent process should keep running.

### Actual behavior

Plugins can't be loaded at the moment and if a plugin crashes It crashed the parent process.

### Steps to reproduce

Run the application and kill httpApi process.

`feature/child_process_module`

lsilvs:
### What was the problem?
Application was crashing if a child process exited

### How did I fix it?
I've implemented a reload mechanism that reload child process when it crashes without impacting the rest of application

### How to test it?
Run the application with `httpApi` and `chain` modules as child process:
`NODE_ENV=test LISK_CHILD_PROCESS_MODULES=httpApi,chain node src/index.js -n devnet`

Kill `httpApi` process (it should reload the module and keep running normally):
`pgrep -f http_api | xargs kill -9`

Kill `chain` process (it should reload the module and keep running normally):
`pgrep -f chain | xargs kill -9`

### Review checklist

* The PR resolves #3142
* Linting passes
* Tests pass


framework/src/controller/controller.js
```diff
@@ -127,6 +130,7 @@ class Controller {
 			'lisk',
 			['ready'],
 			{
+				isLiskReady: () => this.liskReady,
```
nazarhussain: Why we don't need this action? We must not create actions preferably if the it can be sorted through events. And `lisk:ready` event is the right solution for it.

framework/src/controller/controller.js
```diff
@@ -228,12 +232,17 @@ class Controller {
...
 			this.logger.error(
 				`Module ${moduleAlias}(${name}:${version}) exited with code: ${code} and signal: ${signal}`
 			);
-			// TODO: Reload child instead of exiting the parent process
-			process.exit(1);
+
+			await this.bus.unregisterChannel(moduleAlias);
+
+			// Reload the module
+			this._loadChildProcessModule(alias, Klass, options);
```
nazarhussain: Once child process is `exit`, we must not load it again at least for this release. We can develop more dedicated reload strategy later.

framework/src/modules/chain/index.js
```diff
@@ -85,10 +85,15 @@ module.exports = class ChainModule extends BaseModule {
 
 	async load(channel) {
 		this.chain = new Chain(channel, this.options);
+		const isLiskReady = await channel.invoke('lisk:isLiskReady');
 
-		channel.once('lisk:ready', () => {
+		if (isLiskReady) {
```
nazarhussain: Not the right way to do it, `lisk:ready` event is triggered after `load` is called. So it must be listened as of previous implementation. I suspect you changed this because of module reload, if yes then as I told earlier, we should develop that strategy as an enhancement later. And leave that implication as it was before.

framework/src/controller/controller.js
lsilvs: We need this when reloading the module

framework/src/controller/controller.js
lsilvs: This was discussed and decided to reload the faulty module. It can be improved in next versions but it works pretty well as it is.

nazarhussain: No its not the right solution, it can lead into infinite reloading and exhaust the system memory.

framework/src/modules/chain/index.js
lsilvs: Yes, it was introduced to allow modules to be reloaded in case it crashes and I don't see the reason of reverting this implementation if it works.

framework/src/controller/controller.js
nazarhussain: For current release its good if parent process exit wit proper logs if child process is crashed. Later we can develop a strategy, how to reload modules, or even have configurable options with module load retry etc. This topic needed to have more thought process and a refined solution.

framework/src/modules/chain/index.js
nazarhussain: If child process cause some memory leaks crash, then this infinite loading will exhaust the resources.

lsilvs: I agree and we plan to improve this already. This is the first implementation of child process and wont be the final one. Knowing that this implementation can be unstable, I've created a property in the config file to completely disable it if required.

nazarhussain: I think this change is not related to stability. If there is an issue and any of code crash the whole application exited. That's exactly the default behaviour we have from start and we should keep it consistent in this release. In next release we develop this resilience with a detailed and extensive approach.

framework/src/controller/controller.js
```diff
@@ -228,12 +232,17 @@ class Controller {
 
 		const child = child_process.fork(program, parameters);
 
-		child.on('exit', (code, signal) => {
+		this.childrenList.push(child);
```
shuse2: shouldn't we remove from childrenList when it exits?

framework/src/controller/bus.js
```diff
@@ -133,6 +134,37 @@ class Bus extends EventEmitter2 {
...
+	/**
+	 * Unregister channel from bus.
+	 *
+	 * @async
+	 * @param {string} moduleAlias - Alias for module used during registration
+	 *
+	 * @throws {Error} If channel not registered.
+	 */
+	async unregisterChannel(moduleAlias) {
+		Object.keys(this.events).forEach(eventName => {
```
shuse2: maybe it's a bit dangerous to dynamically delete within `forEach`?
prefer to get the `eventName` which needs to be deleted first and then delete?
same for actions

framework/src/controller/channels/child_process_channel.js
```diff
@@ -137,9 +139,18 @@ class ChildProcessChannel extends BaseChannel {
 	 * @returns {Promise<void>}
 	 */
 	async cleanup() {
+		if (this.pubSocket && typeof this.pubSocket.close === 'function') {
```
yatki: Why do we check for `this.pubSocket.close` function?

framework/src/controller/controller.js
yatki: Where do we handle killing all child processes when a SIGINT received?

framework/src/controller/controller.js
yatki: Why not awaiting this? The promise might get rejected.

framework/src/modules/chain/index.js
```diff
@@ -85,10 +85,15 @@ module.exports = class ChainModule extends BaseModule {
...
 			this.chain.bootstrap();
```
yatki: Isn't bootstrap function is a promise? Shouldn't we await it?

framework/src/modules/chain/index.js
```diff
@@ -85,10 +85,15 @@ module.exports = class ChainModule extends BaseModule {
...
-		});
+		} else {
+			channel.once('lisk:ready', () => {
+				this.chain.bootstrap();
```
yatki: same here: await

framework/src/modules/http_api/index.js
```diff
@@ -41,9 +41,15 @@ class HttpAPIModule extends BaseModule {
 
 	async load(channel) {
 		this.httpApi = new HttpApi(channel, this.options);
-		channel.once('lisk:ready', () => {
+		const isLiskReady = await channel.invoke('lisk:isLiskReady');
+
+		if (isLiskReady) {
 			this.httpApi.bootstrap();
```
yatki: same here: await

framework/src/modules/http_api/index.js
```diff
@@ -41,9 +41,15 @@ class HttpAPIModule extends BaseModule {
...
-		});
+		} else {
+			channel.once('lisk:ready', () => {
+				this.httpApi.bootstrap();
```
yatki: same here: await

framework/src/controller/controller.js
MaciejBaj: We can improve the solution later on, but the resilience should be there.

framework/src/controller/controller.js
lsilvs: `Controller.cleanup`

framework/src/controller/channels/child_process_channel.js
lsilvs: To avoid `Uncaught TypeError`?

framework/src/controller/controller.js
yatki: Great, just one more comment. I think we should also pass the kill signal to child processes. 

https://github.com/LiskHQ/lisk/blob/ec96e8de10104a360b9f0e06bee9a7f55f8806f2/framework/src/controller/controller.js#L281

framework/src/controller/channels/child_process_channel.js
yatki: if this.pubSocket is truthy that means it's already a socket object that has a close method. I'd prefer to not have that control for the sake of simplicity. Because person who writes the tests can think it's relevant with the logic etc.

framework/src/controller/bus.js
lsilvs: Any particular reason for that? Why is it dangerous?

shuse2: As discussed, i think it's fine

framework/src/controller/controller.js
lsilvs: By default it sends `SIGTERM` signal. What you reckon?

yatki: I think if main application was killed with SIGINT, children should be exited with SIGINT too.

lsilvs: To do that we need to change the way we're exiting the application. We can improve that in the next version. Could you open an issue for that, pls?

MaciejBaj:
Requested changes are addressed.

framework/src/controller/controller.js
yatki: https://github.com/LiskHQ/lisk/issues/3185 @lsilvs @MaciejBaj

PR #3169, opened 2019-03-22
Fix Unit Tests - Closes #3134

### LINKED ISSUE #3134: Fix Unit Tests
### Expected behavior

Unit tests pass

### Actual behavior

Unit tests has been broken in feature branch while we were replacing usages with elements.

1.7

pablitovicente:
### What was the problem?

Unit tests were failing

### How did I fix it?

- By fixing unit tests

### How to test it?

- Unit test pipeline should be green
- Pay attention to the few skipped unit tests so we can solve those in a different PR

### Review checklist

* The PR resolves #3134
* Linting passes
* Tests pass


framework/test/mocha/unit/modules/chain/logic/round.js
```diff
@@ -96,6 +96,7 @@ describe('round', () => {
 			round = new Round(scope, task);
 			done();
 		});
+		sinonSandbox.reset();
```
shuse2: this should be in afterEach

pablitovicente: This test was not atomic so I went for beforEach to avoid having to dig deeper into this test, can we split this to a new issue so the rest of the PR can move forward?

PR #3180, opened 2019-03-25
Update api peer endpoints use network module - Closes#2879

### LINKED ISSUE #2879: Update api/peer endpoints to use the network module
### Description
Update the /api/controllers/peer.js in chain module to use new network module created.
Also, it should remove the unused functions, and fix/remove tests related to the loader.

### Dependent issue
#2874

Also, depends on changes made in v1.6

ishantiw:
### What was the problem?

`/api/controllers/peer.js` in chain module is using Peers class from submodules to get the list of peers. It should use Network module to get a list of peers for `api/peers` endpoint.

### How did I fix it?

Use helper functions for filtering in api/controller/peer.js and use network module to invoke `getNetworkStatus` to get the list of peers.

### How to test it?

Run functional and unit tests.

### Review checklist

* The PR resolves #2879 
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/http_api/controllers/peers.js
+++ b/framework/src/modules/http_api/controllers/peers.js
@@ -76,8 +76,8 @@ PeersController.getPeers = async function(context, next) {
 		return next(null, {
 			data: peersByFilters,
 			meta: {
-				offset: filters.offset,
-				limit: filters.limit,
+				offset: params.offset.value,
+				limit: params.limit.value,
 				count: peersCount,
 			},
 		});
--- a/framework/src/modules/http_api/controllers/peers.js
+++ b/framework/src/modules/http_api/controllers/peers.js
@@ -69,7 +69,7 @@ PeersController.getPeers = async function(context, next) {
 	filters = _.pickBy(filters, v => !(v === undefined || v === null));
 
 	try {
-		const peerList = await channel.invoke('network:getNetworkStatus', filters);
+		const peerList = await channel.invoke('network:getNetworkStatus');
 		const peersByFilters = getByFilter(peerList, filters);
 		const peersCount = getCountByFilter(peerList, filters);
 
--- a/framework/src/modules/http_api/controllers/peers.js
+++ b/framework/src/modules/http_api/controllers/peers.js
@@ -16,6 +16,7 @@
 
 const _ = require('lodash');
 const swaggerHelper = require('../helpers/swagger');
+const { getByFilter, getCountByFilter } = require('../helpers/filter_peers');
 
 // Private Fields
 let channel;
@@ -68,22 +69,12 @@ PeersController.getPeers = async function(context, next) {
 	filters = _.pickBy(filters, v => !(v === undefined || v === null));
 
 	try {
-		const data = await channel.invoke('chain:getPeers', {
-			parameters: filters,
-		});
-
-		const clonedData = _.cloneDeep(data);
-		const filteredData = clonedData.map(peer => {
-			const { updated, ...filtered } = peer;
-			return filtered;
-		});
-
-		const peersCount = await channel.invoke('chain:getPeersCountByFilter', {
-			parameters: _.cloneDeep(filters),
-		});
+		const peerList = await channel.invoke('network:getNetworkStatus', filters);
+		const peersByFilters = getByFilter(peerList, filters);
+		const peersCount = getCountByFilter(peerList, filters);
 
 		return next(null, {
-			data: filteredData,
+			data: peersByFilters,
 			meta: {
 				offset: filters.offset,
 				limit: filters.limit,
--- a/framework/src/modules/http_api/controllers/peers.js
+++ b/framework/src/modules/http_api/controllers/peers.js
@@ -17,7 +17,7 @@
 const _ = require('lodash');
 const swaggerHelper = require('../helpers/swagger');
 const { getByFilter, getCountByFilter } = require('../helpers/filter_peers');
-
+const swaggerHelper = require('../helpers/swagger');
 // Private Fields
 let channel;
 
--- a/framework/src/modules/http_api/controllers/peers.js
+++ b/framework/src/modules/http_api/controllers/peers.js
@@ -16,7 +16,6 @@
 
 const _ = require('lodash');
 const swaggerHelper = require('../helpers/swagger');
-const { getByFilter, getCountByFilter } = require('../helpers/filter_peers');
 // Private Fields
 let channel;
 
@@ -68,9 +67,11 @@ PeersController.getPeers = async function(context, next) {
 	filters = _.pickBy(filters, v => !(v === undefined || v === null));
 
 	try {
-		const peerList = await channel.invoke('network:getNetworkStatus');
-		const peersByFilters = getByFilter(peerList, filters);
-		const peersCount = getCountByFilter(peerList, filters);
+		const peersByFilters = await channel.invoke('network:getPeers', filters);
+		const peersCount = await channel.invoke(
+			'network:getPeersCountByFilter',
+			filters
+		);
 
 		return next(null, {
 			data: peersByFilters,
--- a/framework/src/modules/http_api/controllers/peers.js
+++ b/framework/src/modules/http_api/controllers/peers.js
@@ -17,7 +17,6 @@
 const _ = require('lodash');
 const swaggerHelper = require('../helpers/swagger');
 const { getByFilter, getCountByFilter } = require('../helpers/filter_peers');
-const swaggerHelper = require('../helpers/swagger');
 // Private Fields
 let channel;
 
--- a/framework/src/modules/http_api/helpers/filter_peers.js
+++ b/framework/src/modules/http_api/helpers/filter_peers.js
@@ -111,14 +111,14 @@ const getByFilter = (peers, filter) => {
 
 	// Sorting
 	if (filter.sort) {
-		const sort_arr = String(filter.sort).split(':');
-		const auxSortField = _.includes(allowedFields, sort_arr[0])
-			? sort_arr[0]
+		const sortArray = String(filter.sort).split(':');
+		const auxSortField = _.includes(allowedFields, sortArray[0])
+			? sortArray[0]
 			: null;
-		const sort_field = sort_arr[0] ? auxSortField : null;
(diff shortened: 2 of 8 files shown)


framework/src/modules/network/network.js
```diff
@@ -153,7 +155,24 @@ module.exports = class Network {
...
 					data: action.params.data,
 				}),
-			getNetworkStatus: () => this.p2p.getNetworkStatus(),
+			getNetworkStatus: () => {
+				const {
+					connectedPeers,
+					newPeers,
+					triedPeers,
+				} = this.p2p.getNetworkStatus();
+				const uniquerPeersList = _.uniqBy(
```
shuse2: prefer not to use `lodash`, probably easily filterable by `Map`?

framework/src/modules/http_api/helpers/filter_peers.js
```diff
@@ -0,0 +1,159 @@
...
+		'httpPort',
+		'state',
+		'os',
+		'version',
+		'protocolVersion',
+		'broadhash',
+		'height',
+		'nonce',
+	];
+	const limit = filter.limit ? Math.abs(filter.limit) : null;
```
shuse2: I think limit has the default of 20 and max 100?

framework/src/modules/http_api/helpers/filter_peers.js
```diff
@@ -0,0 +1,159 @@
+/*
+ * Copyright © 2018 Lisk Foundation
```
shuse2: can we put this file other than helper? 🙏

framework/src/modules/http_api/helpers/filter_peers.js
```diff
@@ -0,0 +1,159 @@
...
+ *
```
shuse2: Also, can we refactor this file to meet the standard as `elements`?

framework/src/modules/http_api/helpers/filter_peers.js
ishantiw: Where should I keep it in your opinion, because I kept it here after discussing with @jondubois and I'm not sure exactly where it should reside, maybe in chain's `submodules`? But it already has `peer` file

framework/src/modules/http_api/helpers/filter_peers.js
ishantiw: Actually, this function is used for two things, one to get all the peer based on all the filters and other is the number of count of peers for those filters without taking `limit` and `offset` into account. Also, http_api is already taking care of the default value for `limit` as 10 and `offset` as 0.

framework/src/modules/http_api/helpers/filter_peers.js
shuse2: mmm definitely not in chain, but lets keep it here and think the strategy of folder structure later.

ishantiw:
`"@liskhq/lisk-p2p": "0.1.0-alpha.1"` is not suitable to handle seed peers as it's not adding it to any peer list.  We need to use `"@liskhq/lisk-p2p": "0.1.0-alpha.3"` but that is throwing error due to a bug (https://github.com/LiskHQ/lisk-elements/issues/1207) which needs to be fixed in elements

framework/src/modules/http_api/helpers/filter_peers.js
```diff
@@ -0,0 +1,155 @@
...
+		'version',
+		'protocolVersion',
+		'broadhash',
+		'height',
+		'nonce',
+	];
+	const limit = filter.limit ? Math.abs(filter.limit) : null;
+	const offset = filter.offset ? Math.abs(filter.offset) : 0;
+
+	peers = peers.filter(peer => {
```
shuse2: does this need to be reassigned?

framework/src/modules/http_api/helpers/filter_peers.js
```diff
@@ -0,0 +1,155 @@
...
+		let passed = true;
+		_.each(filter, (value, key) => {
```
shuse2:
```suggestion
		Object.entries(filter).forEach(([key, value]) => {
```

framework/src/modules/http_api/helpers/filter_peers.js
```diff
@@ -0,0 +1,155 @@
...
+			// Every filter field need to be in allowed fields, exists and match value
+			if (
+				_.includes(allowedFields, key) &&
```
shuse2:
```suggestion
				allowedFields.includes(key) &&
```

framework/src/modules/http_api/helpers/filter_peers.js
```diff
@@ -0,0 +1,155 @@
...
+			}
+			return true;
+		});
+		return passed;
+	});
+
+	// Sorting
+	if (filter.sort) {
+		const sortArray = String(filter.sort).split(':');
+		const auxSortField = _.includes(allowedFields, sortArray[0])
```
shuse2:
```suggestion
		const auxSortField = allowedFields.includes(sortArray[0])
```

framework/src/modules/http_api/helpers/filter_peers.js
ishantiw: It needs to be reassigned because below, the filtered list is being used to sort and filter based on `limit` and `offset`. But for clarity, I have refactored this by assigning it to a var and using it further

framework/src/modules/http_api/helpers/filter_peers.js
```diff
@@ -0,0 +1,157 @@
...
+
+/**
+ * Sorts peers.
+ *
+ * @todo Add @param tags
+ * @todo Add @returns tag
+ * @todo Add description of the function
+ */
+
+const sortPeers = (field, asc) => (a, b) => {
```
diego-G: We should cover this with unit tests.

framework/test/mocha/functional/http/get/peers.js
```diff
@@ -125,8 +125,11 @@ describe('GET /peers', () => {
 			});
 		});
 	});
-
-	describe('pass data from a real peer', () => {
+	/**
+	 * Skipping this GET /api/peers tests as of now because we are using new p2p library and it needs a different apporach to setup the functional test
+	 */
+	/* eslint-disable mocha/no-skipped-tests */
+	describe.skip('pass data from a real peer', () => {
```
diego-G: I don't think it's good approach. It should be quite straightforward to rewrite this file.

framework/test/mocha/network/index.js
```diff
@@ -28,8 +28,11 @@ const WSPORTS = [];
 _.range(TOTAL_PEERS).map(index => {
 	return WSPORTS.push(5000 + index);
 });
-
-describe(`Start a network of ${TOTAL_PEERS} nodes with address "127.0.0.1", WS ports 500[0-9] and HTTP ports 400[0-9] using separate databases`, () => {
+/**
+ * Skipping network tests as of now because we are using new p2p library and it needs a different apporach to setup a network
+ */
+/* eslint-disable mocha/no-skipped-tests */
+describe.skip(`Start a network of ${TOTAL_PEERS} nodes with address "127.0.0.1", WS ports 500[0-9] and HTTP ports 400[0-9] using separate databases`, () => {
```
diego-G: I don't think it's good approach. We should keep as much network tests as possible. I'm in discussions with @jondubois to make an effort now at the beginning. We should rely on them as the last resort in order to remain in the right path.

framework/src/modules/chain/chain.js
```diff
@@ -123,9 +123,9 @@ module.exports = class Chain {
 
 			// TODO: Use the channel application state for update node info
 			// Publish an event whenever the system headers are updated.
-			this.system.on('update', nodeInfo => {
-				this.channel.publish('chain:system:updateNodeInfo', nodeInfo);
-			});
+			// this.system.on('update', nodeInfo => {
```
diego-G: We should merge https://github.com/LiskHQ/lisk/pull/3261 first and rebase this one to avoid misunderstandings.

framework/test/mocha/functional/http/get/peers.js
ishantiw: As discussed with @shuse2 , we decided to resolve these tests in another PR properly, that's why we are skipping it as of now. Also, the problem is not just refactoring from where I can see, the problem is the way we are starting up the test nodes where it is not able to get the peers list from the config file and also it needs a fix from https://github.com/LiskHQ/lisk/pull/3243 related to nodeInfo, so I think this PR should go after rebasing it from it.

framework/test/mocha/network/index.js
ishantiw: same as https://github.com/LiskHQ/lisk/pull/3180#discussion_r272490126

framework/src/modules/http_api/controllers/peers.js
```diff
@@ -68,25 +68,15 @@ PeersController.getPeers = async function(context, next) {
 	filters = _.pickBy(filters, v => !(v === undefined || v === null));
 
 	try {
-		const data = await channel.invoke('chain:getPeers', {
```
diego-G: If we remove this call, we don't need to expose the actions from chain module. `getPeers` & `getPeersCountByFilter` can be removed from:
https://github.com/LiskHQ/lisk/blob/3a9d826dd84b222049b19df4a9eb508ed7624374/framework/src/modules/chain/chain.js#L218
https://github.com/LiskHQ/lisk/blob/3a9d826dd84b222049b19df4a9eb508ed7624374/framework/src/modules/chain/index.js#L61

framework/src/modules/network/helpers/filter_peers.js
```diff
@@ -0,0 +1,186 @@
...
+		...connectedPeers,
+		...newPeers,
+		...triedPeers,
+	].reduce((uniquePeers, peer) => {
+		const found = uniquePeers.find(findPeer => findPeer.ip === peer.ipAddress);
+
+		if (!found) {
+			const { ipAddress, ...peerWithoutIp } = peer;
+
+			return [...uniquePeers, { ip: ipAddress, ...peerWithoutIp }];
```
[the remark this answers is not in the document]
shuse2: Good point, especially in the huge data.
In general, if the data is small, Immutable way would be better for readability, but for big data, we should avoid this

PR #3186, opened 2019-03-26
Integration test for InMemoryChannel - Partially Closes #3118

### LINKED ISSUE #3118: Integration test `ChildProcessChannel` class and `InMemoryChannel` class
### Expected behavior

The `ChildProcessChannel` class and `InMemoryChannel` class should be covered by integration tests.

### Actual behavior

There is no such test coverage.

## Acceptance Criteria

- ~~Add Integration tests for InMemoryChannel #3186~~
- Add Integration tests for ChildProcessChannel

### Additional Information
#3674 needs to be addressed to write the integration test

yatki:
### What was the problem?
InMemoryChannel didn't have integration tests.
### How did I fix it?
I added integration tests.
### Review checklist

* The PR partially solves #3118 
* Linting passes
* Tests pass

### DIFF
--- a/framework/test/jest/config/setup.js
+++ b/framework/test/jest/config/setup.js
@@ -15,6 +15,8 @@
  */
 
 require('jest-extended');
-require('jest-chain');
+
+// jest chain breaks promise handling.
+// require('jest-chain');
 
 process.env.NODE_ENV = 'test';
--- a/framework/test/jest/specs/integration/in_memory_channel.spec.js
+++ b/framework/test/jest/specs/integration/in_memory_channel.spec.js
@@ -2,9 +2,9 @@ const InMemoryChannel = require('../../../../src/controller/channels/in_memory_c
 const Bus = require('../../../../src/controller/bus');
 const Event = require('../../../../src/controller/event');
 
-const logger = jest.fn(() => ({
+const logger = {
 	info: jest.fn(),
-}));
+};
 
 const config = {
 	ipc: {
@@ -38,10 +38,11 @@ describe('InMemoryChannel', () => {
 	describe('after registering itself to the bus', () => {
 		let inMemoryChannelAlpha;
 		let inMemoryChannelBeta;
+		let bus;
 
 		beforeEach(async () => {
 			// Arrange
-			const bus = new Bus(
+			bus = new Bus(
 				{
 					wildcard: true,
 					delimiter: ':',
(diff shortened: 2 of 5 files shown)


framework/test/jest/specs/integration/in_memory_channel.spec.js
```diff
@@ -0,0 +1,200 @@
...
+	events: ['alpha1', 'alpha2'].map(
+		event => new Event(`${'alphaAlias'}:${event}`)
+	),
+	actions: {
+		multiplyByTwo: action => action.params * 2,
+		multiplyByThree: action => action.params * 3,
+	},
+};
+
+const beta = {
```
shuse2: can we put these constants into the scope of describe?

framework/test/jest/specs/integration/in_memory_channel.spec.js
```diff
@@ -0,0 +1,200 @@
...
+					`${alpha.moduleAlias}:${eventName}`,
+					alphaEventData
+				);
+			});
+		});
+
+		describe('#invoke', () => {
+			it('should be able to invoke its own actions.', async () => {
+				// Act && Assert
+				await expect(
```
shuse2: do we need await here? and below.

framework/test/jest/config/setup.js
```diff
@@ -15,6 +15,8 @@
  */
 
 require('jest-extended');
-require('jest-chain');
+
```
shuse2: probably this will be cleared once the other PR merged in right?

framework/test/jest/specs/integration/in_memory_channel.spec.js
limiaspasdaniel: @shuse2 `invoke` is an async function and to be able to assert it with `resolves` the usage of `await` is needed. See [here](https://jestjs.io/docs/en/tutorial-async#async-await)

framework/test/jest/config/setup.js
yatki: Yeap

framework/test/jest/specs/integration/in_memory_channel.spec.js
yatki: @shuse2 actually I didn't have a strong opinion against this but now I kind of have :D 

Moving variables into describe block, increases the `complexity` score of describe blocks. I think intuitively having them outside of describe tells us that they are not being mutated by test hooks. I prefer to keep only the variables that we are mutating inside the describe block.

@lsilvs @limiaspasdaniel what do you guys think about this?

For instance when 
image

framework/test/jest/specs/integration/in_memory_channel.spec.js
yatki: @shuse2 Problem is assertion is failing but with a Promise :D so if you don't await it you get the following problem.

image

framework/test/jest/specs/integration/in_memory_channel.spec.js
lsilvs: I have no strong opinion on this. LGTM.

PR #3195, opened 2019-03-26
Add functionality for handling exceptions present in the blockchain - Closes #3181

### LINKED ISSUE #3181: Add functionality for handling exceptions present in the blockchain
### Expected behavior
There are some transactions in the testnet and mainnet networks which are handled in a special manner. These transactions should be handled in a similar fashion, as present in the old behavior.
### Actual behavior
There is no logic present for handling exceptions.

SargeKhan:
### What was the problem?
The exceptions in the testnet and mainnet were not handled in the new implementation of transaction processing.
### How did I fix it?
Added exception handling logic for various type of exceptions defined in testnet.
### How to test it?
Run integration tests for transactions/exceptions/
### Review checklist

* The PR resolves #3181 
* Linting passes
* Tests pass


framework/test/mocha/integration/transactions/3_3_votes.js
```diff
@@ -31,7 +31,7 @@ describe('system test (type 3) - voting with duplicate submissions', () => {
 	let t = 0;
 
 	/* eslint-disable no-loop-func */
-	while (i < 30) {
+	while (i < 1) {
```
shuse2: i think i've seen this before too, but probably want to put it back to 30?

framework/test/mocha/integration/transactions/exceptions/duplicate_signatures.js
```diff
@@ -0,0 +1,272 @@
...
+	};
+
+	exceptions.duplicatedSignatures = {
+		'15181013796707110990': [
+			'2ec5bbc4ff552f991262867cd8f1c30a417e4596e8343d882b7c4fc86288b9e53592031f3de75ffe8cf4d431a7291b76c758999bb52f46a4da62a27c8901b60a',
+			'36d5c7da5f54007e22609105570fad04597f4f2b00d46baba603c213eaed8de55e9f3e5d0f39789dbc396330b2d9d4da46b7d67187075e86220bc0341c3f7802',
+		],
+	};
+
+	localCommon.beforeBlock('system_inert_transactions', lib => {
```
shuse2: shouldn't this be in the some scope? beforeEach or after?

framework/test/mocha/integration/common.js
```diff
@@ -60,6 +60,22 @@ function createBlock(library, transactions, timestamp, keypair, previousBlock) {
 	return block;
 }
 
+function createValidBlockWithSlotOffset(library, transactions, slotOffset, cb) {
```
shuse2: does this have to use callback style?

framework/test/mocha/integration/transactions/exceptions/multisignature.js
```diff
@@ -0,0 +1,200 @@
+const expect = require('chai').expect;
```
shuse2:
```suggestion
const { expect } = require('chai');
```
Also, header required

framework/test/mocha/integration/transactions/exceptions/duplicate_signatures.js
```diff
@@ -0,0 +1,272 @@
+const expect = require('chai').expect;
```
shuse2:
```suggestion
const { expect } = require('chai');
```
Also, header required

framework/test/mocha/integration/transactions/exceptions/multisignature.js
```diff
@@ -0,0 +1,200 @@
...
+const Bignum = require('bignumber.js');
+const { transfer } = require('@liskhq/lisk-transactions');
+const localCommon = require('../../common');
+const accountFixtures = require('../../../fixtures/accounts');
+
+const exceptions = global.exceptions;
+
+describe('exceptions for multisingature transactions', () => {
```
mitsujutsu: multisignature

framework/test/mocha/integration/common.js
SargeKhan: All the functions in this file are following callback approach, and I would like to keep this consistency.

framework/test/mocha/integration/transactions/exceptions/duplicate_signatures.js
SargeKhan: Nope, this is a strangely implemented functionality for setting up application for integration tests. This function internally has before and after scope functions which sets up/kills the application, server, and databases. Hence, we can call it inside the describe block.

framework/src/modules/chain/logic/check_transaction_against_exceptions.js
```diff
@@ -0,0 +1,197 @@
...
+const checkSignatures = (transactionResponse, transaction) => {
+	if (!exceptions.signatures.includes(transaction.id)) {
+		return false;
+	}
+
+	if (!transactionResponse.errors.length > 1) {
+		return false;
+	}
+
+	if (transactionResponse.errors[0].dataPath !== '.signature') {
```
mitsujutsu: Are we sure it will always be the case that the first error will match the exception `.dataPath`? For this and the other checker functions in this file.

SargeKhan: Yes, we expect that the transactions will fail only with one error. That's why we have the previous `if` condition.

PR #3218, opened 2019-03-28
Lisk Node failed to start with snapshot mode - Closes #3214

### LINKED ISSUE #3214: Lisk Node failed to start with snapshot mode
### Expected behavior
Node should start snapshot when ran with
`node src/index.js -s 1000 -n testnet`
### Actual behavior
Node failed to start when run `node src/index.js -s 1000 -n testnet`
### Steps to reproduce
`node src/index.js -s 1000 -n testnet`
1.6.0

lsilvs:
### What was the problem?
During snapshot `config.peers.enabled` is set to false preventing to create the webSocket server and crashing when trying to listen. 

### How did I fix it?
I check if peers is enabled before creating and listening to webSocket server.

### How to test it?
`node src/index.js -s 1000 -n testnet`

### Review checklist

* The PR resolves #3214
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/chain/chain.js
+++ b/framework/src/modules/chain/chain.js
@@ -153,27 +153,35 @@ module.exports = class Chain {
 				channel: this.channel,
 			};
 
-			// Lookup for peers ips from dns
-			scope.config.peers.list = await lookupPeerIPs(
-				scope.config.peers.list,
-				scope.config.peers.enabled
-			);
-
 			await bootstrapStorage(scope, global.constants.ACTIVE_DELEGATES);
 			await bootstrapCache(scope);
 
 			scope.bus = await createBus();
 			scope.logic = await initLogicStructure(scope);
 			scope.modules = await initModules(scope);
-			scope.webSocket = await createSocketCluster(scope);
+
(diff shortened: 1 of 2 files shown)


ManuGowda:
@lsilvs tests? I am afraid we had some tests before as @4miners informed me, if not then its good to have a test to check of the snapshot works as expected, at integration level.

ManuGowda:
otherwise the PR LGTM 👍

lsilvs:
@ManuGowda The integration test is there but it tests the functions directly. This issue was present when starting the application which should be tested in the functional level but it's tricky as it requires starting the network and waiting for at least one round before running the snapshotting process.

ManuGowda:
> @ManuGowda The integration test is there but it tests the functions directly. This issue was present when starting the application which should be tested in the functional level but it's tricky as it requires starting the network and waiting for at least one round before running the snapshotting process.

I think it will not be a big issue as we a rounds test waiting for multiple rounds, so for this use case we can just wait for 1 round and see if it works fine, my only worry is to avoiding this type issues at the base level so that we don't delay in releases, see what best you can do.

PR #3231, opened 2019-04-01
Add int range for snapshotting - Closes #2843

### LINKED ISSUE #2843: Only accept integer range values for snapshotting rounds
### Expected behavior
We should accept only the integer range values for snapshotting rounds. If need to specify the full all available rounds we should use `-1` for it. 

To run full rounds snapshot we should be able to do;

```
node app.js -s -1
```

### Actual behavior
Currently we are providing different values, integer for valid rounds and non-integer for full rounds. 
https://github.com/LiskHQ/lisk/blob/cb68ebb96aacc3dccd532de6f926a1c39a2b1411/modules/loader.js#L772

e.g. Currently we can run something like this 

```
node app.js -s all
or
node app.js -s highest
or 
node app.js -s my-name
```

all

michielmulders:
### What was the problem?
Currently, we could run the following for snapshotting
```
node app.js -s all
or
node app.js -s highest
or 
node app.js -s my-name
```

This would result in snapshotting all rounds. 
We want to be able to pass an integer range.

### How did I fix it?
Possible breaking change on commander package to start the application with snapshotting enabled.
`.option('-s, --snapshot [round]', 'verify snapshot')` --> passing a property to snapshotting is now optional.

If you do not pass a property, it results in all rounds being included (max).
Otherwise, you can pass a number greater than `0`.

### How to test it?
Run any of the commands:
`npm start -- -s -n testnet` -> ✅(all rounds)
`npm start -- -s 5 -n testnet` -> ✅(number you specified)
`npm start -- -s all -n testnet` -> ❌(throws error: only integer allowed >= 0)
`npm start -- -s -1 -n testnet` -> ❌(`commander` throws error: invalid flag)

### Review checklist

* The PR resolves #2843 
* Linting passes
* Tests pass


framework/src/modules/chain/submodules/loader.js
```diff
@@ -755,8 +755,14 @@ __private.createSnapshot = height => {
 	}
 
 	const snapshotRound = library.config.loading.snapshotRound;
+	if (!snapshotRound && Number.isNaN(parseInt(snapshotRound))) {
```
fchavant: wouldn't `Number.isNaN(parseInt(snapshotRound))` be `false` for e.g. `-3`?

michielmulders: It's not possible to pass this as `commander` considers this as a flag and fails.
But I can add it as an extra check in the code?

fchavant: An extra check or at least a comment would be nice: at the moment the code does not quite correspond to the error message.

michielmulders: It throws `error: unknown option `-1'` which is pretty clear I think. Should make you think about not snapshotting for negative numbers

fchavant: Looking at the code it's not so clear (`Number.isNaN(parseInt(snapshotRound))` vs. `'Unable to create snapshot, snapshot parameters should be an integer equal to or greater than zero'`); maybe it is just me though, feel free to ignore.

framework/src/modules/chain/submodules/loader.js
```diff
@@ -755,8 +755,18 @@ __private.createSnapshot = height => {
 	}
 
 	const snapshotRound = library.config.loading.snapshotRound;
+	// Negative number not possible as `commander` does not recognize this as valid flag (throws error)
+	if (
+		(!snapshotRound && Number.isNaN(parseInt(snapshotRound))) ||
+		parseInt(snapshotRound) < 0
+	) {
+		throw new Error(
+			'Unable to create snapshot, "--snapshot" parameter should be an integer equal to or greater than zero'
```
nazarhussain: > parameter should be an integer equal to or greater than zero

This definition does not apply to examples you mentioned in the PR. So you can actually pass snapshot parameter without any value and it will be considered as all rounds. Please fix this comment or enforce user to pass `0` incase he wants to snapshot all rounds.

framework/src/modules/chain/helpers/config.js
```diff
@@ -70,7 +70,7 @@ function Config(packageJson, parseCommandLineOptions = true) {
 		.option('-a, --address <ip>', 'listening host name or ip')
 		.option('-x, --peers [peers...]', 'peers list')
 		.option('-l, --log <level>', 'log level')
-		.option('-s, --snapshot <round>', 'verify snapshot')
+		.option('-s, --snapshot [round]', 'verify snapshot')
```
nazarhussain: Revert this change to make it a required attribute.

framework/src/modules/chain/submodules/loader.js
```diff
@@ -755,6 +755,13 @@ __private.createSnapshot = height => {
 	}
 
 	const snapshotRound = library.config.loading.snapshotRound;
+	// Negative number not possible as `commander` does not recognize this as valid flag (throws error)
+	if (Number.isNaN(parseInt(snapshotRound)) || parseInt(snapshotRound) < 1) {
+		throw new Error(
+			'Unable to create snapshot, "--snapshot" parameter should be an integer greater than zero or empty which corresponds to all rounds'
+		);
+	}
```
nazarhussain: Adjust this message accordingly to change.

nazarhussain:
This change requires to updated in the scripts for snapshotting. @michielmulders Please create an issue for Lisk scripts and get @fchavant on board for it.

nazarhussain:
@shuse2 @MaciejBaj This change may impact people if they are using custom scripts for snapshotting. We should highlight it in release notes.

4miners:
> This change requires to updated in the scripts for snapshotting. @michielmulders Please create an issue for Lisk scripts and get @fchavant on board for it.

This is not required as we are no longer using `snapshotting` feature to make snapshots (since https://github.com/LiskHQ/lisk/issues/3109).

PR #3232, opened 2019-04-01
Rewrite reserved keywords

### LINKED ISSUE #2981: Avoid usage of variable name which is more like reserved keywords
### Expected behavior
use variable name more explicit, in this case it should be `account` related naming convention.
### Actual behavior
As per the discussion here https://github.com/LiskHQ/lisk/pull/2972/files#r260342366

use of variable `object` is not correct, use variable name more explicit, in this case it should be `account` related naming convention.
As part of this issue, you can also identify other places where have such usage and fix it.

https://github.com/LiskHQ/lisk/blob/9bd1d7b2b16a59db07ee4bb9f6b82a3d73c4456c/framework/src/modules/chain/api/controllers/accounts.js#L44
### Steps to reproduce
https://github.com/LiskHQ/lisk/blob/9bd1d7b2b16a59db07ee4bb9f6b82a3d73c4456c/framework/src/modules/chain/api/controllers/accounts.js#L44

michielmulders:
### What was the problem?
Doing chores: Rewrite variable names for reserved keywords like object / default / ...

### How did I fix it?
Used more specific naming.
Checked for:
`const/let object`
`const/let default`
`const/let debugger`
`const/let native`
`const/let class`

Did not change occurrences like:
`const objects` or `const defaults`

### Review checklist

* The PR resolves #2981 
* Linting passes
* Tests pass


framework/src/controller/event.js
```diff
@@ -65,10 +65,13 @@ class Event {
 	 * @return {module.Event}
 	 */
 	static deserialize(data) {
-		let object = null;
-		if (typeof data === 'string') object = JSON.parse(data);
-		else object = data;
-		return new Event(`${object.module}:${object.name}`, object.data);
+		let parsedEvent = null;
+		if (typeof data === 'string') parsedEvent = JSON.parse(data);
```
shuse2: Although it was used before this PR, if should consistently use `{}`

PR #3234, opened 2019-04-01
Broadcast verified transactions among peers in feature/improve_transaction_processing - Closes #3190

### LINKED ISSUE #3190: Broadcast verified transactions among peers in feature/improve_transaction_processing
### Expected behavior
The verified transactions should be broadcasted to the peers after the transactions are verified in the transaction pool. 
### Actual behavior
The new transaction pool does not broadcast transactions.

pablitovicente:
### What was the problem?

When transactions were added to the pool they were not being broadcasted

### How did I fix it?

By adding the call to `bus.message()`

### How to test it?

When sending transactions broadcaster should receive them too

### Review checklist

* The PR resolves #3190
* Linting passes
* Tests pass


framework/src/modules/chain/submodules/blocks/process.js
```diff
@@ -19,6 +19,7 @@ const async = require('async');
 const { Status: TransactionStatus } = require('@liskhq/lisk-transactions');
 const slots = require('../../helpers/slots');
 const definitions = require('../../schema/definitions');
+const InitTrs = require('../../logic/init_transaction');
```
SargeKhan: InitTransaction is passed through via scope to the constructor function in the rest of the application. We should follow the same practice.

pablitovicente:
Updated tests to include initTransactions @SargeKhan could you please check?

PR #3243, opened 2019-04-02
Update loader module in chain to use network module and fix the tests - Closes #2875

### LINKED ISSUE #2875: Update loader module in chain to use network module and fix the tests
### Description
Update the loader.js in chain module to use new network module created.
Also, it should remove the unused functions, and fix/remove tests related to the loader.

### Dependent issue
#2874

jondubois:
### What was the problem?

The loader was interacting with individual peers directly (e.g. for synching).

### How did I fix it?

- The loader module now interacts with the network via the new network module.
- Fixed a few issues related to the network module.

### How to test it?

- Run the functional and integration tests.
- Sync the node with testnet and check that the loader functionality still works.

### Review checklist

* The PR resolves #2875
* Linting passes
* Tests pass


jondubois:
@ManuGowda I added you because it would be nice if QA team could test this branch on the network to verify that synching still works and that it doesn't introduce any new issues.

I was able to verify that synching still works by running on my local machine and synching with testnet (it was correctly adding blocks to the DB table). I did notice some disconnection errors (which is expected because my node was not exposed to the internet). It would be good to check that these errors do not show up when running the node on a remote instance as part of testnet.

framework/src/modules/chain/submodules/loader.js
```diff
@@ -184,70 +185,54 @@ __private.syncTimer = function() {
...
-					if (err) {
-						modules.peers.remove(peer);
-						return setImmediate(waterCb, err);
-					}
-					return library.schema.validate(
-						res,
-						definitions.WSSignaturesResponse,
-						validateErr => setImmediate(waterCb, validateErr, res.signatures)
-					);
+__private.loadSignatures = async function() {
```
ManuGowda: I am emphasising more on naming at some places, here also I think we can name it better `getSignaturesFromNetwork` which is more verbose and clear. I know that we used the names before, I think we can correct the naming of function and logs accordingly for more meaningful name.

framework/src/modules/chain/submodules/loader.js
```diff
@@ -184,70 +185,54 @@ __private.syncTimer = function() {
...
+	library.logger.info('Loading signatures from the network');
+
+	// TODO: Add target module to procedure name. E.g. chain:getSignatures
+	const result = await library.channel.invoke('network:request', {
+		procedure: 'getSignatures',
```
ManuGowda: one example for above comment is, we call procedure `getSignatures` but the root function name is `loadSignatures` which is incosistant, so `getSignaturesFromNetwork` would make more sense. this comment is just an explanation for above comment

framework/src/modules/chain/submodules/loader.js
```diff
@@ -184,70 +185,54 @@ __private.syncTimer = function() {
...
+	});
+
+	const validate = promisify(library.schema.validate.bind(library.schema));
+
+	await validate(result, definitions.WSSignaturesResponse);
```
ManuGowda: are we not handling error if the validation throws some error here?

framework/src/modules/chain/submodules/loader.js
```diff
@@ -184,70 +185,54 @@ __private.syncTimer = function() {
...
-					return library.schema.validate(
-						res,
-						definitions.WSSignaturesResponse,
-						validateErr => setImmediate(waterCb, validateErr, res.signatures)
-					);
+__private.loadSignatures = async function() {
+	library.logger.info('Loading signatures from the network');
+
+	// TODO: Add target module to procedure name. E.g. chain:getSignatures
+	const result = await library.channel.invoke('network:request', {
```
ManuGowda: use of destructuring?

framework/src/modules/chain/submodules/loader.js
```diff
@@ -256,96 +241,67 @@ __private.loadSignatures = function(cb) {
...
-								module: 'loader',
-								transaction,
-							});
-
-							library.logger.warn(
-								`Transaction ${id} is not valid, peer removed`,
-								peer.string
-							);
-							modules.peers.remove(peer);
+__private.loadTransactions = async function() {
```
ManuGowda: same here `getTransactionsFromNetwork`

framework/src/modules/chain/submodules/loader.js
```diff
@@ -256,96 +241,67 @@ __private.loadSignatures = function(cb) {
...
-								return setImmediate(eachSeriesCb);
-							}
-						);
-					},
-					waterCb
+	const validate = promisify(library.schema.validate.bind(library.schema));
+
+	try {
+		await validate(result, definitions.WSTransactionsResponse);
+	} catch (validateErrors) {
```
ManuGowda: ❤️

ManuGowda: please do the same for signatures

framework/src/modules/chain/submodules/loader.js
```diff
@@ -256,96 +241,67 @@ __private.loadSignatures = function(cb) {
...
+
+	const transactions = result.transactions;
+
+	transactions.forEach(transaction => {
+		const transactionId = transaction ? transaction.id : 'null';
+
+		try {
+			library.logic.transaction.objectNormalize(transaction);
+		} catch (error) {
+			library.logger.debug('Transaction normalization failed', {
```
ManuGowda: `logger.debug`? not `logger.error`?

framework/src/modules/chain/submodules/loader.js
```diff
@@ -256,96 +241,67 @@ __private.loadSignatures = function(cb) {
...
+					false,
+					addSequenceCb
 				);
-			},
-		],
-		err => setImmediate(cb, err)
-	);
+			});
+		} catch (error) {
+			library.logger.debug(error);
```
ManuGowda: same as above

framework/src/modules/network/network.js
```diff
@@ -69,63 +84,83 @@ module.exports = class Network {
...
+
 		this.p2p.on(EVENT_DISCOVERED_PEER, peerInfo => {
 			this.logger.info(
 				`Discovered peer ${peerInfo.ipAddress}:${peerInfo.wsPort}`
 			);
 		});
+
 		this.p2p.on(EVENT_FAILED_TO_FETCH_PEER_INFO, error => {
-			this.logger.debug(error.message);
+			this.logger.debug(error.message || error);
```
ManuGowda: why `logger.debug`? why not `logger.error`?  same for below.

framework/src/modules/network/network.js
```diff
@@ -69,63 +84,83 @@ module.exports = class Network {
...
-					request.procedure,
+					sanitizedProcedure,
 					request.data
 				);
-				request.end(result); // Send the response back to the peer.
 				this.logger.info(`Responsed to peer request ${request.procedure}`);
+				request.end(result); // Send the response back to the peer.
 			} catch (error) {
-				request.error(error); // Send an error back to the peer.
 				this.logger.debug(
```
ManuGowda: same here as well.

framework/src/modules/chain/chain.js
```diff
@@ -272,6 +295,40 @@ module.exports = class Chain {
...
+					// Discounting maxium compression setting used in middleware
+					// Maximum transport payload = 2000000 bytes
+					const query = action.params[0] || {};
+					this.scope.modules.blocks.utils.loadBlocksData(
+						{
+							limit: 34, // 1977100 bytes
+							lastId: query.lastBlockId,
+						},
+						(err, data) => {
+							(data || []).forEach(block => {
```
shuse2: prefer early return if data` is undefined

framework/src/controller/bus.js
```diff
@@ -106,9 +106,6 @@ class Bus extends EventEmitter2 {
 	 * @throws {Error} If event name does not exist to bus.
 	 */
 	emit(eventName, eventValue) {
-		if (!this.getEvents().includes(eventName)) {
```
shuse2: please create an issue to recreate the check

framework/src/modules/network/network.js
```diff
@@ -69,63 +84,83 @@ module.exports = class Network {
...
 			);
 		});
+
 		this.p2p.on(EVENT_DISCOVERED_PEER, peerInfo => {
 			this.logger.info(
 				`Discovered peer ${peerInfo.ipAddress}:${peerInfo.wsPort}`
 			);
 		});
+
 		this.p2p.on(EVENT_FAILED_TO_FETCH_PEER_INFO, error => {
```
ishantiw: I was about to fix it in my PR as it was debugging a lot of undefined messages. 👍

framework/src/modules/network/network.js
```diff
@@ -37,20 +40,32 @@ module.exports = class Network {
...
 
-		const initialNodeInfo = await this.channel.invoke('chain:getNodeInfo');
+		// TODO: Nonce overwrite should be removed once the Network module has been fully integreated into core and the old peer system has been fully removed.
+		// We need this because the old peer system which runs in parallel will conflict with the new one if they share the same nonce.
+		const moduleNonce = randomstring.generate(16);
+		const sanitizeNodeInfo = nodeInfo => (
+			{
+				...nodeInfo,
+				state: 2, // TODO: Delete state property
+				nonce: moduleNonce,
```
ishantiw: Shouldn't we use nonce coming from the config? `this.options.config.nonce`?

framework/src/modules/chain/submodules/blocks.js
```diff
@@ -78,7 +78,8 @@ class Blocks {
 				scope.schema,
 				scope.components.storage,
 				scope.sequence,
-				scope.genesisBlock
+				scope.genesisBlock,
+				scope.channel
```
diego-G: Why do we need this?

framework/src/modules/network/network.js
```diff
@@ -12,8 +12,11 @@ const {
 	EVENT_REQUEST_RECEIVED,
```
diego-G: I miss tests for this entire file.

framework/test/mocha/unit/modules/chain/submodules/loader.js
```diff
@@ -455,286 +379,4 @@ describe('loader', () => {
 			});
 		});
 	});
-
-	describe('__private.loadBlocksFromNetwork', () => {
```
diego-G: Why do we remove this instead of adapting it to the new times?

framework/src/modules/chain/submodules/loader.js
diego-G: We should cover now this with unit tests.

framework/src/modules/chain/submodules/loader.js
diego-G: We should cover now this with unit tests.

framework/src/modules/chain/submodules/loader.js
```diff
@@ -15,6 +15,7 @@
 'use strict';
```
diego-G: I would suggest to clean and move the entire file to a ES6 class while removing callbacks and using async await when needed.

framework/src/modules/chain/chain.js
```diff
@@ -272,6 +295,40 @@ module.exports = class Chain {
...
+			blocks: async action =>
+				new Promise(resolve => {
+					// Get 34 blocks with all data (joins) from provided block id
+					// According to maxium payload of 58150 bytes per block with every transaction being a vote
+					// Discounting maxium compression setting used in middleware
+					// Maximum transport payload = 2000000 bytes
+					const query = action.params[0] || {};
+					this.scope.modules.blocks.utils.loadBlocksData(
+						{
+							limit: 34, // 1977100 bytes
```
diego-G: We should research and update this number. If I remember well it's not accurate. Block size varies depending on the transaction associated to it. We might be able to increase the number significantly.

framework/src/modules/chain/chain.js
```diff
@@ -272,6 +295,40 @@ module.exports = class Chain {
 					action.params[0],
 					action.params[1]
 				),
+			blocks: async action =>
```
diego-G: I would suggest to pass the requested number of blocks as a variable, or returning a dynamic number depending on the sizes of the requested blocks.

diego-G:
There is an error when syncing with Testnet. The application is not able to connect to the peers, therefore non receiving any blocks. For reference:
`Failed to load blocks from the network. Error loading blocks: Request failed due to no peers found in peer selection`

framework/src/modules/chain/submodules/loader.js
jondubois: No because the function is `async` so the caolan `async.retry` module will capture the error.

framework/src/modules/chain/submodules/loader.js
jondubois: Can't because it we need to pass it to the `validate` function.

framework/src/modules/chain/chain.js
```diff
@@ -181,7 +181,9 @@ module.exports = class Chain {
 			scope.bus.message('bind', scope);
 
 			// Listen to websockets
-			await scope.webSocket.listen();
+			if (scope.config.peers.enabled) {
+				await scope.webSocket.listen();
```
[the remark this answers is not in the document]
jondubois: That's for the old peer system. We can't remove this logic right now because the tests would fail. I just added this `if` check because of a bug in the config (if the old peers system is turned off in the main config, it used to throw an error here).

framework/src/modules/chain/submodules/loader.js
```diff
@@ -184,70 +185,54 @@ __private.syncTimer = function() {
...
-					);
+__private.loadSignatures = async function() {
+	library.logger.info('Loading signatures from the network');
+
+	// TODO: Add target module to procedure name. E.g. chain:getSignatures
+	const result = await library.channel.invoke('network:request', {
+		procedure: 'getSignatures',
+	});
+
+	const validate = promisify(library.schema.validate.bind(library.schema));
```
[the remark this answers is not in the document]
jondubois: Yes.

framework/src/modules/network/network.js
jondubois: Eventually yes, but for now in order for it to work with both peer systems, we need each one to have a different nonce that way peers will treat us as if we are two different nodes.

framework/src/modules/chain/submodules/blocks.js
jondubois: Because the file `lisk/framework/src/modules/chain/submodules/blocks/process.js` needs to invoke a network action - This may be refactored later but right now it's the simplest thing to do.

framework/test/mocha/unit/modules/chain/submodules/loader.js
jondubois: These unit tests and the underlying logic were poorly implemented. This refactoring improves this a bit but it's still not in a sufficiently clean/settled state to be unit tested.

framework/src/modules/chain/submodules/loader.js
jondubois: I think it's better to focus more on functional tests at this stage because the implementation details are still not in a settled state.

framework/src/modules/chain/submodules/loader.js
jondubois: Makes sense. I did this for some functions but it's safer if we do this in multiple steps.

framework/src/modules/chain/chain.js
jondubois: We should raise a new issue.

framework/src/modules/chain/chain.js
jondubois: Yes I think as a new issue.

framework/src/modules/chain/chain.js
shuse2: For this current protocol, let's keep it this way because it's not really a protocol change, and we don't want to risk here.
However, please create an issue to address this, and we can tackle this when we change the p2p protocol

framework/src/modules/chain/chain.js
shuse2: @jondubois @diego-G same as above, we don't want to change the behavior now, so please create a new issue, and we will tackle later

jondubois:
@shuse2 I raised an issue about allowing some modules to publish any event https://github.com/LiskHQ/lisk/issues/3263

framework/src/modules/chain/submodules/blocks/process.js
```diff
@@ -428,32 +328,33 @@ Process.prototype.loadBlocksOffset = function(
...
-			(err, res) => {
-				err = err || res.error;
-				if (err) {
-					modules.peers.remove(peer);
-					return setImmediate(seriesCb, err);
-				}
-				return setImmediate(seriesCb, null, res.blocks);
-			}
-		);
+	async function getBlocksFromNetwork() {
```
shuse2: if we keep using this `async.waterfall` we need to pass the callback?

framework/src/modules/chain/submodules/loader.js
```diff
@@ -255,96 +239,62 @@ __private.loadSignatures = function(cb) {
...
-								module: 'loader',
-								transaction,
-							});
-
-							library.logger.warn(
-								`Transaction ${id} is not valid, peer removed`,
-								peer.string
-							);
-							modules.peers.remove(peer);
+__private.getTransactionsFromNetwork = async function() {
```
shuse2: i don't think this can be `async` because it's used in `async` library which expects callback style

framework/src/modules/chain/submodules/loader.js
```diff
@@ -183,70 +184,53 @@ __private.syncTimer = function() {
...
-					if (err) {
-						modules.peers.remove(peer);
-						return setImmediate(waterCb, err);
-					}
-					return library.schema.validate(
-						res,
-						definitions.WSSignaturesResponse,
-						validateErr => setImmediate(waterCb, validateErr, res.signatures)
-					);
+__private.getSignaturesFromNetwork = async function() {
```
shuse2: i don't think this can be async because it's used in async library which expects callback style

framework/test/mocha/unit/modules/chain/submodules/loader.js
diego-G: So how do you assure it's working well?

framework/src/modules/chain/submodules/loader.js
diego-G: If the implementation details should be clear enough because what we want it's to replicate the exact behaviour that it's currently working.
Anyways, I might agree but then I miss new functional test cases in the PR and we cannot skip any network tests.

framework/src/modules/chain/submodules/loader.js
diego-G: We can proceed as you desire. Just open consequents issues to convert the file to the ES6 standard and cover with tests.

shuse2: @diego-G #3280

framework/src/modules/chain/submodules/blocks/process.js
```diff
@@ -240,108 +242,6 @@ __private.receiveForkFive = function(block, lastBlock, cb) {
...
-						// SQL error occurred
-						library.logger.error(err.stack);
-						return setImmediate(waterCb, 'Blocks#getCommonBlock error');
-					});
-			},
-		],
-		(err, res) => {
-			// If comparison failed and current consensus is low - perform chain recovery
-			if (comparisonFailed && modules.transport.poorConsensus()) {
-				return modules.blocks.chain.recoverChain(cb);
```
shuse2: this logic needs to stay

shuse2: #3281

jondubois: Ah yep, we need the `recoverChain(...)` feature but it should be possible to use it as part of the `loadBlocksFromNetwork` logic instead - It would be nice to remove the `commonBlocks` logic.

framework/src/modules/chain/chain.js
diego-G: @jondubois would you mind to open the issue?

diego-G:
@jondubois @shuse2 in order to move on with PRs from P2P epic, what would you say if at least we create thoughtful test skeletons that we will cover in the close future?

jondubois:
@diego-G I think we need more (and better) functional tests but it should be a separate issue which we can work towards gradually.

I don't think that we should write unit tests (or the skeleton) until the logic has settled into a desirable working state first - Until we finish migrating to the new Network module, our internal logic is still likely to change and we should keep this flexibility.

PR #3246, opened 2019-04-02
Modify validation chain and update config.json - Closes #2959

### LINKED ISSUE #2959: Update schema validation support for chain module on provided options
### Expected behavior
Provided validation for chain modules needs changes, e.g.:
- default value incorrect types
- validation for peer objects
- inconsistent type for wsPort in `config/{network}`

### Actual behavior

Validation present but not complete.

1.6.0

michielmulders:
### What was the problem?
Review and update validation for `chain` component.

### How did I fix it?
Changes:
- `wsPort` needs to be an integer, however, few `/config/{network)/config.json` configurations still had a string (while some others had the correct integer).
- Added extra validation (see changes commit)

### How to test it?
Start network and log chain config to see how the default config is applied for each network type.

### Review checklist

* The PR resolves #2959
* Linting passes
* Tests pass


framework/src/modules/chain/defaults/config.js
```diff
@@ -325,6 +341,10 @@ const DefaultConfig = {
 					properties: {
 						blackList: {
 							type: 'array',
+							items: {
+								type: 'string',
+								format: 'ipOrFQDN',
```
nazarhussain: I am not sure if that format is valid here. As in the list when you specify the domain name during the startup we translate it to the IP and in application life cycle we always keep the IP. So if we match it with the blackList, that list must contains only IPs. 

Please verify from the code usage for its format.

PR #3247, opened 2019-04-02
Move broadhash computation to chain module - Close #3141

### LINKED ISSUE #3141: Move broadhash computation logic from applicationState.update() to chain module
### Expected behavior

Application state class should be as simple as possible only managing the state and producing the events after updates.

### Actual behavior

Application state uses crypto library and blocks to calculate the broadhash. This should be performed in the chain module where the business logic belongs.

`1.6`

diego-G:
### How to test it?
```
npm run jest:unit -- framework/test/jest/specs/unit/controller/application_state.spec.js
npx mocha framework/test/mocha/unit/modules/chain/submodules/blocks.js 
npx mocha framework/test/mocha/unit/modules/chain/submodules/blocks/
```

### Review checklist

* The PR resolves #3141
* Linting passes
* Tests pass


framework/test/jest/specs/unit/controller/application_state.spec.js
```diff
@@ -0,0 +1,197 @@
...
+					.mockImplementationOnce(() =>
+						Promise.reject(new Error(errorMessage))
+					),
+			};
+			try {
+				// Act
+				await applicationState.update(newState);
+			} catch (error) {
+				// Assert
+				expect(error).toBeInstanceOf(Error);
```
limiaspasdaniel: This try catch can be simplified as the following:

`await expect(applicationState.update(newState).rejects.toThrow(errorMessage)`

Note that `toThrow` already parses the `Error` object's message and asserts it agains `errorMessage` var. This way there is no need to assert line 124

limiaspasdaniel: Also I would check that the error is being logged

framework/test/jest/specs/unit/controller/application_state.spec.js
```diff
@@ -0,0 +1,197 @@
...
+			});
+
+			afterEach(async () => {
+				// Clean
+				spies.get.mockRestore();
+			});
+
+			it('should call get four times', async () => {
+				// Assert
+				expect(spies.get).toHaveBeenCalledTimes(4);
```
limiaspasdaniel: Just wondering,why should it be called 4 times?

framework/test/jest/specs/unit/controller/application_state.spec.js
```diff
@@ -0,0 +1,197 @@
...
+			});
+
+			it('should update broadhash', async () => {
+				// Assert
+				expect(applicationState.state.broadhash).toBe(newState.broadhash);
+				expect(updatedState.broadhash).toBe(newState.broadhash);
```
limiaspasdaniel: Why this assert? `applicationState.state` and `updatedState` are the same object.

framework/test/jest/specs/unit/controller/application_state.spec.js
```diff
@@ -0,0 +1,197 @@
...
+			});
+
+			it('should update height', async () => {
+				// Assert
+				expect(applicationState.state.height).toBe(newState.height);
+				expect(updatedState.height).toBe(newState.height);
```
limiaspasdaniel: Same here

framework/test/mocha/unit/modules/chain/submodules/blocks.js
```diff
@@ -414,4 +434,77 @@ describe('blocks', () => {
 			return expect(__private.loaded).to.be.true;
 		});
 	});
+
+	describe('calculateNewBroadhash()', () => {
+		describe('when there is a prolem getting blocks from the db', () => {
```
limiaspasdaniel: Typo on `prolem`

framework/src/modules/chain/submodules/blocks/chain.js
```diff
@@ -837,14 +837,9 @@ Chain.prototype.deleteLastBlock = function(cb) {
...
-					{},
-					{
-						limit: 5,
-						sort: 'height:desc',
-					}
-				)
-					.then(blocks => {
+				return modules.blocks
+					.calculateNewBroadhash()
+					.then(newState => {
```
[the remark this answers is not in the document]
diego-G: It's not available inside the `async.series`.

framework/test/jest/specs/unit/controller/application_state.spec.js
diego-G: Three times inside the method plus another one from the beforeEach.

framework/src/modules/chain/submodules/blocks/chain.js
```diff
@@ -837,14 +837,9 @@ Chain.prototype.deleteLastBlock = function(cb) {
...
 						// Listen for the update of step to move to next step
 						library.channel.once('lisk:state:updated', () => {
 							seriesCb();
```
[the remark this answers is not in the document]
diego-G: The aim is to follow the event-driven approach. So for now, we need to explicitly wait for the next notification update to continue the `async.series`.
As for the use of await, it's not available inside the `async.series.`

framework/test/jest/specs/unit/controller/application_state.spec.js
```diff
@@ -0,0 +1,202 @@
...
+			expect(applicationState.logger).toBe(logger);
+			expect(applicationState.stateChannel).toBe(undefined);
+			expect(applicationState.state).toEqual(mockedState);
+		});
+	});
+
+	describe('#get state', () => {
+		it('should call get state', () => {
+			// Arrange
+			const spy = jest.spyOn(applicationState, 'state', 'get');
```
shuse2: no need to spy imo.
As long as `state` is equaled to mocked state, test should be sufficient

framework/test/jest/specs/unit/controller/application_state.spec.js
```diff
@@ -0,0 +1,202 @@
...
+
+			// Act
+			applicationState.channel = channel;
+
+			// Assert
+			expect(spy).toHaveBeenCalled();
+			expect(applicationState.stateChannel).toBe(channel);
+
+			// Clean
+			spy.mockRestore();
```
shuse2: same as above

framework/test/jest/specs/unit/controller/application_state.spec.js
```diff
@@ -0,0 +1,202 @@
...
+		});
+	});
+
+	describe('#set channel', () => {
+		it('should set the channel', () => {
+			// Arrange
+			const spy = jest.spyOn(applicationState, 'channel', 'set');
```
shuse2: same as above

framework/test/jest/specs/unit/controller/application_state.spec.js
```diff
@@ -0,0 +1,202 @@
...
+
+			beforeEach(() => {
+				applicationState.channel = {
+					publish: jest
+						.fn()
+						.mockImplementation(() => Promise.reject(new Error(errorMessage))),
+				};
+			});
+
+			it('should return error', () => {
```
shuse2:
```suggestion
			it('should throw an error', () => {
```

framework/test/jest/specs/unit/controller/application_state.spec.js
```diff
@@ -0,0 +1,202 @@
...
+				expect(applicationState.update(newState)).rejects.toThrow(errorMessage);
+			});
+
+			it('should log the error', async () => {
```
shuse2: do we really need to keep this check for logging too?

framework/test/jest/specs/unit/controller/application_state.spec.js
```diff
@@ -0,0 +1,202 @@
...
+			};
+
+			// Act
+			result = await applicationState.update(newState);
+			updatedState = applicationState.state;
+		});
+
+		afterEach(async () => {
+			// Clean
+			spies.get.mockRestore();
```
shuse2: same as above

framework/test/jest/specs/unit/controller/application_state.spec.js
```diff
@@ -0,0 +1,202 @@
...
+			// Assert
+			expect(updatedState.broadhash).toBe(newState.broadhash);
+		});
+
+		it('should update height', async () => {
+			// Assert
+			expect(updatedState.height).toBe(newState.height);
+		});
+
+		it('should print notification update in logs', async () => {
```
shuse2: same as above

framework/test/jest/specs/unit/controller/application_state.spec.js
diego-G: I guess so as long as the log remains.

framework/src/controller/application_state.js
```diff
@@ -81,33 +79,18 @@ class ApplicationState {
...
 	 *
+	 * @param {broadhash, height} parameters - broadhash and height to update
+	 *
 	 * @returns {Promise.<boolean, Error>}
 	 */
-	async update(blocks) {
-		if (!blocks) {
-			throw new TypeError('Argument blocks should be an array.');
-		}
+	async update({ broadhash, height }) {
```
nazarhussain: In this scope the `broadhash` and `height` both are required and coupled to each other. So we we add some assertion here that user must pass both to update the system state. Otherwise we throw some error.

framework/src/modules/chain/submodules/blocks/chain.js
nazarhussain: We are calling `.calculateNewBroadhash()` and then using variable name `newState`, which leads to understanding that first method is returning a state, which is not true. We may change it to 

```
then ({broadhash, height} => {
```
That will make clear what first method is returning.

Issue #3152, opened 2019-03-22
Change the API response to return an array of errors


SargeKhan:
### Expected behavior
The API response should return all the errors encountered when processing a request. Therefore, the structure of the API response should be:
```
{
   "code": 404,
   "error": [{
      message: "<ERROR_MESSAGE>"
   },
   {
      ...
   }]
}
```
### Actual behavior
The structure of API response in case of an error is not very extensible. It contains only one property with the name `message` which is of type string. In cases where there are multiple errors, it is impossible to return all the errors in this format. The current structure of the API response in case of an error is as follow:
```
{
   "code": 404,
   "message": "<ERROR_MESSAGE>"
}
```
 
### Which version(s) does this affect? (Environment, OS, etc...)
1.7

pablitovicente:
We should also check `processUnconfirmedTransaction()` in transaction pool as it's only returning one error

pablitovicente:
We should fix any tests tagged with `[feature/improve_transactions_processing_efficiency]` on this PR or if scope goes to much up then we should split in different issue blocked by this one

mitsujutsu:
> We should also check `processUnconfirmedTransaction()` in transaction pool 

Yes that's fixed in the branch I'm working on. I am in the process of updating tests.

pablitovicente:
@mitsujutsu be aware that PR #3268 fixes integration and some unit tests so you should check that branch (is about to be merged tough)

mitsujutsu:
@pablitovicente Ok I am currently working on functional tests so that should not be an issue.

PR #3248, opened 2019-04-02
Convert error strings to objects - Closes #3189 

### LINKED ISSUE #3189: All of the Error needs to converted to error object
### Expected behavior
String should not be used as an error

### Actual behavior
String is used as error through out the code

1.7-

mitsujutsu:
### What was the problem?

Application was inconsistently throwing strings instead of Error objects in some areas of the code.

### How did I fix it?

Converted thrown strings to Error Objects and related changes.

API related error changes to be handled in separate PR. 

### How to test it?

npm run mocha:unit

### Review checklist

* The PR resolves #3189 
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/chain/logic/block.js
+++ b/framework/src/modules/chain/logic/block.js
@@ -234,10 +234,12 @@ class Block {
 		});
 		const report = this.scope.schema.validate(block, Block.prototype.schema);
 		if (!report) {
-			throw `Failed to validate block schema: ${this.scope.schema
-				.getLastErrors()
-				.map(err => err.message)
-				.join(', ')}`;
+			throw new Error(
+				`Failed to validate block schema: ${this.scope.schema
+					.getLastErrors()
+					.map(err => err.message)
+					.join(', ')}`
+			);
 		}
 		try {
 			const {
@@ -251,9 +253,7 @@ class Block {
 				throw invalidTransactionResponse.errors;
 			}
 		} catch (errors) {
-			const error =
-				Array.isArray(errors) && errors.length > 0 ? errors[0] : errors;
-			throw error;
+			throw errors;
 		}
 
 		return block;
--- a/framework/src/modules/chain/logic/block.js
+++ b/framework/src/modules/chain/logic/block.js
@@ -241,19 +241,14 @@ class Block {
 					.join(', ')}`
 			);
 		}
-		try {
-			const {
-				transactionsResponses,
-			} = modules.processTransactions.validateTransactions(block.transactions);
-			const invalidTransactionResponse = transactionsResponses.find(
-				transactionResponse =>
-					transactionResponse.status !== TransactionStatus.OK
-			);
-			if (invalidTransactionResponse) {
-				throw invalidTransactionResponse.errors;
-			}
-		} catch (errors) {
-			throw errors;
+		const {
+			transactionsResponses,
+		} = modules.processTransactions.validateTransactions(block.transactions);
+		const invalidTransactionResponse = transactionsResponses.find(
+			transactionResponse => transactionResponse.status !== TransactionStatus.OK
+		);
+		if (invalidTransactionResponse) {
+			throw invalidTransactionResponse.errors;
 		}
 
 		return block;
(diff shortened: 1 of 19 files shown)


framework/src/modules/chain/submodules/accounts.js
```diff
@@ -201,7 +201,7 @@ Accounts.prototype.setAccountAndGet = function(data, cb, tx) {
 
 	return taskPromise
 		.then(taskPromiseData => setImmediate(cb, null, taskPromiseData))
-		.catch(taskPromiseErr => setImmediate(cb, taskPromiseErr));
```
SargeKhan: Won't the error here already be an object?

framework/src/modules/chain/submodules/blocks/chain.js
```diff
@@ -488,7 +488,8 @@ __private.loadSecondLastBlockStep = function(secondLastBlockId, tx) {
 			{ id: secondLastBlockId },
 			(err, blocks) => {
 				if (err || !blocks.length) {
-					library.logger.error('Failed to get loadBlocksPart', err);
+					const error = Array.isArray(err) && err.length > 0 ? err[0] : err;
+					library.logger.error('Failed to get loadBlocksPart', error);
 					return setImmediate(reject, err || 'previousBlock is null');
```
SargeKhan: 'previousBlock is null' should be wrapped inside an error object as well. Right?

framework/src/modules/chain/logic/block.js
```diff
@@ -251,9 +253,7 @@ class Block {
 				throw invalidTransactionResponse.errors;
 			}
 		} catch (errors) {
-			const error =
-				Array.isArray(errors) && errors.length > 0 ? errors[0] : errors;
-			throw error;
+			throw errors;
```
pablitovicente: Are the items contained in `errors` Error instances?

mitsujutsu: Yes or more specifically `TransactionError` from elements

limiaspasdaniel:
Sorry for jumping in here. 

Please bear in mind that **Error** objects are not correctly logged in our current application state. 

For example: https://github.com/LiskHQ/lisk/blob/949b41e80eb30992385f90b5b59bf7006b388a73/framework/src/controller/controller.js#L294
The controller in this case needs to be prepared to receive objects instead of plain `strings`.
You can reproduce this problem by for example running an instance of the application and starting another one with the same port. You'll see how the following log shows:
`Caused error during cleanup - {}` demonstrating what I've stated above.
I don't know if this is included in this PR's scope but definitely needs to be addressed perhaps in another issue.

framework/src/modules/chain/logic/block.js
```diff
@@ -251,9 +253,7 @@ class Block {
 				throw invalidTransactionResponse.errors;
 			}
 		} catch (errors) {
```
shuse2: no need to catch here anymore

mitsujutsu:
> Is it only 18 files? I feel like there should be more...

There is few more in transport but I will change those for the next issue: #3152 as it is related.

mitsujutsu:
> Is it only 18 files? I feel like there should be more...

That is most. There are few left in transport but I will update it in next API Error PR since it is related.

> A quick search in the repository reveals that there are still some places where we are calling callback with error strings. For-example,
> 
> ```
> framework/test/mocha/unit/modules/chain/scripts/json_history.js|182 col 7| cb('Error occurred during change.');
> framework/test/mocha/unit/modules/chain/submodules/multisignatures.js|285 col 6| cb('Err', null);
> framework/test/mocha/unit/modules/chain/submodules/delegates.js|890 col 6| cb('generateDelegateList error');
> framework/test/mocha/common/utils/wait_for.js|65 col 12| return cb('Failed to load blockchain');
> framework/test/mocha/common/utils/wait_for.js|81 col 12| return cb('Server is not responding');
> framework/test/mocha/functional/ws/transport/handshake.js|76 col 11| return cb('Socket should not connect but it did');
> framework/src/modules/chain/submodules/loader.js|846 col 22| return waterCb('Failed to get random peer from network');
> framework/src/modules/chain/submodules/loader.js|887 col 22| return waterCb(`Failed to load blocks from: ${peer.string}`);
> ```
> 
> Please double check the codebase for any missed cases.

The ones left in `src` are for loader.js `waterCb`. It is an `async waterfall` in which the error message is logged at the end of the waterfall, therefore we can leave as a message. I will update the unit ones for consistency but I think they pass as is.

PR #3257, opened 2019-04-03
App doesn't start when using child process on mainnet - Closes #3256

### LINKED ISSUE #3256: App doesn't start when using child process on mainnet
### Expected behavior
Expect to load HTTP API as child process on mainnet when IPC is enabled

### Actual behavior
It crashes with error: `Error: spawn E2BIG`

### Steps to reproduce
Enable `ipc` in config file and run the application on mainnet

1.6

lsilvs:
### What was the problem?
Application not starting when trying to load HTTP API as child process on mainnet network

### How did I fix it?
I've changed the way the config and options are sent to the child process. Before it was using command line arguments and now is using the channel created between parent and child. I also added a todo comment to review the config and options to avoid sending unnecessary data to the child.

### How to test it?
Enable `ipc` in config file and run the application with mainnet network

### Review checklist

* The PR resolves #3256
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/controller/application.js
+++ b/framework/src/controller/application.js
@@ -119,20 +119,14 @@ class Application {
 		__private.modules.set(this, {});
 		__private.transactions.set(this, {});
 
-		// TODO: move this configuration to module especific config file
-		const childProcessModules = process.env.LISK_CHILD_PROCESS_MODULES
-			? process.env.LISK_CHILD_PROCESS_MODULES.split(',')
-			: ['httpApi'];
-
 		this.registerModule(ChainModule, {
 			genesisBlock: this.genesisBlock,
 			constants: this.constants,
-			loadAsChildProcess: childProcessModules.includes(ChainModule.alias),
 		});
 
 		this.registerModule(HttpAPIModule, {
 			constants: this.constants,
-			loadAsChildProcess: childProcessModules.includes(HttpAPIModule.alias),
+			loadAsChildProcess: true,
 		});
 	}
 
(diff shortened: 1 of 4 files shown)


framework/src/controller/child_process_loader.js
```diff
@@ -26,9 +25,13 @@ const loadModule = async () => {
 	channel.publish(`${moduleAlias}:loading:finished`);
 };
 
+process.on('message', data => {
```
ManuGowda: `process.on('message', {loadModule , config, moduleOptions} => {` ?

shuse2:
```suggestion
process.on('message', ({loadModule , config, moduleOptions}) => {
```

framework/src/controller/application.js
```diff
@@ -119,20 +119,14 @@ class Application {
...
-
 		this.registerModule(ChainModule, {
 			genesisBlock: this.genesisBlock,
 			constants: this.constants,
-			loadAsChildProcess: childProcessModules.includes(ChainModule.alias),
 		});
 
 		this.registerModule(HttpAPIModule, {
 			constants: this.constants,
-			loadAsChildProcess: childProcessModules.includes(HttpAPIModule.alias),
```
shuse2: shouldn't this be still optional?

lsilvs: For now it's optional by using `ipc` config in config file.
Next version it will use the module specific config for this.

PR #3261, opened 2019-04-04
Adapt network module to new applicaton state - Closes #3260

### LINKED ISSUE #3260: Adapt network module to use application state
### Expected behavior

Every module initiates the application state and subscribe to updates.

### Actual behavior

Network module uses the former system requesting it to chain module.

`1.6+`

diego-G:
### Review checklist

* The PR resolves #3260
* Linting passes
* Tests pass


framework/src/modules/network/network.js
```diff
@@ -37,26 +37,23 @@ module.exports = class Network {
...
+		const applicationState = await this.channel.invoke(
+			'lisk:getApplicationState'
+		);
 
 		const p2pConfig = {
 			...this.options,
 			nodeInfo: {
-				...initialNodeInfo,
+				...applicationState,
 				wsPort: this.options.nodeInfo.wsPort,
```
ishantiw: Do we need this anymore? it's coming from the application state already. Or, there is a chance that it could vary?

diego-G: Application state contains broadhash and height which are constantly being updated by chain module.

ishantiw: Oh, I was talking about `wsPort` that why we are explicitly assigning it from `this.options.nodeInfo.wsPort`. We are getting it from `applicationState` anyways so we can remove this line?

diego-G: @jondubois wants to overwrite it for now until config refactoring is done. We have some temporary config in here that overwrites the one used by the application. https://github.com/LiskHQ/lisk/blob/feature/create_network_module/framework/src/modules/network/defaults/index.js

ishantiw: Oh yeah so as of now we need it until config refactoring is done

PR #3275, opened 2019-04-05
Narrow parameter list for end user - Closes #3160

### LINKED ISSUE #3160: Narrow the list of supported by framework network parameters
### Expected behavior
For now, expose the following set of network parameters (old constants.js #3159), as configurable by the application-level:

- REWARDS
   - MILESTONES
   - OFFSET (Start rewards at first block of the second round)
   - DISTANCE (Distance between each milestone)
- MAX_TRANSACTIONS_PER_BLOCK
- EPOCH_TIME
- BLOCK_TIME (currently defined in slots.interval)
#### Optional 
- ACTIVE_DELEGATES
- MAX_VOTES_PER_ACCOUNT

Narrow the set of parameters, which is possible to set in the application-level in the initial release. Temporarily, throw the "Unsupported network parameter" error from the framework-level in case of an attempt to set the unsupported parameter. 

The set of supported network parameters will increase gradually with the next releases. 
 
### Actual behavior
It's possible to override all of the network parameters (old constants.js #3159). 

`*`

limiaspasdaniel:
### What was the problem?
The user was able to modify all the constants at the application level.
### How did I fix it?
By limiting the number of constants the user is able to modify at the application level as stated in #3160 
### How to test it?
Run all tests
### Review checklist

* The PR resolves #3160
* Linting passes
* Tests pass


framework/src/controller/application.js
```diff
@@ -46,6 +46,68 @@ const registerProcessHooks = app => {
 	process.once('exit', (error, code) => app.shutdown(code, error));
 };
 
+/* As for now we don't support this constants to be configured
+by the user. This will change in the future */
+const nonConfigurableConstants = {
```
shuse2: maybe put this value in `/controller/constants.js`?

framework/src/controller/schema/constants.js
```diff
@@ -3,177 +3,31 @@ module.exports = {
...
-					DAPP_DEPOSIT: '10000000',
-				},
-			},
-			MAX_PAYLOAD_LENGTH: {
-				type: 'integer',
-				min: 1,
-				default: 1024 * 1024,
-				description:
-					'Maximum transaction bytes length for 25 transactions in a single block',
-			},
```
michielmulders: Are we throwing away the validation rules we wrote here?

limiaspasdaniel: We don't need to validate as those values won't be changed by the end user. We will slowly move those values back to the original file when we remove hardcodings in the codebase

framework/src/controller/application.js
limiaspasdaniel: There is also `/controller/schema` constants which also contains default values... Wouldn't creating another file named constants also inside `/controller` be confusing?

limiaspasdaniel: Just discovered doing it however you say simplifies loading those constants in unit tests setup. This is actually why now they're failing

michielmulders:
@limiaspasdaniel conflicts!

framework/src/controller/schema/constants.js
```diff
@@ -3,177 +3,31 @@ module.exports = {
...
-			MAX_SHARED_TRANSACTIONS: {
-				type: 'integer',
+			BLOCK_TIME: {
+				type: 'number',
 				min: 1,
-				default: 100,
-				description:
-					'Maximum number of in-memory transactions/signatures shared across peers',
+				default: 10,
+				description: 'Slote time interval in seconds',
```
michielmulders: > Typo

framework/src/controller/application.js
```diff
@@ -6,6 +6,10 @@ const version = require('../version');
 const validator = require('./helpers/validator');
 const applicationSchema = require('./schema/application');
 const constantsSchema = require('./schema/constants');
+
+/* As for now we don't support this constants to be configured
+by the user. This will change in the future */
+const nonConfigurableConstants = require('./constants');
```
nazarhussain: Please rename the file to `non_configureable_constants.js`

nazarhussain: BTW if rest are `configureable constants` then these are configurations, not constants. So I suggest to move the rest of them to configurations and then later describe which are network configurations and which are node configurations.

framework/test/jest/specs/unit/controller/schema/__snapshots__/constants.spec.js.snap
```diff
@@ -412,70 +87,5 @@ Object {
...
-      "SIGNATURE",
-      "DELEGATE",
-      "VOTE",
-      "MULTI",
-      "DAPP",
-      "IN_TRANSFER",
-      "OUT_TRANSFER",
-    ],
-    "type": "object",
-  },
```
nazarhussain: Just thought that for the sake for more safety we should not remove the schema even though these are hard coded values now. But still this schema validation does not harm.

framework/src/controller/application.js
limiaspasdaniel: I agree with you but I also think that then `non_configurable_constants` are indeed constants and the rest as you said are `configurations`.

shuse2: I think constants are always non configurable...=0

limiaspasdaniel: Agreed. Then we could move `schema/constants.js` to `schema/config.js` and move and rename `non_configurable_constants` to `schema/constants.js`. But we still need to pass them to `global.constants` as it would be too much refactor for the scope of this PR

limiaspasdaniel: So `schema/config.js` would have what it has now, plus: 

```
module.exports = {
	constants: {
		id: 'constants',
		type: 'object',
		required: [
			'EPOCH_TIME',
			'BLOCK_TIME',
			'MAX_TRANSACTIONS_PER_BLOCK',
			'REWARDS',
		],
		properties: {
			EPOCH_TIME: {
				type: 'string',
				format: 'date-time',
				description:
					'Timestamp indicating the start of Lisk Core (`Date.toISOString()`)',
			},
			BLOCK_TIME: {
				type: 'number',
				min: 1,
				description: 'Slot time interval in seconds',
			},
			MAX_TRANSACTIONS_PER_BLOCK: {
				type: 'integer',
				min: 1,
				description: 'Maximum number of transactions allowed per block',
			},
			REWARDS: {
				$ref: 'rewards',
				default: {
					MILESTONES: [
						'500000000', // Initial Reward
						'400000000', // Milestone 1
						'300000000', // Milestone 2
						'200000000', // Milestone 3
						'100000000', // Milestone 4
					],
					OFFSET: 2160,
					DISTANCE: 3000000,
				},
			},
		},
		additionalProperties: false,
	},
	rewards: {
		id: 'rewards',
		type: 'object',
		required: ['MILESTONES', 'OFFSET', 'DISTANCE'],
		description: 'Object representing LSK rewards milestone',
		properties: {
			MILESTONES: {
				type: 'array',
				items: {
					type: 'string',
					format: 'amount',
				},
				default: [
					'500000000', // Initial Reward
					'400000000', // Milestone 1
					'300000000', // Milestone 2
					'200000000', // Milestone 3
					'100000000', // Milestone 4
				],
				description: 'Initial 5, and decreasing until 1',
			},
			OFFSET: {
				type: 'integer',
				min: 1,
				description: 'Start rewards at block (n)',
			},
			DISTANCE: {
				type: 'integer',
				min: 1,
				description: 'Distance between each milestone',
			},
		},
		additionalProperties: false,
	},
	default: {
		EPOCH_TIME: new Date(Date.UTC(2016, 4, 24, 17, 0, 0, 0)).toISOString(),
		BLOCK_TIME: 10,
		MAX_TRANSACTIONS_PER_BLOCK: 25,
		REWARDS: {
			MILESTONES: [
				'500000000', // Initial Reward
				'400000000', // Milestone 1
				'300000000', // Milestone 2
				'200000000', // Milestone 3
				'100000000', // Milestone 4
			],
			OFFSET: 2160, // Start rewards at first block of the second round
			DISTANCE: 3000000, // Distance between each milestone
		},
	},
};
```

Which are indeed meant to be configured by the user.

limiaspasdaniel: And `schema/constants.js` which have exactly the same content as the content now present on `non_configurable_constants.js`

limiaspasdaniel: @shuse2 @nazarhussain @michielmulders please let me know your position on this :)

michielmulders: Ok, for me the naming makes more sense now! 👍

shuse2: it looks good but does `rewards` need to be defined outside?

limiaspasdaniel: @shuse2 I can move it inside, it was outside to avoid nesting but it's the only reference we have

limiaspasdaniel:
@shuse2 when this PR becomes ready, merge it after https://github.com/LiskHQ/lisk/pull/3288 gets merged first.

PR #3277, opened 2019-04-05
Fix snapshotting validation and rewrite -s to --rebuild

### LINKED ISSUE #2843: Only accept integer range values for snapshotting rounds

michielmulders:
### What was the problem?
Wrong validation check for snapshotting (CLI params are always strings).
Rewrite `-s` option to `--rebuild` (no `-r` as that's reserved for configuring Redis).

### How did I fix it?
Updated validation.

### How to test it?
Run tests.
Run snapshotting: `NODE_ENV=test npm start -- -n testnet --rebuild 40`
Or with `-b`

### Review checklist

* Linting passes
* Tests pass


README.md
```diff
@@ -346,31 +346,31 @@ npm start -- [options]
...
+|                                      | LISK_FORGING_WHITELIST | forging.access.whiteList | Comma separated list of IPs to enable access to forging endpoints. Format `192.168.99.100,172.169.99.77`                                                                                                                                                                                                                                         |
+| <pre nowrap>--rebuild</pre>          |                        |                          | Rebuild the chain from Genesis block up to specified round number, must be a positive integer equal to or greater than `0`. When `0` is passed, this corresponds to the inclusion of all rounds. Any other number equals to its corresponding round. Bear in mind this mode disables all the network features of the node to ensure reliability. |
```
diego-G: I'm not so sure about the wording here. I'd suggest `--validate -v`. What we do it's to process again and therefore validate every block until the typed round.

michielmulders: So, rename the functions as well: `createSnapshot` -> `rebuildChain`.
Your comment about documentation wording is correct. We both validate and rebuild up to the specified number.

4miners:
Files like `snapshot.json`, `pm2-snapshot.json` should also be renamed.

framework/src/modules/chain/submodules/loader.js
```diff
@@ -741,36 +741,31 @@ __private.validateOwnChain = cb => {
  * Snapshot creation - performs rebuild of accounts states from blockchain data
```
diego-G: Please, make another round to check for `snapshot` matches across the codebase. Careful with the concept of round snapshot which is different is meant to stay.

michielmulders: Obsolete and removed (with test), thanks!

PR #3287, opened 2019-04-08
Fix functional and unit tests - Closes #3152

### LINKED ISSUE #3152: Change the API response to return an array of errors

mitsujutsu:
### What was the problem?

Functional and unit tests failing after implementation of `improving transactions processing efficiency`.

API endpoints must be updated to return array of errors from `elements`.

### How did I fix it?

- Update `POST` API endpoints for transaction and signature

- Added transaction timestamp validation to `process_transactions` sub-module

- `VoteTransaction` in elements also checks for null `username` value

- Fix related unit tests

- Fix related functional tests

### How to test it?

Run functional tests related to API.

### Review checklist

* The PR resolves #3152 #3297 #3377 
* Linting passes
* Tests pass


mitsujutsu:
@pablitovicente Can you try building and linking elements and then run the functional tests again? I will check the signature issue and the unit tests.

pablitovicente:
> @pablitovicente Can you try building and linking elements and then run the functional tests again? I will check the signature issue and the unit tests.

Elements is already linked in the repository maybe try cleaning and building again in your end?

mitsujutsu:
It seems things broke after the merge from development. `npm run build` fails with `lisk-transaction-pool` on `feature/improve_transactions_processing_efficiency`

framework/src/modules/chain/submodules/blocks/process.js
```diff
@@ -348,6 +354,7 @@ Process.prototype.getCommonBlock = function(peer, height, cb) {
 			if (comparisonFailed && modules.transport.poorConsensus()) {
 				return modules.blocks.chain.recoverChain(cb);
 			}
+			if (err) err = new Error(err);
```
shuse2: prefer
```
if (err) {
  return setImmediate(cb, new Error(err));
}
```

framework/src/modules/http_api/controllers/signatures.js
```diff
@@ -55,20 +55,31 @@ SignaturesController.postSignature = async function(context, next) {
...
-
-		if (processingError.test(data.message)) {
-			error = new ApiError(data.message, apiCodes.PROCESSING_ERROR);
-		} else if (badRequestBodyError.test(data.message)) {
-			error = new ApiError(data.message, apiCodes.BAD_REQUEST);
+		if (data.code === apiCodes.PROCESSING_ERROR) {
+			error = new ApiError(
+				'Error processing signature',
+				apiCodes.PROCESSING_ERROR,
+				data.error
```
shuse2: isn't this `data.error` a signle Error object?
ApiError should accept array of error as third parameter?

same for below

framework/src/modules/http_api/helpers/error_handlers.js
```diff
@@ -14,30 +14,32 @@
...
+ * @param {array | object} e
  * @see Parent: {@link helpers}
  * @todo Add description for the params
  */
 // TODO: Move this functionality to logger component
-function convertErrorsToString(errors) {
-	if (Array.isArray(errors) && errors.length > 0) {
-		return errors
-			.filter(e => e instanceof Error)
+function convertErrorsToString(e) {
```
shuse2: prefer to use `errors`, but any reason changed?

framework/src/modules/http_api/schema/swagger.yml
```diff
@@ -2434,6 +2434,43 @@ definitions:
...
+        description: Expected value
+        type: 
+            - string
+            - number
+      actual: 
+        description: Actual value
+        type: 
+            - string
+            - number
+      jse_shortmsg:
```
shuse2: Let's keep in mind to remove this property

framework/src/modules/chain/submodules/blocks/process.js
pablitovicente: @shuse2 this case is special and needs to stay as it is because err can be null; changing this might require refactoring older methods.

shuse2: yes, im just saying about the format =)

framework/src/modules/http_api/controllers/signatures.js
mitsujutsu: It has been changed to an `array` from a single object, so user can see more detailed errors from `elements`.

framework/src/modules/http_api/helpers/error_handlers.js
mitsujutsu: this is a temporary file that I think should be moved to `logger`. the reason I changed is because it can also refer to a single error or string.

framework/src/modules/http_api/controllers/signatures.js
shuse2: would it be possible to change the property name to `data.errors`?

PR #3288, opened 2019-04-08
Simplify the default value generation for all configurations - Closes #3273

### LINKED ISSUE #3273: Simplify the default value generation for all configurations
### Expected behavior
The default values for all configurations must be specified collectively for each module and component as one object. 

### Actual behavior

Currently due to nature of `ajv` default keyword, we have to repeat the default values for each nested level properties and objects. 

https://github.com/LiskHQ/lisk/blob/4b936d244e33691ee2b79b2e7f11df95ed742e73/framework/src/modules/http_api/defaults/config.js#L4-L8

https://github.com/LiskHQ/lisk/blob/4b936d244e33691ee2b79b2e7f11df95ed742e73/framework/src/modules/http_api/defaults/config.js#L207-L208

We need to implement a custom logic for default value generation and not use builtin `default` keyword from `ajv`.

https://github.com/LiskHQ/lisk/blob/4b936d244e33691ee2b79b2e7f11df95ed742e73/framework/src/controller/helpers/validator/index.js#L33

1.7.0

nazarhussain:
### What was the problem?
There is redundant default values for nested objects due to nature of `ajv` implementation of default keyword. 

### How did I fix it?

Added custom logic to handle default values any configuration object. 

### How to test it?

Run all tests

### Review checklist

* The PR resolves #3273
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/components/cache/defaults/config.js
+++ b/framework/src/components/cache/defaults/config.js
@@ -3,13 +3,11 @@ const defaultConfig = {
 	properties: {
 		enabled: {
 			type: 'boolean',
-			default: false,
 			env: 'LISK_CACHE_ENABLED',
 		},
 		host: {
 			type: 'string',
 			format: 'ipOrFQDN',
-			default: '127.0.0.1',
 			env: 'LISK_REDIS_HOST',
 			arg: '-r,--redis',
 		},
@@ -17,23 +15,27 @@ const defaultConfig = {
 			type: 'integer',
 			minimum: 1,
 			maximum: 65535,
-			default: 6380,
 			env: 'LISK_REDIS_PORT',
 		},
 		db: {
 			type: 'integer',
 			minimum: 0,
 			maximum: 15,
-			default: 0,
 			env: 'LISK_REDIS_DB_NAME',
 		},
 		password: {
 			type: ['string', 'null'],
-			default: null,
 			env: 'LISK_REDIS_DB_PASSWORD',
 		},
 	},
 	required: ['enabled', 'host', 'port', 'db', 'password'],
+	default: {
+		enabled: true,
+		host: '127.0.0.1',
+		port: 6380,
+		db: 0,
+		password: null,
+	},
 };
 
 module.exports = defaultConfig;
--- a/framework/src/components/cache/defaults/config.js
+++ b/framework/src/components/cache/defaults/config.js
@@ -30,7 +30,7 @@ const defaultConfig = {
 	},
 	required: ['enabled', 'host', 'port', 'db', 'password'],
 	default: {
-		enabled: true,
+		enabled: false,
 		host: '127.0.0.1',
 		port: 6380,
 		db: 0,
--- a/framework/src/controller/schema/constants.js
+++ b/framework/src/controller/schema/constants.js
@@ -30,13 +30,11 @@ module.exports = {
 				type: 'number',
 				format: 'oddInteger',
 				min: 1,
-				default: 101,
 				description: 'The default number of delegates allowed to forge a block',
 			},
 			BLOCK_SLOT_WINDOW: {
 				type: 'integer',
 				min: 1,
-				default: 5,
 				description: 'The default number of previous blocks to keep in memory',
 			},
 			ADDITIONAL_DATA: {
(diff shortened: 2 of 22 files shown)


shuse2:
@nazarhussain Doesn't this go against the JSON schema v7?

nazarhussain:
No its not against JSON schema standard. We just removed the redundant default values and improved the implementation of default from root key.

shuse2:
@nazarhussain I couldn't find the reference 
I wasn't sure this reference applies to only object or can be in root level too
https://json-schema.org/understanding-json-schema/reference/generic.html

framework/src/components/cache/defaults/config.js
```diff
@@ -3,37 +3,39 @@ const defaultConfig = {
...
 		},
 		password: {
 			type: ['string', 'null'],
-			default: null,
 			env: 'LISK_REDIS_DB_PASSWORD',
 		},
 	},
 	required: ['enabled', 'host', 'port', 'db', 'password'],
+	default: {
+		enabled: true,
```
michielmulders: This was `false` before (cache enabled)?
Did we also not discuss to change `LISK_CACHE_ENABLED` to `LISK_ENABLE_CACHE` as it seems like a state instead a switch we can turn on/off?

framework/src/controller/schema/constants.js
```diff
@@ -133,330 +160,253 @@ module.exports = {
...
+					'VOTE',
+					'MULTI',
+					'DAPP',
+					'IN_TRANSFER',
+					'OUT_TRANSFER',
+				],
+				properties: {
+					SEND: {
+						type: 'integer',
+						enum: [0],
```
michielmulders: As this will only have one value, I suggest looking at `const` for JSON schema: http://json-schema.org/latest/json-schema-validation.html#rfc.section.6.1.3

`const: 0`

michielmulders: Same applies to other tx types.

nazarhussain:
@shuse2 Here is the reference 

http://json-schema.org/latest/json-schema-validation.html#rfc.section.10.2

framework/src/controller/schema/constants.js
nazarhussain: Yes current version of `ajv` introduced the support of `const` keyword. We can use it here.

framework/src/components/cache/defaults/config.js
nazarhussain: @shuse2 If we want to change variable names this release is the right candidate for it. The change on our side is very easy, need to communicate with devops team if we do it.

shuse2: yes we should make it clean! =)

framework/src/controller/schema/constants.js
nazarhussain: @limiaspasdaniel May be this is related to changes you are doing.

framework/src/components/cache/defaults/config.js
nazarhussain: @michielmulders Please create a separate issue for refactoring the variable name. Mention this variable in the issue and then we ask everyone else if they want to suggest another improvement.

framework/src/components/cache/defaults/config.js
```diff
@@ -3,37 +3,39 @@ const defaultConfig = {
...
-			default: null,
 			env: 'LISK_REDIS_DB_PASSWORD',
 		},
 	},
 	required: ['enabled', 'host', 'port', 'db', 'password'],
+	default: {
+		enabled: false,
+		host: '127.0.0.1',
+		port: 6380,
+		db: 0,
```
shuse2: redis db name is `integer`?

framework/src/components/cache/defaults/config.js
```diff
@@ -3,37 +3,39 @@ const defaultConfig = {
...
+		password: null,
```
shuse2: does this have to be defined? or just backward compatibility?
Ideally it can be optional right?

framework/src/components/cache/defaults/config.js
nazarhussain: Yes redis dbs are always numbers starting from zero.

framework/src/components/cache/defaults/config.js
nazarhussain: I just kept it exactly the same as before to avoid any unseen issue. 

https://github.com/LiskHQ/lisk/blob/0d2df7d376030122faae763a936e5a1a76f066bf/config/default/config.json#L27

It depends on redis library implementation how it treat this value. I believe it's just omitted if null.

Issue #2780, opened 2019-01-23
Move Lisk-specific transaction types to application-level


nazarhussain:
### Expected behavior
The `chain` module should only have standard transactions required to run the DPOS system, register the second signature and multisignature accounts:
- type 0
- type 1
- type 2
- type 3
- type 4

All other transaction types i.e. "dapp registration", "dapp in-transfer", dapp out-transfer" are specific only for the Lisk blockchain application platform, therefore should be elevated to the application-level (`lisk/src`). 
- type 5
- type 6
- type 7

Unit Test needs to be implemented, too

### Actual behavior
Currently, all transaction types are part of the `chain` module.

MaciejBaj:
Re-opening, the transactions type 5,6,7 should be in the application-level

shuse2:
DappRegistration needs to be duplicated for now to avoid major release

PR #3294, opened 2019-04-09
Implement design pattern for protocol change - Closes #3251

### LINKED ISSUE #3251: Implement design pattern for protocol change
### Expected behavior
Transaction types 6 (dapp in-transfer) and 7 (dapp out-transfer) to be disabled using `matcher` pattern.

The `matcher` pattern is like a toggle `on/off` for a transaction based on a custom condition. Helpful while introducing new or disabling existing transaction types on different, specified, chain heights.

Registering new transaction using the `matcher` pattern should look like:

`registerTransaction(DappInTransfer, matcher: height => height < 6901027)`

### Actual behavior
Currently `dapp in-transfer` and `dapp out-transfer` are hard-coded disabled using exceptions approach.

```
	precedent: {
		disableDappTransfer: 6901027, // Disable Dapp Transfer at this block height
	},
```

1.7

lsilvs:
### Review checklist

* The PR resolves #3251
* Linting passes
* Tests pass


framework/src/modules/chain/logic/init_transaction.js
```diff
@@ -102,7 +102,8 @@ const outTransferAsset = raw => {
 };
 
 class Transaction {
-	constructor(transactions = {}) {
+	constructor({ registeredTransactions = {}, applicationState = {} }) {
+		this.state = applicationState;
```
shuse2: I think applicationState, which are created by the module itself shouldn't be used in the module itself.
it should be using the state that the module has (which should be always correct)
It will create the complexity in the timing management and race condition

lsilvs: Ok. Using `lastBlock` now but I had to use `bindModules` inside initTransaction =\

lisk/src/index.js
```diff
@@ -97,10 +97,21 @@ try {
 	const { TRANSACTION_TYPES } = constants;
 
 	app.registerTransaction(TRANSACTION_TYPES.DAPP, DappTransaction);
-	app.registerTransaction(TRANSACTION_TYPES.IN_TRANSFER, InTransferTransaction);
+	app.registerTransaction(
+		TRANSACTION_TYPES.IN_TRANSFER,
+		InTransferTransaction,
+		{
+			isAllowedAt: context =>
```
MaciejBaj: matcher

lisk/src/index.js
```diff
@@ -97,10 +97,21 @@ try {
...
+				context.blockHeight < config.exceptions.precedent.disableDappTransfer,
+		}
+	);
 	app.registerTransaction(
 		TRANSACTION_TYPES.OUT_TRANSFER,
-		OutTransferTransaction
+		OutTransferTransaction,
+		{
+			isAllowedAt: context =>
```
MaciejBaj: matcher

framework/src/modules/chain/submodules/blocks/process.js
```diff
@@ -114,9 +113,7 @@ __private.receiveBlock = function(block, cb) {
  */
 __private.receiveForkOne = function(block, lastBlock, cb) {
 	let tmp_block = _.clone(block);
-	tmp_block.transactions = tmp_block.transactions.map(aTransaction =>
-		library.logic.initTransaction.jsonRead(aTransaction)
-	);
+	tmp_block.transactions = library.logic.initTransaction.fromBlock(tmp_block);
```
shuse2: not related to this PR, but can we change it to camelCase?

lisk/src/index.js
```diff
@@ -16,19 +14,28 @@ try {
...
 	const genesisBlock = require(`../config/${NETWORK}/genesis_block`);
 
-	const constants = {
-		TRANSACTION_TYPES: constantsSchema.default.TRANSACTION_TYPES,
-	};
-
 	const app = new Application(genesisBlock, config);
 
-	const { TRANSACTION_TYPES } = constants;
+	const { TRANSACTION_TYPES } = app.constants;
```
shuse2:
```suggestion
	const { constants: { TRANSACTION_TYPES } } = app;
```

framework/src/modules/chain/logic/init_transaction.js
```diff
@@ -101,30 +117,54 @@ const outTransferAsset = raw => {
...
 				Number(transactionType),
-				transactions[transactionType]
+				registeredTransactions[transactionType]
 			);
 		});
 	}
 
+	fromBlock(block) {
+		if (!block.transactions) {
+			return undefined;
```
shuse2: shouldnt it return empty array here?

framework/src/modules/chain/submodules/process_transactions.js
```diff
@@ -236,6 +263,75 @@ class ProcessTransactions {
...
+			return {
+				transactionsResponses: [...failedResponses, ...successfulResponses],
+			};
+		};
+	}
+
+	/**
+	 * Get current state from modules.blocks.lastBlock
+	 */
+	static _getCurrentContext() {
```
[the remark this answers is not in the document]
lsilvs: We've changed to `static` because `this` looses context when `checkAllowedTransactions` is used with `composeProcessTransactionSteps`

framework/src/modules/chain/submodules/process_transactions.js
```diff
@@ -236,6 +263,75 @@ class ProcessTransactions {
...
+					if (index === 0) {
+						// previousValue === transactions argument in the first iteration
+						// First iteration includes raw transaction objects instead of formatted responses.
+						return fn(previousValue);
+					}
+
+					previousValue = await previousValue;
+
+					// Keep track of transactions that failed in the current step
+					failedResponses = [
```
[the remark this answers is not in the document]
lsilvs: `failedResponses` is an accumulator, that's why is being reassigned.

framework/src/modules/chain/submodules/process_transactions.js
```diff
@@ -155,6 +155,33 @@ class ProcessTransactions {
...
 	}
 
+	// eslint-disable-next-line class-methods-use-this
+	checkAllowedTransactions(transactions, context) {
+		return {
+			transactionsResponses: transactions.map(transaction => {
+				const allowed =
+					!transaction.matcher ||
+					transaction.matcher(
+						context || ProcessTransactions._getCurrentContext()
```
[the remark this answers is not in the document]
lsilvs: Same as https://github.com/LiskHQ/lisk-sdk/pull/3294#discussion_r278834468

framework/src/modules/chain/submodules/blocks/verify.js
```diff
@@ -110,6 +118,16 @@ __private.checkTransactions = async (transactions, checkExists) => {
 			!checkTransactionExceptions.checkIfTransactionIsInert(transaction)
 	);
 
+	const nonAllowedTxResponses = modules.processTransactions
+		.checkAllowedTransactions(nonInertTransactions, context)
+		.transactionsResponses.find(
```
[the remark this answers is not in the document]
lsilvs: It's using `find` to avoid using `response[0].errors` next

framework/src/modules/chain/logic/transaction_pool.js
```diff
@@ -19,6 +19,7 @@ const {
 	Status: TransactionStatus,
 	TransactionError,
 } = require('@liskhq/lisk-transactions');
+const processTransactionLogic = require('../logic/process_transaction');
```
pablitovicente:
```suggestion
const { composeProcessTransactionSteps } = require('../logic/process_transaction');
```

and then adjust usages below

lsilvs: If I do this way I can't `spy` it during the tests.
I know it's not a good reasoning but there is no easy solution =|

framework/src/modules/chain/logic/init_transaction.js
lsilvs: it will return empty array if `block.transactions` is an empty array. But if there no `transactions` property I reckon is better to return `undefined`

framework/src/modules/chain/logic/init_transaction.js
```diff
@@ -163,25 +203,7 @@ class Transaction {
...
-
-	jsonRead(rawTx) {
-		const TransactionClass = this.transactionClassMap.get(rawTx.type);
-		if (!TransactionClass) {
-			throw new Error('Transaction type not found.');
-		}
-		return new TransactionClass(rawTx);
+		return this.fromJson(_.omitBy(transactionJSON, _.isNull));
 	}
 }
```
nazarhussain: This file export a class `Transaction` but the file name `init_transaction` suggest it do some actions.

framework/src/controller/application.js
```diff
@@ -224,6 +224,10 @@ class Application {
 		);
 		assert(Transaction, 'Transaction implementation is required');
 
+		if (options.matcher) {
+			Transaction.matcher = options.matcher;
+		}
```
nazarhussain: I would suggest this block to be changed as: 

```
const defaultMatcher = scope => scope.transaction.type === transactionType;

Transaction.matcher = options.matcher || Transaction.matcher || defaultMatcher; 
```

This way it will be more flexible for users to specify the matcher, first just as type, second as static property of class itself and on third level as custom matcher in options.

framework/src/controller/application.js
```diff
@@ -224,6 +224,10 @@ class Application {
...
+
 		const transactions = this.getTransactions();
 		transactions[transactionType] = Object.freeze(Transaction);
```
nazarhussain: We are freezing couple of stuff right before we start the application. I would suggest to move this logic to that part. 

https://github.com/LiskHQ/lisk-sdk/blob/bb52221265d0f053109eb864b6593f9b389f6a34/framework/src/controller/application.js#L282-L284

framework/src/modules/chain/logic/init_transaction.js
```diff
@@ -101,30 +117,54 @@ const outTransferAsset = raw => {
...
+
+	fromJson(rawTx) {
+		const TransactionClass = this.transactionClassMap.get(rawTx.type);
+
+		if (!TransactionClass) {
+			throw new Error('Transaction type not found.');
+		}
+
+		const transaction = new TransactionClass(rawTx);
+		transaction.matcher = TransactionClass.matcher;
```
nazarhussain: Do we really need to assign class property to instance, would it not come along with `new TransactionClass`, it not then we should revisit the logic where we assign `matcher` to the class itself.

framework/src/modules/chain/logic/process_transaction.js
```diff
@@ -0,0 +1,65 @@
...
+
+const { Status: TransactionStatus } = require('@liskhq/lisk-transactions');
+
+/**
+ * Executes each step from left to right and pipes the transactions that succeed to the next
+ * step. Finally collects all responses and formats them accordingly.
+ * @param steps
+ * @returns {function(*=): {transactionsResponses: *[]}}
+ */
+const composeProcessTransactionSteps = (...steps) => async transactions => {
```
nazarhussain: This name does not match with its usage. You are not using it just to compose the process transactions steps, but verify steps as well. So better may be just `composeTransactionSteps`

framework/src/modules/chain/submodules/blocks/verify.js
```diff
@@ -110,6 +118,16 @@ __private.checkTransactions = async (transactions, checkExists) => {
 			!checkTransactionExceptions.checkIfTransactionIsInert(transaction)
 	);
 
+	const nonAllowedTxResponses = modules.processTransactions
```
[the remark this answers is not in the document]
lsilvs: `find` will return `undefined` if it can't find. That's why I can't use destructuring here.

framework/src/modules/chain/logic/init_transaction.js
lsilvs: There is an issue for that already.

REF.: https://github.com/LiskHQ/lisk-sdk/pull/3440#discussion_r278441522

framework/src/controller/application.js
lsilvs: The `scope` only contains the block basic informations ({ blockVersion, blockHeight, blockTimestamp })

framework/src/modules/chain/submodules/process_transactions.js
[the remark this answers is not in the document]
shuse2: this is not used in this function though...?
Also, to maintain the scope, you can use lambda function

framework/src/controller/application.js
nazarhussain: It should contain the raw JSON of the transaction as well, so user can have flexible matcher based on transaction content as well.

framework/src/modules/chain/logic/init_transaction.js
lsilvs: No, the instance won't have access to `matcher` unless it's assigned like it's done

framework/src/controller/application.js
lsilvs: I agree that it would concentrate similar actions in one place but it would add an unnecessary step imo.

framework/src/controller/application.js
shuse2: transaction itself shouldn't  be used for matcher.
Only thing that can change the blockchain protocol should be block

framework/src/modules/chain/logic/init_transaction.js
shuse2: I feel it's safer to return empty array, because if we try to loop through, it wont throw an error

framework/src/modules/chain/submodules/process_transactions.js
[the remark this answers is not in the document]
lsilvs: I mean, when I try to use `this._getCurrentContext()` inside `checkAllowedTransactions` when using `composeTransactionSteps`, the `this` looses context.

framework/src/modules/chain/submodules/process_transactions.js
```diff
@@ -236,6 +263,75 @@ class ProcessTransactions {
...
+					return fn(
+						transactions.filter(transaction =>
+							previousValue.transactionsResponses
+								.filter(response => response.status === TransactionStatus.OK)
+								.map(transactionResponse => transactionResponse.id)
+								.includes(transaction.id)
+						)
+					);
+				},
+				transactions
```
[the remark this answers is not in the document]
lsilvs: I agree this function can be improved but I would rather do this in the next iteration.

framework/src/modules/chain/logic/init_transaction.js
shuse2: above it's the difference between static function and instance method

framework/src/modules/chain/submodules/process_transactions.js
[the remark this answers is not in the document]
shuse2: ah i see. I think this usage of `static` is wrong.
`this._getCurrentContext()` should not be allowed since it is not instance method.
It should be used as `ProcessTransaction._getCurrentContext()`
Im not sure if it applies to this case since the syntax is invalid, but in general if it looses the context, we can use lambda to insert or bind the context

framework/src/controller/application.js
nazarhussain: We may introduce new version of transactions, or transaction processing itself can be different for any particular transactions (as in our case exceptions), so passing the transaction JSON to the scope of matcher is very useful. 

At the end matcher just need to resolve the transaction class, no matter through which data it should do it. It does not impact the processing of transaction. The more flexible user facing interface could, the more use cases can be achieved by community (which we don't know right now).

shuse2: I think we shouldnt introduce something that we know it won't work...
one progress of blockchain is only happen by a block, and protocol change can only happen in the block.
Therefore, we should not introduce this at this point

framework/src/modules/chain/logic/init_transaction.js
nazarhussain: As I earlier said, if that's not the case we should update Transaction class in a way that whenever and wherever in the code I initiate the object I always get the matcher along. 

Where we are defining 

```js
Transaction.matcher = matcher;
```

Change it to 

```js
Object.defineProperty(TransactionClass.prototype, 'matcher', {get: ()=> matcher});
```

The idea is once you setup the class once, you don't need to remember to copy data attributes to instances manually.

framework/src/controller/application.js
nazarhussain: The matcher can't just be used for protocol change, it can be used to handle exceptions and other more generalised cases as well. And its not true that passing transaction JSON itself will break any thing. 

So I don't clearly agree to 

> I think we shouldnt introduce something that we know it won't work...

There is no reason that it would not work.

shuse2: Exception handling might happen in the similar way, but it will not be in this matcher property.
This issue is for design pattern for protocol change, and for this case, passing transaction is unnecessary and will not work because you cannot change protocol in the middle of a block

nazarhussain: In short, One use case of `matcher` is to handle protocol changes. It's not restricted only to that use case. So keep it flexible let others to came up with use cases which we can't think of right now.

lisk/src/index.js
```diff
@@ -16,19 +14,30 @@ try {
...
 
-	const { TRANSACTION_TYPES } = constants;
+	const {
+		constants: { TRANSACTION_TYPES },
+	} = app;
 
 	app.registerTransaction(TRANSACTION_TYPES.DAPP, DappTransaction);
-	app.registerTransaction(TRANSACTION_TYPES.IN_TRANSFER, InTransferTransaction);
+	app.registerTransaction(
+		TRANSACTION_TYPES.IN_TRANSFER,
```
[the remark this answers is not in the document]
lsilvs: This can be done along with https://github.com/LiskHQ/lisk-sdk/issues/2780

framework/src/modules/chain/submodules/process_transactions.js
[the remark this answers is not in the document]
lsilvs: Binding will required a big code refactoring and will break the tests. Would be better to do this in another issue

PR #3312, opened 2019-04-12
Update broadcaster to use network module - Close #2878

### LINKED ISSUE #2878: Update broadcaster to use network module and fix the tests
### Description
Update the broadcaster.js in chain module to use new network module created.
Also, it should remove the unused functions, and fix/remove tests related to the broadcaster.

### Dependent issue
#2874

diego-G:
### Review checklist

* The PR resolves #2878
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/chain/logic/broadcaster.js
+++ b/framework/src/modules/chain/logic/broadcaster.js
@@ -40,7 +40,7 @@ let library;
  * @todo Add description for the params
  */
 class Broadcaster {
-	constructor(broadcasts, force, peers, transaction, logger) {
+	constructor(broadcasts, force, peers, transaction, logger, channel) {
 		library = {
 			logger,
 			logic: {
@@ -72,6 +72,8 @@ class Broadcaster {
 				object: 'signature',
 			},
 		];
+
+		this.channel = channel;
 
 		if (broadcasts.active) {
 			jobsQueue.register(
@@ -117,6 +119,13 @@ class Broadcaster {
 			} else {
 				peers = params.peers.slice(0, params.limit);
 			}
+
+			// Broadcast using Elements P2P library via network module
+			await this.channel.invoke('network:send', {
+				event: options.api,
+				data: options.data,
+			});
+
 			library.logger.debug('Begin broadcast', options);
 			peers.forEach(peer => peer.rpc[options.api](options.data));
 			library.logger.debug('End broadcast');
--- a/framework/src/modules/chain/logic/broadcaster.js
+++ b/framework/src/modules/chain/logic/broadcaster.js
@@ -21,7 +21,6 @@ const { MAX_PEERS } = global.constants;
 
 let modules;
 let library;
-let self;
 
 /**
  * Main Broadcaster logic.
@@ -74,12 +73,10 @@ class Broadcaster {
 			},
 		];
 
-		self = this;
-
 		if (broadcasts.active) {
 			jobsQueue.register(
 				'broadcasterReleaseQueue',
-				this.releaseQueue,
+				async () => this.releaseQueue(),
 				this.config.broadcastInterval
 			);
 		} else {
@@ -96,7 +93,7 @@ class Broadcaster {
 	 * @returns {SetImmediate} error, peers
 	 * @todo Add description for the params
 	 */
-	async getPeers(params) {
+	getPeers(params) {
 		params.limit = params.limit || this.config.peerLimit;
 		const peers = library.logic.peers.listRandomConnected(params);
 		return peers;
@@ -116,7 +113,7 @@ class Broadcaster {
 		params.limit = params.limit || this.config.broadcastLimit;
 		try {
 			if (!params.peers) {
-				peers = await this.getPeers(params);
+				peers = this.getPeers(params);
 			} else {
 				peers = params.peers.slice(0, params.limit);
 			}
@@ -169,7 +166,7 @@ class Broadcaster {
 	 * @returns {Promise} null, boolean|undefined
 	 * @todo Add description for the params
 	 */
-	async filterQueue() {
+	filterQueue() {
 		library.logger.debug(`Broadcasts before filtering: ${this.queue.length}`);
 
 		this.queue = this.queue.filter(broadcast => {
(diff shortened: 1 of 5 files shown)


framework/src/modules/chain/logic/broadcaster.js
```diff
@@ -95,135 +92,97 @@ class Broadcaster {
...
- * @param {Object} options
- * @returns {Object[]} Queue private variable with new data
- * @todo Add description for the params
- */
-Broadcaster.prototype.enqueue = function(params, options) {
-	options.immediate = false;
-	return self.queue.push({ params, options });
-};
+			// Broadcast using Elements P2P library via network module
+			await this.channel.invoke('network:send', {
```
nazarhussain: I was expecting this and I don't like this approach. I already shared many times that avoid using `actions` and try more `events`. With these action approach the `chain` module is dependent on the `network` module. Although the original idea of modules were they must be pluggable. Chain module works as it is, we plugin network module and single node now have connection to others. 

So first I suggest the approach to remove all `peers` information from the chain module and then afterwards try more `events`. e.g. What you want to broadcast a *new block* create an event for it. Then use `channel.publish('chain:block:new', block);`, on network module listen to this event and then do what ever you want with it, publish or discard or what ever you want. 

If required we should discuss in detail but ty our self to be more align with *event driven* approach. That's not just for the this module, this practice should be followed fo all modules. @MaciejBaj @shuse2

shuse2: I agree to the even driven approach, butt this is specific case where the chain module needs to tell exactly what needs to be done.
Network module is not listening to any chain specific event now.

jondubois: In this case, I think that using `invoke` is better for sending data to the network because it allows the Network module to remain module-agnostic (ideally, the Network module should not need to know ahead of time what modules or custom events are available on the node).

That way node operators will be able to add any custom modules to their nodes and those modules will be able to send custom requests and messages to each other directly over the network.

The current principle is that modules which need to send messages over the network need to be aware of the Network module but not the other way around. This dependency makes sense IMO because you can't have a Dapp without some kind of network layer.

If testability is the concern (e.g. ability to run a node without the network for testing purposes), that should not be a problem; it's still possible mock the Network module for testing if necessary.

Also, the Network module can still be substituted for other kinds of network modules (e.g. with different transports, P2P protocols, network topology, etc...) provided that they adhere to the same simple interface. The module currently only exposes about 8 different methods and this interface should serve as a standard for any third-party who wants to create a custom network module.

nazarhussain: > Network module to remain module-agnostic 

That also applies to other modules as well, network module is no exception. 

> Network module can still be substituted for other kinds

Yes it can be substituted even in the event driven approach. 

> The current principle is that modules which need to send messages over the network need to be aware of the Network

That's where the basic different of understanding is. Network module itself should encapsulate all network related logic, other modules do the job of their concerns. I am saying this from very start, network module is not a network proxy. It should and must encapsulate all business logic related to network communication of node. Should listen to activities in system and perform network operations, and serve network related information to other modules. 

If you want to use it just as proxy then why not just drop in the library to every module and use network functionality as library.

jondubois: If we make the Network module consume events published by CustomModule, the Network module will depend on CustomModule. It's a loose dependency because we could later replace CustomModule with NewCustomModule (if NewCustomModule publishes the same events in the same format).

However, the dependency problem in this scenario is not about whether CustomModule can be replaced with another module (it can), the problem is that the Network module needs to know about the existence of CustomModule to begin with. The Network module should not know which modules exist on the node; that way any new custom third-party module can be added later without having to change the Network module at all.

> If you want to use it just as proxy then why not just drop in the library to every module and use network functionality as library.

- It would create a lot of duplicate connections (since each module would maintain its own separate network with its own peer connections).
- There needs to be a way to ban malicious peers from a central place on the node. Many modules can report violations but the final decision to ban a peer (and for how long) must be done in one place.
- There needs to be a single consistent P2P protocol for all modules to share. If each module ran a potentially different version of the `lisk-p2p` library, we would be forcing module authors to manage their own P2P network versioning. One benefit of having a single network protocol shared by many modules is interoperability between different modules at the network level. We can upgrade the P2P protocol (and make significant changes) without changing the Network module API; so any third-party custom module created by community members do not need to change when we upgrade the P2P protocol.

framework/src/modules/chain/logic/broadcaster.js
```diff
@@ -249,106 +208,106 @@ __private.filterQueue = function(cb) {
...
- */
-__private.squashQueue = function(broadcasts) {
-	const grouped = _.groupBy(broadcasts, broadcast => broadcast.options.api);
-	const squashed = [];
+		library.logger.debug(`Broadcasts after filtering: ${this.queue.length}`);
 
-	self.routes.forEach(route => {
-		if (Array.isArray(grouped[route.path])) {
-			const data = {};
+		return null;
```
jondubois: Can't we just do `return;` without the `null`?

framework/src/modules/chain/logic/broadcaster.js
```diff
@@ -249,106 +208,106 @@ __private.filterQueue = function(cb) {
...
+	 * @throws {Error}
+	 * @todo Add description for the params
+	 */
+	// eslint-disable-next-line class-methods-use-this
+	async releaseQueue() {
+		library.logger.info('Releasing enqueued broadcasts');
+
+		if (!this.queue.length) {
+			library.logger.info('Queue empty');
+			return null;
```
jondubois: Same here, why not just `return;`?

framework/src/modules/chain/submodules/transport.js
```diff
@@ -398,20 +399,20 @@ Transport.prototype.broadcastHeaders = cb => {
...
-	if (!broadcast) return;
+	if (!broadcast) return null;
 
 	// Check if we are free to broadcast
 	if (__private.broadcaster.maxRelays(block)) {
 		library.logger.debug(
 			'Transport->onBroadcastBlock: Aborted - max block relays exhausted'
 		);
-		return;
+		return null;
```
jondubois: As before why not `return;`?

framework/src/modules/chain/logic/broadcaster.js
nazarhussain: Module concept is clear, it must have its own special domain. To me proxying stuff around does not fit in the definition of a module. We will refactor this concept as separate issue https://github.com/LiskHQ/lisk-sdk/issues/3358

PR #3356, opened 2019-04-15
Fix bug sorting API /delegates/forgers - Closes #3304

### LINKED ISSUE #3304: Issues with Next Forging Turn (/api/delegates/forgers)
### Expected behavior
The api call /delegates/forgers should return always the correct order (ordered by 'nextSlot') of the current delegates list, to get a consistent value when a delegate will forge its next block

### Actual behavior
The api call /delegates/forgers returns inconsistent values. It is not ordered by 'nextSlot'. Well, sometimes it seems to be ordered correctly, but not always.

### Steps to reproduce
call /delegates/forgers multiple times and compare the returned list

1.6.0-rc.3

michielmulders:
### What was the problem?
Custom sorting passed was not correct for `/delegates/forgers` and returned inconsistent results when trying: `/delegates/forgers?limit=101`. 

### How did I fix it?
`.sort((prev, next) => (prev.nextSlot > next.nextSlot ? 1 : -1));`

### How to test it?
Run func tests for `http_api/controllers/delegates.js`.

### Review checklist

* The PR resolves #3304 
* Linting passes
* Tests pass

### DIFF
--- a/framework/test/mocha/functional/http/get/delegates.js
+++ b/framework/test/mocha/functional/http/get/delegates.js
@@ -642,12 +642,10 @@ describe('GET /delegates', () => {
 
 		it('using limit=101 should sort forgers ascending using nextSlot', async () => {
 			return forgersEndpoint.makeRequest({ limit: 101 }, 200).then(res => {
-				const forgers = _.clone(res.body.data);
-				expect(
-					forgers
-						.reverse()
-						.sort((prev, next) => (prev.nextSlot > next.nextSlot ? 1 : -1))
-				).to.be.eql(res.body.data);
+				const nextSlots = _(res.body.data)
+					.map('nextSlot')
+					.value();
+				expect(_.clone(nextSlots).sort()).to.be.eql(nextSlots);
 			});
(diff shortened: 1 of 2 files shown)


framework/test/mocha/functional/http/get/delegates.js
```diff
@@ -640,6 +640,17 @@ describe('GET /delegates', () => {
 			});
 		});
 
+		it('using limit=101 should sort forgers ascending using nextSlot', async () => {
+			return forgersEndpoint.makeRequest({ limit: 101 }, 200).then(res => {
+				const forgers = _.clone(res.body.data);
+				expect(
+					forgers
+						.reverse()
+						.sort((prev, next) => (prev.nextSlot > next.nextSlot ? 1 : -1))
```
shuse2: this doesn't seem to test what it should test.
If this logic is wrong, source and test will still say it is correct

michielmulders: Updated test to make it independent

PR #3362, opened 2019-04-15
Fix version reference for lisk-framework - Closes #3361

### LINKED ISSUE #3361: Wrong reference of the version from lisk folder
### Expected behavior
package.json should reference to existing package version

### Actual behavior
It is referencing `0.1.0` but it should be `0.1.0-alpha.0`

shuse2:
### What was the problem?
lisk was referencing non-existing version

### How did I fix it?
Update to use the exact same version as lisk-framework package.json

### How to test it?
```
npm run clean:node_modules
rm -rf node_modules
npm ci
```

### Review checklist

* The PR resolves #3361 
* Linting passes
* Tests pass


framework/.npmignore
```diff
@@ -0,0 +1,13 @@
...
+.lintstagedrc.json
+.nycrc
+.prettierrc.json
+.prettierignore
+cypress.json
+index.html
+Jenkinsfile*
+
+test/
+logs/
```
nazarhussain: We already have a `src` directory so why not we simplify the ignore file by

```
* 

!src/**/*
!README.md
```

This way every thing will be ignored except content in `src` directory and all README files.

PR #3365, opened 2019-04-16
Expose a sample Genesis Block from Framework

### LINKED ISSUE #3308: Expose a sample Genesis Block from Framework
### Expected behavior

SDK user must be able to start the blockchain application with the minimum code required. 

```js
const {Application, sampleGenesisBlock } = require('lisk-framework');

const app = new Application('myApp', sampleGenesisBlock);

app.run().then(console.info).catch(console.error);
```

To make this possible we should expose one genesis block as part of the framework, which everyone can start with easily. 

### Actual behavior

Right now the genesis block is the required data field to start any application. And we are not shipping any sample genesis block for user. So he had to go to one extra step to first create a genesis block to start any application. 

1.7

MaciejBaj:
### What was the problem?
For the development of Alpha SDK purposes, Framework should expose a genesis block. 

### How to test it?
Check if `const { GenesisBlockDevnet } = require('lisk-framework');`  works

### Review checklist

* The PR resolves #3308 
* Linting passes
* Tests pass


framework/src/index.js
```diff
@@ -1,4 +1,5 @@
 const Application = require('./controller/application');
+const GenesisBlockDevnet = require('./modules/chain/defaults/genesis_block_devnet');
```
nazarhussain: Its an object not a constructor function so we use camel case `genesis*`

nazarhussain: As of now we have no reference or usage of concept `network` int he framework. Network is just the part of the Lisk Core implementation. So I suggest we simply rename it to `sampleGenesisBlock` and don't opt Network naming convention in the framework (at least in current state).

framework/src/modules/chain/defaults/genesis_block_devnet.json
```diff
@@ -0,0 +1,1760 @@
+{
```
nazarhussain: As of the separation of concern, the genesis block concept in current implementation is part of the application/controller layer, and passed to every module as static data option. 

So we should keep that file to application level.

nazarhussain: Shouldn't we have a different genesis block been part of framework? Or we use devnet genesis block which comes from Lisk Core our network implementation.

framework/src/index.js
MaciejBaj: Done.

MaciejBaj: I moved `genesis_block_devnet` to `framework/samples` directory I exported all the samples from there. 

It is a Genesis Block for development purposes, so `devnet` genesis block is correct to me.

framework/src/modules/chain/defaults/genesis_block_devnet.json
MaciejBaj: It is a Genesis Block for development purposes, so devnet genesis block is correct to me.

README.md
```diff
@@ -66,9 +66,9 @@ To create a blockchain application, you need to provide an entry point of your a
 It is quite simple to have a working blockchain application, mirroring the configuration of the existing Lisk network. This can be done by copying the following three lines of code to your `index.js`:
 
 ```js
-const { Application, SampleGenesisBlock } = require('lisk-framework');
+const { Application, genesis_block_devnet } = require('lisk-framework');
```
nazarhussain:
```suggestion
const { Application, genesisBlockDevnet } = require('lisk-framework');
```

README.md
```diff
@@ -66,9 +66,9 @@ To create a blockchain application, you need to provide an entry point of your a
...
 
-const app = new Application('my-app', SampleGenesisBlock);
+const app = new Application('my-app', genesis_block_devnet);
```
nazarhussain:
```suggestion
const app = new Application('my-app', genesisBlockDevnet);
```

framework/samples/index.js
```diff
@@ -0,0 +1,5 @@
+const genesis_block_devnet = require('./genesis_block_devnet');
```
nazarhussain:
```suggestion
const genesisBlockDevnet = require('./genesis_block_devnet');
```

framework/samples/index.js
```diff
@@ -0,0 +1,5 @@
...
+
+module.exports = {
+	genesis_block_devnet,
```
nazarhussain:
```suggestion
	genesisBlockDevnet,
```

framework/src/index.js
```diff
@@ -9,6 +10,7 @@ const validator = require('./controller/helpers/validator');
 module.exports = {
 	Application,
 	version,
+	...samples,
```
nazarhussain: Would be nice if we keep the sample namespace, may be we have to export more sample data/fixtures in future as well as part of framework.

PR #3367, opened 2019-04-16
Fix framework reference - Closes #3366

### LINKED ISSUE #3366: Framework is referencing the framework keyword itself
### Expected behavior
Framework should not reference itself as framework

### Actual behavior
It's using framework in the path

shuse2:
### What was the problem?
Framework itself was referencing it's own

### How did I fix it?
Change to relative path

### How to test it?
try to run the `lisk` app

### Review checklist

* The PR resolves #3366
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/chain/submodules/blocks.js
+++ b/framework/src/modules/chain/submodules/blocks.js
@@ -20,7 +20,7 @@ const { BLOCK_RECEIPT_TIMEOUT, EPOCH_TIME } = global.constants;
 const {
 	CACHE_KEYS_BLOCKS,
 	CACHE_KEYS_TRANSACTIONS,
-} = require('../../../../../framework/src/components/cache');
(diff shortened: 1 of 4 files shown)


framework/src/modules/chain/submodules/blocks.js
```diff
@@ -20,7 +20,7 @@ const { BLOCK_RECEIPT_TIMEOUT, EPOCH_TIME } = global.constants;
 const {
 	CACHE_KEYS_BLOCKS,
 	CACHE_KEYS_TRANSACTIONS,
-} = require('../../../../../framework/src/components/cache');
+} = require('../../../components/cache');
```
nazarhussain: I am not sure if we keep these or not, as its easily accessible by `scope.components.cache.CACHE_KEYS_BLOCKS`. But probably refactor this separately.

PR #3369, opened 2019-04-16
Implement Unit tests for Chain module - Closes #3285

### LINKED ISSUE #3285: Implement Unit tests for Chain module
### Expected behavior
`init_steps` and `chain.js` are covered by unit tests
### Actual behavior
They are not.
`development`

yatki:
### What was the problem?
We didn't have unit tests for chain module
### How did I fix it?
I added unit tests for chain module
### Review checklist

* The PR resolves #3285 
* Linting passes
* Tests pass


framework/src/modules/chain/chain.js
```diff
@@ -122,15 +126,11 @@ module.exports = class Chain {
...
-			if (!this.options.genesisBlock) {
-				throw Error('Failed to assign nethash from genesis block');
-			}
-
 			// TODO: For socket cluster child process, should be removed with refactoring of network module
 			this.options.loggerConfig = loggerConfig;
 
 			const self = this;
-			const scope = {
+			self.scope = {
```
shuse2: why instead of deleting `self`, use self here?

yatki: I had to move `self.scope = scope` line to top in order to have a reference to scope variable to check if `createSocketCluster` is calling the right variable. I could use `this` instead but deleting `self` doesn't seem possible because `Sequence` instance is using chain instance in the onWarning method.

PR #3412, opened 2019-04-18
Add custom data column for accounts entity - Closes #3027

### LINKED ISSUE #3027: Allow custom transactions to add custom info to accounts
### Expected behavior
I expect custom transactions to be able to update custom data into accounts table.

Once developer can create their own custom transactions, they might need to update an account state with custom data in order to reflect the transaction purpose.

michielmulders:
### How did I fix it?
Added a data field to account for storing an account state.
Currently, not exposed via API (`/accounts` & `/delegates`).

### How to test it?
Run all tests.

### Review checklist

* The PR resolves #3027 
* Linting passes
* Tests pass


michielmulders:
Should this new `data` field be documented somewhere to keep track of this for if we decide to fully support this?

framework/src/modules/chain/components/storage/sql/migrations/updates/20190410112400_add_data_field_mem_accounts.sql
```diff
@@ -0,0 +1,26 @@
...
+
+
+/*
+  DESCRIPTION: Add data field for mem_accounts column
+
+  PARAMETERS: None
+*/
+
+-- Add data column to trs table as jsonb
+ALTER TABLE "mem_accounts" ADD COLUMN IF NOT EXISTS "data" jsonb;
```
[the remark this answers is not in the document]
michielmulders: The reason behind the change of the name to `data` is that it conflicts with the `asset` definition and would require multiple tests to be written + adds complexity for future code. It is better to distinct this field and give it a generic name `data`.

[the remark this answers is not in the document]
nazarhussain: But it tend to be used from custom transaction, as those transaction can store custom data to transaction as `asset` then why not same for account?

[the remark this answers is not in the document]
nazarhussain: These naming conventions are very important for the SDK user experience, so we should think a bit further on it. @MaciejBaj What's your opinion here?

framework/src/modules/http_api/controllers/accounts.js
```diff
@@ -46,6 +46,7 @@ function accountFormatter(totalSupply, account) {
 		'balance',
 		'u_balance',
 		'secondPublicKey',
+		'data',
```
lsilvs: Will the `data` be exposed in the API?

michielmulders: Nope, removed! Thanks!

framework/src/modules/chain/components/storage/sql/migrations/updates/20190410112400_add_data_field_mem_accounts.sql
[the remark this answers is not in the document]
michielmulders: `asset` it will be!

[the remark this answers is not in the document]
MaciejBaj: I prefer to have `asset` field for the custom fields of both accounts and transactions. 
`data` was already used in transactions for storing additional data in utf-8 chars.

michielmulders:
Reference build green: https://jenkins.lisk.io/blue/organizations/jenkins/lisk-sdk-framework/detail/3027-asset-property-accounts/1/pipeline/

framework/test/mocha/unit/modules/chain/components/storage/entities/account.js
```diff
@@ -260,6 +263,7 @@ describe('ChainAccount', () => {
 				productivity: 0,
 				votedDelegatesPublicKeys: null,
 				membersPublicKeys: null,
+				asset: {},
```
lsilvs: Can you add another test to make sure this field is populated, pls?

framework/src/components/storage/entities/account.js
```diff
@@ -381,12 +383,27 @@ class Account extends BaseEntity {
...
+					}
+					return account;
+				};
+
+				if (expectedResultCount === 1) {
+					return parseResponse(resp);
+				}
+
+				return resp.map(parseResponse);
+			});
```
nazarhussain: You don't need all this custom code here, its too expensive and unnecessary to map over hundred of result set. You already added `asset` only to `get_extended.sql` file. So if someone called in other case there will be no `asset` property in response and that's what intended.

framework/src/components/storage/entities/account.js
```diff
@@ -47,6 +47,7 @@ const sqlFiles = {
  * @property {string} rewards
  * @property {string} vote
  * @property {number} productivity
+ * @property {json} asset
```
nazarhussain: Asset must go to `ExtendedAccount` not the `BasicAccount` type definition.

framework/src/modules/chain/components/storage/entities/account.js
```diff
@@ -150,6 +142,37 @@ class ChainAccount extends AccountEntity {
...
+				{
+					...data,
+				},
+			];
+		}
+
+		accounts = accounts.map(account => {
+			const parsedAccount = _.defaults(account, defaultCreateValues);
+			parsedAccount.asset = parsedAccount.asset
+				? JSON.stringify(parsedAccount.asset)
```
nazarhussain: You don't need to stringify it manually, underlaying postgresql adapter must handle it. Please test it by just passing as an object.

michielmulders: Correct catch, thanks!

PR #3427, opened 2019-04-23
Validate overflowed string in transfer asset data

### LINKED ISSUE #3326: Ajv character length validation inconsistent with core z-schema
### Expected behavior

Validation of string length should be consistent between `lisk-elements` and `lisk-core`. 

### Actual behavior

Elements uses `ajv` for schema validation whereas core uses `z-schema`.

Checking the length of strings including unique characters such as `现` produces different results between these two libraries. For example: `5v1ug8nzc66kz7kva471i51rzmq3o59ceyjokd3ak4u8liqkoo65jv1k6y1pu1u现` in the transfer asset data check passes in `ajv` but not in `z-schema`.

### Steps to reproduce

Run in core:

` framework/test/mocha/functional/http/post/0.transfer.js`

Run in elements:

`test/0_transfer_transaction.ts`

mitsujutsu:
### What was the problem?

`elements/lisk-transactions` was not checking for overflowing string including special characters such as `现` for transer asset data.

### How did I fix it?

Added validator to check byte length for transfer asset data.

### How to test it?

`npm run test`

### Review checklist

* The PR resolves #3326 
* Linting passes
* Tests pass


elements/lisk-transactions/src/utils/validation/validation.ts
```diff
@@ -160,6 +161,9 @@ export const validateTransferAmount = (data: string) =>
 	isGreaterThanZero(new BigNum(data)) &&
 	!isGreaterThanMaxTransactionAmount(new BigNum(data));
 
+export const validateTransferData = (data: string): boolean =>
```
shuse2: I think, in general `validate` keyword is used to throw something right now.
Maybe change to `isvalidTransferData`?

PR #3434, opened 2019-04-24
Use getNodeStatus to retrieve lastBlock in HTTP API

### LINKED ISSUE #3235: Incorrect API responses during snapshotting
### Expected behavior
When API endpoints are called during the snapshotting process - response should be correct (based on the currently processed block).
### Actual behavior
Responses are based on the last block of the chain - so data returned by some API endpoints are not correct according to the current state. Affected places:
https://github.com/LiskHQ/lisk/blob/4f10a534c5205cfe1f9dc62a7eb8329333d26f9e/framework/src/modules/http_api/controllers/accounts.js#L112
https://github.com/LiskHQ/lisk/blob/4f10a534c5205cfe1f9dc62a7eb8329333d26f9e/framework/src/modules/http_api/controllers/node.js#L66
https://github.com/LiskHQ/lisk/blob/4f10a534c5205cfe1f9dc62a7eb8329333d26f9e/framework/src/modules/http_api/controllers/delegates.js#L167
https://github.com/LiskHQ/lisk/blob/4f10a534c5205cfe1f9dc62a7eb8329333d26f9e/framework/src/modules/http_api/controllers/delegates.js#L191
### Steps to reproduce
Run snapshotting and query affected API endpoints.
`1.6.0-alpha.4`

limiaspasdaniel:
### What was the problem?
Responses are based on the last block of the chain - so data returned by some API endpoints are not correct according to the current state while in **snapshot** mode
### How did I fix it?
By calling `modules.lastBlock.get` via a `Chain` action `getNodeStatus` instead of using the `Storage` component.
### How to test it?
Start the node in snapshot mode and check that the last block height is the current block height that is being snapshotted.
### Review checklist

* The PR resolves #3235 
* Linting passes
* Tests pass


shuse2:
wouldn't it better if `calculateSupply` and other functions (which are always used in combination of last block) are included in the nodeStatus?

limiaspasdaniel:
I agree with you, however I think we could do this in a separate PR as this one fixes a bug.

PR #3440, opened 2019-04-24
Merge Feature/improve transactions processing efficiency

shuse2:
### Description
Merge Improve transactions processing efficiency with development branch.

Still several issues need to be addressed, but it is working atomically.


framework/src/modules/chain/logic/init_transaction.js
```diff
@@ -0,0 +1,188 @@
...
+	}
+	const outTransfer = {
+		dappId: raw.ot_dappId,
+		transactionId: raw.ot_outTransactionId,
+	};
+
+	return { outTransfer };
+};
+
+class Transaction {
```
limiaspasdaniel: I think the name of the class is not self-describing enough. My proposal is: `TransactionInitializer` or something similar. Also, this file is exporting only this class while having a different name.

framework/test/mocha/integration/transactions/exceptions/sigantures.js
```diff
@@ -0,0 +1,200 @@
+/*
```
mitsujutsu: should probably fix this file name

shuse2: it seems like test itself is checking the signatures exception.
Probably description needs to be changed

framework/src/modules/chain/logic/init_transaction.js
shuse2: Originally, we thought this would be temporal class, but as far as chain module concern, there is only one Transaction in general way, and i think it's a good name?

framework/test/mocha/integration/transactions/exceptions/sigantures.js
mitsujutsu: oh I meant just the typo `sigantures` --> `signatures` :)

framework/src/modules/chain/logic/init_transaction.js
limiaspasdaniel: What I find misleading is the mismatching names of the file and the content of it. The file is named `init_transaction` while containing a class named `Transaction`. I think naming it `transaction` is more consistent if you want to keep the name of the class untouched, however afaict, this class is responsible for instantiating other different classes that hold different transaction implementations so other names such as `TransactionInitializer` I find them more suitable for the purpose of the class.

framework/src/modules/chain/logic/state_store/index.js
```diff
@@ -0,0 +1,38 @@
...
+	constructor(storage, cb) {
+		this.entities = {
+			Account: storage.entities.Account,
+			Transaction: storage.entities.Transaction,
+			Round: storage.entities.Round,
+		};
+
+		this.stateStore = {};
+
+		return setImmediate(cb, null, this);
```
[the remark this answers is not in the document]
shuse2: Only for backward compatibility, but its a good point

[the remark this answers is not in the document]
limiaspasdaniel: I understand, but if it's new code we should avoid this. I have checked and it's only used once so it should be fairly easy to change.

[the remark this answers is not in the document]
limiaspasdaniel: Maybe to not to slow down the merge of this PR we can open an issue for tackle this along with `TransactionInitializer` in the future

framework/src/modules/chain/logic/init_transaction.js
shuse2: yea the file and the function is mismatching and confusing.
We should definitely fix that. Let's open another issue for this

framework/src/modules/chain/logic/state_store/index.js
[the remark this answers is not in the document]
shuse2: Sure. but for cb style, it's not causing any harm right now, and it's consistent with other code, so I think we can keep it for now.
We would definitely clean up on next version

framework/src/modules/chain/logic/init_transaction.js
limiaspasdaniel: Issue: https://github.com/LiskHQ/lisk-sdk/issues/3447

PR #3446, opened 2019-04-25
Save and load tried peers to/from database in network module

### LINKED ISSUE #2882: Update DB to save existing peers
### Description
Update the database saving mechanism of existing peers in network module.
Also, it should remove the unused functions, and fix/remove tests related.

### Dependent issue
#2874

ishantiw:
### Description

- Add a property in `P2PConfig` to add `triedPeers` explicitly in the constructor in P2P in elements.
- Add and initialize storage component in the network module to access Peers in DB.
- Load `triedPeers` from DB saved previously on the startup and pass it to P2P constructor.
- Save peers to DB on the node shut down.

### Review checklist

* The PR resolves #2882 
* Linting passes
* Tests pass


framework/src/modules/network/network.js
```diff
@@ -47,35 +50,76 @@ module.exports = class Network {
...
+				throw new Error('Cannot bootstrap the storage component');
+			}
+		} catch (err) {
+			throw err;
+		}
+
+		// Load peers from the database that were tried or connected the last time node was running
+		let triedPeers;
+		try {
+			triedPeers = await this.storage.entities.Peer.get();
```
lsilvs: If you don't pass `limit` to `get()` it will use the default that is only 10.
I reckon you want to get all peers from the database, if so use `storage.entities.Peer.get({}, { limit: null })`

ishantiw: Oh yeah, we want to load all the peers from DB, I will add it in the query 👍

framework/src/modules/network/network.js
```diff
@@ -47,35 +50,76 @@ module.exports = class Network {
...
+		const storageConfig = await this.channel.invoke(
+			'app:getComponentConfig',
+			'storage'
+		);
+		const dbLogger =
+			storageConfig.logFileName &&
+			storageConfig.logFileName === loggerConfig.logFileName
+				? this.logger
+				: createLoggerComponent(
+						Object.assign({}, loggerConfig, {
```
shuse2: prefer to
```
{
  ...loggerConfig,
   logFileName: storageConfig.logFileName,
}
```

framework/src/modules/network/network.js
```diff
@@ -47,35 +50,76 @@ module.exports = class Network {
...
+
+		this.storage = createStorageComponent(storageConfig, dbLogger);
+		this.storage.registerEntity('Peer', Peer);
+
+		try {
+			const status = await this.storage.bootstrap();
+			if (!status) {
+				throw new Error('Cannot bootstrap the storage component');
+			}
+		} catch (err) {
```
shuse2: if it's catch and re-throwing, no need to catch here

framework/src/modules/network/network.js
```diff
@@ -47,35 +50,76 @@ module.exports = class Network {
...
+			throw err;
+		}
+
+		// Load peers from the database that were tried or connected the last time node was running
+		let triedPeers;
+		try {
+			triedPeers = await this.storage.entities.Peer.get({}, { limit: null });
```
shuse2: same as above

framework/src/modules/network/network.js
```diff
@@ -47,35 +50,76 @@ module.exports = class Network {
...
+		}
 		// TODO: Nonce overwrite should be removed once the Network module has been fully integreated into core and the old peer system has been fully removed.
 		// We need this because the old peer system which runs in parallel will conflict with the new one if they share the same nonce.
 		const moduleNonce = randomstring.generate(16);
 		const sanitizeNodeInfo = nodeInfo => ({
 			...nodeInfo,
 			state: 2, // TODO: Delete state property
 			nonce: moduleNonce,
-			wsPort: this.options.wsPort,
+			wsPort: this.options.default.wsPort,
```
shuse2: I think we should not use `default` as it could be overwritten by custom config.
Same for below

framework/src/modules/network/network.js
```diff
@@ -227,6 +271,18 @@ module.exports = class Network {
...
+		const peersToSave = this.p2p.getNetworkStatus().connectedPeers.map(peer => {
+			const { ipAddress, ...peerWithoutIp } = peer;
+
+			return {
+				ip: ipAddress,
+				...peerWithoutIp,
+			};
+		});
+
+		await this.storage.entities.Peer.create(peersToSave);
```
shuse2: dont we need to remove too?

framework/src/modules/network/network.js
ishantiw: But at this point, `this.options.wsPort` is `undefined`

framework/src/modules/network/network.js
ishantiw: Yes, I will add it 👍

framework/src/modules/network/network.js
jondubois: They should be combined at some point. The fields from the config file should override the default modules ones. Could be raised as a separate issue if it's not already doing that.

ishantiw: It's fixed in PR #3453 , so I have reverted these changes now.

shuse2:
@ishantiw 
Test is not passing because
```
dropdb lisk_dev
createdb lisk_dev
node ./test/test_app
```
then
```
2019-04-29T21:31:16.441Z] ERROR: lisk-framework/74486 on shuse2-mbp.fritz.box (/Users/stoda/Documents/70_lisk/lisk-sdk/framework/src/components/storage/adapters/pgp_adapter.js:79 in error): relation "peers" does not exist (err.code=42P01)
    error: relation "peers" does not exist
        at Connection.parseE (/Users/stoda/Documents/70_lisk/lisk-sdk/node_modules/pg/lib/connection.js:554:11)
        at Connection.parseMessage (/Users/stoda/Documents/70_lisk/lisk-sdk/node_modules/pg/lib/connection.js:379:19)
        at Socket.<anonymous> (/Users/stoda/Documents/70_lisk/lisk-sdk/node_modules/pg/lib/connection.js:119:22)
        at Socket.emit (events.js:189:13)
        at Socket.EventEmitter.emit (domain.js:441:20)
        at addChunk (_stream_readable.js:284:12)
        at readableAddChunk (_stream_readable.js:265:11)
        at Socket.Readable.push (_stream_readable.js:220:10)
        at TCP.onStreamRead [as onread] (internal/stream_base_commons.js:94:17)
```
This error will occur

framework/src/modules/network/components/storage/sql/migrations/runtime.sql
```diff
@@ -0,0 +1,34 @@
...
+
+
+/*
+  DESCRIPTION: Runtime queries - executed at node start (right after migrations).
+
+  PARAMETERS: None
+*/
+
+-- Overwrite unconfirmed tables with state from confirmed tables
+DROP TABLE mem_accounts2u_delegates;
```
shuse2: probably we don't need this

framework/src/modules/network/components/storage/entities/index.js
```diff
@@ -0,0 +1,20 @@
...
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+'use strict';
+
+module.exports = {
+	Peer: require('./peer'),
+	Migration: require('./migration'),
```
shuse2: can we actually register 2 different migration?

framework/src/modules/network/components/storage/sql/migrations/runtime.sql
ishantiw: removed migration stuff completely from network module

framework/src/modules/network/components/storage/entities/index.js
ishantiw: now it's only registered once in chain module

PR #3448, opened 2019-04-25
Fix commander with latest framework changes - Closes #3315

### LINKED ISSUE #3315: Align node commands with latest Lisk changes
## Description
With the recent changes in [Lisk](https://github.com/LiskHQ/lisk) the configuration will have changes which has impact on commander `node` commands. This issue is to ensure the commander is inline with the changes in core.

ManuGowda:
### What was the problem?
After the smoke test from the Ops team, found out few bugs in the new commander node commands, also feature requests and improvements.
### How did I fix it?
Fixed the bugs identified

- [x] cache should take the whole path for filename
- [x] support multiple core version like <1.5.0, >1.5.0 & <2.0.0, >= 2.0.0
- [x] handle cache directory if doesnt exists
- [x] supporting cutom url for release and snapshot
- [x] support for multiple installation of same and different networks
- [x] option to start by default during install and `no-start` option
- [ ] support custom configuration files
- [x] Handle upgrade from existing version to `2.0.0`, since new commander implementation manages everything via centralized pm2 and the existing releases will not be registered under new pm2 we will have problems

### How to test it?
Run all the node commands to ensure it works as expected.
### Review checklist

* The PR resolves #3315 
* Linting passes
* Tests pass


commander/src/utils/core/commons.ts
```diff
@@ -158,3 +170,84 @@ export const validateVersion = async (
 		throw new Error(error.message);
 	}
 };
+
+export const getSemver = (str: string): string => {
```
shuse2: semver library doesn't support this?

commander/src/utils/core/commons.ts
```diff
@@ -158,3 +170,84 @@ export const validateVersion = async (
...
+	return result[0];
+};
+
+export const dateDiff = (date1: Date, date2: Date): number => {
+	const MINUTES_OR_SECONDS = 60;
+	const HOURS = 24;
+	const INT_RANGE = 1000;
+
+	return (
+		(((new Date(date1) as unknown) as number) -
```
shuse2: this conversion is wrong as it's subtracting the Date instances

commander/src/utils/core/commons.ts
```diff
@@ -158,3 +170,84 @@ export const validateVersion = async (
...
+	};
+};
+
+const getEnvByKey = (
+	instances: ReadonlyArray<Instance>,
+	key: string,
+	defaultValue: number,
+): number => {
+	const maxValue = instances
+		.map(app => ((app as unknown) as { readonly [key: string]: number })[key])
```
shuse2: no other way to do this...? 😢

shuse2:
Also, why node/upgrade is removed?

commander/src/utils/node/commons.ts
```diff
@@ -158,3 +164,24 @@ export const validateVersion = async (
 		throw new Error(error.message);
 	}
 };
+
+export const getSemver = (str: string): string => {
```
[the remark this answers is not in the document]
ManuGowda: no, it only extracts if you give something like this `semver.coerce('dummy 42.6.7.9.3-alpha dummy')` for us, we need to extract the version from URL, so it can't parse it.

commander/src/commands/core/install.ts
```diff
@@ -119,56 +125,80 @@ export default class InstallCommand extends BaseCommand {
...
+		'node:install --network=mainnet --snapshotUrl=https://testnet-snapshot.lisknode.io/blockchain.db.gz custom-mainnet',
 	];
 
 	static flags = {
-		...BaseCommand.flags,
-		network: flagParser.string({
-			...commonFlags.network,
-			default: NETWORK.MAINNET,
-			options: [NETWORK.MAINNET, NETWORK.TESTNET, NETWORK.BETANET],
+		json: flagParser.boolean({
```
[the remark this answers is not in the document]
ManuGowda: it's from the base class, so I can't avoid it, so hiding it, is there a way I can not list it?

commander/src/utils/node/commons.ts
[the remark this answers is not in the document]
ManuGowda: it will not work for this `semver.coerce('https://lisk-releases.ams3.digitaloceanspaces.com/lisk-core/lisk-1.1.0-5556904-Linux-x86_64.tar.gz')`

commander/src/utils/core/commons.ts
ManuGowda: it works https://runkit.com/manugowda/5ccbed7828b3a8001ada0e0c

shuse2: it works then no need to cast the type, but if it's required to cast the type, let's just do it as typescript suggests

commander/src/commands/core/start/index.ts
```diff
@@ -32,6 +33,17 @@ export default class StartCommand extends BaseCommand {
 		},
 	];
 
+	static flags = {
```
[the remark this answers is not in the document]
shuse2: TypeScript type checking complains, so we will keep it as hidden for now

commander/test/utils/download.ts
```diff
@@ -12,44 +16,34 @@ describe('download utils', () => {
 
 	describe('#download', () => {
 		let existsSyncStub: any = null;
```
shuse2: i wasn't part of this, no need to use `null`, also we only use `undefined` in the code.
Also, the type can be `SinonStub`

commander/test/utils/download.ts
```diff
@@ -1,9 +1,13 @@
 import { expect } from 'chai';
-import fs from 'fs';
+import fs from 'fs-extra';
 import * as axios from 'axios';
 import * as downloadUtil from '../../src/utils/download';
 import * as workerProcess from '../../src/utils/worker-process';
 
+const url =
```
shuse2: this should be defined within the `describe`

PR #3451, opened 2019-04-26
Move exceptions.js file to be part of config.json to simplify config structure - Closes #3309

### LINKED ISSUE #3309: Move exceptions.js file to be part of config.json to simplify config structure
### Expected behavior

There should be only one configuration file for one network. 

### Actual behavior

Right now we have one file `exceptions.js` for each network, which actually is part of configurations at key `modules.chain.exceptions`. Using the following code; 

https://github.com/LiskHQ/lisk/blob/c75591a52864d408828c29d410859902c5bffd13/src/index.js#L79

To avoid any confusion we should merge the exception file into actual `config.json` for each network and remove the above code line. This will simplify the config structure and help to understand it easily that exceptions are actually options for chain module. 

1.7

nazarhussain:
### What was the problem?

Exceptions were kept as separate file, although these are part the chain configuration from now on. 

### How did I fix it?

Moved the exceptions to configurations and added the reference to readme file. 

### How to test it?

Run all tests. 

### Review checklist

* The PR resolves #3309
* Linting passes
* Tests pass


lisk/EXCEPTIONS.md
```diff
@@ -0,0 +1,231 @@
+## Exceptions
```
shuse2: file name should be in small letters? I think no reason to be in capital

nazarhussain: We use capital names for all `.md` files in the project. 

https://github.com/LiskHQ/lisk-sdk/tree/development/docs
https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/modules/README.md

any many others, so why kept this one as capital case as well.

lisk/EXCEPTIONS.md
```diff
@@ -0,0 +1,231 @@
...
+
+During the development of Lisk Core, we found some edge cases when the existing data on the network becomes invalid. Either it was because of a bug or some protocol change. To maintain state of the chain we have to consider that particular data valid in that particular scope. This concept of making some invalid data to be valid in the network is known to us as exceptions. To see the full list of available exception categories see the following [exception schema](https://github.com/LiskHQ/lisk-sdk/blob/ca7a626f5d231f07ac784b4f6b58ed72dfa9aaa3/framework/src/modules/chain/defaults/config.js#L110)
+
+Below is the list of exceptions on each network with some details.
```
lsilvs: This file will be hard to maintain as we will need to remember to update it every time we update the exceptions. I would rather add the details to the `schema` itself in the `description` attribute.

shuse2: I hope we don't have to maintain this much, so I think it should be fine, although documentation should be improved for readability in the future.

@MaciejBaj what do you think?

nazarhussain: Yes it seems hard to maintain for individuals, but exceptions are something special case in our application. And when we encounter any such exception, reviewer must keep track of it in this file or separate document.

PR #3460, opened 2019-04-26
Fix skipped integration tests for second_signature and delegates

### LINKED ISSUE #3457: Fee and Amount incorrectly set to 0 for validated transactions received from blocks
### Expected behavior

Upon receiving transactions from blocks, amount and fee should remain the same as they are already validated. 

### Actual behavior

When receiving a block, transactions are being re-initialized in `transaction.jsonRead` from `init_transaction` module.

During this process fees were sometimes incorrectly being defaulted to 0 due to a validation error.

### Steps to reproduce

Run integration tests: 
`transactions/2_delegates/5_same_account_from_pool_and_peer.js`
`transactions/1_second_signature/1_1_second_signature_from_pool_peer.js`

1.7

mitsujutsu:
### What was the problem?

In some skipped integration tests, simulated received blocks contained transactions which had `amount` and `fee` as `BigNum` object.

This caused the `amount` and `fee` to default to `0`, which then invalidated the transaction and failed the test.

### How did I fix it?

Added utility function for converting transactions in block to proper JSON objects.  

### How to test it?

`npm run mocha:integration transactions/1_second_signature/1_1_second_signature_from_pool_peer.js`

`npm run mocha:integration transactions/2_delegates/5_same_account_from_pool_and_peer.js`

### Review checklist

* The PR resolves #3457 
* Linting passes
* Tests pass


shuse2:
When receiving the block, non of the transaction property should be `BigNum` object.
Also, we have validation before doing the process of he transaction.
If this happens in the source code level, I think it's not a problem of the validation but rather programming error.
If this happens only in the test code, I think it's testing the wrong thing (especially in the integration)
In either case, I think we shouldn't change the validation logic, but fix the test scenario

Issue #3263, opened 2019-04-04
Allow some modules to publish any event


jondubois:
### Expected behavior

Sometimes a module may want to publish events which it only knows at runtime.
For example, the Network module should be able to publish any kind of event on the node (depending on messages it receives from the network).

### Actual behavior

Currently, the module needs to specify the complete list of events that it exposes during boot. This can be useful in terms of imposing constraints for some modules but it's too restrictive for modules like the network module. There should be a way for a module to specify that it can publish any event (maybe a wildcard `'*'`?).

In the `feature/create_network_module` branch, the constraint was disabled completely so that any module can publish any event.

diego-G:
@shuse2 @jondubois possibly duplication of https://github.com/LiskHQ/lisk/issues/3097

shuse2:
Related to this https://github.com/LiskHQ/lisk-sdk/pull/3465#discussion_r279243590
It has to be re-restricted with the above possibility and security.

Also, affects #3154

nazarhussain:
Any module can invoke action or subscribe to events of others, but in any case it should not be able to publish events of others, by impersonating or any other case.

jondubois:
@nazarhussain Even with custom events, modules cannot impersonate each other because the prefix of each event will still contain the module name (e.g. the `postBlock` event published by the `chain` module is `chain:postBlock`). I agree with this aspect completely.

If the Network module publishes an event which originates from a Chain module which running on a remote peer, it might look like `network:chain:postBlock`. In this case, we still know that it's the network module which is publishing the event.

>> When you are not defining the events first, then how subscribers aware of those and use those?

Module developers need to read a module's documentation to find out what the payload of each event will look like; so they can also find out what events are available on the module from the same documentation.

On the pub/sub implementation side, there is no need to specify or restrict what channels a module can publish to. You can have dynamic channels where any client can subscribe or publish to any channel (whether it exists or not) - This is the approach used by Redis, SocketCluster, Faye, and most pub/sub systems and pub/sub SaaS services. The principle of creating channels/topics statically (from the beginning) is more related to Message Queues; systems like RabbitMQ, NSQ, MQTT and Kafka. But MQs are not suitable for our use case; they are too heavy; especially for cross-process communication on the same node.

jondubois:
Some possible approaches (this may change):

Subscribing to a remote event (from a remote module on a peer):

```js
// Subscribe to the postBlock event from a remote chain module
[...]

Issue #3154, opened 2019-03-22
Improve module actions security


lsilvs:
### Expected behavior
I expect Modules to be able to define which actions can be accessed for which module.

### Actual behavior
Any action can be invoke by any module

### Which version(s) does this affect? (Environment, OS, etc...)
1.6

lsilvs:
REF.: https://github.com/LiskHQ/lisk/pull/3126#pullrequestreview-217685289

PR #3465, opened 2019-04-28
Merge Feature/create network module

shuse2:
### Description
Merging feature/create_network_module
In the new network module, internally it uses lisk-p2p library.
Any of the interaction directly to the peer is removed from chain, and it's handled within the network module.


framework/src/controller/bus.js
```diff
@@ -179,10 +179,6 @@ class Bus extends EventEmitter2 {
 	 * @throws {Error} If event name does not exist to bus.
 	 */
 	publish(eventName, eventValue) {
-		if (!this.getEvents().includes(eventName)) {
-			throw new Error(`Event ${eventName} is not registered to bus.`);
-		}
-
```
nazarhussain: Why to remove this check? Its important check for the architecture that modules must register events and can only publish those.

lisk/config/alphanet/config.json
```diff
@@ -431,16 +431,16 @@
...
 			}
+		},
+		"network": {
+			"list": [
+				{
+					"ip": "127.0.0.1",
+					"wsPort": 5000
+				}
+			],
+			"wsPort": 5000
```
nazarhussain: Please update the `update_config.js` to apply this change in the user configuration file. See this script for reference: 

https://github.com/LiskHQ/lisk-sdk/blob/2efbe9c6c4d3780e639d1850e09cb3b423def02e/lisk/scripts/update_config.js#L274-L281

framework/src/modules/network/index.js
```diff
@@ -0,0 +1,66 @@
...
+ * Network module specification
+ *
+ * @namespace Framework.Modules
+ * @type {module.NetworkModule}
+ */
+module.exports = class NetworkModule extends BaseModule {
+	/* eslint-disable-next-line no-useless-constructor */
+	constructor(options) {
+		super(options);
+	}
```
nazarhussain: You can skip defining constructor if there is nothing custom for this class.

framework/src/modules/network/index.js
```diff
@@ -0,0 +1,66 @@
...
+
+	static get alias() {
+		return 'network';
+	}
+
+	static get info() {
+		return {
+			author: 'LiskHQ',
+			version: '0.1.0',
+			name: 'lisk-core-network',
```
nazarhussain: May be we should update this naming convention for all modules, for reference this name will reflect the npm package name when we load module directly from a custom npm package, or when we publish these modules as independent nam packages.

framework/src/modules/network/index.js
```diff
@@ -0,0 +1,66 @@
...
+	async load(channel) {
+		const options = {
+			...config,
+			...this.options,
+		};
+		this.network = new Network(options);
+
+		channel.once('chain:ready', async () => {
+			await this.network.bootstrap(channel);
+			channel.publish('network:ready');
```
nazarhussain: As earlier discussed with @shuse2 as well, suggest not use `ready` events with modules, rather use `network:bootstrap`. As we may remove this logic in future and `load` will resolve once the module is fully ready.

framework/src/modules/network/index.js
```diff
@@ -0,0 +1,66 @@
...
+			getPeersCountByFilter: action =>
+				this.network.actions.getPeersCountByFilter(action),
+		};
+	}
+
+	async load(channel) {
+		const options = {
+			...config,
+			...this.options,
+		};
```
nazarhussain: You don't need this here, all the options with their defaults will be merged already and passed to module. If you need config schema in the `Network` class, you can require it there explicitly.

lisk/config/alphanet/config.json
shuse2: Created separate issue #3466

framework/src/modules/network/index.js
shuse2: Created separate issue #3467

framework/src/controller/bus.js
shuse2: This line has to be put it back with consideration of #3263 and #3154 
Main reason is Network module needs to support `network:chain:postBlock` without knowing where it's going to.
It's dynamically generated by `procedure`

nazarhussain: Network module can call action of any module or subscribe to their events, but it can't publish events of others.

shuse2: yes, but it's publishing events under their own namespace `network:*`

nazarhussain: https://github.com/LiskHQ/lisk-sdk/issues/3263#issuecomment-487477409

framework/src/modules/network/index.js
shuse2: It's bit confusing where `bootstrap` comes after `ready`...

nazarhussain: Yes and to remove the confusion we earlier discussed that bootstrap for each module would be synchronous and `app:ready` will be fired at the end when everything ready.

framework/test/mocha/common/generatePeerHeader.js
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+'use strict';
+
+// Initialize global variables
+require('../setup');
+
+const randomstring = require('randomstring');
```
pablitovicente: Maybe here could be better to use `https://www.npmjs.com/package/cuid` ?

elements/lisk-p2p/test/integration/p2p.ts
```diff
@@ -101,7 +103,7 @@ describe('Integration tests for P2P library', () => {
 					// Set a different discoveryInterval for each node; that way they don't keep trying to discover each other at the same time.
 					discoveryInterval: DISCOVERY_INTERVAL + index * 11,
 					nodeInfo: {
-						wsPort: NETWORK_START_PORT + index,
+						wsPort: nodePort,
```
michielmulders: Possibly rename the variable to `wsPort` for clarity (at line 94)?

michielmulders: And other places as well, just minor that can be considered.

framework/test/mocha/common/generatePeerHeader.js
shuse2: this is because of the current protocol

framework/src/controller/bus.js
jondubois: A module should be allowed to publish custom events.

Upon thinking about it more, I don't think that there is any value in restricting what kinds of events a module can publish.

From my perspective, the only purpose of the feature (of explicitly declaring available events from the module via `get events() { }`) is documentation. My interpretation of the declaration is that it should mean ‘this module definitely supports those events’ but it should not mean `this module ONLY supports those events`.

framework/src/modules/network/index.js
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	get events() {
+		return ['bootstrap', 'remoteEvent'];
+	}
+
+	get actions() {
+		return {
+			request: async action => this.network.actions.request(action),
+			send: action => this.network.actions.send(action),
```
nazarhussain: I suggest we revisit the naming convention here to align it with what we have already for the framework and more readable. 

For `event` model:

```js
// node 1 - chain
channel.invoke('network:send', data);


// node 2 - chain 
channel.subscribe('network:receive', data=> { doSomethingWith(data) });
```

For `request` model, since its just re-routing to existing actions, it must use same naming convention as:

```js

// node 1 - chain

get actions() {
	myAction: async (action) => {

	}
}


// If invokded on same node 
const result = await channel.invoke('chain:myAction');


// If invoked from other node 2
const result = await channel.invoke('network:invoke', {remoteAction: 'chain:myAction'})
```

In short if we change `remoteEvent` event to `receive` and `request` to `invokeRemote ` that will make it more readable and easy for end user.

shuse2: Updated to match the framework
instead i changed to
`request` => `invoke`
`send` => `publish`
`remoteEvent` => `subscribe`

ishantiw: Network sounds like a channel within the channel (although it is a channel to communicate with the network =P) with the above naming, maybe something like below to make it more readable, 

`request` => `requestRemote`
`send` => `publishToNetwork` or `sendToNetwork` or just `send`
`remoteEvent` => `receive`

framework/src/modules/chain/chain.js
```diff
@@ -194,6 +174,25 @@ module.exports = class Chain {
 
 			self.logger.info('Modules ready and launched');
 
+			this.channel.subscribe(
+				'network:subscribe',
```
jondubois: This name still doesn't seem quite right. We should raise another issue to adjust the naming.

jondubois: Actually we can use this existing issue: https://github.com/LiskHQ/lisk-sdk/issues/3263

PR #3471, opened 2019-04-29
Move the application-level project to TypeScript - Closes #3391

### LINKED ISSUE #3391: Move the application-level project to TypeScript
### Expected behavior
Provide the TypeScript setup for Lisk Core (types, tsconfig.json, tslint.json)

As the type 5,6,7 transactions will be moved to Lisk Core #2780 , the remaining source files should also be written in TypeScript, i.e: "lisk/src/index.js".

The rest of the files - scripts and corresponding tests can remain in Javascript.

yatki:
### What was the problem?
lisk-core folder was not using Typescript.
### How did I fix it?
Now it uses typescript.
### Review checklist

* The PR resolves #3391 
* Linting passes
* Tests pass


lisk/package.json
```diff
@@ -28,9 +28,11 @@
 	"scripts": {
 		"start": "node src/index.js",
```
shuse2: this start also need to be changed, and related build, docker, pm2 setting need to be changed accordingly

tslint.json
```diff
@@ -28,5 +21,6 @@
 		"strict-type-predicates": false,
 		// We need private class properties to start with an underscore.
 		"variable-name": [true, "allow-leading-underscore"]
-	}
+	},
+	"linterOptions": { "exclude": ["framework"] }
```
shuse2: why do we need this option?

yatki: If we run tslint command in the root level, we can ignore framework folder with that option.

lisk/package.json
```diff
@@ -24,13 +24,15 @@
 		"node": "10.15.3",
 		"npm": "6.4.1"
 	},
-	"main": "src/index.js",
+	"main": "dist/index.js",
 	"scripts": {
-		"start": "node src/index.js",
+		"start": "node dist/index.js",
```
fchavant: [pm2-lisk.json(https://github.com/LiskHQ/lisk-sdk/blob/3391-typescript-lisk-core/lisk/build/target/etc/pm2-lisk.json#L5), [Dockerfile](https://github.com/LiskHQ/lisk-sdk/blob/3391-typescript-lisk-core/lisk/Dockerfile#L48) and [docker-compose.yml](https://github.com/LiskHQ/lisk-sdk/blob/3391-typescript-lisk-core/lisk/docker/docker-compose.yml#L19) must be updated also.

lisk/Dockerfile
```diff
@@ -3,19 +3,19 @@ FROM node:10-alpine AS builder
 ENV NODE_ENV=production
 
 RUN apk --no-cache upgrade && \
-    apk --no-cache add alpine-sdk python2 libtool autoconf automake
+	apk --no-cache add alpine-sdk python2 libtool autoconf automake
```
fchavant: Why change the indentation? both `apk [...]` commands are not aligned any more now.

yatki: @fchavant That's prettier's doing actually. It replaced whitespaces with tabs since we are using tabs everywhere. I say let's keep it that way if it doesn't create a technical problem, what do you think?

fchavant: Why does prettier run on files that are not `*.ts`, `*.js` or `*.json`?

yatki: The prettier extension on my IDE 😅

fchavant: Can it not be configured? I'd really like to keep the `Dockerfile` as-is (more readable; with less changes) if possible.

tslint.json
```diff
@@ -8,18 +7,12 @@
...
 		"no-class": false,
-		"no-delete": true,
 		"no-expression-statement": false,
 		"no-if-statement": false,
-		"no-let": true,
-		"no-loop-statement": true,
 		"no-method-signature": false,
-		"no-mixed-interface": true,
 		"no-object-mutation": false,
-		"no-parameter-reassignment": true,
```
mitsujutsu: `no-parameter-reassignment` and `no-var-keyword` are worth keeping imo. are we planning to re-introduce later?

yatki: No, we are actually keeping them. Just, they are enabled by default, so there is no need those statements to be in tslint.json.

lisk/Dockerfile
yatki: I reverted it.

PR #3477, opened 2019-04-29
Remove VError usage - Closes #3406

### LINKED ISSUE #3406: Remove VError usage from elements
### Description
`VError` was used as base class of all the errors to be more easily track the bugs.

However, the feature itself is not used, and exposing the internal error property to the user, which will be included in the APIError message.

Therefore, it should be removed.

mitsujutsu:
### What was the problem?

VError was adding extraneous fields to API errors. 
VError functionality was not widely used in the code.

### How did I fix it?

Removed VError usage

### How to test it?

`npm run test` in elements and framework

### Review checklist

* The PR resolves #3406 
* Linting passes
* Tests pass


elements/lisk-transactions/src/errors.ts
```diff
@@ -27,7 +26,8 @@ export class TransactionError extends VError {
 		actual?: string | number,
 		expected?: string | number,
 	) {
-		super(message);
+		super();
+		this.message = message;
```
shuse2: i think we can still use
```
super(message)
```
in this case too

mitsujutsu: Tried this but was causing an issue with swagger. The message property was removed if you call `super(message)` here.

PR #3478, opened 2019-04-29
Fix network config - Closes #3466

### LINKED ISSUE #3466: Update network config and update_config script accordingly
### Expected behavior
Network config should be extensible to coming changes and update_config script should be updated for the release 2.0

### Actual behavior
Network config is outdated, and update_config is not up to date

2.0-

shuse2:
### What was the problem?
Network config was outdated with the new structure.

### How did I fix it?
Update the keys for the network configs and fix the update_config script.

### How to test it?
Create sample script to migrate
`LISK_NETWORK=mainnet node scripts/update_config.js config_sample.json 1.6.0 2.0.0`
Also, check the config keys

### Review checklist

* The PR resolves #3466 
* Linting passes
* Tests pass


framework/src/modules/network/defaults/config.js
```diff
@@ -8,16 +8,13 @@ const defaultConfig = {
 			env: 'LISK_WS_PORT',
 			arg: '--port,-p',
 		},
-		address: {
+		hostAddress: {
```
nazarhussain: Why we need to change the config, not the usage? Similar name used in the http_api module as well, should we change it there as well?

framework/src/modules/network/network.js
```diff
@@ -61,16 +61,16 @@ module.exports = class Network {
...
+		const blacklistedPeers = this.options.blacklistedPeers
 			? this.options.access.blackList.map(peer => ({
 					ipAddress: peer.ip,
 					wsPort: peer.wsPort,
 			  }))
 			: [];
 		const p2pConfig = {
 			nodeInfo: initialNodeInfo,
-			hostAddress: this.options.address,
+			hostAddress: this.options.hostAddress,
```
nazarhussain:
```suggestion
			hostAddress: this.options.address,
```

We are using `address` config option for p2p layer from start, no need to change it just for underlying library.

lisk/scripts/update_config.js
```diff
@@ -286,11 +292,29 @@ history.version('1.6.0-rc.0', version => {
...
+		config = moveElement(
+			config,
+			'peers.access.blackList',
+			'modules.network.blacklistedPeers'
+		);
+		config = moveElement(
+			config,
+			'peers.options.timeout',
+			'modules.network.connectTimeout'
+		);
```
nazarhussain: You also renamed `httpServerSetTimeout` to `ackTimeout`, Please add an entry for this as well.

shuse2: actTimeout and httpServerSetTimeout is not related, and httpServerSetTimeout was not used

framework/src/modules/network/network.js
shuse2: This value is not very useful, but I think it's better to keep it consistent?

nazarhussain: Yes keep it consistent to what it was before, if we see `address` to be changed to `hostAddress` then we will do it for http_api module as well.

nazarhussain: Seems you forget to revert this change; 

```suggestion
			hostAddress: this.options.address,
```

framework/test/fixtures/config/devnet/config.json
```diff
@@ -439,7 +439,7 @@
 			}
 		},
 		"network": {
-			"list": [
+			"seedPeers": [
 				{
 					"ip": "127.0.0.1",
```
jondubois: Do we want to keep the `ip` field or rename it to `ipAddress` to match the p2p library?
Either way is OK for me, just something to think about.

PR #3490, opened 2019-05-02
Bug fix ENV and Args passed via CLI - Closes #3475

### LINKED ISSUE #3475: Passing ENV var via CLI does not work to modify app behavior
### Expected behavior
Should accept ENV var to modify app behavior.

### Actual behavior
Schema validation error:
```js
[ { keyword: 'type',
       dataPath: '.modules.http_api.httpPort',
       schemaPath:
        '#/properties/modules/properties/http_api/properties/httpPort/type',
       params: [Object],
       message: 'should be integer' } ] }
```
--> Probably CLI passing args as string and not converting them to specified type.

Also, `--http-port` option does not work as this gets converted to `--httpPort`. 
Starting application with `node lisk/src/index.js --httpPort 3000 | npx bunyan` does work.

2.0.0

michielmulders:
### What was the problem?
Currently, the passed `ENV` variables via the CLI are strings by default. However, our config parser/validator expects certain values to have a specific type (`string`, `integer`, `array`).

This PR also solves the issue with CLI argument `--http-port` as an alias gets created by `yargs`, so it passes both `['--http-port', '--httpPort']`. An extra parsing line has been added to catch this exception which only occurs for `arg` names with a hyphen. 

### How did I fix it?
Decided to modify the logic for `validator/keywords/env` to parse the provided value to the right type (`automagically` solution).

_Why this solution?_ The other solution is to add a custom formatter for each type to every `ENV` variable we define which is overkill. 

### How to test it?
Try starting the app with several ENV vars:
- Array: ` LISK_FORGING_WHITELIST=10.10.1.1,13.13.13.13 node lisk/src/index.js | npx bunyan`
- Integer: `LISK_HTTP_PORT=3000 node lisk/src/index.js | npx bunyan`
- Boolean: ` LISK_API_PUBLIC=true node lisk/src/index.js | npx bunyan`

Test `node lisk/src/index.js --http-port 3000 | npx bunyan` option

### Review checklist

* The PR resolves #3475 
* Linting passes
* Tests pass


framework/src/controller/validator/keywords/env/index.js
```diff
@@ -18,7 +18,23 @@ const compile = (schema, parentSchema) => {
 			  };
 
 	return function(data, dataPath, object, key) {
-		const variableValue = process.env[envVariable.name];
+		let variableValue = process.env[envVariable.name];
+
+		// Formatting logic to automatically format the CLI value to the expected type
+		// Only format if the variable exist
+		if (process.env[envVariable.name] !== undefined) {
```
yatki: I'd prefer to use `variableValue` instead of `process.env[envVariable.name]`, since it's easier to read.

framework/src/controller/validator/keywords/env/index.js
```diff
@@ -18,7 +18,23 @@ const compile = (schema, parentSchema) => {
...
+			// eslint-disable-next-line default-case
+			switch (parentSchema.type) {
+				case 'array':
+					variableValue = process.env[envVariable.name].split(',');
+					break;
+				case 'integer':
+					variableValue = parseInt(process.env[envVariable.name]);
```
yatki:
```suggestion
					variableValue = parseInt(variableValue, 10);
```

framework/src/controller/validator/keywords/env/index.js
michielmulders: Ah good catch!

framework/src/controller/validator/keywords/env/index.js
```diff
@@ -18,7 +18,23 @@ const compile = (schema, parentSchema) => {
 			  };
 
 	return function(data, dataPath, object, key) {
-		const variableValue = process.env[envVariable.name];
+		let variableValue = process.env[envVariable.name];
```
shuse2:
```suggestion
		const variableValue = process.env[envVariable.name];
```

framework/src/controller/validator/keywords/env/index.js
yatki: See: https://eslint.org/docs/rules/radix

framework/src/controller/validator/keywords/env/index.js
yatki: I don't think that's possible because he is mutating the value in the lines below.

shuse2: oops, sorry

michielmulders: Must stay `let` here as I modify the `variableValue` if it needs to be parsed.

michielmulders:
@nazarhussain targets both, for `args`, only [one line extra](https://github.com/LiskHQ/lisk-sdk/pull/3490/files#diff-1ed813279c95e21b978c1886864dae2bR117) added to fix.

`this.listOfArgs.add(arg.replace(/(^--)(.*)/, '$2'));`

framework/src/controller/validator/keywords/env/index.js
```diff
@@ -18,7 +18,23 @@ const compile = (schema, parentSchema) => {
...
+
+		// Formatting logic to automatically format the CLI value to the expected type
+		// Only format if the variable exist
+		if (variableValue !== undefined) {
+			// eslint-disable-next-line default-case
+			switch (parentSchema.type) {
+				case 'array':
+					variableValue = variableValue.split(',');
```
nazarhussain: This change will break the following implementation. 

https://github.com/LiskHQ/lisk-sdk/blob/32950c965f96e8b54bcaf9a6667089a11318dc7b/framework/src/modules/network/defaults/config.js#L49

https://github.com/LiskHQ/lisk-sdk/blob/32950c965f96e8b54bcaf9a6667089a11318dc7b/framework/src/controller/validator/keywords/formatters.js#L18-L28

I suggest not to parse array values on this level. Leave it to actual implementation. String to array conversion can be different in different cases.

framework/test/jest/unit/specs/controller/helpers/validator/keywords/env.spec.js
```diff
@@ -36,6 +36,44 @@ describe('validator keyword "env"', () => {
...
+			properties: {
+				prop2: {
+					type: 'integer',
+					env: 'PROP2',
+				},
+			},
+		};
+
+		const data = { prop2: '999' };
+		process.env.PROP2 = '999';
```
shuse2: process.env need to be cleaned up in afterEach

michielmulders: Fixed!

PR #3496, opened 2019-05-02
Modify boolean validation logic env - Closes #3495

### LINKED ISSUE #3495: Improve handling boolean ENV var passed via CLI
### Expected behavior
For type boolean, we check if the variable is equal to `"true"`. If yes, we return `true`, else `false`. This implies that `this-is-wrong` renders `false`.

We want to change the behavior to throw a meaningful error.

### Steps to reproduce
Start app with `LISK_API_PUBLIC=invalid-param`.

2.0.0

michielmulders:
### How did I fix it?
Return error if value for type boolean is not equal to `"true"` or `"false"`.
Added test, and updated others accordingly.

### How to test it?
Start app with `LISK_API_PUBLIC=truuuee TS_NODE_PROJECT=./tsconfig.json ts-node src/index.ts | bunyan`

### Review checklist

* The PR resolves #3495 
* Linting passes
* Tests pass


framework/src/controller/validator/keywords/env/index.js
```diff
@@ -45,7 +45,17 @@ const compile = (schema, parentSchema) => {
...
+						variableValue = true;
+					} else if (variableValue.toLowerCase() === 'false') {
+						variableValue = false;
+					} else {
+						throw new Error(
+							`Failed to apply value for option ${
+								envVariable.name
+							}, use "true" or "false"`
+						);
+					}
```
yatki:
```js
if (variableValue.toLowerCase() !== 'true' && variableValue.toLowerCase() !== 'false') {
   throw new Error(`Failed to apply value for option ${envVariable.name}, use "true" or "false"`);
} 
          
variableValue = variableValue.toLowerCase() === 'true';
break;
```
how about this ? :D

michielmulders: Shorter!

shuse2: `['true', 'false'].includes(variableValue.toLowerCase())`?

PR #3501, opened 2019-05-03
Refactor init_transaction DB read into elements - Closes #3429

### LINKED ISSUE #3429: `init_transaction` module not extensible to custom transactions
### Expected behavior

`logic/init_transaction` module should be extensible to accommodate for new custom transaction types created by sdk user.

### Actual behavior

`logic/init_transaction` module directly maps known transaction `type`s to raw db format. 

See issue #2424 

1.7

mitsujutsu:
### What was the problem?

When reading transactions from db, mapping from db format to JSON should not be done in `framework`, as it would not work for custom transactions.

### How did I fix it?

- Moved dbRead functionality to `elements`.
- Remove type validations from elements and swagger 

### Review checklist

* The PR resolves #3429 #3536
* Linting passes
* Tests pass


framework/src/modules/chain/logic/init_transaction.js
```diff
@@ -163,43 +48,15 @@ class Transaction {
...
-
-		if (asset) {
-			transactionJSON.asset = Object.assign(transactionJSON.asset, asset);
-		}
+		const transactionJSON = this.transactionClassMap(
+			parseInt(raw.t_type)
+		).dbRead(raw);
 
-		return this.fromJson(_.omitBy(transactionJSON, _.isNull));
+		return this.fromJson(transactionJSON);
```
pablitovicente: why are we omitting the cleanup of `nulls` now?

mitsujutsu: this logic was abstracted to `baseTransaction.dbRead`

elements/lisk-transactions/src/base_transaction.ts
```diff
@@ -454,6 +456,52 @@ export abstract class BaseTransaction {
...
+			readonly requesterPublicKey: string;
+			readonly [key: string]: string | number | object | null;
+		} = {
+			id: raw.t_id,
+			height: raw.b_height,
+			blockId: raw.b_id || raw.t_blockId,
+			type: parseInt(raw.t_type, 10),
+			timestamp: parseInt(raw.t_timestamp, 10),
+			senderPublicKey: raw.t_senderPublicKey,
+			requesterPublicKey: raw.t_requesterPublicKey,
```
shuse2: I think this key doesn't exist, so we can remove?
please check on the storage

elements/lisk-transactions/src/base_transaction.ts
```diff
@@ -454,6 +456,52 @@ export abstract class BaseTransaction {
...
+		).reduce(
+			(tx, key) => {
+				// Check for null or undefined
+				if (transactionJSON[key]) {
+					tx[key] = transactionJSON[key];
+				}
+
+				return tx;
+			},
+			{} as any,
```
shuse2: do we need `any` here? prefer to avoid `any` in source code

elements/lisk-transactions/src/base_transaction.ts
```diff
@@ -454,6 +456,52 @@ export abstract class BaseTransaction {
...
+			confirmations: parseInt(raw.confirmations || 0, 10),
+			asset: {},
+		};
+
+		const transaction = Object.keys(
+			{ ...transactionJSON, asset: this.assetDbRead(raw) } || {},
+		).reduce(
+			(tx, key) => {
+				// Check for null or undefined
+				if (transactionJSON[key]) {
```
shuse2: `0` or `''` will be falsy here too

elements/lisk-transactions/src/base_transaction.ts
mitsujutsu: I thought so too but it still exists in `storage/entities/transactions` line 135. Also the original `dbRead` function included it.

elements/lisk-transactions/src/base_transaction.ts
mitsujutsu: This is to avoid tslint complaining about `no index signature`

shuse2:
Please rebase to `release/2.1.0` branch

mitsujutsu:
> Please verify why the new blocks aren't forged with a new tr type registered.

This is unrelated to new transaction type being registered, and rather has to do with enabling delegates in configurations.

elements/lisk-transactions/src/0_transfer_transaction.ts
```diff
@@ -260,4 +260,16 @@ export class TransferTransaction extends BaseTransaction {
 
 		return errors;
 	}
+
+	// tslint:disable:next-line: prefer-function-over-method no-any
+	protected assetDbRead(raw: any): object | undefined {
```
nazarhussain: We don't use `db` term anywhere in the project anymore. Now we use `storage` everywhere. Please refactor all the code to update reference of `db` with `storage`

shuse2: this is intended for 2.0, and we don't want to refactor this now.
Also, this method will be removed in 3.0.

framework/src/modules/chain/logic/init_transaction.js
```diff
@@ -163,41 +49,21 @@ class Transaction {
...
+		if (!transactionClass) {
+			return null;
 		}
 
-		const asset = assetDbRead(raw);
-
-		if (asset) {
-			transactionJSON.asset = Object.assign(transactionJSON.asset, asset);
-		}
+		const transactionJSON = transactionClass.prototype.dbRead(raw);
```
shuse2: prefer to instanciate and call dbRead, otherwise it should be static

framework/src/modules/chain/logic/init_transaction.js
```diff
@@ -163,41 +49,21 @@ class Transaction {
 
 	// TODO: remove after https://github.com/LiskHQ/lisk/issues/2424
 	dbRead(raw) {
-		if (!raw.t_id) {
+		if (
+			raw.t_id === undefined ||
+			raw.t_id === null ||
+			raw.t_type === undefined ||
+			raw.t_type === null
```
shuse2: falsy check `(!raw.t_id  || !raw.t_type)` is not enough?

mitsujutsu: !raw.t_type would evaluate to true if type = 0

shuse2: ah yes... sorry ='(

framework/src/modules/chain/logic/init_transaction.js
mitsujutsu: Since you first need the JSON to intialize, and you call `dbRead` to get it calls `assetDbRead`, I would have to first strip the `raw.asset` from the db object, then instantiate. I think this way is cleaner. Making it `static` does not work since `dbRead` calls `this.assetDbRead`.

elements/lisk-transactions/src/0_transfer_transaction.ts
nazarhussain: @shuse2 Do we have `db` in the `2.0`? if not then why we have it in transaction? 


So these are the methods with others have to implement in their custom transactions. So we should straight up the naming convention right now.

shuse2: It's the implementation for backward compatibility, and just moving the function to the transaction
it will be removed on next hardfork

nazarhussain: These are the methods with others have to implement in their custom transactions?

nazarhussain: BTW We renamed the db to storage a long time earlier. Why we even followed this naming convention from first place.

shuse2: this is not even `storage` convention.
It's legacy left from `db`, and the `storage` is not related
and yes user have to implement it, otherwise sync process doesn't work well ='(
correct? @mitsujutsu

nazarhussain: If user have to implement `assetDbRead` method in their custom transaction after 2.0 release. Then we must fix this naming convention right now.

mitsujutsu: That's correct user will have to implement it, but only until the hardfork. This is necessary as a work around to allow for custom tx in the meantime.

PR #3522, opened 2019-05-06
Use redis cache from within http module - Closes #3513

### LINKED ISSUE #3513: Use redis cache from within http module
### Expected behavior

Redis cache should only be used by http-module

### Actual behavior

Redis cache is being used in `submodules/transactions / getTransactionsCount()`

### Steps to reproduce

- Start the application
- GET `http://localhost:4000/api/node/status`

pablitovicente:
### What was the problem?

Cache component was being used from within the chain module and http module

### How did I fix it?

Given that cache is used by the HTTP module clearing is now taken care by subscribing to the `chain:blocks:change` event

### Review checklist

* The PR resolves #3513
* Linting passes
* Tests pass


framework/src/modules/chain/submodules/blocks.js
```diff
@@ -148,29 +148,11 @@ class Blocks {
 	// eslint-disable-next-line class-methods-use-this
 	async onNewBlock(block) {
 		if (components && components.cache && components.cache.isReady()) {
```
SargeKhan: I guess this condition should also be in http_api module.

framework/src/modules/chain/index.js
```diff
@@ -61,6 +61,7 @@ module.exports = class ChainModule extends BaseModule {
 			'delegates:fork',
 			'loader:sync',
 			'dapps:change',
+			'invalidate_cache',
```
SargeKhan: I don't think we need to create a new event. The chain module shouldn't worry about firing a new event for invalidating cache. In case of a block addition/deletion, it should simply call: 
```
library.channel.publish('chain:blocks:change', block);
```
And the http_api module should decide based on this event whether it needs to invalidate cache or not.

framework/src/modules/chain/submodules/blocks.js
pablitovicente: If the check is not here the publish never happens so the subscriber will never fire it's callback

framework/src/modules/chain/index.js
pablitovicente: I like this idea but aren't we potentially overloading the `chain:blocks:change` what if we want to take another action in the future for this message?

SargeKhan: I think that `chain:blocks:change` is an event. There can be multiple actions for one event. One action could be to reset the cache, another could be to update Lisk Nano etc.. If in the future we want to add another action, we can simply add another listener for this event.
Please note that I am not recommending to use the same action to perform all the tasks, I am suggesting to add multiple event handlers aka actions for `chain:blocks:change` event.

pablitovicente: Changed HTTP module to listen for `chain:blocks:change` for clearing cache `onNewBlock`

framework/src/modules/chain/submodules/blocks.js
pablitovicente: Due to the other change this change was also made for validating the status of the cache

framework/src/modules/chain/submodules/blocks/chain.js
```diff
@@ -432,40 +423,7 @@ class Chain {
...
-				library.logger.error(
-					['Cache - Error clearing keys', 'on chain afterSave function'].join(
-						' '
-					)
-				);
-			}
-		}
-	}
-
+	library.channel.publish('chain:confirmed_transactions:update', block);
```
shuse2: maybe better event name or payload?
it seems like it should send the transaction but it's sending block...

pablitovicente: It needs all the transactions; I will update it to send only the transactions instead of the rest of the block.

pablitovicente: Done! Please review again.

pablitovicente:
> Look good to me but I think Lisk Hub may have this issue again #2403

Maybe let's work on that issue on its own; the changes do not change cache clearing timing only from where the cache is cleared. What you think?

PR #3527, opened 2019-05-06
Reference to application specific setting should be removed from framework - Closes #3402

### LINKED ISSUE #3402: Reference to application specific setting should be removed from framework
### Description
`.build` file reference and git `commit` reference are used in the chain module.
However, those are application specific property and it should be removed from the chain module.

`.build` file is not used in the framework layer.
git `commit` reference are not used because the framework will be in the npm registry.

yatki:
### What was the problem?
The chain module was generating application related information such as lastCommitId and build version.
### How did I fix it?
I moved application related logic to lisk folder, passed lastCommitId and buildVersion as config parameters to Application, and used them in the chain module.
### Review checklist

* The PR resolves #3402 
* Linting passes
* Tests pass


framework/src/controller/application.js
```diff
@@ -101,6 +101,8 @@ class Application {
 	 * @param {string} [config.app.version] - Version of the application
 	 * @param {string} [config.app.minVersion] - Minimum compatible version on the network
 	 * @param {string} [config.app.protocolVersion] - Compatible protocol version application is using
+	 * @param {string} [config.app.lastCommitId] - Last commit id coming from application repository
+	 * @param {string} [config.app.buildVersion] - Build version of the application
```
nazarhussain: Please add JSON-Schema for these new attributes in application_config_schema

framework/src/controller/application.js
```diff
@@ -167,6 +169,8 @@ class Application {
 
 		this.registerModule(ChainModule, {
 			registeredTransactions: this.getTransactions(),
+			lastCommitId: this.config.app.lastCommitId,
+			buildVersion: this.config.app.buildVersion,
```
nazarhussain: Both of these values are used in the `http_api` module as well, so don't pass it only to chain module.

See in the function `_compileAndValidateConfigurations` and you will find an object `appConfigToShareWithModules`, please add these values to it, so it would be shared by all modules. 

Then make sure the `http_api` module must use these values in the API.

nazarhussain: The `.build` file for build version is generated at the time of compiling the binary. We must document it extensively for application development, so they know some examples how we are doing it and how they can do it to identify the build.

lisk/tslint.json
```diff
@@ -1,6 +1,8 @@
 {
 	"extends": ["../tslint.json"],
 	"rules": {
-		"no-null-keyword": false
+		"no-null-keyword": false,
+		"no-let": false,
+		"no-console": false
```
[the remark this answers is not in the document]
yatki: @lsilvs We don't have a proper logger at the moment, thus still using console statements. Until we introduce a logger, I decided to disable that rule instead of suppressing the error messages.

framework/src/controller/application.js
yatki: @nazarhussain where do you think we should put that information? I think lots of things are not documented at the moment. So I'm not sure if it make sense just to highlight that info somewhere. Do we have a parent issue somewhere that we can add this as a child issue, so we can address this later?

lisk/src/helpers/git.ts
```diff
@@ -14,52 +14,52 @@
...
+const getLastCommitIdFromRevisionFile = (): string => {
+	// REVISION file is being created in the root folder during build process.
 	try {
 		return fs
 			.readFileSync('REVISION')
 			.toString()
 			.trim();
 	} catch (error) {
-		throw new Error('Not a git repository and no revision file found.');
+		throw new Error('REVISION file found.');
```
shuse2: it should be `not found`

lisk/tslint.json
```diff
@@ -1,6 +1,7 @@
 {
 	"extends": ["../tslint.json"],
 	"rules": {
-		"no-null-keyword": false
+		"no-null-keyword": false,
+		"no-let": false
```
shuse2: prefer to keep it true

lisk/src/helpers/build.ts
```diff
@@ -0,0 +1,32 @@
...
+
+const getBuildVersion = (): string => {
+	// .build file is being created in the root folder during build process.
+	try {
+		return fs
+			.readFileSync(path.join(__dirname, '../../', '.build'), 'utf8')
+			.toString()
+			.trim();
+	} catch (error) {
+		throw new Error('.build file found.');
```
shuse2: same as above `not found`

framework/test/mocha/unit/modules/chain/chain.fixtures.js
```diff
@@ -6,12 +6,18 @@ const storageConfig = {
...
+const buildVersion = '#buildVersion';
+const peerList = ['peerList'];
+
 const chainOptions = {
 	genesisBlock: {},
 	loading: {},
 	network: {
 		enabled: false,
 	},
+	lastCommitId: gitLastCommitId,
```
shuse2: this is not in the chain option anymore?

framework/test/mocha/unit/modules/chain/chain.js
```diff
@@ -189,6 +185,10 @@ describe('Chain', () => {
 			return expect(chain.scope.lastCommit).to.be.equal(gitLastCommitId);
 		});
 
+		it('should get build version from config.buildVersion', () => {
+			return expect(chain.scope.build).to.be.equal(buildVersion);
```
shuse2: lastCommit and build can be removed from chain if it's not used

shuse2:
Please target the PR to `release/2.0.0` branch

PR #3530, opened 2019-05-06
Improve actions security - Closes #3154

### LINKED ISSUE #3154: Improve module actions security

michielmulders:
### What was the problem?
Users could find out who the delegates are by calling action `getForgingStatusForAllDelegates`.

### How did I fix it?
Use `invokePublic` for p2p requests to another node and check if an action is public or not.
Only `public=false` action is now `getForgingStatusForAllDelegates`.

### How to test it?
`npm run jest:unit`

OR: Try synching with testnet and modify some actions to `public: false` in `chain/index.js` for `blocks` and `blocksCommon` action (which is used for synching).

### Review checklist

* The PR resolves #3154 
* Linting passes
* Tests pass


michielmulders:
@yatki Can you roll your eye over the Jest tests I modified, nothing major. I see there are still a lot of open tests that should be covered for this functionality, but as this PR is quite urgent for `release/2.0.0` I haven't written them yet. @shuse2 What do you think?

framework/test/jest/unit/specs/controller/bus.spec.js
```diff
@@ -85,7 +85,18 @@ describe('Bus', () => {
 			const actions = ['action1', 'action2'];
 
 			// Act
-			await bus.registerChannel(moduleAlias, [], actions);
+			await bus.registerChannel(moduleAlias, [], actions, {
```
yatki: For the sake of readability, please move this object to a top level variable, set it in a beforeEach hook and reuse that variable, mutate it in an `Arrange` section if needed.

framework/test/jest/unit/specs/controller/channels/base_channel.spec.js
```diff
@@ -156,6 +156,13 @@ describe('Base Channel', () => {
 		});
 	});
 
+	describe('#invokePublic', () => {
+		it('should throw TypeError', () => {
+			// Assert
+			expect(baseChannel.invokePublic()).rejects.toBeInstanceOf(TypeError);
```
yatki: Don't forget to `return` or `await` expect statement when fulfilling promises. Otherwise it can give a false positive result by giving an unhandled promise rejection warning.

See: https://jestjs.io/docs/en/asynchronous#resolves-rejects

framework/src/controller/controller.js
```diff
@@ -161,10 +161,18 @@ class Controller {
 			'app',
 			['ready', 'state:updated'],
 			{
-				getComponentConfig: action => this.config.components[action.params],
-				getApplicationState: () => this.applicationState.state,
-				updateApplicationState: action =>
-					this.applicationState.update(action.params),
+				getComponentConfig: {
+					handler: action => this.config.components[action.params],
+					public: true,
```
lsilvs: This should be false

framework/src/controller/controller.js
```diff
@@ -161,10 +161,18 @@ class Controller {
...
+				},
+				getApplicationState: {
+					handler: () => this.applicationState.state,
+					public: true,
```
lsilvs: This should be false

framework/src/controller/controller.js
```diff
@@ -161,10 +161,18 @@ class Controller {
...
+				},
+				updateApplicationState: {
+					handler: action => this.applicationState.update(action.params),
+					public: true,
```
lsilvs: This should be false

framework/src/modules/chain/index.js
```diff
@@ -66,32 +66,85 @@ module.exports = class ChainModule extends BaseModule {
 
 	get actions() {
```
lsilvs: We should only make public few actions that are required in the current protocol imo. Everything else shouldn't be exposed.

framework/src/modules/network/index.js
```diff
@@ -49,12 +49,30 @@ module.exports = class NetworkModule extends BaseModule {
 
 	get actions() {
```
lsilvs: Same as before

framework/src/modules/chain/index.js
michielmulders: Made everything `false` except for `blocks` and `blocksCommon`.

framework/src/modules/network/index.js
michielmulders: Not sure which ones should be public here?
Made everything public except for `getPeersCountByFilter`.
@shuse2 @lsilvs

framework/src/modules/chain/index.js
shuse2: getTransaction and getSignarures also should be public

framework/src/modules/network/index.js
shuse2: everything can be false i think, it will not be called from peers directly.
Also, `emitisPublic`?

michielmulders: Removed this as it was unused. Waiting for _greenies_ (y)

framework/src/modules/network/index.js
```diff
@@ -49,12 +49,26 @@ module.exports = class NetworkModule extends BaseModule {
...
+				handler: async action => this.network.actions.request(action),
+				isPublic: false,
+			},
+			emit: {
+				handler: action => this.network.actions.emit(action),
+				isPublic: false,
+			},
+			getNetworkStatus: {
+				handler: () => this.network.actions.getNetworkStatus(),
+				isPublic: true,
```
shuse2: this is also should be false

framework/src/modules/network/index.js
```diff
@@ -49,12 +49,26 @@ module.exports = class NetworkModule extends BaseModule {
...
-			getPeers: action => this.network.actions.getPeers(action),
-			getPeersCountByFilter: action =>
-				this.network.actions.getPeersCountByFilter(action),
+			request: {
+				handler: async action => this.network.actions.request(action),
+				isPublic: false,
+			},
+			emit: {
+				handler: action => this.network.actions.emit(action),
+				isPublic: false,
```
shuse2: do we need to say `isPublic: false` when the default is false?

michielmulders: You are right about the action default. My initial thought was that I want people to explicitly think about the publicness of the action they are creating.

Removed the `isPublic:false`.

framework/src/modules/network/index.js
michielmulders: No, it's currently needed for discovering peers as this also holds a list of the connected peers.
Without, network tests fail, but @jondubois is changing this. When he's doing his PR, he should make this false again.

framework/test/jest/unit/specs/controller/bus.spec.js
```diff
@@ -59,7 +72,7 @@ describe('Bus', () => {
 			const events = ['event1', 'event2'];
 
 			// Act
-			await bus.registerChannel(moduleAlias, events, []);
+			await bus.registerChannel(moduleAlias, events, [], channelOptions);
```
yatki: ❤️

framework/test/jest/unit/specs/controller/channels/child_process_channel.spec.js
```diff
@@ -382,7 +391,7 @@ describe('ChildProcessChannel Channel', () => {
 	describe('#_rejectWhenAnySocketFailsToBind', () => {
 		beforeEach(() => childProcessChannel.registerToBus(socketsPath));
 
-		it('should reject if any of the sockets receive an "error" event', () =>
+		it('should reject if any of the sockets receive an "error" event', async () =>
```
yatki: here function is async but .rejects wasn't awaited.

framework/test/jest/unit/specs/controller/channels/child_process_channel.spec.js
```diff
@@ -431,7 +440,7 @@ describe('ChildProcessChannel Channel', () => {
 	describe('#_rejectWhenTimeout', () => {
 		beforeEach(() => childProcessChannel.registerToBus(socketsPath));
 
-		it('should reject with an Error object with proper message', () =>
+		it('should reject with an Error object with proper message', async () =>
```
yatki: same above.

michielmulders: removed async as it's a one-liner with implicit return

framework/test/jest/unit/specs/controller/channels/child_process_channel.spec.js
michielmulders: removed async as it's a one-liner with implicit return

framework/src/controller/channels/base_channel.js
```diff
@@ -106,6 +106,14 @@ class BaseChannel {
 		throw new TypeError('This method must be implemented in child classes. ');
 	}
 
+	// Call action of any moduleAlias through controller
+	// Specified as moduleName:actionName
+	// Specified action must be defined as publicly callable
+	// eslint-disable-next-line no-unused-vars, class-methods-use-this
+	async invokePublic(actionName, params) {
```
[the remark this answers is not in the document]
shuse2: Created new issue regards to this to tackle separately #3553

nazarhussain:
@shuse2 I don't think its a good idea to postpone these changes to separate issue. And if that's to be done for 2.0.0 we should have done in this PR.

PR #3537, opened 2019-05-07
Transactions not being instantiated on block receive - Closes #3535

### LINKED ISSUE #3535: Transactions not being instantiated when receiving a block
### Expected behavior
Valid blocks with valid transactions should be processed successfully
### Actual behavior
When receiving a block from a peer, a call to `objectNormalize` is performed. This function also used in many places elsewhere expects transactions to be instantiated using their class implementations but the current code doesn't do this and it crashes.

Also, when broadcasting a block transactions are not explicitly converted to JSON by calling `.toJSON()` transaction's class function. Socketcluster is somewhat stringifying the transaction ES6 class and parsing it at JSON before sending.
### Steps to reproduce
Run to nodes, one forging, one now and check the how `framework/src/modules/chain/submodules/transport.js:570` throws an error.
`release/2.0.0`

limiaspasdaniel:
### What was the problem?
Transactions were not being instantiated correctly when receiving a block from another peer causing the block to be rejected.
### How did I fix it?
- Converting raw transactions to its class implementation upon receiving a block.
- Also now transactions are explicitly converted to JSON by calling `toJSON` in its class implementation before being broadcasted
- Integration tests for matcher to test when a node explicitly asks for transactions, asks for blocks, receives transactions and receives blocks.

### How to test it?
Run matcher integration tests `npx mocha framework/test/mocha/integration/matcher.js`
### Review checklist

* The PR resolves #3535 
* Linting passes
* Tests pass


framework/src/modules/chain/submodules/process_transactions.js
```diff
@@ -82,6 +82,12 @@ class ProcessTransactions {
 
 	// eslint-disable-next-line class-methods-use-this
 	async checkPersistedTransactions(transactions) {
+		const responseObject = {
+			transactionsResponses: [],
+		};
+
+		if (transactions.length < 1) return responseObject;
```
shuse2:
```
if (!transactions.length) {
  return {
    transactionsReponses: [],
  }
}
```

framework/src/modules/chain/submodules/process_transactions.js
```diff
@@ -117,6 +123,7 @@ class ProcessTransactions {
 		];
 
 		return {
+			...responseObject,
```
shuse2: we probably don't need to initialise with this?

PR #3569, opened 2019-05-09
Add waiting threshold before forging block if node did not receive the block from previous slot - Closes #3564

### LINKED ISSUE #3564: Add waiting threshold before forging block if node did not receive the block from previous slot
### Expected behavior
The application should provide a configurable parameter for block produces to additionally wait in their slot in case they did not receive block from previous slot. This functionality is added to ensure that delegate forges block on the tip of the chain despite network delay, propagation delay, etc.
### Actual behavior
The delegate node always forges block in the first second of the delegate slot and this functionality is not configurable.
3.0.0

nazarhussain:
### What was the problem?
Node was forging block without waiting for previous blocks on their turn. 

### How did I fix it?
Added a configurable option for threshold to wait before forging. 

### How to test it?
Run all tests. 

### Review checklist

* The PR resolves #3564
* Linting passes
* Tests pass


framework/src/modules/chain/submodules/delegates.js
```diff
@@ -587,35 +589,55 @@ __private.forge = function(cb) {
...
+					`Waiting for ${
+						library.config.forging.waitThreshold
+					} seconds for the last block before forging`
+				);
+				library.logger.debug('Slot information', {
+					currentSlot,
+					lastBlockSlot,
+				});
+				return setTimeout(() => {
+					__private.generateBlock(delegateKeypair, currentSlot, cb);
```
SargeKhan: I don't think we should use setTimeout here for two reasons:
1- It's possible that we receive the block from the previous slot before this setTimeout is executed and if the lastBlockSlot was the last slot of the round, we will call __private.generateBlock with invalid keypair slot based because the round is changed.
2- __private.forge is executed every second, therefore we don't need to setTimeout anyway.

nazarhussain: Not convinced with point 1 as even though we receive block meanwhile the currentSlot and keyapir for that slot does not change. 

Yes for the second point, yes we can skip in case we want to wait.

SargeKhan: If it's the end of the round and the last block is not the most recent one, then this line will have different result: 
```
const round = slots.calcRound(lastBlock.height + 1);
```
Therefore, the round parameter passed to this function will be different:
```
__private.getDelegateKeypairForCurrentSlot(currentSlot, round)
```
Therefore, the response returned from this function will be invalid because it's from a previous round.

SargeKhan:
The failing tests on the jenkins doesn't look like a randomly failing test. Can you verify it? The logs are:
```
lisk-framework: /home/lisk/workspace/lisk-sdk_PR-3569/framework/test/mocha/unit/modules/chain/chain.fixtures.js

lisk-framework:   22:2  error  Duplicate key 'forging'  no-dupe-keys

lisk-framework: ✖ 1 problem (1 error, 0 warnings)

lisk-framework: npm ERR! code ELIFECYCLE

lisk-framework: npm ERR! errno 1

lisk-framework: npm ERR! lisk-framework@0.1.0-alpha.8 lint: `eslint .`

lisk-framework: npm ERR! Exit status 1

lisk-framework: npm ERR! 

lisk-framework: npm ERR! Failed at the lisk-framework@0.1.0-alpha.8 lint script.

lisk-framework: npm ERR! This is probably not a problem with npm. There is likely additional logging output above.

lisk-framework: npm ERR! A complete log of this run can be found in:

lisk-framework: npm ERR!     /home/lisk/.npm/_logs/2019-05-21T12_51_31_611Z-debug.log

```

PR #3579, opened 2019-05-09
Refactor redundant code in Account submodule/logic #3524

### LINKED ISSUE #3524: Refactor redundant code in Account submodule/logic
### Expected behavior
Expect to call storage call whenever possible inside Chain module in order to clean up Account submodules/logic.

### Actual behavior
A lot of functions in Account submodules/logic can be replaced by storage call directly making the code cleaner and more readable.

2.0

lsilvs:
### What was the problem?
Unnecessary function declaration when storage could be used.

### How did I fix it?
Replaced redundant code by storage call directly where possible.

`logic.account.merge` is still present and can be replaced but it will be tackled in another issue

### How to test it?
Run test suite

### Review checklist

* The PR resolves #3524
* Linting passes
* Tests pass


framework/src/modules/chain/submodules/blocks/utils.js
```diff
@@ -625,52 +625,53 @@ class Utils {
...
+				params.generatorPublicKey = account.publicKey;
+				params.delegates = ACTIVE_DELEGATES;
 
-			if (filter.start !== undefined) {
-				params.fromTimestamp = Math.floor(
-					(filter.start - new Date(EPOCH_TIME).getTime()) / 1000
-				);
-				params.fromTimestamp = params.fromTimestamp.toFixed();
-			}
+				if (filter.start !== undefined) {
```
yatki: I know it was like this before but maybe `!filter.start` would be a better check, because it would work if filter.start is null :D Same goes below.

framework/src/modules/chain/submodules/blocks/utils.js
```diff
@@ -625,52 +625,53 @@ class Utils {
...
-				.catch(delegateBlocksRewardsErr => {
-					library.logger.error(delegateBlocksRewardsErr.stack);
-					return setImmediate(cb, 'Blocks#aggregateBlocksReward error');
-				});
-		});
+				// Get calculated rewards
+				return library.storage.entities.Account.delegateBlocksRewards(params)
+					.then(rows => {
+						let data = rows[0];
+						if (data.delegate === null) {
```
yatki: `!data.delegate` ?

framework/src/modules/chain/submodules/loader.js
```diff
@@ -473,13 +473,12 @@ __private.loadBlockChain = function() {
...
 				resetMemTables(seriesCb) {
-					library.logic.account.resetMemTables(err => {
-						if (err) {
-							throw err;
-						} else {
-							return setImmediate(seriesCb);
-						}
-					});
+					library.storage.entities.Account.resetMemTables()
+						.then(() => seriesCb())
```
yatki: `.then(seriesCb)` ?

framework/src/modules/chain/submodules/loader.js
```diff
@@ -870,7 +869,12 @@ __private.rebuildAccounts = height => {
 	async.series(
 		{
 			resetMemTables(seriesCb) {
-				library.logic.account.resetMemTables(seriesCb);
+				library.storage.entities.Account.resetMemTables()
+					.then(() => seriesCb())
```
yatki: same above.

framework/test/mocha/unit/modules/chain/submodules/delegates.js
```diff
@@ -140,45 +140,37 @@ describe('delegates', () => {
...
+			const account = await library.components.storage.entities.Account.getOne({
+				publicKey: testDelegate.publicKey,
+			});
+			expect(__private.keypairs[testDelegate.publicKey]).to.not.be.undefined;
+			expect(account.publicKey).to.equal(testDelegate.publicKey);
 
-					expect(__private.keypairs[testDelegate.publicKey]).to.be.undefined;
-					expect(data.publicKey).to.equal(testDelegate.publicKey);
-				}
+			const data = await library.modules.delegates.updateForgingStatus(
```
yatki: Having // Arrange // Act // Assert comments on top of this line would be really nice to differentiate which action is mutating data.

Arrange above line 143
Act above line 149
Assert above line 155

framework/test/mocha/unit/modules/chain/submodules/delegates.js
```diff
@@ -140,45 +140,37 @@ describe('delegates', () => {
...
+				publicKey: testDelegate.publicKey,
+			});
+			expect(__private.keypairs[testDelegate.publicKey]).to.be.undefined;
+			expect(account.publicKey).to.equal(testDelegate.publicKey);
 
-					expect(__private.keypairs[testDelegate.publicKey]).to.not.be
-						.undefined;
-					expect(data.publicKey).to.equal(testDelegate.publicKey);
-				}
+			const data = await library.modules.delegates.updateForgingStatus(
```
yatki: same above.

framework/src/modules/chain/submodules/loader.js
shuse2: this should be better because if then returns something it will be treated as error

framework/src/modules/chain/submodules/loader.js
shuse2: same as above

framework/test/mocha/unit/modules/chain/submodules/delegates.js
shuse2: This test will be rewritten soon, so lets skip this for now.

Issue #3533, opened 2019-05-07
API node/status - invalid networkHeight


4miners:
### Expected behavior
`networkHeight` should be always correct.
### Actual behavior
`networkHeight` is incorrect (for about 1 second) when API called just after forging a block.

<img width="613" alt="Screen Shot 2019-05-07 at 11 26 09" src="https://user-images.githubusercontent.com/22824146/57289383-89868300-70bb-11e9-88f9-8ed37d383409.png">

### Steps to reproduce
Tested on a network with 10 nodes, all of them are on the same height and moving forward, the cache was disabled.
Call API endpoint http://lon1-foonet-000.liskdev.net:4000/api/node/status frequently and observe the results.
### Which version(s) does this affect? (Environment, OS, etc...)
`2.0.0-alpha.0`

ManuGowda:
This issue is fixed and verified in `alpha.8` tag
<img width="835" alt="Screen Shot 2019-05-20 at 11 41 12" src="https://user-images.githubusercontent.com/6449871/58012186-36b5be00-7af4-11e9-90d0-130e4d680af5.png">

PR #3587, opened 2019-05-10
Unstable broadhash due to stale info about peers in `getNetworkStatus()`

### LINKED ISSUE #3574: Unstable broadhash consensus
### Expected behavior
Broadhash consensus should be stable.
### Actual behavior
Broadhash consensus is unstable which leads to missed blocks.
### Steps to reproduce
Tested on network with 10 peers, observe logs.
`2.0.0-alpha.0`

ishantiw:
### What was the problem?

Broadhash consensus is unstable which leads to missed blocks. #3574 

`networkHeight` is incorrect (for about 1 second) when API called just after forging a block. #3533 

### How did I fix it?

- Assigning correct `state` information (`0` for `disconnected` and `2` for `connected`) for `connectedPeers` and `triedPeers` which are not connected.
- Update `triedPeers` info on update status event on p2p-lib
- Don't let the node to add itself in its `newPeers` list

### How to test it?

Test on network with 10 peers, observe logs.

### Review checklist

* The PR resolves #3574 #3533
* Linting passes
* Tests pass

### DIFF
--- a/elements/lisk-p2p/src/p2p.ts
+++ b/elements/lisk-p2p/src/p2p.ts
@@ -194,11 +194,12 @@ export class P2P extends EventEmitter {
 		};
 
 		this._handlePeerConnectAbort = (peerInfo: P2PPeerInfo) => {
-			// Re-emit the message to allow it to bubble up the class hierarchy.
 			const peerId = constructPeerIdFromPeerInfo(peerInfo);
 			if (this._triedPeers.has(peerId)) {
 				this._triedPeers.delete(peerId);
 			}
+
+			// Re-emit the message to allow it to bubble up the class hierarchy.
 			this.emit(EVENT_CONNECT_ABORT_OUTBOUND, peerInfo);
 		};
 
--- a/elements/lisk-p2p/src/p2p.ts
+++ b/elements/lisk-p2p/src/p2p.ts
@@ -451,7 +451,7 @@ export class P2P extends EventEmitter {
 					!(
 						['127.0.0.1', '0.0.0.0', 'localhost'].includes(
 							incomingPeerInfo.ipAddress,
-						) && incomingPeerInfo.wsPort !== this.nodeInfo.wsPort
+						) && incomingPeerInfo.wsPort === this.nodeInfo.wsPort
 					)
 				) {
 					this._newPeers.set(peerId, incomingPeerInfo);
@@ -522,7 +522,7 @@ export class P2P extends EventEmitter {
 				!this._newPeers.has(peerId) &&
 				!(
 					['127.0.0.1', '0.0.0.0', 'localhost'].includes(peerInfo.ipAddress) &&
-					peerInfo.wsPort !== this.nodeInfo.wsPort
+					peerInfo.wsPort === this.nodeInfo.wsPort
 				)
 			) {
 				this._newPeers.set(peerId, peerInfo);
--- a/elements/lisk-p2p/src/p2p.ts
+++ b/elements/lisk-p2p/src/p2p.ts
@@ -172,7 +172,23 @@ export class P2P extends EventEmitter {
 			this.emit(EVENT_MESSAGE_RECEIVED, message);
 		};
 
-		this._handlePeerConnect = (peerInfo: P2PPeerInfo) => {
+		this._handlePeerConnect = (peerInfo: P2PDiscoveredPeerInfo) => {
+			const peerId = constructPeerIdFromPeerInfo(peerInfo);
+			const foundTriedPeer = this._triedPeers.get(peerId);
+			// On successful connection remove it from newPeers list
+			this._newPeers.delete(peerId);
+
+			if (foundTriedPeer) {
+				const updatedPeerInfo = {
+					...peerInfo,
+					ipAddress: foundTriedPeer.ipAddress,
+					wsPort: foundTriedPeer.wsPort,
+				};
+				this._triedPeers.set(peerId, updatedPeerInfo);
+			} else {
+				this._triedPeers.set(peerId, peerInfo);
+			}
+
 			// Re-emit the message to allow it to bubble up the class hierarchy.
 			this.emit(EVENT_CONNECT_OUTBOUND, peerInfo);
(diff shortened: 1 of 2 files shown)


elements/lisk-p2p/src/p2p.ts
```diff
@@ -433,7 +445,15 @@ export class P2P extends EventEmitter {
 					this.emit(EVENT_NEW_PEER, incomingPeerInfo);
 				}
 
-				if (!this._newPeers.has(peerId) && !this._triedPeers.has(peerId)) {
+				if (
+					!this._newPeers.has(peerId) &&
+					!this._triedPeers.has(peerId) &&
+					!(
+						['127.0.0.1', '0.0.0.0', 'localhost'].includes(
```
shuse2: maybe set this as constants?
```
const LOCAL_ADDRESSES = ['127.0.0.1', '0.0.0.0', 'localhost'];
```

ishantiw: Done

elements/lisk-p2p/src/p2p.ts
```diff
@@ -103,6 +103,7 @@ export const EVENT_FAILED_TO_ADD_INBOUND_PEER = 'failedToAddInboundPeer';
 export const EVENT_NEW_PEER = 'newPeer';
 
 export const NODE_HOST_IP = '0.0.0.0';
```
diego-G: This is redundant after introducing `NODE_HOST_IP_LIST.

framework/src/modules/network/filter_peers.js
```diff
@@ -168,18 +168,22 @@ const getCountByFilter = (peers, filter) => {
...
-		);
-
-		if (!found) {
+	// Assign state 2 to the connected peers
+	const connectedList = connectedPeers.map(peer => {
+		const { ipAddress, options, minVersion, nethash, ...peerWithoutIp } = peer;
+
+		return { ip: ipAddress, ...peerWithoutIp, state: 2 };
+	});
+	// For the peers that are not present in connectedList should be assigned state 0
```
diego-G: Shouldn't it be 1 as in previous versions? https://github.com/LiskHQ/lisk-sdk/blob/40e2e86021ad14661dea78932d638a2a5bb53835/framework/src/modules/chain/logic/peer.js#L197

elements/lisk-p2p/src/p2p.ts
ishantiw: We are using `NODE_HOST_IP` only for http server that requires only one ip, if we remove it then it will look something like `this._httpServer.listen(this._nodeInfo.wsPort, NODE_HOST_IP_LIST[1])`
which doesn't look so nice. 

How about we rename them like below?,
`NODE_HOST_IP` => `NODE_HTTP_HOST_IP`
`NODE_HOST_IP_LIST` => `NODE_VALID_HOST_IP_LIST`

framework/src/modules/network/filter_peers.js
ishantiw: Oh yeah need to change disconnected peer's state to `1` 👍

elements/lisk-p2p/src/p2p.ts
```diff
@@ -433,7 +446,14 @@ export class P2P extends EventEmitter {
 					this.emit(EVENT_NEW_PEER, incomingPeerInfo);
 				}
 
-				if (!this._newPeers.has(peerId) && !this._triedPeers.has(peerId)) {
+				if (
+					!this._newPeers.has(peerId) &&
+					!this._triedPeers.has(peerId) &&
+					!(
```
jondubois: Is this new condition needed?
If the node tries to connect to itself, it should be rejected on line `365` and it should never reach this code.

elements/lisk-p2p/src/p2p.ts
```diff
@@ -497,7 +517,14 @@ export class P2P extends EventEmitter {
 
 		discoveredPeers.forEach((peerInfo: P2PDiscoveredPeerInfo) => {
 			const peerId = constructPeerIdFromPeerInfo(peerInfo);
-			if (!this._triedPeers.has(peerId) && !this._newPeers.has(peerId)) {
+			if (
+				!this._triedPeers.has(peerId) &&
+				!this._newPeers.has(peerId) &&
+				!(
+					NODE_HOST_IP_LIST.includes(peerInfo.ipAddress) &&
+					peerInfo.wsPort === this.nodeInfo.wsPort
```
jondubois: This condition wouldn't work in a production environment. The node's own IP will not be `127.0.0.1` (nor `localhost`), it will look like a regular IP address. The only way to know which IP is our own is during the handshake when we get the nonce.

To make it work properly, we would need to add separate logic for self-discovery and the node would store its own IP address somewhere.

Given the amount of complexity involved, I think it would be simpler to just allow the node to add itself as a peer.

elements/lisk-p2p/src/p2p.ts
ishantiw: That's true but somehow it's still being added to `newPeers` list reported by @shuse2 so at some point the nonce is different for the same IP, is it because we are having two nonces (network and config)?

elements/lisk-p2p/src/p2p.ts
ishantiw: It's just an extra check to avoid adding these IPs in `newPeers` list which are showing up sometimes, if it adds itself with its public IP then it's fine

elements/lisk-p2p/src/p2p.ts
ishantiw: Removed this condition as discussed, as it will be covered by nonce condition already

elements/lisk-p2p/src/p2p.ts
ishantiw: same as above

elements/lisk-p2p/src/p2p.ts
ishantiw: removed this condition as discussed, so it isn't relevant anymore

PR #3589, opened 2019-05-10
Fix broadcaster usage for transaction - Closes #3588

### LINKED ISSUE #3588: Transactions are not broadcasted after received by postTransactions
### Expected behavior
Transactions should be broadcasted to rest of the network after received by API.
### Actual behavior
They are not.
### Steps to reproduce
Send the transaction to a node that has forging disabled, check for its presence on other nodes.
`2.0.0-alpha.0`

shuse2:
### What was the problem?
Syntax of the awaiting the promise was wrong, and it was to parsed before queuing

### How did I fix it?
parse to normal object before queueing and use correct syntax for waiting promise

### How to test it?
run 2 nodes and send a transaction

### Review checklist

* The PR resolves #3588 
* Linting passes
* Tests pass


framework/src/modules/chain/logic/broadcaster.js
```diff
@@ -253,13 +253,11 @@ class Broadcaster {
...
 			const squashedBroadcasts = this.squashQueue(broadcasts);
 
-			// eslint-disable-next-line no-restricted-syntax
-			for await (const squashedBroadcast of squashedBroadcasts) {
-				return this.broadcast(
-					squashedBroadcast.params,
-					squashedBroadcast.options
-				);
-			}
+			await Promise.all(
```
diego-G: This is not necessary.

framework/src/modules/chain/submodules/transport.js
```diff
@@ -306,16 +306,17 @@ Transport.prototype.onUnconfirmedTransaction = function(
...
 ) {
 	if (broadcast && !__private.broadcaster.maxRelays(transaction)) {
+		const transactionJSON = transaction.toJSON();
 		__private.broadcaster.enqueue(
 			{},
 			{
 				api: 'postTransactions',
 				data: {
-					transaction,
+					transaction: transactionJSON,
```
diego-G: TBH, I'm not so sure what we are fixing here. Anyways, in the case we do fix something, we should perform the same action in other places like in here https://github.com/LiskHQ/lisk-sdk/blob/61969851e484a18e119810851177f85091c3e774/framework/src/modules/chain/submodules/transport.js#L288

shuse2: signature is always an object so it should not need the toJSON

framework/src/modules/chain/logic/broadcaster.js
shuse2: the usage of the `for await` should be for async generator or async iterables

PR #3598, opened 2019-05-13
API returns incorrect error for postTransactions - Closes #3586

### LINKED ISSUE #3586: API returns incorrect error for postTransactions
### Expected behavior
API should return the correct error message.
### Actual behavior
API always returns `Invalid transaction body` error even when the transaction failed from a different reason.
```
10:01:51.954Z  WARN lisk-framework: Balance queue 775
10:01:51.954Z DEBUG lisk-framework: Received transaction 1944531498676575052 from public client
10:01:51.960Z DEBUG lisk-framework: Transaction 1944531498676575052 Error: Transaction pool is full
```
```
{ APIError: Invalid transaction body
       at axios_1.default.request.then.catch (lisk-core/qa/node_modules/@liskhq/lisk-api-client/dist-node/api_resource.js:43:27)
       at process._tickCallback (internal/process/next_tick.js:68:7)
     jse_shortmsg: 'Invalid transaction body',
     jse_info: {},
     message: 'Invalid transaction body',
     name: 'APIError',
     errno: 409,
     errors: {} }
```
### Steps to reproduce
Send transaction when the transaction pool is full.
`2.0.0-alpha.0`

pablitovicente:
### What was the problem?

`submodules/transport.js` was always returning `Invalid transaction body`  as error message

### How did I fix it?

Returning original `err` which was passed to the method

### How to test it?

- Build should be green
- Send additional transactions when the transaction pool is ull

### Review checklist

* The PR resolves #3586
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/chain/submodules/transport.js
+++ b/framework/src/modules/chain/submodules/transport.js
@@ -709,7 +709,7 @@ Transport.prototype.shared = {
 				if (err) {
 					return setImmediate(cb, null, {
 						success: false,
-						message: 'Invalid transaction body',
+						message: err.message || 'Invalid transaction body',
 						errors: err,
 					});
 				}
--- a/framework/test/mocha/unit/modules/chain/submodules/transport.js
+++ b/framework/test/mocha/unit/modules/chain/submodules/transport.js
@@ -1731,6 +1731,31 @@ describe('transport', () => {
 								.which.is.equal(receiveTransactionError);
 						});
 					});
+
+					describe('when __private.receiveTransaction fails with "Transaction pool is full"', () => {
+						const receiveTransactionError = 'Transaction pool is full';
+
+						beforeEach(done => {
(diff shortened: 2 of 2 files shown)


4miners:
@pablitovicente Could you also check for other similar issues, like for example `Invalid signatures body`?

pablitovicente:
> @pablitovicente Could you also check for other similar issues, like for example `Invalid signatures body`?

Yes it should be cover by the fix but will send different transactions types to test.

pablitovicente:
> > @pablitovicente Could you also check for other similar issues, like for example `Invalid signatures body`?
> 
> Yes it should be cover by the fix but will send different transactions types to test.

@shuse2 @4miners I have additionally checked the method `receiveSignatures()` in transport and it always returns errors without changing them. Additionally the signatures use a transaction with a place already reserved in the transaction pool so the scenario would never be valid with the current implementation when posting signatures.

PR #3599, opened 2019-05-13
For future compatibility, we need to allow receiving RPCs and events on the outbound socket - Closes #3582

### LINKED ISSUE #3582: For future compatibility, we need to allow receiving RPCs and events on the outbound socket
### Expected behavior

Outbound connections on the Peer class should be able to handle RPCs and events.

### Actual behavior

Right now the lisk-p2p module only handles RPCs and remote events on the inbound socket.
The node should be also able to handle RPCs and remote events on its outbound socket because according to the LIP protocol, a node will usually only have an inbound or outbound connection but not both.

jondubois:
### What was the problem?

Outbound sockets could not receive RPCs or remote events.

### How did I fix it?

Added RPC and remote event handlers on the outbound socket.

### How to test it?

Since we don't currently use this feature (because it's for the next LIP version of the protocol), it didn't make sense to add a test case for it. To test, you can temporarily change the code in the request and send methods on the `Peer` class to use the `_inboundSocket` instead of the `_outboundSocket` then run the tests and they should still all pass.

### Review checklist

* The PR resolves #3582
* Linting passes
* Tests pass

### DIFF
--- a/elements/lisk-p2p/src/peer.ts
+++ b/elements/lisk-p2p/src/peer.ts
@@ -510,6 +510,19 @@ export class Peer extends EventEmitter {
 				reason,
 			});
 		});
+
+		// Bind RPC and remote event handlers
+		outboundSocket.on(REMOTE_EVENT_RPC_REQUEST, this._handleRawRPC);
+		outboundSocket.on(REMOTE_EVENT_MESSAGE, this._handleRawMessage);
+		outboundSocket.on('postBlock', this._handleRawLegacyMessagePostBlock);
+		outboundSocket.on(
+			'postSignatures',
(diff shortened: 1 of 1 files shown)


jondubois:
@diego-G for inbound sockets, by the time we receive the socket inside `p2p.js`, the socket is already fully connected so there is no point listening to the `'connect'` or `'connectionAbort'` events.

That said, for the 'close' events, for consistency I agree that we should handle and propagate the inbound socket `'close'` event.

I've raised a new issue: https://github.com/LiskHQ/lisk-sdk/issues/3602
We don't currently use those events right now so it's not a blocker for us but it's easy to fix.

PR #3604, opened 2019-05-13
Propagate inbound socket close event on Peer to the top level P2P class - Closes #3602

### LINKED ISSUE #3602: Propagate inbound socket close event on Peer to the top level P2P class
### Expected behavior

We should be listening to and propagating close events from inbound sockets.

### Actual behavior

We are listening for close events on the outbound socket `EVENT_CLOSE_OUTBOUND` but not doing the same for inbound sockets on peers. There is no `EVENT_CLOSE_INBOUND`.

jondubois:
### What was the problem?

- The inbound socket close event was not being emitted on top level P2P class.
- We were removing the peer from the peer pool if the outbound socket was disconnected; this is fine for v2.0 but not suitable for v2.1 LIP protocol version since peers will no longer always be both inbound and outbound.

### How did I fix it?

- Propagated 'close' event to P2P class.
- For connections state, use constants that are consistent with WebSocket connection states.
- Used strings as enums for connection states for ease of debugging.
- Only remove a peer from the PeerPool if it is disconnected both inbound and outbound.

### How to test it?

Run `lisk-p2p` tests.

### Review checklist

* The PR resolves #3602
* Linting passes
* Tests pass

### DIFF
--- a/elements/lisk-p2p/src/p2p.ts
+++ b/elements/lisk-p2p/src/p2p.ts
@@ -66,6 +66,7 @@ export { P2PRequest };
 import { selectForConnection, selectPeers } from './peer_selection';
 
 import {
+	EVENT_CLOSE_INBOUND,
 	EVENT_CLOSE_OUTBOUND,
 	EVENT_CONNECT_ABORT_OUTBOUND,
 	EVENT_CONNECT_OUTBOUND,
@@ -84,6 +85,7 @@ import {
 import { checkPeerCompatibility } from './validation';
 
 export {
+	EVENT_CLOSE_INBOUND,
 	EVENT_CLOSE_OUTBOUND,
 	EVENT_CONNECT_ABORT_OUTBOUND,
 	EVENT_CONNECT_OUTBOUND,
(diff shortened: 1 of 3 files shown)


elements/lisk-p2p/src/peer.ts
```diff
@@ -88,9 +89,9 @@ type SCServerSocketUpdated = {
 } & SCServerSocket;
 
 export enum ConnectionState {
-	CONNECTING = 0,
-	CONNECTED = 1,
-	DISCONNECTED = 2,
+	CONNECTING = 'connecting',
```
shuse2: why change this to string now?

elements/lisk-p2p/src/p2p.ts
```diff
@@ -186,11 +189,16 @@ export class P2P extends EventEmitter {
...
 		};
 
-		this._handlePeerClose = (closePacket: P2PClosePacket) => {
+		this._handlePeerCloseOutbound = (closePacket: P2PClosePacket) => {
 			// Re-emit the message to allow it to bubble up the class hierarchy.
 			this.emit(EVENT_CLOSE_OUTBOUND, closePacket);
 		};
 
+		this._handlePeerCloseInbound = (closePacket: P2PClosePacket) => {
+			// Re-emit the message to allow it to bubble up the class hierarchy.
```
mitsujutsu: Is it re-emitted up a hierarchy or to the network module?

elements/lisk-p2p/src/peer.ts
jondubois: To make debugging simpler. Also it's not part of the protocol; it's only used internally by the P2P library so it's non-breaking.

PR #3605, opened 2019-05-14
Refactor entry modules - Closes #3580

### LINKED ISSUE #3580: Move state management and entry modules to the top level
### Description

- Loader
- Delegates
- Transport
should be moved to the root level, and all the repeating task based on time should be managed in the chain.js to avoid circular dependencies.

shuse2:
### What was the problem?
Entry modules was in the same level as other modules and had circular dependency

### How did I fix it?
Move 3 entry modules to root level, delegates, transport and loader.
Also, split delegates management and forging into 2 files.

### How to test it?
- Run the tests
- start the application

### Review checklist

* The PR resolves #3580 
* Linting passes
* Tests pass


framework/src/modules/chain/chain.js
```diff
@@ -152,6 +155,19 @@ module.exports = class Chain {
 			this.scope.bus = await createBus();
 			this.scope.logic = await initLogicStructure(this.scope);
 			this.scope.modules = await initModules(this.scope);
+			// TODO: Global variable forbits to require on top
+			const Loader = require('./loader');
+			const { Forge } = require('./forge');
```
nazarhussain: `forge` is a verb not correct for a module name. Rename it to `forger`

framework/src/modules/chain/chain.js
```diff
@@ -152,6 +155,19 @@ module.exports = class Chain {
...
+			const { Delegates } = require('./submodules/delegates');
+			const Transport = require('./transport');
+			this.loader = new Loader(this.scope);
+			this.forge = new Forge(this.scope);
+			this.transport = new Transport(this.scope);
+			// TODO: should not add to scope
+			this.scope.modules.delegates = new Delegates(this.scope);
+			this.scope.modules.loader = this.loader;
+			this.scope.modules.forge = this.forge;
+			this.scope.modules.transport = this.transport;
```
nazarhussain: If these are modules then what's wrong to load and return from in `init_modules` step?

framework/src/modules/chain/chain.js
```diff
@@ -163,7 +179,14 @@ module.exports = class Chain {
 			// Fire onBind event in every module
 			this.scope.bus.message('bind', this.scope);
 
-			self.logger.info('Modules ready and launched');
+			// After binding, it should immeately load blockchain
```
nazarhussain:
```suggestion
			// After binding, it should immediately load blockchain
```

framework/src/modules/chain/chain.js
```diff
@@ -163,7 +179,14 @@ module.exports = class Chain {
...
+			await this.loader.loadBlockChain();
+			this.logger.info('Modules ready and launched');
```
nazarhussain: This seems the wrong order. Blockchain is ready or modules?

framework/src/modules/chain/loader.js
```diff
@@ -17,28 +17,22 @@
 const async = require('async');
```
nazarhussain: I don't understand how moving different modules of chain to one level up make any difference. Same you can do but keeping the files where these belongs to.

framework/src/modules/chain/chain.js
```diff
@@ -295,4 +318,64 @@ module.exports = class Chain {
 
 		this.logger.info('Cleaned up successfully');
 	}
+
+	_startLoader() {
```
nazarhussain: Move these actions to their respective modules, don't thick the initialization logic file.

framework/src/modules/chain/chain.js
```diff
@@ -295,4 +318,64 @@ module.exports = class Chain {
...
+						if (syncError) {
+							this.logger.error('Sync timer', syncError);
+						}
+					}
+				);
+			}
+		}, syncInterval);
+	}
+
+	async _startForge() {
```
nazarhussain: Same here

framework/src/modules/chain/submodules/blocks/process.js
```diff
@@ -347,28 +347,11 @@ class Process {
 	/**
 	 * Handle newly received block.
 	 *
-	 * @listens module:transport~event:receiveBlock
 	 * @param {block} block - New block
 	 * @todo Add @returns tag
 	 */
 	// eslint-disable-next-line class-methods-use-this
-	onReceiveBlock(block) {
```
nazarhussain: Are we going to remove this `on<Event>` convention? If that's the case then there are many `onBind` handlers in the code.

framework/test/mocha/unit/modules/chain/transport.js
```diff
@@ -23,21 +23,19 @@ const {
...
+const Block = require('../../../fixtures/blocks').Block;
 const {
 	registeredTransactions,
-} = require('../../../../common/registered_transactions');
-const InitTransaction = require('../../../../../../src/modules/chain/logic/init_transaction');
-const ProcessTransactions = require('../../../../../../src/modules/chain/submodules/process_transactions');
-const processTransactionLogic = require('../../../../../../src/modules/chain/logic/process_transaction');
+} = require('../../../common/registered_transactions');
+const InitTransaction = require('../../../../../src/modules/chain/logic/init_transaction');
+const ProcessTransactions = require('../../../../../src/modules/chain/submodules/process_transactions');
```
nazarhussain: `ProcessTransactions` is really not a good name for a module or a class. It should be renamed to `TransactionsProcessor` or `TransactionsLoader`

framework/src/modules/chain/chain.js
shuse2: It's still temporally, and I want to remove this `init_modules` step afterwards

framework/src/modules/chain/loader.js
shuse2: For more clarity of dependency tree, but a lot more coming afterwards

framework/src/modules/chain/chain.js
shuse2: It's not initialisation, chain.js will be orchestrating the modules

framework/src/modules/chain/submodules/blocks/process.js
shuse2: yes we are going to remove `bus` in general, but step by step

framework/test/mocha/unit/modules/chain/transport.js
shuse2: This will be change in coming PRs

framework/src/modules/chain/chain.js
```diff
@@ -152,6 +155,19 @@ module.exports = class Chain {
 			this.scope.bus = await createBus();
 			this.scope.logic = await initLogicStructure(this.scope);
 			this.scope.modules = await initModules(this.scope);
+			// TODO: Global variable forbits to require on top
+			const Loader = require('./loader');
```
pablitovicente: - I think `requires` would be better at the top level of the file
- giving that this method gets too extensive maybe create a method `loadDependencies()` that does the same but then returns the value and here it's only assigned to this?

framework/src/modules/chain/chain.js
pablitovicente: `_startForging` maybe ?

framework/src/modules/chain/chain.js
shuse2: Let's do this loadDependencies at the end of the technical debt because it will change a lot

PR #3620, opened 2019-05-14
Fix transaction related dependency - Closes #3614

### LINKED ISSUE #3614: Combine all the transaction related logic into one module and expose used functionality
### Expected behavior
Within transactions module, it should not have any external dependency, and only exposing the used function from other module

- init_transaction.js
- process_transaction.js
- check_transaction_against_exceptions.js
- broadcaster.js
- transaction_pool.js
- transactions.js
- process_transactions.js
- multisignatures.js

### Actual behavior
Transaction logic spread across the chain module

shuse2:
### What was the problem?
Transaction related logic had dependency everywhere and logic was scattered.

### How did I fix it?
- Transaction pool depends on blocks
- blocks still temporally depends on transaction pool
- removed proxy `transaction.js` and move API related logic to node status (only used for node status endpoint)
- Removed dependency of blocks (_getCurrentContext) from process_transactions
- Changed process transactions to function
- Directly expose transaction pool as module
- Skipped/replaced dapp related/using tests
- Removed all callbacks in transactions and transaction pool

### How to test it?
- Run the application
- Check the changed tests carefully
- Run the tests

### Review checklist

* The PR resolves #3614  
* Linting passes
* Tests pass


framework/src/modules/chain/transactions/votes.js
```diff
@@ -89,7 +93,7 @@ const updateSenderRoundInformationWithAmountForTransaction = function(
 			.map(vote => vote.slice(1));
 
 		// Votes inside the transaction should not be incase it's an exception, but the existing votes should be updated
-		if (!exceptions.roundVotes.includes(transaction.id)) {
+		if (!(exceptions.roundVotes || []).includes(transaction.id)) {
```
SargeKhan: exceptions default configuration should set empty array for roundVotes property.

framework/src/modules/chain/loader.js
```diff
@@ -570,13 +566,16 @@ __private.getTransactionsFromNetwork = async function() {
...
-				modules.transactions.processUnconfirmedTransaction(
-					transaction,
-					false,
-					addSequenceCb
-				);
+				try {
+					await modules.transactionPool.processUnconfirmedTransaction(
+						transaction
+					);
+					setImmediate(addSequenceCb);
```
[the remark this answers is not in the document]
shuse2: I think usage of balanceSequenceAdd is wrong, and it's from 2.0- 😢

framework/src/modules/chain/loader.js
```diff
@@ -541,9 +537,9 @@ __private.getTransactionsFromNetwork = async function() {
 
 	const transactions = result.transactions;
 	try {
-		const {
-			transactionsResponses,
-		} = modules.processTransactions.validateTransactions(transactions);
+		const { transactionsResponses } = validateTransactions(exceptions)(
```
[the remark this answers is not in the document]
shuse2: in the process of validateTransactions, it calls the exception_handler functions, and it excepts to exists.
Maybe I should change the exceptions handler to accepts undefined keys?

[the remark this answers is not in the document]
SargeKhan: i was suggesting to pass an empty object here. It makes more sense because no transaction should be considered an exception when they are loaded from network.

framework/src/modules/chain/chain.js
```diff
@@ -319,6 +307,53 @@ module.exports = class Chain {
 		this.logger.info('Cleaned up successfully');
 	}
 
+	async _initModules() {
```
lsilvs: Can't we move all this logic to `./init_steps/init_modules`?

framework/src/modules/chain/submodules/blocks/chain.js
```diff
@@ -28,6 +32,7 @@ let modules;
 let library;
 let self;
 const __private = {};
+const exceptions = global.exceptions;
```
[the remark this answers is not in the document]
shuse2: we are going to remove them in the block PR

framework/src/modules/chain/init_steps/init_logic_structs.js
```diff
@@ -33,27 +23,15 @@ module.exports = async ({
 			err ? reject(err) : resolve(object);
 		});
 	});
-	const initTransactionLogic = new InitTransaction({ registeredTransactions });
 
 	const blockLogic = await new Promise((resolve, reject) => {
-		new Block(ed, schema, initTransactionLogic, (err, object) => {
+		new Block(ed, schema, (err, object) => {
```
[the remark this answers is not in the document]
shuse2: we are going to remove them in the block PR

framework/src/modules/chain/chain.js
shuse2: we can move to the init_modules file, but it will not be dynamic anymore.
so probably it's easier to access all the required parameters here

shuse2:
I have separated `transaction_manager` to different folder and renamed it to `interface_adapters/transaction_interface_adapter.js` after discussion with @SargeKhan

framework/src/modules/chain/submodules/blocks/chain.js
```diff
@@ -64,11 +69,13 @@ class Chain {
 			balancesSequence,
 			logic: {
 				block,
-				initTransaction,
 			},
 			channel,
 		};
 		self = this;
+		self.modules = {
```
SargeKhan: Can we assign interfaceAdapters to a different variable name. Lest we will have `self.modules` and `module` variables which have different values.

framework/src/modules/chain/transport.js
```diff
@@ -77,7 +77,7 @@ class Transport {
 			scope.config.nonce,
 			scope.config.broadcasts,
 			scope.config.forging.force,
-			scope.logic.transactionPool,
+			scope.modules.transactionPool,
```
SargeKhan: Why do we bind transactionPool in the onBind function this file already receives it in the constructor?

framework/src/modules/chain/submodules/blocks/chain.js
shuse2: true, but this will be changed in the coming block PR

framework/src/modules/chain/transport.js
shuse2: This is because I didn't want to touch the bind method (since it will be removed), and it will break the test ='(
forgers, transport and loader needs to be refactored after the block PR

PR #3626, opened 2019-05-15
Node fail to sync when block contains multisignature transaction - Closes #3612

### LINKED ISSUE #3612: Node fail to sync when block contains multisignature transaction
### Expected behavior
The node should be able to sync when a block contains multisignature registration transaction.
### Actual behavior
It fails with following message:
```
14:03:52.617Z  INFO lisk-framework: Block 5753985202868417271 loaded from the network height: 17
14:03:52.625Z DEBUG lisk-framework:
    Block processing failed { id: '1354458273227907673',
      err:
       'Transaction: 4767118963976999553 failed at .multisignature.keysgroup: \'.multisignature.keysgroup\' should NOT have fewer than 1 items',
      module: 'blocks',
      block:
       { id: '1354458273227907673',
         version: 1,
         timestamp: 93719360,
         height: 18,
         previousBlock: '5753985202868417271',
         numberOfTransactions: 10,
         totalAmount: 0,
         totalFee: 15000000000,
         reward: 0,
         payloadLength: 3130,
         payloadHash:
          '7377aae41a8af606f505d18c63ce6994bf04c1844f192a87c822aaf35e6ca526',
         generatorPublicKey:
          '85b07e51ffe528f272b7eb734d0496158f2b0f890155ebe59ba2989a8ccc9a49',
         generatorId: '6214967903930344618L',
         blockSignature:
          'f8670420bd65a232723d6e1b4de4266e9fe05e67ecb1821a7e4d53c5c0cd31a2da048899631b70fff50f2ff1f774a42ec9470f14b30d2861a7acefa55894ee08',
         confirmations: NaN,
         totalForged: '15000000000',
         transactions:
          [ [MultisignatureTransaction],
            [MultisignatureTransaction],
            [MultisignatureTransaction],
            [MultisignatureTransaction],
            [MultisignatureTransaction],
            [MultisignatureTransaction],
            [MultisignatureTransaction],
            [MultisignatureTransaction],
            [MultisignatureTransaction],
            [MultisignatureTransaction] ] } }
14:03:52.628Z DEBUG lisk-framework: Perform chain recovery due to poor consensus
14:03:52.628Z  WARN lisk-framework: Chain comparison failed, starting recovery
```
### Steps to reproduce
Check logs on `fra1-alpha2-001.liskdev.net` machine.
`2.0.0-alpha.2`

SargeKhan:
### What was the problem?
The node did not return the block with the correct type for multisignature keys group property. Previously, it was a comma-separated list of public keys with + sign. Whereas now, it was returning an array of publicKey with + sign.
### How did I fix it?
Fixed the mapper to match the correct property type.
### How to test it?
Sync with the node.
### Review checklist

* The PR resolves #3612 
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/chain/submodules/blocks/utils.js
+++ b/framework/src/modules/chain/submodules/blocks/utils.js
@@ -510,7 +510,10 @@ Utils.prototype._parseStorageObjToLegacyObj = function(block) {
 			v_votes: t.asset && t.asset.votes ? t.asset.votes.join(',') : null,
 			m_min: _.get(t, 'asset.multisignature.min', null),
 			m_lifetime: _.get(t, 'asset.multisignature.lifetime', null),
-			m_keysgroup: _.get(t, 'asset.multisignature.keysgroup', null),
+			m_keysgroup:
+				t.asset && t.asset.multisignature && t.asset.multisignature.keysgroup
+					? t.asset.multisignature.keysgroup.join(',')
+					: null,
 			dapp_name: _.get(t, 'asset.dapp.name', null),
 			dapp_description: _.get(t, 'asset.dapp.description', null),
 			dapp_tags: _.get(t, 'asset.dapp.tags', null),


SargeKhan:
I am confused for the same reason. In v.2.0.0, when fetching blocks, the multisignature transactions are returned in this format:
```
{
    b_id: '1354458273227907673',
    b_version: 1,
    b_timestamp: 93719360,
    b_height: 18,
    b_previousBlock: '5753985202868417271',
    b_numberOfTransactions: 10,
    b_totalAmount: '0',
    b_totalFee: '15000000000',
    b_reward: '0',
    b_payloadLength: 3130,
    b_payloadHash: '7377aae41a8af606f505d18c63ce6994bf04c1844f192a87c822aaf35e6ca526',
    b_generatorPublicKey: '85b07e51ffe528f272b7eb734d0496158f2b0f890155ebe59ba2989a8ccc9a49',
    b_blockSignature: 'f8670420bd65a232723d6e1b4de4266e9fe05e67ecb1821a7e4d53c5c0cd31a2da048899631b70fff50f2ff1f774a42ec9470f14b30d2861a7acefa55894ee08',
    t_id: '9245873320649117706',
    t_type: 4,
    t_timestamp: 93719351,
    t_senderPublicKey: '9880c2140e89e786e01aed10853069486f09e30241c80793a6ce94267663aea9',
    t_senderId: '12155606898691001926L',
    t_recipientId: null,
    t_amount: '0',
    t_fee: '1500000000',
    t_signature: '2bb9abbecccbc328a4a6d42f077d472673b35e07e0eec90fc0b74d990115d024ba9e59e91d8591f9026a5c6daa747da34b3f86b9bc191eb2df508ed12931da0c',
    t_signSignature: 'b8f4761db9871eac57e20214bb12fb520633e4c4088442e8c031d45254062b0164e266ff3d222f24ecdcac3f3b6a9ea92469bae36894fd764dfab435202a4708',
    t_requesterPublicKey: null,
    t_signatures: '1b6589cfe17f6f4198a046a7a3d7a350441f0c9ba40b134972ad225970b3fa8a7c57bafb0216076013ad5856dcf8d340a39b05c60a171b1e48e357b23196eb0f,2e28ca7a790bdba72bb59823579a0cd769a29acb16f5c244654422c2f873b8c3be4562059e4f5ea15897870131d21ab2ad23e3fc86859e1243949796b3cb0407',
    tf_data: null,
    s_publicKey: null,
    d_username: null,
    v_votes: null,
    m_min: 2,
    m_lifetime: 1,
    m_keysgroup: [
      '+bdc21ac0e253b609325dcf4153bc985ec8791c515b84fdad4051351b068e8ef5',
      '+4b83ca58a52001d0a9871de334f21e46883f350b27481c98e871dc8da558ace9'
    ],
    dapp_name: null,
    dapp_description: null,
    dapp_tags: null,
    dapp_type: null,
    dapp_link: null,
    dapp_category: null,
    dapp_icon: null,
    in_dappId: null,
    ot_dappId: null,
    ot_outTransactionId: null
  }
```
And the definition of dbRead for multisig transaction in v.1.6 to v.2.0 (The definition in v.2.0 is moved to lisk-transactions in elements but the behaviour of both of these functions is same) is as follow:
```
Multisignature.prototype.dbRead = function(raw) {
	if (!raw.m_keysgroup) {
		return null;
	}
	const multisignature = {
		min: raw.m_min,
		lifetime: raw.m_lifetime,
	};

	if (typeof raw.m_keysgroup === 'string') {
		multisignature.keysgroup = raw.m_keysgroup.split(',');
	} else {
		multisignature.keysgroup = [];
	}

	return { multisignature };
};
```
Note that in the case of m_keysgroup is an array, the transaction will have an empty keysgroup. 

Therefore, if the definition of reading multisignature transaction is not changed, then it must be concluded that loading blocks from peers were returning data in a different/new format. Hence, I made the change here.

ManuGowda:
> After investigating, it seems to me that syncing from a node running a version greater than 1.5 never worked.
> 
> @ManuGowda I suggest syncing old versions with new version and new version with old versions during every release to make sure it works properly

@lsilvs when we test syncing and snapshotting against a network, the network always has a mix of old and new versions, so the test cases cover all the scenario, and in a real scenario when we release any new version it does the same.

SargeKhan:
@ManuGowda a diverse network of different versions will not reproduce this bug. It will not because of the fact that even if one version doesn't follow protocol, the node may connect with a different peer and sync with the chain.
Consequently, I agree with @lsilvs that we should add a new test scenario for syncing in QA.

PR #3647, opened 2019-05-16
Change log level of relevant network events to debug - Closes #3623

### LINKED ISSUE #3623: Redundant log info without any context
### Expected behavior
Please make sure the rite log level for the logs, and add the context to message, otherwise, it will be redundant and of no use.
If you think it's not of use remove them rather than keep a log which doesn't add any value.
### Actual behavior
Lisk log has redundant log information and at times without any context, the following are such logs

1. `INFO lisk-framework: Received inbound request for procedure updateMyself`, this log information shouldn't be `info` level and also if we want to keep this log, we need to add the context like `Received inbound request from peer 104.248.130.161  for procedure updateMyself`
2. `INFO lisk-framework: Received inbound message for event postTransactions`, same with this log, no context.
3. `INFO lisk-framework: Received inbound message for event postBlock`
4. `INFO lisk-framework: Received inbound request for procedure list`
2.0.0

ishantiw:
### What was the problem?

Too many logs coming from the network module's events that are more appropriate for the debug log level.

### How did I fix it?

Change the log level of these events to debug.

### How to test it?

Start the application and check the logs

### Review checklist

* The PR resolves #3623 
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/network/network.js
+++ b/framework/src/modules/network/network.js
@@ -203,7 +203,7 @@ module.exports = class Network {
 		});
 
 		this.p2p.on(EVENT_REQUEST_RECEIVED, async request => {
-			this.logger.info(
+			this.logger.trace(
 				`Incoming request event: Received inbound request for procedure ${
 					request.procedure
 				}`
--- a/framework/src/modules/network/network.js
+++ b/framework/src/modules/network/network.js
@@ -142,29 +142,33 @@ module.exports = class Network {
 
 		this.p2p.on(EVENT_CLOSE_OUTBOUND, closePacket => {
 			this.logger.debug(
-				`Outbound connection of peer ${closePacket.peerInfo.ipAddress}:${
-					closePacket.peerInfo.wsPort
-				} was closed with code ${closePacket.code} and reason: ${
-					closePacket.reason
-				}`
+				`Peer disconnect event: Outbound connection of peer ${
+					closePacket.peerInfo.ipAddress
+				}:${closePacket.peerInfo.wsPort} was closed with code ${
+					closePacket.code
+				} and reason: ${closePacket.reason}`
 			);
 		});
 
 		this.p2p.on(EVENT_CONNECT_OUTBOUND, peerInfo => {
 			this.logger.info(
-				`Connected to peer ${peerInfo.ipAddress}:${peerInfo.wsPort}`
+				`Peer connect event: Connected to peer ${peerInfo.ipAddress}:${
+					peerInfo.wsPort
+				}`
 			);
 		});
 
 		this.p2p.on(EVENT_DISCOVERED_PEER, peerInfo => {
 			this.logger.info(
-				`Discovered peer ${peerInfo.ipAddress}:${peerInfo.wsPort}`
+				`New peer found event: Discovered peer ${peerInfo.ipAddress}:${
+					peerInfo.wsPort
+				}`
 			);
 		});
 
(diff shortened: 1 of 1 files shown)


framework/src/modules/network/network.js
```diff
@@ -216,11 +222,15 @@ module.exports = class Network {
 					sanitizedProcedure,
 					request.data
 				);
-				this.logger.info(`Responsed to peer request ${request.procedure}`);
+				this.logger.debug(
```
jondubois: This one should happen very often so maybe it should be `logger.info(...)`?
Also request above is `info` so the response is even less important so should be `info`.

framework/src/modules/network/network.js
```diff
@@ -229,7 +239,11 @@ module.exports = class Network {
 		});
 
 		this.p2p.on(EVENT_MESSAGE_RECEIVED, async packet => {
-			this.logger.info(`Received inbound message for event ${packet.event}`);
+			this.logger.debug(
```
jondubois: Also happens very often; maybe `logger.info`?

framework/src/modules/network/network.js
```diff
@@ -216,11 +222,15 @@ module.exports = class Network {
 					sanitizedProcedure,
 					request.data
 				);
-				this.logger.info(`Responsed to peer request ${request.procedure}`);
+				this.logger.debug(
+					`Peer request fulfilled event: Responsed to peer request ${
```
jondubois: Should be `Responded` instead of `Responsed`.

framework/src/modules/network/network.js
```diff
@@ -229,7 +239,11 @@ module.exports = class Network {
 		});
 
 		this.p2p.on(EVENT_MESSAGE_RECEIVED, async packet => {
-			this.logger.info(`Received inbound message for event ${packet.event}`);
+			this.logger.debug(
+				`Message Received event: Received inbound message for event ${
```
jondubois: The second `Received` does not need to be capitalized.

framework/src/modules/network/network.js
ManuGowda: if it happens very often shouldn't it be `trace`? otherwise we will see so many logs at info and debug levels, @fix whats your thought here?

framework/src/modules/network/network.js
```diff
@@ -187,8 +191,8 @@ module.exports = class Network {
 		});
 
 		this.p2p.on(EVENT_UPDATED_PEER_INFO, peerInfo => {
-			this.logger.info(
-				`Updated info of peer ${peerInfo.ipAddress}:${
+			this.logger.debug(
```
jondubois: Log level should be lowered here as well.

ishantiw:
@ManuGowda @jondubois I have changed the log level to trace as discussed for less trivial events

framework/src/modules/network/network.js
jondubois: Sorry I meant the first :(

framework/src/modules/network/network.js
```diff
@@ -200,7 +204,9 @@ module.exports = class Network {
 
 		this.p2p.on(EVENT_REQUEST_RECEIVED, async request => {
 			this.logger.info(
```
jondubois: This one should be trace too because it's very often.
Also `EVENT_MESSAGE_RECEIVED` is similar so it should be trace.

framework/src/modules/network/network.js
ishantiw: changed it to trace

PR #3650, opened 2019-05-16
Share a single ws connection between peers - Closes #3342

### LINKED ISSUE #3342: Share WS connection between Peers
### Expected behavior
2 Peers will always share the same WS connection. Both for client and server calls. 
https://github.com/LiskHQ/lisk/blob/development/api/ws/rpc/connect.js
Contains logic responsible for establishing Client WS connection. It should be moved and merged with establishing Server WS connection - https://github.com/LiskHQ/lisk/blob/development/api/ws/workers/peers_update_rules.js.
Remaining logic of `connect.js` - `registerRPC` should redirect all `peer.rpc` client calls to the corresponding worker which will handle the request and return the results. A corresponding worker will be identified by storing `workerId` field - worker process which is responsible for P2P communication with selected Peer.

Advantages:
- One source of truth about the Peers WS connections stored in the worker,
- Easier to control connect/disconnect peers actions handling,
- Decreased resources used by P2P WS connection,
- Logic ensuring the existence of 1 WS connection per-Peer only.

### Actual behavior
- Client WS are stored as part of `peer` object on Master process. Server Connections are stored on worker process. Data integrity between processes needs to be kept which seems to be a problem currently.

1.0.0

diego-G:
### What was the problem?

Every couple of peers created two sockets (bidirectional by nature) for the same purpose that we can achieve with only one.

### How did I fix it?

Creating two new classes (OutboundPeer & InboundPeer) which extends Peer class.

### How to test it

`npm test`

### Review checklist

* The PR resolves #3342
* Linting passes
* Tests pass


elements/lisk-p2p/src/inbound_peer.ts
```diff
@@ -0,0 +1,98 @@
...
+		peerSocket?: SCServerSocket,
+	) {
+		super(peerInfo, peerConfig);
+		this._socket = peerSocket ? peerSocket : undefined;
+		if (this._socket) {
+			this._bindHandlersToInboundSocket(this._socket);
+		}
+	}
+
+	public set inboundSocket(scServerSocket: SCServerSocket) {
```
jondubois: Mayeb this setter should be renamed to just `socket` instead of `inboundSocket`?

elements/lisk-p2p/src/inbound_peer.ts
```diff
@@ -0,0 +1,98 @@
...
+		if (this._socket) {
+			this._unbindHandlersFromInboundSocket(this._socket);
+		}
+		this._socket = scServerSocket as SCServerSocketUpdated;
+		this._bindHandlersToInboundSocket(this._socket);
+	}
+
+	public disconnect(): void {
+		if (!this._socket) {
+			throw new Error();
```
jondubois: Missing error message.

elements/lisk-p2p/src/outobund_peer.ts
```diff
@@ -0,0 +1,191 @@
+/*
```
jondubois: The file name is `outobund_peer.ts`, it should be `outbound_peer.ts`.

elements/lisk-p2p/src/outobund_peer.ts
```diff
@@ -0,0 +1,191 @@
...
+		peerSocket?: SCClientSocket,
+	) {
+		super(peerInfo, peerConfig);
+		this._socket = peerSocket ? peerSocket : undefined;
+		if (this._socket) {
+			this._bindHandlersToOutboundSocket(this._socket);
+		}
+	}
+
+	public set outboundSocket(scClientSocket: SCClientSocket) {
```
jondubois: Maybe this should be `socket` instead of `outboundSocket`?

elements/lisk-p2p/src/peer.ts
```diff
@@ -55,6 +55,12 @@ interface ClientOptionsUpdated {
 	readonly connectTimeout?: number;
 }
 
+export type SCServerSocketUpdated = {
```
jondubois: If this is only used in `inbound_peer.ts`, it could be moved to that file and not exported.

elements/lisk-p2p/src/peer.ts
```diff
@@ -43,7 +43,7 @@ import {
 } from './validation';
 
 // This interface is needed because pingTimeoutDisabled is missing from ClientOptions in socketcluster-client.
-interface ClientOptionsUpdated {
+export interface ClientOptionsUpdated {
```
jondubois: If this is only used in `outbound_peer.ts`, it can be moved to that file and not exported.

elements/lisk-p2p/src/peer.ts
```diff
@@ -340,43 +298,30 @@ export class Peer extends EventEmitter {
 	}
 
 	public connect(): void {
-		if (!this._outboundSocket) {
-			this._outboundSocket = this._createOutboundSocket();
+		if (!this._socket) {
+			throw new Error();
```
jondubois: Missing error message.

elements/lisk-p2p/src/peer.ts
```diff
@@ -340,43 +298,30 @@ export class Peer extends EventEmitter {
...
-			this._unbindHandlersFromInboundSocket(this._inboundSocket);
-		}
-	}
-
-	public dropOutboundConnection(code: number = 1000, reason?: string): void {
-		if (this._outboundSocket) {
-			this._outboundSocket.destroy(code, reason);
-			this._unbindHandlersFromOutboundSocket(this._outboundSocket);
+		if (!this._socket) {
+			throw new Error();
```
jondubois: Missing error message.

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -511,11 +452,7 @@ export class PeerPool extends EventEmitter {
 
 	private _removePeerIfFullyDisconnected(peerId: string): void {
 		const peer = this.getPeer(peerId);
-		if (
-			peer &&
-			peer.state.inbound === ConnectionState.CLOSED &&
-			peer.state.outbound === ConnectionState.CLOSED
-		) {
+		if (peer && peer.state === ConnectionState.CLOSED) {
```
jondubois: 👍

elements/lisk-p2p/src/peer.ts
diego-G: It is used also in the base class https://github.com/LiskHQ/lisk-sdk/blob/6c5ea4ca8da1d29addbcdbebab363d569b5ab7da/elements/lisk-p2p/src/peer.ts#L149 .

elements/lisk-p2p/src/inbound_peer.ts
```diff
@@ -0,0 +1,98 @@
...
+
+import { SCServerSocket } from 'socketcluster-server';
+
+export class InboundPeer extends Peer {
+	protected _socket: SCServerSocketUpdated | undefined;
+
+	public constructor(
+		peerInfo: P2PDiscoveredPeerInfo,
+		peerConfig?: PeerConfig,
+		peerSocket?: SCServerSocket,
```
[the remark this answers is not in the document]
diego-G: If we want to do this we need to disable TSLINT rule `(A required parameter cannot follow an optional parameter.ts(1016))` or convert parameters to destructured object. I would go for the second option but we must implement that all over across the library as a pattern. Shall we open an issue for it?

elements/lisk-p2p/src/peer.ts
diego-G: Same as `SCServerSocketUpdated` type.

elements/lisk-p2p/test/unit/outbound_peer.ts
```diff
@@ -0,0 +1,67 @@
...
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+
+import { expect } from 'chai';
+
+import { OutboundPeer } from '../../src/peer';
+import { P2PDiscoveredPeerInfo } from '../../src/p2p_types';
+
+describe('outound_peer', () => {
```
mitsujutsu: `outbound`

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -51,9 +50,11 @@ import {
 	EVENT_OUTBOUND_SOCKET_ERROR,
 	EVENT_REQUEST_RECEIVED,
 	EVENT_UPDATED_PEER_INFO,
+	InboundPeer,
+	OutboundPeer,
 	Peer,
 	PeerConfig,
-} from './peer';
+} from './peer/';
```
shuse2: no need of `/` at the end

elements/lisk-p2p/test/unit/peer_discovery.ts
```diff
@@ -65,6 +66,10 @@ describe('peer discovery', () => {
 				});
 			});
 
+			afterEach(async () => {
+				sandbox.restore();
```
shuse2: it's restore `afterEach` in global level

elements/lisk-p2p/test/unit/errors.ts
```diff
@@ -12,6 +12,7 @@
  * Removal or modification of this copyright notice is prohibited.
  *
  */
+
```
shuse2: maybe prettier setting is different?

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -392,89 +389,36 @@ export class PeerPool extends EventEmitter {
...
-
-		return true;
-	}
-
 	public addOutboundPeer(
 		peerId: string,
 		peerInfo: P2PDiscoveredPeerInfo,
-		socket: SCClientSocket,
-	): boolean {
+		socket?: SCClientSocket,
```
shuse2: do we need this `socket` option?

elements/lisk-p2p/src/peer/index.ts
```diff
@@ -0,0 +1,17 @@
...
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+export * from './inbound';
+export * from './base';
```
shuse2: we don't need to expose `base`?

elements/lisk-p2p/src/peer/outbound.ts
```diff
@@ -0,0 +1,340 @@
...
+	}
+
+	public connect(): void {
+		if (!this._socket) {
+			this._socket = this._createOutboundSocket();
+		}
+		this._socket.connect();
+	}
+
+	public disconnect(code: number = 1000, reason?: string): void {
```
shuse2: it seems like there are lots of redundant code between outbound/inbound peer, why not put it in base?

elements/lisk-p2p/test/unit/peer_discovery.ts
diego-G: Global restore doesn't apply over this test that's why I've added it.

elements/lisk-p2p/test/unit/errors.ts
diego-G: Can you further elaborate your suggestion?

elements/lisk-p2p/src/peer_pool.ts
diego-G: Yes, we do as per discussion with @jondubois.

elements/lisk-p2p/src/peer/outbound.ts
diego-G: There is not so much. We rewrite the base methods that are different depending on the kind of peer. We decided to create two different classes to have a better view of every kind. 

Besides tslint doesn't allow more than one class per file.

elements/lisk-p2p/src/peer/index.ts
diego-G: We do need because another files like p2p or peer_pool use methods from the base. Anyways we should rearrange these kind of things in short https://github.com/LiskHQ/lisk-sdk/issues/3678

elements/lisk-p2p/test/unit/peer_discovery.ts
shuse2: can you check why it's not applied here...? maybe config is wrong

elements/lisk-p2p/src/peer_pool.ts
shuse2: @jondubois is it because we still keep the socket from discovery now?

elements/lisk-p2p/src/peer/index.ts
shuse2: mmm ok let's clean up in the #3678

elements/lisk-p2p/src/peer/inbound.ts
```diff
@@ -0,0 +1,115 @@
...
+				reason,
+			});
+		};
+		this._socket = peerSocket;
+		this._bindHandlersToInboundSocket(this._socket);
+	}
+
+	public set socket(scServerSocket: SCServerSocket) {
+		if (this._socket) {
+			this._unbindHandlersFromInboundSocket(this._socket);
```
jondubois: For the inbound peer, there will always be an existing socket so we don't need to do this check.

elements/lisk-p2p/src/peer/inbound.ts
```diff
@@ -0,0 +1,115 @@
...
+		}
+		this._socket = scServerSocket as SCServerSocketUpdated;
+		this._bindHandlersToInboundSocket(this._socket);
+	}
+
+	public disconnect(code: number = 1000, reason?: string): void {
+		if (!this._socket) {
+			throw new Error('Peer socket does not exist');
```
jondubois: As above, the socket always exits on the inbound peer.

elements/lisk-p2p/src/peer/outbound.ts
```diff
@@ -0,0 +1,340 @@
...
+		if (!this._socket) {
+			throw new Error('Peer socket does not exist');
```
jondubois: Previously, when disconnecting the peer, if the outbound socket did not exist, this was a no-op (fail-safe).
If we throw an error here, then the peer_pool will need to check if the socket exists on the peer before invoking disconnect (which we are not doing now).

I think it's better in this case to make it a no-op if the socket does not exist (since the absence of the socket means that it's disconnected already so it's not exactly a failure).

elements/lisk-p2p/src/peer_pool.ts
jondubois: @shuse2 We need now, but yes we should aim to remove it when we change discovery.

elements/lisk-p2p/src/peer/outbound.ts
```diff
@@ -0,0 +1,339 @@
...
+	new Promise<ConnectAndFetchResponse>(
+		(resolve, reject): void => {
+			const legacyNodeInfo = nodeInfo
+				? convertNodeInfoToLegacyFormat(nodeInfo)
+				: undefined;
+			// Add a new field discovery to tell the receiving side that the connection will be short lived
+			const requestPacket = {
+				procedure,
+			};
+			// Ideally, we should JSON-serialize the whole NodeInfo object but this cannot be done for compatibility reasons, so instead we put it inside an options property.
```
shuse2: do we need to do this for all the options?
I think it was originally something like,
```
const clientOptions = {
...basicPeerInfo,
...peerConfig,
query: querystring.stringify({
					...legacyNodeInfo,
					options: JSON.stringify(legacyNodeInfo),
				}),
}
```

PR #3653, opened 2019-05-17
"Error writing cache count for transactions - Cache Disabled" when /api/node/status is called - Closes #2482

### LINKED ISSUE #2482: "Error writing cache count for transactions - Cache Disabled" when /api/node/status is called

pablitovicente:
### What was the problem?

- loader.js was enabling/disabling cache regardless of the cache being enabled/disabled

### How did I fix it?

- By adding extra checks in loader.js

### How to test it?

- Sync a node until loader tries to enable the cache and the cache config is off (or redis is turned off)
- `GET api/node/status` and you shouldn't see the error message in the console attempting to write/read from the cache

### Review checklist

* The PR resolves #2482
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/chain/submodules/loader.js
+++ b/framework/src/modules/chain/submodules/loader.js
@@ -867,7 +867,7 @@ __private.loadBlocksFromNetwork = function(cb) {
  */
 __private.sync = function(cb) {
 	library.logger.info('Starting sync');
-	if (components.cache.cacheReady) {
+	if (components.cache.options.enabled && components.cache.cacheReady) {
 		components.cache.disable();
 	}
 
@@ -915,7 +915,7 @@ __private.sync = function(cb) {
 			__private.blocksToSync = 0;
 
 			library.logger.info('Finished sync');
-			if (!components.cache.cacheReady) {
+			if (components.cache.options.enabled && !components.cache.cacheReady) {
 				components.cache.enable();
 			}
 			return setImmediate(cb, err);
--- a/framework/src/modules/chain/submodules/loader.js
+++ b/framework/src/modules/chain/submodules/loader.js
@@ -867,7 +867,7 @@ __private.loadBlocksFromNetwork = function(cb) {
  */
 __private.sync = function(cb) {
 	library.logger.info('Starting sync');
-	if (components.cache.options.enabled && components.cache.cacheReady) {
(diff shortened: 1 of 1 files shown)


framework/src/modules/chain/submodules/loader.js
```diff
@@ -867,7 +867,7 @@ __private.loadBlocksFromNetwork = function(cb) {
  */
 __private.sync = function(cb) {
 	library.logger.info('Starting sync');
-	if (components.cache.cacheReady) {
+	if (components.cache.options.enabled && components.cache.cacheReady) {
```
lsilvs: I would remove the second condition as it isn't necessary and creates confusion

```suggestion
	if (components.cache.options.enabled) {
```

framework/src/modules/chain/submodules/loader.js
```diff
@@ -915,7 +915,7 @@ __private.sync = function(cb) {
 			__private.blocksToSync = 0;
 
 			library.logger.info('Finished sync');
-			if (!components.cache.cacheReady) {
+			if (components.cache.options.enabled && !components.cache.cacheReady) {
```
lsilvs: Same as above

```suggestion
			if (components.cache.options.enabled) {
```

framework/src/modules/chain/submodules/loader.js
pablitovicente: Cache can be enabled and not ready (for example redis was not up at boot or went down)

framework/src/modules/chain/submodules/loader.js
pablitovicente: Cache can be enabled and not ready (for example redis was not up at boot or went down)

ManuGowda:
@pablitovicente do we have tests covered for this scenario?

pablitovicente:
> @pablitovicente do we have tests covered for this scenario?

I think this scenario only happens after a full sync so I don't think it's practical to test from within SDK

Issue #3664, opened 2019-05-17
Rate Limit RPC/Events for specified messages from Peers 


mitsujutsu:
### Expected behavior

P2P layer should limit the number of incoming calls from peers for specified message types (e.g. `getTransactions`, `getBlocks`) within a 10 second window to avoid spamming a node as per [LIP 0004 ](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0004.md).

### Actual behavior

P2P layer should not have knowledge of module specific event types.

However it should be responsible for keep track of the 10-second interval for each peer and count the number of calls in order to limit and apply penalty as necessary.

### Proposed options

1. Set a generalized limit for all incoming messages regardless of message type, with a higher threshold within the P2P layer, and apply a 10 point penalty in case of breach.

2. For modules which require penalties for specific message types, they can implement their own rate-limiting functionality from outside the P2P layer. For example this can be handled between the `chain` and `network` module. 

3. Pass a map of `event` types and relevant penalty scores to P2P constructor. However this may not work depending on the order in which the modules are loaded. 

### Which version(s) does this affect? (Environment, OS, etc...)
2.0

PR #3665, opened 2019-05-17
Implement peer banning mechanism - Closes #3343 #3664

### LINKED ISSUE #3343: Add banning mechanism
### Description

With phase 2 from [LIP0004](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0004.md), we have to introduce a robust mechanism in lisk-p2p. There is a ban score for every peer that is initially 0 and at most 100. It is increased every time a peer sends invalid information or too many messages. Once the ban score reaches 100, a peer is banned for a period of `banTime` seconds and the ban score is reset to 0. The default value of `banTime` is 86400, which corresponds to 24 h, and it can be changed by the user in the configuration.

- Ban scores are stored in memory and not broadcast.

- Banned peers are further removed from the collections `triedPeers` and `newPeers` and stored in a separate data structure to be able to prevent reconnection of the banned peer for `banTime` seconds.

- The peers that are declared whitelisted or fixed and the seed nodes are never banned.

- Blacklisted peers stay permanently banned.

- For keeping track of the number of Events/RPCs in a 10-second window for a message type, it is sufficient to use a counter that is reset to 0 every 10 seconds instead of counting the number of calls in a sliding 10 s window.

Banned peers need to be maintained and checked during peer selection and discovery.

mitsujutsu:
### What was the problem?

No mechanism was in place for penalizing peers bad behavior as per[ LIP 0004](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0004.md).

### How did I fix it?

Added banning functionality to P2P library.

### Review checklist

* The PR resolves #3343 #3664 
* Linting passes
* Tests pass


framework/src/modules/network/network.js
```diff
@@ -42,6 +44,8 @@ const { Peer } = require('./components/storage/entities');
 
 const hasNamespaceReg = /:/;
 
+const DEFAULT_PEER_BAN_TIME = 86400;
```
diego-G: Shall we get this from the module's config constants?

elements/lisk-p2p/src/p2p.ts
```diff
@@ -158,7 +169,7 @@ export class P2P extends EventEmitter {
 		this._isActive = false;
 		this._newPeers = new Map();
 		this._triedPeers = new Map();
-
+		this._bannedPeers = new Set();
```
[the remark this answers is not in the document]
mitsujutsu: In this case I think we only need a Set as the `_bannedPeers` list is simply a list of peerIds. No further information is necessary.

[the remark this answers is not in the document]
diego-G: What about [WeakMap](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakMap)? We have used it across the Core Framework several times.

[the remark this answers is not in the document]
shuse2: I don't think we need WeakMap.
the `Set` should be the right data structure used

framework/src/modules/network/network.js
diego-G: We can do this all at once when tackling this issue https://github.com/LiskHQ/lisk-sdk/issues/3678. I Resolve this conversation for now.

elements/lisk-p2p/src/p2p.ts
```diff
@@ -332,9 +356,21 @@ export class P2P extends EventEmitter {
 		return this._nodeInfo;
 	}
 
-	/* tslint:disable:next-line: prefer-function-over-method */
-	public applyPenalty(penalty: P2PPenalty): void {
-		penalty;
+	public applyPenalty(peerPenalty: P2PPenalty): void {
+		// tslint:disable-next-line no-let
+		let exempt;
```
shuse2: prefer not to use `let` =)
I think in this case it's need `find` or `filter`

elements/lisk-p2p/src/peer.ts
```diff
@@ -151,20 +163,29 @@ export class Peer extends EventEmitter {
...
+		this._reputation = DEFAULT_REPUTATION_SCORE;
 		this._peerInfo = peerInfo;
 		this._peerConfig = peerConfig ? peerConfig : {};
 		this._ipAddress = peerInfo.ipAddress;
 		this._wsPort = peerInfo.wsPort;
 		this._id = constructPeerId(this._ipAddress, this._wsPort);
 		this._height = peerInfo.height ? peerInfo.height : 0;
+		this._callCounter = 0;
+		this._counterResetInterval = setInterval(
+			() => (this._callCounter = 0),
```
shuse2: isn't it suppose to be
`() => { this._callCounter = 0; }` or it's another syntax?

elements/lisk-p2p/src/peer.ts
```diff
@@ -294,6 +315,13 @@ export class Peer extends EventEmitter {
 		};
 	}
 
+	public applyPenalty(penalty: number): void {
+		this._reputation -= penalty;
```
shuse2: what happens if the peer is already banned?

elements/lisk-p2p/src/peer.ts
```diff
@@ -628,6 +656,29 @@ export class Peer extends EventEmitter {
...
+		this.disconnect(FORBIDDEN_CONNECTION, FORBIDDEN_CONNECTION_REASON);
+		this.emit(EVENT_BAN_PEER, this._id);
+	}
+
+	private _unbanPeer(): void {
+		this.emit(EVENT_UNBAN_PEER, this._id);
+	}
+
+	private _attachRateInfo = (packet: any) => {
+		packet.rateInfo = {
```
shuse2: prefer not to mutate.
```
return {
...packet,
  rateInfo: {
  }
}
```

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -509,6 +532,15 @@ export class PeerPool extends EventEmitter {
 		return this._peerMap.delete(peerId);
 	}
 
+	public updatePeerScore(peerPenalty: P2PPenalty): void {
```
shuse2: maybe use consistently use `applyPenalty`?

elements/lisk-p2p/src/peer.ts
mitsujutsu: Problem is what if `packet` has deeply nested fields. This will only do a shallow copy.

shuse2: it's a shallow copy but originally it was mutating directly.
also, why `packet` is any?

elements/lisk-p2p/src/peer.ts
mitsujutsu: If peer is already banned it would not exist in `peer_pool`. The `updatePeerScore` checks for this.

elements/lisk-p2p/src/peer.ts
mitsujutsu: Yes they achieve the same result.

elements/lisk-p2p/src/peer.ts
mitsujutsu: In this case I used `any` rather than `unknown` to allow for assigning property.  Other option is I can use lodash `merge`

elements/lisk-p2p/src/p2p.ts
```diff
@@ -448,6 +481,14 @@ export class P2P extends EventEmitter {
 					return;
 				}
 
+				if (this._bannedPeers.has(peerId)) {
+					this._disconnectSocketDueToFailedHandshake(
+						socket,
+						FORBIDDEN_CONNECTION,
+						FORBIDDEN_CONNECTION_REASON,
+					);
```
shuse2: needs return here?

elements/lisk-p2p/test/integration/p2p.ts
```diff
@@ -281,6 +281,29 @@ describe('Integration tests for P2P library', () => {
 			});
 		});
 
+		describe('Peer banning mechanism', () => {
```
shuse2: ❤️

elements/lisk-p2p/test/integration/p2p.ts
```diff
@@ -116,6 +116,7 @@ describe('Integration tests for P2P library', () => {
 					ackTimeout: 5000,
 					// Set a different discoveryInterval for each node; that way they don't keep trying to discover each other at the same time.
 					discoveryInterval: DISCOVERY_INTERVAL + index * 11,
+					peerBanTime: 5000,
```
diego-G: Please, add another test case to check the ban expires after this interval.

framework/src/modules/network/network.js
diego-G: Once arrived to this point, we should add this to this file https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/modules/network/defaults/config.js

mitsujutsu:
> I think we're not properly passing the config parameters to the network module and later to the library. Please, rebase your branch to the feature.

Configuration issue to be handled to separate PR as per discussion with @diego-G

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -145,6 +156,11 @@ export class Peer extends EventEmitter {
 		this._wsPort = peerInfo.wsPort;
 		this._id = constructPeerId(this._ipAddress, this._wsPort);
 		this._height = peerInfo.height ? peerInfo.height : 0;
+		this._reputation = DEFAULT_REPUTATION_SCORE;
+		this._callCounter = 0;
```
jondubois: Where doest the `_callCounter` get incremented?

Also I was thinking that this could be more advanced; we could count per peer **and** per peer event/procedure type. So the _callCounter could be a `Map` which keeps track of counts for each different kind of event.
But we have to be sure to cleanup this map correctly.

elements/lisk-p2p/test/integration/p2p.ts
```diff
@@ -168,6 +169,63 @@ describe('Integration tests for P2P library', () => {
 			});
 		});
 
+		describe('Peer banning mechanism', () => {
```
jondubois: Nice 👍

framework/src/modules/network/network.js
```diff
@@ -290,7 +306,8 @@ module.exports = class Network {
 
 				return getCountByFilter(peerList, action.params);
 			},
-			applyPenalty: action => this.p2p.applyPenalty(action.params),
+			applyPenalty: action =>
+				this.p2p.applyPenalty(action.params.peerId, action.params.penalty),
```
jondubois: We should make the param name consistent with the one in `requestFromPeer` and `emitToPeer` actions.
`requestFromPeer` and `emitToPeer` use `action.params.peer` but `applyPenalty` uses `action.params.peerId`.

Maybe we should rename all to `action.params.peerId`?

PR #3677, opened 2019-05-20
commander core:install command hangs after installation - Closes #3676

### LINKED ISSUE #3676: commander core:install command hangs after installation
### Expected behavior
`lisk core:install` command shouldn't hang, it should successfully install and start node
### Actual behavior
`lisk core:install` command hangs
### Steps to reproduce
`lisk core:install`
2.2.0

ManuGowda:
### What was the problem?
Core commands were hanging due to PM2 exception were not handled
### How did I fix it?
Handled exception in PM2
### How to test it?
Run all core commands `lisk core --help`
### Review checklist

* The PR resolves #3676 
* Linting passes
* Tests pass

### DIFF
--- a/commander/src/utils/core/pm2.ts
+++ b/commander/src/utils/core/pm2.ts
@@ -264,11 +264,7 @@ export const describeApplication = async (
 		const application = await describePM2(name);
 		disconnect();
 
-		if (application) {
-			return extractProcessDetails(application);
-		}
-
-		return false;
+		return extractProcessDetails(application);
 	} catch (error) {
 		disconnect();
 
--- a/commander/src/utils/core/pm2.ts
+++ b/commander/src/utils/core/pm2.ts
@@ -258,7 +258,7 @@ export const listApplication = async (): Promise<
 
 export const describeApplication = async (
 	name: string,
-): Promise<PM2ProcessInstance | boolean> => {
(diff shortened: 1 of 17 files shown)


commander/src/utils/core/pm2.ts
```diff
@@ -258,10 +258,16 @@ export const listApplication = async (): Promise<
...
+	try {
+		await connectPM2();
+		const application = await describePM2(name);
+		disconnect();
+
+		return extractProcessDetails(application);
+	} catch (error) {
+		disconnect();
+
+		return false;
```
shuse2: prefer to return consistent type.
maybe
`Promise<PM2ProcessInstance | undefined>` or rethrow error is bit better

ManuGowda: the problem with throwing would be difficult to handle the case.. here i am returning boolean, that way I can handle the condition, something like this
```
const instance = await describeApplication(name);

		if (!instance) {
			this.log(
				`Lisk Core instance: ${name} doesn't exists, Please install using lisk core:install`,
			);

			return;
		}
		// tslint:disable-next-line await-promise
		await StopCommand.run([name]);
		// tslint:disable-next-line await-promise
		await StartCommand.run([name]);
```

shuse2: then let's say `undefined` =)

PR #3679, opened 2019-05-21
Refactor blocks related modules not to have circular dependency - Closes #3616

### LINKED ISSUE #3616: Combine all the block related logic into one module and expose used functionality
### Expected behavior
Within blocks module, it should not have dependency to transport, forge, loader, and no internal circular dependency.

- blocks/
- blocks.js
- block_reward.js
- block.js
- block_version.js
- block_slots.js

### Actual behavior
It has circular dependency within the modules and using upper layer modules

shuse2:
### What was the problem?
Block related logic was in everywhere, and had circular dependency to other modules, siblings modules and to itself.

### How did I fix it?
- Change all the dependent functions to be a simple function.
- Create a root file and only exposes what is used from outside
- Move logic related to block inside of block

### How to test it?
- Integration test and functional test should be as it is
- Unit test should be up to date

### Review checklist

* The PR resolves #3616 
* Linting passes
* Tests pass


framework/src/modules/chain/blocks/blocks.js
```diff
@@ -0,0 +1,558 @@
...
+		this._shouldNotBeActive();
+		this._isActive = true;
+		try {
+			const block = await blocksProcess.generateBlock({
+				keypair,
+				timestamp,
+				transactions,
+				lastBlock: this._lastBlock,
+				storage: this.storage,
+				exceptions: this.exceptions,
```
SargeKhan: generateBlock should never be passed with exceptions. If we need to do it due to consistency, then it should be an empty array.

framework/src/modules/chain/blocks/chain.js
```diff
@@ -0,0 +1,434 @@
...
+ *
+ * @private
+ * @param {Object} block - Block object
+ * @param {Object} transactions - Transaction object
+ * @param {Object} sender - Sender account
+ * @param {function} cb - Callback function
+ * @returns {function} cb - Callback function from params (through setImmediate)
+ * @returns {Object} cb.err - Error if occurred
+ */
+const applyBlockTransactions = async (
```
SargeKhan: Maybe change the name to applyGenesisBlockTransactions

framework/src/modules/chain/blocks/chain.js
```diff
@@ -0,0 +1,434 @@
...
+				if (tickErr) {
+					return reject(tickErr);
+				}
+				return resolve();
+			},
+			tx
+		);
+	});
+	// TODO: recover below
+	// library.bus.message('newBlock', block);
```
SargeKhan: Is this being done elsewhere?

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -0,0 +1,558 @@
...
+				blockReward: this.blockReward,
+				exceptions: this.exceptions,
+			});
+		}
+		this._isActive = false;
+		this.logger.info('Blockchain ready');
+	}
+
+	async recoverChain() {
+		this._lastBlock = await blocksChain.deleteLastBlock();
```
lsilvs: `deleteLastBlock` function expects 6 params but none is being passed here.

framework/src/modules/chain/blocks/chain.js
```diff
@@ -0,0 +1,439 @@
...
+	let secondLastBlock;
+	await storage.entities.Block.begin('Chain:deleteBlock', async tx => {
+		secondLastBlock = await loadSecondLastBlock(
+			storage,
+			interfaceAdapters,
+			genesisBlock,
+			oldLastBlock.previousBlock,
+			tx
+		);
+		await undoConfirmedStep(storage, slots, oldLastBlock, tx);
```
lsilvs: This function expects `storage, slots, block, exceptions, tx` as params but `exceptions` is missing here

framework/src/modules/chain/loader.js
```diff
@@ -881,37 +366,90 @@ __private.loadBlocksFromNetwork = function(cb) {
...
+									throw new Error(
+										`Failed chain recovery after failing to load blocks while network consensus was low. ${recoveryError}`
 									);
 								}
-								__private.blocksToSync = lastValidBlock.height;
-								loaded = lastValidBlock.id === lastBlock.id;
-								// Reset counter after a batch of blocks was successfully loaded from the network
-								failedAttemptsToLoad = 0;
-								return waterCb();
+								throw new Error(
```
lsilvs: This will throw when `modules.blocks.recoverChain()` succeeds.
Is this the correct behaviour?

shuse2: It seems in either cases, it is on error path
https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/modules/chain/loader.js#L896
https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/modules/chain/loader.js#L905

lsilvs: But note that [here](https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/modules/chain/loader.js#L894) it uses `return` before the function meaning the second one isn't executed.

shuse2: Yes second one is not executed, but `waterCb` is called with the first argument (error string), so I think throw is correct here

PR #3698, opened 2019-05-22
Support asset field in BaseTransaction - Closes #3697

### LINKED ISSUE #3697: Support asset field in BaseTransaction
### Expected behavior
BasicTransaction should store the asset property and implement the default behaviour for `assetToJSON` function, so the transactions don't necessarily need to implement the following constructor:
```
public constructor(rawTransaction: unknown) {
		super(rawTransaction);
		const tx = (typeof rawTransaction === 'object' && rawTransaction !== null
			? rawTransaction
			: {}) as Partial<TransactionJSON>;
		// Initializes to empty object if it doesn't exist
		this.asset = (tx.asset || {}) as TransferAsset;
}
```
All of the transaction types implement the `assetToJSON` function in the same way, which also can be moved to the BasicTransaction:
```
	public assetToJSON(): TransferAsset {
		return this.asset;
	}
```

`2.0`

MaciejBaj:
### What was the problem?
Each time while defining a transaction extending BasicTransaction there is a need for implementing the complex constructor supporting the asset field and assetToJSON function which is the same for all transaction types.
### How did I fix it?
- assign `asset` field in the constructor of BasicTransaction
- implement default `assetToJSON` method in BasicTransaction
### Review checklist

* The PR resolves #3697 
* Linting passes
* Tests pass


elements/lisk-transactions/src/base_transaction.ts
```diff
@@ -521,6 +521,14 @@ export abstract class BaseTransaction {
 		]);
 	}
 
+	public assetToJSON(): object {
+		return this.asset;
+	}
+
+	protected assetToBytes(): Buffer {
```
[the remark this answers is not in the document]
shuse2: @MaciejBaj Please address this comment because this will not work properly

elements/lisk-transactions/src/base_transaction.ts
```diff
@@ -521,6 +521,14 @@ export abstract class BaseTransaction {
...
+		return Buffer.from(JSON.stringify(this.asset), 'utf-8');
```
MaciejBaj:
```suggestion
		return Buffer.from(JSON.stringify(this.asset).split('').sort().toString(), 'utf-8');
```

elements/lisk-transactions/src/base_transaction.ts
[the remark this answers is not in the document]
MaciejBaj: #bdd3fdf

elements/lisk-transactions/src/utils/asset.ts
```diff
@@ -0,0 +1,71 @@
...
+type Primitive = string | boolean | number | symbol | null | Date | undefined;
+interface FlatObject {
+	// tslint:disable-next-line
+	readonly [key: string]: any;
+}
+
+export const flatten = (target: object): FlatObject => {
+	const output = {};
+
+	const step = (
```
shuse2: prefer to define this outside

elements/lisk-transactions/src/utils/asset.ts
```diff
@@ -0,0 +1,71 @@
...
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+
+type Primitive = string | boolean | number | symbol | null | Date | undefined;
+interface FlatObject {
+	// tslint:disable-next-line
+	readonly [key: string]: any;
```
shuse2: `unknown` is not an option?

elements/lisk-transactions/src/utils/asset.ts
```diff
@@ -0,0 +1,71 @@
...
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+
+type Primitive = string | boolean | number | symbol | null | Date | undefined;
+interface FlatObject {
+	// tslint:disable-next-line
```
shuse2: if it's absolutely required, then `no-any` should be added

elements/lisk-transactions/src/utils/asset.ts
```diff
@@ -0,0 +1,71 @@
...
+	const output = {};
+
+	const step = (
+		currentObject: object,
+		previousKey: string,
+		currentDepth: number,
+	): void => {
+		Object.entries(currentObject).forEach(([key, value]) => {
+			const isArray: boolean = Array.isArray(value);
+			const type: string = Object.prototype.toString.call(value);
```
shuse2: why not `typeof` here?

elements/lisk-transactions/src/utils/asset.ts
```diff
@@ -0,0 +1,71 @@
...
+
+export const flatten = (target: object): FlatObject => {
+	const output = {};
+
+	const step = (
+		currentObject: object,
+		previousKey: string,
+		currentDepth: number,
+	): void => {
+		Object.entries(currentObject).forEach(([key, value]) => {
```
shuse2: instead of `forEach`, I think intended effect is `reduce`

elements/lisk-transactions/src/utils/asset.ts
```diff
@@ -0,0 +1,71 @@
...
+	step(target, '', 1);
+
+	return output;
+};
+
+export const primitiveToString = (
+	primitive: Primitive | Array<Primitive | object>,
+): string => {
+	// Buffer.from throws for undefined
+	if (primitive === undefined) {
```
shuse2: shouldn't we remove `undefined` first?

MaciejBaj:
@shuse2 @SargeKhan I would prefer to keep the simplest version accepting the risk of possible confusion in the case of non-deterministic order of keys in assets. The proper research is currently ongoing. I added the relevant `FixMe` to the `assetToBytes` function.

shuse2:
For 2.1, we will go with this temporal solution.

PR #3700, opened 2019-05-22
Do not require nonce when broadcasting any object - Close #3699

### LINKED ISSUE #3699: Do not require nonce when broadcasting any object
### Expected behavior

We don't require property `nonce` when broadcasting any kind of object. However, the emit and its validation should be kept for backwards compatibility.

### Actual behavior
We require it as mandatory for transactions and signatures.

`2.0.0+`

diego-G:
### Review checklist

* The PR resolves #3699
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/chain/schema/definitions.js
+++ b/framework/src/modules/chain/schema/definitions.js
@@ -74,7 +74,7 @@ module.exports = {
 	WSSignaturesList: {
 		id: 'WSSignaturesList',
 		type: 'object',
-		required: ['nonce', 'signatures'],
+		required: ['signatures'],
 		properties: {
 			nonce: {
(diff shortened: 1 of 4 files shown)


framework/src/modules/chain/logic/broadcaster.js
```diff
@@ -99,7 +97,6 @@ class Broadcaster {
 		// Broadcast using Elements P2P library via network module
 		const wrappedData = {
 			...data,
-			nonce: this.nonce,
```
shuse2: You have to keep sending the nonce, otherwise it will be rejected by old nodes

shuse2:
@diego-G Unit test is failing for the changes

PR #3701, opened 2019-05-22
Add capability to register database migration from the application/module layer - Closes #3197

### LINKED ISSUE #3197: Add capability to register database migration from the application/module layer
### Expected behavior
Expect `Application` class to expose a method called `registerMigrations` in order to allow applications to define migrations to be run along with the framework migrations.

Also, each module should be able to define its own migrations through the `migrations` attribute in the module class. By doing so, all chain related migrations should be moved to `ChainModule` class and all p2p related migrations should be moved to `NetworkModule` class.

Migrations should be handled by the `Controller` which will [before loading the modules] get all registered migrations files, order them by date/time and execute the ones not yet executed.

The migration name convention should follow the convention formerly used by Lisk Core: `yyyyMMddHHmmss_name_of_migration.sql`

### Actual behavior
Migrations are handled by Chain Module

### Parent
#3146 

2.2.0

lsilvs:
### What was the problem?
All migrations were being handled by Chain Module and there was no way to run application/modules specific migrations.

### How did I fix it?
Migrations logic was moved to Controller and now it has its own database table called `internal_migrations` to isolate it from app/modules migrations and allow updating database/code together.

`registerMigrations` was introduced in Application class in order to pass all registered migrations during Controller loading.

Chain-specific migrations was moved to ChainModule class.
Network-specific migrations was moved to NetworkModule class.
Migration-specific migrations was moved to Framework.

Because `20160723182900_create_schema.sql` file had migrations belonging to `migrations`, `chain` and `network`, it was split into 3 files and another migration (`20190521000002_migrate_migrations_table.sql`) was created to take care of the split.

See bellow current list of migrations and which module it belongs to:

```
       id       |                                name                          module
----------------+----------------------------------------------------+--------------------------
 20160723182900 | create_schema                                      | [framework|chain|network]
 20160723182901 | create_views                                       | [chain]
 20160724114255 | create_memory_tables                               | [chain]
 20160724132825 | upcase_memory_table_addresses                      | [chain]
 20160725173858 | alter_mem_accounts_columns                         | [chain]
 20160908120022 | add_virgin_column_to_mem_accounts                  | [chain]
 20160908215531 | protect_mem_accounts_columns                       | [chain]
 20161007153817 | create_memory_table_indexes                        | [chain]
 20161016133824 | add_broadhash_column_to_peers                      | [network]
 20170113181857 | add_constraints_to_peers                           | [network]
 20170124071600 | recreate_trs_list_view                             | [chain]
 20170319001337 | create_indexes                                     | [chain]
 20170321001337 | create_rounds_fees_table                           | [chain]
 20170328001337 | recreate_trs_list_view                             | [chain]
 20170403001337 | calculate_blocks_rewards                           | [chain]
 20170408001337 | create_index                                       | [chain]
 20170422001337 | recreate_calculate_blocks_rewards                  | [chain]
 20170428001337 | recreate_trs_lisk_view                             | [chain]
 20170614155841 | unique_delegates_constraint                        | [chain]
 20170921105500 | recreate_revert_mem_account_trigger                | [chain]
 20171207000001 | remove_peers_dapp_table                            | [network]
 20171207000006 | create_transfer_trs_table                          | [chain]
 20171207000007 | recreate_full_block_list_view                      | [chain]
 20171227155620 | rename_port_to_ws_port                             | [network]
 20180205000000 | underscore_patch                                   | [framework]
 20180205000001 | drop_rewards_related_functions                     | [chain]
 20180205000002 | add_height_column_to_peers                         | [network]
 20180205000003 | create_rounds_rewards_table                        | [chain]
 20180205000004 | apply_round_exception_mainnet                      | [chain]
 20180214164336 | change_case_for_blocks_columns_in_mem_accounts     | [chain]
 20180227120000 | enforce_uppercase_trs_recipienId                   | [chain]
 20180327170000 | support_long_peer_version_numbers                  | [network]
 20180419001337 | alter_mem_blockid_column                           | [chain]
 20180423001337 | remove_virgin_column                               | [chain]
 20180503114500 | add_row_id_to_trs_list_view                        | [chain]
 20180814001337 | add_indexes_for_trs_table                          | [chain]
 20180901001337 | recreate_full_blocks_list_view                     | [chain]
 20180903001337 | rename_rate_to_rank_in_delegates                   | [chain]
 20181023001337 | change_round_type_in_mem_rounds                    | [chain]
 20181106000001 | remove_duplicate_..._constraint_in_dapps           | [chain]
 20181106000002 | remove_duplicate_..._constraint_in_votes           | [chain]
 20181106000003 | remove_duplicate_..._constraint_in_transfer        | [chain]
 20181106000004 | remove_duplicate_..._constraint_in_intransfer      | [chain]
 20181106000005 | remove_duplicate_..._constraint_in_multisignatures | [chain]
 20181106000006 | change_wsport_type_in_peers                        | [network]
 20190103000001 | drop_peers_clock                                   | [network]
 20190104000001 | add_recipient_public_key_to_full_block_list        | [chain]
 20190111111557 | add_protocolVersion_column_to_peers                | [network]
 20190204170819 | migrate_trs_dependant_tables_to_trs_trable         | [chain]
 20190313102300 | drop_blocks_list_and_trs_list_views                | [chain]
 20190319111600 | remove_unconfirmed_state                           | [chain]
 20190319111700 | remove_unconfirmed_state_from_mem_accounts_trigger | [chain]
 20190410112400 | add_asset_field_mem_accounts                       | [chain]
```

### How to test it?
`dropdb lisk_dev`
`createdb lisk_dev`
`node framework/test/test_app/index.js -n devnet`

### Review checklist

* The PR resolves #3197
* Linting passes
* Tests pass


framework/src/controller/migrations/migration.js
```diff
@@ -253,31 +274,82 @@ class Migration extends BaseEntity {
...
+	 * Promise object that resolves with an array of objects `{id, name, path, file}`.
+	 */
+	async readPending(migrationsObj, savedMigrations) {
+		// const updatesPath = `${this.sqlDirectory}/updates`;
+		return Object.keys(migrationsObj).reduce((prev, namespace) => {
+			const curr = migrationsObj[namespace]
+				.map(migrationFile => {
+					const migration = migrationFile.match(/(\d+)_(.+).sql/);
+					return (
+						migration && {
```
shuse2: not sure about this syntax, what does it do?
is it suppose to be like this?
```
{
   ...migration,
   id:
}
```

framework/src/modules/network/migrations/sql/20160723182900_create_schema.sql
```diff
@@ -0,0 +1,41 @@
...
+  "id" SERIAL NOT NULL PRIMARY KEY,
+  "ip" INET NOT NULL,
+  "port" SMALLINT NOT NULL,
+  "state" SMALLINT NOT NULL,
+  "os" VARCHAR(64),
+  "version" VARCHAR(11),
+  "clock" BIGINT
+);
+
+CREATE TABLE IF NOT EXISTS "peers_dapp"(
```
shuse2: 🤕

framework/src/modules/network/migrations/index.js
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+'use strict';
+
+const path = require('path');
+
+const migrations = [
+	path.join(path.dirname(__filename), './sql/20160723182900_create_schema.sql'),
```
shuse2: do we need to list it one by one?

framework/src/controller/controller.js
```diff
@@ -189,6 +192,16 @@ class Controller {
...
 	}
 
+	async _loadMigrations(applicationMigrationsObj) {
+		const storageConfig = this.config.components.storage;
+		this.storage = createStorageComponent(storageConfig, this.logger);
+		this.storage.registerEntity('Migration', MigrationEntity);
+		await this.storage.bootstrap();
+		await this.storage.entities.Migration.applyInternal();
+		await this.storage.entities.Migration.applyList(applicationMigrationsObj);
+		return true;
```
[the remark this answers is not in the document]
lsilvs: We either return here or in the line before without the await. I will change to the second option ;)

framework/src/controller/migrations/migration.js
lsilvs: It returns the object if `migration` is true.
We can't use destructuring here because `migration` will be an array and we want to build an object.

framework/src/modules/network/migrations/index.js
lsilvs: We could read the folder and build the array from all files inside.
I don't have preferences. Let me know if you do.

framework/src/controller/migrations/migration.js
shuse2: I think `map` should change the input to output one to one.
so in `true` case it's fine, but in `false` case it will be undefined. (probably needs flatMap in this case)

framework/src/modules/network/migrations/index.js
shuse2: I think it's a bit easier to manage, as we don't need to add the file then create the entry

framework/src/controller/migrations/migration.js
lsilvs: It's still fine because right after we're filtering the map result

framework/src/controller/migrations/migration.js
```diff
@@ -218,7 +224,19 @@ class Migration extends BaseEntity {
 	 *
 	 * @returns {Promise<boolean>} Promise object that resolves with a boolean.
 	 */
-	async hasMigrations() {
+	async hasInternalMigrationsTable() {
+		const hasMigrations = await this.adapter.execute(
+			"SELECT table_name hasMigrations FROM information_schema.tables WHERE table_name = 'internal_migrations';"
```
nazarhussain: Its bad practice to use mixed case name without quotes. 

```suggestion
			"SELECT table_name \"hasMigrations\" FROM information_schema.tables WHERE table_name = 'internal_migrations';"
```

framework/src/controller/migrations/migration.js
```diff
@@ -218,7 +224,19 @@ class Migration extends BaseEntity {
...
+	}
+
+	/**
+	 * Verifies presence of the 'migrations' OID named relation.
+	 *
+	 * @returns {Promise<boolean>} Promise object that resolves with a boolean.
+	 */
+	async hasMigrationsTable() {
 		const hasMigrations = await this.adapter.execute(
 			"SELECT table_name hasMigrations FROM information_schema.tables WHERE table_name = 'migrations';"
```
nazarhussain: Same here. 

```suggestion
			"SELECT table_name \"hasMigrations\" FROM information_schema.tables WHERE table_name = 'migrations';"
```

framework/src/controller/migrations/migration.js
```diff
@@ -253,31 +274,82 @@ class Migration extends BaseEntity {
...
-				.map(f => {
-					f.file = this.adapter.loadSQLFile(f.path, this.sqlDirectory);
-					return f;
+				.sort((a, b) => a.id - b.id) // Sort by migration ID, ascending
+				.map(migration => {
+					migration.file = this.adapter.loadSQLFile(migration.path, '');
+					return migration;
+				})
+		);
+	}
```
nazarhussain: Its not a good approach to maintain two migrations tables while there is no clear physical segregation. Every migration file runs on same database and modify schema of same database. 

You can use the same migrations table to run framework migrations with the namespace `framework`.

framework/src/controller/migrations/migration.js
```diff
@@ -243,8 +264,8 @@ class Migration extends BaseEntity {
 	 * @returns {Promise<Array<Object>>}
 	 * Promise object that resolves with an array of objects `{id, name, path, file}`.
 	 */
-	readPending(lastMigrationId) {
-		const updatesPath = path.join(__dirname, '../sql/migrations/updates');
+	readPendingInternalMigrations(lastMigrationId) {
+		const updatesPath = path.join(__dirname, './sql/updates');
```
nazarhussain: All these updates for schema does not belongs to framework. These are all part of the `chain` module. Without chain module, there is no database schema related to framework. So move these all migrations to `chain` migrations. 

Afterwards framework would only need one table to exists `migrations`. That you can check/create without any migration. As its the prerequisite for running the migrations.

framework/src/controller/migrations/migration.js
lsilvs: I did the separation because I couldn't find a way to migrate the `migrations` table using the same logic.

framework/src/controller/migrations/migration.js
lsilvs: There is no `chain` migration here. All sql files in this folder contains framework-related migrations only.

framework/src/controller/application.js
```diff
@@ -255,6 +259,20 @@ class Application {
...
+	 * Register migrations with the application
+	 *
+	 * @param {Object} namespace - Migration namespace
+	 * @param {Array} migrations - Migrations list. Format ['yyyyMMddHHmmss_name_of_migration.sql']
+	 */
+	registerMigrations(namespace, migrations) {
+		assert(namespace, 'Namespace is required');
+		assert(migrations instanceof Array, 'Migrations list should be an array');
+		const currentMigrations = this.getMigrations() || {};
+		currentMigrations[namespace] = migrations;
```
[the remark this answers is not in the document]
lsilvs: I don't like this approach as it allows the app to register module migrations outside the `registerModule` method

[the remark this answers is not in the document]
lsilvs: Could you give an example where it would be useful?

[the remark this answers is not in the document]
nazarhussain: Yeah one can register it any case because `registerMigrations` is on the app level. And we are not limiting that call to track either it was called from registerModule or outside.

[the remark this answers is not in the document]
nazarhussain: Its related to point above. 

For my custom app I created multiple migrations to introduced different schema used by my custom transactions. I have different networks running for my app, For one network (decided based on some env variable) I want to have a different database structure. In the app startup, based on any business logic I want to register one migration for other I don't want. That will cover the first point. 

For second point `overrideAll` will give flexibility to users to reset migrations for any namespace and register new migrations. 

We are making this change for the framework (not for any module). And any framework should stay flexible.

framework/src/controller/migrations/migration.js
nazarhussain: `migrations` table not required any migration. Its a pre-requisite come along the framework. If you want to introduce some change in the migrations table for future you still can register any SQL file with the namespace `lisk` or `lisk-framework` that will any case should and must be executed in the timely order. So no need of this new concept internal migrations.

lsilvs: Current `migrations` table has no `namespace` column. Hence, it requires migration. And when the table changes its schema itself you can't use the same code/logic for before and after the migration is run.

framework/src/controller/application.js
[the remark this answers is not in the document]
lsilvs: Before registering the migration we are checking if the `namespace` was already used. That way you can't register module migrations twice (which is good imo)

framework/src/controller/migrations/migration.js
nazarhussain: Each migration run in a consolidated database transaction. So you can change structure of migrations table within the migration file. Then left would be when adding entry to migrations table. That you can have two versions to maintain backward compatibility for old migrations.

framework/src/controller/application.js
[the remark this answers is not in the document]
nazarhussain: That's a valid case for modules, but you can register migrations from the app level as well. So one can call `app.registerMigrations` as many times he want. And with each call we override his old register migrations is not a good approach. And we can't restrict developer to call `app.registerMigrations` only once.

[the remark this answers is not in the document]
lsilvs: What is the difference between calling `registerMigrations` multiple times based on your business logic and constructing a variable which holds the migrations based on your business logic and call `registerMigrations` only once?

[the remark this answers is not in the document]
nazarhussain: Difference is when you have an interface function, you can't restrict them to only call it once. Its not a good approach. So one can call `app.registerMigrations` as many times they want to register as many migrations.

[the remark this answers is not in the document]
lsilvs: > Before registering the migration we are checking if the `namespace` was already used. That way you can't register module migrations twice (which is good imo)

Actually I've just realised I'm not checking if the `namespace` was registered before. I'll add this validation and it should be good to go imo.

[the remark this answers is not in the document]
shuse2: I don't think a solid usecase in running registerMigrations twice, and if the computation is required, it should be able to be done prior to calling the first one.
We can implement this feature when we need them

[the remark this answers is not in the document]
shuse2: same as above

framework/src/controller/migrations/migration.js
shuse2: I think solution that @lsilvs have is a good solution unless if you have alternative.
He said, migration file to alter migrations table is possible, but code cannot be adopted.
1.6 => 2.0 works, but fresh starting database will not work in this case.

lsilvs:
@nazarhussain I only agree with your last point (separate the `20190521000002_migrate_migrations_table.sql` file into its own module migration files).
Everything else I will leave as it is. Feel free to propose your implementation in another PR.

lsilvs:
Actually `20190521000002_migrate_migrations_table.sql` is already in the correct place as it updates the `migrations` table and modules migrations shouldn't touch `migrations` table imo.

lsilvs:
@nazarhussain I refactored the code to get rid of the table which kept the framework migrations. Let me know what do you reckon.

PR #3705, opened 2019-05-22
P2P errors while sync on testnet - Closes #3651

### LINKED ISSUE #3651: P2P errors while syncing with testnet
### Expected behavior
While syncing with testnet or mainnet, the node shouldn't make too many disconnects or the other errors mentioned below.
### Actual behavior
Too many P2P errors while syncing with test network.

`14:46:17.406Z ERROR lisk-framework: Event 'rpc-request' was aborted due to a bad connection`
`14:46:13.092Z ERROR lisk-framework: Unable to match an address to the peer`
`14:46:13.073Z ERROR lisk-framework: Connection id does not match with corresponding peer`
### Steps to reproduce
Start `2.0.0-alpha.4` node with test network.
2.0.0

ishantiw:
### What was the problem?

There are a lot of error messages from P2P while syncing in testnet.

### How did I fix it?

I found that this event is mostly happening when we call `applyNodeInfo` in P2P library whenever a node wants to apply its `nodeInfo` to its peers. Some peers from the old version have this issue and also the peers that are disconnected.

Below are the errors are commonly seen in the corresponding version of peers,

```
Error: Unable to match an address to the peer

Version of connected peer: '1.6.0-rc.3' and also some older version


Error: Connection id does not match with the corresponding peer

Verison of connected peer: '1.6.0-rc.4' 
```

Lower the log level of `EVENT_FAILED_TO_PUSH_NODE_INFO` from `error` to `trace`.

### How to test it?

Run the application and check the logs

### Review checklist

* The PR resolves #3651
* Linting passes
* Tests pass

### DIFF
--- a/elements/lisk-p2p/src/errors.ts
+++ b/elements/lisk-p2p/src/errors.ts
@@ -84,9 +84,16 @@ export class InvalidPeerError extends Error {
 }
 
 export class RequestFailError extends Error {
-	public constructor(message: string) {
+	public peerId: string;
+	public peerVersion: string;
+	public constructor(message: string, peerId?: string, peerVersion?: string) {
 		super(message);
 		this.name = 'RequestFailError';
+		this.peerId = peerId || '';
+		this.peerVersion = peerVersion || '';
+		this.message = `${this.message}: PeerId: ${this.peerId}: Peer Version: ${
+			this.peerVersion
+		}`;
 	}
 }
 
--- a/elements/lisk-p2p/src/errors.ts
+++ b/elements/lisk-p2p/src/errors.ts
@@ -98,9 +98,11 @@ export class RequestFailError extends Error {
 		this.errorResponse = errorResponse || new Error(message);
 		this.peerId = peerId || '';
 		this.peerVersion = peerVersion || '';
-		this.message = `${this.message}: PeerId: ${this.peerId}: Peer Version: ${
-			this.peerVersion
-		}`;
+		this.message = peerId
+			? `${this.message}: PeerId: ${this.peerId}: Peer Version: ${
+					this.peerVersion
+			  }`
+			: message;
 	}
 }
 
--- a/elements/lisk-p2p/src/errors.ts
+++ b/elements/lisk-p2p/src/errors.ts
@@ -84,16 +84,17 @@ export class InvalidPeerError extends Error {
 export class RequestFailError extends Error {
 	public peerId: string;
 	public peerVersion: string;
-	public errorResponse: Error;
+	public response: Error;
 	public constructor(
 		message: string,
-		errorResponse?: Error,
+		response?: Error,
 		peerId?: string,
 		peerVersion?: string,
 	) {
 		super(message);
 		this.name = 'RequestFailError';
-		this.errorResponse = errorResponse || new Error(message);
+		// The request was made and the peer responded with error
+		this.response = response || new Error(message);
 		this.peerId = peerId || '';
 		this.peerVersion = peerVersion || '';
 		this.message = peerId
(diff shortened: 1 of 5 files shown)


elements/lisk-p2p/src/peer.ts
```diff
@@ -391,7 +392,13 @@ export class Peer extends EventEmitter {
 					},
 					(err: Error | undefined, responseData: unknown) => {
 						if (err) {
-							reject(err);
+							reject(
```
shuse2: this makes the `test/mocha/functional/ws/transport/blocks.js` to fail because this `err` was not a `Error`, but it seems it's a string (ie err.message doesn't exist)

elements/lisk-p2p/src/errors.ts
```diff
@@ -84,9 +84,23 @@ export class InvalidPeerError extends Error {
...
 
 export class RequestFailError extends Error {
-	public constructor(message: string) {
+	public peerId: string;
+	public peerVersion: string;
+	public errorResponse: Error;
+	public constructor(
+		message: string,
+		errorResponse?: Error,
+		peerId?: string,
```
jondubois: If we use `peerId` here, then we should also use `peerId` in `RPCResponseError`.

ishantiw: Yes, it should use peerId as its more precise, I'll update it for RPC response as well

elements/lisk-p2p/src/errors.ts
```diff
@@ -98,9 +98,11 @@ export class RequestFailError extends Error {
 		this.errorResponse = errorResponse || new Error(message);
 		this.peerId = peerId || '';
 		this.peerVersion = peerVersion || '';
-		this.message = `${this.message}: PeerId: ${this.peerId}: Peer Version: ${
-			this.peerVersion
-		}`;
+		this.message = peerId
```
jondubois: Shouldn't peerId always exist?

ishantiw: As we discussed that it's being used in other places of the p2p-lib where we don't use peerId so we will keep it optional

shuse2: `this.message = peerId` seems very weird. shouldn't it be message?

elements/lisk-p2p/src/peer.ts
```diff
@@ -391,7 +392,15 @@ export class Peer extends EventEmitter {
 					},
 					(err: Error | undefined, responseData: unknown) => {
```
shuse2: in general, should this response type `Error` from SC? @jondubois

framework/test/mocha/functional/ws/transport/blocks.js
```diff
@@ -180,10 +182,10 @@ describe('WS transport blocks', () => {
 			} catch (err) {
 				__testContext.debug(
 					'> Error / Response:'.grey,
-					JSON.stringify(err),
+					JSON.stringify(err.errorResponse),
```
shuse2: why not checking err.message here?

elements/lisk-p2p/src/errors.ts
ishantiw: That's a conditional statement, when `peerId` is available then we build the message

framework/test/mocha/functional/ws/transport/blocks.js
ishantiw: Because we are only capturing the message here, whereas in `errorResponse` we are actually capturing the error or string whatever is coming from the response from the peer.

ishantiw:
@shuse2 I referred to Axios, [Axios error handling](https://github.com/axios/axios#handling-errors). they wrap the error coming from the server as a response. So I have renamed `error.errorResponse` to `error.response` to make it more clear and simple (i.e., when the request was made and the peer responded with error).

PR #3708, opened 2019-05-22
P2P does not consider broadcastLimit - Closes #3690

### LINKED ISSUE #3690: P2P does not consider broadcastLimit
### Expected behavior
When p2p library send information to the network, it should consider the broadcast limit set by the config.
It should send to random connected peers with the broadcast limit.

https://github.com/LiskHQ/lisk-sdk/blob/v1.6.0/framework/src/modules/chain/logic/peers.js#L326

### Actual behavior
It sends to the 

### Steps to reproduce
send transaction to a node, and log the selectedPeers in peer_pool

2.0-

jondubois:
### What was the problem?

- There was no default limit regarding the number of peers to send a packet to and it could not be configured.
- The same peer selection function was used for both request and send. The one needed for send does not need to account for height.
- The implementation of the peer selection function was overly complex.
- The peer selection functions (for send, request and also for connect) had a fixed set of arguments so were not future-proof in case we want to add new properties in the future.
- Added support for a new `emitPeerLimit` (integer) under the Network module config (this should be used instead of the top level `broadcastLimit`).

### How did I fix it?

- Made the sendPeerLimit configurable and specified a default in the library (`25`).
- Cleaned up interface names.
- Pass a single object to peer selection functions; that way the properties can be specified in any order; this is more future-proof.
- Use `lodash.shuffle` to randomize random selection since we already rely on it as part of discovery (instead of our inefficient solution).
- The peer limit passed to the selection functions is an upper bound; don't throw an error if there are fewer available peers than the limit. The network should be able to work with only a few peers.

### How to test it?

In `lisk-p2p` library:

`npm run test:integration`

### Review checklist

* The PR resolves #3690
* Linting passes
* Tests pass

### DIFF
--- a/elements/lisk-p2p/src/p2p.ts
+++ b/elements/lisk-p2p/src/p2p.ts
@@ -54,8 +54,6 @@ import {
 	P2PNetworkStatus,
 	P2PNodeInfo,
 	P2PPeerInfo,
-	P2PPeerSelectionForRequestFunction,
-	P2PPeerSelectionForSendFunction,
 	P2PPenalty,
 	P2PRequestPacket,
 	P2PResponsePacket,
@@ -280,7 +278,7 @@ export class P2P extends EventEmitter {
 			peerSelectionForConnection: config.peerSelectionForConnection
 				? config.peerSelectionForConnection
 				: selectForConnection,
-			peerSendLimit: config.peerSendLimit
+			sendPeerLimit: config.sendPeerLimit
 		});
 
 		this._bindHandlersToPeerPool(this._peerPool);
--- a/elements/lisk-p2p/src/p2p.ts
+++ b/elements/lisk-p2p/src/p2p.ts
@@ -108,6 +108,7 @@ export const EVENT_NEW_PEER = 'newPeer';
 
 export const NODE_HOST_IP = '0.0.0.0';
 export const DEFAULT_DISCOVERY_INTERVAL = 30000;
+export const DEFAULT_SEND_PEER_LIMIT = 25;
 
 const BASE_10_RADIX = 10;
 
@@ -278,7 +279,7 @@ export class P2P extends EventEmitter {
 			peerSelectionForConnection: config.peerSelectionForConnection
 				? config.peerSelectionForConnection
 				: selectPeersForConnection,
-			sendPeerLimit: config.sendPeerLimit
+			sendPeerLimit: config.sendPeerLimit === undefined ? DEFAULT_SEND_PEER_LIMIT : config.sendPeerLimit,
 		});
 
 		this._bindHandlersToPeerPool(this._peerPool);
--- a/elements/lisk-p2p/src/p2p.ts
+++ b/elements/lisk-p2p/src/p2p.ts
@@ -64,7 +64,7 @@ import {
 import { P2PRequest } from './p2p_request';
 export { P2PRequest };
 import {
-	selectForConnection,
+	selectPeersForConnection,
 	selectPeersForRequest,
 	selectPeersForSend,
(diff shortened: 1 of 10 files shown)


elements/lisk-p2p/src/p2p.ts
```diff
@@ -65,7 +63,11 @@ import {
 
 import { P2PRequest } from './p2p_request';
 export { P2PRequest };
-import { selectForConnection, selectPeers } from './peer_selection';
+import {
+	selectForConnection,
```
ManuGowda: why not consistent with the naming here? `selectPeersForConnection`?

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -117,6 +120,8 @@ export class PeerPool extends EventEmitter {
 		this._peerSelectForSend = peerPoolConfig.peerSelectionForSend;
 		this._peerSelectForRequest = peerPoolConfig.peerSelectionForRequest;
 		this._peerSelectForConnection = peerPoolConfig.peerSelectionForConnection;
+		this._sendPeerLimit = peerPoolConfig.sendPeerLimit === undefined ? DEFAULT_SEND_PEER_LIMIT : peerPoolConfig.sendPeerLimit;
```
ManuGowda: `this._sendPeerLimit = peerPoolConfig.sendPeerLimit ? DEFAULT_SEND_PEER_LIMIT : peerPoolConfig.sendPeerLimit;`

elements/lisk-p2p/src/peer_selection.ts
```diff
@@ -69,54 +76,26 @@ export const selectPeers = (
...
-	if (numOfPeers > processedPeers.length) {
-		throw new NotEnoughPeersError(
-			`Requested number of peers: '${numOfPeers}' is more than the available number of good peers: '${
-				processedPeers.length
-			}'`,
-		);
-	}
-
-	if (numOfPeers === processedPeers.length) {
+	if (peerLimit === undefined) {
```
ManuGowda: `if (peerLimit) {`

elements/lisk-p2p/src/peer_pool.ts
jondubois: For correctness, if the user explicitly sets it to `0`, my thinking was that it's better if `0` peers are selected; then an error will be thrown later because no peers were found.

elements/lisk-p2p/src/peer_selection.ts
jondubois: Also because of the `0` issue.

elements/lisk-p2p/src/peer_selection.ts
```diff
@@ -25,11 +32,11 @@ interface HistogramValues {
 }
 
 /* tslint:enable: readonly-keyword */
-export const selectPeers = (
-	peers: ReadonlyArray<P2PDiscoveredPeerInfo>,
-	nodeInfo?: P2PNodeInfo,
-	numOfPeers: number = 0,
+export const selectPeersForRequest = (
```
ishantiw: Shouldn't this request function should always return one peer? we can avoid unnecessary conditions to check the value of `peerLimit` and should only default to `1`?

ishantiw: unless we want to support in the future to request multiple peers at a time?

jondubois: Now yes. But in the future (or for a sidechain) we could request from multiple peers and calculate some kind of consensus on the responses. This is based on a discussion I had a while back with @diego-G

PR #3709, opened 2019-05-22
Create data structure to store block headers for BFT - Closes #3691

### LINKED ISSUE #3691: Create data structure to store block headers for BFT
### Expected behavior
We need a data structure that stores the blockheaders in memory in sorted order and easy to user interface. It would be mix of a queue or stack or mix of both. 

3.0.0

nazarhussain:
### What was the problem?
There was a need to store block headers in-memory with certain order. 

### How did I fix it?
Created a data structure specifically for the use case. 

### How to test it?

Run unit tests 

### Review checklist

* The PR resolves #3691
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/chain/logic/bft.js
+++ b/framework/src/modules/chain/logic/bft.js
@@ -16,28 +16,28 @@
 
 const assert = require('assert');
 
-const __private = {
-	items: new WeakMap(),
-	size: new WeakMap(),
-};
-
 class HeadersList {
 	constructor({ size }) {
 		assert(size, 'Must provide size of the queue');
-		__private.size.set(this, size);
-		__private.items.set(this, []);
+		this.__items = [];
+		this.__size = size;
 	}
 
 	get items() {
-		return __private.items.get(this);
+		return this.__items;
+	}
+
+	// eslint-disable-next-line class-methods-use-this
+	set items(value) {
+		throw new Error('You can\'t set the items directly use "list.add"');
 	}
 
 	get length() {
-		return __private.items.get(this).length;
+		return this.items.length;
 	}
 
 	get size() {
-		return __private.size.get(this);
+		return this.__size;
 	}
 
 	set size(newSize) {
@@ -46,7 +46,7 @@ class HeadersList {
 			this.items.splice(0, currentSize - newSize);
 		}
 
-		__private.size.set(this, newSize);
+		this.__size = newSize;
 	}
 
 	get first() {
@@ -58,11 +58,10 @@ class HeadersList {
 	}
 
 	add(blockHeader) {
-		const items = this.items;
 		const first = this.first;
 		const last = this.last;
 
-		if (items.length) {
+		if (this.items.length) {
 			assert(
 				blockHeader.height === last.height + 1 ||
 					blockHeader.height === first.height - 1,
@@ -73,18 +72,16 @@ class HeadersList {
 
 		if (first && blockHeader.height === last.height + 1) {
 			// Add to the end
-			items.push(blockHeader);
+			this.items.push(blockHeader);
 		} else {
 			// Add to the start
-			items.unshift(blockHeader);
+			this.items.unshift(blockHeader);
 		}
 
 		// If the list size is already full remove one item
-		if (items.length > this.size) {
-			items.shift();
+		if (this.items.length > this.size) {
+			this.items.shift();
 		}
-
-		__private.items.set(this, items);
 
 		return this;
 	}
@@ -94,26 +91,21 @@ class HeadersList {
 			beforeHeight = this.last.height - 1;
 		}
 
-		const items = this.items;
 		const removeItemsCount = this.last.height - beforeHeight;
-		let itemsToReturn;
 
-		if (removeItemsCount < 0 || removeItemsCount >= items.length) {
-			itemsToReturn = items.splice(0, items.length);
-		} else {
-			itemsToReturn = items.splice(
-				items.length - removeItemsCount,
-				removeItemsCount
-			);
+		if (removeItemsCount < 0 || removeItemsCount >= this.items.length) {
+			return this.items.splice(0, this.items.length);
 		}
 
-		__private.items.set(this, items);
-		return itemsToReturn;
+		return this.items.splice(
+			this.items.length - removeItemsCount,
+			removeItemsCount
+		);
 	}
 
 	empty() {
 		const items = [...this.items];
-		__private.items.set(this, []);
+		this.__items = [];
 		return items;
 	}
 }
--- a/framework/src/modules/chain/logic/bft.js
+++ b/framework/src/modules/chain/logic/bft.js
@@ -1,0 +1,79 @@
+/*
+ * Copyright © 2018 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+'use strict';
+
+const assert = require('assert');
+
+const __private = {
+	items: new WeakMap(),
+};
+
+class HeadersList {
+	constructor({ size }) {
+		assert(size, 'Must provide size of the queue');
+		this.size = size;
+		__private.items.set(this, []);
+	}
+
+	get items() {
+		return __private.items.get(this);
+	}
+
+	get length() {
+		return __private.items.get(this).length;
+	}
+
+	get first() {
+		return this.items[0];
+	}
+
+	get last() {
+		return this.items[this.length - 1];
+	}
+
+	add(blockHeader) {
+		const items = this.items;
+		const first = this.first;
+		const last = this.last;
+
+		if (items.length) {
+			assert(
+				blockHeader.height === last.height + 1 ||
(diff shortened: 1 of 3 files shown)


framework/src/modules/chain/logic/bft.js
```diff
@@ -0,0 +1,107 @@
...
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+'use strict';
+
+const assert = require('assert');
+
+const __private = {
```
shuse2: prefer not to use `__private` anymore

framework/src/modules/chain/logic/bft.js
```diff
@@ -0,0 +1,107 @@
...
+	get first() {
+		return this.items[0];
+	}
+
+	get last() {
+		return this.items[this.length - 1];
+	}
+
+	add(blockHeader) {
+		const items = this.items;
```
shuse2: these reassignment is not necessary

framework/src/modules/chain/logic/bft.js
```diff
@@ -0,0 +1,107 @@
...
+
+		return this;
+	}
+
+	remove({ beforeHeight } = {}) {
+		if (!beforeHeight) {
+			beforeHeight = this.last.height - 1;
+		}
+
+		const items = this.items;
```
shuse2: same as above

framework/src/modules/chain/logic/bft.js
```diff
@@ -0,0 +1,107 @@
...
+		}
+
+		// If the list size is already full remove one item
+		if (items.length > this.size) {
+			items.shift();
+		}
+
+		__private.items.set(this, items);
+
+		return this;
```
shuse2: are we going to chain this function?

framework/src/modules/chain/logic/bft.js
```diff
@@ -0,0 +1,107 @@
...
+		if (!beforeHeight) {
+			beforeHeight = this.last.height - 1;
+		}
+
+		const items = this.items;
+		const removeItemsCount = this.last.height - beforeHeight;
+		let itemsToReturn;
+
+		if (removeItemsCount < 0 || removeItemsCount >= items.length) {
+			itemsToReturn = items.splice(0, items.length);
```
shuse2: prefer early return without `let`

framework/src/modules/chain/logic/bft.js
```diff
@@ -0,0 +1,107 @@
+/*
```
shuse2: probably better to create
`chain/bft/headers_queue.js`

framework/src/modules/chain/logic/bft.js
nazarhussain: Why not?

framework/src/modules/chain/logic/bft.js
nazarhussain: I don't like putting everything on top level, that does not provide any advantage, rather create ambiguity in current logical structure.

shuse2: everything will be in the `chain` level so it will be solved

framework/src/modules/chain/logic/bft.js
shuse2: Eventually, we will move everything to typescript, and it will be easier to migrate if it just follow the standard convention. (ie: not use WeakMap)
Also, there is already private method in Node12 too, no need to hack the way to use it anymore

framework/src/modules/chain/logic/bft.js
```diff
@@ -0,0 +1,107 @@
...
+'use strict';
+
+const assert = require('assert');
+
+const __private = {
+	items: new WeakMap(),
+};
+
+class HeadersList {
+	constructor({ size }) {
```
SargeKhan: We agreed that BFT can rollback an indefinite number of blocks, and so the headers list should have not have an initialization size of the list. Rather, it should maintain a list which is dynamic.

framework/src/modules/chain/logic/bft.js
nazarhussain: In both cases that will be easier to migrate, if we clearly know which are private elements.

shuse2: it's not hard, but normal convention is easier in a way that we don't need to touch the code.
Also, it's a hack way to use the private feature which doesn't exist in node.js yet (pre-node12)

framework/src/modules/chain/logic/bft.js
nazarhussain: List have fixed size, but you can still add headers from previous height. So you are on 500 blocks, you removed 450 and added 200 more from past. So the list size will be 250 now. But the actual size of the list will strict not load up million block headers in memory.

framework/src/modules/chain/logic/bft.js
nazarhussain: Yes

framework/src/modules/chain/logic/bft.js
nazarhussain: Yes its should not be on chain level, as we have segerations of business logic, helpers and submodules. So we should put it where it belongs.

shuse2: i meant we won't have separation of logic and submodule anymore

framework/src/modules/chain/logic/bft.js
nazarhussain: Its a normal convention to store private data attributes for any class instance.

framework/src/modules/chain/logic/bft.js
nazarhussain: Yes it is to avoid recurring calls to get items from weakmap with object reference.

framework/src/modules/chain/logic/bft.js
nazarhussain: Same as above.

framework/test/mocha/fixtures/blocks.js
```diff
@@ -74,7 +74,26 @@ const GenesisBlock = stampit(Block, {
 	},
 });
 
+const BlockHeader = stampit({
```
limiaspasdaniel: Do we need to use stampit necessarily here?

nazarhussain: Yes stampit is used for every fixture and should be used.

framework/src/modules/chain/logic/bft.js
nazarhussain: When we remove submodules and logic in the chain module, then we can refactor restructure this as well. For now its good to use the structure which is in place.

framework/src/modules/chain/logic/bft.js
shuse2: Im approving this PR, but I will create meeting to discuss the usage of WeapMap.

framework/src/modules/chain/logic/bft.js
```diff
@@ -0,0 +1,113 @@
...
+ */
+
+'use strict';
+
+const assert = require('assert');
+
+class HeadersList {
+	constructor({ size }) {
+		assert(size, 'Must provide size of the queue');
+		this.__items = [];
```
shuse2: With the new convention please use `this._items = []`

framework/src/modules/chain/logic/bft.js
SargeKhan: I don't understand your comment. Can you explain it in a different way? Also, what will be the appropriate initialization value for the HeadersList in BFT according to you?

nazarhussain: We initialize the list with 500 blocks, in certain cases if we need a larger list we can extend it by just setting the new size `list.size = 10000`.

framework/src/modules/chain/logic/bft.js
```diff
@@ -0,0 +1,113 @@
...
+		if (first && blockHeader.height === last.height + 1) {
+			// Add to the end
+			this.items.push(blockHeader);
+		} else {
+			// Add to the start
+			this.items.unshift(blockHeader);
+		}
+
+		// If the list size is already full remove one item
+		if (this.items.length > this.size) {
```
SargeKhan: I think removing items from the blockHeadersList should be based on the heightFinalized property irrespective of the size of the list because we don't want to remove blocks unless they are finalized. Or am i missing something?

nazarhussain: We can use size property to truncate the items in the list if and when required. `finalizedHeight` is not a decision factor keeping the headers in the list. As it height can be finalized within 152 blocks but in any case we need 303 last block for verification.

framework/test/mocha/unit/modules/chain/logic/bft.js
```diff
@@ -0,0 +1,323 @@
...
+				]);
+			});
+
+			it('should remove last item from the list if passed without height', async () => {
+				list.remove();
+				expect(list.items).to.be.eql([header1, header2, header3, header4]);
+			});
+			it('should remove all items upto and including the provided height', async () => {
+				list.remove({ beforeHeight: 2 });
+				expect(list.items).to.be.eql([header1, header2]);
```
SargeKhan: This expectation does not match the test description

nazarhussain: It matches what ever height we pass in `list.remove()` it should remove all blocks above that height. The list contains 5 blocks, we asked to remove blocks above height 2 and so we have `[header1, header2]` left in the list.

SargeKhan: The test description states that it removes the headers including the provided height. 
`should remove all items upto and including the provided height`

PR #3716, opened 2019-05-23
Add defualt create value for asset field - Closes #3715

### LINKED ISSUE #3715: Add default create values account state store asset
### Expected behavior
Add `asset: null` so state store can track difference in `asset` field as well.
Bug fix for `hello_world` example @MaciejBaj 

### Steps to reproduce
Run tests.

release/2.0

michielmulders:
### What was the problem?
No default create values.

### How did I fix it?
Add default for state_store for account.

### How to test it?
Run tests.

### Review checklist

* The PR resolves #3715 
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/chain/components/storage/entities/account.js
+++ b/framework/src/modules/chain/components/storage/entities/account.js
@@ -37,6 +37,7 @@ const defaultCreateValues = {
 	nameExist: false,
 	multiMin: 0,
 	multiLifetime: 0,
+	asset: null,
 };
 
 const readOnlyFields = ['address'];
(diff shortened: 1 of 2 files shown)


shuse2:
@michielmulders can you target this PR to 2.1 too?

framework/src/modules/chain/components/storage/entities/account.js
```diff
@@ -37,6 +37,7 @@ const defaultCreateValues = {
 	nameExist: false,
 	multiMin: 0,
 	multiLifetime: 0,
+	asset: null,
```
SargeKhan: Maybe makes more sense to initialize with empty object?

michielmulders: This fix is applied in 2.1 release

PR #3719, opened 2019-05-23
Support {} behavior for asset field accounts - Closes #3717

### LINKED ISSUE #3717: Modify asset behavior for accounts in Postgres
### Expected behavior
Populate with `{ }` empty object whenever asset field value is empty.

### Actual behavior
Currently, we store `null` in Postgres whenever the account the asset field is not populated.

Why? To save space in DB.
However, it would be better to stay consistent with `trs` table as it stores the asset field as `{ }` if it is not populated. + it is better for user as he doesn't have to check if it is `null` and convert himself to `{ }`.

`release/2.0`

michielmulders:
### What was the problem?
Currently, the account field in `accounts` column holds `null` if we don't pass a JSON object for `asset` field.

### How did I fix it?
Modify behavior to store empty object `{}`.

### How to test it?
Run tests and check in DB if all accounts have `{}` under `asset` field.

### Review checklist

* The PR resolves #3717 
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/chain/components/storage/sql/migrations/updates/20190410112400_add_asset_field_mem_accounts.sql
+++ b/framework/src/modules/chain/components/storage/sql/migrations/updates/20190410112400_add_asset_field_mem_accounts.sql
@@ -19,7 +19,7 @@
 */
 
  -- Add asset column to trs table as jsonb
-ALTER TABLE "mem_accounts" ADD COLUMN IF NOT EXISTS "asset" jsonb;
+ALTER TABLE "mem_accounts" ADD COLUMN IF NOT EXISTS "asset" jsonb NOT NULL DEFAULT '{}'::jsonb;
 
 -- Create index for asset field. Using `gin` index as it's more efficient for keys or key/value search.
 CREATE INDEX IF NOT EXISTS "mem_accounts_asset" ON "mem_accounts" USING gin ("asset");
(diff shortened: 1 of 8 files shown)


michielmulders:
Or we include this in `release/2.0`?

Current migration `ALTER TABLE "mem_accounts" ADD COLUMN IF NOT EXISTS "asset" jsonb NOT NULL DEFAULT '{}'::jsonb;`

MaciejBaj:
Shouldn't affect any block processing functions

michielmulders:
@MaciejBaj no should not

framework/src/modules/chain/components/storage/sql/migrations/updates/20190410112400_add_asset_field_mem_accounts.sql
```diff
@@ -19,7 +19,7 @@
 */
 
  -- Add asset column to trs table as jsonb
-ALTER TABLE "mem_accounts" ADD COLUMN IF NOT EXISTS "asset" jsonb;
+ALTER TABLE "mem_accounts" ADD COLUMN IF NOT EXISTS "asset" jsonb NOT NULL DEFAULT '{}'::jsonb;
```
lsilvs: If this PR is merged into `2.1`, updating from `2.0` to `2.1` won't apply this change.

framework/src/modules/chain/components/storage/sql/migrations/updates/20190524120600_add_default_value_asset_mem_account.sql
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+
+/*
+   DESCRIPTION: Alter null value to empty JSON object {} for asset field for mem_accounts column
+   PARAMETERS: None
+*/
+
+ -- Alter asset field to empty JSON
+UPDATE mem_accounts SET asset = '{}'::json;
```
lsilvs:
```suggestion
UPDATE mem_accounts SET asset = '{}'::json WHERE asset IS NULL;

```

Issue #3426, opened 2019-04-23
Replace sodium-native with @liskhq/lisk-cryptography


MaciejBaj:
### Expected behavior
There should be one library wrapping `libsodium` used in Lisk SDK.

### Actual behavior
There are two libraries wrapping `libsodium` used in Lisk SDK currently:
"@liskhq/lisk-cryptography" and "sodium-native".

Both of them expose a similar API.

### Which version(s) does this affect? (Environment, OS, etc...)
`1.6.0`

PR #3722, opened 2019-05-23
Updated signature and heightPrevious / heightPrevoted with support old implementation - Closes #3558, #3556

### LINKED ISSUE #3556: Introduce heightPrevious and heightPrevoted properties in block headers
### Expected behavior
"Introduce BFT consensus protocol" LIP introduces two new properties, namely "heightPrevious" and "heightPrevoted". These properties should be included in the block header, used in calculating bytes for block and saved in the database.
### Actual behavior
"heightPrevious" and "heightPrevoted" do not exist in the block header.
3.0.0

### LINKED ISSUE #3558: Include height property in calculating the signature of the block
### Expected behavior
The height property defined in block header should be included in the getBytes() function which in turn is used to calculate the signature of the block.
### Actual behavior
The height property is not used to calculate the signature.
3.0.0

michielmulders:
### What was the problem?
Addresses both the changes needed for heightPrevious and heightPrevoted properties, and supports both old an new implementation based upon `bftUpgradeHeight` property which can be set to tell the application to switch to the new block headers with heightPrevious/prevoted included.
--> Uses migration to alter when upgrading, does not affect old behavior.

As the `getBytes()` function is needed for this change, it's included in the PR. It adds the `height`, `heightPrevious`, and `heightPrevoted` to the signature.

_Edit: Property names changed to: `maxHeightPreviouslyForged` and `prevotedConfirmedUptoHeight`._

### How to test it?
Run network, check for both new block header properties to be present in the database for `blocks`.

### Review checklist

* The PR resolves #3556, #3558 
* Linting passes
* Tests pass


framework/src/modules/chain/logic/block/index.js
```diff
@@ -0,0 +1,210 @@
...
+			ed,
+			schema,
+			transaction,
+			bftUpgradeHeight,
+		};
+
+		this.blockOld = new BlockOld(ed, schema, transaction);
+		this.blockBFT = new BlockBFT(ed, schema, transaction);
+
+		this.attachSchema();
```
nazarhussain: All of the code in this class can be avoided by; 

```suggestion
		this.attachSchema();
		
		['create', 'sign', 'getHash', 'verifyBytes'].map(key => {
		   this[key] = (...args) => {
		      return this.blockLogicForHeight(args[0].height)[key](args); 
		   }
		});
```
```

File: framework/src/modules/chain/logic/block/index.js [hunk 2]
```diff
@@ -0,0 +1,210 @@
+const BlockOld = require('./block');
+const BlockBFT = require('./block_bft');
+
+class Block {
+	constructor(ed, schema, transaction, bftUpgradeHeight) {
```
nazarhussain: This class is behaving like proxy to the right block logic. You can explore using javascript proxy class that may be helpful in this context. 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy

framework/src/modules/chain/logic/block/block_bft.js
```diff
@@ -0,0 +1,573 @@
...
+ * @requires crypto
+ * @requires helpers/bignum
+ * @requires logic/block_reward
+ * @param {Object} ed
+ * @param {ZSchema} schema
+ * @param {Transaction} transaction
+ * @returns {SetImmediate} error, this
+ * @todo Add description for the params
+ */
+class BlockBFT {
```
nazarhussain: Why don't we just extend the basic `Block` and override the functions which have different implementation. e.g. `getBytes` and `create`

framework/src/modules/chain/logic/block/block_bft.js
```diff
@@ -0,0 +1,573 @@
...
+			8 + // totalFee (long)
+			8 + // reward (long)
+			4 + // payloadLength (int)
+			32 + // payloadHash
+			32 + // generatorPublicKey
+			64 + // blockSignature or unused
+			4; // unused
+		let bytes;
+		try {
+			const byteBuffer = new ByteBuffer(capacity, true);
```
shuse2: ByteBuffer library is actually not needed.
Please use the native buffer.
Original code will be changed in https://github.com/LiskHQ/lisk-sdk/issues/3426

michielmulders:
Currently, block version exceptions hardcoded in `modules/chain/blocks/block_version.js` -> Need a strategy for inserting when other PRs related to this are completed. From height 8, block version 2 starts.

michielmulders:
Tests failing for now as exceptions are hardcoded in `block_version.js`. Blocked by PR to inject dynamically and being able to stub in the tests. Code can be reviewed.

framework/src/modules/chain/blocks/block.js
```diff
@@ -160,7 +235,7 @@ const objectNormalize = (block, exceptions = {}) => {
  * @returns {block} block
  * @todo Add description for the params
  */
-const create = ({
+const createV1 = ({
```
shuse2: probably it's better to have 2 separate file with the same interface

framework/src/modules/chain/blocks/block_version.js
```diff
@@ -19,7 +19,20 @@
  *
  * @property {number} currentBlockVersion - Current block version used for forging and verify
  */
-const currentBlockVersion = 1;
+const currentBlockVersion = 2;
+
+const exceptions = {
```
shuse2: this is defined in the config already
can't we use it?

framework/src/modules/chain/logic/block/block_bft.js
michielmulders: #3807 issue responsible for this change

framework/src/modules/chain/blocks/block.js
```diff
@@ -120,6 +120,81 @@ const getBytes = block => {
...
+ * @returns {!Array} Contents as an ArrayBuffer
+ * @todo Add description for the function and the params
+ */
+const getBytesV2 = block => {
+	const capacity =
+		4 + // version (int)
+		4 + // timestamp (int)
+		4 + // height (int)
+		4 + // maxHeightPreviouslyForged (int)
+		4 + // prevotedConfirmedUptoHeight (int)
```
nazarhussain: The LIP suggest that these new properties to be added after `previousBlock`, not before it.

https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#additional-block-header-properties

> The properties `height`, `heightPrevious`, `heightPrevoted` should be included in the byte array in this order directly after the `previousBlock` property.

framework/src/modules/chain/blocks/block_version.js
```diff
@@ -19,7 +19,20 @@
...
+		'1': {
+			start: 1,
+			end: 7,
+		},
+		'2': {
+			start: 8,
+			end: 6901027,
+		},
+	},
+}; // TODO BFT: define strategy to inject this variable from chain config
```
nazarhussain: These exceptions comes from the configuration, and these will be different for each network, so you can't use static definition here.

framework/src/components/storage/entities/block.js
```diff
@@ -167,6 +169,8 @@ class Block extends BaseEntity {
 		this.addField('rowId', 'number');
 		this.addField('id', 'string', { filter: filterType.TEXT });
 		this.addField('height', 'number', { filter: filterType.NUMBER });
+		this.addField('maxHeightPreviouslyForged', 'number');
```
nazarhussain:
```suggestion
		this.addField('maxHeightPreviouslyForged', 'number', { filter: filterType.NUMBER });
```

framework/src/components/storage/entities/block.js
```diff
@@ -167,6 +169,8 @@ class Block extends BaseEntity {
...
+		this.addField('prevotedConfirmedUptoHeight', 'number');
```
nazarhussain:
```suggestion
		this.addField('prevotedConfirmedUptoHeight', 'number', { filter: filterType.NUMBER });
```

framework/src/modules/chain/blocks/process.js
```diff
@@ -124,6 +127,10 @@ class BlocksProcess {
 			maxPayloadLength: this.constants.maxPayloadLength,
 			keypair,
 			timestamp,
+			prevotedConfirmedUptoHeight: 1,
+			maxHeightPreviouslyForged: 1,
+			height: context.blockHeight,
```
shuse2: I think it's more clear not to use the context here

framework/src/modules/chain/blocks/process.js
```diff
@@ -92,7 +92,10 @@ class BlocksProcess {
 		const context = {
 			blockTimestamp: timestamp,
 			blockHeight: lastBlock.height + 1,
-			blockVersion: blockVersion.currentBlockVersion,
+			blockVersion: blockVersion.getBlockVersion(
+				lastBlock.height + 1,
+				this.exceptions
```
shuse2: you shouldn't need to pass the `exceptions` here because it should only generate the latest block version

framework/src/modules/chain/blocks/blockV1.js
```diff
@@ -0,0 +1,465 @@
+/*
```
nazarhussain: We use `underscore` convention for file names, rather camelcase.

framework/src/modules/chain/blocks/blockV2.js
```diff
@@ -0,0 +1,471 @@
+/*
```
nazarhussain: Same file name issue here as well.

framework/src/modules/chain/blocks/blockV1.js
```diff
@@ -0,0 +1,465 @@
...
+		'totalAmount',
+		'totalFee',
+		'reward',
+		'transactions',
+		'version',
+	],
+};
+
+module.exports = {
+	getBytesV0,
```
nazarhussain: This is `blocks_v1.js` file, why it even have `getBytesV0`, `storageReadV0`?

framework/src/modules/chain/blocks/blockV2.js
```diff
@@ -0,0 +1,471 @@
...
+		'totalAmount',
+		'totalFee',
+		'reward',
+		'transactions',
+		'version',
+	],
+};
+
+module.exports = {
+	getBytesV2,
```
nazarhussain: Since the file itself is `blocks_v2.js`, we should export the function names from there with out postfix, so simple `getBytes`, `create`, `dbRead`

framework/src/modules/chain/blocks/blockV1.js
```diff
@@ -0,0 +1,465 @@
...
+
+const { Status: TransactionStatus } = require('@liskhq/lisk-transactions');
+const {
+	getAddressFromPublicKey,
+	hexToBuffer,
+	signDataWithPrivateKey,
+	hash,
+	verifyData,
+} = require('@liskhq/lisk-cryptography');
+const _ = require('lodash');
```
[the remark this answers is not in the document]
michielmulders: Not really scope of this PR, but changed it to meet the new standards.

framework/src/modules/chain/blocks/blockV1.js
```diff
@@ -0,0 +1,465 @@
...
+	createV1,
+	dbReadV0,
+	dbReadV1,
+	storageReadV0,
+	storageReadV1,
+	verifySignature,
+	getId,
+	getHash,
+	objectNormalize,
+	sign,
```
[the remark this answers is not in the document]
michielmulders: Version 0 and 1 are same (compatible), sometimes we have block version `0` but this can be verified by version 1 function.

framework/src/modules/chain/blocks/block_version.js
```diff
@@ -49,7 +49,32 @@ const isValid = (version, height, exceptions = {}) => {
 	return Number(exceptionVersion) === version;
 };
 
+const getBlockVersion = (height, exceptions = {}) => {
+	if (height === undefined || Object.entries(exceptions).length === 0) {
```
limiaspasdaniel: `if(height === undefined || !Object.keys(exceptions))` I think is easier to read and also  `Object.entries` is very slow

framework/src/modules/chain/blocks/blockV2.js
```diff
@@ -0,0 +1,471 @@
...
+
+/**
+ * Creates bytebuffer out of block data used for signatures.
+ *
+ * @param {block} block
+ * @throws {Error}
+ * @returns {!Array} Contents as an ArrayBuffer
+ * @todo Add description for the function and the params
+ */
+const getBytesV2 = block => {
```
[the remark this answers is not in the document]
michielmulders: This has been changed already, I think Github messed up.
File is now called `block_v2.js` and has no `V(#)` references anymore, but good catch!

framework/src/modules/chain/blocks/block_version.js
```diff
@@ -19,7 +19,7 @@
  *
  * @property {number} currentBlockVersion - Current block version used for forging and verify
  */
-const currentBlockVersion = 1;
+const currentBlockVersion = 2;
```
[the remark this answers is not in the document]
michielmulders: Yes, should be, open issue for that?

framework/src/modules/chain/blocks/block_version.js
michielmulders: `Object.keys` works better thanks! Need the explicit check tho.

limiaspasdaniel:
Overall looks good! Tests are not being executed, please make sure they can be run.

framework/src/modules/chain/blocks/block_version.js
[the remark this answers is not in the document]
michielmulders: https://github.com/LiskHQ/lisk-sdk/issues/3828

framework/test/mocha/unit/modules/chain/blocks/block.js
```diff
@@ -294,6 +351,37 @@ describe('block', () => {
 	});
 
 	describe('create', () => {
+		describe('createV1', () => {
```
SargeKhan: Maybe it makes sense to have two different test files for checking block v1 and v2?

nazarhussain: I agree, the the unit test must follow the exact same file structure as source files.

michielmulders: Addressed this comment, thanks! :)

framework/test/mocha/unit/modules/chain/blocks/block_version.js
```diff
@@ -159,18 +219,18 @@ describe('block_version', () => {
...
 
-			it('should return true for version = 1, height = 102', async () => {
+			it('should return true for version = 2, height = 102', async () => {
 				const height = 102;
-				const version = 1;
+				const version = 2;
 
 				return expect(
 					blockVersion.isValid(version, height, exceptions)
 				).to.equal(true);
```
SargeKhan: I don't understand why it's returning true when the exceptions are not present at all?

SargeKhan: Can you double check these tests?

framework/test/mocha/unit/modules/chain/blocks/verify.js
```diff
@@ -316,9 +316,9 @@ describe('blocks/verify', () => {
 	describe('verifyVersion', () => {
 		describe('when there are no exceptions for block versions', () => {
 			describe('when block height provided', () => {
-				it('should return no error when block version = 1', async () => {
+				it('should return no error when block version = 2', async () => {
```
SargeKhan: If exceptions are not present for block versions, doesn't it mean that we don't have block version updates at heights? If so, why do the block_version return an error?

framework/test/mocha/unit/modules/chain/blocks/block_v1.js
```diff
@@ -93,11 +93,23 @@ describe('block', () => {
 			transactions: [],
 			blockSignature:
 				'8a727cc77864b6fc81755a1f4eb4796b68f4a943d69c74a043b5ca422f3b05608a22da4a916ca7b721d096129938b6eb3381d75f1a116484d1ce2be4904d9a0e',
-			height: 69,
+			height: 6,
 			id: '3920300554926889269',
 			relays: 1,
 		},
 		transactions: [],
+		exceptions: {
```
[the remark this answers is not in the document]
michielmulders: I was passing this to check against specific heights, but not needed anymore now tests are split between two files (v1 and v2).

framework/test/mocha/unit/modules/chain/blocks/block_v1.js
```diff
@@ -93,11 +93,23 @@ describe('block', () => {
 			transactions: [],
 			blockSignature:
 				'8a727cc77864b6fc81755a1f4eb4796b68f4a943d69c74a043b5ca422f3b05608a22da4a916ca7b721d096129938b6eb3381d75f1a116484d1ce2be4904d9a0e',
-			height: 69,
+			height: 6,
```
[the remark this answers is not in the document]
michielmulders: Same here, was when tests were together. Can be put back to the normal value yes for clarity.

framework/test/mocha/unit/modules/chain/blocks/block_version.js
michielmulders: For this test, there is one exception present and the current block version is set to 2.
So it returns true for these tests.

framework/test/mocha/unit/modules/chain/blocks/verify.js
michielmulders: It returns always the current block version which is now set to `2`.

framework/test/mocha/unit/modules/chain/blocks/block_version.js
michielmulders: Currently `Currentblockversion = 2` so the test are returning true correctly.

PR #3725, opened 2019-05-24
Implement BFT fork choice rules

### LINKED ISSUE #3559: Implement the fork choice rules described in "Introduce BFT consensus protocol" LIP
### Expected behavior
The LIP "Introduce BFT consensus protocol" introduces fork choice rules for checking if the block should be accepted/rejected. These rules should be implemented.
### Actual behavior
The fork choice rules based on "Introduce BFT consensus protocol" are not implemented.
3.0.0

limiaspasdaniel:
### What was the problem?
New fork choice rules needed to be implemented to comply with the new BFT consensus protocol
### How did I fix it?
Implementing Fork Choice Rules according to [LIP-0014](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md)
### How to test it?
Run unit tests for `blocks/process.js`
### Review checklist
* The PR resolves #3559 
* Linting passes
* Tests pass


framework/src/modules/chain/submodules/blocks/process.js
```diff
@@ -427,11 +435,242 @@ class Process {
...
+		if (lastBlock.id === block.id) {
+			// Case 1: same block received twice
+			return this._handleSameBlockReceived(block);
+		}
+
+		if (
+			lastBlock.height + 1 === block.height &&
+			block.previousBlock === lastBlock.id
+		) {
+			// Case 2: correct block received
```
nazarhussain: As per the LIP its the right order. But if we just reorder the logic to handle good block in start, and then all invalid cases afterwards. That will be more easy to read.

framework/src/modules/chain/submodules/blocks/process.js
```diff
@@ -427,11 +435,242 @@ class Process {
...
+			return this._handleGoodBlock(block);
+		}
+
+		if (
+			lastBlock.height === block.height &&
+			lastBlock.heightPrevoted === block.heightPrevoted &&
```
nazarhussain: Please use the different names for these attributes as we discussed earlier in a meeting. https://github.com/LiskHQ/lisk-sdk/blob/64488f43f7e56d6b963780aabd0f034c6eeb8895/framework/test/mocha/fixtures/blocks.js#L77-L84

framework/src/modules/chain/submodules/blocks/process.js
```diff
@@ -427,11 +435,242 @@ class Process {
...
+
+			// Two competing blocks by different delegates at the same height.
+			if (
+				forgingSlotLastBlock < forgingSlotNewBlock &&
+				!this._receivedInSlot(lastBlock, lastBlockReceivedAt) &&
+				this._receivedInSlot(block, newBlockReceivedAt)
+			) {
+				// Case 4: Tie break
+				return this._handleDoubleForgingTieBreak(block, lastBlock);
+			}
```
nazarhussain: We are missing a default action under the parent if condition. its based on assumption that no other condition is possible, we should double check it.

framework/src/modules/chain/submodules/blocks/process.js
```diff
@@ -427,11 +435,242 @@ class Process {
...
+
+		if (
+			lastBlock.height + 1 === block.height &&
+			block.previousBlock === lastBlock.id
+		) {
+			// Case 2: correct block received
+			return this._handleGoodBlock(block);
+		}
+
+		if (
```
nazarhussain: IfI would prefer to not have nested condition. That's make it complicated to read and maintain. A suggestion is to make these flat. 

```js
const isDuplicateBlock = lastBlock.height === block.height && lastBlock.heightPrevoted === block.heightPrevoted && lastBlock.previousBlock === block.previousBlock; 

if(isDuplicateBlock && generatorPublicKey === block.generatorPublicKey) {
    return this._handleDoubleForging(block);
}

if(isDuplicateBlock && forgingSlotLastBlock < forgingSlotNewBlock &&
 				!this._receivedInSlot(lastBlock, lastBlockReceivedAt) &&
 				this._receivedInSlot(block, newBlockReceivedAt))  {
   return this._handleDoubleForgingTieBreak(block, lastBlock);
}

```

framework/src/modules/chain/submodules/blocks/process.js
```diff
@@ -427,11 +435,242 @@ class Process {
...
+	 * @returns {*}
+	 * @private
+	 */
+	// eslint-disable-next-line class-methods-use-this
+	_handleDoubleForging(block) {
+		library.logger.warn(
+			'Delegate forging on multiple nodes',
+			block.generatorPublicKey
+		);
+		library.logger.info('Last block stands');
```
nazarhussain: Please elaborate this log message further along with the ids of both blocks.

framework/src/modules/chain/submodules/blocks/process.js
```diff
@@ -427,11 +435,242 @@ class Process {
...
+				'Case 4 fork recovery failed',
+				convertErrorsToString(error)
+			);
+			throw error;
+		}
+
+		// If the new block is correctly validated and verified,
+		// last bloc kis deleted and new block is added to the tip chain
+		library.logger.info('Deleting last block due to case 4');
+		await deleteLastBlock();
```
nazarhussain: The LIP says that we keep reference of the old block, and in case we can't process the new block, we restore the old one.

framework/src/modules/chain/submodules/blocks/process.js
```diff
@@ -427,11 +435,242 @@ class Process {
...
+	}
+
+	/**
+	 * Update last receipt and process newly received block.
+	 * @param block
+	 * @private
+	 */
+	// eslint-disable-next-line class-methods-use-this
+	_processBlock(block) {
+		return promisify(__private.receiveBlock)(block); // TODO: Convert __private.receiveBlock to async, move implementation here to async and remove from __private.
```
nazarhussain: Why to promising every time? Please create a proxy function `__private.receiveBlockPromisify` and use it here.

framework/src/modules/chain/submodules/blocks/process.js
```diff
@@ -427,11 +435,242 @@ class Process {
...
+			}
+		} catch (error) {
+			// Should there be any errors while validating and verifying,
+			// last block is kept as the tip of the chain
+			library.logger.error(
+				'Case 4 fork recovery failed',
+				convertErrorsToString(error)
+			);
+			throw error;
+		}
```
nazarhussain: Don't we already have this validation logic? Can't we use `modules.blocks.verify.procssBlock` here?

framework/src/modules/chain/submodules/blocks/process.js
michielmulders: Yea, Jan suggested creating a temporary table to store it in case something goes wrong.

framework/src/modules/chain/submodules/blocks/process.js
```diff
@@ -427,11 +435,242 @@ class Process {
...
+	 */
+	// eslint-disable-next-line class-methods-use-this
+	_forkChoice(block) {
+		// Current slot number based on current time since LiskEpoch ~ Slot number at the time the new block is received
+		// Better to do it here rather than in the Sequence so reciving time is more accurate
+		const newBlockReceivedAt = slots.getTime();
+
+		// Execute in sequence via sequence
+		return library.sequence.add(callback => {
+			this._forkChoiceTask(block, newBlockReceivedAt)
```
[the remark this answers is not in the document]
limiaspasdaniel: You're right, I have moved it to the upper scope.

framework/src/modules/chain/submodules/blocks/process.js
```diff
@@ -427,11 +435,242 @@ class Process {
...
+	 */
+	async _forkChoiceTask(block, newBlockReceivedAt) {
+		const lastBlock = modules.blocks.lastBlock.get();
+
+		const forgingSlotLastBlock = slots.getSlotNumber(lastBlock.timestamp);
+		const forgingSlotNewBlock = slots.getSlotNumber(block.timestamp);
+
+		// Slot number when lastBlock was received.
+		const lastBlockReceivedAt =
+			modules.blocks.lastReceipt.get() || lastBlock.timestamp;
```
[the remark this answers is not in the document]
limiaspasdaniel: As per the current implementation, lastReceipt is not updated if the node starts explicitly asking blocks when syncing. We need to see what to do with this so it is consistent when syncing, asking for blocks, or receiving blocks passively (which in this particular case, lastReceipt is indeed updated).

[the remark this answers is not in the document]
limiaspasdaniel: Also, lastReceipt is not based on Lisk's Epoch time while block.timestamp is. This is inconvenient and I would suggest updating lastReceipt.update() function so it is indeed based on it. We now need these two properties to be consistent regarding Epoch times so we can calculate if a block was received in its designed slot.

framework/src/modules/chain/submodules/blocks/process.js
limiaspasdaniel: Do we need to create a table for this? If process block fails it will return an exception that can be caught and last block can be restored back again.

framework/src/modules/chain/submodules/blocks/process.js
limiaspasdaniel: `this._processBlock()` call here is actually executing the old function:

```
__private.receiveBlock = function(block, cb) {
	library.logger.info(
		`Received new block id: ${block.id} height: ${
			block.height
		} round: ${slots.calcRound(block.height)} slot: ${slots.getSlotNumber(
			block.timestamp
		)} reward: ${block.reward}`
	);

	// Update last receipt
	modules.blocks.lastReceipt.update(); // TODO: Probably not needed anymore.
	// Start block processing - broadcast: true, saveBlock: true
	modules.blocks.verify.processBlock(block, true, true, cb);
};
```

Which is intended to be refactored and removed from _private after compatibility window is over. It is not only calling modules.blocks.verify.processBlock but also logging and updating lastReceipt as you can see above.

michielmulders: @shuse2 is working on this right?

michielmulders: I'm waiting for that PR to make my changes as it will cause a lot of conflicts.

limiaspasdaniel:
@nazarhussain, regarding your suggestion to make the logic more readable:

I like the format and it definitely reads better and clearer, however, I wouldn't nest functions inside functions, this complicates testing. What I can do instead is move those functions as private functions with the following format `this._functionName` as we are not using `_private` anymore. This way we can still apply your

framework/src/modules/chain/submodules/blocks/process.js
limiaspasdaniel: I have changed the log so it includes both block ids and a clear message.

framework/src/modules/chain/submodules/blocks/process.js
limiaspasdaniel: It definitely improves the current implementation. I have updated the code with your suggested changes. Thanks!

framework/src/modules/chain/submodules/blocks/process.js
limiaspasdaniel: Good catch, done.

framework/src/modules/chain/submodules/blocks/process.js
limiaspasdaniel: Done.

framework/src/modules/chain/submodules/blocks/process.js
```diff
@@ -427,11 +444,244 @@ class Process {
...
 	}
+
+	/**
+	 * Wrap of fork choice rule logic so it can be added to Sequence and properly tested
+	 * @param block
+	 * @param newBlockReceivedAt - Time when the new block was received since Lisk Epoch
+	 * @return {Promise}
+	 * @private
+	 */
+	async _forkChoiceTask(block, newBlockReceivedAt) {
```
shuse2: can we change this all fork choice rules to different file, and be a pure function?

framework/src/modules/chain/submodules/blocks/process.js
nazarhussain: Yes I agree we don't need store those blocks to table, just keep in memory. Even if node crashes, it can sync to network. There is no fatal, non-recoverable lost.

framework/src/modules/chain/submodules/blocks/process.js
limiaspasdaniel: I think we can do this when the compatibility window is over as this file currently needs logic that is shared between old fork choice rules and new ones, and it will be easier to extract as old functions will be removed.

framework/src/modules/chain/submodules/blocks/process.js
michielmulders: To clarify this, didn't we decide on storing those blocks in a temporary table as that's more convenient?

limiaspasdaniel: I think in this particular case were we only keep the backup of the last block in memory, maybe we don't need to do that. If something fails in between and last block is lost, it will sync again

framework/test/mocha/unit/modules/chain/submodules/blocks/process.js
```diff
@@ -1596,165 +1598,419 @@ describe('blocks/process', () => {
...
+						);
+						expect(loggerStub.debug.args[0][1]).to.equal('2');
+						done();
 					});
-					expect(__private.receiveForkFive.args[0][1]).to.deep.equal({
-						id: '2',
-						previousBlock: '1',
-						height: 2,
+
+					it('should log debug message and call a callback', done => {
```
michielmulders: Is it possible you are not checking the log debug message for this test?

limiaspasdaniel: Indeed. I am currently improving the tests so I will push the changes again shortly.

framework/src/modules/chain/blocks/block_slots.js
```diff
@@ -115,6 +115,16 @@ class BlockSlots {
 		return nextSlot + this.blocksPerRound;
 	}
 
+	/**
+	 * Check if timestamp is contained within a slot time window
+	 * @param slot
+	 * @param time
+	 * @returns {boolean}
+	 */
+	timeFallsInSlot(slot, time) {
```
shuse2: Maybe better naming?
`IsWithinTimeslot` or something that indicates it returns boolean?

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -302,6 +304,26 @@ class Blocks extends EventEmitter {
 
 	// Process a block from the P2P
 	async receiveBlockFromNetwork(block) {
+		this.logger.info(
```
shuse2: Blocks module is too low level to have `info` logs.
Within module, it should be maximum `debug`

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -302,6 +304,26 @@ class Blocks extends EventEmitter {
...
+				block.height
+			} round: ${this.slots.calcRound(
+				block.height
+			)} slot: ${this.slots.getSlotNumber(block.timestamp)} reward: ${
+				block.reward
+			} version: ${block.version}`
+		);
+
+		const receiveBlockImplementations = {
+			1: this.receiveBlockFromNetworkV1,
```
shuse2: I think we have similar idea.
Let's align with team tomorrow how to switch between the codes

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -449,6 +456,120 @@ class Blocks extends EventEmitter {
...
+	 * @private
+	 */
+	async _forkChoiceTask(block, newBlockReceivedAt) {
+		// Cases are numbered following LIP-0014 Fork choice rule.
+		// See: https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#applying-blocks-according-to-fork-choice-rule
+		// Case 2 and 1 have flipped execution order for better readability. Behavior is still the same
+
+		if (forkChoiceRule.isValidBlock(this._lastBlock, block)) {
+			// Case 2: correct block received
+			return this._handleGoodBlock(block);
```
shuse2: maybe `_handleValidBlock`?

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -584,6 +705,130 @@ class Blocks extends EventEmitter {
...
+		this.logger.debug('Block already processed', block.id);
+	}
+
+	/**
+	 * Block received is correct
+	 * @param block
+	 * @returns {Promise}
+	 * @private
+	 */
+	_handleGoodBlock(block) {
```
shuse2: async?

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -584,6 +705,130 @@ class Blocks extends EventEmitter {
...
+
+	/**
+	 * Block IDs are the same ~ Blocks are equal
+	 * @param block
+	 * @returns {*}
+	 * @private
+	 */
+	// eslint-disable-next-line class-methods-use-this
+	_handleSameBlockReceived(block) {
+		this.logger.debug('Block already processed', block.id);
```
shuse2: is this just for log or TODO?
I think we don't need to create a function just for log 😉

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -584,6 +705,130 @@ class Blocks extends EventEmitter {
...
+	/**
+	 * Handle discarded block determined by the fork choice rule
+	 * @param block
+	 * @return {*}
+	 * @private
+	 */
+	// eslint-disable-next-line class-methods-use-this
+	_handleDiscardedBlock(block) {
+		// Discard newly received block
+		return this.logger.warn(
```
shuse2: log does not return anything (ie no return required.
Also, same as above, warn level should not be used within this module

framework/src/modules/chain/blocks/block_slots.js
limiaspasdaniel: Definitely better, thanks :)

framework/src/modules/chain/blocks/blocks.js
limiaspasdaniel: I will change info logs to debugs for new code, however refactored and legacy code still uses info in this file.

framework/src/modules/chain/blocks/blocks.js
limiaspasdaniel: A mix of both and also for the sake of consistency. I am unsure on how in the function will evolve in the future or of new logic will need to be added but in my opinion, leaving it this way readability on `_forkChoiceTask` increases.

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -449,6 +456,120 @@ class Blocks extends EventEmitter {
...
+					block.height
+				} round: ${this.slots.calcRound(
+					block.height
+				)} slot: ${this.slots.getSlotNumber(block.timestamp)} reward: ${
+					block.reward
+				} version: ${block.version} with error: ${error}`
+			);
+			this._isActive = false;
+			throw error;
+		}
```
nazarhussain: Add a block `finally` to set `this._isActive = false;` or similar postrequisits. That will make it more understandable and align with promise convention.

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -113,6 +115,11 @@ class Blocks extends EventEmitter {
 			blockReward: this.blockReward,
 			constants: this.constants,
 		});
+
+		this._receiveBlockImplementations = {
+			1: this._receiveBlockFromNetworkV1.bind(this),
```
michielmulders: Can we avoid using `bind()` ?

framework/src/modules/chain/blocks/fork_choice_rule.js
```diff
@@ -0,0 +1,99 @@
...
+	lastBlock.height === currentBlock.height &&
+	lastBlock.prevotedConfirmedUptoHeight ===
+		currentBlock.prevotedConfirmedUptoHeight &&
+	lastBlock.previousBlock === currentBlock.previousBlock;
+
+/**
+ * Determine if Case 3 fulfills
+ * @param lastBlock
+ * @param currentBlock
+ * @return {*|boolean}
```
michielmulders: Is this not just a boolean being returned? (Also for the other functions)

framework/src/modules/chain/blocks/blocks.js
shuse2:
```
1: block => this. _receiveBlockFromNetworkV1(block)
```

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -459,6 +466,121 @@ class Blocks extends EventEmitter {
 		});
 	}
 
+	async _receiveBlockFromNetworkV2(block) {
+		// Current time since Lisk Epoch
+		// Better to do it here rather than in the Sequence so receiving time is more accurate
+		const newBlockReceivedAt = this.slots.getTime();
+
+		// Execute in sequence via sequence
```
shuse2:
```suggestion
		// Execute in sequence

File: framework/src/modules/chain/blocks/blocks.js [hunk 24]
```diff
@@ -17,6 +17,7 @@
 const EventEmitter = require('events');
 const { cloneDeep } = require('lodash');
 const blocksUtils = require('./utils');
+const { convertErrorsToString } = require('../helpers/error_handlers');
```
shuse2: No need to use this anymore, our new logger can put object as first parameter

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -594,6 +716,130 @@ class Blocks extends EventEmitter {
...
+		// last block is deleted and new block is added to the tip of the chain
+		this.logger.debug(
+			`Deleting last block with id: ${
+				lastBlock.id
+			} due to Fork Choice Rule Case 4`
+		);
+		const previousLastBlock = cloneDeep(this._lastBlock);
+
+		// Deletes last block and assigns new received block to this._lastBlock
+		await this.recoverChain();
```
shuse2: maybe better to consistently use the utils or rename this function

limiaspasdaniel: I think I will rename the function because its name is misleading

framework/src/modules/chain/blocks/blocks.js
limiaspasdaniel: Applied your suggestions, please give your input on `framework/src/modules/chain/blocks/blocks.js:777`

framework/src/modules/chain/blocks/blocks.js
limiaspasdaniel: Done

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -594,6 +716,130 @@ class Blocks extends EventEmitter {
...
+		const check = await this.blocksVerify.normalizeAndVerify(
+			block,
+			lastBlock,
+			this._lastNBlockIds
+		);
+
+		if (!check.verified) {
+			const errors = convertErrorsToString(check.errors);
+
+			this.logger.error(
```
shuse2: Ah sorry I meant
```
this.logger.error( { errors, blockId: block.id }, 'Fork Choice Case 4 recovery failed in verification and normalization');
```

framework/src/modules/chain/blocks/blocks.js
limiaspasdaniel: Solved

framework/src/modules/chain/blocks/fork_choice_rule.js
limiaspasdaniel: Solved

framework/src/modules/chain/blocks/blocks.js
limiaspasdaniel: I was checking Bunyan docs and you can do this.logger.error(error, message), the first argument being specifically for error objects

shuse2: yes you can, also i thought having meta data in object is probably better than message because its easier to filter

limiaspasdaniel: @shuse2 please see #3818

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -301,7 +307,7 @@ class Blocks extends EventEmitter {
 		this.logger.info('Blockchain ready');
 	}
 
-	async recoverChain() {
+	async deleteLastBlockAndGet() {
 		this._lastBlock = await this.blocksChain.deleteLastBlock(this._lastBlock);
```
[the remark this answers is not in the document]
limiaspasdaniel: I agree with you, we can address this in another issue

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -322,22 +343,13 @@ class Blocks extends EventEmitter {
...
-					this._isActive = false;
-					setImmediate(cb);
+					await this._processReceivedBlock(block);
 				} catch (error) {
-					this._isActive = false;
-					this.logger.error(error);
 					setImmediate(cb, error);
+					return;
 				}
+				setImmediate(cb);
```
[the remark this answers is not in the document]
limiaspasdaniel: This is legacy code, not the scope of this PR to address it.

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -123,6 +124,11 @@ class Blocks extends EventEmitter {
 			blockReward: this.blockReward,
 			constants: this.constants,
 		});
+
+		this._receiveBlockImplementations = {
+			1: block => this._receiveBlockFromNetworkV1(block),
```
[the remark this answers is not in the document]
michielmulders: `this` scope is not preserved when using shorthand, ignore this comment.

PR #3727, opened 2019-05-24
Move the type check from validateAsset to validate as a static field - Closes #3726

### LINKED ISSUE #3726: Don't assign a transaction type as a private variable
### Expected behavior
Transaction type to compare the type against, for example `const TRANSACTION_TRANSFER_TYPE = 0;` for type 0 transaction is assigned in a way that can be overridden by the transactions that inherit after `BaseTransction`. 

An example would be implementing the valid type as a static field: 
```
class TransferTransaction {
   protected static TYPE = 10;
 ...
```

### Actual behavior

The valid type is assigned as a private variable. Each function comparing a type (`validate`) needs to be re-implemented and cannot be overridden if a BaseTransction cannot access a valid transaction type.

`2.0.0`

MaciejBaj:
### What was the problem?
Transaction type check was implemented in `validateAsset` function while not being part of the asset. I moved it to `validate` in BaseTransaction.

Valid transaction type needs to be assigned as a static field instead of a private constant.

### Review checklist

* The PR resolves #3726
* Linting passes
* Tests pass

### DIFF
--- a/elements/lisk-transactions/src/0_transfer_transaction.ts
+++ b/elements/lisk-transactions/src/0_transfer_transaction.ts
@@ -30,8 +30,6 @@ import {
 	verifyBalance,
 } from './utils';
 
-const TRANSACTION_TRANSFER_TYPE = 0;
-
 export interface TransferAsset {
 	readonly data: string;
 }
@@ -49,6 +47,7 @@ export const transferAssetFormatSchema = {
 
 export class TransferTransaction extends BaseTransaction {
 	public readonly asset: TransferAsset;
+	protected static TYPE = 0;
 
 	public constructor(rawTransaction: unknown) {
 		super(rawTransaction);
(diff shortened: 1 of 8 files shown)


elements/lisk-transactions/src/base_transaction.ts
```diff
@@ -264,6 +266,20 @@ export abstract class BaseTransaction {
 			errors.push(idError);
 		}
 
+		/* tslint:disable */
+		if (this.type !== (<typeof BaseTransaction>this.constructor).TYPE) {
```
MaciejBaj:
```
ERROR: (no-angle-bracket-type-assertion) elements/lisk-transactions/src/base_transaction.ts[269, 22]: Type assertion using the '<>' syntax is forbidden. Use the 'as' syntax instead.
```

MaciejBaj: @shuse2 @lsilvs

shuse2: try
`(this.constructor as typeof BaseTransaction)`

elements/lisk-transactions/src/0_transfer_transaction.ts
```diff
@@ -49,6 +47,7 @@ export const transferAssetFormatSchema = {
 
 export class TransferTransaction extends BaseTransaction {
 	public readonly asset: TransferAsset;
+	protected static TYPE = 0;
```
shuse2: probably these should be public as it can be used in the framework too

elements/lisk-transactions/src/base_transaction.ts
lsilvs: Shouldn't this check be done in the constructor?

MaciejBaj: No

PR #3729, opened 2019-05-24
Remove callbacks from chain account logic - Closes #3540

### LINKED ISSUE #3540: Use async/await in chain module logic
### Expected behavior

Use async/await consistently across the application.

Don't introduce any changes to rounds! The parallel changes will be introduced to [Lisk SDK 3.0](https://github.com/LiskHQ/lisk-sdk/projects/7)

### Actual behavior

There is the use of callbacks in the constructor and methods of logic classes. Also, the sequence of callbacks is not consistent.

Related to #3396

`*`

ishantiw:
### Description

Remove the use of callback from the existing code in chain logic folder. 

Changes,

- Since `logic/Account.js` has only a function that is used on `rounds/round.js`  so we have moved `logic/Account.js` to `/rounds` folder
- Minor change to Account constructor, removed callback from the constructor.
- Related changes to the imports and tests.

### Review checklist

* The PR resolves #3540 
* Linting passes
* Tests pass


framework/src/modules/chain/init_steps/init_logic_structs.js
```diff
@@ -16,13 +16,9 @@
...
 	const Block = require('../logic/block.js');
-	const Account = require('../logic/account.js');
+	const Account = require('../rounds/account.js');
 
-	const accountLogic = await new Promise((resolve, reject) => {
-		new Account(storage, schema, logger, (err, object) => {
-			err ? reject(err) : resolve(object);
-		});
-	});
+	const accountLogic = new Account(storage, schema, logger);
```
shuse2: would be nice if we can remove from here, and instantiate within the rounds folder

framework/src/modules/chain/init_steps/init_logic_structs.js
```diff
@@ -14,16 +14,4 @@
...
-		new Account(storage, schema, logger, (err, object) => {
-			err ? reject(err) : resolve(object);
-		});
-	});
-
-	return {
-		account: accountLogic,
-	};
-};
+module.exports = () => {};
```
pablitovicente: Do we need to keep this file, maybe it's better to remove it?

framework/test/mocha/common/application.js
```diff
@@ -167,12 +167,9 @@ async function __init(sandbox, initScope) {
 		await cache.bootstrap();
 
 		scope.bus = await initSteps.createBus();
-		scope.logic = await initSteps.initLogicStructure(scope);
+		// scope.logic = await initSteps.initLogicStructure(scope);
```
pablitovicente: I would say let's remove the line instead of leaving it commented out

ishantiw:
@pablitovicente Addressed

PR #3731, opened 2019-05-27
API returns incorrect error for postTransactions - Closes #3586

### LINKED ISSUE #3586: API returns incorrect error for postTransactions

pablitovicente:
### What was the problem?

The error being used in postTransactions was not generic enough (as the detailed errors are always in the `errors` array )

### How did I fix it?

By changing the previous error to a more generic one

### How to test it?

- Build should be green

### Review checklist

* The PR resolves #3586
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/chain/submodules/transport.js
+++ b/framework/src/modules/chain/submodules/transport.js
@@ -144,10 +144,7 @@ __private.receiveSignature = function(signature, cb) {
  * @param {Array} transactions - Array of transactions
  * @param {string} extraLogMessage - Extra log message
  */
-__private.receiveTransactions = function(
-	transactions = [],
-	extraLogMessage
-) {
+__private.receiveTransactions = function(transactions = [], extraLogMessage) {
 	transactions.forEach(transaction => {
 		if (transaction) {
 			transaction.bundled = true;
@@ -208,9 +205,7 @@ __private.receiveTransaction = async function(
 	}
 
 	return library.balancesSequence.add(balancesSequenceCb => {
-		library.logger.debug(
-			`Received transaction ${transaction.id}`
-		);
+		library.logger.debug(`Received transaction ${transaction.id}`);
 
 		modules.transactions.processUnconfirmedTransaction(
 			transaction,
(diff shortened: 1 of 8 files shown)


framework/src/modules/chain/submodules/transport.js
```diff
@@ -698,7 +693,7 @@ Transport.prototype.shared = {
 				if (err) {
 					return setImmediate(cb, null, {
 						success: false,
-						message: err.message || 'Invalid transaction body',
+						message: err.message || 'Transaction was rejected with errors',
```
ManuGowda: if i understand correctly, @4miners found that the `err.message` was undefined as `err` was `type` `string`.
so can you handle the issue?

pablitovicente: @ManuGowda that assumption was not correct; err is sometimes an error and other times an array. so that part is correct.

ManuGowda: hmmm ok.. so that case as well.. `err.message` is wrong rite.. if its an array then `err.message` will result in undefined know? am i missing something?

PR #3733, opened 2019-05-27
Too many self connect errors from P2P

### LINKED ISSUE #3651: P2P errors while syncing with testnet

ishantiw:
### What was the problem?

Too many self connect errors

```
09:30:35.899Z DEBUG lisk-framework: Peer disconnect event: Outbound connection of peer 134.209.90.103:5000 was closed with code 4101 and reason: Peer cannot connect to itself
09:30:35.900Z ERROR lisk-framework: Event 'rpc-request' was aborted due to a bad connection
09:30:35.900Z DEBUG lisk-framework: Socket connection closed with status code 4101 and reason: Peer cannot connect to itself
```

### How did I fix it?

Since a node is finding itself during discovery from other peers so we add it to the blacklist the first time we see it.

### How to test it?

`npm run test:integration`

### Review checklist

* The PR resolves #3651
* Linting passes
* Tests pass

### DIFF
--- a/elements/lisk-p2p/src/p2p.ts
+++ b/elements/lisk-p2p/src/p2p.ts
@@ -154,10 +154,7 @@ export class P2P extends EventEmitter {
 
 	public constructor(config: P2PConfig) {
 		super();
-		this._config = {
-			...config,
-			blacklistedPeers: config.blacklistedPeers ? config.blacklistedPeers : [],
-		};
+		this._config = config;
 		this._isActive = false;
 		this._newPeers = new Map();
 		this._triedPeers = new Map();
--- a/elements/lisk-p2p/src/p2p.ts
+++ b/elements/lisk-p2p/src/p2p.ts
@@ -422,28 +422,14 @@ export class P2P extends EventEmitter {
 						INVALID_CONNECTION_SELF_REASON,
 					);
 
-					// Ignore these steps for local network tests since all the IPs will be localhost
-					if (socket.remoteAddress !== '127.0.0.1') {
-						const selfWSPort = queryObject.wsPort
-							? +queryObject.wsPort
-							: this._nodeInfo.wsPort;
-						const selfP2PInfo = {
-							ipAddress: socket.remoteAddress,
-							wsPort: selfWSPort,
-						};
-						// When a node tries to connect itself then add it to blacklist so that it doesn't connect itself.
-						this._config.blacklistedPeers = this._config.blacklistedPeers
-							? [...this._config.blacklistedPeers, selfP2PInfo]
-							: [selfP2PInfo];
+					const selfWSPort = queryObject.wsPort
+						? +queryObject.wsPort
+						: this._nodeInfo.wsPort;
 
-						const selfPeerId = constructPeerId(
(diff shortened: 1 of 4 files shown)


elements/lisk-p2p/src/p2p.ts
```diff
@@ -406,6 +412,29 @@ export class P2P extends EventEmitter {
 						INVALID_CONNECTION_SELF_REASON,
 					);
 
+					// Ignore these steps for local network tests since all the IPs will be localhost
+					if (socket.remoteAddress !== '127.0.0.1') {
```
shuse2: Probably this is only helpful in the case of local?
maybe instead, when you check the nonce and found out it's urself, just add the the blacklist?

elements/lisk-p2p/src/p2p_types.ts
```diff
@@ -72,7 +72,8 @@ export interface P2PClosePacket {
 }
 
 export interface P2PConfig {
-	readonly blacklistedPeers?: ReadonlyArray<P2PPeerInfo>;
+	// tslint:disable-next-line:readonly-keyword
+	blacklistedPeers: P2PPeerInfo[];
```
shuse2: do we need to change it here?
maybe holding instance type should be changed but not the P2P config?

elements/lisk-p2p/src/p2p.ts
```diff
@@ -406,6 +412,29 @@ export class P2P extends EventEmitter {
...
+						const selfWSPort = queryObject.wsPort
+							? +queryObject.wsPort
+							: this._nodeInfo.wsPort;
+						const selfP2PInfo = {
+							ipAddress: socket.remoteAddress,
+							wsPort: selfWSPort,
+						};
+						// When a node tries to connect itself then add it to blacklist so that it doesn't connect itself.
+						this._config.blacklistedPeers = this._config.blacklistedPeers
```
jondubois: Maybe `blacklistedPeers` should be stored as a private property on the `P2P` class (and always use that one instead of the one on the config)? That way we don't need to mutate the config object.

ishantiw:
As per discussion, I have added conditions after discovery and on incoming connections to avoid adding a node to itself in `newPeer` or `triedPeer` list, if exist then delete it. Adding an integration test scenario, to check whether the `peerInfo` update is applied properly in all the other nodes and update their `newPeer`, `triedPeer` or `connectedPeer` list if exist.

ishantiw:
Added scenario to check itself in both p2p-lib and functional test of the framework

PR #3740, opened 2019-05-28
Enable forging on Genesis Delegates while starting devnet - Closes #3734

### LINKED ISSUE #3734: Enable forging on Genesis Delegates while starting Devnet
### Expected behavior
Devnet should start forging by default on all Genesis Delegates/
### Actual behavior
Need to pass the genesis delegates passphrases in the application entry point.

`2.0.0`

MaciejBaj:
### What was the problem?
Each time I start an application created using Alpha SDK I need to specify the full set of delegates in the entry file of a project. 
### How did I fix it?

- Create a devnet genesis block containing all possible defaults 
- Assign chain default configuration using devnet sample 
- Assign http_api default configuration using devnet sample 
- Assign network default configuration using devnet sample
- Assign components defaults configuration using devnet sample

### How to test it?
Start a sample app and verify whether the "Forgin enabled" logs are there.
### Review checklist

* The PR resolves #3734 
* Linting passes
* Tests pass

### DIFF
--- a/sdk/src/samples/config_devnet.json
+++ b/sdk/src/samples/config_devnet.json
@@ -1,12 +1,12 @@
 {
 	"app": {
-		"ipc": {
-			"enabled": true
-		},
-		"label": "hello-world-app",
+		"label": "devnet-alpha-sdk",
 		"minVersion": "0.0.0",
 		"version": "0.0.0",
-		"protocolVersion": "0.0"
+		"protocolVersion": "1.1",
+		"ipc": {
+			"enabled": true
+		}
 	},
 	"components": {
 		"logger": {
--- a/sdk/src/samples/config_devnet.json
+++ b/sdk/src/samples/config_devnet.json
@@ -2,7 +2,11 @@
 	"app": {
 		"ipc": {
 			"enabled": true
-		}
+		},
+		"label": "hello-world-app",
+		"minVersion": "0.0.0",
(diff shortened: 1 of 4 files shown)


MaciejBaj:
@nazarhussain I will export the config filled with all of the possible defaults from lisk-sdk and use it in sample app by passing as a second parameter after GenesisBlock. 
I will remove the reference to this sample from the defaults.

sdk/src/samples/config_devnet.json
```diff
@@ -0,0 +1,543 @@
...
+			"logFileName": "logs/devnet/lisk.log",
+			"consoleLogLevel": "info"
+		},
+		"storage": {
+			"logFileName": "logs/devnet/lisk_db.log",
+			"host": "localhost",
+			"port": 5432,
+			"database": "lisk_dev",
+			"user": "lisk",
+			"password": "password",
```
nazarhussain: Suggest not to set username/password here. So while running the sample, user can use their own PostgreSQL user, rather than creating a new user. 

If usage of this file is intended for internal demo, then ok.

MaciejBaj: That password was "always" present here - the default `lisk` user set up is described in the Postgres installation manual.

MaciejBaj: I prefer to leave it this way, so there is a default setting.

PR #3756, opened 2019-05-31
Remove type requirement from register transaction - Closes #3523

### LINKED ISSUE #3523: Don’t use transaction type at the time of transaction registration
### Expected behavior

Transaction type should not be used as unique identifier for a transaction registration, we should be able to register transaction just specifying the matcher. 

```js
app.registerTransaction(MyTransaction, myMatcher);
```

This will give flexibility for the framework to run the matcher on transaction data itself that contains the type itself. So with this structure `type` will be ordinary transaction property not some identification key. 

Since framework transactions registered first and resolved by matcher first, so no problem if someone use custom transaction with same type value it will never be executed. 

In the long run we need to get rid of transaction type as identifying key. 

### Actual behavior

Current implementation of transaction registration use `type` as identifier that limits the flexibility of matcher. Also cause collision if we introduce new transaction in framework. 

2.0.0

mitsujutsu:
### What was the problem?

Transaction type required as unique identifier for a transaction registration in controller. 

### How did I fix it?

Modified `registerTransaction` in application controller to use static `Transaction.TYPE` from each transaction class in elements.

### Review checklist

* The PR resolves #3523 
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/controller/application.js
+++ b/framework/src/controller/application.js
@@ -225,7 +225,7 @@ class Application {
 	 * @param {number} transactionType - Unique integer that identifies the transaction type
 	 * @param {constructor} Transaction - Implementation of @liskhq/lisk-transactions/base_transaction
 	 */
-	registerTransaction(Transaction, matcher) {
+	registerTransaction(Transaction, { matcher }) {
 		assert(Transaction, 'Transaction implementation is required');
 
 		assert(
@@ -250,7 +250,7 @@ class Application {
 		}
 		const transactions = this.getTransactions();
 
-		transactions[Transaction.type] = Object.freeze(Transaction);
+		transactions[Transaction.TYPE] = Object.freeze(Transaction);
 		__private.transactions.set(this, transactions);
 	}
 
--- a/framework/src/controller/application.js
+++ b/framework/src/controller/application.js
@@ -16,6 +16,7 @@
 
 const assert = require('assert');
 const {
+	BaseTransaction,
 	TransferTransaction,
 	SecondSignatureTransaction,
 	DelegateTransaction,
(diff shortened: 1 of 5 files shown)


framework/src/controller/application.js
```diff
@@ -232,26 +225,32 @@ class Application {
 	 * @param {number} transactionType - Unique integer that identifies the transaction type
 	 * @param {constructor} Transaction - Implementation of @liskhq/lisk-transactions/base_transaction
 	 */
-	registerTransaction(transactionType, Transaction, options = {}) {
-		// TODO: Validate the transaction is properly inherited from base class
+	registerTransaction(Transaction, matcher) {
```
nazarhussain: To make sure future improvements may not break the API. 

```suggestion
	registerTransaction(Transaction, { matcher }) {
```

framework/src/controller/application.js
```diff
@@ -232,26 +225,32 @@ class Application {
...
 			Object.defineProperty(Transaction.prototype, 'matcher', {
-				get: () => options.matcher,
+				get: () => matcher,
 			});
 		}
-
 		const transactions = this.getTransactions();
-		transactions[transactionType] = Object.freeze(Transaction);
+
+		transactions[Transaction.type] = Object.freeze(Transaction);
```
nazarhussain:
```suggestion
		transactions[Transaction.TYPE] = Object.freeze(Transaction);
```

framework/src/controller/application.js
```diff
@@ -232,26 +225,32 @@ class Application {
...
-			!Object.keys(this.getTransactions()).includes(transactionType.toString()),
-			`A transaction type "${transactionType}" is already registered.`
+			!Object.keys(this.getTransactions()).includes(Transaction.TYPE),
+			`A transaction type "${Transaction.TYPE}" is already registered.`
 		);
-		assert(Transaction, 'Transaction implementation is required');
 
-		if (options.matcher) {
+		if (matcher) {
 			Object.defineProperty(Transaction.prototype, 'matcher', {
```
nazarhussain: This line refers to the fact that `matcher` can be defined in the `Transaction` class itself, instead of specifying it at time of registration.  Yes when you define at time of registration, it must override actual getter if exists. 

@MaciejBaj We should document it properly.

limiaspasdaniel:
@mitsujutsu I believe this breaks matcher integration tests. Could you please take a look at it?

PR #3757, opened 2019-05-31
Update BigNum to use @liskhq/bignum - Closes #3067

### LINKED ISSUE #3067: Update BigNum to use @liskhq/bignum
### Description

Update Bignum to use @liskhq/bignum package to be consistent with lisk-elements.

pablitovicente:
### What was the problem?

- Instead of @liskhq/bignum being used bignum helper was being used

### How did I fix it?

- Updated to use @liskhq/bignum package to be consistent with lisk-elements.

### How to test it?

- Build should be green

### Review checklist

* The PR resolves #3067
* Linting passes
* Tests pass


framework/src/controller/validator/formats.js
```diff
@@ -367,8 +366,8 @@ const validationFormats = {
 		type: 'number',
 		validate: value => {
 			const { TOTAL_AMOUNT } = global.constants;
-			if (new Bignum(value).isPositive()) {
-				return new Bignum(value).isLessThanOrEqualTo(TOTAL_AMOUNT);
+			if (new BigNum(value).isPositive()) {
+				return new BigNum(value).isLessThanOrEqualTo(TOTAL_AMOUNT);
```
lsilvs: Wouldn't be `lessThanOrEqualTo`?

```suggestion
				return new BigNum(value).lessThanOrEqualTo(TOTAL_AMOUNT);
```

pablitovicente: Good catch!

framework/src/modules/chain/rounds/round.js
```diff
@@ -303,31 +303,27 @@ class Round {
...
-		const feesRemaining = new Bignum(roundFees.toPrecision(15)).minus(
-			fees.multipliedBy(ACTIVE_DELEGATES)
+			.floor(BigNum.ROUND_FLOOR);
+		const feesRemaining = new BigNum(roundFees.toPrecision(15)).minus(
+			fees.times(ACTIVE_DELEGATES)
 		);
 		const rewards =
-			new Bignum(roundRewards[index].toPrecision(15)).integerValue(
-				Bignum.ROUND_FLOOR
+			new BigNum(roundRewards[index].toPrecision(15)).floor(
```
shuse2: I think `floor` doesn't take the input in the `@liskhq/bignum`

pablitovicente: As we checked the parameter was inert but removed for clarity of code.

framework/src/modules/chain/rounds/round.js
```diff
@@ -303,32 +303,26 @@ class Round {
...
-			fees.multipliedBy(ACTIVE_DELEGATES)
+			.floor(BigNum.ROUND_FLOOR);
+		const feesRemaining = new BigNum(roundFees.toPrecision(15)).minus(
+			fees.times(ACTIVE_DELEGATES)
 		);
 		const rewards =
-			new Bignum(roundRewards[index].toPrecision(15)).integerValue(
-				Bignum.ROUND_FLOOR
-			) || 0;
+			new BigNum(roundRewards[index].toPrecision(15)).floor() || 0;
```
lsilvs: Don't we need `BigNum.ROUND_FLOOR` here as in the other uses of `.floor`?

```suggestion
			new BigNum(roundRewards[index].toPrecision(15)).floor(BigNum.ROUND_FLOOR) || 0;
```

pablitovicente: In the library floor() does not take any arguments.

pablitovicente: https://github.com/LiskHQ/bignum/blob/804cff6af6f8f840663a361ef36f031979003209/bignum.js#L1018

pablitovicente: Hmm actually 0 decimal places will check more

pablitovicente: Yeah it's good as before it was returning `integerValue()` that as per docs `Returns a BigNumber whose value is the value of this BigNumber rounded to an integer using rounding mode` so `floor()` is correct here

framework/src/modules/chain/rounds/round.js
```diff
@@ -303,32 +303,26 @@ class Round {
...
+				.floor(BigNum.ROUND_FLOOR);
 		}
 
-		const fees = new Bignum(roundFees.toPrecision(15))
+		const fees = new BigNum(roundFees.toPrecision(15))
 			.dividedBy(ACTIVE_DELEGATES)
-			.integerValue(Bignum.ROUND_FLOOR);
-		const feesRemaining = new Bignum(roundFees.toPrecision(15)).minus(
-			fees.multipliedBy(ACTIVE_DELEGATES)
+			.floor(BigNum.ROUND_FLOOR);
```
shuse2: there are several places still uses input for `floor` function

pablitovicente: Found only in rounds in the code, will fix and push; did you found more in other files?

shuse2: yes there were several places 😉

pablitovicente: Changes pushed. Resolving

pablitovicente: @shuse2 check that many of the occurrences are `Math.floor()` which do take the number to be floored.

Issue #3346, opened 2018-11-22
Update peer selection (for connection) mechanism for selection of multiple peers


shuse2:
### Description

According to [LIP0004](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0004.md), Update peer selection mechanism based on `newPeers` and `triedPeers` list and consider fixedlist and blacklist peers while selecting any peer. 
Also, peer selection must take banned peers into account while selecting peers.

jondubois:
This line of the LIP is particularly relevant to this issue: `Select a peer from collection triedPeers with probability r and a peer from the collection newPeers with probability 1-r. The peer in each of the collections is selected uniformly at random.`

PR #3764, opened 2019-06-03
Introduce peer populator - Closes #3648

### LINKED ISSUE #3648: Introduce peer populator to replenish outbound connections
### Expected behavior

Since peers on the network should only be partially interconnected, a max of 20 outbound connections must be targeted. The populator takes charge of connecting to peers independently from the discovery process interval. A different interval should be introduced to allow that. The populator tries to connect the node to as many peers as possible until reaching the configured number of outbound connections. Therefore, a `maxOutboundConnections` parameter should be configurable, discouraging the users to modify it in any non-experimental scenario. The proposed default value is `20`. 

### Actual behavior

 Node tries to connect to all peers creating a full mesh. Connection is coupled to the discovery process being triggered sequentially after.

diego-G:
### How to test it?

`npm run test:integration`

### Review checklist

* The PR resolves #3648
* Linting passes
* Tests pass


elements/lisk-p2p/src/p2p.ts
```diff
@@ -115,9 +115,11 @@ export const EVENT_NEW_PEER = 'newPeer';
 export const NODE_HOST_IP = '0.0.0.0';
 export const DEFAULT_DISCOVERY_INTERVAL = 30000;
 export const DEFAULT_BAN_TIME = 86400;
+export const DEFAULT_POPULATOR_INTERVAL = 10000;
 export const DEFAULT_SEND_PEER_LIMIT = 25;
 
 const BASE_10_RADIX = 10;
+const MAX_OUTBOUND_CONNECTIONS = 20;
```
shuse2: Probably `DEFAULT_MAX_OUTBOUND_CONNECTIONS`?

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -373,18 +375,27 @@ export class PeerPool extends EventEmitter {
 		return discoveredPeers;
 	}
 
-	public selectPeersAndConnect(
+	public triggerNewConnections(
 		peers: ReadonlyArray<P2PDiscoveredPeerInfo>,
```
shuse2: In the LIP, connection selection is based on each list with some condition, but would that be in another PR?

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -439,6 +450,21 @@ export class PeerPool extends EventEmitter {
 		return peer;
 	}
 
+	public getPeersCountByKind(): P2PPeersCount {
+		const peersCount = { outbound: 0, inbound: 0 };
```
shuse2: I think what you want to use here is `reduce`

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -373,18 +375,27 @@ export class PeerPool extends EventEmitter {
...
-			return this.addOutboundPeer(peerId, peerInfo);
-		});
+			const peersToConnect = this._peerSelectForConnection({
+				peers: shuffledPeers,
+			});
+			peersToConnect.forEach((peerInfo: P2PDiscoveredPeerInfo) => {
+				const peerId = constructPeerIdFromPeerInfo(peerInfo);
 
-		return peersToConnect;
+				return this.addOutboundPeer(peerId, peerInfo);
```
shuse2: this return doesn't mean anything i guess

elements/lisk-p2p/src/peer_pool.ts
diego-G: We plan to implement that through this issue https://github.com/LiskHQ/lisk-sdk/issues/3346

elements/lisk-p2p/src/p2p.ts
```diff
@@ -115,9 +115,11 @@ export const EVENT_NEW_PEER = 'newPeer';
 export const NODE_HOST_IP = '0.0.0.0';
 export const DEFAULT_DISCOVERY_INTERVAL = 30000;
 export const DEFAULT_BAN_TIME = 86400;
+export const DEFAULT_POPULATOR_INTERVAL = 10000;
```
mitsujutsu: Is this interval specified in LIP somewhere?

diego-G: No, it's not. We can decide what to set by default.

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -373,18 +375,27 @@ export class PeerPool extends EventEmitter {
...
 		peers: ReadonlyArray<P2PDiscoveredPeerInfo>,
-	): ReadonlyArray<P2PDiscoveredPeerInfo> {
-		const peersToConnect = this._peerSelectForConnection({ peers });
-
-		peersToConnect.forEach((peerInfo: P2PDiscoveredPeerInfo) => {
-			const peerId = constructPeerIdFromPeerInfo(peerInfo);
+	): void {
+		const peersCount = this.getPeersCountByKind();
+		// Trigger new connections only if the maximum of outbound connections has not been reached
+		if (peersCount.outbound < this._maxOutboundConnections) {
```
jondubois: The decision for selecting peers is currently split between this code and the `this._peerSelectForConnection` function .

Ideally, the `_peerSelectForConnection` function should receive the full list of peers and return a subset of that list as output. You can pass new arguments to the `_peerSelectForConnection` function to give it more information to make the decision but the ideally the entire selection decision should happen inside `_peerSelectForConnection`.

This is because `_peerSelectForConnection` is customizable by the user of the library (e.g. our Network module) so we should give that function more control over the selection-for-connection logic.

Note that we may want to change the input arguments to the `_peerSelectForConnection` function in the future; maybe we will want to pass `_newPeers` and `_triedPeers` separately?

PR #3765, opened 2019-06-03
Update forger.js to use async/await - Closes #3748

### LINKED ISSUE #3748: Update forger.js to use async/await
### Expected behavior
forger.js should use async/await style completely

### Actual behavior
it is still using callback style

2.2-

pablitovicente:
### What was the problem?

- forger.js was still using callbacks

### How did I fix it?

- Removed callback based code and using async/await
- Updated test

### How to test it?

- Build should be green

### Review checklist

* The PR resolves #3748
* Linting passes
* Tests pass


framework/src/modules/chain/forger.js
```diff
@@ -283,89 +262,85 @@ class Forger {
...
+			this.logger.info(
+				`Forged new block id: ${forgedBlock.id} height: ${
+					forgedBlock.height
+				} round: ${this.slots.calcRound(
+					forgedBlock.height
+				)} slot: ${this.slots.getSlotNumber(forgedBlock.timestamp)} reward: ${
+					forgedBlock.reward
+				}`
+			);
+			return;
```
shuse2: probably no need to return here

framework/src/modules/chain/forger.js
```diff
@@ -183,89 +181,70 @@ class Forger {
...
-					publicKeyBytes,
-					privateKeyBytes,
-				} = getPrivateAndPublicKeyBytesFromPassphrase(passphrase);
-
-				const keypair = {
-					publicKey: publicKeyBytes,
-					privateKey: privateKeyBytes,
-				};
-
-				if (keypair.publicKey.toString('hex') !== encryptedItem.publicKey) {
```
lsilvs: This was removed maybe by mistake?

framework/test/mocha/unit/modules/chain/forger.js
```diff
@@ -479,18 +481,19 @@ describe('forge', () => {
 
 				forgeModule.config.forging.delegates = [accountDetails];
 
-				forgeModule.loadDelegates(err => {
+				try {
+					await forgeModule.loadDelegates();
+				} catch (err) {
```
lsilvs: This is giving false positive as it never catches and the expectation is never evaluated.

Try using the following instead:
```return expect(forgeModule.loadDelegates()).to.be.rejectedWith(`Invalid encryptedPassphrase for publicKey: ${accountDetails.publicKey}. Public keys do not match`);```

File: framework/src/modules/chain/forger.js [hunk 4]
```diff
@@ -183,89 +181,76 @@ class Forger {
...
+			const filters = {
+				address: getAddressFromPublicKey(keypair.publicKey.toString('hex')),
+			};
+
+			const options = {
+				extended: true,
+			};
+
+			// eslint-disable-next-line no-await-in-loop
+			const accounts = await this.storage.entities.Account.get(
```
lsilvs:
```suggestion
			const [account] = await this.storage.entities.Account.get(
```

framework/test/mocha/unit/modules/chain/forger.js
```diff
@@ -191,39 +191,30 @@ describe('forge', () => {
...
+			it('should not load any delegates when forging.delegates list is undefined', async () => {
 				forgeModule.config.forging.delegates = undefined;
 
-				forgeModule.loadDelegates(err => {
-					expect(err).to.not.exist;
-					expect(Object.keys(forgeModule.keypairs).length).to.equal(0);
-					done();
-				});
+				await forgeModule.loadDelegates();
+				expect(Object.keys(forgeModule.keypairs).length).to.equal(0);
```
lsilvs:
```suggestion
				return expect(Object.keys(forgeModule.keypairs).length).to.equal(0);
```

framework/test/mocha/unit/modules/chain/forger.js
```diff
@@ -479,18 +592,31 @@ describe('forge', () => {
...
+					publicKey:
+						'141b16ac8d5bd150f16b1caa08f689057ca4c4434445e56661831f4e671b7c0a',
+					encryptedPassphrase:
+						'iterations=1&salt=8c79d754416acccb567a42cf62b2e3bb&cipherText=73f5827fcd8eeab475abff71476cbce3b1ecacdeac55a738bb2f0a676d8e543bb92c91e1c1e3ddb6cef07a503f034dc7718e39657218d5a955859c5524be06de5954a5875b4c7b1cd11835e3477f1d04&iv=aac6a3b77c0594552bd9c932&tag=86231fb20e7b263264ca68b3585967ca&version=1',
+				};
+
+				forgeModule.config.forging.delegates = [accountDetails];
+
+				try {
+					await forgeModule.loadDelegates();
```
lsilvs: If this call doesn't throw anything this test will give a false positive.
Would be nice to add a falsy expectation after this line to make sure the test never reaches it.

framework/src/modules/chain/forger.js
pablitovicente: Committed and pushed locally as I wanted to run tests.

framework/test/mocha/unit/modules/chain/forger.js
pablitovicente: Added falsy expectation on all tests using try/catch and testing for failures.

PR #3779, opened 2019-06-04
Implement maximum number of inbound peer connections - Closes #3720

### LINKED ISSUE #3720: Implement incoming connections limit
### Expected behavior

When an incoming peer connection arises, the peer pool should check if the current incoming connections are below the maximum of incoming connections allowed (`maxInConnections`). Otherwise, an eviction method must be triggered to randomly remove an ongoing connection and therefore allow the fresh one.

The parameter `maxInConnections` must be configurable, discouraging the users to modify it in any non-experimental scenario. The proposed default value is `100`.

### Actual behavior

The node allows as much connections as the hardware supports.

`2.0.0+`

diego-G:
### How to test it?

`npm run test:integration`

### Review checklist

* The PR resolves #3720
* Linting passes
* Tests pass

### DIFF
--- a/elements/lisk-p2p/src/peer_pool.ts
+++ b/elements/lisk-p2p/src/peer_pool.ts
@@ -235,7 +235,7 @@ export class PeerPool extends EventEmitter {
 
 	public applyNodeInfo(nodeInfo: P2PNodeInfo): void {
 		this._nodeInfo = nodeInfo;
-		const peerList = this.getAllPeers();
+		const peerList = this.getPeers();
 		peerList.forEach(peer => {
 			this._applyNodeInfoOnPeer(peer, nodeInfo);
 		});
@@ -380,13 +380,13 @@ export class PeerPool extends EventEmitter {
 	public triggerNewConnections(
 		peers: ReadonlyArray<P2PDiscoveredPeerInfo>,
 	): void {
-		const disconnectPeers = peers.filter(
+		const disconnectedPeers = peers.filter(
 			peer => !this._peerMap.has(constructPeerIdFromPeerInfo(peer)),
 		);
-		const { outbound } = this.getPeersCountByKind();
+		const { outbound } = this.getPeersCountPerKind();
 		// Trigger new connections only if the maximum of outbound connections has not been reached
 		const peersToConnect = this._peerSelectForConnection({
-			peers: disconnectPeers,
+			peers: disconnectedPeers,
 			peerLimit: this._maxOutboundConnections - outbound,
 		});
 		peersToConnect.forEach((peerInfo: P2PDiscoveredPeerInfo) => {
@@ -400,7 +400,7 @@ export class PeerPool extends EventEmitter {
 		peerInfo: P2PDiscoveredPeerInfo,
 		socket: SCServerSocket,
 	): Peer {
-		const inboundPeers = this.getAllPeers(InboundPeer);
+		const inboundPeers = this.getPeers(InboundPeer);
 		if (inboundPeers.length >= this._maxInboundConnections) {
 			this.removePeer(shuffle(inboundPeers)[0].id);
 		}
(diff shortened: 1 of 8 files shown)


elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -473,8 +480,13 @@ export class PeerPool extends EventEmitter {
 		return this.getAllPeers().map(peer => peer.peerInfo);
 	}
 
-	public getAllPeers(): ReadonlyArray<Peer> {
-		return [...this._peerMap.values()];
+	public getAllPeers(kind?: any): ReadonlyArray<Peer> {
```
mitsujutsu: maybe specify `InboundPeer` or `OutboundPeer` instead of `any`

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -473,8 +480,15 @@ export class PeerPool extends EventEmitter {
 		return this.getAllPeers().map(peer => peer.peerInfo);
 	}
 
-	public getAllPeers(): ReadonlyArray<Peer> {
-		return [...this._peerMap.values()];
+	public getAllPeers(
+		kind?: typeof OutboundPeer | typeof InboundPeer,
```
mitsujutsu: maybe can use `unknown` here if the idea is to allow other types of Peers in the future.

elements/lisk-p2p/test/integration/p2p.ts
```diff
@@ -1319,7 +1314,7 @@ describe('Integration tests for P2P library', () => {
 
 	describe('Network with a maximum number of outbound/inbound connections', () => {
 		const NETWORK_PEER_COUNT_WITH_LIMIT = 30;
-		const TEN_CONNECTIONS = 10;
+		const TEN_CONNECTIONS = 5;
```
mitsujutsu: Update variable name here as well

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -473,8 +483,15 @@ export class PeerPool extends EventEmitter {
...
```
jondubois: This should be a string enum 'outbound' or 'inbound' for consistency. Also this is confusing because in JavaScript `typeof OutboundPeer` would evaluate to 'function'.

shuse2: What was the conclusion about this one? Although in tyepescript, this seems to be valid

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -378,11 +380,14 @@ export class PeerPool extends EventEmitter {
 	public triggerNewConnections(
 		peers: ReadonlyArray<P2PDiscoveredPeerInfo>,
 	): void {
-		const peersCount = this.getPeersCountByKind();
+		const disconnectPeers = peers.filter(
+			peer => !this._peerMap.has(constructPeerIdFromPeerInfo(peer)),
+		);
+		const { outbound } = this.getPeersCountByKind();
```
shuse2: not related to this PR, but `getPeersCountByKind ` seems to be bad function name.
It seems like it takes the `kind` as input, but it does not

elements/lisk-p2p/src/peer_pool.ts
shuse2: function name should be adjusted here too because when we supply `kind` it's not `getAllPeers` anymore

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -378,11 +380,14 @@ export class PeerPool extends EventEmitter {
 	public triggerNewConnections(
 		peers: ReadonlyArray<P2PDiscoveredPeerInfo>,
 	): void {
-		const peersCount = this.getPeersCountByKind();
+		const disconnectPeers = peers.filter(
```
shuse2: does this mean this is peers that is not connected?
disconnectedPeers might be better choice
'

elements/lisk-p2p/src/peer_pool.ts
diego-G: We have no better option and it is valid Typescript-wise.

jondubois: Because this method is only used internally by the PeerPool and not exposed publicly, I was convinced that it's OK in this case.

PR #3780, opened 2019-06-04
Remove callback from loader js and use async - Closes #3747

### LINKED ISSUE #3747: Update loader.js to async/await
### Expected behavior
loader.js should use async/await style completely

### Actual behavior
it is still using callback style

2.2-

ishantiw:
### What was the problem?

loader.js should use async/await style completely

### How did I fix it?

Removed callbacks and use async await

### How to test it?

N/A

### Review checklist

* The PR resolves #3747
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/chain/loader.js
+++ b/framework/src/modules/chain/loader.js
@@ -332,106 +332,118 @@ __private.getTransactionsFromNetwork = async () => {
  * @returns {Promise} void
  * @todo Add description for the params
  */
+__private.getBlocksFromNetwork = async () => {
+	const lastBlock = modules.blocks.lastBlock;
+	// TODO: If there is an error, invoke the applyPenalty action on the Network module once it is implemented.
+	// TODO: Rename procedure to include target module name. E.g. chain:blocks
+	let data;
+	try {
+		const response = await library.channel.invoke('network:request', {
+			procedure: 'blocks',
+			data: {
+				lastBlockId: lastBlock.id,
+			},
+		});
+		data = response.data;
+	} catch (p2pError) {
+		library.logger.error('Failed to load block from network', p2pError);
+		return [];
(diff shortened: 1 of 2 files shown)


framework/src/modules/chain/loader.js
```diff
@@ -159,43 +158,35 @@ class Loader {
...
-				__private.blocksToSync = 0;
+		const consensusBefore = await modules.peers.calculateConsensus();
 
-				library.logger.info('Finished sync');
-				if (components.cache.ready) {
-					components.cache.enable();
-				}
-				return setImmediate(cb, err);
-			}
+		library.logger.debug(
```
pablitovicente: shouldn't this be before the call to `calculateConsensus()` ? Small difference but upon potential errors might give more clues about the failure?

ishantiw: We want to calculate consensus just before loading blocks from the network and just after it. So that we can see the difference before and after the sync.

framework/src/modules/chain/loader.js
```diff
@@ -338,118 +329,116 @@ __private.getTransactionsFromNetwork = async function() {
...
  */
-__private.loadBlocksFromNetwork = function(cb) {
+__private.loadBlocksFromNetwork = async () => {
 	// Number of failed attempts to load from the network.
 	let failedAttemptsToLoad = 0;
 	// If True, own node's db contains all the blocks from the last block request.
 	let loaded = false;
+	while (!loaded && failedAttemptsToLoad < 5) {
+		// eslint-disable-next-line no-loop-func
+		const loadingSteps = async () => {
```
shuse2: maybe better to remove this internal function definition and just call step by step is enough
It looks more complicated than it should be

PR #3782, opened 2019-06-04
Generate delegate list using `round_delegates` table - Closes #3631

### LINKED ISSUE #3631: Generate delegate list using `round_delegates` table
### Expected behavior
We need to have `round_delegates` table to store delegate list that generated at the beginning of the round.
### Acceptance Criteria
- [x] Create `round_delegates` table.

Table Schema: 
```
roundNo - integer - primary key
Delegates - json
```

- [x] Get active delegates from `mem_accounts` table sorted by vote desc.
- [x] Store **unshuffled** delegateList in `delegate_rounds` table. RoundNo must be the round that list was generated. List will be stored in the JSON field.
- [x] **Unshuffled** delegateList should be stored in the memory
- [x] Shuffle delegateList with the `activeRound` number.
v3.0.0

yatki:
### What was the problem?
We were using mem_rounds table for generating delegate list.
### How did I fix it?
Now we have `round_delegates` table which requires less db calls.
### Review checklist

* The PR resolves #3631 
* Linting passes
* Tests pass


framework/src/modules/chain/submodules/dpos/delegates.js
```diff
@@ -0,0 +1,75 @@
+const { hash } = require('@liskhq/lisk-cryptography');
+
+global.constants = global.constants || {};
```
shuse2: it should not use `global`

framework/src/modules/chain/submodules/dpos/delegates.js
```diff
@@ -0,0 +1,75 @@
+const { hash } = require('@liskhq/lisk-cryptography');
```
shuse2: missing license header, and file path should be updated

framework/src/modules/chain/submodules/dpos/delegates.js
```diff
@@ -0,0 +1,75 @@
...
+
+global.constants = global.constants || {};
+global.exceptions = global.exceptions || {};
+
+const { ACTIVE_DELEGATES = 101 } = global.constants;
+const { ignoreDelegateListCacheForRounds = [] } = global.exceptions;
+
+const shuffleActiveDelegateList = async (round, list) => {
```
shuse2: this function is not `async`

framework/src/modules/chain/submodules/dpos/index.js
```diff
@@ -0,0 +1,15 @@
+const { Delegates } = require('./delegates');
+
+class Dpos {
```
shuse2: index should only import/export

framework/src/modules/chain/components/storage/entities/dpos.js
```diff
@@ -0,0 +1,140 @@
...
+ * @property {number} [round_ne]
+ * @property {number} [round_gt]
+ * @property {number} [round_gte]
+ * @property {number} [round_lt]
+ * @property {number} [round_lte]
+ * @property {number} [round_in]
+ * @property {json} [delegatePublicKeys]
+ */
+
+class Dpos extends BaseEntity {
```
shuse2: name is not RoundDelegates?

framework/src/modules/chain/dpos/dpos.js
```diff
@@ -0,0 +1,15 @@
+const Delegates = require('./delegates');
+
+const __methods = {
```
shuse2: prefer not to use weakmap. in Node 12 there will be private method, but we shouldnt use it now

framework/src/modules/chain/dpos/dpos.js
```diff
@@ -0,0 +1,15 @@
...
+	delegates: new WeakMap(),
+};
+
+module.exports = class Dpos {
+	constructor({ storage, logger }) {
+		__methods.delegates.set(this, new Delegates({ storage, logger }));
```
shuse2: this.delegates = new Delegates({ storage, logger });

framework/src/modules/chain/components/storage/entities/round_delegates.js
```diff
@@ -0,0 +1,140 @@
...
+		this.addField('round', 'number', { filter: NUMBER });
+
+		this.extendDefaultOptions({ sort: '' });
+
+		this.sqlDirectory = path.join(path.dirname(__filename), '../sql');
+
+		this.SQLs = this.loadSQLFiles('dpos', sqlFiles, this.sqlDirectory);
+	}
+
+	async get(round) {
```
SargeKhan: Is this interface consistent with other entities? I think `get` function returns multiple rows, whereas getOne returns a single element. Maybe we should stick with it?
Also, @nazarhussain, should this be in `chain` component or framework level componenet?

framework/src/modules/chain/components/storage/entities/round_delegates.js
```diff
@@ -0,0 +1,140 @@
...
+	 * Create round object
+	 *
+	 * @param {Object} data
+	 * @param {Object} [_options]
+	 * @param {Object} [tx] - Transaction object
+	 * @return {null}
+	 */
+	// eslint-disable-next-line no-unused-vars
+	create({ round, delegatePublicKeys }, _options = {}, tx = null) {
+		assert(round, 'Round must be an object');
```
SargeKhan: Should the round be a number or an object?
Also, maybe the expression passed to assert should check if it's a number

framework/src/modules/chain/components/storage/entities/round_delegates.js
```diff
@@ -0,0 +1,140 @@
...
+		assert(
+			Array.isArray(delegatePublicKeys),
+			'delegatePublicKeys must be an array of strings'
+		);
+
+		const attributes = Object.keys(this.fields).filter(
+			fieldname => fieldname !== 'id'
```
SargeKhan: We don't have an id field in the database table corresponding to this entity. So this check is unnecessary

framework/src/modules/chain/components/storage/entities/round_delegates.js
```diff
@@ -0,0 +1,140 @@
...
+		this.sqlDirectory = path.join(path.dirname(__filename), '../sql');
+
+		this.SQLs = this.loadSQLFiles('dpos', sqlFiles, this.sqlDirectory);
+	}
+
+	async get(round) {
+		const result = await this.adapter.executeFile(this.SQLs.getRoundDelegates, {
+			round,
+		});
+		return result[0].delegatePublicKeys;
```
SargeKhan: Is the response returned here in the JSON format or comma-separated values?

framework/src/modules/chain/dpos/delegates.js
```diff
@@ -0,0 +1,89 @@
...
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+const { hash } = require('@liskhq/lisk-cryptography');
+
+const shuffleActiveDelegateList = (round, list) => {
```
SargeKhan: Maybe `shuffleDelegateListForRound` makes more sense since it's not necessary that this function is only called for current active round.

framework/src/modules/chain/dpos/delegates.js
```diff
@@ -0,0 +1,89 @@
...
+		if (this.delegateListCache[round]) {
+			return this.delegateListCache[round];
+		}
+
+		let delegatePublicKeys = await this.storage.entities.RoundDelegates.get(
+			round
+		);
+
+		if (!delegatePublicKeys.length) {
+			delegatePublicKeys = await this.getDelegatePublicKeysSortedByVote();
```
SargeKhan: Keeping in mind that behavior of this function will change based on the protocol version, which will mean that `RoundDelegates` table will insert entries for publicKeys sorted by vote with the current round number in the earlier protocol version and current round number -2 in the next protocol version, do you think maybe it makes more sense to separate creating entry in RoundDelegates in a different function which will be called independently?

framework/src/modules/chain/components/storage/entities/round_delegates.js
yatki: it is an JSON array.

framework/src/modules/chain/dpos/delegates.js
yatki: I think when we need to get the delegate list from the round -2, we can just pass it to the `generateActiveDelegateList` function. `getDelegatePublicKeysSortedByVote` doesn't need to know which round we are in. It's just for returning the top voted 101 delegates. Also `return shuffleDelegateListForRound(round, list)` function makes sure that list is shuffled for that round.

So I think this is fine.

framework/src/modules/chain/components/storage/entities/round_delegates.js
yatki: It's a number. I fixed it.

framework/src/modules/chain/components/storage/entities/round_delegates.js
yatki: @SargeKhan Actually I'd prefer to name this function `getRoundDelegates` because it doesn't return the row but just delegate pks. And So far we don't need a get or getOne method in the entity. I'd prefer to add them when we actually need them. What do you think?

SargeKhan: Yea, getRoundDelegates is fine too

Issue #3278, opened 2019-04-06
Update Ajv, ZSchema usages to use @liskhq/lisk-validator


shuse2:
### Expected behavior
It should use one consistent library, @liskhq/lisk-validator

### Actual behavior
Currently using 2 libraries Ajv and ZSchema to do the same thing

PR #3783, opened 2019-06-05
Update transport.js to use async/await - Closes #3746

### LINKED ISSUE #3746: Update transport.js to use async/await
### Expected behavior
transport.js should use async/await

### Actual behavior
It is till using the callback style

2.2-

lsilvs:
### What was the problem?
`transport.js` still using callbacks

### How did I fix it?
I've replaced the callbacks functions by making them async

### How to test it?
Run test suite

### Review checklist

* The PR resolves #3746
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/chain/transport.js
+++ b/framework/src/modules/chain/transport.js
@@ -319,34 +319,36 @@ class Transport {
 					};
 				}
 
-				return modules.blocks
-					.loadBlocksDataWS({
+				try {
+					const data = await modules.blocks.loadBlocksDataWS({
 						limit: 34, // 1977100 bytes
 						lastId: query.lastBlockId,
-					})
-					.then(data => {
-						_.each(data, block => {
-							if (block.tf_data) {
-								try {
-									block.tf_data = block.tf_data.toString('utf8');
-								} catch (e) {
-									library.logger.error(
-										'Transport->blocks: Failed to convert data field to UTF-8',
-										{
-											block,
-											error: e,
-										}
-									);
-								}
+					});
+
+					_.each(data, block => {
+						if (block.tf_data) {
+							try {
+								block.tf_data = block.tf_data.toString('utf8');
+							} catch (e) {
+								library.logger.error(
+									'Transport->blocks: Failed to convert data field to UTF-8',
+									{
+										block,
+										error: e,
+									}
+								);
 							}
-						});
-						return { blocks: data, success: true };
-					})
-					.catch(err => ({
+						}
+					});
+
+					return { blocks: data, success: true };
+				} catch (err) {
+					return {
 						blocks: [],
 						message: err,
 						success: false,
-					}));
+					};
+				}
 			},
 
 			/**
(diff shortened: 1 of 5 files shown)


framework/src/modules/chain/transport.js
```diff
@@ -423,19 +423,30 @@ class Transport {
...
+					const err = library.schema.getLastErrors();
+					const error = new TransactionError(err[0].message);
+					return {
+						success: false,
+						code: 400,
+						errors: [error],
+					};
+				}
+
+				return modules.transactionPool
```
pablitovicente: why no await here?

framework/src/modules/chain/transport.js
```diff
@@ -231,74 +229,75 @@ class Transport {
 			 * @todo Add @returns tag
 			 * @todo Add description of the function
 			 */
-			blocksCommon(query, cb) {
+			async blocksCommon(query) {
 				query = query || {};
-				return library.schema.validate(
+
+				const valid = library.schema.validate(
```
shuse2: isn't this async?

framework/src/modules/chain/transport.js
```diff
@@ -308,19 +307,19 @@ class Transport {
...
+					return {
 						success: false,
 						message: 'Invalid lastBlockId requested',
-					});
+					};
 				}
 
-				modules.blocks
+				return modules.blocks
 					.loadBlocksDataWS({
```
shuse2: this is also already Promise, so prefer to use async await

framework/src/modules/chain/transport.js
```diff
@@ -359,61 +356,64 @@ class Transport {
...
+			async postBlock(query) {
 				if (!library.config.broadcasts.active) {
 					return library.logger.debug(
 						'Receiving blocks disabled by user through config.json'
 					);
 				}
 				query = query || {};
-				return library.schema.validate(
+
+				const valid = library.schema.validate(
```
shuse2: same as above. Maybe we can just import controller validator for now?

framework/src/modules/chain/transport.js
```diff
@@ -445,22 +456,25 @@ class Transport {
...
 					query,
-					definitions.WSSignaturesList,
-					err => {
-						if (err) {
-							return library.logger.debug('Invalid signatures body', err);
-						}
-						return __private.receiveSignatures(query.signatures);
-					}
+					definitions.WSSignaturesList
```
shuse2: same as above

framework/src/modules/chain/transport.js
```diff
@@ -549,25 +555,25 @@ class Transport {
...
-			postTransactions(query) {
+			async postTransactions(query) {
 				if (!library.config.broadcasts.active) {
 					return library.logger.debug(
 						'Receiving transactions disabled by user through config.json'
 					);
 				}
-				return library.schema.validate(
+
+				const valid = library.schema.validate(
```
shuse2: same as above

framework/src/modules/chain/transport.js
lsilvs: Because this function doesn't throw anything.
Instead, it always return an object.

framework/src/modules/chain/transport.js
lsilvs: No. It's the sync mode of `z-schema`

REF.: https://github.com/zaggino/z-schema#sync-mode

framework/src/modules/chain/transport.js
lsilvs: This will be done as part of https://github.com/LiskHQ/lisk-sdk/issues/3278

framework/src/modules/chain/transport.js
lsilvs: same as above

framework/src/modules/chain/transport.js
lsilvs: same as above

PR #3784, opened 2019-06-05
Convert BlockReward Class functions to pure functions - Closes #2412

### LINKED ISSUE #2412: Convert BlockReward class methods to static methods
### Expected behavior
[block_reward.js#L35](https://github.com/LiskHQ/lisk/blob/development/logic/block_reward.js#L35) Here we don't need to have class properties. We can directly use constants in the methods.

But removing properties causes eslint exceptions (class-method-use-this).

We need to define all methods as static classes and refactor its usages.
### Actual behavior

We are using unnecessary class properties.

1.3.0+

ishantiw:
### What was the problem?

BlockReward class is not holding and mostly using pure functions within it. 

### How did I fix it?

Extracted all the functions from the BlockReward class and export them.

### How to test it?

`npm run mocha:unit test/mocha/unit/modules/chain/blocks/block_reward.js`

`npm run mocha:unit`

### Review checklist

* The PR resolves #2412
* Linting passes
* Tests pass


framework/src/modules/chain/blocks/block_reward.js
```diff
@@ -38,117 +38,101 @@ const parseHeight = height => {
...
+		const reward = rewards[i];
+		supply = supply.plus(new BigNum(reward[0]).times(reward[1]));
 	}
-}
 
-module.exports = { BlockReward };
+	return supply;
+};
+
+module.exports = { calcMilestone, calcSupply, calcReward };
```
lsilvs:
```suggestion
module.exports = { calculateMilestone, calculateSupply, calculateReward };
```

ishantiw: Yeah the names were not nice, I changed the names as per your suggestion

ishantiw:
@pablitovicente I feel what you are saying, when I changed it, I felt somewhere that we lose some clarity of what function is all about. However, in most of the cases, we are actually using it as a member of Blocks class and using it as `Blocks.blockRewards.calculateMilestone()`. However, we can still use them directly but without destructured way. `const blockRewards = require('./block_rewards')` and then use functions like `blockRewards.calculateRewards()` that will make sure that you have some context about the function.

PR #3790, opened 2019-06-06
Fix peerInfo update events across p2p lib - Closes #3624

### LINKED ISSUE #3624: Unprocessed transactions were not added to empty or non full blocks
### Expected behavior
Blocks should be full when we have enough `unprocessed` transactions in the queue.
### Actual behavior
Node doesn't process transactions even though there are enough `unprocessed` transactions in the queue
This block contains only 4 transactions even though during that point of time there were many `unprocessed` transactions http://alpha4-seed-01.liskdev.net:6040/block/13226538873268144676

Same here, the below link shows empty blocks even though there were 100's of `unprocessed` transactions
http://alpha4-seed-01.liskdev.net:6040/block/558195309841882412
http://alpha4-seed-01.liskdev.net:6040/block/15706719315041840789
http://alpha4-seed-01.liskdev.net:6040/block/17560937340718727312
### Steps to reproduce

- Run stress test and you will be able to observe the issue (deploy at least 50 nodes)

2.0.0

ishantiw:
### What was the problem?

Peer's updatePeerInfo is called in multiple places in PeerPool that is creating inconsistency across the peer lists in p2p-lib for a particular peer. Also, updating peerInfo from discovery. This leads to low broadhash consensus due to stale info about the connected peers.

### How did I fix it?

PeerInfo can only be updated by the peer itself using `fetchStatus` and `updateMyself` RPCs and not by the peerlist of other peers.

### How to test it?

`npm t` in `lisk-p2p`

### Review checklist

* The PR resolves #3624 
* Linting passes
* Tests pass


elements/lisk-p2p/src/peer.ts
```diff
@@ -495,14 +497,29 @@ export class Peer extends EventEmitter {
 		return outboundSocket;
 	}
 
+	private async updatePeerOnConnect(): Promise<void> {
```
jondubois: Private method should start with an underscore.

PR #3793, opened 2019-06-07
Update Ajv, ZSchema usages to use @liskhq/lisk-validator - Closes #3278

### LINKED ISSUE #3278: Update Ajv, ZSchema usages to use @liskhq/lisk-validator

pablitovicente:
### What was the problem?

- Validator library exists in framework instead of using @liskhq/lisk-validator

### How did I fix it?

- By replacing usages of validator library with @liskhq/lisk-validator
- Note that `framework/src/controller/validator/index.js` was not removed as it has some cases that's used for setting defaults and not related directly to validation of application input but configuration. Changes for this should be discussed and worked in a separated issue.

### How to test it?

- Build should be green

### Review checklist

* The PR resolves #3278
* Linting passes
* Tests pass


elements/lisk-validator/src/validation.ts
```diff
@@ -242,3 +242,18 @@ export const validateFee = (data: string): boolean =>
...
+		return false;
+	}
+
+	const csvAsArray = data.split(',');
+
+	if (csvAsArray.length > 0 && csvAsArray.length <= maxItemCount) {
+		return true;
+	}
+
+	return false;
```
lsilvs: `isCsv` will return `false` for any CSV with more than `maxItemCount`.
Maybe use a better name for this function which express that restriction?

elements/lisk-validator/src/validation.ts
```diff
@@ -242,3 +242,18 @@ export const validateFee = (data: string): boolean =>
 	isNumberString(data) &&
 	isGreaterThanZero(new BigNum(data)) &&
 	!isGreaterThanMaxTransactionAmount(new BigNum(data));
+
+export const isCsv = (data: string): boolean => {
```
lsilvs: Can we test this function in `elements/lisk-validator/test/validation.ts`?

framework/src/modules/chain/loader.js
```diff
@@ -195,8 +194,14 @@ class Loader {
...
 		});
 
-		const validate = promisify(this.schema.validate.bind(this.schema));
-		await validate(result, definitions.WSSignaturesResponse);
+		const validate = validator.validate(
+			definitions.WSSignaturesResponse,
+			result
+		);
+
+		if (!validate) {
```
lsilvs: In some places we're using `if (!validate)` and in other places `if (validator.errors)`.
Prefer using it consistently.

pablitovicente: Tried to maintain the changes low due to four PRs changing the same code making merges very complex and prone to errors. Will wait for other PRs to merge before making this change.

elements/lisk-validator/src/validation.ts
pablitovicente: I would prefer not to change this behaviour now as the implementation comes from framework and is used in the schemas there so I need to research first why it was like this in the first place. Maybe work it in a separate issue?

elements/lisk-validator/src/validation.ts
pablitovicente: Added test for `isCsv` resolving.

framework/src/modules/chain/loader.js
pablitovicente: Found this case only here so I went ahead and replaced. Let me know if you spotted more cases I couldn't find any more. Resolving.

framework/src/modules/chain/loader.js
```diff
@@ -195,8 +194,14 @@ class Loader {
...
 
-		const validate = promisify(this.schema.validate.bind(this.schema));
-		await validate(result, definitions.WSSignaturesResponse);
+		const validate = validator.validate(
+			definitions.WSSignaturesResponse,
+			result
+		);
+
+		if (validator.errors) {
+			throw validate.errors;
```
lsilvs:
```suggestion
			throw validator.errors;
```

elements/lisk-validator/src/formatter.ts
```diff
@@ -121,3 +129,7 @@ validator.addKeyword('uniqueSignedPublicKeys', {
 	compile: () => (data: ReadonlyArray<string>) =>
 		new Set(data.map((key: string) => key.slice(1))).size === data.length,
 });
```
nazarhussain: The keyword `validator.addKeyword('uniqueSignedPublicKeys'` does not have a meta-schema. So completely unaware what it actually is what it accepts. Look at https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/controller/validator/keywords/arg/meta_schema.js for a sample. Also I don't think its a use case of a keyword, rather its just a validation format which requires uniqueness on keys.

elements/lisk-validator/src/validation.ts
```diff
@@ -242,3 +242,18 @@ export const validateFee = (data: string): boolean =>
...
+	const maxItemCount = 1000;
```
nazarhussain: I don't thinks its a good idea to mention max item count for a generic formatter like `isCSV`. This should and only check if data is a csv or not. If you want to limit CSV items for particular use case then create specific format for it. e.g. `isValidListOfKeys`. Or this is right use case to introduce a `keyword` e.g. `CSVMaxItem`. 

One can also specify a regular expression as `pattern` directly on the JSON-schema to limit the items in CSV string.

elements/lisk-validator/src/formatter.ts
```diff
@@ -17,7 +17,9 @@ import * as Ajv from 'ajv';
 // tslint:disable-next-line no-require-imports
 import addKeywords = require('ajv-merge-patch');
```
nazarhussain: This is very complicated stuff, dynamically modifying the JSON-schema runtime and then validate data against that particular in-memory schema. We should avoid using it, unless we have a particular use-case which can't be handled otherwise. Instead we should always recommend to use a statically defined JSON-Schema.

framework/src/controller/application.js
```diff
@@ -114,7 +115,10 @@ class Application {
 	 * @throws Framework.errors.SchemaValidationError
 	 */
 	constructor(genesisBlock, config = {}) {
-		validator.validate(genesisBlockSchema, genesisBlock);
+		liskValidator.validate(genesisBlockSchema, genesisBlock);
+		if (liskValidator.errors) {
+			throw liskValidator.errors;
+		}
```
nazarhussain: Validation must always throw error in case of failure, its so much redundant to use `if` everywhere. If one don't want to throw error, then he can use `try/catch` for his code block.

framework/src/modules/chain/blocks/block.js
```diff
@@ -134,10 +134,11 @@ const objectNormalize = (block, exceptions = {}) => {
...
 	});
-	try {
-		validator.validate(blockSchema, block);
-	} catch (schemaError) {
-		throw schemaError.errors;
+
+	validator.validate(blockSchema, block);
+
+	if (validator.errors) {
+		throw validator.errors;
```
nazarhussain: A lot of redundant if conditions introduced. Validator must throw error in case of failure.

elements/lisk-validator/src/formatter.ts
```diff
@@ -29,7 +31,13 @@ import {
 	validateTransferAmount,
 } from './validation';
 
-export const validator = new Ajv({ allErrors: true, removeAdditional: 'all' });
+export const validator = new Ajv({
+	allErrors: true,
+	removeAdditional: 'all',
```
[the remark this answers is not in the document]
pablitovicente: @shuse2 I think you implemented this configuration originally could you please answer to @nazarhussain ?

elements/lisk-validator/src/formatter.ts
```diff
@@ -29,7 +31,13 @@ import {
...
+	schemaId: 'auto',
```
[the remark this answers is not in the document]
pablitovicente: @nazarhussain it's using the same configuration for this as it was implemented by you in https://github.com/LiskHQ/lisk-sdk/blob/86b78e02b9313150c761dbfccb37ba249e759ec6/framework/src/controller/validator/index.js#L31 why you think now it should be changed but before it was ok?

elements/lisk-validator/src/formatter.ts
```diff
@@ -121,3 +129,7 @@ validator.addKeyword('uniqueSignedPublicKeys', {
 	compile: () => (data: ReadonlyArray<string>) =>
 		new Set(data.map((key: string) => key.slice(1))).size === data.length,
 });
+
+validator.addFormat('hex', isHexString);
+
+validator.addFormat('csv', isCsv);
```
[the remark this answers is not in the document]
pablitovicente: I think this is outside the scope of this PR for which only methods that were missing were added to this library. Please open an issue with suggestions for lisk-validator

elements/lisk-validator/src/formatter.ts
pablitovicente: I think this is outside the scope of this PR for which only methods that were missing were added to this library. Please open an issue with suggestions for lisk-validator

elements/lisk-validator/src/validation.ts
pablitovicente: This honours the original implementation in framework. I think this is outside the scope of this PR for which only methods that were missing were added to this library. Please open an issue with suggestions for lisk-validator

framework/src/controller/application.js
pablitovicente: I think this is outside the scope of this PR for which only methods that were missing were added to this library. Please open an issue with suggestions for lisk-validator

framework/src/modules/chain/blocks/block.js
pablitovicente: I think this is outside the scope of this PR for which only methods that were missing were added to this library. Please open an issue with suggestions for lisk-validator

elements/lisk-validator/src/formatter.ts
pablitovicente: I think this is outside the scope of this PR for which only methods that were missing were added to this library. Please open an issue with suggestions for lisk-validator

elements/lisk-validator/src/formatter.ts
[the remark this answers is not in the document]
shuse2: Originally we implemented this for transaction validation, and we didn't want the additional field if exists.
I think in general we shouldn't allow additional field when validation unless we have specific usecase to allow them

elements/lisk-validator/test/validation.ts
```diff
@@ -599,4 +600,21 @@ describe('validation', () => {
 			});
 		});
 	});
+
+	describe.only('isCsv', () => {
```
ishantiw: I think you forgot to remove `only`, all the other tests are skipped.

pablitovicente: Removed

elements/lisk-validator/src/formatter.ts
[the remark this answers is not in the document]
nazarhussain: Yeah I know, and that's that I explained `based on earlier experience`. When we make it `auto`, someone use `id` in schema, someone used `$id`. I have to make it consistent in all schema files. So in new implementation we should not use `auto`.

elements/lisk-validator/src/formatter.ts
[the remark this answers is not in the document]
nazarhussain: @pablitovicente  For the feedback which you agree and think its out of scope of the PR, please create issue for it and resolve feedback with reference of the issue.

elements/lisk-validator/src/validation.ts
nazarhussain: The above code is introduced in the current PR, so its the right scope to do it right. This validator `isCsv` is generic one, it must not have such limits in it as explained above.

elements/lisk-validator/src/formatter.ts
nazarhussain: If you agree to feedback and think its out of scope, then consolidate all such feedback in an issue and resolve them with issue reference.

framework/src/controller/application.js
nazarhussain: This feedback is related to the code introduced in this PR, so not out of scope. If you are replacing the previous implementation with new then this is important instead of adding a lot of conditions in this PR.

elements/lisk-validator/src/formatter.ts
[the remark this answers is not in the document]
nazarhussain: > I think in general we shouldn't allow additional field

The question is what would happen if there are additional fields. With this configuration you silently removing additional attributes in the source data and not throwing error.

elements/lisk-validator/src/formatter.ts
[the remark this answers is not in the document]
pablitovicente: https://github.com/LiskHQ/lisk-sdk/issues/3823

elements/lisk-validator/src/validation.ts
pablitovicente: Code was originally in framework I just moved it but made the changes. Check is now done in validation.ts and also in the blocksCommon function.

elements/lisk-validator/src/formatter.ts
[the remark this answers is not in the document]
pablitovicente: Removed as requested.

elements/lisk-validator/src/formatter.ts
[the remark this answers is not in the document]
pablitovicente: #3825

elements/lisk-validator/src/formatter.ts
pablitovicente: Removed. Some tests of validator were using but I changed them to `allOf`

elements/lisk-validator/src/formatter.ts
```diff
@@ -29,95 +30,122 @@ import {
...
-		return false;
-	}
-});
-
-validator.addFormat('signedPublicKey', data => {
-	try {
-		const action = data[0];
-		if (action !== '+' && action !== '-') {
-			return false;
+	public validate(schema: object, data: object): [] {
```
nazarhussain: You can also create a method `validateAndThrow` or pass third param to `validate` to throw instead of return. This will simplify the usage in other code files. Just a suggestion.

framework/src/controller/application.js
pablitovicente: As discussed I changed lisk-validator to wrap ajv for it to return any errors during validation and return them. Resolving and starting fixing/rewriting tests.

elements/lisk-validator/src/formatter.ts
```diff
@@ -29,95 +30,122 @@ import {
...
+		const valid = this.validator.validate(schema, data);
+		if (!valid) {
```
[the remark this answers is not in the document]
pablitovicente: Committed from FS as the change also needs removal of const

elements/lisk-validator/src/formatter.ts
pablitovicente: On Friday we discussed that the library should not throw and I do agree with that if you think it should throw or an alternate method should throw please issue a new PR against development.

elements/lisk-validator/src/formatter.ts
```diff
@@ -29,95 +30,122 @@ import {
...
-	}
-});
-
-validator.addFormat('signedPublicKey', data => {
-	try {
-		const action = data[0];
-		if (action !== '+' && action !== '-') {
-			return false;
+	public validate(schema: object, data: object): [] {
+		const valid = this.validator.validate(schema, data);
```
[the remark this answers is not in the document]
pablitovicente: Committed directly from FS.

elements/lisk-validator/src/formatter.ts
```diff
@@ -29,95 +30,122 @@ import {
...
-validator.addFormat('transferAmount', validateTransferAmount);
-
-validator.addFormat('nonTransferAmount', validateNonTransferAmount);
-
-validator.addFormat('fee', validateFee);
-
-validator.addFormat('emptyOrPublicKey', data => {
-	if (data === null || data === '') {
-		return true;
+class LiskValidator {
```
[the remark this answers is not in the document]
pablitovicente: Renamed file to lisk_validator.ts and updated related files to match the change.

elements/lisk-validator/src/formatter.ts
```diff
@@ -29,95 +30,122 @@ import {
...
+
+		this.validator.addFormat('amount', isNumberString);
+
+		this.validator.addFormat('transferAmount', validateTransferAmount);
+
+		this.validator.addFormat('nonTransferAmount', validateNonTransferAmount);
+
+		this.validator.addFormat('fee', validateFee);
+
+		this.validator.addFormat('emptyOrPublicKey', data => {
```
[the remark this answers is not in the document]
pablitovicente: Moved code around as per requested.

elements/lisk-validator/src/formatter.ts
```diff
@@ -29,95 +30,122 @@ import {
...
+			}
+
+			try {
+				validatePublicKey(data);
+
+				return true;
+			} catch (error) {
+				return false;
+			}
+		});
```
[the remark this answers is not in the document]
pablitovicente: File was changed so please feel free to commit any changes to code and test you consider.

elements/lisk-validator/src/validation.ts
```diff
@@ -242,3 +242,17 @@ export const validateFee = (data: string): boolean =>
...
 	!isGreaterThanMaxTransactionAmount(new BigNum(data));
+
+export const isCsv = (data: string): boolean => {
+	if (typeof data !== 'string') {
+		return false;
+	}
+
+	const csvAsArray = data.split(',');
+
+	if (csvAsArray.length > 0) {
```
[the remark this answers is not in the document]
pablitovicente: Sorry I do not understand your comment this is keeping the validation that was previously in place in framework. Please if you can point me to the code or tests for the existing functionality in framework I might be able to understand your feedback about this point better.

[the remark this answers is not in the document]
nazarhussain: This formatter is checking two things, one provided value is a csv and that csv have particular minimum length. I was suggesting to split these two things separate and check the second check by `minLength` specifier of son-schema.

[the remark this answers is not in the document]
nazarhussain: But since this formatter `isCSV` is going to be removed in other PR from @limiaspasdaniel so let's keep it as it is for now.

PR #3798, opened 2019-06-07
Simplify loadBlocksData method - Closes #2752

### LINKED ISSUE #2752: Simplify loadBlocksData method
Currently `modules.blocks.utils.loadBlocksData` is used in 2 different places with different params and expected returned data.

First:
```
modules.blocks.chain.__private.popLastBlock
    modules.blocks.chain.__private.loadSecondLastBlockStep
        modules.blocks.utils.loadBlocksPart
            modules.blocks.utils.loadBlocksData
```

Second:
```
modules.transport.shared.blocks
    modules.blocks.utils.loadBlocksData
```

The first case can be entirely replace by `library.storage.entities.Block.getOne()`
Once the first case is replaced, the methods `modules.blocks.chain.__private.loadSecondLastBlockStep` and `modules.blocks.utils.loadBlocksPart` can be completely removed and `modules.blocks.utils.loadBlocksData` can be refactored to cater only for `modules.transport.shared.blocks`.

Additionally, `loadBlocksData` should be renamed to express better its use case.

This refactoring will greatly simplify the code and understanding of those methods.

1.5

lsilvs:
### What was the problem?
The function `blocks->chain->popLastBlock()` was unnecessarily calling `loadSecondLastBlock` once it could get this data from Storage directly.

### How did I fix it?
I've replaced `loadSecondLastBlock` call by `storage.entities.Block.get` and consequentially I've removed `loadBlocksData`, `loadBlocksPart` and `loadSecondLastBlock` as there were no use for them anymore.

### How to test it?
Run complete test suite

### Review checklist

* The PR resolves #2752
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/chain/blocks/chain.js
+++ b/framework/src/modules/chain/blocks/chain.js
@@ -17,7 +17,7 @@
 const { cloneDeep } = require('lodash');
 const { Status: TransactionStatus } = require('@liskhq/lisk-transactions');
 const transactionsModule = require('../transactions');
-const { loadSecondLastBlock } = require('./utils');
+const { storageRead } = require('./block');
 
 const TRANSACTION_TYPES_VOTE = 3;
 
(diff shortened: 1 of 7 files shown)


framework/src/modules/chain/blocks/utils.js
```diff
@@ -96,7 +96,7 @@ const parseStorageObjToLegacyObj = block => {
 
 /**
  * Generates a list of full blocks for another node upon sync request from that node, see: modules.transport.internal.blocks.
- * NOTE: changing the original method loadBlocksData() could potentially change behaviour (popLastBlock() uses it for instance)
+ * NOTE: changing the original method loadBlocksDataWS() could potentially change behaviour (popLastBlock() uses it for instance)
```
shuse2: Probably name should be changed back to `loadBlocksData`

framework/src/modules/chain/blocks/chain.js
```diff
@@ -417,22 +417,28 @@ const popLastBlock = async (
...
+
+		const secondLastBlock = storageRead(storageResult);
+		secondLastBlock.transactions = interfaceAdapters.transactions.fromBlock(
+			secondLastBlock
+		);
+
 		await undoConfirmedStep(storage, slots, oldLastBlock, exceptions, tx);
 		await backwardTickStep(roundsModule, oldLastBlock, secondLastBlock, tx);
 		await deleteBlock(storage, oldLastBlock.id, tx);
+		return secondLastBlock;
```
shuse2: does `begin` returns the result from inner function?

framework/src/modules/chain/blocks/utils.js
lsilvs: I reckon it's fine to keep `loadBlocksDataWS` as it's only used by Network module but if we want to change the name I would suggest changing it for something more descriptive.

shuse2: yes true, it's only used for syncing.

framework/src/modules/chain/blocks/chain.js
lsilvs: Yes. I've just double checked and it does return the `secondLastBlock` in this case

PR #3804, opened 2019-06-11
Replace sodium-native with @liskhq/lisk-cryptography - Closes #3426

### LINKED ISSUE #3426: Replace sodium-native with @liskhq/lisk-cryptography

lsilvs:
### What was the problem?
`sodium-native` was still as dependency of Lisk Framework

### How did I fix it?
I've removed `framework/src/modules/chain/helpers/ed.js` entirely as it wasn't being used, I've replaced the use of `sodium-native` on test files by using `@liskhq/lisk-cryptography` and finally I've removed `sodium-native` from `package.js`

### How to test it?
Run regular test suite

### Review checklist

* The PR resolves #3426
* Linting passes
* Tests pass


framework/test/mocha/integration/blocks/process/on_receive_block.js
```diff
@@ -187,12 +187,15 @@ describe('integration test (blocks) - process receiveBlockFromNetwork()', () =>
...
 
 	function getKeypair(passphrase) {
-		return library.ed.makeKeypair(
-			crypto
-				.createHash('sha256')
-				.update(passphrase, 'utf8')
-				.digest()
-		);
+		const {
+			publicKeyBytes: publicKey,
```
michielmulders: Just wondering why we are renaming the `publicKeyBytes` to `publicKey`.
Maybe use the naming with bytes as that better represents the content of the variable?

lsilvs: I've renamed it to keep the same interface as before.

framework/test/mocha/unit/modules/chain/logic/test_data/multisignature.js
```diff
@@ -38,7 +35,10 @@ const senderHash = crypto
 	.createHash('sha256')
 	.update(validSender.passphrase, 'utf8')
 	.digest();
-const senderKeypair = ed.makeKeypair(senderHash);
+
+const senderKeypair = getPrivateAndPublicKeyBytesFromPassphrase(
```
shuse2: this will be `senderKeypair.publicKeyBytes` but it seems below didnt change is that ok?

framework/package.json
```diff
@@ -89,7 +89,6 @@
 		"redis": "2.8.0",
 		"semver": "5.6.0",
 		"socket.io": "2.2.0",
-		"sodium-native": "2.2.4",
```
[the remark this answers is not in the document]
lsilvs: humm, isn't it weird to have a dependency in the `package.json` without `require` in the code?
Why is `sodium-native` optional in `lisk-cryptography`? Shouldn't it be mandatory?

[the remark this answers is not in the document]
shuse2: lisk-cryptography supports both browser and server, so it's first option is to use `tweetnacl`, but if sodium-native is installed, it uses the faster version
That's why we kept it as optional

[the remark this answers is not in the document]
lsilvs: Gotcha! I will put it back then ;)

framework/test/mocha/unit/modules/chain/logic/test_data/multisignature.js
lsilvs: Just double checked and it seems these files inside `test_data/` folder are not used. I will remove them all.

lsilvs:
@shuse2 I've created separated issues for that

> There are `crypto` library usage in the block.js.
https://github.com/LiskHQ/lisk-sdk/issues/3808

> Also, `ByteBuffer` library usage can be removed in this PR too
https://github.com/LiskHQ/lisk-sdk/issues/3807

PR #3806, opened 2019-06-11
Move peer.js to util without storing state - Closes#3751

### LINKED ISSUE #3751: Change peer.js to utils without holding states
### Expected behavior
peers.js is for calculating the consensus percent, but it relies on the `broadhash` from the blcok.
It should take the broadhash as an input and calculate on demand.
Then, it can be move to the helper/util file.
Also, it should use the async/await format

### Actual behavior
it still has dependency to the block with previous standards

2.2-

ishantiw:
### What was the problem?

`chain/submodules/peers.js` is for calculating the consensus percent, but it relies on the `broadhash` from the blocks.
It should take the broadhash as input and calculate on demand.

Move it to `chain/helpers/peer.js`
Also, it should use the async/await format

### How did I fix it?

* Change function definition of `Peers.calculateConsensus` and other related functions to accept broadhash in `peers.js`
* Pass broadhash in chain.js forger.js and loader.js when calling the above functions.
* Move `peers.js` in unit tests to helpers folder
* Update the related test

### How to test it?

`npm run mocha:unit`
`npm run mocha:integration`

### Review checklist

* The PR resolves #3751 
* Linting passes
* Tests pass


ishantiw:
@pablitovicente The app was not starting because peers object was `undefined` and that's why all the  Jest tests were failing. I fixed it now.

framework/src/modules/chain/helpers/peers.js
```diff
@@ -43,28 +39,19 @@ const MAX_PEERS = 100;
  */
 class Peers {
 	constructor(scope) {
-		library = {
+		this.library = {
```
pablitovicente: do we need to keep the `library` here? why not set them as properties directly?

ishantiw:
@shuse2 @pablitovicente Resolved all your comments.

framework/src/modules/chain/chain.js
```diff
@@ -402,6 +405,23 @@ module.exports = class Chain {
 			},
 			syncInterval
 		);
+
+		/**
+		 * Periodically calculate consensus
+		 */
+		this.logger.trace('Peers ready');
```
shuse2: Let's make this separate function `_calculateConsensus`.
Also this log is not relevant anymore

framework/src/modules/chain/chain.js
```diff
@@ -402,6 +405,23 @@ module.exports = class Chain {
...
+		const calculateConsensus = async () => {
+			const consensus = await this.scope.peers.calculateConsensus(
```
shuse2: we can use `this.peers`

framework/src/modules/chain/chain.js
```diff
@@ -402,6 +405,23 @@ module.exports = class Chain {
...
+				this.blocks.broadhash
+			);
+			return this.logger.debug(`Broadhash consensus: ${consensus} %`);
+		};
+
+		jobQueue.register(
+			'calculateConsensus',
+			calculateConsensus,
+			this.scope.peers.broadhashConsensusCalculationInterval
```
shuse2: same as above

framework/src/modules/chain/peers.js
```diff
@@ -0,0 +1,96 @@
...
+		if (this.config.forging.force) {
+			return false;
+		}
+		const consensus = await this.calculateConsensus(broadhash);
+		return consensus < this.constants.minBroadhashConsensus;
+	}
+}
+
+// Export
+module.exports = Peers;
```
shuse2: Prefer to export as object
```
module.exports = {
  Peers,
}
```

framework/src/modules/chain/peers.js
```diff
@@ -0,0 +1,96 @@
...
+ * Main peers methods. Initializes library with scope content.
+ *
+ * @class
+ * @memberof modules
+ * @see Parent: {@link chain}
+ * @requires helpers/jobs_queue
+ * @param {scope} scope - App instance
+ */
+class Peers {
+	constructor(scope) {
```
shuse2: Maybe, as other files, we should deconstruct and supply what's needed

framework/src/modules/chain/peers.js
```diff
@@ -0,0 +1,101 @@
...
+ * Main peers methods. Initializes library with scope content.
+ *
+ * @class
+ * @memberof modules
+ * @see Parent: {@link chain}
+ * @param {scope} scope - App instance
+ */
+class Peers {
+	constructor({
+		components: { logger },
```
shuse2: maybe it's better to make the constructer as flat as other ones?
```
{
 logger,
  channel,
}
```
erc

framework/src/modules/chain/chain.js
```diff
@@ -477,6 +480,20 @@ module.exports = class Chain {
...
+		 * Periodically calculate consensus
+		 */
+		jobQueue.register(
+			'calculateConsensus',
+			async () => this._calculateConsensus(this.peers),
+			this.peers.broadhashConsensusCalculationInterval
+		);
+	}
+
+	async _calculateConsensus(peers) {
```
shuse2: It would be simple to register jobQueue in this function it self
```
jobQueue.register(
			'calculateConsensus',
			async () => {
                             const consensus = await this.peers.calculateConsensus(this.blocks.broadhash);
		              return this.logger.debug(`Broadhash consensus: ${consensus} %`);
                          },
			this.peers.broadhashConsensusCalculationInterval
		);
```

ishantiw: Yeah I was thinking to do the same but from your previous comment I extracted it out, also I thought it that it might be used somewhere else too in chain.

framework/src/modules/chain/chain.js
```diff
@@ -178,6 +179,7 @@ module.exports = class Chain {
 			this.channel.subscribe('network:bootstrap', async () => {
 				this._startLoader();
 				await this._startForging();
+				this._calculateConsensus();
```
shuse2: can we move this up before the last `await`?

framework/src/modules/chain/forger.js
```diff
@@ -331,9 +331,13 @@ class Forger {
 			});
 			return;
 		}
-		const isPoorConsensus = await this.peersModule.isPoorConsensus();
+		const isPoorConsensus = await this.peersModule.isPoorConsensus(
+			this.blocksModule.broadhash
+		);
 		if (isPoorConsensus) {
-			const consensusErr = `Inadequate broadhash consensus before forging a block: ${this.peersModule.getLastConsensus()} %`;
+			const consensusErr = `Inadequate broadhash consensus before forging a block: ${await this.peersModule.getLastConsensus(
```
shuse2: maybe it's better to split the consensus calculation in different line
```
const consensus = await this.peersModule.getLastConsensus(this.blocksModule.broadhash);
consensusErr = `.... ${consensus}`;
```

framework/src/modules/chain/forger.js
```diff
@@ -342,7 +346,9 @@ class Forger {
 		}
 
 		this.logger.info(
-			`Broadhash consensus before forging a block: ${this.peersModule.getLastConsensus()} %`
+			`Broadhash consensus before forging a block: ${await this.peersModule.getLastConsensus(
```
shuse2: same as above

PR #3810, opened 2019-06-12
Replace crypto with @liskhq/lisk-cryptography - Closes #3808

### LINKED ISSUE #3808: Replace crypto with @liskhq/lisk-cryptography
### Expected behavior
Expect to use `@liskhq/lisk-cryptography` to manipulate hashes 

### Actual behavior
`framework/src/modules/chain/blocks/block.js` and `framework/src/modules/chain/blocks/verify.js` use `crypto`

`development`

lsilvs:
### What was the problem?
`crypto` lib being used where `@liskhq/lisk-cryptography` should be used instead

### How did I fix it?
By replacing `crypto` with `@liskhq/lisk-cryptography` on `framework/src/modules/chain/blocks/block.js` and `framework/src/modules/chain/blocks/verify.js`

### How to test it?
Run complete test suite

### Review checklist

* The PR resolves #3808
* Linting passes
* Tests pass


framework/src/modules/chain/blocks/block.js
```diff
@@ -210,31 +209,34 @@ const create = ({
...
 
 		totalFee = totalFee.plus(transaction.fee);
 		totalAmount = totalAmount.plus(transaction.amount);
 
 		blockTransactions.push(transaction);
-		payloadHash.update(bytes);
+		transactionsBytesArray.push(transactionBytes);
 	}
 
+	const transactionsBuffer = Buffer.concat(transactionsBytesArray);
```
shuse2: does it have same result as the previous implementation?
This is so much better than I was thinking =p

lsilvs: It does :)

framework/src/modules/chain/blocks/block.js
```diff
@@ -23,7 +23,6 @@ const {
 	verifyData,
 } = require('@liskhq/lisk-cryptography');
 const _ = require('lodash');
-const crypto = require('crypto');
```
diego-G: If we're not using anymore this library we should remove it from the tests as well and ultimately from the `package.json`. Is that feasible @lsilvs ?

PR #3816, opened 2019-06-13
Implement procedure to return the highest common block id - Closes #3769

### LINKED ISSUE #3766: Implement get common blocks from peer
### Expected behavior
Implement get common blocks from a peer. It will be used by BSM and FSM with issues: #3563 and #3562

`feature/BFT`

### LINKED ISSUE #3769: Implement mechanism for finding commonBlock
### Expected behavior
~Peer should be able to send 202 block ids to a specific peer in the network in order to find a common block within this range. If no common block can be found, this means the node can not provide it and the BFT process should be aborted.~

~Also, add functionality for verifying the commonBlock.~

**Remarks:** This functionality can be reused by Block Sync Mechanism @limiaspasdaniel 

**Goal:** Implement functionality as a chain action.
~Requirements: @jondubois PR for targetting specific peer is needed for this to be able to work.~

`feature/introduce-bft`

limiaspasdaniel:
### What was the problem?
`blocksCommon` implementation was outdated and it was also not taking into account the full list of block identifiers sent by the requester.
### How did I fix it?
Implementation:
- Rename `blocksCommon` to `getHighestCommonBlockId`
- Create a public action in `Chain` module
- Ditch input CSV format for JSON arrays.
- Validate the request query using JSON Schema to check if:
	- an array named "ids" is present
	- the array is not empty (at least 1 block ids needs to included)
	- the array contains no duplicated ids
- Return null if no common block is found.
- Feed back validation errors if the input has an invalid format or content

Testing:
- Update unit tests
- Update functional tests

### How to test it?
Run the following tests:
- `framework/test/mocha/unit/modules/chain/transport.js`
- `framework/test/mocha/functional/ws/transport/blocks.js`
### Review checklist
* The PR resolves #3769 
* Linting passes
* Tests pass


framework/src/modules/chain/transport.js
```diff
@@ -223,80 +223,35 @@ class Transport {
...
+			 * @return {Promise<Array<String>>}
 			 */
-			async blocksCommon(query) {
-				query = query || {};
-
+			async getCommonBlocks(query) {
 				const valid = library.schema.validate(
 					query,
-					definitions.WSBlocksCommonRequest
+					definitions.WSGetCommonBlocksRequest
```
shuse2: maybe rename this definition to `GetCommonBlocksRequest`? it's not related to WS anymore

framework/src/modules/chain/transport.js
```diff
@@ -223,80 +223,35 @@ class Transport {
...
-					return {
-						success: true,
-						common: parsedRow,
-					};
-				} catch (error) {
-					library.logger.error(error.stack);
-					throw new Error('Failed to get common block');
+					return (await library.storage.entities.Block.get({
+						id_in: query.ids,
+					})).map(commonRawBlock => commonRawBlock.id);
```
shuse2: Probably we should decide on what format we send the response.

shuse2: Shouldn't we return heightest common block id?

limiaspasdaniel: Please see https://github.com/LiskHQ/lisk-sdk/pull/3816/commits/fb01dd27ff96fa56e495cf970642fcd88cf5c341 's message and let me know what you think.

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -493,6 +493,24 @@ class Blocks extends EventEmitter {
 		}
 	}
 
+	/**
+	 * Returns the largest common block between ids.ids and the database blocks table
```
michielmulders: Typo? Prefer IDs (also for the variable naming in the `getHighestCommonBlock` function.

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -493,6 +493,24 @@ class Blocks extends EventEmitter {
...
+	 * @return {Promise<BasicBlock|undefined>}
+	 */
+	async getHighestCommonBlock(ids) {
+		try {
+			const block = await this.storage.entities.Block.getMatchingHighestBlock(
+				ids
+			);
+			return block;
+		} catch (e) {
+			const errMessage = 'Failed to read common blocks from storage';
```
michielmulders: Failed to read common blocks implies that you couldn't find a common block.
However, an error is only thrown in case of bad network connection or unable to access storage layer.
Prefer to throw more generic message like `Failed to access storage layer/or network`.

framework/src/modules/chain/index.js
```diff
@@ -126,8 +126,9 @@ module.exports = class ChainModule extends BaseModule {
 				handler: async action => this.chain.actions.blocks(action),
 				isPublic: true,
 			},
-			blocksCommon: {
-				handler: async action => this.chain.actions.blocksCommon(action),
+			getHighestCommonBlockId: {
```
michielmulders: Security wise, this action needs to be public but this also implies someone can write a custom script that tries to overload the node by requesting to find a common ID by sending continuously fake IDs. We should take care of a rate-limiting security measure of some sort.

framework/src/modules/chain/schema/definitions.js
```diff
@@ -99,14 +99,18 @@ module.exports = {
...
+		id: 'getHighestCommonBlockIdRequest',
 		type: 'object',
 		required: ['ids'],
 		properties: {
 			ids: {
-				type: 'string',
-				format: 'csv',
+				type: 'array',
+				items: {
+					type: 'string',
```
michielmulders:
```suggestion
					type: 'string',
					format: 'id',
```

michielmulders: Formatting using ID

framework/src/modules/chain/transport.js
```diff
@@ -205,7 +205,7 @@ class Transport {
 	}
 
 	/**
-	 * @property {function} blocksCommon
+	 * @property {function} getCommonBlocks
```
michielmulders: @limiaspasdaniel

framework/src/modules/chain/chain.js
```diff
@@ -268,8 +269,32 @@ module.exports = class Chain {
 			}),
 			blocks: async action =>
 				this.scope.modules.transport.shared.blocks(action.params || {}),
-			blocksCommon: async action =>
-				this.scope.modules.transport.shared.blocksCommon(action.params || {}),
+
+			getHighestCommonBlockId: async action => {
+				const valid = this.scope.schema.validate(
```
michielmulders: Would be nice if we wouldn't need all this logic inside the chain actions to keep this clean?

limiaspasdaniel: We definitely need to think what to do with this, if Transport is a good way to hold action query schema validation and logic routing. I am personally not a big fan of the current state of transport.js as it doesn't only hold transport logic strictly.

From my perspective, Transport could hold action query schema validation and response formatting and in case the validation succeeds could route the query to the desired logic stored in another chain submodule, but as of now, this is not the case. We have to discuss this further.

framework/src/modules/chain/components/storage/entities/block.js
```diff
@@ -183,6 +183,22 @@ class ChainBlock extends BlockEntity {
 			{}
 		);
 	}
+
+	/**
+	 * Get the matching highest block for the providen array of block ids
+	 * @param {Array} blockIds - An array of block ids
+	 * @return {Promise<BasicBlock>}
+	 */
+	async getMatchingHighestBlock(blockIds) {
```
shuse2: Isn't this too specific to the logic? 
Probably we can just put this logic in the `blocks.js`

limiaspasdaniel: I don't have a strong opinion regarding this, for me it is correct in both ways. @nazarhussain discussed this with me and can give his input.

framework/src/modules/chain/chain.js
SargeKhan: Isn't tranport doing the validation schema currently for all the other actions?

SargeKhan:
After discussion with @shuse2, the point i raised turned out to be invalid. So, please forget about my last comment.
I'll review the PR again.

framework/src/modules/chain/components/storage/entities/block.js
shuse2: @nazarhussain what's your opinion on this?
I feel like so far most of the functions on entities are generic, and we should keep it that way

nazarhussain: After looking at the implementation I agree to @shuse2. Since it don't have any custom SQL or complex logic behind, so can be used only in the implementation as well. 

```
entities.Block.get(
 			{
 				id_in: blockIds,
 			},
 			{ sort: 'height:desc', limit: 1 }
 		)
```

limiaspasdaniel: @shuse2 @nazarhussain logic has been moved back to its place.Done

PR #3818, opened 2019-06-13
Add bunyan standard serializer - Closes #3838

### LINKED ISSUE #3838: Add Bunyan standard serializer
### Expected behavior
To be able to log error objects along with metadata and a description message, by following this format:
`const anErrorInstance = new Error('Storage read error');
`logger.error({err: anErrorInstance, blockId: '1234', 'Failure when generating a new block'}`

This produces the following output: 

image

### Actual behavior
When trying to log error object along with metadata and a description message, errors are not being properly printed.
`development`

limiaspasdaniel:
### What was the problem?
There is no possibility to log error objects, metadata and log description at the same time because the Error instance is not being serialized.
### How did I fix it?
As described in [Bunyan's documentation](https://github.com/trentm/node-bunyan#log-method-api), I have enabled the Bunyan's standard serializer which allows error logging along with metadata with by for example doing:

```
const anErrorInstance = new Error('Storage read error');
logger.error({err: anErrorInstance, blockId: '1234', 'Failure when generating a new block'}
```

Note that the way to tell Bunyan to pretty print error instances is by adding the `err` key to the object literal in the first argument.

This produces the following nice output when Bunyan is executed with `-o short` CLI argument:
Captura de pantalla 2019-06-13 a la(s) 3 12 30 p  m

This is the output if the serializer is not set:
`13:14:57.327Z ERROR lisk-framework: Failure when generating a new block (err={}, blockId=1)`
### Review checklist

* The PR resolves #3838
* Linting passes
* Tests pass


pablitovicente:
I would first check what's the impact of the serialisation option in bunyan's performance as I think a similar output might be achieved by multiple calls to `logger.err` or any other bunyan methods being used.

limiaspasdaniel:
I think the performance might be the same as doing `logger.error(anErrorInstance)` in another line should be performing the same operation against the error instance.

pablitovicente:
> I think the performance might be the same as doing `logger.error(anErrorInstance)` in another line should be performing the same operation against the error instance.

It's a little different as when serialisers are enabled bunyan's code calls `_applySerializers` as an extra step but given that we don't have numbers to measure difference between enabled them or not I don't we can get to a conclusion about the difference.

shuse2:
I think considering the fact that err doesn't / shouldn't happen that often, it would be more helpful to have serialized format than the performance

limiaspasdaniel:
> > I think the performance might be the same as doing `logger.error(anErrorInstance)` in another line should be performing the same operation against the error instance.
> 
> It's a little different as when serialisers are enabled bunyan's code calls `_applySerializers` as an extra step when logging but given that we don't have numbers to measure difference between enabled them or not I don't we can get to a conclusion about the difference.

Absolutely @pablitovicente, however, @shuse2 also has a good point on this, as an error shouldn't happen that often, readability, in my opinion, is more valuable

pablitovicente:
> > > I think the performance might be the same as doing `logger.error(anErrorInstance)` in another line should be performing the same operation against the error instance.
> > 
> > 
> > It's a little different as when serialisers are enabled bunyan's code calls `_applySerializers` as an extra step when logging but given that we don't have numbers to measure difference between enabled them or not I don't we can get to a conclusion about the difference.
> 
> Absolutely @pablitovicente, however, @shuse2 also has a good point on this, as an error shouldn't happen that often, readability, in my opinion, is more valuable

Yeah I do agree that added value is good!

framework/src/components/logger/logger.js
```diff
@@ -61,6 +61,7 @@ const createLogger = ({ fileLogLevel, consoleLogLevel, logFileName }) => {
 		name: 'lisk-framework',
 		streams,
 		src: consoleSrc || fileSrc,
+		serializers: bunyan.stdSerializers,
```
pablitovicente: What if instead of using all serialisers we only use `bunyan.stdSerializers.err` ?

pablitovicente:
```suggestion
		serializers: {err: bunyan.stdSerializers.err},
```

limiaspasdaniel: Good one @pablitovicente

limiaspasdaniel: Great!

limiaspasdaniel:
@shuse2 ready to merge

shuse2:
Closing due to github error. It is merged to development

PR #3824, opened 2019-06-14
Inbound peer eviction method - Closes #3721

### LINKED ISSUE #3721: Introduce specific eviction method for incoming connections
### Expected behavior
The Bitcoin eviction method was proposed in the LIP.
https://github.com/bitcoin/bitcoin/blob/17d644901bbd770578619545bbd8bc930fe6f2d9/src/net.cpp#L1000

### Actual behavior

We randomly remove an ongoing incoming connection when another peer attempts to connect to us and we currently hold the maximum allowed incoming connections.

`2.0.0+`

mitsujutsu:
### What was the problem?

In the current implementation, when the maximum number of inbound peers are connected and a new peer attempts to connect to the node, we randomly select and remove an ongoing inbound peer connection.

### How did I fix it?

In the new protocol, we use the strategy of protecting a number of peers for several distinct characteristics for increased security. 

Netgroups: We calculate a randomized value `keyedNetgroup` based on IP group for each eviction candidate. We protect 4 peers with the smallest  `keyedNetgroup` values. An attacker cannot predict which netgroups will be protected.
* To be started in bucket system milestone

- [x] Peer latency: We protect a number of peers with optimal latency. An attacker cannot manipulate this metric without physically moving nodes closer to the target.

- [x] Useful work: We protect peers who perform useful work, such as peers who have recently sent blocks and transactions. An attacker cannot manipulate this metric without performing useful work.

- [x] Connection time: We protect half of the remaining nodes who have been connected the longest. This precludes attacks that start later.

### Review checklist

* The PR resolves #3721 
* Linting passes
* Tests pass

### DIFF
--- a/elements/lisk-p2p/src/p2p.ts
+++ b/elements/lisk-p2p/src/p2p.ts
@@ -125,6 +125,7 @@ const BASE_10_RADIX = 10;
 const DEFAULT_MAX_OUTBOUND_CONNECTIONS = 20;
 const DEFAULT_MAX_INBOUND_CONNECTIONS = 100;
 const DEFAULT_OUTBOUND_SHUFFLE_INTERVAL = 300000;
+const DEFAULT_EVICTION_PROTECTION = 0.58;
 
 const selectRandomPeerSample = (
 	peerList: ReadonlyArray<P2PPeerInfo>,
@@ -363,8 +364,11 @@ export class P2P extends EventEmitter {
 			outboundShuffleInterval: config.outboundShuffleInterval
 				? config.outboundShuffleInterval
 				: DEFAULT_OUTBOUND_SHUFFLE_INTERVAL,
-			evictionProtectionEnabled: config.evictionProtectionEnabled === false ? false : true,
-			peerProtectionRatio: config.peerProtectionRatio ? config.peerProtectionRatio : 0.58,
+			evictionProtectionEnabled:
+				config.evictionProtectionEnabled === false ? false : true,
+			evictionProtectionRatio: config.evictionProtectionRatio
+				? config.evictionProtectionRatio
+				: DEFAULT_EVICTION_PROTECTION,
 		});
 
 		this._bindHandlersToPeerPool(this._peerPool);
--- a/elements/lisk-p2p/src/p2p.ts
+++ b/elements/lisk-p2p/src/p2p.ts
@@ -125,7 +125,8 @@ const BASE_10_RADIX = 10;
 const DEFAULT_MAX_OUTBOUND_CONNECTIONS = 20;
 const DEFAULT_MAX_INBOUND_CONNECTIONS = 100;
 const DEFAULT_OUTBOUND_SHUFFLE_INTERVAL = 300000;
-const DEFAULT_EVICTION_PROTECTION = 0.58;
+const DEFAULT_PEER_PROTECTION_FOR_LATENCY_AND_USEFULNESS = 0.068;
+const DEFAULT_PEER_PROTECTION_FOR_LONGEVITY = 0.5;
 
 const selectRandomPeerSample = (
 	peerList: ReadonlyArray<P2PPeerInfo>,
@@ -364,11 +365,18 @@ export class P2P extends EventEmitter {
 			outboundShuffleInterval: config.outboundShuffleInterval
 				? config.outboundShuffleInterval
 				: DEFAULT_OUTBOUND_SHUFFLE_INTERVAL,
-			evictionProtectionEnabled:
-				config.evictionProtectionEnabled === false ? false : true,
-			evictionProtectionRatio: config.evictionProtectionRatio
-				? config.evictionProtectionRatio
-				: DEFAULT_EVICTION_PROTECTION,
+			latencyProtectionRatio:
+				typeof config.latencyProtectionRatio === 'number'
+					? config.latencyProtectionRatio
+					: DEFAULT_PEER_PROTECTION_FOR_LATENCY_AND_USEFULNESS,
+			productivityProtectionRatio:
+				typeof config.productivityProtectionRatio === 'number'
+					? config.productivityProtectionRatio
+					: DEFAULT_PEER_PROTECTION_FOR_LATENCY_AND_USEFULNESS,
+			longevityProtectionRatio:
+				typeof config.longevityProtectionRatio === 'number'
+					? config.longevityProtectionRatio
+					: DEFAULT_PEER_PROTECTION_FOR_LONGEVITY,
 		});
 
 		this._bindHandlersToPeerPool(this._peerPool);
--- a/elements/lisk-p2p/src/p2p.ts
+++ b/elements/lisk-p2p/src/p2p.ts
@@ -334,6 +334,7 @@ export class P2P extends EventEmitter {
 			// Re-emit the error to allow it to bubble up the class hierarchy.
 			this.emit(EVENT_INBOUND_SOCKET_ERROR, error);
 		};
+
 		this._peerPool = new PeerPool({
 			connectTimeout: config.connectTimeout,
 			ackTimeout: config.ackTimeout,
@@ -362,6 +363,8 @@ export class P2P extends EventEmitter {
 			outboundShuffleInterval: config.outboundShuffleInterval
 				? config.outboundShuffleInterval
 				: DEFAULT_OUTBOUND_SHUFFLE_INTERVAL,
+			evictionProtectionEnabled: config.evictionProtectionEnabled === false ? false : true,
+			peerProtectionRatio: config.peerProtectionRatio ? config.peerProtectionRatio : 0.58,
 		});
 
 		this._bindHandlersToPeerPool(this._peerPool);
--- a/elements/lisk-p2p/src/p2p.ts
+++ b/elements/lisk-p2p/src/p2p.ts
@@ -126,7 +126,8 @@ const BASE_10_RADIX = 10;
 const DEFAULT_MAX_OUTBOUND_CONNECTIONS = 20;
 const DEFAULT_MAX_INBOUND_CONNECTIONS = 100;
 const DEFAULT_OUTBOUND_SHUFFLE_INTERVAL = 300000;
-const DEFAULT_PEER_PROTECTION_FOR_LATENCY_AND_USEFULNESS = 0.068;
+const DEFAULT_PEER_PROTECTION_FOR_LATENCY = 0.068;
+const DEFAULT_PEER_PROTECTION_FOR_USEFULNESS = 0.068;
 const DEFAULT_PEER_PROTECTION_FOR_LONGEVITY = 0.5;
 
 const selectRandomPeerSample = (
@@ -397,11 +398,11 @@ export class P2P extends EventEmitter {
 			latencyProtectionRatio:
 				typeof config.latencyProtectionRatio === 'number'
 					? config.latencyProtectionRatio
-					: DEFAULT_PEER_PROTECTION_FOR_LATENCY_AND_USEFULNESS,
+					: DEFAULT_PEER_PROTECTION_FOR_LATENCY,
 			productivityProtectionRatio:
 				typeof config.productivityProtectionRatio === 'number'
 					? config.productivityProtectionRatio
-					: DEFAULT_PEER_PROTECTION_FOR_LATENCY_AND_USEFULNESS,
+					: DEFAULT_PEER_PROTECTION_FOR_USEFULNESS,
 			longevityProtectionRatio:
 				typeof config.longevityProtectionRatio === 'number'
 					? config.longevityProtectionRatio
--- a/elements/lisk-p2p/src/p2p_request.ts
+++ b/elements/lisk-p2p/src/p2p_request.ts
@@ -32,12 +32,23 @@ export class P2PRequest {
 		data: unknown,
 		peerId: string,
 		rate: number,
+		productivity: {
+			// tslint:disable-next-line: readonly-keyword
+			requestCounter: number;
+			// tslint:disable-next-line: readonly-keyword
+			responseCounter: number;
+			// tslint:disable-next-line: readonly-keyword
+			responseRate: number;
+			// tslint:disable-next-line: readonly-keyword
+			lastResponded: number;
+		},
 		respondCallback: (responseError?: Error, responseData?: unknown) => void,
 	) {
 		this._procedure = procedure;
 		this._data = data;
 		this._peerId = peerId;
 		this._rate = rate;
+		productivity.requestCounter += 1;
 		this._respondCallback = (
 			responseError?: Error,
 			responsePacket?: P2PResponsePacket,
@@ -48,10 +59,16 @@ export class P2PRequest {
 				);
 			}
 			this._wasResponseSent = true;
+			// We assume peer performed useful work and update peer response rate
+			if (!responseError && responsePacket) {
+				productivity.lastResponded = Date.now();
+				productivity.responseCounter += 1;
+				productivity.responseRate =
+					productivity.responseCounter / productivity.requestCounter;
+			}
 			respondCallback(responseError, responsePacket);
 		};
 		this._wasResponseSent = false;
-		this._peerId = peerId;
 	}
 
 	public get procedure(): string {
--- a/elements/lisk-p2p/src/p2p_request.ts
+++ b/elements/lisk-p2p/src/p2p_request.ts
@@ -16,6 +16,23 @@
 import { RPCResponseAlreadySentError } from './errors';
 import { P2PResponsePacket } from './p2p_types';
 
+interface RequestOptions {
+	readonly procedure: string;
+	readonly data: unknown;
+	readonly id: string;
+	readonly rate: number;
+	productivity: {
+		// tslint:disable-next-line: readonly-keyword
+		requestCounter: number;
+		// tslint:disable-next-line: readonly-keyword
+		responseCounter: number;
+		// tslint:disable-next-line: readonly-keyword
+		responseRate: number;
+		// tslint:disable-next-line: readonly-keyword
+		lastResponded: number;
+	};
+}
+
 export class P2PRequest {
 	private readonly _procedure: string;
 	private readonly _data: unknown;
@@ -28,27 +45,14 @@ export class P2PRequest {
 	private readonly _rate: number;
 
 	public constructor(
-		procedure: string,
-		data: unknown,
-		peerId: string,
-		rate: number,
-		productivity: {
-			// tslint:disable-next-line: readonly-keyword
-			requestCounter: number;
-			// tslint:disable-next-line: readonly-keyword
-			responseCounter: number;
-			// tslint:disable-next-line: readonly-keyword
-			responseRate: number;
-			// tslint:disable-next-line: readonly-keyword
-			lastResponded: number;
-		},
+		options: RequestOptions,
 		respondCallback: (responseError?: Error, responseData?: unknown) => void,
 	) {
-		this._procedure = procedure;
-		this._data = data;
-		this._peerId = peerId;
-		this._rate = rate;
-		productivity.requestCounter += 1;
+		this._procedure = options.procedure;
+		this._data = options.data;
+		this._peerId = options.id;
+		this._rate = options.rate;
+		options.productivity.requestCounter += 1;
 		this._respondCallback = (
 			responseError?: Error,
 			responsePacket?: P2PResponsePacket,
@@ -61,10 +65,11 @@ export class P2PRequest {
 			this._wasResponseSent = true;
 			// We assume peer performed useful work and update peer response rate
 			if (!responseError && responsePacket) {
-				productivity.lastResponded = Date.now();
-				productivity.responseCounter += 1;
-				productivity.responseRate =
-					productivity.responseCounter / productivity.requestCounter;
+				options.productivity.lastResponded = Date.now();
+				options.productivity.responseCounter += 1;
+				options.productivity.responseRate =
+					options.productivity.responseCounter /
+					options.productivity.requestCounter;
 			}
 			respondCallback(responseError, responsePacket);
 		};
--- a/elements/lisk-p2p/src/peer/base.ts
+++ b/elements/lisk-p2p/src/peer/base.ts
@@ -91,6 +91,13 @@ export const DEFAULT_CONNECT_TIMEOUT = 2000;
 export const DEFAULT_ACK_TIMEOUT = 2000;
 export const DEFAULT_REPUTATION_SCORE = 100;
 export const DEFAULT_RATE_INTERVAL = 1000;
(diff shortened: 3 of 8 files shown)


mitsujutsu:
@jondubois @diego-G "Netgrouping" portion of evicting inbound peers is closely tied to bucket system, and will be handled together with bucket implementation.

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -96,6 +96,8 @@ interface PeerPoolConfig {
 	readonly maxOutboundConnections: number;
 	readonly maxInboundConnections: number;
 	readonly outboundShuffleInterval?: number;
+	readonly evictionProtectionEnabled?: boolean;
+	readonly evictionProtectionRatio?: number;
```
jondubois: I think that we should make `peerBanTime`, `outboundShuffleInterval`, `evictionProtectionEnabled` and `evictionProtectionRatio` compulsory because otherwise we have to do a lot of type casting like `this.someProperty as number` in our code and we lose type safety.

We can keep `ackTimeout` and `connectTimeout` as optional because those are passed to SocketCluster so we don't need to worry about type casting them.

elements/lisk-p2p/src/p2p.ts
```diff
@@ -125,6 +125,7 @@ const BASE_10_RADIX = 10;
 const DEFAULT_MAX_OUTBOUND_CONNECTIONS = 20;
 const DEFAULT_MAX_INBOUND_CONNECTIONS = 100;
 const DEFAULT_OUTBOUND_SHUFFLE_INTERVAL = 300000;
+const DEFAULT_EVICTION_PROTECTION = 0.58;
```
jondubois: Does this mean that we protect 58% of peers in our list?
If so, it seems like a lot. Can you double check with Jan?

elements/lisk-p2p/src/peer/inbound.ts
```diff
@@ -27,14 +26,16 @@ import { SCServerSocket } from 'socketcluster-server';
 
 export const EVENT_CLOSE_INBOUND = 'closeInbound';
 export const EVENT_INBOUND_SOCKET_ERROR = 'inboundSocketError';
-
+const DEFAULT_PING_INTERVAL = 500;
```
jondubois: We should ping less often by default. Maybe once every 1 minute? We don't want to allocate too many resources to pinging; it creates a lot of messages in the network.

elements/lisk-p2p/src/peer/inbound.ts
```diff
@@ -86,6 +97,7 @@ export class InboundPeer extends Peer {
 			'postTransactions',
 			this._handleRawLegacyMessagePostTransactions,
 		);
+		inboundSocket.on('pong', this._handlePong);
```
jondubois: We should define a new constant for 'ping' and 'pong' events at the top of the file.

elements/lisk-p2p/src/peer/outbound.ts
```diff
@@ -165,6 +165,10 @@ export class OutboundPeer extends Peer {
 			});
 		});
 
+		outboundSocket.on('ping', () => {
+			this.emit('pong', Date.now());
```
jondubois: We cannot trust the time which the peer provides in the pong event which they emit. We need to calculate the delay ourselves; start the timer when we do the ping and stop the timer when we receive the pong from the peer.

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -490,16 +496,75 @@ export class PeerPool extends EventEmitter {
 		})();
 	}
 
+	private _selectPeersForEviction(peers: Peer[]): Peer[] {
+		const PEER_PROTECTION_PERCENTAGE = 0.08;
```
jondubois: I think there is a naming issue here. It's difficult to understand how many peers of each kind are being protected.

elements/lisk-p2p/src/p2p.ts
mitsujutsu: The percentage is derived from the following [bitcoin strategy of eviction](https://github.com/bitcoin/bitcoin/blob/17d644901bbd770578619545bbd8bc930fe6f2d9/src/net.cpp#L1000). Bitcoin allows maximum 117 incoming connections. Out of those, 4 are protected by `netgroup`, 8 by `latency`, 8 by `useful work`, making 20 peers. Half of the remaining peers (97 / 2 = 48.5 peers) are protected for connected time. (20 + 48.5) / 117 = ~0.58

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -128,10 +135,19 @@ export class Peer extends EventEmitter {
 	protected readonly _ipAddress: string;
 	protected readonly _wsPort: number;
 	private readonly _height: number;
-	private _reputation: number;
+	protected _reputation: number;
+	protected _latency: number;
+	protected _connectTime: number;
+	public productivity: {
```
jondubois: Can we also make this one private or protected and expose it as a getter? Also, maybe in the getter we can return a clone of the object to prevent outside code from tempering with the internal productivity object?

elements/lisk-p2p/src/peer/inbound.ts
```diff
@@ -106,5 +124,6 @@ export class InboundPeer extends Peer {
 			'postTransactions',
 			this._handleRawLegacyMessagePostTransactions,
 		);
+		inboundSocket.off('pong', this._handlePong);
```
jondubois: We should use `EVENT_PONG` for consistency.

elements/lisk-p2p/src/peer/inbound.ts
```diff
@@ -46,13 +51,25 @@ export class InboundPeer extends Peer {
...
 			this.emit(EVENT_CLOSE_INBOUND, {
 				peerInfo,
 				code,
 				reason,
 			});
 		};
+		this._pingStart = Date.now();
+		this._handlePong = () => {
+			const latency = this._pingStart - Date.now();
+			this._latency = latency;
```
jondubois: A malicious peer could ignore the first ping, and then emit a pong at the exact same time as it expects us to send the next ping (so basically it emits the pong before it actually received the ping) so it could get a low latency score. Alternatively, the peer could just constantly spam us with pongs every few milliseconds.

We could:
1. Randomize the ping interval a bit.
2. Add a callback to the `socket.emit(event, callback)` and then the peer just sends a response instead of emitting a separate pong; that way each response (pong) is coupled to its request (ping) so it cannot be gamed by spamming pongs.

jondubois: It's a lot more difficult to game than before though.

elements/lisk-p2p/test/integration/p2p.ts
```diff
@@ -1475,4 +1475,82 @@ describe('Integration tests for P2P library', () => {
...
+				p2pNodeList.forEach(async (p2p, i) => {
+					// Disregard early peers who may be re-discovered by already evicted peers
+					if (i <= 3) {
+						return;
+					}
+					const inboundPeers = p2p['_peerPool']
+						.getPeers(InboundPeer)
+						.map(peer => peer.wsPort);
+					const lastPeerPort = 5009;
+					expect(inboundPeers).to.not.include(lastPeerPort);
```
diego-G: I don't get the scenario. We should think of something more robust than this.

elements/lisk-p2p/src/peer/inbound.ts
```diff
@@ -27,14 +26,20 @@ import { SCServerSocket } from 'socketcluster-server';
 
 export const EVENT_CLOSE_INBOUND = 'closeInbound';
 export const EVENT_INBOUND_SOCKET_ERROR = 'inboundSocketError';
+export const EVENT_PING = 'ping';
```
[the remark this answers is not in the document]
mitsujutsu: I changed it to event `response` instead of a `PONG` event due to @jondubois comment regarding a security issue. An attacker can potentially spam `PONG` events and fake a low latency, whereas with the event response it is explicitly coupled with the `PING` event and cannot be faked as easily.

elements/lisk-p2p/test/integration/p2p.ts
mitsujutsu: This scenario is testing specifically for the `connectTime` peer protection category, where peers who connected the earliest are less likely to be evicted.  I am testing for this as it is difficult to simulate peers with varying `latency` as well as `responseRate` in these integration tests, without potentially using something like `rewire`. I agree that we should have more robust tests eventually. Thoughts? @jondubois

elements/lisk-p2p/src/p2p.ts
```diff
@@ -362,6 +364,11 @@ export class P2P extends EventEmitter {
 			outboundShuffleInterval: config.outboundShuffleInterval
 				? config.outboundShuffleInterval
 				: DEFAULT_OUTBOUND_SHUFFLE_INTERVAL,
+			evictionProtectionEnabled:
+				config.evictionProtectionEnabled === false ? false : true,
```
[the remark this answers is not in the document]
mitsujutsu: In my opinion it's not necessary as we presume the network module will follow the recommended LIP 0004 procedure for eviction, and we want to discourage modification of that.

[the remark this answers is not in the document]
diego-G: Then, why do we need the `config.evictionProtectionRatio`?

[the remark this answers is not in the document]
mitsujutsu: For users of `lisk-p2p` working on projects unrelated to lisk or even blockchain, they may want to use a ratio that is totally different for their use-case.

elements/lisk-p2p/test/integration/p2p.ts
diego-G: I've received this error when executing the test:
```
(node:3368) UnhandledPromiseRejectionWarning: AssertionError: expected [ 5003, 5002, 5000, 5001, 5009 ] to not include 5009
    at p2pNodeList.forEach (/Users/diego-g/git/lisk-sdk/elements/lisk-p2p/test/integration/p2p.ts:1551:34)
    at Array.forEach (<anonymous>)
    at Context.it (/Users/diego-g/git/lisk-sdk/elements/lisk-p2p/test/integration/p2p.ts:1542:17)
    at callFn (/Users/diego-g/git/lisk-sdk/elements/lisk-p2p/node_modules/mocha/lib/runnable.js:372:21)
    at Test.Runnable.run (/Users/diego-g/git/lisk-sdk/elements/lisk-p2p/node_modules/mocha/lib/runnable.js:364:7)
    at Runner.runTest (/Users/diego-g/git/lisk-sdk/elements/lisk-p2p/node_modules/mocha/lib/runner.js:455:10)
    at /Users/diego-g/git/lisk-sdk/elements/lisk-p2p/node_modules/mocha/lib/runner.js:573:12
    at next (/Users/diego-g/git/lisk-sdk/elements/lisk-p2p/node_modules/mocha/lib/runner.js:369:14)
    at /Users/diego-g/git/lisk-sdk/elements/lisk-p2p/node_modules/mocha/lib/runner.js:379:7
    at next (/Users/diego-g/git/lisk-sdk/elements/lisk-p2p/node_modules/mocha/lib/runner.js:303:14)
    at Immediate.<anonymous> (/Users/diego-g/git/lisk-sdk/elements/lisk-p2p/node_modules/mocha/lib/runner.js:347:5)
    at runCallback (timers.js:705:18)
    at tryOnImmediate (timers.js:676:5)
    at processImmediate (timers.js:658:5)
(node:3368) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). (rejection id: 1)
(node:3368) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
```

jondubois: This is difficult to test. One way to make it easier/more predictable would be to add support for additional config options to allow us to enable and disable each of the 3 characteristics individually then each test could test just one characteristic and turn off the others:

- Peer latency: We protect a number of peers with optimal latency. An attacker cannot manipulate this metric without physically moving nodes closer to the target.

  - Useful work: We protect peers who perform useful work, such as peers who have recently sent blocks and transactions. An attacker cannot manipulate this metric without performing useful work.

  - Connection time: We protect half of the remaining nodes who have been connected the longest. This precludes attacks that start later.

elements/lisk-p2p/test/integration/p2p.ts
```diff
@@ -1475,4 +1475,82 @@ describe('Integration tests for P2P library', () => {
...
+				p2pNodeList
+					.filter(p2p => p2p.isActive)
+					.map(async p2p => await p2p.stop()),
+			);
+			await wait(100);
+		});
+
+		describe('Inbound peer evictions', () => {
+			it('should evict inbound peer which connected the latest', async () => {
+				p2pNodeList.forEach(async (p2p, i) => {
```
diego-G: The test fails. Careful with the `async` inside a forEach.
```suggestion
				p2pNodeList.forEach((p2p, i) => {
```

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -91,6 +98,13 @@ export const DEFAULT_CONNECT_TIMEOUT = 2000;
 export const DEFAULT_ACK_TIMEOUT = 2000;
 export const DEFAULT_REPUTATION_SCORE = 100;
 export const DEFAULT_RATE_INTERVAL = 1000;
+export const DEFAULT_PRODUCTIVITY_RESET_INTERVAL = 2000;
+export const DEFAULT_PRODUCTIVITY = {
+	requestCounter: 0,
+	responseCounter: 0,
+	responseRate: 0,
+	lastResponded: 0,
```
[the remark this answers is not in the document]
mitsujutsu: I think `0` is better since there is no response yet upon initialization. Either way I think it makes no difference.

elements/lisk-p2p/test/integration/p2p.ts
mitsujutsu: I have created new configurations to enable.disable protection by category (latency, connecttime, response, etc). For the integration test I only enable connectTime, and limit the scope of peers to make the behavior predictable.

elements/lisk-p2p/src/peer/outbound.ts
```diff
@@ -165,6 +167,13 @@ export class OutboundPeer extends Peer {
 			});
 		});
 
+		outboundSocket.on(
+			EVENT_PING,
+			(_: undefined, res: (_: undefined, data: string) => void) => {
+				res(undefined, 'pong');
```
shuse2: EVENT_PING is const, so we should make this `const` too

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -92,15 +92,41 @@ interface PeerPoolConfig {
...
+): Peer[] => {
+	// tslint:disable-next-line no-magic-numbers
+	if (options.percentage > 1 || options.percentage < 0) {
+		return peers;
+	}
+	const peerCount = Math.ceil(peers.length * options.percentage);
+	const sign = !!options.asc ? 1 : -1;
+
+	return peers
+		.sort((a: any, b: any) =>
```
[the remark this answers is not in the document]
mitsujutsu: No index signature on type `Peer`

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -92,15 +92,41 @@ interface PeerPoolConfig {
...
+			a[options.category] > b[options.category] ? sign : sign * -1,
+		)
+		.slice(peerCount, peers.length);
```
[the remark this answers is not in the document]
mitsujutsu: It's slicing out the beginning where the ideal peers are sorted at top.

elements/lisk-p2p/src/p2p_request.ts
```diff
@@ -16,6 +16,23 @@
 import { RPCResponseAlreadySentError } from './errors';
 import { P2PResponsePacket } from './p2p_types';
 
+interface RequestOptions {
+	readonly procedure: string;
+	readonly data: unknown;
+	readonly id: string;
+	readonly rate: number;
+	productivity: {
```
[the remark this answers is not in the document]
mitsujutsu: To have access to `end` of request, so we know a response was sent. We are modifying peer.productivity in this function and thus is not readonly.

elements/lisk-p2p/src/peer/outbound.ts
mitsujutsu: I will change it to empty string as it's just a placeholder string (there is no `pong` event)

shuse2: I think it would be better to return pong for convention rather than returning empty string, or maybe follow socket cluster protocol?

elements/lisk-p2p/src/peer_pool.ts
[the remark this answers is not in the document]
shuse2: Can we add them...?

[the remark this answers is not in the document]
mitsujutsu: `Peer` is a class type so maybe we would have to create a new interface just to add index signature for it. I can do so if it better than using `any`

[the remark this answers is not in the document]
shuse2: yes definitely better than `any` =p

elements/lisk-p2p/src/p2p.ts
```diff
@@ -390,6 +394,18 @@ export class P2P extends EventEmitter {
 			outboundShuffleInterval: config.outboundShuffleInterval
 				? config.outboundShuffleInterval
 				: DEFAULT_OUTBOUND_SHUFFLE_INTERVAL,
+			latencyProtectionRatio:
+				typeof config.latencyProtectionRatio === 'number'
```
diego-G: Do we really need to check this? I think the typing should enforce it by itself.

elements/lisk-p2p/src/p2p.ts
```diff
@@ -178,6 +180,7 @@ export class P2P extends EventEmitter {
 	private readonly _handleInboundSocketError: (error: Error) => void;
 	private readonly _peerHandshakeCheck: P2PCheckPeerCompatibility;
 
+	// tslint:disable-next-line: cyclomatic-complexity
```
diego-G: I would avoid disabling the cyclomatic-complexitiy. This should be our boundary to find a simpler way or breaking the function into pieces.

elements/lisk-p2p/test/integration/p2p.ts
```diff
@@ -1763,4 +1765,83 @@ describe('Integration tests for P2P library', () => {
...
+		afterEach(async () => {
+			await Promise.all(
+				p2pNodeList
+					.filter(p2p => p2p.isActive)
+					.map(async p2p => await p2p.stop()),
+			);
+			await wait(100);
+		});
+
+		describe('Inbound peer evictions', () => {
```
diego-G: I would strongly suggest to add at least two cases more to cover the three different variables.

elements/lisk-p2p/src/p2p.ts
mitsujutsu: I agree however my additions to this file in this PR are minimal, and cyclomatic complexity only popped up after merging feature branch. I would like to suggest we tackle this in a separate PR.

diego-G: Agree. Please, open an issue/task to not to forget and link it to this comment.

elements/lisk-p2p/test/integration/p2p.ts
mitsujutsu: It's hard to test for `latency` in a predictable manner without getting varying results in this integration test.. Also the same for `responseRate`. Since it is integration test I tried to only manipulate the p2p configurations and node start times, without calling internal functions directly to test the behavior. The only metric that is easily tested for in this case is `connectTime`.

diego-G: What do you think about at least creating a test skeleton that you're able to think of now that it's a fresh topic? Then we can cover later. Otherwise we will forget.

elements/lisk-p2p/src/p2p.ts
mitsujutsu: It's necessary because say a user enters `0`. In this case `latencyProtectionRatio` is evaluated to false which would default to `0.068`. Instead if users enters `0` I want to disable latency protection.

elements/lisk-p2p/src/p2p.ts
```diff
@@ -390,6 +394,18 @@ export class P2P extends EventEmitter {
...
 				? config.outboundShuffleInterval
 				: DEFAULT_OUTBOUND_SHUFFLE_INTERVAL,
+			latencyProtectionRatio:
+				typeof config.latencyProtectionRatio === 'number'
+					? config.latencyProtectionRatio
+					: DEFAULT_PEER_PROTECTION_FOR_LATENCY_AND_USEFULNESS,
+			productivityProtectionRatio:
+				typeof config.productivityProtectionRatio === 'number'
+					? config.productivityProtectionRatio
+					: DEFAULT_PEER_PROTECTION_FOR_LATENCY_AND_USEFULNESS,
```
jondubois: Can we separate `DEFAULT_PEER_PROTECTION_FOR_LATENCY_AND_USEFULNESS` into two constants?

elements/lisk-p2p/src/peer/inbound.ts
```diff
@@ -46,12 +56,25 @@ export class InboundPeer extends Peer {
...
+			}
 			this.emit(EVENT_CLOSE_INBOUND, {
 				peerInfo,
 				code,
 				reason,
 			});
 		};
+		this._pingStart = Date.now();
+		this._sendPing = () => {
+			clearTimeout(this._pingTimeoutId);
```
jondubois: The `clearTimeout(this._pingTimeoutId)` line is not needed here because the `_sendPing` function only executes when the timeout has been triggered; when the last timeout triggers, it is automatically cleared.

We only need to `clearTimeout` on socket disconnect.

elements/lisk-p2p/src/peer/inbound.ts
```diff
@@ -46,12 +56,25 @@ export class InboundPeer extends Peer {
...
+			if (this._pingTimeoutId) {
+				clearTimeout(this._pingTimeoutId);
+			}
 			this.emit(EVENT_CLOSE_INBOUND, {
 				peerInfo,
 				code,
 				reason,
 			});
 		};
+		this._pingStart = Date.now();
```
jondubois: This `_pingStart ` is not needed here. It should be defined as a scoped `const pingStart` inside the `_sendPing` function (no `this` needed); keeping it only inside the scope of the `_sendPing` function would guarantee that it cannot be mixed up with a different ping. It's possible that our node could be waiting for the response from multiple parallel pings at the same time so the calculation for each ping/pong must have its own independent scope.

PR #3833, opened 2019-06-17
Create ChainMeta entity to persist BFT information - Closes #3832

### LINKED ISSUE #3832: Create ChainMeta entity to persist BFT information
### Expected behavior
We. need some `persistence` logic and store to load up the BFT state while reloading the application. For that we need to utilise the current persistence layer of PostgreSQL. So create an entity `ChainMeta` which will store *any meta information* as *key value* store. This will give a flexible and extensible way to store BFT information for now and other fields if required in the future. 

This should help us to persist information in particular way: 

```js
await ChainMeta.upsert({key: 'BFT_finalizedHeight', value: 56});

await ChainMeta.get({key: 'BFT_finalizedHeight'});
```

3.0.0

nazarhussain:
### What was the problem?
There were no persistence location to store meta information, for now for BFT. But for future any other information about the chain. 

### How did I fix it?

Created a table to behave like a key/value store to persist any meta information. 

### How to test it?

```
npm run mocha:unit test/mocha/unit/modules/chain/components/storage/entities/chain_meta.js
```

### Review checklist

* The PR resolves #3832
* Linting passes
* Tests pass


framework/src/modules/chain/components/storage/entities/chain_meta.js
```diff
@@ -0,0 +1,196 @@
...
+	/**
+	 * Update or create the key value pair
+	 *
+	 * @param {Object} data
+	 * @param {string} data.key
+	 * @param {string} data.value
+	 * @param {Object} [tx] - Database transaction object
+	 * @return {Promise.<null, Error>}
+	 */
+	upsert({ key, value }, tx) {
```
lsilvs: Same as above

framework/src/modules/chain/components/storage/entities/chain_meta.js
```diff
@@ -0,0 +1,196 @@
...
+		return this.adapter.executeFile(
+			this.SQLs.get,
+			params,
+			{ expectedResultCount },
+			tx
+		);
+	}
+}
+
+module.exports = ChainMeta;
```
lsilvs: Don't we need a `delete` method?

michielmulders:
@nazarhussain Would it be possible to store the temp blocks and transactions for fast chain switching as a value inside this metachain? Haven't checked the limitations of the storage capacity for one key-value pair.

framework/src/modules/chain/components/storage/entities/chain_meta.js
```diff
@@ -0,0 +1,196 @@
...
+ * @property {string} key
+ * @property {string} value
+ */
+
+/**
+ * ChainMeta Filters
+ * @typedef {Object} filters.ChainMeta
+ * @property {string} [key]
+ * @property {string} [key_eql]
+ * @property {string} [key_ne]
```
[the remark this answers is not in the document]
nazarhussain: The `key` have a filter type boolean, so it only have `eql` and `ne` filters. That's done intentionally to use it as a key store, where you specify full key. 

```
this.addField('key', 'string', { filter: BOOLEAN });
```

Will see if in future we need to provide matching on the key, then we can just change the filter type.

nazarhussain:
@michielmulders Yes its possible but suggest to use temp file for that purpose if required.

framework/src/modules/chain/components/storage/entities/chain_meta.js
```diff
@@ -0,0 +1,196 @@
...
+	/**
+	 * Update the key value pair
+	 *
+	 * @param {Object} data
+	 * @param {string} data.key
+	 * @param {string} data.value
+	 * @param {Object} [tx] - Database transaction object
+	 * @return {Promise.<null, Error>}
+	 */
+	update({ key, value }, tx) {
```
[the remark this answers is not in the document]
nazarhussain: This entity is for particular use so need interfaces with particular definition. I removed the `update` and `create` from here. See the new implementation.

framework/src/modules/chain/components/storage/entities/chain_meta.js
```diff
@@ -0,0 +1,196 @@
...
+	/**
+	 * Create the key value pair
+	 *
+	 * @param {Object} data
+	 * @param {string} data.key
+	 * @param {string} data.value
+	 * @param {Object} [tx] - Database transaction object
+	 * @return {Promise.<null, Error>}
+	 */
+	create({ key, value }, tx) {
```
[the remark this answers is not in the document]
nazarhussain: Please see the new implementation;

framework/src/modules/chain/components/storage/entities/chain_meta.js
nazarhussain: Added

nazarhussain:
@LucasIsasmendi This entity is designed for particular use so need interfaces with particular definition. The interfaces which it inherits from the base entity have same parameter definition. Rest are complete specific to this entity usage, so have different parameter definition. 

Please review the PR again.

nazarhussain:
@LucasIsasmendi The tests are green. Please review.

framework/src/modules/chain/components/storage/entities/chain_meta.js
```diff
@@ -0,0 +1,183 @@
...
+	 * Fetch the key value or resolve to null
+	 *
+	 * @param {string} key - Key to fetch
+	 * @param {Object} [tx] - Database transaction object
+	 * @return {Promise<{key: string, value: string} | {}>}
+	 */
+	async getKey(key, tx) {
+		assert(key, 'Must provide the key to get');
+
+		return this.getOne({ key }, {}, tx)
```
limiaspasdaniel: Can we use `await` here?

nazarhussain: The main function itself is a `async` and it must be used with `await` where ever called. So retuning promise is fine to me.

framework/src/modules/chain/chain.js
```diff
@@ -45,8 +45,11 @@ const { Transport } = require('./transport');
 const syncInterval = 10000;
 const forgeInterval = 1000;
 
-// Begin reading from stdin
-process.stdin.resume();
+// TODO: https://github.com/LiskHQ/lisk-sdk/issues/3839
+// Don't read from stdin in test environment
+if (process.env.NODE_ENV !== 'test') {
```
shuse2: ❤️

framework/src/modules/chain/migrations/sql/20190617105815_create_chain_meta_table.sql
```diff
@@ -0,0 +1,25 @@
...
+
+
+/*
+  DESCRIPTION: Creates meta table for BFT.
+
+  PARAMETERS: None
+*/
+
+CREATE TABLE IF NOT EXISTS "chain_meta" (
+  "key" VARCHAR(40) UNIQUE NOT NULL PRIMARY KEY,
```
shuse2: Primary key is unique and not null ?

framework/src/modules/chain/migrations/sql/20190617105815_create_chain_meta_table.sql
```diff
@@ -0,0 +1,25 @@
...
+  "value" TEXT NOT NULL
```
shuse2: why not JSON field here?

framework/test/jest/integration/specs/modules/chain/components/storage/entities/chain_meta.spec.js
```diff
@@ -0,0 +1,248 @@
...
+const getAllMeta = async adapter =>
+	adapter.execute('select * from chain_meta;');
+
+describe('ChainMeta', () => {
+	const storage = new StorageSandbox(
+		config.components.storage,
+		'lisk_test_chain_module_storage_chain_meta'
+	);
+	let adapter;
+	let ChainMetaEntity;
```
shuse2: Prefer consistent ordering of `const` and `let`

framework/src/modules/chain/components/storage/entities/chain_meta.js
```diff
@@ -0,0 +1,183 @@
...
+			.then(data => data.value)
+			.catch(error => {
+				if (error.message === 'No data returned from the query.') {
```
shuse2: No better way to check this? ie) check with error type etc
and do we really want to catch this error?

framework/src/modules/chain/components/storage/entities/chain_meta.js
```diff
@@ -0,0 +1,183 @@
...
+
+	/**
+	 * Delete the keys with following conditions
+	 *
+	 * @param {filters.ChainMeta} filters
+	 * @param {Object} [options]
+	 * @param {Object} [tx]
+	 * @returns {Promise.<boolean, Error>}
+	 */
+	delete(filters, _options, tx = null) {
```
shuse2: no `deleteKey`?

framework/test/jest/integration/specs/modules/chain/components/storage/entities/chain_meta.spec.js
```diff
@@ -0,0 +1,248 @@
...
+			await ChainMetaEntity.delete({ key: data1.key });
+
+			expect(await getAllMeta(ChainMetaEntity.adapter)).toEqual([data2]);
+		});
+
+		it('should not throw error if no matching record found', async () => {
+			const nonExistingKey = 'nonExistingKey';
+
+			await expect(
+				ChainMetaEntity.delete({ key: nonExistingKey })
```
shuse2: why not await inside here?

framework/src/modules/chain/migrations/sql/20190617105815_create_chain_meta_table.sql
nazarhussain: Its upto usage, if you want to store serialized json or just primitive values. We don't provide feature to search in the value, so no need to store it as JSON.

framework/test/jest/integration/specs/modules/chain/components/storage/entities/chain_meta.spec.js
nazarhussain: Didn't get it.

framework/src/modules/chain/components/storage/entities/chain_meta.js
nazarhussain: Yes we want to return null in that case so this check is needed. Othercase is to get list of records and based on length decide the return value.

framework/src/modules/chain/components/storage/entities/chain_meta.js
nazarhussain: `get`, `getOne`, `delete` are consistent interfaces from BaseEntity, so used the same.

framework/test/jest/integration/specs/modules/chain/components/storage/entities/chain_meta.spec.js
shuse2: Sorry 😉 Instead of

```
const x = '';
let y = '';
const z = 3;
```

prefer below
```
const x = '';
const z = 3;
let y = '';
```

framework/src/modules/chain/components/storage/entities/chain_meta.js
shuse2: but getOne is throwing error, but getKey is not throwing when no key found?
Both are pretty much same feature, but seems inconsistent

framework/test/jest/integration/specs/modules/chain/components/storage/entities/chain_meta.spec.js
nazarhussain: We want to check if returned promise resolved to null, that's one way to do it in Jest. Other case is to resolve to value first and match later.

framework/src/modules/chain/migrations/sql/20190617105815_create_chain_meta_table.sql
nazarhussain: 👍 

For reference https://www.postgresql.org/docs/9.4/ddl-constraints.html

framework/src/modules/chain/components/storage/entities/chain_meta.js
nazarhussain: We need different behaviour for convenient use so have different interfaces.

PR #3835, opened 2019-06-17
Update discovery according to LIP protocol - Closes #3649

### LINKED ISSUE #3649: Update discovery according to LIP protocol
### Expected behavior

Update discovery to the new LIP protocol.

### Actual behavior

Current approach tries to discover and connect to all peers in the network.

jondubois:
### What was the problem?

Discovery was not consistent with P2P LIP.

### How did I fix it?

Made discovery consistent with LIP so that it picks up peers as it connects to them and updated tests.

### How to test it?

`npm run test`

### Review checklist

* The PR resolves #3649
* Linting passes
* Tests pass


elements/lisk-p2p/src/peer/base.ts
```diff
@@ -405,15 +401,24 @@ export class Peer extends EventEmitter {
 		}
 	}
 
-	public async fetchStatus(): Promise<P2PDiscoveredPeerInfo> {
+	public async discoverPeers(): Promise<ReadonlyArray<P2PPeerInfo>> {
+		const discoveredPeerInfoList = await this.fetchPeers();
+		discoveredPeerInfoList.forEach(peerInfo => {
```
shuse2: isn't this firing maximum 1000 events? would that be ok?

elements/lisk-p2p/src/p2p.ts
```diff
@@ -785,30 +693,30 @@ export class P2P extends EventEmitter {
...
+			if (!this._newPeers.has(peerId)) {
+				this._newPeers.set(peerId, seedInfo);
 			}
 		});
 
-		await this._startDiscovery();
-		this._startPopulator();
+		await this._startPeerServer();
+
+		if (this._isActive) {
```
shuse2: this if seems strange.
It's throwing in the above if it's active, and it's checking _isActive

elements/lisk-p2p/src/peer/outbound.ts
```diff
@@ -168,7 +146,11 @@ export class OutboundPeer extends Peer {
 
 		outboundSocket.on('connect', async () => {
 			this.emit(EVENT_CONNECT_OUTBOUND, this._peerInfo);
-			await this._updatePeerOnConnect();
+			try {
```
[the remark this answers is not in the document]
jondubois: The `EVENT_FAILED_TO_COLLECT_PEER_DETAILS_ON_CONNECT` event will be emitted in addition to events that are already emitted at the peer level. That said, I noticed that we are not currently emitting an event on the peer for the `discoverPeers -> fetchPeers` call so I added a new `EVENT_FAILED_TO_FETCH_PEERS`.

elements/lisk-p2p/src/p2p.ts
jondubois: Ah yep it does look strange, I will add a comment. It's needed because there is an edge case which can cause `_isActive` to be false if the P2P library is stopped before it fully started (while in the middle of _startPeerServer) because promises resolve on the next tick. There could be a millisecond gap between when `this._isActive = true` is set inside `_startPeerServer` and when this line is hit and it's possible that `_isActive` could have been set to `false` during this time.

elements/lisk-p2p/src/p2p.ts
```diff
@@ -291,38 +309,23 @@ export class P2P extends EventEmitter {
 		};
 
 		// When peer is fetched for status after connection then update the peerinfo in triedPeer list
-		this._handleDiscoveredPeer = (detailedPeerInfo: P2PDiscoveredPeerInfo) => {
+		this._handleDiscoveredPeer = (detailedPeerInfo: P2PPeerInfo) => {
 			const peerId = constructPeerIdFromPeerInfo(detailedPeerInfo);
-			const foundTriedPeer = this._triedPeers.get(peerId);
 			// Remove the discovered peer from newPeer list on successful connect and discovery
```
ishantiw: We can remove this comment or update it

elements/lisk-p2p/src/p2p.ts
```diff
@@ -785,30 +708,31 @@ export class P2P extends EventEmitter {
 		if (this._isActive) {
 			throw new Error('Cannot start the node because it is already active');
 		}
-		await this._startPeerServer();
 
 		// Fetch status of all the seed peers and then start the discovery
```
ishantiw: This comment is no longer relevant

elements/lisk-p2p/src/p2p.ts
```diff
@@ -785,30 +708,31 @@ export class P2P extends EventEmitter {
...
-		const seedPeerInfos = await this._fetchSeedPeerStatus(
-			this._config.seedPeers,
-		);
+		const seedPeerInfos = this._config.seedPeers;
 
 		// Add seed's peerinfos in tried peer as we already tried them to fetch status
```
ishantiw: same here

PR #3840, opened 2019-06-19
Remove bus from chain module - Closes #3813

### LINKED ISSUE #3813: Remove bus from chain module
### Expected behavior
 Since we introduced the new architecture supporting the inter-module communication through the channels, we should use them to facilitate the interactions within the chain modules itself.  

### Actual behavior
To communicate between modules we use the bus (as a helper).

### Steps to reproduce
- the bus helper is removed
- the bus usages are replaced by channel.invoke calls
`*`

lsilvs:
### What was the problem?
Chain using a custom implementation called ` Bus` to communicate via "events" where it should be using modules Channel.

### How did I fix it?
I've removed the custom implementation and replaced it by `channel.publish` where it was used.

### How to test it?
Run test suit.

### Review checklist

* The PR resolves #3813
* Linting passes
* Tests pass


framework/src/modules/chain/rounds/rounds.js
```diff
@@ -214,7 +213,7 @@ class Rounds {
 				.then(() => {
 					if (scope.finishRound) {
 						return promised.land().then(() => {
-							library.bus.message('finishRound', round);
+							library.channel.publish('chain:rounds:change', { number: round });
```
limiaspasdaniel: Maybe this can be called `chain:rounds:finished` or `finish`?

lsilvs: Changing the event key is out of scope of this PR. Please open an issue for that if you feel it's important.

PR #3842, opened 2019-06-20
Update sequence in chain module - Closes #3753

### LINKED ISSUE #3753: Update sequence usage to be async
### Expected behavior
Remove balance sequence.
Update sequence class to be async/await format

2.2-

shuse2:
### What was the problem?
Sequence was still using the callback style.
Also, balance sequence was not required after removal of unconfirmed state

### How did I fix it?
- Remove balance sequence
- Update sequence to only accept async function
- Add unit test for sequence

### How to test it?
- `npm jest:unit`
- run application

### Review checklist

* The PR resolves #3753 
* Linting passes
* Tests pass


framework/src/modules/chain/utils/sequence.js
```diff
@@ -28,81 +35,76 @@ const util = require('util');
...
-					setImmediate(task.done, err, res);
-				}
-				setImmediate(cb);
-			},
-		];
-		if (task.args) {
-			args = args.concat(task.args);
+		try {
+			const result = await task.worker();
+			task.done.resolve(result);
```
lsilvs: 😮

framework/test/mocha/integration/blocks/process/on_receive_block.js
```diff
@@ -218,22 +218,15 @@ describe('integration test (blocks) - process receiveBlockFromNetwork()', () =>
...
-		);
+		library.sequence
+			.add(async () => {
+				const rows = storage.adapter.db.query(
+					new PQ('SELECT "id" FROM blocks ORDER BY "height" DESC LIMIT 10;')
+				);
+				return rows.map(r => r.id);
+			})
+			.then(ids => cb(null, ids))
+			.catch(err => __testContext.debug(err.stack));
```
SargeKhan: not calling callback here

framework/test/mocha/integration/blocks/process/on_receive_block.js
```diff
@@ -458,13 +451,17 @@ describe('integration test (blocks) - process receiveBlockFromNetwork()', () =>
...
 						blockFromPreviousRound.height = mutatedHeight;
-						return library.modules.blocks.receiveBlockFromNetwork(
-							blockFromPreviousRound
-						);
+						try {
+							await library.modules.blocks.receiveBlockFromNetwork(
+								blockFromPreviousRound
+							);
+						} catch (err) {
+							// expected error
```
SargeKhan: I don't understand why  you made this code change

framework/src/modules/chain/utils/sequence.js
```diff
@@ -28,81 +35,76 @@ const util = require('util');
...
+	add(worker) {
+		if (!util.types.isAsyncFunction(worker)) {
+			throw new Error('Worker must be an async function.');
 		}
+		let done;
+		const workerPromise = new Promise((resolve, reject) => {
+			done = { resolve, reject };
+		});
+		const task = { worker, done };
+		this.queue.push(task);
```
SargeKhan: We can simply wrap the promise and remove the need to use closures. Like,
```
const workerPromise = new Promise((resolve, reject) => {
 try {
  resolve(await worker);
 } catch (e) {
  reject(e);
 }
});
this.queue.push(workerPromise);
return workerPromise;
```

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -332,20 +327,18 @@ class Blocks extends EventEmitter {
 					await this._updateBroadhash();
 					this._lastBlock = newBlock;
 					this._isActive = false;
-					setImmediate(cb);
+					return;
 				} catch (error) {
 					this._isActive = false;
 					this.logger.error(error);
-					setImmediate(cb, error);
+					throw error;
```
SargeKhan: In case of the old implementation, the error in sequence is passed to the done parameter passed to the sequence.add, whereas now the error will bubble up in the call stack. Maybe we don't want that? I think we should replicate the functionality of sequences 100%

framework/src/modules/chain/chain.js
```diff
@@ -436,21 +429,13 @@ module.exports = class Chain {
...
-							if (syncError) {
-								this.logger.error('Sync timer', syncError);
-								return cb(syncError);
-							}
-							return cb();
+					this.scope.sequence.add(async () => {
+						try {
+							await this.loader.sync();
+						} catch (error) {
+							this.logger.error(error, 'Sync timer');
```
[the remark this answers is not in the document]
shuse2: In the new logger, object should be first, then it logs better =)

framework/src/modules/chain/utils/sequence.js
shuse2: When the tick is called, we want to make sure that it doesn't throw, but it resolve/reject the workerPromise.
If we do this way, I wasn't sure how to do that

framework/src/modules/chain/chain.js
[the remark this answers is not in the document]
SargeKhan: That's weird. No?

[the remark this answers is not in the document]
shuse2: `bunyan` supports this idea and seems to be ok

framework/src/modules/chain/chain.js
```diff
@@ -428,21 +421,13 @@ module.exports = class Chain {
...
-								.then(() => sequenceCB())
-								.catch(err => sequenceCB(err));
-						},
-						syncError => {
-							if (syncError) {
-								this.logger.error('Sync timer', syncError);
-								return cb(syncError);
-							}
-							return cb();
+					await this.scope.sequence.add(async () => {
```
SargeKhan: i think we also need to return await from here

shuse2: In the jobQueue, the result of the promise is not use and it just need to wait for the execution.
Why do you think it needs to return promise here?

SargeKhan: because multiple nextSync jobs may be executed if we don't return promise from here. Come to think of it, it will not cause any problems because the preceding `if` statement will return false if syncing process is executing.

framework/src/modules/chain/transport/transport.js
```diff
@@ -652,27 +648,21 @@ class Transport {
...
+				true
+			);
+			return transaction.id;
+		} catch (err) {
+			this.logger.debug(`Transaction ${id}`, convertErrorsToString(err));
+			if (transaction) {
+				this.logger.debug('Transaction', transaction);
 			}
-		});
+			throw err;
```
shuse2: @SargeKhan throwing error was correct behavior because balance sequence was promisified

PR #3846, opened 2019-06-21
 Synchronizer design proposal - Closes #3773

### LINKED ISSUE #3773: Block Synchronizer class to combine Block Synchronizaton Mechanism logic
### Expected behavior
Create class will be stored under `framework/chain/blocks_synchronizer.js`,
and define interface for the class

`feature/BFT`

limiaspasdaniel:
Draft PR to introduce a logic blueprint that will be responsible for implementing the two synchronization mechanisms introduced in BFT LIP-0014 starting [here](https://rubygems.pkg.github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#moving-to-a-different-chain)

The idea would be to expose only the Synchronizer interface and let it decide when to use Block Synchronization Mechanism or Fast Chain Switching as defined in the link above.

Entities in the application that require checking if the application is syncing would only need to call the `isActive` method of the `Synchronizer` class. 

image

The folder structure could be as the following:
```
src
│   ├── modules
│   │   ├── chain
│   │   │   ├── blocks
│   │   │   │   ├── synchronizer
│   │   │   │   │   ├── block_synchronization_mechanism.js
│   │   │   │   │   ├── fast_chain_switching_mechanism.js
│   │   │   │   │   ├── index.js
│   │   │   │   │   └── synchronizer.js
```

Using the Synchronizer would be as simple as doing: 
```
const synchronizer = new Synchronizer(dependencies)
synchronizer.run()
```

Note: possibly we could modify the scope of #3773 to include this.

* The PR resolves #3773 
* Linting passes
* Tests pass


michielmulders:
I like this idea as it allows for interfaces to be defined in `Synchronizer` and can be overwritten by the BSM or FCS, and even share some of the logic. The only thing that pops up in my mind now is where we will be instantiating the synchronizer class? In `Chain`?

limiaspasdaniel:
@michielmulders 
> I like this idea as it allows for interfaces to be defined in `Synchronizer` and can be overwritten by the BSM or FCS, and even share some of the logic.

`BlockSynchronizationMechanism` and `FastChainSwitchingMechanism` classes are not meant to inherit `Synchronizer` and overwrite any of its methods. My idea for `Synchronizer` would be to be some sort of orchestrator that only decides which syncing method to use and provides an easy-to-use interface. This way if the logic in any of both methods changes we don't need to change its usages.

For logic sharing, both mechanisms mainly use logic present in other submodules especially in `blocks.js`

> The only thing that pops up in my mind now is where we will be instantiating the synchronizer class? In `Chain`?

Yes, it could be initialized in `chain.js`

pablitovicente:
> @michielmulders
> 
> > I like this idea as it allows for interfaces to be defined in `Synchronizer` and can be overwritten by the BSM or FCS, and even share some of the logic.
> 
> `BlockSynchronizationMechanism` and `FastChainSwitchingMechanism` classes are not meant to inherit `Synchronizer` and overwrite any of its methods. My idea for `Synchronizer` would be to be some sort of orchestrator that only decides which syncing method to use and provides an easy-to-use interface. This way if the logic in any of both methods changes we don't need to change its usages.
> 
> For logic sharing, both mechanisms mainly use logic present in other submodules especially in `blocks.js`
> 
> > The only thing that pops up in my mind now is where we will be instantiating the synchronizer class? In `Chain`?
> 
> Yes, it could be initialized in `chain.js`

Check the code I shared a few weeks ago does something very similar to this with the ability to switch on the strategies on the fly.

limiaspasdaniel:
@pablitovicente what do you mean by `on the fly`?

pablitovicente:
> @pablitovicente what do you mean by `on the fly`?

This case might be different but my small prototype was for switching block processing strategies and it allowed to define rules for when to switch the module processing the data. Will upload to gist later!

shuse2:
@limiaspasdaniel Since `Synchronizer` will be using blocks, dpos etc, I think it should be out of `blocks` folder. (more or less replacement of loader.js`

nazarhussain:
Yes the ideology is nice to split the logic `BlockSynchronizationMechanism` and `FastChainSwitchingMechanism` completely, but in generalized way. This will not help us just for now, also helpful to introduce another synchronisation method as well. 

We need to look into these points; 

1. There is slight difference in `strategy` and `proxy` pattern. 
2. In strategy pattern, `strategy` is just an interface which can't be initiated but can be extended to implement real strategies. Based on some context you initiate that particular strategy and assign to that context once. And later the context use that particular strategy always without knowing which strategy it is, for context its always base strategy interface. 
3. In proxy pattern, you have one interface, when invoked it make decision based on business logic, to which other object it forwards the call. 
4. Looking at the code below, seems its more of proxy pattern rather strategy. 
```js
// Initialisation of both logics 
this.currentStrategy = null;
this.blockSynchronizationMechanism = new BlockSynchronizationMechanism();
this.fastChainSwitchingMechanism = new FastChainSwitchingMechanism();

// Runtime decided on each call
this.currentStrategy = this._determineStrategy();
return this.currentStrategy.run();
```
5. If it would be strategy would look something similar to: 
```js
// Context Initialisation
this.strategy = this._determineStrategy();

// Runtime no decision on strategy just call one interface 
return this.strategy.run();
```
5. Above is just the thoughts on pattern naming convention, the idea is good to go. You may look into `Proxy`, `Mediator` or `Chain of Responsibility` patterns. These are more aligned with this concept and to me `Proxy` is more accurate in our case. 
6. We are using ES6 classes a lot already, unless have particular reason we should not avoid it. So extending class to share code is also not a problem.

limiaspasdaniel:
@nazarhussain thanks for your input :) 

4. Agreed, this solution is closer to a Proxy pattern rather than the Strategy one.

However, I don't see the conclusion to your comment. Perhaps the `strategy` naming in the code isn't the most accurate to what the pattern really is, but I can't find a better name for it. Perhaps `currentMechanism` could also fit there. Please, give your input on that.

As you pointed, in this case, the mechanism to follow is decided on each call, as it depends entirely on the input, making it closer to a Proxy pattern. I assume `run()` will need to accept data such as height, heightPrevoted, etc to be able to do it.

Apart from that, regarding:

> @limiaspasdaniel Since `Synchronizer` will be using blocks, dpos etc, I think it should be out of `blocks` folder. (more or less replacement of loader.js

@shuse2 Good point, I will move it to the upper scope.

nazarhussain:
@limiaspasdaniel The conclusion is that its good to :) 

> Above is just the thoughts on pattern naming convention, the idea is good to go.

limiaspasdaniel:
@shuse2 moved the files to the upper scope

framework/src/modules/chain/synchronizer/synchronizer.js
```diff
@@ -0,0 +1,44 @@
+const BlockSynchronizationMechanism = require('./block_synchronization_mechanism');
+const FastChainSwitchingMechanism = require('./fast_chain_switching_mechanism');
+
+class Synchronizer {
+	constructor() {
+		this.currentStrategy = null;
+		this.blockSynchronizationMechanism = new BlockSynchronizationMechanism();
+		this.fastChainSwitchingMechanism = new FastChainSwitchingMechanism();
```
yatki: I think, there is no need to initialise both classes in the constructor? Maybe we can do it in  `_determineStrategy` ?

limiaspasdaniel: My idea was to not to have to initialize the strategy every time `run()` is called, which can be potentially (but hopefully _NOT_) called multiple times

yatki: If we execute `run` in synchronizer, wouldn't that mean to create mechanism instances with a clear state? I think you are planning to clear the state of the mechanism instances and reuse them. But if the instances are assigned to a variable and not used, gc won't be able to free the memory.

I don't know what you are planning to do there. Just wanted to remind that. I'd create the instance and dispose it in the `run`.

limiaspasdaniel: That's a good point @yatki, thanks for the input. Unfortunately, at this phase, I am unsure of which state will the instances hold and if it would be needed to be cleared on each run.

I'll keep your point in mind :D

limiaspasdaniel:
@shuse2 we can merge this PR to have the logic skeleton ready for further development and also close its related issue.

PR #3851, opened 2019-06-24
Create submodule/bft to initiate the BFT logic at application startup - Closes #3693

### LINKED ISSUE #3693: Create submodule/bft to initiate the BFT logic at application startup
### Expected behavior

Create submodule/bft to initiate the BFT logic at application startup. That requires loading up minimum block headers from the database to perform BFT block verification. 

3.0.0

nazarhussain:
### Review checklist

* The PR resolves #3693
* Linting passes
* Tests pass


framework/src/modules/chain/bft/index.js
```diff
@@ -18,10 +18,7 @@
  * @namespace BFT
  */
 
-const HeadersList = require('./headers_list');
-const BFT = require('./bft');
-
 module.exports = {
-	...HeadersList,
-	...BFT,
+	...require('./headers_list'),
```
shuse2: prefer no dynamic require.

nazarhussain:
@limiaspasdaniel @shuse2 @SargeKhan Please review this PR. Git somehow didn't tracked the changes well here. The old `bft.js` file is moved as `consensus_manager.js` without noticeable change. And current `bft.js` have complete new code for integration.

framework/src/modules/chain/bft/consensus_manager.js
```diff
@@ -0,0 +1,297 @@
...
+		// Set constants
+		this.activeDelegates = activeDelegates;
+
+		// Threshold to consider a block pre-voted
+		this.preVoteThreshold = Math.ceil((this.activeDelegates * 2) / 3);
+
+		// Threshold to consider a block pre-committed (or finalized)
+		this.preCommitThreshold = Math.ceil((this.activeDelegates * 2) / 3);
+
+		// Limit to of blocks to must have to make any prevote-precommit verification on blocks (1 block less than 3 rounds)
```
limiaspasdaniel: Maybe `limit of blocks`?

framework/src/modules/chain/bft/consensus_manager.js
```diff
@@ -0,0 +1,297 @@
...
+		this.processingThreshold = this.activeDelegates * 3 - 1;
+
+		// Maximum headers to store (5 rounds)
+		this.maxHeaders = this.activeDelegates * 5;
+
+		this.headers = new HeadersList({ size: this.maxHeaders });
+
+		// Height up to which blocks are finalized
+		this.finalizedHeight = finalizedHeight || 0;
```
limiaspasdaniel: Maybe we could set this in the constructor `{ finalizedHeight = 0, activeDelegates } = {}`

framework/src/modules/chain/bft/consensus_manager.js
```diff
@@ -0,0 +1,297 @@
...
+	}
+
+	/**
+	 * Add block header to BlockHeaderManager
+	 *
+	 * @param {BlockHeader} blockHeader
+	 * @return {Block_headers_manager}
+	 */
+	addBlockHeader(blockHeader) {
+		debug('addBlockHeader invoked');
```
limiaspasdaniel: Are we still using `require('debug')`?

framework/src/modules/chain/bft/consensus_manager.js
```diff
@@ -0,0 +1,297 @@
...
+			maxPreCommitHeight: 0,
+		};
+
+		// Get first block of the round when delegate was active
+		const delegateMinHeightActive =
+			(header.activeSinceRound - 1) * this.activeDelegates + 1;
+
+		// If delegate is new then first block of the round will be considered
+		// if it forged before then we probably have the last commit height
+		// delegate can't pre-commit and block before above mentioned conditions
```
limiaspasdaniel:
```suggestion
		// delegate can't pre-commit a block before the above mentioned conditions
```

framework/src/modules/chain/bft/consensus_manager.js
```diff
@@ -0,0 +1,297 @@
...
+	/**
+	 * Update the pre-voted confirmed and finalized height
+	 */
+	updatePreVotedAndFinalizedHeight() {
+		debug('updatePreVotedAndFinalizedHeight invoked');
+		if (this.headers.length === 0) {
+			return;
+		}
+
+		const higherPairVoted = Object.entries(this.preVotes)
```
limiaspasdaniel: `Object.entries` is the slowest of the methods to iterate over object keys. I suggest using `Object.keys` or `for..in`

framework/src/modules/chain/bft/consensus_manager.js
```diff
@@ -0,0 +1,297 @@
...
+			.reverse()
+			.find(pair => pair[1] >= this.preVoteThreshold);
+
+		this.prevotedConfirmedHeight = higherPairVoted
+			? parseInt(higherPairVoted[0])
+			: this.prevotedConfirmedHeight;
+
+		const higherPairCommitted = Object.entries(this.preCommits)
```
limiaspasdaniel: Same here

framework/src/modules/chain/bft/consensus_manager.js
```diff
@@ -0,0 +1,297 @@
...
+	 * Use existing block headers and re-compute all information
+	 */
+	recompute() {
+		this.state = {};
+		this.finalizedHeight = 0;
+		this.prevotedConfirmedHeight = 0;
+		this.preVotes = {};
+		this.preCommits = {};
+
+		this.headers.items.forEach(header => {
```
limiaspasdaniel: Avoid using `forEach` when possible. I suggest using `for..of`

framework/src/modules/chain/bft/errors.js
```diff
@@ -0,0 +1,55 @@
...
+'use strict';
+
+const { FrameworkError } = require('../../../errors');
+
+class BFTError extends FrameworkError {}
+
+class BFTChainDisjointError extends BFTError {
+	constructor() {
+		super(
+			'Violates disjointness condition. If delegate forged a block of higher height earlier and later the block with lower height'
```
limiaspasdaniel:
```suggestion
			'Violation of disjointness condition. If delegate forged a block of higher height earlier and later the block with lower height'
```

To keep wording consistency

framework/src/modules/chain/bft/errors.js
```diff
@@ -0,0 +1,55 @@
...
+		);
+	}
+}
+
+class BFTLowerChainBranchError extends BFTError {
+	constructor() {
+		super(
+			'Violates the condition that delegate must choose the branch with largest prevotedConfirmedUptoHeight'
```
limiaspasdaniel:
```suggestion
			'Violation of the condition that delegate must choose the branch with largest prevotedConfirmedUptoHeight'
```

framework/src/modules/chain/bft/index.js
```diff
@@ -18,10 +18,15 @@
...
+const { extractBFTBlockHeaderFromBlock, BFT } = require('./bft');
+const BFTErrors = require('./errors');
 
 module.exports = {
-	...HeadersList,
-	...BFT,
+	HeadersList,
+	ConsensusManager,
+	extractBFTBlockHeaderFromBlock,
+	BFT,
```
shuse2: do we really need to export everything?
Prefer to export only thing it should be used externally

framework/src/modules/chain/bft/errors.js
```diff
@@ -0,0 +1,55 @@
...
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+'use strict';
+
+const { FrameworkError } = require('../../../errors');
+
+class BFTError extends FrameworkError {}
```
shuse2: what is the point of extending `FrameworkError`? Prefer to just extend `Error`

framework/src/modules/chain/bft/bft.js
```diff
@@ -14,273 +14,136 @@
...
-			.find(pair => pair[1] >= this.preVoteThreshold);
-
-		this.prevotedConfirmedHeight = higherPairVoted
-			? parseInt(higherPairVoted[0])
-			: this.prevotedConfirmedHeight;
-
-		const higherPairCommitted = Object.entries(this.preCommits)
-			.reverse()
-			.find(pair => pair[1] >= this.preCommitThreshold);
+	async _initConsensusManager() {
```
shuse2: initialization returning finalized height seems weird to me.
Probably getting finalized height and intializing consensus manager should be separated

framework/src/modules/chain/bft/consensus_manager.js
```diff
@@ -0,0 +1,297 @@
...
+ * @typedef {Object} BlockHeader
+ * @property {string} blockId
+ * @property {int} height
+ * @property {int} maxHeightPreviouslyForged
+ * @property {int} prevotedConfirmedUptoHeight
+ * @property {int} activeSinceRound
+ * @property {string} delegatePublicKey
+ */
+
+class ConsensusManager {
```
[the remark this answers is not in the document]
shuse2: I also felt `ConsensusManager` is kind of misleading, too. Although it does reflect what it's doing.
Im not sure if ConsensusManager is good name to expose to the external either.

Maybe, instead of swapping, we can come up with better name for this class?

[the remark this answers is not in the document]
limiaspasdaniel: For me, it feels that as it is being managed by `BFT.js`, hence `Manager` isn't a name that reflects its purspose (?). 

> Maybe, instead of swapping, we can come up with better name for this class?
Agreed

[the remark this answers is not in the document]
nazarhussain: We discussed for this name convention for a long time and then used it. I originally named it as `FinalityManager` and then on request of @SargeKhan changed it to `ConsensusManager`. 

To me `FinalityManager` and `ConsensusManager` names both relates to the business logic it includes. `BFT` itself is very broad term which covers every thing so better suited for integration logic. This file only have particular scope to manage one particular aspect in the BFT.

framework/src/modules/chain/bft/consensus_manager.js
nazarhussain: Yes we do, why we wouldn't. 😃

framework/src/modules/chain/bft/consensus_manager.js
nazarhussain: Although this part of logic was reviewed and merged way log, not part of this PR. I will see if we can improve it for now or create separate issue.

framework/src/modules/chain/bft/consensus_manager.js
nazarhussain: Any particular reason for not using a default iterator function?

framework/src/modules/chain/bft/index.js
nazarhussain: ُPlease elaborate further for **External**?

framework/src/modules/chain/bft/errors.js
nazarhussain: We have a base error constructor and every error in framework should be extended by it. There is no harm in it and could be useful in future.

framework/src/modules/chain/bft/bft.js
nazarhussain: Yeah sure.

framework/src/modules/chain/bft/consensus_manager.js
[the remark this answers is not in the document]
limiaspasdaniel: In my opinion the name of `Byzantine Fault Tolerant(nce)` for an integration class it's not self-describing enough. For me it lacks a `verb` after it to expose what is its purpose, that is why I proposed adding the `Manager` or something similar word to the integration logic.

framework/src/modules/chain/bft/consensus_manager.js
limiaspasdaniel: Thanks, I realized it was old logic after seeing your comment

framework/src/modules/chain/bft/consensus_manager.js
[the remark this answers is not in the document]
nazarhussain: @limiaspasdaniel This logic applies to every module we have. `blocks`, `dpos`

framework/src/modules/chain/bft/consensus_manager.js
limiaspasdaniel: `forEach` is proven to be the slowest way to iterate over an array, so in my opinion not using it by default is a good exercise.

framework/src/modules/chain/bft/index.js
shuse2: external meaning outside of this folder

framework/src/modules/chain/bft/consensus_manager.js
nazarhussain: @limiaspasdaniel Created a separate issue to handle this and add more tests.

framework/src/modules/chain/bft/consensus_manager.js
nazarhussain: @limiaspasdaniel Created a separate issue to handle this and add more tests.

framework/src/modules/chain/bft/bft.js
```diff
@@ -14,273 +14,135 @@
...
-			delegateLastBlock.height >= blockHeader.height
-		) {
-			// Violation of the fork choice rule as delegate moved to different chain
-			// without strictly larger maxHeightPreviouslyForged or larger height as
-			// justification. This in particular happens, if a delegate is double forging.
-			throw new Error(
-				'Violation of fork choice rule, delegate moved to a different chain'
-			);
-		}
+	async loadBlocks({ fromHeight, tillHeight }) {
```
SargeKhan: Shouldn't this function also be a private function?

framework/src/modules/chain/bft/bft.js
```diff
@@ -14,273 +14,135 @@
...
+		const rows = await this.BlockEntity.get(
+			{ height_gte: fromHeight, height_lte: tillHeight },
+			{ limit: null, sort: 'height:asc' }
+		);
 
-		if (delegateLastBlock.height > blockHeader.maxHeightPreviouslyForged) {
-			throw new Error('Violates disjointness condition');
-		}
+		rows.forEach(row => {
+			if (row.version !== '2') return;
```
SargeKhan: For me, it makes more sense to throw an error here.

framework/src/modules/chain/bft/bft.js
nazarhussain: I kept it public if one need to load blocks to BFT, may be synchroniser or some other logic.

framework/src/modules/chain/bft/bft.js
nazarhussain: Its the edge case for migration. When we migrate on a particular height, we assume that height  - 202 as finalized height. While loading we have to load all blocks higher from finalized height. In that case we can encounter blocks with older version and we need to just ignore those blocks to make BFT ready for further processing.

framework/src/modules/chain/bft/bft.js
```diff
@@ -14,273 +14,135 @@
...
-			delegateState.maxPreCommitHeight + 1
+	async init() {
+		this.finalityManager = await this._initFinalityManager();
+		const finalizedHeight = this.finalityManager.finalizedHeight;
+		const lastBlockHeight = await this._getLastBlockHeight();
+
+		const loadFromHeight = Math.max(
+			finalizedHeight,
+			lastBlockHeight - this.constants.activeDelegates * 2,
+			this.constants.startingHeight
```
SargeKhan: What does this constant signify?

nazarhussain: Its the migration height for the network where we introduce the BFT. That probably will the height to introduce block version 2

SargeKhan:
And please review the tests, the jenkins is failing on your code related changes

framework/src/modules/chain/bft/finality_manager.js
```diff
@@ -0,0 +1,297 @@
...
+	updatePreVotesPreCommits(lastBlockHeader) {
+		debug('updatePreVotesPreCommits invoked');
+		// Update applies particularly in reference to last block header in the list
+		const header = lastBlockHeader || this.headers.last;
+
+		// If delegate forged a block with higher or same height previously
+		// that means he is forging on other chain and we don't count any
+		// pre-votes and pre-commits from him
+		if (header.maxHeightPreviouslyForged >= header.height) {
+			return;
```
yatki: eslint(consistent-return) :D Seriously why eslint is not complaining as much as i do :D

yatki: Also please avoid undefined returns.

framework/src/modules/chain/bft/finality_manager.js
```diff
@@ -0,0 +1,297 @@
...
+		this.state[delegatePublicKey] = delegateState;
+	}
+
+	/**
+	 * Update the pre-voted confirmed and finalized height
+	 */
+	updatePreVotedAndFinalizedHeight() {
+		debug('updatePreVotedAndFinalizedHeight invoked');
+		if (this.headers.length === 0) {
+			return;
```
yatki: same above.

framework/src/modules/chain/bft/finality_manager.js
```diff
@@ -0,0 +1,297 @@
...
+			blockHeader.prevotedConfirmedUptoHeight !== this.prevotedConfirmedHeight
+		) {
+			throw new BFTInvalidAttributeError(
+				'Wrong prevotedConfirmedHeight in blockHeader.'
+			);
+		}
+
+		// Find top most block forged by same delegate
+		const delegateLastBlock = this.headers
+			.top(this.processingThreshold)
```
yatki: I couldn't fine proper documentation for array.top. Is it a standart array method?

framework/test/jest/unit/specs/modules/chain/bft/utils.spec.js
```diff
@@ -0,0 +1,49 @@
...
+} = require('../../../../../../mocha/fixtures/blocks');
+
+const {
+	validateBlockHeader,
+} = require('../../../../../../../src/modules/chain/bft/utils');
+
+describe('utils', () => {
+	describe('validateBlockHeader', () => {
+		it('should be ok for valid headers', async () => {
+			const header = blockHeaderFixture();
```
yatki: please use // Arrange & Act & Assert comments in the tests for better readability.

framework/src/modules/chain/bft/bft.js
```diff
@@ -14,273 +14,135 @@
...
-
-		// If delegate is new then first block of the round will be considered
-		// if it forged before then we probably have the last commit height
-		// delegate can't pre-commit and block before above mentioned conditions
-		const minPreCommit = Math.max(
-			delegateMinHeightActive,
-			delegateState.maxPreCommitHeight + 1
+	async init() {
+		this.finalityManager = await this._initFinalityManager();
+		const finalizedHeight = this.finalityManager.finalizedHeight;
```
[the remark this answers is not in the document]
nazarhussain: If ES-Lint didn't complaint then we should be ok with the syntax. right?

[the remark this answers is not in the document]
yatki: Seems like it's disabled in .eslintrc ok let's ignore it. We can refactor it in future.

framework/src/modules/chain/bft/bft.js
```diff
@@ -14,273 +14,135 @@
...
-		this.prevotedConfirmedHeight = 0;
-		this.preVotes = {};
-		this.preCommits = {};
-
-		this.headers.items.forEach(header => {
-			this.updatePreVotesPreCommits(header);
-			this.updatePreVotedAndFinalizedHeight();
-		});
+	async _getLastBlockHeight() {
+		const lastBlock = await this.BlockEntity.get(
```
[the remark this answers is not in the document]
nazarhussain: I used it before :) it throws error if no record found. And trying to catch and see if the error was particularly that one or some other was not a clean solution.

framework/src/modules/chain/bft/errors.js
nazarhussain: @yatki I would do it on your request 😄

framework/src/modules/chain/bft/finality_manager.js
nazarhussain: Because you are more tedious reviewer than eslint 😄

nazarhussain: I would change that, but that change was not part of this PR. Its just the issue Git didn't tracked the file rename very well.

framework/src/modules/chain/bft/finality_manager.js
nazarhussain: Same above, this code was not changed in this PR, didn't tracked well by Git. But still changing on your request.

framework/src/modules/chain/bft/finality_manager.js
nazarhussain: Its the method from `HeadsList`

framework/test/jest/unit/specs/modules/chain/bft/utils.spec.js
nazarhussain: I felt code itself is very sell explanatory and not needed such comments for few lines of blocks.

framework/src/modules/chain/bft/finality_manager.js
yatki: Let's ignore this comment too.

framework/src/modules/chain/bft/bft.js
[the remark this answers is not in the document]
yatki: Hmm. Why throw error in the first place. But anyway, I see, it's not part of this PR. Let's ignore this comment too.

framework/src/modules/chain/bft/errors.js
nazarhussain: While refactoring, I noticed that its very tedious to match against long string messages to see if that particular error occurred. So its better to have different error constructors for each error type for now. 

I agree to the approach that we must have less error constructors, but for that we need to introduce error codes in these system to pinpoint the exact error quickly rather matching long string messages. That usually end up in doing regular expression matching on the error messages.

yatki: If we have a separate issue to follow up on this, we can ignore it for now. 👍  @nazarhussain could you please create the issue, and reference to this discussion?

nazarhussain: @yatki Created a separate issue to refine that aspect fo the SDK https://github.com/LiskHQ/lisk-sdk/issues/3899

PR #3853, opened 2019-06-24
Automatically cleanup round_delegates table - Closes #3683

### LINKED ISSUE #3683: Automatically cleanup round_delegates table
### Expected behavior
We don't need to store delegate lists before the round which contains finalised block. So we need to run a **non-blocking** cleanup function (eg. child process) at the end of every round to cleanup redundant data from `round_delegates` table.
### Actual behavior
We don't have this functionality at the moment.

SargeKhan:
### What was the problem?
The delegate list which is saved in the dpos module for each round needs to be sliced when all blocks in the round are finalized. This functionality was non-existent in the current implementation
### How did I fix it?
Created events which listen for finalized blocks and remove the delete lists for rounds when the blocks are finalized.
### How to test it?
This is a draft PR, tests will be added after the system is finalized.
### Review checklist

* The PR resolves #3683 
* Linting passes
* Tests pass


SargeKhan:
@nazarhussain @yatki please give feedback on the events I am using in this PR to achieve the goal of removing removable delegate lists for rounds. If we agree on events used in this function, i shall add the relevant tests for the functionality.

framework/src/modules/chain/dpos/dpos.js
```diff
@@ -1,11 +1,27 @@
...
+		this.delegate.on(EVENT_ROUND_FINISHED, () => {
+			this.onRoundFinish();
+		});
 	}
 
 	async getRoundDelegates(round) {
 		return this.delegates.getRoundDelegates(round);
 	}
+
+	async onBlockFinalized({ height }) {
```
shuse2: I think the naming of this function should be more descriptive.
rather than calling it `onBlockFinalized` maybe `updateFinalizedRound`?

framework/src/modules/chain/dpos/dpos.js
```diff
@@ -1,11 +1,27 @@
-const Delegates = require('./delegates');
+const { Delegates, EVENT_ROUND_FINISHED } = require('./delegates');
 
 module.exports = class Dpos {
```
shuse2: It's out of scope, but probably better to export as object as other modules

framework/src/modules/chain/dpos/dpos.js
```diff
@@ -1,11 +1,27 @@
...
 	async getRoundDelegates(round) {
 		return this.delegates.getRoundDelegates(round);
 	}
+
+	async onBlockFinalized({ height }) {
+		this.finalizedBlockRound = this.slots.calcRound(height);
+	}
+
+	async onRoundFinish() {
+		const delegateListOffsetForRound = 2;
```
yatki: shouldn't `2` be a configurable parameter?

framework/src/modules/chain/dpos/dpos.js
SargeKhan: It is going to be used as an event handler.

framework/src/modules/chain/bft/bft.js
```diff
@@ -38,8 +40,9 @@ const validateBlockHeader = blockHeader =>
  * @property {string} delegatePublicKey
  */
 
-class BFT {
+class BFT extends EventEmitter {
```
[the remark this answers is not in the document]
SargeKhan: This is the general pattern we follow in our application, I am merely building on top of the existing event declaration/handling style.

framework/src/modules/chain/dpos/dpos.js
SargeKhan: yes, this parameter is used in multiple places. It should be defined in process_V2.js which is out of scope of this PR. So, i merely set a private variable for it. Maybe I can add a TODO comment for it.

framework/src/modules/chain/chain.js
```diff
@@ -570,11 +581,16 @@ module.exports = class Chain {
 		this.blocks.on(EVENT_NEW_BROADHASH, ({ broadhash, height }) => {
 			this.channel.invoke('app:updateApplicationState', { broadhash, height });
 		});
+
+		this.bft.on(EVENT_BLOCK_FINALIZED, ({ height }) => {
+			this.dpos.onBlockFinalized({ height });
+		});
```
yatki:
```js
this.bft.on(EVENT_BLOCK_FINALIZED, this.dpos.onBlockFinalized);
```

Here, since the method is not used like above, and even if it was used like that, I'd still recommend to rename the method to clarify what method is doing. Because event name is already telling which event was triggered. I can't understand what method is doing from `onBlockFinalized` but `updateFinalizedRound` is more clear, I think. The same goes for updateFinishedRound.

What do you think?

framework/src/modules/chain/components/storage/entities/round_delegates.js
```diff
@@ -99,8 +99,9 @@ class RoundDelegates extends BaseEntity {
 			.map(k => `"${this.fields[k].fieldName}"`)
 			.join(',');
 
-		const values = [{ round, delegatePublicKeys }];
-
+		const values = [
+			{ round, delegatePublicKeys: JSON.stringify(delegatePublicKeys) },
```
yatki: ❤️

framework/src/modules/chain/chain.js
limiaspasdaniel: I agree with Mehmet's idea. I think `onBlockFinalized` is more of an event handler so maybe you could event write it like this: 

`this.bft.on(EVENT_BLOCK_FINALIZED, this.dpos.onBlockFinalized)` and the function implementation could call `updateFinishedRound`. This way if eventually, anything else needs to be performed it can be appended and called.

Or you could directly call `updateFinishedRound`.

SargeKhan: The reason I am partial towards keeping generic names is to hide the internal logic of the modules from the encapsulating chain module. If we use generic method names, the chain only needs to know that dpos module requires to know when a block is finalized. It doesn't care what is the purpose and how dpos will use this information. For me, this decouples chain module and dpos module implementation such that if in the future, we want to updateFinalizedRound as well as perfom another action `onBlockFinalized` event, the code in chain module will need no update.

SargeKhan: Since @shu, @limiaspasdaniel and @yatki side with have descriptive names in chain module, please respond to the above point i made, if you still don't agree, i'll change the naming of the event handlers

framework/src/modules/chain/dpos/dpos.js
```diff
@@ -1,11 +1,27 @@
...
 
 	async getRoundDelegates(round) {
 		return this.delegates.getRoundDelegates(round);
 	}
+
+	async onBlockFinalized({ height }) {
+		this.finalizedBlockRound = this.slots.calcRound(height);
+	}
+
+	async onRoundFinish() {
```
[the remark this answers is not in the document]
SargeKhan: I made the comment an argument for generic names in another comment, please resolve this comment if you agree with my rationale there.

framework/src/modules/chain/bft/bft.js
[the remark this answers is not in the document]
SargeKhan: Secondly, the inheritance of BFT class and Delegate class from EventEmitter in this PR is merely to declare (without runtime errors) the integration logic in the chain module for event handlers in dpos. The proper system and structure for event emitters can be discussed in an STM. Would you agree?

framework/src/modules/chain/chain.js
limiaspasdaniel: I think your comment is aligned with what I mentioned: 
> this.bft.on(EVENT_BLOCK_FINALIZED, this.dpos.onBlockFinalized) and the function implementation could call updateFinishedRound. This way if eventually, anything else needs to be performed it can be appended and called.

So all good from my side.

framework/src/modules/chain/bft/bft.js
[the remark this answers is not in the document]
yatki: We can have an STM but as i said earlier, I have a strong opinion on this. Other devs may repeat the same pattern and it's just more work for future. However, i don't want to prevent the PR to be merged. This can be easily refactored. If we merge this PR like this, would be great if we could have an issue to follow up on this.

framework/src/modules/chain/chain.js
yatki: @SargeKhan I think you make sense :D 👍

framework/src/modules/chain/bft/bft.js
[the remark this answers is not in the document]
SargeKhan: I created issue #3904 to further discuss this point.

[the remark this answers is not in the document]
yatki: @SargeKhan could you please the issue and reference it to this discussion?

PR #3855, opened 2019-06-24
Remove relayLimit from broadcaster - Closes #3819

### LINKED ISSUE #3819: Remove relayLimit from broadcaster
### Expected behavior

We should remove the relayLimit to ensure good propagation of blocks after having implemented partial view of the network.

### Actual behavior

Currently we stop broadcasting a block after it has been relayed beyond the relayLimit threshold.

2.3

jondubois:
### What was the problem?

The relayLimit can cause issues with propagation. When a block, transaction or signature has already been processed, it will not be relayed so the relayLimit is not required.

### How did I fix it?

- Removed the `maxRelays(...)` check so that the broadcaster does not account for relayLimit when relaying a block, transaction or signature.
- The node will always set the `relayLimit` to `1`; this is to prevent breaking backwards compatibility with existing nodes. Also, keeping the `relayLimit` at `1` will prevent other nodes from hitting the relay limit.

### How to test it?

Run tests

### Review checklist

* The PR resolves #3819
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/chain/defaults/config.js
+++ b/framework/src/modules/chain/defaults/config.js
@@ -33,13 +33,8 @@ const defaultConfig = {
 					minimum: 1,
 					maximum: 25,
 				},
-				relayLimit: {
-					type: 'integer',
-					minimum: 1,
-					maximum: 100,
-				},
 			},
-			required: ['broadcastInterval', 'releaseLimit', 'relayLimit'],
+			required: ['broadcastInterval', 'releaseLimit'],
 		},
 		transactions: {
 			type: 'object',
@@ -248,7 +243,6 @@ const defaultConfig = {
 			active: true,
 			broadcastInterval: 5000,
 			releaseLimit: 25,
-			relayLimit: 3,
 		},
 		transactions: {
 			maxTransactionsPerQueue: 1000,
--- a/framework/src/modules/chain/transport.js
+++ b/framework/src/modules/chain/transport.js
@@ -28,7 +28,7 @@ const exceptions = global.exceptions;
 const { MAX_SHARED_TRANSACTIONS } = global.constants;
 
 function incrementRelays(packet) {
-	if (!packet.relays) {
+	if (!Number.isInteger(packet.relays)) {
 		packet.relays = 0;
 	}
 	packet.relays++;
--- a/framework/src/modules/chain/transport.js
+++ b/framework/src/modules/chain/transport.js
@@ -108,7 +108,9 @@ class Transport {
 	 */
 	// eslint-disable-next-line class-methods-use-this
 	onSignature(signature, broadcast) {
-		if (broadcast && !__private.broadcaster.maxRelays(signature)) {
+		if (broadcast) {
+			// TODO: Remove the relays property as part of the next hard fork. This needs to be set to a fixed value for backwards compatibility.
+			signature.relays = 1;
 			__private.broadcaster.enqueue(
 				{},
 				{
@@ -132,7 +134,9 @@ class Transport {
 	 */
 	// eslint-disable-next-line class-methods-use-this
 	onUnconfirmedTransaction(transaction, broadcast) {
-		if (broadcast && !__private.broadcaster.maxRelays(transaction)) {
+		if (broadcast) {
+			// TODO: Remove the relays property as part of the next hard fork. This needs to be set to a fixed value for backwards compatibility.
+			transaction.relays = 1;
 			const transactionJSON = transaction.toJSON();
 			__private.broadcaster.enqueue(
 				{},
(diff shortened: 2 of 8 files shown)


framework/src/modules/chain/transport.js
```diff
@@ -108,7 +108,9 @@ class Transport {
 	 */
 	// eslint-disable-next-line class-methods-use-this
 	onSignature(signature, broadcast) {
-		if (broadcast && !__private.broadcaster.maxRelays(signature)) {
+		if (broadcast) {
+			// TODO: Remove the relays property as part of the next hard fork. This needs to be set to a fixed value for backwards compatibility.
+			signature.relays = 1;
```
shuse2: If we are going to backward compatible way, I think we should keep track of release limit, otherwise with previous implementation, it will propagate too much

framework/test/mocha/unit/modules/chain/transport.js
```diff
@@ -801,40 +801,30 @@ describe('transport', () => {
...
-							maxRelays: sinonSandbox.stub().returns(false),
 							enqueue: sinonSandbox.stub(),
 						};
 						transportInstance.onSignature(SAMPLE_SIGNATURE_1, true);
 					});
 
-					it('should call __private.broadcaster.maxRelays with signature', async () => {
-						expect(__private.broadcaster.maxRelays.calledOnce).to.be.true;
+					it('should call __private.broadcaster.enqueue with {} and {api: "postSignatures", data: {signature: signature}} as arguments', async () => {
+						expect(__private.broadcaster.enqueue.calledOnce).to.be.true;
```
diego-G: Please, remove async and return from the test.

framework/test/mocha/unit/modules/chain/transport.js
```diff
@@ -855,55 +845,43 @@ describe('transport', () => {
...
-								__private.broadcaster.enqueue.calledWith(
-									{},
-									{
-										api: 'postTransactions',
-										data: { transaction: transaction.toJSON() },
-									}
-								)
-							).to.be.true;
-						});
+					it('should call __private.broadcaster.enqueue with {} and {api: "postTransactions", data: {transaction}}', async () => {
```
diego-G: Same as above.

framework/test/mocha/unit/modules/chain/transport.js
```diff
@@ -855,55 +845,43 @@ describe('transport', () => {
...
+					});
 
-						it('should call library.channel.publish with "chain:transactions:change" and transaction as arguments', async () => {
-							expect(library.channel.publish).to.be.calledOnce;
-							expect(library.channel.publish).to.be.calledWith(
-								'chain:transactions:change',
-								transaction.toJSON()
-							);
-						});
+					it('should call library.channel.publish with "chain:transactions:change" and transaction as arguments', async () => {
```
diego-G: Same as above.

framework/src/modules/chain/transport.js
```diff
@@ -27,6 +27,13 @@ const transactionsModule = require('./transactions');
 const exceptions = global.exceptions;
 const { MAX_SHARED_TRANSACTIONS } = global.constants;
 
+function incrementRelays(packet) {
```
diego-G: Could you make a static function inside the class?

diego-G: @shuse2 which alternative do you propose or what's your point of view here?

shuse2: if it doesn't rely on the state, it should be just a simple function.
We don't use `static` function unless design/UX related point

framework/src/modules/chain/defaults/config.js
```diff
@@ -33,13 +33,8 @@ const defaultConfig = {
...
 					maximum: 25,
 				},
-				relayLimit: {
-					type: 'integer',
-					minimum: 1,
-					maximum: 100,
-				},
 			},
-			required: ['broadcastInterval', 'releaseLimit', 'relayLimit'],
+			required: ['broadcastInterval', 'releaseLimit'],
```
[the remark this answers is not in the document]
diego-G: @jondubois disregard this comment. I've noticed those tests are disabled since  https://github.com/LiskHQ/lisk-sdk/commit/ec996cbf53b407d35eaaa6474f3bdd23288f7a78 @shuse2 or @MaciejBaj , any plan to reenable them?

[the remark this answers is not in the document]
shuse2: I already re-enabled in my PR for 2.1 https://github.com/LiskHQ/lisk-sdk/pull/3880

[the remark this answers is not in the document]
diego-G: Is 2.1.0 merged into development ? In that case, please @jondubois update our feature branch.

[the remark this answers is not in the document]
shuse2: This PR is not merged yet =)

framework/test/mocha/unit/modules/chain/transport.js
jondubois: It seems that I cannot remove them. I get these eslint errors:

```
809:133  error  Unexpected synchronous test  mocha/no-synchronous-tests
866:113  error  Unexpected synchronous test  mocha/no-synchronous-tests
879:110  error  Unexpected synchronous test  mocha/no-synchronous-tests
```

PR #3858, opened 2019-06-25
Add more propagation tests - Closes #3812

### LINKED ISSUE #3812: Add a P2P integration test scenario to cover message propagation in partially connected network
### Expected behavior

1. There needs to be additional test cases in the P2P integration test scenario https://github.com/LiskHQ/lisk-sdk/blob/5ee62078e683dc89f0eada152f2eab917f6fda34/elements/lisk-p2p/test/integration/p2p.ts#L1318 to check that messages and requests each an appropriate number of peers.

2. We should mock some additional 'message propagation' functionality into a new test scenario so that each P2P node re-broadcasts a message to its peers during a certain time period, then at the end, we can verify that all the nodes in the network have received the message. The network should have at least 30 nodes.

### Actual behavior

1. For the partial mesh network, currently, we only check peer discovery and connections, but we should also check the `p2p.send(...)` and `p2p.request(...)` methods.

2. Also, we do not currently have an integration test scenario to verify that messages propagate correctly through the network. 

### Steps to reproduce

Run tests.

Feature branch: `feature/implement-lip-p2p`

diego-G:
### How did I fix it

- f9b8e34
- Change `forEach` to `for of` across P2P integration tests
- Add special test cases for the scenario with limited connections by peer

### How to test it?
`~/lisk-sdk/elements/lisk-p2p$ npm run test:integration`

### Review checklist

* The PR resolves #3812
* Linting passes
* Tests pass

### DIFF
--- a/elements/lisk-p2p/test/integration/p2p.ts
+++ b/elements/lisk-p2p/test/integration/p2p.ts
@@ -86,9 +86,9 @@ describe('Integration tests for P2P library', () => {
 		});
 
 		it('should set the isActive property to true for all nodes', () => {
-			p2pNodeList.forEach(p2p => {
+			for (let p2p of p2pNodeList) {
 				expect(p2p).to.have.property('isActive', true);
-			});
+			}
 		});
 
 		describe('P2P.request', () => {
@@ -150,7 +150,7 @@ describe('Integration tests for P2P library', () => {
 				// Wait for a few cycles of discovery.
 				await wait(POPULATOR_INTERVAL * 10);
 
-				p2pNodeList.forEach(p2p => {
+				for (let p2p of p2pNodeList) {
 					const { connectedPeers } = p2p.getNetworkStatus();
 
 					const peerPorts = connectedPeers
@@ -161,7 +161,7 @@ describe('Integration tests for P2P library', () => {
 					);
 
 					expect(peerPorts).to.be.eql(expectedPeerPorts);
-				});
+				}
 			});
 		});
 
@@ -228,14 +228,14 @@ describe('Integration tests for P2P library', () => {
 
 				beforeEach(async () => {
 					collectedMessages = [];
-					p2pNodeList.forEach(p2p => {
+					for (let p2p of p2pNodeList) {
 						p2p.on('messageReceived', message => {
 							collectedMessages.push({
 								nodePort: p2p.nodeInfo.wsPort,
 								message,
 							});
 						});
-					});
+					}
 				});
 
 				it('should send messages to peers within the network; should reach multiple peers with even distribution', async () => {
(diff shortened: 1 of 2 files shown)


elements/lisk-p2p/test/integration/p2p.ts
```diff
@@ -1349,47 +1308,186 @@ describe('Integration tests for P2P library', () => {
...
+			});
+
+			it('should receive a message in the correct format', async () => {
+				const firstP2PNode = p2pNodeList[0];
+				firstP2PNode.send({ event: 'bar', data: 'test' });
+
+				await wait(100);
+
+				expect(collectedMessages).to.be.an('array');
+				expect(collectedMessages.length).to.be.greaterThan(6);
```
mitsujutsu: Where does `6` come from?

diego-G: The number of collected messages was sometimes 7,8 or 9. I don't know if this can be an issue. What do you think @jondubois?

elements/lisk-p2p/test/integration/p2p.ts
```diff
@@ -1349,47 +1308,186 @@ describe('Integration tests for P2P library', () => {
...
+		});
+
+		describe('P2P.send', () => {
+			let collectedMessages: Array<any> = [];
+
+			beforeEach(() => {
+				collectedMessages = [];
+				for (let p2p of p2pNodeList) {
+					p2p.on('messageReceived', message => {
+						collectedMessages.push({
```
jondubois: Here we're just recording the messages. It would be good to add another test case for propagation.

We could add a special propagation handler like this:

```js
let propagationStart = Date.now();
p2p.on('messageReceived', message => {
  let propagationDuration = Date.now() - propagationStart;
  // All nodes will propagate messages to each other during 200ms
  if (message.event === 'propagate' && propagationDuration < 200) {
     p2p.send({ event: 'propagate', data: message.data + 1 });
  }
});
```

Then inside the test case you can do (just once; this simulates forging a block and then sending it to the network):

```
firstP2PNode.send({ event: 'propagate', data: 0 });
```

jondubois: Also, we should simulate an additional case related to what happens if a node receives a message which it has already seen (this is what happens if our node has already seen a particular block); in that case we should not propagate the message again.

To do this, we should keep a Map or Set of all messages received during the test case. The first time we see a message, we propagate it, otherwise we do nothing.

elements/lisk-p2p/test/integration/p2p.ts
diego-G: Reverted

PR #3865, opened 2019-06-26
Update application state structure to include prevotedConfirmedUptoHeight  - Closes #3774

### LINKED ISSUE #3774: Update Application State interface to include prevotedConfirmedUptoHeight
### Expected behavior
Update Application State interface to include `prevotedConfirmedUptoHeight` so network:getPeers returns the required information to perform peer selection on Block Synchronization Mechanism's step 1.
`feature/BFT`

limiaspasdaniel:
### What was the problem?
`prevotedConfirmedUptoHeight` was needed to be included in the Application State to be able to inform the peers about it. Step 1 of [Block Synchronization Mechanism](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#block-synchronization-mechanism)
### How did I fix it?
- Added `prevotedConfirmedUptoHeight` to the Application State
- Made EVENT_NEW_BLOCK to be fired when a block is received from the network passively and applied. For some reason, this was not the case before.
### How to test it?
Run `framework/test/jest/unit/specs/controller/application_state.spec.js`
### Review checklist
* The PR resolves #3774 
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/controller/application_state.js
+++ b/framework/src/controller/application_state.js
@@ -33,6 +33,7 @@ const __private = {
  * - height
  * - nethash
  * - prevotedConfirmedUptoHeight
+ * - broadhash
  * - nonce
  *
  * @class
@@ -65,6 +66,7 @@ class ApplicationState {
 			height: 1,
 			prevotedConfirmedUptoHeight: 0,
 			nethash,
+			broadhash: nethash,
 			nonce,
 		});
 	}
@@ -78,17 +80,19 @@ class ApplicationState {
 	}
 
 	/**
-	 * Updates prevotedConfirmedUptoHeight and height values.
+	 * Updates the application state.
 	 *
-	 * @param {prevotedConfirmedUptoHeight, height} parameters - prevotedConfirmedUptoHeight and height to update
-	 *
-	 * @returns {Promise.<boolean, Error>}
+	 * @param height
+	 * @param prevotedConfirmedUptoHeight
+	 * @param broadhash
+	 * @return {Promise<boolean, Error>}
 	 * @throws assert.AssertionError
 	 */
-	async update({ height, prevotedConfirmedUptoHeight }) {
+	async update({ height, prevotedConfirmedUptoHeight, broadhash }) {
 		assert(height, 'height is required to update application state.');
 		try {
 			const newState = this.state;
+			newState.broadhash = broadhash || this.state.broadhash;
 			newState.prevotedConfirmedUptoHeight =
 				prevotedConfirmedUptoHeight || this.state.prevotedConfirmedUptoHeight;
 			newState.height = height;
--- a/framework/src/controller/application_state.js
+++ b/framework/src/controller/application_state.js
@@ -85,11 +85,12 @@ class ApplicationState {
 	 * @returns {Promise.<boolean, Error>}
 	 * @throws assert.AssertionError
 	 */
-	async update({ height, prevotedConfirmedUptoHeight = 0 }) {
+	async update({ height, prevotedConfirmedUptoHeight }) {
 		assert(height, 'height is required to update application state.');
 		try {
 			const newState = this.state;
-			newState.prevotedConfirmedUptoHeight = prevotedConfirmedUptoHeight;
+			newState.prevotedConfirmedUptoHeight =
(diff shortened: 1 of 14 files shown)


framework/src/controller/application_state.js
```diff
@@ -78,19 +78,18 @@ class ApplicationState {
...
 	 * @throws assert.AssertionError
 	 */
-	async update({ broadhash, height }) {
-		assert(broadhash, 'broadhash is required to update application state.');
+	async update({ height, prevotedConfirmedUptoHeight = 0 }) {
 		assert(height, 'height is required to update application state.');
 		try {
 			const newState = this.state;
-			newState.broadhash = broadhash;
+			newState.prevotedConfirmedUptoHeight = prevotedConfirmedUptoHeight;
```
michielmulders: Concern about this line:
The default value for `prevotedConfirmedUpToHeight` is `0`.

What if we don't want to update the `prevotedConfirmedUpToHeight`, then it will set the value to `0`. Or is this case not possible atm?

I was thinking about adding a check that the new value must be bigger than the current `prevotedConfirmed`, but when kicking in Block Sync Mechanism, the new value can be lower than the current one.

Or remove the default value for `prevotedConfirmedUpToHeight` as you already set the initial state to `0` in `line 66`.

framework/test/jest/unit/specs/controller/application_state.spec.js
```diff
@@ -93,7 +93,7 @@ describe('Application State', () => {
 		describe('when there is an error', () => {
 			// Arrange
 			const newState = {
-				broadhash: 'xxx',
+				prevotedConfirmedUptoHeight: 0,
```
michielmulders: Can you tell why this is throwing an error?

framework/src/controller/application_state.js
limiaspasdaniel: That's a good point @michielmulders and I agree with you. The concrete implementation of this function was too tied to the concept of broadhash (you couldn't update another property of the state without not sending the broadhash ) and now it remains the same behavior for `prevotedConfirmedUpToHeight`.

At the moment there is only one place where the application state is updated but if someone else wants to update the application state from another part of the code this will result in wrongly setting the attribute to `0`.

> I was thinking about adding a check that the new value must be bigger than the current prevotedConfirmed, but when kicking in Block Sync Mechanism, the new value can be lower than the current one.

This shouldn't be part of the logic of the application state. I like better your idea of removing the default of `0`, but then if someone else updates the application state without passing the attribute it will result in a null value.

My opinion is that the application state `update` function should be generic, have schema validation (to not pass wrong formats or different attributes), and it should not require certain attributes to be passed explicitly. But this is the scope of another issue.

pablitovicente: I would also consider getters and setter for having an strict control over each shared property exposed by application state.

framework/test/jest/unit/specs/controller/application_state.spec.js
limiaspasdaniel: In this case we are checking what happens when there is an error thrown by publishing an event using the `channel`

framework/src/controller/application_state.js
michielmulders: Getters and setters I agree with! 👍 
And @limiaspasdaniel , I would remove the default value for the input parameters in the update function, but leave the initialisation of the state to `0` so a `null` value is not possible.

limiaspasdaniel:
@pablitovicente @michielmulders brought back the broadhash as it was getting out of the scope of this PR. Tests are green, please review again :)

framework/src/modules/chain/chain.js
```diff
@@ -558,6 +558,14 @@ module.exports = class Chain {
 				);
 			}
 			this.channel.publish('chain:blocks:change', block);
+
+			// If block version is 2
```
pablitovicente: the comment mentions block version but the if checks for `loader.syncing()` can this comment be improved with more details?

limiaspasdaniel: Actually, it can be removed. Good catch!

limiaspasdaniel: Related issue: #3875

michielmulders: Agree, it feels not related to this line of code like this.

PR #3873, opened 2019-06-26
Move transactions to the ready queue only if all transaction can be processed together

### LINKED ISSUE #3871: Move transactions to the ready queue only if all transaction can be processed together
### Expected behavior
During the processing of a group of transactions, failed transactions also update the related accounts. They can update the account state in such a way that succeeding invalid transactions become valid. Hence, in order to guarantee transactions are processable together and can be added to the ready queue, all transactions should be passed. 
### Actual behavior
After processing transactions, failed transactions were removed, while passed transactions were put in the ready queue.
release/2.0.0

SargeKhan:
### What was the problem?
Invalid transactions got stuck in the transaction pool.
### How did I fix it?
Fixed the issue by removing the invalid (failed) transactions from the transaction pool and re-processing the passed transactions without adding any new transactions to the ready queue. 
### How to test it?
Implemented the tests which tackle this scenario, we can also test this feature by creating many conflicting transactions which cannot be processed together.
### Review checklist

* The PR resolves #3871 
* Linting passes
* Tests pass

### DIFF
--- a/elements/lisk-transaction-pool/src/transaction_pool.ts
+++ b/elements/lisk-transaction-pool/src/transaction_pool.ts
@@ -560,31 +560,35 @@ export class TransactionPool extends EventEmitter {
 			queueCheckers.checkTransactionForId(failedTransactions),
 		);
 
-		// Keep transactions in the ready queue which still exist
-		this._queues.ready.enqueueMany(
-			this._queues.ready.removeFor(
-				queueCheckers.checkTransactionForId(passedTransactions),
-			),
-		);
+		// Empty failed transactions list gurantees that all transactions can be processed together
+		// So move all passed transactions to the ready queue
+		if (failedTransactions.length === 0) {
+			// Keep transactions in the ready queue which still exist
+			this._queues.ready.enqueueMany(
+				this._queues.ready.removeFor(
+					queueCheckers.checkTransactionForId(passedTransactions),
+				),
+			);
 
-		// Move processeable transactions from the verified queue to the ready queue
-		this._queues.ready.enqueueMany(
-			this._queues.verified.removeFor(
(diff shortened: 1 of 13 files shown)


pablitovicente:
@SargeKhan integration tests are failing and do not seem to be random failures. Could you please check?

elements/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -560,31 +560,35 @@ export class TransactionPool extends EventEmitter {
...
 
-		// Keep transactions in the ready queue which still exist
-		this._queues.ready.enqueueMany(
-			this._queues.ready.removeFor(
-				queueCheckers.checkTransactionForId(passedTransactions),
-			),
-		);
+		// Empty failed transactions list gurantees that all transactions can be processed together
+		// So move all passed transactions to the ready queue
+		if (failedTransactions.length === 0) {
```
shuse2: I think in case of
```
@verified queue
1. Valid Tx
2. Invalid Tx
```
failedTransactions length will be 1, and it will not move to ready queue?

SargeKhan: No, it will not be moved. I think if we optimize for these scenarios we will do it at the cost of complicating the logic. Complications will arise due to the fact that a) we'll be removing valid transactions from the verified queue and the pending queue and b) valid transactions will be only removed until we encounter first invalid transaction c) the function will get overcomplicated after dynamic fees because we reorder transactions from ready, pending and verified queues before processing them together.
This function is executed every second, and this scenario is a very narrow edge case. Hence, optimizing it does add many benefits for me. Do you disagree?

PR #3876, opened 2019-06-27
Use nodejs crypto library for checksum verification - Closes #3872

### LINKED ISSUE #3872: Use nodejs crypto library for checksum verification
### Expected behavior
To avoid inconsistent behaviors it's better to use Nodejs `crypto` module which provides a way to verify the checksum  
### Actual behavior
`shasum` is used for verifying the checksum of the Lisk core binary, this results in inconsistent behavior in the ubunut instance. 
### Steps to reproduce
- Use a bare Linux machine or Install Docker on your machine and run `docker run -it ubuntu:18.04 /bin/bash`
- Create lisk user and install commander `2.2.0`
- Extract and install Lisk core `./lisk core:install --network=testnet testnet-01`
- You will receive an error `→ /bin/sh: 1: shasum: not found`
Lisk Commander 2.2.0

ManuGowda:
### What was the problem?
`shasum` was incompatible with `linux` when the machine had bare minimum installation.
### How did I fix it?
Using Nodejs `crypto` library to verify the checksum to support all OS.
### How to test it?
- Use a bare Linux machine or Install Docker on your machine and run docker run -it ubuntu:18.04 /bin/bash
- Create lisk user and install commander 2.2.0
- Extract and install Lisk core ./lisk core:install --network=testnet testnet-01
- The installation should succeed

### Review checklist

* The PR resolves #3872 
* Linting passes
* Tests pass

### DIFF
--- a/commander/package-lock.json
+++ b/commander/package-lock.json
@@ -121,7 +121,7 @@
 		},
 		"@liskhq/bignum": {
 			"version": "1.3.1",
-			"resolved": "https://npm.lisk.io/@liskhq%2fbignum/-/bignum-1.3.1.tgz",
+			"resolved": "https://registry.npmjs.org/@liskhq/bignum/-/bignum-1.3.1.tgz",
 			"integrity": "sha512-q9+NvqbpmXOqpPmV8Y+XSEIUJFMZDGyfW6rkN9Ej3nzPb/qurY/Ic2UPTeTTaj8+q/bcw5JUwTb86hi7PIziDg==",
 			"requires": {
 				"@types/node": "11.11.2"
@@ -129,14 +129,14 @@
 			"dependencies": {
 				"@types/node": {
 					"version": "11.11.2",
-					"resolved": "https://npm.lisk.io/@types%2fnode/-/node-11.11.2.tgz",
+					"resolved": "https://registry.npmjs.org/@types/node/-/node-11.11.2.tgz",
 					"integrity": "sha512-iEaHiDNkHv4Jrm9O5T37OYEUwjJesiyt6ZlhLFK0sbo4CLD0jyCOB4Pc2F9iD3MbW2397SLNxZKdDGntGaBjQQ=="
 				}
 			}
 		},
 		"@liskhq/lisk-api-client": {
 			"version": "2.0.1",
-			"resolved": "https://npm.lisk.io/@liskhq%2flisk-api-client/-/lisk-api-client-2.0.1.tgz",
+			"resolved": "https://registry.npmjs.org/@liskhq/lisk-api-client/-/lisk-api-client-2.0.1.tgz",
 			"integrity": "sha512-EScNNZz5Ee5P3jRMpKytQEMACZut2I5kT6Z+DCK3VcdhRC2h6r+e+wPx4hNNJcgoIMjLysI2e6c79T5wtbRryA==",
 			"requires": {
 				"@types/node": "10.12.21",
@@ -145,7 +145,7 @@
 		},
 		"@liskhq/lisk-constants": {
 			"version": "1.2.1",
-			"resolved": "https://npm.lisk.io/@liskhq%2flisk-constants/-/lisk-constants-1.2.1.tgz",
+			"resolved": "https://registry.npmjs.org/@liskhq/lisk-constants/-/lisk-constants-1.2.1.tgz",
 			"integrity": "sha512-myu4rawbXCiPnbwuKUngBT3SvD2JG7Yvdfb5RU6q89YvDfnWai6HCoGID8Rj4StvrX+PktiijRX78rpoaVC8kw==",
 			"requires": {
 				"@types/node": "10.12.21"
@@ -153,7 +153,7 @@
 		},
 		"@liskhq/lisk-cryptography": {
 			"version": "2.1.0",
-			"resolved": "https://npm.lisk.io/@liskhq%2flisk-cryptography/-/lisk-cryptography-2.1.0.tgz",
+			"resolved": "https://registry.npmjs.org/@liskhq/lisk-cryptography/-/lisk-cryptography-2.1.0.tgz",
 			"integrity": "sha512-syxHTng4pd7LaeXVSrwBqM54wgcATjVHO/W9fvTjq3NfuTER1Y7vwlIpzXebPG5Dr4CNT8C/Q+7nAIx/gEKQLA==",
 			"requires": {
 				"@liskhq/bignum": "1.3.1",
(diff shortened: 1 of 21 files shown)


fchavant:
@ManuGowda 
1. shouldn't there be a system test for that?
2. how can this be tested before the PR is merged? (`oclif-dev pack` seems to always pull the published version)

commander/src/utils/download.ts
```diff
@@ -88,5 +72,13 @@ export const extract = async (
 export const downloadAndValidate = async (url: string, cacheDir: string) => {
 	await download(url, cacheDir);
 	await download(`${url}.SHA256`, cacheDir);
-	await validateChecksum(`${url}.SHA256`, cacheDir);
+
+	const { fileName, fileDir } = getDownloadedFileInfo(url, cacheDir);
+	const dataBuffer = fs.readFileSync(`${fileDir}/${fileName}`);
```
fchavant: It looks like the whole file would be read into RAM. Since it is pretty large already wouldn't it be nicer to read it as a stream instead?

elements/lisk-cryptography/src/hash.ts
```diff
@@ -43,3 +43,13 @@ export const hash = (data: Buffer | string, format?: string): Buffer => {
 		'Unsupported data format. Currently only Buffers or `hex` and `utf8` strings are supported.',
 	);
 };
+
+export const verifyChecksum = (
+	data: Buffer | string,
+	checksum: string,
+): boolean => {
+	const dataHash = hash(data);
+	const hex = bufferToHex(dataHash);
```
fchavant: How does this work? The output of `sha256sum` looks like `48be5c5bff9fbf6077844165242856b9db6cdbf44e10cd9aeb8ae18f46acbac2  lisk-2.0.0-rc.0-Linux-x86_64.tar.gz` (notice the `\t<file name>`)

ManuGowda: when u generate the hex from buffer of tar file it will be just the hex `48be5c5bff9fbf6077844165242856b9db6cdbf44e10cd9aeb8ae18f46acbac2`, so it should be fine.
on the other hand, what you might be referring to `SHA256` file which has the format you mentioned, in that case i am already doing 
```
const content = fs.readFileSync(`${fileDir}/${fileName}.SHA256`, 'utf8');
const checksum = content.split(' ')[0];
```

elements/lisk-cryptography/src/hash.ts
```diff
@@ -43,3 +43,13 @@ export const hash = (data: Buffer | string, format?: string): Buffer => {
 		'Unsupported data format. Currently only Buffers or `hex` and `utf8` strings are supported.',
 	);
 };
+
+export const verifyChecksum = (
```
shuse2: maybe if it's this simple, then probably we don't need to add this function.
This is not really specific to checksum either

ManuGowda: ok then I will not add the changes in cryptography 👍 
it will not be useful in general so

commander/package.json
```diff
@@ -119,6 +119,7 @@
 		"pm2": "3.5.0",
 		"semver": "5.6.0",
 		"strip-ansi": "4.0.0",
+		"tar": "^4.4.10",
```
shuse2: please use exact version

commander/package.json
```diff
@@ -139,6 +140,7 @@
 		"@types/sinon": "7.0.5",
 		"@types/sinon-chai": "3.2.2",
 		"@types/strip-ansi": "3.0.0",
+		"@types/tar": "^4.0.1",
```
shuse2: same as above

commander/src/commands/core/install.ts
```diff
@@ -255,7 +256,11 @@ export default class InstallCommand extends BaseCommand {
 								);
 
 								createDirectory(installDir);
-								await extract(fileDir, fileName, installDir);
+								await tar.x({
```
fchavant: This is not really related, is it?

commander/package-lock.json
```diff
@@ -119,6 +119,95 @@
...
+			},
+			"dependencies": {
+				"@types/node": {
+					"version": "11.11.2",
+					"resolved": "https://registry.npmjs.org/@types/node/-/node-11.11.2.tgz",
+					"integrity": "sha512-iEaHiDNkHv4Jrm9O5T37OYEUwjJesiyt6ZlhLFK0sbo4CLD0jyCOB4Pc2F9iD3MbW2397SLNxZKdDGntGaBjQQ=="
+				}
+			}
+		},
+		"@liskhq/lisk-api-client": {
```
shuse2: siblings package shouldn't be in the `package-lock`.
Please use 
`npx lerna boostrap` in the root level to update the `package-lock`

commander/src/utils/download.ts
```diff
@@ -14,9 +14,34 @@
...
 import fs from 'fs-extra';
+import * as tar from 'tar';
 import { dateDiff, getDownloadedFileInfo } from './core/commons';
-import { exec, ExecResult } from './worker-process';
+
+export const verifyChecksum = async (
+	filePath: string,
+	expectedChecksum: string,
+): Promise<boolean> => {
+	const shasum = crypto.createHash('sha256');
```
shuse2: prefer to use `hash` and `bufferToHex` from `lisk-cryptography`
We shouldn't try not to use `crypto` directly

commander/src/commands/core/install.ts
ManuGowda: i just removed the old function and now we can use the npm package for the extract. so all good

commander/src/commands/core/upgrade.ts
```diff
@@ -174,7 +175,11 @@ export default class UpgradeCommand extends BaseCommand {
 									cacheDir,
 								);
 
-								await extract(fileDir, fileName, installationPath);
+								await tar.x({
```
shuse2: why not using `extract` now?

commander/src/commands/core/install.ts
shuse2: same as below, why not use `extract`?

commander/src/utils/download.ts
ManuGowda: the problem with using the `hash` function was that its not a buffer.. its stream here.. so

commander/src/commands/core/upgrade.ts
ManuGowda: tar has the extract function directly.. we dont need an extra wrapper just for calling it

commander/src/utils/download.ts
shuse2: we can read the file and then create the hash. we don't need to do it while streaming i think

commander/src/commands/core/upgrade.ts
shuse2: I think you already defined the extra wrapper, and probably it would be easier if we need to change library (like this case) if we keep the wrapper

commander/package-lock.json
ManuGowda: @shuse2 i tried the `npx lerna boostrap` in the root and it dint resolve the package lock issue

commander/src/commands/core/upgrade.ts
ManuGowda: ok ll move it back.

PR #3880, opened 2019-06-27
Add validation of custom transaction - Closes #3877

### LINKED ISSUE #3877: Update BaseTransaction validation in framework
### Expected behavior
After https://github.com/LiskHQ/lisk-sdk/issues/3821, framework will not have sufficient check for the BaseTransaction.
However, we need to make sure that it has executable transaction type when custom transaction is supplied.

### Actual behavior
Currently it's too strict that it only allows exact same instance, and after the issue, it will allow anything and fail on runtime

2.1-

shuse2:
### What was the problem?
Registering custom transaction did not have validation.

### How did I fix it?
Adding validation for
- The first prototype chain is BaseTransaction after generic Function
- It has at least satisfies the public base transaction methods and properties

### How to test it?
From SDK, use Application but use lisk-transactions/BaseTransaction and try to register
or 
`npm t`

### Review checklist

* The PR resolves #3877 
* Linting passes
* Tests pass


framework/src/controller/application.js
```diff
@@ -233,10 +252,31 @@ class Application {
...
+		const publicPropertiesOfBaseTransaction = getPublicProperties(
+			BaseTransaction
+		);
+
+		assert(
+			publicPropertiesOfBaseTransaction.every(prop =>
+				publicPropertiesOfBase.includes(prop)
+			),
+			'Transaction must use compatible BaseTransaction.'
+		);
```
nazarhussain: Suggest to use the same pattern we used to validate other object. `JSON-Schema`.

Define the full JSON-Schema for what you expect a transaction to be look like and then match with it. 

This will help to check the types as well:

https://github.com/epoberezkin/ajv-keywords#types

This package is already in the dependency tree, so safe to use.

nazarhussain:
```js
{
	instanceOf: 'BaseTransaction',
	required: ['apply', 'undo'],
	properties: {
		apply: {
			typeof: 'function',
		},
		undo: {
			typeof: 'function',
		},
	},
};
```

shuse2: `instanceOf` will not work unless we create custom one with instance, but it defeats the purpose of not using exact match.

other than that, I think we can define all the function/properties with types, but I think this is too redundant in a way that all of the type are defined in the TypeScript already.
also, for property of instance, it has it's own validation

therefore, I think checking dynamically make more sense.

nazarhussain: I am still not convinced on doing it this way. Using `Object.getOwnPropertyNames` to match a class object is seems weird and not completely under the hood. Unless we have descriptive details of what we are matching, things will get difficult to understand later. And would also be difficult to refer to user that what a transaction could be, and what are required interfaces to implement and what are optional. If you have a JSON-Schema that will make it very clear and descriptive.

Also the this code  `Object.getPrototypeOf(instance);` and `Base.name === BaseTransaction.name`  is deceptive. It refers to the fact Transaction can be extended by any thing which name is `BaseTransaction`. So if we have the code below then it will pass that check, although its invalid BaseTransaction. 

```
class BaseTransaction(){}
class MyTransaction extends BaseTransaction{}
```

> That all of the type are defined in the TypeScript already.

The above reasoning is not correct as well under the context. User is not bound to use typescript where transpiler check and validate the types. Even if he does type matching, then question will come back to root point, against which BaseTransaction it will match, one which is part of the Framework or one which is in the transactions package. 

I suggest this issue need more thought process to define a correct pattern to handle breaking changes in transaction.

nazarhussain: I would tend to highlight my initial thought here as well. We just need a check that transaction is extended by `BaseTransaction` which `lisk-sdk` supports and exports. All custom transactions should must be extended by that one only. 

If someone have any client side logic or standalone script using `lisk-transactions` only, that's not an issue. But for any custom application and transaction for the application one must and only user the `lisk-sdk` package. 

@MaciejBaj

shuse2: true, we shouldn't need to check if the class extends base transaction or not.
At this point it just requires to have exact same public signature.
1. we cannot check static type in the schema. we need another schema, but it's checked already so it should be ok.
2. we should not check the property, because we don't know when it will be assigned from the class. Also, possible that it requires some parameter in the constructor for some specific custom transaction.

Let's try with schema even though it is redundant.
If it causes too much trouble, I will move it back to dynamic check later time.

framework/test/jest/unit/specs/controller/application.spec.js
```diff
@@ -163,25 +158,17 @@ describe.skip('Application', () => {
...
+			const app = new Application(genesisBlock, config);
 
 			// Act && Assert
 			expect(() => app.registerTransaction()).toThrow(
 				'Transaction implementation is required'
 			);
 		});
 
-		it('should throw error when transaction does not extend BaseTransaction.', () => {
+		it('should throw error when transaction does satisify TransactionInterface.', () => {
```
lsilvs:
```suggestion
		it('should throw error when transaction does not satisfy TransactionInterface.', () => {
```

elements/lisk-transactions/src/utils/validation/schema.ts
```diff
@@ -65,6 +65,64 @@ export const transaction = {
...
+		'undo',
+		'prepare',
+		'addMultisignature',
+		'addVerifiedMultisignature',
+		'isExpired',
+		'fromSync',
+	],
+	properties: {
+		toJSON: {
+			typeof: 'function',
```
nazarhussain: Please add one test case to check this particular keyword `typeof`, or another new keyword specifically used in this schema.

shuse2: Addressed. Good catch =)

shuse2:
Further discussion is moved to #3935

PR #3882, opened 2019-06-27
Maintain received time of last received applied block - Closes #3557

### LINKED ISSUE #3557: The received time of the most recent block in the blockchain should be maintained
### Expected behavior
"Introduce BFT consensus protocol" dictates that every node should maintain the time it received the most recent block.

Maintain in blocks state.
### Actual behavior
The node does not maintain the time at which block is received from the network.
3.0.0

limiaspasdaniel:
### What was the problem?
The node does not maintain the time at which the block is received from the network.
In our codebase, `lastReceipt` was wrongly updated also when forging or syncing. In the new BFT consensus protocol `lastReceipt` is only tied to blocks that are received from the network and applied successfully, and this needs a change in the codebase and the logic responsible of maintaining this information.
### How did I fix it?
`lastReceipt` still holds the time a block was received from the network but no necessarily meaning it was applied successfully.

Furthermore, the following property
```javascript
this._lastReceivedAndAppliedBlock = {
			receivedTime,
			id,
		};
```
now holds the time and id of the last block that was received from the network AND happened to be applied successfully.

Note that now this value is not exposed to other entities in our application because the functions that used to require it will no longer be used, such as `isStale()` 

### How to test it?
Run `fork_choice_rule.spec.js` (Jest unit) and `blocks/blocks.js` (Mocha unit)
### Review checklist

* The PR resolves #3557
* Linting passes
* Tests pass


framework/src/modules/chain/blocks/fork_choice_rule.js
```diff
@@ -1,3 +1,59 @@
...
+
+/**
+ * Wrapper for readability. Returns whether a block was received in its designated
+ * forging slot.
+ * @param slots
+ * @param receivedBlock
+ * @param receivedTime
+ * @return {boolean}
+ */
+const isBlockReceivedWithinForgingSlot = (slots, receivedBlock, receivedTime) =>
```
nazarhussain: If full block not required, suggest to only pass the required attribute as `blockTime`

framework/src/modules/chain/blocks/fork_choice_rule.js
```diff
@@ -1,3 +1,59 @@
...
+ *
+ * IMPORTANT: If the last block applied on the chain was not received from the network
+ * but instead it was explicitly synced or forged, the return value of this function is true
+ * This is evaluated in the first line.
+ * @param slots
+ * @param lastAppliedBlock
+ * @param lastReceivedAndAppliedBlock
+ * @return {boolean}
+ */
+const islastAppliedBlockReceivedWithinForgingSlot = (
```
nazarhussain: This should be renamed to `isLastAppliedBlockReceivedWithinForgingSlot

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -452,21 +488,20 @@ class Blocks extends EventEmitter {
...
-		const newBlockReceivedAt = this.slots.getTime();
+		this._lastReceipt = Math.floor(this.slots.getTime() / 1000);
 
 		// Execute in sequence
 		return this.sequence.add(async () => {
 			this._shouldNotBeActive();
 			this._isActive = true;
-			const result = await this._forkChoiceTask(block, newBlockReceivedAt);
+			const result = await this._forkChoiceTask(block);
 			this._isActive = false;
```
SargeKhan: forkChoiceTask should declare finally block which sets the isActive property to true, as we discussed.

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -505,14 +540,53 @@ class Blocks extends EventEmitter {
...
+				} round: ${this.slots.calcRound(
+					block.height
+				)} slot: ${this.slots.getSlotNumber(block.timestamp)} reward: ${
+					block.reward
+				} version: ${block.version}`
+			);
+			this._lastReceivedAndAppliedBlock.receivedTime = this._lastReceipt;
+			this._lastReceivedAndAppliedBlock.id = block.id;
+			this._lastBlock = newBlock;
+			this._isActive = false;
```
yatki: already setting it to false in finally. We can remove it from here.

framework/test/mocha/unit/modules/chain/blocks/blocks.js
```diff
@@ -888,12 +922,85 @@ describe('blocks', () => {
...
 
 		it('should assign the new last block to _lastBlock and _isActive to false', async () => {
-			await blocksInstance._processReceivedBlock(block);
+			await blocksInstance._processReceivedBlock[block.version](block);
+			expect(blocksInstance._lastBlock).to.equal(block);
+			expect(blocksInstance._isActive).to.be.false;
+		});
+
+		it('should error log that apply the new received block failed with the error', async () => {
+			const error = new Error('This is an error');
```
yatki: Please use // Arrange & Act & Assert comments when test gets bigger. 

You can also move to error message in line 944 to arrange part as a variable, to make assertion easy to read.

framework/test/mocha/unit/modules/chain/blocks/blocks.js
```diff
@@ -888,12 +922,85 @@ describe('blocks', () => {
...
+		it('should call processBlock with the new block', async () => {
+			await blocksInstance._processReceivedBlock[block.version](block);
+			expect(stubs.processBlock).to.be.called;
+		});
+
+		it('should debug log that the block has been successfully applied if so', async () => {
+			await blocksInstance._processReceivedBlock[block.version](block);
+
+			expect(loggerStub.debug).to.be.calledWith(
+				`Successfully applied new received block id: ${block.id} height: ${
```
yatki: same goes here.

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -488,22 +464,27 @@ class Blocks extends EventEmitter {
...
 			this._shouldNotBeActive();
 			this._isActive = true;
-			const result = await this._forkChoiceTask(block);
-			this._isActive = false;
-			return result;
+			try {
+				await this._forkChoiceTask(block);
+			} catch (e) {
+				this.logger.error(e);
+			} finally {
```
yatki: no need to have finally here. We can simply set isActive after try/catch block.

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -161,22 +140,19 @@ class Blocks extends EventEmitter {
...
+		// Remove receivedAt property..
+		const { receivedAt, ...block } = this._lastBlock;
+		return block;
 	}
 
 	get isActive() {
 		return this._isActive;
 	}
 
+	// TODO: Remove after fork
```
yatki: what does remove after fork mean? Could you please elaborate the comment.

framework/src/modules/chain/blocks/fork_choice_rule.js
```diff
@@ -19,35 +19,30 @@ const forgingSlot = (slots, block) => slots.getSlotNumber(block.timestamp);
...
  * @return {boolean}
  */
 const islastAppliedBlockReceivedWithinForgingSlot = (
 	slots,
-	lastAppliedBlock,
-	lastReceivedAndAppliedBlock
+	lastAppliedBlock
 ) => {
-	if (lastAppliedBlock.id !== lastReceivedAndAppliedBlock.id) return true;
+	// If the blockk doesn't have the property `receivedAt` it meants it was forged
```
yatki: typo: `blockk`, `meants`  :D https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker :D 🖖

framework/src/modules/chain/blocks/blocks.js
yatki: ignore this comment bitte.

framework/src/modules/chain/blocks/blocks.js
limiaspasdaniel: When the new block version `2` is activated, the class property `_lastReceipt` will no longer be used or exposed, therefore we will be able to remove it.
Notice how `receiveBlockFromNetworkV2` now doesn't use this attribute whatsoever

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -64,7 +64,37 @@ class Blocks extends EventEmitter {
...
+		};
+
+		/**
+		 * Represents the receipt time of the last block that was received
+		 * from the network.
+		 * Doesn't necessarily mean that it was applied.
+		 * @type {number}
+		 * @private
+		 */
 		this._lastReceipt = null;
```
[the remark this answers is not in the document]
limiaspasdaniel: After discussion with @SargeKhan we no longer need this attribute and instead `lastBlock` will have an attribute `receivedAt` if the block was received from the network. Its `getter` will remove the attribute upon retrieval. This greatly simplifies the logic and allows `_lastReceipt` to be removed after the next chain upgrade.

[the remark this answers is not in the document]
limiaspasdaniel: please see https://github.com/LiskHQ/lisk-sdk/pull/3882#discussion_r299852116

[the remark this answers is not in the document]
limiaspasdaniel: Please see https://github.com/LiskHQ/lisk-sdk/pull/3882#discussion_r299852116

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -131,6 +161,12 @@ class Blocks extends EventEmitter {
 			1: block => this._receiveBlockFromNetworkV1(block),
 			2: block => this._receiveBlockFromNetworkV2(block),
 		};
+
+		this._processReceivedBlock = {
+			0: block => this._processReceivedBlockV1(block),
+			1: block => this._processReceivedBlockV1(block),
+			2: block => this._processReceivedBlockV2(block),
```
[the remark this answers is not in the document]
limiaspasdaniel: After https://github.com/LiskHQ/lisk-sdk/pull/3882#discussion_r299852116 we only have one implementation of _processReceivedBlock so this is no longer a problem.

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -505,14 +540,53 @@ class Blocks extends EventEmitter {
 		}
 	}
 
+	// PRE: Block has been validated and verified before
+	async _processReceivedBlockV2(block) {
+		try {
+			const newBlock = await this.blocksProcess.processBlock(
```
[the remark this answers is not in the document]
limiaspasdaniel: Solved after discussion with @SargeKhan https://github.com/LiskHQ/lisk-sdk/pull/3882#discussion_r299852116

framework/src/modules/chain/blocks/blocks.js
limiaspasdaniel: Solved.

framework/src/modules/chain/blocks/fork_choice_rule.js
limiaspasdaniel: I don't use VSCode hehe but thanks for the good catch. Solved :)

framework/test/mocha/unit/modules/chain/blocks/blocks.js
limiaspasdaniel: Done

framework/test/jest/unit/specs/modules/chain/blocks/fork_choice_rule.spec.js
```diff
@@ -89,35 +89,59 @@ describe('Fork Choice Rule', () => {
...
-				timestamp: 5000,
+				timestamp: lastReceivedAndAppliedBlock.receivedTime,
 				generatorPublicKey: 'abc',
+				receivedAt: 300000,
 			};
-			const current = {
-				height: last.height,
-				prevotedConfirmedUptoHeight: last.prevotedConfirmedUptoHeight,
-				previousBlock: last.previousBlock,
+			const receivedBlock = {
```
yatki:
```js
const receivedBlock = {...lastAppliedBlock, id: '2', timestamp: 200000, receivedAt: 200000} 
```

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -452,23 +464,27 @@ class Blocks extends EventEmitter {
 
 	async _receiveBlockFromNetworkV2(block) {
 		// Current time since Lisk Epoch
-		// Better to do it here rather than in the Sequence so receiving time is more accurate
-		const newBlockReceivedAt = this.slots.getTime();
+		block.receivedAt = Math.floor(this.slots.getTime() / 1000);
```
nazarhussain: Please use `slots.getSlotTime()` which will make it consistent to the forging timestamp. So what ever value it returns in seconds or miliseconds it will be consistent across code.

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -452,23 +464,27 @@ class Blocks extends EventEmitter {
...
 
 		// Execute in sequence
 		return this.sequence.add(async () => {
 			this._shouldNotBeActive();
 			this._isActive = true;
-			const result = await this._forkChoiceTask(block, newBlockReceivedAt);
-			this._isActive = false;
-			return result;
+			try {
+				await this._forkChoiceTask(block);
```
shuse2: this was throwing before, but it should be caught?

framework/src/modules/chain/blocks/fork_choice_rule.js
```diff
@@ -1,3 +1,51 @@
...
+ * @param lastAppliedBlock
+ * @return {boolean}
+ */
+const isLastAppliedBlockReceivedWithinForgingSlot = (
+	slots,
+	lastAppliedBlock
+) => {
+	// If the block doesn't have the property `receivedAt` it meants it was forged
+	// or synced, therefore we assume it was "received in the correct slot"
+	if (!lastAppliedBlock.receivedAt) return true;
```
shuse2: prefer
```
if (!lastAppliedBlock.receivedAt) {
  return true;
}
```
or 
```
return !lastAppliedBlock.receivedAt ? true : isBlockReceivedWithinForgingSlot(slots, lastAppliedBlock)
```

framework/src/modules/chain/blocks/blocks.js
limiaspasdaniel: I wonder what happens when an error is thrown in a Sequence? Do you have more info about the expected behavior?

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -452,23 +464,27 @@ class Blocks extends EventEmitter {
 
 	async _receiveBlockFromNetworkV2(block) {
 		// Current time since Lisk Epoch
-		// Better to do it here rather than in the Sequence so receiving time is more accurate
-		const newBlockReceivedAt = this.slots.getTime();
+		block.receivedAt = this.slots.getSlotTime(this.slots.getSlotNumber());
```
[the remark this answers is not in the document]
limiaspasdaniel: We want to check if the block was received within a slot time window. If the block was received in the next slot the time would've been set to the first second of the next slot and the condition would be evaluated correctly.

framework/src/modules/chain/blocks/fork_choice_rule.js
limiaspasdaniel: Done :)

framework/src/modules/chain/blocks/blocks.js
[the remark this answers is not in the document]
SargeKhan: Setting the receivedAt time to the slot beginning time does not add any value. However, it adds a little bit of obscurity, at least for me, because it goes against the name of the property. Which is `receivedAt`. I think if we want to keep the value as the beginning of the round then perhaps property name requires an update.

framework/src/modules/chain/blocks/blocks.js
shuse2: If error happens within the sequence, outside promise will return `Promise.reject(err)`.
I doubled check the V1 implementation, and it's correct to catch the error 🙏

PR #3883, opened 2019-06-27
Update peer selection method for fetchPeers response - Closes #3333

### LINKED ISSUE #3333: Update peer discovery messages in response to fetching peers list
### Description

Update method that selects and returns its peer list in response to `list` rpc call that will be renamed to `getPeers`. As of now, it selects 100 peers and returns it. But according to , it should be implemented as per below.

`getPeers` RPC is used to query a connected peer for a list of known IP addresses of Lisk nodes. In response, the peer sends a message containing up to 1000 IP addresses. 
The IP addresses are selected as follows:

- Let `n` be a random number in the interval,` n = [min(1000, #known peers *0.25), min(1000, #known peers*0.5)]`
- where known peers are any peers in the `triedPeers` or `newPeers` collection

- `select max(n,min(100, #known peers))` addresses at random from all known peers, i.e., peers in the `triedPeers` or `newPeers` collection

- The list returned is shuffled for security reasons

Note: IP address of peers that declared that their IP address should not be advertised (see advertiseAddress flag defined in the next subsection) are never returned.

mitsujutsu:
### What was the problem?

Original implementation of peer selection in response to fetch peers simply selecting 100 random peers from known peers list for any given node.

### How did I fix it?

RPC will only return a fraction of the known peers (25 % to 50 % of the peers) so that it is difficult to know which peers a Lisk node knows about. For instance, an attacker trying to poison the newPeers collection with a lot of IP addresses from malicious nodes cannot easily find out how many IP addresses were actually added to that collection. Furthermore, the list returned is shuffled for security reasons to ensure that it is impossible to infer the currently connected peers from the list.

### Review checklist

* The PR resolves #3333 #3923 
* Linting passes
* Tests pass


shuse2:
@jondubois @mitsujutsu 
Since this would be breaking change, probably we should merge into different branch, or wait for 2.3 to be merge back to development?

elements/lisk-p2p/src/p2p.ts
```diff
@@ -722,8 +721,20 @@ export class P2P extends EventEmitter {
...
 	private _handleGetPeersRequest(request: P2PRequest): void {
-		// TODO later: Remove fields that are specific to the current Lisk protocol.
-		const peers = this._pickRandomPeers(MAX_PEER_LIST_BATCH_SIZE).map(
+		// TODO: Get this from peerbook
+		const knownPeers = [
+			...this._newPeers.values(),
+			...this._triedPeers.values(),
+		];
+		/* tslint:disable no-magic-numbers*/
+		const min = Math.ceil(Math.min(1000, knownPeers.length * 0.25));
```
ishantiw: do you think we can have a const defined for 1000 and 100 if it makes sense? otherwise maybe its unnecessary

mitsujutsu: I thought about it but there are 2 other numbers which would necessitate a total 4 new const. I am not sure if there would be much value added.

mitsujutsu:
> @jondubois @mitsujutsu
> Since this would be breaking change, probably we should merge into different branch, or wait for 2.3 to be merge back to development?

Is it breaking if only the length/order of returned peer list is changed?

mitsujutsu:
> LGTM. Is it possible to cover the behaviour with an integration test?

It is difficult to do in an integration test as we would need to launch a network with > 100 nodes to properly test the behavior. Nodes with less than 100 known peers will simply send all known peers.

elements/lisk-p2p/test/integration/p2p.ts
```diff
@@ -1847,4 +1847,130 @@ describe('Integration tests for P2P library', () => {
 			it.skip('should not evict peers with high responseRate', async () => {});
 		});
 	});
+
+	describe('Connected network: Minimum peer discovery threshold is set to 1', () => {
```
diego-G: Please, try to merge both scenarios into one single describe. Perhaps even add them to one of the existing discovery describes.

elements/lisk-p2p/test/integration/p2p.ts
```diff
@@ -1847,4 +1847,130 @@ describe('Integration tests for P2P library', () => {
...
+					.filter(p2p => p2p.isActive)
+					.map(async p2p => await p2p.stop()),
+			);
+			await wait(100);
+		});
+
+		it('should return list of peers with at most the minimum discovery threshold', async () => {
+			const firstP2PNode = p2pNodeList[0];
+			const { newPeers } = firstP2PNode.getNetworkStatus();
+			expect(newPeers.length).to.eql(1);
```
diego-G: Please, convert this to a variable and try to use a higher number like 3 or 4. Then use `atMost` instead of `eql`.

elements/lisk-p2p/test/integration/p2p.ts
```diff
@@ -1847,4 +1847,130 @@ describe('Integration tests for P2P library', () => {
...
+					.filter(p2p => p2p.isActive)
+					.map(async p2p => await p2p.stop()),
+			);
+			await wait(100);
+		});
+
+		it('should return list of peers with less than maximum discovery response size', async () => {
+			const firstP2PNode = p2pNodeList[0];
+			const { newPeers } = firstP2PNode.getNetworkStatus();
+			expect(newPeers.length).to.be.lessThan(3);
```
diego-G: As above, convert this to a variable. You can reuse it in the title and the P2PConfig object.

elements/lisk-p2p/test/integration/p2p.ts
mitsujutsu: Both scenarios require new configurations which could alter behavior of nodes. Therefore it wouldn't fit into other previous describes. I can put both describe blocks in a wrapper describe.

jondubois:
Aside from the new merge conflict, this PR looks good but it's been put on hold to allow us to focus on the `v2.3` release. This should be part of `v3.0` release.

PR #3890, opened 2019-06-28
Improve create and sign any transaction functionality in Lisk Transactions - Closes #3739

### LINKED ISSUE #3739: Implement create and sign any transaction functionality in Lisk Transactions
### Expected behavior
There should be functionality in Lisk Transactions allowing to create and sing any transaction - passing that transaction as a parameter. The functionality should verify only the required properties of transactions (common for all of the transactions):
```
- receipientId/receipentPublicKey,
- fee
- passphrase 
```
For that property set, the function creates and signs a transaction, adds the following properties:
```
- receipientId (if receipentPublicKey was passed),
- signature
- senderPublicKey (derived
- senderId (derived from the senderPublicKey)
- timestamp
- amount (0 by default)
- type
```
The output is ready to send over the `POST /api/transactions` API.
### Actual behavior
Missing functionality.

`2.0.0`

lsilvs:
### What was the problem?
There were no easy way to sign and get a transaction ready to be broadcasted using `BaseTransaction`

### How did I fix it?
By setting default values and getting the `senderId` and `senderPublicKey` from the passphrase during `sign` method. The only required property is `recipientId`.

### How to test it?
Implement a custom transaction extending from BaseTransaction and then:
```
const tx = new MyCustomTransaction({});
tx.sign(passphrase);
tx.stringify();
```

### Review checklist

* The PR resolves #3739
* Linting passes
* Tests pass


framework/src/modules/http_api/controllers/node.js
```diff
@@ -326,4 +317,27 @@ async function _getNetworkHeight() {
 	return parseInt(networkHeight);
 }
 
+/**
+ * Parse transaction instance to raw data
+ *
+ * @returns Object
+ * @private
+ */
+function _parseTransaction(transaction) {
```
pablitovicente: maybe improve name is it parses and returns a new object?

lsilvs: Any suggestion?

pablitovicente: `normalizeTransaction` maybe?

PR #3891, opened 2019-06-28
Implement method to identify chain recovery mode with BFT - Closes #3561

### LINKED ISSUE #3561: Implement method to identify chain recovery mode with BFT
### Expected behavior
The chain recovery mechanism is triggered based on the conditions specified in the "Introduce BFT consensus protocol" LIP. This specification should be implemented in the framework.

Write an algorithm to decide whether:
- Fast chain switch mechanism
- Synchronization

### Actual behavior
This functionality is non-existent in the current state.
3.0.0

nazarhussain:
### What was the problem?
The implementation responsible to identify which chain recovery mechanism should be used was pending. 

### How did I fix it?
Implemented as per the LIP.

https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#moving-to-a-different-chain

### Review checklist

* The PR resolves #3561
* Linting passes
* Tests pass


framework/src/modules/chain/blocks/blocks.js
```diff
@@ -74,6 +76,7 @@ class Blocks extends EventEmitter {
 		this.genesisBlock = genesisBlock;
 		this.interfaceAdapters = interfaceAdapters;
 		this.slots = slots;
+		this.bft = bft;
```
shuse2: this is circular dependency

framework/src/modules/chain/bft/utils.js
```diff
@@ -16,6 +16,15 @@
...
 const { validate } = require('../../../../src/controller/validator');
+const {
+	verifySignature,
+	verifyVersion,
+	verifyReward,
+	verifyId,
+	verifyPayload,
+	verifyForkOne,
+	verifyBlockSlot,
+} = require('../blocks/verify');
```
shuse2: under `blocks/` should not be used outside of the `blocks` module

framework/src/modules/chain/bft/utils.js
```diff
@@ -26,6 +35,38 @@ const { validate } = require('../../../../src/controller/validator');
...
 
+/**
+ * Perform all checks outlined in Step 1 of the section "Processing Blocks" except for checking height, parentBlockID and delegate slot (block B may be in the future and assume different delegates that are not active in the round of block A). If any check fails, the peer that sent block B is banned and the node aborts the process of moving to a different chain.
+ *
+ * https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#moving-to-a-different-chain
+ *
+ * @param lastBlock
+ * @param block
+ */
+const verifyBlockForChainSwitching = (lastBlock, block) => {
```
shuse2: probably this is not responsibility of bft

nazarhussain: Its BFT to decide for chain switching, so bft have its own validation to do it.

framework/src/modules/chain/synchronizer/synchronizer.js
```diff
@@ -1,44 +1,162 @@
 const BlockSynchronizationMechanism = require('./block_synchronization_mechanism');
 const FastChainSwitchingMechanism = require('./fast_chain_switching_mechanism');
```
limiaspasdaniel: As discussed, in my opinion, these dependencies should be instantiated in `Chain.js` and registered into Synchronizer there. Doing the way it is right now will make Synchronizer dependencies grow if one of the mechanisms needs more of them.

framework/src/modules/chain/synchronizer/synchronizer.js
```diff
@@ -1,44 +1,162 @@
...
+		const blockRound = this.slots.calcRound(receivedBlock.height);
+		const delegateList = await this.dpos.getRoundDelegates(blockRound);
+		if (delegateList.includes(receivedBlock.generatorPublicKey)) {
+			return this.fastChainSwitchingMechanism;
+		}
+
+		return null;
+	}
+
+	async lastBlock() {
```
limiaspasdaniel:
```suggestion
	/**
	* @private
	*/
	async _lastBlock() {
```

framework/src/modules/chain/synchronizer/synchronizer.js
nazarhussain: If we initiate these in the chain, we have to register logic dynamically when to choose which strategy. Which I feel is not bad idea either, but considering that for reasonable time in future we are not going to introduce new sync mechanism, this approach is fine for now.

limiaspasdaniel: I understand, however, the purpose of having a Synchronizer on top was to isolate implementation details from other parts of the code. For me even if it is not fully complete, it makes more sense to:

1. Instantiate both mechanism classes with its corresponding dependencies
2. Register them to the Synchronizer, so the Synchronizer shouldn't have to include the dependencies of the mechanisms registered in it. If in the future (possibly), we improve or change a mechanism, we might have to add/remove dependencies and that will mean we will have to modify the upper Synchronizer class, this will make the purpose of the Synchronizer empty.

IMHO it makes more sense to do it this way since the beginning than have it to refactor it later on.

nazarhussain: As said earlier, If I see even smaller percent of chance to have new synchronisation mechanism or any SDK feature to register custom sync mechanism, I would have gone with the direction you suggested. 

But for now we have only two and we keep go with these two and not exposing it as feature for the SDK to register custom sync mechanism. @MaciejBaj Can suggest if we plan for any such feature for future. 

For suggested above reasons, I think with current approach. If we plan for such feature for SDK, then we will brainstorm further on how to expose the sync selection logic for dynamic use.

limiaspasdaniel: IMO, it’s not meant to be exposed to the higher layer of the SDK for now but to ease future development on both mechanisms and to simplify its design.

As I said, doing it the way you propose will induce extra development if any of the dependencies of each mechanism change in the future (which we don’t know, and doesn’t necessarily mean we add more mechanisms) and Synchronizer will have to be modified. In my view, doing it the way I propose since the beginning can save us time in the future and the difference in implementation is very simple, just registering the mechanisms to the Synchronizer in Chain.js.

I find more valuable to go the way I propose for the reasons I stated than having to revisit this in the future.

nazarhussain: I don't see the difference of maintaining effort for both approaches. If there is a change, you have to maintain equally in both cases. 

Will see if I can came up with a middle ground.

framework/src/modules/chain/synchronizer/fast_chain_switching_mechanism.js
```diff
@@ -1,14 +1,59 @@
...
+
+	/**
+	 * Check if this sync mechanism is valid for the received block
+	 *
+	 * @param {object} receivedBlock - The blocked received from the network
+	 * @return {Promise.<Boolean|undefined>} - If the mechanism applied to received block
+	 * @throws {Error} - In case want to abort the sync pipeline
+	 */
+	async isValidFor(receivedBlock) {
+		const lastBlock = await this.storage.entities.Block.getLastBlock();
```
shuse2: it shouldn't access the database to ask for the last block. It should ask the blocks module

framework/src/modules/chain/synchronizer/fast_chain_switching_mechanism.js
```diff
@@ -1,14 +1,59 @@
...
+
+		// 3. Step: Check whether B justifies fast chain switching mechanism
+		const TWO_ROUNDS = this.constants.activeDelegates * 2;
+		if (Math.abs(receivedBlock.height - lastBlock.height) > TWO_ROUNDS) {
```
shuse2: In case of receivedBlock = 100 and lastBlock = 150, should also trigger the sync?

framework/src/modules/chain/synchronizer/fast_chain_switching_mechanism.js
```diff
@@ -1,14 +1,59 @@
...
+	 *
+	 * @param {object} receivedBlock - The blocked received from the network
+	 * @return {Promise.<Boolean|undefined>} - If the mechanism applied to received block
+	 * @throws {Error} - In case want to abort the sync pipeline
+	 */
+	async isValidFor(receivedBlock) {
+		const lastBlock = await this.storage.entities.Block.getLastBlock();
+
+		// 3. Step: Check whether B justifies fast chain switching mechanism
+		const TWO_ROUNDS = this.constants.activeDelegates * 2;
```
shuse2: this is not constant, so it should be camel case

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism/block_synchronization_mechanism.spec.js
```diff
@@ -14,130 +14,236 @@
...
 
-describe('BlockSynchronizationMechanism', () => {
-	const stubs = {
-		channel: {
-			invoke: jest.fn(),
-		},
-		modules: {
+describe('block_synchronization_mechanism', () => {
+	afterEach(async () => {
+		jest.clearAllMocks();
```
shuse2: isn't this automatic? @yatki

framework/test/jest/unit/specs/modules/chain/synchronizer/synchronizer.spec.js
```diff
@@ -0,0 +1,376 @@
...
+	logger: loggerMock,
+	exceptions: {},
+	blockReward: {},
+	maxTransactionsPerBlock,
+	maxPayloadLength,
+};
+
+describe('synchronizer', () => {
+	afterEach(async () => {
+		jest.clearAllMocks();
```
shuse2: same as above

framework/test/jest/unit/specs/modules/chain/synchronizer/synchronizer.spec.js
```diff
@@ -0,0 +1,376 @@
...
+const {
+	Block: blockFixture,
+} = require('../../../../../../mocha/fixtures/blocks');
+
+jest.mock('../../../../../../../src/modules/chain/blocks/verify');
+
+const maxPayloadLength = 10000;
+const maxTransactionsPerBlock = 25;
+
+const storageMock = {
```
shuse2: setup should be in the `describe`

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -83,6 +84,7 @@ class Blocks extends EventEmitter {
 		this.genesisBlock = genesisBlock;
 		this.interfaceAdapters = interfaceAdapters;
 		this.slots = slots;
+		this.synchronizer = synchronizer;
```
shuse2: blocks should not reference synchronizer imo. Dependency should be opposite

framework/src/modules/chain/synchronizer/synchronizer.js
```diff
@@ -14,47 +14,175 @@
...
 		}
 
-		this.currentStrategy = this._determineStrategy();
+		// Choose the right mechanism to sync
+		const validMechanism = await this._determineSyncMechanism(receivedBlock);
 
-		return this.currentStrategy.run();
+		if (!validMechanism) {
+			return this.logger.info(
+				"Can't determine sync mechanism at the moment for block",
```
limiaspasdaniel:
```suggestion
				"Sync mechanism could not be determined for the given block",
```

I think this way is clearer

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism/block_synchronization_mechanism.spec.js
```diff
@@ -14,130 +14,236 @@
...
+	});
+
+	describe('BlockSynchronizationMechanism', () => {
+		const activeDelegates = 101;
```
limiaspasdaniel: This shouldn't be hardcoded by instead retrieved from constants/config

framework/test/jest/unit/specs/modules/chain/synchronizer/fast_chain_switching_mechanism/fast_chain_switching_mechanism.spec.js
```diff
@@ -0,0 +1,162 @@
...
+const {
+	FastChainSwitchingMechanism,
+} = require('../../../../../../../../src/modules/chain/synchronizer/fast_chain_switching_mechanism');
+const {
+	Block: blockFixture,
+} = require('../../../../../../../mocha/fixtures/blocks');
+
+describe('fast_chain_switching_mechanism', () => {
+	afterEach(async () => {
+		jest.clearAllMocks();
```
limiaspasdaniel: Jest already clears all mocks after each run so this line can be removed

framework/test/jest/unit/specs/modules/chain/synchronizer/fast_chain_switching_mechanism/fast_chain_switching_mechanism.spec.js
```diff
@@ -0,0 +1,162 @@
...
+	});
+
+	describe('FastChainSwitchingMechanism', () => {
+		const activeDelegates = 101;
```
limiaspasdaniel: Same as the previous comments, this number should be retrieved from constants/config.

framework/test/jest/unit/specs/modules/chain/synchronizer/synchronizer.spec.js
limiaspasdaniel: Same here

framework/test/jest/unit/specs/modules/chain/synchronizer/synchronizer.spec.js
```diff
@@ -0,0 +1,376 @@
...
+};
+
+describe('synchronizer', () => {
+	afterEach(async () => {
+		jest.clearAllMocks();
+	});
+
+	describe('Synchronizer', () => {
+		let sync;
+		let sync1;
```
limiaspasdaniel: Perhaps we could rewrite this variable names to `oneSynchronizer` and `anotherSynchronizer`

framework/src/modules/chain/blocks/blocks.js
limiaspasdaniel: In that case we could fire an event instead of injecting the dependency.

framework/test/jest/unit/specs/modules/chain/synchronizer/fast_chain_switching_mechanism/fast_chain_switching_mechanism.spec.js
nazarhussain: It didn't cleared so why I have to add it here.

framework/test/jest/unit/specs/modules/chain/synchronizer/synchronizer.spec.js
nazarhussain: It didn't cleared so why I have to add it here.

framework/test/jest/unit/specs/modules/chain/synchronizer/fast_chain_switching_mechanism/fast_chain_switching_mechanism.spec.js
limiaspasdaniel: @yatki do you know why is this happening?

framework/src/modules/chain/synchronizer/fast_chain_switching_mechanism.js
nazarhussain: Why we add dependency of `blocks` module here just to get the last block, which we can easily get from the storage and storage is already here.

framework/src/modules/chain/synchronizer/fast_chain_switching_mechanism.js
nazarhussain: https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#moving-to-a-different-chain
> Check whether |B.height-A.height|<=202. If the checks fails, abort the process of moving to a different chain.

With this definition if gap is more than 2 rounds we will abort. In the example you shared the gap is just 50 blocks so we will trigger the fast sync.

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism/block_synchronization_mechanism.spec.js
nazarhussain: It not clearing the mocks automatically, so I have to add these lines.

framework/test/jest/unit/specs/modules/chain/synchronizer/synchronizer.spec.js
nazarhussain: Same as above.

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism/block_synchronization_mechanism.spec.js
nazarhussain: In jest structure there is no feature to get a compiled configuration and constants. So if you even look at all other tests in the Jest directory you will find hardcoded values everywhere. We may update it but not in scope of this PR.

framework/test/jest/unit/specs/modules/chain/synchronizer/fast_chain_switching_mechanism/fast_chain_switching_mechanism.spec.js
nazarhussain: Same as above.

framework/src/modules/chain/blocks/blocks.js
nazarhussain: BTW I have no idea, how this change appears in this PR. As I didn't made it.

framework/src/modules/chain/synchronizer/fast_chain_switching_mechanism.js
shuse2: we should only get the last block from the blocks module because it's managed in memory.
when to update last block is the responsibility of `blocks`
This problem happened already multiple times, so let's consistently use the blocks

nazarhussain: Our logic here does not refer to something in memory. We need to get last persisted block, for which storage is the right source of truth. As said, I don't want to add a new dependency to blocks just to fetch last persisted block, which is already can be done through storage.

framework/src/modules/chain/synchronizer/synchronizer.js
```diff
@@ -14,47 +14,175 @@
...
+					this.activeMechanism.constructor.name
+				} is already running`
+			);
+		}
+
+		const lastBlock = await this.storage.entities.Block.getLastBlock();
+
+		// Moving to a Different Chain
+		// 1. Step: Validate new tip of chain
+		const result = this._verifyBlockBeforeSync(lastBlock, receivedBlock);
```
SargeKhan: The checks included in this step are validation steps according to the definition of validate and verify, so i think we should use validate here.

framework/src/modules/chain/synchronizer/synchronizer.js
```diff
@@ -14,47 +14,175 @@
...
+	async run(receivedBlock) {
+		if (this.activeMechanism) {
+			throw new Error(
+				`Blocks Sychronizer with ${
+					this.activeMechanism.constructor.name
+				} is already running`
+			);
+		}
+
+		const lastBlock = await this.storage.entities.Block.getLastBlock();
```
SargeKhan: Here as well, i think we should get the lastBlock from the blocks module

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -846,6 +845,27 @@ class Blocks extends EventEmitter {
...
+	/**
+	 * Handle the fast switching mechanism
+	 *
+	 * @param lastBlock
+	 * @param block
+	 * @return {Promise<void>}
+	 * @private
+	 */
+	// eslint-disable-next-line no-unused-vars, class-methods-use-this, no-empty-function
+	async _handleFastChainSwitching(lastBlock, block) {}
```
SargeKhan: what do we mean by handle here?

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -846,6 +845,27 @@ class Blocks extends EventEmitter {
...
+
+	/**
+	 * Handle chain switching via sync
+	 * @param lastBlock
+	 * @param block
+	 * @return {Promise<void>}
+	 * @private
+	 */
+	// eslint-disable-next-line no-unused-vars, class-methods-use-this, no-empty-function
+	async _handleChainSwitchingViaSync(lastBlock, block) {}
```
SargeKhan: same question here

framework/test/jest/unit/specs/modules/chain/synchronizer/fast_chain_switching_mechanism/fast_chain_switching_mechanism.spec.js
SargeKhan: if it's a constant, it should be at least written in upper case

framework/src/modules/chain/chain.js
```diff
@@ -383,6 +394,37 @@ module.exports = class Chain {
...
+
+		const fastChainSwitchMechanism = new FastChainSwitchingMechanism({
+			storage: this.storage,
+			logger: this.logger,
+			slots: this.slots,
+			dpos: {},
+			activeDelegates: this.options.constants.ACTIVE_DELEGATES,
+		});
+
+		this.synchronizer = new Synchronizer({
```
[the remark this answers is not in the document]
nazarhussain: It was exactly like this before as you described. Modified on request of @limiaspasdaniel

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism/block_synchronization_mechanism.spec.js
limiaspasdaniel: According to @yatki there is a way to overcome this with the current codebase. Could you elaboreate more please @yatki ?

framework/src/modules/chain/chain.js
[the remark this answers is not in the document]
limiaspasdaniel: I think by doing it the way you propose, the single purpose of the Synchronizer gets invalidated. By doing it the way @nazarhussain is doing it and after long discussing on how to do it, this way, Synchronizer is not populated with too coupled logic and coupled dependencies that come from each individual mechanism. From my point of view, Synchronizer shouldn't be aware of the dependencies of the mechanisms registered to it. This makes it independent, separates its concerns from the ones from its registered mechanisms and improves the readability of the code, as one could simply read the bootstrap initialization logic and register and implement a new synchronization mechanism in an easy way.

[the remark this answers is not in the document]
limiaspasdaniel: IMO firing an event from Blocks is not incompatible with the current Synchronizer implementation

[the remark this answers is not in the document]
nazarhussain: I tend to agree with @limiaspasdaniel If we put all together in Synchronizer then we simply need two functions rather two distinct isolation for different mechanisms.

framework/src/modules/chain/blocks/blocks.js
limiaspasdaniel: `_handle` follows the same naming consistency across all logic that gets triggered as a result of the fork choice rules

framework/src/modules/chain/blocks/blocks.js
limiaspasdaniel: Same as https://github.com/LiskHQ/lisk-sdk/pull/3891#discussion_r308749002

framework/src/modules/chain/synchronizer/synchronizer.js
nazarhussain: If you see functions inside `verifySignature`, `verifyVersion`, `verifyReward`, `verifyId`, `verifyPayload`, then the name make sense. Otherwise we would have `validate*` for all other functions as well.

framework/src/modules/chain/chain.js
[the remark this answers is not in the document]
SargeKhan: Maybe I am mistaken but for me, if the synchronizer class constructor initializes two synchronization mechanisms and using the same `isValidFor` interface to choose b/w the two strategies then no strong coupling will be introduced. That being said, I am okay with them initialized from the chain directly.
Although, I would still prefer triggering the synchronizer from the blocks module by firing event rather than declaring the `_handleFastChainStrategy` and `_handleSynchronizationStrategy` funcitons within block processing

framework/src/modules/chain/blocks/blocks.js
SargeKhan: Isn't the handling part going to be implemented in each synchronization mechanism?

nazarhussain:
@shuse2 @SargeKhan @limiaspasdaniel Can you guys suggest event name to trigger from `blocks` for synchronisation. I thought but felt its an action not an event in reference to blocks.

framework/src/modules/chain/chain.js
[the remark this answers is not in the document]
limiaspasdaniel: If the Synchronizer initialized the mechanisms within its constructor, it means it needs to have those dependencies available to be passed down to those mechanisms, hence having a strong coupling to them, at least dependency-wise. However, I agree with your last point of sending an event from the `blocks` module.

framework/src/modules/chain/blocks/blocks.js
limiaspasdaniel: Agreed, I thought this was old code but it is indeed introduced here. @nazarhussain could you elabore more on why we need this here?

framework/src/modules/chain/synchronizer/synchronizer.js
SargeKhan: I agree it's being done incorrectly for block validation because it is an old code. But recently we had strictly defined the distinction b/w validate and verify which can be seen when we are checking transactions. There we group all nondb related checks in the validate step and db related checks in the verify/apply steps.

SargeKhan: Hence we should follow the new approach

nazarhussain: Is there any issue created to refactor all code for block verification? If not then we should create it and also make refactor this as well under same issue.

framework/test/jest/unit/specs/modules/chain/synchronizer/fast_chain_switching_mechanism/fast_chain_switching_mechanism.spec.js
nazarhussain: Its passed as function argument and argument name is exactly `activeDelegates`

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -32,6 +32,7 @@ const EVENT_NEW_BLOCK = 'EVENT_NEW_BLOCK';
 const EVENT_DELETE_BLOCK = 'EVENT_DELETE_BLOCK';
 const EVENT_BROADCAST_BLOCK = 'EVENT_BROADCAST_BLOCK';
 const EVENT_NEW_BROADHASH = 'EVENT_NEW_BROADHASH';
+const EVENT_FORK_DETECTED = 'EVENT_FORK_DETECTED';
```
SargeKhan: This is not correct as this event is only triggered in the case when a block with priority has detected. Maybe something like: `EVENT_PRIORITY_CHAIN_DETECTED` would be a more accurate name for this event?

nazarhussain: @SargeKhan  I asked earlier for that name yesterday, I wish you had told this earlier.

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism/block_synchronization_mechanism.spec.js
yatki: @nazarhussain, you can avoid using this call by simply moving mock definitions into beforeEach statements or adding `clearMocks: true` option to jest.config.js. 

Notice the assignment below is in `describe` block. and jest.fn() creates a mock function only once. If you move this assignment into beforeEach you won't need to call `jest.clearAllMocks` because in each hook they'll be reassigned.

```js
	const slotsMock = {
			getSlotNumber: jest.fn(),
		};
```
@shuse2 @limiaspasdaniel I think this also answers your comment above.

framework/test/jest/unit/specs/modules/chain/synchronizer/fast_chain_switching_mechanism/fast_chain_switching_mechanism.spec.js
yatki: @limiaspasdaniel @nazarhussain Because mock file assignments are not in beforeEach blocks. I didn't enable `clearMocks` rule intentionally in jest.config.js to be aware and avoid these kind of assignments. I explained it more above on shu's comment.

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism/block_synchronization_mechanism.spec.js
yatki: @nazarhussain I didn't get why you cannot get the values from config? If getting config function is async, In `before` hook, you can await it, no?

framework/src/modules/chain/synchronizer/synchronizer.js
SargeKhan: I have created an issue for to rename functions for block here https://github.com/LiskHQ/lisk-sdk/issues/4025. Please update the function name here since this is new implementation.

nazarhussain: @SargeKhan  Done. Please review.

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism/block_synchronization_mechanism.spec.js
nazarhussain: So that's actually reassigning the mock objects completely, not clearing old mock values and states. Which I felt completely wrong in reference to clearing mocks. I don't feel its a good idea to create variables with `let` and then assigning their mocks in `beforeEach`. Using `const` with definition and clearing mocks after each is right way.

nazarhussain: @shuse2 @yatki Adding `clearMocks: true` in the jest config works for these tests then 7 other tests starts failing. So let's add this config and remove `clearAllMocks` as separate issue.

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism/block_synchronization_mechanism.spec.js
nazarhussain: Nothing is impossible to do, its just that we don't have that structure available to load configurator and get the compiled configuration. Hope you are working on those utility functions for jest, when we have it we will replace the hardcoded values. Remember this hardcoded value is not only in this test, already used on several other places in jest.

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism/block_synchronization_mechanism.spec.js
nazarhussain: Here is the issue. https://github.com/LiskHQ/lisk-sdk/issues/4026

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism/block_synchronization_mechanism.spec.js
limiaspasdaniel: Can we have an issue to remember to take care of this then, @nazarhussain ?

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -42,10 +57,32 @@ class BlockSynchronizationMechanism {
...
+	/**
+	 * Check if this sync mechanism is valid for the received block
+	 *
+	 * @param {object} receivedBlock - The blocked received from the network
+	 * @return {Promise.<Boolean|undefined>} - If the mechanism applied to received block
+	 * @throws {Error} - In case want to abort the sync pipeline
+	 */
+	// eslint-disable-next-line no-unused-vars
+	async isValidFor(receivedBlock) {
+		// 2. Step: Check whether current chain justifies triggering the block synchronization mechanism
```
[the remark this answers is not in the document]
nazarhussain: No chain module is not choosing mechanism, its synchroniser calling each of mechanism to choose right one.

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism/block_synchronization_mechanism.spec.js
nazarhussain: Here you go https://github.com/LiskHQ/lisk-sdk/issues/4028

limiaspasdaniel: Thanks!

framework/src/modules/chain/chain.js
```diff
@@ -574,6 +614,20 @@ module.exports = class Chain {
 			}
 		});
 
+		this.blocks.on(EVENT_PRIORITY_CHAIN_DETECTED, ({ block }) => {
+			this.logger.info(
+				'Received EVENT_PRIORITY_CHAIN_DETECTED. Triggering synchronizer.'
+			);
+			this.synchronizer
```
shuse2: why not async await here?

framework/src/modules/chain/components/storage/entities/block.js
```diff
@@ -188,6 +188,16 @@ class ChainBlock extends BlockEntity {
 			{}
 		);
 	}
+
+	/**
+	 * Get the last block from storage
+	 *
+	 * @param {Object} [tx] - Transaction Object
+	 * @return {Promise<BasicBlock|ExtendedBlock, NonSupportedFilterTypeError|NonSupportedOptionError>}
```
shuse2: does it return `error`? probably misleading here because it throws.
probably Error is not necessary here

framework/src/modules/chain/chain.js
nazarhussain: Not sure if use `await` or keep the promise resolved in detached mode as its a long process.

shuse2: ah i got ur point. 👍

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism/block_synchronization_mechanism.spec.js
```diff
@@ -14,130 +14,238 @@
...
+			storage: storageMock,
+			slots: slotsMock,
+			bft: bftMock,
+			activeDelegates,
+		};
+
+		let syncMechanism;
+
+		beforeEach(() => {
+			syncMechanism = new BlockSynchronizationMechanism(syncParams);
```
yatki: Please use Arrange & Act & Assert comments in general, same goes for the tests below.

framework/test/jest/unit/specs/modules/chain/synchronizer/fast_chain_switching_mechanism/fast_chain_switching_mechanism.spec.js
yatki: As I stated above, if you could move jest.fn inside beforeEach blocks, you wouldn't need this call.

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism/block_synchronization_mechanism.spec.js
```diff
@@ -14,130 +14,238 @@
...
-				prevotedConfirmedUptoHeight: 0,
-			};
-
-			const peers = [
-				{
-					lastBlockId: '12343245',
-					prevotedConfirmedUptoHeight: 1,
-					height: 67,
-					ip: '127.0.0.1',
+		const storageMock = {
```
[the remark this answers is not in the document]
nazarhussain: Discussed with @shuse2 already, I don't think recreating mocks every time is right approach, when we have `jest.clearAllMocks` available.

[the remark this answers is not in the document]
nazarhussain: If that would have been the right approach to re-create mocks before each test then there should be a `jest.clearAllMocks` function at all.

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism/block_synchronization_mechanism.spec.js
nazarhussain: The tests are very much clear and each test block is not more than 5-6 lines. No complexity to understand, I don't like such tedious job of adding these comments in each 5-6 lines of code block.

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism/block_synchronization_mechanism.spec.js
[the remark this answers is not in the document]
yatki: If we are creating something that needs to be cleared in `each` test, I think right place to define them in beforeEach hooks. But anyway, that's ok. Just then instead of having the call, we should later replace it by setting `clearMocks: true` in jest config.

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism/block_synchronization_mechanism.spec.js
yatki: These are opinions. That's why we voted and agreed to use them as a team. And that's why I'm trying to remind everyone this each time to create this habit. You are free to ignore my feedbacks.

shuse2: We have agreed on enforcing or making as an option after the testing standard discussion.
Therefore im dismissing this review

Issue #3885, opened 2019-06-27
networkHeight in /api/node/status unreliable


simonmorgenthaler:
### Expected behavior
/api/node/status should return "networkHeight" which is the currently valid height in the network. This height could be used to verify if the own node is on the correct height.

### Actual behavior
networkHeight only rarely contains the correct height in the network. Most often it is 0.

### Steps to reproduce
call /api/node/status mutiple times and check "networkHeight" in the returned data. Most often it is 0. Only rarely it returns the correct height.

### Which version(s) does this affect? (Environment, OS, etc...)
core 2.0.0-rc.0 on testnet

shuse2:
Possibly related to #3881

Issue #3338, opened 2019-04-15
Add missing unit tests for p2p


ishantiw:
### Description

There are lots of tests missing in p2p and peer modules and also few tests in others. Add all the relevant test cases.

Below is the list of modules that are missing test cases.

- [ ] `validation.ts` related issue #1038 
- [ ] `errors.ts` related issue #1043
- [ ] `peer.ts`
- [ ] `peer_pool.ts`
- [ ] `p2p_request.ts`
- [ ] `p2p.ts`

shuse2:
Closing since All of the critical unit tests are already covered

PR #3895, opened 2019-07-01
API misreporting the connected peer count and state property is wrong - Closes #3881 #3885 #3907

### LINKED ISSUE #3881: API misreporting the connected peer count and state property is wrong
### Expected behavior

The connected peer count should be accurate and not fluctuate randomly.

### Actual behavior

When getting the peer count from the API, it misreports the number of connected peers.

This issue may be caused by the fact that `state` is not a native property of `P2PPeerInfo` (it doesn't come from the peer itself); it's something that our own node injects into the P2PPeerInfo instance for each peer. It's possible that nodes are sharing each other's `P2PPeerInfo` list through the discovery process and overriding each other's peer `state` properties.

The `state` property is needed for backwards compatibility with existing 1.x nodes but we should tyr to not rely on it to perform calculations. We should use a different approach for getting connected peers here: https://github.com/LiskHQ/lisk-sdk/blob/4797f8d7c4a4d245e69c8531139b00018da4fc20/framework/src/modules/http_api/controllers/peers.js#L72

^ Either that, or we should come up with a better mechanism for storing the peer state. Maybe like we discussed in the last P2P meeting and add a wrapper around the P2PPeerInfo object to store private properties (the ones that we do not want to share with peers); the `state` could go in there instead of inside the P2PPeerInfo.

### Steps to reproduce

On testnet, check the peer count as reported by the API. The one which comes from here: https://github.com/LiskHQ/lisk-sdk/blob/4797f8d7c4a4d245e69c8531139b00018da4fc20/framework/src/modules/http_api/controllers/peers.js#L81

`2.0.0-rc.0`

ishantiw:
### What was the problem?

The connected peer count should be accurate and not fluctuate randomly.

### How did I fix it?

All the connectedPeers are coming from `PeerPool` that has all the peers that are connected but sometimes during the discovery, PeerPool adds a peer without successfully connecting and that's why the number fluctuates with every discovery round.

### How to test it?

Run the application and hit `/api/peers` endpoint

### Review checklist

* The PR resolves #3881 #3885 #3907
* Linting passes
* Tests pass

### DIFF
--- a/elements/lisk-p2p/src/peer_pool.ts
+++ b/elements/lisk-p2p/src/peer_pool.ts
@@ -432,7 +432,7 @@ export class PeerPool extends EventEmitter {
 	}
 
 	public getConnectedPeers(): ReadonlyArray<Peer> {
-		const peers = [...this._peerMap.values()];
+		const peers = this.getAllPeers();
 
 		return peers.filter(
 			peer =>
--- a/elements/lisk-p2p/src/peer_pool.ts
+++ b/elements/lisk-p2p/src/peer_pool.ts
@@ -427,6 +427,20 @@ export class PeerPool extends EventEmitter {
 		});
 	}
 
+	public getAllConnectedPeerInfos(): ReadonlyArray<P2PDiscoveredPeerInfo> {
+		return this.getConnectedPeers().map(peer => peer.peerInfo);
+	}
+
+	public getConnectedPeers(): ReadonlyArray<Peer> {
+		const peers = [...this._peerMap.values()];
+
+		return peers.filter(
+			peer =>
+				peer.state.outbound === ConnectionState.CONNECTED ||
+				peer.state.inbound === ConnectionState.CONNECTED,
+		);
+	}
+
 	public getAllPeerInfos(): ReadonlyArray<P2PDiscoveredPeerInfo> {
 		return this.getAllPeers().map(peer => peer.peerInfo);
 	}
--- a/elements/lisk-p2p/test/unit/peer_pool.ts
+++ b/elements/lisk-p2p/test/unit/peer_pool.ts
@@ -1,0 +1,105 @@
+/*
+ * Copyright © 2018 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import { expect } from 'chai';
+import { PeerPool } from '../../src/peer_pool';
+import {
+	selectPeersForConnection,
+	selectPeersForRequest,
+	selectPeersForSend,
+} from '../../src/peer_selection';
+import { Peer, ConnectionState } from '../../src/peer';
+import { initializePeerList } from '../utils/peers';
+import { P2PDiscoveredPeerInfo } from '../../src/p2p_types';
+
+describe.only('peerPool', () => {
+	const peerPool = new PeerPool({
+		peerSelectionForConnection: selectPeersForConnection,
+		peerSelectionForRequest: selectPeersForRequest,
+		peerSelectionForSend: selectPeersForSend,
+		sendPeerLimit: 25,
+	});
+
+	describe('#constructor', () => {
+		it('should be an object and instance of PeerPool', () => {
+			return expect(peerPool)
+				.to.be.an('object')
+				.and.be.instanceof(PeerPool);
+		});
+	});
+
+	describe('#getConnectedPeers', () => {
+		const peerList: ReadonlyArray<Peer> = initializePeerList();
+
+		let activePeersInfoList: ReadonlyArray<P2PDiscoveredPeerInfo>;
+		beforeEach(async () => {
+			sandbox.stub(peerList[0], 'state').get(() => {
+				return {
+					inbound: ConnectionState.CONNECTED,
+					outbound: ConnectionState.CONNECTED,
+				};
+			});
+			sandbox.stub(peerList[1], 'state').get(() => {
+				return {
+					inbound: ConnectionState.CONNECTED,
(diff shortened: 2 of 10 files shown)


elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -140,24 +142,33 @@ export class PeerPool extends EventEmitter {
...
 			this.emit(EVENT_DISCOVERED_PEER, peerInfo);
 		};
 
 		this._handlePeerConnect = async (peerInfo: P2PDiscoveredPeerInfo) => {
+			const peerId = constructPeerIdFromPeerInfo(peerInfo);
+			// Add a peer to a list that are successfully connected
+			this._connectedPeers.set(peerId, peerInfo);
+
 			// Re-emit the message to allow it to bubble up the class hierarchy.
 			this.emit(EVENT_CONNECT_OUTBOUND, peerInfo);
```
4miners: Seems redundant as this event is fired just before `EVENT_DISCOVERED_PEER` or am I wrong here?

ishantiw: You are right it's not needed

diego-G: True. Besides, we have to check if the list has that peer before setting it.

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -170,6 +181,9 @@ export class PeerPool extends EventEmitter {
 		};
 		this._handlePeerInfoUpdate = (peerInfo: P2PDiscoveredPeerInfo) => {
 			// Re-emit the error to allow it to bubble up the class hierarchy.
+			const peerId = constructPeerIdFromPeerInfo(peerInfo);
+			// Add a peer to a list that are successfully connected
+			this._connectedPeers.set(peerId, peerInfo);
```
diego-G: Same as above.

elements/lisk-p2p/src/peer_pool.ts
ishantiw: @diego-G we don't have to check explicitly if the peerId exists or not because when we set, if it exists then it will update the value else it will create a new entry for it

ishantiw:
Based on the commit https://github.com/LiskHQ/lisk-sdk/commit/bbb40beb3ea6adb1cf7c02115e4743ec25fee868 in 2.2.0. I have updated my PR because its better to use state value coming from socket itself instead of maintaining a separate connected peer list. @4miners Most of the tests for PeerPool are covered in integration test in P2P because PeerPool is mostly dealing with instances of Peers that has sockets. But in future we want to add unit tests for which I created an issue sometime back https://github.com/LiskHQ/lisk-sdk/issues/3338

elements/lisk-p2p/src/peer_pool.ts
diego-G: @ishantiw would you address @4miners suggestion in this PR?

ishantiw: This PR is not adding new code but fixing the existing code, another thing is that PeerPool doesn't have any unit tests, most of its functionality is covered in integration tests. If I write unit tests for PeerPool then I will have to cover all the tests which will take the time or maybe I can just write unit tests for the functions? Another thing is, a lot of these unit tests will not be there or will change in 2.2 version. That was the reason that @jondubois recommended to have low priority on unit tests

4miners: @ishantiw Yes, you can cover by unit tests the new or modified code only, the rest can be done in a separate issue.

elements/lisk-p2p/test/unit/peer_pool.ts
```diff
@@ -0,0 +1,105 @@
...
+				.and.be.instanceof(PeerPool);
+		});
+	});
+
+	describe('#getConnectedPeers', () => {
+		const peerList: ReadonlyArray<Peer> = initializePeerList();
+
+		let activePeersInfoList: ReadonlyArray<P2PDiscoveredPeerInfo>;
+		beforeEach(async () => {
+			sandbox.stub(peerList[0], 'state').get(() => {
```
4miners: We are adding to the list only connected peers, can we have here all the possible cases? (inbound + outbound disconnected, only inbound connected, etc.)

framework/test/mocha/unit/modules/http_api/controllers/node.js
```diff
@@ -194,6 +196,53 @@ describe('node/api', () => {
...
+			});
+		});
+
+		describe('When network:getPeers returns a blank list of peers', () => {
+			beforeEach(async () => {
+				channelStub.invoke.withArgs('network:getPeers').returns([]);
+				networkHeight = await getNetworkHeight();
+			});
+
+			it('should return correct networkHeight based on majority', async () => {
```
4miners: Incorrect description - there is no majority when the list is empty.

framework/test/mocha/unit/modules/http_api/controllers/node.js
```diff
@@ -194,6 +196,53 @@ describe('node/api', () => {
 		});
 	});
 
+	describe('_networkHeight', () => {
+		const MAJORITY_HEIGHT = 456;
+
+		let networkHeight;
+		describe('When there are set of majority peers with same height', () => {
+			beforeEach(async () => {
+				const defaultPeers = [
```
4miners: Can we test more cases? (1 peer only, another height with count same as the majority, majority height way lower than other non-majority heights, etc.)

ishantiw:
I have added unit tests for both `PeerPool` in P2P lib and `http_api/controller/node.js` related to network height and covering the code change in this PR.

4miners:
@jondubois Please review! 🤗

framework/test/mocha/unit/modules/http_api/controllers/node.js
```diff
@@ -194,6 +200,214 @@ describe('node/api', () => {
...
+					{
+						height: MAJORITY_HEIGHT,
+					},
+					{
+						height: MAJORITY_HEIGHT,
+					},
+					{
+						height: MAJORITY_HEIGHT + 1,
```
4miners: `MAJORITY_HEIGHT` here and `MAJORITY_HEIGHT - 1` makes more sense, but this is fine also.

Issue #3875, opened 2019-06-26
Refactor the `update` function of the Application State to make it generic


limiaspasdaniel:
### Expected behavior
The implementation should be written in a more generic way, allowing the user to update whichever property is available and updateable in the state.

For that, we could define an Application State schema that defines which properties are read-only, their formats, etc. 

`update()` would validate this schema and override these properties if it is allowed with the values passed in the arguments.

`framework/src/controller/application_state.js`

### Actual behavior
The `update` function of the Application State is currently forcing to update certain properties every time is called. There are certain situations where we want to update one property and not the one that is being enforced.

The implementation is too specific and tied to those properties. See: https://github.com/LiskHQ/lisk-sdk/pull/3900#discussion_r299879856

### Which version(s) does this affect? (Environment, OS, etc...)
`feature/development`

pablitovicente:
I would also consider using getter/setters for this instead of an schema.

limiaspasdaniel:
Thanks @pablitovicente. Isn't it easier to maintain a fairly simple and short schema that takes care of all of this at once by using JSON Schema definition that having to use getters and setters individually?

pablitovicente:
But with getter/setters we have kind of proxies if in the future we need to perform additional operations upon an application state change while if we have an schema we can validate and then use branching for doing the same.

jondubois:
I think this is a duplicate of https://github.com/LiskHQ/lisk-sdk/issues/3222 - The P2P team has picked it up for the current sprint. Can we close this issue?

About the schema, currently P2P relies on a flexible schema with custom properties. Also as it is now, the application state will only be updated internally by trusted modules running on the current node so we don't need to do validation.

PR #3900, opened 2019-07-01
Implement peer selection algorithm for block sync mechanism - Closes #3767

### LINKED ISSUE #3767: Implement peer selection algorithm step for Block Sync Mechanism
### Expected behavior
Implement peer selection algorithm for Block Sync Mechanism as stated in [LIP-0014](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#block-synchronization-mechanism)

Requires: #3774 

Request a list of all connected peers from the P2P module including T.height and T.heightPrevoted of the block T at the tip of their chain. From this information, determine a set of bestPeers that are at the tip of the chain according to the fork choice rule as follows:

a) Among all peers, choose the subset of peers with largest `T.heightPrevoted`.

b) From the peers in a) choose those with largest `T.height`.

c) Form the peers in b) choose the largest set which has the same `blockID`. Ties are broken in favor of smaller `blockID`. This set is the set `bestPeers` and we define `B` to be the block at the tip of the chain of this set of peers.

If the following expression evaluates to `false`, the block synchronization mechanism is aborted:

    (A.heightPrevoted<B.heightPrevoted) || (A.height<B.height && A.heightPrevoted==B.heightPrevoted)

`feature/BFT`

limiaspasdaniel:
### What was the problem?
Peer selection algorithm for Block Synchronization Mechanism was not implemented
### How did I fix it?
By implementing the peer selection algorithm defined in LIP-004 step 1 of [Block Synchronization Mechanism](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#block-synchronization-mechanism)
### Review checklist

* The PR resolves #3767
* Linting passes
* Tests pass


framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -1,14 +1,114 @@
...
+	 * @example
+	 *
+	 * Input: [{id: 1, height: 2}, {id: 2, height: 3}, {id: 3, height: 3}]
+	 * Output: [{id: 2, height: 3}, {id: 3, height: 3}]
+	 */
+	// eslint-disable-next-line class-methods-use-this
+	_computeLargestSubsetMaxBy(arrayOfObjects, condition) {
+		const absoluteMax = condition(maxBy(arrayOfObjects, condition));
+		const largestSubset = [];
+		// eslint-disable-next-line no-restricted-syntax
```
yatki: Not part of the PR but we can disable this rule or modify it to ignore `BinaryExpression[operator='of']`.

limiaspasdaniel: As discussed in the meeting we have to come up with a conclusion on this topic :) Resolving for now

framework/src/controller/application_state.js
```diff
@@ -85,16 +85,18 @@ class ApplicationState {
...
 	 */
-	async update({ height, prevotedConfirmedUptoHeight, broadhash }) {
+	async update({ height, prevotedConfirmedUptoHeight, broadhash, blockId }) {
 		assert(height, 'height is required to update application state.');
 		try {
 			const newState = this.state;
 			newState.broadhash = broadhash || this.state.broadhash;
 			newState.prevotedConfirmedUptoHeight =
 				prevotedConfirmedUptoHeight || this.state.prevotedConfirmedUptoHeight;
+			newState.blockId = blockId || this.state.blockId;
```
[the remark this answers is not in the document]
limiaspasdaniel: > maybe we can set default values for the params like?:
> 
> ```js
> async update({ height, prevotedConfirmedUptoHeight, broadhash, blockId = this.state.blockId })
> ```
> 
> same goes for other params.

Agreed, definitely cleaner, thanks.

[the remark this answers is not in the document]
limiaspasdaniel: > Also, I know it's out of this PR's scope, not relevant what you did but the whole update function can be simplified as below ¯\_(ツ)_/¯ :
> 
> ```js
> async update(params) {
>  const newState = {...this.state, ...params};
> __private.state.set(this, newState);
> this.logger.debug('Application state', this.state);	
> await this.stateChannel.publish('app:state:updated', this.state);
> return true;
> }
> ```
> 
> Also, I don't like catching errors just to log them.

Agreed, for that, I created an issue to improve it: #3875

elements/lisk-p2p/src/p2p_types.ts
```diff
@@ -62,6 +62,8 @@ export interface P2PNodeInfo {
 	readonly protocolVersion: string;
 	readonly nethash: string;
 	readonly wsPort: number;
+	readonly blockId: string;
```
nazarhussain: Its too generic, we must make it sepcific for understanding. `lastBlockId`, `finalizedBlockId` or any thing which is specifically been communicated here.

limiaspasdaniel: Agreed, will change it to `lastBlockId`

limiaspasdaniel: Done

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -1,14 +1,116 @@
...
+	 * @return {Array<Object>}
+	 * @private
+	 *
+	 * @example
+	 *
+	 * Input: [{id: 1, height: 2}, {id: 2, height: 3}, {id: 3, height: 3}]
+	 * Output: [{id: 2, height: 3}, {id: 3, height: 3}]
+	 */
+	// eslint-disable-next-line class-methods-use-this
+	_computeLargestSubsetMaxBy(arrayOfObjects, condition) {
```
michielmulders: This function is difficult to understand if you don't know where it is coming from.
Need to specify what this `condition` parameter is: `peer => peer.height`.  (More or less a lambda / anonymous function)

+ split line 104 into two lines as we use condition as an argument and then call it again on the result.
First, find the peer with the max height and then get the `absoluteMax` out.

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism.spec.js
```diff
@@ -0,0 +1,183 @@
...
+			 * c) Form the peers in b) choose the largest set which has the same blockID.
+			 * 		Ties are broken in favor of smaller blockID.
+			 *
+			 * This set is the set bestPeers and we define B to be the block at the
+			 * tip of the chain of this set of peers.
+			 *
+			 * @link https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#block-synchronization-mechanism
+			 */
+
+			it('should successfully select the best peer according to the steps defined in LIP-0014', () => {
```
SargeKhan: Does it make sense to have a similar structure of CSV file for test scenarios in selecting best peers?

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism.spec.js
```diff
@@ -0,0 +1,183 @@
...
+	beforeEach(() => {
+		blockSynchronizationMechanism = new BlockSynchronizationMechanism({
+			channel: stubs.channel,
+			modules: stubs.modules,
+		});
+	});
+
+	describe('constructor()', () => {});
+
+	describe('_computeBestPeer()', () => {
```
shuse2: current convetion is
`#_computeBestPeer` I think

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism.spec.js
```diff
@@ -0,0 +1,183 @@
+const BlockSynchronizationMechanism = require('../../../../../../../src/modules/chain/synchronizer/block_synchronization_mechanism');
```
shuse2: Requires header

framework/test/jest/unit/specs/controller/application_state.spec.js
```diff
@@ -196,7 +196,7 @@ describe('Application State', () => {
 
 			it('should call get four times', async () => {
 				// Assert
-				expect(spies.get).toHaveBeenCalledTimes(4);
+				expect(spies.get).toHaveBeenCalledTimes(5);
```
shuse2: why this increase?

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -1,14 +1,116 @@
...
+	 * From an input list of peers, computes the best peer to continue working with
+	 * according to the set of rules defined in Step 1. of Block Synchroniztion Mechanism
+	 * @link https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#block-synchronization-mechanism
+	 * @param peers
+	 * @return {Array<Object>}
+	 * @private
+	 */
+	_computeBestPeer(peers) {
+		// Largest subset of peers with largest prevotedConfirmedUptoHeight
+		let largestSubset = this._computeLargestSubsetMaxBy(
```
shuse2: prefer no let

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -1,14 +1,116 @@
...
 	}
 
-	// eslint-disable-next-line class-methods-use-this,no-empty-function
-	async run() {}
+	async run() {
+		this.active = true;
+		const { connectedPeers: peers } = await this.channel.invoke(
+			'network:getNetworkStatus'
+		);
+		if (!peers.length) throw new Error('Connected peers list is empty');
```
shuse2: if syntax should be
```
if () {
}
```

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -1,14 +1,116 @@
...
+		// eslint-disable-next-line no-unused-vars
+		const bestPeer = this._computeBestPeer(peers);
+		// TODO: handle bestPeer and move on to step 2 defined in
+		// https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#block-synchronization-mechanism
+		// ...
+		this.active = false;
```
shuse2: if it throws above, it will be active forever 😢

elements/lisk-p2p/test/unit/peer_selection.ts
```diff
@@ -25,6 +25,8 @@ describe('peer selector', () => {
 		let peerList = initializePeerInfoList();
 		const nodeInfo: P2PNodeInfo = {
 			height: 545777,
+			lastBlockId: '1234',
```
shuse2: we are not adding any changes to p2p library, but we need to update the test?
I think p2p doesn't know anything about those values.

framework/src/modules/chain/chain.js
```diff
@@ -372,7 +376,15 @@ module.exports = class Chain {
 			rewardMileStones: this.options.constants.REWARDS.MILESTONES,
 			totalAmount: this.options.constants.TOTAL_AMOUNT,
 			blockSlotWindow: this.options.constants.BLOCK_SLOT_WINDOW,
+			synchronizer: new Synchronizer(),
 		});
+
+		this.blocks.synchronizer.addStrategy(
```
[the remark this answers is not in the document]
limiaspasdaniel: IMO Synchronizer shouldn't be aware of the mechanism registered. This means also that if the mechanisms require more dependencies, Synchronizer should need to add them to, depending entirely on its child mechanisms. I think by instantiating, adding dependencies and registering the mechanisms to the Synchronizer at bootstrap time (chain.js) makes more sense in this case. This will prevent modifying the Synchronizer in the  future if mechanisms are altered.

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism.spec.js
limiaspasdaniel: I think this case is different as the volume of data and input is way less than in BFT case and so it's easier to read as a JSON object. I agree with you that perhaps I can add more tests cases to cover more scenarios.

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism.spec.js
limiaspasdaniel: Perhaps we can add this to Notion, as I have seen other tests using this format.

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
limiaspasdaniel: This function will be completed when all the steps in BSM are implemented separately. I will keep this in mind. Thanks

elements/lisk-p2p/test/unit/peer_selection.ts
limiaspasdaniel: Agreed, this was before I removed the properties from P2P as @ishantiw told me it is made to support generic ones atm.

yatki:
I think tests were updated after I approved the PR. I'll go through the tests tomorrow morning.

framework/test/jest/unit/specs/controller/application_state.spec.js
limiaspasdaniel: Removed the test case as it is difficult to maintain and will be refactored in #3875

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
limiaspasdaniel: Done

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism.spec.js
limiaspasdaniel: Done

limiaspasdaniel:
@SargeKhan https://github.com/LiskHQ/lisk-sdk/pull/3900#discussion_r299499260
> and also maybe have assertions for all properties. Only checking that height exists doesn't make sense anymore.

We will take care of this in #3875. I don't think adding assertions to individual properties makes sense here. @yatki  already has proposed a way to generalize the behavior of this function so any allowed attribute can be passed as long as it complies with the format.

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism.spec.js
```diff
@@ -0,0 +1,199 @@
...
+	let blockSynchronizationMechanism;
+
+	beforeEach(() => {
+		blockSynchronizationMechanism = new BlockSynchronizationMechanism({
+			channel: stubs.channel,
+			modules: stubs.modules,
+		});
+	});
+
+	describe('#constructor', () => {});
```
yatki: why constructor test suite is empty?

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism.spec.js
```diff
@@ -0,0 +1,199 @@
...
+
+	describe('#_computeBestPeer', () => {
+		it('should accept an array of peers as input', () => {
+			expect(blockSynchronizationMechanism._computeBestPeer.length).toEqual(1);
```
yatki: prefer to use toHaveLength matcher.

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism.spec.js
```diff
@@ -0,0 +1,199 @@
...
+				prevotedConfirmedUptoHeight: 0,
+			};
+
+			const peers = [
+				{
+					lastBlockId: '12343245',
+					prevotedConfirmedUptoHeight: 1,
+					height: 67,
+					ip: '127.0.0.1',
+				},
```
yatki: I'm just speculating now, maybe putting another peer object into the peers array could improve the test. Because it's returning one of the good peers randomly, right? so, this test doesn't cover that scenario properly. We need to give 2 good and 1 bad peers and make sure that the best peer is always one of those 2 good peers.

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism.spec.js
```diff
@@ -0,0 +1,199 @@
...
+			 * @link https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#block-synchronization-mechanism
+			 */
+
+			it('should successfully select the best peer according to the steps defined in LIP-0014', () => {
+				stubs.modules.blocks.lastBlock = {
+					height: 0,
+					prevotedConfirmedUptoHeight: 0,
+				}; // So ForkChoiceRule.isDifferentChain returns is TRUTHY
+
+				const peers = [
```
yatki: To improve readability, I would move these peers to a fixture file.

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism.spec.js
limiaspasdaniel: Added the skeleton as these tests will be completed once we have both mechanism in place.

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism.spec.js
limiaspasdaniel: Good catch. This behavior is being tested at line 85 for a bigger set of peers. so the scenario is covered.

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism.spec.js
limiaspasdaniel: Done

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism.spec.js
limiaspasdaniel: Done

framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism/peers.js
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+'use strict';
+
+/**
+ * Fixtures used by Block Synchronization Mechanism's Peer selection algorithm
+ * It represents an array of Peers supposedly returned by the network module
+ * @type {*[]}
+ */
+module.exports = [
```
yatki: ❤️ 👍

PR #3918, opened 2019-07-04
Unreacheable peers are kept as connected for too long - Close #3917

### LINKED ISSUE #3917: Unreacheable peers are kept as connected for too long
### Expected behavior
When peer for some reason stops being reachable, for example, because of:
- lost internet connection
- crashed Lisk Core instance
- destroyed VM

It should be marked as disconnected in less time, for example 30 seconds would be acceptable.
### Actual behavior
Peers are marked as disconnected after 15 minutes, until then they stays as connected with all peer info preserved. This can affect broadcasts, consensus calculation, `networkHeight`, etc. `1.6.0` is marking peers as disconnected properly.
### Steps to reproduce
Have a network with `X` peers, destroy `Y` VM instances or kill `Y` Lisk Core clients. After this query `api/peers` endpoint of some node, check amount of peers with `state: 2` (it will report `X` connected`).
`release/2.0.0`

jondubois:
### What was the problem?

Peers where not marked as disconnected after they were shut down.
Because the pingTimeout (heartbeat) was disabled, the low level WebSocket was not detecting the disconnection for several minutes.

### How did I fix it?

Removed the `pingTimeoutDisabled` option so that it uses the default ping behaviour.

### How to test it?

Run a network, shutdown some nodes (not gracefully).

### Review checklist

* The PR resolves #3917
* Linting passes
* Tests pass

### DIFF
--- a/elements/lisk-p2p/src/peer.ts
+++ b/elements/lisk-p2p/src/peer.ts
@@ -43,19 +43,7 @@ import {
 	validateRPCRequest,
 } from './validation';
 
-// This interface is needed because pingTimeoutDisabled is missing from ClientOptions in socketcluster-client.
-interface ClientOptionsUpdated {
-	readonly hostname: string;
-	readonly port: number;
-	readonly query: string;
-	readonly autoConnect: boolean;
-	readonly autoReconnect: boolean;
-	readonly pingTimeoutDisabled: boolean;
-	readonly multiplex: boolean;
-	readonly ackTimeout?: number;
-	readonly connectTimeout?: number;
-}
-
+type ClientOptions = socketClusterClient.SCClientSocket.ClientOptions;
 type SCClientSocket = socketClusterClient.SCClientSocket;
 
 // Local emitted events.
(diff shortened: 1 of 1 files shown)


elements/lisk-p2p/src/peer.ts
```diff
@@ -670,7 +657,7 @@ export const connectAndRequest = async (
 				procedure,
 			};
 			// Ideally, we should JSON-serialize the whole NodeInfo object but this cannot be done for compatibility reasons, so instead we put it inside an options property.
-			const clientOptions: ClientOptionsUpdated = {
+			const clientOptions: ClientOptions = {
 				hostname: basicPeerInfo.ipAddress,
```
diego-G: Why don't we use `pingTimeoutDisabled: false, pingTimeout: 20000,` as suggested in the commit 6e0a8543f0b10733489b55951aba407935632443 ?

4miners: @diego-G `pingTimeoutDisabled: false` is the default, so we don't need to set it explicitly.

jondubois: @diego-G as Marius mentioned, it's safer and more compatible to stick to the defaults in SC.

diego-G: The timeout too? What about adding a simple scenario checking the behaviour?

PR #3921, opened 2019-07-04
Improve usability of StateStore to get and set new values for accounts - Closes #3913

### LINKED ISSUE #3913: Improve usability of StateStore to get and set new values for accounts
### Expected behavior
Expect to be able to mutate the state of the account returned by `store.account.get(accountId)` and use it to set the new values by calling `store.account.set(accountId, account)`

### Actual behavior
It doesn't set the new values if the account used as second parameter of `store.account.set(accountId, account)` is the account returned by `store.account.get(accountId)`

### Steps to reproduce
```
const account = store.account.get(accountId);
account.asset = { newAttr: true };
store.account.set(accountId, account);
```

2.x

lsilvs:
### What was the problem?
StateStore was returning the object pointer allowing users to mutate its state without using `set` method and consequentially breaking the `finalize` method.

### How did I fix it?
By deep cloning the object before returning it in `cache`, `get` and `getOrDefault` methods.

### How to test it?
```
const account = store.account.get(accountId);
account.asset = { newAttr: true };
store.account.set(accountId, account);
```

### Review checklist

* The PR resolves #3913
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/modules/chain/logic/state_store/account_store.js
+++ b/framework/src/modules/chain/logic/state_store/account_store.js
@@ -51,12 +51,12 @@ class AccountStore {
 	async cache(filter) {
 		const result = await this.account.get(filter, { extended: true }, this.tx);
 		this.data = _.uniqBy([...this.data, ...result], this.primaryKey);
-		return result;
+		return _.cloneDeep(this.data);
 	}
 
 	createSnapshot() {
-		this.originalData = _.clone(this.data);
-		this.updatedKeys = _.clone(this.updatedKeys);
+		this.originalData = _.cloneDeep(this.data);
+		this.updatedKeys = _.cloneDeep(this.updatedKeys);
 	}
 
 	restoreSnapshot() {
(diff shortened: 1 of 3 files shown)


framework/src/modules/chain/submodules/blocks/chain.js
```diff
@@ -317,9 +317,9 @@ Chain.prototype.applyGenesisBlock = function(block, cb) {
 __private.applyTransactions = function(transactions, cb) {
 	modules.processTransactions
 		.applyTransactions(transactions)
-		.then(({ stateStore }) => {
+		.then(async ({ stateStore }) => {
```
shuse2: Probably it's better to be consistent with then/catch or async/await

framework/src/modules/chain/logic/state_store/account_store.js
```diff
@@ -111,7 +111,7 @@ class AccountStore {
 
 		const updatedKeys = Object.entries(updatedElement).reduce(
 			(existingUpdatedKeys, [key, value]) => {
-				if (value !== this.data[elementIndex][key]) {
+				if (!_.isEqual(value, this.data[elementIndex][key])) {
```
shuse2: why do we need `isEqual` here? 😢

lsilvs: Because `{} === {}` is `false` in JS 😢

PR #3930, opened 2019-07-05
Set websocket maxPayload on P2P library - Closes #3335

### LINKED ISSUE #3335: Enforce limits on the size of inbound messages/requests in P2P
### Expected behavior

Messages and requests which are too large should be rejected to prevent DoS against nodes. Ideally, this value should be configurable.

This is particularly important when dealing with peer info. Right now there is no limit as to how big the custom `options` object in `P2PDiscoveredPeerInfo` can be; so peers could just spam us with lots of data inside the options object to fill up our memory (since we keep a list of `P2PDiscoveredPeerInfo`).

### Actual behavior

There is no size limit on messages and requests.

diego-G:
# How to test it?

`npm run test:integration`

### Review checklist

* The PR resolves #3335
* Linting passes
* Tests pass

### DIFF
--- a/elements/lisk-p2p/src/peer.ts
+++ b/elements/lisk-p2p/src/peer.ts
@@ -36,7 +36,6 @@ import { P2PRequest } from './p2p_request';
 
 import * as socketClusterClient from 'socketcluster-client';
 import { SCServerSocket } from 'socketcluster-server';
-import { DEFAULT_WS_MAX_PAYLOAD } from './p2p';
 import {
 	validatePeerInfo,
 	validatePeerInfoList,
@@ -47,7 +46,7 @@ import {
 interface ClientOptionsUpdated
 	extends socketClusterClient.SCClientSocket.ClientOptions {
 	readonly wsEngineServerOptions: {
-		readonly maxPayload: number;
+		readonly maxPayload?: number;
 	};
 }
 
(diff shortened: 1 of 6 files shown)


elements/lisk-p2p/src/peer.ts
```diff
@@ -36,14 +36,29 @@ import { P2PRequest } from './p2p_request';
...
 import {
 	validatePeerInfo,
 	validatePeerInfoList,
 	validateProtocolMessage,
 	validateRPCRequest,
 } from './validation';
 
-type ClientOptions = socketClusterClient.SCClientSocket.ClientOptions;
+// This interface is needed because pingTimeoutDisabled is missing from ClientOptions in socketcluster-client.
+interface ClientOptionsUpdated {
```
ishantiw: shouldn't we extend the ClientOptions and just add missing fields?

elements/lisk-p2p/src/peer.ts
```diff
@@ -36,14 +36,29 @@ import { P2PRequest } from './p2p_request';
 
 import * as socketClusterClient from 'socketcluster-client';
 import { SCServerSocket } from 'socketcluster-server';
+import { DEFAULT_WS_MAX_PAYLOAD } from './p2p';
```
ishantiw: Maybe we can avoid this import by using the incoming payload value in PeerPool because we are instantiating Peer object only their and we can use whatever value is flowing from P2PConfig

elements/lisk-p2p/src/peer.ts
```diff
@@ -475,6 +482,9 @@ export class Peer extends EventEmitter {
 			multiplex: false,
 			autoConnect: false,
 			autoReconnect: false,
+			wsEngineServerOptions: {
+				maxPayload: this._peerConfig.wsMaxPayload,
+			},
```
jondubois: This will not be used. For the client, we should use `maxPayload: this._peerConfig.wsMaxPayload` as a top level property otherwise it will be ignored. The approach above is only for the server-side.

elements/lisk-p2p/src/peer.ts
```diff
@@ -677,6 +687,9 @@ export const connectAndRequest = async (
 				multiplex: false,
 				autoConnect: false,
 				autoReconnect: false,
+				wsEngineServerOptions: {
+					maxPayload: peerConfig.wsMaxPayload,
+				},
```
jondubois: Same here.

PR #3932, opened 2019-07-05
Transaction pool behavior breaks backward compatibility causing hard-fork - Closes #3914

### LINKED ISSUE #3914: Transaction pool behavior breaks backward compatibility causing hard-fork
### Expected behavior
Transactions should be processed the same way as in the previous versions (for example `1.6.0`) so backward compatibility is preserved.
### Actual behavior
Transactions are processed differently (or in different order) resulting with breaking backward compatibility with older versions (hard-fork).
Screen Shot 2019-07-03 at 12 32 22

Suppose the account have 1 LSK balance and we sent following transactions in this particular order:

1. Spend from the account 0.5 LSK + 0.1 LSK fee
2. Spend from the account 0.5 LSK + 0.1 LSK fee
3. Credit the account with 0.3 LSK

Last 3 transactions are included in the same block. When older nodes are processing this block they apply those transactions in order `1 - (0.5 + 0.1) - (0.5 + 0.1) + 0.3` - this is not possible because account balance cannot go negative.
```
[ERR] 2019-07-03 10:16:08 | Failed to apply transaction: 1485790066807584278 to unconfirmed state of account - Account does not have enough LSK: 8772818227501265890L balance: 0.4
[ERR] 2019-07-03 10:16:08 | Transaction - {"id":"1485790066807584278","height":5962,"type":0,"timestamp":98038652,"senderPublicKey":"21cae3b2639038d77fe8fc6bba2a27c0ac3e9dead18d8d349aadb9e3c7d798a6","senderId":"8772818227501265890L","recipientId":"13604394588210584148L","amount":"50000000","fee":"10000000","signature":"5ff6347e9d4e17635e7b764254a28b4b66e0e2aa4014191cd98df1c8066b945defe805db8a1029c244a705ef6f4033f2f6977b136931c4a766920738d0f4d807","signatures":[],"confirmations":null,"asset":{},"blockId":"3406159478353082838","ready":true}
[dbg] 2019-07-03 10:16:08 | Block processing failed - {"id":"3406159478353082838","err":"Error: Failed to apply transaction: 1485790066807584278 to unconfirmed state of account - Account does not have enough LSK: 8772818227501265890L balance: 0.4","module":"blocks","block":{"id":"3406159478353082838","version":1,"timestamp":98038660,"height":5962,"previousBlock":"16526343828936722422","numberOfTransactions":3,"totalAmount":"130000000","totalFee":"30000000","reward":"500000000","payloadLength":351,"payloadHash":"74052bd734cc960b40eb68c225bf68b49881c5e9cb862492a9ba2a00d2c5bd9b","generatorPublicKey":"fab7b58be4c1e9542c342023b52e9d359ea89a3af34440bdb97318273e8555f0","generatorId":"9528507096611161860L","blockSignature":"f8bc80fa0244905a83e467cd205d2a89c9bb741999e97d24367dd9ee5011d0b9f9916b7d4324ce52111039ab9993610a87542a4370668658c44a8d476797b608","confirmations":null,"totalForged":"530000000","transactions":[{"id":"17565598885086760784","height":5962,"type":0,"timestamp":98038652,"senderPublicKey":"12c23f3942907327d5d6422f069f13f056ae9ab6752fc2c2344075abec03a813","senderId":"7069564171607021764L","recipientId":"8772818227501265890L","amount":"30000000","fee":"10000000","signature":"b9bbff723bbff9204aee2b8ebb582e70b6c5104738bc0f7fd19b8d9ac308ee85a62bc9b3e5e3190738ac61c5328acc686ff6ff542e6c68d72e9e4816cc528803","signatures":[],"confirmations":null,"asset":{},"blockId":"3406159478353082838","ready":true},{"id":"9239255295860703076","height":5962,"type":0,"timestamp":98038652,"senderPublicKey":"21cae3b2639038d77fe8fc6bba2a27c0ac3e9dead18d8d349aadb9e3c7d798a6","senderId":"8772818227501265890L","recipientId":"4547759333925451495L","amount":"50000000","fee":"10000000","signature":"ee1d52de8c3d6bbd3b10f7803df99e5e9c64c0936b8fe5487a3f870dcfb3ae8abff9487729c4fce1137db8527d4bca30b66adef0788aa56bd19136aafe89ce04","signatures":[],"confirmations":null,"asset":{},"blockId":"3406159478353082838","ready":true},{"id":"1485790066807584278","height":5962,"type":0,"timestamp":98038652,"senderPublicKey":"21cae3b2639038d77fe8fc6bba2a27c0ac3e9dead18d8d349aadb9e3c7d798a6","senderId":"8772818227501265890L","recipientId":"13604394588210584148L","amount":"50000000","fee":"10000000","signature":"5ff6347e9d4e17635e7b764254a28b4b66e0e2aa4014191cd98df1c8066b945defe805db8a1029c244a705ef6f4033f2f6977b136931c4a766920738d0f4d807","signatures":[],"confirmations":null,"asset":{},"blockId":"3406159478353082838","ready":true}]}}
```
### Steps to reproduce
Send spend transaction(s) that causing account balance to go negative and credit transaction(s) that makes account balance to go positive again. Those needs to be included in one block.
`release/2.0.0`

nazarhussain:
### What was the problem?

There was behavioural change to verify total spending of account in a block, due to removal of unconfirmed state. 

### How did I fix it?
Extend the transaction verification to explicitly verify the total spending per account. 

### Review checklist

* The PR resolves #3914
* Linting passes
* Tests pass


nazarhussain:
@shuse2 @SargeKhan @4miners Please review the code, if its good, I will add more tests.

framework/src/modules/chain/submodules/process_transactions.js
```diff
@@ -276,6 +286,71 @@ class ProcessTransactions {
...
+	_verifyTotalSpending(transactions, stateStore) {
+		const spendingErrors = [];
+
+		// Group the transactions per senderId to calculate total spending
+		const senderTransactions = transactions.reduce((rv, x) => {
+			(rv[x.senderId] = rv[x.senderId] || []).push(x);
+			return rv;
+		}, {});
+
+		// We need to get the transaction id which cause exceeding the sufficient balance
```
4miners: Why we need the transaction ID? If only for logging then it's not worth the additional complexity.

framework/src/modules/chain/submodules/process_transactions.js
```diff
@@ -18,6 +18,7 @@ const {
 	Status: TransactionStatus,
 	TransactionError,
 } = require('@liskhq/lisk-transactions');
+const BigNum = require('bignumber.js');
```
ManuGowda: why not use `@liskhq/bignum` library?

framework/src/modules/chain/submodules/process_transactions.js
ManuGowda: Agree.. 
Error -> `senderAggregatedTransactionSpending` > `accountBalance`
Success -> `senderAggregatedTransactionSpending` < `accountBalance`

framework/src/modules/chain/submodules/process_transactions.js
nazarhussain: Its not added as dependency for the framework in `2.0.0`. And don't want to add it just for this patch.

pablitovicente: We're using @liskhq/bignum in 2.2 @ManuGowda so we'll need to update when merging back :)

framework/src/modules/chain/submodules/process_transactions.js
nazarhussain: We need transaction id to throw error object. Its its not the case then accumulating the amount will be much easier.

framework/test/mocha/integration/delegates/forge.js
```diff
@@ -0,0 +1,113 @@
...
+				const account = random.account();
+				const transaction = transfer({
+					amount: new Bignum(NORMALIZER).multipliedBy(1).toString(),
+					recipientId: account.address,
+					passphrase: accountFixtures.genesis.passphrase,
+				});
+				await addTransactionsAndForge(library, [transaction], 0);
+
+				// Transfer 0.5 LSK to a random account - The account balance would be 0.4
+				const trs_06 = transfer({
```
shuse2: Minor thing but the variable name should be camel case

framework/test/mocha/integration/blocks/verify/verify.js
```diff
@@ -954,15 +955,84 @@ describe('blocks/verify', () => {
...
+						passphrase: accountFixtures.genesis.passphrase,
+					}),
+				],
+				genesisBlock
+			);
+			// Process the block so account have 1 LSK
+			await promisifyProcessBlock(block, true, true);
+
+			// Transfer 0.5 LSK to a random account - The account balance would be 0.4
+			const trs_05_1 = transfer({
```
shuse2: same as below

framework/src/modules/chain/submodules/process_transactions.js
```diff
@@ -276,6 +286,71 @@ class ProcessTransactions {
...
+			senderSpending[senderId] = new BigNum(0);
+
+			senderTransactions[senderId].forEach(transaction => {
+				senderSpending[senderId] = senderSpending[senderId]
+					.plus(transaction.amount)
+					.plus(transaction.fee);
+
+				if (senderBalance.lt(senderSpending[senderId])) {
+					spendingErrors.push({
+						status: TransactionStatus.FatalError,
```
shuse2: TransactionStatus does not have `FatalError`, it should be `TransactionStatus.FAIL`

framework/src/modules/chain/submodules/process_transactions.js
```diff
@@ -276,6 +286,71 @@ class ProcessTransactions {
...
+		// Group the transactions per senderId to calculate total spending
+		const senderTransactions = transactions.reduce((rv, x) => {
+			(rv[x.senderId] = rv[x.senderId] || []).push(x);
+			return rv;
+		}, {});
+
+		// We need to get the transaction id which cause exceeding the sufficient balance
+		// So we can't sum up all transactions together at once
+		const senderSpending = {};
+		Object.keys(senderTransactions).forEach(senderId => {
```
pablitovicente: Can we change this to a `for of` giving that it's way faster and we're starting to focus on processing more transactions per block?

framework/test/mocha/integration/blocks/verify/verify.js
ManuGowda: How about making it more simple
```
const transactions = [0.5, 0.5, 0.3].map(amount => transfer({
				amount: new Bignum(NORMALIZER).multipliedBy(amount).toString(),
				recipientId: random.account().address,
				passphrase: account.passphrase,
			}));
```

framework/test/mocha/integration/delegates/forge.js
ManuGowda: same comment as above.

4miners: We can have more meaningful variable names here, for example, `spend06`, `spend04`, `credit03`

framework/test/mocha/integration/delegates/forge.js
```diff
@@ -0,0 +1,113 @@
...
+
+				// Transfer 0.3 LSK to that account - The account balance would be 0.1
+				const trs_03 = transfer({
+					amount: new Bignum(NORMALIZER).multipliedBy(0.3).toString(),
+					recipientId: account.address,
+					passphrase: accountFixtures.genesis.passphrase,
+				});
+
+				// Add the transactions to pool and forge
+				await addTransaction(library, trs_06);
```
4miners: This is the happy path. The order should be `credit03`, `spend06`, `spend04`. Transaction pool should not select those transactions to be included in one block, because the block will be invalid when it hit block processing (`_verifyTotalSpending`). I don't see any changes related to transaction pool in this PR, so I assume the test with such an order will fail and block with all 3 transactions will be created - which is not an expected behavior.

framework/test/mocha/integration/blocks/verify/verify.js
michielmulders: Only works for `trs_05_01` and `trs_05_02`

framework/test/mocha/integration/delegates/forge.js
4miners: I tested with the correct order and it forge a block with all 3 transactions and fail to process it:
```
	Add transaction ID: 11584432913339092061
	Add transaction ID: 12970355266971244115
	Add transaction ID: 15590495324329756252
		Last block height: 2 Last block ID: 16453677599573165466 Last block timestamp: 10 Next slot: 2 Next delegate PK: 74583aba9c0b92e4f08c8c75e6df341c255ca007971195ff64d6f909dc4b7177 Next block timestamp: 20
Error: the array [
  "Transaction: 12970355266971244115 failed at .amount: Account does not have enough LSK for total spending. balance: 100000000, spending: 120000000"
] was thrown, throw an Error :)
```

framework/test/mocha/integration/delegates/forge.js
```diff
@@ -0,0 +1,113 @@
...
+				await addTransaction(library, trs_04);
+				await addTransaction(library, trs_03);
+				await fillPool(library);
```
4miners: This line shouldn't be here, as `fillPool` is called inside of `forge`. However, when we remove this call and we have in the transaction pool a transaction that should not be included in a block because of `_verifyTotalSpending` - the empty block is forged. The block should not be empty, it should just not contain the invalid transaction. This is a regression because in earlier versions the behavior was correct, I'm not sure if this is a result of this pull request or other changes.

framework/src/modules/chain/submodules/process_transactions.js
nazarhussain: @pablitovicente We discussed in a meeting previously and we will define this as coding practice , until then use of `forEach` is permissible.

framework/test/mocha/integration/delegates/forge.js
nazarhussain: @4miners I tested it yesterday, there was difference of behaviour in `forge` add `addTransactionAndForge`, the later was not working in any case, not even with single transactions. I didn't debugged it why its not, but used `fillPool` and `forge` separately in this test.

framework/test/mocha/integration/delegates/forge.js
nazarhussain: Let me add more tests and confirm the behaviour.

nazarhussain: @4miners In any order you pass, the transactions are validated individually in the pool. And then the block is generated with sorted transactions and invalid transition is removed.

4miners: @nazarhussain Invalid transaction is not removed. It forges a block which contains all of the transactions, then it passes it to block processing and it fails there, so the `forge` call fails as the error propagates.

nazarhussain: This is what I saw during the debugging and while calling the `fillPool`;

```
%% before fill pool 

{ received: Queue { _transactions: [], _index: {} },
  validated: Queue { _transactions: [], _index: {} },
  verified:
   Queue {
     _transactions:
      [ [TransferTransaction],
        [TransferTransaction],
        [TransferTransaction] ],
     _index:
      { '2051428277466744290': [TransferTransaction],
        '17229733069888235767': [TransferTransaction],
        '4415635931455692879': [TransferTransaction] } },


%% after fill pool 

{ received: Queue { _transactions: [], _index: {} },
  validated: Queue { _transactions: [], _index: {} },
  verified:
   Queue {
     _transactions: [ [TransferTransaction], [TransferTransaction] ],
     _index:
      { '2051428277466744290': [TransferTransaction],
        '4415635931455692879': [TransferTransaction] } },
  pending: Queue { _transactions: [], _index: {} },
  ready: Queue { _transactions: [], _index: {} } }
```

From this debug information I tend to think it it removes the invalid transaction from the pool and keep rest of two in the `verified` queue.

4miners: Please change the order of the transactions to `credit03`, `spend06`, `spend04` and re-run this test - it will fail to forge.

framework/src/modules/chain/submodules/process_transactions.js
```diff
@@ -137,24 +138,72 @@ class ProcessTransactions {
...
-			stateStore.transaction.add(transaction);
-			return transactionResponse;
+		return {
+			transactionsResponses: transactions.map(transaction => {
+				const transactionResponse = transaction.apply(stateStore);
+
+				roundInformation.apply(stateStore, transaction);
+				stateStore.transaction.add(transaction);
+
+				transactionResponse.status = TransactionStatus.OK;
```
SargeKhan: Maybe add a comment here which is something like this: 
```suggestion
                                 // We are overriding the status of transaction because it's from genesis block 
				transactionResponse.status = TransactionStatus.OK;
```

framework/test/mocha/common/utils/blocks_and_transactions_helper.js
```diff
@@ -0,0 +1,250 @@
...
+	id: t.id,
+	amount: new Bignum(t.amount).toFixed(),
+	senderId: t.senderId,
+	recipientId: t.recipientId,
+});
+
+/**
+ * Blocks and Transactions Helper
+ */
+class BAT {
```
shuse2: Prefer descriptive name

framework/test/mocha/common/utils/blocks_and_transactions_helper.js
```diff
@@ -0,0 +1,250 @@
...
+
+function createDebitTransaction(account, amount) {
+	return transfer({
+		amount: new Bignum(NORMALIZER).multipliedBy(amount).toString(),
+		recipientId: random.account().address,
+		passphrase: account.passphrase,
+	});
+}
+
+function createCreditTransaction(account, amount) {
```
shuse2: don't we have this function somewhere in commons?

framework/test/mocha/integration/blocks/blocks_and_transactions.js
```diff
@@ -0,0 +1,386 @@
...
+describe('blocks processing & transactions pool consistency', () => {
+	let library;
+
+	localCommon.beforeBlock('blocks_transactions_processing', lib => {
+		library = lib;
+	});
+
+	describe('total spending', () => {
+		describe('should not include transactions which exceed total spending per account balance', () => {
+			it('when we debit the account first', async () => {
```
shuse2: `when` and `it` seems opposite here

framework/test/mocha/integration/blocks/blocks_and_transactions.js
```diff
@@ -0,0 +1,386 @@
...
+				expect(errors).to.be.undefined;
+				// Last block should only contain all transactions marked as EXPECT.OK and no transactions marked as EXPECT.FAIL
+				// Transactions with smallest amount are first as we sort them by amount while forging
+				expect(bat.getTransactionsInLastBlock()).to.deep.equal(
+					bat.getValidSortedTransactions()
+				);
+			});
+		});
+
+		describe('other cases', () => {
```
shuse2: other cases should be more descriptive, probably it's just opposite?

framework/test/mocha/common/utils/blocks_and_transactions_helper.js
```diff
@@ -0,0 +1,250 @@
...
+		const transactionsInPool = this._transactions
+			.filter(t => isTransactionInPool(this._library, t.data.id))
+			.map(t => t.data);
+
+		// We return only transaction ID, amount (in string format), sender and recipient
+		return transactionsInPool.map(formatTransaction);
+	}
+
+	async enqueueTransactions() {
+		return Promise.mapSeries(this._transactions, t =>
```
shuse2: prefer
```
for (const t of this._transactions) {
  await addTransaction(this._library, t.data);
}
```

framework/src/modules/chain/logic/block.js
```diff
@@ -57,16 +57,14 @@ class Block {
...
-	 * @param {Object} data
-	 * @returns {block} block
-	 * @todo Add description for the params
+	 * @param {Array} transactions Unsorted collection of transactions
+	 * @returns {Array} transactions Sorted collection of transactions
+	 * @static
 	 */
-	create(data) {
-		const transactions = data.transactions.sort((a, b) => {
+	static sortTransactions(transactions) {
```
shuse2: do we need to change this now?
Prefer to keep it as it was if it's unnecessary for the conflict

framework/test/mocha/common/utils/blocks_and_transactions_helper.js
SargeKhan: Yes, I think so too. We have most of these functions already declared elsewhere. Adding this class does not make sense to me. It duplicates code and adds technical debt. I'd prefer if we reused existing functions.

framework/src/modules/chain/logic/state_store/account_store.js
```diff
@@ -56,12 +56,14 @@ class AccountStore {
...
-		this.updatedKeys = _.clone(this.updatedKeys);
+		this.originalUpdatedKeys = _.clone(this.updatedKeys);
 	}
 
 	restoreSnapshot() {
 		this.data = this.originalData;
-		this.updatedKeys = {};
+		this.updatedKeys = this.originalUpdatedKeys;
+		this.originalKeys = {};
+		this.originalUpdatedKeys = [];
```
nazarhussain: In constructor you initialised it as an `objet`, why its been converted to array on reset?

framework/src/modules/chain/logic/state_store/account_store.js
```diff
@@ -56,12 +56,14 @@ class AccountStore {
...
 		this.originalData = _.clone(this.data);
-		this.updatedKeys = _.clone(this.updatedKeys);
+		this.originalUpdatedKeys = _.clone(this.updatedKeys);
 	}
 
 	restoreSnapshot() {
 		this.data = this.originalData;
-		this.updatedKeys = {};
+		this.updatedKeys = this.originalUpdatedKeys;
+		this.originalKeys = {};
```
nazarhussain: There is no `originalKeys` defined in the constructor and never used anywhere.

framework/test/mocha/common/utils/blocks_and_transactions_helper.js
nazarhussain: We used same existing code, just wrapped it to write better and quick tests scenarios. So I don't think its something of a problem here. If we had re-invented the `transfer` again and have to maintain on two places then it was a technical debt. 


@4miners

framework/src/modules/chain/logic/block.js
nazarhussain: Yes we need to extract that logic from deep nested blocks, to verify the integrity of blocks in integration tests.

framework/test/mocha/common/utils/blocks_and_transactions_helper.js
```diff
@@ -0,0 +1,250 @@
...
+		this._transactions = this._transactions
+			.filter(t => t.expect)
+			.map(t => {
+				t.data = t.type(this._account, t.amount);
+				return t;
+			});
+	}
+
+	async createBlock() {
+		const promisifyGetNextForger = util.promisify(localCommon.getNextForger);
```
yatki: can we promisify these outside of the function?

framework/src/modules/chain/logic/state_store/account_store.js
SargeKhan: The purpose of these changes was to reset the data stored the `originalData` and `originalUpdatedKeys`. It got messed up after the .swo file. I am fixing it, again.

framework/src/modules/chain/logic/state_store/account_store.js
SargeKhan: Thanks for this catch!

framework/test/mocha/common/utils/blocks_and_transactions_helper.js
4miners: `no-restricted-syntax`, `no-await-in-loop`, if we finally want to allow this - change the rules first.

framework/test/mocha/unit/modules/chain/logic/process_transactions.js
```diff
@@ -0,0 +1,79 @@
...
+				transactionsResponses: [
+					...step1Response.transactionsResponses,
+					...step2Response.transactionsResponses,
+				],
+			});
+		});
+
+		it('should only pass successfull transactions to the next step', async () => {
+			// Assert
+			expect(step2).to.have.been.calledWith([]);
```
yatki: Can we add one OK transaction to the step1response to make sure it's returning the OK transaction?

framework/test/mocha/common/utils/blocks_and_transactions_helper.js
yatki: I also kind of feel like this BAT class needs its own unit tests. Isn't it too complex for a helper? I'm not sure if we can get rid of it in 2.0 but maybe in upcoming versions we should avoid having helper classes of helpers :D

framework/test/mocha/common/utils/blocks_and_transactions_helper.js
yatki: I'd just avoid using bluebird methods.

framework/test/mocha/common/utils/blocks_and_transactions_helper.js
```diff
@@ -0,0 +1,250 @@
...
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+'use strict';
+
+const util = require('util');
+const Promise = require('bluebird');
```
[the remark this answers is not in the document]
4miners: Why? Bluebird is still a better implementation than the native one - it's more flexible, has better performance and uses less memory.

References:
https://softwareengineering.stackexchange.com/a/279003
https://pub.clevertech.biz/native-javascript-promises-vs-bluebird-9e58611be22
https://github.com/petkaantonov/bluebird/tree/master/benchmark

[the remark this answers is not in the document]
shuse2: https://gist.github.com/joseluisq/a87adfd46d86d82ba787919a062f6e36
Those references are outdated with the recent node versions =)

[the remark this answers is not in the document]
4miners: @shuse2 Not true, run the tests I referenced to. The ones you pointed to are testing only one case which is far from the normal application workload. Even if we treat them as meaningful - the difference is negligible, and we shouldn't drop Bluebird with all its features and flexibility because of micro-optimizations. We have much bigger problems to solve, bottlenecks are not here.

framework/test/mocha/common/utils/blocks_and_transactions_helper.js
```diff
@@ -0,0 +1,250 @@
...
+const forge = util.promisify(localCommon.forge);
+const isTransactionInPool = localCommon.transactionInPool;
+
+const addTransactionsAndForge = util.promisify(
+	localCommon.addTransactionsAndForge
+);
+
+const TYPE = {
+	RECEIVED: createCreditTransaction, // CREDIT
+	SPEND: createDebitTransaction, // DEBIT
```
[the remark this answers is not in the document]
4miners: Maybe revert it to `CREDIT` and `DEBIT`, as it was before?

framework/test/mocha/unit/modules/chain/logic/process_transactions.js
nazarhussain: @yatki We didn't created these tests in this PR, just moved from one file to other. We can improve all tests later as discussed.

framework/test/mocha/common/utils/blocks_and_transactions_helper.js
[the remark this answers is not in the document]
nazarhussain: I have no strict opinion on this topic. But what I can suggest is `bluebird` is a widely used and a good library with large user base. Using it in project should not be a problem, unless we have a clear benchmark available which justifies that using it cause problems to our product.

framework/test/mocha/common/utils/blocks_and_transactions_helper.js
nazarhussain: @yatki If you look at many other tests you will see its not too complex. I agree that tests and expectations can be more declarative, but we can do it as separate issue.

nazarhussain: Created a separate issue to refactor these tests. Please share your thoughts to it.

framework/test/mocha/common/utils/blocks_and_transactions_helper.js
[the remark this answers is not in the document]
shuse2: I agree to refactor later in another issue, but one of our goals is to reduce 3rd party usage.
Therefore, we will remove the `bluebird` later along with `async` library.

framework/src/modules/chain/submodules/process_transactions.js
```diff
@@ -139,22 +140,73 @@ class ProcessTransactions {
...
+				!transactionsResponseWithSpendingErrors
+					.map(({ id }) => id)
+					.includes(transaction.id)
+		);
+
+		const transactionsResponses = transactionsWithoutSpendingErrors.map(
+			transaction => {
+				stateStore.account.createSnapshot();
+				const transactionResponse = transaction.apply(stateStore);
+				if (transactionResponse.status !== TransactionStatus.OK) {
```
ManuGowda: can you explain a bit why we have this logic of
```
if not ok -> do something
if ok -> do something
if not ok -> do something
```
this will have three checks each time, rather why not just do
```
if ok -> do something
else if (not ok) -> rest do here
```
?

yatki: @ManuGowda Because if it's not ok in the first condition we are mutating the transactionResponse.status. It's very ugly and not ideal. However, that's the case.. For the moment we can ignore this.

MaciejBaj: Bc of looking for exceptions being "do something" in this case: `if not ok -> updateTransactionResponseForExceptionTransactions`

ManuGowda: anyway, it was not convincing, if  `TransactionStatus.OK` is really --" "-- ok, then we don't need to look for exceptions. but as @yatki explained it's not something u can understand straightaway unless u have the context.

PR #3938, opened 2019-07-08
Remove coverall and use Jenkins - Closes #3322

### LINKED ISSUE #3322: Introduce the coverage metrics without using coverall
### Description

We had multiple issue getting the stable coverage from `coverall`.
We should investigate and re-introduce the coverage on github.

shuse2:
### What was the problem?
All the coverage settings was done in the coveralls, but it was not showing the disired property.
Also, we had Jenkins tools enabled, but not properly working.

### How did I fix it?
Update to collect all the coverage results and show it on the Jenkins.
Also, merged sdk and framework Jenkins test to show the total coverage

### How to test it?
See the Jenkins for sdk and framework

### Review checklist

* The PR resolves #3322 
* Linting passes
* Tests pass


Jenkinsfile.sdk
```diff
@@ -45,7 +45,30 @@ pipeline {
 	}
 	post {
 		always {
-			sh 'rm -rf */node_modules'
+			script {
+				sh '''
+					rm -rf .nyc_output; mkdir -p .nyc_output
+					rm -rf coverage; mkdir -p coverage
```
yatki: It's very trivial thing but can we rename the folders we create on CI with a `ci_` prefix. Because at first glance it feels like we are deleting the folder that was created by test runners then creating the same folder :D Also the line below `cp -r ./commander/.nyc_output/* ./.nyc_output` looks weird at first.

So maybe just `.ci_nyc_output` and `ci_coverage` are more explicit, then we can know we are manually creating these folders to create reports. Would be great if we could rename the other folders we create with jenkins as well.

framework/Jenkinsfile
```diff
@@ -58,21 +58,16 @@ def teardown_mocha(test_name) {
...
 			dir('framework') {
 				sh """
-				rm -rf coverage_mocha_${test_name}; mkdir -p coverage_mocha_${test_name}
-				npx istanbul report --root framework/test/mocha/.coverage-unit/ --dir framework/test/mocha/.coverage-unit/
-				cp framework/test/mocha/.coverage-unit/lcov.info coverage_mocha_${test_name}/ || true
-				npx istanbul report cobertura --root framework/test/mocha/.coverage-unit/ --dir framework/test/mocha/.coverage-unit/
-				cp framework/test/mocha/.coverage-unit/cobertura-coverage.xml coverage_mocha_${test_name}/ || true
-				curl http://localhost:4000/coverage/download --output functional-coverage.zip
-				unzip functional-coverage.zip lcov.info -d coverage_mocha_${test_name}/functional/
+				mkdir -p coverage_mocha_${test_name}
```
yatki: Don't we need to delete this folder or at least clear it ?

framework/Jenkinsfile
```diff
@@ -59,20 +59,15 @@ def teardown_mocha(test_name) {
 			dir('framework') {
 				sh """
 				rm -rf coverage_mocha_${test_name}; mkdir -p coverage_mocha_${test_name}
-				npx istanbul report --root framework/test/mocha/.coverage-unit/ --dir framework/test/mocha/.coverage-unit/
-				cp framework/test/mocha/.coverage-unit/lcov.info coverage_mocha_${test_name}/ || true
-				npx istanbul report cobertura --root framework/test/mocha/.coverage-unit/ --dir framework/test/mocha/.coverage-unit/
-				cp framework/test/mocha/.coverage-unit/cobertura-coverage.xml coverage_mocha_${test_name}/ || true
-				curl http://localhost:4000/coverage/download --output functional-coverage.zip
-				unzip functional-coverage.zip lcov.info -d coverage_mocha_${test_name}/functional/
+				cp test/mocha/.coverage-${test_name}/* coverage_mocha_${test_name}/ || true
```
yatki: Actually having test_name variable here is redundant. Because all mocha tests (unit, integration, functional) are creating their coverage reports to `mocha/.coverage-unit` folder. But I confirmed with @Nazgolze, since all parallel tests are executed in different node it's not a problem. Technically, two test suites can be executed in the same node as well but not at the same time and `cleanWs` function in Jenkinsfile, ensures workspace is cleaned before.

I'd suggest to hard code unit here. I'm gonna create an issue to use different folders for different test types when running mocha tests and reference it to this discussion.

yatki: After having the discussion, since we don't need coverage reports other than integration and deprecating mocha gradually, we can ignore this comment.

Issue #3151, opened 2019-03-22
Benchmark the framework parameters to suggest its values


MaciejBaj:
### Expected behavior
For all of the exposed parameters by `chain` module suggest the minimum and maximum values. Design the best-fitting benchmarking methods for:
- REWARDS
   - MILESTONES
   - OFFSET (Start rewards at first block of the second round)
   - DISTANCE (Distance between each milestone)
- MAX_TRANSACTIONS_PER_BLOCK
- EPOCH_TIME
- BLOCK_TIME (currently defined in slots.interval)

### Which version(s) does this affect? (Environment, OS, etc...)
`1.6.0`

nazarhussain:
@MaciejBaj @shuse2 

Had tested following constants or genesis configurations and these works well.

**EPOCH_TIME**
- Changed it to "2009-06-15T13:45:30Z"
- Started the app
- Call `curl localhost:4000/api/blocks`
- The timestamp of the firs block returned `311214370`
- If you test the value by `moment.unix(moment("2009-06-15T13:45:30Z").unix()+ 311214370).format()` 
- It returns the `2019-04-26T16:11:40+02:00` clock time which is the accurately correct for that block.

nazarhussain:
To benchmark the right min/max range for `MAX_TRANSACTIONS_PER_BLOCK` and `BLOCK_TIME` is very difficult for the reason that both are co-related to each other. As well different physical parameters can effect the performance as well. 

I suggest to do this after benchmarking of the new transaction processing.

MaciejBaj:
We should be able to give "healthy" suggestions on the value ranges.

pablitovicente:
Limits on values for BLOCK_TIME and MAX_TRANSACTIONS_PER_BLOCK (these values were obtained by trying against a single node (node specs: 1GB Ram, 1CPU core, 25 GB HDD and later confirmed against a network consisting of 50 nodes distributed in 10 different regions) it’s important to note that these values do not represent limits in the SDK but recommended values for the current defaults for other configuration properties and objects configurations used by the SDK and lisk-core. 


| BLOCK_TIME | MAX_TRANSACTIONS_PER_BLOCK| 
| -------------| ------------------------------------| 
| 2 | 25| 
| 4 | 50| 
| 6 | 75| 
| 8 | 110| 
| 10 | 150| 


- With all this configurations blocks were generated correctly, on time and filling all blocks with the configured number of transactions. 

[...]

PR #3940, opened 2019-07-09
application_config_schema.js not being used and Application/Schema invalid syntax - Closes #3939

### LINKED ISSUE #3939: application_config_schema.js not being used and Application/Schema invalid syntax
### Expected behavior

- `application_config_schema.js` is validated in Application as `genesisBlockSchema` is.
- Suggested values in https://github.com/LiskHQ/lisk-sdk/issues/3151 are added to `application_config_schema.js`
- Schema definition is valid

### Actual behavior

- `application_config_schema.js`  is not being validated.
- schema is using `min` and `max` instead of `minimum` and `maximum` for limits which is invalid and pass validation with invalid values

### Steps to reproduce

Make changes in `application_config_schema.js` setting values that should fail validation and application will run

pablitovicente:
### What was the problem?

- application_config_schema.js is validated in Application as genesisBlockSchema is.
- Suggested values in #3151 are added to application_config_schema.js
- Schema definition was using `min` and `max` instead of `minimum` and `maximum`

### How did I fix it?

- application_config_schema.js is being validated.
- schema is using min and max instead of minimum and maximum for limits which is invalid and pass validation with invalid values

### How to test it?

- Build should be green
- Change `framework/test/fixtures/config/devnet/config.json` with values that will fail validation and application should not start

### Review checklist

* The PR resolves #3939
* Linting passes
* Tests pass

### DIFF
--- a/framework/src/controller/application.js
+++ b/framework/src/controller/application.js
@@ -29,7 +29,11 @@ const Controller = require('./controller');
 const version = require('../version');
 const validator = require('./validator');
 const configurator = require('./default_configurator');
-const { genesisBlockSchema, constantsSchema } = require('./schema');
+const {
+	genesisBlockSchema,
+	constantsSchema,
+	applicationConfigSchema,
+} = require('./schema');
 
 const { createLoggerComponent } = require('../components/logger');
 
@@ -114,6 +118,8 @@ class Application {
 	 */
 	constructor(genesisBlock, config = {}) {
 		validator.validate(genesisBlockSchema, genesisBlock);
+
+		validator.validate(applicationConfigSchema, config);
 
 		// Don't change the object parameters provided
 		let appConfig = _.cloneDeep(config);
--- a/framework/src/controller/application.js
+++ b/framework/src/controller/application.js
@@ -29,11 +29,7 @@ const Controller = require('./controller');
 const version = require('../version');
 const validator = require('./validator');
 const configurator = require('./default_configurator');
-const {
-	genesisBlockSchema,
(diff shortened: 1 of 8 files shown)


framework/src/controller/application.js
```diff
@@ -29,7 +29,11 @@ const Controller = require('./controller');
 const version = require('../version');
 const validator = require('./validator');
 const configurator = require('./default_configurator');
-const { genesisBlockSchema, constantsSchema } = require('./schema');
+const {
+	genesisBlockSchema,
+	constantsSchema,
+	applicationConfigSchema,
```
nazarhussain: @pablitovicente If that's the only issue, this config schema is already been loaded to configurator, you don't need to do it here.

pablitovicente: Hmm will check that code then as it was not working. It was possible to specify invalid values but maybe due to the combination of wrong schema definition... will double check.

pablitovicente: Removed application schema validation from Application. Schema fix was enough. Resolving.

framework/src/controller/schema/application_config_schema.js
```diff
@@ -159,12 +164,12 @@ module.exports = {
 								},
 								OFFSET: {
 									type: 'integer',
-									min: 1,
+									minimum: 1,
```
nazarhussain: For future reference, here is the specification for this change. 

https://json-schema.org/latest/json-schema-validation.html#rfc.section.6.2.4

PR #3945, opened 2019-07-11
Custom transaction fails to validate if `FEE` member is not defined - Closes #3941

### LINKED ISSUE #3941: Custom transaction fails to validate if `FEE` member is not defined
### Expected behavior
Expect custom transaction to not crash during validation if `FEE` member is not defined.

### Actual behavior
The PR https://github.com/LiskHQ/lisk-sdk/pull/3893 introduced a new static member called `FEE` to `BaseTransaction`. This property should be implemented by custom transactions to define its own fee but it crashes during validation if it's not defined.

### Steps to reproduce
Create a custom transaction extending `BaseTransaction` like:
```
class CustomTransaction extends BaseTransaction {
	static get TYPE () {
		return 10;
	}

	validateAsset() {
		return[];
	}
}
```
Then:
```
const tx = new CustomTransaction({
  recipientId: '10881167371402274308L',
});

tx.sign('passphrase');
tx.validate();
```

`validate()` will throw `eq() not a number: undefined`

2.1

lsilvs:
### What was the problem?
Custom transaction crashing during validation if FEE member is not defined.

### How did I fix it?
By setting a default value for `FEE`

### How to test it?
Create a custom transaction extending `BaseTransaction` like:
```
class CustomTransaction extends BaseTransaction {
	static get TYPE () {
		return 10;
	}

	validateAsset() {
		return[];
	}
}
```
Then:
```
const tx = new CustomTransaction({
  recipientId: '10881167371402274308L',
});

tx.sign('passphrase');
tx.validate();
```

It should return successful transaction response

### Review checklist

* The PR resolves #3941
* Linting passes
* Tests pass

### DIFF
--- a/elements/lisk-transactions/src/base_transaction.ts
+++ b/elements/lisk-transactions/src/base_transaction.ts
@@ -115,7 +115,7 @@ export abstract class BaseTransaction {
 	public receivedAt?: Date;
 
 	public static TYPE: number;
-	public static FEE: string;
+	public static FEE = '0';
 
 	protected _id?: string;
 	protected _senderId?: string;
@@ -143,12 +143,21 @@ export abstract class BaseTransaction {
 		const tx = (typeof rawTransaction === 'object' && rawTransaction !== null
 			? rawTransaction
 			: {}) as Partial<TransactionJSON>;
+
 		this.amount = new BigNum(
 			isValidNumber(tx.amount) ? (tx.amount as string | number) : '0',
 		);
+
 		this.fee = new BigNum(
-			isValidNumber(tx.fee) ? (tx.fee as string | number) : '0',
+			isValidNumber(tx.fee)
+				? (tx.fee as string | number)
+				: (this.constructor as typeof BaseTransaction).FEE,
 		);
+
+		this.type =
+			typeof tx.type === 'number'
+				? tx.type
+				: (this.constructor as typeof BaseTransaction).TYPE;
 
 		this._id = tx.id;
 		this.recipientId = tx.recipientId || '';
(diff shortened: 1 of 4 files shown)


elements/lisk-transactions/test/base_transaction.ts
```diff
@@ -501,6 +504,15 @@ describe('Base transaction class', () => {
 			expect(status).to.eql(Status.OK);
 		});
 
+		it('should return a successful transaction response with a valid transaction with basic impl', async () => {
+			transactionWithBasicImpl.sign('passphrase');
```
shuse2: shouldn't we check if we add the invalid fee, it becomes `0`?

elements/lisk-transactions/src/base_transaction.ts
```diff
@@ -115,7 +115,7 @@ export abstract class BaseTransaction {
 	public receivedAt?: Date;
 
 	public static TYPE: number;
-	public static FEE: string;
+	public static FEE = '0';
```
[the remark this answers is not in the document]
lsilvs: We don't have default value for `TYPE`

elements/lisk-transactions/test/base_transaction.ts
lsilvs: I've added a new test case scenario ;)

elements/lisk-transactions/src/base_transaction.ts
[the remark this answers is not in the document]
lsilvs: ah! You mean variable type, not transaction type? haha

If so, tslint said I shouldn't use
```
no-inferrable-types -> Type string trivially inferred from a string literal, remove type annotation
```

elements/lisk-transactions/src/base_transaction.ts
```diff
@@ -143,13 +143,22 @@ export abstract class BaseTransaction {
...
+
 		this.fee = new BigNum(
-			isValidNumber(tx.fee) ? (tx.fee as string | number) : '0',
+			isValidNumber(tx.fee)
+				? (tx.fee as string | number)
+				: (this.constructor as typeof BaseTransaction).FEE,
 		);
 
+		this.type =
+			typeof tx.type === 'number'
```
pablitovicente: Maybe not to be fixed in this PR but should we discuss a way to let the user know we're defaulting to something or maybe create an issue to document this behaviour?

PR #3946, opened 2019-07-11
Prevent crashing the application when a block is received and node is syncing or forging - Closes #3896

### LINKED ISSUE #3896: Fatal error Block process cannot be executed in parallel
### Expected behavior
There shouldn't be any fatal errors and the block processing should be completed without any errors.
### Actual behavior
```
10:04:50.247Z FATAL lisk-framework:
    System error: unhandledRejection : { message: 'Block process cannot be executed in parallel',
      stack:
       'Error: Block process cannot be executed in parallel\n    at Blocks._shouldNotBeActive (/home/lisk/lisk-2.0.0-rc.0-f5bc2760c-e629652-Linux-x86_64/node_modules/lisk-framework/src/modules/chain/blocks/blocks.js:564:10)\n    at Object.sequence.add [as worker] (/home/lisk/lisk-2.0.0-rc.0-f5bc2760c-e629652-Linux-x86_64/node_modules/lisk-framework/src/modules/chain/blocks/blocks.js:316:9)\n    at Sequence._tick (/home/lisk/lisk-2.0.0-rc.0-f5bc2760c-e629652-Linux-x86_64/node_modules/lisk-framework/src/modules/chain/utils/sequence.js:72:30)\n    at Timeout.nextSequence [as _onTimeout] (/home/lisk/lisk-2.0.0-rc.0-f5bc2760c-e629652-Linux-x86_64/node_modules/lisk-framework/src/modules/chain/utils/sequence.js:52:15)\n    at ontimeout (timers.js:436:11)\n    at tryOnTimeout (timers.js:300:5)\n    at listOnTimeout (timers.js:263:5)\n    at Timer.processTimers (timers.js:223:10)' }
```
### Steps to reproduce
- Deploy a network from release branch 2.2.0(sdk) and core(development) branch
- Enable forging
2.2.0

limiaspasdaniel:
### What was the problem?
When receiving a block from the network in `receiveBlockFromNetwork` function, the application was crashing with a fatal error if the node was syncing or forging at that moment (aka `blocks.js#_isActive was set to true`)
### How did I fix it?
By adding back `forger.#forge` call to the Sequence. Somehow it was removed from it in `2.2`
Added a unit test case to cover this particular scenario.
### How to test it?
Run these tests `npm run mocha:unit framework/test/mocha/unit/modules/chain/chain.js`
### Review checklist

* The PR resolves #3896 
* Linting passes
* Tests pass


pablitovicente:
> https://github.com/LiskHQ/lisk-sdk/blob/release/2.0.0/framework/src/modules/chain/submodules/blocks/process.js#L560-L565
> 
> I think we shouldn't change the `receiveBlockFromNetwork` in block, but it should be rejected in transport layer as it was before

I like this idea if the changes are not massive - due to the possible accessibility issues of the `active` variable -  it would be nice for this to happen in network. Maybe through ApplicationState? But I think that if the changes for this are too big it could be ok if we create an issue to solve it in a minor version.

limiaspasdaniel:
> [/2.0.0/framework/src/modules/chain/submodules/blocks/process.js@`release`#L560-L565](https://github.com/LiskHQ/lisk-sdk/blob/release/2.0.0/framework/src/modules/chain/submodules/blocks/process.js#L560-L565) [](/LiskHQ/lisk-sdk/blob/HEAD@{2019-07-11T16:13:19Z}/2.0.0/framework/src/modules/chain/submodules/blocks/process.js#L560-L565)
> 
> I think we shouldn't change the `receiveBlockFromNetwork` in block, but it should be rejected in transport layer as it was before

@shuse2 you mean `transport.js`? That file points to `process.js`

limiaspasdaniel:
Also @shuse2 @pablitovicente shouldn't we add the task to the Sequence queue no matter what, and then the task itself could check if `active` = true. This way we don't miss received blocks. Doing it the way you propose I think could change the behavior.

shuse2:
If the current status is `syncing` it shouldn't add to the sequence. If we do, the process queue will be something like
```
height: 10 (sync)
height: 10 (sync)
height: 200000 (received)
height: 10 (sync)
height: 10 (sync)
height: 10 (sync)
```
but this received is unnecessary, and only useful in case where it's pushed exactly when the sync is finishing

framework/test/mocha/unit/modules/chain/chain.js
```diff
@@ -329,4 +334,117 @@ describe('Chain', () => {
 			return expect(stubs.modules.module2.cleanup).to.have.been.called;
 		});
 	});
+
+	describe('#_startForging', () => {
```
yatki: Please always use `// Arrange & Act & Assert` comments as much as possible. It will also help us to organize our code in better way.

framework/test/mocha/unit/modules/chain/chain.js
```diff
@@ -329,4 +334,117 @@ describe('Chain', () => {
...
+			expect(chain.forger.loadDelegates).to.be.called;
+		});
+
+		describe('in sequence', () => {
+			it('should register a task in Jobs Queue named "nextForge" and a designated interval', async () => {
+				await chain._startForging();
+				expect(stubs.jobsQueue.register).to.be.calledWith('nextForge');
+			});
+
+			it('should halt if no delegates are enabled', async () => {
```
yatki: I think these tests that contain `let done;` and `await workerPromise` is following really bad pattern. We shouldn't have to rely on this kind of tests. Because in several tests we'd need to arrange that `workerPromise` which is redundant for the things we want to check.

I suggest to extract `nextForge` job to a separate function. Write unit tests for that function. And when it's necessary we can just check if `stubs.jobsQueue.register` was called with correct arguments.

framework/src/modules/chain/chain.js
```diff
@@ -456,21 +456,22 @@ module.exports = class Chain {
...
-					this.logger.debug('Client not ready to forge');
-					return;
-				}
-				try {
-					await this.forger.forge();
-				} catch (error) {
-					this.logger.error(error);
-					throw error;
-				}
+				await this.scope.sequence.add(async () => {
```
yatki: let's move this to a separate method for making testing easier.

framework/test/mocha/unit/modules/chain/chain.js
limiaspasdaniel: @shuse2 @yatki I have created this issue #3958 so this PR can be merged and in order to apply the above suggestion.

PR #3957, opened 2019-07-15
Bus sockets timeout - Closes #3942

### LINKED ISSUE #3942: Bus sockets setup timeout error
### Expected behavior
Expect to clean axon socket when application exits.

### Actual behavior
Socket keeps open and application crash when trying to start again
```
Faced error in application Error: Bus sockets setup timeout
    at Timeout.setTimeout [as _onTimeout] (/lisk-framework/src/controller/bus.js:375:12)
    at ontimeout (timers.js:436:11)
    at tryOnTimeout (timers.js:300:5)
    at listOnTimeout (timers.js:263:5)
    at Timer.processTimers (timers.js:223:10)
```

### Steps to reproduce
Create a new application using SDK 2.1.0 with the following code:
```
const { Application, genesisBlockDevnet, configDevnet } = require('lisk-sdk');

const app = new Application(genesisBlockDevnet, configDevnet);

app
	.run()
	.then(() => app.logger.info('App started...'))
	.catch(error => {
		console.error('Faced error in application', error);
		process.exit();
	});
```

Start the application.
Exit the application.
Start the application again.

2.1

michielmulders:
### What was the problem?
Bus sockets time out whenever you stop the application (`CMD+C`) and try to restart it.
The reason for this is the path where the sockets are stored is longer than the allowed 104 characters for MacOS (or 108 for Unix). So, the cleanup did happen upon shutting down the application, but because the path is too long, it couldn't clean up those files.

This resulted in having to manually remove the `/tmp` folder.

### How did I fix it?
1. Made the `tempPath` configurable as the user wishes.
2. By default, the path gets created inside the `/tmp` folder of the system.
+ Had to slice the default app `label` as it was using the full payload hash of the genesis block. @lsilvs suggested to use only the first 7 characters as that makes it unique (referring to Github commit slice). 

### How to test it?
Run tests:
- `npm run jest:unit -- test/jest/unit/specs/controller/application.spec.js`
- `framework/test/jest/unit/specs/controller/controller.spec.js`
- Modify the `/config/devnet/config.json` and check if this folder path contains files when the app is running:
```json
{
	"app": {
		"ipc": {
			"enabled": true
		},
		"tempPath": "/tmp/custom-folder"
	},
...
}
```

#### Calculations for determining unix socket path length
Note for future reference about calculations for limiting `tempPath` and `label` config parameters.

Unix socket path length:
- MacOS 104
- Unix 108

We choose to support MacOS and use a buffer of 1 character = 103.
-> Max length app label: 30
-> Max length tempPath: 50

Longest possible path:
`${tempPath:50}/${label:30}/sockets/lisk_pub.sock` = 103 chars 👍 

### Review checklist

* The PR resolves #3942 
* Linting passes
* Tests pass


framework/test/jest/unit/specs/controller/application.spec.js
```diff
@@ -76,6 +76,32 @@ describe('Application', () => {
 			expect(app.config.app.label).toBe(config.app.label);
 		});
 
+		it('should set default tempPath if not provided', () => {
```
yatki: ❤️

framework/src/controller/schema/application_config_schema.js
```diff
@@ -56,6 +61,15 @@ module.exports = {
 					example: '968d7b5b97a5bfad8f77614dc8a9918de49f6c6e',
 					description: 'The version of Lisk Core that the peer node runs on.',
 				},
+				tempPath: {
+					type: 'string',
+					format: 'path',
+					minLength: 1,
+					maxLength: 50,
```
lsilvs: Can you add a comment here explaining why we have max length limit, bitte?

michielmulders: Ok, will add a description for this and expanded description with the calculations on the issue itself, thanks! 

`Restrictions due to unix domain socket path length`.

PR #3959, opened 2019-07-15
Add missing unit tests for transactions processing - Closes #3951

### LINKED ISSUE #3951: Add missing unit tests for transactions processing
### Expected behavior
All the files related to transaction processing must have unit test. 

### Actual behavior
There are missing unit test for `submodule/transactions.js`

2.0.0

nazarhussain:
### What was the problem?

There were no unit test for `submoduels/transactions_processing.js`

### How did I fix it?

Added the unit tests 

### How to test it?

`cd framework && npm run mocha:unit test/mocha/unit/modules/chain/submodules/process_transactions.js`

### Review checklist

* The PR resolves #3951
* Linting passes
* Tests pass

### DIFF
--- a/framework/test/mocha/unit/modules/chain/submodules/process_transactions.js
+++ b/framework/test/mocha/unit/modules/chain/submodules/process_transactions.js
@@ -23,13 +23,23 @@ const ProcessTransactions = rewire(
 
 describe('ProcessTransactions', () => {
 	let processTransactions;
-	const dummyTransactions = [
+
+	const trs1 = [
 		{
 			id: 'aTransactionId',
 			matcher: () => true,
 			type: 0,
 		},
 	];
+
+	const trs2 = [
+		{
+			id: 'aTransactionId',
+			matcher: () => true,
+			type: 0,
+		},
+	];
+
 	const dummyState = {
 		version: 1,
 		height: 1,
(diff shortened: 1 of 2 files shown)


framework/test/mocha/unit/modules/chain/submodules/process_transactions.js
```diff
@@ -41,16 +80,435 @@ describe('ProcessTransactions', () => {
...
-			logic: {},
+		processTransactions = new ProcessTransactions(() => {
+			processTransactions.onBind(paramScope);
+
+			// Add matcher to transactions
+			trs1.matcher = () => true;
+			trs2.matcher = () => true;
+
+			// Add prepare steps to transactions
+			trs1.prepare = sinonSandbox.stub();
```
yatki: prefer stubbing with `stub.trs1Prepare = sinonSandbox.stub(trs1, 'prepare')`. Same goes with other stubbings. 

This way, `sinonSandbox.restore()` could work. At the moment, sinon is not able to track the stubs, and the reason tests are working is we are reassigning new stubs to these methods in every beforeEach hook.

Then instead of 
```js
			trs1.validate = sinonSandbox.stub().returns(validResponse);
```

we would use 
```js
			stubs.trs1Validate.returns(validResponse);
```

framework/test/mocha/unit/modules/chain/submodules/process_transactions.js
```diff
@@ -41,16 +80,435 @@ describe('ProcessTransactions', () => {
...
+
+		roundInformationMock.apply.reset();
+		roundInformationMock.undo.reset();
+
+		StateStoreStub.resetHistory();
+		stateStoreMock.restoreSnapshot.reset();
+		stateStoreMock.createSnapshot.reset();
+		stateStoreMock.account.restoreSnapshot.reset();
+		stateStoreMock.account.createSnapshot.reset();
+		stateStoreMock.transaction.add.reset();
```
yatki: `sinonSandbox.reset` should be covering all these reset calls. Are they really needed?

framework/test/mocha/unit/modules/chain/submodules/process_transactions.js
```diff
@@ -41,16 +80,435 @@ describe('ProcessTransactions', () => {
...
+
+			expect(result).to.be.eql({ transactionsResponses: [] });
+		});
+		it('should invoke entities.Transaction to check persistence of transactions', async () => {
+			storageMock.entities.Transaction.get.resolves([trs1, trs2]);
+
+			await processTransactions.checkPersistedTransactions([trs1, trs2]);
+
+			expect(storageMock.entities.Transaction.get).to.be.calledOnce;
+			expect(storageMock.entities.Transaction.get).to.be.calledWithExactly({
```
yatki: It's trivial but you can combine these two assertions with `.and` matcher. But it's fine this way too but would be more readable if we combine them?

framework/test/mocha/unit/modules/chain/submodules/process_transactions.js
nazarhussain: @yatki I love using `sinonSandbox.stub(trs1, 'prepare')` this convention, but problem in this case is we don't have `prepare`, `apply` on those trs objets. So I have to assign those.

framework/test/mocha/unit/modules/chain/submodules/process_transactions.js
nazarhussain: The call to `sinonSandbox.reset` is not resetting all stubs, don't exactly debugged why. Will look at it later.

framework/test/mocha/unit/modules/chain/submodules/process_transactions.js
nazarhussain: I prefer to have more atomic expectations.

nazarhussain:
@shuse2 @SargeKhan @yatki Remember the context for this PR. The tests can be better with some tweaking of the implementation. But at this point of release I don't want to change any single line in the code itself, just for tests.

PR #3963, opened 2019-07-16
Chai-as-promised assertions are giving false positives - Closes #3956

### LINKED ISSUE #3956: Chai-as-promised assertions are giving false positives
### Expected behavior
assertions that use `eventually` matcher should be awaited/returned. Otherwise they are giving false positives.

Example: 
```js
// file: framework/test/mocha/unit/modules/chain/submodules/blocks/verify.js
// Correct
it('should throw error when transaction is already confirmed', async () => {
  await expect(
    __private.checkTransactions(dummyBlock, true)
  ).to.eventually.be.rejectedWith('memo');
});
```

The test above will fail because ıt's not being rejected with `memo`.
### Actual behavior

We are not awaiting/returning the async assertions. So tests are giving false positives.

Example:

```js
// Wrong
it('should throw error when transaction is already confirmed', async () => {
  await expect(
    __private.checkTransactions(dummyBlock, true)
  ).to.eventually.be.rejectedWith('memo');
});
```

The test above will **PASS** even though it's not being rejected with `memo`. But it shouldn't pass.

I suggest to solve this in 2.0

yatki:
### What was the problem?
Async assertions were not awaited or returned thus giving false possitive in mocha tests.
### How did I fix it?
I returned the async assertions
### Review checklist

* The PR resolves #3956 
* Linting passes
* Tests pass

### DIFF
--- a/framework/test/mocha/integration/state_store/round_store.js
+++ b/framework/test/mocha/integration/state_store/round_store.js
@@ -63,7 +63,7 @@ describe('system test - round store', () => {
 
 	describe('cache', () => {
 		it('should throw ', async () => {
-			expect(roundStore.cache()).to.eventually.throw(
+			return expect(roundStore.cache()).to.eventually.throw(
 				'cache cannot be called for round'
 			);
 		});
--- a/framework/test/mocha/integration/state_store/round_store.js
+++ b/framework/test/mocha/integration/state_store/round_store.js
@@ -63,7 +63,7 @@ describe('system test - round store', () => {
 
 	describe('cache', () => {
 		it('should throw ', async () => {
-			return expect(roundStore.cache()).to.eventually.throw(
+			return expect(roundStore.cache()).to.eventually.rejectedWith(
 				'cache cannot be called for round'
 			);
 		});
--- a/framework/test/mocha/unit/modules/chain/components/storage/entities/account.js
+++ b/framework/test/mocha/unit/modules/chain/components/storage/entities/account.js
@@ -178,7 +178,9 @@ describe('ChainAccount', () => {
 		it('should create an account object successfully', async () => {
 			const account = new accountFixtures.Account();
 
-			await expect(AccountEntity.create(account)).to.eventually.be.fulfilled;
+			await expect(
+				AccountEntity.create(account)
+			).to.eventually.be.fulfilled.and.deep.equal([]);
 
 			const accountResult = await AccountEntity.getOne(
 				{
@@ -197,7 +199,9 @@ describe('ChainAccount', () => {
 			const account = new accountFixtures.Account();
 			account.asset = { lisk: 'test-asset' };
 
-			await expect(AccountEntity.create(account)).to.eventually.be.fulfilled;
+			await expect(
+				AccountEntity.create(account)
+			).to.eventually.be.fulfilled.and.deep.equal([]);
 
 			const accountResult = await AccountEntity.getOne(
 				{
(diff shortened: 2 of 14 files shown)


framework/test/mocha/integration/state_store/round_store.js
```diff
@@ -63,7 +63,7 @@ describe('system test - round store', () => {
 
 	describe('cache', () => {
 		it('should throw ', async () => {
-			expect(roundStore.cache()).to.eventually.throw(
+			return expect(roundStore.cache()).to.eventually.throw(
```
lsilvs: Tests failing for this one. Changing to the following should fix it:

```suggestion
			return expect(roundStore.cache()).to.be.rejectedWith(
```

yatki: @lsilvs cache is an async function right? So I think eventually should stay there.

lsilvs: When you use `rejected` or `fulfilled` the `eventually` is implicit, I guess. But should work with it as well.

framework/test/mocha/unit/modules/chain/submodules/blocks/verify.js
```diff
@@ -1794,8 +1794,8 @@ describe('blocks/verify', () => {
 					});
 
 					it('should throw error when transaction is already confirmed', async () => {
-						expect(__private.checkTransactions(dummyBlock, true)).to.eventually
-							.rejected;
+						return expect(__private.checkTransactions(dummyBlock, true)).to
+							.eventually.rejected;
```
lsilvs: Would be nice to be explicit here and add the expected message using `rejectedWith` instead of `rejected` only

framework/test/mocha/unit/modules/chain/submodules/blocks/verify.js
```diff
@@ -1835,8 +1835,8 @@ describe('blocks/verify', () => {
 						modules.processTransactions.verifyTransactions.resolves(
 							invalidTransactionsResponse
 						);
-						expect(__private.checkTransactions(dummyBlock, true)).to.eventually
-							.throw;
+						return expect(__private.checkTransactions(dummyBlock, true)).to
+							.eventually.throw;
```
lsilvs: Same as above

framework/test/mocha/unit/modules/chain/submodules/blocks/verify.js
```diff
@@ -1859,8 +1859,8 @@ describe('blocks/verify', () => {
 					],
 				});
 
-				expect(__private.checkTransactions(dummyBlock, false)).to.eventually.be
-					.rejected;
+				return expect(__private.checkTransactions(dummyBlock, false)).to
+					.eventually.be.rejected;
```
lsilvs: Same as above

framework/test/mocha/unit/modules/chain/logic/broadcaster.js
```diff
@@ -483,7 +483,9 @@ describe('Broadcaster', () => {
 				.rejects(new Error('failed to broadcast'));
 			broadcaster.filterQueue = filterQueueStub;
 			broadcaster.enqueue(params, options);
-			expect(broadcaster.releaseQueue()).rejectedWith('failed to broadcast');
+			return expect(broadcaster.releaseQueue()).to.eventually.be.rejectedWith(
+				'failed to broadcast2'
```
lsilvs:
```suggestion
				'failed to broadcast'
```

framework/test/mocha/unit/modules/chain/components/storage/entities/account.js
```diff
@@ -219,9 +224,11 @@ describe('ChainAccount', () => {
 				new accountFixtures.Account(),
 			];
 
-			expect(AccountEntity.create(accounts)).to.be.fulfilled;
+			await expect(
+				AccountEntity.create(accounts)
+			).to.eventually.be.fulfilled.and.deep.equal([]);
 
-			accounts.forEach(async account => {
+			return accounts.map(async account => {
```
lsilvs: I think we need a `Promise.all` here

PR #3968, opened 2019-07-16
Implement netgrouping for peer eviction and buckets - Closes #3874

### LINKED ISSUE #3874: Create ip prefix calculation function for peer bucket
### Description

In order to create peer bucket within new/tried peer list we need to calculate /16 IPv4 prefix of peer's IP address.

```
group = /16 IPv4 prefix of peer's IP address
k = Hash(random_secret, IP) % 4
b = Hash(random_secret, group, k) % 64
```

mitsujutsu:
### What was the problem?

No functionality in place for grouping peers into `netgroups` for inbound peer eviction, as well as for `peer bucket` functionality.

### How did I fix it?

Added utility file with functions `getNetgroup` for use in inbound peer eviction, `getBucket` for peer bucket functionality and other functions related to IP addresses. 

A peer's `bucket` is computed as follows:

```
group = /16 IPv4 prefix of peer's IP address
k = Hash(random_secret, IP) % 4
bucket = Hash(random_secret, group, k) % 64
```

A peer's `netgroup` is computed as follows:
```
netGroup = Hash(random_secret, IP group of peer)
```
For Lisk implementation, we will put local and private IP addresses into their own buckets respectively, while all other `IPv4` addresses are computed regularly. 

### How to test it?

npm run test

### Review checklist

* The PR resolves #3874 
* Linting passes
* Tests pass


elements/lisk-p2p/src/utils.ts
```diff
@@ -0,0 +1,150 @@
...
+		secretBytes,
+		networkBytes,
+		groupABytes,
+		groupBBytes,
+	]);
+
+	return hash(netgroupBytes).readUInt32BE(0);
+};
+
+export const getBucket = (
```
ishantiw: Since we need MOD_128 for `newPeers` so can we make this function to accept `numOfBuckets` and use it to get a bucket?

elements/lisk-p2p/src/utils.ts
```diff
@@ -0,0 +1,150 @@
...
+	address: string,
+	secret: number,
+): number | undefined => {
+	if (!isIPv4(address)) {
```
ishantiw: shouldn't we validate the IPs when we receive them in the discovery process and use the valid ones to calculate bucket?

mitsujutsu: IMO it wouldn't hurt to do an extra check here, since this is the entry point to other functions like `getBytes` below which depend on a valid IPv4 address.

elements/lisk-p2p/src/utils.ts
```diff
@@ -0,0 +1,146 @@
...
+	NET_OTHER,
+}
+
+/* tslint:disable no-magic-numbers */
+export const getByte = (address: string, n: number): number | undefined => {
+	if (n > 3) {
+		return undefined;
+	}
+
+	return parseInt(address.split('.')[n], 10);
```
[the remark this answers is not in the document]
mitsujutsu: It won't happen in this flow since isIPv4 check occurs before this function is ever called, but I will add it in case a user wants to use it independently.

elements/lisk-p2p/src/utils.ts
```diff
@@ -0,0 +1,287 @@
...
+const BYTES_128 = 128;
+const NEW_PEERS = 'new';
+const TRIED_PEERS = 'tried';
+
+export enum NETWORK {
+	NET_IPV4 = 0,
+	NET_PRIVATE,
+	NET_LOCAL,
+	NET_OTHER,
+}
```
jondubois: Can we make this a string enum for consistency with other parts of the lisk-p2p code?
The underlying reason is that integer enums are a pain to debug. The performance gain of using integers for this is insignificant when you consider the big picture.

elements/lisk-p2p/src/utils.ts
```diff
@@ -0,0 +1,287 @@
...
+	}
+
+	if (isIPv4(address)) {
+		return NETWORK.NET_IPV4;
+	}
+
+	return NETWORK.NET_OTHER;
+};
+
+interface GroupBytes {
```
jondubois: Maybe `IPGroupBytes` or `AddressGroupBytes` is more descriptive?

elements/lisk-p2p/src/utils.ts
```diff
@@ -0,0 +1,287 @@
...
+	readonly networkBytes: Buffer;
+	readonly targetABytes: Buffer;
+	readonly targetBBytes: Buffer;
+	readonly targetCBytes: Buffer;
+	readonly targetDBytes: Buffer;
+	readonly sourceABytes?: Buffer;
+	readonly sourceBBytes?: Buffer;
+}
+
+export const getGroupBytes = (options: {
```
jondubois: It's difficult to describe in words what this function does. Maybe it shouldn't exist. It's a small function and half of it is potentially cut off by an if condition.

elements/lisk-p2p/src/utils.ts
mitsujutsu: I would prefer this to remain an integer as the byte value is written into the buffer to be hashed later. In the bitcoin implementation they also use integers, and I would like to keep it consistent with that approach.

elements/lisk-p2p/src/utils.ts
```diff
@@ -0,0 +1,287 @@
...
+	address: string,
+	groupNumber: number,
+): number | undefined => {
+	if (groupNumber > 3) {
+		return undefined;
+	}
+
+	const group = parseInt(address.split('.')[groupNumber], 10);
+
+	if (!Number.isInteger(group)) {
```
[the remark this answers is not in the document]
mitsujutsu: That's what it was originally and it's a good point, but after @ishantiw comment I thought some of these `util` functions can be used by the user independently for their own purposes, in which case it might help to re-validate. What do you think?

elements/lisk-p2p/src/utils.ts
mitsujutsu: There was some overlap in logic between `getNetgroup` and `getBucket` which this function covered. I agree it's hard to verbalize the purpose of it. In this case maybe it's better to have a small overlap to make it easier to understand.

elements/lisk-p2p/src/utils.ts
```diff
@@ -0,0 +1,224 @@
...
+	const secondMod = peerListType === NEW_PEERS ? BYTES_128 : BYTES_64;
+	const secretBytes = Buffer.alloc(SECRET_BUFFER_LENGTH);
+	secretBytes.writeUInt32BE(secret, 0);
+	const network = getNetwork(targetAddress);
+	const networkBytes = Buffer.alloc(NETWORK_BUFFER_LENGTH);
+	networkBytes.writeUInt8(network, 0);
+
+	// Get bytes of ip address to bucket
+	const {
+		aBytes: targetABytes,
```
jondubois: It would be nice to add a comment somewhere to explain what ABytes, BBytes, CBytes and DBytes mean to save the user having to dig through the code.

elements/lisk-p2p/src/utils.ts
```diff
@@ -0,0 +1,225 @@
...
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import { createHash } from 'crypto';
```
shuse2: can we use `hash` function from `lisk-cryptography`?

elements/lisk-p2p/src/utils.ts
```diff
@@ -0,0 +1,225 @@
...
+import { isIPv4 } from 'net';
```
shuse2: can we use `lisk-validator`, it has validation method for IPv4 I believe?

elements/lisk-p2p/src/utils.ts
```diff
@@ -0,0 +1,225 @@
...
+			  ])
+			: Buffer.concat([
+					secretBytes,
+					networkBytes,
+					targetABytes,
+					targetBBytes,
+					kBytes,
+			  ]);
+
+	return cryptoHashSha256(bucketBytes).readUInt32BE(0) % secondMod;
```
shuse2: in general, do we need to use `0` offset?

elements/lisk-p2p/src/utils.ts
mitsujutsu: @jondubois brought up a good point that since we only need the `sha256` hash from node's `crypto` module, we should use that instead of installing a new dependency to keep the library lightweight. The same logic would apply to your comment for `lisk-validator` below.

shuse2: I would prefer to consistently use the same hash function for all the cases.
For example if we move to other hash algorithm, we should be able to change and detect everywhere. Also, if someone is using lisk-p2p, most likely they are using lisk-cryptography and lisk-validator in some form =)

elements/lisk-p2p/src/utils.ts
mitsujutsu: you don't need it to work but linter complains without it

elements/lisk-p2p/package-lock.json
```diff
@@ -144,6 +144,43 @@
...
+			},
+			"dependencies": {
+				"@types/node": {
+					"version": "11.11.2",
+					"resolved": "https://npm.lisk.io/@types%2fnode/-/node-11.11.2.tgz",
+					"integrity": "sha512-iEaHiDNkHv4Jrm9O5T37OYEUwjJesiyt6ZlhLFK0sbo4CLD0jyCOB4Pc2F9iD3MbW2397SLNxZKdDGntGaBjQQ=="
+				}
+			}
+		},
+		"@liskhq/lisk-cryptography": {
```
shuse2: I think it should not include this in the package-lock json.
can you try
`npx lerna bootstrap` in the root level?

elements/lisk-p2p/package.json
```diff
@@ -47,6 +47,7 @@
 		"enableLocalIPs": "sudo ./scripts/enableTestLocalIPs.sh 10 19"
 	},
 	"dependencies": {
+		"@liskhq/lisk-cryptography": "2.1.1",
```
diego-G: As we've done with `lodash`, I'd only include the specific module we use, which in this case is `hash`.

mitsujutsu: I think this is only possible for lodash due to the way they designed the library. But to address @jondubois earlier concern of `sodium-native` being huge, it's actually an optional dependency for `lisk-cryptography` and not installed by default.

mitsujutsu:
> There is room to optimize the distribution for newPeers but the current distribution should suffice for our use case.

I agree. Improvements can be worked on once `sourceGroup` hashing is back in effect.

PR #3981, opened 2019-07-18
Rate limit WebSocket messages - Closes #3975

### LINKED ISSUE #3975: Rate limit WebSocket messages
**Story**
As a node user, I would like to restrict traffic/ messages which I receive from any peer so that I can be safeguarded from the spamming.

**Acceptance Criteria**
1. Accept messages/ transaction blocks from each peer when it is not exceeded your Rate limit.
2. If a specific peer is reached his maximum Rate limit, then the specific peer connection should be interrupted.
3. The rate limit  (receiving limit) can vary from peer to peer.

### Expected behavior

There should be a simple rate-limiting mechanism to prevent nodes from spamming our WebSocket connection.

### Actual behavior

There is no rate limiting to prevent remote clients from spamming a node's WebSocket connection.

If we have a sufficiently large machine (or multiple machines), we can spam a specific node with WebSocket messages to make it consume CPU.

jondubois:
### What was the problem?

There was no rate limiting at the WebSocket level to block malicious peers.

### How did I fix it?

- Track the raw WebSocket message rate. If a peer tries to send more than `wsMaxMessageRate` messages per second, they will be disconnected with a forbidden message.
- Adjusted how the p2p request and p2p message rate is calculated to be consistent with the calculation of the `wsMaxMessageRate` and to increase precision.

### How to test it?

New integration test cases were added.

`npm run test:integration`

### Review checklist

* The PR resolves #3975
* Linting passes
* Tests pass


elements/lisk-p2p/src/peer/base.ts
```diff
@@ -173,21 +184,53 @@ export class Peer extends EventEmitter {
...
+				this.disconnect(FORBIDDEN_CONNECTION, FORBIDDEN_CONNECTION_REASON);
+
+				return;
+			}
+
+			this._rpcRates = new Map(
+				[...this._rpcCounter.entries()].map(entry => {
+					const rate = entry[1] / this._rateInterval;
+
+					return [entry[0], rate] as any;
```
ishantiw: isn't `entry[0]` is string type and we can use it as key for `rpcRates` type as `Map<string, number>`?

elements/lisk-p2p/src/p2p.ts
```diff
@@ -124,6 +124,7 @@ export const DEFAULT_POPULATOR_INTERVAL = 10000;
 export const DEFAULT_SEND_PEER_LIMIT = 25;
 // Max rate of WebSocket messages per second per peer.
 export const DEFAULT_WS_MAX_MESSAGE_RATE = 100;
+export const DEFAULT_WS_MAX_MESSAGE_RATE_PENALTY = 10;
```
mitsujutsu: Should penalty be `100`? Otherwise the actual number of messages allowed would be 10x

jondubois: @mitsujutsu `DEFAULT_WS_MAX_MESSAGE_RATE` means the rate above which we disconnect, not ban. Banning is more serious; we want to restrict it to intentional DoS attempts, not one-offs. For example, if someone is running a node which has bad configs or custom logic, we want to give the operator of that node a chance to fix their configs before they are banned from the network.

elements/lisk-p2p/test/integration/p2p.ts
```diff
@@ -1030,6 +976,193 @@ describe('Integration tests for P2P library', () => {
...
+							}
+						}
+						let peerRates = requestRates.get(p2p.nodeInfo.wsPort) || [];
+						peerRates.push(request.rate);
+						requestRates.set(p2p.nodeInfo.wsPort, peerRates);
+					});
+				}
+			});
+
+			it('should track the request rate correctly when receiving requests', async () => {
```
diego-G: Shall we add another test case to check if the peer interrupts the channel?

elements/lisk-p2p/src/p2p.ts
mitsujutsu: I see. So in practice they are allowed to exceed the max rate 10 times before they are banned?

elements/lisk-p2p/test/integration/p2p.ts
jondubois: @diego-G Ok resolved.

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -175,21 +187,54 @@ export class Peer extends EventEmitter {
...
+
+			if (this._wsMessageRate > this._peerConfig.wsMaxMessageRate) {
+				this.disconnect(FORBIDDEN_CONNECTION, FORBIDDEN_CONNECTION_REASON);
+				this.applyPenalty(this._peerConfig.wsMaxMessageRatePenalty);
+
+				return;
+			}
+
+			this._rpcRates = new Map(
+				[...this._rpcCounter.entries()].map(entry => {
```
ishantiw: maybe for clarity we can use `.map([key, rate]) =>` or `.map([key, value]) =>`?

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -175,21 +187,54 @@ export class Peer extends EventEmitter {
...
+					const rate = entry[1] / this._rateInterval;
+
+					return [entry[0], rate] as any;
+				}),
+			);
+			this._rpcCounter = new Map();
+
+			this._messageRates = new Map(
+				[...this._messageCounter.entries()].map(entry => {
```
ishantiw: same here

elements/lisk-p2p/src/peer/base.ts
ishantiw: we don't need to typecast as `any`, it doesn't give me warning

jondubois: It gives me warnings in vscode if I don't put `any`.

PR #3995, opened 2019-07-23
Add examples and tools for protocol specification tests

shuse2:
### Description
Adding the protocol-specs folder with the example of how to test the protocol.
Also, added the tool to create simple block fixtures.

### Review checklist

* The PR resolves #INSERT_ISSUE_NUMBER
* Linting passes
* Tests pass


protocol-specs/v1/01_balance_transfer.feature
```diff
@@ -0,0 +1,32 @@
+Feature: Balance transfer transaction included in a block
+
+Scenario: a valid balance transfer transaction
+    Given the valid blockchain with an account A with 2 LSK
```
ManuGowda: where do we have the association for the feature? meaning, if the feature spec says `Given the valid blockchain with an account A with 2 LSK`, how can one correlate the state of `Account A having 2 LSK`?
if there is no logical mapping and we just define feature specification it's not the correct way of handling BDD or feature specification.

A `BDD` or `Feature spec` should allow
- The specification to be modifiable, i.e if a user wants to change the from `Given the valid blockchain with an account A with 2 LSK` to `Given the valid blockchain with an account A with 10 LSK` or change the behavior from `When a block with a transfer transaction from account A to B with amount 1 LSK is processed` to `When a block with a transfer transaction from account X to Z with amount 10 LSK is processed`.
- The behaviors should be reusable, i.e, reoccurring behavior patterns as `Given` statement, or `Background` statement should be reusable.

The current proposal does not address these concerns, we should ensure the specification proposal should address these issues.

yatki: I'd advise not to use to feature files like in this PR. Ideally that for `Given the valid blockchain with an account A with 2 LSK` step `2` is a variable. So we need a handler function for  `Given the valid blockchain with an account A with ${amount} LSK` that updates the context by the matching arguments of the step. This way, we could reuse the steps. 

Also in a scenario there can be multiple assertions. I'm not sure how we can achieve having multiple assertions.

Maybe I didn't get the whole idea but i don't think it really matters anyway. When I see feature files I'd like to see step definitions and reuse them :)

I'm totally up for using feature files, I think it brings value but I'd prefer to have a proper use of them, not another customised solution.

shuse2: Maybe it's not good to use `feature` file, but what we want to achieve here is to have the `spec.json` file which is the fixture that cannot be changed once it's created for testing the protocol in the neutral way (ie any language can use it to test their implementation).

However, we need a guidance and a documentation of what the fixture is testing.
In order to have that, I think the gherkin format is good for it.

It might be confusing in a way that it sounds like it should use the cucumber, so we could change it to different format.
this feature file is not meant to be directly used for testing.

ManuGowda: I think it's good to have a planning first approach than, implementation first as the `spec of spec` going to become the standard defacto for the future protocol implementation. Let's have a broader brainstorming session before we jump into solutions. If we want to use `BDD` or `Feature` specification standard then let's not find a hack, rather find an approach which can help everyone to understand it better to make it a success.

PR #3997, opened 2019-07-23
Change license from GPL v3 to Apache 2.0

### LINKED ISSUE #3994: Change license from GPLv3 to Apache 2.0
Together with Lisk SDK 2.0.0 release, we are changing the licensing scheme from [GPLv3](https://opensource.org/licenses/GPL-3.0) to [Apache 2.0](https://opensource.org/licenses/Apache-2.0)

We want to change the license of the newly unified Lisk SDK product to a more permissive one. We decided on Apache 2.0, because:
- It is a permissive (same as MIT, BSD-2) and therefore better facilitates the use of the Lisk SDK within closed-source/commercial applications
- It is wordier so there is less room for misinterpretation
- It does not force hardware vendors to do anything
- It explicitly grants patents (whereas MIT does not)
- It has the Apache Foundation behind it
 
Checklist:

- [ ] Change license footer, with amendment to reflect change of license
- [ ] Add license field to `package.json`
- [ ] Add license badge to `README.md`
- [ ] Replace `LICENSE` file within project
`2.0.0`

MaciejBaj:
### Review checklist

* The PR resolves #3994
* Linting passes
* Tests pass


fchavant:
@MaciejBaj we use pm2 (for instance in [commander](https://github.com/LiskHQ/lisk-sdk/blob/4738ff326dfca1d01f65923ec917113256c70d86/commander/package.json#L119)) which is distributed under [AGPLv3](https://github.com/Unitech/pm2/blob/master/LICENSE); doesn't switching to Apache-2.0 make us imcompatible with the GPL? Can we keep on using PM2?

MaciejBaj:
@fchavant should be fine as it's only a tool we use, don't use/modify the code.

fchavant:
@MaciejBaj IANAL however we import `pm2` and use its API which is AGPLv3 licensed and I think that is the same as linking against the GPL which is not possible using Apache-2.0 (see https://www.apache.org/licenses/GPL-compatibility.html)

PR #4008, opened 2019-07-25
Merge development to feature/introduce_bft_consensus branch - Closes #4011

### LINKED ISSUE #4011: Merge development to feature/introduce_bft_consensus branch
### Expected behavior
Test should pass after the merge

### Actual behavior
There are many conflicts in the merge, so we have to resolve these all.

nazarhussain:
### What was the problem?
Conflicts: 
- chain/blocks/blocks.js
- chain/blocks/block.js
- chain/chain.js
- framework/errors.js

### How did I fix it?

Resolved the conflicts by refactoring a bit of logic. 

### Review checklist

* The PR resolves #4011
* Linting passes
* Tests pass


framework/src/modules/chain/blocks/block_v1.js
```diff
@@ -30,6 +30,9 @@ const BigNum = require('@liskhq/bignum');
 const { validator } = require('@liskhq/lisk-validator');
 const { validateTransactions } = require('../transactions');
 const blockVersion = require('./block_version');
+
+// TODO: Due to UNKNOWN issue the require for utils returned empty object
+delete require.cache[require.resolve('./utils')];
```
SargeKhan: This is wrong. We shouldn't try to find a hack to resolve this thing. The problem occurs due to circular dependency introduced by utils.js file depending on block.js file which in turn depends on blockV1.js and blockV2.js and they import utils.js again.
A simpler fix would be to duplicate code and declare sortTransactions function within blockV1 and blockV2. I think it makes sense since it is possible that different versions of blocks sort transactions differently.

SargeKhan: Ignorre this comment, since it was fixed in the later commit.

nazarhussain: Its already resolved.

michielmulders: Is this not removed in one of the other commits?

nazarhussain: Yes its removed, this change here is outdated.

Issue #3987, opened 2019-07-22
Fix and unskip the tests for applyTransaction function defined in transactionHandler


SargeKhan:
### Expected behavior
There should be unit tests for the applyTransaction function in the transactionHandler.js file and they should not be skipped.
### Actual behavior
The tests defined for applyTransactions are skipped because they are not fixable with the current code structure.
### Steps to reproduce
Check the skipped tests.
### Which version(s) does this affect? (Environment, OS, etc...)
release/2.2.0

**Grooming Notes**

- It is really difficult to mock and stub now. We want to make the code more testable.
- Goal: fix tests but also refactor code in order to unskip skipped tests
- Needs discussion as in new block processor we want a new kind of interface on how we use state store. Figure out a way to make it easier to test.
- Also, we have to update the usages of this function.
- How big? For applyTransaction, multiple tests are skipped 15.

### Tasks list
- [ ] Refactor transactionHandler file to make code more testable.
- [ ] Update the usages of transactionHandler functions, and relevant tests.
- [ ] Unskip tests for applyTransactions function.
- [ ] Ensure transactionHandler file is covered with unit tests.

PR #4010, opened 2019-07-25
Snapshotting and syncing fails on 2.2 - Closes #4003

### LINKED ISSUE #4003: Snapshotting and syncing fails on 2.2
### Expected behavior
Expect to successfully finish snapshotting and syncing when executed on testnet

### Actual behavior
It fails with the following message:
```
07:36:45.518Z DEBUG lisk-framework: Using delegate list from the cache for round 5320
07:36:45.519Z  INFO lisk-framework:
    Fork { delegate:
       '410ebf1736f1e7f8c5a3ce5364626e62731b105943ed94d2294d33304a496fd3',
      block:
       { id: '5010245806505876363',
         timestamp: 8719180,
         height: 537249,
         previousBlock: '14954058649670137792' },
      cause: 3 }
```

### Steps to reproduce
Start snapshotting or syncing on testnet

`2.2`

lsilvs:
### What was the problem?
Exceptions was not being passed causing syncing and snapshotting to fails

### How did I fix it?
Added the `exceptions` object to `votes.apply` inside `applyTransactions`

### How to test it?
Perform syncing and snapshotting on testnet.

### Review checklist

* The PR resolves #4003 
* Linting passes
* Tests pass

### DIFF
--- a/framework/test/mocha/integration/process_transactions.js
+++ b/framework/test/mocha/integration/process_transactions.js
@@ -26,6 +26,7 @@ const transactionsModule = require('../../../src/modules/chain/transactions');
 const {
 	TransactionInterfaceAdapter,
 } = require('../../../src/modules/chain/interface_adapters');
+const votes = require('../../../src/modules/chain/transactions/votes');
 
 const interfaceAdapters = {
 	transactions: new TransactionInterfaceAdapter(registeredTransactions),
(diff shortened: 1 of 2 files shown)


framework/test/mocha/integration/process_transactions.js
```diff
@@ -346,6 +347,16 @@ describe('processTransactions', () => {
 						);
 					});
 				});
+
+				it('should call votes.apply with correct params', async () => {
```
SargeKhan: Tests which spy or stub should be unit tests. No?

SargeKhan: Here we can test that stateStore value is updated as expected if we want this to be an integration test

lsilvs: I added this test here because the unit test for this method was skipped as reported in the issue https://github.com/LiskHQ/lisk-sdk/issues/3987

shuse2: I think spy is fine for integration test

PR #4017, opened 2019-07-29
Add get minimal peers list data for discovery process - Closes #3948

### LINKED ISSUE #3948: Only send back the port and ip when peer requests our peer list
### Expected behavior

When doing discovery, the peer should only give us a list of IPs and ports of its peers; it should not include the other fields from `P2PPeerInfo`.

We will fetch the remaining `P2PPeerInfo` fields after we connect to the peer and make a `status` call to it.

### Actual behavior

Right now, for peers in our newPeers list, it's possible that the fields could be wrong because a different peer could give us fake values for the fields (they come from third-party peers and not the peer itself).

Currently, we shouldn't trust fields of `P2PPeerInfo` objects in the `newPeers` list.

diego-G:
### What was the problem?

Discovery process was handling a lot unneeded information which also could make us prone to attacks.  

### How did I solve it?

Create a new method only used in discovery process. Peers are able to send back peer lists and minimal peer list.

### How to manually test it?

`npm run test:integration`

### Review checklist

- [X] The PR resolves #3948
- [X] Relevant integration / functional tests are added
- [X] Commit messages follow the [commit guidelines](CONTRIBUTING.md#git-commit-messages)


elements/lisk-p2p/src/peer/base.ts
```diff
@@ -93,7 +93,7 @@ export const REMOTE_EVENT_MESSAGE = 'remote-message';
 
 export const REMOTE_RPC_UPDATE_PEER_INFO = 'updateMyself';
 export const REMOTE_RPC_GET_NODE_INFO = 'status';
-export const REMOTE_RPC_GET_ALL_PEERS_LIST = 'list';
+export const REMOTE_RPC_GET_MINIMAL_PEERS_LIST = 'minimalList';
```
ishantiw: why we are giving a new RPC name? we have `getPeers` RPC call that has the same requirement.

elements/lisk-p2p/src/p2p.ts
```diff
@@ -759,26 +764,40 @@ export class P2P extends EventEmitter {
...
-				};
-			},
-		);
-		const protocolPeerInfoList: ProtocolPeerInfoList = {
-			success: true,
-			peers,
-		};
-
-		request.end(protocolPeerInfoList);
+		if (request.procedure === REMOTE_RPC_GET_MINIMAL_PEERS_LIST) {
```
ishantiw: As mentioned above, we can use `REMOTE_RPC_GET_MINIMAL_PEERS_LIST ` for `getPeers` RPC for the 2.3 nodes and else part to support old versions

elements/lisk-p2p/src/peer/base.ts
diego-G: We're keeping them separated. I'll rename back the `REMOTE_RPC_GET_ALL_PEERS_LIST` to make it clear.

elements/lisk-p2p/src/p2p.ts
```diff
@@ -27,7 +27,7 @@ interface SCServerUpdated extends SCServer {
 import {
 	constructPeerId,
 	constructPeerIdFromPeerInfo,
-	REMOTE_RPC_GET_ALL_PEERS_LIST,
+	REMOTE_RPC_GET_MINIMAL_PEERS_LIST,
```
pablitovicente: This variable name seems to indicate the list is minimal i.e. reduced. Maybe rename to `REMOTE_RPC_GET_BASIC_PEERS_LIST ` ?

elements/lisk-p2p/src/p2p.ts
```diff
@@ -759,26 +764,40 @@ export class P2P extends EventEmitter {
...
-			peers,
-		};
-
-		request.end(protocolPeerInfoList);
+		if (request.procedure === REMOTE_RPC_GET_BASIC_PEERS_LIST) {
+			const basicPeers = this._pickRandomPeers(randomPeerCount).map(
+				(peerInfo: P2PPeerInfo): P2PPeerInfo =>
+					// Discovery process only require minmal peers data
+					({
+						ipAddress: peerInfo.ipAddress,
```
lsilvs: Shouldn't it return `ip` instead of `ipAddress`?

```suggestion
						ip: peerInfo.ipAddress,
```

elements/lisk-p2p/src/p2p.ts
```diff
@@ -759,26 +764,40 @@ export class P2P extends EventEmitter {
...
+						wsPort: peerInfo.wsPort,
+					}),
+			);
+			const peerInfoList: P2PBasicPeerInfoList = {
+				success: true,
+				peers: basicPeers,
+			};
+			request.end(peerInfoList);
+		} else {
+			// TODO: Remove fields that are specific to the current Lisk protocol.
```
lsilvs: I guess this comment doesn't apply anymore, right?

elements/lisk-p2p/src/p2p.ts
diego-G: In this case @lsilvs, we need `ipAddress` instead... @jondubois we should get rid of this inconsistency to avoid misunderstandings.

elements/lisk-p2p/src/p2p.ts
diego-G: @ishantiw could you shed us some light here? I blame this TODO and it comes from https://github.com/LiskHQ/lisk-sdk/commit/7ee21334aad36a7fbd0c244a2042c9785c124c2c

elements/lisk-p2p/src/p2p.ts
ishantiw: in p2p-lib is ipAddress and in framework context it's `ip`, but yes we need to get rid of this inconsistency as in framework its more prone to errors because of absence of TS ;)

elements/lisk-p2p/src/p2p.ts
ishantiw: I think this comment can be removed

elements/lisk-p2p/src/p2p.ts
```diff
@@ -759,26 +764,40 @@ export class P2P extends EventEmitter {
...
+			const peers = this._pickRandomPeers(randomPeerCount).map(
+				(peerInfo: P2PPeerInfo): ProtocolPeerInfo => {
+					const { ipAddress, ...peerInfoWithoutIp } = peerInfo;
+
+					// The options property is not read by the current legacy protocol but it should be added anyway for future compatibility.
```
ishantiw: Also this comment can be removed

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -91,7 +91,7 @@ export const REMOTE_EVENT_MESSAGE = 'remote-message';
 
 export const REMOTE_RPC_UPDATE_PEER_INFO = 'updateMyself';
 export const REMOTE_RPC_GET_NODE_INFO = 'status';
-export const REMOTE_RPC_GET_ALL_PEERS_LIST = 'list';
+export const REMOTE_RPC_GET_BASIC_PEERS_LIST = 'basicList';
```
ishantiw: But we have to support `getPeers` RPC in LIP and it's doing the same job as `getPeers` is suppose to

diego-G: We still do. It is answer from the peer pool.

elements/lisk-p2p/src/validation.ts
```diff
@@ -115,6 +132,23 @@ export const validatePeerInfoList = (
...
+	if (!rawBasicPeerInfoList) {
+		throw new InvalidRPCResponseError('Invalid response type');
+	}
+	const { peers } = rawBasicPeerInfoList as RPCPeerListResponse;
+
+	if (Array.isArray(peers)) {
+		const peerList = peers.map<P2PPeerInfo>(validateBasicPeerInfo);
+
+		return peerList;
+	} else {
```
mitsujutsu: Prefer avoiding `else` by checking `!Array.isArray(peers))`

jondubois: I agree that it's better but if we do this we should also update `validatePeerInfoList` because it also has a similar else condition. Maybe we can clean this up later as part of a different PR?

diego-G: @mitsujutsu I agree with your comment but in this case, the tests fail when we follow your approach.

PR #4018, opened 2019-07-29
Remove stdin and gc - Closes #3839

### LINKED ISSUE #3839: Investigate the impact and remove process.stdin.resume()
### Expected behavior
Currently we have a usage of `process.stdin.resume()` in chain module. Which actually come from the legacy code. 

https://github.com/LiskHQ/lisk-sdk/blob/942d320795d13cf0ad83e33d8e78c07357ae8700/framework/src/modules/chain/chain.js#L47-L48

While writing storage tests in `jest` I found out the test runner hangs up after the successful execution and it jest option `--detectOpenHandles` leads to above line. 

For now I had limited it not block stdin during the tests. We should investigate the impact of it in the production environment and then remove it. 

### Actual behavior
We should not be holding off the `stdin` in any case.

3.0.0

shuse2:
### What was the problem?
`process.stdin.resume` and `gc` was used without reason.
We do not use `--expose-gc` so, `gc` function is not exposed. Therefore, it is no longer used.
Also, we do not make application on hold by `stdin.resume` anywhere. Therefore, it's no longer needed.

### How did I solve it?
Remove those two functions.

### How to manually test it?
Start the application and see if it works fine
`node framework/test/test_app`

### Review checklist

- [x] The PR resolves #3839 

### DIFF
--- a/framework/src/modules/chain/chain.js
+++ b/framework/src/modules/chain/chain.js
@@ -48,15 +48,6 @@ const { Transport } = require('./transport');
 const syncInterval = 10000;
 const forgeInterval = 1000;
 
-// Begin reading from stdin
-process.stdin.resume();
-
-if (typeof gc !== 'undefined') {
-	setInterval(() => {
-		gc(); // eslint-disable-line no-undef
-	}, 60000);
-}
-
 /**
  * Chain Module
  *


ManuGowda:
@shuse2 I think `gc` was used during the `memory leak` testing, we used the `--expose-gc` flag and captured the snapshot for memory leak detection.

shuse2:
memory leak detection doesn't use this `gc` and it is till able to take heapdump

shuse2:
@nazarhussain 
We are not using the flag on any network, so it will never be triggered forcefully.

PR #4019, opened 2019-07-29
Store deleted block in temp table - Closes #3772

### LINKED ISSUE #3772: Blocks need to be stored in temporary table on block deletion
### Expected behavior
Whenever we have found and verified the `commonBlock`, we will store the blocks from this commonHeight til the height of the tip in a temporary table so we can start applying the newly received blocks.

**Implement the functionality to**:
- delete blocks from the current `blocks` table
- create a backup of the deleted blocks in a temp table

`feature/improve-bft`

SargeKhan:
### What was the problem?
Blocks which are deleted as part of fork recovery mechanism in BFT need to be stored in the temp table. This temp table was non existent in the application. 
### How did I solve it?
Created the required temp table, its migration and exposed this functionality from chain module.
### How to manually test it?
Run the unit tests for entity as well as the deleteLastBlockAndStoreInTemp function
### Review checklist

- [ ] The PR resolves #3772 


SargeKhan:
I have written tests in jest while importing some libraries from mocha. This is an antipattern, i concede but i choose to favor having tests than not. Feel free to suggest a better alternative.

SargeKhan:
Another note, the diff on github is not correct as it shows too many changes. I am listing my commits which need to be reviewed for this PR to be merged:
- declare database entity for block_temp table 322aabe
- create sql querires for crd operation on block_temp table 811e72c
- create migration for block_temp table 273e57c
- integrate block_temp entity in application startup and test faa3385
- add test for block_temp entity 384ed4a
- fix tests for deleteLastBlockAndStoreInTemp 781c314

framework/src/modules/chain/blocks/chain.js
```diff
@@ -141,16 +141,48 @@ class BlocksChain {
...
+					this.storage,
+					this.interfaceAdapters,
+					this.genesisBlock,
+					this.roundsModule,
+					this.slots,
+					lastBlock,
+					this.exceptions,
+					tx
+				);
+				const parsedDeletedBlock = parseBlockToJson(lastBlock);
```
nazarhussain: You returned `previousBlock` but to store used `lastBlock`, any particular reason for it?

framework/src/modules/chain/blocks/chain.js
```diff
@@ -141,16 +141,48 @@ class BlocksChain {
...
+				const blockTempEntry = {
+					id: parsedDeletedBlock.id,
+					height: parsedDeletedBlock.height,
+					fullBlock: parsedDeletedBlock,
+				};
+				await this.storage.entities.BlockTemp.create(blockTempEntry, {}, tx);
```
nazarhussain: Would be nice to have `storage.entities.TempBlock` rather `storage.entities.BlockTemp`

framework/src/modules/chain/blocks/chain.js
```diff
@@ -195,11 +224,29 @@ const saveBlock = async (storage, block, tx) => {
...
 
-	parsedBlock.transactions.map(transaction => {
+	parsedBlock.transactions.forEach(transaction => {
 		transaction.blockId = parsedBlock.id;
 		return transaction;
 	});
 
+	parsedBlock.transactions = parsedBlock.transactions.map(transaction =>
+		transaction.toJSON()
+	);
```
nazarhussain: We should override the block id after this block 

```
	parsedBlock.transactions.forEach(transaction => {
 		transaction.blockId = parsedBlock.id;		 		
 		return transaction;
 	});

```

framework/src/modules/chain/components/storage/entities/block_temp.js
```diff
@@ -0,0 +1,170 @@
...
+ * @property {number} [height_eql]
+ * @property {number} [height_ne]
+ * @property {number} [height_gt]
+ * @property {number} [height_gte]
+ * @property {number} [height_lt]
+ * @property {number} [height_lte]
+ * @property {number} [height_in]
+ */
+
+class BlockTemp extends BaseEntity {
```
nazarhussain: Please rename it to `TempBlock`

framework/src/modules/chain/components/storage/entities/block_temp.js
```diff
@@ -0,0 +1,170 @@
...
+	delete: 'block_temp/delete.sql',
+	get: 'block_temp/get.sql',
+};
+
+/**
+ * Block temp
+ * @typedef {Object} block_temp
+ * @property {string} id
+ * @property {number} height
+ * @property {json} block
```
nazarhussain:
```suggestion
 * @property {json} fullBlock
```

framework/src/modules/chain/components/storage/entities/block_temp.js
```diff
@@ -0,0 +1,170 @@
...
+	 * @param {Object} data
+	 * @param {Object} [_options]
+	 * @param {Object} [tx] - Transaction object
+	 * @return {null}
+	 */
+	// eslint-disable-next-line no-unused-vars
+	create({ height, id, fullBlock }, _options = {}, tx = null) {
+		assert(height && Number.isInteger(height), 'height must be a number');
+		assert(id && typeof id === 'string', 'id must be a string');
+		assert(typeof fullBlock === 'object', 'block must be an object');
```
nazarhussain: Typeof works good only for primitives, I would prefer to use 
``
fullBlock instanceof Object
``

framework/src/modules/chain/components/storage/sql/block_temp/create.sql
```diff
@@ -0,0 +1,19 @@
...
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+INSERT INTO block_temp (
+	${fields:raw}
```
nazarhussain: As we have always fixed set of fields, then would prefer to hardcode the field names in SQL. When you have dynamic fields based on conditions then we calculate it JS and pass it as params.

framework/src/modules/chain/init_steps/bootstrap_storage.js
```diff
@@ -37,6 +38,7 @@ module.exports = async ({ components: { storage, logger } }, accountLimit) => {
 			replaceExisting: true,
 		});
 		storage.registerEntity('ChainMeta', ChainMeta);
+		storage.registerEntity('BlockTemp', BlockTemp);
```
nazarhussain: As mentioned above 
```suggestion
		storage.registerEntity('TempBlock', TempBlock);
```

framework/src/modules/chain/loader.js
```diff
@@ -283,19 +283,12 @@ class Loader {
...
-		} catch (p2pError) {
-			this.logger.error('Failed to load block from network', p2pError);
-			return [];
-		}
+		const { data } = await this.channel.invoke('network:request', {
+			procedure: 'blocks',
+			data: {
+				lastBlockId: lastBlock.id,
+			},
+		});
```
nazarhussain: Why this change appears in this PR?

framework/src/modules/chain/migrations/sql/20190705101215_create_block_temp_table.sql
```diff
@@ -0,0 +1,26 @@
...
+
+
+/*
+  DESCRIPTION: Creates meta table for BFT.
+
+  PARAMETERS: None
+*/
+
+CREATE TABLE IF NOT EXISTS "block_temp" (
+	"id" character varying(20) NOT NULL,
```
nazarhussain:
```suggestion
	"id" VARCHAR(20) PRIMARY KEY,
```

framework/src/modules/chain/migrations/sql/20190705101215_create_block_temp_table.sql
```diff
@@ -0,0 +1,26 @@
...
+  "height" integer NOT NULL,
```
nazarhussain:
```suggestion
  "height" INT NOT NULL,
```

framework/src/modules/chain/transactions/transactions_handlers.js
```diff
@@ -168,7 +168,7 @@ const applyTransactions = (storage, exceptions) => async (transactions, tx) => {
 				);
 			}
 			if (transactionResponse.status === TransactionStatus.OK) {
-				votes.apply(stateStore, transaction);
+				votes.apply(stateStore, transaction, exceptions);
```
nazarhussain: Don't think this change belongs to scope of this PR.

nazarhussain:
@SargeKhan Change the base branch on GitHub and then set base again, hopefully this will fix the problem on diff.

framework/src/modules/chain/migrations/sql/20190705101215_create_block_temp_table.sql
michielmulders: Regarding Nazar's comments, the table name in the SQL files should be changes as well to `temp_block`

framework/src/modules/chain/components/storage/entities/block_temp.js
limiaspasdaniel: I think the `,` can be removed

framework/src/modules/chain/components/storage/entities/block_temp.js
limiaspasdaniel: Why is `_options` underscored?

framework/src/modules/chain/blocks/chain.js
limiaspasdaniel: I think `blockToJSON` or `convertBlockToJSON` suits as a better name, as `parse` is only meant to describe the act of reading and understanding the contents of the block

framework/src/modules/chain/components/storage/entities/block_temp.js
limiaspasdaniel: I think `json` should be `object` instead.

framework/src/modules/chain/blocks/chain.js
```diff
@@ -195,11 +224,29 @@ const saveBlock = async (storage, block, tx) => {
 	parsedBlock.previousBlockId = parsedBlock.previousBlock;
 	delete parsedBlock.previousBlock;
 
-	parsedBlock.transactions.map(transaction => {
+	parsedBlock.transactions.forEach(transaction => {
```
limiaspasdaniel: Avoid using forEach if possible.

framework/src/modules/chain/blocks/chain.js
```diff
@@ -195,11 +224,29 @@ const saveBlock = async (storage, block, tx) => {
...
+};
+
+/**
+ * Save block with transactions to database.
+ *
+ * @param {Object} block - Full normalized block
+ * @param {function} cb - Callback function
+ * @returns {Function|afterSave} cb - If SQL transaction was OK - returns safterSave execution, if not returns callback function from params (through setImmediate)
+ * @returns {string} cb.err - Error if occurred
+ */
```
limiaspasdaniel: These `jsdocs` are outdated. I don't know if it's the scope of this PR but maybe we could update them.

framework/src/modules/chain/blocks/chain.js
SargeKhan: @nazarhussain I am keeping the same interface as `deleteLastBlock` since soon we might be replacing the functionality of deleteLastBlock with thiis function.

SargeKhan: @limiaspasdaniel what do you mean? I don't understand your point?

framework/src/modules/chain/blocks/chain.js
SargeKhan: What do you suggest?

framework/src/modules/chain/blocks/chain.js
SargeKhan: Why so?

framework/src/modules/chain/blocks/chain.js
limiaspasdaniel: Ignore my comment sorry, I read the name again and it indeed makes sense :)

framework/src/modules/chain/blocks/chain.js
limiaspasdaniel: `for (transaction of parsedBlock.transactions)`. However we haven't agreed about this yet so feel free to adopt it if the scope of the PR doesn't get too big.

framework/src/modules/chain/components/storage/entities/block_temp.js
SargeKhan: simply using the existing interface defined in the rest of the application

framework/src/modules/chain/components/storage/sql/block_temp/create.sql
SargeKhan: Can you give me an example? I looked at create queries for other entities and they follow the same practice. I am not sure how you'd like me to change this particular query.

limiaspasdaniel:
@SargeKhan something is wrong with the tests, could you please take a look?

framework/src/modules/chain/blocks/chain.js
nazarhussain: Because I am not sure if `transaction.toJSON()` is preserving that custom attribute we assigned just above manually.

framework/src/modules/chain/blocks/chain.js
nazarhussain: I posted earlier to Slack about this topic but it was not concluded. Let's conclude the topic outside PR and then all follow same pattern.

framework/src/modules/chain/components/storage/entities/temp_block.js
```diff
@@ -0,0 +1,170 @@
...
+
+const sqlFiles = {
+	create: 'temp_block/create.sql',
+	delete: 'temp_block/delete.sql',
+	get: 'temp_block/get.sql',
+};
+
+/**
+ * Block temp
+ * @typedef {Object} temp_block
```
nazarhussain:
```suggestion
 * @typedef {Object} TempBlock
```

framework/src/modules/chain/components/storage/entities/temp_block.js
```diff
@@ -0,0 +1,170 @@
...
+ * @property {string} id
+ * @property {number} height
+ * @property {object} fullBlock
+ */
+
+/**
+ * Round Filters
+ * @typedef {Object} filters.Round
```
nazarhussain:
```suggestion
 * @typedef {Object} filters.TempBlock
```

framework/src/modules/chain/components/storage/entities/temp_block.js
```diff
@@ -0,0 +1,170 @@
...
+
+		this.sqlDirectory = path.join(path.dirname(__filename), '../sql');
+
+		this.SQLs = this.loadSQLFiles('temp_block', sqlFiles, this.sqlDirectory);
+	}
+
+	/**
+	 * Create temp_block row entry
+	 *
+	 * @param {Object} data
```
nazarhussain: * @param {TempBlock} data

framework/src/modules/chain/components/storage/entities/temp_block.js
```diff
@@ -0,0 +1,170 @@
...
+
+		return this.adapter.executeFile(
+			this.SQLs.create,
+			{ createSet, fields },
+			{ expectedResultCount: 0 },
+			tx
+		);
+	}
+
+	async get(filters = {}, options = {}, tx = null) {
```
nazarhussain: Add JS doc for this function 

```
         * @param {filters. TempBlock|filters.TempBlock[]} [filters = {}]
	 * @param {Object} [options = {}] - Options to filter data
	 * @param {Number} [options.limit=10] - Number of records to fetch
	 * @param {Number} [options.offset=0] - Offset to start the records
	 * @param {string} [options.sort] - Sort key for transaction e.g. amount:asc, amount:desc
	 * @param {Boolean} [options.extended=false] - Get extended fields for entity
	 * @param {Object} tx - Database transaction object
	 * @return {Promise.< TempBlock[], Error>}
```

framework/src/modules/chain/components/storage/entities/temp_block.js
```diff
@@ -0,0 +1,170 @@
...
+		return this._getResults(filters, options, tx);
+	}
+
+	async getOne(filters = {}, options = {}, tx = null) {
```
nazarhussain: Please here as well, add JS doc.

framework/src/modules/chain/components/storage/entities/temp_block.js
```diff
@@ -0,0 +1,170 @@
...
+			params,
+			{ expectedResultCount },
+			tx
+		);
+	}
+
+	/**
+	 * Delete records with following conditions
+	 *
+	 * @param {filters.Round} filters
```
nazarhussain:
```suggestion
	 * @param {filters.TempBlock} filters
```

Issue #3970, opened 2019-07-16
Generate 32-bit entropy random secret on node startup


mitsujutsu:
### Expected behavior

Upon starting a lisk node, a 32-bit entropy random secret should be generated. The secret will be used as a seed when determining peer `buckets` and `netgroups`.

This feature did not exist prior to v2.3.

### Actual behavior

There is no secret generated.

### Which version(s) does this affect? (Environment, OS, etc...)

2.3

PR #4023, opened 2019-07-31
Netgroup inbound peer eviction and secret generation - Closes #3976

### LINKED ISSUE #3976: Use netgrouping for inbound evictions
### Expected behavior

[Related](https://github.com/LiskHQ/lisk-sdk/pull/3968)

Inbound peers should be evicted based on netgroup.

### Actual behavior

Currently peers in peer pool not yet evicted based on netgroup.

v 2.3

mitsujutsu:
### What was the problem?

Eviction of inbound peers based on `netgroup` value was not implemented. 

Node required a random 32-bit entropy secret on startup.

### How did I solve it?

Peers are now evicted based on 4 categories including `netgroup` value as per new protocol.
Secret generated on node startup.

### How to manually test it?

Start node. Set `netgroupProtectionRatio` to a high value for `P2PConfig`. Check which peers are evicted.

### Review checklist

- [x] The PR resolves #3976 #3970 


elements/lisk-p2p/src/p2p_types.ts
```diff
@@ -40,6 +40,7 @@ export interface P2PPenalty {
 export interface P2PPeerInfo {
 	readonly ipAddress: string;
 	readonly wsPort: number;
+	readonly secret?: number;
```
jondubois: We should not put the secret inside the P2PPeerInfo or else it will be shared with other peers and they can use it to guess our buckets.

elements/lisk-p2p/src/p2p.ts
```diff
@@ -620,6 +626,7 @@ export class P2P extends EventEmitter {
 					wsPort,
 					height: queryObject.height ? +queryObject.height : 0,
 					version: queryObject.version,
+					secret: this._config.secret,
```
jondubois: It shouldn't be in the PeerInfo because we share it with the network.

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -184,6 +184,9 @@ export class Peer extends EventEmitter {
 		this._id = constructPeerId(this._ipAddress, this._wsPort);
 		this._height = peerInfo.height ? (peerInfo.height as number) : 0;
 		this._reputation = DEFAULT_REPUTATION_SCORE;
+		this._netgroup = peerInfo.secret
```
jondubois: Maybe secret should be in peerConfig not peerInfo?

framework/src/modules/network/network.js
```diff
@@ -95,6 +97,13 @@ module.exports = class Network {
 			{},
 			{ limit: null }
 		);
+
+		// If tried peers were stored in db retrieve pervious secret, else generate new one
+		this.secret =
+			previousPeers && previousPeers.length > 0
+				? previousPeers[0].secret
```
jondubois: We should not storing our node's secret on the peers.
Maybe it's better to not persist the secret at all.

The purpose of persisting the old secret is because generating a new secret will cause the balance of peers across buckets to change and it might cause evictions of some of our old peers; this is not a problem. Also we have tests to prove that the bucket system has approximately even distribution so there shouldn't be too many evictions (and we can definitely tolerate some; we will rediscover peers anyway).

So maybe we should not persist the secret; that way maybe we don't even need to pass it to the peer objects or put it in the peerInfo.

mitsujutsu: I think it would be better to persist the secret rather than not, but I agree the benefit is marginal. We can put it off until 3.0.

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -566,9 +571,19 @@ export class PeerPool extends EventEmitter {
 		})();
 	}
 
-	// TODO: Protect peers by netgroup
 	private _selectPeersForEviction(): Peer[] {
 		const peers = [...this.getPeers(InboundPeer)];
+		// Cannot predict which netgroups will be protected
+		const filteredPeersByNetgroup = this._peerPoolConfig.netgroupProtectionRatio
```
shuse2: shouldn't `netgroupProtectionRatio` always exist?

mitsujutsu:
> Shouldn't this issue also save the secret to the database?

After discussion with @jondubois we decided to delay persistence to db until a later release. The effect was minimal enough that it can be seen as an optimization rather than necessity.

elements/lisk-p2p/src/peer_pool.ts
mitsujutsu: User has the option to set it to `0` which means they want to disable it.

.github/ISSUE_TEMPLATE/ISSUE_TEMPLATE_BUG_REPORT.md
```diff
@@ -1,10 +1,9 @@
 ---
-name: 
+name:
```
diego-G: Is this intended?

.github/ISSUE_TEMPLATE/ISSUE_TEMPLATE_FEATURE.md
```diff
@@ -1,10 +1,9 @@
 ---
-name: 
+name:
```
diego-G: Same as above.

elements/lisk-p2p/src/p2p.ts
```diff
@@ -12,7 +12,7 @@
  * Removal or modification of this copyright notice is prohibited.
  *
  */
-
+import { randomBytes } from 'crypto';
```
diego-G: Can we import this from `lisk-cryptography` as we have discussed here? https://github.com/LiskHQ/lisk-sdk/pull/4030#discussion_r309680239

elements/lisk-p2p/src/p2p.ts
```diff
@@ -131,11 +131,14 @@ const BASE_10_RADIX = 10;
...
 export const DEFAULT_MAX_INBOUND_CONNECTIONS = 100;
 export const DEFAULT_OUTBOUND_SHUFFLE_INTERVAL = 300000;
+export const DEFAULT_PEER_PROTECTION_FOR_NETGROUP = 0.034;
 export const DEFAULT_PEER_PROTECTION_FOR_LATENCY = 0.068;
 export const DEFAULT_PEER_PROTECTION_FOR_USEFULNESS = 0.068;
 export const DEFAULT_PEER_PROTECTION_FOR_LONGEVITY = 0.5;
 export const DEFAULT_MIN_PEER_DISCOVERY_THRESHOLD = 100;
 export const DEFAULT_MAX_PEER_DISCOVERY_RESPONSE_SIZE = 1000;
+export const SECRET_LENGTH = 4;
+export const DEFAULT_SECRET = randomBytes(SECRET_LENGTH).readUInt32BE(0);
```
diego-G: We don't have to export this two values.

diego-G: Could you also use a more explicit name or comment its meaning?

elements/lisk-p2p/test/integration/p2p.ts
```diff
@@ -2110,9 +2110,6 @@ describe('Integration tests for P2P library', () => {
 					(n: Number[]) => n.includes(5003) || n.includes(5004),
 				);
 			});
-
-			it.skip('should not evict peers with low latency', async () => {});
```
diego-G: Why do we remove them? Are we implementing these cases in another place? Are they not relevant anymore?

.github/ISSUE_TEMPLATE/ISSUE_TEMPLATE_BUG_REPORT.md
mitsujutsu: this is possibly due to merge / linting, either way it should not be a problem.

elements/lisk-p2p/src/p2p.ts
mitsujutsu: I also responded to your linked thread, but in short `getRandomBytes` is not exported to top level. Therefore here we can use `crypto` instead.

shuse2: lisk-cryptography is not exposing those functions.
Please expose and update the lisk-cryptography version to `2.3.0-alpha.0`

elements/lisk-p2p/test/integration/p2p.ts
mitsujutsu: Firstly they do not belong in this describe block as the config is only for `connectTime`. Secondly yes these should be tested outside of integration as it's not possible to fake high ping node in integration.

elements/lisk-p2p/src/p2p.ts
mitsujutsu: `DEFAULT_SECRET` is used in some unit tests where a real 32 bit entropy secret was required. I will unexport `SECRET_LENGTH`

mitsujutsu:
> I've left few comments. Besides, I'd like to relapse if the features we are adding in this PR is actually covered with any of the existing integration tests.

`netgrouping` itself is already covered by unit tests. The new code in this PR is the use of `netgroup` in protecting some inbound peers from eviction. The functionality is identical with protecting peers for `connectTime`, which is already covered by integration test.

elements/lisk-p2p/package-lock.json
```diff
@@ -146,23 +146,23 @@
 		},
 		"@liskhq/bignum": {
 			"version": "1.3.1",
-			"resolved": "https://registry.npmjs.org/@liskhq/bignum/-/bignum-1.3.1.tgz",
+			"resolved": "https://npm.lisk.io/@liskhq%2fbignum/-/bignum-1.3.1.tgz",
```
shuse2: it shouldnt update to the `npm.lisk.io`
probably u need to do  `npm config set registry https://registry.npmjs.com/` and re run the package.lock

PR #4033, opened 2019-08-01
Standarize eslint - Closes #3749

### LINKED ISSUE #3749: Remove framework specific eslint rules
### Expected behavior
Framework should not have special rules for the eslint

### Actual behavior
It has custom eslint

2.2-

shuse2:
### What was the problem?
framework had several eslint rules that is not consistent with other packages

### How did I solve it?
Remove most of the custom rules.
Changed to `warn` level for the rules that has many usage

### How to manually test it?
Test should pass (only changed the rule)

### Review checklist

- [x] The PR resolves #3749 


framework/src/components/storage/index.js
```diff
@@ -14,6 +14,7 @@
 
 'use strict';
 
+const path = require('path');
```
pablitovicente: Why do you extract path if it's only used on the branch for new relic?

framework/test/jest/.eslintrc.json
```diff
@@ -5,6 +5,8 @@
 		"jest/globals": true
 	},
 	"rules": {
+		"callback-return": "error",
+		"no-unused-expressions": "off",
```
pablitovicente: Maybe double check if this should be on?

framework/.eslintrc.json
```diff
@@ -47,19 +47,11 @@
 			}
 		],
 		"callback-return": "error",
-		"generator-star-spacing": [
```
diego-G: Please, confirm this was on purpose since the commit just points to the other rule.

framework/.prettierrc.json
```diff
@@ -1,5 +1,5 @@
 {
 	"singleQuote": true,
-	"trailingComma": "es5",
+	"trailingComma": "all",
```
diego-G: Why don't we get the template as we do in the libraries? They look alike. https://github.com/LiskHQ/lisk-sdk/blob/development/templates/.prettierrc.json.tmpl

framework/test/jest/unit/specs/controller/application.spec.js
```diff
@@ -60,6 +60,7 @@ describe('Application', () => {
 		it('should validate genesisBlock', () => {
 			// Act
 
+			// eslint-disable-next-line no-new
```
diego-G: If it is completely necessary to use new in this test, I would suggest to add the rule to the Jest test suite as you've done for Mocha.

framework/src/components/storage/index.js
shuse2: no global import..
I think `path` won't hurt to be required in the root

framework/src/controller/controller.js
```diff
@@ -117,7 +117,7 @@ class Controller {
 		const pidPath = `${this.config.dirs.pids}/controller.pid`;
 		const pidExists = await fs.pathExists(pidPath);
 		if (pidExists) {
-			const pid = parseInt(await fs.readFile(pidPath));
+			const pid = parseInt(await fs.readFile(pidPath), 10);
```
diego-G: I would prefer to use `Number(inputValue`) because it is easier to read. https://github.com/airbnb/javascript#coercion--numbers

framework/.eslintrc.json
```diff
@@ -30,9 +30,6 @@
 				]
 			}
 		],
-		"callback-return": "error",
```
diego-G: You've removed more rules than the commit indicates. 

Besides, there are many Eslint disable statements. I would prefer to leave the rule until we're really able to remove them. Otherwise, we should create an issue to act on all this Eslint disables comments.

framework/.eslintrc.json
```diff
@@ -31,7 +31,6 @@
 			}
 		],
 		"no-restricted-properties": "off",
-		"no-template-curly-in-string": "off",
```
diego-G: Same as for the other rules.

framework/.eslintrc.json
```diff
@@ -2,13 +2,11 @@
 	"extends": ["../.eslintrc.json"],
 	"rules": {
 		"camelcase": "off",
-		"implicit-arrow-linebreak": "off",
 		"max-len": "off",
 		"no-param-reassign": "off",
 		"no-plusplus": "off",
 		"no-underscore-dangle": "off",
 		"strict": "off",
-		"no-restricted-properties": "off",
```
diego-G: Small typo in the commit headline.

framework/.eslintrc.json
```diff
@@ -3,11 +3,11 @@
 	"rules": {
 		"camelcase": "off",
 		"max-len": "off",
-		"no-param-reassign": "off",
```
diego-G: Could you explain why do you change this? Also you don't need to wrap it into an array. https://eslint.org/docs/user-guide/configuring#configuring-rules

framework/.eslintrc.json
```diff
@@ -4,7 +4,6 @@
 		"camelcase": "off",
 		"max-len": "off",
 		"no-param-reassign": ["warn"],
-		"no-plusplus": "off",
```
diego-G: Same as for the other rules.

framework/.eslintrc.json
```diff
@@ -1,12 +1,12 @@
 {
 	"extends": ["../.eslintrc.json"],
 	"rules": {
-		"camelcase": "off",
+		"strict": "off",
```
diego-G: Commit's headline is misleading again. Why do you suddenly modify rule levels here?

framework/test/mocha/common/utils/wait_for.js
```diff
@@ -51,6 +51,7 @@ function blockchainReady(retries, timeout, baseUrl, doNotLogRetries, cb) {
 
 	baseUrl = baseUrl || __testContext.baseUrl;
 
+	// eslint-disable-next-line wrap-iife
```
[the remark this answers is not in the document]
shuse2: I didn't want to rewrite the whole thing 😉 
And this will be removed while we move jest, so I kept it as it is for now

framework/test/jest/unit/specs/controller/application.spec.js
shuse2: In normal cases, we shouldn't use the constructor to have side effect, that's why it should be removed in general (but this requires too much refactoring)

framework/src/controller/controller.js
shuse2: Im just standardising the eslint rule across the packages.
Also, https://eslint.org/docs/rules/radix has valid point

framework/.eslintrc.json
shuse2: If we add `off`, there will be more code that uses the invalid rules.
Therefore, I would prefer to turn it off, and fix the rule one by one when we do refactoring

framework/.eslintrc.json
shuse2: some rule levels are `warn` because it's not possible to change in one PR, but we don't want to keep the rule. Therefore, put it to warn and we can refactor when we change the code

framework/.eslintrc.json
shuse2: aligning with other packages, but also it has valid reasoning
https://eslint.org/docs/rules/no-param-reassign#disallow-reassignment-of-function-parameters-no-param-reassign

framework/.eslintrc.json
shuse2: yes it is removed since it's not used

framework/test/jest/unit/specs/controller/application.spec.js
diego-G: Jest tests are completely new. They should serve as example... Do we plan to refactor them as well?

framework/src/controller/controller.js
diego-G: I've checked the `radix` example and it is only useful for code written in pre-ECMAScript 5 which is not our case. We don't need the extra `, 10)`. In any case, I suggest to move to `Number` as Airbnb states is good too and it is less verbose.

framework/test/mocha/common/utils/wait_for.js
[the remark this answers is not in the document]
diego-G: Your call. But in this case, the issues's estimation is too high. Disabling rules is very bad practise instead of keeping them together in the `.eslint.json` file. It will be more complicated to track.

framework/src/controller/controller.js
shuse2: Im fine changing the rule, but let's change it across the packages in another issue

framework/test/mocha/common/utils/wait_for.js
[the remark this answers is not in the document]
shuse2: at the same time, we won't add a new code like this, and I think that should be better =)
Definitely estimation was too heigh 😉

framework/test/jest/unit/specs/controller/application.spec.js
shuse2: Some of the cases are exceptions, in this case it's using as a check, but 90% of the case shouldn't apply to this.
For exceptional cases, we should disable by line like this, and it's consistent across packages too

PR #4046, opened 2019-08-06
Add util functions to be used in integration test - Closes #3955

### LINKED ISSUE #3955: Add util functions to be used in integration test
### Description
While we migrate to the Jest integration tests, we need to have various helper function to setup the environment.

### List of functionality
To be added

### Acceptance criteria
- List all the functionality used in current mocha integration tests in this issue
- Decide the necessary functionality to be implemented
- Reimplement the functionality in `jest/integration/utils`
- Add one sample tests

SargeKhan:
### What was the problem?
The util function for integration tests in jests was missing
### How did I solve it?
Added util functions and a few tests.
### How to manually test it?
Run the tests
### Review checklist

- [x] The PR resolves #3955 


shuse2:
Also, can you delete `.github/ISSUE_TEMPLATE/ISSUE_TEMPLATE_FEATURE.md` and `.github/ISSUE_TEMPLATE/ISSUE_TEMPLATE_BUG_REPORT.md` (although it's not related to this PR)
It seems I created redundant file by mistake

framework/test/jest/integration/specs/modules/chain/genesis_block.spec.js
```diff
@@ -0,0 +1,199 @@
...
+const genesisBlock = require('../../../../../fixtures/config/devnet/genesis_block');
+
+describe('genesis block', () => {
+	const dbName = 'genesis_block';
+	const TRANSACTION_TYPES_DELEGATE_REGISTRATION = 2;
+	let storage;
+	let chainModule;
+
+	beforeAll(async () => {
+		storage = new StorageSandbox(storageConfig(dbName));
```
yatki: can we call storageConfig inside StorageSandbox and just provide the config object to storageSandbox ?

.github/ISSUE_TEMPLATE/ISSUE_TEMPLATE_FEATURE.md
```diff
@@ -1,10 +1,9 @@
 ---
-name: 
+name:
```
yatki: why github templates are in this PR ? :D

framework/test/jest/integration/utils/configs/config_storage.js
```diff
@@ -0,0 +1,33 @@
...
+	port: 5432,
+	database,
+	user: user || 'lisk',
+	password: password || 'password',
+	min: 10,
+	max: process.env.LISK_TEST_DB_MAX_CONNECTIONS || 2,
+	poolIdleTimeout: 30000,
+	reapIntervalMillis: 1000,
+	logEvents: ['error'],
+	logFileName: 'logs/lisk_db.log',
```
yatki: I would set the default values for properties then overwrite the properties with rest parameter ? Also storageConfig is used inside SandboxStorage only right. Can we move this object inside that and get the config from SandboxStorage constructor.

framework/test/jest/integration/utils/configs/config_cache.js
```diff
@@ -0,0 +1,27 @@
...
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+'use strict';
+
+const cacheConfig = () => ({
```
yatki: Why don't we set the object to variable directly?

SargeKhan: This ensures that the cacheConfig is not overwritten by any of the tests.

.github/ISSUE_TEMPLATE/ISSUE_TEMPLATE_FEATURE.md
SargeKhan: These are added due to the merging the development branch into this branch. You can refer to [this commit](https://github.com/LiskHQ/lisk-sdk/pull/4046/commits/b2d5f5a0e925ddf52d707a6ac53eeead4d1ae0e0) to see. Do you know how to fix this? Linking: @shuse2

framework/test/jest/integration/utils/chain/chain.js
```diff
@@ -0,0 +1,67 @@
...
+		constants: genesisConfig(),
+		genesisBlock,
+		registeredTransactions: defaultTransactions(),
+	};
+	const chainModule = new ChainModule(options);
+
+	return chainModule;
+};
+
+const createMockChannel = databaseName => {
```
nazarhussain: This mock does not belongs to chain module. Such mocks should be defined on application level and shared among tests for all modules.

framework/test/jest/integration/utils/chain/chain.js
```diff
@@ -0,0 +1,67 @@
...
+			return {};
+		}),
+		subscribe: jest.fn((event, listener) => {
+			listener({ data: {} });
+		}),
+	};
+	return channel;
+};
+
+const createDefaultLoadedChainModule = async databaseName => {
```
nazarhussain: Can't understand `default loaded chain module`. What's default? Is there custom chain modules as well? 

```suggestion
const createLoadedChainModule = async databaseName => {
```

framework/test/jest/integration/utils/chain/transaction.js
```diff
@@ -0,0 +1,107 @@
...
+				keysgroup: signedPublicKeys,
+			},
+		},
+	});
+	transaction.sign(passphrase);
+	return transaction.toJSON();
+};
+
+const getTransaction = async (storage, id) =>
+	storage.entities.Transaction.getOne({ id }, { extended: true });
```
nazarhussain: Similar to previous comment, I don't why we need such 1 line wrappers.  If we have to pass `storage` and `id` then why can't we have that 1 line in the test itself.

framework/test/jest/integration/utils/chain/chain.js
shuse2: by `default`, it meant default config 😉

framework/test/jest/integration/utils/configs/config_genesis.js
```diff
@@ -0,0 +1,67 @@
...
+		DELEGATE: 2,
+		VOTE: 3,
+		MULTI: 4,
+		DAPP: 5,
+		IN_TRANSFER: 6,
+		OUT_TRANSFER: 7,
+	},
+	UNCONFIRMED_TRANSACTION_TIMEOUT: 10800, // 1080 blocks
+	EXPIRY_INTERVAL: 30000,
+});
```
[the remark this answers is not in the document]
shuse2: I think internally those are still genesis_config although it's not exposed

framework/test/jest/integration/specs/modules/chain/genesis_block.spec.js
```diff
@@ -0,0 +1,201 @@
...
+
+describe('genesis block', () => {
+	const dbName = 'genesis_block';
+	const TRANSACTION_TYPES_DELEGATE_REGISTRATION = 2;
+	let storage;
+	let chainModule;
+
+	beforeAll(async () => {
+		storage = new StorageSandbox(storageConfig({ database: dbName }));
+		await storage.bootstrap();
```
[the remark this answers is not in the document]
shuse2: this storage is only used for probing, and the usage is different.
Chain module itself creates own storage.

framework/test/jest/integration/utils/chain/transaction.js
shuse2: The purpose is to wrap the storage usage as much as possible, so that even if it's used in everywhere, we can change the test easily

framework/test/jest/integration/specs/modules/chain/genesis_block.spec.js
```diff
@@ -0,0 +1,201 @@
...
+				await chainModule.unload();
+			});
+
+			it('should save genesis block to the database', async () => {
+				const block = await getBlock(storage, genesisBlock.id);
+				expect(block.id).toEqual(genesisBlock.id);
+				expect(block.height).toEqual(1);
+			});
+
+			it('should have genesis transactions in database', async () => {
```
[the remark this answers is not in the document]
shuse2: Yes possibly, and I tend to agree to your point. However, this test was chose to migrate just to see the util functions are working properly.
We can improve the test on another cycle

framework/test/jest/integration/utils/index.js
```diff
@@ -0,0 +1,27 @@
...
+ */
+
+'use strict';
+
+const { defaultTransactions } = require('./default_transactions');
+const storageUtils = require('./storage');
+const configUtils = require('./configs');
+const chainUtils = require('./chain');
+
+module.exports = {
```
pablitovicente: I would export as sub objects like ```{
storageUtils,
configUtils,
chainUtils,
}``` maybe makes reading some tests easier?

framework/test/jest/integration/specs/modules/chain/genesis_block.spec.js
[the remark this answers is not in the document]
nazarhussain: This `storage` is used and passed to many functions. So you have different storage object in the chain module and using different in tests. Would be nice if you simply re-use the storage object initialized within the chain module.

framework/test/jest/integration/utils/configs/config_genesis.js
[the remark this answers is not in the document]
nazarhussain: Internally these are all constants not genesisConfig.

framework/test/jest/integration/utils/chain/chain.js
nazarhussain: In that case would be nice to to rename it to `createAndLoadChainModule`

framework/test/jest/integration/utils/chain/transaction.js
nazarhussain: In that chase, I would suggest to group them for `storageUtils` not the `chainUtils`.

shuse2: I think getting the transaction is using storage, but not really responsibility of `storage`.
How storage is used is defined in each utils

framework/test/jest/integration/specs/modules/chain/genesis_block.spec.js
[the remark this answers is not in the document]
shuse2: I prefer not to use the `storage` within the chain module for not intended effect on the test itself

framework/test/jest/integration/utils/chain/transaction.js
nazarhussain: If you wrap this code `storage.entities.Transaction.get()`, its a storage util function, nothing else.

framework/test/jest/integration/specs/modules/chain/genesis_block.spec.js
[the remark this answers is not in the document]
nazarhussain: Didn't understood 'not to use storage within the chain module'. Its already used, I am just pointing for testing purpose we should use the same storage object which chain module have, rather creating different storage object.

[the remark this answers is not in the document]
shuse2: I meant I prefer not to use the same storage instance for setting up the test and for using internally in the chain module for not intended effect on the test itself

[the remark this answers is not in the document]
nazarhussain: Any particular reason for this preference and have duplicate connections? I don't see there is any change of behaviour in tests if we use same storage instance. With this approach every integration will start additional connections, and then we run parallel tests and then more connections increased, without any reason.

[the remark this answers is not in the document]
shuse2: That is a good point too, for increasing the connection.
However, I dont think it's good practice to use an instance of testing subject.
For setup and probing, I feel safer to use other method than what we are testing.

framework/test/jest/integration/utils/chain/transaction.js
shuse2: I think it make more sense to bundle by domain, it will be easier to find the code too.
It bit strange other utils for account are in one place, and other function related to account is in other place

nazarhussain: Yes I agree and domain is already well defined. So why we have all database entities together in once place. We didn't put `Block` entity to block module and `Transaction` entity to transaction module. 

Please follow same pattern here, and group all `storage` related functions to `storageUtils`

framework/test/jest/integration/utils/configs/config_genesis.js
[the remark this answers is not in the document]
nazarhussain: This also needs to refactored. As said internally these are all constants. 

https://github.com/LiskHQ/lisk-sdk/blob/c86a887442d5f49fb696715f4dfb27c478a063a8/framework/src/modules/chain/chain.js#L98

https://github.com/LiskHQ/lisk-sdk/blob/8c569ec7a450e5ae2a17cd306251026e1c8a0bc4/framework/src/controller/application.js#L153

So I suggest to explicitly use scope of the module itself in which these are all constants. Or split them exactly in the way we have on application level to genesis config and immutable constants.

PR #4054, opened 2019-08-07
npx mocha doesn't use mocha.opts file - Closes #3413

### LINKED ISSUE #3413: npx mocha doesn't use mocha.opts file
### Expected behavior
It should
### Actual behavior
When running `npx mocha pathToTests` it seems Mocha is not using the mocha.opts file which requires to run `setup.js` in order to load for example global constants used by the tests.
### Steps to reproduce
Try to run `npx mocha framework/test/mocha/integration/blocks/process/process.js`
`development`

pablitovicente:
### What was the problem?

Running mocha through npx didn't work

### How did I solve it?

- By linking the `mocha.opts` file to the `/test/mocha.opts` folder where mocha is searching for.
- by updating lisk-mocha-runner 

Another options I tested but do not work with `npx`

- to remove mocha.opts (deprecated but still supported by Mocha) and replace with `.mocharc`

### How to manually test it?

in `framework/` run: 
- mocha test/mocha/unit/controller/helpers/validator.js
- npm run mocha:unit test/mocha/unit/controller/helpers/validator.js
- npx mocha test/mocha/unit/controller/helpers/validator.js
and all should be able to run that test or any other test

### Review checklist

- [ ] The PR resolves #3413


framework/test/mocha/common/lisk-mocha-runner/index.js
```diff
@@ -142,8 +142,6 @@ if (testPathPattern && testPathPattern.indexOf('-') === 0) {
 	testPathPattern = null;
 }
 
-mochaCliOptions.unshift('--', '--opts', 'test/mocha/mocha.opts');
```
yatki: ❤️

pablitovicente:
> Instead



> Instead of moving the source file, I'd prefer to create a symlink. This way we could know that file is not actually belong there. Because it sort of doesn't fit to current folder structure with having mocha/jest folders.. and having a mocha related file in the top folder?

This was my first approach but was not convinced to have the link instead of the file but this is a good reasoning to have the symlink so I will change it to a symlink.

framework/test/mocha/common/lisk-mocha-runner/index.js
lsilvs: I suspect this is causing https://github.com/LiskHQ/lisk-sdk/issues/4124

PR #4055, opened 2019-08-07
Remove block dependency from block/utils - Closes #4012

### LINKED ISSUE #4012: Remove block dependency from blocks/utils
### Expected behavior
Blocks/Utils must be pure utility functions without have any dependency on any other module. 

### Actual behavior

Currently the `block/utils.js` is requiring `block.js` which is requiring the `block_v1.js` and `block_v2.js`. It violates the concepts of utility functions. Also in case we need to use some utility function inside `block_v1.js` or `block_v2.js`, it creates a circular dependency and failed. 

https://github.com/LiskHQ/lisk-sdk/blob/8920e9dd4bd1be8aa578d294395fe46d50b0f970/framework/src/modules/chain/blocks/utils.js#L20

2.3.0

michielmulders:
### What was the problem?
Passing block dependency to several files in the `block` folder creating circular dependency.
To solve this now, we pass the `blocksLogic` dependency.

However, as @nazarhussain indicated, we should create pure functions as util functions. Currently, they do processing and `I/O` with DB. 

### How did I solve it?
Pass `blocksLogic` dependency directly as argument.

### Review checklist

- [x] The PR resolves #4012

### DIFF
--- a/framework/src/modules/chain/blocks/blocks.js
+++ b/framework/src/modules/chain/blocks/blocks.js
@@ -17,7 +17,7 @@
 const EventEmitter = require('events');
 const { cloneDeep } = require('lodash');
 const blocksUtils = require('./utils');
-const Block = require('./block');
+const blocksLogic = require('./block');
 const { BlocksProcess } = require('./process');
 const { BlocksVerify } = require('./verify');
 const blockVersion = require('./block_version');
@@ -290,7 +290,7 @@ class Blocks extends EventEmitter {
 			return;
 		}
 		try {
-			this._lastBlock = await Block.loadLastBlock(
+			this._lastBlock = await blocksLogic.loadLastBlock(
 				this.storage,
 				this.interfaceAdapters,
 				this.genesisBlock,
@@ -577,7 +577,7 @@ class Blocks extends EventEmitter {
 		this._isActive = true;
 
 		try {
-			const normalizedBlocks = Block.readDbRows(
+			const normalizedBlocks = blocksLogic.readDbRows(
 				blocks,
 				this.interfaceAdapters,
 				this.genesisBlock,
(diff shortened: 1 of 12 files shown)


michielmulders:
To avoid circular dependency @shuse2

shuse2:
I would prefer to clean up the code location than just injecting for that purpose

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -293,6 +294,7 @@ class Blocks extends EventEmitter {
 				this.storage,
 				this.interfaceAdapters,
 				this.genesisBlock,
+				blocksLogic,
```
nazarhussain: Agree that its the easiest way to solve. But I don't like that direction. We have to resolve and not pass any other module to utils. To me when util functions require instance of other modules then these are actually not standalone utils functions. We should rather move `loadLastBlock` to `blockLogic` where it relates more accurately.

michielmulders: The loadLastBlock function is not the issue, it is that `dbRead` and `storageRead` require `blocksLogic`.

michielmulders:
@nazarhussain I moved the functions that were not pure util functions to `block.js`.
Also, implemented back the `sortTransactions` function which was initially causing the circular dep.

Currently, the dep tree looks like:
`block` -> `block_v1/2` -> `utils`

`blocks`
    -> `utils`
    -> `block`

`utils` -> No dependencies!

**Full dep list: `blocks` -> `block` -> `block_v1/2` -> `utils` END**

Before:
`utils` required `block` which caused the circular dependency.

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -17,6 +17,7 @@
 const EventEmitter = require('events');
 const { cloneDeep } = require('lodash');
 const blocksUtils = require('./utils');
+const Block = require('./block');
```
shuse2:
```suggestion
const blocksLogic = require('./block');
```

PR #4061, opened 2019-08-09
Emit event when a block is finalized from the bft module - Closes #3908

### LINKED ISSUE #3908: Emit event when a block is finalized from the bft module
### Expected behavior
Once a block is finalized, the bft module should fire an event to ensure that other modules perform their own logic based on this blockchain state change.
### Actual behavior
Event is currently not fired from the bft module.

feature/introduce_bft_consensus

pablitovicente:
### What was the problem?

An event was not being emitted when finalized height changed in finality manager

### How did I solve it?

- by extending event emitter in finality manager
- by emitting an event on change of finalized height
- by listening to the event in the BFT class

### Review checklist

- [x] The PR resolves #3908


lsilvs:
Would be possible to use `channel.publish` and `channel.subscribe` instead?

pablitovicente:
> Would be possible to use `channel.publish` and `channel.subscribe` instead?

There was a talk about this and it seems for now we'll use `emit()` I do agree that maybe before completing `bft` we can agree on changing to `channel.subscribe/publish` on all the code? @nazarhussain @shuse2 ?

shuse2:
Since BFT will be separate library in the future, so I think it would be better to keep it as emit

framework/src/modules/chain/bft/finality_manager.js
```diff
@@ -95,13 +98,18 @@ class FinalityManager {
...
 		this.updatePreVotesPreCommits(blockHeader);
 
+		// Store current finalizedHeight
+		const currentFinalizedHeight = this.finalizedHeight;
 		// Update the pre-voted confirmed and finalized height
 		this.updatePreVotedAndFinalizedHeight();
 
+		if (currentFinalizedHeight !== this.finalizedHeight) {
+			this.emit(EVENT_NEW_FINALIZED_HEIGHT, blockHeader);
+		}
```
nazarhussain: @pablitovicente Please move this same logic inside function `updatePreVotedAndFinalizedHeight` This way no matter when and from where the finalized height is changed. The event will be emitted. In one of my PR I am refactoring a bit of code to only call `updatePreVotedAndFinalizedHeight` once during the `recompute`. So that will also fix emitting this event multiple times during recompute process.

framework/src/modules/chain/bft/finality_manager.js
```diff
@@ -25,6 +26,7 @@ const {
 	BFTInvalidAttributeError,
 } = require('./errors');
 
+const EVENT_NEW_FINALIZED_HEIGHT = 'EVENT_NEW_FINALIZED_HEIGHT';
```
nazarhussain: I would rather suggest it to change `EVENT_BFT_FINALIZED_HEIGHT_CHANGED` This way its very clear where ever used from which module and entity that even belongs.

framework/src/modules/chain/bft/finality_manager.js
pablitovicente: ok

framework/src/modules/chain/bft/finality_manager.js
pablitovicente: ok

framework/src/modules/chain/bft/finality_manager.js
pablitovicente: Done and also renamed variables for the code to make a bit more sense.

framework/src/modules/chain/bft/bft.js
```diff
@@ -15,7 +15,10 @@
 'use strict';
 
 const EventEmitter = require('events');
-const { FinalityManager } = require('./finality_manager');
+const {
+	EVENT_BFT_FINALIZED_HEIGHT_CHANGED,
```
nazarhussain: One thing left, should not we export this event down from BFT module. So chain can utilize it.

pablitovicente: Good catch!

PR #4065, opened 2019-08-09
P2P 2.3 incompatibility with older version - Closes #4060

### LINKED ISSUE #4060: P2P 2.3 incompatibility with 2.0 and 1.x related to peer lists
### Expected behavior

- The 2.3 node should work with old 2.0 and 1.x peers.

### Actual behavior

- v2.3 nodes need to expose both the old `list` endpoint and the new `getPeers` (minimal) endpoints.
- v2.3 nodes need to call the old `list` endpoint when trying to fetch peers from an old v2.0 or v1.x peer.
- There is an incompatibility issue related to `ipAddress` vs `ip` properties in old versions.

### Steps to reproduce

Try to connect a v2.3 node with old v2.0 and v1.6 nodes.

v2.3

ishantiw:
### What was the problem?
The `2.3` node should work with old `2.0` and `1.x` peers. There is an incompatibility issue related to `ipAddress` vs `ip` properties in old versions. Also, the old version only supports `list` RPC to get peers list.

### How did I solve it?

- Created incoming and outgoing `peerInfo` sanitization function.
- To fetch peers call RPC based on the version if its lower than `2.3`

### How to manually test it?

`npm t` and also connect with older versions and see if it's fetching peers

### Review checklist

- [x] The PR resolves #4060 

### DIFF
--- a/elements/lisk-p2p/src/peer/base.ts
+++ b/elements/lisk-p2p/src/peer/base.ts
@@ -35,9 +35,8 @@ import { constructPeerIdFromPeerInfo, getNetgroup } from '../utils';
 import * as socketClusterClient from 'socketcluster-client';
 import { SCServerSocket } from 'socketcluster-server';
 import {
-	isLowerVersionPeer,
-	validateBasicPeersInfoList,
 	validatePeerInfo,
+	validatePeersInfoList,
 	validateProtocolMessage,
 	validateRPCRequest,
 } from '../validation';
@@ -91,7 +90,6 @@ export const REMOTE_EVENT_MESSAGE = 'remote-message';
 
 export const REMOTE_RPC_UPDATE_PEER_INFO = 'updateMyself';
 export const REMOTE_RPC_GET_NODE_INFO = 'status';
-export const REMOTE_RPC_GET_MINIMAL_PEERS_LIST = 'getPeers';
 export const REMOTE_RPC_GET_PEERS_LIST = 'list';
 
 export const DEFAULT_CONNECT_TIMEOUT = 2000;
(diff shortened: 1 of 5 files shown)


elements/lisk-p2p/src/validation.ts
```diff
@@ -39,6 +43,7 @@ import { constructPeerIdFromPeerInfo } from './utils';
 
 const IPV4_NUMBER = 4;
 const IPV6_NUMBER = 6;
+export const IPADDRESS_FIELD_SUPPORT_VERSION = '2.3.0';
```
jondubois: `IP_ADDRESS_FIELD_SUPPORT_VERSION` is better.

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -510,7 +512,9 @@ export class Peer extends EventEmitter {
 	public async fetchPeers(): Promise<ReadonlyArray<P2PPeerInfo>> {
 		try {
 			const response: P2PResponsePacket = await this.request({
-				procedure: REMOTE_RPC_GET_PEERS_LIST,
+				procedure: isLowerVersionPeer(this._peerInfo as P2PDiscoveredPeerInfo)
+					? REMOTE_RPC_GET_PEERS_LIST
+					: REMOTE_RPC_GET_MINIMAL_PEERS_LIST,
 			});
 
 			return validateBasicPeersInfoList(response.data);
```
jondubois: Maybe we should always remove other properties (except for ip/ipAddress and wsPort), even if the peer is the old version and it gives us the full list. The idea for the original feature was that we can't trust PeerInfo which we haven't connected to so that is why we don't store the other fields. Maybe we should keep this aspect.

elements/lisk-p2p/src/validation.ts
shuse2: Is this SDK version or Core version?
I think this version doesnt relates to this library, so Prefer not to use hardcoded version , too.

PR #4075, opened 2019-08-13
Improve logger naming for components - Closes #3532

### LINKED ISSUE #3532: Improve the logger name for JSON logger
### Expected behavior

Every logger entry should be easily identifiable to which module in the framework it belongs to. 

### Actual behavior

Right now every log entry have a name `lisk-framework`, which is less intuitive while debugging in the logs. 

```
{
  "name": "lisk-framework",
  "hostname": "Nazars-MacBook-Pro.local",
  "pid": 44216,
  "level": 30,
  "msg": "Releasing enqueued broadcasts",
  "time": "2019-05-07T08:47:57.530Z",
  "src": {
    "file": "/Users/nazar/Hub/Lightcurve/Lisk/dev/lisk-sdk/framework/src/modules/chain/logic/broadcaster.js",
    "line": 245,
    "func": "releaseQueue"
  },
  "v": 0
}
```

```
{
  "name": "lisk-framework",
  "hostname": "Nazars-MacBook-Pro.local",
  "pid": 44553,
  "level": 30,
  "msg": "Loading module lisk-framework-chain:0.1.0 with alias \"chain\"",
  "time": "2019-05-07T08:55:54.538Z",
  "src": {
    "file": "/Users/nazar/Hub/Lightcurve/Lisk/dev/lisk-sdk/framework/src/controller/controller.js",
    "line": 219,
    "func": "_loadInMemoryModule"
  },
  "v": 0
}
```

As you look into both entries, both shows the name 'lisk-framework`, it would be nice if we can have different names of logger for different chunks of the application. 

2.0.0

michielmulders:
### What was the problem?
All logs were logged under `lisk-framework`.

### How did I solve it?
Modify the wrapper and creator function for the logger to accept a new parameter name.
This name identifies the component you are in.

### How to manually test it?
Start application and monitor the logs 👍 

### Review checklist

- [x] The PR resolves #3532 
- [x] Documentation has been added/updated - **[Updated internal logger docs](https://www.notion.so/lightcurve/Logger-c88c595e2931409fa72ecfffe6b20d4b)**


framework/src/components/logger/logger.js
```diff
@@ -58,7 +65,7 @@ const createLogger = ({ fileLogLevel, consoleLogLevel, logFileName }) => {
 			: [];
 	const streams = [...consoleStream, ...fileStream];
 	return bunyan.createLogger({
-		name: 'lisk-framework',
+		name: loggerName,
```
shuse2: everything except this new `loggerName` has default value, so probably it should provide default, too.
Also, createLogger with `loggerName` seems redundant, although it kinda applies to other property too 😢

framework/src/components/logger/index.js
```diff
@@ -18,13 +18,13 @@ const { createLogger } = require('./logger');
 const { config: defaultConfig } = require('./defaults');
 const validator = require('../../controller/validator');
 
-function createLoggerComponent(options = {}) {
+function createLoggerComponent(name, options = {}) {
```
nazarhussain: Idea of such factory methods to accept single object was that, we can define its configuration as JSON files if required at any time. So why not we just consider `name` as part of options.

framework/src/components/logger/logger.js
```diff
@@ -58,7 +65,7 @@ const createLogger = ({ fileLogLevel, consoleLogLevel, logFileName }) => {
 			: [];
 	const streams = [...consoleStream, ...fileStream];
 	return bunyan.createLogger({
-		name: 'lisk-framework',
+		name,
```
nazarhussain: This seems to follow right pattern as of Bunyan documentation. But I suggest another path. I suggest that logger name should be fixed `Lisk-framework` for all logger instances. The extra attribute we are trying to set should be set as `module: 'lisk-chain'` or  `module: 'lisk-network'` etc. Which is called bounded fields in terms of Bunyan. 


Having fixed static name for all log entries is very essential. Think of a fact that one have a log analysis service e.g. Logstash where he redirects all logs from many other products. And now if he want to find all log entries from Lisk, he can find it only by having static name `lisk-framework` in all entries.

nazarhussain: I suggest to have a bound field `module` or `tag` to identify user specific identifier.

framework/src/components/logger/index.js
michielmulders: I don't want name to go through the validation, that's the reason I kept it separate

framework/src/components/logger/logger.js
michielmulders: Add an extra property `module` (in class module) with shorthand module alias to used as `module` name.

framework/src/components/logger/index.js
michielmulders: Discussion with Nazar: Add name to options and so, to schema object => validate for alphabetical and some chars like `-:`

nazarhussain: Additionally since we fixed the name to `lisk-framework`, so this option should not be name rather `moduleInfo`

framework/src/modules/http_api/index.js
```diff
@@ -41,6 +41,7 @@ class HttpAPIModule extends BaseModule {
 			author: 'LiskHQ',
 			version: '0.1.0',
 			name: 'lisk-framework-http-api',
+			module: 'http_api',
```
nazarhussain: Its not required to add redundant information. You can access same with module `alias`.

```suggestion
```

framework/src/modules/network/index.js
```diff
@@ -37,6 +37,7 @@ module.exports = class NetworkModule extends BaseModule {
 			author: 'LiskHQ',
 			version: '0.1.0',
 			name: 'lisk-framework-network',
+			module: 'network',
```
nazarhussain: Same goes here...

PR #4082, opened 2019-08-14
Fix invalid schema - Closes #4077

### LINKED ISSUE #4077: Invalid schema usage for validating transaction
### Expected behavior
When fetching the transaction from peer, it should use correct schema

### Actual behavior
It is using invalid schema.

### Steps to reproduce
Try to get any transaction from peer

2.2-

shuse2:
### What was the problem?
Transaction response was validated against wrong schema

### How did I solve it?
Change the schema to validate

### How to manually test it?
Start a node and connect to a network which has transaction inside

### Review checklist

- [x] The PR resolves #4077 

### DIFF
--- a/framework/test/jest/unit/specs/modules/chain/loader.spec.js
+++ b/framework/test/jest/unit/specs/modules/chain/loader.spec.js
@@ -118,6 +118,9 @@ describe('Loader', () => {
 					error = err;
 				}
 				expect(error.length).toBe(1);
+				expect(error[0].message).toBe(
+					"should have required property 'transactions'",
+				);
 			});
 		});
 	});
--- a/framework/test/jest/unit/specs/modules/chain/loader.spec.js
+++ b/framework/test/jest/unit/specs/modules/chain/loader.spec.js
@@ -1,0 +1,124 @@
+/*
+ * Copyright © 2019 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
+ * for licensing information.
+ *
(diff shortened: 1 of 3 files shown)


framework/test/jest/unit/specs/modules/chain/loader.spec.js
```diff
@@ -0,0 +1,124 @@
...
+			});
+
+			it('should not throw an error', async () => {
+				let error;
+				try {
+					await loader._getTransactionsFromNetwork();
+				} catch (err) {
+					error = err;
+				}
+				expect(error).toBe(undefined);
```
ManuGowda: why not to use chai's `to.not.throw` ? something like this `expect(function () {}).to.not.throw();`

framework/test/jest/unit/specs/modules/chain/loader.spec.js
```diff
@@ -0,0 +1,124 @@
...
+			});
+
+			it('should throw an error', async () => {
+				let error;
+				try {
+					await loader._getTransactionsFromNetwork();
+				} catch (err) {
+					error = err;
+				}
+				expect(error.length).toBe(1);
```
ManuGowda: expecting an error would be more concise than just checking the length.

framework/test/jest/unit/specs/modules/chain/loader.spec.js
shuse2: with `await`, it doesn't work well

PR #4085, opened 2019-08-14
Implement isEmpty and truncate for temp_block - Closes #3776

### LINKED ISSUE #3776: Implement functionality for removing blocks stored in temp table
### Expected behavior
Should be able to remove blocks stored in temp table and implement a check to see if the table is present/holds records.

`feature/improve-bft`

michielmulders:
### What was the problem?
Need functionality for truncating the `temp_block` table and checking if the table is`isEmpty`. 

### How did I solve it?
Added functionality in `temp_block` entity in `chain` and added relevant integration tests.
[Related PR](https://github.com/LiskHQ/lisk-sdk/pull/4019/files).

### How to manually test it?
Run `npm run jest:integration -- test/jest/integration/specs/modules/chain/components/storage/entities/block_temp.spec.js`

### Review checklist

- [x] The PR resolves #3776 
- [x] All new code is covered with ~unit tests~ integration tests

### DIFF
--- a/framework/src/modules/chain/components/storage/entities/temp_block.js
+++ b/framework/src/modules/chain/components/storage/entities/temp_block.js
@@ -220,7 +220,7 @@ class TempBlock extends BaseEntity {
 	isEmpty(tx = null) {
 		return this.adapter
 			.executeFile(this.SQLs.isEmpty, {}, {}, tx)
-			.then(result => (result[0] ? result[0].bool : true));
+			.then(([result]) => (result ? result.bool : true));
 	}
 }
 
--- a/framework/src/modules/chain/components/storage/entities/temp_block.js
+++ b/framework/src/modules/chain/components/storage/entities/temp_block.js
@@ -219,8 +219,8 @@ class TempBlock extends BaseEntity {
 	 */
 	isEmpty(tx = null) {
 		return this.adapter
-			.executeFile(this.SQLs.isEmpty, {}, { expectedResultCount: 1 }, tx)
-			.then(result => result.bool);
+			.executeFile(this.SQLs.isEmpty, {}, {}, tx)
+			.then(result => (result[0] ? result[0].bool : true));
 	}
 }
 
--- a/framework/src/modules/chain/components/storage/entities/temp_block.js
+++ b/framework/src/modules/chain/components/storage/entities/temp_block.js
@@ -28,6 +28,8 @@ const sqlFiles = {
 	create: 'temp_block/create.sql',
 	delete: 'temp_block/delete.sql',
 	get: 'temp_block/get.sql',
+	truncate: 'temp_block/truncate.sql',
+	isEmpty: 'temp_block/isEmpty.sql',
 };
 
 /**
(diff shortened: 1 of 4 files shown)


framework/test/jest/integration/specs/modules/chain/components/storage/entities/block_temp.spec.js
```diff
@@ -234,4 +234,20 @@ describe('TempBlock', () => {
 			await expect(TempBlockEntity.delete({ height })).resolves.toBeNull();
 		});
 	});
+
+	describe('truncate', () => {
+		it('should truncate all rows from table', async () => {
```
shuse2: don't we need a setup for this?

framework/test/jest/integration/specs/modules/chain/components/storage/entities/block_temp.spec.js
```diff
@@ -234,4 +234,20 @@ describe('TempBlock', () => {
...
+			await TempBlockEntity.truncate();
+
+			expect(await TempBlockEntity.get()).toEqual([]);
+		});
+	});
+
+	describe('count', () => {
```
shuse2: description is wrong

framework/test/jest/integration/specs/modules/chain/components/storage/entities/block_temp.spec.js
```diff
@@ -234,4 +234,20 @@ describe('TempBlock', () => {
...
+		it('should count all rows in table', async () => {
```
shuse2: same as above

framework/test/jest/integration/specs/modules/chain/components/storage/entities/block_temp.spec.js
```diff
@@ -234,4 +234,20 @@ describe('TempBlock', () => {
...
+			const isEmpty = await TempBlockEntity.isEmpty();
+
+			expect(isEmpty).toEqual(false); // Row 1 and Row 2
```
shuse2: it should check both existing case and non-existing case

framework/test/jest/integration/specs/modules/chain/components/storage/entities/block_temp.spec.js
michielmulders: Before every test, a table gets created with two transactions in. Should be good! :)

framework/src/modules/chain/components/storage/entities/temp_block.js
```diff
@@ -193,6 +195,33 @@ class TempBlock extends BaseEntity {
...
+	/**
+	 * Check if table is empty (true)
+	 *
+	 * @param {Object} [tx]
+	 * @returns {Promise.<boolean, Error>}
+	 */
+	isEmpty(tx = null) {
+		return this.adapter
+			.executeFile(this.SQLs.isEmpty, {}, {}, tx)
+			.then(([result]) => (result ? result.bool : true));
```
SargeKhan: In case there are empty rows returned from the database, the result will be undefined which is falsy so the function will return true. Also, in the case when query returns non-empty rows, the value of result will be false hence the function will still return true. Hence, I am unsure when `([result]) => (result ? result.bool : true)` will return `result.bool`, and what actually is `result.bool`?
The tests seem to be correct, but  I am stiill confused by how it works :D

framework/src/modules/chain/components/storage/sql/temp_block/truncate.sql
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+/*
+  DESCRIPTION: Clear table temp_block
+  ONLY makes sure that only this table gets truncated
+
+  PARAMETERS: none
+*/
+
+TRUNCATE TABLE ONLY temp_block
```
SargeKhan: What does `ONLY` mean here?

lsilvs: > If ONLY is specified before the table name, only that table is truncated. If ONLY is not specified, the table and all its descendant tables (if any) are truncated.

framework/src/modules/chain/components/storage/entities/temp_block.js
michielmulders: I'm not sure what you mean with empty rows here?

When the table contains rows, it will return `false`. 
When there are no rows, no result is returned (that's why I removed `expectedResultCount` as it can be 0 or 1). 

So, if it contains rows -> return `false`.
No rows, result is undefined -> return `true`.

The `result.bool` comes from pgpromise as I'm only returning a bool as a result, it appends it under the `bool` property.

SargeKhan: By empty, I meant no rows. I was confused how value of the query result is assigned to result.bool property, but now I understand. Thanks for the explanation!

PR #4086, opened 2019-08-14
PeerPool removePeer should allow code or reason as parameter - Closes #4083

### LINKED ISSUE #4083: PeerPool removePeer should allow code or reason as parameter
### Expected behavior

`removePeer` should accept `reason` string for disconnect reason so logs are more descriptive.

### Actual behavior

Currently when inbound peer evicted it shows message: `INbound connection of peer 157.230.222.179:5000 was closed with code 1000 and reason: undefined`

### Steps to reproduce

SSH into alphanet 2.3.0 node and view logs. 

2.3

jondubois:
### What was the problem?

The logs did not show the disconnect reason most of the time.

### How did I solve it?

If the status reason is not provided by the other side of the socket, then use the default reason for that code from SocketCluster if it exists.

### How to manually test it?

Run tests and check against testnet.

### Review checklist

- [x] The PR resolves #4083

### DIFF
--- a/elements/lisk-p2p/src/peer/inbound.ts
+++ b/elements/lisk-p2p/src/peer/inbound.ts
@@ -58,8 +58,10 @@ export class InboundPeer extends Peer {
 		this._handleInboundSocketError = (error: Error) => {
 			this.emit(EVENT_INBOUND_SOCKET_ERROR, error);
 		};
-		this._handleInboundSocketClose = (code, reasonString) => {
-			const reason = reasonString ? reasonString : socketErrorStatusCodes[code];
+		this._handleInboundSocketClose = (code, reasonMessage) => {
+			const reason = reasonMessage
+				? reasonMessage
+				: socketErrorStatusCodes[code];
 			if (this._pingTimeoutId) {
 				clearTimeout(this._pingTimeoutId);
 			}
--- a/elements/lisk-p2p/src/peer/inbound.ts
+++ b/elements/lisk-p2p/src/peer/inbound.ts
@@ -61,7 +61,7 @@ export class InboundPeer extends Peer {
 		this._handleInboundSocketClose = (code, reasonMessage) => {
 			const reason = reasonMessage
 				? reasonMessage
-				: socketErrorStatusCodes[code];
+				: socketErrorStatusCodes[code] || 'Unknown reason';
 			if (this._pingTimeoutId) {
 				clearTimeout(this._pingTimeoutId);
 			}
--- a/elements/lisk-p2p/src/peer/inbound.ts
+++ b/elements/lisk-p2p/src/peer/inbound.ts
@@ -23,6 +23,10 @@ import {
 import { P2PDiscoveredPeerInfo } from '../p2p_types';
 
 import { SCServerSocket } from 'socketcluster-server';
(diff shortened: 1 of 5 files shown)


elements/lisk-p2p/src/peer/inbound.ts
```diff
@@ -23,6 +23,10 @@ import {
 import { P2PDiscoveredPeerInfo } from '../p2p_types';
 
 import { SCServerSocket } from 'socketcluster-server';
+const socketErrorStatusCodes = {
+	...(SCServerSocket as any).errorStatuses,
```
ishantiw: Do you think it might throw an error if `errorStatuses` is `undefined`?

elements/lisk-p2p/src/peer/inbound.ts
```diff
@@ -40,7 +44,7 @@ export class InboundPeer extends Peer {
 	protected readonly _handleInboundSocketError: (error: Error) => void;
 	protected readonly _handleInboundSocketClose: (
 		code: number,
-		reason: string,
+		reason: string | undefined,
```
ishantiw: Maybe instead of `undefined` we can give a message like `Unknown Reason`?

elements/lisk-p2p/src/peer/outbound.ts
```diff
@@ -47,6 +47,11 @@ export const EVENT_CLOSE_OUTBOUND = 'closeOutbound';
 export const EVENT_OUTBOUND_SOCKET_ERROR = 'outboundSocketError';
 export const RESPONSE_PONG = 'pong';
 
+const socketErrorStatusCodes = {
+	...(socketClusterClient.SCClientSocket as any).errorStatuses,
```
ishantiw: same here

elements/lisk-p2p/src/peer/outbound.ts
```diff
@@ -151,13 +156,19 @@ export class OutboundPeer extends Peer {
...
-		outboundSocket.on('close', (code: number, reason: string) => {
-			this.emit(EVENT_CLOSE_OUTBOUND, {
-				peerInfo: this._peerInfo,
-				code,
-				reason,
-			});
-		});
+		outboundSocket.on(
+			'close',
+			(code: number, reasonMessage: string | undefined) => {
```
ishantiw: instead of `undefined` use `Unknown Reason`

elements/lisk-p2p/src/peer/inbound.ts
jondubois: The `errorStatuses` object is always defined; it's part of the SC API.

elements/lisk-p2p/src/peer/inbound.ts
jondubois: This comes from SC so we can't control this input but good point. I made the change below.

elements/lisk-p2p/src/peer/outbound.ts
jondubois: As above, it will always exist. I had to do this casting because the DefinitelyTyped definition is incomplete.

jondubois:
@mitsujutsu Ready to review.

PR #4094, opened 2019-08-15
core:status command returns inconsistent output - Closes #4084

### LINKED ISSUE #4084: core:status command returns inconsistent output
### Expected behavior
`lisk core:status` always returns the same kind of output (e.g. always JSON)

### Actual behavior
`lisk core:status` may return JSON or a plain text error message (if there are no installed instances)

### Steps to reproduce
```
lisk core:status  # not JSON
lisk core:install lisk-mainnet
lisk core:status  # JSON
```

`2.2.2`

ManuGowda:
### What was the problem?
lisk `core:status` may return JSON or a plain text error message (if there are no installed instances)
### How did I solve it?
Return JSON result consistently
### How to manually test it?
`lisk core:status`
### Review checklist

- [ ] The PR resolves #4084 

### DIFF
--- a/commander/src/commands/core/status.ts
+++ b/commander/src/commands/core/status.ts
@@ -41,9 +41,9 @@ export default class StatusCommand extends BaseCommand {
 			const instance = await describeApplication(name);
 
 			if (!instance) {
-				this.log(
-					`Lisk Core instance: ${name} doesn't exists, Please install using lisk core:install`,
-				);
+				this.print({
+					message: `Lisk Core instance: ${name} doesn't exists, Please install using lisk core:install`,
+				});
 
 				return;
 			}
@@ -52,9 +52,10 @@ export default class StatusCommand extends BaseCommand {
 			const instances = await listApplication();
 
 			if (!instances.length) {
-				this.log(
-					'Lisk Core instances not available, use lisk core:install --help',
-				);
+				this.print({
+					message:
+						'Lisk Core instances not available, use lisk core:install --help',
+				});
 			} else {
 				const toDisplay = [
 					'name',
--- a/commander/src/commands/core/status.ts
+++ b/commander/src/commands/core/status.ts
@@ -52,7 +52,7 @@ export default class StatusCommand extends BaseCommand {
 
 			if (!instance) {
 				this.print({
-					message: `Lisk Core instance: ${name} doesn't exists, Please install using lisk core:install`,
+					message: `Lisk Core instance: ${name} doesn't exists, Please install using lisk core:install --help`,
 				});
 
 				return;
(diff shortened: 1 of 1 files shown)


ManuGowda:
@fchavant build: https://jenkins.lisk.io/job/lisk-qa/job/lisk-commander-build-tag/37/

shuse2:
@ManuGowda probably it's better to have consistent message as @fchavant mentioned

ManuGowda:
Can you guys suggest the message format, I am not a native speaker so 🤷‍♂

Nazgolze:
Well, `ls` in Linux has this:
```ls -T
ls: option requires an argument -- 'T'
Try 'ls --help' for more information.
```
So maybe `"message": "Lisk Core instances not available, try 'lisk core:install --help' for more information"` would be acceptable

Nazgolze:
Of course, the `ls` example is a little different. That's me not providing the right input, which can be handled differently from the program erring. I suspect it's also more common in the former case to print the usage/help instead of telling the user to do it, but I don't have a strong inclination either way

I'll demonstrate the difference of what I mean (invalid input vs program error):
```
jg@mustang ~ $ ping -z
ping: invalid option -- 'z'

Usage
  ping [options] <destination>

Options:
  <destination>      dns name or ip address
  -a                 use audible ping
  -A                 use adaptive ping
  -B                 sticky source address
  -c <count>         stop after <count> replies
  -D                 print timestamps
  -d                 use SO_DEBUG socket option
  -f                 flood ping
  -h                 print help and exit
  -I <interface>     either interface name or address
  -i <interval>      seconds between sending each packet
  -L                 suppress loopback of multicast packets
  -l <preload>       send <preload> number of packages while waiting replies
  -m <mark>          tag the packets going out
  -M <pmtud opt>     define mtu discovery, can be one of <do|dont|want>
  -n                 no dns name resolution
  -O                 report outstanding replies
  -p <pattern>       contents of padding byte
  -q                 quiet output
  -Q <tclass>        use quality of service <tclass> bits
  -s <size>          use <size> as number of data bytes to be sent
  -S <size>          use <size> as SO_SNDBUF socket option value
  -t <ttl>           define time to live
  -U                 print user-to-user latency
  -v                 verbose output
  -V                 print version and exit
  -w <deadline>      reply wait <deadline> in seconds
  -W <timeout>       time to wait for response

IPv4 options:
  -4                 use IPv4
  -b                 allow pinging broadcast
  -R                 record route
  -T <timestamp>     define timestamp, can be one of <tsonly|tsandaddr|tsprespec>

IPv6 options:
  -6                 use IPv6
  -F <flowlabel>     define flow label, default is random
  -N <nodeinfo opt>  use icmp6 node info query, try <help> as argument

For more details see ping(8).
jg@mustang ~ $ ping 30.0.0.3
ping: socket: Operation not permitted
jg@mustang ~ $
```

PR #4100, opened 2019-08-16
Add network ready event - Closes #4090

### LINKED ISSUE #4090: Network Ready event for P2P layer
### Description

P2P network layer should emit event `NETWORK_READY` after peers found and connected to.

After a specified period of time, if no peers found the node should crash with error. 

### Motivation

Upon node startup and sync, nodes were requesting data from the network too early resulting in `RequestFailError: Request failed due to no peers found in peer selection` error. This issue is connected to [this core issue.](https://github.com/LiskHQ/lisk-core/issues/114)

### Acceptance Criteria

Run 2.3 node with fresh db and connect to testnet. Check logs and make sure there is no `RequestFailError`.

mitsujutsu:
### What was the problem?

Node was attempting to load transactions and signatures before it was connected to any peers on p2p network. This caused an error message: `RequestFailError: Request failed due to no peers found in peer selection`.

### How did I solve it?

1. Created `EVENT_NETWORK_READY` which the network module listens for before publishing `network:bootstrap` and calling `loader.loadTransactionsAndSignatures`.

2. `peerSelectForRequest` now simply shuffles peerList and selects one, instead of using height histogram. 

### How to manually test it?

Start 2.3.0 node and connect to testnet. Observe that there is no `RequestFailError`.

### Review checklist

- [x] The PR resolves #4090 

### DIFF
--- a/elements/lisk-p2p/src/peer_pool.ts
+++ b/elements/lisk-p2p/src/peer_pool.ts
@@ -182,6 +182,7 @@ export class PeerPool extends EventEmitter {
 	private readonly _maxOutboundConnections: number;
 	private readonly _maxInboundConnections: number;
 	private readonly _peerSelectForSend: P2PPeerSelectionForSendFunction;
+	private readonly _peerSelectForRequest: P2PPeerSelectionForRequestFunction;
 	private readonly _peerSelectForConnection: P2PPeerSelectionForConnectionFunction;
 	private readonly _sendPeerLimit: number;
 	private readonly _outboundShuffleIntervalId: NodeJS.Timer | undefined;
@@ -191,6 +192,7 @@ export class PeerPool extends EventEmitter {
 		this._peerMap = new Map();
 		this._peerPoolConfig = peerPoolConfig;
 		this._peerSelectForSend = peerPoolConfig.peerSelectionForSend;
+		this._peerSelectForRequest = peerPoolConfig.peerSelectionForRequest;
 		this._peerSelectForConnection = peerPoolConfig.peerSelectionForConnection;
 		this._maxOutboundConnections = peerPoolConfig.maxOutboundConnections;
 		this._maxInboundConnections = peerPoolConfig.maxInboundConnections;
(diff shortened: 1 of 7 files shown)


framework/src/modules/network/network.js
```diff
@@ -166,6 +167,10 @@ module.exports = class Network {
 		});
 
 		// ---- START: Bind event handlers ----
+		this.p2p.on(EVENT_NETWORK_READY, () => {
+			this.logger.debug('Network has connected to peers');
```
ManuGowda: `Node has connected to network`?

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -309,22 +307,15 @@ export class PeerPool extends EventEmitter {
 	}
 
 	public async request(packet: P2PRequestPacket): Promise<P2PResponsePacket> {
-		const listOfPeerInfo = [...this._peerMap.values()].map(
-			(peer: Peer) => peer.peerInfo as P2PDiscoveredPeerInfo,
-		);
-		const selectedPeers = this._peerSelectForRequest({
```
ishantiw: Should we change the selection function instead of completely bypassing selection function? because then the custom selection function for the request will not work. So let's modify the existing selection functions to simply return one incoming peerlist for both send and request and remove the height logic

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -309,21 +309,19 @@ export class PeerPool extends EventEmitter {
 	}
 
 	public async request(packet: P2PRequestPacket): Promise<P2PResponsePacket> {
-		const listOfPeerInfo = [...this._peerMap.values()].map(
-			(peer: Peer) => peer.peerInfo as P2PDiscoveredPeerInfo,
-		);
+		const listOfPeerInfo = [...this._peerMap.values()]
+			.filter(p => p.state === ConnectionState.OPEN)
```
jondubois: I don't think we need this filtering since:

- We delete the peer from our peerMap as soon as we detect that they've disconnected.
- Even if the peer state is connected, it still does not fully guarantee that the peer is actually connected right now because WebSocket/TCP relies on a ping/pong mechanism to detect unintentional lost connection so there can be a delay between the lost connection and when it is detected (and the socket/peer state is updated).

PR #4104, opened 2019-08-16
Fix syncing logic for reverting - Closes #4103

### LINKED ISSUE #4103: Syncing should revert a block when loading from network failed
### Expected behavior
Syncing should revert the block when it fails to load blocks from the network, but not when it fails to process.

https://github.com/LiskHQ/lisk-sdk/blob/v1.6.0/framework/src/modules/chain/submodules/blocks/process.js#L254-L343

### Actual behavior

Behavior is opposite. It reverts a block when it fails to process, but it does not revert on the network load fails.

### Steps to reproduce
Build a blockchain in a network, and connect to another network with the same dataset.

2.2-

shuse2:
### What was the problem?
When the chain is on fork, it doesn't rollback if it fails to fetch block from network, instead it rolls back when it fails to process

### How did I solve it?
Move the logic to the network fail.

### How to manually test it?
Build some blockchain data (around height 100), and connect to different network.

### Review checklist

- [ ] The PR resolves #4103 

### DIFF
--- a/framework/src/modules/chain/loader.js
+++ b/framework/src/modules/chain/loader.js
@@ -370,7 +370,10 @@ class Loader {
 				failedAttemptsToLoad += 1;
 				// eslint-disable-next-line no-await-in-loop
 				await this._handleCommonBlockError(err);
-				this.logger.warn({ error: err });
+				this.logger.warn(
+					{ error: err },
+					'Failed to load blocks from the network.',
+				);
 			}
 		}
 	}
@@ -391,7 +394,6 @@ class Loader {
 				);
 			}
 		}
-		this.logger.warn({ error }, 'Failed to load blocks from the network.');
 	}
 }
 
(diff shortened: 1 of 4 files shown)


framework/src/modules/chain/loader.js
```diff
@@ -385,24 +355,49 @@ class Loader {
...
-					failedAttemptsToLoad += 1;
-					this.logger.error(convertErrorsToString(err));
-				}
+				failedAttemptsToLoad += 1;
+				this.logger.error(convertErrorsToString(err));
 			}
 		}
 	}
+
+	async _handleBlocksFromNetworkError(error) {
```
4miners: This function is triggering chain recovery, the name is confusing and doesn't reflect the behavior. Also, with this logic, we have nested try-catch block (3 levels). Can we refactor this in a better way?

shuse2: yes, i found a bug in this way, so im gonna clean it up

PR #4109, opened 2019-08-16
Syncing halted due to block max payload size exceeded - Closes #4083

### LINKED ISSUE #4083: PeerPool removePeer should allow code or reason as parameter

jondubois:
### What was the problem?

The payload limit was too low.

### How did I solve it?

- Increased the message payload limit.
- Added separate limits for PeerInfo and PeerInfoLists.
- Allow user to configure all these limits from config file.

### How to manually test it?

Run tests

### Review checklist

- [ ] The PR resolves #4083


elements/lisk-p2p/src/validation.ts
```diff
@@ -117,19 +123,33 @@ export const validatePeerInfo = (rawPeerInfo: unknown): P2PPeerInfo => {
 
 	const { ip, ...peerInfoUpdated } = peerInfo;
 
+	const byteSize = getByteSize(peerInfoUpdated);
```
ishantiw: I think we can move this check right at the beginning of `validatePeerInfo` function before going into all the validations or transformation of `peerInfo`

elements/lisk-p2p/src/p2p.ts
```diff
@@ -779,18 +788,17 @@ export class P2P extends EventEmitter {
 			.minimumPeerDiscoveryThreshold
 			? this._config.minimumPeerDiscoveryThreshold
 			: DEFAULT_MIN_PEER_DISCOVERY_THRESHOLD;
-		const maximumPeerDiscoveryResponseSize = this._config
-			.maximumPeerDiscoveryResponseSize
-			? this._config.maximumPeerDiscoveryResponseSize
-			: DEFAULT_MAX_PEER_DISCOVERY_RESPONSE_SIZE;
+		const peerDiscoveryResponseLength = this._config.peerDiscoveryResponseLength
```
mitsujutsu: I think we can leave `maximum` in the name here, as it describes the maximum and not the actual response length.

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -429,6 +431,9 @@ export class PeerPool extends EventEmitter {
 			ackTimeout: this._peerPoolConfig.ackTimeout,
 			wsMaxMessageRate: this._peerPoolConfig.wsMaxMessageRate,
 			wsMaxMessageRatePenalty: this._peerPoolConfig.wsMaxMessageRatePenalty,
+			maxPeerDiscoveryResponseLength: this._peerPoolConfig
```
mitsujutsu: Does peer need to know this?

elements/lisk-p2p/src/p2p.ts
jondubois: @mitsujutsu I separated those two. I renamed the variables. The max length of the peer list that we accept from peers can be different from the length that we provide to peers. I'm worried that if we don't separate those it could make updates more difficult in the future.

elements/lisk-p2p/src/peer_pool.ts
jondubois: The validation function which needs it is inside the peer list. It's a pure function so we can't pass it in another way.

elements/lisk-p2p/src/validation.ts
jondubois: Ok. I changed it so that it just removes peers that are too large from the list when validating the list instead of rejecting the entire list if just one peer is too big. So then I was able to implement it as you recommend.

PR #4112, opened 2019-08-19
Load minimum block headers on block deletion - Closes #3694

### LINKED ISSUE #3694: Load minimum block headers on block deletion
### Expected behavior

Integrate submodule/bft with the chain that it should trigger activity on block deletion to maintain minimum block headers in the BFT. 

3.0.0

nazarhussain:
### What was the problem?

The interface to delete block headers from BFT was missing

### How did I solve it?

Added the new interfaces and tests as well. 

### How to manually test it?

Run bft unit tests. 

### Review checklist

- [ ] The PR resolves #3694 


framework/src/modules/chain/bft/bft.js
```diff
@@ -84,12 +85,62 @@ class BFT extends EventEmitter {
...
+
+		assert(
+			!blockHeights.some(h => h <= this.finalityManager.finalizedHeight),
+			'Can not delete block below or same as finalized height',
+		);
+
+		const removeFromHeight = Math.min(...blockHeights);
+
+		this.finalityManager.removeBlockHeaders({
+			aboveHeight: removeFromHeight - 1,
```
michielmulders: Can you clarify with a comment why `removeFromHeight - 1` is applied here? Thanks! :)

framework/src/modules/chain/bft/bft.js
```diff
@@ -84,12 +85,62 @@ class BFT extends EventEmitter {
...
+		});
+
+		// Make sure there are 2 rounds of block headers available
+		if (
+			this.finalityManager.maxHeight - this.finalityManager.minHeight <
+			this.constants.activeDelegates * 2
```
michielmulders: Great use of constants! 👍

framework/src/modules/chain/bft/bft.js
```diff
@@ -136,23 +187,30 @@ class BFT extends EventEmitter {
...
 	 * Load blocks into consensus manager fetching from storage
 	 *
-	 * @param {integer} fromHeight - The start height to fetch and load
-	 * @param {integer} tillHeight - The end height to fetch and load
+	 * @param {int} fromHeight - The start height to fetch and load
+	 * @param {int} tillHeight - The end height to fetch and load
 	 * @return {Promise<void>}
 	 */
-	async loadBlocks({ fromHeight, tillHeight }) {
+	async loadBlocksFromStorage({ fromHeight, tillHeight }) {
```
michielmulders: Do we want to put this `blocks` related function inside of BFT? Can we move it to blocksLogic?

framework/src/modules/chain/bft/bft.js
```diff
@@ -84,12 +85,62 @@ class BFT extends EventEmitter {
...
 		});
 	}
 
+	/**
+	 * When blocks deleted send those to BFT to update BFT state
+	 *
+	 * @param {Array.<Object>} blocks - List of all blocks
+	 * @return {Promise<void>}
+	 */
+	async onDeleteBlocks(blocks) {
```
shuse2: `onNewBlock` takes single block, but why `onDeleteBlocks` takes array?
Probably it should be consistent

framework/src/modules/chain/bft/bft.js
```diff
@@ -161,11 +219,9 @@ class BFT extends EventEmitter {
 	}
 }
 
-const exportedInterface = {
+module.exports = {
```
shuse2: ❤️

framework/test/jest/unit/specs/modules/chain/bft/bft.spec.js
```diff
@@ -21,11 +21,19 @@ const {
...
 
-jest.mock('../../../../../../../src/modules/chain/bft/finality_manager');
-
-const { BFT, extractBFTBlockHeaderFromBlock } = bftModule;
+const {
+	BFT,
+	extractBFTBlockHeaderFromBlock,
+} = require('../../../../../../../src/modules/chain/bft/bft');
+
+const generateBlocks = ({ startHeight, numberOfBlocks }) => {
```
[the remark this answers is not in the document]
nazarhussain: I want to have named arguments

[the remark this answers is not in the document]
shuse2: I meant
({ startHeight, numberOfBlocks }) => `{` <= this one

framework/src/modules/chain/bft/bft.js
```diff
@@ -66,7 +67,7 @@ class BFT extends EventEmitter {
 	 *
 	 * @return {Promise<void>}
 	 */
-	async init() {
+	async bootstrap() {
```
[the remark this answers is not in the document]
nazarhussain: Just to align with functions with places. We use `bootstrap` a lot. e.g. Chain bootstrap, cache bootstrap, storage bootstrap.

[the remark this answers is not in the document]
shuse2: Since this is not `module` nor `components`, I think it make sense to keep `init`.

framework/src/modules/chain/bft/bft.js
nazarhussain: `onNewBlock` works when you receive or forge new block. While in fork recovery, we know we may delete multiple blocks, so why I make these different. If you prefer it to be consistent then I will make `onDeleteBlocks` to work with one bock only.

framework/src/modules/chain/bft/bft.js
nazarhussain: Because of `removeBlockHeaders` have `aboveHeight` param. Which removes block headers above a point. So if you want to delete the current one, you have to do a step back.

framework/src/modules/chain/bft/bft.js
SargeKhan: Using event might lead to unexpected behavior. BFT module decides the validity of the blocks based on the information stored internally. Since a block is added/deleted atomically and in a synchronous manner and the addition/removal of block header should also be performed synchronously during the processing of the block addition/deletion. Otherwise, if we rely on events, we may verify an invalid block header in bft because some blockheader was not yet added to bft module through events.

framework/src/modules/chain/bft/bft.js
nazarhussain: Yes its BFT related logic. We can probably make it private, I kept it public just in case we came up with scenario to load blocks directly from storage somewhere.

framework/src/modules/chain/bft/bft.js
SargeKhan: Also, I also think this function should be called after each block deletoin.

framework/src/modules/chain/bft/bft.js
```diff
@@ -84,12 +85,62 @@ class BFT extends EventEmitter {
...
+		}
+	}
+
+	/**
+	 * Load new block to BFT
+	 *
+	 * @param {Object} block - The block which is forged
+	 * @return {Promise<void>}
+	 */
+	async onNewBlock(block) {
```
SargeKhan: Here as well.

framework/src/modules/chain/bft/bft.js
nazarhussain: Can't understand your comment, what do you meant by automatically? 

> Since a block is added/deleted atomically 

Also we can't make it synchronous as we have to perform IO call to load blocks.

framework/src/modules/chain/bft/bft.js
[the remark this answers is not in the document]
nazarhussain: Even thought it meant the same no matter `init` or `bootstrap`. So why not use the consistent name.

framework/src/modules/chain/bft/bft.js
SargeKhan: *Atomically. as part of block processing. Sorry for the confusion.

framework/src/modules/chain/bft/bft.js
nazarhussain: @SargeKhan Its not an event. Its a function that you can invoke or bound to an event.

SargeKhan: We have the convention of using `on` for functions which are triggered via events in the chain module. This function should be executed as part of block processing. Perhaps we can update the name to `addNewBlockHeader` in order to add further clarity?

framework/src/modules/chain/bft/bft.js
nazarhussain: Block processing it self is an asynchronous task, so its not necessary to make this one synchronous. And we really can't make it synchronous as we have to load the blocks from storage as well.

SargeKhan: Maybe synchronous is an incorrect word. What I mean is that each block is processed in a series and after processing one block, the blockchain state is valid; As we process one block, we apply all the state relevant for that block as part of processing i.e, applying transaction, saving transactions and blocks. The same thing happens for block deletion. And the BFT module should behave in a similar fashion. A block header should be added/removed from the BFT module as part of a single block addition/deletion rather than deleting a set of blocks and then removing those headers from the BFT module.
Blocks and transactions are also applied asynchronously, but they are executed within an atomic block processing task, same should go for BFT header addition/deletion.

nazarhussain: I explained in very detail to @shuse2 why I designed it this way. We know a clear use case of fork elimination which requires deleting multiple blocks, which is obvious and will happen often. And since deleting blocks headers from BFT is a compute intensive task so we should not perform it every block, rather collective. 

If you guys still we should do it for single block, then we can use `[block]` as parameter, by keeping the same implementation. So we can decide while integrating BFT which approach to use.

framework/src/modules/chain/bft/bft.js
nazarhussain: I don't see any such conventions around project in current state. But idea was similar that `onNewBlock` should be bound to event outside BFT.

framework/src/modules/chain/bft/bft.js
michielmulders: Would it make sense to add this info as a `//` comment above this line of code? Thanks!

nazarhussain:
@shuse2 Some tests in the commander are failing, so why PR not getting green.

framework/src/modules/chain/bft/finality_manager.js
```diff
@@ -117,6 +117,25 @@ class FinalityManager extends EventEmitter {
...
 
+	/**
+	 * Remove block headers above given height (exclusive)
+	 * If no param provided will remove last block header
+	 *
+	 *
+	 * @param {int} aboveHeight -  Height from above remove headers
+	 */
+	removeBlockHeaders({ aboveHeight }) {
+		debug('removeBlockHeaders invoked');
```
SargeKhan: Is this something new we started using in the application?

shuse2: We have been using this in http API, but general consensus is this can be used in lowest layer

PR #4120, opened 2019-08-20
Sync error on testnet - Closes #4119

### LINKED ISSUE #4119: Sync error on testnet
### Expected behavior
I expect to be able to sync with testnet

### Actual behavior
The following error occur at round 134:
`
Chain recovery failed chain recovery after failing to load blocks Error: Account with address = 1690488743833175495L does not exist
`
### Steps to reproduce
Start syncing process for testnet

lsilvs:
### What was the problem?
Syncing failing on testnet due to account not being cached in StateStore. StateStore uses `storage.entities.Account.get` which default limit result is 101 entries meaning it was only caching maximum 101 accounts.

### How did I solve it?
By adding `limit: null` to remove the limit when calling `storage.entities.Account.get`.
The same approach was introduced to `storage.entities.Transaction.get` inside StateStore to keep consistency.

### How to manually test it?
Start syncing with test. Is should pass round 134

### Review checklistd

- [ ] The PR resolves #4119

### DIFF
--- a/framework/src/modules/chain/transactions/votes_weight.js
+++ b/framework/src/modules/chain/transactions/votes_weight.js
@@ -206,30 +206,35 @@ const undo = (stateStore, transaction, exceptions = {}) => {
 
 const prepare = async (stateStore, transaction) => {
 	// Get delegate public keys whom sender voted for
-	const senderDelegatePks =
+	const senderDelegatePublicKeys =
 		stateStore.account.getOrDefault(transaction.senderId)
 			.votedDelegatesPublicKeys || [];
 
 	const recipientId = getRecipientAddress(stateStore, transaction);
 
 	// Get delegate public keys whom recipient voted for
-	const recipientDelegatePks =
+	const recipientDelegatePublicKeys =
 		(recipientId &&
 			stateStore.account.getOrDefault(recipientId).votedDelegatesPublicKeys) ||
 		[];
 
 	// Get unique public keys from merged list
-	const uniqPksToBeCached = [
-		...new Set([...senderDelegatePks, ...recipientDelegatePks]),
+	const uniqPublicKeysToBeCached = [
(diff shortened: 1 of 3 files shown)


framework/src/modules/chain/transactions/votes_weight.js
```diff
@@ -206,30 +206,35 @@ const undo = (stateStore, transaction, exceptions = {}) => {
...
+	 * because StateStore uses `storage.entities.Account.get` which default limit result is 101 entries
+	 * meaning it will only cache maximum 101 accounts.
+	 */
+	return Promise.all(
+		uniqPublicKeysToBeCached
+			.map(delegatePublicKey => ({
+				address: getAddressFromPublicKey(delegatePublicKey),
+			}))
+			.map(filter => stateStore.account.cache(filter)),
+	);
```
nazarhussain: This change comes with a big performance degradation. This will result into minimum 1 to 34 public keys. So in worst case scenario you are converting 1 SQL query to 34 different queries. That I would not recommend to do. 

Apart from this change, this line actually not makes any sense to SDK developers.
https://github.com/LiskHQ/lisk-sdk/blob/3a866573edf27f824837cc478c0abdd2aacee264/framework/src/modules/chain/state_store/account_store.js#L53

For a custom transaction developer, they can use `cache(filters)` in the `prepare` step. And what ever matching records will come for `filters` they will be cached. We had not documented anywhere that it will not cache more than 101 accounts, and which actually does not makes sense in the prepare steps. 

So two solutions can be done. 
1. Add `limit: null` inside cache method and add condition that user must provide `filters` for the cache. 
2. Or add `options` as second parameter to `cache` method so user can set himself what options needed for storage.

lsilvs: After discussion with the team the `limit: null` solution was chosen.

lsilvs: After discussion with the team the limit: null solution was chosen.

PR #4127, opened 2019-08-22
Fix invalid use of callback

shuse2:
### What was the problem?
Sequence was expecting the callback to be called, but it was throwing.
This was solved by changing sequence to async await in 2.2, but it still exists in 2.0.1.

### How did I solve it?
Try catch the async-await and call the callback.

### Review checklist


ManuGowda:
@shuse2 is it possible to cover this with tests? otherwise PR LGTM

shuse2:
@ManuGowda it is possible but this code is removed already in 2.2, so it should be sufficient to test manually imo

PR #4128, opened 2019-08-22
Fix slow http api/account - Closes #4122

### LINKED ISSUE #4122: /api/accounts performance is too low
### Expected behavior
/api/accounts endpoint should be less than 100ms

### Actual behavior
Each accounts request takes around 300ms
`/api/accounts?address=xxxL` should be less than 

### Steps to reproduce
call
`http://localhost:8000/api/accounts?address=xxxL`

hotfix/2.3.1

shuse2:
### What was the problem?
`getNodeStatus` call uses transaction count and it's slow

### How did I solve it?
Change not to use `getNodeStatus` function
Also, include `asset` in default account sql and remove extended flag

### How to manually test it?
Call endpoint `/api/accounts=address=XXXL` with mainnet data

### Review checklist

- [ ] The PR resolves #4122 

### DIFF
--- a/framework/src/modules/chain/index.js
+++ b/framework/src/modules/chain/index.js
@@ -122,6 +122,9 @@ module.exports = class ChainModule extends BaseModule {
 			getNodeStatus: {
 				handler: async () => this.chain.actions.getNodeStatus(),
 			},
+			getLastBlock: {
+				handler: async () => this.chain.actions.getLastBlock(),
+			},
 			blocks: {
 				handler: async action => this.chain.actions.blocks(action),
 				isPublic: true,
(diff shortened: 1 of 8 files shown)


framework/src/modules/chain/index.js
```diff
@@ -122,6 +122,9 @@ module.exports = class ChainModule extends BaseModule {
 			getNodeStatus: {
 				handler: async () => this.chain.actions.getNodeStatus(),
 			},
+			getLastBlock: {
+				handler: async () => this.chain.actions.getLastBlock(),
```
michielmulders: Would it make sense to make this even public?

shuse2: I think it will need to be in 3.0, but let's not do it in this PR

PR #4134, opened 2019-08-23
Calculate consensus only from outbound connected peers - Closes #3969

### LINKED ISSUE #3969: Avoid low broadhash consensus attacks in partial mesh network
### Expected behavior

Broadhash consensus should be greater than 50% for forging a block.

### Actual behavior

Since in partial view of the network, a node can have a max of 20 outbound connections and 100 inbound connections. We take these peers (120 peers in total) into account to calculate broadhash consensus, there is a possibility that an attacker can start many malicious nodes and connect to a particular node in the network and manipulate its broadhash consensus. The worse case will be if the attacker knows the IP of any delegate then it can manipulate the broadhash consensus of that delegate by creating a lot of incoming connections and make him miss its slot to forge. 

We can avoid it by only trusting the outbound connections and only use them for calculating broadhash consensus. Also, with a low(10 connections) outbound connection, there is a possibility that a node is connected to more than 10 malicious nodes that can also change its broadhash consensus. So we can use 40 or 50 outbound connections that will be big enough to avoid this kind of attack.

### Steps to reproduce

Start a large number of nodes and connect to a node in the network.

`v2.3`

diego-G:
### How to manually test it?

```
npx mocha --opts test/mocha/mocha.opts test/mocha/unit/modules/chain/peers.js
npx mocha --opts test/mocha/mocha.opts test/mocha/unit/modules/http_api/controllers/node.js
npx mocha --opts test/mocha/mocha.opts test/mocha/unit/modules/network/lookup_peers_ips.js 
```

### Review checklist

- [x] The PR resolves #3969


elements/lisk-p2p/src/p2p.ts
```diff
@@ -531,13 +530,34 @@ export class P2P extends EventEmitter {
...
-			newPeers: [...this._peerBook.newPeers],
-			triedPeers: [...this._peerBook.triedPeers],
-			connectedPeers: this._peerPool.getAllConnectedPeerInfos(),
-			connectedUniquePeers: this._peerPool.getUniqueConnectedPeers(),
-		};
+	public getConnectedPeers(): ReadonlyArray<P2PDiscoveredPeerInfo> {
+		return this._peerPool.getAllConnectedPeerInfos();
+	}
+
+	public getUniqueOutboundConnectedPeers(): ReadonlyArray<
```
mitsujutsu: should we modify this to allow for both `outbound` and `inbound`?

diego-G: For now, we don't need `Inbound` due to security reasons. That's why I haven't done it like that.

framework/src/modules/network/index.js
```diff
@@ -60,21 +60,15 @@ module.exports = class NetworkModule extends BaseModule {
...
-			getConnectedPeers: {
-				handler: action => this.network.actions.getConnectedPeers(action),
-			},
-			getPeersCountByFilter: {
-				handler: action => this.network.actions.getPeersCountByFilter(action),
+			getPeersCount: {
+				handler: action => this.network.actions.getPeersCount(action),
 			},
-			getConnectedPeersCountByFilter: {
+			getUniqueOutboundConnectedPeersCount: {
```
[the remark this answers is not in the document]
diego-G: Let's discuss it when they need it. `getUniqueOutboundConnectedPeersCount` action must be removed when we get rid of broadhash consensus concept.

PR #4151, opened 2019-08-27
Protocol-spec second signature - closes #4148

### LINKED ISSUE #4148: Protocol-spec second signature
### Description

- Do not include genesis block in output
- Basic Second signature acceptance (account should end up with 5 lisk less and have second public key)
- Combination of second signature and transfer transactions in a same block (account should remain unchanged. Transaction should not be in the block)
- Double registration (i.e. an account with already registered signature tries to register again) (account should remain unchanged. Transaction should not be in the block)

### Acceptance Criteria

JSON generator for: 

- [ ] Do not include genesis block in output
- [x] Basic Second signature acceptance
- [x] Combination of second signature and transfer transactions in a same block
- [x] Double registration (i.e. an account with already registered signature tries to register again)

### Additional Information

Split from https://github.com/LiskHQ/lisk-sdk/issues/4087

pablitovicente:
### What was the problem?

- Genesis block was being used and was requested to be removed
- No protocol spec generators for account's second signature existed

### How did I solve it?

By adding the required generators and its JSON output

### How to manually test it?

- Build should be green 
- run `protocol-specs/index.js` and check output in `generator_outputs/block_processing_second_signature`

### Review checklist

- [x] The PR resolves #4148


protocol-specs/generators/block_processing_second_signature/index.js
```diff
@@ -0,0 +1,2436 @@
...
+const BigNum = require('@liskhq/bignum');
+const BaseGenerator = require('../base_generator');
+const defaultConfig = require('../../config/devnet');
+const { createBlock } = require('../../utils/blocks');
+
+const { genesisBlock } = defaultConfig;
+
+// Computed within Client application
+// TODO: Compute the initial account state here
+const initialAccountState = [
```
shuse2: do we need to compare all the account?
I think only the target account is good enough

pablitovicente: I followed the existing pattern in the code. I will see how to change.

pablitovicente: I removed all the accounts and left only the ones affected by the cases. Please review again @shuse2

limiaspasdaniel: As a side note: I think we need to rethink this initial account state. The way I obtained it was running devnet and converting accounts table to JSON, but this might cause issues in the future.

pablitovicente: @limiaspasdaniel yes; this feedback seems to indicate that we want to re-work the accounts states. I will consider different options for the delegates spec to see how a different approach could work out.

Issue #4045, opened 2019-08-05
Unit tests for eviction of peers from peer buckets


ishantiw:
### Description

After the implementation of peer buckets its important to have an extensive test for peer eviction from the buckets of each peer lists that have different strategies of eviction.

### Motivation

Since each of these peer list's have different eviction strategy on their buckets we need to test it by filling up the buckets with valid IPs and test the strategy based on time or random evictions.

### Acceptance Criteria

- Should test eviction based on time in `newPeers` and random eviction if no old peers
- Should test eviction based on random when the bucket is full in `triedPeers` and moved to `newPeers`

ishantiw:
Unit test cases for eviction strategies are covered in PR #4160 

- Unit test for eviction based on random strategy 998a34c738f805c4b9ba472a0342be469480e7e7
- Unit test for eviction based on time in NewPeers list bc61dceea279cd1e313c77b1d2e4898a18753318
- Unit test for eviction based on random selection in TriedPeers list 879392a95c6b0e1d7ece8308b2674b3df69ba792

PR #4160, opened 2019-08-27
Create base class peer list - Closes #4048

### LINKED ISSUE #4048: Create base class for peer list for peer directory
### Description

A lot of redundant and similar methods in new/triedPeers.

### Motivation

We need a base class and new/tried Peers can extend from it.

### Acceptance Criteria

All the tests related to new/Tried Peers should pass and should implement the base class method that differs.

ishantiw:
### Review checklist

- [x] The PR resolves #4048 #4045 


elements/lisk-p2p/src/peer_directory/base_list.ts
```diff
@@ -0,0 +1,213 @@
...
+		return {
+			success: true,
+			isEvicted: true,
+			evictedPeer: evictedPeer.peerInfo,
+		};
+	}
+
+	// This action is called when a peer is disconnected
+	public failedConnectionAction(incomingPeerInfo: P2PPeerInfo): boolean {
+		const success = this.removePeer(incomingPeerInfo);
```
mitsujutsu: this variable can be success or fail, so maybe `result` instead of `sucess`, or simply return this.removePeer(incomingPeerInfo)

elements/lisk-p2p/src/peer_directory/base_list.ts
```diff
@@ -0,0 +1,213 @@
...
+
+		return success;
+	}
+
+	public evictPeer(bucketId: number): CustomPeerInfo {
```
mitsujutsu: Is this necessary if it's only wrapping?

elements/lisk-p2p/src/peer_directory/new_peers.ts
```diff
@@ -12,193 +12,48 @@
  * Removal or modification of this copyright notice is prohibited.
```
mitsujutsu: I propose renaming this and `tried_peers` to `new_list` and `tried_list` to stay in conjunction with `base_list`

elements/lisk-p2p/src/peer_directory/tried_peers.ts
```diff
@@ -222,26 +113,8 @@ export class TriedPeers {
 		};
```
mitsujutsu: this name `newTriedPeerInfo` is confusing, I suggest `updatedTriedPeerInfo`

elements/lisk-p2p/src/peer_directory/peer_book.ts
```diff
@@ -31,18 +42,34 @@ export class PeerBook {
...
 
 	public get newPeers(): ReadonlyArray<P2PPeerInfo> {
-		return this._newPeers.newPeersList();
+		return this._newPeers.peersList();
 	}
 
 	public get triedPeers(): ReadonlyArray<P2PDiscoveredPeerInfo> {
-		return this._triedPeers.triedPeersList();
+		return this._triedPeers.peersList() as ReadonlyArray<P2PDiscoveredPeerInfo>;
 	}
```
mitsujutsu: not from this PR but imo it would be nicer if the order of listed functions after `getAllPeers` below is: `getPeer`, `updatePeer`, `addPeer`, `removePeer`, `upgradePeer`, `downgradePeer`.

elements/lisk-p2p/src/peer_directory/new_peers.ts
diego-G: I like this proposal. I'd go even further because `new` is not exactly what we have since we move back `tried` peers to `new`. Could we find a better adjective for `new` peers?

elements/lisk-p2p/src/peer_directory/base_list.ts
ishantiw: It seems like a wrapper in `base_list` but I'm using it as a function that will be used to override this function with your own eviction strategies. Whereas the function that is being called inside is just one eviction strategy that is by default in `base_list`

mitsujutsu: I see. Can `evictRandomlyFromBucket` be protected in this case?

ishantiw: Yes thats a good point, we can make it protected

elements/lisk-p2p/src/peer_directory/peer_book.ts
diego-G: @ishantiw I'd make it in this PR. Why not? It is straightforward change.

ishantiw: Oh I've made the changes in this PR already

elements/lisk-p2p/src/peer_directory/base_list.ts
```diff
@@ -0,0 +1,213 @@
...
+import { P2PPeerInfo } from '../p2p_types';
+import { constructPeerIdFromPeerInfo, getBucket, PEER_TYPE } from '../utils';
+
+export interface PeerListConfig {
+	readonly peerBucketCount: number;
+	readonly peerBucketSize: number;
+	readonly secret: number;
+	readonly peerType: PEER_TYPE;
+}
+export interface CustomPeerInfo {
```
diego-G: I'd extend from the regular peer info instead of creating a new one. In that case, we should move this to `p2p_types`.

ishantiw: Actually this type is confined to peer_directory and we never want to disclose `CustomPeerInfo` to upper classes so this is the reason to keep it here. 
And we are not extending it from peerInfo but we are keeping it as a field and this is more like wrapper around it. Its because if a custom field is overwritten by peerInfo coming from the network it can manipulate or affect the peer_bucket functionality

elements/lisk-p2p/src/peer_directory/new_list.ts
```diff
@@ -0,0 +1,103 @@
...
+	// Evict a peer when a bucket is full based on the time of residence in a peerlist
+	private _evictionBasedOnTimeInBucket(
+		bucketId: number,
+		peerList: Map<string, CustomPeerInfo>,
+	): CustomPeerInfo | undefined {
+		// tslint:disable-next-line:no-let
+		let evictedPeer: CustomPeerInfo | undefined;
+
+		[...this.peerMap.values()].forEach(peersMap => {
+			[...peersMap.keys()].forEach(peerId => {
```
jondubois: As before, `this.peerMap` and `peersMap` can be confused as being the same.

elements/lisk-p2p/src/peer_directory/base_list.ts
```diff
@@ -0,0 +1,213 @@
...
+	readonly dateAdded: Date;
+}
+
+export interface AddPeerOutcome {
+	readonly success: boolean;
+	readonly isEvicted: boolean;
+	readonly evictedPeer?: P2PPeerInfo;
+}
+// Base peer list class is covering a basic peer list that has all the functionality to handle buckets with default eviction strategy
+export class BasePeerList {
```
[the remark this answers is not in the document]
jondubois: Also the file names should be more related to the class names if possible.

PR #4161, opened 2019-08-27
Dpos.getRoundDelegates returns outdated list - Closes #4147

### LINKED ISSUE #4147: Dpos.getRoundDelegates returns outdated list
### Expected behavior
I expect to get the correct list of delegates when calling `Dpos.getRoundDelegates` after a new round is reverted and previous round changes the delegate list.

### Actual behavior
It always uses the cached list ignoring possible changes on the list when a block is reverted.

### Steps to reproduce

1. Call `Dpos.getRoundDelegates(R2)` where `R2` is the beginning of a new round
2. Revert the last blocks so that we return to the previous round (`R1`)
3. Unvote an active delegate and vote for a new one during the previous round (`R1`)
4. Call `Dpos.getRoundDelegates(R2)` again in the beginning of the next round (`R2`)

Because `Dpos.getRoundDelegates` caches every new list and never clears it, it returns an outdated list ignoring the changes made after the last block of the round is reverted.

Step 4 should return a new delegate list including the new voted delegate and excluding the unvoted delegate.

yatki:
### What was the problem?
Dpos.getRoundDelegates was returning an outdated delegate list after reverting the last block of previous round.
### How did I solve it?
I made sure on undo operations we are invalidating the cache for upcoming rounds.
### Review checklist

- [x] The PR resolves #4147 
- [x] Relevant integration / functional tests are added - Will be done in https://github.com/LiskHQ/lisk-sdk/issues/4192


lsilvs:
`Dpos.getRoundDelegates` unit tests are missing. Can you add them, bitte?

framework/src/modules/chain/dpos/delegates_list.js
```diff
@@ -48,6 +48,12 @@ class DelegatesList extends EventEmitter {
 		this.exceptions = exceptions;
 	}
 
+	deleteDelegateListFromCache(round) {
+		if (this.delegateListCache[round]) {
+			delete this.delegateListCache[round];
```
lsilvs: I reckon it would be safer to clear everything after `round`

lsilvs: Also I think we need to clear `RoundDelegates` entry in the database?

framework/src/modules/chain/components/storage/entities/round_delegates.js
```diff
@@ -71,6 +71,10 @@ class RoundDelegates extends BaseEntity {
 		this.SQLs = this.loadSQLFiles('dpos', sqlFiles, this.sqlDirectory);
 	}
 
+	/**
+	 * @returns {string[]} delegatePublicKeys
+	 */
+
```
michielmulders: Minor thingy:
- Remove empty line between description and function declaration
- This code has been there, but it is very hard to know for someone reading the code what `result[0]` means? Please add some clarification if possible.

framework/src/modules/chain/dpos/delegates_info.js
```diff
@@ -85,6 +85,16 @@ class DelegatesInfo {
 
 		// Perform updates that only happens in the end of the round
 		if (this._isLastBlockOfTheRound(block)) {
+			if (undo) {
```
michielmulders: Can you also remove the `@todo` mention and add a small description for the `_update` function as it's a generic name. (out of scope+-)

framework/src/modules/chain/dpos/delegates_list.js
```diff
@@ -88,11 +88,30 @@ class DelegatesList extends EventEmitter {
 	}
 
 	async deleteDelegateListUntilRound(round) {
+		this.deleteDelegateListsFromCache(r => r < round);
```
michielmulders: Can you pick a more meaningful name for `r`? :)
Something like `roundFromCache`?

framework/src/modules/chain/dpos/delegates_list.js
```diff
@@ -88,11 +88,30 @@ class DelegatesList extends EventEmitter {
...
 		await this.storage.entities.RoundDelegates.delete({
 			round_lt: round,
 		});
 	}
 
+	async deleteDelegateListAfterRound(round, tx) {
+		this.deleteDelegateListsFromCache(r => r > round);
```
michielmulders: Same here, would prefer a more descriptive name

framework/src/modules/chain/dpos/delegates_list.js
```diff
@@ -88,11 +88,30 @@ class DelegatesList extends EventEmitter {
...
+			{
+				round_gt: round,
+			},
+			tx,
+		);
+	}
+
+	deleteDelegateListsFromCache(filterCb) {
+		Object.keys(this.delegateListCache)
+			.filter(filterCb)
```
michielmulders: Is this a filter function? Mabye `filterFn`?

framework/src/modules/chain/dpos/dpos.js
```diff
@@ -24,7 +24,7 @@ module.exports = class Dpos {
 		this.delegatesList = new DelegatesList({
 			storage,
 			logger,
-			slots,
+      slots,
```
michielmulders: formatting issue

michielmulders: Fixed!

framework/src/modules/chain/dpos/delegates_list.js
yatki: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter#Syntax in the api description they called it callback so did i :D

michielmulders: I hate any reference to the callback era :D I saw some other implementations where they call it `lambdaFn` `fn` `filterFn`, feel free to pick what feels appropriate or use the standard by Mozilla :)

framework/src/modules/chain/processor/processor.js
```diff
@@ -157,8 +157,8 @@ class Processor {
 
 	async create(values) {
 		this.logger.debug({ data: values }, 'creating block');
-		const heghestVersion = Math.max.apply(null, Object.keys(this.processors));
-		const processor = this.processors[heghestVersion];
+		const highestVersion = Math.max.apply(null, Object.keys(this.processors));
```
shuse2: 🙇

framework/test/jest/unit/specs/modules/chain/dpos/getRoundDelegates/utils.js
```diff
@@ -0,0 +1,24 @@
+const { hash } = require('@liskhq/lisk-cryptography');
```
shuse2: please add the license header

framework/test/jest/unit/specs/modules/chain/dpos/getRoundDelegates/getRoundDelegates.spec.js
```diff
@@ -0,0 +1,128 @@
+const { when } = require('jest-when');
```
shuse2: please add the license header

framework/src/modules/chain/components/storage/entities/round_delegates.js
```diff
@@ -120,7 +128,7 @@ class RoundDelegates extends BaseEntity {
 	 * @param {Object} [tx]
 	 * @returns {Promise.<boolean, Error>}
 	 */
-	delete(filters, _options, tx = null) {
+	delete(filters, tx = null) {
```
SargeKhan: Why we remove the options parameter? Maybe it makes sense to follow the consistent interface for the entities.

framework/src/modules/chain/dpos/delegates_info.js
```diff
@@ -76,7 +76,11 @@ class DelegatesInfo {
...
 	 */
 	async _update(block, undo, tx) {
-		// @todo add proper description
+		/**
+		 * If the block height is 1, that means the block is
+		 * the genesis block, in that case we don't have to
+		 * update anything in the accounts.
+		 */
 		if (block.height === 1) {
 			return false;
```
SargeKhan: what does returning false/true signify here? Also, out of scope, but don't we need to update the `_updateVotedDelegatesVoteWeight` for geneisBlock?

framework/src/modules/chain/dpos/delegates_list.js
```diff
@@ -88,11 +88,30 @@ class DelegatesList extends EventEmitter {
...
+		await this.storage.entities.RoundDelegates.delete(
+			{
+				round_gt: round,
+			},
+			tx,
+		);
+	}
+
+	deleteDelegateListsFromCache(filterFn) {
+		Object.keys(this.delegateListCache)
```
SargeKhan: Probably makes more sense to do to more functionaly. Something like:
```
this.delegateListCache = Object.keys(this.delegateListCache).filter(filterFn)
    .reduce((newDelegateListCache, roundFromCacheToKeep) => { 
        newDelegateListCache[roundFromCacheToKeep] = this.delegateListCahe[roundFromCacheToKeep];
       return newDelegateListCache;
    }, {})

File: framework/src/modules/chain/dpos/delegates_info.js [hunk 14]
```diff
@@ -85,6 +89,16 @@ class DelegatesInfo {
 
 		// Perform updates that only happens in the end of the round
 		if (this._isLastBlockOfTheRound(block)) {
+			if (undo) {
```
SargeKhan: Does it maybe make more sense to have another function. Something like: `updateDelegateListAtTheEndOfRound`. This function, when called with undo can delete the delegate list, otherwise, in case of apply, it will create the delegate list for the new round.

framework/test/jest/unit/specs/modules/chain/dpos/getRoundDelegates/getRoundDelegates.spec.js
```diff
@@ -0,0 +1,128 @@
...
+				.calledWith(roundNo)
+				.mockResolvedValue([]);
+			stubs.storage.entities.Account.get.mockResolvedValue(delegateAccounts);
+			shuffledList = shuffleDelegateListForRound(roundNo, delegatePublicKeys);
+
+			// Act
+			list = await dpos.getRoundDelegates(roundNo);
+		});
+
+		it('should return shuffled delegate list by using delegate accounts', () => {
```
SargeKhan: Imagine the current round is 2 and we have the entry for the current round in the round_delegates table. If getRoundDelegates function is called with round number 4 or 5, what will happen? Will it throw an error or will be generate invalid delegateList for the round in the future?

framework/test/jest/unit/specs/modules/chain/dpos/getRoundDelegates/utils.js
```diff
@@ -0,0 +1,24 @@
...
+
+const shuffleDelegateListForRound = (round, list) => {
```
lsilvs: Shouldn't we avoid having implementation to test implementation? Maybe here we should use a known list instead of re-implementing the `shuffleDelegateListForRound`

framework/src/modules/chain/components/storage/entities/round_delegates.js
yatki: But it's not used anywhere in the method? :D

framework/src/modules/chain/dpos/delegates_info.js
yatki: Nothing actually. It's just the termination of the method. 

Based on the previous rounds implementation it seems we were not doing it. But if there is a case that genesisBlock can contain vote transactions from a delegate to someone else maybe we should do it? @shuse2 ?

framework/src/modules/chain/dpos/delegates_list.js
yatki: But here, the filtered rounds have to be deleted. This function is keeping everything. Also, we'll get rid of cache mechanism #4217. So this code gonna be removed very very soon :D

framework/test/jest/unit/specs/modules/chain/dpos/getRoundDelegates/utils.js
yatki: delegateList changes by roundNumber and to sake of avoiding having hardcoded values in tests I'm generating round number randomly. Thus, I wanted to have the shuffle logic as utils. Also I wanted to avoid using original implementation just in case something breaks there. 

I think this is a better way to go and maybe mowing shuffle logic to elements is the way to go?

framework/src/modules/chain/components/storage/entities/round_delegates.js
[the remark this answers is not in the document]
yatki: As i said earlier, it wasn't being used. If you and @SargeKhan didn't like this, I can revert it instead of modifying other functions in this pr.

framework/src/modules/chain/dpos/delegates_info.js
yatki: I agree with this also your other concern if we provide a future round number to getRoundDelegates method make sense. I'll create a separate issue to tackle that.https://github.com/LiskHQ/lisk-sdk/issues/4218

framework/test/jest/unit/specs/modules/chain/dpos/getRoundDelegates/getRoundDelegates.spec.js
yatki: https://github.com/LiskHQ/lisk-sdk/issues/4218

framework/src/modules/chain/dpos/delegates_list.js
SargeKhan: We are reassigning the variable: `this.delegateListCache` in the code I shared above. The value of this object will not contain the delegateList for rounds which fail the filter condition. So, it's irrelevant if we delete keys from the old delegateListCache object. Right? 
Also, this is not refactoring of some old code. I am suggesting changes in the code which was added as part of this PR. So, doesn't it make sense to add code which is cleaner?

framework/src/modules/chain/dpos/delegates_info.js
SargeKhan: GenesisBlock contains delegate registeration as well as vote transactions. I wonder how it was syncing with the testnet. @lsilvs what do you think?

framework/src/modules/chain/dpos/delegates_list.js
yatki: @SargeKhan filter conditions that were used in the other functions were returning only the rounds to be deleted. That's what i meant so reduce only iterates on the rounds to be deleted in that case. But we can negate those conditions. That's not a problem.

What I mean is, the whole caching logic will be deleted, so this function will be deleted completely. So there is no need to refactor it to be functional.

framework/src/modules/chain/dpos/delegates_info.js
shuse2: dpos module only do the apply for `reward` and `fee`, which does not exists on the genesis block now.
Therefore, in our case it should work now.

framework/test/jest/unit/specs/modules/chain/dpos/getRoundDelegates/utils.js
shuse2: This shuffle logic is essential part of the DPOS module. I don't think it make sense to move to any other parts.
I also think instead of generating the implementation by implementation, it's better to use fixture in this case
Since it's very hard to test, we should add specific test for this function (although it's not public function) with fixture.

framework/test/jest/unit/specs/modules/chain/dpos/getRoundDelegates/getRoundDelegates.spec.js
```diff
@@ -0,0 +1,144 @@
...
+			},
+		};
+
+		dpos = new Dpos({
+			...stubs,
+			activeDelegates: constants.ACTIVE_DELEGATES,
+		});
+	});
+
+	describe('When non-shuffled delegate public keys for the round IS in the cache', () => {
```
shuse2: Probably it's better to add a case where it should fail if table and cache doesn't exist but in the middle of the round?

framework/src/modules/chain/dpos/delegates_info.js
lsilvs: Among all updates performed during `Dpos.apply`, only `_updateProducedBlocks` would be relevant for genesis block because genesis block is not the last block of the round. As the genesis block forger is not even a delegate account, we don't update its `producedBlocks` column.

PS: Note that `_updateVotedDelegatesVoteWeight` is not related to vote transactions included in the block. It's related to the accounts that the block forger voted for.

framework/test/jest/unit/specs/modules/chain/dpos/getRoundDelegates/getRoundDelegates.spec.js
shuse2: Since the implementation is changed, agreed to test it on the coming PR

framework/test/jest/unit/specs/modules/chain/dpos/undo.spec.js
```diff
@@ -95,6 +96,22 @@ describe('dpos.undo()', () => {
 	});
 
 	describe('Given block is NOT the last block of the round', () => {
+		it('should delete delegate list for rounds which are after the current round', async () => {
```
lsilvs:
```suggestion
		it('should NOT delete delegate list for rounds which are after the current round', async () => {
```

PR #4165, opened 2019-08-27
Fix the number accuracy of vote calculations in round - Closes #2423

### LINKED ISSUE #2423: Fix the number accuracy of vote calculations in round
### Expected behavior
The numbers calculations during the round tick should be done accurately with the help of Bignumber. 

### Actual behavior
It is identified that if `amount` of the voter cross the `Number.MAX_SAFE_INTEGER` then calculations are not done accurately, and we get minor differences in votes.  

https://github.com/LiskHQ/lisk/blob/3050d2845ff64e570d873281ccfc90fed30b27d1/logic/round.js#L151-L153

This difference can accumulate up to effecting the order in the delegate list. 

### Steps to reproduce
1. Have an account which have a balance more than `Number.MAX_SAFE_INTEGER`
2. Vote from that account to a delegate 
3. Notice the difference of `delegate.vote`, it will not match the `amount` you voted with. 

`> 1.0.0 `

**NOTES GROOMING MEETING**
- Notes: at logic/round.js:145 updateVotes()
- This happened on testnet already. For mainnet nobody has this amount of Lisk tokens.
- In genesis, we do a vote transaction from genesis account that holds 100 million to all delegates.
- At end of the round, we do vote weight summation.
- As it is an old ticket, we need to investigate if this is still reproducible in the current code (DPOS module).

nazarhussain:
### What was the problem?
There was issue in calculating vote weight in old rounds module. If the vote value is > Number.MAX_SAFE_INTEGER. 

### How did I solve it?
Since we rewrite the whole rounds module to dpos, so just adding a unit test to verify the behaviour in new module. 

### Review checklist

- [x] The PR resolves #2423

### DIFF
--- a/framework/test/jest/unit/specs/modules/chain/dpos/apply.spec.js
+++ b/framework/test/jest/unit/specs/modules/chain/dpos/apply.spec.js
@@ -434,6 +434,63 @@ describe('dpos.apply()', () => {
 			});
 		});
 
+		// Reference: https://github.com/LiskHQ/lisk-sdk/issues/2423
+		describe('When summarizing round return value which is greater than Number.MAX_SAFE_INTEGER ', () => {
+			beforeEach(async () => {
+				feePerDelegate = new BigNum(Number.MAX_SAFE_INTEGER.toString()).add(
+					randomInt(10, 1000),
+				);
+				totalFee = new BigNum(feePerDelegate).mul(constants.ACTIVE_DELEGATES);
+
+				rewardPerDelegate = new BigNum(Number.MAX_SAFE_INTEGER.toString()).add(
+					randomInt(10, 1000),
+				);
+
+				stubs.storage.entities.Round.summedRound.mockResolvedValue([
+					{
+						fees: totalFee.toString(), // dividable to ACTIVE_DELEGATE count
+						rewards: delegatesWhoForged.map(() => rewardPerDelegate),
+						delegates: delegatesWhoForged.map(account => account.publicKey),
+					},
+				]);
+
+				getTotalEarningsOfDelegate = account => {
+					const blockCount = delegatesWhoForged.filter(
+						d => d.publicKey === account.publicKey,
+					).length;
+					const reward = new BigNum(rewardPerDelegate).mul(blockCount);
+					const fee = new BigNum(feePerDelegate).mul(blockCount);
+					return {
+						reward,
+						fee,
+					};
+				};
+			});
+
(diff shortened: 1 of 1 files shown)


framework/test/jest/unit/specs/modules/chain/dpos/apply.spec.js
```diff
@@ -434,6 +434,63 @@ describe('dpos.apply()', () => {
 			});
 		});
 
+		// Reference: https://github.com/LiskHQ/lisk-sdk/issues/2423
+		describe('When summarizing round return value which is greater than Number.MAX_SAFE_INTEGER ', () => {
+			beforeEach(async () => {
+				feePerDelegate = new BigNum(Number.MAX_SAFE_INTEGER.toString()).add(
+					randomInt(10, 1000),
```
shuse2: I think this shouldn't be randomize. in some cases, it will be the same as using basic number and in some case it will be different

framework/test/jest/unit/specs/modules/chain/dpos/apply.spec.js
```diff
@@ -434,6 +434,63 @@ describe('dpos.apply()', () => {
...
+				);
+				totalFee = new BigNum(feePerDelegate).mul(constants.ACTIVE_DELEGATES);
+
+				rewardPerDelegate = new BigNum(Number.MAX_SAFE_INTEGER.toString()).add(
+					randomInt(10, 1000),
```
shuse2: same as above

framework/test/jest/unit/specs/modules/chain/dpos/apply.spec.js
nazarhussain: Just tried to have a number larger than `Number.MAX_SAFE_INTEGER`. And instead of a fixed number tried to randomize it. I personally experience having random fixtures may catch some issues which you miss with a single static value.

shuse2: I think the test should be always deterministic rather than hoping some random number can catch the bug.

nazarhussain: Yes tests are always deterministic, but using a known range of input is a good practice compared to a static value. That's what we are doing everywhere in the tests, e.g. look at any fixtures files. Also this `randomInt` is already used inside the same file, its not thing new I introduced here.

PR #4166, opened 2019-08-28
Protocol-spec delegate registration - Closes #4159

### LINKED ISSUE #4159: Protocol-spec delegate registration
### Description

- Basic delegate registration acceptance
- Delegate registration for delegate who already exists
- Delegate registration with existing username
- Delegate registration with forbidden name

### Acceptance Criteria

JSON generator for: 

- [x] Basic delegate registration acceptance
- [x] Delegate registration for user who already exists
- [x] Delegate registration with existing username
- [x] Delegate registration with forbidden name

### Additional Information

Split from https://github.com/LiskHQ/lisk-sdk/issues/4087

pablitovicente:
### What was the problem?

There were no protocol spec generator and JSON for delegate registration

### How did I solve it?

By adding the generator and generating the JSON for the delegate registration spec

### Review checklist

- [x] The PR resolves #4159

### DIFF
--- a/protocol-specs/generators/block_processing_delegate/index.js
+++ b/protocol-specs/generators/block_processing_delegate/index.js
@@ -102,7 +102,7 @@ const accounts = {
 	},
 };
 
-const generateTestCasesValidBlockSecondSignatureTx = () => {
+const generateTestCasesValidBlockDelegateRegistration = () => {
 	const chainSimulator = new ChainStateSimulator(
 		genesisBlock,
 		initialAccountsState,
@@ -141,15 +141,69 @@ const generateTestCasesValidBlockSecondSignatureTx = () => {
 	};
 };
 
-const validBlockWithSecondSignatureTxSuite = () => ({
+const generateTestCasesInvalidBlockDelegateRegistrationSecondTime = () => {
+	const chainSimulator = new ChainStateSimulator(
+		genesisBlock,
+		initialAccountsState,
+		accounts,
+	);
+
+	chainSimulator
+		.transfer('50')
+		.from('16313739661670634666L')
+		.to('10881167371402274308L')
+		.forge();
+
+	chainSimulator
+		.transfer('30')
+		.from('10881167371402274308L')
+		.to('2222471382442610527L')
+		.forge();
+
+	chainSimulator
+		.registerDelegate('RadioHead')
+		.for('2222471382442610527L')
+		.forge();
+
+	chainSimulator
+		.registerDelegate('RadioHead')
+		.for('2222471382442610527L')
+		.forge(true);
+
+	const chainAndAccountStates = chainSimulator.getScenario();
+
+	return {
+		initialState: {
+			chain: chainAndAccountStates.chain,
+			accounts: chainAndAccountStates.initialAccountsState,
+		},
+		input: chainAndAccountStates.chain.slice(2),
+		output: {
+			chain: chainAndAccountStates.chain,
+			accounts: chainAndAccountStates.finalAccountsState,
+		},
+	};
+};
+
+const validBlockWithDelegateRegistrationSuite = () => ({
 	title: 'Valid block processing',
 	summary: 'A valid block with a delegate registration',
 	config: 'mainnet',
 	runner: 'block_processing_delegate',
 	handler: 'valid_block_processing_delegate_registration_tx',
-	testCases: generateTestCasesValidBlockSecondSignatureTx(),
+	testCases: generateTestCasesValidBlockDelegateRegistration(),
+});
+
+const invalidBlockWithSecondDelegateRegistrationSuite = () => ({
+	title: 'Invalid block processing',
+	summary: 'An invalid block with a second delegate registration',
+	config: 'mainnet',
+	runner: 'block_processing_delegate',
+	handler: 'invalid_block_processing_second_delegate_registration_tx',
+	testCases: generateTestCasesInvalidBlockDelegateRegistrationSecondTime(),
 });
 
 module.exports = BaseGenerator.runGenerator('block_processing_delegate', [
-	validBlockWithSecondSignatureTxSuite,
+	validBlockWithDelegateRegistrationSuite,
+	invalidBlockWithSecondDelegateRegistrationSuite,
 ]);
--- a/protocol-specs/generators/block_processing_delegate/index.js
+++ b/protocol-specs/generators/block_processing_delegate/index.js
@@ -165,11 +165,10 @@ const generateTestCasesInvalidBlockDelegateRegistrationSecondTime = () => {
 		.for('2222471382442610527L')
 		.forge();
 
-	const invalidBlock = true;
 	chainStateBuilder
 		.registerDelegate('RadioHead')
 		.for('2222471382442610527L')
-		.forge(invalidBlock);
+		.forgeInvalidInputBlock();
 
 	const chainAndAccountStates = chainStateBuilder.getScenario();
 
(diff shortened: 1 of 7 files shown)


protocol-specs/generators/block_processing_delegate/index.js
```diff
@@ -0,0 +1,155 @@
...
+const validBlockWithSecondSignatureTxSuite = () => ({
+	title: 'Valid block processing',
+	summary: 'A valid block with a delegate registration',
+	config: 'mainnet',
+	runner: 'block_processing_delegate',
+	handler: 'valid_block_processing_delegate_registration_tx',
+	testCases: generateTestCasesValidBlockSecondSignatureTx(),
+});
+
+module.exports = BaseGenerator.runGenerator('block_processing_delegate', [
```
michielmulders: Not exactly related, but looking at the base_generator code, the `generatorName` is unused.

protocol-specs/generators/block_processing_delegate/index.js
```diff
@@ -0,0 +1,155 @@
...
+		},
+		input: chainAndAccountStates.chain.slice(2),
+		output: {
+			chain: chainAndAccountStates.chain,
+			accounts: chainAndAccountStates.finalAccountsState,
+		},
+	};
+};
+
+const validBlockWithSecondSignatureTxSuite = () => ({
```
michielmulders: Would be nice to use a generic function in `base_generator` that replaces the custom method to create the suite data.

```js
const addressFromPubKeySuite = () => ({
	title: 'Address generation',
	summary: 'Address generation from a public key',
	config: 'mainnet',
	runner: 'address_generation',
	handler: 'address_from_pub_key',
	testCases: [generateTestCasesForAddressFromPubKey()],
});
```

into

```js
BaseGenerator.runGenerator(
	'address_generation',
	[
		suiteGenerator({
			title: 'xxx',
			summary: 'yyy',
			config: 'mainnet',
			runner: 'address_generation',
			handler: 'address_from_pub_key',
		}, [generateTestCasesForAddressFromPubKey()
	])
]);
```

michielmulders: This maybe does not look great tho, but it prevents from creating a method that does more or less the same for every generator.

protocol-specs/utils/chain_state_simulator.js
```diff
@@ -0,0 +1,219 @@
...
+		this.fixedPoint = 10 ** 8;
+	}
+
+	transfer(amount) {
+		return {
+			from: addressFrom => ({
+				to: addressTo => {
+					const amountBedows = `${amount * this.fixedPoint}`;
+					const transferTx = new TransferTransaction(
+						transfer({
```
michielmulders: Confusing with the transfer function that surrounds this.
Maybe import `transfer` from `lisk-txs` as `transferLisk`

protocol-specs/utils/chain_state_simulator.js
```diff
@@ -0,0 +1,219 @@
...
+			initialAccountStore: cloneDeep(initialAccountsStates),
+			pendingTransactions: [],
+			appliedTransactions: [],
+		};
+		this.round = 0;
+		this.slot = 0;
+		this.fixedPoint = 10 ** 8;
+	}
+
+	transfer(amount) {
```
michielmulders: Creds for the nice code!

protocol-specs/generators/block_processing_delegate/index.js
pablitovicente: Yes I did notice this but I prefer to leave that for another issue.

protocol-specs/generators/block_processing_delegate/index.js
pablitovicente: Yeah there's room for improvement for the original implementation. Would you create an issue including the previous comment related to the protocol-specs?

protocol-specs/utils/chain_state_simulator.js
pablitovicente: Changed to `transferLisk `

protocol-specs/generators/block_processing_delegate/index.js
```diff
@@ -0,0 +1,210 @@
...
+		genesisBlock,
+		initialAccountsState,
+		accounts,
+	);
+
+	chainStateBuilder
+		.transfer('50')
+		.from('16313739661670634666L')
+		.to('10881167371402274308L')
+		.forge();
```
shuse2: `to` maybe only relevant to transfer transaction or maybe there are other usecase in mind?
to be easily combine multiple transactions, make it make sense to do like
```
chainStateBuilder
  .transfer('16313739661670634666L', '50', '10881167371402274308L')
  .registerSecondSignature('16313739661670634666L', 'passphrase')
  .forge()
```

limiaspasdaniel: This is opinionated but I feel that the way Pablo implemented it feels more natural. If doing it the way you propose, you need to know the order of the parameters. Also reading `transfer` `from` `to` resembles the way you say it in plain English.

pablitovicente: The idea is to experiment how this will work. One of the points is what Dani says but also with the current implementation we don't need any validation or schema as there's only one way to call the methods. I would like for us to give it a try to this to check how it feels.

protocol-specs/generators/block_processing_delegate/index.js
```diff
@@ -0,0 +1,258 @@
...
+		.to('2222471382442610527L')
+		.forge();
+
+	chainStateBuilder
+		.registerDelegate('RadioHead')
+		.for('2222471382442610527L')
+		.forge();
+
+	chainStateBuilder
+		.registerDelegate('RadioHead')
```
limiaspasdaniel: Hahahaha good one

protocol-specs/utils/chain_state_builder.js
```diff
@@ -0,0 +1,282 @@
...
+			vote: this.fixedPoint * 1,
+			multisignature: this.fixedPoint * 5,
+		};
+	}
+
+	transfer(amount) {
+		return {
+			from: addressFrom => ({
+				to: addressTo => {
+					const amountBedows = `${amount * this.fixedPoint}`;
```
limiaspasdaniel:
```suggestion
					const amountBeddows = `${amount * this.fixedPoint}`;
```

protocol-specs/utils/chain_state_builder.js
```diff
@@ -0,0 +1,282 @@
...
+
+					const transferTx = new TransferTransaction(
+						transferLisk({
+							amount: amountBedows,
```
limiaspasdaniel:
```suggestion
							amount: amountBeddows,
```

protocol-specs/utils/chain_state_builder.js
```diff
@@ -0,0 +1,282 @@
...
+	updateAccountBalancesAfterTransfer(from, to, amount) {
+		const newAccountStoreState = cloneDeep(
+			this.state.accountStore.slice(-1)[0],
+		);
+		const sender = this.findAccountByAddress(from, newAccountStoreState);
+		const recipient = this.findAccountByAddress(to, newAccountStoreState);
+
+		if (!sender) {
+			throw new Error(
+				'Sender does not exists so it would not be possible to transfer form this account. Check the values passed to the constructor',
```
limiaspasdaniel:
```suggestion
				'Sender does not exists so it would not be possible to transfer from this account. Check the values passed to the constructor',
```

protocol-specs/utils/chain_state_builder.js
```diff
@@ -0,0 +1,282 @@
...
+
+		if (!sender && !recipient) {
+			throw new Error(
+				'Both sender and recipient were not found in the account store state. This means that something is wrong with the values passed to the constructor.',
+			);
+		}
+
+		// Update sender balance
+		sender.balance = parseInt(
+			new BigNum(sender.balance.toString())
```
shuse2: you should not `parseInt` for the balance, and it should be string, and same for other cases which uses BigNum

pablitovicente: removed usage of `parseInt`

Issue #3859, opened 2019-06-25
Fix cyclomatic complexity in P2P constructor


mitsujutsu:
### Expected behavior
[Reference comment thread](https://github.com/LiskHQ/lisk-sdk/pull/3824#discussion_r297162071)

There should be no need to disable tslint warning for cyclomatic complexity.

### Actual behavior

In feature branch for p2p, we disable tslint warning.

PR #4169, opened 2019-08-29
Allow inbound connections and get status from outbound

ishantiw:
### What was the problem?

SDK Version 2.3 nodes are disconnecting constantly from older versions because the way it makes outbound and inbound connections. It doesn't allow both inbound and outbound connections with a specific peer whereas older versions want to make both outbound/inbound connections when you connect with them. 

Also, older versions don't send `updateMyself` to their inbounds so the new versions connected to them don't get any update if they don't have inbound with them

### How did I solve it?

1) Create a periodic job to ask for status voluntarily from the outbound connections
2) Allow inbound connection from older nodes even if we have outbound connection with them.

### How to manually test it?

Connect to the network and check the consensus and connected nodes

### Review checklist

- [ ] The PR resolves #INSERT_ISSUE_NUMBER


elements/lisk-p2p/src/p2p.ts
```diff
@@ -727,6 +732,15 @@ export class P2P extends EventEmitter {
 				}
 
 				const existingPeer = this._peerPool.getPeer(peerId);
+				// Allow connections with lower version to have incoming connection even if it has outbound
+				if (
+					existingPeer &&
```
mitsujutsu: can we also use `getUniqueOutboundConnections` and check it?

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -509,6 +521,7 @@ export class PeerPool extends EventEmitter {
 		// Clear periodic eviction of outbound peers for shuffling
 		if (this._outboundShuffleIntervalId) {
 			clearInterval(this._outboundShuffleIntervalId);
+			clearInterval(this._outboundUpdateStatusId);
```
diego-G: What is the reason to update the status ID at this point?

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -682,6 +702,22 @@ export class PeerPool extends EventEmitter {
 		}
 	}
 
+	private async _updateOutboundConnections(): Promise<void> {
+		try {
+			await Promise.all(
+				this.getConnectedPeers(OutboundPeer)
+					.filter(
```
mitsujutsu: do we need this filter if it's already `getConnectedPeers(OutboundPeer)`

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -443,9 +453,11 @@ export class PeerPool extends EventEmitter {
 
 		// Throw an error because adding a peer multiple times is a common developer error which is very difficult to identify and debug.
 		if (this._peerMap.has(peer.id)) {
-			throw new Error(`Peer ${peer.id} was already in the peer pool`);
+			this._peerMapDuplicates.set(peer.id, peer);
```
diego-G: Why don't we remove the peer straight away in here? I don't think we need the `_peerMapDuplicates` list. Could you explain why we need it?

mitsujutsu: I am not sure if necessary but should check to see if it's also already in `_peerMapDuplicates` and throw if so?

elements/lisk-p2p/src/p2p.ts
ishantiw: we don't need that check because indirectly we are already checking it with `this._peerPool.getPeer(peerId)` if it exists or not

elements/lisk-p2p/src/peer_pool.ts
ishantiw: statusId is created in the constructor for setInterval so that we can clear it when we are stopping the p2p node on cleanup

elements/lisk-p2p/src/peer_pool.ts
ishantiw: If we remove the peer that means we are removing the outbound connection of that peer and then only making inbound that will again create the same problem which we are trying to solve with this fix.

elements/lisk-p2p/src/p2p.ts
```diff
@@ -727,6 +732,15 @@ export class P2P extends EventEmitter {
...
+					isVersionLessThan(incomingPeerInfo.version, this._nodeInfo.version)
+				) {
+					this._peerPool.addInboundPeer(incomingPeerInfo, socket);
```
jondubois: We should check that it's not already in the inbound 'duplicatePeerMap' list or else we can end up with the same peer two or more times in the  'duplicatePeerMap' and this could be a memory leak or attack vector.

ishantiw: Since its setting in a map, if it finds the same key then it will overwrite it with the new value so we should be afraid of duplicates

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -424,6 +434,14 @@ export class PeerPool extends EventEmitter {
 		socket: SCServerSocket,
 	): Peer {
 		const inboundPeers = this.getPeers(InboundPeer);
+		const peerId = constructPeerIdFromPeerInfo(peerInfo);
+
+		if (this._peerMapDuplicates.has(peerId) && this._peerMap.has(peerId)) {
+			throw new Error(
+				`Peer ${peerId} already has inbound and outbound connections.`,
+			);
```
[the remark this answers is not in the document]
ishantiw: This vulnerability was also present in the old version, if we try to add already existing peer in peerMap it will crash. However, it was checked in p2p.js before we call `addIbound` thats why it was not happening. But yes, it will crash the node with this error being thrown so now I will not throw error but will simply emit event and return.

elements/lisk-p2p/src/p2p.ts
ishantiw: I am only creating the peer inbound object when there is only one connection present or none.

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -442,10 +453,13 @@ export class PeerPool extends EventEmitter {
 		const peer = new InboundPeer(peerInfo, socket, peerConfig);
 
 		// Throw an error because adding a peer multiple times is a common developer error which is very difficult to identify and debug.
-		if (this._peerMap.has(peer.id)) {
-			throw new Error(`Peer ${peer.id} was already in the peer pool`);
+		// For older versions allow them to make an inbound connection even if they have an outbound
+		if (this._peerMap.has(peer.id) && !this._peerMapDuplicates.has(peer.id)) {
```
diego-G:
```suggestion
		if (this._peerMap.has(peer.id)) {
		   this._peerMap.set(peer.id, peer);
		   if(!this._peerMapDuplicates.has(peer.id)){
		       this._peerMapDuplicates.set(peer.id, peer);
		   }
              }		
```

elements/lisk-p2p/src/p2p.ts
```diff
@@ -601,6 +608,7 @@ export class P2P extends EventEmitter {
 	private async _startPeerServer(): Promise<void> {
 		this._scServer.on(
 			'connection',
+			// tslint:disable-next-line:cyclomatic-complexity
```
diego-G: To remove via https://github.com/LiskHQ/lisk-sdk/issues/3859

elements/lisk-p2p/src/p2p.ts
```diff
@@ -727,14 +735,37 @@ export class P2P extends EventEmitter {
 				}
 
 				const existingPeer = this._peerPool.getPeer(peerId);
+				const existingDuplicatePeer = this._peerPool.getDuplicatePeer(peerId);
+				// Allow connections with lower version to have incoming connection even if it has outbound
+				// tslint:disable-next-line:no-let
+				let isInboundCreated = false;
+				if (
```
diego-G: I guess we can simplify here merging this condition with the one below. In that case, we might still be able to keep the `cyclomatic-complexity`.

PR #4171, opened 2019-08-30
If an existing peer tries to connect to us, the inbound connection is kept even though they are not added to our peer list - Closes #4170

### LINKED ISSUE #4170: If an existing peer tries to connect to us, the inbound connection is kept even though they are not added to our peer list
### Expected behavior

If a peer tries to connect to us and we already have them as a peer, the new inbound socket should be disconnected. This was identified by @4miners as a cause for low consensus when interacting in a mixed network which has older 2.0 and 1.x nodes.

### Actual behavior

Currently we don't add the peer but we forget to disconnect the socket so the peer thinks that we added them and it can lead to consensus issues with old nodes.

lisk-sdk 2.3

jondubois:
### What was the problem?

In one of the `if` conditions where we refuse to add the inbound peer, we forget to disconnect their socket.

### How did I solve it?

- Disconnect the inbound socket with a relevant code/message.
- The test case related to peer discovery had to be adjusted to account for a small change in discovery time.

### How to manually test it?

A new integration test case was added.

Run `npm test`

### Review checklist

- [ ] The PR resolves #4170

### DIFF
--- a/elements/lisk-p2p/src/disconnect_status_codes.ts
+++ b/elements/lisk-p2p/src/disconnect_status_codes.ts
@@ -35,8 +35,10 @@ export const INCOMPATIBLE_PEER_CODE = 4104;
 export const INCOMPATIBLE_PEER_UNKNOWN_REASON =
 	'Peer is incompatible with the node for unknown reasons';
 
-// First case to follow HTTP status codes
 export const FORBIDDEN_CONNECTION = 4403;
 export const FORBIDDEN_CONNECTION_REASON = 'Peer is not allowed to connect';
+
+export const DUPLICATE_CONNECTION = 4404;
+export const DUPLICATE_CONNECTION_REASON = 'Peer has a duplicate connection';
 
 export const EVICTED_PEER_CODE = 4418;
--- a/elements/lisk-p2p/src/peer_pool.ts
+++ b/elements/lisk-p2p/src/peer_pool.ts
@@ -189,28 +189,15 @@ export class PeerPool extends EventEmitter {
 	private readonly _peerSelectForConnection: P2PPeerSelectionForConnectionFunction;
 	private readonly _sendPeerLimit: number;
 	private readonly _outboundShuffleIntervalId: NodeJS.Timer | undefined;
-	private readonly _inboundPeerConfig: PeerConfig;
-	private readonly _outboundPeerConfig: PeerConfig;
+	private readonly _peerConfig: PeerConfig;
 
 	public constructor(peerPoolConfig: PeerPoolConfig) {
 		super();
 		this._peerMap = new Map();
 		this._peerPoolConfig = peerPoolConfig;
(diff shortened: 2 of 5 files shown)


elements/lisk-p2p/src/disconnect_status_codes.ts
```diff
@@ -35,8 +35,10 @@ export const INCOMPATIBLE_PEER_CODE = 4104;
 export const INCOMPATIBLE_PEER_UNKNOWN_REASON =
 	'Peer is incompatible with the node for unknown reasons';
 
-// First case to follow HTTP status codes
 export const FORBIDDEN_CONNECTION = 4403;
 export const FORBIDDEN_CONNECTION_REASON = 'Peer is not allowed to connect';
 
+export const DUPLICATE_CONNECTION = 4404;
```
diego-G: Errors look too similar. https://github.com/LiskHQ/lisk-sdk/blob/decb81e6489730810a18f7968db424860ebf7c2d/elements/lisk-p2p/src/disconnect_status_codes.ts#L38

jondubois: The first/original connection can be outbound so the suggested name would create confusion because it sounds like there are two duplicate inbound connections which is not the case.

diego-G: Sorry, it wasn't a suggestion. I've just showed a similar variable that we are introducing in another PR https://github.com/LiskHQ/lisk-sdk/pull/4169. Shall we change this variable's name or the other?

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -188,11 +189,37 @@ export class PeerPool extends EventEmitter {
...
 	private readonly _sendPeerLimit: number;
 	private readonly _outboundShuffleIntervalId: NodeJS.Timer | undefined;
+	private readonly _inboundPeerConfig: PeerConfig;
+	private readonly _outboundPeerConfig: PeerConfig;
 
 	public constructor(peerPoolConfig: PeerPoolConfig) {
 		super();
 		this._peerMap = new Map();
 		this._peerPoolConfig = peerPoolConfig;
+		this._inboundPeerConfig = {
```
mitsujutsu: why does `inbound` have no config for `banTime` and `wsMaxPayload` but outbound does

jondubois: `wsMaxPayload` is not necessary for inbound because it is set on the SocketCluster server (not on the socket). It seems that `banTime` is not used in any of the peer classes. I guess we can remove it.

elements/lisk-p2p/src/disconnect_status_codes.ts
jondubois: Ah, I think it may be the same error. We should change it in the other PR so that they match.

PR #4188, opened 2019-09-02
Solve JavaScript heap out of memory - Closes #4183

### LINKED ISSUE #4183: JavaScript heap out of memory error
### Expected behavior
Expect to sync with mainnet

### Actual behavior
It crashes with the following error:

> FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory

### Steps to reproduce
Using `feature/introduce_bft_consensus` branch in Lisk SDK and `release/2.1.0` branch in Lisk Core.
Link SDK in Core: `npm link lisk-sdk`
Start core on mainnet: `LISK_CONSOLE_LOG_LEVEL=debug node dist/index.js -n mainnet | npx bunyan -o short`

michielmulders:
### What was the problem?
It crashes with the following error:

> FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory

### How did I solve it?
Separate both prepare statements, use `for ... of`, and skip caching for genesis block (preventing 8k DB calls -> causing the issue previously).

### How to manually test it?
Start application and see if it runs correctly.

### Review checklist

- [x] The PR resolves #4183 

### DIFF
--- a/framework/src/modules/chain/transactions/transactions_handlers.js
+++ b/framework/src/modules/chain/transactions/transactions_handlers.js
@@ -115,6 +115,7 @@ const applyGenesisTransactions = storage => async (
 		tx,
 	});
 
+	// Avoid merging both prepare statements into one for...of loop as this slows down the call dramatically
 	// eslint-disable-next-line no-restricted-syntax
 	for (const transaction of transactions) {
 		// eslint-disable-next-line no-await-in-loop
--- a/framework/src/modules/chain/transactions/transactions_handlers.js
+++ b/framework/src/modules/chain/transactions/transactions_handlers.js
@@ -115,8 +115,17 @@ const applyGenesisTransactions = storage => async (
 		tx,
 	});
 
-	await Promise.all(transactions.map(t => t.prepare(stateStore)));
-	await Promise.all(transactions.map(t => votesWeight.prepare(stateStore, t)));
+	// eslint-disable-next-line no-restricted-syntax
+	for (const transaction of transactions) {
+		// eslint-disable-next-line no-await-in-loop
+		await transaction.prepare(stateStore);
+	}
+
+	// eslint-disable-next-line no-restricted-syntax
+	for (const transaction of transactions) {
+		// eslint-disable-next-line no-await-in-loop
+		await votesWeight.prepare(stateStore, transaction);
+	}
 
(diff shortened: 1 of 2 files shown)


framework/src/modules/chain/transactions/votes_weight.js
```diff
@@ -224,9 +224,13 @@ const prepare = async (stateStore, transaction) => {
 	];
 
 	const cacheFilter = senderRecipientVotedPublicKeys.map(publicKey => ({
-		address: getAddressFromPublicKey(publicKey),
+		publicKey,
```
shuse2: I think we should use consistently primary key search otherwise we will always need more index

framework/src/modules/chain/transactions/votes_weight.js
```diff
@@ -224,9 +224,13 @@ const prepare = async (stateStore, transaction) => {
...
 	}));
 
+	if (cacheFilter.length === 0) {
```
shuse2: shouldn't this check goes before creating filter?

limiaspasdaniel: @shuse2 what do you mean?

shuse2: I mean maybe we can just check the length of `senderRecipientVotedPublicKeys`?

limiaspasdaniel: True @michielmulders

michielmulders: Yes we can do this as well so we don't have to loop the empty array

framework/src/modules/chain/transactions/transactions_handlers.js
```diff
@@ -115,8 +115,17 @@ const applyGenesisTransactions = storage => async (
 		tx,
 	});
 
-	await Promise.all(transactions.map(t => t.prepare(stateStore)));
-	await Promise.all(transactions.map(t => votesWeight.prepare(stateStore, t)));
+	// eslint-disable-next-line no-restricted-syntax
```
lsilvs: It would be good to have a comment explaining why we can't run in parallel and why we need two for loops here. (`to avoid someone merging this into one for loop in the future`)

Issue #4158, opened 2019-08-27
Replace rounds.backwardTick by dpos.undo


lsilvs:
### Description

Replace `rounds.backwardTick` by `dpos.undo`
### Motivation

Rounds module is being replaced by Dpos module as part of BFT implementation

### Acceptance Criteria

- Every use of `rounds.backwardTick` should be replaced by `dpos.undo`
- Tests should be updated to use `dpos.undo` instead of `rounds.backwardTick`
- Should sync on testnet at least up to height 10000

### Additional Information
--

yatki:
https://github.com/LiskHQ/lisk-sdk/issues/4147#issuecomment-528440010

Issue #3685, opened 2019-05-21
Remove previous Round implementation


yatki:
### Expected behavior
After developing the new Round module, 

1. we won't need following tables:
- mem_rounds
- mem_round_snapshot
- mem_votes_snapshot 

2. Previous Round submodule files & tests.

### Actual behavior
We still have them at the moment.
### Which version(s) does this affect? (Environment, OS, etc...)

**List of Task** 

- [ ] Remove the above-mentioned tables
- [ ] Remove rounds modules files & tests

PR #4191, opened 2019-09-03
Replace rounds.tick by dpos.apply - Closes #4157

### LINKED ISSUE #4157: Replace rounds.tick by dpos.apply
### Description

Replace `rounds.tick` by `dpos.apply`

### Motivation

Rounds module is being replaced by Dpos module as part of BFT implementation

### Acceptance Criteria

- Every use of `rounds.tick` should be replaced by `dpos.apply`
- Tests should be updated to use `dpos.apply` instead of `rounds.tick`
- Should sync on testnet at least up to height 20000

### Additional Information
--

lsilvs:
### What was the problem?
`rounds.tick` still being used in the codebase

### How did I solve it?
By replacing every use of `rounds.tick` by `dpos.apply`.

Because `rounds.backwardTick` is still used (will be replace by https://github.com/LiskHQ/lisk-sdk/issues/4158), some integration tests had to be skipped because they were expecting some side effects from `rounds.tick` that is not used anymore.

### How to manually test it?
Green build
Should sync on testnet at least up to height 20000

### Review checklist

- [ ] The PR resolves #4157


framework/src/modules/chain/blocks/chain.js
```diff
@@ -192,22 +192,19 @@ const applyConfirmedStep = async (storage, slots, block, exceptions, tx) => {
  * @param {function} tx - Database transaction
```
michielmulders: Can you update the params and the description for them? 
The most important one is the `shouldSave` flag

framework/src/modules/chain/blocks/verify.js
```diff
@@ -491,12 +491,7 @@ class BlocksVerify {
 		);
 	}
 
-	async reloadRequired(blocksCount, memRounds) {
```
michielmulders: Can you add some comment here what the goal of this function is?

framework/src/modules/chain/dpos/delegates_info.js
```diff
@@ -162,24 +162,40 @@ class DelegatesInfo {
...
-				const { fee, reward } = earnings;
-				const amount = fee.plus(reward);
-
-				const filters = {
-					publicKey_in: delegateAccount.votedDelegatesPublicKeys,
-				};
-				const field = 'voteWeight';
-				const value = amount.toString();
+			delegatesWithEarnings
+				.filter(({ delegatePublicKey }) => {
```
michielmulders: The complexity here for the `.filter` and `.map` functions are too high. 
I suggest to apply some functional programming here where we extract the logic into private functions, example:

```
async _updateVotedDelegatesVoteWeight(roundSummary, undo, tx) {
    const delegatesWithEarnings = this._getDelegatesWithTheirEarnings(
        roundSummary,
    );

    return Promise.all(
        delegatesWithEarnings
            .filter(({ delegatePublicKey }) => _delegatesWithVotesForDelegates)
            .map(({ delegatePublicKey, earnings }) => _updateDelegateVoteWeight),
    );
}
```

framework/src/modules/chain/dpos/delegates_info.js
```diff
@@ -162,24 +162,40 @@ class DelegatesInfo {
 		);
```
michielmulders: This comment is related to the naming and responsibility of the function itself.

What does `_updateVotedDelegatesVoteWeight` do?
- Find delegates with earnings for this round (a separate function which is good)
- Find delegates that voted for other delegates
- Calculate the fees + reward
- Update the weight for votes

First of all, in my opinion, that's too much responsibility for a private function (bit out of scope, but as we are replacing it, why not think about this)

Next, the naming of the function does not include everything that happens inside. Actually, the last part that happens in the `.map` function is related to the function name `_updateVotedDelegatesVoteWeight` -> Therefore I suggest to split this.

framework/src/modules/chain/dpos/delegates_info.js
```diff
@@ -212,13 +228,20 @@ class DelegatesInfo {
 
 			const delegateAccounts = await this.storage.entities.Account.get(
 				{ publicKey_in: summedRound.delegates },
-				{},
+				{ extended: true },
```
michielmulders: Do we really need `extended` here? :)

framework/src/modules/chain/dpos/delegates_list.js
```diff
@@ -74,18 +74,17 @@ class DelegatesList extends EventEmitter {
...
-			});
-		}
-
-		const { ignoreDelegateListCacheForRounds = [] } = this.exceptions;
 
-		if (!ignoreDelegateListCacheForRounds.includes(round)) {
-			// If the round is not an exception, cache the round.
-			this.delegateListCache[round] = delegatePublicKeys;
+			// If the round is not an exception, create entry in the database and cache the list.
+			if (!ignoreDelegateListCacheForRounds.includes(round)) {
```
michielmulders: Nice workaround! 👍

framework/src/modules/chain/dpos/delegates_info.js
lsilvs: This is out of scope of this PR but a valid concern.

From your comment:
> Find delegates that voted for other delegates

This is just a filter to avoid a `if` inside the `map`.
But I agree that it can be a separated private function. Will update.


> Calculate the fees + reward

It doesn't calculate them.
The calculation happens inside `_getDelegatesWithTheirEarnings`


> that's too much responsibility for a private function

Apart from moving the `filter` to a separated function, what else could be extracted.

framework/src/modules/chain/dpos/delegates_info.js
lsilvs: Yes. `votedDelegatesPublicKeys` is only returned when using `extended: true`

framework/src/modules/chain/blocks/verify.js
lsilvs: I discussed with @shuse2 before and he said we should probably redefine this function to be more useful/meaningful to BFT

framework/src/modules/chain/dpos/delegates_list.js
lsilvs: Just cross my mind that we probably don't need it anymore as the exception is being handled in `verifyBlockForger`.

@shuse2 What you reckon?

framework/src/modules/chain/dpos/delegates_info.js
michielmulders: > Calculate the fees + reward
Could use a wrapper function or pure function to do the summation. I think the final number should be coming in here and we shoudn't be doing any "calculations" inside here anymore (even a simple summation).

> What else could be extracted.
Possibly the content in `.map` function could be also private function that focuses only on changing the vote weight for delegates.

> But I agree that it can be a separated private function. Will update.
Danke!

framework/src/modules/chain/dpos/delegates_list.js
shuse2: yes we shouldnt do this on the generate part

framework/test/jest/unit/specs/modules/chain/dpos/apply.spec.js
```diff
@@ -526,21 +530,22 @@ describe('dpos.apply()', () => {
...
 						},
 						expect.objectContaining(partialData),
+						{},
 						stubs.tx,
 					);
 				});
 			});
 
-			it('should should multiple "totalFee" with "fee_factor" and add "fee_bonus"', async () => {
+			it('should multiple "totalFee" with "fee_factor" and add "fee_bonus"', async () => {
```
yatki: multiply*

framework/src/modules/chain/blocks/chain.js
```diff
@@ -192,22 +192,19 @@ const applyConfirmedStep = async (storage, slots, block, exceptions, tx) => {
...
  * @returns {Promise<reject|resolve>}
  */
-const saveBlockStep = async (storage, roundsModule, block, shouldSave, tx) => {
+const saveBlockStep = async (
+	storage,
+	roundsModule,
```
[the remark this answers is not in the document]
yatki: Ah for backwardTick right?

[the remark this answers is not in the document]
lsilvs: It's not used but it will be removed as part of https://github.com/LiskHQ/lisk-sdk/issues/3685

framework/src/modules/chain/dpos/delegates_info.js
```diff
@@ -138,15 +138,15 @@ class DelegatesInfo {
 				const factor = undo ? -1 : 1;
 				const amount = fee.plus(reward);
 				const data = {
-					...account,
-					balance: account.balance.plus(amount.times(factor)),
-					fees: account.fees.plus(fee.times(factor)),
-					rewards: account.rewards.plus(reward.times(factor)),
+					balance: account.balance.plus(amount.times(factor)).toString(),
```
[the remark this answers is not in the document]
lsilvs: This is `BigNum` that's why we need `toString` before update.

framework/src/modules/chain/rounds/round.js
```diff
@@ -86,7 +86,9 @@ class Round {
 		return new Promise((resolve, reject) => {
 			const data = {
 				publicKey: self.scope.block.generatorPublicKey,
-				producedBlocks: self.scope.backwards ? -1 : 1,
+				// Decrement only for backwards because the forward happens during Dpos.apply
+				// This will be completely removed after Dpos.undo is in place (LiskHQ/lisk-sdk/issues/4158)
+				producedBlocks: self.scope.backwards ? -1 : 0,
```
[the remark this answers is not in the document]
lsilvs: This entire file will be removed as part of https://github.com/LiskHQ/lisk-sdk/issues/3685

framework/src/modules/chain/dpos/delegates_info.js
```diff
@@ -239,16 +230,26 @@ class DelegatesInfo {
...
-					({ publicKey }) => publicKey === item.publicKey,
-				),
-			}));
+			const totalFee = new BigNum(summedRound.fees);
+
+			// Aggregate forger infor into one object
+			const uniqForgersInfo = uniqDelegateListWithRewardsInfo.map(
+				forgerInfo => ({
+					...forgerInfo,
+					earnings: this._calculateRewardAndFeeForDelegate({
```
michielmulders: Looks good, thanks!

Issue #4206, opened 2019-09-05
Transaction count wasn't cached TypeError


ManuGowda:
### Expected behavior
Transaction count should be cached.
### Actual behavior
`
{"name":"lisk-framework","hostname":"ams1-explorer-001.ed148.net","pid":12758,"level":40,"msg":"Transaction count wasn't cached TypeError: Cannot destructure property confirmed of 'undefined' or 'null'.\n    at _getConfirmedTransactionCount (/home/lisk/lisk-main/node_modules/lisk-framework/src/modules/http_api/controllers/node.js:102:26)\n    at process._tickCallback (internal/process/next_tick.js:68:7)","time":"2019-08-29T12:28:07.855Z","v":0}
`
### Steps to reproduce
- Enable cache before running the node
- After enabling cache, make an API call to `api/node/status`
- Observe logs for `Transaction count wasn't cached TypeError`
### Which version(s) does this affect? (Environment, OS, etc...)
2.3.x

PR #4210, opened 2019-09-05
Fix cache logging and destruction - Closes #4198, #4206

### LINKED ISSUE #4198: Improve cache logging
### Expected behavior
Either change the log level to `trace` or remove it completely if its not required.
### Actual behavior
For every single block and transaction, an info log is logged which is reduandant and doesn't add any value.
`
Cache - Keys with patterns: '/api/blocks*,/api/transactions*' cleared from cache on 'chain:blocks:change'
`
### Steps to reproduce
- Enable caching before running the node
- Also send multiple requests to `api/node/status` and other endpoints to see the cache related to logs
2.1.x

shuse2:
### What was the problem?
Caching was returning undefined or null, but it was not handled properly.
Also, cache logs were in `debug` level, but it should've been `trace`

### How did I solve it?
- Change cache related logs to `trace`
- Properly handle result of `cache`

### How to manually test it?
- Call /node/status endpoint with cache enabled
- Set to debug level and cache enabled. It should not log any cache related things

### Review checklist

- [ ] The PR resolves #4198 #4206 

### DIFF
--- a/framework/src/modules/http_api/controllers/node.js
+++ b/framework/src/modules/http_api/controllers/node.js
@@ -121,7 +121,10 @@ async function _getConfirmedTransactionCount() {
 			);
 		} catch (error) {
 			// Ignore error and just put warn
-			library.components.logger.debug(error, "Transaction count wasn't cached");
+			library.components.logger.debug(
+				error,
+				'Failed to cache Transaction count',
+			);
 		}
 	}
 	return confirmed;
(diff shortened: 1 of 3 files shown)


framework/src/modules/http_api/controllers/node.js
```diff
@@ -124,7 +121,7 @@ async function _getConfirmedTransactionCount() {
 			);
 		} catch (error) {
 			// Ignore error and just put warn
-			library.components.logger.warn("Transaction count wasn't cached", error);
+			library.components.logger.debug(error, "Transaction count wasn't cached");
```
ManuGowda: this is invalid log message, it should be `Failed to set transaction count` or something like that

shuse2: I think it is correct as meaning, but since it's the same error message so ill update it

PR #4211, opened 2019-09-05
Fix event handling - Closes #4208

### LINKED ISSUE #4208: Handle transport endpoints from malicious payload
### Expected behavior
- If the payload is invalid, handle the exception and log accordingly and do not let it crash.
- Also, please fix the `Reason: [object Object]`
- Cover all the endpoint with a integration test, with positive and negative scenarios
### Actual behavior
When a user emits a malicious payload to transport endpoint, the application validates and throws an error, and the error not being handled causes the node to crash due to `unhandledRejection`
```
13:19:20.796Z DEBUG lisk-framework:
    Received post block broadcast request in unexpected format { errors:
       [ { keyword: 'type',
           dataPath: '',
           schemaPath: '#/type',
           params: [Object],
           message: 'should be object' } ],
      module: 'transport',
      query:
       '{"block":{"version":1,"reward":200000000,"payloadHash":"e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855","timestamp":103574530,"previousBlock":"13940227048585692600","generatorPublicKey":"eaa049295d96618c51eb30deffe7fc2cc8bfc13190cb97f3b513dd060b000a46","blockSignature":"a41a2786df5b446de04f4d369481f3e126f20e87b50a6557fda3ddb5ccfc2390adfd1015284bceae653660af0dca2fa76bdfbcde97a48ea994923e89ec949400EXTRADATAHERETOCRASHTHENODE","height":9171510,"id":"8390758977495702841","relays":4,"totalAmount":0,"totalFee":0,"numberOfTransactions":0,"payloadLength":0,"transactions":[]},"nonce":"3WRejz1OirEgUb4Z"}' }
13:19:20.803Z FATAL lisk-framework:
    System error: unhandledRejection : { message: '[object Object]',
      stack:
       'Error: [object Object]\n    at Transport.postBlock (/home/lisk/lisk-main/node_modules/lisk-framework/src/modules/chain/transport/transport.js:366:10)\n    at Immediate.channel.subscribe (/home/lisk/lisk-main/node_modules/lisk-framework/src/modules/chain/chain.js:183:22)\n    at runCallback (timers.js:706:11)\n    at tryOnImmediate (timers.js:676:5)\n    at processImmediate (timers.js:658:5)\n    at process.topLevelDomainCallback (domain.js:126:23)' }
13:19:20.807Z  INFO lisk-framework: Cleaning network...
13:19:20.805Z  INFO lisk-framework: Cleaning chain...
13:19:20.805Z DEBUG lisk-framework: Cache - Clean up database
13:19:20.806Z DEBUG lisk-framework: Cache - Quit database
13:19:20.806Z  INFO lisk-framework: Cleaned up successfully
13:19:20.804Z  INFO lisk-framework: Cleanup controller...
13:19:20.804Z ERROR lisk-framework: Reason: [object Object]
```
### Steps to reproduce
- Create socket client and emit malformed payload to `postBlock` endpoint
2.3.x

shuse2:
### What was the problem?
- network:event was subscribed, but it shouldn't throw an error

### How did I solve it?
- wrap with try-catch, and expect the error to happen. Also, it should log.

### How to manually test it?
- Use P2P library and call the endpoint with invalid data

### Review checklist

- [ ] The PR resolves #4208 

### DIFF
--- a/framework/src/modules/chain/chain.js
+++ b/framework/src/modules/chain/chain.js
@@ -170,21 +170,30 @@ module.exports = class Chain {
 
 			// Avoid receiving blocks/transactions from the network during snapshotting process
 			if (!this.options.loading.rebuildUpToRound) {
-				this.channel.subscribe('network:event', ({ data: { event, data } }) => {
-					if (event === 'postTransactions') {
-						this.transport.postTransactions(data);
-						return;
-					}
-					if (event === 'postSignatures') {
-						this.transport.postSignatures(data);
-						return;
-					}
-					if (event === 'postBlock') {
-						this.transport.postBlock(data);
-						// eslint-disable-next-line no-useless-return
-						return;
-					}
-				});
+				this.channel.subscribe(
+					'network:event',
+					async ({ data: { event, data } }) => {
+						try {
+							if (event === 'postTransactions') {
+								await this.transport.postTransactions(data);
+								return;
+							}
+							if (event === 'postSignatures') {
+								await this.transport.postSignatures(data);
+								return;
+							}
+							if (event === 'postBlock') {
+								await this.transport.postBlock(data);
+								return;
+							}
+						} catch (error) {
+							this.logger.info(
+								{ error, event, data },
+								'Received invalid event message',
+							);
+						}
+					},
+				);
 			}
 		} catch (error) {
 			this.logger.fatal('Chain initialization', {
--- a/framework/src/modules/chain/chain.js
+++ b/framework/src/modules/chain/chain.js
@@ -187,8 +187,8 @@ module.exports = class Chain {
 								return;
 							}
 						} catch (error) {
-							this.logger.info(
-								{ error, event, data },
+							this.logger.warn(
+								{ error, event },
 								'Received invalid event message',
 							);
 						}
--- a/framework/src/modules/chain/transport/transport.js
+++ b/framework/src/modules/chain/transport/transport.js
@@ -363,28 +363,16 @@ class Transport {
 					query,
 				},
 			);
-			throw new Error(errors);
+			throw errors;
 		}
 
 		let block;
-		let success = true;
-		try {
-			block = blocksUtils.addBlockProperties(query.block);
+		block = blocksUtils.addBlockProperties(query.block);
 
-			// Instantiate transaction classes
-			block.transactions = this.interfaceAdapters.transactions.fromBlock(block);
+		// Instantiate transaction classes
+		block.transactions = this.interfaceAdapters.transactions.fromBlock(block);
 
-			block = blocksUtils.objectNormalize(block);
-		} catch (e) {
-			success = false;
-			this.logger.debug('Block normalization failed', {
-				err: e.toString(),
-				module: 'transport',
-				block: query.block,
-			});
-
-			// TODO: If there is an error, invoke the applyPenalty action on the Network module once it is implemented.
-		}
+		block = blocksUtils.objectNormalize(block);
 		// TODO: endpoint should be protected before
 		if (this.loaderModule.syncing()) {
 			return this.logger.debug(
@@ -392,10 +380,7 @@ class Transport {
 				block.id,
 			);
 		}
-		if (success) {
-			return this.blocksModule.receiveBlockFromNetwork(block);
-		}
-		return null;
+		return this.blocksModule.receiveBlockFromNetwork(block);
 	}
 
 	/**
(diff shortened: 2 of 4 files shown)


framework/src/modules/chain/chain.js
```diff
@@ -170,21 +170,30 @@ module.exports = class Chain {
...
+							if (event === 'postSignatures') {
+								await this.transport.postSignatures(data);
+								return;
+							}
+							if (event === 'postBlock') {
+								await this.transport.postBlock(data);
+								return;
+							}
+						} catch (error) {
+							this.logger.info(
```
ishantiw: Shouldn't the log level of this error be `debug` or `error`?

ishantiw: maybe we can remove `data` to be logged as it clutters up the logs if someone constantly sends malformed data

shuse2: I thought error will be too much since it's handled properly, but user should know about those event.
Maybe info is too low on this case, so i increased it to warn

framework/src/modules/chain/chain.js
```diff
@@ -170,21 +170,30 @@ module.exports = class Chain {
...
+								return;
+							}
+							if (event === 'postSignatures') {
+								await this.transport.postSignatures(data);
+								return;
+							}
+							if (event === 'postBlock') {
+								await this.transport.postBlock(data);
+								return;
+							}
```
ManuGowda: `return` at the end instead of returning each condition?

framework/src/modules/chain/transport/transport.js
```diff
@@ -363,39 +363,24 @@ class Transport {
...
 			);
-			throw new Error(errors);
+			throw errors;
 		}
 
 		let block;
-		let success = true;
-		try {
-			block = blocksUtils.addBlockProperties(query.block);
+		block = blocksUtils.addBlockProperties(query.block);
```
ManuGowda: why not just `let block = blocksUtils.addBlockProperties(query.block);`?

framework/src/modules/chain/chain.js
shuse2: It should be fine, but I think in this case it's bit more clear it doesn't go to other cases

framework/src/modules/chain/transport/transport.js
shuse2: I would even do the `const block`, but since this is patch case and not to change a lot (because there is alot of those cases in this file), i would prefer not to touch it now

framework/src/modules/chain/chain.js
diego-G: I agree with @ManuGowda. Using `else if` + 1 single return is cleaner.

framework/src/modules/chain/transport/transport.js
```diff
@@ -363,39 +363,24 @@ class Transport {
...
-			success = false;
-			this.logger.debug('Block normalization failed', {
-				err: e.toString(),
-				module: 'transport',
-				block: query.block,
-			});
+		// Instantiate transaction classes
+		block.transactions = this.interfaceAdapters.transactions.fromBlock(block);
 
-			// TODO: If there is an error, invoke the applyPenalty action on the Network module once it is implemented.
```
diego-G: Please, keep this TODO.

framework/test/mocha/functional/ws/transport/transport.js
```diff
@@ -32,7 +32,7 @@ describe('WS transport', () => {
 		let p2p;
 
 		function postTransaction(transactionToPost) {
-			p2p.send({
+			return p2p.send({
```
diego-G: Do we really need this return? I actually don't like the structure of the entire describe. I suggest to clean it:
- Its `beforeEach` is misleading
- Remove the use of done.
- The inner `describe` block is not necessary
- I would remove `postTransaction()` to ease the understanding of every test or move it to a more general handler. If you don't agree, we should follow this pattern later on and do the same thing when sending.

shuse2: This PR is not intended to refactor the test.
All of the functional tests will be recovered in jest, so not needed to rewrite for this patch imo

framework/src/modules/chain/transport/transport.js
shuse2: ah nvm i think it's mistake will fix it

Issue #4214, opened 2019-09-05
Whitelisted peers should not be evicted


mitsujutsu:
### Expected behavior

Whitelisted peers should not be evicted through inbound eviction mechanism.

### Actual behavior

Currently does not check if peer is whitelisted.

### Steps to reproduce

Whitelist a peer and check eviction logs. 

### Which version(s) does this affect? (Environment, OS, etc...)
2.3

Issue #4213, opened 2019-09-05
Shuffling should not evict fixed peer from peer pool


ishantiw:
### Expected behavior

We should not disconnect or evict with a fixed peer even during the shuffling.

### Actual behavior

We also consider evicting fixed peer during shuffling.

### Steps to reproduce

Run and monitor peer connect/disconnect event and check if outbound peer that is in the fixed list is being evicted.

### Which version(s) does this affect? (Environment, OS, etc...)
`2.3`

PR #4216, opened 2019-09-05
Fix white/fixed/blacklist config and manage it in peer pool for connections- Closes#4197#4213#4214

### LINKED ISSUE #4197: Issue with fixed/white peers config
### Expected behavior

1. Adding fixed peers in config file causes the node to crash.
2. When checking if a fixed peer is in peerMap, we should use the id instead of checking by reference because otherwise it will always be false: https://github.com/LiskHQ/lisk-sdk/blob/e542623b28e88f6a16a85971af212285b133f145/elements/lisk-p2p/src/peer_pool.ts#L383
3. Write a unit/integration test to ensure if the config changes work
### Actual behavior

The new fixed peers config doesn't work.
The crash error is:

```
Network initialization { message: 'Cannot read property \'split\' of undefined',
      stack:
       'TypeError: Cannot read property \'split\' of undefined\n    at Object.exports.getIPGroup (/Users/ishantiwari/Documents/lisk-repos/sdk/test-sdk/lisk-sdk/elements/lisk-p2p/dist-node/utils/misc.js:16:29)\n    at Object.exports.isLocal (/Users/ishantiwari/Documents/lisk-repos/sdk/test-sdk/lisk-sdk/elements/lisk-p2p/dist-node/utils/misc.js:49:40)\n    at Object.exports.getNetwork (/Users/ishantiwari/Documents/lisk-repos/sdk/test-sdk/lisk-sdk/elements/lisk-p2p/dist-node/utils/misc.js:51:17)\n    at Object.exports.getNetgroup (/Users/ishantiwari/Documents/lisk-repos/sdk/test-sdk/lisk-sdk/elements/lisk-p2p/dist-node/utils/misc.js:65:29)\n    at new Peer (/Users/ishantiwari/Documents/lisk-repos/sdk/test-sdk/lisk-sdk/elements/lisk-p2p/dist-node/peer/base.js:33:34)\n    at new OutboundPeer (/Users/ishantiwari/Documents/lisk-repos/sdk/test-sdk/lisk-sdk/elements/lisk-p2p/dist-node/peer/outbound.js:11:9)\n    at PeerPool.addOutboundPeer (/Users/ishantiwari/Documents/lisk-repos/sdk/test-sdk/lisk-sdk/elements/lisk-p2p/dist-node/peer_pool.js:209:22)\n    at forEach (/Users/ishantiwari/Documents/lisk-repos/sdk/test-sdk/lisk-sdk/elements/lisk-p2p/dist-node/peer_pool.js:184:24)\n    at Array.forEach (<anonymous>)\n    at PeerPool.triggerNewConnections (/Users/ishantiwari/Documents/lisk-repos/sdk/test-sdk/lisk-sdk/elements/lisk-p2p/dist-node/peer_pool.js:179:56)\n    at P2P._startPopulator (/Users/ishantiwari/Documents/lisk-repos/sdk/test-sdk/lisk-sdk/elements/lisk-p2p/dist-node/p2p.js:394:24)\n    at P2P.start (/Users/ishantiwari/Documents/lisk-repos/sdk/test-sdk/lisk-sdk/elements/lisk-p2p/dist-node/p2p.js:453:18)\n    at process._tickCallback (internal/process/next_tick.js:68:7)' }
```

### Steps to reproduce

Run on testnet with fixed peers in config file.

SDK 2.3

ishantiw:
### What was the problem?

- Adding fixed/blacklist/whitelist peers in config file causes the node to crash
- Whitelisted peers should not be evicted through inbound eviction mechanism.
- We should not disconnect or evict with a fixed peer even during the shuffling.

### How did I solve it?

- Valid schema in config file for different lists
- Pass white/black/fixed list to peerpool to manage connections based on these lists

### How to manually test it?

Run the app and use the list to see if the particular behavior of a peer list works or not

### Review checklist

- [x] The PR resolves #4197 #4213 #4214  

### DIFF
--- a/elements/lisk-p2p/src/p2p.ts
+++ b/elements/lisk-p2p/src/p2p.ts
@@ -477,7 +477,7 @@ export class P2P extends EventEmitter {
 					? config.rateCalculationInterval
 					: DEFAULT_RATE_CALCULATION_INTERVAL,
 			secret: config.secret ? config.secret : DEFAULT_RANDOM_SECRET,
-			peerList: this._sanitizedPeerLists,
+			peerLists: this._sanitizedPeerLists,
 		});
 
 		this._bindHandlersToPeerPool(this._peerPool);
--- a/elements/lisk-p2p/src/p2p.ts
+++ b/elements/lisk-p2p/src/p2p.ts
@@ -477,6 +477,7 @@ export class P2P extends EventEmitter {
 					? config.rateCalculationInterval
 					: DEFAULT_RATE_CALCULATION_INTERVAL,
 			secret: config.secret ? config.secret : DEFAULT_RANDOM_SECRET,
+			peerList: this._sanitizedPeerLists,
 		});
 
 		this._bindHandlersToPeerPool(this._peerPool);
--- a/elements/lisk-p2p/src/p2p.ts
+++ b/elements/lisk-p2p/src/p2p.ts
@@ -898,6 +898,10 @@ export class P2P extends EventEmitter {
 			}
 		});
 
+		// According to LIP, add whitelist peers to triedPeer by upgrading them initially.
+		this._sanitizedPeerLists.whitelisted.forEach(whitelistPeer =>
+			this._peerBook.upgradePeer(whitelistPeer),
+		);
 		await this._startPeerServer();
 
 		// We need this check this._isActive in case the P2P library is shut down while it was in the middle of starting up.
--- a/elements/lisk-p2p/src/p2p.ts
+++ b/elements/lisk-p2p/src/p2p.ts
@@ -127,7 +127,7 @@ export const DEFAULT_POPULATOR_INTERVAL = 10000;
 export const DEFAULT_SEND_PEER_LIMIT = 25;
 // Max rate of WebSocket messages per second per peer.
 export const DEFAULT_WS_MAX_MESSAGE_RATE = 100;
-export const DEFAULT_WS_MAX_MESSAGE_RATE_PENALTY = 10;
+export const DEFAULT_WS_MAX_MESSAGE_RATE_PENALTY = 100;
 export const DEFAULT_RATE_CALCULATION_INTERVAL = 1000;
 export const DEFAULT_WS_MAX_PAYLOAD = 3048576; // Size in bytes
 
--- a/elements/lisk-p2p/src/peer_pool.ts
+++ b/elements/lisk-p2p/src/peer_pool.ts
@@ -601,7 +601,13 @@ export class PeerPool extends EventEmitter {
 	}
 
 	private _selectPeersForEviction(): Peer[] {
-		const peers = [...this.getPeers(InboundPeer)];
+		const peers = [...this.getPeers(InboundPeer)].filter(peer => {
+			const found = this._peerLists.whitelisted.find(
+				whitePeer => constructPeerIdFromPeerInfo(whitePeer) === peer.id,
+			);
+
+			return found ? false : true;
+		});
 		// Cannot predict which netgroups will be protected
 		const filteredPeersByNetgroup = this._peerPoolConfig.netgroupProtectionRatio
 			? filterPeersByCategory(peers, {
@@ -659,7 +665,15 @@ export class PeerPool extends EventEmitter {
 		}
 
 		if (kind === OutboundPeer) {
-			const selectedPeer = shuffle(peers)[0];
+			const selectedPeer = shuffle(
+				peers.filter(peer => {
+					const found = this._peerLists.fixedPeers.find(
+						fixedPeer => constructPeerIdFromPeerInfo(fixedPeer) === peer.id,
+					);
+
+					return found ? false : true;
+				}),
+			)[0];
 			if (selectedPeer) {
 				this.removePeer(
 					selectedPeer.id,
--- a/elements/lisk-p2p/src/peer_pool.ts
+++ b/elements/lisk-p2p/src/peer_pool.ts
@@ -401,14 +401,18 @@ export class PeerPool extends EventEmitter {
 	): void {
 		// Try to connect to disconnected peers without including the fixed ones which are specially treated thereafter
 		const disconnectedNewPeers = newPeers.filter(
-			peer =>
-				!this._peerMap.get(constructPeerIdFromPeerInfo(peer)) ||
-				!fixedPeers.includes(peer),
+			newPeer =>
+				!this._peerMap.has(constructPeerIdFromPeerInfo(newPeer)) ||
+				!fixedPeers
+					.map(fixedPeer => fixedPeer.ipAddress)
+					.includes(newPeer.ipAddress),
 		);
 		const disconnectedTriedPeers = triedPeers.filter(
-			peer =>
-				!this._peerMap.get(constructPeerIdFromPeerInfo(peer)) ||
-				!fixedPeers.includes(peer),
+			triedPeer =>
+				!this._peerMap.has(constructPeerIdFromPeerInfo(triedPeer)) ||
+				!fixedPeers
+					.map(fixedPeer => fixedPeer.ipAddress)
+					.includes(triedPeer.ipAddress),
 		);
 		const { outboundCount } = this.getPeersCountPerKind();
 		const disconnectedFixedPeers = fixedPeers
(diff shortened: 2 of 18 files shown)


elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -398,17 +402,17 @@ export class PeerPool extends EventEmitter {
 		// Try to connect to disconnected peers without including the fixed ones which are specially treated thereafter
 		const disconnectedNewPeers = newPeers.filter(
 			peer =>
-				!this._peerMap.has(constructPeerIdFromPeerInfo(peer)) ||
+				!this._peerMap.get(constructPeerIdFromPeerInfo(peer)) ||
```
shuse2: why change to `get`? in this case `has` seems more suitable

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -597,7 +601,13 @@ export class PeerPool extends EventEmitter {
 	}
 
 	private _selectPeersForEviction(): Peer[] {
-		const peers = [...this.getPeers(InboundPeer)];
+		const peers = [...this.getPeers(InboundPeer)].filter(peer => {
+			const found = this._peerLists.whitelisted.find(
```
shuse2:
```
const found = !!this._peerLists.whitelisted.find(
	whitePeer => constructPeerIdFromPeerInfo(whitePeer) === peer.id,
);
return !found;
```
is probably fine? i could be even one line, but not so readable too

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -655,7 +665,15 @@ export class PeerPool extends EventEmitter {
...
 
 		if (kind === OutboundPeer) {
-			const selectedPeer = shuffle(peers)[0];
+			const selectedPeer = shuffle(
+				peers.filter(peer => {
+					const found = this._peerLists.fixedPeers.find(
+						fixedPeer => constructPeerIdFromPeerInfo(fixedPeer) === peer.id,
+					);
+
+					return found ? false : true;
```
shuse2: sane as avove

framework/src/modules/network/defaults/config.js
```diff
@@ -52,27 +52,63 @@ const defaultConfig = {
...
 		fixedPeers: {
 			type: 'array',
 			items: {
-				type: 'string',
-				format: 'ip',
+				type: 'object',
+				properties: {
+					ip: {
+						type: 'string',
+						format: 'ipOrFQDN',
```
shuse2: do we allow FQDN? I think we only check for the seed peers, same for below

framework/src/modules/network/defaults/config.js
```diff
@@ -52,27 +52,63 @@ const defaultConfig = {
 		blacklistedPeers: {
 			type: 'array',
 			items: {
-				type: 'string',
-				format: 'ip',
+				type: 'object',
+				properties: {
+					ip: {
+						type: 'string',
+						format: 'ipOrFQDN',
```
shuse2: wasn't blackList suppose to be list of IPs not including port?

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -398,17 +402,17 @@ export class PeerPool extends EventEmitter {
 		// Try to connect to disconnected peers without including the fixed ones which are specially treated thereafter
 		const disconnectedNewPeers = newPeers.filter(
 			peer =>
-				!this._peerMap.has(constructPeerIdFromPeerInfo(peer)) ||
+				!this._peerMap.get(constructPeerIdFromPeerInfo(peer)) ||
 				!fixedPeers.includes(peer),
```
jondubois: We should check using peer id instead of by peer object reference.

framework/src/modules/network/defaults/config.js
ManuGowda: think to have a consistent definition of peer makes it clear for usability and configuration.

shuse2: FQDN cannot be used for these things.
Seed peers are only used at the beginning, therefore we converted to the ip at the beginning.
However, other ip are the runtime validation, and we should not use the DNS resolver.

ishantiw: Yes, I have reverted the changes to use only IP for blacklist. For the seedPeers makes sense to use FQDN because we have seed peers by default in the config when someone starts a node whereas for the fixed and white list are optional lists that will be used by the user for their use-cases.

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -602,12 +606,13 @@ export class PeerPool extends EventEmitter {
 
 	private _selectPeersForEviction(): Peer[] {
 		const peers = [...this.getPeers(InboundPeer)].filter(peer => {
-			const found = this._peerLists.whitelisted.find(
+			const found = !!this._peerLists.whitelisted.find(
```
diego-G: This is hard to read. Is there any way to improve readability here?

elements/lisk-p2p/src/p2p.ts
```diff
@@ -897,6 +898,10 @@ export class P2P extends EventEmitter {
 			}
 		});
 
+		// According to LIP, add whitelist peers to triedPeer by upgrading them initially.
+		this._sanitizedPeerLists.whitelisted.forEach(whitelistPeer =>
```
diego-G: Is it possible to cover it by any integration test?

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -597,7 +605,14 @@ export class PeerPool extends EventEmitter {
 	}
 
 	private _selectPeersForEviction(): Peer[] {
-		const peers = [...this.getPeers(InboundPeer)];
+		const peers = [...this.getPeers(InboundPeer)].filter(peer => {
```
diego-G: I'd expand the test coverage for eviction feature.

framework/src/modules/network/network.js
```diff
@@ -111,19 +111,18 @@ module.exports = class Network {
 		);
 
 		const seedPeers = await lookupPeersIPs(this.options.seedPeers, true);
-		const blacklistedPeers = this.options.blacklistedPeers
-			? this.options.blacklistedPeers.map(peer => ({
-					ipAddress: peer.ip,
-			  }))
-			: [];
+		const blacklistedPeers = this.options.blacklistedPeers || [];
```
diego-G: I suggest to rename this to `blacklistedIPs`.

elements/lisk-p2p/src/peer_pool.ts
ishantiw: The last one was more readable of course, @shuse2 agrees then I can revert the changes. I'm fine with both.

shuse2: find returns `undefined` if not found, and changing it to `boolean`.
I think it's common expression in JS

elements/lisk-p2p/src/peer_pool.ts
ishantiw: I've tested it in integration test to see if white list from config is making to the sanitized list but since in the sanitized list we remove ourselves from whitelist, we need not local ip to test it. So we can move it to QA testing instead of testing it locally

elements/lisk-p2p/src/p2p.ts
ishantiw: I get your point but we don't want to test `new/triedPeer` in the integration tests. We already tested that when we do `upgradePeer` then it goes to `triedPeers` in peer_directory unit tests.

framework/src/modules/network/network.js
ishantiw: we will make this change in `3.0`

elements/lisk-p2p/src/p2p.ts
```diff
@@ -477,6 +477,7 @@ export class P2P extends EventEmitter {
 					? config.rateCalculationInterval
 					: DEFAULT_RATE_CALCULATION_INTERVAL,
 			secret: config.secret ? config.secret : DEFAULT_RANDOM_SECRET,
+			peerList: this._sanitizedPeerLists,
```
jondubois: This should be plural `peerLists` because there are multiple lists.

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -109,6 +110,7 @@ interface PeerPoolConfig {
 	readonly wsMaxMessageRatePenalty: number;
 	readonly rateCalculationInterval: number;
 	readonly secret: number;
+	readonly peerList: PeerLists;
```
jondubois: Plural `peerLists` if it's multiple lists.

elements/lisk-p2p/src/peer_pool.ts
mitsujutsu: can also do something like `.filter(peer => this._peerLists.whitelisted.every(p  => constructPeerIdFromPeerInfo(p) !== peer.id)` to make it shorter

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -597,7 +605,14 @@ export class PeerPool extends EventEmitter {
 	}
 
 	private _selectPeersForEviction(): Peer[] {
-		const peers = [...this.getPeers(InboundPeer)];
+		const peers = [...this.getPeers(InboundPeer)].filter(peer => {
+			const found = !!this._peerLists.whitelisted.find(
+				whitePeer => constructPeerIdFromPeerInfo(whitePeer) === peer.id,
```
mitsujutsu: prefer `whitelistedPeer`

PR #4234, opened 2019-09-09
Split block processing steps - Closes #3869

### LINKED ISSUE #3869: Split block processing steps into separate functions and expose them from module/blocks
### Expected behavior
Block processing includes block and transactions validation, verification and applying into the blockchain. Each step is an expensive operation and should be called only once for each block. As part of BFT, we execute these steps at different stages, and block processing should allow calling these functions independently.
### Actual behavior
process.js file only exports applyBlock and processBlock functions which encapsulate all the work related to validating, verifying and applying a block. It does not allow executing these steps separately. 

feature/introduce_bft_consensus

limiaspasdaniel:
### What was the problem?
Block processing steps needed to be split into a more granular level in order to allow BFT implementations to call the logic separately
### How did I solve it?
By adding a new pipeline `validateDetached` responsible for validating a detached block (block that is not sequential to the last block in the system) and also by reordering pipeline logic under `block_processor_v2`
### How to manually test it?
Run `blocks.spec.js` tests
### Review checklist
- [ ] The PR resolves #3869 


framework/src/modules/chain/block_processor_v2.js
```diff
@@ -179,9 +206,13 @@ class BlockProcessorV2 extends BaseBlockProcessor {
 		// TODO: Remove validate new since it's no longer required
 		this.validateNew.pipe([() => Promise.resolve()]);
 
-		this.verify.pipe([data => this.blocksModule.verify(data)]);
+		this.verify.pipe([({ block }) => this.bft.verifyNewBlock(block)]);
```
shuse2: we don't need to verify block here anymore?
it feels bit weird to have `verify` in apply step

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -242,8 +255,6 @@ class Blocks extends EventEmitter {
 	async verify({ block, skipExistingCheck }) {
```
shuse2: do we still need this as public function after adding `verifyInMemory`?

framework/src/modules/chain/block_processor_v2.js
SargeKhan: The block must be broadcasted prior to performing any database related checks. That is what is described in the LIP. I concede that it looks odd to have verify in apply but perhaps we should restructure the blocks module a little bit to adjust to that?
Also, we opened this PR because as draft because we want to have a discussion on how to approach this topic - should we update the blocks module to reflect the LIP or should we go on a tangent from what is described in the LIP.
Please suggest alternatives. :)

PR #4237, opened 2019-09-09
Protocol-spec generator should not generate different JSON on each run - Closes #4193

### LINKED ISSUE #4193: Protocol-spec generator should not generate different json on each run
### Description

Generators output different JSON (due to transactions timestamps) on each run and they should always output the same JSON regardless of the time being run

### Acceptance Criteria

- [x] Confirm if transaction timestamps are causing this and fix the date

pablitovicente:
# What was the problem?

A timestamp was not being set so generated transaction would always have a different signature and id.

### How did I solve it?

By fixing a configurable timestamp for all transactions.

### How to manually test it?

Run the generators several times and compare the outputs

### Review checklist

- [x] The PR resolves #4193


protocol-specs/generators/base_generator.js
```diff
@@ -33,7 +33,7 @@ const runGenerator = (generatorName, suiteCreators) => {
 
 		const fullPath = `${dirPath}/${suiteResult.handler}.json`;
 
-		fs.writeFileSync(fullPath, JSON.stringify(suiteResult, null, 2));
+		fs.writeFileSync(fullPath, `${JSON.stringify(suiteResult, null, 2)}\u000A`);
```
shuse2: why do we need to do this way? maybe comment would be nice

protocol-specs/generators/block_processing_second_signature/index.js
```diff
@@ -292,12 +303,16 @@ const generateTestCasesInvalidBlockSecondSignatureTxSecondTime = () => {
 
 	const secondSignature =
 		'erupt sponsor rude supreme vacant delay salute allow laundry swamp curve brain';
+	const secondPassphraseObject = registerSecondPassphrase({
```
shuse2: do we need to use this function at all?

protocol-specs/generators/block_processing_second_signature/index.js
```diff
@@ -200,22 +207,26 @@ const generateTestCasesValidBlockSecondSignatureTx = () => {
...
 const generateTestCasesinvalidBlockWithSecondSignatureAndFundsTxSuite = () => {
 	const amount = '5500000000';
-	const transferTx = new TransferTransaction(
-		transfer({
-			amount,
-			passphrase: accounts.genesis.passphrase,
-			recipientId: accounts.existingDelegate.address,
-		}),
-	);
+	const transferObject = transfer({
```
shuse2: same as below. Do we need to mix this function and the instance?
It was meant to be used one way or another

pablitovicente: It was faster to implement this way with the least amount of changes I will take the extra time for doing it from JSON/Object literal to instance.

protocol-specs/generators/base_generator.js
pablitovicente: Prettier was removing last line and linter was complaining about \n will look for workarounds or add comment.

pablitovicente: Fixed guess was my IDE being funky

protocol-specs/generators/block_processing_second_signature/index.js
pablitovicente: Fixed.

protocol-specs/generators/block_processing_second_signature/index.js
pablitovicente: Removed and replaced by object and `getKeys`

PR #4240, opened 2019-09-10
Unstable P2P integration test - Closes#4219

### LINKED ISSUE #4219: p2p integration tests fail sometimes and are unstable
### Expected behavior

P2P integration tests should pass and should be stable

### Actual behavior

P2P integration tests fail sometimes on Jenkins or locally.

### Steps to reproduce

Run `npm t` 

`2.3`

ishantiw:
### Review checklist

- [x] The PR resolves #4219 


elements/lisk-p2p/test/integration/custom_peer_selection.ts
```diff
@@ -84,17 +84,13 @@ describe('Custom peer selection', () => {
...
-								ipAddress: '127.0.0.1',
-								wsPort: NETWORK_START_PORT + ((index + 1) % NETWORK_PEER_COUNT),
-							},
-					  ];
-
 			const nodePort = NETWORK_START_PORT + index;
+			const seedPeers = [...new Array(NETWORK_PEER_COUNT / 2).keys()]
+				.map(index => ({
+					ipAddress: '127.0.0.1',
+					wsPort: NETWORK_START_PORT + ((index + 2) % NETWORK_PEER_COUNT),
```
jondubois: Why not make all nodes seed peer of each other? That way we can completely factor out the discovery aspect from this test since it's not related.

The `peerSelectionForConnection` feature would be affected by the discovery but it doesn't appear to be tested in this file so we should try to make the discovery happen as fast as possible with more seed peers.

elements/lisk-p2p/test/integration/maximum_payload.ts
```diff
@@ -34,20 +34,20 @@ describe('Maximum payload', () => {
 		for (let i = 0; i < 6000; i++) {
 			dataLargerThanMaxPayload.push(`message${i}`);
 		}
-		p2pNodeList = [...new Array(NETWORK_PEER_COUNT).keys()].map(index => {
+		p2pNodeList = [...new Array(NETWORK_PEER_COUNT / 2).keys()].map(index => {
```
jondubois: Also it's not clear why only half. It seems more arbitrary and adds an element of randomness depending on which nodes start first.

ishantiw: So the reason for such a way of choosing seed peers is, many times the first node tries to make a connection with other nodes that already have an incoming connection with it. It was working before without the hotfix because it was still counting the outbound connection to others even though it was disconnected from there end, but now it throws duplicate connection error and it marks them as disconnected, hence giving the right count of outbound/inbound

elements/lisk-p2p/test/integration/custom_peer_selection.ts
ishantiw: same reason as the comment https://github.com/LiskHQ/lisk-sdk/pull/4240#discussion_r322816575

ishantiw: > The peerSelectionForConnection feature would be affected by the discovery but it doesn't appear to be tested in this file so we should try to make the discovery happen as fast as possible with more seed peers

Before in this test, only one seed peer for there with one node. At least now it's half of the nodes as seed.

PR #4246, opened 2019-09-11
Create function to compute BFT headers on block creation - Closes #4233

### LINKED ISSUE #4233: Create function to compute BFT headers on block creation
### Description
When forging a new block, we need to compute two new properties and include them in the header:
- `maxHeightPreviouslyForged`
- `prevotedConfirmedUptoHeight` 

Currently, the `BFT` module doesn't provide an interface to compute those properties so **this must be implemented**

### Motivation
We need those properties in order to create a block successfully under the new protocol.

### Acceptance Criteria
- These two properties are properly computed and edge cases are considered.
- Unit tests are implemented

### Additional Information

limiaspasdaniel:
### What was the problem?
There was no way to compute these BFT headers required for block forging.
### How did I solve it?
By implementing a function to handle this computation inside `BFT` module.
### How to manually test it?
Run Jest tests on `bft.spec.js` file.
### Review checklist

- [ ] The PR resolves #4233 

### DIFF
--- a/framework/src/modules/chain/bft/bft.js
+++ b/framework/src/modules/chain/bft/bft.js
@@ -148,7 +148,7 @@ class BFT extends EventEmitter {
 	 * @param delegatePubKey
 	 * @return {Promise<{prevotedConfirmedUptoHeight: number, maxHeightPreviouslyForged: (number|*)}>}
 	 */
-	async computeBlockHeadersForNewBlock(delegatePubKey) {
+	async computeHeadersForNewBlock(delegatePubKey) {
 		const blockHeaders = cloneDeep(this.finalityManager.headers.items);
 		const lastBlockForgedByDelegate = blockHeaders
 			.reverse()
--- a/framework/src/modules/chain/bft/bft.js
+++ b/framework/src/modules/chain/bft/bft.js
@@ -147,21 +147,8 @@ class BFT extends EventEmitter {
 	 * @param delegatePubKey
 	 * @return {Promise<{prevotedConfirmedUptoHeight: number, maxHeightPreviouslyForged: (number|*)}>}
 	 */
-	async computeHeadersForNewBlock(delegatePubKey) {
-		const [
-			lastBlockForgedByDelegate,
-		] = this.finalityManager.headers.items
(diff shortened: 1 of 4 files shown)


framework/src/modules/chain/bft/bft.js
```diff
@@ -141,6 +142,28 @@ class BFT extends EventEmitter {
...
 	}
 
+	/**
+	 * Computes maxHeightPreviouslyForged and prevotedConfirmedUptoHeight properties that are necessary
+	 * for creating a new block
+	 * @param delegatePubKey
+	 * @return {Promise<{prevotedConfirmedUptoHeight: number, maxHeightPreviouslyForged: (number|*)}>}
+	 */
+	async computeHeadersForNewBlock(delegatePubKey) {
+		const blockHeaders = cloneDeep(this.finalityManager.headers.items);
```
SargeKhan: Why are we required to clone the headers?

framework/src/modules/chain/bft/bft.js
```diff
@@ -141,6 +142,28 @@ class BFT extends EventEmitter {
...
+		const lastBlockForgedByDelegate = blockHeaders
+			.reverse()
+			.find(blockHeader => blockHeader.delegatePublicKey === delegatePubKey);
```
SargeKhan: Is this method of getting the lastBlockForgedByDelegate valid in all cases? Imagine a scenario where a delegate forged on a forked chain. In this case, the `lastBlockForgedByDelegate` should correspond to that block; this block will not be part of the blockHeaders.

framework/test/jest/unit/specs/modules/chain/bft/bft.spec.js
```diff
@@ -386,6 +391,134 @@ describe('bft', () => {
...
 		});
 
+		describe('computeHeadersForNewBlock()', () => {
+			let bft;
+
+			beforeEach(async () => {
+				bft = new BFT(bftParams);
+
+				jest
+					.spyOn(bft, '_getLastBlockHeight')
```
SargeKhan: Do we need to spy on mocked implementations?

PR #4251, opened 2019-09-12
Move generate delegate list to dpos apply and undo - Closes #4218

### LINKED ISSUE #4218: Move generating delegate list logic into Dpos.apply and Dpos.undo
### Expected behavior
- getRoundDelegates should only return record from db. throw exception otherwise
- apply call should **upsert** the delegate list for the next round if the block is the last block of the round. 
### Actual behavior
- getRoundDelegates implementation allows to provide a future round number which returns a wrong delegate list and writes it to the database.
3.0.0

michielmulders:
### What was the problem?
Generate delegate list should be part of apply/undo in DPoS.
Also, `getDelegateList` should only get the delegate list for a round that exists, otherwise throw an error. Also, `getDelegateList` should not be responsible for creating a list when it doesn't exist (DPoS responsibility).

### How did I solve it?
Add generate delegate list as a step for DPoS.apply or undo.

### Review checklist

- [x] The PR resolves #4218 


framework/src/modules/chain/dpos/delegates_list.js
```diff
@@ -64,22 +64,37 @@ class DelegatesList extends EventEmitter {
...
 		}
 
 		return delegatePublicKeys;
 	}
 
+	async createRoundDelegateList(round, tx) {
+		const delegatePublicKeys = await this.getDelegatePublicKeysSortedByVoteWeight();
+
+		// Delete delegate list and create new updated list
+		await this.storage.entities.RoundDelegates.delete(
```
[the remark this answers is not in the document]
michielmulders: It's this or either select and see if the table holds records (upsert).
Prefer the delete-create.

framework/src/modules/chain/dpos/delegates_info.js
```diff
@@ -196,7 +204,7 @@ class DelegatesInfo {
 		const round = this.slots.calcRound(block.height);
 		const nextRound = this.slots.calcRound(block.height + 1);
 
-		return round < nextRound || block.height === 1;
+		return round < nextRound;
```
yatki: ❤️

framework/src/modules/chain/dpos/delegates_info.js
```diff
@@ -82,21 +82,29 @@ class DelegatesInfo {
...
 				 * If we are reverting the block, new transactions
 				 * can change vote weight of delegates, so we need to
 				 * invalidate the cache for the next rounds.
 				 */
-				const round = this.slots.calcRound(block.height);
 				await this.delegatesList.deleteDelegateListAfterRound(round, tx);
+			} else {
+				// Create round delegate list
+				const nextRound = round + 1;
+				await this.delegatesList.createRoundDelegateList(nextRound, tx);
```
yatki: I think we have to create the round delegate list after applying rewards fees and updating the votedDelegateVoteWeights after await Promise all.

framework/src/modules/chain/dpos/delegates_list.js
```diff
@@ -53,33 +53,54 @@ class DelegatesList extends EventEmitter {
...
 
+	async createRoundDelegateList(round, tx) {
+		const delegatePublicKeys = await this.getDelegatePublicKeysSortedByVoteWeight(
+			tx,
+		);
+
+		// Delete delegate list and create new updated list
+		await this.storage.entities.RoundDelegates.delete(
+			{
+				round_gte: round,
```
yatki: I'm not sure if we need to use round_gte because undo is already deleting the rounds that is not needed. What would be the case we would still have a round number with greater number.

I'd prefer to delete only round itself to see other problems could happen in dpos. I feel like this could suppress a bigger problem.

michielmulders: You are right about this. We thought more about this as a safety net?

framework/src/modules/chain/dpos/delegates_info.js
michielmulders: Does the order matter? We can also create the list before this Promise.all?

shuse2: ah that's true.
delegate weight changes after this `Promise.all`, so the round delegate list should be generated after that

framework/src/modules/chain/dpos/delegates_list.js
yatki: I would say real safety is to see if we mess up in update function. If somebody breaks the functionality that takes care of deleting the list on undo, we may never notice it until we need to revert a block. In general I feel less safe when a function is doing another function's job. :)

framework/test/jest/unit/specs/modules/chain/dpos/apply.spec.js
```diff
@@ -99,6 +113,25 @@ describe('dpos.apply()', () => {
 
 			expect(stubs.storage.entities.Account.update).not.toHaveBeenCalled();
 		});
+
+		it('should call RoundDelegates.delete method when applying genesis block', async () => {
```
lsilvs: Maybe change the description and make sure `create` is called after `delete`?

```suggestion
		it('should save round 1 active delegates list in RoundDelegates entity after applying genesis block', async () => {
```

framework/test/jest/unit/specs/modules/chain/dpos/apply.spec.js
```diff
@@ -280,6 +325,28 @@ describe('dpos.apply()', () => {
 			});
 		});
 
+		it('should call create for RoundDelegates for last block of round', async () => {
```
lsilvs: Same here

```suggestion
		it('should save next round active delegates list in RoundDelegates entity after applying last block of round', async () => {
```

framework/test/jest/unit/specs/modules/chain/dpos/apply.spec.js
michielmulders: This functionality is part of: https://www.npmjs.com/package/jest-extended#tohavebeencalledbefore Need to check if we can include this.

framework/src/modules/chain/dpos/delegates_list.js
michielmulders: Changed `gte` to just the `round` number. :)

framework/test/jest/unit/specs/modules/chain/dpos/apply.spec.js
michielmulders: Nvm, this is already present in SDK.

framework/src/modules/chain/dpos/delegates_info.js
```diff
@@ -81,23 +81,18 @@ class DelegatesInfo {
 		 * the genesis block, in that case we don't have to
 		 * update anything in the accounts.
 		 */
-		if (block.height === 1) {
+		if (this._isGenesisBlock(block)) {
+			const round = 1;
+			await this.delegatesList.createRoundDelegateList(round, tx);
```
yatki: @shuse2 @lsilvs do we need to create a delegate list for the first round?

framework/src/modules/chain/dpos/delegates_list.js
```diff
@@ -53,33 +53,54 @@ class DelegatesList extends EventEmitter {
...
-		const accounts = await this.storage.entities.Account.get(filters, options);
+		const accounts = await this.storage.entities.Account.get(
+			filters,
+			options,
+			tx,
+		);
 		return accounts.map(account => account.publicKey);
 	}
 
 	async generateActiveDelegateList(round) {
```
yatki: I think we should rename this function now. Or maybe we can move the whole logic inside `getRoundDelegates` and get rid of this since now it's pretty simple what it's doing?

framework/src/modules/chain/dpos/delegates_list.js
```diff
@@ -53,33 +53,54 @@ class DelegatesList extends EventEmitter {
...
-				round,
-				delegatePublicKeys,
-			});
+			throw new Error(`No delegate list found for round: ${round}`);
 		}
 
 		return delegatePublicKeys;
 	}
 
+	async createRoundDelegateList(round, tx) {
```
yatki: It is still possible to call this function with a round number from future. I think it's better the fetch the latest record and check provided round number and eventually update or insert accordingly.

Also I suggest to rename this `generateActiveDelegateList` since it has kind of took over the current `generateActiveDelegateList`'s job. And as I mentioned above, move the logic in existing `generateActiveDelegateList` method into `getRoundDelegates`.

framework/test/jest/unit/specs/modules/chain/dpos/getRoundDelegates/getRoundDelegates.spec.js
```diff
@@ -71,37 +71,18 @@ describe('dpos.getRoundDelegates()', () => {
...
 				.calledWith(roundNo)
 				.mockResolvedValue([]);
 			stubs.storage.entities.Account.get.mockResolvedValue(delegateAccounts);
-
-			// Act
-			list = await dpos.getRoundDelegates(roundNo);
 		});
 
-		it('should return shuffled delegate list by using delegate accounts', () => {
+		it('should throw error when round is not in round_delegates table', async () => {
```
yatki: ❤️

framework/src/modules/chain/dpos/delegates_info.js
michielmulders: Screenshot 2019-09-13 at 21 49 45

We decided to add this as otherwise we got the following error.

framework/src/modules/chain/dpos/delegates_list.js
michielmulders: Good catch, I don't think we are checking for this now. Also, we should add an extra test for that.

framework/src/modules/chain/dpos/delegates_info.js
shuse2: Just for the record, we do need to get the delegate list for the first round, and we were doing it before too 😉

framework/src/modules/chain/dpos/delegates_list.js
lsilvs: In what case it would be called with a round from the future?

framework/src/modules/chain/dpos/delegates_list.js
michielmulders: There is one function using this logic which relies on the unshuffled logic.
Renamed functions to:
- `getForgersForRound`
- `getActiveDelegatesForRound`

framework/src/modules/chain/dpos/delegates_list.js
yatki: @lsilvs Not in the current business logic. However, method should throw exception if a round from future was provided, just as a precaution.

lsilvs: This method is not exposed outside Dpos module.
Adding a comment saying `createRoundDelegateList` should only be called after each round finishes is enough imo.

framework/test/jest/unit/specs/modules/chain/dpos/apply.spec.js
```diff
@@ -179,6 +214,18 @@ describe('dpos.apply()', () => {
...
+				.calledWith(
+					{
+						isDelegate: true,
+					},
+					{
+						limit: constants.ACTIVE_DELEGATES,
+						sort: ['voteWeight:desc', 'publicKey:asc'],
+					},
+				)
+				.mockResolvedValue(delegateAccounts);
```
yatki: I'm not sure if these accounts are sorted by voteweight and publickey can you make sure that array is sorted.

framework/test/jest/unit/specs/modules/chain/dpos/getRoundDelegates/getRoundDelegates.spec.js
```diff
@@ -71,37 +73,18 @@ describe('dpos.getRoundDelegates()', () => {
...
-		it('should return shuffled delegate list by using delegate accounts', () => {
-			// Assert
-			expect(stubs.storage.entities.Account.get).toHaveBeenCalledWith(
-				{ isDelegate: true },
-				{
-					limit: constants.ACTIVE_DELEGATES,
-					sort: ['voteWeight:desc', 'publicKey:asc'],
-				},
-			);
-			expect(list).toEqual(shuffledDelegatePublicKeys);
```
yatki: please move this test scenario to "apply.spec.js" since it's actually part of apply behaviour.

michielmulders: This test has been removed as an error should be thrown when the round delegates table is not in the db?

PR #4254, opened 2019-09-12
The nodeInfo was removed from _peerSelectForConnection again - Closes #4253

### LINKED ISSUE #4253: The nodeInfo was removed from _peerSelectForConnection again
### Expected behavior

https://github.com/LiskHQ/lisk-sdk/blob/d45c301d61c1e8f86ce34db29a1f326fdac3a967/elements/lisk-p2p/src/peer_pool.ts#L406-L409

### Actual behavior

The selectForConnection function doesn't receive the `nodeInfo` anymore. This issue probably came back due to a merge conflict. We should add it back and add a test to prevent another regression.

Note that the nodeInfo is optional because the P2P library does not force the user to provide it but if it does exist, then we must pass it to the function.

jondubois:
### What was the problem?

The selectForConnection function wasn't passed the nodeInfo.

### How did I solve it?

- Added back nodeInfo to the selectForConnection function argument.
- Added unit test coverage for all the selection functions to prevent regression.

### How to manually test it?

`npm run test:unit`

### Review checklist

- [x] The PR resolves #4253


shuse2:
@diego-G 
This is not really IO related test so I think unit test should be fine

jondubois:
> We can have a test case where we can use peer objects from `utils.ts` and use them for selection based on some `nodeInfo` value. If it's not much then we can do it in this PR.

I didn't want to add a new integration test case because the unit test case will make sure that the selectForConnection function has access to the nodeInfo object which was the main problem. Also if I add an integration test now, there will be a huge conflict with the development branch.

elements/lisk-p2p/test/unit/peer_pool.ts
```diff
@@ -160,6 +161,79 @@ describe('peerPool', () => {
 		});
 	});
 
+	describe('#request', () => {
+		let caughtError: Error;
+		beforeEach(async () => {
+			// @ts-ignore
+			peerPool['_peerSelectForRequest'] = sandbox
```
shuse2: why do we need this `ts-ignore` here?

jondubois: My TS compiler complains if I don't. Maybe it doesn't allow assigning to private properties?

shuse2:
@jondubois It's failing on unit test

elements/lisk-p2p/test/unit/peer_pool.ts
```diff
@@ -160,6 +161,76 @@ describe('peerPool', () => {
...
+				.returns([] as ReadonlyArray<P2PPeerInfo>);
+			try {
+				await peerPool.request({ procedure: 'proc', data: 123 });
+			} catch (err) {
+				caughtError = err;
+			}
+		});
+
+		it('should call _peerSelectForRequest with all the necessary options', async () => {
+			expect(peerPool['_peerSelectForRequest']).to.be.calledWith({
```
diego-G: Prefer to write `(peerPool as any)._peerSelectForRequest`. @mitsujutsu told me to follow this approach. Please, review every case from this pull request.

elements/lisk-p2p/test/unit/peer_pool.ts
```diff
@@ -160,6 +161,76 @@ describe('peerPool', () => {
...
+				peers: [],
+				nodeInfo: peerPool.nodeInfo,
+				peerLimit: 1,
+				requestPacket: { procedure: 'proc', data: 123 },
+			});
+		});
+
+		it('should throw an error if no peers are found', async () => {
```
diego-G: I miss an extra case when no error is thrown.

jondubois: This is outside the scope of this issue. This issue is about `nodeInfo` being absent from the function. We have a different issue already open about unit tests for `PeerPool` and it should be addressed as part of that issue.

diego-G: I suggest to add a test skeleton now to not to forget the case. We should do this kind of things frequently.

PR #4257, opened 2019-09-13
Extract and use common setup for integration tests - Closes #4200

### LINKED ISSUE #4200: Reorganize integration tests based on features
### Description

Currently, integration tests are all in one file and broken up based on the network topology (partial mesh vs full mesh). We should break it up based on features. Each feature has its own test file and can specify its own P2P config/topology.

### Motivation

It will make it easier for people to understand the P2P library and to make changes if we break up the integration tests based on features instead of network topology.

### Acceptance Criteria

The integration tests are separated into files based on features.

ishantiw:
### What was the problem?

Every test scenario was using network setup with almost 90% common configuration. A lot of lines were added in each test file and also difficult to track every variable if it is used differently in a test case.

### How did I solve it?

Extracted a setup file that exposes `createNetwork` and `destroyNetwork` functions that creates a network with default settings. Highly configurable at granular level taking into `index`, 'startPort' and `networkSize` into account.

All the tests have been modified to use this setup and we can easily see and track what values we have changed for a particular test case.

### How to manually test it?

`npm run test:integration`

### Review checklist

- [ ] The PR resolves #4200 


elements/lisk-p2p/test/integration/custom_peer_selection.ts
```diff
@@ -25,11 +25,14 @@ import {
 	P2PPeerSelectionForConnectionInput,
 } from '../../src/p2p_types';
 
+import {
```
diego-G: Why not using `createNetwork` & `destroyNetwork` here?

elements/lisk-p2p/test/utils/network_setup.ts
```diff
@@ -0,0 +1,142 @@
...
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import { P2P } from '../../src/index';
+import { wait } from './helpers';
+import { platform } from 'os';
+export const NETWORK_START_PORT = 5000;
+export const NETWORK_PEER_COUNT = 10;
+export const POPULATOR_INTERVAL = 50;
+export const DEFAULT_MAX_OUTBOUND_CONNECTIONS = 20;
```
diego-G: Let's get these defaults from `src/constants.js`

elements/lisk-p2p/test/integration/custom_peer_selection.ts
ishantiw: Oh, I somehow I missed to add the changes while committing, will do

elements/lisk-p2p/test/utils/network_setup.ts
```diff
@@ -0,0 +1,142 @@
...
+				broadhash: nodeInfoConstants.broadhash,
+				nonce: `${nodeInfoConstants.nonce}${nodePort}`,
+				...customNodeInfoObject,
+			},
+			...customConfigObject,
+		};
+
+		return new P2P(p2pConfig);
+	});
+	await Promise.all(p2pNodeList.map(async p2p => await p2p.start()));
```
shuse2: it shouldn't wait inside of `map`. Since this is promise.all, it can just return the promise directly

elements/lisk-p2p/test/utils/network_setup.ts
```diff
@@ -0,0 +1,142 @@
...
+
+	return p2pNodeList;
+};
+
+export const destroyNetwork = async (
+	p2pNodeList: ReadonlyArray<P2P>,
+	networkDestroyWaitTime?: number,
+) => {
+	await Promise.all(
+		p2pNodeList.filter(p2p => p2p.isActive).map(async p2p => await p2p.stop()),
```
shuse2: same as above

elements/lisk-p2p/test/integration/custom_node_info.ts
```diff
@@ -14,73 +14,28 @@
...
+		const customModules = () => ({
+			modules: {
+				names: ['test', 'crypto'],
+				active: true,
+			},
 		});
-		await Promise.all(p2pNodeList.map(async p2p => await p2p.start()));
-		await wait(1000);
+
+		p2pNodeList = await createNetwork({ customNodeInfo: customModules });
```
jondubois: This is a good example where the `createNetwork` utility makes the test easier to follow because there aren't many custom properties.

elements/lisk-p2p/test/integration/blacklist_whitelist_fixed_peers.ts
```diff
@@ -84,45 +74,33 @@ describe('Blacklisted/fixed/whitelisted peers', () => {
...
-						protocolVersion: '1.0.1',
-						minVersion: '1.0.0',
-						os: platform(),
-						height: 0,
-						broadhash:
-							'2768b267ae621a9ed3b3034e2e8a1bed40895c621bbb1bbd613d92b9d24e54b5',
-						nonce: `O2wTkjqplHII${nodePort}`,
-					},
-				});
+			const customConfig = (index: number) => ({
```
jondubois: We shouldn't pass any callback functions to the `createNetwork`; that makes the code difficult to reason about. It forces the developer to dig into the boilerplate logic and keep a lot of things in their mind at the same time which distracts them from the important aspects of the test scenario.

elements/lisk-p2p/test/utils/network_setup.ts
```diff
@@ -0,0 +1,142 @@
...
+	) => any[];
+}
+
+export const createNetwork = async ({
+	networkSize,
+	startNodePort,
+	networkCreationWaitTime,
+	customConfig,
+	customNodeInfo,
+	customSeedPeers,
```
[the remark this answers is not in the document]
ishantiw: The reason for injecting these functions is the way we create config for each p2p instance in the network. I saw that we are at most using `index`, `startPort` and `networkSize` to generate some config values. So it's easier to generate any custom config value using these params. I understand that a new developer at first won't be able to understand what are these custom functions. But if any developer wants to add any new integration test, we assume that it has basic knowledge basic working of p2p-lib, about `P2PConfig` and creating a `P2P` instance. 

Also, if a new developer wants to create a new integration test, it will be hard for him to know the consistent `P2Pconfig` to be used because then it has to go through many files to know what config other tests are using (especially with so many hardcoded values and sometimes doesn't make sense and unrelated to a test case). With a single file where all the networks are being created, it's easier for it to understand how it works and it doesn't need to know what other tests are doing. So IMO, even though we are giving up some readability but we are gaining more flexibility, consistency, and scalability. Also, it was actually harder for me to read a test_scenario while debugging when the config code was taking the most chunk of a file.

The other advantage is, we have been dealing with a lot of issues related to unstable integration tests. From my experience, it was hard to find out what is different in the config in a test case that is making it fail. With this approach we are sure, what is the exact difference for a test case.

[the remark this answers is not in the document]
jondubois: I feel that the `createNetwork` function is currently almost complex enough to require a test case of its own; which is a red flag. Test logic should be very obvious.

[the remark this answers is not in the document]
jondubois: One problem in Lisk core network test was that often we didn't know if the failing test was because of our lisk-core code or because we messed something up in the test util... Very often there was something wrong in the util.

elements/lisk-p2p/test/utils/network_setup.ts
```diff
@@ -15,11 +15,13 @@
...
+import {
+	DEFAULT_MAX_OUTBOUND_CONNECTIONS,
+	DEFAULT_MAX_INBOUND_CONNECTIONS,
+} from '../../src';
 export const NETWORK_START_PORT = 5000;
 export const NETWORK_PEER_COUNT = 10;
 export const POPULATOR_INTERVAL = 50;
-export const DEFAULT_MAX_OUTBOUND_CONNECTIONS = 20;
-export const DEFAULT_MAX_INBOUND_CONNECTIONS = 100;
 export const DEFAULT_CONNECTION_TIMEOUT = 500;
```
diego-G: What about this `DEFAULTS`? Shall we maybe rename or add them to `src/constants.ts`?

ishantiw: All the other values are actually not defaulted constant values but have been changed to run tests faster. I can rename it and remove `DEFAULT` for the ones that are not coming from constants

elements/lisk-p2p/test/utils/network_setup.ts
```diff
@@ -0,0 +1,144 @@
...
+	minVersion: '1.0.0',
+	os: platform(),
+	height: 0,
+	broadhash: '2768b267ae621a9ed3b3034e2e8a1bed40895c621bbb1bbd613d92b9d24e54b5',
+	nonce: `O2wTkjqplHII`,
+};
+
+interface TestNetworkConfig {
+	networkSize?: number;
+	startNodePort?: number;
```
mitsujutsu: Do we actually need this? Are there cases where it isn't `5000`?

elements/lisk-p2p/test/utils/network_setup.ts
```diff
@@ -0,0 +1,144 @@
...
+	networkCreationWaitTime?: number;
```
mitsujutsu: prefer `networkDiscoveryWaitTime` to be more specific

elements/lisk-p2p/test/utils/network_setup.ts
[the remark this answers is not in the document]
mitsujutsu: I prefer to not have `customSeedPeers` as that boilerplate is not too much overhead on each test and the value add of the function doesn't justify the added complexity. I prefer @jondubois suggestion of having separate `customNodeInfo`, `customConfig` utility functions. `destroyNetwork` is nice and I think we can keep that as is.

elements/lisk-p2p/test/utils/network_setup.ts
ishantiw: Maybe we don't need it but I see a lot of values been calculated using this value. So, in general, I wanted to make it configurable if someone wants to run two parallel integration tests on different ports then it will be helpful to modify this value.

elements/lisk-p2p/test/utils/network_setup.ts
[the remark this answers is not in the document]
ishantiw: So I have simplified to only one function for `customConfig` as per suggested by @jondubois . It's still open for discussion if we think it's more complex to use then we can move the logic to each and every file. But by looking at each test, I think there is not much change in the config and its not so complex to use.

elements/lisk-p2p/test/integration/duplicate_peers.ts
```diff
@@ -26,53 +26,9 @@ describe('Disconnect duplicate peers', () => {
...
-					broadhash:
-						'2768b267ae621a9ed3b3034e2e8a1bed40895c621bbb1bbd613d92b9d24e54b5',
-					nonce: `O2wTkjqplHII${nodePort}`,
-				},
-			});
-		});
-		await Promise.all(p2pNodeList.map(async p2p => await p2p.start()));
-
-		await wait(1000);
+		p2pNodeList = await createNetwork({});
```
shuse2: probably `createNetwork` could've have default to `{}` so that it doesn't need to provide, but not so important point

PR #4259, opened 2019-09-13
Remove previous Round implementation - Closes #3685

### LINKED ISSUE #3685: Remove previous Round implementation

lsilvs:
### What was the problem?
Rounds module not used anymore

### How did I solve it?
Removed `mem_rounds`, `mem_round_snapshot` and `mem_votes_snapshot` tables.
Removed `modules/chain/round` folder
Removed Storage Round entity
Removed StateStore Round entity
Removed `forks_stat` table and its usages.
Removed/Adjust all tests accordingly

### How to manually test it?
Make sure the application starts.
Make sure it sync with testnet at least up to height 20000.
All tests should be green

### Review checklist

- [ ] The PR resolves #3685


framework/test/mocha/common/modules_loader.js
```diff
@@ -87,26 +86,8 @@ const modulesLoader = new function() {
...
-							new Account(scope.components.storage, scope.components.logger);
-
-							return waterCb();
-						},
-					],
-					() => {
-						new Logic(scope.ed, this.transactions, cb);
-					},
-				);
+				new Logic(scope.ed, this.transactions, cb);
```
michielmulders: Do we need a switch statement here as there is only one real option 'Block' and the 'default'?

framework/src/modules/chain/blocks/verify.js
```diff
@@ -65,14 +57,8 @@ class BlocksVerify {
 	}
 
 	async verifyBlockForger(block) {
-		// Check if block was generated by the right active delagate. Otherwise, fork 3
-		// DATABASE: Read only to mem_accounts to extract active delegate list
-		try {
-			await this.dposModule.verifyBlockForger(block);
-		} catch (error) {
-			this.roundsModule.fork(block, 3);
```
SargeKhan: Do we do this somewhere else now?

framework/test/mocha/integration/delegates/synchronous_tasks.js
```diff
@@ -45,8 +45,6 @@ describe.skip('integration test (delegates) - synchronous tasks', () => {
 			durationMs = intervalMs + 1;
 
 			before(done => {
-				library.modules.rounds.onBlockchainReady =
```
SargeKhan: Dpos module implements a replacement function event for `onBlockchainReady`?

lsilvs: It was only used to return whether the Rounds module was loaded or not: `__private.loaded = true;`

framework/src/modules/chain/blocks/verify.js
lsilvs: I don't think so. It will be implemented as part of fork choice rule? @shuse2

framework/test/mocha/common/modules_loader.js
lsilvs: Not really. Will replace by if/else ;)

framework/test/mocha/integration/delegates/synchronous_tasks.js
SargeKhan: It was an event. Previously, (perhaps long time ago) this event was listened for in other parts of the application. For example, the application waiting for this event before accepting HTTP requests. Is that no longer the case?

framework/src/modules/chain/blocks/verify.js
shuse2: we don't need to do this `fork` record anymore because it's different fork choice rule and we can read it from the logs

SargeKhan: We maintained a database table for forks encountered by a node to gauge the health of the network. Do we no longer need it anymore? If so, do we/did we remove the corresponding table as well?

shuse2: We definitely should log the fork choices clearly, but I don't think we need to save it to the database, and it was caused the problem, too.
Let's remove the corresponding table as well

lsilvs: Will remove the table `forks_stat` and its usage as well ;)

framework/test/mocha/integration/delegates/synchronous_tasks.js
lsilvs: I couldn't find any use of it anymore. I reckon it's safe to remove it.

pablitovicente: Yes, I think it's safe to remove if I recall correctly this was a left over from tech-debt effort and was left there to be removed.

framework/src/modules/chain/dpos/delegates_info.js
```diff
@@ -219,7 +219,9 @@ class DelegatesInfo {
 			// summedRound always returns 101 delegates,
 			// that means there can be recurring public keys for delegates
 			// who forged multiple times.
-			const [summedRound] = await this.storage.entities.Round.summedRound(
+			const [
+				summedRound,
+			] = await this.storage.entities.RoundDelegates.summedRound(
```
yatki: ❤️

framework/test/mocha/unit/modules/chain/components/storage/entities/round_delegates.js
```diff
@@ -0,0 +1,85 @@
...
+	});
+
+	beforeEach(() => seeder.seed(storage));
+
+	afterEach(() => {
+		sinonSandbox.restore();
+		return seeder.reset(storage);
+	});
+
+	describe('summedRound()', () => {
```
yatki: I see that it's maybe more straight forward to write tests in mocha. But would be better if we could write this tests in jest. Maybe create an issue for later to migrate it. Or if it's not a big effort do it in this PR ?

framework/test/mocha/unit/modules/chain/components/storage/entities/round_delegates.js
```diff
@@ -0,0 +1,85 @@
...
+			expect(result).to.have.lengthOf(1);
+			expect(result[0]).to.have.all.keys('fees', 'rewards', 'delegates');
+			expect(result[0].rewards).to.be.an('array');
+			expect(result[0].delegates).to.be.an('array');
+			expect(result[0].fees).to.be.eql(
+				new BigNum(computedBlocks[0].totalFee)
+					.plus(computedBlocks[1].totalFee)
+					.toString(),
+			);
+			expect(result[0].rewards).to.be.eql(computedBlocks.map(b => b.reward));
```
yatki: this assertion ensures that rewards is an array so `to.be.an('array')` assertions are redundant.

yatki:
@lsilvs actually this pr was closing 3686 it seems title is wrong ?

yatki:
hmm since the connected issue was closed cannot connect the issue to the pr with zenhub anymore. anyway ignore this comment :D

framework/test/mocha/unit/modules/chain/components/storage/entities/round_delegates.js
lsilvs: In this case I just renamed this file from `rounds.js` to `round_delegates.js` :)

framework/test/mocha/unit/modules/chain/components/storage/entities/round_delegates.js
lsilvs: Same as above. Just file renaming here

PR #4272, opened 2019-09-17
When propagating a block with p2p.send, the block is more likely to reach inbound peers - Closes #4271

### LINKED ISSUE #4271: When propagating a block with p2p.send, the block is more likely to reach inbound peers
### Expected behavior

When using `p2p.send()`, it should reach outbound and inbound peers with equal probability no matter what the proportion of max inbound and max outbound peers is.

### Actual behavior

When propagating a block with `p2p.send(...)`, the block is more likely to reach inbound peers than outbound since each node has more inbound slots by default.

2.3

jondubois:
### What was the problem?

Send was not selecting an even proportion of inbound vs outbound peers. The ratio of max inbound connections vs max outbound connections was skewing the proportion of peers who received a block in favor of peers which consume more inbound connections (old peers).

### How did I solve it?

- Add a new `kind` property to each PeerInfo object which is passed to the request and send selection functions. The `kind` can be either `inbound` or `outbound`.
- Modified the default selection function to select half of its peers from inbound list and half from the outbound list.
- Lowered the `DEFAULT_SEND_PEER_LIMIT` to 24 (instead of 25; this was done to get an even number of inbound and outbound peers).

### How to manually test it?

- A new unit test was added to test the selection function.
- A new integration test was added to test that new `kind` property is present on the PeerInfo objects which are passed to the selectForSend and selectForRequest functions.

`npm test`

### Review checklist

- [ ] The PR resolves #4271 

### DIFF
--- a/elements/lisk-p2p/src/peer_pool.ts
+++ b/elements/lisk-p2p/src/peer_pool.ts
@@ -118,6 +118,9 @@ export const MAX_PEER_DISCOVERY_PROBE_SAMPLE_SIZE = 100;
 export const EVENT_REMOVE_PEER = 'removePeer';
 export const INTENTIONAL_DISCONNECT_STATUS_CODE = 1000;
 
+// TODO: Move to events.ts.
+export const EVENT_FAILED_TO_SEND_MESSAGE = 'failedToSendMessage';
+
 // TODO: Move these to constants.ts
 export const PEER_KIND_OUTBOUND = 'outbound';
 export const PEER_KIND_INBOUND = 'inbound';
@@ -378,7 +381,11 @@ export class PeerPool extends EventEmitter {
 
 		selectedPeers.forEach((peerInfo: P2PDiscoveredPeerInfo) => {
 			const selectedPeerId = constructPeerIdFromPeerInfo(peerInfo);
-			this.sendToPeer(message, selectedPeerId);
+			try {
+				this.sendToPeer(message, selectedPeerId);
+			} catch (error) {
+				this.emit(EVENT_FAILED_TO_SEND_MESSAGE, error);
+			}
 		});
 	}
 
--- a/elements/lisk-p2p/src/peer_pool.ts
+++ b/elements/lisk-p2p/src/peer_pool.ts
@@ -118,6 +118,10 @@ export const MAX_PEER_DISCOVERY_PROBE_SAMPLE_SIZE = 100;
 export const EVENT_REMOVE_PEER = 'removePeer';
 export const INTENTIONAL_DISCONNECT_STATUS_CODE = 1000;
 
+// TODO: Move these to constants.ts
+export const PEER_KIND_OUTBOUND = 'outbound';
+export const PEER_KIND_INBOUND = 'inbound';
+
 export enum PROTECTION_CATEGORY {
 	NET_GROUP = 'netgroup',
 	LATENCY = 'latency',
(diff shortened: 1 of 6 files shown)


elements/lisk-p2p/test/unit/peer_selection.ts
```diff
@@ -14,15 +14,60 @@
...
+		const nodeInfo: P2PNodeInfo = {
+			height: 545777,
+			nethash: '73458irc3yb7rg37r7326dbt7236',
+			os: 'linux',
+			version: '1.1.1',
+			protocolVersion: '1.1',
+			wsPort: 5000,
+		};
+
+		it('returned array should contain an even number of inbound and outbound peers', () => {
```
mitsujutsu: `it should return an array containing...`

elements/lisk-p2p/src/peer_selection.ts
```diff
@@ -63,8 +63,23 @@ export const selectPeersForRequest = (
 
 export const selectPeersForSend = (
 	input: P2PPeerSelectionForSendInput,
-): ReadonlyArray<P2PDiscoveredPeerInfo> =>
-	shuffle(input.peers).slice(0, input.peerLimit);
+): ReadonlyArray<P2PDiscoveredPeerInfo> => {
```
ishantiw: What happens if there only 10 outbound connections? even though we have more inbound peers that we can send to, we will still send to (5 + 12) peers.
Do you think we should check at the end if we have all 24 peers to send? and maybe select more peers from other `kind` to fill it up.

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -118,6 +118,10 @@ export const MAX_PEER_DISCOVERY_PROBE_SAMPLE_SIZE = 100;
 export const EVENT_REMOVE_PEER = 'removePeer';
 export const INTENTIONAL_DISCONNECT_STATUS_CODE = 1000;
 
+// TODO: Move these to constants.ts
```
diego-G: Why don't we make it now?

elements/lisk-p2p/src/peer_selection.ts
```diff
@@ -63,8 +63,23 @@ export const selectPeersForRequest = (
...
+	const shuffledPeers = shuffle(input.peers);
+	// tslint:disable: no-magic-numbers
+	const halfPeerLimit = Math.round((input.peerLimit as number) / 2);
+
+	// TODO: Get outbound string from constants.ts file.
+	const selectedOutboundPeers = shuffledPeers
+		.filter((peerInfo: P2PDiscoveredPeerInfo) => peerInfo.kind === 'outbound')
```
diego-G: I recommend to follow the pattern we introduced in `getPeers()` from PeerPool. We use `typeOf` and `instanceOf`. That way we don't have to create _kind_ constants.

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -353,9 +363,11 @@ export class PeerPool extends EventEmitter {
 	}
 
 	public send(message: P2PMessagePacket): void {
-		const listOfPeerInfo = [...this._peerMap.values()].map(
-			(peer: Peer) => peer.peerInfo as P2PDiscoveredPeerInfo,
-		);
+		const listOfPeerInfo = [...this._peerMap.values()].map((peer: Peer) => ({
+			...(peer.peerInfo as P2PDiscoveredPeerInfo),
+			kind:
```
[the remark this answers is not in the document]
jondubois: I was going to call it type but we have a function on the peerPool called `getPeersByKind(inboundOrOutbound)`.

elements/lisk-p2p/src/peer_selection.ts
jondubois: The object provided here is of type `P2PPeerInfo/P2PDiscoveredPeerInfo` so we can't use `instanceof`. Another reason for this which we discussed in the meeting is that in the near future we could also add this `kind` property to each P2PPeerInfo object which is returned from the `p2p.getConnectedPeers()` method.

elements/lisk-p2p/src/peer_selection.ts
jondubois: Good point. I guess we can add a step before this logic to check if `input.peers.length` is less than `input.peerLimit`; if so, then we just return all of them.

jondubois: In that case, it would send to (10 + 12 peers) because `halfPeerLimit` will be 12. Still 2 short of 24 though so I agree we should pad the list with peers from the other kind.

elements/lisk-p2p/src/peer_pool.ts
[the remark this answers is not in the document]
diego-G: Type would be misleading in Typescript. That was the reason to use `kind`.

PR #4273, opened 2019-09-17
Fix the expectation of the signature response - Closes #4268

### LINKED ISSUE #4268: Remove loading signatures and providing signature endpoint
### Expected behavior
Since we don't manage signature outside of the transaction pool, we don't need to separately provide the signature.
Therefore,

getSignature and getSignature from the network should be removed.

### Actual behavior
It is expecting to have `res[0].transactionId`.
This bug seems to be true since 1.x
https://github.com/LiskHQ/lisk-sdk/blob/1dd941cce6bd7a9cf7fbc3073236f2aa8949d362/framework/src/modules/chain/submodules/loader.js#L227-L231

Also, since signature is provided with transaction object, it always result in error (existing signature)

### Steps to reproduce
Start the node while network having signature

1.x-

shuse2:
### What was the problem?
Response handling of the signature was incorrect

### How did I solve it?
Correct the signature response format.

### How to manually test it?
Connect to the network with pending multisignature transaction

### Review checklist

- [ ] The PR resolves #4268 

### DIFF
--- a/framework/src/modules/chain/loader.js
+++ b/framework/src/modules/chain/loader.js
@@ -201,7 +201,7 @@ class Loader {
 				// eslint-disable-next-line no-await-in-loop
 				await this.transactionPoolModule.getTransactionAndProcessSignature({
 					signature,
-					transactionId: signature.transactionId,
+					transactionId: signaturePacket.transaction,
 				});
 			}
 		}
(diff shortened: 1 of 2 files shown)


framework/src/modules/chain/loader.js
```diff
@@ -201,7 +201,7 @@ class Loader {
 				// eslint-disable-next-line no-await-in-loop
 				await this.transactionPoolModule.getTransactionAndProcessSignature({
 					signature,
-					transactionId: signature.transactionId,
+					transactionId: signaturePacket.transaction,
```
ManuGowda: Can you please add a TODO to address the `signaturePacket.transaction` in the future

shuse2:
Closing because this fix is not to the point and not urgent

PR #4289, opened 2019-09-23
Add bft header - Closes #4247

### LINKED ISSUE #4247: Integrate usage of `computeHeadersForNewBlock` on `create` pipeline in BlockProcessorV2
### Description
In order to construct a block object when forging a new block, we need to call the function `computeHeadersForNewBlock` responsible for computing required BFT headers and use the returned values to include them in the block header.

In order to do that, we need to include this function call into `create` pipeline on `block_processor_v2` to it's able to forward those properties to the `_create` private method

### Motivation
When creating a new block, BFT properties are required to be computed and included in the block header.

### Acceptance Criteria
- All new blocks created by Block Processor V2 need to include `maxHeightPreviouslyForged` and `prevotedConfirmedUptoHeight` as part of their headers. This needs to be integration tested.

shuse2:
### What was the problem?
BFT header was not added during the creation of block

### How did I solve it?
Add the header while creating the block

### How to manually test it?
Start the node with devnet setting, and observe block header changes after 101 blocks

### Review checklist

- [ ] The PR resolves #4247 


framework/src/modules/chain/chain.js
```diff
@@ -211,7 +211,6 @@ module.exports = class Chain {
 			});
 
 			this.logger.info('Modules ready and launched');
-			await this.bft.init();
```
lsilvs: Unit test still expecting this:

```js
expect(chain.bft.init).to.have.been.calledOnce;
```

framework/src/modules/chain/processor/processor.js
```diff
@@ -65,7 +65,10 @@ class Processor {
 		await this._processGenesis(genesisBlock, blockProcessor, {
 			skipSave: false,
 		});
-		await blockProcessor.init.run();
+		await this.blocksModule.init();
+		for (const processor of Object.values(this.processors)) {
+			await processor.init.run();
+		}
```
limiaspasdaniel: Are these changes necessary under the scope of this PR?

shuse2: that's a good point. It was not necessary, but at the same time, bft was initialized from root, but blocks were initialized in each. (which should be opposite) so I just fixed it...

shuse2: Moved to #4297

PR #4290, opened 2019-09-23
Remove signature loading - Closes #4268

### LINKED ISSUE #4268: Remove loading signatures and providing signature endpoint

shuse2:
### What was the problem?
On the startup, signature endpoint was called, but it was never successful.
When signature is received from the network, it will be injected in the transaction.
Therefore, when we start up the node and get the transactions, signatures will be there already if it was propagated.

### How did I solve it?
Remove the signatures fetch on startup.

### How to manually test it?
Observe no error while connecting to the network where it has the multisignature in the pool.

### Review checklist

- [ ] The PR resolves #4268 

### DIFF
--- a/framework/src/modules/chain/chain.js
+++ b/framework/src/modules/chain/chain.js
@@ -564,7 +564,7 @@ module.exports = class Chain {
 	}
 
 	_startLoader() {
-		this.loader.loadTransactions();
+		this.loader.loadUnconfirmedTransactions();
 		if (!this.options.syncing.active) {
 			return;
 		}
--- a/framework/src/modules/chain/chain.js
+++ b/framework/src/modules/chain/chain.js
@@ -564,7 +564,7 @@ module.exports = class Chain {
 	}
 
 	_startLoader() {
-		this.loader.loadTransactionsAndSignatures();
+		this.loader.loadTransactions();
 		if (!this.options.syncing.active) {
 			return;
 		}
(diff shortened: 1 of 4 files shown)


framework/src/modules/chain/chain.js
```diff
@@ -564,7 +564,7 @@ module.exports = class Chain {
 	}
 
 	_startLoader() {
-		this.loader.loadTransactionsAndSignatures();
+		this.loader.loadTransactions();
```
lsilvs: Would be nice to take the opportunity and give this function a better name like `loadPendingTransactions` or something like that.

framework/src/modules/chain/chain.js
```diff
@@ -564,7 +564,7 @@ module.exports = class Chain {
 	}
 
 	_startLoader() {
-		this.loader.loadTransactionsAndSignatures();
+		this.loader.loadUnconfirmedTransactions();
```
lsilvs: ❤️

PR #4292, opened 2019-09-23
Full peers list growing excessively large (> 100K) - Closes #4279

### LINKED ISSUE #4279: Full peers list growing excessively large (> 100K)
### Expected behavior

Node should store a reasonably sized List of peers.
 

### Actual behavior

Nodes on several versions are storing over 100k peer entiries, in their sets, many of which are even doubles due to disconnecting and reconnecting activity of attackers.

### Steps to reproduce

It appears these sets are also growing rapidly still. Here are the results of our open API nodes that have become sick after attaining a peers set this large. The issue can be fixed by rebuilding the node in most cases but still this should not be happening. I've also included other (unknown owner) open API nodes' results on testnet with those IPs half censored to demonstrate which versions this appears to be affecting.

https://test-01.liskapi.io/api/peers?limit=100&offset=100

{"meta":{"offset":100,"limit":100,"count":111040}, "version":"2.1.3-rc.0" (12 hours ago)

{"meta":{"offset":100,"limit":100,"count":140157},"version":"2.1.3-rc.0" (Most recent call)

https://test-02.liskapi.io/api/peers?limit=100&offset=100

{"meta":{"offset":100,"limit":100,"count":141534},"data":[{"version":"2.1.3-rc.0", }
(This one also expanded approx 20k or more over night.

Here are some other pubic API nodes results

{"meta":{"offset":0,"limit":1,"count":104920},"data":[{" version":"2.0.1-rc.0"

{"meta":{"offset":0,"limit":1,"count":121263},"data":[{"version":"2.1.3-rc.0"

{"meta":{"offset":0,"limit":1,"count":127036},"data":[{""version":"2.1.1-rc.0"

{"meta":{"offset":0,"limit":1,"count":120085},"data":[{"version":"2.1.3-rc.0"

{"meta":{"offset":0,"limit":1,"count":141497},"data":[{"version":"2.1.3-rc.0"

{"meta":{"offset":0,"limit":1,"count":125670},"data":[{"version":"2.1.3-rc.0"

For certain 2.0.1, 2.1.2, and 2.1.3. I do also confirm finding a 2.1.2 version nodewith the problem.

Special Thanks to Przemer and LiskPoland.PL for playing a major part in this report.

jondubois:
### What was the problem?

The peer list was growing too big. There was an issue in the PeerBook bucket management which caused PeerInfo eviction to not work correctly in some situations; an expiry issue was causing P2PPeerInfo objects to accumulate faster than old ones were evicted.

### How did I solve it?

- Wrote a unit test to reproduce the issue.
- Perform time-based eviction on a per-bucket basis to guarantee that P2PPeerInfo instances cannot accumulate faster than old ones are evicted. This fix decreases expiry precision slightly but should improve performance.
- Made the getBucket function support customisable bucket count (this is useful for the unit tests but it doesn't affect current functionality).
- addPeer method now returns an array of evicted peers instead of just one peer because time-based expiry can cause multiple peers to be evicted.
- Cleaned up unit test boilerplate.

### How to manually test it?

`npm run test:unit`

### Review checklist

- [ ] The PR resolves #4279


elements/lisk-p2p/src/utils/misc.ts
```diff
@@ -126,10 +124,10 @@ export const getBucket = (options: {
 	readonly secret: number;
 	readonly peerType: PEER_TYPE;
 	readonly targetAddress: string;
+	readonly bucketCount: number;
 }): number => {
-	const { secret, targetAddress, peerType } = options;
+	const { secret, targetAddress, peerType, bucketCount } = options;
```
ishantiw: I'm not sure if it works fine with any value of bucket count because so far we are using `64` and `128`. Maybe we can cover it up in unit tests?

jondubois: I already updated the unit tests to test this and removed the stubs.

elements/lisk-p2p/src/peer_directory/new_list.ts
```diff
@@ -50,54 +50,54 @@ export class NewList extends PeerList {
 	}
 
 	// Extend eviction of NewPeers
-	public evictPeer(bucketId: number): CustomPeerInfo {
+	public evictPeers(bucketId: number): ReadonlyArray<CustomPeerInfo> {
```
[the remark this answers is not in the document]
jondubois: Actually yes, removing only one is better. If all the peers just happen to all expire at the same time it would be a problem.

elements/lisk-p2p/src/utils/misc.ts
ishantiw: I meant, tests with different values of `bucketCount` and see if the bucket number is returned within the range. I only see tests covering for `MAX_TRIED_BUCKETS` which is `128` Let's say having a test if choose a `bucketCount` of 12 and see if `getBucket` returns a number between 0 and 11

elements/lisk-p2p/src/utils/misc.ts
```diff
@@ -126,10 +124,10 @@ export const getBucket = (options: {
 	readonly secret: number;
```
ishantiw: Can we change the name to `getBucketId`. `getBucket`? It's a bit misleading as it only returns a mere number.

elements/lisk-p2p/src/peer_directory/new_list.ts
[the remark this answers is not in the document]
ishantiw: Yes, it will be weird if one new peer deletes all the peers in a bucket if all of them are 30 days older. It should be a peer for a peer strategy.

elements/lisk-p2p/src/utils/misc.ts
mitsujutsu: I think it will work. It seems @jondubois added `bucketCount` for unit test purposes only, and in the actual code it will always be `64` and `124`. The edge case in which it will not work is if `0` is passed.

ishantiw: If we create a new peer list and use a value other than 64 or 128 and if the function return a value outside the given `bucketCount` then it could be a problem. If we are giving `bucketCount` required value to this function then we have to make sure it should not only work for limited values. It will be nice if we can have a unit test covering a few random values of `bucketCount`.

mitsujutsu: @ishantiw It's a good point I don't see the need to change `secondMod` to `bucketCount` as the tests only used `MAX_TRIED_BUCKETS` and in the code it will always be `64` or `128`. This is why I originally had it as using `secondMod`.  However I don't think it's necessary to add more unit tests to check if other values would work.

jondubois: I made the secondMod configurable because the higher level PeerBucket logic assumed that the getBucket function allowed this. It could be useful in the future if we want to make the bucket values configurable.

Right now the distribution of the `getBucket` function is tested with two different numbers: 64 and 128; if it works with both of those then it will work with any other value because the second modulo expression doesn't treat those two numbers as special. I agree with @mitsujutsu that the main special case is 0.

elements/lisk-p2p/src/utils/misc.ts
```diff
@@ -122,14 +120,14 @@ export const getNetgroup = (address: string, secret: number): number => {
...
+export const getBucketId = (options: {
 	readonly secret: number;
 	readonly peerType: PEER_TYPE;
 	readonly targetAddress: string;
+	readonly bucketCount: number;
 }): number => {
-	const { secret, targetAddress, peerType } = options;
+	const { secret, targetAddress, peerType, bucketCount } = options;
 	const firstMod = peerType === PEER_TYPE.NEW_PEER ? BYTES_16 : BYTES_4;
-	const secondMod = peerType === PEER_TYPE.NEW_PEER ? BYTES_128 : BYTES_64;
```
shuse2: shouldn't this be used instead of bucketCount?

elements/lisk-p2p/src/peer_directory/new_list.ts
```diff
@@ -70,34 +67,28 @@ export class NewList extends PeerList {
 		// Second eviction strategy: Default eviction based on base class
 		return this.evictRandomlyFromBucket(bucketId);
 	}
+
 	// Evict a peer when a bucket is full based on the time of residence in a peerlist
-	private _evictionBasedOnTimeInBucket(
+	private _evictPeerBasedOnTimeInBucket(
```
shuse2: isn't `Peer` kind of redundant...? I think only thing we can evict is `Peer` 😉

elements/lisk-p2p/src/peer_directory/new_list.ts
```diff
@@ -70,34 +67,28 @@ export class NewList extends PeerList {
...
-					evictedPeer = peer;
-				}
-			});
-		});
+		const bucket = this.peerMap.get(bucketId);
+		if (!bucket) {
+			return undefined;
+		}
+
+		for (const [peerId, peer] of bucket) {
```
shuse2: 👍

elements/lisk-p2p/src/utils/misc.ts
jondubois: The PeerList expects the `getBucket` function to be able to work with a variable number of buckets, `secondMod` represents the number of buckets; there is no need to couple the number of buckets to the kind of peer list (this could be configured later). The `firstMod` part is not really necessary; the bucket system could work fine without it but it helps to make sure that the two lists (tried vs new) map to different buckets compared to each other (to make it harder to predict one list from the other).

elements/lisk-p2p/src/peer_directory/new_list.ts
jondubois: `_evictPeerBasedOnTimeInBucket` was chosen to be consistent with the `evictPeer` method.

elements/lisk-p2p/src/peer_directory/peer_list.ts
```diff
@@ -28,8 +28,7 @@ export interface CustomPeerInfo {
 
 export interface AddPeerOutcome {
 	readonly success: boolean;
-	readonly isEvicted: boolean;
-	readonly evictedPeer?: P2PPeerInfo;
+	readonly evictedPeer: P2PPeerInfo | undefined;
```
diego-G: I prefer `isEvicted`. The new alternative is redundant -> `AddPeerOutcome.evictedPeer`.

elements/lisk-p2p/src/peer_directory/new_list.ts
```diff
@@ -50,18 +50,15 @@ export class NewList extends PeerList {
 	}
 
 	// Extend eviction of NewPeers
-	public evictPeer(bucketId: number): CustomPeerInfo {
+	public evictPeer(bucketId: number): CustomPeerInfo | undefined {
```
diego-G: This `undefined` doesn't seem ok to me. We should always evict at least a peer when we call this method, as per its name states. I suggest to split the action in two:
- when adding a peer, pick the bucket and check if it's full
-  if it's full, then evict a peer or several

ishantiw: Yes, this function should always evict some peer to make space for the new one. For `newPeers`, we have two strategies and the second strategy will always return one evictedPeer. So overall, this function should always return one evictedPeer

elements/lisk-p2p/src/peer_directory/new_list.ts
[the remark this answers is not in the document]
diego-G: `Any address` means to me to expire every address. Once you start cleaning a bucket, I'd clean it completely. Please @janhack, could you shed us some light here?

elements/lisk-p2p/src/peer_directory/peer_list.ts
```diff
@@ -190,23 +185,23 @@ export class PeerList {
...
-	protected evictRandomlyFromBucket(bucketId: number): CustomPeerInfo {
-		const peerList = this.peerMap.get(bucketId);
-		if (!peerList) {
-			throw new Error(`No Peers exist for bucket Id: ${bucketId}`);
+	protected evictRandomlyFromBucket(
+		bucketId: number,
+	): CustomPeerInfo | undefined {
+		const bucket = this.peerMap.get(bucketId);
+		if (!bucket) {
+			return undefined;
```
ishantiw: I think we should throw an error in this case because it is not a possible scenario and make this function always return an evictedPeer. By code, a bucket must exist only then the eviction can be triggered but TS always think `Map.get` can return `undefined` which is not true in this case

jondubois: It would be a problem if evict was called on an empty bucket.

Before, if that happened, it would throw an error https://github.com/LiskHQ/lisk-sdk/pull/4292/files#diff-29e3f653688df876f4d60b83818a6cd8L200 - But that error was not handled anywhere so it adds risk.

If the function tried to evict a specific peer based on a peer id, then it would make sense to throw an error but in this case we evict based on the bucket ID so it's OK to make that operation fail-safe if the bucket is empty.

ishantiw: I agree that we need to then handle the error everywhere. I have one more solution to make this function always return a peer.
As we already check if the bucket exists or not in `addPeer` function, We can pass `bucket` instead of `bucketId` from `addPeer` function to the `evictPeer` function. I think we can reduce some computation as well to compute the bucketId again and again

elements/lisk-p2p/src/peer_directory/new_list.ts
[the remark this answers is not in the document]
janhack: The main purpose of the `newPeers` collection is to store a reasonable number of possible peers (<=32 per bucket, <=4096 in total). This collection should not grow over these limits (to avoid spamming of peers or poising the `newPeers` with addresses controlled by an adversary), but also not to small so a node has sufficient peers to try as backup. I think the eviction should only evict peers if necessary (i.e., the bucket is full) and only as many peers as necessary (only one peer if one peer needs to be added).

A different topic is, if we should maybe introduce some `newPeers` housekeeping. This could mean that we go over all addresses in `newPeers`, make a connection attempt (to ensure that there is actually a Lisk Node at this address) and update a timestamp if the attempt is successful. As part of this housekeeping we could also remove peers if they have not been reachable for a certain period of time. I would think that it would be better to do this separately from the peer eviction.

jondubois:
I've made some changes in a way which is consistent with the other parts of the PeerBook logic. The focus of this PR is to fix the PR overflow issue.

elements/lisk-p2p/src/peer_directory/peer_list.ts
```diff
@@ -147,16 +147,16 @@ export class PeerList {
 
 		if (!bucket) {
 			return {
+				success: false,
 				wasPeerAdded: false,
```
ishantiw: I thought `success` value already implies if a peer was added.

elements/lisk-p2p/src/peer_directory/peer_list.ts
```diff
@@ -147,16 +147,16 @@ export class PeerList {
...
-				wasPeerEvicted: false,
 				evictedPeer: undefined,
```
ishantiw: If make this optional then we don't need to explicitly assign `undefined`

elements/lisk-p2p/src/peer_directory/peer_list.ts
jondubois: I changed it back to `success` in a new commit but the definition of success is slightly different. Success doesn't mean that peer was added because if peer was added but it failed to evict while bucket was full, then that's also not success.

elements/lisk-p2p/src/peer_directory/tried_list.ts
```diff
@@ -111,7 +110,6 @@ export class TriedList extends PeerList {
 		};
 
 		bucket.set(incomingPeerId, updatedTriedPeerInfo);
```
mitsujutsu: Not directly related to this PR but you can simply do `bucket.set(incomingPeerId, { ...foundPeer, numOfConnectionFailures: numOfConnectionFailures + 1})`. This way you don't need all references to `peerInfo`, `dateAdded` and `updatedPeerInfo` above

elements/lisk-p2p/src/peer_directory/peer_list.ts
ishantiw: I think failed to evict is an exception and should throw en error but since we are not throwing errors and handling them anywhere we can keep it as it is

elements/lisk-p2p/src/utils/misc.ts
mitsujutsu: A reminder to add the check for `0` `bucketCount` value as discussed above

elements/lisk-p2p/src/peer_directory/peer_list.ts
```diff
@@ -28,8 +28,8 @@ export interface CustomPeerInfo {
 
 export interface AddPeerOutcome {
 	readonly success: boolean;
-	readonly isEvicted: boolean;
-	readonly evictedPeer?: P2PPeerInfo;
+	readonly wasPeerAdded: boolean;
```
mitsujutsu: prefer `isAdded`

elements/lisk-p2p/src/peer_directory/peer_list.ts
mitsujutsu: In that case should we throw an error and handle as opposed to adding another `boolean` outcome?

jondubois:
@mitsujutsu I implemented all your recommendation except the last one about success: false. There was a previous discussion about this. I was going to make it throw errors everywhere but this was doing to be a bigger change (out of scope of this PR) so we agreed to make everything failsafe for now for consistency and to keep this PR focused on the overflow issue. We can fix it in a separate issue. @diego-G  has a proposal which we can look into.

Issue #2954, opened 2019-02-25
Remove unwanted old database functions


ManuGowda:
### Expected behavior
Remove unwanted old database related logic which is no longer required after storage changes.
### Actual behavior
There are few old database logic functions such as
https://github.com/LiskHQ/lisk/blob/ebfc1349419c6263c26b6c6d59cff7b97223e90f/framework/src/modules/chain/logic/transaction.js#L1221
### Which version(s) does this affect? (Environment, OS, etc...)

**Grooming Meeting Notes**
Remove dbRead and storageRead and put it in other blocks logic related file or merge the code.

lsilvs:
It can only be done after https://github.com/LiskHQ/lisk/issues/2424 as `modules.blocks.process.loadBlocksFromPeer` still uses it.

REF.: https://github.com/LiskHQ/lisk/issues/2424#issuecomment-467018480

PR #4293, opened 2019-09-23
Remove inefficient sync - Closes #2424 and  #2954

### LINKED ISSUE #2424: Inefficient blocks' data transmitting over the network
### Expected behavior
We should pass a proper block object through the network instead of raw SQL rows from `full_blocks_list` view.
### Actual behavior
We have the function `modules.blocks.process->loadBlocksFromPeer` which is called as part of the sync process to ask a remote peer for blocks and then process them.

The function is calling `peer.rpc.blocks` and this on remote peer calls `modules.transport.shared.blocks` which calls `modules.blocks.utils.loadBlocksData`. This function executes a query `blocks.loadBlocksData` (this query is using `full_blocks_list` view to retrieve the data) and returns all the results from this query directly.

Because of this logic node is passing raw SQL data through the network, instead of blocks objects. For example, if blocks contain 25 transactions there will be 25 rows from the database, all of them containing the block data itself, which is redundant and inefficient.

Then on the node that asked for blocks, we're calling `modules.blocks.utils.readDbRows` to create proper block object from the response. This is also not the best approach, as this function shouldn't be used in this content.

### Steps to reproduce
N/A
`>= 1.0.0`

shuse2:
### What was the problem?
blocks endpoint was returning in block in table format instead of JSON format.

### How did I solve it?
Change the blocks endpoint to return JSON format

### How to manually test it?
1. Run one node with forging
2. wait for a while to build up the height
3. sync with this node

### Review checklist

- [ ] The PR resolves #2424 and #2954 

### DIFF
--- a/framework/src/modules/chain/blocks/block_v2.js
+++ b/framework/src/modules/chain/blocks/block_v2.js
@@ -19,43 +19,6 @@ const { omitBy, isNull } = require('lodash');
 const BigNum = require('@liskhq/bignum');
 
 /**
- * Creates block object based on raw data.
- *
- * @param {Object} raw
- * @returns {null|block} Block object
- * @todo Add description for the params
- */
-const dbRead = raw => {
-	if (!raw.b_id) {
-		return null;
-	}
-	const block = {
-		id: raw.b_id,
-		version: parseInt(raw.b_version, 10),
-		timestamp: parseInt(raw.b_timestamp, 10),
-		height: parseInt(raw.b_height, 10),
-		maxHeightPreviouslyForged: parseInt(raw.b_maxHeightPreviouslyForged, 10),
-		prevotedConfirmedUptoHeight: parseInt(
-			raw.b_prevotedConfirmedUptoHeight,
-			10,
-		),
-		previousBlock: raw.b_previousBlock,
-		numberOfTransactions: parseInt(raw.b_numberOfTransactions, 10),
-		totalAmount: new BigNum(raw.b_totalAmount),
-		totalFee: new BigNum(raw.b_totalFee),
-		reward: new BigNum(raw.b_reward),
-		payloadLength: parseInt(raw.b_payloadLength, 10),
-		payloadHash: raw.b_payloadHash,
-		generatorPublicKey: raw.b_generatorPublicKey,
-		generatorId: getAddressFromPublicKey(raw.b_generatorPublicKey),
-		blockSignature: raw.b_blockSignature,
-		confirmations: parseInt(raw.b_confirmations, 10),
-	};
-	block.totalForged = block.totalFee.plus(block.reward).toString();
-	return block;
-};
-
-/**
  * Creates block object based on raw database block data.
  *
  * @param {Object} raw Raw database data block object
@@ -98,6 +61,5 @@ const storageRead = raw => {
 };
 
 module.exports = {
-	dbRead,
 	storageRead,
 };
--- a/framework/src/modules/chain/blocks/utils.js
+++ b/framework/src/modules/chain/blocks/utils.js
@@ -100,7 +100,6 @@ const parseStorageObjToLegacyObj = block => {
 
 /**
  * Generates a list of full blocks for another node upon sync request from that node, see: modules.transport.internal.blocks.
- * NOTE: changing the original method loadBlocksDataWS() could potentially change behaviour (popLastBlock() uses it for instance)
  * so that's why this new method was added
  * @param {Object} filter - Filter options
  * @param {Object} filter.limit - Limit blocks to amount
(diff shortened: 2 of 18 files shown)


framework/src/modules/chain/blocks/block_v2.js
```diff
@@ -18,43 +18,6 @@ const { getAddressFromPublicKey } = require('@liskhq/lisk-cryptography');
...
 const BigNum = require('@liskhq/bignum');
 
-/**
- * Creates block object based on raw data.
- *
- * @param {Object} raw
- * @returns {null|block} Block object
- * @todo Add description for the params
- */
-const dbRead = raw => {
```
michielmulders: Finally get rid of it! :)

framework/src/modules/chain/transport/transport.js
```diff
@@ -235,27 +234,10 @@ class Transport {
...
-								block,
-								error: e,
-							},
-						);
-					}
-				}
-			});
-
+			const data = await this.blocksModule.loadBlocksFromLastBlockId(
+				query.lastBlockId,
```
michielmulders: I liked the interface where you could pass an object with limit and lastId.
It looks a bit odd now what this `34` refers to?

framework/test/jest/unit/specs/modules/chain/blocks/blocks.spec.js
```diff
@@ -1703,12 +1704,113 @@ describe('blocks', () => {
 		});
 	});
 
-	describe('loadBlocksDataWs', () => {
-		it.todo('a');
-	});
+	describe('loadBlocksFromLastBlockId', () => {
+		describe('when called without lastBlockId', () => {
+			it('should reject with error', async () => {
```
michielmulders: There is no test for checking when called without storage object.

framework/test/jest/unit/specs/modules/chain/blocks/blocks.spec.js
```diff
@@ -1703,12 +1704,113 @@ describe('blocks', () => {
...
+				id: 'block-id',
+			};
+
+			beforeEach(async () => {
+				stubs.dependencies.storage.entities.Block.get.mockResolvedValue([
+					validLastBlock,
+				]);
+			});
+
+			it('should use limit 1 as default', async () => {
```
michielmulders: This is a confusing test. We say `limit 1` however I don't see any reference towards checking if it got called with `limit 1`.

michielmulders: If the storage layer replaces an undefined limit to 1, pls add a comment to clarify this.

framework/src/modules/chain/blocks/utils.js
```diff
@@ -112,66 +31,44 @@ const parseStorageObjToLegacyObj = block => {
...
-		throw new Error('Invalid filter: Received both id and lastId');
+const loadBlocksFromLastBlockId = async (storage, lastBlockId, limit) => {
+	if (!lastBlockId) {
+		throw new Error('lastBlockId needs to be specified');
 	}
-	if (filter.id) {
-		params.id = filter.id;
-	} else if (filter.lastId) {
-		params.lastId = filter.lastId;
+	if (!limit) {
```
lsilvs: What about add a default value and remove this check?

framework/src/modules/chain/blocks/utils.js
```diff
@@ -112,66 +31,44 @@ const parseStorageObjToLegacyObj = block => {
...
-		{ id: filter.lastId || null },
-		{ limit: params.limit },
-		tx,
+	const [lastBlock] = await storage.entities.Block.get(
+		{ id: lastBlockId },
+		{ limit },
 	);
-	if (!rows.length) {
+	if (!lastBlock) {
 		throw new Error('Invalid lastBlockId requested');
```
michielmulders: Would it make sense to also add the `lastBlockId` in the Error message? Can be beneficial info for debugging.

framework/src/modules/chain/blocks/utils.js
```diff
@@ -112,66 +31,44 @@ const parseStorageObjToLegacyObj = block => {
...
 	}
 
 	// Get height of block with supplied ID
-	const rows = await storage.entities.Block.get(
-		{ id: filter.lastId || null },
-		{ limit: params.limit },
-		tx,
+	const [lastBlock] = await storage.entities.Block.get(
+		{ id: lastBlockId },
+		{ limit },
```
lsilvs: Why use limit here if getting by ID?

framework/test/jest/unit/specs/modules/chain/blocks/blocks.spec.js
```diff
@@ -1703,12 +1704,113 @@ describe('blocks', () => {
...
+						extended: true,
+						limit: null,
+						sort: ['height'],
+					},
+				);
+			});
+		});
+
+		describe('when called with invalid lastBlockId', () => {
+			beforeEach(async () => {
```
michielmulders: No need for `beforeEach`, would prefer to add this setup directly to the test under `// Arrange` section.

framework/src/modules/chain/blocks/utils.js
```diff
@@ -112,66 +31,44 @@ const parseStorageObjToLegacyObj = block => {
...
-		{
-			extended: true,
-			limit: null,
-			sort: ['height'],
-		},
-		tx,
-	);
+	const filter = {
+		height_gt: lastBlockHeight,
+		height_lt: fetchUntilHeight,
```
lsilvs: I think you want less than or equal to here

```suggestion
		height_lte: fetchUntilHeight,
```

framework/src/modules/chain/blocks/utils.js
```diff
@@ -112,66 +31,44 @@ const parseStorageObjToLegacyObj = block => {
...
+	};
 
-	let parsedBlocks = [];
-	blockRows.forEach(block => {
-		parsedBlocks = parsedBlocks.concat(parseStorageObjToLegacyObj(block));
+	const blocks = await storage.entities.Block.get(filter, {
+		extended: true,
+		limit: null,
```
lsilvs:
```suggestion
		limit,
```

framework/test/jest/unit/specs/modules/chain/blocks/blocks.spec.js
```diff
@@ -1703,12 +1704,113 @@ describe('blocks', () => {
...
+							extended: true,
+							limit: null,
+							sort: ['height'],
+						},
+					)
+					.mockResolvedValue(validBlocks);
+			});
+
+			it('should use the storage with correct filter', async () => {
+				const blocks = await blocksInstance.loadBlocksFromLastBlockId(
```
michielmulders: We do not check if this first `Block.get` call is actually called in the happy path. Can you add this check? :)
```
const [lastBlock] = await storage.entities.Block.get(
    { id: lastBlockId },
    { limit },
);
```

framework/src/modules/chain/loader.js
```diff
@@ -333,18 +334,19 @@ class Loader {
 	 * @returns {Promise} void
 	 * @todo Add description for the params
 	 */
-	async _getValidatedBlocksFromNetwork(blockRows) {
+	async _getValidatedBlocksFromNetwork(blocks) {
 		const { lastBlock } = this.blocksModule;
 		let lastValidBlock = lastBlock;
 		// TODO: this should be removed and the block should be received from the network using *normal* block property names
```
lsilvs:
```suggestion

```

framework/src/modules/chain/transport/transport.js
```diff
@@ -235,27 +234,10 @@ class Transport {
...
-							{
-								block,
-								error: e,
-							},
-						);
-					}
-				}
-			});
-
+			const data = await this.blocksModule.loadBlocksFromLastBlockId(
```
lsilvs: Param `storage` missing. Maybe use object as param?

framework/test/jest/unit/specs/modules/chain/blocks/blocks.spec.js
shuse2: the API is called without storage. so not possible to have that case 😉

framework/test/jest/unit/specs/modules/chain/blocks/blocks.spec.js
shuse2: in the `blocks. loadBlocksFromLastBlockId`, limit is set to `1` as default params

framework/src/modules/chain/blocks/utils.js
shuse2: actually it is provided in the calling function. maybe we don't need this check at all, but it was just for safety

framework/src/modules/chain/transport/transport.js
shuse2: this function doesn't require storage as input (it's managed in blocks module)

framework/test/jest/unit/specs/modules/chain/blocks/blocks.spec.js
shuse2: I think when we add the similar test case, this would be easier and ensure that stub is reset

michielmulders: It's personal taste! If it's a small stub, I prefer it to be inside the test itself, but the test section below this one is a large stub which makes sense to put in a beforeEach block. For me it's fine to keep this kind of consistency. I'll resolve the comment

shuse2:
@michielmulders it's addressed in the test `should call storage for the lastBlockId`

PR #4304, opened 2019-09-25
Update max height previously forged - Closes #3696

### LINKED ISSUE #3696: Update the block processing to persist maxHeightPreviouslyForged
### Expected behavior

Update the block processing to persist maxHeightPreviouslyForged, this value should be updated at the end of the steps of forging a block. 

This property needs to be updated and saved before broadcasting a block.

> If a delegate forges a block B to be added to the current chain, then B.heightPrevious must be the largest height at which the delegate forging the block previously forged a block on any chain or 0 if the delegate did not forge a block before. To be able to provide this information correctly, even in the case of node crashes, it is important that the largest height of a block previously forged is persistently stored before any forged block is broadcast to the network.

3.0.0

shuse2:
### What was the problem?
- Max height previously forged was not persisted when creating a block
- Wrong maxPreviouslyForgedHeight was used while creating a block

### How did I solve it?
- Add a function to save previously forged height per delegate
- Use the saved value for BFT header

### How to manually test it?
- Start the node and observe chain_meta table that maxHeightPreviouslyForged is saved
- Check that newly created block has the correct previously forged height per delegate

### Review checklist

- [ ] The PR resolves #3696 


framework/src/modules/chain/block_processor_v2.js
```diff
@@ -224,6 +224,13 @@ class BlockProcessorV2 extends BaseBlockProcessor {
 				return this.bftModule.computeBFTHeaderProperties(delegatePublicKey);
 			},
 			(data, bftHeader) => this._create({ ...data, ...bftHeader }),
+			async (data, block) => {
```
michielmulders: Can you add a comment above the create.pipe briefly explaining the three steps that are happening in the create pipeline?

framework/src/modules/chain/bft/bft.js
```diff
@@ -151,11 +152,44 @@ class BFT extends EventEmitter {
...
+	 * Saving a height which delegate last forged. this needs to be saved before broadcasting
+	 * so it needs to be outside of the DB transaction
+	 * @param delegatePublicKey
+	 * @param height
+	 */
+	async saveMaxHeightPreviouslyForged(delegatePublicKey, height) {
+		const previouslyForged = await this._getPreviouslyForgedMap();
+		const updatedPreviouslyForged = {
+			...previouslyForged,
+			[delegatePublicKey]: height,
```
lsilvs: Don't we need to check if `previouslyForged[delegatePublicKey].height` is less than `height` before updating?

shuse2: I think it should always update to the last forging block regardless of height

shuse2: @lsilvs You are right. It's maxPreviouslyForgedHeight so it should check (confirmed with research team)

PR #4308, opened 2019-09-25
Return full common block instead of ID property - Closes #4307

### LINKED ISSUE #4307: Return full common block instead of only ID property
### Description
We need to return the full common block as the response for getHighestCommonBlock Chain action

### Motivation
After reading the LIP-0014 again, we need the `height` property of the highest common block returned by the Chain action mentioned above ([Step 3](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#block-synchronization-mechanism) of Block Synchronization Mechanism). Makes sense then to return the full block and choose which properties to pick from the requester side.

### How to do it
Rename
https://github.com/LiskHQ/lisk-sdk/blob/07f1cb75cdde08acdc40a1e8855c913e84fa63a4/framework/src/modules/chain/chain.js#L310

to `getHighestCommonBlock`

and change this line to return the full block:
https://github.com/LiskHQ/lisk-sdk/blob/07f1cb75cdde08acdc40a1e8855c913e84fa63a4/framework/src/modules/chain/chain.js#L333

Then fix the tests related to this function.

### Acceptance Criteria
- Chain action getHighestCommonBlock returns the full block
- Tests are updated to match new implementation

limiaspasdaniel:
### What was the problem?
Refer to #4307
### How did I solve it?
Returning the full block instead of only its ID.
### How to manually test it?
Run `blocks.spec.js`
### Review checklist

- [ ] The PR resolves #4307

### DIFF
--- a/framework/src/modules/chain/chain.js
+++ b/framework/src/modules/chain/chain.js
@@ -330,7 +330,7 @@ module.exports = class Chain {
 					action.params.ids,
 				);
 
-				return commonBlock || null;
+				return commonBlock;
 			},
 		};
 	}
--- a/framework/src/modules/chain/chain.js
+++ b/framework/src/modules/chain/chain.js
@@ -316,10 +316,13 @@ module.exports = class Chain {
 				if (valid.length) {
 					const err = valid;
 					const error = `${err[0].message}: ${err[0].path}`;
-					this.logger.debug('getHighestCommonBlock request validation failed', {
-						err: error,
-						req: action.params,
-					});
+					this.logger.debug(
+						{
+							err: error,
+							req: action.params,
+						},
+						'getHighestCommonBlock request validation failed',
+					);
 					throw new Error(error);
 				}
 
--- a/framework/src/modules/chain/chain.js
+++ b/framework/src/modules/chain/chain.js
@@ -307,22 +307,19 @@ module.exports = class Chain {
 			}),
 			getLastBlock: async () => this.blocks.lastBlock,
 			blocks: async action => this.transport.blocks(action.params || {}),
-			getHighestCommonBlockId: async action => {
+			getHighestCommonBlock: async action => {
 				const valid = validator.validate(
-					definitions.getHighestCommonBlockIdRequest,
+					definitions.getHighestCommonBlockRequest,
 					action.params,
 				);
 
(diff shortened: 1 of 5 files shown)


framework/src/modules/chain/chain.js
```diff
@@ -307,30 +307,27 @@ module.exports = class Chain {
...
 					const err = valid;
 					const error = `${err[0].message}: ${err[0].path}`;
-					this.logger.debug(
-						'getHighestCommonBlockId request validation failed',
-						{
-							err: error,
-							req: action.params,
-						},
-					);
+					this.logger.debug('getHighestCommonBlock request validation failed', {
```
shuse2: I think syntax is opposite.
It shoukd be 
debug({}, 'message')

framework/test/mocha/unit/modules/chain/chain.js
```diff
@@ -183,10 +183,10 @@ describe('Chain', () => {
 					id: '123',
 					height: '1',
 				});
-				const response = await chain.actions.getHighestCommonBlockId({
+				const response = await chain.actions.getHighestCommonBlock({
 					params: { ids: ['1', '2'] },
 				});
-				expect(response).to.equal('123');
+				expect(response).to.deep.equal({ id: '123', height: '1' });
```
shuse2:
```suggestion
				expect(response).to.eql({ id: '123', height: '1' });
```
Probably it's the same?

limiaspasdaniel: It's the same in Jest, but in Mocha if you do just `equal` it will perform reference comparison and hence fail.

framework/src/modules/chain/chain.js
limiaspasdaniel: True, done

framework/src/modules/chain/chain.js
```diff
@@ -330,7 +330,7 @@ module.exports = class Chain {
 					action.params.ids,
 				);
 
-				return commonBlock ? commonBlock.id : null;
+				return commonBlock || null;
```
lsilvs: Do we need `null` here?

```suggestion
				return commonBlock;
```

PR #4309, opened 2019-09-25
Restore blocks from temp_block table - Closes #3775

### LINKED ISSUE #3775: Blocks need to be restored from temporary table upon failure or application restart
### Expected behavior
Whenever the node crashes or upon failure during one of the sync processes we roll back to the original state by applying the blocks in the temp table and continue the operating of the node.

Functions to be implemented:
- `restoreBlocks()`

1. Read blocks from temp table
2. Apply blocks (blocks verify process - do not care about how it is implemented - just call this)
3. Remove temp blocks after the apply process is successful

### Design For Issue
Place the `restoreBlocks` in `blocks/blocks`.
The synchronizer class should injects `blocks module` as a dependency so we can access `restoreBlocks` inside.

`feature/improve-bft`

### Tasks list
- [ ] Add functionality for applying block from blocks temporary table.
- [ ] Restore blocks from the temporary table when the node restarts.
- [ ] Restore blocks from the temporary table when the synchronization fails.
- [ ] Write integration tests.

michielmulders:
### What was the problem?
No functionality for restoring and re-applying blocks.

### How did I solve it?
`restoreBlocks` util function that can be used inside of BSM or FSM in case we want to restore the original chain.

**Why not implement in synchronizer class?**
The original idea was to work with a base class (-> synchronizer) that holds the common logic for both sync mechanisms. However, now the synchronizer class has been converted in some orchestrator for registering both sync mechanisms and running them, we can not place `restoreBlocks` functionality inside it (creates a circular dependency).

Therefore, a utils file has been created that holds only one util function. We expect more util functions that hold common logic between both sync mechanisms. The disadvantage of this approach currently, we have to inject blocks and processor module to `restoreBlocks`.

### How to manually test it?
Will be implemented with one of the fork choices where the chain has to be stored.
Covered with unit tests.

### Review checklist

- [x] The PR resolves #3775 

### DIFF
--- a/framework/src/modules/chain/processor/processor.js
+++ b/framework/src/modules/chain/processor/processor.js
@@ -184,7 +184,7 @@ class Processor {
 	}
 
 	// processValidated processes a block assuming that statically it's valid
-	async processValidated(block) {
+	async processValidated(block, removeFromTempTable = false) {
 		return this.sequence.add(async () => {
 			this.logger.debug(
 				{ id: block.id, height: block.height },
@@ -194,6 +194,7 @@ class Processor {
 			const blockProcessor = this._getBlockProcessor(block);
 			return this._processValidated(block, lastBlock, blockProcessor, {
 				skipBroadcast: true,
+				removeFromTempTable,
 			});
 		});
 	}
@@ -236,7 +237,7 @@ class Processor {
 		block,
 		lastBlock,
 		processor,
-		{ skipSave, skipBroadcast } = {},
+		{ skipSave, skipBroadcast, removeFromTempTable = false } = {},
 	) {
 		await this.storage.entities.Block.begin('Chain:processBlock', async tx => {
 			await processor.verify.run({
@@ -262,6 +263,10 @@ class Processor {
 				this.channel.publish('chain:processor:newBlock', {
 					block: cloneDeep(block),
 				});
+			}
+			if (removeFromTempTable) {
+				// Remove block from temp tabel
+				await this.blocksModule.removeBlockFromTempTable(block.id);
 			}
 			return block;
 		});
--- a/framework/src/modules/chain/processor/processor.js
+++ b/framework/src/modules/chain/processor/processor.js
@@ -184,7 +184,7 @@ class Processor {
 	}
 
 	// processValidated processes a block assuming that statically it's valid
-	async processValidated(block, removeFromTempTable = false) {
+	async processValidated(block, { removeFromTempTable = false } = {}) {
 		return this.sequence.add(async () => {
 			this.logger.debug(
 				{ id: block.id, height: block.height },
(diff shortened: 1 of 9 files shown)


framework/src/modules/chain/processor/processor.js
```diff
@@ -184,7 +184,7 @@ class Processor {
 	}
 
 	// processValidated processes a block assuming that statically it's valid
-	async processValidated(block) {
+	async processValidated(block, removeFromTempTable = false) {
```
limiaspasdaniel: Can we include this property in an object literal? This way we can name the parameters.

michielmulders: Yes will do, thanks!

framework/src/modules/chain/synchronizer/utils.js
```diff
@@ -0,0 +1,43 @@
...
+ */
+const restoreBlocks = async (blocksModule, processorModule, tx) => {
+	const tempBlocks = await blocksModule.getTempBlocks(tx);
+
+	if (tempBlocks.length === 0) {
+		throw new Error('Temp_block table is empty');
+	}
+
+	for (const block of tempBlocks) {
+		await processorModule.processValidated(block, true);
```
limiaspasdaniel: Need to update this for the new argument :)

framework/src/modules/chain/synchronizer/utils.js
```diff
@@ -0,0 +1,45 @@
...
+ * 3. Each block gets deleted from temp_block table when its being applied
+ *
+ * @param {Object} tx - database transaction
+ * @return {Promise<Boolean>|Error} - returns true when successfully restoring blocks
+ */
+const restoreBlocks = async (blocksModule, processorModule, tx) => {
+	const tempBlocks = await blocksModule.getTempBlocks(tx);
+
+	if (tempBlocks.length === 0) {
+		throw new Error('Temp_block table is empty');
```
shuse2: do we need to throw error in the case it doesn't exist?
Maybe we can just return false and end the execution?

framework/test/jest/unit/specs/modules/chain/synchronizer/utils.spec.js
```diff
@@ -0,0 +1,75 @@
...
+			// Arrange
+			const blocks = [{ id: 'block1' }, { id: 'block2' }];
+			blocksMock.getTempBlocks = jest.fn().mockReturnValue(blocks);
+
+			// Act
+			const result = await restoreBlocks(blocksMock, processorMock, stubs.tx);
+
+			// Assert
+			expect(blocksMock.getTempBlocks).toHaveBeenCalledWith(stubs.tx);
+			expect(processorMock.processValidated).toHaveBeenCalledTimes(2);
```
shuse2: probably these cases are worth to have another `it` phrase to describe?

michielmulders: Agree, makes sense, thanks!

PR #4311, opened 2019-09-25
Update inbound peer eviction and add advanced test coverage - Closes #4280

### LINKED ISSUE #4280: Advanced test coverage related to inbound eviction
### Description

More tests are needed to verify that inbound peer eviction is working under unusual scenarios.
Some things to check (note that this is not an exhaustive list):

- Verify that many bad peers connecting to a node quickly will not cause all good peers to be evicted.
- Verify that old version peers are not given more protection than new version peers.

Related issue: https://github.com/LiskHQ/lisk-sdk/issues/4229

### Motivation

Inbound peer eviction/protection is a complex feature which requires more thorough testing.

### Acceptance Criteria

- The test cases cover advanced or malicious scenarios like the ones described above.
- The goal of this issue is not to add test cases to verify obvious functionality. The tests should aim to uncover flaws. They need to focus on edge cases.

mitsujutsu:
### What was the problem?

- Some categories in inbound eviction process were dependent on each other causing minor change in eviction candidate results.

- There were no advanced unit tests for inbound eviction process.

### How did I solve it?

Decoupled `netgroup`, `latency` and `responseRate` peer eviction processing. 
Added required tests simulating different inbound peer scenarios.

The expected number of eviction candidates for each test are based on the following estimates derived from default protection ratios assuming 100 peers:  

`netgroup`: 4 peers, `latency`: 7 peers, `usefulness`: 7 peers, `longevity`: 41 peers

### How to manually test it?

npm run test

### Review checklist

- [ ] The PR resolves #4280 

### DIFF
--- a/elements/lisk-p2p/src/peer_pool.ts
+++ b/elements/lisk-p2p/src/peer_pool.ts
@@ -77,7 +77,12 @@ import { constructPeerIdFromPeerInfo, getUniquePeersbyIp } from './utils';
 interface FilterPeersOptions {
 	readonly category: PROTECTION_CATEGORY;
 	readonly percentage: number;
-	readonly asc: boolean;
+	readonly protectBy: PROTECT_BY;
+}
+
+export enum PROTECT_BY {
+	HIGHEST = 'highest',
+	LOWEST = 'lowest',
 }
 
 // Returns an array of peers to be protected
@@ -89,18 +94,15 @@ export const filterPeersByCategory = (
 	if (options.percentage > 1 || options.percentage < 0) {
 		return peers;
 	}
-	const protectedPeerIndex = Math.max(
-		1,
-		Math.floor(peers.length * (1 - options.percentage)),
-	);
-	const sign = !!options.asc ? 1 : -1;
+	const numberOfProtectedPeers = Math.ceil(peers.length * options.percentage);
+	const sign = options.protectBy === PROTECT_BY.HIGHEST ? -1 : 1;
 
 	// tslint:disable-next-line no-any
 	return peers
 		.sort((peerA: any, peerB: any) =>
 			peerA[options.category] > peerB[options.category] ? sign : sign * -1,
 		)
-		.slice(protectedPeerIndex, peers.length);
+		.slice(0, numberOfProtectedPeers);
 };
 
 export enum PROTECTION_CATEGORY {
@@ -612,8 +614,8 @@ export class PeerPool extends EventEmitter {
 			? filterPeersByCategory(peers, {
 					category: PROTECTION_CATEGORY.NET_GROUP,
 					percentage: this._peerPoolConfig.netgroupProtectionRatio,
-					asc: true,
-			  })
+					protectBy: PROTECT_BY.HIGHEST,
+			  }).map(peer => peer.id)
 			: [];
 
 		// Cannot manipulate without physically moving nodes closer to the target.
@@ -621,8 +623,8 @@ export class PeerPool extends EventEmitter {
 			? filterPeersByCategory(peers, {
 					category: PROTECTION_CATEGORY.LATENCY,
 					percentage: this._peerPoolConfig.latencyProtectionRatio,
-					asc: false,
-			  })
+					protectBy: PROTECT_BY.LOWEST,
+			  }).map(peer => peer.id)
 			: [];
 
 		// Cannot manipulate this metric without performing useful work.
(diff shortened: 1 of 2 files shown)


elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -80,26 +80,30 @@ interface FilterPeersOptions {
 	readonly asc: boolean;
 }
 
-const filterPeersByCategory = (
+// Returns an array of peers to be protected
+export const filterPeersByCategory = (
```
ishantiw: Can we make this function to accept argument `sort`, that takes `ASC` and `DESC` enum instead of because when we give `asc` as `false` then it should not mean that we have to return descending. Also, in the framework api module has a sort field that accepts `ASC` and `DESC`.

elements/lisk-p2p/test/unit/peer_pool.ts
```diff
@@ -73,6 +79,216 @@ describe('peerPool', () => {
...
+		describe('when node with netgroup protection disabled has 100 inbound peers', () => {
+			beforeEach(() => {
+				(peerPool as any)._peerPoolConfig.netgroupProtectionRatio = 0;
+				sandbox.stub(peerPool, 'getPeers').returns(originalPeers as Peer[]);
+			});
+
+			it('should return expected amount of eviction candidates', async () => {
+				const selectedPeersForEviction = (peerPool as any)._selectPeersForEviction();
+
+				expect(selectedPeersForEviction.length).to.eql(44);
```
ishantiw: how we are expecting that there will be 44 peers? I see in most of the places we are using a specific number and we don't know why we are expecting that number. If it's possible can we break it down like `peersWithPoorLatency` and we can use `expect(selectedPeersForEviction.length).to.eql(peersWithPoorLatency.length)`

mitsujutsu: If we compare to a `peersWithPoorLatency`, how do we define the expected threshold for `poorLatency`? The specific numbers can be calculated using the `DEFAULT` values. For example for this test, we have `netgroup` protection disabled. Using default protection ratios, the breakdown for 100 peers is: 4 by `netgroup`, 7 by `latency`, 7 by `usefulness `and 41 by `longevity` (simply multiply the ratio with 100). Since `netgroup` is disabled, that leaves `7 (latency) + 7(usefulness) + 41 (longevity) = 55. 100 - 55 = 45.` The 1 difference is a rounding difference, I opted to round down.

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -603,53 +607,58 @@ export class PeerPool extends EventEmitter {
...
 					percentage: this._peerPoolConfig.longevityProtectionRatio,
-					asc: true,
+					asc: false,
 			  })
-			: filteredPeersByResponseRate;
+			: [];
 
-		return filteredPeersByConnectTime;
+		return unprotectedPeers.filter(
+			peer => !protectedPeersByConnectTime.includes(peer),
```
jondubois: It's better to avoid passing object references to `Array.includes()` in case the peer object is a clone then then references will not match (even though both objects represent the same thing).

Also, using `includes()` inside a loop could be expensive if there are a lot of peers.

You could add an additional step before this and create a `new Set` containing of all the projected peer IDs for the category. Then inside the `filter` loop, you just do `set.has(...)`; set lookup is very fast regardless of how many items are in the set.

This should also be done in other cases above where `include` is used.

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -603,53 +607,58 @@ export class PeerPool extends EventEmitter {
...
-			return filteredPeersByResponseRate;
-		}
+			: [];
+
+		const uniqueProtectedPeers = [
+			...new Set([
+				...protectedPeersByNetgroup,
+				...protectedPeersByLatency,
+				...protectedPeersByResponseRate,
+			]),
```
jondubois: Why put the Set into an array? The Set is much faster for lookup especially if there are a lot of peers.

In the worst case, `Array.includes()` iterates over every element whereas `Set.has()` hashes directly to the key. Also, we should store the peer ids instead of the whole `Peer` instances or else it can lead to mismatched references if one object is a clone (which is a common thing that pure functions do); it's safer to always use primitive types for lookups.

elements/lisk-p2p/src/peer_pool.ts
jondubois: Maybe instead of `asc` or `desc` it can just be an enum `highest` or `lowest` because it's not clear from outside how the sorting relates to the selection (at first, it's not clear if the function protects the first `x%` or the last `x%`? If the developer didn't read the implementation of the function, they won't know whether they should use asc or desc so `highest` and `lowest` is more clear).

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -603,53 +607,58 @@ export class PeerPool extends EventEmitter {
...
-		const filteredPeersByConnectTime = this._peerPoolConfig
+		// Protect *the remaining half* of peers by longevity, precludes attacks that start later.
+		const protectedPeersByConnectTime = this._peerPoolConfig
 			.longevityProtectionRatio
-			? filterPeersByCategory(filteredPeersByResponseRate, {
+			? filterPeersByCategory(unprotectedPeers, {
 					category: PROTECTION_CATEGORY.CONNECT_TIME,
 					percentage: this._peerPoolConfig.longevityProtectionRatio,
-					asc: true,
+					asc: false,
```
jondubois: Doesn't this mean that those with the shortest connected time are more likely to be protected? We need to protect the oldest ones with the longest connected time.

elements/lisk-p2p/test/unit/peer_pool.ts
```diff
@@ -73,6 +79,216 @@ describe('peerPool', () => {
...
+			const filteredPeers = filterPeersByCategory(originalPeers as any, {
+				category: PROTECTION_CATEGORY.RESPONSE_RATE,
+				percentage: 0.2,
+				asc: true,
+			});
+
+			expect(filteredPeers.filter(p => p.responseRate === 1).length).to.eql(2);
+		});
+
+		it('should protect peers with lowest connectTime value when sorted by descending', async () => {
```
jondubois: Highest connectTime is a more important to test case.

elements/lisk-p2p/src/peer_pool.ts
mitsujutsu: No, it means those who connected at the smallest `Date` number e.g. longest connected time.

elements/lisk-p2p/test/unit/peer_pool.ts
mitsujutsu: `connectTime` is the exact time of connect, not `duration` of connection time.

jondubois: Ah nevermind then  ;p

elements/lisk-p2p/test/unit/peer_pool.ts
ishantiw: I mean if it's possible then can we actually make few variables (`noOfPeersbyNetgroup`, `noOfPeersbyLatency`, .etc) and then sum them up before using it in expect? Maybe just for a test so that its clear what's going on if I'm new and I read the tests directly.

mitsujutsu: It's a good point. I will add a comment detailing the above calculation before the `expect` to make it clear.

PR #4319, opened 2019-09-26
State store inconsistent return value - Closes #4317

### LINKED ISSUE #4317: StateStore.<entity>.cache() inconsistent return values
When using the `StateStore.<entity>.cache()` method one expect to have a consistent return value.

When we look at the return value of [TransactionStore.cache()](https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/modules/chain/state_store/transaction_store.js#L35) we can see the return value is the fetched data. Now if you take a look at [AccountStore.cache()](https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/modules/chain/state_store/account_store.js#L54) method you can see the entire cached data is returned.

This should be considered a bug and fixed without waiting a major release despite the breaking changes.

JesusTheHun:
As mentionned into LiskHQ#4317 the return values of the state store are inconsistent. This tiny patch fixes it.

### DIFF
--- a/framework/src/modules/chain/state_store/account_store.js
+++ b/framework/src/modules/chain/state_store/account_store.js
@@ -51,7 +51,7 @@ class AccountStore {
 	async cache(filter) {
 		const result = await this.account.get(filter, { extended: true }, this.tx);
 		this.data = _.uniqBy([...this.data, ...result], this.primaryKey);
-		return _.cloneDeep(this.data);
+		return _.cloneDeep(result);
 	}
 
 	createSnapshot() {
--- a/framework/src/modules/chain/state_store/transaction_store.js
+++ b/framework/src/modules/chain/state_store/transaction_store.js
@@ -32,7 +32,7 @@ class TransactionStore {
 			this.tx,
 		);
 		this.data = _.uniqBy([...this.data, ...result], this.primaryKey);
-		return result;
+		return _.cloneDeep(result);
 	}
 
 	add(element) {


framework/src/modules/chain/state_store/account_store.js
```diff
@@ -51,7 +51,7 @@ class AccountStore {
 	async cache(filter) {
 		const result = await this.account.get(filter, { extended: true }, this.tx);
 		this.data = _.uniqBy([...this.data, ...result], this.primaryKey);
-		return _.cloneDeep(this.data);
+		return result;
```
shuse2: Since account state will be possible for mutation, it would be better to return the cloned version of the result

shuse2:
@JesusTheHun Thank you for the contribution. I think it's correct that it should've been consistent with other store.
I just made a small comment regards to the change

framework/src/modules/chain/state_store/account_store.js
JesusTheHun: AFAIK the data comes straight from the database and I see no additional reference anywhere, making a clone seems unnecessary. One could argue there is never too much safety but account will carry assets, possibly a lot, so we should consider the performance aspect.

Of course if you still believe it's necessary I'll make the change.
Note that [transaction store does not clone the result](https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/modules/chain/state_store/transaction_store.js#L35).

shuse2: If someone mutates the account in the `prepare` step, it will be mutated directly without using `set` function.
Meanwhile, whenever account is `get`, it is cloned.
Therefore, I think it should be consistent.

lsilvs: Yeah, it needs to be cloned.

REF.: https://github.com/LiskHQ/lisk-sdk/issues/3913

JesusTheHun: Okay I have read some code. This is going to be a discussion.

From what I understand you manage an internal initial copy so you can find out what has been modified. The only purpose I've found to this is so you can make a partial update later on.

Knowing that most modifications will be made to the asset field, which you won't be able to update partially since it's a userland structure.

Based on those element I don't see the purpose of all this mechanics. But I see you have to workaround this design with the extensive use of deep clone.

What am I missing ?

shuse2: the copying is not really related to how it's handled for the partial update structure.
The design that we provide is if you want to make any changes to account state from the custom transaction, it needs to use `set`.
However, if you pass back the reference, in other part of code, it can be changed unknowingly.
To avoid that, we want to ensure that we return by value

JesusTheHun: Ok. I've updated the PR

shuse2:
@JesusTheHun would you be able to sign your commits? https://help.github.com/en/articles/signing-commits

JesusTheHun:
> @JesusTheHun would you be able to sign your commits? 

I squashed & signed.

PR #4322, opened 2019-09-26
Make usage of delegate list BFT compliant - Closes #3681

### LINKED ISSUE #3681: Make usage of delegate list BFT compliant
### Expected behavior
BFT requires us to use delegate list generated 2 rounds prior to active round. So generateDelegateList logic should support that.
### Actual behavior
We are currently using the delegate list that generated before the round for the round.
### Acceptance Criteria

- [ ] DelegateList is generated at the beginning of each round (after processing genesisBlock)
- [ ] For round `r`, the delegateList was generated in round `r - ROUND_DIFFERENCE` will be used and this will be written in sql query explicitly. ROUND_DIFFERENCE = 2.
- [ ] For the round 1 and 2, delegateList was provided in genesisBlock will be used.
- [ ] During migration, for migration round `m`, the list was generated in round `m - ROUND_DIFFERENCE` will be used, but list will be shuffled with activeRoundNo `m`.
 (This is also backward compatible)

lsilvs:
### What was the problem?
BFT required to use delegate list generated 2 rounds prior to current round in order to validate block forger.

### How did I solve it?
By moving the logic which verifies block forger into block processor and added the param `roundOffset` to `Dpos.verifyBlockForger`.

For `BlockProcessorV0` the `roundOffset` is equal to 0 and for `BlockProcessorV2` the `roundOffset` is equal to 2.

### How to manually test it?
Application should start and forge blocks as usual
New unit tests were added to `Dpos.verifyBlockForger`

### Review checklist

- [ ] The PR resolves #3681


framework/src/modules/chain/dpos/delegates_list.js
```diff
@@ -128,13 +128,18 @@ class DelegatesList extends EventEmitter {
...
 	 */
-	async verifyBlockForger(block) {
+	async verifyBlockForger(block, roundOffset = 0) {
 		const currentSlot = this.slots.getSlotNumber(block.timestamp);
-		const round = this.slots.calcRound(block.height);
-		const delegateList = await this.getForgerPublicKeysForRound(round);
+		const currentRound = this.slots.calcRound(block.height);
+
+		// Delegate list is generated from round 1 hence `roundToVerify` can't be less than 1
+		const roundToVerify = Math.max(currentRound - roundOffset, 1);
```
yatki: wouldn't it be better if we throw an exception? If we provide roundOffset = 3, seems like the first 2 rounds will be calculated as round 1?

framework/src/modules/chain/block_processor_v2.js
```diff
@@ -165,10 +165,19 @@ const validateSchema = ({ block }) => {
...
+	constructor({
+		blocksModule,
+		bftModule,
+		dposModule,
+		logger,
+		constants,
+		exceptions,
+	}) {
 		super();
+		const verifyBlockForgerOffset = 2;
```
yatki: wouldn't it be better if this comes from config?

framework/src/modules/chain/dpos/delegates_list.js
lsilvs: That's the idea. From the LIP:

> Note that for a new blockchain, this delay of 2 rounds can be achieved by letting the active delegates of the first three rounds be the delegates defined in the genesis block.

framework/src/modules/chain/block_processor_v2.js
lsilvs: This is a protocol attribute and changing it would require a hard-fork.

yatki: but rest of the implementation in bft is based on constants.  I don't have a strong opinion on this but maybe @shuse2 can comment.

michielmulders: We could define it in `constants.js` but don't expose it as indeed this is hard-forking.

framework/src/modules/chain/blocks/verify.js
```diff
@@ -40,9 +40,8 @@ const verifyPreviousBlockId = (block, lastBlock, genesisBlock) => {
 };
 
 class BlocksVerify {
-	constructor({ storage, exceptions, slots, dposModule, genesisBlock }) {
+	constructor({ storage, exceptions, slots, genesisBlock }) {
```
michielmulders: Do we have a test for verifying the attributes the constructor expects?

framework/src/modules/chain/block_processor_v2.js
shuse2: It could be in the genesis config, but since it is very specific to v2 and hard to predict what it will affect in the future, let's keep it here or in `constants.js`

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -201,7 +200,6 @@ class Blocks extends EventEmitter {
 	async verifyInMemory({ block, lastBlock }) {
 		verifyPreviousBlockId(block, lastBlock, this.genesisBlock);
 		validateBlockSlot(block, lastBlock, this.slots);
-		await this.blocksVerify.verifyBlockForger(block);
```
shuse2: ❤️

framework/src/modules/chain/blocks/verify.js
lsilvs: Not that I'm aware of.

framework/src/modules/chain/block_processor_v2.js
lsilvs: I will leave it here for now. We can move it to constants when it is used somewhere else.

Issue #3771, opened 2019-06-03
Implement request and apply blocks of new chain step for Block Sync Mechanism


limiaspasdaniel:
### Expected behavior
Implement request and apply blocks of new chain step for Block Sync Mechanism

We request blocks from `P` and apply them according to the section "Processing Blocks" of the LIP until we have block `B` at tip of the chain.

In order to do this we need to issue a network call `blocks` which returns 34 blocks. This is pretty old code and maybe we can update it to rename the endpoint to `getBlocks` and rethink how many blocks to return.

a) If all blocks up to `B` are successfully applied, the block synchronization mechanism terminates.

b) If the node is not able to obtain all blocks up to block `B` from `P`, but synchronization stops at a block `T` (i.e., `T` is the tip of the current chain), then the node checks whether `T` has priority over `A` (where A is tip of our original chain)  in terms of the fork choice rule (condition at the beginning of the section):

`(A.heightPrevoted<T.heightPrevoted) || A.height<T.height && A.heightPrevoted==T.heightPrevoted)`

If the check fails, `P` is banned, the chain with tip `A` is restored and block synchronization is restarted from Step 1. If the check passes, block synchronization is also restarted from Step 1.

### Which version(s) does this affect? (Environment, OS, etc...)
`feature/BFT`

PR #4341, opened 2019-09-30
Revert to last common block in BSM - Closes #3770

### LINKED ISSUE #3770: Implement reverting to last common block step for Block Sync Mechanism
### Expected behavior
Implement reverting to last common block step for Block Sync Mechanism

a) The node requests the last common block `C` from `P` (Peer).
b) If `C.height<chainHeightFinalized`, then `P` is banned and the block synchronization is restarted from Step 1. BFT module will provide `chainHeightFinalized`

c) The node deletes all blocks of the current chain until block `C` is the tip of the chain. All these blocks are stored persistently so the previous chain can be restored in case synchronizing with `P` fails. This logic will not be implemented inside Block Synchronizer but instead shared with Fast Chain Switcher

`feature/BFT`

limiaspasdaniel:
### What was the problem?
Refer to #3770 
### How did I solve it?
By implementing all the steps mentioned in **Step 3** under [LIP-0014#Block Synchronization Mechanism](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#block-synchronization-mechanism)

A note on finding the common block:

The LIP doesn't mention how to specifically handle this scenario.

The way for finding a common block between two peers in our implementation is the following:
- A node issues a network call `getHighestCommonBlock` including a payload that must contain a non-empty array of block IDs
- If the targeted peer finds a block ID in common with one of those IDs present in its chain, it returns the full block back to the requester.
- If the peer is unable to find a common block it simply returns undefined.

As the difference in height between both chains can be way bigger than 202 blocks (`activeDelegates`*2), instead of sending an array of IDs of consecutive blocks, an array of block IDs of the first block of `N` rounds is sent, `N` being a number to discuss.

This behavior is repeated until the maximum number of retries is reached.

### How to manually test it?
Tests will be added for `run` public interface once #3771 is implemented as stated in https://github.com/LiskHQ/lisk-sdk/pull/4341#issuecomment-539056621
### Review checklist

- [x] The PR resolves #3770 


framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -42,47 +43,199 @@ class BlockSynchronizationMechanism {
...
 			const { connectedPeers: peers } = await this.channel.invoke(
-				'network:getUniqueOutboundConnectedPeers',
+				'network:getNetworkStatus',
 			);
 			if (!peers.length) {
 				throw new Error('Connected peers list is empty');
 			}
-			// eslint-disable-next-line no-unused-vars
+
 			const bestPeer = this._computeBestPeer(peers);
```
michielmulders: I like the implementation below there, however, we need to discuss about `_computeBestPeer` with P2P team @limiaspasdaniel

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -42,47 +43,199 @@ class BlockSynchronizationMechanism {
...
+	 * @param {Error } error - An error object containing the reason for applying
+	 * a penalty and restarting sync
+	 * @private
+	 */
+	_applyPenaltyAndRestartSync(peer, receivedBlock, error) {
+		this.channel.invoke('network:applyPenalty', {
+			peerId: peer.id,
+			penalty: 100,
+		});
+		this.channel.publish('chain:processor:sync', { block: receivedBlock });
```
michielmulders: Do we restart the sync as the peer violated some block validation check?

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -42,47 +43,199 @@ class BlockSynchronizationMechanism {
...
+	 * @param {Object} peer - The selected peer to target.
+	 * @return {Promise<void>}
+	 * @private
+	 */
+	async _revertToLastCommonBlock(receivedBlock, peer) {
+		const lastCommonBlock = await this._requestLastCommonBlock(peer);
+
+		if (!lastCommonBlock) {
+			// TODO: Should we apply penalty here?
+			this._applyPenaltyAndRestartSync(
```
michielmulders: In case we don't find a common block from the peer that just send us the conflicting block, we can be sure something went wrong or he modified his database? Check with science if we should apply a penalty and how we should handle this edge case.

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -42,47 +43,199 @@ class BlockSynchronizationMechanism {
...
+		let numberOfBlocks = 0; // Keeps track of the number of block IDs to be included in a request
+		let currentRound = Math.floor(
+			lastBlock.height / this.constants.activeDelegates,
+		); // Keeps track of the round number it is  being used to compute the first block of the round to be included in the request payload
+		let highestCommonBlock; // Holds the common block returned by the peer if found.
+
+		while (
+			!highestCommonBlock &&
+			numberOfRequests < requestLimit &&
+			currentRound * this.constants.activeDelegates < this.bft.finalizedHeight
```
michielmulders: Why do we want this to be lower than height finalized?

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -42,47 +43,199 @@ class BlockSynchronizationMechanism {
...
+		const requestLimit = 10; // Maximum number of requests to be made to the remote peer
+
+		let numberOfRequests = 0; // Keeps track of the number of requests made to the remote peer
+		let numberOfBlocks = 0; // Keeps track of the number of block IDs to be included in a request
+		let currentRound = Math.floor(
+			lastBlock.height / this.constants.activeDelegates,
+		); // Keeps track of the round number it is  being used to compute the first block of the round to be included in the request payload
+		let highestCommonBlock; // Holds the common block returned by the peer if found.
+
+		while (
```
michielmulders: This nested while loop structure looks a bit complex, maybe we can find an easier way to rewrite this? The cyclomatic complexity of this function is really high and needs a lot of comments to understand the idea that's happening here. I would go for a more functional approach (but I understand this is a draft!) 👍

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
limiaspasdaniel: Agreed, we can have a discussion about it but this PR didn't introduce this so I would keep it for a separate meeting.

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
limiaspasdaniel: Yes, this function is being used to do that and also ban the peer.

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -42,47 +43,199 @@ class BlockSynchronizationMechanism {
...
+	/**
+	 * Requests the last common block in common with the targeted peer.
+	 * In order to do that, sends a set of network calls which include a set of block ids
+	 * corresponding to the first block of descendent consecutive rounds (starting from the last one)
+	 * @param peer - The peer to target.
+	 * @return {Promise<Object | undefined>}
+	 * @private
+	 */
+	async _requestLastCommonBlock(peer) {
+		const [lastBlock] = await this.storage.entities.Block.get({
```
shuse2: last block should be taken from the blocks module

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -42,47 +43,199 @@ class BlockSynchronizationMechanism {
...
-			this.active = false;
+
+			const blocks = await this.storage.entities.Block.get(
+				{ height_in: blockHeights },
+				{
+					sort: 'height:asc',
+				},
+			);
+
+			const blockIds = blocks.map(block => block.id);
```
shuse2: didn't we have this logic in the blocks module? and i think it should be in the blocks module

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -42,47 +43,199 @@ class BlockSynchronizationMechanism {
...
+			this._applyPenaltyAndRestartSync(
+				peer,
+				new Error(
+					'The last common block height is less than the finalized height of the current chain',
+				),
+			);
+		}
+
+		try {
+			await deleteBlocksAfterHeight(
```
shuse2: probably this name should say it is saving to temp table?

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
shuse2: I agree, i think it's bit too complex here

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
limiaspasdaniel: Solved after talking with Jan

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
limiaspasdaniel: Actually I had to switch the condition to `>`. If the block height is below `chainHeightFinalized`  we stop the loop as stated in Step 3 of BSM section in the LIP

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
limiaspasdaniel: Moved the inner while to a separate function. Please check again when you have time :)

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -42,47 +43,199 @@ class BlockSynchronizationMechanism {
 	}
 
 	async run(receivedBlock) {
+		this.active = true;
 		try {
-			this.active = true;
 			const { connectedPeers: peers } = await this.channel.invoke(
-				'network:getUniqueOutboundConnectedPeers',
+				'network:getNetworkStatus',
```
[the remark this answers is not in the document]
limiaspasdaniel: Changed it to `getPeers` as it returns outbound and inbound peers after talking with @ishantiw

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
limiaspasdaniel: Changed it :)

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
limiaspasdaniel: Moved inner while loop to a separate function within same file as it is specific logic to BSM

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -42,47 +43,199 @@ class BlockSynchronizationMechanism {
...
+			{
+				procedure: 'getLastBlock',
+				peerId: peer.id,
+			},
+		);
+
+		const { valid: validBlock, err } = await this._blockDetachedStatus(
+			networkLastBlock,
+		);
+		const notInDifferentChain = ForkChoiceRule.isDifferentChain(
```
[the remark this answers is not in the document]
limiaspasdaniel: Done :)

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -113,20 +115,47 @@ class BlockSynchronizationMechanism {
...
-				lastCommonBlock.height,
-			);
-		} catch (e) {
-			this.logger.debug(
-				`Failure deleting blocks after height ${
-					lastCommonBlock.height
-				} while reverting chain to last common block`,
-			);
-			throw e;
+		await deleteBlocksAfterHeightAndBackup(
```
michielmulders: Not 100% sure about the naming here? `AndBackup`?

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -113,20 +115,47 @@ class BlockSynchronizationMechanism {
...
+			lastCommonBlock.height,
+		);
+	}
+
+	/**
+	 * Returns a list of block heights corresponding to the first block of a defined number
+	 * of rounds (listSizeLimit)
+	 *
+	 * @param listSizeLimit
+	 * @param _currentRound
```
michielmulders: Multiline comment whoa!

- Can you add description for `listSizeLimit`
- I think that we don't append variables with `_` as the function is already privately defined?

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -113,20 +115,47 @@ class BlockSynchronizationMechanism {
...
+		let numberOfBlocks = 0; // Keeps track of the number of block IDs to be included in a request
+		let currentRound = _currentRound;
+		let currentHeight = currentRound * this.constants.activeDelegates;
+
+		while (
+			numberOfBlocks < listSizeLimit &&
+			currentHeight > this.bft.finalizedHeight
+		) {
+			let height = currentRound * this.constants.activeDelegates;
+			if (height <= this.bft.finalizedHeight) {
```
michielmulders: Nice!

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -138,53 +167,32 @@ class BlockSynchronizationMechanism {
...
 
 		let numberOfRequests = 0; // Keeps track of the number of requests made to the remote peer
-		let numberOfBlocks = 0; // Keeps track of the number of block IDs to be included in a request
-		let currentRound = Math.floor(
-			lastBlock.height / this.constants.activeDelegates,
-		); // Keeps track of the round number it is  being used to compute the first block of the round to be included in the request payload
 		let highestCommonBlock; // Holds the common block returned by the peer if found.
+		let currentRound = Math.floor(
+			this.blocks.lastBlock.height / this.constants.activeDelegates,
+		);
```
michielmulders: Removed the comment here, intentionally?

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
limiaspasdaniel: This is after @shuse2's comment, to specify that blocks are being backed up to temp table.

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
limiaspasdaniel: I think it makes sense to append `_` to a variable name when it needs to be cloned inside the function scope. It's done like this in the Storage layer.

limiaspasdaniel: Done :)

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
limiaspasdaniel: Added it back

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -42,47 +45,210 @@ class BlockSynchronizationMechanism {
...
-				});
-				this.channel.publish('chain:processor:sync', { block: receivedBlock });
-				throw err;
+	/**
+	 * Returns a list of block heights corresponding to the first block of a defined number
+	 * of rounds (listSizeLimit)
+	 *
+	 * @param listSizeLimit - The size of the array to be computed
+	 * @param _currentRound
+	 * @return {Promise<*>}
```
michielmulders: does this resolve an array of blockIDs all the time? If yes, pls change to `@return {Promise<[Number]>}`

framework/src/modules/chain/synchronizer/utils.js
```diff
@@ -42,6 +42,29 @@ const restoreBlocks = async (blocksModule, processorModule, tx) => {
...
 
+/**
+ * Deletes blocks of the current chain after the desired height exclusive and
+ * backs them up in temp_block database table.
+ * @param {Object} processorModule
+ * @param {Object} blocksModule
+ * @param {Number} desiredHeight - The height desired to delete blocks after.
+ * @return {Promise<void>} - Promise is resolved when blocks are successfully deleted
+ */
+const deleteBlocksAfterHeightAndBackup = async (
```
michielmulders: I think I added a comment about the naming of this function? `AndBackup` maybe `saveToTemp` to keep consistency or is that not related? @limiaspasdaniel

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
shuse2: blockIDs are string 😉

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -42,47 +45,210 @@ class BlockSynchronizationMechanism {
...
+	 * @return {Promise<Object | undefined>}
+	 * @private
+	 */
+	async _requestLastCommonBlock(peer) {
+		const blocksPerRequestLimit = 10; // Maximum number of block IDs to be included in a single request
+		const requestLimit = 10; // Maximum number of requests to be made to the remote peer
+
+		let numberOfRequests = 0; // Keeps track of the number of requests made to the remote peer
+		let highestCommonBlock; // Holds the common block returned by the peer if found.
+		let currentRound = Math.floor(
```
shuse2: please use the common `slot` function

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -42,47 +45,210 @@ class BlockSynchronizationMechanism {
...
+
+			// Request the highest common block with the previously computed list
+			// to the given peer
+			const { data } = await this.channel.invoke('network:requestFromPeer', {
+				procedure: 'getHighestCommonBlock',
+				peerId: peer.id,
+				data: blockIds,
+			});
+
+			highestCommonBlock = data; // If no common block, data is undefined.
```
shuse2: no need to reassign here.
maybe like
`const { data: highestCommonBlock } = ...`?

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -178,11 +344,21 @@ class BlockSynchronizationMechanism {
 			height: peers[0].height,
 		};
 
-		if (!ForkChoiceRule.isDifferentChain(this.blocks.lastBlock, peersTip)) {
+		const forkStatus = await this.processorModule.forkStatus.run({
+			block: peersTip,
```
shuse2: peersTip needs to have `version` to select the correct block_processor.
Also, I think `run` is the method for pipeline, and `processorModule.forkStatus` should be fine (although it should be exposed from `processor/processor.js`

framework/src/modules/chain/synchronizer/utils.js
limiaspasdaniel: I think `backup` is a synonym to `saveToTemp` and it better wraps the meaning of backing blocks up.

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -42,47 +45,210 @@ class BlockSynchronizationMechanism {
...
+	/**
+	 * Returns a list of block heights corresponding to the first block of a defined number
+	 * of rounds (listSizeLimit)
+	 *
+	 * @param listSizeLimit - The size of the array to be computed
+	 * @param _currentRound
+	 * @return {Promise<*>}
+	 * @private
+	 */
+	async _computeBlockHeightsList(listSizeLimit, _currentRound) {
```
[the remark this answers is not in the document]
limiaspasdaniel: I agree it's simpler, however filling an array with listSizeLimit shouldn;t be necessary?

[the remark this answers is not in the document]
shuse2: `new Array(listSizeLimit).fill(0)` is necessary to create initial array to loop through?

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
limiaspasdaniel: It is reassigned in order to check if the variable holds any value. If it does, the outer while loop stops.

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
limiaspasdaniel: Done :)

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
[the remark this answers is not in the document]
limiaspasdaniel: Done :)

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
limiaspasdaniel: Done :)

limiaspasdaniel:
> Draft looks good now! But needs testing/fixing unit tests

Tests will be done once draft is ready to avoid rewriting the tests. Thanks!

limiaspasdaniel:
After discussion with @yatki and @shuse2 we concluded unit testing is needed only for the public interface, in this case the `run` method, which will be testable once the next issue #3771 is implemented. @shuse2 will create an issue to tackle these unit tests for the whole method.

Hence, this PR is ready for final review.

PR #4364, opened 2019-10-02
Improve isValidNumber function validation - Closes #3902

### LINKED ISSUE #3902: Misleading validation function name
### Expected behavior
Expect the `isValidNumber` function inside `lisk-transactions` to return `true` when `isValidNumber('-1')`

Also expect to have tests for `isValidNumber` function.

### Actual behavior
Currently it returns `true` for `isValidNumber(-1)` but returns `false` for `isValidNumber('-1')`.
This is happening because `isValidNumber` uses `isNumberString` which has a regex allowing only `0-9` numbers.

### Steps to reproduce
```
const { isValidNumber } = require('./validation.ts');

isValidNumber('-1'); //? false
```

2.*

### Addional Information
- isValidNumber is used to check only the positive number now, so the naming should be isValidPositiveNumber

alepop:
### What was the problem?
Expect the `isValidNumber` function inside `lisk-transactions` to return `true` when `isValidNumber('-1')`

Also expect to have tests for `isValidNumber` function.

### How did I solve it?
- Add optional `-` sign for regexp validation
- Add test block for `isValidNumber` function

### How to manually test it?
run `npm test`

### Review checklist

- [x] The PR resolves #3902

### DIFF
--- a/elements/lisk-transactions/src/utils/validation/validation.ts
+++ b/elements/lisk-transactions/src/utils/validation/validation.ts
@@ -153,7 +153,7 @@ export const isNumberString = (str: string) => {
 		return false;
 	}
 
-	return /^[0-9]+$/g.test(str);
+	return /^-?[0-9]+$/g.test(str);
 };
 
 export const validateNonTransferAmount = (data: string) =>
--- a/elements/lisk-transactions/test/utils/validation/validation.ts
+++ b/elements/lisk-transactions/test/utils/validation/validation.ts
@@ -30,6 +30,7 @@ import {
 	isGreaterThanMaxTransactionId,
 	isNumberString,
 	isValidInteger,
+	isValidNumber,
 	isNullByteIncluded,
 } from '../../../src/utils/validation/validation';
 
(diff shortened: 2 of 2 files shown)


elements/lisk-transactions/src/utils/validation/validation.ts
```diff
@@ -153,7 +153,7 @@ export const isNumberString = (str: string) => {
 		return false;
 	}
 
-	return /^[0-9]+$/g.test(str);
+	return /^-?[0-9]+$/g.test(str);
```
yatki: I think isValidNumber still a misleading name. Because `123.123` would fail with this regex but it's a valid number. I was going to suggest to rename it to `isValidInteger` but seems like we already have it. So, I suggest either we would improve this to validate all numbers or remove this function?

https://www.regextester.com/96606

yatki:
@lsilvs `isValidAmount` ?

alepop:
Ok. I will change it to `isValidAmount` and remove negative allowance

yatki:
@lsilvs in the original issue you wrote that the function should return `true` when `'-1'` was provided. But now seems like you are saying something different. Just to be sure could you please confirm, what needs to be done or update the original issue?

lsilvs:
> @lsilvs in the original issue you wrote that the function should return `true` when `'-1'` was provided. But now seems like you are saying something different. Just to be sure could you please confirm, what needs to be done or update the original issue?

The issue is about the misleading name. We can keep this function and make it return `true` when `'-1'` but than we need to create a new one and update the usage for `amount` and `fee`

nazarhussain:
@alepop Please resolve the conflicts and update your branch.

nazarhussain:
@lsilvs @yatki It seems an old thread, was there any conclusion on it?

shuse2:
With the latest development, validation functionality was removed from the transactions library, and moved to lisk-validator.
Therefore, closing this PR for now, and I will update the issue with better description.

PR #4372, opened 2019-10-07
Refactor blocks endpoint - Closes #4369

### LINKED ISSUE #4369: Refactor `blocks` endpoint to retrieve blocks from a peer
### Description

Rename `blocks` endpoint (Chain action) to a meaningful name (e.g. `getBlocks`) and reconsider the number of blocks to return. As for now, it is set to 34 blocks probably due to payload size limits.

It would also be a plus to add schema validation for the action request format, the same way we're doing with action `getHighestCommonBlock`.

### Motivation

`blocks` endpoint to retrieve blocks from a peer can be improved now that we need to use it only for BFT implementation purposes (BSM and FSC).

### Acceptance Criteria
- It is clear what the action does and its return value.
- It validates input schema and returns simple types.

### Additional Information

limiaspasdaniel:
### What was the problem?
Refer to #4369 
### How did I solve it?
- Renaming action to `getBlocksFromID`
### How to manually test it?
To do
### Review checklist

- [ ] The PR resolves #4369 


framework/src/modules/chain/transport/transport.js
```diff
@@ -215,37 +215,42 @@ class Transport {
...
-				message: err,
-				success: false,
-			};
+			throw new Error(error);
 		}
+
+		// Get 34 blocks with all data (joins) from provided block id
+		// According to maximum payload of 58150 bytes per block with every transaction being a vote
+		// Discounting maximum compression setting used in middleware
+		// Maximum transport payload = 2000000 bytes
```
limiaspasdaniel: @shuse2 @ishantiw 
I am tagging you here to catch up on what is the current maximum payload size set for our P2P layer as this will directly affect the number of blocks we can return.

ishantiw: Right now by default max payload size is `3048576` bytes

limiaspasdaniel: @ishantiw can you elaborate on why is that? Thanks!

ishantiw: Initially, we had `maxPayloadLimit` around `2000000 bytes` but when we were syncing with this limit, there were some packets having size higher than this limit and it was not syncing after that as it was rejected. To support syncing until the current height we decided to increase it to keep it `3048576 bytes`~3mb.

limiaspasdaniel: Thanks @ishantiw. @shuse2 does this statement `// According to maximum payload of 58150 bytes per block with every transaction being a vote
` still applies?

framework/src/modules/chain/schema/definitions.js
```diff
@@ -99,6 +99,17 @@ module.exports = {
 			type: 'object',
 		},
 	},
+	getBlocksFromIDRequest: {
+		id: 'getBlocksFromIDRequest',
+		type: 'object',
+		required: ['ids'],
```
shuse2: there is no property `ids`.

framework/src/modules/chain/transport/transport.js
shuse2: I think since we change the syncing format on this branch, it should be better.
We need to check size of a block with full vote with 25 transactions.

framework/src/modules/chain/transport/transport.js
```diff
@@ -215,37 +215,42 @@ class Transport {
...
+
+		const data = await this.blocksModule.loadBlocksFromLastBlockId(
```
shuse2: probably we can just return here?

limiaspasdaniel: True 😄

framework/src/modules/chain/transport/transport.js
limiaspasdaniel: After taking with @shuse2 let's leave it to 34 blocks and discuss this in another spike issue

framework/src/modules/chain/chain.js
```diff
@@ -307,7 +307,8 @@ module.exports = class Chain {
 				lastBlock: this.blocks.lastBlock,
 			}),
 			getLastBlock: async () => this.blocks.lastBlock,
-			blocks: async action => this.transport.blocks(action.params || {}),
+			getBlocksFromID: async action =>
```
ishantiw: `getBlocksFromId`? and also at other places

framework/src/modules/chain/transport/transport.js
```diff
@@ -215,37 +215,32 @@ class Transport {
 	 */
 
 	/**
-	 * Description of blocks.
-	 *
-	 * @todo Add @param tags
-	 * @todo Add description of the function
+	 * Returns a set of full blocks starting on the ID defined in the payload up to
```
ishantiw: `...from the block ID defined in...`

PR #4373, opened 2019-10-07
Emit rounds change event when round finishes in Dpos module - Closes #4264

### LINKED ISSUE #4264: Emit rounds change event when round finishes in Dpos module
### Description

When a round finishes it should emit the `chain:rounds:change` event using `channel.publish`.

### Motivation

This event is used by `http_api` to clear delegate list cache and was emitted by Rounds modules but it was removed as part of https://github.com/LiskHQ/lisk-sdk/pull/4191/files#diff-1e4cc7d62ac5555da5eef6a3b0882828L216

### Acceptance Criteria

The event `chain:rounds:change` with the corresponding round should be emitted after each round finishes.
The event should use `channel.publish`

### Additional Information

Note that Dpos module has a method called `onRoundFinish` but is never executed.
As part of this issue, the logic to trigger this method should be added and the event should be emitted inside it.

lsilvs:
### What was the problem?
Event `chain:rounds:change` not being emitted

### How did I solve it?
By using `channel` on Dpos module and emitting the event when last block of the round is applied or undone.

### How to manually test it?
Test if delegate list cache on HTTP API is cleared after the last block of the round is applied.

### Review checklist

- [ ] The PR resolves #4264


framework/test/jest/unit/specs/modules/chain/dpos/apply.spec.js
```diff
@@ -60,6 +60,11 @@ describe('dpos.apply()', () => {
 			error: jest.fn(),
 		};
 
+		stubs.channel = {
+			subscribe: jest.fn(),
+			publish: jest.fn(),
```
yatki: now publishing event is part of behaviour of `apply` and `undo`. So, could you please add one test case per each method to check if stubs are called with right parameters if the block is the last block of the round.

framework/src/modules/chain/dpos/delegates_list.js
```diff
@@ -39,9 +36,8 @@ const shuffleDelegateListForRound = (round, list) => {
 	return delegateList;
 };
 
-class DelegatesList extends EventEmitter {
+class DelegatesList {
```
yatki: ❤️ 🙇

shuse2:
@lsilvs 
Related to not extending event, I think not extending the event is good, but since dpos will be moved to the elements, I prefer not to use the channel.
Maybe, add event listener variable and emit from there?

framework/src/modules/chain/dpos/delegates_info.js
```diff
@@ -340,4 +347,4 @@ class DelegatesInfo {
 	}
 }
 
-module.exports = { DelegatesInfo };
+module.exports = { DelegatesInfo, EVENT_ROUND_CHANGED };
```
michielmulders: Would it make sense to export like this to clearly distinguish between both parameters?
```
module.exports = {
    DelegatesInfo,
    events: {
        EVENT_ROUND_CHANGED,
    }
}
```

michielmulders: Allows for more events in the future, otherwise, we have to refactor the interface everywhere if we added it as a third parameter.

framework/src/modules/chain/dpos/dpos.js
```diff
@@ -14,13 +14,21 @@
 
 'use strict';
 
-const { DelegatesList, EVENT_ROUND_FINISHED } = require('./delegates_list');
-const { DelegatesInfo } = require('./delegates_info');
+const { DelegatesList } = require('./delegates_list');
+const { DelegatesInfo, EVENT_ROUND_CHANGED } = require('./delegates_info');
```
michielmulders: If you decide to use `events` object, need to destructure here 💃

framework/test/mocha/integration/blocks/verify/verify.js
```diff
@@ -423,11 +423,15 @@ describe('blocks/verify', () => {
...
 			});
 
-			it('should fail when block generator is invalid (fork:3)', async () => {
+			it('should fail when block generator is invalid', async () => {
 				try {
 					await library.modules.processor.process(block2);
 				} catch (err) {
-					expect(err.message).equal('Failed to verify slot: 3377288');
+					expect(err.message).equal(
+						`Failed to verify slot: 3377288. Block ID: ${
```
michielmulders: Would you like to decouple this error message by using `${block2.timestamp}` instead of `3377288`? 🥇

framework/src/modules/chain/dpos/delegates_info.js
yatki: I prefer to keep it simple @michielmulders. I think it's good the way it is. I wish we could move all events to a single constants file but it's a trivial topic for the moment and can be easily refactored in future imo.

michielmulders: Agree, we discussed this as well that it should go in some constants file. To be honest, instead of postponing this I would make more sense to do it now no?

yatki: For me, it's up to @lsilvs. I'm ok with either way. I think maybe tackling this with a refactoring ticket makes more sense because we don't follow a common pattern when it comes to constant names. Also, there are constants that are changing application behaviour and some of them are just implementation details... So, I don't have a strong opinion on this :)

framework/test/mocha/integration/blocks/verify/verify.js
lsilvs: Slot and timestamp aren't the same think and there is no slot data in blocks

framework/src/modules/chain/dpos/delegates_info.js
lsilvs: I agree that it would be nice to have all events in one single file but would prefer to do that in a separate ticket. Will leave as it's for now.

framework/src/modules/chain/dpos/dpos.js
lsilvs: See answer above

lsilvs:
@shuse2 

> Related to not extending event, I think not extending the event is good, but since dpos will be moved to the elements, I prefer not to use the channel.

I thought the whole idea of the new architecture was to use channels everywhere to communicate between modules.

> Maybe, add event listener variable and emit from there?

Can you elaborate more this one, bitte?

shuse2:
@lsilvs 
> I thought the whole idea of the new architecture was to use channels everywhere to communicate between modules.
Yes, definitely within framework, that's true, but dpos, blocks and bft will be move to the elements. Although it could be done later too, but since we are changing now, i thought it would be better not to introduce it from the beginning.

> Maybe, add event listener variable and emit from there?
> Can you elaborate more this one, bitte?

for this one, i was thinking like this
```
this.events.emit('ROUND_CHANGE');
this.event.on('ROUND_CHANGE', () => {});
```

framework/test/mocha/integration/blocks/verify/verify.js
michielmulders: Ahhh I thought it was the timestamp from the fixed block, OK!

framework/src/modules/chain/dpos/dpos.js
```diff
@@ -14,30 +14,36 @@
...
 			slots,
 			activeDelegates,
 			logger,
+			events: this.events,
 			delegatesList: this.delegatesList,
 			exceptions,
 		});
 
-		this.delegatesList.on(EVENT_ROUND_FINISHED, () => {
+		this.events.on(EVENT_ROUND_CHANGED, () => {
```
shuse2: not related particular to this PR, but this should be
```
this.events.on(EVENT_ROUND_CHANGED, async () => {
  try {
    await this.onRoundFinish();
  } catch (err) {
    this.logger.error({ err }, 'Failed to apply round finish');
  }
});
```

framework/test/jest/unit/specs/modules/chain/dpos/apply.spec.js
lsilvs: `channel` was replaced by `events` and appropriate tests were added.

Issue #4314, opened 2019-09-26
Backup and restore peer reputation and productivity to and from the PeerBook


jondubois:
### Description

Depends on https://github.com/LiskHQ/lisk-sdk/issues/4313

In order to keep track of a peer across multiple connections over time, there needs to be a mechanism for backing up the last known `reputation` and `productivity` of the peer and to restore it when that peer connects to us the next time.

### Motivation

Currently, we only keep track of the `reputation` and `productivity` of a peer while it's active. Once the peer disconnects, we clear that information. For the purpose of banning and tracking the performance of a peer, it would be useful to be able to track that peer across multiple connections; and to hold onto that info for as long as possible. The `PeerBook` stores information about `triedPeers` across multiple connections, so the `P2PPeerInfo` for the peer is a good place to keep track of this information.

### Acceptance Criteria

- When a peer reconnects, our node creates the new peer instance using the last known productivity and reputation of the peer (reputation and productivity are not reset).
- Requires a new integration test scenario.

jondubois:
Something to discuss and think about is whether we should backup the `P2PPeerInfo` only when the peer becomes disconnected or if we should update it periodically based on more granular events.

The benefit of updating more often is that the `PeerBook` would hold the latest, most up to day info possible.

PR #4374, opened 2019-10-07
Use new data structure to capture peer info to handle internal parameters - Closes #4313

### LINKED ISSUE #4313: More detailed P2PPeerInfo
### Description

Currently, `P2PPeerInfo` (which we keep in the `PeerBook`) only holds information about a peer which we can share with other peers. There should also be a way to store information about a peer which the node wants to keep only for its own private use. For example: A peer's `reputation` and `productivity` should not be shared with other peers.

### Motivation

Being able to track private information about a peer on a `P2PPeerInfo` instance will open up the following opportunities:

- Allows a peer's state to be tracked across multiple connections over time. If a peer connects to our node or we connect to them, instead of creating a new `Peer` instance from scratch with a fresh `reputation` and `productivity` score, the `P2PPeerInfo` from the `PeerBook` could be used to initialize the `Peer` instance based on the last known `reputation` and `productivity` for that peer.

- The internal P2P library API has a `p2p.getConnectedPeers()` method which exposes `P2PPeerInfo` objects; it would be useful for these `P2PPeerInfo` objects include as much info about the peer as possible; then it could be used to get detailed information about the network from the node's perspective.

### Acceptance Criteria

Based on an internal poll, the P2P team voted to follow this format for the new `P2PPeerInfo` interface:

`interface P2PPeerInfo {sharedState: P2PSharedState, internalState: P2PInternalState}`

What is currently referred to as `P2PPeerInfo` in the current code (which a node shares with other peers) should be renamed to `P2PSharedState`. The new `P2PPeerInfo` then becomes the combination of the `sharedState` and `internalState` so it encapsulates complete information about the peer.

Note that the step of backing up the `P2PPeerInfo` inside the `PeerBook` and then restoring it when the peer reconnects is outside the scope of this issue.

ishantiw:
### What was the problem?

P2PPeerInfo (which we keep in the PeerBook) only holds information about a peer that we can share with other peers. There should also be a way to store information about a peer which the node wants to keep only for its own private use.

### How did I solve it?

We will use the interface below to capture the internal and shared state of a peer along with `peerId` as we are using it at many places and every time we are using it we are constructing it. 

`interface P2PPeerInfo {peerId: string, ipAddress: string, wsPort: number, sharedState: P2PSharedState, internalState: P2PInternalState}`

- `peerId`: Since we construct the `peerId` from `ipAddress` and `wsPort`, we can assign `peerId` right from the beginning to any incoming peer.
- `ipAddress`: IP address of a peer
- `wsPort`: web socket port of a peer
- `sharedState`: This field will store different values of a peer that are communicated across the network. Also, removing `height` value as a mandatory field from it.
- `internalState`: This field will be an object holding values related to internal management of a peer and be more stateful about a peer so that we don't lose valuable information for performance and productivity once we disconnected with them. As of now all the fields are optional but it will be fully implemented and used in across the library in the issue #4314

`ProtocolPeerInfo` interface is modified and will accept both `ip` and `ipAddress` to have backward compatibility and will only contain `ipAddress` after v3.0.

### How to manually test it?

`npm t`

### Review checklist

- [x] The PR resolves #4313 


elements/lisk-p2p/src/p2p_types.ts
```diff
@@ -35,26 +35,60 @@ export interface P2PPenalty {
...
+	readonly internalState?: P2PInternalState;
+}
+
 export interface P2PPeersCount {
 	readonly outboundCount: number;
 	readonly inboundCount: number;
 }
 
-export interface P2PDiscoveredPeerInfo extends P2PPeerInfo {
+export interface P2PDiscoveredSharedPeerInfo extends P2PSharedState {
```
diego-G: I'd call it `P2PDiscoveredSharedState`.

elements/lisk-p2p/src/p2p_types.ts
```diff
@@ -35,26 +35,60 @@ export interface P2PPenalty {
...
-export interface P2PPeerInfo {
+export interface P2PSharedState {
 	readonly ipAddress: string;
 	readonly wsPort: number;
 	// tslint:disable-next-line: no-mixed-interface
 	readonly [key: string]: unknown;
 }
 
+export enum ConnectionKind {
+	OUTBOUND = 'outbound',
```
diego-G: If you introduce this, we must remove `PEER_KIND_OUTBOUND` and `PEER_KIND_INBOUND` from `constants.ts`.

elements/lisk-p2p/src/p2p_types.ts
```diff
@@ -35,26 +35,60 @@ export interface P2PPenalty {
...
-export interface P2PDiscoveredPeerInfo extends P2PPeerInfo {
+export interface P2PDiscoveredSharedPeerInfo extends P2PSharedState {
 	readonly height: number;
 	readonly updatedAt?: Date;
 	readonly os?: string;
 	readonly version: string;
 	readonly protocolVersion: string;
 }
 
+export interface P2PDiscoveredPeerInfo extends P2PPeerInfo {
```
diego-G: Can we try to remove this?

elements/lisk-p2p/src/p2p_types.ts
```diff
@@ -35,26 +35,60 @@ export interface P2PPenalty {
 	readonly penalty: number;
 }
 
-export interface P2PPeerInfo {
+export interface P2PSharedState {
```
diego-G: I would prefer to not to have this interface and keep the values at the root. Please, assess how long will take to change the approach.

elements/lisk-p2p/src/p2p_types.ts
```diff
@@ -35,26 +35,60 @@ export interface P2PPenalty {
...
 	readonly [key: string]: unknown;
 }
 
+export enum ConnectionKind {
+	OUTBOUND = 'outbound',
+	INBOUND = 'inbound',
+	NONE = 'none',
+}
+
+export enum PeerType {
```
diego-G: Move this to `constants.ts`

diego-G: What about rename to `PeerKind`?

elements/lisk-p2p/src/utils/validate.ts
```diff
@@ -44,40 +44,53 @@ const IPV4_NUMBER = 4;
...
+	peerInfo: P2PPeerInfo,
 	nodeInfo: P2PNodeInfo,
 ): boolean => {
-	if (!peerInfo.nethash) {
+	if (!peerInfo.sharedState) {
 		return false;
 	}
 
-	return peerInfo.nethash === nodeInfo.nethash;
+	if (!peerInfo.sharedState.nethash) {
```
mitsujutsu: You can do this check and one above in one block with `||`

elements/lisk-p2p/src/p2p_types.ts
```diff
@@ -165,14 +195,17 @@ export interface P2PCompatibilityCheckReturnType {
...
 export type P2PCheckPeerCompatibility = (
-	headers: P2PDiscoveredPeerInfo,
+	headers: P2PPeerInfo,
 	nodeInfo: P2PNodeInfo,
 ) => P2PCompatibilityCheckReturnType;
 
 // This is a representation of the inbound peer object according to the current protocol.
 // TODO later: Switch to LIP protocol format.
+// TODO: Include peerId as field
 export interface ProtocolPeerInfo {
```
diego-G: Please, reuse the new types or extend them to have a better understanding of what we can remove in the near future. Review the comments and place all the candidates to remove at the bottom for example.

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -151,11 +149,8 @@ export class Peer extends EventEmitter {
 		this._peerConfig = peerConfig;
 		this._ipAddress = peerInfo.ipAddress;
 		this._wsPort = peerInfo.wsPort;
-		this._id = constructPeerIdFromPeerInfo({
-			ipAddress: this._ipAddress,
-			wsPort: this._wsPort,
-		});
-		this._height = peerInfo.height ? (peerInfo.height as number) : 0;
+		this._id = peerInfo.peerId;
```
diego-G: Shouldn't this come from `this._peerInfo`? Assess if this can be tackle here or we should open a specific issue.

elements/lisk-p2p/src/p2p_types.ts
```diff
@@ -35,26 +35,60 @@ export interface P2PPenalty {
...
+	readonly isBanned: boolean;
+	readonly productivity?: number;
+	readonly reputation?: number;
+	readonly connectionKind?: ConnectionKind;
+}
+
+export interface P2PPeerInfo {
+	// String to uniquely identify each peer
+	readonly peerId: string;
+	readonly sharedState: P2PSharedState;
```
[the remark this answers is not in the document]
ishantiw: actually this naming is the outcome of the poll we did and everyone voted for these names 😅

elements/lisk-p2p/src/p2p_types.ts
```diff
@@ -35,26 +35,60 @@ export interface P2PPenalty {
...
+	readonly peerType: PeerType;
+	readonly isBanned: boolean;
+	readonly productivity?: number;
+	readonly reputation?: number;
+	readonly connectionKind?: ConnectionKind;
+}
+
+export interface P2PPeerInfo {
+	// String to uniquely identify each peer
+	readonly peerId: string;
```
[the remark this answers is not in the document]
ishantiw: I have updated the description and now we have `ipAddress`, `wsPort` and `peerId` at root level for easy access and simplicity.

elements/lisk-p2p/src/utils/validate.ts
ishantiw: I separated it because typescript was throwing an error, not only we want to check if `peerInfo.sharedState` exist to check `peerInfo.sharedState.nethash` but also have to check if `peerInfo.sharedState` is undefined.

elements/lisk-p2p/src/p2p_types.ts
ishantiw: Removed

elements/lisk-p2p/src/peer/base.ts
ishantiw: created #4401 issue 👆

elements/lisk-p2p/src/p2p_types.ts
ishantiw: I have tried to reuse `P2PSharedPeerInfo` as much as I could to shorten `P2PNodeInfo` and `ProtocolNodeInfo`, couldn't update `ProtocolPeerInfo` because most of its fields are optional and TS throws an error as `P2PSharedPeerInfo` has some mandatory values

elements/lisk-p2p/src/p2p.ts
```diff
@@ -416,15 +457,19 @@ export class P2P extends EventEmitter {
...
-						...detailedPeerInfo,
-						ipAddress: error.peerInfo.ipAddress,
-						wsPort: error.peerInfo.wsPort,
-					};
-					const isUpdated = this._peerBook.updatePeer(updatedPeerInfo);
-					if (isUpdated) {
-						// If found and updated successfully then upgrade the peer
-						this._peerBook.upgradePeer(updatedPeerInfo);
+					// Don't update peerInfo when we already have connection with that peer
+					if (!this._peerPool.getPeer(error.peerInfo.peerId)) {
```
shuse2: `hasPeer` would be better?

elements/lisk-p2p/src/p2p_types.ts
```diff
@@ -35,26 +35,52 @@ export interface P2PPenalty {
 	readonly penalty: number;
 }
 
+export interface P2PSharedState {
+	readonly height: number;
+	readonly updatedAt?: Date;
+	readonly os?: string;
```
shuse2: should we share `updatedAt` and `os`?

elements/lisk-p2p/src/utils/misc.ts
```diff
@@ -139,8 +143,10 @@ export const getUniquePeersbyIp = (
 	return [...peerMap.values()];
 };
 
-export const constructPeerIdFromPeerInfo = (peerInfo: P2PPeerInfo): string =>
-	`${peerInfo.ipAddress}:${peerInfo.wsPort}`;
+export const constructPeerIdFromPeerInfo = (
```
diego-G: It's not PeerInfo anymore. We should rename the method to `constructPeerId` for example.

elements/lisk-p2p/src/p2p_types.ts
```diff
@@ -35,51 +37,87 @@ export interface P2PPenalty {
...
+	readonly version: string;
+	readonly protocolVersion: string;
+	// tslint:disable-next-line: no-mixed-interface
+	readonly [key: string]: unknown;
+}
+
+export interface P2PInternalState {
+	readonly dateAdded?: Date;
```
diego-G: This parameter looks like overlapping with https://github.com/LiskHQ/lisk-sdk/blob/a32eb03c8d0aa8836ca00f3d1089e34d845cc515/elements/lisk-p2p/src/peer_book/base_list.ts#L31 Shall we remove one of them? Shall we remove `CustomPeerInfo` interface and reuse the new `P2PPeerInfo`?

elements/lisk-p2p/src/p2p_types.ts
```diff
@@ -35,51 +37,87 @@ export interface P2PPenalty {
...
+	readonly peerKind?: PeerKind;
+	readonly isBanned?: boolean;
```
diego-G: This is not used.

elements/lisk-p2p/src/p2p_types.ts
ishantiw: In future we might have more values that we need within peer bucket so its better to still keep it and also `dateAdded` is removed from P2PPeerInfo and its a required field in bucket and also triedPeers use `maxConnectRetries`

elements/lisk-p2p/src/p2p_types.ts
ishantiw: Right now only connectionKind value is being used, the rest of the values in internalState will be handled in #4314 issue, including if a peer is banned or not

elements/lisk-p2p/src/p2p_types.ts
```diff
@@ -35,51 +37,87 @@ export interface P2PPenalty {
...
+	readonly productivity?: number;
+	readonly reputation?: number;
```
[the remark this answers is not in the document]
ishantiw: Yeah this will be moved to `peer.Info.internalState` from `base.ts` in #4314

PR #4377, opened 2019-10-08
Increase unit test coverage across P2P peer book - Closes #4227

### LINKED ISSUE #4227: Add unit test coverage for P2P peer directory folder
### Expected behavior

Every file from `lisk-p2p` library is well covered with unit test. The minimum must be 80% at least to achieve green reports.  

### Actual behavior

`peer_directory/` counts with only 75% coverage.

### Steps to reproduce

`npm t`

`+2.3`

diego-G:
### How to manually test it?
`npm t`
`npm run test:unit`

### Review checklist
- [x] The PR resolves #4227


elements/lisk-p2p/src/peer_book/base_list.ts
```diff
@@ -121,66 +128,25 @@ export class PeerList {
...
-				isAdded: false,
-				evictedPeer: undefined,
-			};
-		}
-
-		const newPeer = this.initPeerInfo(peerInfo);
-
-		if (bucket.size < this.peerListConfig.peerBucketSize) {
-			bucket.set(incomingPeerId, newPeer);
+	public makeSpace(ipAddress: string): CustomPeerInfo | undefined {
```
shuse2: why it's not `addPeer` anymore? if the space is empty and just simply adding peer, would it use this function?

elements/lisk-p2p/test/unit/peer/base.ts
```diff
@@ -399,24 +391,17 @@ describe('peer/base', () => {
...
 
-			it('should throw an error', async () => {
-				try {
-					await defaultPeer.fetchPeers();
-				} catch (e) {
-					expect(e).to.be.an.instanceOf(RPCResponseError);
-					expect(e.message).to.be.eql('Failed to fetch peer list of peer');
-					expect(e.peerId).to.be.eql(defaultPeerInfo.ipAddress);
-				}
+			it('should throw an error', () => {
```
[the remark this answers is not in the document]
diego-G: Thank you for pointing out. I found out we need `return` when using `.to.be.rejectedWith`, otherwise if it fails we have an unhandled rejection. However, I don't see the need of `async`.

elements/lisk-p2p/src/peer_book/base_list.ts
diego-G: `makeSpace()` checks if the bucket is full and only in that case removes a peer using eviction strategies and returns that evicted peer. If the bucket is not full, it just returns undefined without removing any peer. After that, `addPeer()` is executed.

elements/lisk-p2p/src/peer_book/utils.ts
```diff
@@ -0,0 +1,135 @@
...
+						targetBBytes,
+					]),
+			  ).readUInt32BE(0) % firstMod
+			: hash(
+					Buffer.concat([secretBytes, networkBytes, addressBytes]),
+			  ).readUInt32BE(0) % firstMod;
+
+	kBytes.writeUInt32BE(k, 0);
+
+	// New peers: b = Hash(random_secret, source_group, k) % 128
```
ishantiw: These comments are out of place

elements/lisk-p2p/src/peer_book/base_list.ts
```diff
@@ -75,22 +70,35 @@ export class PeerList {
...
+		const peer = bucket.get(incomingPeerId);
 
-		if (!bucket) {
-			return false;
+		return peer ? peer.peerInfo : undefined;
+	}
+
+	public addPeer(peerInfo: P2PPeerInfo): CustomPeerInfo | undefined {
+		if (this.getPeer(peerInfo)) {
+			throw new Error('Peer already exists');
```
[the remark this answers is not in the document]
diego-G: The bucket must always exist. When you add a peer the only exception that might arise is the peer already exists.

elements/lisk-p2p/src/peer_book/utils.ts
diego-G: Where would you place them?

elements/lisk-p2p/src/peer_book/base_list.ts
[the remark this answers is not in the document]
mitsujutsu: Currently it looks like whereever `addPeer` is called there is a check to see if it already exists using `getPeer`

elements/lisk-p2p/src/peer_book/base_list.ts
mitsujutsu: I am not sure about the naming. maybe `checkBucket`

elements/lisk-p2p/src/peer_book/utils.ts
```diff
@@ -0,0 +1,135 @@
...
+	const bucketPeerIds = Array.from(bucket.keys());
+	const randomPeerIndex = Math.floor(Math.random() * bucketPeerIds.length);
+	const randomPeerId = bucketPeerIds[randomPeerIndex];
+	const evictedPeer = bucket.get(randomPeerId);
+	bucket.delete(randomPeerId);
+
+	return evictedPeer;
+};
+
+export const evictAnOldPeerFromBucket = (
```
mitsujutsu: prefer `expirePeerFromBucket`

elements/lisk-p2p/src/utils/misc.ts
```diff
@@ -119,82 +117,6 @@ export const getNetgroup = (address: string, secret: number): number => {
 	return hash(netgroupBytes).readUInt32BE(0);
 };
```
mitsujutsu: These functions can remain here and the new `util` functions can also move here, instead of having `util` files on different levels.

elements/lisk-p2p/src/peer_book/utils.ts
```diff
@@ -0,0 +1,135 @@
+/*
```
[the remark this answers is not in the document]
mitsujutsu: prefer to keep it where the `getBucketId` function is.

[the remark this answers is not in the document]
diego-G: I've moved `getBucketId()` util as well. Before it was placed into `utils/misc`. I don't see a big issue having these utils separated under `peer_book/`.  They are very specific and I don't foresee them used in another place.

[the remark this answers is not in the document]
mitsujutsu: I think they can all be in the `utils/misc` file in the base level where it was previously. imo there is no need for two levels of `utils` files.

elements/lisk-p2p/src/peer_book/base_list.ts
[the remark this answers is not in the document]
diego-G: I've added this commit https://github.com/LiskHQ/lisk-sdk/pull/4377/commits/ba6b41de56528d91685c08aab46c31525da88bba .

elements/lisk-p2p/src/peer_book/base_list.ts
diego-G: At the end, all of them are incomplete. MakeSpace is simply and we just need to write in the header that it does it if there is not space.

ishantiw: Yes, let's keep it simple as of now

Issue #4375, opened 2019-10-08
Cover block synchronisation mechanism with unit/integration tests


shuse2:
### Description

After all of the block syncrhonization mechanism implementation is done, it should follow test guideline and cover all the cases from the public interface `run`.

### Acceptance Criteria

- Coverage of the synchronization file should be more than 80%
- All the external module which uses IO should be stubbed.

PR #4378, opened 2019-10-08
Request and apply blocks of new chain [BSM] - Closes #3771

### LINKED ISSUE #3771: Implement request and apply blocks of new chain step for Block Sync Mechanism

limiaspasdaniel:
### What was the problem?
Refer to #3771 
### How did I solve it?
Refer to #3771 
### How to manually test it?
Test will be implemented on #4375 
### Review checklist
- [ ] The PR resolves #3771 


framework/src/modules/chain/processor/processor.js
```diff
@@ -158,9 +158,8 @@ class Processor {
 		});
 	}
 
-	async forkStatus(receivedBlock) {
+	async forkStatus(receivedBlock, lastBlock = this.blocksModule.lastBlock) {
```
michielmulders: I don't like this to have this static getter being used as default value directly in the parameter definition (personal opinion).

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -91,6 +227,8 @@ class BlockSynchronizationMechanism {
 	/**
 	 * Reverts the current chain so the new tip of the chain corresponds to the
 	 * last common block.
+	 *
+	 * Returns the last common block
```
michielmulders: Minor: Add this return message behind `@return {Promise<object>}`

framework/src/modules/chain/synchronizer/utils.js
```diff
@@ -51,10 +51,15 @@ const restoreBlocks = async (blocksModule, processorModule, tx) => {
...
 const deleteBlocksAfterHeightAndBackup = async (
+	logger,
 	processorModule,
 	blocksModule,
 	desiredHeight,
 ) => {
+	logger.debug(
+		{ height: desiredHeight },
+		'Deleting blocks after target height',
+	);
```
michielmulders: would it make sense to add another logger.debug message after the while loop telling something like `blocks deleted successfully`?

framework/src/modules/chain/transport/transport.js
```diff
@@ -215,37 +215,32 @@ class Transport {
...
 	/**
-	 * Description of blocks.
-	 *
-	 * @todo Add @param tags
-	 * @todo Add description of the function
+	 * Returns a set of full blocks starting from the ID defined in the payload up to
+	 * the current tip of the chain.
+	 * @param {object} payload
+	 * @param {string} payload.blockId - The ID of the starting block
+	 * @return {Promise<*|Promise<*>>}
```
michielmulders: What are the possible returns? `this.blocksModule.loadBlocksFromLastBlockId(payload.blockId, 34);` ?

framework/src/modules/chain/transport/transport.js
```diff
@@ -215,37 +215,32 @@ class Transport {
...
+			payload,
+		);
 
-		try {
-			const data = await this.blocksModule.loadBlocksFromLastBlockId(
-				query.lastBlockId,
-				34, // 1977100 bytes
+		if (validationResult.length) {
+			const err = validationResult;
+			const error = `${err[0].message}: ${err[0].path}`;
```
michielmulders: Manu is trying to make strict usage of only `err` throughout the code.
Can we do:
```
const error = `${validationResult[0].message}: ${validationResult[0].path}`;
```

or 

```
const errList = validationResult;
const err = `${errList[0].message}: ${errList[0].path}`;
```

michielmulders: then you can do below

```
this.logger.debug(
	{
		err,
		req: payload,
	},
        ...
);

throw new Error(err);
```

framework/test/mocha/functional/ws/transport/transport.js
```diff
@@ -127,27 +127,23 @@ describe('WS transport', () => {
...
+					data: { blockId },
 				});
-				expect(data)
-					.to.have.property('blocks')
-					.to.be.an('array');
+				expect(data).to.be.an('array');
 			});
 
-			it('using valid headers should be ok', async () => {
+			it('using a valid payload should be ok', async () => {
```
michielmulders: I see it was using this `it using a valid ...` format but would be nice if we could refactor to: `it( 'should use a valid payload')`

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -60,12 +63,142 @@ class BlockSynchronizationMechanism {
...
+	 * @return {Promise<void | boolean>}
+	 * @private
+	 */
+	async _requestAndApplyBlocksToCurrentChain(
+		receivedBlock,
+		lastCommonBlock,
+		peer,
+	) {
+		this.logger.debug(
+			{ peer, fromID: lastCommonBlock.id, toID: receivedBlock.id },
```
ManuGowda: whole peer object would be too much info.. only peer id?

framework/src/modules/chain/processor/processor.js
limiaspasdaniel: Sometimes we want to pass an optional lastBlock to statically compare two blocks (BSM)

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -60,12 +63,142 @@ class BlockSynchronizationMechanism {
...
+			lastCommonBlock.id,
+			receivedBlock.id,
+		);
+		const tipBeforeApplying = cloneDeep(this.blocks.lastBlock);
+
+		if (!listOfFullBlocks.length) {
+			return this._applyPenaltyAndRestartSync(
+				peer,
+				receivedBlock,
+				new Error("Peer didn't return any block after calling getBlocksFromId"),
```
ManuGowda: Instead of an error object, how about defining error codes for the penalty?
example: PEER_PENALTY_EMPTY_BLOCKS = `code here`

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -60,12 +63,142 @@ class BlockSynchronizationMechanism {
...
+			);
+		}
+
+		try {
+			await this._applyBlocksToCurrentChain(listOfFullBlocks);
+		} catch (e) {
```
ManuGowda: how about `err` instead of `e`? that way you can use { err } below in logger

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -60,12 +63,142 @@ class BlockSynchronizationMechanism {
...
+			this.logger.debug(
+				{ err: e },
+				'Applying obtained blocks to the current chain failed',
```
ManuGowda: `Failed to apply blocks obtained to current chain`?

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -60,12 +63,142 @@ class BlockSynchronizationMechanism {
...
+			const forkStatus = await this.processorModule.forkStatus(
+				this.blocks.lastBlock, // New tip of the chain
+				tipBeforeApplying, // Previous tip of the chain
+			);
+
+			if (!forkStatus === FORK_STATUS_DIFFERENT_CHAIN) {
+				return this._applyPenaltyAndRestartSync(
+					peer,
+					receivedBlock,
+					new Error(
```
ManuGowda: same here, using the error code?

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -60,12 +63,142 @@ class BlockSynchronizationMechanism {
...
+						'New tip of the chain has no preference over the previous tip before synchronizing',
+					),
+				);
+			}
+
+			this.logger.debug('Restarting block synchronization');
```
ManuGowda: info maybe?

framework/src/modules/chain/synchronizer/utils.js
limiaspasdaniel: Agreed

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -78,6 +211,10 @@ class BlockSynchronizationMechanism {
 	 * @private
 	 */
 	async _applyPenaltyAndRestartSync(peer, receivedBlock, error) {
+		this.logger.debug(
```
ManuGowda: info?

framework/src/modules/chain/blocks/fork_choice_rule.js
```diff
@@ -142,12 +142,17 @@ const isTieBreak = ({ slots, lastAppliedBlock, receivedBlock }) =>
...
  */
 // eslint-disable-next-line class-methods-use-this
-const isDifferentChain = (lastBlock, currentBlock) =>
-	lastBlock.prevotedConfirmedUptoHeight <
-		currentBlock.prevotedConfirmedUptoHeight ||
-	(lastBlock.height < currentBlock.height &&
-		lastBlock.prevotedConfirmedUptoHeight ===
-			currentBlock.prevotedConfirmedUptoHeight);
+const isDifferentChain = (lastBlock, currentBlock) => {
+	if (lastBlock.height === 1) return true;
```
michielmulders: nice one!

framework/src/modules/chain/blocks/schema.js
```diff
@@ -54,15 +54,15 @@ const baseBlockSchema = {
 			type: 'integer',
 		},
 		totalAmount: {
-			type: 'object',
+			type: ['string', 'object'],
```
michielmulders: NOt sure why we expect or string or object? Is it not just a string?

framework/src/modules/chain/processor/processor.js
michielmulders: I agree on the use, but I don't like to have like business logic tied to our input parameters, feels bit odd?
this default value is more meant to be like object = {} or string = '' but`this.blocksModule.lastBlock` is really business logic and the function depends on it

limiaspasdaniel: It is meant to set default values, and in this case, the last block has default value the real last block of the chain. I get your point but I don't think it is misleading.

framework/src/modules/chain/blocks/schema.js
limiaspasdaniel: This is due to the fact that the function responsible to validate `amount` format accepts both object or string, therefore if only setting the type of `totalAmount` to object, when passing a `string` it was failing, so now it is actually correct and compliant with the format definition.

framework/src/modules/chain/block_processor_v2.js
```diff
@@ -95,22 +95,14 @@ const getBytes = block => {
...
 		SIZE_INT64,
 		LITTLE_ENDIAN,
 	);
 
-	const totalFeeBuffer = intToBuffer(
-		block.totalFee.toString(),
-		SIZE_INT64,
-		LITTLE_ENDIAN,
-	);
+	const totalFeeBuffer = intToBuffer(block.totalFee, SIZE_INT64, LITTLE_ENDIAN);
```
shuse2: I think this error was due to the deserialization timing.
All the processor logic assumes block to be deserialized, so i think this is also should be the case

framework/src/modules/chain/blocks/fork_choice_rule.js
shuse2: As coding standard, it should be
```
if (lastBlock.height === 1) {
  return true;
}
```
Also, why if the last block height is 1, then it's different chain?

framework/src/modules/chain/blocks/schema.js
shuse2: As mentioned above

framework/src/modules/chain/processor/processor.js
shuse2: I also agree with @michielmulders 
Default input should be static to be able to test easier.
It could be optional and assigned if not exist?

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -248,7 +421,7 @@ class BlockSynchronizationMechanism {
 	async isValidFor(receivedBlock) {
 		// 2. Step: Check whether current chain justifies triggering the block synchronization mechanism
 		const finalizedBlock = await this.storage.entities.Block.getOne({
-			height_eq: this.bft.finalizedHeight,
+			height_eql: this.bft.finalizedHeight,
```
shuse2: I think we can do this?
```suggestion
			height: this.bft.finalizedHeight,
```

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -60,12 +63,142 @@ class BlockSynchronizationMechanism {
...
+		this.logger.debug(
+			{
+				peerId: peer.id,
+				fromBlockId: lastCommonBlock.id,
+				toBlockId: receivedBlock.id,
+			},
+			'Requesting blocks within ID range from peer',
+		);
+
+		const listOfFullBlocks = await this._requestBlocksWithinIDs(
```
shuse2: shouldn't this loop through until we get to the last block?

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -60,12 +60,140 @@ class BlockSynchronizationMechanism {
...
+		return blocks;
+	}
+
+	/**
+	 * Applies a set of blocks on top of the current chain
+	 * @param {Array<object>} blocks - An array of full blocks
+	 * @return {Promise<void>}
+	 * @private
+	 */
+	async _applyBlocksToCurrentChain(blocks) {
```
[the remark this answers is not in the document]
limiaspasdaniel: @shuse2 it is removed, Idk why Github still shows it as present

framework/src/modules/chain/synchronizer/errors.js
```diff
@@ -0,0 +1,41 @@
+/*
+ * Copyright © 2018 Lisk Foundation
```
michielmulders: 2019 copyright?

framework/src/modules/chain/block_processor_v2.js
```diff
@@ -113,7 +113,7 @@ const getBytes = block => {
 	);
 
 	const payloadLengthBuffer = intToBuffer(
-		block.payloadLength,
+		block.payloadLength.toString(),
```
shuse2: This is number 😉

Issue #4136, opened 2019-08-23
Update Dpos module to consistently use state store


shuse2:
### Description

DPOS module should consistently use state store when it mutates the account and states.
In below, instead of taking the database transaction as second parameter, it should take `stateStore` as second parameter.

https://github.com/LiskHQ/lisk-sdk/blob/feature/introduce_bft_consensus/framework/src/modules/chain/dpos/delegates_info.js#L67
https://github.com/LiskHQ/lisk-sdk/blob/feature/introduce_bft_consensus/framework/src/modules/chain/dpos/delegates_info.js#L83

In stead of saving the rounds delegate in separate table, it should be saved in `chainState` with
```
// One column in ChainState
forgersList: [{
  round: number,
  delegates: string[],
}]
```

Also, when DPoS mutates the account, it should use the account from StateStore

### Motivation

In stead of directly modifying the account to database, it should always use state store to mutate the account.
Currently, we use database transaction to have consistency when updating the rewards and forged block information as in https://github.com/LiskHQ/lisk-sdk/blob/feature/introduce_bft_consensus/framework/src/modules/chain/dpos/delegates_info.js#L106-L122.
However, there introduce several complexity.
- Passing around the DB transaction and hard to follow through what is in transaction and what not
- Transactions are processed using state store (in memory) so those updates has to happen before processing the rounds using the same transaction
- Unnecessary round trip to database while processing a block.

Also, all the state mutation (calculated values which is affected by block/transaction), should be done using state store for the future features.

### Acceptance Criteria

- DPOS does not use storage to mutate the accounts
- Unit test needs to be updated

shuse2:
Considering the fact that account change, which is done by dpos will drastically be changed in the network consensus phase, this issue should be tackled at that time.

PR #4379, opened 2019-10-08
Investigate slow sync on introduce_bft_consensus-dpos.1 branch - Closes #4340

### LINKED ISSUE #4340: Investigate slow sync on introduce_bft_consensus-dpos branch
### Expected behavior
Expect the branch `introduce_bft_consensus-dpos.1` to sync with mainnet in less than 3 days

PS.: `introduce_bft_consensus-dpos.1` branch was branched off `feature/introduce_bft_consensus` after Dpos module implementation and before introducing breaking changes. Because of that, it should still have similar performance when syncing with mainnet (note that due to https://github.com/LiskHQ/lisk-sdk/issues/4194 it's not possible to sync with testnet)

### Actual behavior
It takes more than 5 days to sync up to the height `4 000 000`

### Steps to reproduce
Syncing with mainnet using Lisk Core branch `feature/introduce_bft_consensus` with Lisk SDK branch `introduce_bft_consensus-dpos.1`

`introduce_bft_consensus-dpos.1`

lsilvs:
### What was the problem?
When processing the last block of the round Dpos module updates the `voteWeight` of delegates that forging delegates voted for. This update was using `publicKey` as Storage filter but `publicKey` didn't have database index causing the query to take too long.

### How did I solve it?
I've created missing indexes and made some other improvements/fixes that was left out.

### How to manually test it?
Syncing with mainnet using Lisk Core branch `feature/introduce_bft_consensus` with Lisk SDK branch `4340-investigate-slow-sync`
Full sync should take less than 4 days

### Review checklist

- [ ] The PR resolves #4340

### DIFF
--- a/framework/src/modules/chain/migrations/sql/20191008170600_create_mem_accounts_public_key_index.sql
+++ b/framework/src/modules/chain/migrations/sql/20191008170600_create_mem_accounts_public_key_index.sql
@@ -1,0 +1,21 @@
+/*
+ * Copyright © 2019 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+
+ /*
+  DESCRIPTION: Create index for mem_accounts.publicKey
+  PARAMETERS: None
+*/
+
+CREATE INDEX IF NOT EXISTS "mem_accounts_public_key" ON "mem_accounts" (ENCODE("publicKey", 'hex'));
--- a/framework/src/modules/chain/transactions/transactions_handlers.js
+++ b/framework/src/modules/chain/transactions/transactions_handlers.js
@@ -286,8 +286,18 @@ const undoTransactions = (storage, exceptions) => async (
 		tx,
 	});
 
-	await Promise.all(transactions.map(t => t.prepare(stateStore)));
-	await Promise.all(transactions.map(t => votesWeight.prepare(stateStore, t)));
+	// Avoid merging both prepare statements into one for...of loop as this slows down the call dramatically
+	// eslint-disable-next-line no-restricted-syntax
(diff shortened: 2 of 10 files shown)


framework/src/modules/chain/migrations/sql/20191008170600_create_mem_accounts_public_key_index.sql
```diff
@@ -0,0 +1,21 @@
...
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+
+ /*
+  DESCRIPTION: Create index for mem_accounts.publicKey
+  PARAMETERS: None
+*/
+
+CREATE INDEX IF NOT EXISTS "mem_accounts_public_key" ON "mem_accounts" (ENCODE("publicKey", 'hex'));
```
shuse2: I think it will be nice to keep the publicKey out of index since we will use the state store in #4136

framework/src/modules/chain/transactions/transactions_handlers.js
```diff
@@ -151,8 +151,18 @@ const applyTransactions = (storage, exceptions) => async (transactions, tx) => {
...
+	// eslint-disable-next-line no-restricted-syntax
+	for (const transaction of transactions) {
+		// eslint-disable-next-line no-await-in-loop
+		await transaction.prepare(stateStore);
+	}
+
+	// eslint-disable-next-line no-restricted-syntax
+	for (const transaction of transactions) {
+		// eslint-disable-next-line no-await-in-loop
+		await votesWeight.prepare(stateStore, transaction);
```
shuse2: I think this also happens in the `undo` case too.
Not important for this PR, but it would be faster to change to `await votesWeight.prepare(stateStore, transactions);`

framework/src/modules/chain/migrations/sql/20191008170600_create_mem_accounts_public_key_index.sql
yatki: @shuse2 I agree that maybe creating index for publicKey just for this issue is not necessary but I'd prefer to tackle it after or during working on #4136. Since everything goes to bft branch we can remove this migration file later. But it's up to @lsilvs to decide. This is just my opinion.

framework/src/modules/chain/transactions/transactions_handlers.js
lsilvs: Done

PR #4385, opened 2019-10-11
Restore blocks upon startup - Closes #4348

### LINKED ISSUE #4348: Decide to restore blocks upon startup when temp_block table is not empty
### Expected behavior
Blocks need to be restored from the `temp_block` table upon restart of the node.

If there are blocks left, we need to decide if we want to recover the original chain or try to continue recovering the chain by applying the remaining blocks one by one.

**From LIP:**
b) If the node is not able to obtain all blocks up to block B from P, but synchronization stops at a block T (i.e., T is the tip of the current chain), then the node checks whether T has priority over A in terms of the fork choice rule (condition at the beginning of the section):

`(A.heightPrevoted<T.heightPrevoted) || A.height<T.height && A.heightPrevoted==T.heightPrevoted)`

If A has priority, we restore the original chain, otherwise we continue applying the blocks and remove the blocks from the `temp_block` table.

michielmulders:
### What was the problem?
If node crashes during `restoreBlocks` process, we need to check if we should continue applying those blocks or clear the `temp_block` table based on the fork choice rule. Refer to the `restoreBlocksUponStartup` which has the full description.

### How did I solve it?
`restoreBlocksUponStartup`

### How to manually test it?
Add some blocks to `temp_block`
Start node and check if `temp_block` table is cleared or applies the blocks.

### Review checklist

- [x] The PR resolves #4348 

### DIFF
--- a/framework/test/mocha/unit/modules/chain/chain.js
+++ b/framework/test/mocha/unit/modules/chain/chain.js
@@ -18,6 +18,9 @@
 const rewire = require('rewire');
 
 const Chain = rewire('../../../../../src/modules/chain/chain');
+const {
+	Synchronizer,
+} = require('../../../../../src/modules/chain/synchronizer/synchronizer');
 const { Processor } = require('../../../../../src/modules/chain/processor');
 const { BFT } = require('../../../../../src/modules/chain/bft');
 const {
@@ -35,6 +38,7 @@ describe('Chain', () => {
 		// Arrange
 
 		sinonSandbox.stub(Processor.prototype, 'init').resolves();
+		sinonSandbox.stub(Synchronizer.prototype, 'init').resolves();
 
 		/* Arranging Stubs start */
 		stubs.logger = {


framework/src/modules/chain/chain.js
```diff
@@ -253,6 +254,16 @@ module.exports = class Chain {
 					},
 				);
 			}
+
+			// Check if blocks are left in temp_block table
```
shuse2: why don't we do `await this.synchronizor.init()` instead of explicitly calling separate function?

framework/src/modules/chain/synchronizer/utils.js
```diff
@@ -42,6 +45,44 @@ const restoreBlocks = async (blocksModule, processorModule, tx) => {
...
+const restoreBlocksUponStartup = async (
+	blocksModule,
+	processorModule,
+	storageModule,
+) => {
+	// Get all blocks and find lowest height (next one to be applied)
+	const tempBlocks = await storageModule.entities.TempBlock.get();
+	const blockLowestHeight = tempBlocks.reduce((prev, current) =>
+		prev.height < current.height ? prev : current,
+	);
```
shuse2: initial value missing here

framework/src/modules/chain/synchronizer/utils.js
```diff
@@ -42,6 +45,44 @@ const restoreBlocks = async (blocksModule, processorModule, tx) => {
...
+	const nextTempBlock = parseBlockToJson(blockLowestHeight.fullBlock);
```
shuse2: do we need to do this step?
I think rather than `toJSON`, it should be deserialized to pass to the processor

framework/src/modules/chain/synchronizer/utils.js
```diff
@@ -42,6 +45,44 @@ const restoreBlocks = async (blocksModule, processorModule, tx) => {
...
+	const forkStatus = await processorModule.forkStatus(nextTempBlock);
```
shuse2: in this special case, don't we need to check `nextTempBlock.id === lastBlock.id` too?
https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#block-synchronization-mechanism

framework/src/modules/chain/chain.js
michielmulders: I also want to check if there are blocks left in the temp_block table.
This check can possibly be added to the `run` function for block_sync_mechanism?
I prefer this separated logic tho. @shuse2

framework/src/modules/chain/synchronizer/utils.js
michielmulders: Do you mean the init value for reducer? This is optional + I check in processor if there are blocks left in the temp_block table, if not, this function is not called.

framework/src/modules/chain/chain.js
shuse2: yea, i meant add
`init` function in `synchronizer.js` and check and run there?

framework/src/modules/chain/synchronizer/utils.js
shuse2: yea it is optional so it will be undefined. so the first run of the `prev` will be `undefined` and u r using `undefined.height`?

framework/src/modules/chain/chain.js
limiaspasdaniel: > yea, i meant add
> `init` function in `synchronizer.js` and check and run there?

This makes sense for me.

michielmulders: Does it make sense to still have this `restoreBlocksUponStartup` to separate logic but add a `init` function that calls this `restoreBlocksUponStartup` ?

limiaspasdaniel: Hmmm yeah II thiink it makes sense to separate the logic within Synchronizer to identify it better

framework/src/modules/chain/synchronizer/utils.js
michielmulders: I verified this, by default your initial value is set to the first object in the array so this won't happen.
+ defining a default value gives more overhead as you never know the hight of the blocks in the array as you have to set a very high initial value (made a mistake first by setting `{ height: 0}`.

framework/src/modules/chain/synchronizer/utils.js
```diff
@@ -42,6 +44,47 @@ const restoreBlocks = async (blocksModule, processorModule, tx) => {
...
+	const tempBlocks = await storageModule.entities.TempBlock.get({}, {}, {});
+	const blockLowestHeight = tempBlocks.reduce((prev, current) =>
+		prev.height < current.height ? prev : current,
+	);
+
+	const nextTempBlock = blockLowestHeight.fullBlock;
+	const forkStatus = await processorModule.forkStatus(nextTempBlock);
+	const inDifferentChain =
+		forkStatus === FORK_STATUS_DIFFERENT_CHAIN ||
+		blockLowestHeight.id === blocksModule.lastBlock.id;
```
limiaspasdaniel: Why `blockLowestHeight.id === blocksModule.lastBlock.id`?

michielmulders: https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#block-synchronization-mechanism For this check, correct?

limiaspasdaniel: Good catch! Actually had to update the logic on BSM PR

PR #4389, opened 2019-10-14
Remove delegate rank - Closes #4006

### LINKED ISSUE #4006: Update delegates ranks at the end of the round
### Description

On `undo` step of the round calculation, if the undoing block is the beginning of the round, `rank` of the mem_account should be updated.

### Motivation
In order to maintain the http_api endpoint, we need to keep the `rank` property calculation up to date.

### Acceptance Criteria
- it should not change the rank if the block is not first block of the round
- it should update the rank if the block is first block of the round

### Additional Information

shuse2:
### What was the problem?
Rank was not updated, and contains confusing information

### How did I solve it?
Remove Rank information from account and delegate API endpoint

### How to manually test it?
Call `/api/delegate` endpoints and observe rank does not exist

### Review checklist

- [ ] The PR resolves #4006 


shuse2:
@lsilvs good catch! I fixed in this PR

elements/lisk-api-client/src/resources/delegates.ts
```diff
@@ -31,15 +31,15 @@ export class DelegatesResource extends APIResource {
...
+				sort: 'username:asc',
 			},
 			method: GET,
 		}).bind(this);
 
 		this.getStandby = apiMethod({
 			defaultData: {
 				offset: 101,
-				sort: 'rank:asc',
+				sort: 'username:asc',
```
slaweet: AFAIK, here is a 🐛, @shuse2 

offset: 101, sorted by `username:asc` doesn't return standby delegates.

It could work with `sort: 'voteWeight:desc`

PR #4397, opened 2019-10-16
Make usage of delegate list BFT compliant - Closes #3681

### LINKED ISSUE #3681: Make usage of delegate list BFT compliant

lsilvs:
### What was the problem?
List of delegates used to verify block forger was different from the on used to forge.
The delegate list round offset was not being used to calculate delegates who missed the slot in order to update `missedBlocks` property.

### How did I solve it?
- I've added the property `delegateListRoundOffset` to apply/undo/getForgerPublicKeysForRound.
- Moved the usage of Dpos.apply/undo into each block processor allowing it to call with its own params.
- Added the config property `DELEGATE_LIST_ROUND_OFFSET` to set the `delegateListRoundOffset` default value for Dpos module

### Review checklist

- [ ] The PR resolves #3681


framework/src/modules/chain/block_processor_v0.js
```diff
@@ -126,7 +127,8 @@ class BlockProcessorV0 extends BaseBlockProcessor {
 					blockBytes,
 				}), // validate common block header
 			data => this.blocksModule.verifyInMemory(data),
-			({ block }) => this.dposModule.verifyBlockForger(block),
+			({ block }) =>
+				this.dposModule.verifyBlockForger(block, delegateListRoundOffset),
```
yatki: delegateListRoundOffset is kind of an option to the method. So maybe if pass it inside an object `this.dposModule.verifyBlockForger(block, {delegateListRoundOffset}),` would be more future proof? But we can also do that when there is a new option that comes up so we can leave it as it is now. I don't mind either way.

lsilvs: I thought about that but for apply/undo as we have two params as options. Gonna update that as well

framework/src/modules/chain/blocks/chain.js
```diff
@@ -184,32 +184,6 @@ const applyConfirmedGenesisStep = async (
...
-	block,
-	blockJSON,
-	skipSave,
-	tx,
-) => {
-	if (!skipSave) {
-		await saveBlock(storage, blockJSON, tx);
-	}
-
-	await dposModule.apply(block, tx);
```
shuse2: ❤️

framework/src/modules/chain/dpos/delegates_list.js
```diff
@@ -45,12 +45,16 @@ class DelegatesList {
...
-	 * Get shuffled list of active delegate public keys for a specific round -> forger public keys
+	 * Get shuffled list of active delegate public keys (forger public keys) for a specific round.
+	 * The list of delegates used is the one computed at the end of the round `r - delegateListRoundOffset`
 	 * @param {number} round
+	 * @param {number} delegateListRoundOffset
 	 */
-	async getForgerPublicKeysForRound(round) {
+	async getForgerPublicKeysForRound(round, delegateListRoundOffset) {
+		// Delegate list is generated from round 1 hence `roundWithOffset` can't be less than 1
+		const roundWithOffset = Math.max(round - delegateListRoundOffset, 1);
```
yatki: Do you think adding a check against negative offset values make sense? For instance, if round: 6 and I provide offset: -2 I can get the list for round 8. Just saying..

framework/src/modules/chain/processor/processor.js
```diff
@@ -293,34 +293,41 @@ class Processor {
...
 					block: cloneDeep(block),
 				});
 			}
+
+			if (!skipSave) {
+				const blockJSON = await this.serialize(block);
+				await this.blocksModule.save({ blockJSON, tx });
+			}
+
+			// Apply should always be executed after save as it performs database calculations
```
shuse2:
```suggestion
			// Apply should always be executed after save as it performs database calculations
			// because sumround in the dpos expects to have this processing block in the database in apply step
```

framework/src/modules/chain/dpos/dpos.js
```diff
@@ -20,8 +20,16 @@ const { DelegatesList } = require('./delegates_list');
...
-	constructor({ storage, slots, activeDelegates, logger, exceptions = {} }) {
+	constructor({
+		storage,
+		slots,
+		activeDelegates,
+		delegateListRoundOffset,
+		logger,
+		exceptions = {},
+	}) {
 		this.events = new EventEmitter();
```
yatki: I didn't thank for this before :D ❤️

framework/src/modules/chain/dpos/dpos.js
```diff
@@ -52,33 +60,46 @@ module.exports = class Dpos {
...
 	async onBlockFinalized({ height }) {
 		this.finalizedBlockRound = this.slots.calcRound(height);
 	}
 
 	async onRoundFinish() {
-		// TODO use the configuration variable to set the value of this variable
-		const delegateListOffsetForRound = 2;
 		const disposableDelegateList =
-			this.finalizedBlockRound - delegateListOffsetForRound;
+			this.finalizedBlockRound - this.delegateListRoundOffset;
```
yatki: same negative offset situation. Maybe we can just throw an error in the constructor?

framework/test/mocha/setup.js
```diff
@@ -56,6 +56,7 @@ if (process.env.LOG_DB_EVENTS === 'true') {
 
 testContext.config = config;
 testContext.config.constants = _.cloneDeep(app.constants);
+testContext.config.constants.DELEGATE_LIST_ROUND_OFFSET = 0;
```
yatki: Hmm, maybe adding a comment for setting this constant here would be nice, even I didn't understand it for a second, can you imagine :P :D

framework/src/modules/chain/dpos/delegates_list.js
lsilvs: That's a good point but I would prefer doing this in another ticket to avoid more changes and tests broken in this PR 😬

framework/src/modules/chain/dpos/dpos.js
lsilvs: In this case the constructor is receiving this from config file which validates the value as grater than zero

PR #4411, opened 2019-10-21
Bubble up the peerId when a P2P message packet is received - Closes #4395

### LINKED ISSUE #4395: Add peerId on Message and Response packet
### Description

In order to ban and retarget the peer, `P2PMessagePacket` and `P2PResponsePacket` should have `peerId` information.
It should be propagated to the library user.

It should be
```
{
  data: unknown,
  peerId: string,
}
```

### Motivation

applying penalty and re-targeting requires peerId

### Acceptance Criteria

- Response of the RPC and the event should have `peerId` along with `data`

diego-G:
### Review checklist

- [x] The PR resolves #4395

### DIFF
--- a/elements/lisk-p2p/src/p2p_types.ts
+++ b/elements/lisk-p2p/src/p2p_types.ts
@@ -31,6 +31,25 @@ export interface P2PMessagePacket extends P2PPacket {
 	readonly event: string;
 }
 
+export interface P2PClosePacket {
+	readonly peerInfo: P2PPeerInfo;
+	readonly code: number;
+	readonly reason?: string;
+}
+
+// TODO later: Switch to LIP protocol format.
+export interface ProtocolRPCRequestPacket {
+	readonly data: unknown;
+	readonly procedure: string;
+	readonly type: string;
+}
+
+// TODO later: Switch to LIP protocol format.
+export interface ProtocolMessagePacket {
+	readonly data: unknown;
+	readonly event: string;
+}
+
 export interface P2PPenalty {
 	readonly peerId: string;
 	readonly penalty: number;
@@ -101,11 +120,6 @@ export interface ProtocolNodeInfo extends P2PSharedState {
 	readonly nonce: string;
 	readonly wsPort: number;
 	readonly httpPort: number;
-}
-export interface P2PClosePacket {
-	readonly peerInfo: P2PPeerInfo;
-	readonly code: number;
-	readonly reason?: string;
 }
 
 export interface P2PConfig {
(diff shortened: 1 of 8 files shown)


elements/lisk-p2p/src/p2p_types.ts
```diff
@@ -13,24 +13,40 @@
...
 	readonly procedure: string;
 }
 
-export interface P2PResponsePacket extends P2PPacket {
+export interface P2PResponsePacket {
 	readonly data: unknown;
 }
 
-export interface P2PMessagePacket extends P2PPacket {
+export interface P2PMessagePacket {
```
shuse2: wouldn't it be easier to just include `peerId` in this packet?

elements/lisk-p2p/src/p2p_types.ts
```diff
@@ -13,24 +13,40 @@
...
-}
 
-export interface P2PRequestPacket extends P2PPacket {
+export interface P2PRequestPacket {
 	readonly data?: unknown;
 	readonly procedure: string;
 }
 
-export interface P2PResponsePacket extends P2PPacket {
+export interface P2PResponsePacket {
```
shuse2: It will be nice to have it here too

diego-G: This interface is mainly used in P2PRequest class which intrinsically includes the peer ID.

elements/lisk-p2p/src/p2p_types.ts
diego-G: That is the first option that I've assessed as it is indicated in the issue. The problem is that P2PMessagePacket interface is used as well in the `send()` from peer pool. It doesn't make sense to include the peerId in the message because it would be redundant. When you receive a package from a peer, we are able to grab the peerId at that moment. There is no need to include this information in the package itself.

elements/lisk-p2p/src/p2p_types.ts
shuse2: I meant it would be nice to have consistently returning the peerId on return action here
https://github.com/LiskHQ/lisk-sdk/blob/f5e2e3428957cd96eafc119dcba7a76d0bd99264/elements/lisk-p2p/src/p2p.ts#L610

shuse2: and probably https://github.com/LiskHQ/lisk-sdk/blob/f5e2e3428957cd96eafc119dcba7a76d0bd99264/elements/lisk-p2p/src/p2p.ts#L620
too (this one is kinda redundant)

elements/lisk-p2p/src/p2p_types.ts
shuse2: I think it make sense to separate them if the `send()` is using it.
but for user, it would be nice to have similar response interface for send and request if above comment is handled

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -237,7 +237,7 @@ export class Peer extends EventEmitter {
 			const rate = this._getMessageRate(message);
 			const messageWithRateInfo = {
 				...message,
-				peerId: this._id,
+				peerId: this.id,
```
shuse2: probably this doesn't need to change. Prefer to use private property over public one internally

PR #4414, opened 2019-10-21
Request and apply blocks in chunks for BSM - Closes #4412

### LINKED ISSUE #4412: Request and apply blocks in chunks for BSM
### Description

Currently, BSM requests a set of blocks within a defined pair of block IDs and receives in chunks of `34` blocks repeatedly until the desired block ID is reached.  Then, all the blocks are applied.

We can do this more efficiently by applying every chunk of 34 blocks after it has been received, this way, we can still apply some blocks in case future chunks are not valid or the peer doesn't return any further blocks, 

### Motivation

Applying blocks after every chunk of 34 blocks has been received instead of waiting to receive all blocks guarantees that we can still apply some chunks even if future chunks are invalid.

### Acceptance Criteria

- Blocks are requested in chunks and applied immediately after instead of waiting to receive all of them.

limiaspasdaniel:
### What was the problem?
Refer to #4412 
### How did I solve it?
Refer to #4412 
### How to manually test it?
Will add tests in a separate issue that tackles tests for BSM
### Review checklist

- [ ] The PR resolves #4412 

### DIFF
--- a/framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
+++ b/framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
@@ -107,13 +107,12 @@ class BlockSynchronizationMechanism extends BaseSynchronizer {
 	}
 
 	/**
-	 * Request blocks from `fromID` ID to `toID` ID from an specific peer `peer` and applieis them
+	 * Request blocks from `fromID` ID to `toID` ID from an specific peer `peer` and applies them
 	 *
 	 * @param {object} peerId - The ID of the peer to target
 	 * @param {string} fromId - The starting block ID to fetch from
 	 * @param {string} toId - The ending block ID
 	 * @throws {ApplyPenaltyAndRestartError} - In case the peer hasn't returned any block
-	 * @return {Promise<Array<object>>}
 	 */
 	async _requestAndApplyBlocksWithinIDs(peerId, fromId, toId) {
 		const maxFailedAttempts = 10; // TODO: Probably expose this to the configuration layer?
--- a/framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
+++ b/framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
@@ -151,7 +151,7 @@ class BlockSynchronizationMechanism extends BaseSynchronizer {
 					await this.processorModule.process(deserializedBlock);
 				}
 
-				finished = this.blocks.lastBlock.id === toId || false;
+				finished = this.blocks.lastBlock.id === toId;
 			} else {
 				failedAttempts += 1; // It's only considered a failed attempt if the target peer doesn't provide any blocks on a single request
 			}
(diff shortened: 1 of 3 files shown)


framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -106,6 +106,65 @@ class BlockSynchronizationMechanism extends BaseSynchronizer {
...
+				);
+
+				for (const block of blocks) {
+					const deserializedBlock = await this.processorModule.deserialize(
+						block,
+					);
+					await this.processorModule.process(deserializedBlock);
+				}
+
+				finished = this.blocks.lastBlock.id === toId || false;
```
shuse2: I think first condition is always boolean?

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -133,35 +192,15 @@ class BlockSynchronizationMechanism extends BaseSynchronizer {
 			'Requesting blocks within ID range from peer',
 		);
 
-		const listOfFullBlocks = await this._requestBlocksWithinIDs(
+		await this._requestAndApplyBlocksWithinIDs(
```
shuse2: original function `_requestBlocksWithinIDs` can be move to the fast chain mechanism from base mechanism

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -106,6 +106,65 @@ class BlockSynchronizationMechanism extends BaseSynchronizer {
...
+	 * @throws {ApplyPenaltyAndRestartError} - In case the peer hasn't returned any block
+	 */
+	async _requestAndApplyBlocksWithinIDs(peerId, fromId, toId) {
+		const maxFailedAttempts = 10; // TODO: Probably expose this to the configuration layer?
+		let failedAttempts = 0; // Failed attempt === the peer doesn't return any block or there is a network failure (no response or takes too long to answer)
+		let lastFetchedID = fromId;
+		let finished = false;
+
+		while (!finished && failedAttempts < maxFailedAttempts) {
+			const { data: blocks } = await this.channel.invoke(
```
shuse2: in case of this request fails or network error, it will throw and exit without restarting

limiaspasdaniel: @shuse2 let's tackle this in another issue, as we're currently not handling network timeouts or request failures for both BSM and FCS, and because this is a direct cause for banning, we should spike it.

framework/src/modules/chain/synchronizer/fast_chain_switching_mechanism.js
```diff
@@ -133,6 +133,44 @@ class FastChainSwitchingMechanism extends BaseSynchronizer {
...
+			const { data } = await this.channel.invoke('network:requestFromPeer', {
+				procedure: 'getBlocksFromId',
+				peerId,
+				data: {
+					blockId: lastFetchedID,
+				},
+			}); // Note that the block matching lastFetchedID is not returned but only higher blocks.
+
+			if (data && data.length) {
+				blocks.push(...data); // `data` is an array of blocks.
```
lsilvs: I would replace this comment for a meaningful variable name

limiaspasdaniel: Done! :) Thanks for the suggestion, makes sense.

PR #4419, opened 2019-10-22
Merge dpos improvments into introduce bft

lsilvs:
### What was the problem?
Slow sync in `feature/introduce_bft_consensus` branch

### How did I solve it?
Some improvements were done in https://github.com/LiskHQ/lisk-sdk/issues/4340 but the sync is still slow.
Merging the improvements back to the branch but the issue is still open

### How to manually test it?
--

### Review checklist

- [ ] The PR resolves #INSERT_ISSUE_NUMBER

### DIFF
--- a/framework/src/modules/chain/transport/transport.js
+++ b/framework/src/modules/chain/transport/transport.js
@@ -247,11 +247,11 @@ class Transport {
 			);
 		}
 
-		// TODO: endpoint should be protected before
-		if (this.synchronizer.isActive) {
+		// Should ignore received block if syncing
+		if (this.synchronizer.isActive()) {
 			return this.logger.debug(
-				"Client is syncing. Can't receive block at the moment.",
-				query.block.id,
+				{ id: query.block.id, height: query.block.height },
+				"Client is syncing. Can't process new block at the moment.",
 			);
 		}
 
--- a/framework/src/modules/chain/transport/transport.js
+++ b/framework/src/modules/chain/transport/transport.js
@@ -248,9 +248,9 @@ class Transport {
 		}
 
 		// Should ignore received block if syncing
-		if (this.synchronizer.isActive()) {
+		if (this.synchronizer.isActive) {
 			return this.logger.debug(
-				{ id: query.block.id, height: query.block.height },
+				{ blockId: query.block.id, height: query.block.height },
 				"Client is syncing. Can't process new block at the moment.",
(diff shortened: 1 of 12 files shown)


framework/src/modules/chain/transport/transport.js
```diff
@@ -247,11 +247,11 @@ class Transport {
 			);
 		}
 
-		// TODO: endpoint should be protected before
-		if (this.synchronizer.isActive) {
+		// Should ignore received block if syncing
+		if (this.synchronizer.isActive()) {
```
shuse2: `isActive` is getter so it shouldn't need `()`

limiaspasdaniel: Which now makes me think we shouldn't include verbs in getter functions

shuse2: ah yea that's true

limiaspasdaniel: So or either `synchronizer.active` accesses the `active` variable, or `synchronizer.isActive()` IMHO

framework/src/modules/chain/transport/transport.js
```diff
@@ -247,11 +247,11 @@ class Transport {
...
 			return this.logger.debug(
-				"Client is syncing. Can't receive block at the moment.",
-				query.block.id,
+				{ id: query.block.id, height: query.block.height },
```
limiaspasdaniel:
```suggestion
				{ blockId: query.block.id, height: query.block.height },
```

I think per convention we're using `blockId` instead of just `id` in most places.

framework/src/modules/chain/transport/transport.js
shuse2: variable which we don't want to be changed from outside, so i think it make sense to use method?

lsilvs: Yeah, `isSomething` should be function IMO

limiaspasdaniel: But then we can declare a getter like:
```javascript
get active() {
    return this._active;
}
```

This way you set `active` to "private" and IDEs won't show it as a public property when accessing the synchronizer.

lsilvs: I've removed the `()` as refactoring synchronizer is out of scope of this PR

Issue #3344, opened 2018-11-22
Add lazy push mechanism to reduce overall bandwidth consumption in the network


shuse2:
### Description

This mechanism of first sending only the message identifier and not the full payload is referred to as lazy push.

**Transaction Propagation:** As unconfirmed transactions are not as time critical as blocks, we propose to switch to a lazy push mechanism to broadcast transaction. This means at regular intervals the transactions selected by the transaction pool mechanism of a node will be announced (meesage name: `postTransactionsAnnouncement`) to all connected peers by sending a list of transaction IDs. The peers can then request the full transactions explicitly from the Lisk node.

This involves both the advertising (e.g. pushing the block/transaction IDs to peers) and also the request part where the peer requests for the advertised block or transaction.

General requirements:
- [x] Create broadcast P2P public method + adding a broadcast action to Network module #4423

Lazy push related to Transactions:
- [ ] Trigger `postTransactionsAnnouncement` network event in chain module.
- [ ] Act when receiving a `postTransactionsAnnouncement` event: create checkTransactionsAnnouncement method in chain module.
- [ ] Trigger `getTransactions` network event in chain module.
- [ ] Act when receiving a `getTransactions` event.

Lazy push related to Blocks:
- [ ] TBD

sridharmeganathan:
STM is completed and the team has a fair understanding of the implementation.

4miners:
@MaciejBaj @shuse2 @sridharmeganathan

Before considering the implementation of this issue we need to ensure that following steps will be done:
- we know the actual propagation times for current P2P implementation (`2.0.0`)
- we know the propagation times after the implementation of `LIP 0004`, so we exactly know the effect of it on the network
- we know the impact of lazy push on the propagation or at least we are able to make an assumption that is backed by some real data (simulations)
- after the implementation of this issue, we have to confirm that the actual results match our expectations and network propagation was not degraded significantly
- acceptance criteria are defined (for example worst acceptable propagation time on the network with `X` nodes)

I would recommend implementing LIP `Introduce robust peer selection and banning mechanism` without lazy push mechanism first, otherwise we will not be able to determine its impact.

jondubois:
@4miners Good points. I agree, this part of the LIP seems high risk. That's why I moved it into the second sprint/milestone but I would definitely prefer to consider it as a 'spike' rather than as a definite feature.

[...]

PR #4423, opened 2019-10-23
Create P2P broadcast method - Related to #3344

diego-G:
# How to manually test it?

`elements/lisk-p2p$ npm run test:functional`

### Review checklist

- [x] The PR related to #3344


framework/src/modules/chain/chain.js
```diff
@@ -235,7 +235,7 @@ module.exports = class Chain {
 			if (!this.options.loading.rebuildUpToRound) {
 				this.channel.subscribe(
 					'network:event',
-					async ({ data: { event, data, peerId } }) => {
+					async ({ data: { event, data } }) => {
```
limiaspasdaniel: Does this change mean that `peerId` is at the same level as the data sent by the peer? Just asking.

shuse2: I think original code here was correct.
it `event` and `data` and `peerId` is at the same level

diego-G: Right guys. I got confused because we have two datas in the same line. I've changed the variable name to avoid variable shadowing in the future.

framework/src/modules/chain/chain.js
```diff
@@ -235,7 +235,7 @@ module.exports = class Chain {
 			if (!this.options.loading.rebuildUpToRound) {
 				this.channel.subscribe(
 					'network:event',
-					async ({ data: { event, data, peerId } }) => {
+					async ({ packet: { event, data, peerId } }) => {
```
shuse2: you cannot do this because `data` is wrapped object from the channel.
In this case, root level `data` will not be variable you can use, so it won't be shadowed too

elements/lisk-p2p/test/functional/broadcast.ts
```diff
@@ -0,0 +1,86 @@
...
+	});
+
+	it('should receive a message in the correct format', async () => {
+		const firstP2PNode = p2pNodeList[0];
+		firstP2PNode.broadcast({ event: 'bar', data: 'test' });
+
+		await wait(100);
+
+		expect(collectedMessages).to.be.an('array');
+		expect(collectedMessages.length).to.be.eql(9);
```
ishantiw: There is no context why it should be 9 maybe use `networkSize - 1` so in case we change the number of nodes in the network this will fail.

framework/src/modules/network/index.js
```diff
@@ -60,6 +60,9 @@ module.exports = class NetworkModule extends BaseModule {
 			emit: {
```
ishantiw: Can we change this to `send`?

PR #4425, opened 2019-10-23
Add IP limit for P2P Outbound connections - Closes #4315

### LINKED ISSUE #4315: Do not connect to a peer if we already have an outbound peer with that IP
### Description

Right now we allow connecting to multiple outbound peers from the same IP address but in the logic related to outbound peers, we ignore non-unique peers anyway. 

For example, the `PeerPool` `request` method uses `this.getUniqueOutboundConnectedPeers()` to filter out duplicates.

It would be simpler and cleaner to just not connect to multiple peers from the same IP to begin with. We should be able to remove `this.getUniqueOutboundConnectedPeers()`. Also using `height` to decide which is the best peer is opinionated and is outside the scope of the P2P library.

Note that avoiding multiple peers from the same IP should only apply to outbound peers at this stage. As it stands currently, we should continue to allow it for inbound peers.

### Motivation

Cleaner logic in the `PeerPool`.

### Acceptance Criteria

- Node does not connect to multiple peers from the same IP.
- The `getUniqueOutboundConnectedPeers` method is removed from `PeerPool`.
- The `getUniquePeersbyIp` util function is removed.

valamidev:
### What was the problem?

We allowed to the Node to connect on different ports on the same IP, it could also mean that the target IP has multiple Nodes running or it has multiple registry in the PeerList. With implementing this IP limit we can also simplify the P2P Request / Send actions and improve shape of the network connections. 
Peer selection should also improved to choose the best height possible from same IP.

### How did I solve it?

Extended the default selectPeersForConnection() in the [elements/lisk-p2p/src/utils/select.ts](elements/lisk-p2p/src/utils/select.ts) with a new private function `_removeCommonIPsFromLists`.  
Add a new logical check into the [elements/lisk-p2p/src/peer_pool.ts](elements/lisk-p2p/src/peer_pool.ts) `PeerPool.addOutboundPeer()` method to avoid creating multiple connections for the same IP, exception for localhost `127.0.0.1`  [elements/lisk-p2p/src/constants.ts](elements/lisk-p2p/src/constants.ts) exist to allow localhost testing or creating proxy Nodes.

### How to manually test it?

npm run test

### Review checklist

- [x] The PR resolves #4315


elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -413,23 +413,30 @@ export class PeerPool extends EventEmitter {
 			disconnectedFixedPeers.length -
 			outboundCount;
 
+		const bestPeersbyHeight = getBestPeersbyHeight([
```
ishantiw: prefer `getUniquePeersByIP` as we are going to remove the usage of height in future

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -413,23 +413,30 @@ export class PeerPool extends EventEmitter {
...
-		[...peersToConnect, ...disconnectedFixedPeers].forEach(
-			(peerInfo: P2PPeerInfo) => {
-				const existingPeer = this.getPeer(peerInfo.peerId);
-
-				return existingPeer
-					? existingPeer
-					: this.addOutboundPeer(peerInfo.peerId, peerInfo);
-			},
-		);
+		comparePeerListsByHeight(
```
ishantiw: Only use the function here `getBestPeersbyHeight([...peersToConnect, ...disconnectedFixedPeers])`. I think its enough

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -413,23 +413,30 @@ export class PeerPool extends EventEmitter {
...
+			...disconnectedNewPeers,
+			...disconnectedTriedPeers,
+			...disconnectedFixedPeers,
+		]);
+
 		// This function can be customized so we should pass as much info as possible.
 		const peersToConnect = this._peerSelectForConnection({
-			newPeers: disconnectedNewPeers,
-			triedPeers: disconnectedTriedPeers,
+			newPeers: comparePeerListsByHeight(
```
ishantiw: we don't this here, we can simply filter out the duplicates in the end while making an inbound

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -455,10 +462,19 @@ export class PeerPool extends EventEmitter {
 		return peer;
 	}
 
-	public addOutboundPeer(peerId: string, peerInfo: P2PPeerInfo): Peer {
-		const existingPeer = this.getPeer(peerId);
+	public addOutboundPeer(peerInfo: P2PPeerInfo): P2PPeerInfo {
+		const existingPeer = this.getPeer(peerInfo.peerId);
 		if (existingPeer) {
-			return existingPeer;
+			return peerInfo;
```
ishantiw: its not a `peerInfo`, its an Peer object

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -455,10 +462,19 @@ export class PeerPool extends EventEmitter {
 		return peer;
 	}
 
-	public addOutboundPeer(peerId: string, peerInfo: P2PPeerInfo): Peer {
-		const existingPeer = this.getPeer(peerId);
+	public addOutboundPeer(peerInfo: P2PPeerInfo): P2PPeerInfo {
```
ishantiw: Return Peer Class type

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -455,10 +462,19 @@ export class PeerPool extends EventEmitter {
...
+		const existingPeer = this.getPeer(peerInfo.peerId);
 		if (existingPeer) {
-			return existingPeer;
+			return peerInfo;
+		}
+
+		const connectedPeersFromIP = this.getConnectedPeersByIP(
```
ishantiw: Add a short comment what we are doing here

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -518,8 +534,14 @@ export class PeerPool extends EventEmitter {
 		return peers;
 	}
 
-	public getUniqueOutboundConnectedPeers(): ReadonlyArray<P2PPeerInfo> {
-		return getUniquePeersbyIp(this.getAllConnectedPeerInfos(OutboundPeer));
+	public getConnectedPeersByIP(
```
ishantiw: Instead of returning count simply return if a peer outbound exists or not because the count will always be one if we never allow the second outbound connection

elements/lisk-p2p/src/utils/misc.ts
```diff
@@ -121,7 +121,7 @@ export const getNetgroup = (address: string, secret: number): number => {
 };
 
 // TODO: Remove the usage of height for choosing among peers having same ip, instead use productivity and reputation
-export const getUniquePeersbyIp = (
+export const getBestPeersbyHeight = (
```
ishantiw: Keep the old name `getUniquePeersbyIp ` as we are not going to rely on height in future

elements/lisk-p2p/src/utils/misc.ts
```diff
@@ -152,6 +152,40 @@ export const getUniquePeersbyIp = (
 	return [...peerMap.values()];
 };
 
+export const comparePeerListsByHeight = (
```
ishantiw: same here, `removeCommonIPsFromLists` or something more generic

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -530,8 +536,16 @@ export class PeerPool extends EventEmitter {
 		return peers;
 	}
 
-	public getUniqueOutboundConnectedPeers(): ReadonlyArray<P2PPeerInfo> {
-		return getUniquePeersbyIp(this.getAllConnectedPeerInfos(OutboundPeer));
+	public getIsConnectedPeerByIP(
```
diego-G: I would call it `getConnectedPeersByFilter` and as argument, a new type with the accepted filters must be created. For now, there should be two filters by kind or by IP. Therefore, we don't need to pass the whole P2PPeerInfo object, just the IP string.

elements/lisk-p2p/src/utils/misc.ts
```diff
@@ -152,6 +152,40 @@ export const getUniquePeersbyIp = (
 	return [...peerMap.values()];
 };
 
+export const removeCommonIPsFromLists = (
```
diego-G: This method is using height which is a concept that must be removed from the library. Please, read the description of the issue, it is stated there.

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -530,8 +536,16 @@ export class PeerPool extends EventEmitter {
...
+		peerInfo: P2PPeerInfo,
+		kind: typeof OutboundPeer | typeof InboundPeer,
+	): boolean {
+		const connectedPeers = this.getAllConnectedPeerInfos(kind);
+
+		return connectedPeers.filter(peer => peer.ipAddress === peerInfo.ipAddress)
+			.length > 0
+			? true
```
ManuGowda: the expression `length > 0` already returns a boolean value, so you can remove the conditional ternary operator.

elements/lisk-p2p/src/utils/misc.ts
```diff
@@ -152,6 +152,40 @@ export const getUniquePeersbyIp = (
...
+	peerList: ReadonlyArray<P2PPeerInfo>,
+	controlPeerList: ReadonlyArray<P2PPeerInfo>,
+): ReadonlyArray<P2PPeerInfo> => {
+	const peerMap = new Map<string, P2PPeerInfo>();
+
+	for (const controlPeer of controlPeerList) {
```
ManuGowda: I can not think of a better way to improve this code for now, but let's catch up and see if we can make it better.

framework/src/modules/network/network.js
```diff
@@ -385,18 +385,9 @@ module.exports = class Network {
...
-			getUniqueOutboundConnectedPeersCount: action => {
-				const peers = consolidatePeers({
-					connectedPeers: this.p2p.getUniqueOutboundConnectedPeers(),
-				});
-
-				const { limit, offset, ...filterWithoutLimitOffset } = action.params;
-
-				return filterByParams(peers, filterWithoutLimitOffset).length;
-			},
 			getUniqueOutboundConnectedPeers: () => {
```
diego-G: Let's remove this one too as discussed.

elements/lisk-p2p/src/p2p.ts
```diff
@@ -567,9 +568,9 @@ export class P2P extends EventEmitter {
 		}));
 	}
 	// Make sure you always share shared peer state to a user
-	public getUniqueOutboundConnectedPeers(): ReadonlyArray<ProtocolPeerInfo> {
+	public getOutboundConnectedPeers(): ReadonlyArray<ProtocolPeerInfo> {
```
diego-G: Please, remove this method as we discussed. No-one is using it for now.

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -467,10 +468,18 @@ export class PeerPool extends EventEmitter {
...
-		const existingPeer = this.getPeer(peerId);
+	public addOutboundPeer(peerInfo: P2PPeerInfo): boolean {
+		const existingPeer = this.getPeer(peerInfo.peerId);
 		if (existingPeer) {
-			return existingPeer;
+			return false;
+		}
+
+		// Check if we got already Outbound connection into the IP address of the Peer
+		const connectedPeer = this.getAllConnectedPeerInfos(OutboundPeer).find(
```
diego-G: Rename variable to `outboundConnectedPeer`.

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -467,10 +468,18 @@ export class PeerPool extends EventEmitter {
 		return peer;
 	}
 
-	public addOutboundPeer(peerId: string, peerInfo: P2PPeerInfo): Peer {
-		const existingPeer = this.getPeer(peerId);
+	public addOutboundPeer(peerInfo: P2PPeerInfo): boolean {
```
diego-G: This method should become private.

elements/lisk-p2p/test/functional/actions/request.ts
```diff
@@ -63,23 +64,21 @@ describe('P2P.request', () => {
...
 		expect(response.data)
 			.to.have.property('requestPeerId')
-			.which.is.equal(`127.0.0.1:${secondP2PNode.nodeInfo.wsPort}`);
+			.which.is.equal(`127.0.0.2:${secondP2PNode.nodeInfo.wsPort}`);
 	});
 
 	// Check for even distribution of requests across the network. Account for an error margin.
-	// TODO: Skipping this test as of now because we are removing duplicate IPs so this scenario will not work locally
-	// TODO: #3389 Improve network test to be fast and stable, it can fail randomly depend on network shuffle
-	it.skip('requests made to the network should be distributed randomly', async () => {
```
diego-G: Great! Thanks for bringing back this skipt tests!

elements/lisk-p2p/test/utils/network_setup.ts
```diff
@@ -57,6 +63,15 @@ export const createNetwork = async ({
 	networkDiscoveryWaitTime,
 	customConfig,
 }: TestNetworkConfig = {}) => {
+	const serverSocketPrototype = SCServerSocket.prototype as any;
```
diego-G: I don't like to bring this to the general setup. I'd prefer to keep it just where it's needed. Would it be possible?

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -468,17 +468,17 @@ export class PeerPool extends EventEmitter {
 		return peer;
 	}
 
-	public addOutboundPeer(peerInfo: P2PPeerInfo): boolean {
+	private addOutboundPeer(peerInfo: P2PPeerInfo): boolean {
```
diego-G: When a function is private we write `_` in front.

elements/lisk-p2p/test/functional/actions/send.ts
```diff
@@ -84,12 +84,15 @@ describe('P2P.send', () => {
 
 	it('should receive a message in the correct format', async () => {
 		const firstP2PNode = p2pNodeList[0];
+
+		const connectedPeers = firstP2PNode.getConnectedPeers().length;
+
 		firstP2PNode.send({ event: 'bar', data: 'test' });
 
-		await wait(100);
+		await wait(300);
```
yatki: I think setting arbitrary delays for testing creates brittle tests. Because what if the action takes a bit longer. If not why wasting time waiting :D I think, there should be a better way to make sure the message was transferred.

ManuGowda:
@valamidev the PR has conflicts

elements/lisk-p2p/test/functional/actions/send.ts
valamidev: I found that, some unreliable (we have no lifetime check after we emiting it into the network) operation like the send better to add some wait() to be sure that every peer receive and handle it. This could be ~5-100ms. 300ms is a safe value.

elements/lisk-p2p/test/functional/actions/send_to_peer.ts
```diff
@@ -42,28 +31,31 @@ describe('P2P.sendToPeer', () => {
...
-			.to.have.property('nodePort')
-			.which.is.equal(targetPeer.wsPort);
-		expect(collectedMessages[0]).to.have.property('message');
-		expect(collectedMessages[0].message)
-			.to.have.property('event')
-			.which.is.equal('foo');
-		expect(collectedMessages[0].message)
-			.to.have.property('data')
-			.which.is.equal(123);
+		p2pNodeList[1].on('messageReceived', msg => {
```
yatki: this is a better approach. Just you need to call `done` in the callback because this test would always pass as soon as it's done with the execution of top-level code.

elements/lisk-p2p/src/constants.ts
```diff
@@ -16,6 +16,7 @@ import { getRandomBytes } from '@liskhq/lisk-cryptography';
 
 // General P2P constants
 export const DEFAULT_NODE_HOST_IP = '0.0.0.0';
+export const LOCALHOST_IP = '127.0.0.1';
```
diego-G: Add `DEFAULT` here.

elements/lisk-p2p/src/utils/select.ts
```diff
@@ -106,19 +135,22 @@ export const selectPeersForConnection = (
 	const shuffledTriedPeers = shuffle(input.triedPeers);
 	const shuffledNewPeers = shuffle(input.newPeers);
 
-	return [...Array(input.peerLimit)].map(() => {
-		if (shuffledTriedPeers.length !== 0) {
-			if (Math.random() < r) {
-				// With probability r
-				return shuffledTriedPeers.pop() as P2PPeerInfo;
+	// TODO: Remove the usage of height for choosing among peers having same ip, instead use productivity and reputation
```
diego-G: Please, increase unit test coverage here.

elements/lisk-p2p/test/unit/utils/misc.ts
```diff
@@ -139,77 +136,6 @@ describe('utils/misc', () => {
 		});
 	});
 
-	describe('#getUniquePeersbyIp', () => {
```
diego-G: These unit tests must be replaced by another.

elements/lisk-p2p/test/functional/actions/send.ts
```diff
@@ -83,12 +83,15 @@ describe('P2P.send', () => {
 
 	it('should receive a message in the correct format', async () => {
 		const firstP2PNode = p2pNodeList[0];
+
+		const connectedPeers = firstP2PNode.getConnectedPeers().length;
```
ishantiw: `numOfConnectedPeers`

elements/lisk-p2p/test/integration/discovery.ts
```diff
@@ -78,6 +79,7 @@ describe('Peer discovery', () => {
 
 	after(async () => {
 		await disconnectedNode.stop();
+		await wait(1000);
```
ishantiw: Since we are only stopping one node we can give ~200 ms only

elements/lisk-p2p/test/integration/outbound_shuffling.ts
```diff
@@ -42,27 +42,43 @@ describe('Outbound peer shuffling', () => {
...
-		// Wait for periodic shuffling
+	it('should shuffle outbound peers and close connection with evict', async () => {
 		await wait(500);
-		const {
-			outboundCount: updatedOutbound,
-		} = (p2pNode as any)._peerPool.getPeersCountPerKind();
 
-		expect(updatedOutbound).lt(outboundCount);
+		p2pNodeList.forEach(p2p => {
+			let evictedConnections = collectedEventsCount.get(p2p.nodeInfo.wsPort);
```
ishantiw: `const`

PR #4426, opened 2019-10-24
Reject unnecessary messages - Closes #4424

### LINKED ISSUE #4424: Reject unnecessary messages and ban a peer
### Expected behavior

A node should not send internal messages to a lisk node as it's not needed which creates an extra overhead.

### Actual behavior

Its accepting messages that are internal and not related to `lisk-sdk` and should be rejected and a peer should be banned as its not suppose to send these messages.

`3.0`

ishantiw:
### What was the problem?

Peers can send unnecessary or internal messages that are not part of the protocol.

### How did I solve it?

Rejecting the message and banning the peer.

### How to manually test it?

Send messages like `INVALID` or something and see if the receiving node rejects the connection

### Review checklist

- [x] The PR resolves #4424


elements/lisk-p2p/src/constants.ts
```diff
@@ -55,6 +55,8 @@ export const DEFAULT_PRODUCTIVITY = {
 // Peer inbound constants
 export const DEFAULT_PING_INTERVAL_MAX = 60000;
 export const DEFAULT_PING_INTERVAL_MIN = 20000;
+export const SOCKET_PING_MESSAGE = '#1';
```
diego-G: Where is this coming from? It looks too specific. I would suggest to have a regEx capturing similar cases.

elements/lisk-p2p/src/constants.ts
```diff
@@ -66,6 +68,9 @@ export const DEFAULT_MAX_RECONNECT_TRIES = 3;
 
 // Disconnect statuses
 export const INTENTIONAL_DISCONNECT_CODE = 1000;
+export const INVALID_MESSAGE_RECEIVED_CODE = 4009;
```
diego-G: Why 4009 instead of 4109 or 4409?

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -498,6 +525,7 @@ export class Peer extends EventEmitter {
 		this._wsMessageRate =
 			(this._wsMessageCount * RATE_NORMALIZATION_FACTOR) / this._rateInterval;
 		this._wsMessageCount = 0;
+		this._badMessageCount = 0;
```
diego-G: I suggest to use a more meaningful naming. What is a bad message?

elements/lisk-p2p/src/p2p.ts
```diff
@@ -640,6 +640,20 @@ export class P2P extends EventEmitter {
 	}
 
 	private async _startPeerServer(): Promise<void> {
+		this._scServer.on(
+			'handshake',
```
diego-G: Shall we move here the blacklist check present in `connection` event?

elements/lisk-p2p/test/integration/reject_unwanted_events.ts
```diff
@@ -0,0 +1,77 @@
...
+		}
+
+		for (const p2p of p2pNodeList) {
+			if (p2p.nodeInfo.wsPort === connectedPeerOfNode2.wsPort) {
+				const peerOfNode1 = firstNode['_peerPool']['_peerMap'].get(
+					`${SEED_PEER_IP}:${secondNode.nodeInfo.wsPort}`,
+				);
+				if (peerOfNode1 && peerOfNode1['_socket']) {
+					for (let num = 0; num < 50; num++) {
+						peerOfNode1['_socket'].send(INVALID_MESSAGE_1, {});
```
diego-G: We only reject that specific message '#1'. The PR states different `Send messages like INVALID or something and see if the receiving node rejects the connection`

elements/lisk-p2p/src/constants.ts
ishantiw: Its a socketcluster specific close code on invalid raw messages,
https://github.com/SocketCluster/socketcluster-server/blob/master/scserversocket.js#L106

elements/lisk-p2p/src/constants.ts
```diff
@@ -66,6 +68,9 @@ export const DEFAULT_MAX_RECONNECT_TRIES = 3;
...
+export const INVALID_MESSAGE_RECEIVED_REASON =
+	'Peer should not send bad or internal messages';
```
diego-G:
```suggestion
	'Peer should not send invalid or internal messages';
```

elements/lisk-p2p/src/p2p.ts
diego-G: Isn't it redundant now? My suggestion was to move the code...

PR #4427, opened 2019-10-24
Implement unit tests for Synchronizer class - Closes #4428

### LINKED ISSUE #4428: Cover Synchronizer class with unit tests
### Description
Synchronizer class needs unit tests, to check whether its behavior works as intended
### Acceptance Criteria
The following 
-  `init` checks whether the blocks temp table is empty. If it is, the table is cleared. If it is not, it restores the blocks from it to the main block table if temp table tip has preference over main table tip.
- Required arguments for `run` methods are asserted and an error is thrown if so. This is required as the function input comes directly from the outside (network) and we don't have control over it.
- Registering a mechanism adds a new mechanism to the array of mechanisms and validates its required public interfaces
- Determining the suitable mechanism for the given block guarantees that two mechanisms are not valid at the same time
- Running a mechanism awaits for its execution and passes the receivedBlock and peerId as parameters
- Logging is meaningful
- Activity status (isActive) is set to true immediately after calling the run method.
- Activity status (isActive) is set to false whether there is an error within the `run` method or the syncing mechanisms finishes successfully

limiaspasdaniel:
### What was the problem?
Refer to #4428 
### How did I solve it?
Refer to #4428 
### How to manually test it?
Run ` framework/test/jest/unit/specs/modules/chain/synchronizer/synchronizer.spec.js `
### Review checklist

- [ ] The PR resolves #4428


framework/test/jest/unit/specs/modules/chain/synchronizer/synchronizer.spec.js
```diff
@@ -174,86 +434,104 @@ describe('Synchronizer', () => {
...
+			await synchronizer.run(aReceivedBlock, aPeerId);
+
+			expect(syncMechanism1.isValidFor).toHaveBeenCalledTimes(1);
+			expect(syncMechanism2.isValidFor).toHaveBeenCalledTimes(1);
+			expect(syncMechanism1.run).toHaveBeenCalledWith(
+				await processorModule.deserialize(aReceivedBlock),
+				aPeerId,
 			);
+			expect(syncMechanism2.run).not.toHaveBeenCalled();
+			expect(loggerMock.info).nthCalledWith(2, 'Triggering: Object');
```
shuse2: I prefer to avoid checking logs, is there any other way to check this behavior?

framework/test/jest/unit/specs/modules/chain/synchronizer/synchronizer.spec.js
```diff
@@ -14,151 +14,411 @@
...
 			// Assert
-			expect(utils.restoreBlocksUponStartup).not.toHaveBeenCalled();
+			expect(loggerMock.error).toHaveBeenCalledWith(
+				{ err: error },
+				'Failed to restore blocks from temp table upon startup',
+			);
 		});
 	});
 
 	describe('register()', () => {
```
yatki: please refactor the synchronizer accordingly and remove these tests since the `register` method is redundant. If you need to move the validations to constructor, however, you can test constructor by providing invalid mechanisms to see if necessary checks are being performed.

framework/test/jest/unit/specs/modules/chain/synchronizer/synchronizer.spec.js
```diff
@@ -174,86 +434,104 @@ describe('Synchronizer', () => {
...
+				'A peer ID from the peer sending the block must be provided to the Synchronizer in order to run',
 			);
+			expect(synchronizer.active).toBeFalsy();
 		});
 
-		it('should reject with error if block verification failed', async () => {
-			const validationError = new Error('Block verifyError');
-			processorMock.validateDetached.mockRejectedValue(validationError);
+		it('should validate the block before sync', async () => {
+			jest.spyOn(processorModule, 'validateDetached');
```
yatki: ❤️

framework/test/jest/unit/specs/modules/chain/synchronizer/synchronizer.spec.js
ManuGowda: I am also not a fan of checking the logs.. but for this scenario, it should be fine.
maybe we can use setter and getter to get the mechanism? just a thought

framework/src/modules/chain/synchronizer/synchronizer.js
```diff
@@ -142,15 +159,18 @@ class Synchronizer {
...
 		// Loop through to find first mechanism which return true for isValidFor(receivedBlock)
+		let selectedMechanism;
 
-		// eslint-disable-next-line no-restricted-syntax
-		for await (const mechanism of this.mechanisms) {
+		for (const mechanism of this.mechanisms) {
 			if (await mechanism.isValidFor(receivedBlock)) {
-				return mechanism;
+				if (selectedMechanism) {
+					throw new Error('Two mechanisms cannot be valid at the same time'); // TODO: Add better message
```
yatki: you can remove the todo comment now :D

framework/src/modules/chain/chain.js
```diff
@@ -477,11 +477,9 @@ module.exports = class Chain {
 			blocksModule: this.blocks,
 			processorModule: this.processor,
 			storageModule: this.storage,
+			mechanisms: [blockSyncMechanism, fastChainSwitchMechanism],
```
yatki: ❤️

framework/test/jest/unit/specs/modules/chain/synchronizer/synchronizer.spec.js
yatki: I think logs are important especially for delegates, and if the info we are dumping to the logs is important we should have tests making sure we are still doing that.

framework/src/modules/chain/synchronizer/synchronizer.js
```diff
@@ -164,7 +164,7 @@ class Synchronizer {
 		for (const mechanism of this.mechanisms) {
 			if (await mechanism.isValidFor(receivedBlock)) {
 				if (selectedMechanism) {
-					throw new Error('Mechanisms are conflicting'); // TODO: Add better message
+					throw new Error('Two mechanisms cannot be valid at the same time');
```
ManuGowda: `Two` can be misleading, as one can register more than 2 aswell. it would be better if you could make it `multiple`? that way it's generic.

limiaspasdaniel: Ok will change it :)

PR #4437, opened 2019-10-28
Implement Unit tests for Fast Chain Switching - Closes #4376

### LINKED ISSUE #4376: Cover fast chain switch mechanism with unit/integration tests
### Description

After all of the fast chain switch mechanism implementation is done, it should follow test guideline and cover all the cases from the public interface `run`.

### Acceptance Criteria

- Coverage of the synchronization file should be more than 80%
- All the external module which uses IO should be stubbed.

limiaspasdaniel:
### What was the problem?
Refer to #4376 
### How did I solve it?
Refer to #4376 
### How to manually test it?
Execute `framework/test/jest/unit/specs/modules/chain/synchronizer/fast_chain_switching_mechanism/fast_chain_switching_mechanism.spec.js`
### Review checklist

- [ ] The PR resolves #4376 

### DIFF
--- a/framework/src/modules/chain/synchronizer/fast_chain_switching_mechanism.js
+++ b/framework/src/modules/chain/synchronizer/fast_chain_switching_mechanism.js
@@ -317,6 +317,7 @@ class FastChainSwitchingMechanism extends BaseSynchronizer {
 				const blockInstance = await this.processor.deserialize(block);
 				await this.processor.processValidated(blockInstance);
 			}
+			this.logger.info('Successfully switched chains. Node is now up to date');
 		} catch (err) {
 			this.logger.error({ err }, 'Error while processing blocks');
 			this.logger.debug(
@@ -334,8 +335,6 @@ class FastChainSwitchingMechanism extends BaseSynchronizer {
 			this.logger.debug('Cleaning blocks temp table');
 			await clearBlocksTempTable(this.storage);
 		}
-
-		this.logger.info('Successfully switched chains. Node is now up to date');
 	}
 
 	/**
--- a/framework/src/modules/chain/synchronizer/fast_chain_switching_mechanism.js
+++ b/framework/src/modules/chain/synchronizer/fast_chain_switching_mechanism.js
@@ -366,7 +366,7 @@ class FastChainSwitchingMechanism extends BaseSynchronizer {
 	async _requestLastCommonBlock(peerId) {
 		this.logger.debug({ peerId }, 'Requesting the last common block with peer');
 		const requestLimit = 10; // Maximum number of requests to be made to the remote peer
-		let numberOfRequests = 0; // Keeps track of the number of requests made to the remote peer
+		let numberOfRequests = 1; // Keeps track of the number of requests made to the remote peer
 
 		const heightList = this._computeLastTwoRoundsHeights();
 
--- a/framework/src/modules/chain/synchronizer/fast_chain_switching_mechanism.js
+++ b/framework/src/modules/chain/synchronizer/fast_chain_switching_mechanism.js
@@ -317,7 +317,13 @@ class FastChainSwitchingMechanism extends BaseSynchronizer {
 				const blockInstance = await this.processor.deserialize(block);
 				await this.processor.processValidated(blockInstance);
 			}
-			this.logger.info('Successfully switched chains. Node is now up to date');
+			this.logger.info(
+				{
+					currentHeight: this.blocks.lastBlock.height,
+					highestCommonBlockHeight: highestCommonBlock.height,
+				},
+				'Successfully switched chains. Node is now up to date',
+			);
 		} catch (err) {
 			this.logger.error({ err }, 'Error while processing blocks');
 			this.logger.debug(
--- a/framework/test/jest/unit/specs/modules/chain/synchronizer/fast_chain_switching_mechanism/fast_chain_switching_mechanism.spec.js
+++ b/framework/test/jest/unit/specs/modules/chain/synchronizer/fast_chain_switching_mechanism/fast_chain_switching_mechanism.spec.js
@@ -23,6 +23,7 @@ const {
 const {
 	Synchronizer,
 	FastChainSwitchingMechanism,
+	Errors,
 } = require('../../../../../../../../src/modules/chain/synchronizer');
 const { Slots } = require('../../../../../../../../src/modules/chain/dpos');
 const {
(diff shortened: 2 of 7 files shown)


framework/test/jest/unit/specs/modules/chain/synchronizer/fast_chain_switching_mechanism/fast_chain_switching_mechanism.spec.js
```diff
@@ -14,149 +14,866 @@
...
+			dpos: dposModuleMock,
+			activeDelegates: constants.ACTIVE_DELEGATES,
+		});
+	});
+
+	describe('async run()', () => {
+		const aPeer = '127.0.0.1:5000';
+		let aBlock;
+
+		function checkApplyPenaltyAndRestartIsCalled(
```
shuse2: I think in mocha test there were several function which calls expect, and it was hard to tell where it was actually throwing error. what do you think?
Also, defining in arrow function is fine IMO `const checkApplyPenaltyAndRestartIsCalled = () => {}`

framework/test/jest/unit/specs/modules/chain/synchronizer/fast_chain_switching_mechanism/fast_chain_switching_mechanism.spec.js
```diff
@@ -14,149 +14,866 @@
...
+			jest.spyOn(fastChainSwitchingMechanism, '_queryBlocks');
+			jest.spyOn(fastChainSwitchingMechanism, '_switchChain');
+			jest.spyOn(fastChainSwitchingMechanism, '_validateBlocks');
+			await blocksModule.init();
+			await bftModule.init();
+		});
 
-		let syncMechanism;
+		afterEach(() => {
+			jest.clearAllMocks();
```
shuse2: isn't this done by config? @yatki

framework/test/jest/unit/specs/modules/chain/synchronizer/fast_chain_switching_mechanism/fast_chain_switching_mechanism.spec.js
```diff
@@ -14,149 +14,866 @@
...
+			// Independently of the correct execution of the mechanisms, `active` property should be always
+			// set to false upon finishing the execution
+			expect(fastChainSwitchingMechanism.active).toBeFalsy();
+		});
+
+		it('should request the common block and try 10 times before giving up. If given up, it should apply a penalty to the peer and restart the mechanism', async () => {
```
shuse2: `if given up` seems like a different test. maybe structure like
```
describe('when fail to request common block')
  it('should give up after trying 10 times')
  it('should apply penalty after trying 10 times and restart the mechanism')
```

framework/test/jest/unit/specs/modules/chain/synchronizer/fast_chain_switching_mechanism/fast_chain_switching_mechanism.spec.js
```diff
@@ -14,149 +14,866 @@
...
+				"Peer didn't return a common block or its height is lower than the finalized height of the chain",
+			);
+			expect(fastChainSwitchingMechanism.active).toBeFalsy();
 		});
 
-		describe('#constructor', () => {
-			it('should create instance of FastChainSwitchingMechanism', async () => {
-				expect(syncMechanism).toBeInstanceOf(FastChainSwitchingMechanism);
+		describe('given that the highest common block is found', () => {
+			it('should apply penalty to the peer and restart syncing mechanisms if the height of the common block is smaller than the finalized height', async () => {
```
shuse2: also,
`describe('when the height of the common block is smaller than the finalized height')` might be better?

framework/test/jest/unit/specs/modules/chain/synchronizer/fast_chain_switching_mechanism/fast_chain_switching_mechanism.spec.js
limiaspasdaniel: Yeah, @yatki and I are discussing how to do this in a clearer way. The thing is those expects are repeatedly called on multiple test cases and I didn't want to clone the code multiple times.

framework/test/jest/unit/specs/modules/chain/synchronizer/fast_chain_switching_mechanism/fast_chain_switching_mechanism.spec.js
limiaspasdaniel: You mean remove `given that the highest common block is found` describe? I guess we can imply that the common block has been found with `it('should apply penalty to the peer and restart syncing mechanisms if the height of the common block is smaller than the finalized height'`

framework/src/modules/chain/synchronizer/fast_chain_switching_mechanism.js
```diff
@@ -317,6 +317,7 @@ class FastChainSwitchingMechanism extends BaseSynchronizer {
 				const blockInstance = await this.processor.deserialize(block);
 				await this.processor.processValidated(blockInstance);
 			}
+			this.logger.info('Successfully switched chains. Node is now up to date');
```
ManuGowda: It would be nice to know the details of the node up to date. something like `this.logger.info({ highestCommonBlockHeight, lastHighestBlockHeight }, 'Successfully switched chains. Node is now up to date');`

framework/test/jest/unit/specs/modules/chain/synchronizer/fast_chain_switching_mechanism/fast_chain_switching_mechanism.spec.js
ManuGowda: @limiaspasdaniel just understand your perspective here, are you trying to follow the `Given, When, Then` approach? if that's the case `Given would be your hooks, When would be described, Then can be it` what do you think? if this is clear we can change the sentence accordingly.

limiaspasdaniel: I wasn't strictly following that pattern, especially because we agreed that `it` should always start with `should` but we can come up with a better solution if it is no sufficiently clear right now.

framework/test/jest/unit/specs/modules/chain/synchronizer/fast_chain_switching_mechanism/fast_chain_switching_mechanism.spec.js
limiaspasdaniel: I did this to reset call count on mocks but I am still unable to do it. I will talk to @yatki to find a solution for this.

framework/src/modules/chain/synchronizer/fast_chain_switching_mechanism.js
limiaspasdaniel: Done :)

PR #4440, opened 2019-10-29
Update eager push to only send to 16 peers - Closes #3345

### LINKED ISSUE #3345: Add eager push mechanism to ensure that blocks and transactions propagate quickly
## SCOPE

This issue depends on how https://github.com/LiskHQ/lisk-sdk/issues/3344 is implemented, it's possible that it may not require additional work. The goal of this issue is to make sure that we do not lose propagation time after having implemented lazy push.

### Description

In our current system, we adopt an eager push mechanism: new information like blocks are directly pushed to a subset of the peers. We are going to use eager in the cases where it suited more according to [LIP0004](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0004.md). 

**Transaction Propagation:** Removal of eager push as it's not as time critical as blocks.

**Signature Propagation:** Unchanged (stick to eager push as already present in current protocol)

**Block Propagation:** New valid blocks are forwarded via a `postBlock` message to 16 randomly chosen connected peers while at least 8 `postBlock` messages are sent via outgoing connections

shuse2:
### What was the problem?
Eager push was defined to send 16 peers in which it should at least send 8 to outbound.
Although the selection of outbound was existing, it was sending to total 24 peers.

### How did I solve it?
Update the default to be 16 in both p2p and network

### How to manually test it?
Connect more than 16 nodes, and eager push the information (block)

### Review checklist

- [ ] The PR resolves #3345 


elements/lisk-p2p/test/unit/peer_pool.ts
```diff
@@ -53,7 +53,7 @@ describe('peerPool', () => {
 		peerSelectionForConnection: selectPeersForConnection,
 		peerSelectionForRequest: selectPeersForRequest,
 		peerSelectionForSend: selectPeersForSend,
-		sendPeerLimit: 24,
```
diego-G: Let's use `DEFAULT_SEND_PEER_LIMIT`

elements/lisk-p2p/test/unit/utils/select.ts
```diff
@@ -174,7 +174,7 @@ describe('peer selector', () => {
 			const selectedPeers = selectPeersForSend({
 				peers: peerList,
 				nodeInfo,
-				peerLimit: 24,
+				peerLimit: 16,
```
diego-G: Same as above.

elements/lisk-p2p/test/unit/utils/select.ts
```diff
@@ -194,7 +194,7 @@ describe('peer selector', () => {
 
 			expect(peerKindCounts.inbound)
 				.to.equal(peerKindCounts.outbound)
-				.to.equal(12);
+				.to.equal(8);
```
diego-G: `DEFAULT_SEND_PEER_LIMIT / 2`

framework/src/modules/network/defaults/config.js
```diff
@@ -159,7 +159,7 @@ const defaultConfig = {
 		wsEngine: 'ws',
 		maxOutboundConnections: 20,
 		maxInboundConnections: 100,
-		emitPeerLimit: 25,
+		emitPeerLimit: 16,
```
diego-G: If we are using the default's library, shall we have it here?

diego-G: https://github.com/LiskHQ/lisk-sdk/issues/4442

PR #4441, opened 2019-10-29
Add incentivizing Lisk-BFT Protocol Participation - Closes #4354

### LINKED ISSUE #4354: Add incentivizing Lisk-BFT Protocol Participation
### Description

BFT module should expose a method which verifies if the BFT protocol is being followed as per [LIP-0014](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#incentivizing-lisk-bft-protocol-participation)

This method should be used when validating a block in order to validate the block reward. i.e. the block reward should be 25% of the current reward if the protocol is not followed. Blocks which doesn't follow this rule should be rejected.

This method should be used when creating a new block in order to use the correct block reward. i.e. new block which doesn't follow the protocol should be have its reward reduced to 25% of the current reward.

### Motivation

Incentivize Lisk-BFT Protocol Participation

### Acceptance Criteria

- Unit test the new method with all possible scenarios

### Additional Information
--

lsilvs:
### What was the problem?
`Incentivizing Lisk-BFT Protocol Participation` section of the [LIP-0014](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#incentivizing-lisk-bft-protocol-participation) was not implemented

### How did I solve it?
I've created a method in BFT module called `isBFTProtocolCompliant` which returns whether or not the block follows the BFT protocol.
This method is used in the block processor validation pipeline in order to check if the reward was reduced to 25% if the protocol is not followed and it's also used in the block processor create pipeline in order to reduce the reward if the protocol is not followed.

### How to manually test it?
Forge a block with one of the conditions to make the block not follow the BFT protocol and check the reward value

### Review checklist

- [ ] The PR resolves #4354


framework/test/jest/unit/specs/modules/chain/bft/bft.spec.js
```diff
@@ -583,6 +583,129 @@ describe('bft', () => {
...
 
+		describe('#isBFTProtocolCompliant', () => {
+			let bft;
+			let blocks;
+
+			beforeEach(async () => {
+				bft = new BFT(bftParams);
+				storageMock.entities.Block.get.mockReturnValue([]);
+				await bft.init();
+				storageMock.entities.Block.get.mockClear();
```
yatki: I think, there is no need for this statement here.

framework/test/jest/unit/specs/modules/chain/bft/bft.spec.js
```diff
@@ -583,6 +583,129 @@ describe('bft', () => {
 			});
 		});
 
+		describe('#isBFTProtocolCompliant', () => {
+			let bft;
+			let blocks;
+
+			beforeEach(async () => {
+				bft = new BFT(bftParams);
```
yatki: An `// Arrange` comment at the beginning of the hook would be nice.
```suggestion
				// Arrange
				bft = new BFT(bftParams);
```

framework/test/jest/unit/specs/modules/chain/bft/bft.spec.js
```diff
@@ -583,6 +583,129 @@ describe('bft', () => {
...
+			it('should return FALSE when B.maxHeightPreviouslyForged is equal to B.height', async () => {
+				const block = {
+					height: 203,
+					maxHeightPreviouslyForged: 203,
+				};
+
+				expect(bft.isBFTProtocolCompliant(block)).toBe(false);
+			});
+
+			it('should return FALSE when B.maxHeightPreviouslyForged is greater than B.height', async () => {
```
yatki: I think in general the description of tests should reflect the behavior other than the condition. For instance, in this case, we can say `should return FALSE when a delegate does not constructively participate in the Lisk-BFT protocol`. I know this also sounds a bit shallow but I think you could make it even more clear. My point is we should write the behaviors as test descriptions. You may wanna check DPOS.apply for reference. It's also not perfect but would give you the idea. Please just take another look and see if there are ways to improve the descriptions.

framework/test/jest/unit/specs/modules/chain/bft/bft.spec.js
```diff
@@ -583,6 +583,129 @@ describe('bft', () => {
...
+			});
+
+			it('should return TRUE when B.maxHeightPreviouslyForged is equal to 0', async () => {
+				const block = {
+					height: 102,
+					generatorPublicKey: 'zxc',
+					maxHeightPreviouslyForged: 0,
+				};
+
+				expect(bft.isBFTProtocolCompliant(block)).toBe(true);
```
yatki:
```suggestion
				// Act & Assert
				expect(bft.isBFTProtocolCompliant(block)).toBe(true);
```

framework/test/jest/unit/specs/modules/chain/bft/bft.spec.js
```diff
@@ -583,6 +583,129 @@ describe('bft', () => {
...
+			});
+
+			it('should THROW if block is not provided', async () => {
+				expect(() => bft.isBFTProtocolCompliant()).toThrow(
+					'No block was provided to be verified',
+				);
+			});
+
+			it('should return TRUE when B.maxHeightPreviouslyForged is equal to 0', async () => {
+				const block = {
```
yatki:
```suggestion
				// Arrange
				const block = {
```

PR #4448, opened 2019-10-31
Upgrade to NodeJS 12 - Closes #4392

### LINKED ISSUE #4392: Upgrade to NodeJS 12
### Description

NodeJS 12 is going LTS soon. We should prepare for the upgrade.

### Motivation

NodeJS 12 introduce many improvements: 
- faster V8 version
- async stack traces!
- worker threads still experimental but no any longer behind a flag
- private fields
- faster http parser
- faster async/await
- import/export support

pablitovicente:
### What was the problem?

We were using version 10 of NodeJS but Node 12 is the latest LTS

### How did I solve it?

By upgrading to NodeJS 12

### How to manually test it?

build should be green

### Review checklist

- [ ] The PR resolves #4392


framework/package.json
```diff
@@ -20,8 +20,8 @@
 		"url": "https://github.com/LiskHQ/lisk-sdk/issues"
 	},
 	"engines": {
-		"node": ">=10.4 <=10",
-		"npm": ">=6"
+		"node": ">=12.13.0 <=12",
```
shuse2: let's make it consistent for all the packages
Also, packages in the elements need to be updated too

package.json
```diff
@@ -18,7 +18,7 @@
 		"url": "https://github.com/LiskHQ/lisk-sdk/issues"
 	},
 	"engines": {
-		"node": ">=8.10 <=10",
```
fchavant: This range is different from the others, is this on purpose?

pablitovicente: Fixed based on Shu's comment too.

framework/package.json
pablitovicente: Fixed

pablitovicente:
@shuse2 @fchavant I think I need to also update core and do you know what else? Maybe I should ping @Tschakki on how to update the docs?

fchavant:
@pablitovicente good points

Tschakki:
@pablitovicente You can simply open an issue for this in the lisk-docs repository, and put a reference to this PR inside the issue :)

PR #4456, opened 2019-11-01
Show BFT properties in api/blocks response - Closes #4449

### LINKED ISSUE #4449: Show BFT properties in api/blocks response
### Expected behavior
Add BFT properties to `/api/blocks` response so that end users can validate the BFT properties for validation
### Actual behavior
Currently, we do not have a way to know the BFT properties in blocks to identify the behaviors.
### Steps to reproduce
http://lander-seed-01.liskdev.net:4000/api/blocks
3.0.0

lsilvs:
### What was the problem?
HTTP API endpoint `api/blocks` not returning BFT properties

### How did I solve it?
By adding the properties `maxHeightPrevoted` and `maxHeightPreviouslyForged` to the blocks endpoint.
I also did some code refactoring to simplify the code and remove unnecessary validations already done by swagger.

### How to manually test it?
Start the application
Access `http://localhost:4000/api/blocks`

### Review checklist

- [ ] The PR resolves #4449


framework/src/modules/http_api/controllers/blocks.js
```diff
@@ -40,52 +39,31 @@ function parseBlockFromDatabase(raw) {
 		version: parseInt(raw.version, 10),
 		timestamp: parseInt(raw.timestamp, 10),
 		height: parseInt(raw.height, 10),
-		previousBlock: raw.previousBlockId,
+		previousBlockId: raw.previousBlockId || '',
 		numberOfTransactions: parseInt(raw.numberOfTransactions, 10),
-		totalAmount: new BigNum(raw.totalAmount).toFixed(),
-		totalFee: new BigNum(raw.totalFee).toFixed(),
-		reward: new BigNum(raw.reward).toFixed(),
+		totalAmount: raw.totalAmount,
```
shuse2: I think, most of the stuff, if it's exactly the same you could do below, so that we don't need to update if there is no parsing required
```
const block = {
  ...raw,
  previousBlockId: raw.previousBlockId || '',
}
```

PR #4462, opened 2019-11-04
transaction.id should not be used before assignment - Closes #4461

### LINKED ISSUE #4461: transaction.id should not be used before assignment
### Expected behavior
After id has been not required property by #4038, transaction id should not be used before the assignment.
All the usage should be after the assignment.
Also, calculation of the `id` must be done after schema validation, and during the schema validation, `id` should never be used.
Specifically at this point, it should not be used, https://github.com/LiskHQ/lisk-sdk/blob/development/elements/lisk-transactions/src/base_transaction.ts#L235-L239

### Actual behavior
In validation it is used by `toJSON` right before execution.

### Steps to reproduce
Call validate function with minimal properties.

3.0-

pablitovicente:
### What was the problem?

Transacion.id was throwing error even though not required before signing

### How did I solve it?

By making the getter return empty string if the id doesn't exist

### How to manually test it?

Build should be green

### Review checklist

- [x] The PR resolves #4461

### DIFF
--- a/elements/lisk-transactions/src/base_transaction.ts
+++ b/elements/lisk-transactions/src/base_transaction.ts
@@ -157,7 +157,7 @@ export abstract class BaseTransaction {
 	}
 
 	public get id(): string {
-		return this._id || '';
+		return this._id || 'incalculable-id';
 	}
 
 	public get senderId(): string {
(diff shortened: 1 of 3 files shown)


elements/lisk-transactions/src/base_transaction.ts
```diff
@@ -157,11 +157,7 @@ export abstract class BaseTransaction {
 	}
 
 	public get id(): string {
-		if (!this._id) {
-			throw new Error('id is required to be set before use');
-		}
-
-		return this._id;
+		return this._id || 'incalculable-id';
```
mitsujutsu: why not empty string?

pablitovicente: to be more obvious to the user that something is wrong with its JSON

PR #4464, opened 2019-11-04
API node/status/forging filter doesn't work as expected - Closes #4091

### LINKED ISSUE #4091: API node/status/forging filter doesn't work as expected
### Expected behavior
Endpoint `node/status/forging` should filter `forging` delegates or non-forging delegates
### Actual behavior
Endpoint `node/status/forging` doesn't filter `forging` delegates or non-forging delegates
### Steps to reproduce
- Start node on localhost as `devnet`
- Disable few delegates from forging
- http://localhost:4000/api/node/status/forging?forging=false
- The API endpoint doesn't filter according to the falg

ManuGowda:
### What was the problem?
API `node/status/forging` was not filtering according to forging status.
### How did I solve it?
Filtering the response based on forging status parameter.
### How to manually test it?
- Start the node, enable forging all nodes
- http://localhost:4000/api/node/status/forging?forging=false (should get zero results)
- Disable a few delegates
- http://localhost:4000/api/node/status/forging?forging=false (should get all disabled delegates)

### Review checklist

- [x] The PR resolves #4091 

### DIFF
--- a/framework/src/modules/http_api/schema/swagger.yml
+++ b/framework/src/modules/http_api/schema/swagger.yml
@@ -1190,7 +1190,7 @@ parameters:
     in: query
     description: Forging status to filter
     type: boolean
-    format: bool
+
 definitions:
   Meta:
     type: object
--- a/framework/src/modules/http_api/schema/swagger.yml
+++ b/framework/src/modules/http_api/schema/swagger.yml
@@ -384,6 +384,7 @@ paths:
         - application/json
       parameters:
         - $ref: '#/parameters/publicKey'
+        - $ref: '#/parameters/forging'
       responses:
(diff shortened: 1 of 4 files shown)


framework/src/modules/http_api/schema/swagger.yml
```diff
@@ -1184,7 +1185,12 @@ parameters:
     description: Ending unix timestamp
     type: integer
     minimum: 1
-
+  forging:
+    name: forging
+    in: query
+    description: Forging status to filter
+    type: boolean
+    format: bool
```
shuse2: `format` bool is not there by default. I think `type is enough` in this case

PR #4466, opened 2019-11-05
Introduce advertiseAddress flag for peers - Closes #3661

### LINKED ISSUE #3661: Introduce advertiseAddress flag for peers
### Expected behavior

 An additional flag `advertiseAddress` should be introduced in the system header of a Peer. 
Default setting should be true meaning the IP address of the peer will be advertised in the whole network. 
If this flag is false, then the node has to remember to not to advertise this address.
Remember that the API shouldn't show the peer even if the node is connected.

If `advertiseAddress` flag is set to `false` then:
- `getPeers` should never return the peer with `advertiseAddress` flag set to `false`
- `getNodeInfo` should return `advertiseAddress` as part of node header info 
- Filter `advertiseAddress` peers in `getConnectedPeers`, `getDisconnectedPeers`
### Actual behavior

Every peer is advertised.

`2.1`
Note: Should come from the framework config file.

ManuGowda:
### What was the problem?
Every peer is advertised to rest of the network. If a peer intended to not `advertised` across the network it wasn't possible.
### How did I solve it?
Introducing `advertiseAddress` flag, which is set to `true` by default. if a delegate/node operator wants to not advertise they can update the flag to `false`
### How to manually test it?
- Start three nodes (node1, node2, node3)
- Update `advertiseAddress` flag to `false` on node1, use node2 as the seed node
- Verify if the node3 doesn't send the node1 peer info
### Review checklist

- [ ] The PR resolves #3661 


elements/lisk-p2p/src/p2p.ts
```diff
@@ -182,6 +182,14 @@ const createPeerPoolConfig = (
 	peerLists,
 });
 
+const filteradvertiseAddressPeer = (peer: P2PPeerInfo) => {
```
ishantiw: `filterAdvertiseAddressPeer`

elements/lisk-p2p/test/integration/advertise_address.ts
```diff
@@ -0,0 +1,95 @@
...
+	maxOutboundConnections: 20,
+	maxInboundConnections: 100,
+	nodeInfo: {
+		...nodeInfoConstants,
+		wsPort,
+		advertiseAddress,
+	},
+});
+
+describe.only('Advertise Address', () => {
```
diego-G: Please, remove this.

elements/lisk-p2p/src/p2p.ts
```diff
@@ -182,6 +182,14 @@ const createPeerPoolConfig = (
...
+	if (peer.internalState && !peer.internalState.advertiseAddress) {
```
ishantiw: Can we change this condition to `peer.internalState.advertiseAddress === false` because in runtime if someone doesn't send `advertiseAddress` it could be `undefined` and still return false, i.e., it will not advertise the address who doesn't explicitly mention it to be advertise.

elements/lisk-p2p/test/integration/advertise_address.ts
```diff
@@ -0,0 +1,95 @@
...
+import { P2P } from '../../src/index';
+import {
+	createNetwork,
+	destroyNetwork,
+	POPULATOR_INTERVAL,
+	nodeInfoConstants,
+} from '../utils/network_setup';
+import { wait } from '../utils/helpers';
+
+const p2pConfig = (wsPort: number, advertiseAddress: boolean = true) => ({
```
diego-G: Move this inside the first describe.

elements/lisk-p2p/src/p2p.ts
```diff
@@ -559,12 +567,15 @@ export class P2P extends EventEmitter {
 	// Make sure you always share shared peer state to a user
```
ishantiw: update the comment to something like `Make sure you always share shared peer state to a user and remove private peers`

elements/lisk-p2p/src/p2p.ts
```diff
@@ -593,7 +604,7 @@ export class P2P extends EventEmitter {
 		});
 
 		// Only share the shared state to the user
```
ishantiw: update the comment as above

elements/lisk-p2p/test/integration/advertise_address.ts
```diff
@@ -0,0 +1,95 @@
...
+				.getConnectedPeers()
+				.filter(p => p.wsPort === p2pNode.nodeInfo.wsPort);
+			const disConnectedPeers = p2p
+				.getDisconnectedPeers()
+				.filter(p => p.wsPort === p2pNode.nodeInfo.wsPort);
+			expect(connectedPeers).to.be.empty;
+			expect(disConnectedPeers).to.be.empty;
+		}
+
+		await p2pNode.stop();
```
diego-G: You can move this to the `afterEach`.

elements/lisk-p2p/test/integration/advertise_address.ts
```diff
@@ -0,0 +1,95 @@
+/*
```
diego-G: The test fits better into the functional test suite.

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -343,6 +343,9 @@ export class PeerPool extends EventEmitter {
 		].map(peer => ({
 			...peer.peerInfo,
 			internalState: {
+				advertiseAddress: peer.peerInfo.internalState
```
diego-G: Please, cover this with unit tests.

elements/lisk-p2p/src/utils/validate.ts
```diff
@@ -174,6 +175,10 @@ export const validatePeerInfoSchema = (rawPeerInfo: unknown): P2PPeerInfo => {
 			height: height && isNumeric(height.toString()) ? +height : 0,
 			...restOfProtocolPeer,
 		},
+		internalState: {
```
diego-G: Same as above.

elements/lisk-p2p/test/integration/advertise_address.ts
```diff
@@ -0,0 +1,95 @@
...
+		await wait(100);
+
+		for (let p2p of p2pNodeList) {
+			const connectedPeers = p2p
+				.getConnectedPeers()
+				.filter(p => p.wsPort === p2pNode.nodeInfo.wsPort);
+			const disConnectedPeers = p2p
+				.getDisconnectedPeers()
+				.filter(p => p.wsPort === p2pNode.nodeInfo.wsPort);
+			expect(connectedPeers).to.be.empty;
```
diego-G: Empty or better undefined?

elements/lisk-p2p/src/utils/validate.ts
```diff
@@ -174,6 +175,10 @@ export const validatePeerInfoSchema = (rawPeerInfo: unknown): P2PPeerInfo => {
...
+			advertiseAddress:
+				advertiseAddress === false ? (advertiseAddress as boolean) : true,
```
ishantiw: forgot to mention that this `peerInfo` might have `advertiseAddress` as a string since its coming from the network
we can check `advertiseAddress == false ? false: true`

elements/lisk-p2p/test/integration/advertise_address.ts
ishantiw: remove only

elements/lisk-p2p/src/p2p_types.ts
```diff
@@ -77,6 +78,7 @@ export interface P2PNodeInfo extends P2PSharedState {
 	readonly os: string;
 	readonly nethash: string;
 	readonly wsPort: number;
+	readonly advertiseAddress?: boolean;
```
diego-G: Make this non-optional.

elements/lisk-p2p/src/p2p_types.ts
```diff
@@ -132,6 +134,7 @@ export interface P2PConfig {
 	readonly maxPeerDiscoveryResponseLength?: number;
 	readonly maxPeerInfoSize?: number;
 	readonly secret?: number;
+	readonly advertiseAddress?: boolean;
```
diego-G: Same as above.

elements/lisk-p2p/test/integration/advertise_address.ts
```diff
@@ -0,0 +1,95 @@
...
+	},
+});
+
+describe('Advertise Address', () => {
+	let p2pNodeList: ReadonlyArray<P2P> = [];
+
+	beforeEach(async () => {
+		const customConfig = () => ({
+			advertiseAddress: true,
+			populatorInterval: POPULATOR_INTERVAL,
```
ishantiw: we can remove as its by default

ishantiw: also, we can use a smaller network

elements/lisk-p2p/test/integration/advertise_address.ts
ManuGowda: since I am creating a different peer with a different port for each test, it should be ok stop individually.

elements/lisk-p2p/test/integration/advertise_address.ts
ManuGowda: I think its a configuration setting from the framework, so its an integration and also all other configuration tests are in integration. so it should be fine.

elements/lisk-p2p/src/p2p_types.ts
ManuGowda: as discussed with Ishan, by default the advertiseAddress is handled as true. I understand your intention behind avoid the extra check, but as we discussed, for p2p it's optional and its a configuration from the framework, keeping it optional would be fine in my opinion. Same for below.

diego-G: The final discussion was that it's ok to make it optional when passing the conf value but it's better to make it mandatory internally. Please, make it non-optional in `P2PNodeInfo` iface.

elements/lisk-p2p/src/p2p.ts
```diff
@@ -184,6 +184,15 @@ const createPeerPoolConfig = (
 	peerLists,
 });
 
+const filterAdvertiseAddressPeer = (peer: P2PPeerInfo) => {
```
diego-G: Why don't you make this a private P2P method? I would even aggregate the `filter` action inside the method. Otherwise it feels redundant `.filter(filterAdvertiseAddressPeer)`.

elements/lisk-p2p/test/integration/discovery.ts
```diff
@@ -182,6 +182,7 @@ describe('Peer discovery', () => {
 				os: platform(),
 				height: 10000,
 				nonce: `404`,
+				advertiseAddress: true,
```
diego-G: You don't need this since it's the default value and optional in the config interface.

diego-G: Same happens in other places.

elements/lisk-p2p/src/p2p_types.ts
diego-G: In general, we should avoid optional fields overall when they are booleans. It doesn't make sense to have an optional boolean...

elements/lisk-p2p/src/utils/validate.ts
```diff
@@ -174,6 +175,10 @@ export const validatePeerInfoSchema = (rawPeerInfo: unknown): P2PPeerInfo => {
 			height: height && isNumeric(height.toString()) ? +height : 0,
 			...restOfProtocolPeer,
 		},
+		internalState: {
+			// tslint:disable-next-line no-boolean-literal-compare
+			advertiseAddress: (advertiseAddress as boolean) === false ? false : true,
```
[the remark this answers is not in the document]
ManuGowda: I am not quite sure what difference it will make with `==` in comparison to `===`.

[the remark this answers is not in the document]
ManuGowda: In this case, the value is unknown, so it can be `undefined, string or boolean` so now I am covering all the scenarios.

elements/lisk-p2p/src/p2p.ts
```diff
@@ -739,14 +750,23 @@ export class P2P extends EventEmitter {
...
+				} = queryObject;
 				const incomingPeerInfo: P2PPeerInfo = {
 					sharedState: {
 						...restOfQueryObject,
 						...queryOptions,
 						height: queryObject.height ? +queryObject.height : 0,
 						version: queryObject.version,
 					},
+					internalState: {
+						advertiseAddress: advertiseAddress !== 'false',
```
shuse2:
```suggestion
						advertiseAddress: Boolean(advertiseAddress),
```

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -343,11 +343,14 @@ export class PeerPool extends EventEmitter {
 		].map(peer => ({
 			...peer.peerInfo,
 			internalState: {
-				...peer.peerInfo.internalState,
+				advertiseAddress: peer.peerInfo.internalState
```
shuse2: why the order is switched here?

elements/lisk-p2p/src/utils/validate.ts
```diff
@@ -174,6 +175,12 @@ export const validatePeerInfoSchema = (rawPeerInfo: unknown): P2PPeerInfo => {
 			height: height && isNumeric(height.toString()) ? +height : 0,
 			...restOfProtocolPeer,
 		},
+		internalState: {
+			advertiseAddress:
+				advertiseAddress !== undefined
+					? !(advertiseAddress === 'false' || advertiseAddress === false)
```
shuse2: `Boolean(advertiseAddress)` is fine imo

PR #4467, opened 2019-11-05
Update transaction processing to consistently use state store - Closes #4137, #3987

### LINKED ISSUE #4137: Update transaction module to consistently use state store
### Description

Currently transaction module is creating the state store while processing.
However, state store should be given by the pipeline of processor, and it should use the instance. For example, https://github.com/LiskHQ/lisk-sdk/blob/feature/introduce_bft_consensus/framework/src/modules/chain/transactions/transactions_handlers.js#L149.

In this transaction processing, instead of taking `storage` as parameter, it should take `stateStore` as parameter.
Also, it will affect https://github.com/LiskHQ/lisk-sdk/blob/feature/introduce_bft_consensus/framework/src/modules/chain/transaction_pool/transaction_pool.js#L101-L102

### Motivation

state store should be shared across the process of a block to be consistent.
Since verify and apply is using different state store, it is inefficient and hard to test.

### Acceptance Criteria

- Missing applyTransaction test should be covered
- transaction handler should accept state store as input instead of storage
- All the transaction handler test needs to be updated

shuse2:
### What was the problem?
- For transaction processing, it was using different state store every time, and the caching did not work well.

### How did I solve it?
- For transaction processing, in one atomic step, it now uses the same state store instance.
- Further optimization on the cacheing will be done in future.
- Add 

### How to manually test it?
- All the test should pass

### Review checklist

- [ ] The PR resolves #4137 #3987 


framework/src/modules/chain/blocks/chain.js
```diff
@@ -19,24 +19,6 @@ const transactionsModule = require('../transactions');
...
-			storage.entities.Transaction.create(
-				parsedBlock.transactions,
-				{},
-				saveBlockBatchTx,
-			),
-		);
-	}
-
-	return saveBlockBatchTx.batch(promises);
-};
```
lsilvs: ❤️

framework/src/modules/chain/transaction_pool/transaction_pool.js
```diff
@@ -89,22 +90,28 @@ class TransactionPool extends EventEmitter {
...
+			const stateStore = new StateStore(this.storage);
+			return transactionsModule.composeTransactionSteps(
+				transactionsModule.checkAllowedTransactions(() => {
+					const { version, height, timestamp } = this.blocks.lastBlock;
+					return {
+						blockVersion: version,
+						blockHeight: height,
+						blockTimestamp: timestamp,
+					};
+				}), // TODO: probably wrong
```
lsilvs: I know this comment was not introduced in this PR but this is a very weird message 😄 

This should be investigated or removed IMO

PR #4469, opened 2019-11-05
Move transactions into block - Closes #4138

### LINKED ISSUE #4138: Move transaction module in framework to inside of blocks module
### Description
Currently, transaction module is separated from blocks module, but it should be inside of blocks module like
```
/blocks
  index.js
  /transactions
    transaction_handler.js
    index.js
```

Additionally, `interface_adapter` is only used for transaction, so it also should be in the blocks module

Also, relevant functions which are used from upper layer need to be exported from the blocks module.

### Motivation

Because blocks module is responsible for saving and applying the transaction, transaction module is sub-domain of the blocks module

### Acceptance Criteria

- All tests related to transactions module needs to be moved

shuse2:
### What was the problem?
Concept related to block, transaction and state store was closely related, but scatter across folders

### How did I solve it?
- Move transactions, state-store and interface adapter into blocks module, and exposed minimal function to use them
- Add and update tests to cover them

### How to manually test it?
- Everything should work as before

### Review checklist

- [ ] The PR resolves #4138 


framework/test/jest/unit/specs/modules/chain/blocks/process.spec.js
```diff
@@ -0,0 +1,1014 @@
...
+						blockBytes,
+						defaultReward,
+					);
+				} catch (errors) {
+					expect(errors).toHaveLength(1);
+				}
+			});
+		});
+
+		describe('when payload exceeds maximum', () => {
```
ManuGowda: when block exceeds maximum payload length?

framework/test/jest/unit/specs/modules/chain/blocks/process.spec.js
```diff
@@ -0,0 +1,1014 @@
...
+						blockBytes,
+						defaultReward,
+					);
+				} catch (error) {
+					expect(error.message).toContain('Payload length is too long');
+				}
+			});
+		});
+
+		describe('when payload exceeds maximum', () => {
```
ManuGowda: when exceeds maximum transactions per block limit?

framework/test/jest/unit/specs/modules/chain/blocks/process.spec.js
```diff
@@ -0,0 +1,1014 @@
...
+				// Act & Assert
+				await expect(
+					blocksInstance.verifyInMemory(block, genesisBlock),
+				).rejects.toThrow('Invalid block timestamp');
+			});
+
+			it('should throw when block timestamp is earlier than lastBlock timestamp', async () => {
+				// Arrange
+				const futureTimestamp = slots.getSlotTime(slots.getNextSlot());
+				block = newBlock({ timestamp: futureTimestamp });
```
ManuGowda: is this timestamp earlier? or future timestamp?

framework/src/modules/chain/forger/forger.js
```diff
@@ -20,7 +20,7 @@ const {
 	parseEncryptedPassphrase,
 	getAddressFromPublicKey,
 } = require('@liskhq/lisk-cryptography');
-const { sortTransactions } = require('./transactions');
+const { sortTransactions } = require('./sort');
```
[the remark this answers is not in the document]
shuse2: there could be other sort function here, so i think file name is fine for now

framework/src/modules/chain/blocks/blocks.js
```diff
@@ -416,7 +416,7 @@ class Blocks extends EventEmitter {
 			);
 			return block;
 		} catch (e) {
-			const errMessage = 'Failed to access storage layer';
+			const errMessage = 'Failed to fetch the heigest common block';
```
limiaspasdaniel: `highest`

PR #4472, opened 2019-11-05
Remove nonce from framework chain submodule - Closes #3702

### LINKED ISSUE #3702: Do not include nonce when broadcasting
### Expected behavior
The node broadcast only the necessary data. Remember that we won't have to validate the nonce in the schema either (transactions, signatures and blocks).
https://github.com/LiskHQ/lisk-sdk/blob/008d6d20292aa0e54a82a4c44999e649fa786be0/framework/src/modules/chain/schema/definitions.js#L118

The nonce property should be removed from all of the RPC endpoints response and validation.

### Actual behavior
Nonce is broadcast together with the objects but never used.
https://github.com/LiskHQ/lisk-sdk/blob/008d6d20292aa0e54a82a4c44999e649fa786be0/framework/src/modules/chain/logic/broadcaster.js#L109

`1.0.0+`

diego-G:
### Review checklist

- [x] The PR resolves #3702


elements/lisk-p2p/src/constants.ts
```diff
@@ -24,6 +24,7 @@ export const DEFAULT_WS_MAX_MESSAGE_RATE = 100;
 export const DEFAULT_WS_MAX_MESSAGE_RATE_PENALTY = 10;
 export const DEFAULT_RATE_CALCULATION_INTERVAL = 1000;
 export const DEFAULT_WS_MAX_PAYLOAD = 3048576; // Size in bytes
+export const DEFAULT_NONCE_LENGTH_BYTES = 32;
```
shuse2: I think it's my bad, but this actually should be `8` (it was opposite)
32 bytes will be 64 characters, and we want 16 characters

PR #4473, opened 2019-11-05
Add delegateMinHeightActive needs to be included in calculation of prevote and precommit - Closes #4349

### LINKED ISSUE #4349: Add delegateMinHeightActive needs to be included in calculation of prevote and precommit
### Description
In the BFT module when we compute prevotes and precommits, an important variable [`delegateMinHeightActive`](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#computing-prevotes-and-precommits) is currently hardcoded to `0`. Practically this is not breaking the code execution but of course, it leads to **miscalculated** precommits and prevotes.

As LIP states, `delegateMinHeightActive` is for a currently **active** delegate, the height of **first block of the round** when that delegate has been **continuously** __active__.`

So, a correct `delegateMinHeightActive` information must be provided to the BFT module.

For that, Dpos module should update the account states at the beginning of the rounds. Unfortunately, as LIP states `undo` requires to calculate this information from scratch. 

### Motivation
Delegates should **not** be able to precommit or prevote for blocks when they were not active in that round.

### Acceptance Criteria
-  `delegateMinHeightActive` must be updated once a delegate becomes active and should not be updated as long as that delegate remains active.
- Unit test for dpos module is updated for delegate becoming active
- Unit test for dpos module is updated for delegate going standby and coming back active
- Add unit test for bft module where delegateMinHeightActive is update according to the account state
- Add unit tests in case of  reverting a block.

### Additional Information
- https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#computing-prevotes-and-precommits

**How to choose the right `delegateMinHeightActive` value.**

Let's assume block A which has a height of 820 was received. That means block A is part of round 9. And if delegate was not active in the previous round (8), `delegateMinHeightActive` for that delegate will be updated to `809` (The first block height of the round 9). It doesn't matter which list was used or how the list was stored in the database. Also [`delegateListRoundOffset`](https://github.com/LiskHQ/lisk-sdk/pull/4397/files) variable is not important when calculating `delegateMinHeightActive`. delegateMinHeightActive is always calculated from the received block which gives us the active round. In this example active round is round 9.

yatki:
### What was the problem?
[`delegateMinHeightActive`](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#computing-prevotes-and-precommits) parameter was not being evaluated in the BFT module.
### How did I solve it?
I implemented getDelegateMinHeightActive method in BFT module in order to calculate that parameter.
### Review checklist

- [x] The PR resolves #4349 


framework/src/modules/chain/block_processor_v2.js
```diff
@@ -232,7 +232,18 @@ class BlockProcessorV2 extends BaseBlockProcessor {
...
-			({ block, tx }) => this.bftModule.addNewBlock(block, tx),
+			({ block, tx }) =>
+				this.dposModule.getDelegateMinHeightActive(block, { tx }),
+			({ block, tx }, delegateMinHeightActive) => {
+				const blockHeader = {
+					...block,
+					// This parameter injected to block object to avoid big refactoring
+					// for the moment. `activeSinceRound` will be removed from the block
+					// object with https://github.com/LiskHQ/lisk-sdk/issues/4413
+					activeSinceRound: this.slots.calcRound(delegateMinHeightActive),
```
lsilvs: In BFT module, this property is used to calculate the `delegateMinHeightActive` so I would suggest sending it as `delegateMinHeightActive` directly and skip the conversion to round.

framework/src/modules/chain/bft/bft.js
```diff
@@ -34,7 +34,10 @@ const extractBFTBlockHeaderFromBlock = block => ({
 	maxHeightPreviouslyForged: block.maxHeightPreviouslyForged,
 	prevotedConfirmedUptoHeight: block.prevotedConfirmedUptoHeight,
 	delegatePublicKey: block.generatorPublicKey,
-	activeSinceRound: 0, // TODO: Link the new DPOS with BFT here
+	// This parameter injected to block object to avoid big refactoring
+	// for the moment. `activeSinceRound` will be removed from the block
+	// object with https://github.com/LiskHQ/lisk-sdk/issues/4413
+	activeSinceRound: block.activeSinceRound || 0,
```
shuse2: Can we change this to activeSinceHeight? In the BFT, it should not care about the round

framework/src/modules/chain/bft/bft.js
```diff
@@ -49,12 +52,13 @@ class BFT extends EventEmitter {
...
 	 * @param {integer} startingHeight - The height at which BFT finalization manager initialize
 	 */
-	constructor({ storage, logger, activeDelegates, startingHeight }) {
+	constructor({ storage, logger, slots, activeDelegates, startingHeight }) {
 		super();
 		this.finalityManager = null;
 
 		this.logger = logger;
 		this.storage = storage;
+		this.slots = slots;
```
shuse2: I think once we change to the activeSinceHeight, we don't need `slot` here

framework/src/modules/chain/dpos/dpos.js
```diff
@@ -77,17 +79,48 @@ module.exports = class Dpos {
...
 		);
 	}
 
+	async getLatestDelegateLists({
+		tx,
+		delegateListRoundOffset = this.delegateListRoundOffset,
+	}) {
+		// TODO: Discuss reintroducing a caching mechanism to avoid fetching
+		// active delegate lists multiple times.
+		const previousRounds = this.storage.entities.RoundDelegates.get(
```
shuse2: it needs to await here

framework/src/modules/chain/dpos/dpos.js
```diff
@@ -31,6 +31,8 @@ module.exports = class Dpos {
 		this.events = new EventEmitter();
 		this.delegateListRoundOffset = delegateListRoundOffset;
 		this.finalizedBlockRound = 0;
+		// @todo consider making this a constant and reuse it in BFT module.
+		this.bftRoundOffset = 3;
```
shuse2: would it possible to make the variable name not mention `bft`?

framework/src/modules/chain/bft/bft.js
shuse2: Regards to this, we have decided to refactor on another issue.
@yatki Please open the issue and reference this PR

framework/src/modules/chain/bft/bft.js
shuse2: same as above

framework/src/modules/chain/dpos/dpos.js
```diff
@@ -77,17 +79,99 @@ module.exports = class Dpos {
...
 		);
 	}
 
+	async getActiveDelegateHeights(
+		numberOfRounds = 1,
+		{ tx, delegateListRoundOffset = this.delegateListRoundOffset },
+	) {
+		// TODO: Discuss reintroducing a caching mechanism to avoid fetching
+		// active delegate lists multiple times.
+		let delegateLists = await this.storage.entities.RoundDelegates.get(
```
shuse2: prefer no `let`.
Since type is different after map, it probably should have different name

framework/src/modules/chain/dpos/dpos.js
```diff
@@ -77,17 +79,99 @@ module.exports = class Dpos {
...
+			list.round += delegateListRoundOffset;
+			return list;
+		});
+
+		const delegates = {};
+		for (let i = 0; i < numberOfRounds; i += 1) {
+			const list = delegateLists[i];
+			const activeRound = list.round;
+
+			list.delegatePublicKeys.forEach(publicKey => {
```
shuse2: seems like purpose of this function is more `reduce` than `forEach`

framework/src/modules/chain/dpos/dpos.js
```diff
@@ -77,17 +79,99 @@ module.exports = class Dpos {
...
+		 */
+		delegateLists = delegateLists.map(list => {
+			// eslint-disable-next-line no-param-reassign
+			list.round += delegateListRoundOffset;
+			return list;
+		});
+
+		const delegates = {};
+		for (let i = 0; i < numberOfRounds; i += 1) {
+			const list = delegateLists[i];
```
shuse2:
```suggestion
			const { round, delegatePublicKeys } = delegateLists[i];
```

framework/src/modules/chain/dpos/dpos.js
```diff
@@ -77,17 +79,99 @@ module.exports = class Dpos {
...
+				limit: numberOfRounds + this.delegateActiveRoundLimit,
+			},
+			tx,
+		);
+
+		/**
+		 * Reverting the RoundOffset value, so that
+		 * round no can actually reflect the real round number.
+		 */
+		delegateLists = delegateLists.map(list => {
```
shuse2:
```suggestion
		delegateLists = delegateLists.map(list => ({ ...list, round: list.round + delegateListRoundOffset }));
```

framework/src/modules/chain/dpos/dpos.js
```diff
@@ -77,17 +79,99 @@ module.exports = class Dpos {
...
-			this.finalizedBlockRound - this.delegateListRoundOffset;
+			this.finalizedBlockRound -
+			this.delegateListRoundOffset -
+			this.delegateActiveRoundLimit;
 		await this.delegatesList.deleteDelegateListUntilRound(
 			disposableDelegateList,
 		);
 	}
 
+	async getActiveDelegateHeights(
```
shuse2: test for this function is missing

framework/src/modules/chain/block_processor_v2.js
```diff
@@ -168,7 +168,14 @@ class BlockProcessorV2 extends BaseBlockProcessor {
 		this.constants = constants;
 		this.exceptions = exceptions;
 
-		this.init.pipe([() => this.bftModule.init()]);
+		this.init.pipe([
+			async () => {
+				const delegateMinHeightActiveList = await this.dposModule.getActiveDelegateHeights(
+					2,
```
lsilvs: Prefer using a variable name here for readability

framework/src/modules/chain/block_processor_v2.js
```diff
@@ -231,7 +238,30 @@ class BlockProcessorV2 extends BaseBlockProcessor {
 				this.blocksModule.verify(block, stateStore, { skipExistingCheck }),
 			({ block, stateStore }) => this.blocksModule.apply(block, stateStore),
 			({ block, tx }) => this.dposModule.apply(block, { tx }),
-			({ block, tx }) => this.bftModule.addNewBlock(block, tx),
+			async ({ block, tx }) => {
+				const delegateMinHeightActiveList = await this.dposModule.getActiveDelegateHeights(
+					1,
```
lsilvs: Same as above

framework/src/modules/chain/block_processor_v2.js
```diff
@@ -243,7 +273,13 @@ class BlockProcessorV2 extends BaseBlockProcessor {
 		this.undo.pipe([
 			({ block, stateStore }) => this.blocksModule.undo(block, stateStore),
 			({ block, tx }) => this.dposModule.undo(block, { tx }),
-			({ block }) => this.bftModule.deleteBlocks([block]),
+			async ({ block, tx }) => {
+				const delegateMinHeightActiveList = await this.dposModule.getActiveDelegateHeights(
+					2,
```
lsilvs: Same as above

framework/src/modules/chain/dpos/dpos.js
yatki: That's true but since it is inside a for loop I didn't want to reassign `delegates` variable and didn't use `reduce`.

framework/src/modules/chain/bft/bft.js
```diff
@@ -301,7 +314,30 @@ class BFT extends EventEmitter {
...
 			if (row.height !== 1 && row.version !== 2) return;
 
-			this.finalityManager.addBlockHeader(extractBFTBlockHeaderFromBlock(row));
+			const delegate = delegateMinHeightActiveList[row.generatorPublicKey];
+
+			if (!delegate) {
+				throw new Error('Delegate was not found in the active delegate list!');
+			}
+
+			const delegateMinHeightActive = delegate.activeHeights.filter(
```
shuse2:
```suggestion
			const [delegateMinHeightActive] = delegate.activeHeights.filter(
```
Instead of using `[0]`

framework/src/modules/chain/dpos/dpos.js
```diff
@@ -77,17 +79,97 @@ module.exports = class Dpos {
...
+		const delegateListsWithActualRounds = delegateLists.map(list => ({
+			...list,
+			round: list.round + delegateListRoundOffset,
+		}));
+
+		const delegates = {};
+		for (let i = 0; i < numberOfRounds; i += 1) {
+			const { round, delegatePublicKeys } = delegateListsWithActualRounds[i];
+
+			delegatePublicKeys.forEach(publicKey => {
```
shuse2: prefer to use `for..of`

framework/src/modules/chain/components/storage/sql/round_delegates/get.sql
```diff
@@ -0,0 +1,26 @@
...
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+SELECT
+	"round"::bigint,
+	"delegatePublicKeys"::json
+FROM "round_delegates"
+WHERE
```
shuse2: we don't need `WHERE` here

framework/src/modules/chain/bft/bft.js
```diff
@@ -301,7 +314,34 @@ class BFT extends EventEmitter {
...
+				);
+			}
+
+			const [delegateMinHeightActive] = delegate.activeHeights.filter(
+				height => row.height >= height,
+			);
+
+			if (!delegateMinHeightActive) {
+				throw new Error(
+					'delegateMinHeightActive was not found for the delegate!',
```
shuse2:
```suggestion
					'delegateMinHeightActive was not found for the delegate.',
```

framework/src/modules/chain/dpos/dpos.js
```diff
@@ -77,17 +80,106 @@ module.exports = class Dpos {
...
+			{
+				sort: 'round:desc',
+				// limit: number or requested rounds + last 3 rounds
+				limit: numberOfRounds + this.delegateActiveRoundLimit,
+			},
+			tx,
+		);
+
+		if (
+			delegateLists.length - this.delegateActiveRoundLimit !==
```
shuse2: It needs to consider first 2 rounds where delegate list doesn't exist

framework/src/modules/chain/dpos/dpos.js
```diff
@@ -77,17 +80,106 @@ module.exports = class Dpos {
...
 
+	async getActiveDelegateHeights(
+		numberOfRounds = 1,
+		{ tx, delegateListRoundOffset = this.delegateListRoundOffset },
+	) {
+		// TODO: Discuss reintroducing a caching mechanism to avoid fetching
+		// active delegate lists multiple times.
+		const delegateLists = await this.storage.entities.RoundDelegates.get(
+			{
+				sort: 'round:desc',
```
shuse2: sort and limit is option, not filter, so the first parameter should be `{}`

framework/src/modules/chain/block_processor_v2.js
```diff
@@ -230,7 +240,33 @@ class BlockProcessorV2 extends BaseBlockProcessor {
...
+					numberOfRounds,
+					{ tx },
+				);
+
+				const delegate = delegateMinHeightActiveList.find(
+					d => block.generatorPublicKey === d.publicKey,
+				);
+
+				const delegateMinHeightActive = delegate.activeHeights.filter(
+					height => block.height >= height,
```
shuse2: why do we filter here with block.height? isnt it maximum within the array?

framework/src/modules/chain/block_processor_v2.js
```diff
@@ -230,7 +240,33 @@ class BlockProcessorV2 extends BaseBlockProcessor {
 				this.blocksModule.verify(block, stateStore, { skipExistingCheck }),
 			({ block, stateStore }) => this.blocksModule.apply(block, stateStore),
 			({ block, tx }) => this.dposModule.apply(block, { tx }),
-			({ block, tx }) => this.bftModule.addNewBlock(block, tx),
+			async ({ block, tx }) => {
+				// We only need activeMinHeight value of the delegate who is forging the block.
+				// Since the block is always the latest,
+				// fetching only the latest active delegate list would be enough.
+				const numberOfRounds = 1;
+				const delegateMinHeightActiveList = await this.dposModule.getActiveDelegateHeights(
```
shuse2: maybe this logic can be another public interface in the DPOS?

framework/src/modules/chain/dpos/dpos.js
```diff
@@ -77,12 +80,135 @@ module.exports = class Dpos {
...
 			disposableDelegateList,
 		);
 	}
 
+	async getActiveDelegateHeights(
+		numberOfRounds = 1,
+		{ tx, delegateListRoundOffset = this.delegateListRoundOffset } = {},
+	) {
+		const limit =
+			numberOfRounds + this.delegateActiveRoundLimit + delegateListRoundOffset;
```
shuse2: shouldn't this be `-1`?
`1 + 3 + 2` in the numberOfRounds = 1 case, but it should be 5?

framework/src/modules/chain/block_processor_v2.js
```diff
@@ -230,7 +240,33 @@ class BlockProcessorV2 extends BaseBlockProcessor {
...
+					numberOfRounds,
+					{ tx },
+				);
+
+				const delegate = delegateMinHeightActiveList.find(
```
shuse2: I think you confused yourself by calling this ActiveList, since this is a map.
It would be simpler to do
```
{
  'delegatePublicKey': [active heights]
}
```
and you can probably do
`const [delegateMinHeightActive ] = delegateMinHeightActiveMap[block.generatorPublicKey]`

framework/src/modules/chain/bft/bft.js
```diff
@@ -335,7 +347,30 @@ class BFT extends EventEmitter {
...
+					`Delegate "${
+						row.generatorPublicKey
+					}" was not found in the active delegate list.`,
+				);
+			}
+
+			// If there is no minHeightActive until this point,
+			// we can set the value to 0
+			const [delegateMinHeightActive = 0] = delegate.activeHeights.filter(
+				height => row.height >= height,
```
shuse2: just for remark, this should be the `raw.heigh - offset`

framework/src/modules/chain/bft/bft.js
```diff
@@ -335,7 +347,30 @@ class BFT extends EventEmitter {
...
+				throw new Error(
+					`Delegate "${
+						row.generatorPublicKey
+					}" was not found in the active delegate list.`,
+				);
+			}
+
+			// If there is no minHeightActive until this point,
+			// we can set the value to 0
+			const [delegateMinHeightActive = 0] = delegate.activeHeights.filter(
```
shuse2: please check if this default to `0` make sense with Jan

framework/src/modules/chain/bft/bft.js
```diff
@@ -72,7 +75,7 @@ class BFT extends EventEmitter {
 	 *
 	 * @return {Promise<void>}
 	 */
-	async init() {
+	async init(minActiveHeightsOfDelegates = {}) {
```
limiaspasdaniel: Please, add JSDOC for the parameter :)

framework/src/modules/chain/bft/bft.js
```diff
@@ -116,7 +120,7 @@ class BFT extends EventEmitter {
 	 * @param {Array.<Object>} blocks - List of all blocks
 	 * @return {Promise<void>}
 	 */
-	async deleteBlocks(blocks) {
+	async deleteBlocks(blocks, minActiveHeightsOfDelegates = {}) {
```
limiaspasdaniel: Please, add JSDOC for the parameter :)

framework/src/modules/chain/bft/bft.js
```diff
@@ -316,7 +324,11 @@ class BFT extends EventEmitter {
 	 * @param {int} tillHeight - The end height to fetch and load
 	 * @return {Promise<void>}
 	 */
-	async _loadBlocksFromStorage({ fromHeight, tillHeight }) {
+	async _loadBlocksFromStorage({
```
limiaspasdaniel: Please, add JSDOC for the parameter :)

framework/test/jest/unit/specs/modules/chain/bft/bft.spec.js
```diff
@@ -51,56 +51,72 @@ describe('bft', () => {
 
 	describe('extractBFTBlockHeaderFromBlock', () => {
 		it('should extract particular headers for bft', async () => {
+			// Arrange,
 			const block = blockFixture();
-
+			const delegateMinHeightActive = 304;
```
limiaspasdaniel: Can we use ACTIVE_DELEGATES*3 + 1. This way it's easier to read.

framework/test/jest/unit/specs/modules/chain/bft/bft.spec.js
```diff
@@ -154,59 +170,65 @@ describe('bft', () => {
...
+				const minActiveHeightsOfDelegates = { dummy: 'object' };
 
 				bft._initFinalityManager.mockReturnValue({
 					finalizedHeight,
 					on: jest.fn(),
 				});
 				bft._getLastBlockHeight.mockReturnValue(lastBlockHeight);
 
-				await bft.init();
+				await bft.init(minActiveHeightsOfDelegates);
```
limiaspasdaniel: ❤️

framework/test/jest/unit/specs/modules/chain/dpos/getMinActiveHeightsOfDelegates.spec.js
```diff
@@ -0,0 +1,133 @@
...
+
+			// Assert
+			expect(Object.keys(minActiveHeightsOfDelegates).length <= 3).toBe(true);
+			expect(minActiveHeightsOfDelegates[newDelegate]).toEqual([
+				slots.calcRoundStartHeight(activeRound),
+			]);
+		});
+	});
+
+	describe.skip('Given a delegate was active in previous 1 round', () => {
```
limiaspasdaniel: Do we need to implement these tests for this PR?

yatki: Nope, I'll add them separately.

limiaspasdaniel: Thanks! If so, please could you create an issue and refer it here for the record :)

framework/src/modules/chain/bft/bft.js
```diff
@@ -73,6 +73,7 @@ class BFT extends EventEmitter {
 	/**
 	 * Initialize the BFT module
 	 *
+	 * @param {Object} minActiveHeightsOfDelegates - Minimum active heights of a delegate
```
limiaspasdaniel: Shouldn't it be `number` or maybe I'm mistaken. The same for the rest.  Also I think it's optional param

yatki: No the structure is {[string]: number[] }

yatki: Also actually it's not optional.

yatki:
@shuse2 Created #4499

framework/src/modules/chain/dpos/dpos.js
```diff
@@ -77,12 +80,132 @@ module.exports = class Dpos {
...
+		// TODO: Discuss reintroducing a caching mechanism to avoid fetching
+		// active delegate lists multiple times.
+		let delegateLists = await this.storage.entities.RoundDelegates.get(
+			{},
+			{
+				// IMPORTANT! All logic below based on ordering rounds in
+				// descending order. Change it at your own discretion!
+				sort: 'round:desc',
+				// limit: number or requested rounds + last 3 rounds
+				limit,
```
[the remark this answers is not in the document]
yatki: I removed the comment since the variable declaration is pretty straightforward.

PR #4478, opened 2019-11-08
Request node info from Outbound peers - Closes #4260

### LINKED ISSUE #4260: Request nodeInfo from outbound peers before adding to peer lists
### Description

Upon outbound connections, a lisk node should request `nodeInfo`:`protocolVersion` to see if it is compatible before adding to peer lists. If compatible, add to relevant peer lists. If the version is not compatible, it should disconnect the peer.

### Motivation

Handshake checks minimum `protocolVersion` for inbound peers but not outbound. Currently there is no way to limit outbound connected peers by checking the `protocolVersion`. 

### Acceptance Criteria

Proper functionality implemented in `p2p` and `peerPool`.

mitsuaki-u:
### What was the problem?

Handshake checks minimum protocolVersion for inbound peers but not outbound. 

### How did I solve it?

Upon outbound peer connection and fetching status, we check peer compatibility. 

### How to manually test it?

npm run test

### Review checklist

- [ ] The PR resolves #4260 


elements/lisk-p2p/package.json
```diff
@@ -65,6 +65,7 @@
 		"@types/jsonwebtoken": "8.3.0",
 		"@types/lodash.clonedeep": "4.5.0",
 		"@types/lodash.shuffle": "4.2.6",
+		"@types/lodash.uniq": "4.5.0",
```
shuse2: I just realized that we don't have `clonedeep` and `uniq` installed, but seems like using it

elements/lisk-p2p/src/peer/outbound.ts
```diff
@@ -135,7 +135,7 @@ export class OutboundPeer extends Peer {
 		outboundSocket.on('connect', async () => {
 			this.emit(EVENT_CONNECT_OUTBOUND, this._peerInfo);
 			try {
-				await Promise.all([this.fetchStatus(), this.discoverPeers()]);
+				await Promise.all([this.fetchAndUpdateStatus(), this.discoverPeers()]);
```
shuse2: Instead of calling with `Promise.all` I think it make sense to call the `fetchAndUpdateStatus` first to check compatibility and then ask for the peerList

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -492,17 +496,28 @@ export class Peer extends EventEmitter {
...
 		);
+
+		if (!this._serverNodeInfo) {
+			throw new Error('Missing server node info.');
+		}
+
+		const result = validatePeerCompatibility(newPeerInfo, this._serverNodeInfo);
+
+		if (!result.success && result.errors) {
+			throw result.errors[0];
```
shuse2: `result.errors[0]` is `string`, so I think it should be `throw new Error(result.errors[0]`?
Also, I don't know why only this function is returning `errors` as array but always one item

mitsuaki-u: I'm not sure why but looks like the `validator` function was designed that way. I will change it to single error.

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -492,17 +496,28 @@ export class Peer extends EventEmitter {
...
 			ip: this._ipAddress,
 			wsPort: this._wsPort,
 		};
+
 		const newPeerInfo = validatePeerInfo(
 			protocolPeerInfo,
 			this._peerConfig.maxPeerInfoSize,
 		);
+
+		if (!this._serverNodeInfo) {
```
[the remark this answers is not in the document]
mitsuaki-u: I think it's ok to validate peer info size before?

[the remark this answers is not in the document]
diego-G: It's better to do fast returns before anything else. Btw, `validatePeerInfo ` is throwing an error which it isn't being caught, is it?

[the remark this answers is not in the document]
mitsuaki-u: `validatePeerInfo()` error is being caught in the calling function `fetchAndUpdateStatus`

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -492,17 +496,28 @@ export class Peer extends EventEmitter {
...
+
 		const newPeerInfo = validatePeerInfo(
 			protocolPeerInfo,
 			this._peerConfig.maxPeerInfoSize,
 		);
+
+		const result = validatePeerCompatibility(newPeerInfo, this._serverNodeInfo);
+
+		if (!result.success && result.error) {
+			throw new Error(result.error);
```
ishantiw: we can throw a similar error to `PeerInboundHandshakeError` at least with peer ip and port for logging purpose

PR #4479, opened 2019-11-08
Update broadcaster to send transactionIds - Closes #4460

### LINKED ISSUE #4460: Update broadcaster to store transaction and signature separately
### Description

Since transaction and signature are propagated in different way, it should have different queuing system.
Also, Update the schema to be 
```
{
  ids: []
}
```

### Motivation

Transaction lazy push schema should be efficient and understandable.

### Acceptance Criteria

- Node should be able to receive transaction ids
- Node should be able to receive signatures

### Additional Information
- Need to be cautious when removing from the queue since sometime it is added at the same time

shuse2:
### What was the problem?
- transactionId and signature was queued in the same place

### How did I solve it?
- Update schema to use transactionId
- Update broadcaster to queue transactionid and signature separately
- Sends both transaction and signature separately

### How to manually test it?
Connect 2 nodes and send transaction, signature and blocks

### Review checklist

- [ ] The PR resolves #4460 


framework/src/modules/chain/transport/broadcaster.js
```diff
@@ -14,8 +14,8 @@
 
 'use strict';
 
-const _ = require('lodash');
-const jobsQueue = require('../utils/jobs_queue');
+const ENDPOINT_BORADCAST_TRANSACTIONS = 'postTransactionsAnnouncement';
```
limiaspasdaniel:
```suggestion
const ENDPOINT_BROADCAST_TRANSACTIONS = 'postTransactionsAnnouncement';
```

framework/src/modules/chain/transport/broadcaster.js
```diff
@@ -14,8 +14,8 @@
...
+const ENDPOINT_BORADCAST_SIGNATURES = 'postSignatures';
```
limiaspasdaniel:
```suggestion
const ENDPOINT_BROADCAST_SIGNATURES = 'postSignatures';
```

framework/src/modules/chain/transport/broadcaster.js
```diff
@@ -28,41 +28,28 @@ const jobsQueue = require('../utils/jobs_queue');
...
+		this.transactionIdQueue = [];
+		this.signatureObjectQueue = [];
 
 		if (this.config.active) {
-			jobsQueue.register(
-				'broadcasterReleaseQueue',
-				async () => this.releaseQueue(),
-				this.config.broadcastInterval,
-			);
+			setInterval(async () => {
```
limiaspasdaniel: Just for curiosity, why not using jobsQueue anymore?

framework/src/modules/chain/transport/broadcaster.js
```diff
@@ -71,189 +58,75 @@ class Broadcaster {
...
+		if (this.transactionIdQueue.length > 0) {
+			this.transactionIdQueue = this.transactionIdQueue.filter(id =>
+				this.transactionPool.transactionInPool(id),
+			);
+			const transactionIds = this.transactionIdQueue.slice(
+				0,
+				this.config.releaseLimit,
+			);
+			await this.channel.invoke('network:broadcast', {
+				event: ENDPOINT_BORADCAST_TRANSACTIONS,
```
limiaspasdaniel:
```suggestion
				event: ENDPOINT_BROADCAST_TRANSACTIONS,
```

framework/src/modules/chain/transport/broadcaster.js
```diff
@@ -71,189 +58,75 @@ class Broadcaster {
...
+		if (this.signatureObjectQueue.length > 0) {
+			const signatures = this.signatureObjectQueue.slice(
+				0,
+				this.config.releaseLimit,
 			);
-
-			return this.logger.info(
-				`Broadcasts released: ${squashedBroadcasts.length}`,
+			await this.channel.invoke('network:send', {
+				event: ENDPOINT_BORADCAST_SIGNATURES,
```
limiaspasdaniel:
```suggestion
				event: ENDPOINT_BROADCAST_SIGNATURES,
```

framework/src/modules/chain/transport/transport.js
```diff
@@ -63,20 +62,19 @@ class Transport {
...
 
-		this.broadcaster = new Broadcaster(
-			this.constants.broadcasts,
-			this.transactionPoolModule,
-			this.logger,
-			this.channel,
-			this.storage,
-		);
+		this.broadcaster = new Broadcaster({
+			broadcasts: this.constants.broadcasts,
```
limiaspasdaniel: is `broadcasts` a flag?

framework/src/modules/chain/transport/transport.js
```diff
@@ -321,7 +272,7 @@ class Transport {
 	async handleRPCGetSignatures() {
 		const transactions = this.transactionPoolModule.getMultisignatureTransactionList(
 			true,
-			this.constants.maxSharedTransactions,
+			this.constants.broadcasts.releaseLimit,
```
limiaspasdaniel: Maybe `txReleaseLimit`?

framework/test/jest/unit/specs/modules/chain/transport/transport.spec.js
```diff
@@ -0,0 +1,359 @@
...
+describe('Transport', () => {
+	const defaultBroadcastInterval = 5000;
+	const defaultReleaseLimit = 25;
+
+	let transport;
+	let transactionPoolStub;
+	let synchronizerStub;
+	let blocksStub;
+	let loggerStub;
+	let processorStub;
```
limiaspasdaniel: IMO and following the recent testing guidelines we agreed on, we shouldn't mock processor or blocks module, only the public interfaces that deal with I/O

framework/src/modules/chain/transport/broadcaster.js
shuse2: I think jonsQueue doesn't make sense in a way that it's every 5 sec, so it will never be stuck in the execution. So in this case i thought `setInterval` is enough

framework/src/modules/chain/transport/transport.js
shuse2: this was the public config value existed before.

framework/src/modules/chain/transport/transport.js
shuse2: broadcasts directly comes from the config, which is `object` 😢

framework/test/jest/unit/specs/modules/chain/transport/transport.spec.js
shuse2: I think in the guidelines, it states that we should mock the external modules, in this case, other blocks and processors are external

framework/src/modules/chain/transport/broadcaster.js
```diff
@@ -71,189 +58,75 @@ class Broadcaster {
...
-	async broadcast(params, { api: event, data }) {
-		// Broadcast using Elements P2P library via network module
-		if (event === 'postTransactionsAnnouncement') {
-			return this.channel.invoke('network:broadcast', {
-				event,
-				data,
-			});
+	enqueueTransactionId(transactionId) {
+		if (
+			this.transactionIdQueue.find(id => id === transactionId) !== undefined
```
MaciejBaj: Why not a simple object with transaction ids as keys? Same for signatures. No need for `.find` then, just a check.

shuse2: id queue is `FIFO` queue, so we cannot use map to keep the order

framework/src/modules/chain/transport/schemas.js
```diff
@@ -162,20 +156,15 @@ module.exports = {
 		},
 	},
 	transactionsRequest: {
```
diego-G: The schema names are not following any pattern at all. Can we have one? Otherwise, let's call this new one `transactionIDsRequest` at least.

framework/src/modules/chain/transport/broadcaster.js
```diff
@@ -14,8 +14,8 @@
 
 'use strict';
 
-const _ = require('lodash');
-const jobsQueue = require('../utils/jobs_queue');
+const ENDPOINT_BROADCAST_TRANSACTIONS = 'postTransactionsAnnouncement';
```
diego-G: If you introduce this, you should make it in more places and tests as well. Chain file is using the event and transport spec.

framework/src/modules/chain/transport/transport.js
```diff
@@ -87,20 +85,9 @@ class Transport {
 	 * @emits signature/change
 	 * @todo Add description for the params
 	 */
-	// eslint-disable-next-line class-methods-use-this
-	handleBroadcastSignature(signature, broadcast) {
-		if (broadcast) {
```
diego-G: You've removed this check which likely is still needed.

framework/src/modules/chain/transport/transport.js
```diff
@@ -111,21 +98,9 @@ class Transport {
 	 * @emits transactions/change
 	 * @todo Add description for the params
 	 */
-	// eslint-disable-next-line class-methods-use-this
-	handleBroadcastTransaction(transaction, broadcast) {
-		if (broadcast) {
```
diego-G: Same as above.

framework/test/mocha/unit/modules/chain/transport/broadcaster.js
```diff
@@ -1,439 +0,0 @@
-/*
```
diego-G: Where are the replacement of these unit tests?

framework/src/modules/chain/transport/transport.js
shuse2: This was not used in anywhere it was always called with true

framework/src/modules/chain/transport/transport.js
shuse2: same as above

framework/src/modules/chain/transport/broadcaster.js
shuse2: this constant is not exposed, it is just to make it clear in this file

framework/test/mocha/unit/modules/chain/transport/broadcaster.js
shuse2: since broadcaster is not exposed, it's not explicitly tested, but it's covered by the transport.spec test

framework/src/modules/chain/transport/transport.js
diego-G: If broadcast is disabled via config file, you're gonna fill the queues with transactions which won't be released. We were preventing this with that check.

shuse2: https://github.com/LiskHQ/lisk-sdk/pull/4479/files#diff-eafdfefa460d6943a1a938936ca754b5L595
That's a good point, but actually it was not been handled in that way before.

framework/src/modules/chain/transport/broadcaster.js
diego-G: You could expose it and reuse it.

framework/test/mocha/unit/modules/chain/transport/broadcaster.js
diego-G: Some basic unit tests won't hurt anybody. Your call.

framework/src/modules/chain/transport/schemas.js
shuse2: I will update all of them in another issue

framework/src/modules/chain/transport/broadcaster.js
shuse2: Prefer not to expose this since this will be endpoint definition, and I think it make sense to keep it as it shouldn't be updated by mistake

diego-G: Then, why do you create a constant?? It doesn't make sense to create one here and not in the rest...

framework/src/modules/chain/transport/transport.js
diego-G: A good point forgotten because before wasn't handled?

shuse2: yes before it wasn't handled, and i think no-broadcast was just for testing if i remember correctly.
I will create an issue to remove that

PR #4482, opened 2019-11-09
Add missing apply penalty from LIP-0004 - Closes #3672

### LINKED ISSUE #3672: Invoke `applyPenalty` network actions in framework
### Expected behavior

As per [Lip 0004](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0004.md), we should apply penalties to peers for a list of bad behaviors on the framework level.

All of the banning type are described in https://github.com/LiskHQ/lips/blob/master/proposals/lip-0004.md#banning-mechanism-1

### Notes
- It is blocked by lazy push/eager push because this cannot be fully completed until they are implemented.

2.3

### Acceptance Criteria
- For bad blocks and transaction, peer gets banned after a few transactions( As per Lips 004 )

shuse2:
### What was the problem?
Apply penalty was not completely applied to all the RPC endpoints.

### How did I solve it?
- Add apply penalty on the schema errors.
- Expose peerId on the request endpoints

### How to manually test it?
- run tests
- observe behavior with 2 nodes

### Review checklist

- [ ] The PR resolves #3672 


framework/src/modules/chain/transport/transport.js
```diff
@@ -254,8 +291,11 @@ class Transport {
 		const errors = validator.validate(schemas.signaturesList, query);
 
 		if (errors.length) {
-			this.logger.debug({ err: errors }, 'Invalid signatures body');
-			// TODO: If there is an error, invoke the applyPenalty action on the Network module once it is implemented.
+			this.logger.warn({ err: errors }, 'Invalid signatures body');
```
limiaspasdaniel: Why don't instead of calling applyPenalty when an invalid input is received, we throw a specific error that is caught in the upper scope and then we apply penalty there? I see this is a repeating pattern and we could avoid code repetition.

shuse2: Actually, this cannot be done since each function is exposed as a action, and there is no outer function than controller now

framework/test/mocha/functional/ws/transport/transport.js
```diff
@@ -136,7 +136,9 @@ describe('WS transport', () => {
 			}
 		});
 
-		it('should fail when using an empty payload', async () => {
+		// This cannot be tested in the functional, since it will get banned
+		// eslint-disable-next-line mocha/no-skipped-tests
+		it.skip('should fail when using an empty payload', async () => {
```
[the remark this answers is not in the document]
shuse2: ahh this can be actually removed, I was keeping it until I cover in the unit tests

Issue #4360, opened 2019-10-02
Discover from seed peer when triedPeers < 100


ishantiw:
### Expected behavior

Discover from seed peer when triedPeers < 100

### Actual behavior

Right now we always do discovery from the seed peers on the node startup.

### Steps to reproduce

Start the application and check if the node is using seed peers initially to discover new peers. Or use blank seed peer lists to see if the node is still discovering new peers from the previously saved peers in the database.

### Which version(s) does this affect? (Environment, OS, etc...)
`3.0`

PR #4516, opened 2019-11-15
P2P Disconnect from seed peers after initial discovery - Closes #4359 & #4360

### LINKED ISSUE #4359: Disconnect from seed peers after initial discovery
### Description

Disconnect from seed peers after the initial discovery

### Motivation

We don't want to overwhelm the seed peers as many other nodes are connecting to them for discovery.

The main argument against this issue is that seed peers have a limited number of inbound slots so they already evict peers on their own so there is no need for peers to voluntarily disconnect from them.

### Acceptance Criteria

We need to discuss the pros and cons and decide if we want to implement this feature or not.

valamidev:
## What was the problem?

According the LIP-0004 we should disconnect from SeedPeers after we executed the initial discovery(get their PeerList).

### How did I solve it?

- Add `_fetchSeedPeerList()` method into the [elements/lisk-p2p/src/p2p.ts](elements/lisk-p2p/src/p2p.ts) class which is executed before/parallel within the `_startPopulator()`. 
- Extend PeerPool with `discoverSeedPeers` function and add `_discoverySeedPeer` flag into the  Peer/Base to mark the Peer connection with initial or discovery purpose.
- Remove insertion of SeedPeers from the peerBook at start-up, lately they still can be inserted into the peerBook as triedPeers from PeerPool.  
- Add `_initializePeerBook()` function into [elements/lisk-p2p/src/p2p.ts](elements/lisk-p2p/src/p2p.ts) to handle loading previousPeers and whitelistedPeers in one place.

Exceptions:
- whitelistedPeers which are also seedPeers cannot be disconnected.
- disconnection event triggered when we reach the maxOutboundConnections limit to avoid case when the Node has no outbound connection and could not send request() into the network.

### How to manually test it?

`npm run t`

### Review checklist

- [x] The PR resolves #4359 & #4360


elements/lisk-p2p/src/p2p.ts
```diff
@@ -1008,6 +1022,8 @@ export class P2P extends EventEmitter {
 
 		// We need this check this._isActive in case the P2P library is shut down while it was in the middle of starting up.
 		if (this._isActive) {
+			// Initial Discovery SeedPeers and Disconnect
```
ishantiw: According to LIP, we need to check if the number of previous peers is less than 100 then we do the discovery with seed peers. Otherwise, we choose only previous (tried) peers to do the discovery initially.

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -102,6 +103,7 @@ export interface PeerConfig {
 	readonly wsMaxPayload?: number;
 	readonly maxPeerInfoSize: number;
 	readonly maxPeerDiscoveryResponseLength: number;
+	readonly discoverySeedPeer?: boolean;
```
ishantiw: I think you can use internal state and use peerKind to know if its a seed peer or not

valamidev: To avoid saving this flag and cause unwanted side effects later, I suggest that better to keep outside from internalState.

elements/lisk-p2p/src/p2p.ts
```diff
@@ -982,32 +985,42 @@ export class P2P extends EventEmitter {
...
-		this._sanitizedPeerLists.whitelisted.forEach(whitelistPeer =>
-			this._peerBook.upgradePeer(whitelistPeer),
-		);
 		await this._startPeerServer();
 
 		// We need this check this._isActive in case the P2P library is shut down while it was in the middle of starting up.
 		if (this._isActive) {
+			// Initial Discovery SeedPeers and Disconnect LIP-0004
+			if (this._peerBook.triedPeers.length < DEFAULT_MIN_TRIED_PEER_COUNT) {
+				this._fetchSeedPeerList();
```
ManuGowda: rename function to `connectToSeedPeers`?

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -408,17 +437,17 @@ export class PeerPool extends EventEmitter {
 		const disconnectedTriedPeers = triedPeers.filter(
 			triedPeer => !this._peerMap.has(triedPeer.peerId),
 		);
-		const { outboundCount } = this.getPeersCountPerKind();
-		const disconnectedFixedPeers = fixedPeers
+		const disconnectedFixedPeers = this._peerLists.fixedPeers
 			.filter(peer => !this._peerMap.get(peer.peerId))
 			.map(peer2Convert => peer2Convert);
```
ManuGowda: whats it doing? do we need this?

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -589,6 +620,22 @@ export class PeerPool extends EventEmitter {
 		throw new Error(`Peer not found: ${peerPenalty.peerId}`);
 	}
 
+	private _getAvailableOutboundConnectionSlots(): number {
+		const { outboundCount } = this.getPeersCountPerKind();
+
+		const disconnectedFixedPeers = this._peerLists.fixedPeers
+			.filter(peer => !this._peerMap.get(peer.peerId))
+			.map(peer2Convert => peer2Convert);
```
ManuGowda: whats it doing? do we need this?

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -597,6 +644,20 @@ export class PeerPool extends EventEmitter {
 		}
 	}
 
+	private _disconnectDiscoverySeedPeers(): void {
+		const outboundPeers = this.getPeers(OutboundPeer);
+
+		outboundPeers.forEach(peer => {
+			if (peer.discoverySeedPeer) {
```
ManuGowda: can we name `discoverySeedPeer` as boolean notation, `isDiscoverySeedPeer`?

elements/lisk-p2p/test/functional/advertise_address.ts
```diff
@@ -35,7 +35,7 @@ describe('Advertise Address', () => {
 			},
 		],
 		wsEngine: 'ws',
-		populatorInterval: POPULATOR_INTERVAL,
+		populatorInterval: POPULATOR_INTERVAL + 15,
```
ManuGowda: why extra 15?

valamidev: Avoid `Peer has a duplicate connection` loop error which could rarely happen but make some test failing.

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -396,10 +403,32 @@ export class PeerPool extends EventEmitter {
 		peer.send(message);
 	}
 
+	public discoverSeedPeers(): void {
```
diego-G: Please cover this public method with unit.

elements/lisk-p2p/src/p2p.ts
```diff
@@ -983,39 +982,49 @@ export class P2P extends EventEmitter {
...
 	}
 
-	public async start(): Promise<void> {
-		if (this._isActive) {
-			throw new Error('Cannot start the node because it is already active');
-		}
+	private _initializePeerBook(): void {
+		const newPeersToAdd = [
+			...this._sanitizedPeerLists.previousPeers,
+			...this._sanitizedPeerLists.whitelisted,
```
shuse2: why no `fixedPeer here`?

valamidev: They could, but we are connecting to fixedPeers are not affected by having them in PeerBook. [PeerPool](https://github.com/LiskHQ/lisk-sdk/blob/development/elements/lisk-p2p/src/peer_pool.ts#L432)

valamidev: Implemented

elements/lisk-p2p/src/p2p.ts
```diff
@@ -317,6 +319,18 @@ export class P2P extends EventEmitter {
 			if (this._isNetworkReady()) {
 				this.emit(EVENT_NETWORK_READY);
 			}
+
+			// Once we successfuly connected to a SeedPeer we try to make new connection or initiate eviction from it LIP-0004
+			const isSeedPeer = this._sanitizedPeerLists.seedPeers.find(
+				peer => peer.peerId === peerInfo.peerId,
+			);
+
+			if (isSeedPeer) {
```
ishantiw: So when we successfully connect to a `seedPeer`,  we should not initiate the discovery from others. When we make an outbound connection we `fetchPeers` as part of `connect` event so that's enough to cover discovery from seed

elements/lisk-p2p/test/integration/discovery.ts
```diff
@@ -31,176 +34,231 @@ import {
...
+					protocolVersion: '9.9',
+					minVersion: '9.9.9',
+					os: platform(),
+					height: 10000,
+					nonce: `404`,
+					advertiseAddress: true,
+				},
+			});
+			await disconnectedNode.start();
+			await wait(1000);
```
ishantiw: I think ~200 ms should be enough

elements/lisk-p2p/test/unit/p2p.ts
```diff
@@ -17,7 +17,7 @@ import { P2P } from '../../src/p2p';
 
 describe('p2p', () => {
 	describe('#constructor', () => {
-		const lisk = new P2P({
+		const P2PNode = new P2P({
```
ishantiw: 👍

elements/lisk-p2p/src/p2p.ts
valamidev: Seems this part could be removed.

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -409,17 +431,17 @@ export class PeerPool extends EventEmitter {
 		const disconnectedTriedPeers = triedPeers.filter(
 			triedPeer => !this._peerMap.has(triedPeer.peerId),
 		);
-		const { outboundCount } = this.getPeersCountPerKind();
-		const disconnectedFixedPeers = fixedPeers
-			.filter(peer => !this._peerMap.get(peer.peerId))
-			.map(peer2Convert => peer2Convert);
+		const disconnectedFixedPeers = this._peerLists.fixedPeers.filter(
+			peer => !this._peerMap.get(peer.peerId),
```
mitsuaki-u: Minor but can also use `has` here

PR #4529, opened 2019-11-18
Fix applying penalty on transaction processing error - Closes #4528

### LINKED ISSUE #4528: Node should not apply penalty while processing transaction
### Expected behavior
It should only apply penalty if the transaction values are invalid.

### Actual behavior
If transaction processing fails, it applies penalty, but it can be invalid in various reasons.
this transaction processing could fail on validation, or verification or existence.

### Steps to reproduce
Send already processed transaction

3.0-

shuse2:
### What was the problem?
If transaction fails on the processing, it was applying the penalty.
However, transaction processing can fail in various reason.
- not synced
- already processed
- etc

### How did I solve it?
- Create error type for invalid transaction
- Only apply error in case of this error

### How to manually test it?
Try to send already processed transaction

### Review checklist

- [ ] The PR resolves #4528 

### DIFF
--- a/framework/test/jest/unit/specs/modules/chain/transport/transport.spec.js
+++ b/framework/test/jest/unit/specs/modules/chain/transport/transport.spec.js
@@ -44,6 +44,7 @@ describe('Transport', () => {
 		loggerStub = {
 			info: jest.fn(),
 			warn: jest.fn(),
+			error: jest.fn(),
 			debug: jest.fn(),
 		};
 		storageStub = {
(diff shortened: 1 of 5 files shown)


framework/test/jest/unit/specs/modules/chain/transport/transport.spec.js
```diff
@@ -922,7 +923,7 @@ describe('Transport', () => {
 				).toHaveBeenCalledTimes(2);
 			});
 
-			it('should apply penalty when validateTransactions fails', async () => {
+			it.only('should apply penalty when validateTransactions fails', async () => {
```
limiaspasdaniel: only? 😏

shuse2: 😢

PR #4536, opened 2019-11-20
Update BFT to use state store - Closes #4316

### LINKED ISSUE #4316: Use state store for updating finalizedHeight
### Description

In BFT module, when we update the finalizedHeight, we use database transaction.
https://github.com/LiskHQ/lisk-sdk/blob/feature/introduce_bft_consensus/framework/src/modules/chain/bft/bft.js#L141

However, since this is calculated value from blocks header (state), it should use the state store consistently

Also, `maxHeightPreviouslyForged` should be moved to forger module, and handled in different database, since this has to be handled differently in snapshot

### Motivation

All of the state changes based on blocks or transaction should happen through state store.

### Acceptance Criteria

- Add block and delete block uses state store instead of database transaction

shuse2:
- Add forger info table and move maxHeightPreviouslyForged to this table
- Rename chain meta to chain state table
- Update source code of BFT to use state store to update data

### What was the problem?
- BFT was not consistently using state store
- max height previously forged was stored in the chain state, but it has to be in different table

### How did I solve it?
- Create forger info table so that it can be handled in different way when snapshotting / restoring
- Update BFT to use state store

### How to manually test it?
- All the test should pass
- Observe database info that forger_info is populated
- Observe finality is saved

### Review checklist

- [ ] The PR resolves #4316 


framework/src/modules/chain/components/storage/entities/forger_info.js
```diff
@@ -0,0 +1,184 @@
...
+			.then(result => result);
+	}
+
+	_getResults(filters, options, tx, expectedResultCount = undefined) {
+		this.validateFilters(filters);
+		this.validateOptions(options);
+
+		const mergedFilters = this.mergeFilters(filters);
+		const parsedFilters = this.parseFilters(mergedFilters);
+		const parsedOptions = defaults(
```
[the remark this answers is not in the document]
shuse2: Assuming it was working before, I think this will be the correct order. I didn't change anything from the other implementation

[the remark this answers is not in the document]
limiaspasdaniel: Hope it didn't come from `ChainMeta` entity 🧐😂

limiaspasdaniel:
@shuse2 tests are failing :(

framework/src/modules/chain/components/storage/entities/forger_info.js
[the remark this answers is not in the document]
yatki: Let's not use lodash anymore bitte :D

framework/src/modules/chain/bft/bft.js
```diff
@@ -182,48 +177,6 @@ class BFT extends EventEmitter {
 		);
 	}
 
-	/**
-	 * Computes maxHeightPreviouslyForged and maxHeightPrevoted properties that are necessary
-	 * for creating a new block
-	 * @param delegatePublicKey
-	 * @return {Promise<{maxHeightPrevoted: number, maxHeightPreviouslyForged: (number|*)}>}
-	 */
-	async computeBFTHeaderProperties(delegatePublicKey) {
```
limiaspasdaniel: Where'd this logic go?

shuse2: It's in the processor v2.
maxHeightPreviouslyForged is calculated in v2, and maxHeightPrevoted is getting from getter

limiaspasdaniel: I'm not sure about moving this logic to block processor v2, what's the reason behind it? Feels like it belongs to `bft.js` or at least `bft` folder.

shuse2: maxForgedHeight is related to `BFT`, but managed by forging module.
it's out of `BFT` because `forger_info` module is nothing to do with `bft`, and we will include `dpos` related information in the `forger_info` in the future too

framework/src/modules/chain/block_processor_v2.js
```diff
@@ -280,16 +285,19 @@ class BlockProcessorV2 extends BaseBlockProcessor {
...
 			// Create a block with with basic block and bft properties
-			(data, bftHeader) => this._create({ ...data, ...bftHeader }),
-			async (data, block) => {
-				// Saving maxHeightPreviouslyForged before broadcasting
-				await this.bftModule.saveMaxHeightPreviouslyForged(
+			async data => {
+				const previouslyForged = await this._getPreviouslyForgedMap();
+				const delegatePublicKey = data.keypair.publicKey.toString('hex');
+				const maxHeightPreviouslyForged =
+					previouslyForged[delegatePublicKey] || 0;
```
limiaspasdaniel: Does this have to default to `0` according to what Jan said?

framework/src/modules/chain/block_processor_v2.js
```diff
@@ -372,6 +380,38 @@ class BlockProcessorV2 extends BaseBlockProcessor {
...
+	/**
+	 * Saving a height which delegate last forged. this needs to be saved before broadcasting
+	 * so it needs to be outside of the DB transaction
+	 * @param delegatePublicKey
+	 * @param height
+	 */
+	async _saveMaxHeightPreviouslyForged(delegatePublicKey, height) {
+		const previouslyForgedMap = await this._getPreviouslyForgedMap();
+		const previouslyForgedHeightByDelegate =
+			previouslyForgedMap[delegatePublicKey] || 0;
```
limiaspasdaniel: And same here

framework/src/modules/chain/block_processor_v2.js
shuse2: According to Jan, If the delegate is forging for the first time, 0 should be the default value.
However, if the data is corrupted or missing, it has to be highest value possible. Jan is thinking a good default value in that case, but it will be tackled in different issue

framework/src/modules/chain/block_processor_v2.js
shuse2: same as above

framework/src/modules/chain/block_processor_v2.js
limiaspasdaniel: Alright thanks for the clarification

framework/src/modules/chain/bft/bft.js
```diff
@@ -76,8 +73,8 @@ class BFT extends EventEmitter {
 	 * @param {Object} minActiveHeightsOfDelegates - Minimum active heights of a delegate
```
ManuGowda: can you add the stateStore to docs?

PR #4545, opened 2019-11-21
P2P Update peer buckets with source_group bytes - Closes #4488

### LINKED ISSUE #4488: Use source_group in calculation of bucketId
### Description

Currently we do not take into account `source_group` bytes when calculating `bucketId`.

### Motivation

In order to mitigate peer list poisoning risk, we will include ip address bytes of peer who provides new peer lists into the calculation of `bucketId`. 

### Acceptance Criteria

For a given source_group all IP addresses advertised by this source_group are hashed to at most 16 buckets.

### Additional Information

n/a

mitsuaki-u:
### What was the problem?

Given a `source_group` IP address (a peer who provides a peer list), all IP addresses advertised by this `source_group` are hashed to at most 16 buckets. A single malicious Lisk node cannot completely fill the `newPeers` collection of its peers by advertising many IP addresses, but can only fill up 16 buckets. This will mitigate peer list poisoning risk, as we will include ip address bytes of peer who provides new peer lists into the calculation of `bucketId`. 

### How did I solve it?

- [x] Added `source_group` bytes into `getBucketId`.

- [x] Added auxiliary `map` in `BaseList` for direct `peerId` to `peerInfo`  lookups.

- [x] Updated types in `peer book` and it's lists.

- [x] Updated `p2p` and `peerPool` to include `sourceAddress` when adding new peers to `PeerBook`.

- [x] Refactored functions in `PeerBook`.

- [x] Removed unnecessary try catch in p2p handlers

- [x] Update tests

### How to manually test it?

`npm run test`

### Review checklist

- [x] The PR resolves #4488 


elements/lisk-p2p/src/p2p.ts
```diff
@@ -349,26 +333,16 @@ export class P2P extends EventEmitter {
 		};
 
 		this._handlePeerInfoUpdate = (peerInfo: P2PPeerInfo) => {
-			try {
+			const existingPeer = this._peerBook.getPeer(peerInfo);
```
diego-G: Why do we change this back? The idea is to not to check if the peer exists twice. `addPeer()` throws an error if the peer exists.

elements/lisk-p2p/src/p2p.ts
```diff
@@ -435,31 +409,17 @@ export class P2P extends EventEmitter {
...
-							sharedState: detailedPeerInfo.sharedState
-								? { ...detailedPeerInfo.sharedState }
-								: error.peerInfo.sharedState,
-						};
-						const isUpdated = this._peerBook.updatePeer(updatedPeerInfo);
-						if (isUpdated) {
-							// If found and updated successfully then upgrade the peer
-							this._peerBook.upgradePeer(updatedPeerInfo);
-						}
+				this._peerBook.addPeer(detailedPeerInfo);
```
diego-G: Same as above.

elements/lisk-p2p/src/p2p.ts
```diff
@@ -980,16 +943,7 @@ export class P2P extends EventEmitter {
...
-				this._peerBook.addPeer(peerInfo);
-				this._peerBook.upgradePeer(peerInfo);
-			} catch (error) {
-				if (!(error instanceof ExistingPeerError)) {
-					throw error;
-				}
-
-				this._peerBook.upgradePeer(error.peerInfo);
-			}
+			this._peerBook.upgradePeer(peerInfo);
```
diego-G: I think before upgrading any peer we should add them to the peer book. Is the logic changing?

elements/lisk-p2p/src/peer_book/base_list.ts
```diff
@@ -13,146 +13,192 @@
...
 	readonly secret: number;
 	readonly peerType: PEER_TYPE;
 }
 
-export interface CustomPeerInfo {
-	readonly peerInfo: P2PPeerInfo;
-	readonly dateAdded: Date;
+export type Bucket = Map<string, P2PEnhancedPeerInfo>;
+
+export interface BucketInfo {
```
diego-G: This type looks weird. I was expecting a map of buckets identified by its ids. Do we really need it?

elements/lisk-p2p/src/peer_book/base_list.ts
```diff
@@ -13,146 +13,192 @@
...
+	readonly numOfBuckets: number;
+	readonly bucketSize: number;
 	readonly secret: number;
 	readonly peerType: PEER_TYPE;
 }
 
-export interface CustomPeerInfo {
-	readonly peerInfo: P2PPeerInfo;
-	readonly dateAdded: Date;
+export type Bucket = Map<string, P2PEnhancedPeerInfo>;
```
diego-G: Bucket or Buckets?

elements/lisk-p2p/src/utils/sanitize.ts
```diff
@@ -47,6 +51,20 @@ export const sanitizeInitialPeerInfo = (peerInfo: ProtocolPeerInfo) => ({
 	wsPort: peerInfo.wsPort,
 });
 
+export const sanitizeInternalPeerInfo = (
```
diego-G: `sanitizeInternalPeerInfo` -> `sanitizeEnhancedPeerInfo` or the other way around. Right now, I don't clearly understand the difference in between `internalState` and `enhancedInfo`. Shall we merge them?

elements/lisk-p2p/src/p2p.ts
mitsuaki-u: The usage of `try catch` here is not correct. It is easier to understand by simply checking if peer exists in peerBook.

diego-G: Then you check twice...

elements/lisk-p2p/src/p2p_types.ts
```diff
@@ -65,6 +65,13 @@ export interface P2PPeerInfo {
 	readonly internalState?: P2PInternalState;
 }
 
+export type P2PEnhancedPeerInfo = {
+	readonly dateAdded?: Date;
```
shuse2: does this all needs to be optional? can we initialize it initially?

elements/lisk-p2p/src/peer_book/base_list.ts
mitsuaki-u: This type only represents one `Bucket`.

elements/lisk-p2p/src/peer_book/base_list.ts
```diff
@@ -13,146 +13,192 @@
...
+		sourceAddress?: string,
+	): BucketInfo {
 		const bucketId = getBucketId({
 			secret: this.peerListConfig.secret,
 			peerType: this.peerListConfig.peerType,
-			targetAddress: ipAddress,
-			bucketCount: this.peerListConfig.peerBucketCount,
+			targetAddress,
+			sourceAddress:
+				this.type === PEER_TYPE.NEW_PEER ? sourceAddress : undefined,
```
shuse2: if it's tried peer we don't use source address?

elements/lisk-p2p/src/peer_book/base_list.ts
mitsuaki-u: This is the return type to `calculateBucket` and is required to get `bucket` based on only a `peerId`. See `getBucket` function below.

elements/lisk-p2p/src/peer_book/base_list.ts
```diff
@@ -13,146 +13,192 @@
...
+		const { bucketId, bucket } = this.calculateBucket(
+			incomingPeerInfo.ipAddress,
+			this.type === PEER_TYPE.NEW_PEER
+				? incomingPeerInfo.sourceAddress
+				: undefined,
+		);
+
+		// If bucket is full, evict a peer to make space for incoming peer
+		const evictedPeer =
+			bucket.size === this.peerListConfig.bucketSize
```
shuse2: maybe it's better to compare with `bucket.size >= this.peerListConfig.bucketSize` so that even if it overflow, it will resolve?

elements/lisk-p2p/src/peer_book/base_list.ts
```diff
@@ -13,146 +13,192 @@
...
 
-	public removePeer(peerInfo: P2PPeerInfo): boolean {
-		const bucket = this.getBucket(peerInfo.ipAddress);
-		const incomingPeerId = peerInfo.peerId;
+	public removePeer(incomingPeerInfo: P2PPeerInfo): boolean {
+		const bucket = this.getBucket(incomingPeerInfo.peerId);
 
-		if (bucket.get(incomingPeerId)) {
-			const result = bucket.delete(incomingPeerId);
+		if (bucket && bucket.get(incomingPeerInfo.peerId)) {
```
shuse2: I think you could do `bucket?.get(incomingPeerInfo.peerId)`

elements/lisk-p2p/src/peer_book/base_list.ts
```diff
@@ -13,146 +13,192 @@
...
+			const removedFromBucket = bucket.delete(incomingPeerInfo.peerId);
+			const removedFromPeerLookup = this.peerIdToPeerInfo.delete(
+				incomingPeerInfo.peerId,
+			);
 
-			return result;
+			return removedFromBucket && removedFromPeerLookup;
```
shuse2: shouldn't this be `||` ? although as long as it's consistent, it shouldn't matter tho

elements/lisk-p2p/src/p2p.ts
mitsuaki-u: You only check once if it exists. The overhead of a single `if` check is worth it if you compare the ease of readability.

elements/lisk-p2p/src/peer_book/base_list.ts
mitsuaki-u: Yes only `new peer` uses source address.

elements/lisk-p2p/src/peer_book/base_list.ts
mitsuaki-u: I think it should be `&&` to enforce consistency. Technically it should never happen but maybe strict is better?

elements/lisk-p2p/src/p2p.ts
diego-G: Then we need to follow the same approach everywhere. There are other places where we proceed as I've suggested.

mitsuaki-u: I added `peerBook.hasPeer` for faster checking and applied this approach for the existing `try catch` related to `peerBook.addPeer`.

elements/lisk-p2p/src/p2p_types.ts
mitsuaki-u: It's optional because some of the properties such as `sourceAddress` are initialized in base peer class during fetching and others (e.g. `dateAdded`, `numOfConnectionFailures`) cannot initialized until it `peerInfo` reaches peerBook.

elements/lisk-p2p/src/peer_book/base_list.ts
```diff
@@ -13,146 +13,195 @@
...
+
+export interface BucketInfo {
+	readonly bucketId: number;
+	readonly bucket: Bucket;
 }
 
-// Base list class is covering a basic peer list that has all the functionality to handle buckets with default eviction strategy
 export class BaseList {
-	protected peerMap: Map<number, Map<string, CustomPeerInfo>>;
+	protected bucketIdToBucket: Map<number, Bucket>;
```
ishantiw: Prefer `bucketIdToPeersMap`? Looks a bit weird with bucket word used 2 times

elements/lisk-p2p/src/peer_book/base_list.ts
```diff
@@ -13,146 +13,195 @@
...
+	readonly numOfBuckets: number;
+	readonly bucketSize: number;
 	readonly secret: number;
 	readonly peerType: PEER_TYPE;
 }
 
-export interface CustomPeerInfo {
-	readonly peerInfo: P2PPeerInfo;
-	readonly dateAdded: Date;
+export type Bucket = Map<string, P2PEnhancedPeerInfo>;
```
ishantiw: `PeersBucket`?

elements/lisk-p2p/src/peer_book/base_list.ts
```diff
@@ -13,146 +13,195 @@
...
 
-// Base list class is covering a basic peer list that has all the functionality to handle buckets with default eviction strategy
 export class BaseList {
-	protected peerMap: Map<number, Map<string, CustomPeerInfo>>;
+	protected bucketIdToBucket: Map<number, Bucket>;
+	/* 
+		Auxilliary map for direct peerId => peerInfo lookups
+		Required because peerLists may be provided by discrete sources
+	*/
+	protected peerIdToPeerInfo: Map<string, P2PEnhancedPeerInfo>;
```
ishantiw: `peerLookUp`?

elements/lisk-p2p/src/peer_book/base_list.ts
mitsuaki-u: It is named exactly what it is: a map of `bucketId` to `Bucket` type.

elements/lisk-p2p/src/peer_book/base_list.ts
mitsuaki-u: Prefer `bucket` as `Peers` part would be redundant in a `p2p` bucket.

elements/lisk-p2p/src/peer_book/base_list.ts
mitsuaki-u: I prefer to say exactly what it is for naming.

elements/lisk-p2p/src/peer_book/base_list.ts
```diff
@@ -13,146 +13,195 @@
...
 		return peerListMap;
 	}
 
-	public initPeerInfo = (peerInfo: P2PPeerInfo): CustomPeerInfo => ({
-		peerInfo,
-		dateAdded: new Date(),
-	});
-
-	public getBucket(ipAddress: string): Map<string, CustomPeerInfo> {
+	public calculateBucket(
```
diego-G: Please, organise the methods related to buckets altogether after the ones related to peers for example.

elements/lisk-p2p/src/p2p.ts
```diff
@@ -956,13 +918,14 @@ export class P2P extends EventEmitter {
 					return;
 				}
 
-				try {
-					this._peerBook.addPeer(incomingPeerInfo);
-				} catch (error) {
-					if (!(error instanceof ExistingPeerError)) {
-						throw error;
-					}
+				if (this._peerBook.hasPeer(incomingPeerInfo)) {
```
diego-G: You can check the opposite and avoid the empty return.

elements/lisk-p2p/src/p2p.ts
```diff
@@ -303,26 +304,13 @@ export class P2P extends EventEmitter {
 			this.emit(EVENT_MESSAGE_RECEIVED, message);
 		};
 
-		this._handleOutboundPeerConnect = (peerInfo: P2PPeerInfo) => {
-			try {
+		this._handleOutboundPeerConnect = (peerInfo: P2PEnhancedPeerInfo) => {
```
diego-G: I think you don't need to change the type here. Tests pass using `P2PPeerInfo`.

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -358,17 +359,22 @@ export class Peer extends EventEmitter {
 		);
 	}
 
-	public async fetchPeers(): Promise<ReadonlyArray<P2PPeerInfo>> {
+	public async fetchPeers(): Promise<ReadonlyArray<P2PEnhancedPeerInfo>> {
```
diego-G: Same as above.

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -386,7 +392,7 @@ export class Peer extends EventEmitter {
 		}
 	}
 
-	public async discoverPeers(): Promise<ReadonlyArray<P2PPeerInfo>> {
+	public async discoverPeers(): Promise<ReadonlyArray<P2PEnhancedPeerInfo>> {
```
diego-G: Same as above.

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -358,17 +359,22 @@ export class Peer extends EventEmitter {
...
 			});
 
-			return validatePeerInfoList(
+			const validatedPeers = validatePeerInfoList(
 				response.data,
 				this._peerConfig.maxPeerDiscoveryResponseLength,
 				this._peerConfig.maxPeerInfoSize,
 			);
+
+			return validatedPeers.map(peerInfo => ({
```
diego-G: This looks like a sanitation of `P2PEnhancedPeerInfo`, isn't it? Shall we use `sanitizeEnhancedPeerInfo()` util?

mitsuaki-u: In this case it is adding `sourceAddress`. `sanitizeEnhancedPeerInfo` is to remove `sourceAddress` and other internal fields to peerBook.

PR #4554, opened 2019-11-22
wait_for helper not updated when changing status endpoint - Closes #4553

### LINKED ISSUE #4553: wait_for helper not updated when changing status endpoint
### Expected behavior

wait_for works correctly for tests

### Actual behavior

wait_for is using the property `res.data.loaded` from calling `/api/node/status` but that property was removed by #4544 

### Steps to reproduce

try to run `npm run mocha:functional:post test/mocha/functional/http/post/0.transfer.js` locally

pablitovicente:
### What was the problem?

`wait_for` not updated when removing properties from `/api/node/status`

### How did I solve it?

by using a different property in `wait_for` to check for the HTTP API

### Review checklist

- [ ] The PR resolves #4553

### DIFF
--- a/framework/test/mocha/common/utils/wait_for.js
+++ b/framework/test/mocha/common/utils/wait_for.js
@@ -58,7 +58,7 @@ function blockchainReady(retries, timeout, baseUrl, doNotLogRetries, cb) {
 			.then(res => {
 				retries -= 1;
 				res = JSON.parse(res.body);
-				if (!res.data.loaded && retries >= 0) {
+				if (res.data.syncing && retries >= 0) {
 					if (!doNotLogRetries) {
 						__testContext.debug(
 							`Retrying ${totalRetries -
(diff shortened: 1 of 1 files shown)


framework/test/mocha/common/utils/wait_for.js
```diff
@@ -58,7 +58,7 @@ function blockchainReady(retries, timeout, baseUrl, doNotLogRetries, cb) {
 			.then(res => {
 				retries -= 1;
 				res = JSON.parse(res.body);
-				if (!res.data.loaded && retries >= 0) {
+				if (res.data.syncing && retries >= 0) {
```
shuse2: i think it should check like `res.data.syncing === undefined || res.data.syncing === null`?

pablitovicente: I think this value in the context of tests is always false and never undefined or null

Issue #4549, opened 2019-11-22
Network module use deprecated P2P config values


valamidev:
### Expected behavior
- this.options.address should change the `p2pConfig.hostIp`
- `peerDiscoveryResponseLength` and `maxPeerDiscoveryResponseLength` hold the same configuration and used mixed inside the P2P library.
- `emitPeerLimit` were renamed to `sendPeerLimit`
- Deprecated configurations: `wsEngine` , `discoveryInterval`
### Actual behavior
-  `p2pConfig.hostAddress` which is a un-used variable in P2P, cannot change the listen IP.
### Steps to reproduce
- Change this.options.address has no effect in the P2P library and fall back to 0.0.0.0.
### Which version(s) does this affect? (Environment, OS, etc...)
3.0

valamidev:
Solved in #4556

Issue #4503, opened 2019-11-14
Revisit Peer discovery threshold test


shuse2:
### Expected behavior
Test should not be unstable.

### Actual behavior
Test is unstable, and locally it fails more than 50%.
Also the test is checking the `newPeers`, but the `MINIMUM_PEER_DISCOVERY_THRESHOLD`, should check the result of getPeers.
Same for `MAX_PEER_DISCOVERY_RESPONSE_LENGTH`, it should check the getPeers response. 

### Steps to reproduce
Run peer discovery threshold test.

valamidev:
Functionality is already covered with[ test/unit/peer_book/peer_book.ts](https://github.com/LiskHQ/lisk-sdk/blob/development/elements/lisk-p2p/test/unit/peer_book/peer_book.ts#L365) and edge cases are covered with integration test.

valamidev:
Solved in #4556

PR #4556, opened 2019-11-25
P2P rediscovery seedPeers interval - Closes #4350 & #4549 & #4503

### LINKED ISSUE #4350: Discover peers through seed peers when not enough outbound connections
### Description

Currently, we don't have a mechanism to fill numOfOutboundConnections slots when they are less than the configured value. 

### Motivation

According to [LIP0004](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0004.md), we should check every 30 seconds if the number of outbound connections is lower than `numOfOutboundConnections` and should initiate the discovery using seed peers.

### Acceptance Criteria

Should trigger discovery when the number of outbound connections is lower than configured `numOfOutboundConnections`

valamidev:
### What was the problem?
- Implementing from LIP-0004 `If the node cannot successfully establish min(numOutConnections,20) outgoing connections within 30 seconds, then it also falls back and queries the hardcoded list of peers with a getPeers message and again tries to establish numOutConnections outgoing connection.`

### How did I solve it?
- Add `fallbackSeedPeerDiscoveryInterval` / `DEFAULT_FALLBACK_SEED_PEER_DISCOVERY_INTERVAL = 30000;` configuration into P2P library.  This time based condition allow us to execute every 30 second the [discovery process](https://github.com/LiskHQ/lisk-sdk/blob/development/elements/lisk-p2p/src/peer_pool.ts#L407), this process allow to send getPeers() request to connected seedPeers or initialise new seedPeer connections.

### How to manually test it?
- `npm run t`

### Review checklist

- [x] The PR resolves #4350 & #4549 & #4503


elements/lisk-p2p/test/integration/discovery.ts
```diff
@@ -261,4 +263,46 @@ describe('Network discovery', () => {
...
+			await Promise.all(p2pNodeList.map(p2p => p2p.start()));
+
+			await wait(1000);
+		});
+
+		afterEach(async () => {
+			await destroyNetwork(p2pNodeList);
+		});
+
+		it(`should recreive getPeers multiple times`, async () => {
```
diego-G: Typo

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -405,18 +405,28 @@ export class PeerPool extends EventEmitter {
...
 			this._peerLists.seedPeers.slice(0, openOutboundSlots),
 		);
 
-		seedPeersForConnection.forEach(peer => {
-			this._addOutboundPeer(peer, this._nodeInfo as P2PNodeInfo);
+		seedPeersForDiscovery.forEach(peer => {
+			// LIP-0004 re-discovery SeedPeers when Outboundconnection < maxOutboundconnections
+			const seedPeer = this.getPeer(peer.peerId);
+			if (seedPeer) {
+				setImmediate(
```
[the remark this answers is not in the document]
ishantiw: or use,
```typescript
(async() => {
    try {
         await seedPeer.discoverPeers();
    } catch (err) {
        throw err;
    }
})()
```

framework/src/modules/network/defaults/config.js
```diff
@@ -152,26 +143,7 @@ const defaultConfig = {
 	required: ['wsPort', 'seedPeers'],
 	default: {
 		wsPort: 5000,
```
ishantiw: We can keep other defaults as well?

framework/src/modules/network/defaults/config.js
```diff
@@ -122,15 +116,11 @@ const defaultConfig = {
 		maxInboundConnections: {
 			type: 'integer',
 		},
-		emitPeerLimit: {
+		sendPeerLimit: {
```
ishantiw: please change it in `config_devnet.json`

framework/src/modules/network/defaults/config.js
valamidev: Without typesafety it will always lead the same problem that these defaults are became obsolete or different from P2P defaults.

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -408,19 +408,32 @@ export class PeerPool extends EventEmitter {
...
+		seedPeersForDiscovery.forEach(peer => {
+			// LIP-0004 re-discovery SeedPeers when Outboundconnection < maxOutboundconnections
+			const seedPeer = this.getPeer(peer.peerId);
+			if (seedPeer) {
+				// tslint:disable-next-line: no-floating-promises
+				(async () => {
+					try {
+						await seedPeer.discoverPeers();
+					} catch (err) {
+						throw err;
```
mitsuaki-u: I think you don't need this `try catch` block if you only re-throw the error..

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -408,19 +408,32 @@ export class PeerPool extends EventEmitter {
...
+					}
+				})();
+			} else {
```
mitsuaki-u: Prefer adding outbound peer and returning early if `!seedPeer` to avoid unnecessary `else`

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -409,32 +409,32 @@ export class PeerPool extends EventEmitter {
 	}
 
 	public discoverFromSeedPeers(): void {
-		const openOutboundSlots = this.getAvailableOutboundConnectionSlots();
+		const freeOutboundSlots = this.getAvailableOutboundConnectionSlots();
```
diego-G: What do you think about shortening the name to `getFreeOutboundSlots()`?

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -408,18 +408,31 @@ export class PeerPool extends EventEmitter {
...
 
-		const seedPeersForConnection = shuffle(
-			this._peerLists.seedPeers.slice(0, openOutboundSlots),
+		// Looking after existing seed peer connection(s)
+		this._peerLists.seedPeers.forEach(peer => {
+			const isConnectedSeedPeer = this.getPeer(peer.peerId);
+			if (isConnectedSeedPeer) {
+				// tslint:disable-next-line: no-floating-promises
+				(async () => {
+					await isConnectedSeedPeer.discoverPeers();
```
shuse2: if this throws, it would be unhandled rejection, it should be try-catched and ignored or emit error event

framework/src/modules/network/defaults/config.js
```diff
@@ -141,6 +131,7 @@ const defaultConfig = {
 		},
 		wsMaxPayload: {
 			type: 'integer',
+			maximum: 3048576,
```
shuse2: why do we need to enforce maximum here?

valamidev: Exceed this limit will cause that you will be banned from the network.

elements/lisk-p2p/src/p2p.ts
```diff
@@ -886,6 +896,16 @@ export class P2P extends EventEmitter {
 				this._peerBook.newPeers,
 				this._peerBook.triedPeers,
 			);
+
+			// LIP-0004 re-discovery SeedPeers when Outboundconnection < maxOutboundconnections
+			if (
+				this._nextSeedPeerDiscovery < Date.now() &&
+				this._peerPool.getFreeOutboundSlots() > 0
```
ishantiw: I would prefer `getAvailableOutboundSlots` but its okay now 😅

framework/src/modules/network/defaults/config.js
ishantiw: I know there is no type-safety there but still, it is good to have a list of config options to look at. If someone wants to change it in the future, they will have to check available p2p-lib config options instead of changing the value straight away in this config. Even if someone gives a wrong field name, it will not have any effect on the working of the library. But yes, it is also important to be consistent with the library when we change the config the options. Let's deal with this later.

PR #4558, opened 2019-11-25
Fix not to evict the fixed peer - Closes #4352

### LINKED ISSUE #4352: Incoming connections from fixed peers should never be closed similar to whitelist
### Expected behavior

Incoming connections from fixed peers should never be closed similar to whitelist.

### Actual behavior

Fixed peers incoming connections could be dropped when there is an incoming connection through the eviction process

### Steps to reproduce

Assign a `fixedPeer` and then allow it to connect (incoming) to you first. Check if it's getting evicted at any point.

`3.0`

shuse2:
### What was the problem?
Fixed peer had possibility to be evicted

### How did I solve it?
From the eviction, remove filter out the fixed peer before evicting

### How to manually test it?
Defined the fixed peer, and observe it's not been disconnected from inbound

### Review checklist

- [ ] The PR resolves #4352 

### DIFF
--- a/elements/lisk-p2p/test/unit/peer_pool.ts
+++ b/elements/lisk-p2p/test/unit/peer_pool.ts
@@ -1028,13 +1028,18 @@ describe('peerPool', () => {
 		beforeEach(async () => {
 			(peerPool as any)._peerLists.whitelisted = whitelistedPeers;
 			(peerPool as any)._peerLists.fixedPeers = fixedPeers;
+			(peerPool as any)._peerPoolConfig = {
+				netgroupProtectionRatio: 0,
+				latencyProtectionRatio: 0,
+				productivityProtectionRatio: 0,
+			};
 			sandbox.stub(peerPool as any, 'getPeers').returns(defaultPeers);
 			sandbox.stub(peerPool, 'removePeer');
 		});
 
 		it('should not evict whitelisted peer', async () => {
 			(peerPool as any)._evictPeer(InboundPeer);
-			expect(peerPool.removePeer).not.to.be.calledWith(
+			expect(peerPool.removePeer).not.to.be.calledWithExactly(
 				whitelistedPeers[0].ipAddress,
 				sinon.match.any,
 				sinon.match.any,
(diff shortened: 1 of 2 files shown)


elements/lisk-p2p/test/unit/peer_pool.ts
```diff
@@ -1006,7 +1007,49 @@ describe('peerPool', () => {
...
+		];
+
+		beforeEach(async () => {
+			(peerPool as any)._peerLists.whitelisted = whitelistedPeers;
+			(peerPool as any)._peerLists.fixedPeers = fixedPeers;
+			sandbox.stub(peerPool as any, 'getPeers').returns(defaultPeers);
+			sandbox.stub(peerPool, 'removePeer');
+		});
+
+		it('should not evict whitelisted peer', async () => {
```
diego-G: Can we add another case to check a positive eviction?

elements/lisk-p2p/test/unit/peer_pool.ts
```diff
@@ -1006,7 +1007,49 @@ describe('peerPool', () => {
...
+				sinon.match.any,
+				sinon.match.any,
+			);
+		});
+
+		it('should not evict fixed peer', async () => {
+			(peerPool as any)._evictPeer(InboundPeer);
+			expect(peerPool.removePeer).not.to.be.calledWith(
+				fixedPeers[0].ipAddress,
+				sinon.match.any,
```
diego-G: We're not using this anywhere else. Is it necessary?

shuse2: for this test case, the second and third parameter is not important, so that's why it's using any

diego-G: You can also test it without because you're not utilising `calledWithExactly`.

shuse2: Updated to use called with exactly 😉

elements/lisk-p2p/test/unit/peer_pool.ts
shuse2: Added positive case too

PR #4560, opened 2019-11-25
Fix logic not try to add to peer book if it already exist - Closes #4362

### LINKED ISSUE #4362: Adding an incoming peer even when it is a duplicate connection
### Expected behavior

Should not add peer to peer list when we already have a duplicate connection. 

### Actual behavior

Even after we emit duplicate connection error we try to add peer to the peer list.

`3.0`

shuse2:
### What was the problem?
When inbound socket already exists in the pool, it was still trying to add to the peer book although it will be rejected later.

### How did I solve it?
Update the logic to return if already exists

### How to manually test it?
All test should pass and work as previously

### Review checklist

- [ ] The PR resolves #4362 


elements/lisk-p2p/src/p2p.ts
```diff
@@ -823,12 +823,13 @@ export class P2P extends EventEmitter {
...
-					this.emit(EVENT_NEW_INBOUND_PEER, incomingPeerInfo);
+
+					return;
 				}
 
 				try {
+					this._peerPool.addInboundPeer(incomingPeerInfo, socket);
+					this.emit(EVENT_NEW_INBOUND_PEER, incomingPeerInfo);
 					this._peerBook.addPeer(incomingPeerInfo);
 				} catch (error) {
```
diego-G: Shall we use the same logic of throwing an error when adding a peer to the pool as we do in the peer book?

shuse2: what do you mean by using the same logic as peer book?
I think this logic shouldn't throw error on the ExistingPeerError, which might happen on the peer book, and if anything else happen, the application will throw unhandle error?

ishantiw: We can make it consistent with `addPeer` approach of peerbook.
we don't have to check for `const existingPeer = this._peerPool.getPeer(peerId);` if handle it separately.

```typescript
try {
	this._peerPool.addInboundPeer(incomingPeerInfo, socket);
	this.emit(EVENT_NEW_INBOUND_PEER, incomingPeerInfo);
} catch (err) {
         this._disconnectSocketDueToFailedHandshake(
         socket,
	 DUPLICATE_CONNECTION,
	 DUPLICATE_CONNECTION_REASON);

        return;
}

try {
	this._peerBook.addPeer(incomingPeerInfo);
} catch (error) {
        if (!(error instanceof ExistingPeerError)) {
	     throw error;
	}
}
```

ishantiw:
We have to check if it's present in the `PeerMap` and throw the error before we create the `Inbound` object which is setting some timeouts that are not cleared and hence the tests are stuck. Earlier we had this check in P2P but now we are relying on `AddInboundPeer` function to check and throw the error.

https://github.com/LiskHQ/lisk-sdk/blob/b597070231ae76ffee10282400f8195411a799db/elements/lisk-p2p/src/peer_pool.ts#L464-#L477

So please move this check on top of `AddInboundPeer` function
```typescript
// Throw an error because adding a peer multiple times is a common developer error which is very difficult to identify and debug.
		if (this._peerMap.has(peer.id)) {
			throw new Error(`Peer ${peer.id} was already in the peer pool`);
		}
```

PR #4568, opened 2019-11-27
Add missing unit test for transactions library - Closes #3331

### LINKED ISSUE #3331: Add missing unit test for transactions library
### Expected behavior
All the newly created function should be covered by the unit test.
Also, all of the transaction above type 8 have to use the protocol spec to ensure the validity

### Actual behavior
There are missing unit tests

2.1.0-alpha.2

pablitovicente:
#What was the problem?

- Some statements were not covered by unit tests
- Some tests were not using protocol-spec generated fixtures

### How did I solve it?

- by adding extra tests
- by using protocol-spec generated fixtures to existing tests

### How to manually test it?

Build should be green

### Review checklist

- [x] The PR resolves #3331


elements/lisk-transactions/test/10_delegate_transaction.ts
```diff
@@ -45,7 +56,7 @@ describe('Delegate registration transaction class', () => {
 		});
 
 		it('should set the delegate asset', async () => {
-			expect(validTestTransaction.asset.username).to.eql('0x0');
+			expect(validTestTransaction.asset.username).to.eql('new_delegate');
```
ManuGowda: expect(validTestTransaction.asset.username).to.eql(protocolSpecDelegateFixture.testCases.input.account.transaction.asset.username); ?
more accurate expectation?

elements/lisk-transactions/test/10_delegate_transaction.ts
```diff
@@ -31,7 +34,15 @@ describe('Delegate registration transaction class', () => {
 	let storeAccountFindStub: sinon.SinonStub;
 
 	beforeEach(async () => {
-		validTestTransaction = new DelegateTransaction(validDelegateTransaction);
+		validTestTransaction = new DelegateTransaction({
+			...validDelegateTransaction,
+			networkIdentifier:
+				protocolSpecDelegateFixture.testCases.input.networkIdentifier,
```
ManuGowda: in general, to avoid this nest object extraction you can destruct in describe 
```
const { networkIdentifier, account, transaction } = protocolSpecDelegateFixture.testCases.input;
```
?

elements/lisk-transactions/test/create_signature_object.ts
```diff
@@ -135,4 +135,32 @@ describe('#createSignatureObject', () => {
...
+					transaction: {} as any,
+					passphrase: (passphrase as unknown) as string,
+					networkIdentifier:
+						multisignatureFixture.testCases.input.networkIdentifier,
+				}),
+			).to.throw('Transaction type is required');
+		});
+	});
+
+	it('should throw an Error if type is invalid', () => {
```
ManuGowda: I think the state `if type is invalid` can be frames as `if type is not registerd`? because of the type valid but not registered in the sdk? I know it's maybe a silly thing, it just caught my mind.

pablitovicente: The error thrown by the library under test is `Invalid transaction type.` and there's no registration concept in it. What you suggest happens in the context of the framework and I do agree in framework there should be a test with that wording. I think this should be kept as is in the context of the package.

elements/lisk-transactions/test/10_delegate_transaction.ts
pablitovicente: Fixed.

elements/lisk-transactions/test/10_delegate_transaction.ts
pablitovicente: Fixed.

elements/lisk-transactions/test/errors.ts
```diff
@@ -52,6 +52,61 @@ describe('errors', () => {
...
+				return expect(TxError.toString()).to.match(/actual: actual property/);
+			});
+
+			it('should show provided expected property when present', () => {
+				TxError = new TransactionError(
+					'error message',
+					'transaction id',
+					'.dataPath',
+					'actual property',
+					'expected property',
```
yatki: Just a general suggestion: I'd recommend using more random property names when testing these kinds of error handlings because somebody can write "expected property" in the error message. Eg. instead of `expected property` use `__EXPECTED_PROPERTY_1__`.

elements/lisk-transactions/test/errors.ts
```diff
@@ -52,6 +52,61 @@ describe('errors', () => {
 					.to.have.property('dataPath')
 					.and.be.a('string');
 			});
+
+			it('should show provided actual property when present', () => {
+				TxError = new TransactionError(
+					'error message',
+					'transaction id',
+					'.dataPath',
+					'actual property',
```
yatki: same below.

elements/lisk-transactions/src/register_delegate.ts
```diff
@@ -48,16 +48,6 @@ export const registerDelegate = (
...
-	if (!username || typeof username !== 'string') {
-		throw new Error('Please provide a username. Expected string.');
-	}
-
-	if (username.length > USERNAME_MAX_LENGTH) {
-		throw new Error(
-			`Username length does not match requirements. Expected to be no more than ${USERNAME_MAX_LENGTH} characters.`,
-		);
-	}
-
```
yatki: I see that this piece of code was removed but related tests were not updated or maybe there were no tests? Also is this change part of this PR? Why are we removing this check?

pablitovicente: It was tested but it was duplicated code as the validate() function already does this. So this was being checked two times.

PR #4576, opened 2019-11-29
Remove state from Network module - Closes #4073

### LINKED ISSUE #4073: When a peer asks us for our peer list, we should not report the state property
### Expected behavior

When a peer asks for our peer list, our node should not report the state property.
This will likely cause a hard fork with 1.x nodes so it should be implemented as a major release.

As for the `getPeerList` endpoint, it should only return ip:port as described in https://github.com/LiskHQ/lips/blob/master/proposals/lip-0004.md#peer-discovery-messages

### Actual behavior

Because old 1.x nodes expect the `state` property, we had to also add it in the new `2.x` versions of the SDK for compatibility reasons.

However, security-wise, it's not a good idea to tell peers about which other peers we are connected to; for this reason we hardcoded the value to always be `2`.

2.x

valamidev:
### What was the problem?
- We still had old usage of connection `state` with values `1,2` in the Network module, this old implementation also bring some unnecessary logic into the Network module, like param filtering which is clearly Http API related logic.

### How did I solve it?
- Replaced old getPeers() / getPeersCount function with getConnectedPeers() and getDisconnectedPeers() which has now equal output with the similar P2P function. 
Http API related filtering is moved under the Http API as it should.

- Output of Http API changed this should be handled in related projects.
```
const PEER_STATE_CONNECTED = 'connected'; 
const PEER_STATE_DISCONNECTED = 'disconnected';
```

### How to manually test it?
- `npm run t`

### Review checklist

- [x] The PR resolves #4073 


framework/src/modules/http_api/helpers/utils.js
```diff
@@ -16,9 +16,146 @@
...
+
+		return { ip: ipAddress, ...peerWithoutIp, state: PEER_STATE_CONNECTED };
+	});
+	const disconnectedList = [...disconnectedPeers].map(peer => {
+		const { ipAddress, options, minVersion, nethash, ...peerWithoutIp } = peer;
+
+		return {
+			ip: ipAddress,
+			...peerWithoutIp,
+			state: PEER_STATE_DISCONNECTED,
```
diego-G: Same here.

framework/src/modules/http_api/helpers/utils.js
```diff
@@ -16,9 +16,146 @@
...
+
+	return filteredPeers;
+};
+
+const consolidatePeers = (connectedPeers = [], disconnectedPeers = []) => {
+	// Assign state 2 to the connected peers
+	const connectedList = [...connectedPeers].map(peer => {
+		const { ipAddress, options, minVersion, nethash, ...peerWithoutIp } = peer;
+
+		return { ip: ipAddress, ...peerWithoutIp, state: PEER_STATE_CONNECTED };
```
diego-G: I guess the state is not necessary from now on.

framework/src/modules/http_api/helpers/utils.js
ishantiw: You mean,
`api/peers?state=true`? -> Shows all the connected peers
`api/peers?state=false`? -> Shows all the disconnected peers
`api/peers?` -> Shows all the peers

or,

`api/peers?connected=true`? -> Shows all the connected peers
`api/peers?connected=false`? -> Shows all the disconnected peers
`api/peers?` -> Shows all the peers

framework/src/modules/http_api/schema/swagger.yml
```diff
@@ -2069,14 +2066,11 @@ definitions:
...
-        example: 2
-        format: int32
-        minimum: 0
-        maximum: 2
+        type: string
+        example: 'connected'
         description: |
           The state of the Peer.
-          Available values: Connected, Disconnected, Banned
+          Available values: Connected, Disconnected
```
ishantiw: values will be in lower case?

PR #4577, opened 2019-11-29
Add P2P condition to select peers for connection - Closed #4351

### LINKED ISSUE #4351: Missing a condition for selecting peers for connection when having low triedPeers
### Description

When we select peers for connection, we choose a ratio of `triedPeers` and `newPeers` but there is a condition missing that when we have triedPeers lower than 100 entries then we should select peers uniformly at random from all known peers (i.e., collection triedPeers and new peers).

### Motivation

According to [LIP0004](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0004.md), we should check the condition when triedPeers are less than 100. This will ensure that we are not choosing a big ratio of triedPeers when there are not enough peers in triedPeers in the list.

### Acceptance Criteria

We can add an integration test that checks if the limit (triedPeers < 100) is passed to the p2p config and condition is triggered when the `triedPeers` are less than 100.

diego-G:
### Review checklist

- [x] The PR resolves #4351


elements/lisk-p2p/src/utils/select.ts
```diff
@@ -127,9 +127,12 @@ export const selectPeersForConnection = (
 	}
 
 	// LIP004 https://github.com/LiskHQ/lips/blob/master/proposals/lip-0004.md#peer-discovery-and-selection
-	const x =
-		input.triedPeers.length / (input.triedPeers.length + input.newPeers.length);
 	const minimumProbability = 0.5;
+	const x =
+		input.triedPeers.length + input.newPeers.length < 100
```
shuse2: in the LIP it says ` If the collection triedPeers contains less than 100 entries, then a peer is selected uniformly at random from all known peers`

ishantiw: Oh yeah, it should be only `triedPeers` < 100

PR #4583, opened 2019-12-03
Use `peerId` instead of constructing it on Block Synchronization Mechanism - Closes #4580

### LINKED ISSUE #4580: bestPeer.id returns `undefined:4000`
### Expected behavior
Expect `BlockSynchronizationMechanism.run` to use `bestPeer.peerId` instead of `bestPeer.id`

### Actual behavior
`_computeBestPeer` adds the property `id` to the object `bestPeer` as `${bestPeer.ip}:${bestPeer.wsPort}` but `bestPeer.ip` is `undefined` because P2P now return `ipAddress` instead of `ip`

### Steps to reproduce
Execute `BlockSynchronizationMechanism._computeBestPeer`
It will return `undefined:4000`

`3.0.0-alpha.2`

limiaspasdaniel:
### What was the problem?
Refer to #4580 
### How did I solve it?
Use `peerId` property now available on peer objects instead of computing it manually.

I also improved some logs and updated unit tests
### How to manually test it?
Run `block_synchronization_mechanism.spec.js`
### Review checklist

- [ ] The PR resolves #4580 


framework/test/jest/unit/specs/modules/chain/synchronizer/block_synchronization_mechanism/peers.js
```diff
@@ -26,93 +26,81 @@ module.exports = {
 			blockVersion: 2,
 			maxHeightPrevoted: 1,
 			height: 66,
-			ip: '127.0.0.2',
-			wsPort: '5000',
```
lsilvs: Is this supposed to be a `P2PPeerInfo`?
If so, would be better to just replace `ip` for `ipAddress` and leave `wsPort`

REF.: https://github.com/LiskHQ/lisk-sdk/blob/development/elements/lisk-p2p/src/p2p_types.ts#L73-L80

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -396,28 +404,32 @@ class BlockSynchronizationMechanism extends BaseSynchronizer {
...
-			const blockIds = (await this.storage.entities.Block.get(
-				{
-					height_in: heightList,
-				},
-				{
-					sort: 'height:asc',
-					limit: heightList.length,
-				},
-			)).map(block => block.id);
+			const blockIds = (
```
ManuGowda: looks like the same formatting issue? if its fixing the formatting please ignore this comment.

limiaspasdaniel: Yeah probably prettier got updated, it's fine.

PR #4584, opened 2019-12-03
Prevent application crashing if there is a fork choice rule violation - Closes #4579

### LINKED ISSUE #3992: Update & Enable eslint-plugin-jest
### Expected behavior
With eslint-plugin-jest v22.12.0+ valid-expect rule verifies async assertions, which previously was a big pain since it was so easy to misuse them.

See: https://github.com/jest-community/eslint-plugin-jest/blob/master/docs/rules/valid-expect.md
### Actual behavior
We don't have a rule makes sure async assertions properly awaited/returned.

### LINKED ISSUE #4579: Violation of fork choice rule error
### Expected behavior
Expect to start sync mechanism when a node with only genesis block receive a valid block from the tip of the chain

### Actual behavior
It fails with the message `Violation of fork choice rule`

REF.: https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js#L601

### Steps to reproduce
Send a valid block that will trigger the Block Synchronization Mechanism to a node with only genesis block

`3.0.0-alpha.2`

limiaspasdaniel:
### What was the problem?
Refer to #4579 
### How did I solve it?
Throwing an `AbortError` which is caught within the Synchronizer scope so the synchronizer aborts peacefully
### How to manually test it?
Run `synchronizer.spec.js` and `block_synchronization_mechanism.spec.js` tests with `Jest`
### Review checklist

- [ ] The PR resolves #4579 


framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -595,10 +600,12 @@ class BlockSynchronizationMechanism extends BaseSynchronizer {
 
 		const forkStatus = await this.processorModule.forkStatus(peersTip);
 
-		const inDifferentChain = forkStatus === FORK_STATUS_DIFFERENT_CHAIN;
+		const tipHasPreference = forkStatus === FORK_STATUS_DIFFERENT_CHAIN;
```
lsilvs: ❤️

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -78,6 +79,14 @@ class BlockSynchronizationMechanism extends BaseSynchronizer {
 					block: receivedBlock,
 				});
 			}
+
+			if (error instanceof AbortError) {
+				return this.logger.info(
+					{ error, reason: error.reason },
+					`Aborting synchronization mechanism with reason: ${error.reason}`,
```
ManuGowda: Since we already have reason as part of json we do not need it again, maybe change it to `Aborting synchronization mechanism!!`?

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
```diff
@@ -396,28 +405,32 @@ class BlockSynchronizationMechanism extends BaseSynchronizer {
...
-			const blockIds = (await this.storage.entities.Block.get(
-				{
-					height_in: heightList,
-				},
-				{
-					sort: 'height:asc',
-					limit: heightList.length,
-				},
-			)).map(block => block.id);
+			const blockIds = (
```
ManuGowda: is it because of formatting? can you fix it if its not a formatting issue?

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
limiaspasdaniel: Done :)

framework/src/modules/chain/synchronizer/block_synchronization_mechanism.js
limiaspasdaniel: It's just prettier doing its things :D

Issue #4598, opened 2019-12-04
Chain module events miss direction


JesusTheHun:
When I subscribe to the chain module for an event such as `chain:transactions:confirmed:change`, I received the transactions as an argument, but I don't know if those transactions are newly appended or on the contrary have been _poped_.

I suggest we either add a second parameter to the event to state "appended" or "poped" (I don't like this option but it's a solution), or we add two new events to better describe what happened.

This is required to have a reliable view of blockchain evolution, otherwise we have to store a block height on the listening part, and that's dirty work.

I'm writing a PR for this.

shuse2:
I agree that this is definitely an issue, and it is raised https://github.com/LiskHQ/lisk-sdk/issues/3863 too
In order to tackle this problem, we want to have consistent approach throughout the chain module.
Since this would be big topic, I would suggest to post your idea on #3863 

I think main events that everyone would be interested
- addition/deletion of block
- update on accounts
- update on chain state (delegate list, finality etc)
- addition/deletion of transactions from the pool

If we can define the consistent naming for these events, which can be extended for the future usecases, that would be great

shuse2:
Closing because this issue is resolved in release/4.0.0

PR #4600, opened 2019-12-04
4598 chain module event direction

JesusTheHun:
Implements #4598


JesusTheHun:
I haven't found related tests, if they do exist, please let me know.

PR #4608, opened 2019-12-05
Multi signature protocol specs - Closes #4604

### LINKED ISSUE #4604: Add multi signature protocol specs
### Description

Protocol Specs for changes to multi signature transactions are required

### Motivation

[LIP-0017](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0017.md) introduces changes to multi signature transactions and we need to have Protocol Specs to use for validating the implementation.

### Acceptance Criteria

- Protocol Specs for multi signature added

### Related Issues
#4836

### LINKED ISSUE #4890: Wrong signature serialization in MultiSig Registration
### Expected behavior

When signatures are being serialized they should be prepend with 0x00 or 0x01 based on a signature for a key being present or not.

### Actual behavior

Signatures are not being prepend with 0x00 or 0x01

### Steps to reproduce

Look at protocol spec/LIP/Source code

pablitovicente:
### What was the problem?

No protocol spec existed for changes introduced by LIP 0017

### How did I solve it?

By adding spec generator for multisig transaction

### Review checklist

- [x] The PR resolves #4604
- [x] The PR resolves #4890 


protocol-specs/generators/multisignature/index.js
```diff
@@ -0,0 +1,221 @@
...
+			'ffed38380998a90a2af9501f10182bc2a07922448ab383575b1e34aeddfa5482fa406b6952d377f0278920e3eb8da919e4cf5c68b02eeba5d8b3334fdc0369b6',
+		publicKey:
+			'fa406b6952d377f0278920e3eb8da919e4cf5c68b02eeba5d8b3334fdc0369b6',
+		address: '7086965981385941478L',
+	},
+];
+
+const sortKeysDescending = publicKeys =>
+	publicKeys.sort((publicKeyA, publicKeyB) => {
+		if (publicKeyA < publicKeyB) return 1;
```
shuse2: I think sorting function needs to be done by converting to the big-ending encoded integer
> For two public keys pk1 and pk2, we say that pk1 is larger than pk2, and write pk1 > pk2, if and only if the 256-bit representation of pk1 interpreted as a big-endian encoded integer is larger than the one of pk2.

protocol-specs/generators/multisignature/index.js
```diff
@@ -0,0 +1,221 @@
...
+	// Serialize all signatures
+	registrationTx.signatures.forEach(aSignature => {
+		const signatureBuffer = Buffer.concat([
+			Buffer.alloc(1, aSignature.index),
+			hexToBuffer(aSignature.signature),
+		]);
+		txBuffer = Buffer.concat([txBuffer, signatureBuffer]);
+	});
+	// Sender signs whole transaction
+	const signature = signData(
```
shuse2: I think multi signature doesn't require signature

protocol-specs/generators/multisignature/index.js
```diff
@@ -0,0 +1,221 @@
...
+			account: accounts[1],
+			networkIdentifier,
+			coSigners: [accounts[2], accounts[3], accounts[4], accounts[5]],
+			transaction: registrationTx,
+		},
+		output: signedRegistrationTx,
+	};
+};
+
+generateValidMultisignatureRegistrationTransaction();
```
shuse2: why run here?

protocol-specs/generators/multisignature/index.js
```diff
@@ -0,0 +1,221 @@
...
+	});
+	// Sender signs whole transaction
+	const signature = signData(
+		hash(Buffer.concat([hexToBuffer(networkIdentifier), txBuffer])),
+		accounts[0].passphrase,
+	);
+
+	const signedRegistrationTx = { ...registrationTx, signature };
+
+	const id = getId(Buffer.concat([txBuffer, Buffer.from(signature, 'hex')]));
```
shuse2: If it doesn't require signature, id shouldn't require signature too, but instead it should include `signatures`

protocol-specs/generators/multisignature/index.js
```diff
@@ -0,0 +1,221 @@
...
+	registrationTx.asset.optionalKeys.forEach(aKey => {
+		const account = findAccountByKey(aKey).pop();
+		registrationTx.signatures.push({
+			...createSignatureObject(txBuffer, account),
+			index,
+		});
+		index += 1;
+	});
+	// Serialize all signatures
+	registrationTx.signatures.forEach(aSignature => {
```
shuse2: instead of using `forEach`, i think it will be easier to use `map` or `reduce` to create the `signaturesBuffer`

protocol-specs/generators/multisignature/index.js
pablitovicente: Left by mistake as it was being use for debugging. Removed.

protocol-specs/generators/multisignature/index.js
pablitovicente: Changed to compare using BigInts. Checked with research and the endiandness should not matter in this case as the objective is to have the keys sorted. This LIP is gonna be updated a bit so I think this is going to be one of the corrections.

protocol-specs/generators/multisignature/index.js
pablitovicente: This is for registration of multisig so in this case it does need it. It's not needed for transactions sent from a multisig

shuse2: ah that's true, if the owner account doesn't agree to change to multi sig, it sounds dangerous.

protocol-specs/generators/multisignature/index.js
pablitovicente: This is solved because for registration we need the signature of the account owner.

protocol-specs/generators/multisignature/index.js
pablitovicente: It's a good point. I'm going to try to refactor to be more expressive in general with the whole spec.

pablitovicente: I tried changing to both map and reduce but makes the code look more confusing. Reduce provides both previous and current values which we don't need together. Map makes the assignment of the index more complex as we need the index to be sequential for both mandatory and optional keys.

protocol-specs/generators/multisignature/index.js
```diff
@@ -0,0 +1,433 @@
+/*
+ * Copyright © 2018 Lisk Foundation
```
ManuGowda: 2019

protocol-specs/generators/multisignature/index.js
```diff
@@ -0,0 +1,433 @@
...
+	hexToBuffer,
+	intToBuffer,
+	hash,
+	getFirstEightBytesReversed,
+	bufferToIntAsString,
+	signData,
+} = require('@liskhq/lisk-cryptography');
+const BaseGenerator = require('../base_generator');
+
+const networkIdentifier =
```
ManuGowda: can we define this in config or constants of protocol spec folder and it will be useful to reuse other places as well?

protocol-specs/generators/multisignature/index.js
```diff
@@ -0,0 +1,433 @@
...
+
+const createSignatureObject = (txBuffer, account) => ({
+	signature: signData(
+		hash(Buffer.concat([hexToBuffer(networkIdentifier), txBuffer])),
+		account.passphrase,
+	),
+});
+
+const serializeBasicProperties = tx => {
+	const transactionTimestamp = Buffer.alloc(4);
```
ManuGowda: are we gonna continue using timestamp?

protocol-specs/generators/multisignature/index.js
pablitovicente: I think this might change from spec to spec in the future. Maybe is best to keep isolated on each suite? ping @shuse2 ?

protocol-specs/generators/multisignature/index.js
pablitovicente: In the open PR for multi-signatures it is still there. We can update later if it's being removed.

protocol-specs/generators/multisignature/index.js
pablitovicente: 2020 :)

protocol-specs/generators/multisignature/README.md
```diff
@@ -0,0 +1,20 @@
...
+A set of test generators for updated multisignatures transactions by LIP-0017
+
+### Format
+
+#### Input
+
+- account: sender account
+- transaction: raw transaction
+- network identifier: network identifier used
+- coSigner: accounts that co-signed the transaction
```
AndreasKendziorra: It's not really important, but in some test cases not all entries of `coSigner` are used. Maybe you want to delete them from the JSON?

protocol-specs/generators/multisignature/index.js
shuse2: yes it could differ between spec but also, the test should use the consistent one (doesn't even need to be real), so it should be fine

protocol-specs/generators/multisignature/index.js
```diff
@@ -0,0 +1,436 @@
...
+		publicKey:
+			'fa406b6952d377f0278920e3eb8da919e4cf5c68b02eeba5d8b3334fdc0369b6',
+		address: '7086965981385941478L',
+	},
+};
+
+const sortKeysDescending = publicKeys =>
+	publicKeys.sort((publicKeyA, publicKeyB) => {
+		// eslint-disable-next-line no-undef, new-cap
+		if (BigInt(`0x${publicKeyA}`) > BigInt(`0x${publicKeyB}`)) return 1;
```
shuse2:
```suggestion
		if (BigInt(`0x${publicKeyA}`) > BigInt(`0x${publicKeyB}`)) {
		  return 1;
		}
```

protocol-specs/generators/multisignature/index.js
```diff
@@ -0,0 +1,436 @@
...
+		// eslint-disable-next-line no-undef, new-cap
+		if (BigInt(`0x${publicKeyA}`) < BigInt(`0x${publicKeyB}`)) return -1;
```
shuse2:
```suggestion
		if (BigInt(`0x${publicKeyA}`) < BigInt(`0x${publicKeyB}`)) {
		  return -1;
		}
```

protocol-specs/generators/multisignature/index.js
pablitovicente: Though we need it to match development as we're using them for test right?

protocol-specs/generator_outputs/multisignature_registration_transaction/multisignature_transaction.json
```diff
@@ -0,0 +1,92 @@
+{
```
AndreasKendziorra: you probably want to remove this file now since its a duplicate of  multisignature_registration_transaction.json?

protocol-specs/generators/multisignature/index.js
shuse2: depends on where we use i think. If we use it in the unit test for transactions, it shouldn't matter

protocol-specs/generators/multisignature_registration/index.js
```diff
@@ -0,0 +1,440 @@
...
+	]);
+
+	return buf;
+};
+
+const serializeMemberSignatures = (tx, txBuffer) => {
+	let txBufferCopy = Buffer.alloc(txBuffer.length);
+	txBuffer.copy(txBufferCopy);
+
+	tx.signatures.forEach(aSignature => {
```
AndreasKendziorra: This serialization of the signatures is not as specified in LIP: 0x01 has to be added for each signature and 0x00 for every empty buffer in tx.signatures.

protocol-specs/generators/multisignature_registration/index.js
```diff
@@ -0,0 +1,440 @@
...
+
+	let txBuffer = serializeBasicProperties(tx);
+
+	// Sender signs
+	tx.signatures.push(
+		createSignatureObject(txBuffer, accounts.targetAccount).signature,
+	);
+	// Members sign in order
+	tx.signatures.push(
+		createSignatureObject(txBuffer, accounts.optionalOne).signature,
```
AndreasKendziorra:
```suggestion
		createSignatureObject(txBuffer, accounts.mandatoryTow).signature,
```

protocol-specs/generators/multisignature_registration/index.js
```diff
@@ -0,0 +1,440 @@
...
+	);
+	tx.signatures.push(
+		createSignatureObject(txBuffer, accounts.optionalTwo).signature,
```
AndreasKendziorra:
```suggestion
		createSignatureObject(txBuffer, accounts.mandatoryOne).signature,
```

protocol-specs/generators/multisignature_registration/index.js
pablitovicente: Isn't that type of serialization for other types of transactions originating from a multisignature? This is multisig registration only

protocol-specs/generator_outputs/multisignature_registration_transaction/multisignature_transaction.json
pablitovicente: Done.

protocol-specs/generators/multisignature_registration/index.js
pablitovicente: Had to also fix typo in key name so was not able to commit sorry :)

protocol-specs/generators/multisignature_registration/index.js
pablitovicente: Had to also fix typo in key name so was not able to commit sorry :)

protocol-specs/generators/multisignature/index.js
pablitovicente: Ok maybe lets create an issue to fix this for all specs?

pablitovicente:
> Looks good, but the signature serialization is not as specified.
> And what about the negative test cases? Are they all covered by unit tests, or is planned to add them in a separate PR?

@AndreasKendziorra check my question for the signature serialization please. Negative cases are already covered in implementation and some more are going to be added based on this fixtures.

protocol-specs/generators/multisignature/README.md
pablitovicente: I updated the coSigner to be the correct ones.

protocol-specs/generators/multisignature_registration/index.js
AndreasKendziorra: this serialization has to be applied for every transaction, including transactions from "regular" accounts and multisig registration transactions.
That also means that for transactions from non-multisig accounts the 0x01 has to be prepended to the signature.

pablitovicente: Fixed. Good catch!

pablitovicente:
@AndreasKendziorra could you review again please?

protocol-specs/generators/multisignature_registration/index.js
```diff
@@ -0,0 +1,451 @@
...
+		publicKey:
+			'fa406b6952d377f0278920e3eb8da919e4cf5c68b02eeba5d8b3334fdc0369b6',
+		address: '7086965981385941478L',
+	},
+};
+
+const SIGNATURE_PRESENT = Buffer.from('01', 'hex');
+const SIGNATURE_NOT_PRESENT = Buffer.from('00', 'hex');
+
+const sortKeysDescending = publicKeys =>
```
shuse2: shouldn't this be ascending?

pablitovicente: Name was misleading but sort correct. Updated name.

protocol-specs/generators/multisignature_registration/index.js
```diff
@@ -139,17 +143,30 @@ const serializeMemberSignatures = (tx, txBuffer) => {
 
 	tx.signatures.forEach(aSignature => {
 		const signatureBuffer = Buffer.concat([hexToBuffer(aSignature)]);
-		txBufferCopy = Buffer.concat([txBufferCopy, signatureBuffer]);
+		if (signatureBuffer.lenght === 0) {
+			txBufferCopy = Buffer.concat([
+				txBufferCopy,
```
AndreasKendziorra: Does this really work? I mean having `txBufferCopy` on the left and the right side of the assignment without having `txBufferCopy` initialized before?

pablitovicente: it's initialized above in line 141 :) 

```	
let txBufferCopy = Buffer.alloc(txBuffer.length);
txBuffer.copy(txBufferCopy);
```

AndreasKendziorra: ah, thanks. I overlooked it during the review :)

PR #4615, opened 2019-12-06
Prettier fails to apply format for lisk-sdk - Closes #4614

### LINKED ISSUE #4614: Prettier fails to apply format for lisk-sdk
### Expected behavior
Prettier should be able to apply format at root level and individual packages.
### Actual behavior
The prettier fails apply the format as certain directory and files are not ignored
### Steps to reproduce
- Checkout `development` branch
- Run `npm run format` from root
3.0.0

ManuGowda:
### What was the problem?
The prettier fails apply the format as certain directory and files are not ignored
### How did I solve it?
Fixed the prettierignore and applied outdated prettier changes 
### How to manually test it?
`npm run format`
### Review checklist

- [ ] The PR resolves #4614 

### DIFF
--- a/commander/src/utils/core/database.ts
+++ b/commander/src/utils/core/database.ts
@@ -45,7 +45,9 @@ const isDbRunning = async (
 	installDir: string,
 	port: string,
 ): Promise<boolean> => {
-	const { stderr }: ExecResult = await exec(
+	const {
+		stderr,
+	}: ExecResult = await exec(
 		`${PG_CTL} --pgdata ${DB_DATA} --options '-F -p ${port}' status`,
 		{ cwd: installDir },
 	);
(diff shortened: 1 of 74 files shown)


commander/src/utils/core/database.ts
```diff
@@ -45,7 +45,9 @@ const isDbRunning = async (
 	installDir: string,
 	port: string,
 ): Promise<boolean> => {
-	const { stderr }: ExecResult = await exec(
+	const {
```
pablitovicente: I prefer the other way but if the chosen prettier rules are this format I'm ok

ManuGowda: before it was like this `const { stderr }: ExecResult = await exec(` I think with new prettier update, it might have changed. so it should be ok, as its gonna apply the same to other parts of the code.

elements/lisk-p2p/src/peer_pool.ts
```diff
@@ -475,9 +475,11 @@ export class PeerPool extends EventEmitter {
 			peerLimit,
 		});
 
-		[...peersToConnect, ...disconnectedFixedPeers].forEach(
-			(peerInfo: P2PPeerInfo) =>
-				this._addOutboundPeer(peerInfo, this._nodeInfo as P2PNodeInfo),
+		[
```
pablitovicente: I prefer the previous version but if the chosen prettier rules are set to this I'm ok with it.

PR #4621, opened 2019-12-11
Handle usage of async and await expression - Closes #4619

### LINKED ISSUE #4619: Handle usage of async and await expression
### Expected behavior
Remove `async` expression where it's not required.

Example:
https://github.com/LiskHQ/lisk-sdk/blob/5d5ece5c7306ef918822d64e37ef4d8fa6a2b820/framework/src/modules/chain/blocks/blocks.js#L201
### Actual behavior
`async` function expression is used where the function doesn't have anything related to `await` expression.
### Steps to reproduce
https://github.com/LiskHQ/lisk-sdk/blob/5d5ece5c7306ef918822d64e37ef4d8fa6a2b820/framework/src/modules/chain/blocks/blocks.js#L201
3.0.0

ManuGowda:
### What was the problem?
an `async` function expression is used where the function doesn't have anything related to await.
### How did I solve it?
- Remove unwanted usage if `async` expression
- Add missing `async` expression
### How to manually test it?
Run all the framework test
### Review checklist

- [ ] The PR resolves #4619 


framework/test/mocha/unit/modules/chain/chain.js
```diff
@@ -438,29 +438,6 @@ describe('Chain', () => {
...
-			chain.options.syncing.active = false;
-
-			// Act
-			chain._startLoader();
-
-			// Assert
-			expect(stubs.jobsQueue.register).to.not.be.called;
-		});
-
-		it('should load transactions and signatures', async () => {
```
pablitovicente: Maybe this test needs to be kept and make sure `this.loader.loadUnconfirmedTransactions` is being called?

framework/src/controller/channels/child_process_channel.js
```diff
@@ -175,7 +175,7 @@ class ChildProcessChannel extends BaseChannel {
 		});
 	}
 
-	async cleanup() {
+	cleanup() {
```
nazarhussain: This was kept `async` as pattern to avoid future refactoring. In future we may need to update cleanup logic with async task, then we don't need to update function usage everywhere. Also this pattern should be followed if we or anyone else create different channel.

framework/src/controller/channels/in_memory_channel.js
```diff
@@ -23,9 +23,9 @@ class InMemoryChannel extends BaseChannel {
 		super(moduleAlias, events, actions, options);
 	}
 
-	async registerToBus(bus) {
+	registerToBus(bus) {
```
nazarhussain: This happens same for `registerToBus` function as well. Each channel type should and must implement it `async` to have consistent usage. See the function signature in `BaseChannel`.

The top level code is expecting every channel to have async function implementation for `registerToBus` and we should keep it consistent.

framework/src/controller/bus.js
```diff
@@ -226,7 +226,7 @@ class Bus extends EventEmitter2 {
 		return Object.keys(this.events);
 	}
 
-	async cleanup() {
+	cleanup() {
```
nazarhussain: This function is missed to mention in the `BaseChannel` that we should add. And it should be kept as `async`, not just for this implementation rather for future refactoring and other implementations.

framework/src/controller/controller.js
```diff
@@ -307,7 +307,7 @@ class Controller {
 		this.childrenList.forEach(child => child.kill());
 
 		try {
-			await this.bus.cleanup();
+			this.bus.cleanup();
```
nazarhussain: See the inconsistence pattern here. One place you have `async` cleanup other place you don't. We kept it consistently `async` for this purpose.

framework/src/modules/chain/components/storage/entities/chain_state.js
```diff
@@ -61,7 +61,7 @@ class ChainState extends BaseEntity {
 		});
 	}
 
-	async setKey(key, value, tx) {
+	setKey(key, value, tx) {
```
nazarhussain: That can be applied similar to `getKey` as well.

framework/src/modules/http_api/index.js
```diff
@@ -50,15 +50,15 @@ class HttpAPIModule extends BaseModule {
 		return {};
 	}
 
-	async load(channel) {
+	load(channel) {
```
nazarhussain: These `load` and `unload` methods must be kept as `async` for every module as pattern defined by SDK. Look for `BaseModule` class.

framework/src/modules/http_api/index.js
```diff
@@ -50,15 +50,15 @@ class HttpAPIModule extends BaseModule {
...
 		this.httpApi = new HttpApi(channel, this.options);
 
 		channel.once('app:ready', async () => {
 			await this.httpApi.bootstrap();
 		});
 	}
 
-	async unload() {
+	unload() {
```
nazarhussain: Same as above.

framework/test/mocha/unit/modules/chain/chain.js
nazarhussain: 👍

framework/src/modules/chain/bft/bft.js
```diff
@@ -188,7 +188,7 @@ class BFT extends EventEmitter {
 		return forkChoiceRule.FORK_STATUS_DISCARD;
 	}
 
-	async _initFinalityManager(stateStore) {
+	_initFinalityManager(stateStore) {
```
shuse2: usage of this should be checked too?
same for all other cases

framework/src/modules/chain/components/storage/entities/account.js
```diff
@@ -111,7 +111,7 @@ class ChainAccount extends AccountEntity {
 		return accounts;
 	}
 
-	async update(filters, data, _options, tx) {
+	update(filters, data, _options, tx) {
```
shuse2: I think all the functions which return Promise are better to be `async` since it's more clear that it returns `Promise`

pablitovicente: I get the point but there's some overhead associated with using async vs regular promises so though I get the point but maybe we better keep it when really necessary?

pablitovicente: Although it seems the overhead is about 1% so maybe readability is better in this case...

shuse2: I think that doesn't apply to something that is already returning promise? Or does it apply all the time?

ManuGowda:
@shuse2 @pablitovicente @nazarhussain ok the common understanding is where we are returning promises I will keep the async, just avoid confusion, and remove async where the function is synchronous. @nazarhussain regarding the once which were defined for future I think we can remove them, we can make them `async` when it required. otherwise, the whole codebase can become smelly.

Let me know your views before I make the changes, Danke!

framework/src/modules/chain/components/storage/entities/account.js
pablitovicente: All the time as async wraps promises. But 1% is small enough to be disregarded.

shuse2: Mmm then add long as it's consistent I'm ok for now

framework/test/mocha/unit/modules/chain/chain.js
```diff
@@ -449,14 +449,14 @@ describe('Chain', () => {
...
-			chain._startLoader();
+			await chain._startLoader();
 
 			// Assert
 			expect(stubs.jobsQueue.register).to.not.be.called;
 		});
 
 		it('should load transactions and signatures', async () => {
-			chain._startLoader();
+			await chain._startLoader();
```
pablitovicente: Why did you re-introduce this wrapper?

ManuGowda: the function call couldn't be spied, so I moved it back. and also I checked the whole test suite, for all the channel events where we are subscribing or publish we do not have such tests, so moved it back, maybe we can clean this unwanted private functions when we have a way to test. For this PR I have addressed the missing await for that function.

nazarhussain:
>  once which were defined for future I think we can remove them, we can make them async when it required

In general its right. But I highlighted the particular use cases where the we presented an API or pattern for others to develop further with our SDK. e.g. Custom Modules, Custom Transactions, Entities. So those must be followed by the design defined rather what implementation we have currently in our code base. 

e.g. We say by design that every module `load` and `unload` interfaces will be `async`. So no matter if we are doing only synchronous stuff inside we should keep their declaration `async` so others can read our code base and follow same.

framework/src/modules/chain/chain.js
```diff
@@ -511,7 +511,7 @@ module.exports = class Chain {
 		this.scope.modules.synchronizer = this.synchronizer;
 	}
 
-	_startLoader() {
+	async _startLoader() {
```
shuse2: if this is asyc, the line below shouldve been awaited

framework/src/modules/chain/components/storage/entities/account.js
```diff
@@ -134,7 +134,7 @@ class ChainAccount extends AccountEntity {
 		};
 
 		if (_.isEmpty(objectData)) {
-			return false;
+			return Promise.resolve(false);
```
shuse2: No need to do Promise.resolve, it's automatically wrapped

framework/src/modules/chain/chain.js
ManuGowda: returning the promise now, and waiting in the subscribe, so should be good.

PR #4628, opened 2019-12-16
P2P add protocol message limit and improved message limit penalty

valamidev:
### What was the problem?
- LIP-0004 proposed protocol message limit were missing from the P2P libraries. 
- The current message rate limiting were still allow a 10 second wide flood-attack window. 

### How did I solve it?
- Add `_protocolRCPCounter` and logic to count and prevent P2P protocol messages flooding.
- Extend the messageRate calculation with `messageRateExceedCoeff` which allow to multiple the penalty score for `applyPenalty` this could stop flood-attack in case _immediately_.

### How to manually test it?
`npm run test`

### Review checklist

- [x] The PR resolves

### DIFF
--- a/elements/lisk-p2p/src/constants.ts
+++ b/elements/lisk-p2p/src/constants.ts
@@ -21,6 +21,8 @@ export const DEFAULT_BAN_TIME = 86400000; // Milliseconds in a day -> hours*minu
 export const DEFAULT_POPULATOR_INTERVAL = 10000;
 export const DEFAULT_FALLBACK_SEED_PEER_DISCOVERY_INTERVAL = 30000;
 export const DEFAULT_SEND_PEER_LIMIT = 16;
+// Message types
+export const RCP_REQUEST_TYPE = '/RPCRequest';
 // Max rate of WebSocket messages per second per peer.
 export const DEFAULT_WS_MAX_MESSAGE_RATE = 100;
 export const DEFAULT_WS_MAX_MESSAGE_RATE_PENALTY = 10;
--- a/elements/lisk-p2p/src/constants.ts
+++ b/elements/lisk-p2p/src/constants.ts
@@ -24,7 +24,7 @@ export const DEFAULT_SEND_PEER_LIMIT = 16;
 // Message types
 export const RCP_REQUEST_TYPE = '/RPCRequest';
 // Max rate of WebSocket messages per second per peer.
-export const DEFAULT_WS_MAX_MESSAGE_RATE = 1000;
+export const DEFAULT_WS_MAX_MESSAGE_RATE = 100;
 export const DEFAULT_WS_MAX_MESSAGE_RATE_PENALTY = 10;
 export const DEFAULT_RATE_CALCULATION_INTERVAL = 1000;
 export const DEFAULT_WS_MAX_PAYLOAD = 3048576; // Size in bytes
--- a/elements/lisk-p2p/src/constants.ts
+++ b/elements/lisk-p2p/src/constants.ts
@@ -24,7 +24,7 @@ export const DEFAULT_SEND_PEER_LIMIT = 16;
 // Message types
 export const RCP_REQUEST_TYPE = '/RPCRequest';
 // Max rate of WebSocket messages per second per peer.
-export const DEFAULT_WS_MAX_MESSAGE_RATE = 100;
+export const DEFAULT_WS_MAX_MESSAGE_RATE = 1000;
 export const DEFAULT_WS_MAX_MESSAGE_RATE_PENALTY = 10;
 export const DEFAULT_RATE_CALCULATION_INTERVAL = 1000;
 export const DEFAULT_WS_MAX_PAYLOAD = 3048576; // Size in bytes
--- a/elements/lisk-p2p/src/events.ts
+++ b/elements/lisk-p2p/src/events.ts
@@ -40,6 +40,8 @@ export const EVENT_FAILED_TO_SEND_MESSAGE = 'failedToSendMessage';
 // Peer base
 export const REMOTE_SC_EVENT_RPC_REQUEST = 'rpc-request';
 export const REMOTE_SC_EVENT_MESSAGE = 'remote-message';
+
+// P2P Protocol messages
 export const REMOTE_EVENT_POST_NODE_INFO = 'postNodeInfo';
 export const REMOTE_EVENT_RPC_GET_NODE_INFO = 'getNodeInfo';
 export const REMOTE_EVENT_RPC_GET_PEERS_LIST = 'getPeers';
--- a/elements/lisk-p2p/src/events.ts
+++ b/elements/lisk-p2p/src/events.ts
@@ -49,7 +49,7 @@ export const REMOTE_EVENT_PING = 'ping';
 export const REMOTE_EVENT_PONG = 'pong';
 
 // P2P Protocol messages list
-export const PROTOCOL_REMOTE_EVENT_RCP_EVENTS: Set<string> = new Set([
+export const PROTOCOL_EVENTS_TO_RATE_LIMIT: Set<string> = new Set([
 	REMOTE_EVENT_RPC_GET_NODE_INFO,
 	REMOTE_EVENT_RPC_GET_PEERS_LIST,
 ]);
--- a/elements/lisk-p2p/src/events.ts
+++ b/elements/lisk-p2p/src/events.ts
@@ -48,6 +48,12 @@ export const REMOTE_EVENT_RPC_GET_PEERS_LIST = 'getPeers';
 export const REMOTE_EVENT_PING = 'ping';
 export const REMOTE_EVENT_PONG = 'pong';
 
+// P2P Protocol messages list
+export const PROTOCOL_REMOTE_EVENT_RCP_EVENTS: Set<string> = new Set([
+	REMOTE_EVENT_RPC_GET_NODE_INFO,
+	REMOTE_EVENT_RPC_GET_PEERS_LIST,
+]);
+
 // Inbound peer
 export const EVENT_CLOSE_INBOUND = 'closeInbound';
 export const EVENT_INBOUND_SOCKET_ERROR = 'inboundSocketError';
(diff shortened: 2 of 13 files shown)


elements/lisk-p2p/src/constants.ts
```diff
@@ -21,8 +21,10 @@ export const DEFAULT_BAN_TIME = 86400000; // Milliseconds in a day -> hours*minu
 export const DEFAULT_POPULATOR_INTERVAL = 10000;
 export const DEFAULT_FALLBACK_SEED_PEER_DISCOVERY_INTERVAL = 30000;
 export const DEFAULT_SEND_PEER_LIMIT = 16;
+// Message types
+export const RCP_REQUEST_TYPE = '/RPCRequest';
 // Max rate of WebSocket messages per second per peer.
-export const DEFAULT_WS_MAX_MESSAGE_RATE = 100;
+export const DEFAULT_WS_MAX_MESSAGE_RATE = 1000;
```
ManuGowda: do you think we need more than 100/sec WS message rate? we have less than 10 WS calls per/sec know? 
should we keep the limit lower?

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -136,12 +138,17 @@ export class Peer extends EventEmitter {
 			this._resetProductivity();
 		}, DEFAULT_PRODUCTIVITY_RESET_INTERVAL);
 		this._serverNodeInfo = peerConfig.serverNodeInfo;
+		this._protocolRCPCounter = 0;
+		this._protocolRCPEvents = new Set([
+			REMOTE_EVENT_RPC_GET_NODE_INFO,
+			REMOTE_EVENT_RPC_GET_PEERS_LIST,
```
ManuGowda: can we rename this `REMOTE_EVENT_RPC_GET_PEERS_LIST` to `REMOTE_EVENT_RPC_GET_PEER_LIST`, because `peer list` is already list, `peers list` is weird

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -136,12 +138,17 @@ export class Peer extends EventEmitter {
...
+		]);
 
 		// This needs to be an arrow function so that it can be used as a listener.
 		this._handleRawRPC = (
 			packet: unknown,
 			respond: (responseError?: Error, responseData?: unknown) => void,
-		) => {
+		): void => {
 			// TODO later: Switch to LIP protocol format.
```
ManuGowda: I think this comment is no longer valid, maybe remove it?

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -136,12 +138,17 @@ export class Peer extends EventEmitter {
 			this._resetProductivity();
 		}, DEFAULT_PRODUCTIVITY_RESET_INTERVAL);
 		this._serverNodeInfo = peerConfig.serverNodeInfo;
+		this._protocolRCPCounter = 0;
+		this._protocolRCPEvents = new Set([
+			REMOTE_EVENT_RPC_GET_NODE_INFO,
```
ManuGowda: move this to constants as `P2P RPC EVENTS` and use it here, so that in the future if we have new  RPC events, its easy to modify at one place.

elements/lisk-p2p/test/integration/discovery.ts
```diff
@@ -275,6 +275,7 @@ describe('Network discovery', () => {
 			const customConfig = (index: number) => ({
 				maxOutboundConnections: index % 2 === 1 ? 3 : 20,
 				fallbackSeedPeerDiscoveryInterval: index === 2 ? 100 : 10000,
+				rateCalculationInterval: 100,
```
ManuGowda: `rateLimiterCalculationInterval`?

elements/lisk-p2p/test/integration/penalty_protocol_message_limit.ts
```diff
@@ -0,0 +1,71 @@
...
+
+		await wait(NETWORK_CREATION_WAIT_TIME);
+	});
+
+	afterEach(async () => {
+		await destroyNetwork(p2pNodeList);
+	});
+
+	it(`should apply penalty for limit exceed`, async () => {
+		// Arrange
```
ManuGowda: you can remove this if not required.

elements/lisk-p2p/src/p2p.ts
```diff
@@ -289,10 +290,15 @@ export class P2P extends EventEmitter {
 
 		// This needs to be an arrow function so that it can be used as a listener.
 		this._handlePeerPoolRPC = (request: P2PRequest) => {
+			// Process protocol messages
 			if (request.procedure === REMOTE_EVENT_RPC_GET_PEERS_LIST) {
 				this._handleGetPeersRequest(request);
 			}
 
+			if (request.procedure === REMOTE_EVENT_RPC_GET_NODE_INFO) {
```
nazarhussain: Redundant if for same value `request.procedure`, wrap with `else if` or move to `switch` statement.

elements/lisk-p2p/src/p2p.ts
```diff
@@ -289,10 +290,15 @@ export class P2P extends EventEmitter {
...
+				request.end(this._nodeInfo);
```
nazarhussain: I saw a convention used in code `_handle<Request>`, please use same instead of manipulating `request` object directly.

elements/lisk-p2p/src/events.ts
```diff
@@ -40,12 +40,20 @@ export const EVENT_FAILED_TO_SEND_MESSAGE = 'failedToSendMessage';
...
+
+// P2P Protocol messages
 export const REMOTE_EVENT_POST_NODE_INFO = 'postNodeInfo';
 export const REMOTE_EVENT_RPC_GET_NODE_INFO = 'getNodeInfo';
 export const REMOTE_EVENT_RPC_GET_PEERS_LIST = 'getPeers';
 export const REMOTE_EVENT_PING = 'ping';
 export const REMOTE_EVENT_PONG = 'pong';
 
+// P2P Protocol messages list
+export const PROTOCOL_REMOTE_EVENT_RCP_EVENTS: Set<string> = new Set([
```
nazarhussain: Didn't understand or relate this name with other names above. As per implementation detail it looks more like. `VALID_PROTOCOL_EVENTS` or `COUNTABLE_REMOTE_EVENTS` or `PROTOCOL_EVENTS_TO_RATE_LIMIT`

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -157,6 +159,15 @@ export class Peer extends EventEmitter {
 				return;
 			}
 
+			// Protocol RCP request limiter LIP-0004
+			if (PROTOCOL_REMOTE_EVENT_RCP_EVENTS.has(rawRequest?.procedure)) {
+				if (this._protocolRCPCounter <= PROTOCOL_REMOTE_EVENT_RCP_EVENTS.size) {
+					this._protocolRCPCounter += 1;
```
nazarhussain: This implementation count each `REMOTE_EVENT_RPC_GET_PEERS_LIST` and `REMOTE_EVENT_RPC_GET_NODE_INFO` events together, although the LIP suggest limit only on `getPeers` separately. Can you explain it further why you combined both?

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -157,6 +159,15 @@ export class Peer extends EventEmitter {
...
+				} else {
+					return;
+				}
```
nazarhussain: I would write the above code with guard conditions. 

```
if (this._protocolRCPCounter > PROTOCOL_REMOTE_EVENT_RCP_EVENTS.size) {
    return; 
}

this._protocolRCPCounter += 1;
```
It increase readability.

elements/lisk-p2p/src/peer/base.ts
nazarhussain: If you are trying to guard only 1 `REMOTE_EVENT_RPC_GET_NODE_INFO` and `REMOTE_EVENT_RPC_GET_PEERS_LIST`, then this code will also work in case of two `REMOTE_EVENT_RPC_GET_PEERS_LIST` or two `REMOTE_EVENT_RPC_GET_NODE_INFO` requests.

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -448,15 +438,26 @@ export class Peer extends EventEmitter {
...
 
 		if (
 			this.peerInfo.internalState.wsMessageRate >
-			this._peerConfig.wsMaxMessageRate
+				this._peerConfig.wsMaxMessageRate ||
+			this._protocolRCPCounter > PROTOCOL_REMOTE_EVENT_RCP_EVENTS.size
 		) {
-			this.applyPenalty(this._peerConfig.wsMaxMessageRatePenalty);
+			// Allow to increase penalty based on message rate limit exceeded
+			const messageRateExceedCoeff = Math.floor(
```
nazarhussain: Avoid abbreviations in variable names. I guess you are referring to `Coefficient`  here as `messageRateExceed Coefficient`

elements/lisk-p2p/src/constants.ts
```diff
@@ -21,6 +21,8 @@ export const DEFAULT_BAN_TIME = 86400000; // Milliseconds in a day -> hours*minu
 export const DEFAULT_POPULATOR_INTERVAL = 10000;
 export const DEFAULT_FALLBACK_SEED_PEER_DISCOVERY_INTERVAL = 30000;
 export const DEFAULT_SEND_PEER_LIMIT = 16;
+// Message types
+export const RCP_REQUEST_TYPE = '/RPCRequest';
```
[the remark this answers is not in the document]
valamidev: The name came from how our `rpc-request` event object structure looks like.
```
{
			type: '/RPCRequest',
			procedure: 'getBlocksFromId',
			data: {},
},
```
Change the value of this constant will cause unexpected incompatibility with existing tools or modules.

elements/lisk-p2p/src/peer/base.ts
valamidev: LIP-0004 mention both event as `should happen only at connection create`, with the given implementation we are allowing the expected scenario.

PR #4643, opened 2019-12-19
Update to Node.js v12.14.0 #4640 - Closes #4640

### LINKED ISSUE #4640: Update to Node.js v12.14.0
### Expected behavior
Ref https://github.com/nodejs/node/releases/tag/v12.14.0
### Actual behavior
Current Node.js version `12.13.0`
### Steps to reproduce
https://github.com/LiskHQ/lisk-sdk/blob/development/.nvmrc
development, 3.0.0

ManuGowda:
### What was the problem?
Ref https://github.com/nodejs/node/releases/tag/v12.14.0
### How did I solve it?
Update to Node.js v12.14.0
### How to manually test it?
- Install Node.js v12.14.0
- Build `npm run clean:node_modules; rm -rf ./node_modules; npm ci; npm run bootstrap -- --ci; npm run build`
### Review checklist

- [ ] The PR resolves #4640 


shuse2:
No need to update the requirement to the latest version. 
It's only necessary to update the nvmrc

ManuGowda:
> No need to update the requirement to the latest version.
> It's only necessary to update the nvmrc

its `>=` and we have `<=12` for the ones who don't what to update, so it should be fine I guess?

fchavant:
@shuse2 in this case we really want to use the new `npm` version though

shuse2:
> its >= and we have <=12 for the ones who don't what to update, so it should be fine I guess?
this means between 12.3 and 12.x.x, so if you change the lower end, it will be updated.
For SDK, this version is supported version as library.
Therefore, the range shouldn't be changed. (we should support as much as possible)
but the user (application) should update the version.

In other words, even if we update the version here, if user(application) doesn't update, it doesn't work

ManuGowda:
> > its >= and we have <=12 for the ones who don't what to update, so it should be fine I guess?
> 
> this means between 12.3 and 12.x.x, so if you change the lower end, it will be updated.
> For SDK, this version is supported version as library.
> Therefore, the range shouldn't be changed. (we should support as much as possible)
> but the user (application) should update the version.
> 
> In other words, even if we update the version here, if user(application) doesn't update, it doesn't work

This check `"node": ">=12.13.0 <=12"` which is checking node version should be greater than `12.13.0` and less than `12` 🤔 ? shouldn't that be `>=12 <=12.x.x`? in our case `>=12 <=12.14.0`?

we want the SDK to be supportable between `12 and 12.4.0`, is that correct?

PR #4661, opened 2020-01-06
Rename nethash to networkId in lisk-p2p - Closes #4647

### LINKED ISSUE #4647: Update term nethash to networkId in the p2p library
### Description
All of the variable usage `nethash` should be updated to `networkId` in the p2p library.

### Motivation

`Nethash` is very specific term to the blockchain, and also we introduced the term `network identifier`.

### Acceptance Criteria

No reference to the `nethash` should not exist in the lisk-p2p

ishantiw:
### What was the problem?

`nethash` name has been discontinued in the framework and `networkId` or `networkIdentifier` is being used so it needs to change everywhere in lisk-p2p library.

### How did I solve it?

Renamed `nethash` to `networkId`

### How to manually test it?

`npm t` under `elements/lisk-p2p`

### Review checklist

- [x] The PR resolves #4647 


framework/src/controller/application.js
```diff
@@ -288,13 +288,13 @@ class Application {
 	_compileAndValidateConfigurations() {
 		const modules = this.getModules();
 
-		this.config.app.nethash = this.genesisBlock.payloadHash;
```
shuse2: we can keep this as it was. For the framework, networkId is not the payloadHash.
We only need to change it in the `lisk-p2p`, for the framework, it can remain as it was for now.

shuse2:
Sorry maybe I made it confusing.
We don't need to change the nethash in the framework, but we need to inject the correct information into the P2P library.
As of now, I think it doesn't have the networkId in the p2p library

PR #4673, opened 2020-01-09
Fix NaCL slow/fast module loading

### LINKED ISSUE #4632: NaCl falls back to "slow" making signatures slower
### Expected behavior

Lisk crypto uses "slow" when "process.env.NACL_FAST === 'disable'". In
https://github.com/LiskHQ/lisk-sdk/blob/development/elements/lisk-cryptography/src/nacl/index.ts

### Actual behavior

Using slow when "process.env.NACL_FAST" is undefined.

### Steps to reproduce
Check which version (slow or fast) of the library is used. 

### Proposed solution
Add a check for existence of process.env.NACL_FAST:
```
try {
	if (process.env.NACL_FAST && process.env.NACL_FAST === 'disable') {
		throw new Error('Use tweetnacl');
	}
	// Require used for conditional importing
	// tslint:disable-next-line no-var-requires no-require-imports
	lib = require('./fast');
} catch (err) {
	process.env.NACL_FAST = 'disable';
	// tslint:disable-next-line no-var-requires no-require-imports
	lib = require('./slow');
}
```
Even better: remove the try/catch logic here.

valamidev:
### What was the problem?
- Lisk-cryptography got 2 different implementation for NaCL cryptography library, `fast and slow`, we prefer to use Fast unless it is disabled with `process.env.NACL_FAST=disable`. Due a bug we choose slow when NACL_FAST were undefined which is the default value of a non set environment variable.

### How did I solve it?
- Updated the module import logic statement.

### How to manually test it?
- `npm run test`

### Review checklist

- [x] The PR resolves #4632 


elements/lisk-cryptography/test/nacl/index.ts
```diff
@@ -109,25 +109,11 @@ describe('nacl index.js', () => {
 			return expect(process.env.NACL_FAST).to.eql('disable');
 		});
 
-		it('should load nacl slow if process.env.NACL_FAST is set to enable', () => {
```
shuse2: this test seems to be wrong, but why it was passing?

valamidev: The Test check the interface of the loaded module which is equal for slow and fast implementation.
```
{
  box: [Function],
  openBox: [Function],
  signDetached: [Function],
  verifyDetached: [Function],
  getRandomBytes: [Function],
  getKeyPair: [Function],
  getPublicKey: [Function]
}
```

valamidev: Related issue about tests #4677

PR #4712, opened 2020-01-16
Remove bignum library - Closes #4573

### LINKED ISSUE #4573: Remove BigNum from the dependency
### Description

After Node.10, we have BigInt native support.
Therefore, we should use BigInt from node.js instead of the library

### Acceptance Criteria

All BigNum reference should be removed from framework and elements

shuse2:
### What was the problem?
BigNum was used across the product, but Node.js supports `BigInt`

### How did I solve it?
Remove dependency of BigNum and replace with `BigInt`

### How to manually test it?
All the test should pass

### Review checklist

- [ ] The PR resolves #4573 


elements/lisk-blocks/src/schema.ts
```diff
@@ -52,15 +52,15 @@ export const baseBlockSchema = {
 			type: 'integer',
 		},
 		totalAmount: {
-			type: 'object',
+			typeof: 'bigint',
 			format: 'amount',
```
nazarhussain: The format here is still `amount`, which is actually

```
export const amount = isNumberString;
```

I am not sure how it will behave with `bigint`, as `typeof(bigint) !== 'string'`

elements/lisk-cryptography/src/buffer.ts
```diff
@@ -12,19 +12,38 @@
...
 	byteLength: number,
 	endianness: string = BIG_ENDIAN,
-) => new BigNum(value).toBuffer({ size: byteLength, endian: endianness });
+) => {
+	const buffer = Buffer.alloc(byteLength);
+	if (endianness === 'big') {
+		if (byteLength <= MAX_NUMBER_BYTE_LENGTH) {
+			buffer.writeUIntBE(Number(value), 0, byteLength);
+		} else {
+			buffer.writeBigUInt64BE(BigInt(value));
```
nazarhussain: Why not to provide offset in second condition?

elements/lisk-cryptography/src/buffer.ts
```diff
@@ -12,19 +12,38 @@
...
-import * as BigNum from '@liskhq/bignum';
 
 export const BIG_ENDIAN = 'big';
 export const LITTLE_ENDIAN = 'little';
+const MAX_NUMBER_BYTE_LENGTH = 6;
 
 export const intToBuffer = (
 	value: number | string,
 	byteLength: number,
 	endianness: string = BIG_ENDIAN,
```
nazarhussain: Would be nice to assert endian type. or Use enum instead.

elements/lisk-dpos/src/delegates_info.ts
```diff
@@ -366,8 +366,8 @@ export class DelegatesInfo {
 			uniqDelegateListWithRewardsInfo,
 			totalFee,
 		} = blocksInRounds.reduce(
-			(acc: AccountSummary, fetchedBlock: Block, i) => {
-				acc.totalFee = acc.totalFee.add(fetchedBlock.totalFee);
+			(acc: AccountSummary, fetchedBlock: Block | BlockJSON, i) => {
+				acc.totalFee = acc.totalFee + BigInt(fetchedBlock.totalFee);
```
nazarhussain:
```suggestion
				acc.totalFee = BigInt(acc.totalFee) + BigInt(fetchedBlock.totalFee);
```

elements/lisk-dpos/src/delegates_info.ts
```diff
@@ -377,15 +377,15 @@ export class DelegatesInfo {
...
+						reward: BigInt(fetchedBlock.reward),
 						isGettingRemainingFees: i === blocksInRounds.length - 1,
 					});
 					acc.delegatePublicKeys.push(fetchedBlock.generatorPublicKey);
 
 					return acc;
 				}
 
-				delegate.reward = delegate.reward.add(fetchedBlock.reward);
+				delegate.reward = delegate.reward + BigInt(fetchedBlock.reward);
```
nazarhussain:
```suggestion
				delegate.reward = BigInt(delegate.reward) + BigInt(fetchedBlock.reward);
```

elements/lisk-cryptography/src/buffer.ts
shuse2: for the `writeUintBE` it is required to specify probably because byteLength is required, but on BigUint64BE, it's optional

elements/lisk-dpos/src/delegates_info.ts
shuse2: `totalFee` has `add` meaning that it's already BigNum. therefore, no need to double cast

elements/lisk-dpos/src/delegates_info.ts
shuse2: same as above

elements/lisk-blocks/src/schema.ts
shuse2: As isNumberString, it was always suppose to be a `string`

shuse2:
I think the `protocol-spec`, is detached from the development, and until it actually update the protocol, it's probably better to be remained.
Also, lisk-bignum cannot be deprecated until we rollout the changes because master still depends on it

PR #4733, opened 2020-01-23
Move network to app level

### LINKED ISSUE #3358: Move network communication functionality to the application level
### Expected behavior
Each module in the application must have separation of concern and also spezialisation in what they do. Being the decentralised nature of the application we decided that each module should be able to communicate over the network. So the instance of P2P library should exist on the application level and network communication should be exposed through existing `channel` interface with new interface methods. 

### Actual behavior
Current prototype implementation of network module is completely based on proxy pattern. Its just proxying stuff around which is does make it a candidate of the a separate module. 

1.7.0

ishantiw:
### Description

Move the network to application level for the reasons explained in #3358 

### Review checklist

- [x] The PR resolves #3358 


framework/src/controller/channels/in_memory_channel.js
```diff
@@ -75,6 +75,14 @@ class InMemoryChannel extends BaseChannel {
 		return this.bus.invoke(action.serialize());
 	}
 
+	async invokeFromNetwork(remoteMethod, params) {
```
shuse2: @mitsuaki-u what do u think about the name?

framework/src/controller/channels/in_memory_channel.js
```diff
@@ -75,6 +75,14 @@ class InMemoryChannel extends BaseChannel {
...
+		return this.invoke(`app:${remoteMethod}`, params);
+	}
+
+	async publishToNetwork(actionName, data) {
```
shuse2: I think this change also needs in the child_channel.js?

framework/src/controller/controller.js
```diff
@@ -308,6 +343,7 @@ class Controller {
 
 		try {
 			await this.bus.cleanup();
+			this.network.stopNetwork();
```
ManuGowda: this should be an async function and you need to await here

framework/src/controller/network/network.js
```diff
@@ -347,50 +311,63 @@ module.exports = class Network {
...
+
+	applyPenalty(penaltyPacket) {
+		return this.p2p.applyPenalty({
+			peerId: penaltyPacket.peerId,
+			penalty: penaltyPacket.penalty,
+		});
 	}
 
-	async cleanup() {
+	stopNetwork() {
```
ManuGowda: this should be an `async` `return this.p2p.stop()` stop is promise

framework/src/controller/network/network.js
```diff
@@ -48,46 +45,15 @@ const NETWORK_INFO_KEY_TRIED_PEERS = 'tried_peers_list';
 const DEFAULT_PEER_SAVE_INTERVAL = 10 * 60 * 1000; // 10min in ms
 
 module.exports = class Network {
-	constructor(options) {
-		this.options = options;
-		this.channel = null;
-		this.logger = null;
-		this.storage = null;
+	constructor({ networkConfig, channel, logger, storage }) {
+		this.networkConfig = networkConfig;
```
nazarhussain: What's the need to change `options` to `networkConfig`? I would be rather consistent to use `options` in ever class.

framework/src/controller/network/network.js
nazarhussain: Having `stop` in the `Network` class is good. 

```suggestion
	stop() {
```

framework/src/controller/schema/application_config_schema.js
```diff
@@ -17,7 +17,7 @@
 module.exports = {
 	id: '#/app/config',
 	type: 'object',
-	required: ['app', 'components', 'modules'],
+	required: ['app', 'network', 'components', 'modules'],
```
nazarhussain: I would suggest to move `network` under `app`. It will match the code structure. Top level configurations should only stay with `app`, `components` and `modules`

framework/src/controller/channels/in_memory_channel.js
shuse2: We will keep this name for now

framework/src/controller/channels/in_memory_channel.js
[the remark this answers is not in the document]
ishantiw: `invokeFromNetwork` is an action on the network to call RPC on any peer in the network, `publishToNetwork` is calling the action on the network to `send` or `broadcast` (so implicitly publishing through network module even though it's invoking on the channel)

The idea here is, when a developer wants to invoke anything on the network, then call `invokeFromNetwork` and when wants to publish then use `publishToNetwork` so to differentiate from `invoke` and `publish` that are called locally

framework/src/controller/channels/base_channel.js
```diff
@@ -79,13 +79,28 @@ class BaseChannel {
...
 	// Specified as moduleName:actionName
 	// eslint-disable-next-line no-unused-vars, class-methods-use-this
 	async invoke(actionName, params) {
 		throw new TypeError('This method must be implemented in child classes. ');
 	}
 
+	// Call action network module when requesting from the network or a specific peer
+	// Specified as actionName for request available on the network
+	// eslint-disable-next-line no-unused-vars, class-methods-use-this
+	async invokefromNetwork(actionName, params) {
```
nazarhussain:
```suggestion
	async invokeFromNetwork(actionName, params) {
```

framework/src/controller/schema/application_config_schema.js
```diff
@@ -148,6 +149,134 @@ module.exports = {
 					},
 					additionalProperties: false,
 				},
+				network: {
```
shuse2: maybe no need to nest the properties?

ishantiw: Let's do it once we have chain also moved to the application to have a better understanding

Issue #4745, opened 2020-01-27
Change state store account/chain state get and getOrDefault to async


shuse2:
### Description

Update State Store, functions `get` and `getOrDefault` in `account` and `chain state` to be async function.
Check if there is cached version and if not, get from the database.

### Motivation

- When the `get` function is performant enough, there is no need to do `prepare`, so those 2 functions can be just async.
- In some cases, it's very hard to know all the affected accounts. Especially, in DPoS, it needs to get affected accounts from the delegate.

### Acceptance Criteria

- Test needs to be updated
- All the usage of the `get` or `getOrDefault` need to be updated to `await`

### Additional Information
- Core also needs to be updated.

PR #4749, opened 2020-01-27
Update lisk-dpos to use state store consistently- Closes #4136

### LINKED ISSUE #4136: Update Dpos module to consistently use state store

shuse2:
### What was the problem?
liks-dpos was not using state store and using database transactions to update the data.
It was inconsistent behavior which causes the execution order to be unclear.

* This PR doesn't work until #4745 , because it needs to load the accunts who are voted from delegates at the end of the round

### How did I solve it?
- Update to use state store instead of RoundDelegages table
- Update to create 2 rounds ahead instead of using 2 rounds before
- Remove database transaction creation from the processor

### How to manually test it?
- All the test should pass, and observe processing happen with correct forgers
- Sync up with betanet

### Review checklist

- [ ] The PR resolves #4136 


framework/src/application/node/processor/processor.js
```diff
@@ -268,121 +265,88 @@ class Processor {
...
+		if (!saveOnlyState) {
+			this.channel.publish('app:processor:newBlock', {
+				block: blockJSON,
+			});
+		}
 
-				return block;
-			},
-		);
+		return block;
```
ManuGowda: do we use the return block? or should this function return void?

framework/src/application/node/processor/processor.js
```diff
@@ -268,121 +265,88 @@ class Processor {
...
+		await processor.applyGenesis.run({
+			block,
+			stateStore,
+		});
+		await this.blocksModule.save(block, stateStore, { saveOnlyState });
 
-				return block;
-			},
-		);
+		return block;
```
ManuGowda: same comment, why not void

framework/src/application/node/processor/processor.js
shuse2: In some case, it might need processed block

framework/src/application/node/block_processor_v0.js
```diff
@@ -177,21 +176,27 @@ class BlockProcessorV0 extends BaseBlockProcessor {
...
-				this.dposModule.apply(block, { tx, delegateListRoundOffset }),
+			({ block, stateStore }) =>
+				this.dposModule.apply(block, stateStore, {
+					delegateListRoundOffset,
+				}),
 		]);
 
 		this.applyGenesis.pipe([
-			({ block, stateStore }) =>
+			async ({ block, stateStore }) =>
```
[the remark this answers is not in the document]
shuse2: it is not required, so ill remove

framework/src/application/node/processor/processor.js
shuse2: same as above

elements/lisk-blocks/src/blocks.ts
```diff
@@ -258,6 +258,10 @@ export class Blocks extends EventEmitter {
 		this.dataAccess.resetBlockHeaderCache();
 	}
 
+	public newStateStore(): StateStore {
```
mitsuaki-u: `createStateStore`?

PR #4750, opened 2020-01-27
Remove scope property from chain - Closes #4748

### LINKED ISSUE #4748: Remove scope property from chain
### Description

Remove `scope` variable from SDK

### Motivation

The property `scope` was almost completely removed during refactors for 2.X/3.X but some usages still remain. We should remove it completely so the concept of scope no longer is present in the SDK

### Acceptance Criteria

- No usage of `scope` or `this.scope` exists in chain

pablitovicente:
### What was the problem?

The `scope` variable was almost completely removed during refactors for 2.0/3.0 but chain was still using it in some places

### How did I solve it?

By removing the `scope` variable. Note that to avoid refactoring some files that were expecting some of the nesting that `scope` presented I'm wrapping the required variables in an object when passing around some components.

### Review checklist

- [x] The PR resolves #4748


framework/src/modules/chain/chain.js
```diff
@@ -110,25 +114,25 @@ module.exports = class Chain {
...
-			await bootstrapStorage(this.scope, global.constants.ACTIVE_DELEGATES);
+			this.sequence = new Sequence({
+				onWarning(current) {
+					this.components.logger.warn('Main queue', current);
+				},
+			});
+
+			await bootstrapStorage(
+				{ components: this.components },
+				global.constants.ACTIVE_DELEGATES,
```
shuse2: this can be changed to reference value from `this.config`, and `global` should be removed

framework/src/modules/chain/chain.js
```diff
@@ -349,12 +353,12 @@ module.exports = class Chain {
 	}
 
 	async _initModules() {
-		this.scope.modules = {};
+		this.modules = {};
```
shuse2: is this used anywhere?

pablitovicente: @shuse2 in `async cleanup`

framework/src/modules/chain/chain.js
pablitovicente: Done!

PR #4758, opened 2020-01-28
Change state store account/chain state get and getOrDefault to async - Closes #4745

### LINKED ISSUE #4745: Change state store account/chain state get and getOrDefault to async

pablitovicente:
### What was the problem?

State store account/chain state need to make an attempt at getting accounts being modified in transactions (for example when processing votes)

### How did I solve it?

Check code

### How to manually test it?

Build should be green

### Review checklist

- [x] The PR resolves #4745


elements/lisk-blocks/test/unit/transactions_handlers.spec.ts
```diff
@@ -882,11 +882,11 @@ describe('transactions', () => {
 			trs1.senderId = account1.address;
 			trs2.senderId = account2.address;
 
-			const result = transactionHandlers.verifyTotalSpending(
+			const result = await transactionHandlers.verifyTotalSpending(
 				[trs1, trs2],
 				stateStoreMock,
 			);
-
+			console.log(result);
```
ishantiw: delete this `console.log`

pablitovicente: Removed

PR #4760, opened 2020-01-29
Update crypto not to call fast if slow is specified - Closes #4759

### LINKED ISSUE #4759: Sodium-native breaks the Webpack build process
### Expected behavior
Lisk Element should not require sodium-native by default. This library is marked as optional and should be treated as optional.

### Actual behavior
Lisk Element tries to r[equire sodium-native by default](https://github.com/LiskHQ/lisk-sdk/blob/v3.0.1/elements/lisk-cryptography/src/nacl/index.ts#L26). 

The try-catch statement works fine if you want to check if a library is available. But the issue here is that the target library is indeed available, but it's not compatible with the Webpack build process. So it gets imported, Webpack tries to build the bundle, it can't so it breaks and throws error. The compile process is halted and it never gets to the catch statement. 

### Steps to reproduce
https://jenkins.lisk.io/blue/organizations/jenkins/lisk-desktop/detail/PR-2747/4/pipeline

3.1

shuse2:
### What was the problem?
Webpack build process did not allow to catch error

### How did I solve it?
If the `process.env.NACL_FAST === disable`, then it doesn't read the `fast` module

### How to manually test it?
Try to build with webpack with this change

### Review checklist

- [ ] The PR resolves #4759 

### DIFF
--- a/elements/lisk-cryptography/src/nacl/index.ts
+++ b/elements/lisk-cryptography/src/nacl/index.ts
@@ -14,20 +14,16 @@
  */
 import { NaclInterface } from './nacl_types';
 
-// tslint:disable-next-line no-let
-let lib: NaclInterface;
+// tslint:disable-next-line no-let no-require-imports no-var-requires
+let lib: NaclInterface = require('./slow');
 
 try {
-	if (process.env.NACL_FAST === 'disable') {
-		throw new Error('Use tweetnacl');
+	if (process.env.NACL_FAST !== 'disable') {
+		// tslint:disable-next-line no-var-requires no-require-imports
+		lib = require('./fast');
 	}
-	// Require used for conditional importing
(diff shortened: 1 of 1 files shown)


elements/lisk-cryptography/src/nacl/index.ts
```diff
@@ -14,20 +14,16 @@
...
 
-// tslint:disable-next-line no-let
-let lib: NaclInterface;
+// tslint:disable-next-line no-let no-require-imports no-var-requires
+let lib: NaclInterface = require('./slow');
 
 try {
-	if (process.env.NACL_FAST === 'disable') {
-		throw new Error('Use tweetnacl');
+	if (process.env.NACL_FAST !== 'disable') {
```
reyraa: the NACL_FAST must be a Boolean and fall back to false. There’s no environment variable in React Native.

PR #4769, opened 2020-01-31
Fix BFT condition - Closes #4768

### LINKED ISSUE #4768: BFT verify block header need to check version and prevotes
### Expected behavior
https://github.com/LiskHQ/lisk-sdk/blob/master/framework/src/modules/chain/bft/finality_manager.js#L293

function should check the order of delegateLastBlock and block header by 
```
// Order the two block headers such that b1 must be forged first
   let b1=blockHeader1;
   let b2=blockHeader2;
   if(b1.maxHeightPreviouslyForged>b2.maxHeightPreviouslyForged ||
     (b1.maxHeightPreviouslyForged==b2.maxHeightPreviouslyForged && b1.maxHeightPrevoted>b2.maxHeightPrevoted) ||
     (b1.maxHeightPreviouslyForged==b2.maxHeightPreviouslyForged && b1.maxHeightPrevoted==b2.maxHeightPrevoted && b1.height>b2.height)){
      b1=blockHeader2;
      b2=blockHeader1;
   }
```

and the `BFTForkChoiceRuleError` should check maxHeightPrevoted, not `maxHeightPreviouslyForged`

as described here
https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#detecting-contradicting-block-headers

Also, default maxHeightPreviouslyForged should be the forging height not `0`

### Actual behavior
It's not checking the order of header, and checking the wrong property

shuse2:
### What was the problem?
Refer to #4768 

### How did I solve it?
Update the BFT condition

### How to manually test it?
- Start forging on local, and check if the initial forger_info doesn't contain `0`
- BFT condition depends on fork scenario

### Review checklist

- [ ] The PR resolves #4768 

### DIFF
--- a/framework/src/modules/chain/bft/finality_manager.js
+++ b/framework/src/modules/chain/bft/finality_manager.js
@@ -304,10 +304,27 @@ class FinalityManager extends EventEmitter {
 			return true;
 		}
 
+		// Order the two block headers such that earlierBlock must be forged first
+		let earlierBlock = delegateLastBlock;
+		let laterBlock = blockHeader;
 		if (
-			delegateLastBlock.maxHeightPreviouslyForged ===
-				blockHeader.maxHeightPreviouslyForged &&
-			delegateLastBlock.height >= blockHeader.height
+			earlierBlock.maxHeightPreviouslyForged >
+				laterBlock.maxHeightPreviouslyForged ||
+			(earlierBlock.maxHeightPreviouslyForged ===
+				laterBlock.maxHeightPreviouslyForged &&
+				earlierBlock.maxHeightPrevoted > laterBlock.maxHeightPrevoted) ||
+			(earlierBlock.maxHeightPreviouslyForged ===
+				laterBlock.maxHeightPreviouslyForged &&
+				earlierBlock.maxHeightPrevoted === laterBlock.maxHeightPrevoted &&
+				earlierBlock.height > laterBlock.height)
+		) {
+			earlierBlock = blockHeader;
+			laterBlock = delegateLastBlock;
+		}
+
+		if (
+			earlierBlock.maxHeightPrevoted === laterBlock.maxHeightPrevoted &&
+			earlierBlock.height >= laterBlock.height
 		) {
 			// Violation of the fork choice rule as delegate moved to different chain
 			// without strictly larger maxHeightPreviouslyForged or larger height as
@@ -315,11 +332,11 @@ class FinalityManager extends EventEmitter {
 			throw new BFTForkChoiceRuleError();
 		}
 
-		if (delegateLastBlock.height > blockHeader.maxHeightPreviouslyForged) {
+		if (earlierBlock.height > laterBlock.maxHeightPreviouslyForged) {
 			throw new BFTChainDisjointError();
 		}
 
-		if (delegateLastBlock.maxHeightPrevoted > blockHeader.maxHeightPrevoted) {
+		if (earlierBlock.maxHeightPrevoted > laterBlock.maxHeightPrevoted) {
 			throw new BFTLowerChainBranchError();
 		}
 
--- a/framework/src/modules/chain/bft/finality_manager.js
+++ b/framework/src/modules/chain/bft/finality_manager.js
@@ -307,19 +307,23 @@ class FinalityManager extends EventEmitter {
 		// Order the two block headers such that earlierBlock must be forged first
 		let earlierBlock = delegateLastBlock;
 		let laterBlock = blockHeader;
-		if (
+		const higherMaxHeightPreviouslyForgerd =
 			earlierBlock.maxHeightPreviouslyForged >
-				laterBlock.maxHeightPreviouslyForged ||
-			(earlierBlock.maxHeightPreviouslyForged ===
-				laterBlock.maxHeightPreviouslyForged &&
-				earlierBlock.maxHeightPrevoted > laterBlock.maxHeightPrevoted) ||
-			(earlierBlock.maxHeightPreviouslyForged ===
-				laterBlock.maxHeightPreviouslyForged &&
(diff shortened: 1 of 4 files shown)


framework/src/modules/chain/bft/finality_manager.js
```diff
@@ -304,22 +304,39 @@ class FinalityManager extends EventEmitter {
...
+			(earlierBlock.maxHeightPreviouslyForged ===
+				laterBlock.maxHeightPreviouslyForged &&
+				earlierBlock.maxHeightPrevoted > laterBlock.maxHeightPrevoted) ||
+			(earlierBlock.maxHeightPreviouslyForged ===
+				laterBlock.maxHeightPreviouslyForged &&
+				earlierBlock.maxHeightPrevoted === laterBlock.maxHeightPrevoted &&
+				earlierBlock.height > laterBlock.height)
+		) {
+			earlierBlock = blockHeader;
+			laterBlock = earlierBlock;
```
IkerAlus: Should not this be:
```
laterBlock = delegateLastBlock;
```
?
It looks like here `earlierBlock` and `laterBlock` are going to be the same.

framework/src/modules/chain/bft/finality_manager.js
```diff
@@ -304,22 +304,39 @@ class FinalityManager extends EventEmitter {
...
+				laterBlock.maxHeightPreviouslyForged &&
+				earlierBlock.maxHeightPrevoted > laterBlock.maxHeightPrevoted) ||
+			(earlierBlock.maxHeightPreviouslyForged ===
+				laterBlock.maxHeightPreviouslyForged &&
+				earlierBlock.maxHeightPrevoted === laterBlock.maxHeightPrevoted &&
+				earlierBlock.height > laterBlock.height)
+		) {
+			earlierBlock = blockHeader;
+			laterBlock = delegateLastBlock;
+		}
```
nazarhussain:
```js
const higherMaxHeightPreviouslyForged = earlierBlock.maxHeightPreviouslyForged > laterBlock.maxHeightPreviouslyForged;
const sameMaxHeightPreviouslyForged = earlierBlock.maxHeightPreviouslyForged === laterBlock.maxHeightPreviouslyForged;
const higherMaxHeightPrevoted = earlierBlock.maxHeightPrevoted > laterBlock.maxHeightPrevoted;
const sameMaxHeightPrevoted =earlierBlock.maxHeightPrevoted === laterBlock.maxHeightPrevoted;
const higherHeight = earlierBlock.height > laterBlock.height;
const sameHeight = earlierBlock.height === laterBlock.height;

if(higherMaxHeightPreviouslyForged 
    || (sameMaxHeightPreviouslyForged && higherMaxHeightPrevoted) 
    || (sameMaxHeightPreviouslyForged && sameMaxHeightPrevoted && higherHeight)) {

  [earlierBlock, laterBlock ] = [laterBlock, earlierBlock];
}
```

framework/src/modules/chain/block_processor_v2.js
```diff
@@ -281,19 +281,24 @@ class BlockProcessorV2 extends BaseBlockProcessor {
...
 			// Create a block with with basic block and bft properties
 			async data => {
-				const previouslyForged = await this._getPreviouslyForgedMap();
+				const previouslyForgedMap = await this._getPreviouslyForgedMap();
 				const delegatePublicKey = data.keypair.publicKey.toString('hex');
+				const height = data.previousBlock.height + 1;
+				const previousBlockId = data.previousBlock.id;
 				const maxHeightPreviouslyForged =
-					previouslyForged[delegatePublicKey] || 0;
+					previouslyForgedMap[delegatePublicKey] || height;
```
nazarhussain: I am not sure if that's right. `data.previousBlock` refers to the previous block in the chain, not the the previous block forged by delegate. So in case we don't know what height delegated previously forged, setting it to previous block in the chain does not make sense.

framework/src/modules/chain/bft/finality_manager.js
```diff
@@ -304,22 +304,39 @@ class FinalityManager extends EventEmitter {
 			return true;
 		}
 
+		// Order the two block headers such that earlierBlock must be forged first
+		let earlierBlock = delegateLastBlock;
+		let laterBlock = blockHeader;
```
ManuGowda: later can be future tense and the block is not saved, so why not present tense? something like `currentBlock`?

framework/src/modules/chain/block_processor_v2.js
shuse2: On the LIP it says
> The first block forged by the delegate uses the value max{A.height, A.maxHeightPreviouslyForged} in the maxHeightPreviouslyForged property.

framework/test/jest/unit/specs/modules/chain/block_processor_v2.spec.js
```diff
@@ -92,7 +92,7 @@ describe('block processor v2', () => {
 	describe('create', () => {
 		let block;
 
-		it('should set maxPreviouslyForgedHeight to zero when the delegate did not forge before', async () => {
+		it('should set maxPreviouslyForgedHeight to height when the delegate did not forge before', async () => {
```
ManuGowda: can we keep the scenario where `maxHeightPreviouslyForged` can be zero?

framework/src/modules/chain/block_processor_v2.js
nazarhussain: I think, this condition applies to migrate a node with existing data explicitly. 

The LIP also says very clearly that it should be `0` in case delegate didn't forged the block earlier. 

> If a delegate forges a block B to be added to the current chain, then B.maxHeightPreviouslyForged must be the largest height at which the delegate forging the block previously forged a block on any chain or 0 if the delegate did not forge a block before.

https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#additional-block-header-properties

PR #4773, opened 2020-01-31
Update pull request (PULL_REQUEST_TEMPLATE.md)  template without review checklist

sridharmeganathan:
This PR is created to update the pull request template.  We would like to remove the review checklist from the existing template.


docs/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,13 +1,8 @@
...
-
-### Review checklist
-
-- [ ] The PR resolves #INSERT_ISSUE_NUMBER
-- [ ] All new code is covered with unit tests
-- [ ] Relevant integration / functional tests are added
-- [ ] Commit messages follow the [commit guidelines](CONTRIBUTING.md#git-commit-messages)
-- [ ] Documentation has been added/updated
+### How this has been tested?
+<!--- Please describe how you tested your changes -->
```
shuse2: This should be consistent with the above phrase

sridharmeganathan: Thanks for your feedback. Updated based on your comments

pablitovicente: I think the proper form is `How was it tested?`

docs/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,13 +1,8 @@
 ### What was the problem?
+This PR resolves #INSERT_ISSUE_NUMBER
 
 ### How did I solve it?
```
pablitovicente: And if the other questions do not make reference to `I` then maybe this could be more consistent this way `How was it solved?`

PR #4777, opened 2020-02-03
Fix the cache loading to be consistent - Closes #4776

### LINKED ISSUE #4776: BFT header cache is not working properly
### Expected behavior
* BFT header cache should at least save last 303 blocks
* BFT header should search through last 303 blocks

### Actual behavior
In the start up, it only holds cache up to 202 blocks.
In case the cache is requested with more length than the holding size, it returns incorrect value

### Steps to reproduce
* request cache with higher value than the actual contents
* Check the length of cache after initialization

3.0-

shuse2:
### What was the problem?
The PR resolves #4776 

### How did I solve it?
Update the parameter to be 3 rounds in stead of 2.
Fix the cache `top` mechanism to return value

### How to manually test it?
- Start double forging using 2 nodes, and on the second round, restart one of the nodes, and check the validation is correct

### DIFF
--- a/framework/src/modules/chain/bft/bft.js
+++ b/framework/src/modules/chain/bft/bft.js
@@ -22,7 +22,10 @@ const {
 } = require('./finality_manager');
 const forkChoiceRule = require('./fork_choice_rule');
 const { validateBlockHeader } = require('./utils');
-const { BFT_ROUND_THRESHOLD } = require('./constant');
+const {
+	BFT_MIGRATION_ROUND_OFFSET,
+	BFT_ROUND_THRESHOLD,
+} = require('./constant');
 
 const CHAIN_STATE_FINALIZED_HEIGHT = 'BFT.finalizedHeight';
 const EVENT_BFT_BLOCK_FINALIZED = 'EVENT_BFT_BLOCK_FINALIZED';
@@ -73,7 +76,10 @@ class BFT extends EventEmitter {
 
 		const loadFromHeight = Math.max(
 			finalizedHeight,
-			lastBlockHeight - this.constants.activeDelegates * BFT_ROUND_THRESHOLD,
+			// Search is inclusive, therefore, it should start from one above (ex: 288 - 500, which results total 303)
+			lastBlockHeight -
+				this.constants.activeDelegates * BFT_ROUND_THRESHOLD +
+				1,
 			this.constants.startingHeight,
 		);
 
@@ -121,7 +127,9 @@ class BFT extends EventEmitter {
 			const tillHeight = this.finalityManager.minHeight - 1;
 			const fromHeight =
 				this.finalityManager.maxHeight -
-				this.constants.activeDelegates * BFT_ROUND_THRESHOLD;
+				// Search is inclusive, therefore, it should start from one above (ex: 288 - 500, which results total 303)
+				this.constants.activeDelegates * BFT_ROUND_THRESHOLD +
+				1;
 			await this._loadBlocksFromStorage({
 				fromHeight,
 				tillHeight,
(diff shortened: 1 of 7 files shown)


framework/src/modules/chain/bft/bft.js
```diff
@@ -197,7 +199,8 @@ class BFT extends EventEmitter {
 		// Check BFT migration height
 		// https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#backwards-compatibility
 		const bftMigrationHeight =
-			this.constants.startingHeight - this.constants.activeDelegates * 2;
+			this.constants.startingHeight -
+			this.constants.activeDelegates * BFT_ROUND_THRESHOLD;
```
nazarhussain: This variable `bftMigrationHeight ` is used to initialize the `finalizedHeight `. This value is clearly documented in LIP to be 202 blocks from the height of migraiton. Can we discuss with science team before making this change. 

> we propose that chainMaxHeightFinalized is initialized with the value l-202

https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#backwards-compatibility

framework/src/modules/chain/bft/bft.js
```diff
@@ -72,7 +73,7 @@ class BFT extends EventEmitter {
 
 		const loadFromHeight = Math.max(
 			finalizedHeight,
-			lastBlockHeight - this.constants.activeDelegates * 2,
+			lastBlockHeight - this.constants.activeDelegates * BFT_ROUND_THRESHOLD,
```
nazarhussain: Can we make it more readable, may be assign this value in initialiser. 

```js
this.constants = {
   activeDelegates, 
   startingHeight,
   twoRounds: activeDelegates * 2,
   threeRounds: activeDelegates * 3,
}
```

And then use it here...

```js
lastBlockHeight - this.constants.threeRounds,
```

This looks more clear while reading the code logic.

framework/src/modules/chain/bft/bft.js
```diff
@@ -112,14 +113,15 @@ class BFT extends EventEmitter {
...
 		if (
 			this.finalityManager.maxHeight - this.finalityManager.minHeight <
-			this.constants.activeDelegates * 2
+			this.constants.activeDelegates * BFT_ROUND_THRESHOLD
 		) {
 			const tillHeight = this.finalityManager.minHeight - 1;
 			const fromHeight =
-				this.finalityManager.maxHeight - this.constants.activeDelegates * 2;
+				this.finalityManager.maxHeight -
+				this.constants.activeDelegates * BFT_ROUND_THRESHOLD;
```
nazarhussain: The LIP defined 

> we need to ensure that the block headers of the blocks at heights A.height-302 up to height A.height are contained in memory

https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#reverting-blocks

This becomes having exactly 303 blocks in headers. Above logic will lead to 304 Blocks instead. Remember this `tillHeight` and `fromHeight` are inclusive. If we already have headers from `451-590`, above calculation will make sure to have `287-590`, this range when inclusive become 304 blocks.

framework/src/modules/chain/bft/bft.js
nazarhussain: This calculation also end up loading 304 blocks.

shuse2: Instead of having 2 rounds, 3 rounds, I would prefer to have the 2 different constants and used accordingly

framework/src/modules/chain/bft/headers_list.js
```diff
@@ -112,6 +112,10 @@ class HeadersList {
 	top(size) {
 		assert(size, 'Please provide the size');
 
+		if (this.length <= size) {
+			return [...this._items];
+		}
+
```
nazarhussain: It was a very minor logical error, should have been catched in development/review cycle.  🤦‍♂

PR #4778, opened 2020-02-03
Chain rebuild process is initialized in wrong order and failed to process - Closes #4775

### LINKED ISSUE #4775: Chain rebuild process is initialized in wrong order and failed to process
### Expected behavior

When we want to rebuild the chain and started the application with `--rebuild` option or `modules.chain.loading.rebuildUpToRound` option, it should rebuild the chain upto particular round. 

### Actual behavior

Currently there are three issues; 

1. The rebuilder is initialized before the block processor is started. So the rebuild process is actually failing on very start. 
2. If we fix that order, then rebuild failed due to BFT validation. BFT load up last 202 blocks at the time of initializing and forces that you can only add more blocks on top of the chain. But while rebuilding we actually want to start from block 1. We have to reset the BFT cache from the rebuilder. 
3. The command line option `--rebuild` look for `"type": "integer,null"`. If you start with `node index.js --rebuild` it failed. If you start with `node index.js --rebuild null` it failed. If you start with `node index.js --rebuild 'null'` it failed. So one must have to specify a integer value. There must be way to rebuild to all rounds available. 

### Steps to reproduce

```
cd framework/test/test_app/ && node index.js --rebuild 5000
```

3.0.0

nazarhussain:
### What was the problem?

There were multiple issues in rebuilder initialization. 

### How did I solve it?

Fixed by code refactoring and adding few functionality to round delegates. 

### How to manually test it?

```
cd framework/test/test_app/ && node index.js --rebuild 5000
```

### Review checklist

- [ ] The PR resolves #4775

### DIFF
--- a/framework/src/modules/chain/chain.js
+++ b/framework/src/modules/chain/chain.js
@@ -171,25 +171,6 @@ module.exports = class Chain {
 				this.options.genesisBlock,
 			);
 
-			// Deactivate broadcast and syncing during snapshotting process
-			if (this.options.loading.rebuildUpToRound) {
-				this.options.broadcasts.active = false;
-				this.options.syncing.active = false;
-				await this.rebuilder.rebuild(
-					this.options.loading.rebuildUpToRound,
-					this.options.loading.loadPerIteration,
-				);
-				this.logger.info(
-					{
-						rebuildUpToRound: this.options.loading.rebuildUpToRound,
-						loadPerIteration: this.options.loading.loadPerIteration,
-					},
-					'Successfully rebuild the blockchain',
-				);
-				process.emit('cleanup');
-				return;
-			}
-
 			this.channel.subscribe('app:state:updated', event => {
 				Object.assign(this.scope.applicationState, event.data);
 			});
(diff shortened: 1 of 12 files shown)


framework/src/modules/chain/chain.js
```diff
@@ -171,10 +171,29 @@ module.exports = class Chain {
...
+			});
+
 			// Deactivate broadcast and syncing during snapshotting process
-			if (this.options.loading.rebuildUpToRound) {
+			if (this.options.loading.rebuildUpToRound !== null) {
 				this.options.broadcasts.active = false;
 				this.options.syncing.active = false;
+
+				// Need to reset the BFT to rebuild from start of the chain
+				this.bft.reset();
```
shuse2: this should be done in rebuilder when we do other reset

framework/src/modules/chain/chain.js
```diff
@@ -171,10 +171,29 @@ module.exports = class Chain {
...
+			this.channel.invoke('app:updateApplicationState', {
+				height: this.blocks.lastBlock.height,
+				lastBlockId: this.blocks.lastBlock.id,
+				maxHeightPrevoted: this.blocks.lastBlock.maxHeightPrevoted || 0,
+				blockVersion: this.blocks.lastBlock.version,
+			});
+
 			// Deactivate broadcast and syncing during snapshotting process
-			if (this.options.loading.rebuildUpToRound) {
+			if (this.options.loading.rebuildUpToRound !== null) {
```
shuse2: does this need to be strictly `null`?

framework/src/modules/chain/chain.js
nazarhussain: Rebuilder don't have dependency for BFT. Do think we pass this dependency for this line?

framework/src/modules/chain/chain.js
nazarhussain: Yes because of its default value. 

https://github.com/LiskHQ/lisk-sdk/blob/0ca424501ba2123815f37fe871d44748171e31b3/framework/src/modules/chain/defaults/config.js#L272

framework/src/modules/chain/chain.js
shuse2: yes having dependency in the rebuilder is better than splitting the logic in 2 places IMO

framework/src/modules/chain/components/storage/sql/round_delegates/reset_round_delegates.sql
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+
+/*
+  DESCRIPTION: ?
+
+  PARAMETERS: None
+*/
```
ManuGowda: can we remove them

PR #4788, opened 2020-02-03
Create account class - Closes #4746

### LINKED ISSUE #4746: Create Account class and consistently use outside of lisk-blocks
### Description

Create account class in `lisk-blocks` and except within this library, it should always use this account class obtained through the libraries.

### Motivation
Properties like `balance` or `voteWight` are often converted from `string` to `bigint` across the system, and it is very easy to missed typed. Also, for user, necessity to consider the conversion is unnecessary steps.

### Acceptance Criteria
- Account class should have  `updateBalance(number | string | bigint) => Error` method.
- All the account usage should be converted to expect the account class

### Additional Information
- Core also needs to be updated

ishantiw:
### Description

Following changes required to fulfill #4746  

- Rename interface `Account` to `AccountJSON`
- Create `Account` class with all the properties as `AccountJSON`
- Create `addbalance` function in `Account` class
- Use `AccountJSON` only when reading or writing from the database
- Update `lisk-transactions` library interface Account to match with `Account` class in `lisk-blocks`
- Remove the usage of `BigInt` everywhere when reading a value from `Account` class object.
- Use `addBalance`  to update the balance
- Fix all the relevant tests

### Review checklist

- [x] The PR resolves #4746 

### DIFF
--- a/elements/lisk-blocks/src/account.ts
+++ b/elements/lisk-blocks/src/account.ts
@@ -30,9 +30,11 @@ export const accountDefaultValues = {
 	nameExist: false,
 	multiMin: 0,
 	multiLifetime: 0,
-	votedDelegatesPublicKeys: [],
+	// tslint:disable-next-line:no-null-keyword
+	votedDelegatesPublicKeys: null,
 	asset: {},
-	membersPublicKeys: [],
+	// tslint:disable-next-line:no-null-keyword
+	membersPublicKeys: null,
 };
 
 export class Account {
@@ -79,11 +81,12 @@ export class Account {
 		this.multiLifetime = accountInfo.multiLifetime;
 		this.asset = accountInfo.asset;
 		this.votedDelegatesPublicKeys =
+			!accountInfo.votedDelegatesPublicKeys ||
 			accountInfo.votedDelegatesPublicKeys === null
 				? []
 				: accountInfo.votedDelegatesPublicKeys;
 		this.membersPublicKeys =
-			accountInfo.membersPublicKeys === null
+			!accountInfo.membersPublicKeys || accountInfo.membersPublicKeys === null
 				? []
 				: accountInfo.membersPublicKeys;
 	}
@@ -114,14 +117,14 @@ export class Account {
 			multiMin: this.multiMin,
 			multiLifetime: this.multiLifetime,
 			votedDelegatesPublicKeys:
-				this.votedDelegatesPublicKeys.length > 0
-					? this.votedDelegatesPublicKeys
-					: // tslint:disable-next-line:no-null-keyword
-					  null,
+				this.votedDelegatesPublicKeys.length < 1
+					? // tslint:disable-next-line:no-null-keyword
+					  null
+					: this.votedDelegatesPublicKeys,
 			asset: this.asset,
 			membersPublicKeys:
 				// tslint:disable-next-line:no-null-keyword
-				this.membersPublicKeys.length > 0 ? this.membersPublicKeys : null,
+				this.membersPublicKeys.length < 1 ? null : this.membersPublicKeys,
 		};
 	}
 }
--- a/elements/lisk-blocks/src/account.ts
+++ b/elements/lisk-blocks/src/account.ts
@@ -56,8 +56,14 @@ export class Account {
 		this.multiMin = accountInfo.multiMin;
 		this.multiLifetime = accountInfo.multiLifetime;
 		this.asset = accountInfo.asset;
-		this.votedDelegatesPublicKeys = accountInfo.votedDelegatesPublicKeys;
-		this.membersPublicKeys = accountInfo.membersPublicKeys;
+		this.votedDelegatesPublicKeys =
+			accountInfo.votedDelegatesPublicKeys === null
+				? []
+				: accountInfo.votedDelegatesPublicKeys;
+		this.membersPublicKeys =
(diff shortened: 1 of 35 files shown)


elements/lisk-blocks/src/state_store/account_store.ts
```diff
@@ -156,10 +136,17 @@ export class AccountStore {
...
+		const foundAccount = this._data.find(fn);
+		if (!foundAccount) {
+			return undefined;
+		}
+
+		return foundAccount;
 	}
 
 	public set(primaryValue: string, updatedElement: Account): void {
+		const updatedElementObj = new Account(updatedElement);
```
shuse2: if it's setting the account, there is no need to create new instance

elements/lisk-blocks/src/transactions/votes_weight.ts
```diff
@@ -220,7 +221,7 @@ const updateDelegateVotes = async (
 		const delegatePublicKey = vote.slice(1);
 
 		const senderAccount = await stateStore.account.get(transaction.senderId);
-		const amount = BigInt(senderAccount.balance).toString();
+		const amount = senderAccount.balance.toString();
```
shuse2: maybe just use `bigint` here directly?

elements/lisk-blocks/src/types.ts
```diff
@@ -17,26 +17,28 @@ import {
...
-	readonly isDelegate?: number;
-	readonly fees: string;
-	readonly rewards: string;
+	readonly publicKey: string | undefined;
+	readonly secondPublicKey: string | null;
+	readonly secondSignature: number;
+	readonly username: string | null;
+	readonly isDelegate: number;
+	readonly fees: string | bigint;
+	readonly rewards: string | bigint;
```
shuse2:
```suggestion
	readonly rewards: string;
```

elements/lisk-blocks/src/types.ts
```diff
@@ -17,26 +17,28 @@ import {
...
 	// tslint:disable-next-line readonly-keyword
-	voteWeight: string;
-	readonly nameExist: false;
+	voteWeight: string | bigint;
```
shuse2:
```suggestion
	voteWeight: string;
```

elements/lisk-blocks/src/types.ts
```diff
@@ -17,26 +17,28 @@ import {
...
-	readonly username?: string;
-	readonly isDelegate?: number;
-	readonly fees: string;
-	readonly rewards: string;
+	readonly publicKey: string | undefined;
+	readonly secondPublicKey: string | null;
+	readonly secondSignature: number;
+	readonly username: string | null;
+	readonly isDelegate: number;
+	readonly fees: string | bigint;
```
shuse2:
```suggestion
	readonly fees: string;
```

elements/lisk-blocks/src/types.ts
```diff
@@ -17,26 +17,28 @@ import {
 	TransactionResponse,
 } from '@liskhq/lisk-transactions';
 
-export interface Account {
+export interface AccountJSON {
 	readonly address: string;
-	readonly balance: string;
+	readonly balance: string | bigint;
```
shuse2:
```suggestion
	readonly balance: string;
```

elements/lisk-blocks/src/account.ts
```diff
@@ -0,0 +1,130 @@
...
+	rewards: BigInt(0),
+	voteWeight: BigInt(0),
+	nameExist: false,
+	multiMin: 0,
+	multiLifetime: 0,
+	votedDelegatesPublicKeys: [],
+	asset: {},
+	membersPublicKeys: [],
+};
+export class Account {
```
ManuGowda: space please

elements/lisk-blocks/test/unit/process.spec.ts
```diff
@@ -753,8 +753,8 @@ describe('blocks/header', () => {
 			it('should update vote weight on voted delegate', async () => {
 				const delegateOne = await stateStore.account.get(delegate1.address);
 				const deletateTwo = await stateStore.account.get(delegate2.address);
-				expect(delegateOne.voteWeight).toBe('9889999900');
-				expect(deletateTwo.voteWeight).toBe('9889999900');
+				expect(delegateOne.voteWeight?.toString()).toBe('9889999900');
```
ManuGowda: since the value we expect is to be bigint, shouldn't the test expectation change to `BigInt('9889999900')`? same for below assertions

elements/lisk-blocks/src/state_store/account_store.ts
ishantiw: Sometimes the object coming from the framework is not an Account object when we are running functional tests so I will put a condition to cover that

elements/lisk-blocks/src/state_store/account_store.ts
```diff
@@ -75,7 +75,7 @@ export class AccountStore {
 		);
 
 		if (element) {
-			return cloneDeep(element);
+			return new Account(element.toJSON());
```
ManuGowda: it might not be relevant to this PR but the variable name `element` sounds weird, it's an account?

ishantiw: we can do it in a separate PR as it also requires to be consistent with other stateStore.

PR #4790, opened 2020-02-03
Update framework persistence reads - Closes #4770

### LINKED ISSUE #4770: Update framework node to use data access for read operation from blocks library
### Description
Framework node uses storage for reading from the database, in order to consolidate all the database operations, the node should use blocks library for reading operations.
### Motivation
The motivation is described here https://github.com/LiskHQ/lisk-sdk/issues/4709
### Acceptance Criteria
- Replace all the database read operation in the framework to use data access from blocks library

mitsuaki-u:
### What was the problem?

Framework layer reads directly from storage.

### How did I solve it?

Update storage usage to use data access from lisk-blocks.

### How to manually test it?

npm run test

### Review checklist

- [ ] The PR resolves #4770 


framework/src/application/node/forger/forger.js
```diff
@@ -124,11 +122,11 @@ class Forger {
 			throw new Error('Invalid password and public key combination');
 		}
 
-		const filters = {
-			address: getAddressFromPublicKey(keypair.publicKey.toString('hex')),
-		};
+		const address = getAddressFromPublicKey(keypair.publicKey.toString('hex'));
 
-		const [account] = await this.storage.entities.Account.get(filters);
+		const [account] = await this.blocksModule.dataAccess.getAccountsByAddress(
```
shuse2: can we use the single one `getAccountByAddress`?

framework/src/application/node/forger/forger.js
```diff
@@ -191,11 +189,13 @@ class Forger {
 				);
 			}
 
-			const filters = {
-				address: getAddressFromPublicKey(keypair.publicKey.toString('hex')),
-			};
+			const address = getAddressFromPublicKey(
+				keypair.publicKey.toString('hex'),
+			);
+			const [account] = await this.blocksModule.dataAccess.getAccountsByAddress(
```
shuse2: same as above

framework/src/application/node/node.js
```diff
@@ -333,10 +332,11 @@ module.exports = class Node {
 
 		this.slots = this.blocks.slots;
 		this.dpos = new Dpos({
-			blocks: this.blocks,
+			logger: this.logger,
```
shuse2: no need to inject logger

framework/src/application/node/rebuilder.js
```diff
@@ -23,36 +23,36 @@ class Rebuilder {
...
 		this.logger = logger;
-		this.storage = storage;
 		this.genesisBlock = genesisBlock;
 
 		this.processorModule = processorModule;
 		this.blocksModule = blocksModule;
 		this.constants = {
 			activeDelegates,
 		};
+		this.storage = storage;
```
shuse2: we can remove storage?

framework/src/application/node/rebuilder.js
```diff
@@ -80,7 +80,8 @@ class Rebuilder {
 		const targetHeight = targetRound * this.constants.activeDelegates;
 
 		const limit = loadPerIteration;
-		await this.blocksModule.resetState();
+		this.blocksModule.resetBlockHeaderCache();
```
shuse2: i think `resetState` takes care both of these lines

framework/test/jest/unit/specs/application/node/synchronizer/block_synchronization_mechanism/block_synchronization_mechanism.spec.js
```diff
@@ -1156,22 +1133,20 @@ describe('block_synchronization_mechanism', () => {
 
 	describe('isValidFor', () => {
 		it('should return true if the difference in block slots between the current block slot and the finalized block slot of the system is bigger than ACTIVE_DELEGATES*3', async () => {
-			when(storageMock.entities.Block.getOne)
+			when(storageMock.entities.Block.get)
```
shuse2: this also should be changed to blocks mock?

framework/test/jest/unit/specs/application/node/synchronizer/block_synchronization_mechanism/block_synchronization_mechanism.spec.js
```diff
@@ -1156,22 +1133,20 @@ describe('block_synchronization_mechanism', () => {
...
 			expect(isValid).toBeTruthy();
 		});
 
 		it('should return false if the difference in block slots between the current block slot and the finalized block slot of the system is smaller than ACTIVE_DELEGATES*3', async () => {
-			when(storageMock.entities.Block.getOne)
-				.calledWith({
-					height_eql: bftModule.finalizedHeight,
-				})
-				.mockResolvedValue({ ...genesisBlockDevnet, timestamp: Date.now() });
+			storageMock.entities.Block.get.mockResolvedValue([
```
shuse2: same as above

framework/test/jest/unit/specs/application/node/synchronizer/fast_chain_switching_mechanism/fast_chain_switching_mechanism.spec.js
```diff
@@ -590,16 +590,16 @@ describe('fast_chain_switching_mechanism', () => {
 						},
 					})
 					.mockResolvedValue({ data: highestCommonBlock });
-				when(storageMock.entities.Block.getOne)
+				when(storageMock.entities.Block.get)
```
shuse2: same as above

framework/test/jest/unit/specs/application/node/synchronizer/block_synchronization_mechanism/block_synchronization_mechanism.spec.js
mitsuaki-u: I think we can keep as storage Mock, because in the end the blocks is calling storage.

framework/src/application/node/rebuilder.js
mitsuaki-u: We can't remove it yet because there is a call to `Block.delete`

framework/test/jest/unit/specs/application/node/synchronizer/block_synchronization_mechanism/block_synchronization_mechanism.spec.js
shuse2: I think to separate the concern, and not not affecting the changes from `lisk-blocks`, it's better not to use the storage.
If we keep it, when we change the internal implementation, it will be affected

framework/src/application/node/rebuilder.js
shuse2: ah maybe we can add the function in the `lisk-blocks` too

mitsuaki-u: Will we do writes in separate PR?

shuse2: I think this is the only one that is missing. all other things are already solved

framework/test/jest/unit/specs/application/node/synchronizer/block_synchronization_mechanism/block_synchronization_mechanism.spec.js
shuse2: Since this will introduce huge change, we will create a separate issue to address this

framework/src/application/node/forger/forger.js
```diff
@@ -124,11 +122,11 @@ class Forger {
 			throw new Error('Invalid password and public key combination');
 		}
 
-		const filters = {
-			address: getAddressFromPublicKey(keypair.publicKey.toString('hex')),
-		};
+		const address = getAddressFromPublicKey(keypair.publicKey.toString('hex'));
 
-		const [account] = await this.storage.entities.Account.get(filters);
+		const account = await this.blocksModule.dataAccess.getAccountByAddress(
```
ManuGowda: can we not use publicKey to get the account? and publicKey is also indexed
```
const account = await this.blocksModule.dataAccess.getAccountsByPublicKey([keypair.publicKey.toString('hex')])
```

The reason is we could avoid an extra call to get the address first `getAddressFromPublicKey` from above line and directly use `getAccountsByPublicKey`.

framework/src/application/node/forger/forger.js
```diff
@@ -191,11 +189,13 @@ class Forger {
 				);
 			}
 
-			const filters = {
-				address: getAddressFromPublicKey(keypair.publicKey.toString('hex')),
-			};
+			const address = getAddressFromPublicKey(
```
ManuGowda: same as above comment.

framework/src/application/node/node.js
```diff
@@ -96,15 +91,19 @@ module.exports = class Node {
...
 			// Prepare dependency
 			const processorDependencies = {
 				blocksModule: this.blocks,
 				bftModule: this.bft,
 				dposModule: this.dpos,
-				storage: this.storage,
 				logger: this.logger,
 				constants: this.options.constants,
 				exceptions: this.options.exceptions,
+				storage: this.storage,
```
ManuGowda: I hope after the conclusion with you will remove this dependency as well, correct?

framework/src/application/node/synchronizer/block_synchronization_mechanism.js
```diff
@@ -175,10 +174,8 @@ class BlockSynchronizationMechanism extends BaseSynchronizer {
 	async _handleBlockProcessingError(lastCommonBlock, peerId) {
 		// If the list of blocks has not been fully applied
 		this.logger.debug('Failed to apply obtained blocks from peer');
-		const [tipBeforeApplying] = await this.storage.entities.TempBlock.get(
-			{},
-			{ sort: 'height:desc', limit: 1, extended: true },
-		);
+		const tempBlocks = await this.blocks.dataAccess.getTempBlocks();
+		const [tipBeforeApplying] = tempBlocks.sort((a, b) => b - a);
```
ManuGowda: maybe not to mutate tempBlocks.. we could do `[...tempBlocks].sort((a, b) => b - a);`?

framework/src/application/node/synchronizer/block_synchronization_mechanism.js
```diff
@@ -369,17 +366,9 @@ class BlockSynchronizationMechanism extends BaseSynchronizer {
...
-					{
-						height_in: heightList,
-					},
-					{
-						sort: 'height:asc',
-						limit: heightList.length,
-					},
-				)
-			).map(block => block.id);
+			const blockIds = await this.blocks.dataAccess.getBlockHeadersWithHeights(
```
ManuGowda: the result is not blockIds so can we use the variable name as `blockHeaders`?

framework/src/application/node/synchronizer/utils.js
```diff
@@ -40,8 +40,8 @@ const restoreBlocks = async (blocksModule, processorModule, tx = null) => {
 	return true;
 };
 
-const clearBlocksTempTable = storageModule =>
-	storageModule.entities.TempBlock.truncate();
+const clearBlocksTempTable = blocksModule =>
```
ManuGowda: can we remove this and directly use `blocksModule.dataAccess.clearTempBlocks`? it's not reused as well

ManuGowda:
@mitsuaki-u I still see the usage if storage in `block_processor_v2`
```
const previouslyForgedStr = await this.storage.entities.ForgerInfo.getKey(
			FORGER_INFO_KEY_MAX_HEIGHT_PREVIOUSLY_FORGED,
		);
```
we indent to move this logic to dataAccess? or use storage?

mitsuaki-u:
> @mitsuaki-u I still see the usage if storage in `block_processor_v2`
> 
> ```
> const previouslyForgedStr = await this.storage.entities.ForgerInfo.getKey(
> 			FORGER_INFO_KEY_MAX_HEIGHT_PREVIOUSLY_FORGED,
> 		);
> ```
> 
> we indent to move this logic to dataAccess? or use storage?



> @mitsuaki-u I still see the usage if storage in `block_processor_v2`
> 
> ```
> const previouslyForgedStr = await this.storage.entities.ForgerInfo.getKey(
> 			FORGER_INFO_KEY_MAX_HEIGHT_PREVIOUSLY_FORGED,
> 		);
> ```
> 
> we indent to move this logic to dataAccess? or use storage?

Yes I asked about this earlier but it was decided we only do Accounts and Blocks for now.

shuse2:
@ManuGowda @mitsuaki-u 
Forger info is not context of `blocks`, so it should keep this as it is

framework/src/application/node/node.js
mitsuaki-u: We still need it, otherwise it breaks.

framework/src/application/node/synchronizer/utils.js
mitsuaki-u: That was the original function so I will keep it for now.

PR #4794, opened 2020-02-03
Update to use blocks and dpos in BFT - Closes #4737

### LINKED ISSUE #4737: Update BFT to use blocks instance to obtain block header data
### Description

Currently, BFT is maintaining own cache mechanism.
However, after #4318, it should consistently use the `blocks` instance from `lisk-blocks`

### Motivation

It should always use consistent data source

### Acceptance Criteria

- https://github.com/LiskHQ/lisk-sdk/blob/development/elements/lisk-bft/src/headers_list.ts is removed
- all of the access to block header is done through blocks library

shuse2:
### What was the problem?
The PR resolves #4467 

### How did I solve it?
- Remove direct storage dependency
- Remove required input of min active height of each delegate
- Remove cache management from bft
- consistently use the same property for block header

### How to manually test it?
- All application should work as it is


elements/lisk-bft/src/utils.ts
```diff
@@ -15,25 +15,11 @@
 import { validator } from '@liskhq/lisk-validator';
 
 import { blockHeaderSchema } from './schema';
-import { Block, BlockHeader } from './types';
+import { BlockHeader } from './types';
 
-export const validateBlockHeader = (blockHeader: BlockHeader): boolean => {
+export const validateBlockHeader = (blockHeader: BlockHeader): void => {
```
ishantiw: I think this function is just wrapper around `validator.validate` maybe its better to use it directly instead of adding as a util func

PR #4803, opened 2020-02-06
Fix BFT delete blocks logic - Closes #4800

### LINKED ISSUE #4800: BFT deletion logic throws error which is not error
### Expected behavior
* It should not fail to delete blocks more than 197 blocks.
* It should recompute after filling the cache
* Update `prevotedConfirmedHeight` to `chainMaxHeightFinalized` in order to be compatible with LIP
* Include `height`, `heightPrevoted` and `maxHeightPreviouslyForged` in the forgers list

### Actual behavior
When deleting blocks which makes the block header cache to go below, it fails to refill the cache because of the `chainMaxHeightFinalized` validation.

### Steps to reproduce
Delete more than 197 blocks when it has 500 blocks cache

3.0-

shuse2:
### What was the problem?
This PR resolves #4800 

### How was it solved?
- Update the saved forger info to be consistent with LIP
- Update variable name to be consistent with LIP
- delete block, but only populate the cache first and recompute without validation

### How was it tested?
- Added test case to check the recompute
- Check database to contain correct forger info

### DIFF
--- a/framework/src/modules/chain/bft/bft.js
+++ b/framework/src/modules/chain/bft/bft.js
@@ -118,15 +118,8 @@ class BFT extends EventEmitter {
 		this.finalityManager.removeBlockHeaders({
 			aboveHeight: removeFromHeight - 1,
 		});
-
 		// Make sure there are BFT_ROUND_THRESHOLD rounds of block headers available
-		if (
-			this.finalityManager.maxHeight - this.finalityManager.minHeight <
-			this.constants.activeDelegates * BFT_ROUND_THRESHOLD
-		) {
-			await this._fillCache(minActiveHeightsOfDelegates);
-			this.finalityManager.recompute();
-		}
+		await this._fillCache(minActiveHeightsOfDelegates);
 	}
 
 	addNewBlock(block, stateStore) {
@@ -334,6 +327,13 @@ class BFT extends EventEmitter {
 	}
 
 	async _fillCache(minActiveHeightsOfDelegates) {
+		if (
+			this.finalityManager.maxHeight - this.finalityManager.minHeight >=
+			this.constants.activeDelegates * BFT_ROUND_THRESHOLD
+		) {
+			return;
+		}
+
 		const tillHeight = this.finalityManager.minHeight - 1;
 		const fromHeight =
 			this.finalityManager.maxHeight -
@@ -382,6 +382,7 @@ class BFT extends EventEmitter {
 				extractBFTBlockHeaderFromBlock(blockHeader),
 			);
 		}
+		this.finalityManager.recompute();
 	}
 }
 
--- a/framework/src/modules/chain/bft/bft.js
+++ b/framework/src/modules/chain/bft/bft.js
@@ -325,7 +325,7 @@ class BFT extends EventEmitter {
 	}
 
 	get maxHeightPrevoted() {
-		return this.finalityManager.chainMaxHeightFinalized;
+		return this.finalityManager.chainMaxHeightPrevoted;
 	}
 
 	reset() {
--- a/framework/src/modules/chain/bft/bft.js
+++ b/framework/src/modules/chain/bft/bft.js
@@ -124,17 +124,8 @@ class BFT extends EventEmitter {
 			this.finalityManager.maxHeight - this.finalityManager.minHeight <
 			this.constants.activeDelegates * BFT_ROUND_THRESHOLD
 		) {
-			const tillHeight = this.finalityManager.minHeight - 1;
-			const fromHeight =
-				this.finalityManager.maxHeight -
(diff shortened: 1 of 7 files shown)


framework/src/modules/chain/bft/bft.js
```diff
@@ -334,13 +325,64 @@ class BFT extends EventEmitter {
...
+				height => height >= minimumPossibleActiveHeight,
+			);
+
+			const blockHeader = {
+				...blockJSON,
+				delegateMinHeightActive,
+			};
+			this.finalityManager.headers.add(
+				extractBFTBlockHeaderFromBlock(blockHeader),
+			);
```
ManuGowda: can we abstract this in a function and reuse in `_loadBlocksFromStorage` as well?

framework/src/modules/chain/bft/finality_manager.js
```diff
@@ -58,7 +58,7 @@ class FinalityManager extends EventEmitter {
 		this.finalizedHeight = finalizedHeight;
 
 		// Height up to which blocks have pre-voted
-		this.prevotedConfirmedHeight = 0;
+		this.chainMaxHeightFinalized = 0;
```
nazarhussain: `this.prevotedConfirmedHeight` are different than `this.chainMaxHeightFinalized`

framework/src/modules/chain/bft/finality_manager.js
```diff
@@ -88,7 +88,7 @@ class FinalityManager extends EventEmitter {
 
 		debug('after adding block header', {
 			finalizedHeight: this.finalizedHeight,
-			prevotedConfirmedHeight: this.prevotedConfirmedHeight,
+			chainMaxHeightFinalized: this.chainMaxHeightFinalized,
```
nazarhussain: Its duplicate information both `finalizedHeight` and `chainMaxHeightFinalized` are the same attribute.

framework/src/modules/chain/bft/finality_manager.js
```diff
@@ -191,9 +191,9 @@ class FinalityManager extends EventEmitter {
 			.reverse()
 			.find(key => this.preVotes[key] >= this.preVoteThreshold);
 
-		this.prevotedConfirmedHeight = highestHeightPreVoted
+		this.chainMaxHeightFinalized = highestHeightPreVoted
 			? parseInt(highestHeightPreVoted, 10)
-			: this.prevotedConfirmedHeight;
+			: this.chainMaxHeightFinalized;
```
nazarhussain: Block with highest prevotes is not the finalized height.

framework/src/modules/chain/block_processor_v2.js
```diff
@@ -27,8 +27,7 @@ const {
 const { baseBlockSchema } = require('./blocks');
 const { BaseBlockProcessor } = require('./processor');
 
-const FORGER_INFO_KEY_MAX_HEIGHT_PREVIOUSLY_FORGED =
-	'maxHeightPreviouslyForged';
+const FORGER_INFO_KEY_MAX_HEIGHT_PREVIOUSLY_FORGED = 'previouslyForged';
```
ManuGowda: can you update the variable to the corresponding key

framework/src/modules/chain/bft/finality_manager.js
shuse2: It was meant to be `chainMaxHeightPrevoted`, will update.
This was because when we were discussing the bug, it was hard to communicate/reference the LIP

framework/src/modules/chain/bft/bft.js
shuse2: it's slightly different functionality, and for the patch, i don't want to clean up and change so much.
On my other PR, it will be cleaned up

framework/src/modules/chain/bft/bft.js
```diff
@@ -124,17 +124,8 @@ class BFT extends EventEmitter {
 			this.finalityManager.maxHeight - this.finalityManager.minHeight <
 			this.constants.activeDelegates * BFT_ROUND_THRESHOLD
 		) {
-			const tillHeight = this.finalityManager.minHeight - 1;
-			const fromHeight =
-				this.finalityManager.maxHeight -
-				// Search is inclusive, therefore, it should start from one above (ex: 288 - 500, which results total 303)
-				this.constants.activeDelegates * BFT_ROUND_THRESHOLD +
-				1;
-			await this._loadBlocksFromStorage({
```
nazarhussain: Why we need to keep this function still int he file? actually that function should have been renamed to `fillCache`

shuse2: @ManuGowda Also mentioned this, but this logic is slightly different than fill cache because it fills to the top, and this is patch so i didn't want to clean up. This will be changed in my BFT PR #4737

PR #4804, opened 2020-02-06
Locale should be specified when using toLocaleString - Closes #4747

### LINKED ISSUE #4747: Locale should be specified when using toLocaleString
### Expected behavior
`toLocaleString` should be set to `en-US` in order to avoid having different format (ex: having `,` in the middle)

### Actual behavior
It's set to `undefined` currently

4.0-

pablitovicente:
### What was the problem?
This PR resolves #4747

### How was it solved?
By specifying the locale `en-US`

### How was it tested?
By running tests

### DIFF
--- a/elements/lisk-transactions/src/utils/format.ts
+++ b/elements/lisk-transactions/src/utils/format.ts
@@ -35,7 +35,7 @@ export const convertBeddowsToLSK = (beddowsAmount?: string): string => {
 	const floating =
 		Number(beddowsAmountBigInt % BigInt(FIXED_POINT)) / FIXED_POINT;
 	const floatingPointsSplit = floating
-		.toLocaleString(undefined, {
+		.toLocaleString('en-US', {
 			maximumFractionDigits: LISK_MAX_DECIMAL_POINTS,
 		})
 		.split('.')[1];


elements/lisk-transactions/src/utils/format.ts
```diff
@@ -35,7 +35,7 @@ export const convertBeddowsToLSK = (beddowsAmount?: string): string => {
 	const floating =
 		Number(beddowsAmountBigInt % BigInt(FIXED_POINT)) / FIXED_POINT;
 	const floatingPointsSplit = floating
-		.toLocaleString(undefined, {
+		.toLocaleString('en-US', {
```
ManuGowda: @shuse2 could you give a scenario where this could make a difference, i tried this and i dint make any difference with undefined
```
(1.2e-7).toLocaleString('en-US', {maximumFractionDigits: 8}) === (1.2e-7).toLocaleString(undefined, {maximumFractionDigits: 8})
```
I was trying to think a test scenario for this fix.

shuse2: I think it's pretty hard to change that behavior. It is looking at local locale, and it is the node.js internal behavior. 
As long as `en-US` works, it should be ok IMO

ManuGowda: ok, I was trying to understand the problem through a test, ok no problem 👍

PR #4805, opened 2020-02-06
Remove istanbul-middleware from the application - Closes #2919

### LINKED ISSUE #2919: Remove istanbul-middleware from the application
### Expected behavior
We should come up with a better approach to collect code coverage from functional tests. 

We should also discuss if it's needed or does it give any value.

If it's really needed, maybe at least we could strip that code piece from production build.

### Actual behavior

We are using `istanbul-middleware` in the application source code.

pablitovicente:
### What was the problem?
This PR resolves #2919

### How was it solved?
By removing the middleware

### How was it tested?
Running tests


framework/test/mocha/unit/modules/http_api/init_steps/setup_servers.js
```diff
@@ -52,7 +51,7 @@ describe('init_steps/setup_servers', () => {
 	};
 
 	beforeEach(async () => {
-		sinonSandbox.stub(im);
+		sinonSandbox.stub();
```
shuse2: we can remove this line

PR #4807, opened 2020-02-07
Remove getFirstBlockIdOfLastRounds and SQL file - Closes #4741

### LINKED ISSUE #4741: Create lisk-blocks data access layer function for `getFirstBlockIdOfLastRounds`
### Expected behavior

There should be a function to replicate `getFirstBlockIdOfLastRounds` function in `lisk-blocks`.

### Actual behavior

Currently it uses a SQL file to execute query. This logic should be duplicated in `storage_access`.

### Steps to reproduce
n/a

4.0

mitsuaki-u:
### What was the problem?

This PR resolves #4741

Removal of `loader` module removes necessity of `getFirstBlockIdOfLastRounds` functionality.

### How was it solved?

Remove all usages of `getFirstBlockIdOfLastRounds` and SQL files.

### How was it tested?

npm t


elements/lisk-chain/src/data_access/data_access.ts
```diff
@@ -19,7 +19,6 @@ import {
 	BlockHeaderJSON,
 	BlockInstance,
 	BlockJSON,
-	BlockRound,
```
shuse2: I think we can also remove this `BlockRound` type

PR #4811, opened 2020-02-10
Fix wrong min precommit height and overwrite of finality - Closes #4808

### LINKED ISSUE #4808: BFT properties calculation does not follow the LIP
### Expected behavior
- Finality should not be reset to the initial value while recomputing
- `_getValidMinHeightToCommit` should increment needle height (searching cursor) every time
- it should not throw error when looking for `_getValidMinHeightToCommit`

### Actual behavior
- Finality is reset to original value
- Needle height is only incremented if it doesn't match previously forged height
- if there is no cache, it fails to get `_getValidMinHeightToCommit`

shuse2:
### What was the problem?

This PR resolves #4808 

### How was it solved?
- Remove resetting the finality when recomputing
- Fix not to throw error when cache doesn't exist
- Fix the logic to for height count down

### How was it tested?
- it should not affect the property calculation on the success scenario
- Removed the cache manually (not initialize enough in the code, and observe it doesn't throw)

### DIFF
--- a/framework/src/modules/chain/bft/finality_manager.js
+++ b/framework/src/modules/chain/bft/finality_manager.js
@@ -128,14 +128,16 @@ class FinalityManager extends EventEmitter {
 			maxPreCommitHeight: 0,
 		};
 
-		const heightNotPrevoted = this._getHeightNotPrevoted(header);
+		const minValidHeightToPreCommit = this._getMinValidHeightToPreCommit(
+			header,
+		);
 
 		// If delegate is new then first block of the round will be considered
 		// if it forged before then we probably have the last commit height
 		// delegate can't pre-commit a block before the above mentioned conditions
 		const minPreCommitHeight = Math.max(
 			header.delegateMinHeightActive,
-			heightNotPrevoted + 1,
+			minValidHeightToPreCommit,
 			delegateState.maxPreCommitHeight + 1,
 		);
 
@@ -211,7 +213,7 @@ class FinalityManager extends EventEmitter {
 		return true;
 	}
 
-	_getHeightNotPrevoted(header) {
+	_getMinValidHeightToPreCommit(header) {
 		// We search backward from top block to bottom block in the chain
 
 		// We should search down to the height we have in our headers list
@@ -238,7 +240,7 @@ class FinalityManager extends EventEmitter {
 				const previousBlockHeader = this.headers.get(needleHeight);
 				if (!previousBlockHeader) {
 					debug('Fail to get cached block header');
-					return -1;
+					return 0;
 				}
 
 				// Was the previous block suggested by current block header
@@ -248,14 +250,15 @@ class FinalityManager extends EventEmitter {
 					previousBlockHeader.delegatePublicKey !== header.delegatePublicKey ||
 					previousBlockHeader.maxHeightPreviouslyForged >= needleHeight
 				) {
-					return needleHeight;
+					return needleHeight + 1;
 				}
 				// Move the needle to previous block and consider it current for next iteration
+				// needleHeight = previousBlockHeader.maxHeightPreviouslyForged;
 				currentBlockHeader = previousBlockHeader;
 			}
 			needleHeight -= 1;
 		}
-		return needleHeight;
+		return needleHeight + 1;
 	}
 
 	recompute() {
--- a/framework/src/modules/chain/bft/finality_manager.js
+++ b/framework/src/modules/chain/bft/finality_manager.js
@@ -137,8 +137,9 @@ class FinalityManager extends EventEmitter {
 		// delegate can't pre-commit a block before the above mentioned conditions
 		const minPreCommitHeight = Math.max(
 			header.delegateMinHeightActive,
-			validMinHeightToVoteAndCommit,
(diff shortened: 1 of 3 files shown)


framework/src/modules/chain/bft/finality_manager.js
```diff
@@ -248,20 +253,16 @@ class FinalityManager extends EventEmitter {
 				) {
 					return needleHeight;
 				}
-
 				// Move the needle to previous block and consider it current for next iteration
-				needleHeight = previousBlockHeader.maxHeightPreviouslyForged;
```
nazarhussain: This line was doing the jump to previous block header of that delegate. That's a required behaviour as we don't need to iterate linearly for each block header.

framework/src/modules/chain/bft/headers_list.js
```diff
@@ -126,6 +126,12 @@ class HeadersList {
 	}
 
 	get(height) {
+		if (
+			this.length === 0 ||
+			this._items[height - this.first.height] === undefined
+		) {
+			return undefined;
+		}
```
nazarhussain: That's the default behaviour of array. If index does not exists it will return `undefined`. That's why additional conditions not required here.

framework/src/modules/chain/bft/finality_manager.js
```diff
@@ -248,20 +253,16 @@ class FinalityManager extends EventEmitter {
...
-				needleHeight -= 1;
 			}
+			needleHeight -= 1;
 		}
 		return needleHeight;
 	}
 
 	recompute() {
 		this.state = {};
-		this.finalizedHeight = this._initialFinalizedHeight;
```
nazarhussain: Recomputation meant to re-compute everything, including the finalized height. So what's the synopsis for not resetting its value?

framework/src/modules/chain/bft/finality_manager.js
```diff
@@ -238,6 +239,10 @@ class FinalityManager extends EventEmitter {
 			// maxHeightPreviouslyForged always refers to a height with a block forged by the same delegate.
 			if (needleHeight === currentBlockHeader.maxHeightPreviouslyForged) {
 				const previousBlockHeader = this.headers.get(needleHeight);
+				if (!previousBlockHeader) {
+					debug('Fail to get cached block header');
+					return -1;
+				}
```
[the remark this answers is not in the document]
shuse2: When it resets, it calculates from the lowest one, which might not include the `maxHeightPreviouslyForged` block header, which should be ignored for the preCommitMin.
That's why I added there

framework/src/modules/chain/bft/headers_list.js
shuse2: if `this.first` is `undefined`, it will crash (which should not happen, but it happened on one of the node)

framework/src/modules/chain/bft/finality_manager.js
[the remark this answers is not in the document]
nazarhussain: After reset the list must have `height - processingThreshold`, otherwise all calculations would went wrong, not just on this place.

framework/src/modules/chain/bft/headers_list.js
nazarhussain: Can we simply the condition in that case based with simple; 

```
if (this.length === 0) return undefined; 
```

framework/src/modules/chain/bft/finality_manager.js
```diff
@@ -137,8 +137,9 @@ class FinalityManager extends EventEmitter {
 		// delegate can't pre-commit a block before the above mentioned conditions
 		const minPreCommitHeight = Math.max(
 			header.delegateMinHeightActive,
-			validMinHeightToVoteAndCommit,
+			validMinHeightToVoteAndCommit + 1,
```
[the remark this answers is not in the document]
shuse2:
```
   const heightNotPrevoted = getHeightNotPrevoted(newBlockheader)

   // Add implied precommits by newBlockheader
   const minPrecommitHeight = Math.max(delegateMinHeightActive,
                                       heightNotPrevoted+1,
                                       largestHeightPrecommit+1)
```
directly from LIP it's taken 😉

framework/src/modules/chain/bft/finality_manager.js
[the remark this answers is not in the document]
shuse2: height - processingThreshold refers to the highest height, but in lower-end of the cache, the maxHeightPreviouslyForged might not exist.
ie)
cache height 500 - 297, and 297 refers to the older block than this

framework/src/modules/chain/bft/finality_manager.js
shuse2: Finalized height cannot be recomputed in all the case (from the LIP), therefore, we need to store it in the chain state

framework/src/modules/chain/bft/finality_manager.js
shuse2: Does it work in case of conflicting block header?

nazarhussain: As far as block at `previousBlockHeader.maxHeightPreviouslyForged` is valid block forged by same delegate, it should work.

PR #4816, opened 2020-02-10
pass an array of ip instead of an array of entry

JesusTheHun:
Once the config.json is parsed it is transformed into an entry : 

```typescript
type WhiteListEntry = {
  ip: string;
  wsPort: number;
}
```

And then injected into the config tree.
When this config is used, it is passed as is. Unfortunately the function `CheckIpInList`expects an array of string, not an array of object, resulting into an exception and therefore blocking the legitimate requests.

This PR fixes the passed values.

### DIFF
--- a/framework/src/modules/http_api/controllers/node.js
+++ b/framework/src/modules/http_api/controllers/node.js
@@ -299,7 +299,10 @@ NodeController.getStatus = async (context, next) => {
  */
 NodeController.getForgingStatus = async (context, next) => {
 	if (
-		!checkIpInList(library.config.forging.access.whiteList, context.request.ip)
+		!checkIpInList(
+			library.config.forging.access.whiteList.map(e => e.ip),
+			context.request.ip,
+		)
 	) {
 		context.statusCode = apiCodes.FORBIDDEN;
 		return next(new Error('Access Denied'));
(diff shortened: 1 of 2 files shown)


JesusTheHun:
After a bit more testing I've noticed that loading the `config.json` does not require an object and therefore pass a `Array<string>` instead of an `Array<WhiteListEntry>`.

So we should either edit the `CheckIpInList` to test if it's an `WhiteListEntry` or a `string` (quick & dirty), or make sure the loaded config is a `WhiteListEntry` (more robust)

nazarhussain:
https://github.com/LiskHQ/lisk-sdk/blob/1fb235e5988d330b068c78ef71b26eed5acc3cd6/framework/src/modules/http_api/defaults/config.js#L143-L144

If loading from a config file you can specify an array of IPs. But while loading through env variable parser assuming it would be ip, port set, which is wrong logic. 
https://github.com/LiskHQ/lisk-sdk/blob/1fb235e5988d330b068c78ef71b26eed5acc3cd6/framework/src/modules/http_api/defaults/config.js#L145-L148
https://github.com/LiskHQ/lisk-sdk/blob/1fb235e5988d330b068c78ef71b26eed5acc3cd6/framework/src/controller/validator/keywords/formatters.js#L18

That issue is already fixed in 3.0+ version. 

https://github.com/LiskHQ/lisk-sdk/blob/d15f8a0279be1ec400a9859c5ef0e24c9bb69655/framework/src/modules/http_api/defaults/config.js#L145-L148

PR #4817, opened 2020-02-10
pass an array of ip instead of an array of entry SDK 3.x

JesusTheHun:
Same as https://github.com/LiskHQ/lisk-sdk/pull/4816 but for the SDK 3.x

### DIFF
--- a/framework/src/modules/http_api/controllers/node.js
+++ b/framework/src/modules/http_api/controllers/node.js
@@ -128,7 +128,10 @@ NodeController.getStatus = async (context, next) => {
 
 NodeController.getForgingStatus = async (context, next) => {
 	if (
-		!checkIpInList(library.config.forging.access.whiteList, context.request.ip)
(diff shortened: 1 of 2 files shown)


nazarhussain:
For 3+ version access whitelist and forging access whitelist is already considered as array of IPs without port. 

https://github.com/LiskHQ/lisk-sdk/blob/0f5bef7d2007d999b1526db6130938c4cf57c416/framework/src/modules/http_api/defaults/config.js#L143-L148

https://github.com/LiskHQ/lisk-sdk/blob/0f5bef7d2007d999b1526db6130938c4cf57c416/framework/src/modules/http_api/defaults/config.js#L46-L51

https://github.com/LiskHQ/lisk-sdk/blob/0f5bef7d2007d999b1526db6130938c4cf57c416/framework/src/application/validator/keywords/formatters.js#L31-L35

So the issue is already fixed for 3+ version.

PR #4819, opened 2020-02-11
P2P Worker thread init PoC

valamidev:
### What was the problem?
- P2P server was not utilising multi process.

### How was it solved?
- PeerServer implementation extended with a single Worker_thread  
 (https://nodejs.org/api/worker_threads.html), which aimed to handle the SocketCluster Server and all the inbound connections. 

### How was it tested?

- Integration and Functional test should be able to run in same process.
`npm run test`


shuse2:
Closing for now as this is PoC, and the branch will be kept

PR #4820, opened 2020-02-11
Replace lodash usage in lisk-chain - Closes #4810

### LINKED ISSUE #4810: Remove lodash dependency from chain library
### Description

All-in-one Lodash dependency from the chain library should be removed.
We can keep the `cloneDeep`, but other usage should be replaced by other implementation

### Motivation

We want to keep dependency to the other package to the minimum level

### Acceptance Criteria

- lodash.cloneDeep, is installed separately
- All other usage of lodash should be reoved

mitsuaki-u:
### What was the problem?

`lodash` usage should be limited in `elements` due to security.

This PR resolves #4810 

### How was it solved?

Replaced some `lodash` usage with native functionality. 

### How was it tested?

npm run test


elements/lisk-chain/src/state_store/account_store.ts
```diff
@@ -11,8 +11,10 @@
  *
  * Removal or modification of this copyright notice is prohibited.
  */
-
-import { cloneDeep, isEqual, pick, uniq, uniqBy } from 'lodash';
+// tslint:disable-next-line no-require-imports
+import cloneDeep = require('lodash.clonedeep');
+// tslint:disable-next-line no-require-imports
+import isEqual = require('lodash.isequal');
```
shuse2: it should not use `require`

elements/lisk-chain/src/state_store/transaction_store.ts
```diff
@@ -11,11 +11,12 @@
  *
  * Removal or modification of this copyright notice is prohibited.
  */
-
 import { TransactionJSON } from '@liskhq/lisk-transactions';
-import { clone, cloneDeep, uniqBy } from 'lodash';
+// tslint:disable-next-line no-require-imports
+import cloneDeep = require('lodash.clonedeep');
```
shuse2: same as above

elements/lisk-chain/src/utils.ts
```diff
@@ -27,3 +27,13 @@ export const getId = (blockBytes: Buffer): string => {
 
 	return id;
 };
+
+export const uniqBy = (
```
shuse2: I think instead of making it generic, we can just check in each case. Mostly checking if the id or address exists in the cache, it uses it

elements/lisk-chain/src/state_store/account_store.ts
```diff
@@ -185,7 +188,13 @@ export class AccountStore {
...
 			async ({ updatedItem, updatedKeys }) => {
 				const filter = { [this._primaryKey]: updatedItem[this._primaryKey] };
-				const updatedData = pick(updatedItem, updatedKeys);
+				const updatedData = updatedKeys.reduce((data, key) => {
+					// tslint:disable-next-line:no-any
+					data[key] = (updatedItem as any)[key];
+
+					return data;
+					// tslint:disable-next-line:no-any
+				}, {} as any);
```
shuse2: prefer no `no-any`

elements/lisk-chain/src/state_store/account_store.ts
mitsuaki-u: Since we are not importing whole `lodash` library, we need to use `require` here.

elements/lisk-chain/src/utils.ts
mitsuaki-u: Decision to keep `uniqBy` due to efficiency improvement.

PR #4821, opened 2020-02-11
Merge 3.0.2 development

shuse2:
### What was the problem?

Merging back latest hotfix into development


nazarhussain:
@shuse2 I reviewed in detail explicitly for the BFT changes. Would be nice if for other chances somone else also do a review.

PR #4822, opened 2020-02-12
Block header cached should be refilled if goes over certain threshold - Closes #4793

### LINKED ISSUE #4793: Block header cached should be refilled if goes over certain threshold
### Description

Currently if a block is deleted, there is no re-cache until new block is added, but we want to maintain at least 303 blocks in the cache

### Motivation

In order to compute the BFT header we use 303 bocks header, therefore this amount of cache should be maintained.

### Acceptance Criteria

- threshold should be constructor option but default to 350 (more than 3 rounds for us)
- The minimum block header cache threshold should be updated from 500 to 505

### Additional Information

- Current max block header cache 505 (It is configurable)
- When we delete the blocks from blockchain the block header cache should be refilled (from database) up to the lower bound (minimum lower bound - 303)

**Example: If the block header cache is 505, when we delete 300 block header cache, and when the minimum lower bound is 300 then first 100 block header should be refilled from database**

pablitovicente:
### What was the problem?

This PR resolves #4793

### How was it solved?

- By adding minimum and maximum items options to the cache constructor
- When enough items are removed from the cache and the minimum items value is reached a flag is set
- When the flag is set the data access layer knows it needs to pull data from the DB and push it to the head of the items array

### How was it tested?

- Unit tests


elements/lisk-chain/src/data_access/cache/block.ts
```diff
@@ -18,8 +18,11 @@ import { BlockHeader } from '../../types';
 import { Base } from './base';
 
 export class BlockCache extends Base<BlockHeader> {
-	public constructor(size: number = 500) {
-		super(size);
+	public constructor(
+		minCachedItems: number = 303,
```
shuse2: Since defualt is defined in the `chain.ts` I think this constructor doesn't need the default value

elements/lisk-chain/src/data_access/data_access.ts
```diff
@@ -61,8 +66,29 @@ export class DataAccess {
 		return this._blocksCache.add(blockHeader);
 	}
 
-	public removeBlockHeader(id: string): BlockHeader[] {
-		return this._blocksCache.remove(id);
+	public async removeBlockHeader(id: string): Promise<BlockHeader[]> {
+		const cachedItems = this._blocksCache.remove(id);
+
+		if (this._blocksCache.needsRefill) {
```
shuse2: Prefer early return

elements/lisk-chain/src/data_access/data_access.ts
```diff
@@ -61,8 +66,29 @@ export class DataAccess {
...
+			// The method getBlocksByHeightBetween uses gte & lte so we need to adjust values
+			const upperHeightToFetch = this._blocksCache.items[0]?.height - 1 || 0;
+
+			const lowerHeightToFetch = Math.max(
+				upperHeightToFetch -
+					(this._blocksCache.maxCachedItems - this._blocksCache.minCachedItems),
+				1,
+			);
+
+			const blockHeaders = await this.getBlocksByHeightBetween(
```
shuse2: Before calling this function, it's probably better to check if `upperHeight - lowerHeight > 0` and in that case we don't need to fetch DB

elements/lisk-chain/src/data_access/cache/base.ts
```diff
@@ -48,4 +56,12 @@ export abstract class Base<T> {
 
 		return this.items;
 	}
+
+	public set needsRefill(needRefill: boolean) {
```
shuse2: if we have both setter and getter without mutation, maybe we can just make the property public?

pablitovicente: As discussed keeping as is just for consistency.

elements/lisk-chain/src/data_access/cache/block.ts
pablitovicente: Done.

elements/lisk-chain/src/data_access/data_access.ts
pablitovicente: Done.

elements/lisk-chain/src/data_access/data_access.ts
pablitovicente: Done.

PR #4823, opened 2020-02-12
Improve console logging - Closes #4812

### LINKED ISSUE #4812: Console logger should not require bunyan to parse
### Description

Currently console logging is displayed in the JSON format, but it should be displayed in the human readable way.

### Motivation

Console logging is mostly for development, and in the production, only file logging should be enabled.
Therefore, although it might have slight performance degrade on the console logging, it should be better to display the human readable log.

### Acceptance Criteria

- console log should not require piping `npx bunyan`

shuse2:
### What was the problem?

This PR resolves #4812 

### How was it solved?

Add formatting before writing to stdout

### How was it tested?

Run the application and observe the log


shuse2:
@nazarhussain 
I think it's hacky to use the executable from the application. Let's use bunyan to format only if they provide library API

framework/src/components/logger/logger.js
```diff
@@ -26,6 +27,55 @@ const createDirIfNotExist = filePath => {
...
+	50: setColor('red', 'ERROR'),
+	60: setColor('red', 'FATAL'),
+};
+
+class ConsoleLog {
+	// eslint-disable-next-line
+	write(rec) {
+		try {
+			let log = util.format(
+				'%s [%s] %s %s (module=%s)\n',
```
nazarhussain:
```suggestion
				'%s %s %s: %s (module=%s)\n',
```

framework/src/components/logger/logger.js
```diff
@@ -26,6 +27,55 @@ const createDirIfNotExist = filePath => {
...
+				new Date(rec.time).toISOString(),
```
nazarhussain:
```suggestion
				new Date(rec.time).toLocaleTimeString('en-US', { hour12: false }),
```

shuse2: I think it's more closer to the bunyan, but for log, i think it's better to have the date too

nazarhussain: For a certain session, it will just be the same date repeating every line.

PR #4829, opened 2020-02-13
Add and update yarn, lerna setup - Closes #4828

### LINKED ISSUE #4828: Update to use yarn for the monorepo management
### Description

Update the setting to use yarn for monorepo structure management, and contribution README

### Motivation

Lerna is good to manage versions and commands for the monorepo, but we faced several problems below
- package-lock are in multiple packages, and it often has conflict
- Cannot correctly use audit for the multiple packages
- bootstrapping becomes slower and sometime fails

### Acceptance Criteria

- Yarn can be used for the monorepo management
- package-lock.json should be removed and replaced by yarn.lock

### Additional Information
- Jenkins also needs to be updated

shuse2:
### What was the problem?

This PR resolves #4828 

### How was it solved?
- Remove package-lock json
- Add yarn setup
- Updated readme

### How was it tested?
- Started fresh repo, and check if it works


shuse2:
@pablitovicente 
https://classic.yarnpkg.com/en/docs/cli/workspace/
there is a command, but since we dont generally want to add new package, let's leave it as it is =)

PR #4832, opened 2020-02-14
Add data access actions to framework - Closes #4827

### LINKED ISSUE #4827: Expose new actions from application/node
### Description

In order to support to fetch data in case of missed event from #4826 , it should add framework actions to fetch the blocks, transactions and accounts.

### Motivation

- In order to fetch the data from the custom module in case they missed or want to catch up with the latest information, framework needs to provide the interface.

### Acceptance Criteria

Framework should provide below actions
- app:getBlockByID (id: string)
- app:getBlocksByIDs (ids: string[])
- app:getBlockByHeight (height: number)
- app:getBlocksByHeights (heights: number[])
- app:getBlocksByHeightBetween(from: number, to: number)

- app:getTransactionByID (id: string)
- app:getTransactionByID (ids: string[])

- app:getAccount(by address)
- app:getAccounts(by address[])

### Additional Information
Related issue: #4824

mitsuaki-u:
### What was the problem?

For custom modules, we needed actions which for data access in framework layer.

This PR resolves #4827 

### How was it solved?

Added actions in application and node.


framework/src/application/application.js
```diff
@@ -431,6 +431,16 @@ class Application {
 				'rebuild',
 				'processor:sync',
 				'processor:broadcast',
+				'processor:newBlock',
+				'getAccount',
```
shuse2: this is for the `get***` are not the events, so it should be removed from here.
Also, `processor:newBlock` should be removed

framework/test/jest/unit/specs/application/application.spec.js
```diff
@@ -349,4 +349,61 @@ describe('Application', () => {
...
+		beforeEach(() => {
+			// Arrange
+			app = new Application(genesisBlock, config);
+			app.channel = app._initChannel();
+			events = app.channel.eventsList.map(event => event.name);
+		});
+
+		it('should create getAccount action', () => {
+			// Assert
+			expect(events).toContain('getAccount');
```
shuse2: these are checking `events` rather than `actions`

framework/src/application/node/node.js
```diff
@@ -241,6 +241,24 @@ module.exports = class Node {
 					action.params.password,
 					action.params.forging,
 				),
+			getAccount: async action =>
+				this.chain.dataAccess.getAccountsByAddress([action.params.address]),
```
shuse2: This returns array, not an object.

framework/src/application/node/node.js
```diff
@@ -241,6 +241,24 @@ module.exports = class Node {
...
+			getAccounts: async action =>
+				this.chain.dataAccess.getAccountsByAddress(action.params.address),
+			getBlockByID: async action =>
```
shuse2: Also, in general, these has to be serialized when it goes out of node context

mitsuaki-u: It's still within the context of the node. Do you mean for custom module context?

framework/src/application/node/node.js
```diff
@@ -241,6 +241,61 @@ module.exports = class Node {
...
 				),
+			getAccount: async action => {
+				const [account] = this.chain.dataAccess.getAccountsByAddress([
+					action.params.address,
+				]);
+
+				return account;
+			},
+			getAccounts: async action =>
+				this.chain.dataAccess.getAccountsByAddress(action.params.address),
```
shuse2: accounts also need to be `toJSON`

framework/src/application/node/node.js
shuse2: yes when the data goes out from channel or network, it always should be serialized

PR #4834, opened 2020-02-17
Move test app functionalities to sample app - Closes #4833

### LINKED ISSUE #4833: Move test app out of framework change to sample app
### Description

Move test app from inside of framework/test to the root level, and change it to the sample app.
Also, it should represents how `application` should be built as PoC.

### Motivation

Test app was created to test the framework when we split the `core` and `sdk`.
However, in order to mimic the behavior of application, it should be outside of `framework`, and use `sdk` instead of `framework`

### Acceptance Criteria

- Remove test app and all the functionality should be moved to this sample-app
- It should be command line tool

shuse2:
### What was the problem?

This PR resolves #4833 

### How was it solved?

- Create sample app as CLI and add the same functionality as test app
- Update Jenkins to use this sample app for testing

### How was it tested?

- All the test should pass
- `yarn; yarn build; cd sample-app; ./bin/run run`


sample-app/src/app/index.ts
```diff
@@ -11,20 +11,10 @@
  *
  * Removal or modification of this copyright notice is prohibited.
  */
+import { Application } from 'lisk-sdk';
 
-'use strict';
+export const createApp = (genesisBlock: JSON, config: JSON): Application => {
```
ishantiw:
```typescript
export const createApp = (genesisBlock: JSON, config: JSON): Application => new Application(genesisBlock, config)
```

sample-app/src/utils/config.ts
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright © 2018 Lisk Foundation
```
ishantiw: `2020`

sample-app/src/utils/constants.ts
```diff
@@ -0,0 +1,17 @@
+/*
+ * Copyright © 2018 Lisk Foundation
```
ishantiw: `2020`

shuse2:
@nazarhussain 
It might be valid point, but as a structure, we cannot have nested package.json for the application.
It will not bootstrap correctly, and it requires the valid package.json for this to work.

Also, it's not really part of `SDK`, as this is sample of `SDK` usage.
We definitely don't want to ship this as part of SDK, and this is used for functional tests

sample-app/src/app/index.ts
shuse2: We could do this, but i shouldve added the comment to expand the module/custom transaction here

nazarhussain:
@shuse2 Then we have couple of options; 

1. If want an isolated sample app, we can extract it out as `LiskHQ/lisk-sdk-sample` repository and add it as git submodule 

2. If want to keep in lerna structure to use local packages, we can put it in `sdk/` directory. It will not require to have  `package.json` there as it can refer to all packages without it. The only thing will be we don't need to publish that directory during npm build. 

The first option is more appropriate if you want to have an app looks like `lisk-core` for testing the SDK.

shuse2:
For both of these options, it won't really work for us, but until further clarification, Im going to close this PR.
Essentially what it needs to do is to show the usage close to the `core` using `sdk` as separate package (ie outside of sdk as user) and used for testing the latest code.
Therefore, it shouldn't and it cannot be under `sdk` folder.
- Validate correct interface for typescript
- Validating all the interface that we create for application
- Making sure correct interfaces are exposed from SDK

> If want an isolated sample app, we can extract it out as LiskHQ/lisk-sdk-sample repository and add it as git submodule

We want to test/use the latest and only the latest SDK, which cannot be done using the separate repository. For this purpose, `lisk-core` is sufficient.

> If want to keep in lerna structure to use local packages, we can put it in sdk/ directory. It will not require to have package.json there as it can refer to all packages without it. The only thing will be we don't need to publish that directory during npm build.

If we put it under the `sdk` directory, this whole sample-app will not work, and it's not under the package of `sdk`, it's more of the usage of SDK which we need to test and use.

PR #4875, opened 2020-02-24
Move reward fee distribution to lisk-chain - Closes #4849

### LINKED ISSUE #4849: Remove reward/fee destribution from the lisk-dpos library and move to the lisk-chain to calculate per block
### Description

With the new dynamic fee system, rewards and fee have to be directly given to the delegate who forged the block for each block.
Therefore, no longer related to the DPoS system, and it should be handled in the `lisk-chain`.

### Motivation

motivation is described in https://github.com/LiskHQ/lips/blob/master/proposals/lip-0013.md

### Acceptance Criteria

- Remove all the reward and fee handling from the `lisk-dpos` along with the tests
- Add function to update the block generator account for reward and fees in the apply step after applying all the transactions
- Add function to burn sum of minFee while giving the fee to the block generator
- Accumulate burnt fee in the chain state
- Test if the minimum fee is correctly burnt
- Test if the delegate is rewarded with specified rewards and fees
- Test if the accumulated burnt fee is correctly recorded

### Related issues
#4837

shuse2:
### What was the problem?

This PR resolves #4849
 
### How was it solved?

- Move reward and fee calculation to chain
- Remove reward calculation from dpos
- Update vote weight calculation for the voted delegates for the generator to take the total earning for delegate

### How was it tested?

- Add test scenario on chain
- Update test scenario on dpos


elements/lisk-chain/src/block_reward.ts
```diff
@@ -121,3 +125,94 @@ export const calculateSupply = (
...
+export const getTotalRewardAndFees = (
+	blockInstance: BlockInstance,
+): { readonly totalFee: bigint; readonly totalBurnt: bigint } => {
+	// tslint:disable-next-line no-let
+	let totalFee = BigInt(0);
+	// tslint:disable-next-line no-let
+	let totalBurnt = BigInt(0);
+
+	for (const transaction of blockInstance.transactions) {
+		const minFee = BigInt(0); // TODO: It should get from transaction
```
nazarhussain: We do have an issue for this task, can we add a link to that issue directly in the comment.

elements/lisk-chain/src/block_reward.ts
```diff
@@ -121,3 +125,94 @@ export const calculateSupply = (
...
+		totalFee += transaction.fee - minFee;
+		totalBurnt += minFee;
+	}
```
nazarhussain: Although I am ok with for loop here I think we prefer a more functional approach. 

```js
return blockInstance.transactions.reduce((acc, transaction) => {
   acc.totalFee = acc.totalFee + transaction.fee - minFee;
   acc.totalBurnt = acc.totalBurnt + minFee;
   return acc; 
}, {totalFee: 0, totalBurnt: 0})
```

elements/lisk-chain/src/block_reward.ts
```diff
@@ -121,3 +125,94 @@ export const calculateSupply = (
 
 	return supply;
 };
+
+export const getTotalRewardAndFees = (
+	blockInstance: BlockInstance,
+): { readonly totalFee: bigint; readonly totalBurnt: bigint } => {
```
nazarhussain: I would prefer to not mix the concept of `burntFee` here in this logic. Logically its just the sum of `minFee` of each transaction we burn. So all interfaces should be preferred to return `totalMinFee` and then the caller/forger/implementation would decide if that value is needed to be burnt or do what. The `totalFee` should be total fee of all transactions and the implementation can later subtract `totalFee - totalMinFee` to calculate the `burntFee` for that block. 


```suggestion
): { readonly totalFee: bigint; readonly totalMinFee: bigint } => {
```

elements/lisk-chain/src/block_reward.ts
```diff
@@ -121,3 +125,94 @@ export const calculateSupply = (
...
+	generator.balance += blockInstance.reward;
+	generator.producedBlocks += 1;
+	// If there is no transactions, no need to give fee
+	if (!blockInstance.transactions.length) {
+		stateStore.account.set(generatorAddress, generator);
+
+		return;
+	}
+	const { totalFee, totalBurnt } = getTotalRewardAndFees(blockInstance);
+	generator.balance += totalFee;
```
nazarhussain: Related to the above comment, when we read this line its confusing. Logically `totalFee` meant to be total fee of block, difficult to track in mind that this already subtracted value. 

Would be much understandable if we have: 

```js
generator.balance += (totalFee - totalMinFee);
```

elements/lisk-chain/src/block_reward.ts
```diff
@@ -121,3 +125,94 @@ export const calculateSupply = (
...
+	const totalFeeBurntStr = await stateStore.chainState.get(
+		CHAIN_STATE_KEY_BURNT_FEE,
+	);
+	// tslint:disable-next-line no-let
+	let totalFeeBurnt = totalFeeBurntStr
+		? BigInt(parseInt(totalFeeBurntStr, 10))
+		: BigInt(0);
```
nazarhussain: `BigInt` constructor accepts string values. 

```js
let totalFeeBurnt = BigInt(totalFeeBurntStr || 0);
```

elements/lisk-chain/src/block_reward.ts
```diff
@@ -121,3 +125,94 @@ export const calculateSupply = (
...
+	totalFeeBurnt += totalBurnt;
```
nazarhussain: If we have something like this then its more clear that we are burning the minFee. 

```suggestion
	totalFeeBurnt += totalMinFee;
```

shuse2:
@nazarhussain 
Related to rounds test, it is all covered in the DPoS tests.
Also, in this version, complete vote calculation will be different, so it doesn't make sense to update the rounds test anymore

elements/lisk-chain/test/unit/process.spec.ts
```diff
@@ -745,6 +807,39 @@ describe('blocks/header', () => {
...
+				expect(generator.producedBlocks).toEqual(1);
+			});
+
+			it('should update generator balance with rewards and fees - minFee', async () => {
+				const generator = await stateStore.account.get(
+					getAddressFromPublicKey(block.generatorPublicKey),
+				);
+				let expected = block.reward;
+				for (const tx of block.transactions) {
+					// TODO: substract minFee
```
ManuGowda: is it gonna be addressed in this PR? if not can we have here with issue

elements/lisk-chain/test/unit/process.spec.ts
```diff
@@ -745,6 +807,39 @@ describe('blocks/header', () => {
...
+				expect(generator.balance.toString()).toEqual(expected.toString());
+			});
+
+			it('should update burntFee in the chain state', async () => {
+				const burntFee = await stateStore.chainState.get(
+					CHAIN_STATE_KEY_BURNT_FEE,
+				);
+				let expected = BigInt(0);
+				for (const tx of block.transactions) {
+					// TODO: substract minFee
```
ManuGowda: same

elements/lisk-chain/test/unit/process.spec.ts
shuse2: Updated comment

PR #4877, opened 2020-02-24
Remove second signature registration transaction - Closes #4840

### LINKED ISSUE #4840: Remove second signature registration transaction
### Description

After introducing the new multi-signature transaction ([LIP-0017](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0017.md)) the second signature is deprecated, so it should just simply be removed.

### Motivation

The exact same feature can be achieved with the new multi-signature transaction by adding the second signature as a mandatory key.

### Acceptance Criteria

- All of the code and tests related to the second signature registration should be removed
- Remove registration of the transaction in the application
- Remove commander commands related to `second signature`
- Test in the framework should also be removed
- Test in the commander should also be removed

### Related issues
#4836, #4852

ManuGowda:
### What was the problem?

After introducing the new multi-signature transaction (LIP-0017) the second signature is deprecated, so it should just simply be removed.

### How was it solved?

- All of the code and tests related to the second signature registration should be removed
- Remove registration of the transaction in the application
- Remove commander commands related to the second signature
- Test in the framework should also be removed
- Test in the commander should also be removed

### How was it tested?

- Scan the repository for the usage of the second signature and it should be removed
- The build should be green
- Run the application and send the second signature transaction and it should not be accepted


framework/src/application/storage/migrations/node/20200201000000_remove_second_signature_from_accounts.sql
```diff
@@ -0,0 +1,23 @@
...
+ */
+
+
+ /*
+  DESCRIPTION: Remove rank column from mem accounts
+*/
+
+ALTER TABLE mem_accounts
+DROP COLUMN "secondSignature",
+DROP COLUMN "u_secondSignature",
```
shuse2: this shouldn't exist already

framework/test/jest/unit/specs/application/application.spec.js
```diff
@@ -278,7 +275,6 @@ describe('Application', () => {
 
 			const TransactionWithoutBase = {
 				prototype: {},
```
shuse2: maybe this should use some other transaction instead?

framework/test/utils/legacy/multisig_scenarios.js
```diff
@@ -72,39 +71,12 @@ function Multisig(options) {
 
 	this.multiSigTransaction = multisigTrs.toJSON();
 
-	// TODO: Remove signRawTransaction on lisk-transactions 3.0.0
-	const multisigSecondSignatureTrs = new MultisignatureTransaction({
```
shuse2: should this be removed in this PR?

ManuGowda:
> I don't see the migration of second signatures so if you drop the columns data might be lost. Maybe you should not drop those columns yet.

The migration of the second signature is handled in a separate issue, the SDK will be up to date with not having the second signature, so it should not accept any second signature or support

elements/lisk-chain/test/unit/state_store_account.spec.ts
```diff
@@ -196,79 +194,14 @@ describe('state store / account', () => {
 		});
 	});
 
-	describe('set', () => {
```
shuse2: this should be replaced by some other property because it's checking the `set` functionality

elements/lisk-chain/test/unit/state_store_account.spec.ts
```diff
@@ -196,79 +194,14 @@ describe('state store / account', () => {
...
 	describe('finalize', () => {
 		let txStub = {} as StorageTransaction;
 		let existingAccount;
 		let updatedAccount;
-		let secondPublicKey: string;
-		let secondSignature: number;
 		let accountUpsertObj: object;
 
 		beforeEach(async () => {
-			secondPublicKey =
```
shuse2: same as above

elements/lisk-chain/test/unit/transactions.spec.ts
```diff
@@ -128,10 +128,12 @@ describe('blocks/transactions', () => {
 					}) as TransactionJSON,
 				);
 				const notAllowedTx = chainInstance.deserializeTransaction(
-					registerSecondPassphrase({
-						passphrase: genesisAccount.passphrase,
-						secondPassphrase: 'second-passphrase',
+					registerMultisignature({
```
shuse2: better to use other transactions because multi signature will change, and same for below

elements/lisk-cryptography/src/sign.ts
```diff
@@ -97,112 +95,25 @@ export const verifyMessageWithPublicKey = ({
...
-
-	const verifyFirstSignature = () =>
-		verifyDetached(messageBytes, signatureBytes, publicKeyBytes);
-	const verifySecondSignature = () =>
-		verifyDetached(messageBytes, secondSignatureBytes, secondPublicKeyBytes);
-
-	return verifyFirstSignature() && verifySecondSignature();
-};
-
 export interface SingleOrDoubleSignedMessage {
```
shuse2: does it make sense to keep this?

elements/lisk-transactions/test/base_transaction.spec.ts
```diff
@@ -838,54 +801,6 @@ describe('Base transaction class', () => {
 		});
 	});
 
-	describe('create, sign and stringify transaction', () => {
-		it('should return correct senderId/senderPublicKey when sign with passphrase', () => {
```
shuse2: I think this test should remain

elements/lisk-transactions/test/utils/sign_and_validate.spec.ts
```diff
@@ -12,124 +12,17 @@
...
 // The list of valid transactions was created with lisk-js v0.5.1
 // using the below mentioned passphrases.
 import {
 	validMultisignatureAccount as defaultMultisignatureAccount,
 	validMultisignatureTransaction,
-	validSecondSignatureTransaction,
 } from '../../fixtures';
 
 describe('signAndVerify module', () => {
-	describe('#validateSignature', () => {
```
shuse2: test not related to second signature should remain

framework/test/fixtures/accounts.js
```diff
@@ -62,16 +60,16 @@ accounts.mem_accountsFields = [
 	'asset',
 	'membersPublicKeys',
 	'votedDelegatesPublicKeys',
+	'secondPublicKey',
```
shuse2: why not remove this one?

framework/test/jest/unit/specs/application/application.spec.js
```diff
@@ -17,7 +17,7 @@
 
 const {
 	BaseTransaction: Base,
-	SecondSignatureTransaction,
+	MultisignatureTransaction,
```
shuse2: it's better not to use multisig for test

framework/test/fixtures/accounts.js
ManuGowda: as per Pablo's suggestion, the migration to remove these columns was not addressed in this PR, we will remove these columns from the database in the other issue which address the migration of second signature to multi-sign. I have already update that issue with the migration information

elements/lisk-bft/test/fixtures/accounts.ts
```diff
@@ -40,10 +40,8 @@ export const genesis = {
 export const mem_accountsFields = [
 	'username',
 	'isDelegate',
-	'secondSignature',
```
nazarhussain: Unless these fields are removed from the database table, they should not be excluded from these fixtures.

elements/lisk-chain/src/account.ts
```diff
@@ -16,9 +16,6 @@ import { AccountJSON } from './types';
 export const accountDefaultValues = {
 	publicKey: undefined,
 	// tslint:disable-next-line:no-null-keyword
-	secondPublicKey: null,
-	secondSignature: 0,
```
nazarhussain: The same applies here, this issues should remove second-signature from everwhere we are generating transactions. But until we completely remove it from database we should not remove it from these files.

commander/src/commands/transaction/create/multisignature.ts
```diff
@@ -37,11 +37,10 @@ const processInputs = (
 	lifetime: number,
 	minimum: number,
 	keysgroup: ReadonlyArray<string>,
-) => ({ passphrase, secondPassphrase }: InputFromSourceOutput) =>
+) => ({ passphrase }: InputFromSourceOutput) =>
```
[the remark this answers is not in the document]
ManuGowda: this will not be release anyway with these changes, and this will be updated with pablo's PR. so the economics phase should already address this concern.

PR #4880, opened 2020-02-25
Add MIN_REMAINING_BALANCE static property in the base transaction, and verify after updating the account - Closes #4848

### LINKED ISSUE #4848: Add MIN_REMAINING_BALANCE static property in the base transaction, and verify after updating the account
### Description

With the new dynamic fee system, `MIN_REMAINING_BALANCE ` should be adjustable.
It should add `MIN_REMAINING_BALANCE ` a static property in the base transaction, and after subtracting `balance` from the `sender`, it should check against this `MIN_REMAINING_BALANCE `

### Motivation

Motivation is described in https://github.com/LiskHQ/lips/blob/master/proposals/lip-0025.md

### Acceptance Criteria

- The base transaction should have this property set to 500000 Beddow in `bigint` as a static property
- When subtracting fee from sender account in the apply step, it should check if the remaining `balance` is equal or greater than specified `MIN_REMAINING_BALANCE `
- Test base transaction where sender balance is the same amount as minimum balance
- Test base transaction where sender balance is below the `MIN_REMAINING_BALANCE `
- Test base transaction where sender balance is above the `MIN_REMAINING_BALANCE `

- When recipient exists in transaction (only type 8), check resulting balance is is equal or greater than specified `MIN_REMAINING_BALANCE `
- Test transfer transaction where recipient resulting balance  is below the `MIN_REMAINING_BALANCE `

### Related issues
#4837

nazarhussain:
### What was the problem?

This PR resolves #4848. There LIP0025 introduced a business constraint to have a minimum remaining balance for any account after processing transaction. There was no such check earlier. 

### How was it solved?

Added this check to `BaseTransaction` and also created tests for it. 

### How was it tested?

Run tests for `lisk-transactions`


elements/lisk-transactions/src/utils/verify.ts
```diff
@@ -72,6 +72,25 @@ export const verifyAmountBalance = (
 	return undefined;
 };
 
+export const verifyMinRemainingBalance = (
+	id: string,
+	account: Account,
+	minRemainingBalance: bigint,
+): TransactionError | undefined => {
+	const balance = BigInt(account.balance);
```
shuse2: `account.balance` is always `bigint` now

nazarhussain: To keep the consistent implementation with `verifyAmountBalance` and `verifyBalance` I used `BigInt` here.

elements/lisk-transactions/src/base_transaction.ts
```diff
@@ -318,6 +321,17 @@ export abstract class BaseTransaction {
 
 		errors.push(...assetErrors);
 
+		// Validate minimum remaining balance
+		const minRemainingBalanceError = verifyMinRemainingBalance(
+			this.id,
+			// Get updated account state
+			await store.account.get(this.senderId),
```
shuse2: sender already exist in above, so it's better to move the `store.account.set(sender.address, sender);` below this, or just reuse

elements/lisk-transactions/src/utils/verify.ts
shuse2: it's probably better to make it consistently not use the `BigInt`

elements/lisk-transactions/test/base_transaction.spec.ts
```diff
@@ -20,45 +20,57 @@ import { Status } from '../src/response';
 import { TransactionError, TransactionPendingError } from '../src/errors';
 import {
 	addTransactionFields,
-	MockStateStore as store,
 	TestTransaction,
 	TestTransactionBasicImpl,
 } from './helpers';
 import * as transferFixture from '../fixtures/transaction_network_id_and_change_order/transfer_transaction_validate.json';
+import * as transferSecondSignatureFixture from '../fixtures/transaction_network_id_and_change_order/transfer_transaction_with_second_signature_validate.json';
```
shuse2: this fixture shouldnt be used

elements/lisk-transactions/test/utils/state_store_mock.ts
```diff
@@ -0,0 +1,129 @@
...
+import { Account, TransactionJSON } from '../../src/transaction_types';
+import {
+	StateStoreCache,
+	StateStoreDefaultGetter,
+	StateStoreGetter,
+	StateStoreSetter,
+	StateStoreTransactionGetter,
+} from '../../src/base_transaction';
+
+export const defaultAccount = {
```
shuse2: for the Account class,  i think it's better to use the real `Account` so that it won't have inconsistency in the future

nazarhussain: What you meant by real account?

shuse2: ah nevermind, i was thinking it's `lisk-chain`, but it's transactions
All good here

elements/lisk-transactions/src/base_transaction.ts
nazarhussain: Sender don't have an updated balance. When we called `this.applyAsset(store)` The balance of the sender is changed in the store. The `sender` which exists in this context have old balance. So I have to fetch updated information from the store.

shuse2: ahhh ok that's true, it can changed in the previous step.
Maybe split the await statement above, and make bit more descriptive comment there?

PR #4884, opened 2020-02-26
Remove numberOfTransactions and maxTransactionsPerBlock - Closes #4851

### LINKED ISSUE #4851: Remove number of transactions per block validation from `lisk-chain` and update maxPayloadLength constant
### Description

With the new dynamic fee system, the number of transactions per block check should be removed.

### Motivation

Motivation is described in https://github.com/LiskHQ/lips/blob/master/proposals/lip-0013.md

### Acceptance Criteria

- Number of transactions check needs to be removed from `elements/lisk-chain/src/validate.ts`
- Remove the related tests
- Config for setting the number of transactions per block needs to be removed
- `maxTransactionsPerBlock` constant needs to be removed in config and constructors in chain
- `maxPayloadLength` constant needs to be updated in config and constructors in chain
- Test exceeding `maxPayloadLength`
- Test below `maxPayloadLength`

### Related issues
#4837

ishantiw:
### What was the problem?

This PR resolves #4851 

### How was it solved?

- Removed `numberOfTransactions` check in `elements/lisk-chain/src/validate.ts`
- Remove tests related to `numberOfTransactions`
- Remove `numberOfTransactions` from the config
- Remove `maxTransactionsPerBlock ` from the config and constructors in chain
- Update value of `maxPayloadLength` to be equal to 15 kb.

### How was it tested?

`npm t`


elements/lisk-chain/test/unit/chain.spec.ts
```diff
@@ -35,8 +35,7 @@ describe('chain', () => {
 	const constants = {
 		blockReceiptTimeout: 20,
 		loadPerIteration: 1000,
-		maxPayloadLength: 1024 * 1024,
-		maxTransactionsPerBlock: 25,
+		maxPayloadLength: 15 * 1024,
```
ManuGowda: can we use this from the constants or config? so that if SDK user wants to configure the payload size the test is intact. same comment for rest of the test below.

ishantiw: 1) This commit updates the default value for `maxPayloadLength` from `1 Mb` to `15 kB`  and it is already configurable as was before I think -> 
https://github.com/LiskHQ/lisk-sdk/pull/4884/commits/26c898749d900f6a5b8d42e217861d8f08e78035

2) The issue only says to change it in `elements/lisk-chain` and only to remove the check condition and from the constructor. Also, we still want to accept it in the block header? @shuse2

ManuGowda: ok, in that case, my comment is invalid, what about the comment on point 2, reference to `maxTransactionsPerBlock` exists in code. We dont remove them?

ishantiw: For point 2, only the check should be removed, I reverted the changes where I removed it from the types in lisk-chain library. This is what the issue says and we want to keep it in the `blockHeader`

ManuGowda:
@ishantiw the tests are failing

PR #4886, opened 2020-02-27
Add nonce in the account class, and add nonce verification in the base transaction - Closes #4850

### LINKED ISSUE #4850: Add nonce in the account class, and add nonce verification in the base transaction
### Description

In order to replace the transaction in the transaction pool with a higher fee, `nonce` has to be introduced.

### Motivation

Motivation is described in https://github.com/LiskHQ/lips/blob/master/proposals/lip-0015.md

### Acceptance Criteria

- Add `nonce` as bigint in the account class and also in the database as a `string`
- Add validation against `nonce` against account for equality in the Base Transaction `apply` step
- Increment account nonce during apply after checking the nonce equality
- Decrement account nonce during undo after checking the nonce equality
- Test case where the transaction has the same `nonce` as account `nonce`
- Test case where the transaction has the smaller `nonce` than account `nonce`
- Test case where the transaction has the bigger `nonce` than account `nonce`
- Test case if account nonce is incremented after `apply`
- Test case if account nonce is decrement after `undo`

### Related issues
#4837

nazarhussain:
### What was the problem?

This PR resolves #4850. The base transaction was not verifying the account nonce. 

### How was it solved?

Added account nonce verification in the BaseTransation.

### How was it tested?

Run `lisk-transactions` tests.


ManuGowda:
@nazarhussain there are conflicts in the PR, also it shows 28 commits??

ManuGowda:
why is it listing 75 file changes? overall @nazarhussain changes look good. can we wait for @mitsuaki-u PR get merged and then review this PR?

elements/lisk-transactions/src/utils/verify.ts
```diff
@@ -86,6 +86,24 @@ export const verifyMinRemainingBalance = (
 	return undefined;
 };
 
+export const verifyAccountNonce = (
```
shuse2: we need to detect if it's `===` or `>`, maybe we can have 2 different error for this, so that in transaction pool we can check if it's only one nonce error, it's unprocessable, but if it has more than one error it's invalid?
ie)
```
if (nonce < account.nonce) {
errors.push(new TransactionError());
}
if (nonce !== account.nonce) {
errors.push(new TransactionError());
}
```

shuse2: If we only return one error, in the transaction pool, it would be hard to distinguish (only string comparison) between invalid and unprocessable.
Maybe return 2 errors is better?

nazarhussain: There can't be two errors, either it's higher or lower nonce. And we don't have different error constructors in transactions to differentiate between errors, so we have to do message comparison.

shuse2: if we return 2 errors for lower nonce and nonce not equal,
we can check
* for non-processable case, `errors.lengh === 1 && errors[0].dataPath === '.nonce'` => not equal error
* for processable `errors.length === 0`
everything else is invalid
dataPath and error length is used to handle exceptions now

nazarhussain: That's what exactly I mentioned above, adding a check on the length of errors is more fragile and less readable. There can be a bug or logical error in nonce validation which can result in `errors.lenght === 1` which will turn it into invalid execution.  

So be it more clearly readable at the time of processing what are expecting.

framework/src/application/storage/migrations/node/20200226000000_add_nonce_to_trs.sql
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright © 2019 Lisk Foundation
```
ManuGowda: you can start using 2020 for new files

framework/src/application/storage/migrations/node/20200226000000_add_nonce_to_trs.sql
```diff
@@ -0,0 +1,16 @@
...
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+ALTER TABLE "trs" ADD COLUMN IF NOT EXISTS "nonce" TEXT NOT NULL DEFAULT '0';
+ALTER TABLE "trs" ALTER COLUMN "nonce" DROP DEFAULT;
```
ManuGowda: why are we adding and removing the default?

framework/src/application/storage/migrations/node/20200227102000_remove_timestamp_from_trs.sql
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright © 2019 Lisk Foundation
```
ManuGowda: same

framework/src/components/storage/entities/account.js
```diff
@@ -48,6 +48,7 @@ class Account extends BaseEntity {
 			booleanToInt,
 		);
 		this.addField('balance', 'string', { filter: ft.NUMBER });
+		this.addField('nonce', 'string', { filter: ft.BOOLEAN });
```
ManuGowda: boolean? its number know?

elements/lisk-transactions/src/utils/verify.ts
nazarhussain: I added more details to the error object itself. Now we can do something like. 

```
error.dataPath == '.nonce' && error.actual < error.expected
```

framework/src/application/storage/migrations/node/20200226000000_add_nonce_to_trs.sql
nazarhussain: We removed `default` so every transaction must have its own value, so there may not be any transaction mistakenly added without nonce, DB will catch it.

elements/lisk-transactions/src/utils/verify.ts
shuse2: as long as we dont use message to validate it's good. Another option would be just extend TransactionError and create specific error (I think we had `TransactionPendingError` for multi-sig before

framework/src/components/storage/entities/account.js
nazarhussain: Value is string, filter type is `BOOLEAN`. Which will result in `nonce_eql` and `nonce_ne` filters.

PR #4896, opened 2020-03-02
Create TransactionList to manage transactions based on nonce - Closes #4855

### LINKED ISSUE #4855: Create transaction list class
### Description
In order to manage transactions in order of nonce per account, it should create transaction list class.
Class should have below property
```
{
     address: '',
     transactions: {
       [nonce]: Transaction,
     },
     nonceHeap: MinHeap<nonce>,
     processable: number[]<nonce>,
     maxSize: number,
     minReplacementFeeDifference: number,
}
```
with interfaces
```
add(tx: Transaction, processable: bool = false): bool; // if replaced, remove subsequent transactions from processable
delete(nonce: number): bool; // remove subsequent transactions from processable
getProcessable(): Transaction[]; // return copy of processable transactions
getUnprocessable(): Transaction[]; // return copy of unprocessable transactions
```

### Motivation

Transaction pool need to transactions in order of nonce, and detect processable and non-processable transactions. Also, it should be able to replace transaction by nonce.

### Acceptance Criteria

- It should be efficient to replace by nonce
- It should be ordered by nonce per account
- It should implement function to return transactions which have sequential nonce as an array
- When the new replacement transaction has a lower fee priority than the original, it should be rejected
- Test if getProcessable() returns sequential nonce transactions
- Test if getUnprocessable() does not return nonce in the `processable`
- Test if transaction is replaced by `add`, all subsequent nonce transactions are not processable
- Test if transaction is deleted, the transaction is deleted and all subsequent nonce transactions are not processable
- Test if replacement transaction does not satisfy `minReplacementFeeDifference`, it is not replaced

### Related issues
#4841

shuse2:
### What was the problem?

This PR resolves #4855 

### How was it solved?

- Add transaction list implementation to be used in the transaction pool

### How was it tested?

- All tests are covered in the transaction pool unit test

### DIFF
--- a/elements/lisk-transaction-pool/src/transaction_list.ts
+++ b/elements/lisk-transaction-pool/src/transaction_list.ts
@@ -1,0 +1,246 @@
+/*
+ * Copyright © 2019 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import { EventEmitter } from 'events';
+
+import { MinHeap } from './min_heap';
+import { TransactionObject } from './transaction_pool';
+
+export interface TransactionListOptions {
+	readonly maxSize?: number;
+	readonly minReplacementFeeDifference?: bigint;
+}
+
+const DEFAULT_MAX_SIZE = 64;
+const DEFAULT_REPLACEMENT_FEE_DIFF = BigInt(0);
+
+export const EVENT_TRANSACTION_ADDED = 'EVENT_TRANSACTION_ADDED';
+export const EVENT_TRANSACTION_REMOVED = 'EVENT_TRANSACTION_REMOVED';
+
+export class TransactionList {
+	public readonly address: string;
+	public readonly events: EventEmitter;
+
+	private _processable: Array<bigint>;
+	// Value is not needed here
+	private readonly _transactions: { [nonce: string]: TransactionObject };
+	private readonly _nonceHeap: MinHeap<undefined, bigint>;
+	private readonly _maxSize: number;
+	private readonly _minReplacementFeeDifference: bigint;
+
+	public constructor(address: string, options?: TransactionListOptions) {
+		this.address = address;
+		this.events = new EventEmitter();
+		this._transactions = {};
+		this._nonceHeap = new MinHeap<undefined, bigint>();
+		this._processable = [];
+		this._maxSize = options?.maxSize ?? DEFAULT_MAX_SIZE;
+		this._minReplacementFeeDifference =
+			options?.minReplacementFeeDifference ?? DEFAULT_REPLACEMENT_FEE_DIFF;
+	}
+
+	public get(nonce: bigint): TransactionObject | undefined {
+		return this._transactions[nonce.toString()];
+	}
+
+	public add(tx: TransactionObject, processable: boolean = false): boolean {
+		const replacingTx = this._transactions[tx.nonce.toString()];
+		// If the same nonce already exist in the pool try to replace
+		if (replacingTx) {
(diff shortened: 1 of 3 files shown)


elements/lisk-transaction-pool/src/transaction_list.ts
```diff
@@ -0,0 +1,246 @@
...
+		this.events.emit(EVENT_TRANSACTION_REMOVED, {
+			address: this.address,
+			id: removingTx.id,
+		});
+
+		return true;
+	}
+
+	public promote(txs: ReadonlyArray<Transaction>): boolean {
+		// Promtoe if only all ID are still existing and the same
```
mitsuaki-u: `Promote`

elements/lisk-transaction-pool/test/unit/transaction_list.spec.ts
```diff
@@ -0,0 +1,553 @@
+/*
+ * Copyright © 2019 Lisk Foundation
```
ishantiw: `2020`

elements/lisk-transaction-pool/src/transaction_list.ts
```diff
@@ -0,0 +1,249 @@
...
+		this._nonceHeap.push(incomingTx.nonce, undefined);
+		// If this transaction is processable and it is the first one in the list
+		if (processable && this._processable.length === 0) {
+			this._processable.push(incomingTx.nonce);
+		}
+
+		// If the size exceeds, remove the largest nonce transaction
+		if (this._nonceHeap.count > this._maxSize) {
+			this.remove(highestNonce);
+		}
```
ishantiw: maybe it better if we remove a transaction first when its reaches its `_maxSize ` so that we make sure at no given time the list size exceeds the `_maxSize `. So checking `if (this._nonceHeap.count >= this._maxSize) {` before `this._transactions[incomingTx.nonce.toString()] = incomingTx;`

shuse2: In case like account has nonce: 0, and in pool if it has [40, 41 ...125], and if nonce: 1 comes in, rather than rejecting, maybe it's better to remove 125?

ishantiw: But if nonce `1` comes in, with below condition nonce `125` will be removed
```typescript
if (this._nonceHeap.count >= this._maxSize) {
			this.remove(highestNonce);
		}
```
and now, we can add 1,
```typescript
this._transactions['1'] = incomingTx;
this._nonceHeap.push(incomingTx.nonce, undefined);
// If this transaction is processable and it is the first one in the list
if (processable && this._processable.length === 0) {
   this._processable.push(incomingTx.nonce);
  }
```

ishantiw: In this case, we will never have the list above `_maxSize`.

shuse2: ah you mean just move up the condition?

ishantiw: yes :)

elements/lisk-transaction-pool/src/transaction_list.ts
```diff
@@ -0,0 +1,249 @@
...
+		this.events.emit(EVENT_TRANSACTION_REMOVED, {
+			address: this.address,
+			id: incomingTx.id,
```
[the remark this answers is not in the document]
shuse2: `remove` will handle that event, so this was not needed

elements/lisk-transaction-pool/src/transaction_list.ts
```diff
@@ -0,0 +1,249 @@
...
+		for (const _ of this._processable) {
+			clonedHeap.pop();
+		}
+		const firstNonProcessable = clonedHeap.pop();
+		if (!firstNonProcessable) {
+			return [];
+		}
+		if (this._processable.length !== 0) {
+			const heighestNonce = this._processable[this._processable.length - 1];
+			if (firstNonProcessable.key !== heighestNonce + BigInt(1)) {
```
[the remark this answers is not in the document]
shuse2: this is checking, if there is processable transaction, get the highestProcessableNonce, and the nonProcessableTransaction nonce has to be highestProcessableNonce + 1, otherwise all the nonProcessableTransaction is not possible to promote

PR #4897, opened 2020-03-02
Update commander with dynamic fee and nonce - Closes #4888

### LINKED ISSUE #4888: Update commander to create new transactions
### Description

With the new fee and nonce property for the transaction, the commander should support to add fee and nonce while creating the transaction

### Motivation

Fee and nonce have to be supported.

### Acceptance Criteria

- Each transaction creation commands should have argument of `fee` in lisk and `nonce` as required input.
ex) `transaction:create -t 8 0.001 0 100 123L` as fee, nonce, amount,  and then recipient
- All test should be updated
- Test if the transaction output has nonce and fee properly

### Related Issues
#4837

mitsuaki-u:
### What was the problem?

With new dynamic fee and tx nonce system, commander needed to be updated.

This PR resolves #4888 

### How was it solved?

Updated commander commands and tests.

### How was it tested?

`npm run test` in commander

### DIFF
--- a/commander/src/commands/transaction/create.ts
+++ b/commander/src/commands/transaction/create.ts
@@ -98,8 +98,6 @@ export default class CreateCommand extends BaseCommand {
 		votes: flagParser.string(commonFlags.votes),
 		unvotes: flagParser.string(commonFlags.unvotes),
 		networkIdentifier: flagParser.string(commonFlags.networkIdentifier),
-		nonce: flagParser.string(commonFlags.nonce),
-		fee: flagParser.string(commonFlags.fee),
 	};
 
 	async run(): Promise<void> {
--- a/commander/src/commands/transaction/create.ts
+++ b/commander/src/commands/transaction/create.ts
@@ -23,7 +23,7 @@ import MultisignatureCommand from './create/multisignature';
 import TransferCommand from './create/transfer';
 import VoteCommand from './create/vote';
 
-const MAX_ARG_NUM = 3;
+const MAX_ARG_NUM = 5;
 
 interface TypeNumberMap {
 	readonly [key: string]: string;
@@ -98,8 +98,6 @@ export default class CreateCommand extends BaseCommand {
 		votes: flagParser.string(commonFlags.votes),
 		unvotes: flagParser.string(commonFlags.unvotes),
 		networkIdentifier: flagParser.string(commonFlags.networkIdentifier),
-		nonce: flagParser.string(commonFlags.nonce),
-		fee: flagParser.string(commonFlags.fee),
 	};
 
 	async run(): Promise<void> {
--- a/commander/src/commands/transaction/create.ts
+++ b/commander/src/commands/transaction/create.ts
@@ -81,7 +81,7 @@ export default class CreateCommand extends BaseCommand {
 	`;
 
 	static examples = [
-		'transaction:create --type=0 100 13356260975429434553L',
+		'transaction:create --type=8 100 13356260975429434553L',
 		'transaction:create --type=delegate lightcurve',
 	];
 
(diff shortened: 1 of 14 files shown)


commander/src/commands/transaction/create/transfer.ts
```diff
@@ -74,12 +91,16 @@ export default class TransferCommand extends BaseCommand {
...
-	static examples = ['transaction:create:transfer 100 13356260975429434553L'];
+	static examples = [
+		'transaction:create:transfer 1 100 100 13356260975429434553L',
+	];
 
 	static flags = {
 		...BaseCommand.flags,
 		networkIdentifier: flagParser.string(commonFlags.networkIdentifier),
 		passphrase: flagParser.string(commonFlags.passphrase),
+		nonce: flagParser.string(commonFlags.nonce),
```
shuse2: No need to add to the flag

commander/src/commands/transaction/create/delegate.ts
```diff
@@ -74,15 +91,26 @@ export default class DelegateCommand extends BaseCommand {
...
 		const networkIdentifier = getNetworkIdentifierWithInput(
 			networkIdentifierSource,
 			this.userConfig.api.network,
 		);
+
+		if (!isValidNonce(nonce)) {
+			throw new ValidationError('Enter a valid nonce in number string format.');
+		}
+
+		if (!isValidFee(fee)) {
```
shuse2: fee should be validated after the conversion, otherwise it fails on decimals

commander/src/utils/flags.ts
```diff
@@ -105,6 +109,12 @@ export interface FlagMap {
 }
 
 export const flags: FlagMap = {
+	nonce: {
```
shuse2: we don't need to add it to the flag

commander/src/commands/transaction/create/delegate.ts
mitsuaki-u: It should fail on decimals because we expect a non-decimal input e.g. `transaction:create:delegate 1 100 lightcurve`

shuse2: when the input is `LSK` it should accept decimal points (it should be able to say `0.0001` LSK for fee

commander/src/commands/transaction/create/delegate.ts
```diff
@@ -101,12 +101,16 @@ export default class DelegateCommand extends BaseCommand {
 			throw new ValidationError('Enter a valid nonce in number string format.');
 		}
 
-		if (!isValidFee(fee)) {
+		if (Number.isNaN(Number(fee))) {
```
shuse2: do we need this validation if we are going to validate below anyway?
I think `convertLSKToBeddows` and `isValidFee` both have validation so it should be enough

mitsuaki-u: `convertLSKToBeddows` doesn't check if it's non-numeric like 'a'

commander/src/commands/transaction/create.ts
```diff
@@ -81,7 +81,7 @@ export default class CreateCommand extends BaseCommand {
 	`;
 
 	static examples = [
-		'transaction:create --type=0 100 13356260975429434553L',
+		'transaction:create --type=8 100 13356260975429434553L',
```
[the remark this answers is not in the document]
mitsuaki-u: Each specific example is in it's transaction:create type file e.g. `create/elegate.ts`,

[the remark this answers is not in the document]
ishantiw: It only has one example 😕 
https://github.com/LiskHQ/lisk-sdk/blob/e026d528dd46fa5638c29f8cf3597b0c906b3122/commander/src/commands/transaction/create/delegate.ts#L75

PR #4898, opened 2020-03-02
Add min balance check for recipient - Closes #4848

### LINKED ISSUE #4848: Add MIN_REMAINING_BALANCE static property in the base transaction, and verify after updating the account

shuse2:
### What was the problem?

This PR resolves #4848 

### How was it solved?

Add missing recipient balance check on the transfer transaction

### How was it tested?

New test case is covered in the test

### DIFF
--- a/elements/lisk-chain/test/unit/process.spec.ts
+++ b/elements/lisk-chain/test/unit/process.spec.ts
@@ -625,7 +625,7 @@ describe('blocks/header', () => {
 						nonce: '0',
 						passphrase: genesisAccount.passphrase,
 						recipientId: '123L',
-						amount: '100',
+						amount: '10000000',
 						networkIdentifier,
 					}) as TransactionJSON,
 				);
@@ -650,7 +650,7 @@ describe('blocks/header', () => {
 				).rejects.toMatchObject([
 					expect.objectContaining({
 						message: expect.stringContaining(
-							'Account does not have enough LSK',
+							'Account does not have enough minimum remaining LSK',
 						),
 					}),
 					expect.objectContaining({
@@ -716,7 +716,7 @@ describe('blocks/header', () => {
 						nonce: '0',
 						passphrase: genesisAccount.passphrase,
 						recipientId: '124L',
-						amount: '100',
+						amount: '10000000',
 						networkIdentifier,
 					}) as TransactionJSON,
 				);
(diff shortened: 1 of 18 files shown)


elements/lisk-chain/test/unit/process.spec.ts
```diff
@@ -650,7 +650,7 @@ describe('blocks/header', () => {
 				).rejects.toMatchObject([
 					expect.objectContaining({
 						message: expect.stringContaining(
-							'Account does not have enough LSK',
+							'Account does not have enough minimum remaining LSK',
```
pablitovicente: Maybe show the value for minimum?

elements/lisk-transactions/test/base_transaction.spec.ts
```diff
@@ -584,8 +584,8 @@ describe('Base transaction class', () => {
 			expect((errors as ReadonlyArray<TransactionError>)[0]).toBeInstanceOf(
 				TransactionError,
 			);
-			expect((errors as ReadonlyArray<TransactionError>)[0].message).toEqual(
-				`Account does not have enough LSK: ${defaultSenderAccount.address}, balance: 0`,
+			expect((errors as ReadonlyArray<TransactionError>)[0].message).toContain(
```
pablitovicente: Maybe better to do exact match?

elements/lisk-chain/test/unit/process.spec.ts
shuse2: Updated the error message to include actual and expected

elements/lisk-transactions/test/base_transaction.spec.ts
shuse2: Checking the exact match of error message make the test too fragile IMO

PR #4900, opened 2020-03-03
Remove signature and signSignature, and multi-signature handling from Base Transaction, and replace with signatures verification - Closes #4838

### LINKED ISSUE #4838: Remove signature and signSignature, and multi-signature handling from Base Transaction, and replace with signatures verification
### Description

When introducing the new multi-signature scheme, we no longer need `signSignature` and `signature`, instead, the base transaction needs to implement a new rule for validating the `signatures`.
This has to be done at the `apply` phase of the transaction since it requires the account state

### Motivation
`signSignature`, `signature` and any previous multi-signature handling will not be used anymore, but instead, new `signatures` verification should be enforced

### Acceptance Criteria

- BaseTransaction.apply step should check if the `signatures` property is valid according to the LIP.
- BaseTransaction should not have `signature` property
- BaseTransaction should not have `signSignature` property
- transactions database table should not have `signature` and `signSignature` property
- BaseTransaction should not have `processMultisignature`
- Remove Multi-signature handling cases from transactions library
- All of the transaction creation function is adapted to this change
- Create function to sort `signatures` based on `mandatoryKeys` and `optionalKeys` from account state and array of [{ `signature`: "",  `publicKey`: ""}]
- Test basic account signature case
- Test 1 mandatory key signature case, with the same public key associated with the address (register of multisignature with its original public key)
- Test 1 mandatory key signature case, with the different public key associated with the address (use different passphrase)
- Test 2 mandatory key signature case (second signature)
- Test case where account requires signatures from mix of mandatory and optional keys
- Test only cases where only optional keys are specified
- Test more and fewer `numberOfSignatures` than specified for multi-signature and a basic account
- All of the test related to signatures in mocha should be removed if no longer relevant

### Related issues
#4836

ManuGowda:
### What was the problem?

This PR resolves #4838 

### How was it solved?
- Removed usage of `signature` and `signSignature`
- Replaces `signature` logic with `signatures`
- Update signature verification
- Updated all the test to use `signatures`
- Removed unwanted `multi-signature` code
### How was it tested?

- Build should be green
- Send type 8 to 12 and check if they get accepted


elements/lisk-transactions/src/utils/verify.ts
```diff
@@ -86,72 +85,86 @@ export const verifyMinRemainingBalance = (
...
 
-	if (valid) {
-		return {
-			status: MultisignatureStatus.READY,
-			errors: [],
-		};
+	errors.push(...mandatoryKeysError);
+
+	// tslint:disable-next-line: no-let
+	let validOptionalSig = 0;
```
shuse2: you can check
`(non-empty signatures).length === numberOfSignatures` and then
```
const optionalSignatures = signatures.slice(mandatoryKeys.length + 1);
for (let k = mandatoryKey.length + 1 ....) {
  if (signature === '') {
    continue;
  }
```

elements/lisk-transactions/src/schema.ts
```diff
@@ -100,14 +96,6 @@ export const baseTransaction = {
...
-			type: 'string',
-			format: 'signature',
-		},
-		signSignature: {
-			type: 'string',
-			format: 'signature',
-		},
 		signatures: {
 			type: 'array',
 			uniqueItems: true,
```
shuse2: Remove `uniqueItems` and `format` and checked in the code.
minItem is 1 and max is 64

framework/src/application/schema/genesis_block_schema.js
```diff
@@ -111,9 +111,15 @@ module.exports = {
 						description:
 							'Send relevant data with transaction like delegate, vote, signature, ...',
 					},
-					signature: {
-						type: 'string',
-						format: 'signature',
+					signatures: {
+						type: 'array',
+						uniqueItems: true,
```
shuse2: it cannot be unique items

framework/src/application/schema/genesis_block_schema.js
```diff
@@ -111,9 +111,15 @@ module.exports = {
...
+						items: {
+							type: 'string',
+							format: 'signature',
```
shuse2: format cannot be `signature`

elements/lisk-transactions/fixtures/transactions.json
```diff
@@ -318,7 +318,6 @@
 		},
 		"senderPublicKey": "c094ebee7ec0c50ebee32918655e089f6e1a604b83bcaa760293c61e0f18ab6f",
 		"signature": "8953d989a4be6d2d0ab25dbb11b1268a21ce3681aba6c0fe391b2b7c18179f52af5440aca8b27794eb27d29b41a306f931882190a1ad922c274fb9bc5be63303",
```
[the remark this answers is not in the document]
ManuGowda: its being used, updated it

elements/lisk-transactions/test/utils/verify.spec.ts
```diff
@@ -38,13 +70,208 @@ describe('#verify', () => {
...
+		it('should return empty array when signatures are valid', () => {
+			const { signatures, asset: keys } = defaultTransferTransaction;
+			const [result] = validateKeysSignatures(
+				keys,
+				signatures,
+				defaultTransferTransactionBytes,
+			);
 
-		it('should call validateMultisignature with correct argument', async () => {});
+			expect(result).toBeEmpty;
```
shuse2: `toBeEmpty` should be function call?

elements/lisk-transactions/test/utils/verify.spec.ts
```diff
@@ -38,13 +70,208 @@ describe('#verify', () => {
...
+			);
+
+			expect(result).toBeInstanceOf(TransactionError);
+			expect(result).toHaveProperty(
+				'message',
+				'Transaction has empty signatures',
+			);
+		});
+
+		it('should return error when signatures does not have required number of signatures', () => {
```
shuse2: Prefer to add a case where `signatures` doesn't have enough length

framework/src/application/schema/genesis_block_schema.js
```diff
@@ -111,9 +111,14 @@ module.exports = {
...
 							'Send relevant data with transaction like delegate, vote, signature, ...',
 					},
-					signature: {
-						type: 'string',
-						format: 'signature',
+					signatures: {
+						type: 'array',
+						items: {
+							type: 'string',
+							oneOf: [{ type: 'string', format: 'signature' }],
```
shuse2: same as other case in above (use of `oneOf`

elements/lisk-transactions/src/base_transaction.ts
```diff
@@ -189,16 +174,12 @@ export abstract class BaseTransaction {
 		return this._senderPublicKey;
 	}
 
-	public get signature(): string {
-		if (!this._signature) {
-			throw new Error('signature is required to be set before use');
+	public get signatures(): ReadonlyArray<string> {
```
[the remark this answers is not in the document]
ManuGowda: so we would remove the check and just return the signatures?

elements/lisk-transactions/src/utils/create_base_transaction.ts
```diff
@@ -37,3 +40,16 @@ export const createBaseTransaction = ({
 		senderPublicKey,
 	};
 };
+
+export const getSignaturesBytes = (signatures: ReadonlyArray<string>) => {
+	if (signatures?.length) {
```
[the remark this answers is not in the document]
ManuGowda: so the idea was.. this.signatures will throw error if its not set, thats why i was using the getter to throw it for me.

elements/lisk-transactions/src/utils/verify.ts
```diff
@@ -54,72 +53,89 @@ export const verifyMinRemainingBalance = (
...
-	if (!isMultisignatureAccount(sender)) {
-		return {
-			status: MultisignatureStatus.NONMULTISIGNATURE,
-			errors: [],
-		};
+	const { mandatoryKeys, optionalKeys, numberOfSignatures } = sender.keys;
+	const numMandatoryKeys = mandatoryKeys.length;
+	const nonEmptySignaturesCount = signatures.filter(k => k.length !== 0).length;
+
+	if (nonEmptySignaturesCount !== numberOfSignatures) {
```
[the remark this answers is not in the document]
ManuGowda: why do we need this check? `signatures.length !== numMandatoryKeys + optionalKeys.length`
in what case this will occur?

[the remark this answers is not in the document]
shuse2: I don't think there is any validation regards to the length of the signatures. 
ie) empty element was not injected property, it will throw
```
mandatory: 1, optional: 3, required signature 2
signatures: ['validPK', 'validOptionalForPosition4']
```

elements/lisk-transactions/src/utils/create_base_transaction.ts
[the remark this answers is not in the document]
shuse2: for `this.signatures` is should never throw as it is always required at least one?

framework/src/application/storage/migrations/node/20200229000000_remove_signature_signSignature_from_transactions.sql
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+
+/*
+  DESCRIPTION: Remove signature and signSignature from trs
+*/
+
+ALTER TABLE trs
+DROP COLUMN "signature",
+DROP COLUMN "signSignature";
```
pablitovicente: I think this needs to be added to the list of things to be migrated when going to main net?

elements/lisk-transactions/src/utils/create_base_transaction.ts
```diff
@@ -37,3 +40,16 @@ export const createBaseTransaction = ({
...
+		// tslint:disable-next-line: no-unnecessary-callback-wrapper
+		const signaturesBuffer = signatures.map(signature =>
+			hexToBuffer(signature),
```
pablitovicente: Isn't this missing the prepend of booleans as per LIP? https://github.com/LiskHQ/lips/blob/master/proposals/lip-0017.md#signatures-serialization ?

framework/src/application/storage/migrations/node/20200229000000_remove_signature_signSignature_from_transactions.sql
ManuGowda: I think this is covered under the migration issue. do we need anything extra to be done here?

elements/lisk-transactions/src/utils/verify.ts
```diff
@@ -54,72 +53,89 @@ export const verifyMinRemainingBalance = (
...
-		errors &&
-		errors.length === 1 &&
-		errors[0] instanceof TransactionPendingError
-	) {
-		return {
-			status: MultisignatureStatus.PENDING,
-			errors,
-		};
+	// tslint:disable-next-line: prefer-for-of no-let
+	for (let k = 0; k < numOptionalKeys - 1; k += 1) {
```
pablitovicente: Can you add comments about why are you modifying the limit (`< numOptionalKeys - 1`) and what's the objective of the whole loop?

framework/src/application/storage/migrations/node/20200229000000_remove_signature_signSignature_from_transactions.sql
pablitovicente: Ok if it's covered. Resolving then.

elements/lisk-transactions/src/base_transaction.ts
```diff
@@ -434,7 +424,7 @@ export abstract class BaseTransaction {
 			this.getBasicBytes(),
 		]);
 
-		this._signatures.push(
+		this.signatures.push(
```
shuse2: Let's create another issue to update `sign` function

framework/src/application/schema/genesis_block_schema.js
```diff
@@ -115,7 +115,10 @@ module.exports = {
 						type: 'array',
 						items: {
 							type: 'string',
```
shuse2: probably this `type` is no needed anymore

elements/lisk-transactions/src/utils/create_base_transaction.ts
```diff
@@ -37,3 +40,17 @@ export const createBaseTransaction = ({
 		senderPublicKey,
 	};
 };
+
+export const serializeSignatures = (signatures: ReadonlyArray<string>) => {
+	const signaturesBuffer = signatures.map(signature => {
+		// If signature is empty append 0x00 to byteBuffer
+		if (signature.length === 0) {
+			return Buffer.from('0x00');
```
pablitovicente:
```suggestion
			Buffer.from('00', 'hex');
```

elements/lisk-transactions/src/utils/create_base_transaction.ts
```diff
@@ -37,3 +40,17 @@ export const createBaseTransaction = ({
...
+		}
+
+		// If signature is not empty append 0x01 to byteBuffer
+		return Buffer.concat([Buffer.from('0x01'), hexToBuffer(signature)]);
```
pablitovicente:
```suggestion
		return Buffer.concat([Buffer.from('01', 'hex');, hexToBuffer(signature)]);
```

elements/lisk-transactions/src/base_transaction.ts
ManuGowda: https://github.com/LiskHQ/lisk-sdk/issues/4906

elements/lisk-transactions/src/utils/create_base_transaction.ts
```diff
@@ -37,3 +40,17 @@ export const createBaseTransaction = ({
 		senderPublicKey,
 	};
 };
+
+export const serializeSignatures = (signatures: ReadonlyArray<string>) => {
+	const signaturesBuffer = signatures.map(signature => {
```
pablitovicente: - This can be achieved with only one loop (you can check protocol-specs for multisig for the same functionality)
- it would be ideal to use variables for the `00` and `01` buffers so they make more sense

elements/lisk-transactions/src/base_transaction.ts
```diff
@@ -485,31 +392,24 @@ export abstract class BaseTransaction {
...
-			this._multisignatureStatus === MultisignatureStatus.PENDING ||
-			this._multisignatureStatus === MultisignatureStatus.READY
-				? UNCONFIRMED_MULTISIG_TRANSACTION_TIMEOUT
-				: UNCONFIRMED_TRANSACTION_TIMEOUT;
 		const timeElapsed =
 			// tslint:disable-next-line no-magic-numbers
 			timeNow - Math.floor(this.receivedAt.getTime() / 1000);
 
-		return timeElapsed > timeOut;
+		return timeElapsed > UNCONFIRMED_TRANSACTION_TIMEOUT;
```
pablitovicente: What's this for? Is it related to Transaction Pool changes?

ManuGowda: yes, we have 3-hour timeout for all types of transactions, before it was calculating for multi-sig as well, so removed those specific calculations.

elements/lisk-transactions/src/utils/create_base_transaction.ts
ManuGowda: I had a look at your reference @pablitovicente, I addressed the variables part, but the serialization logic looks clear and good to me. Let me know if it is still a concern.

PR #4904, opened 2020-03-04
BatchError: deadlock detected in Transaction entity tests - Closes #4903

### LINKED ISSUE #4903: BatchError: deadlock detected in Transaction entity tests
### Expected behavior
There should not be any random test failures. 

### Actual behavior
Facing this error `BatchError: deadlock detected` randomly in the `components/storage/entitites/transaction.js` test file. 

### Steps to reproduce

Run the unit tests for the framework. 

3.0.0

nazarhussain:
### What was the problem?

This PR resolves #4903

### How was it solved?

By adding `await` for `except` cases to be resolved properly. 

### How was it tested?

Run `mocha:unit` for `framework.


framework/test/mocha/unit/components/storage/entities/transaction.js
```diff
@@ -263,42 +263,40 @@ describe('Transaction', () => {
...
+
+			expect(res).to.deep.equal([]);
 		});
 
 		it('should throw error for invalid options', async () => {
 			const transaction = new Transaction(adapter);
-			try {
+
+			await expect(() => {
 				transaction.get({}, invalidOptions);
```
shuse2: this internal function is the `async`. I think it will be detached error if internal function throws

nazarhussain:
@pablitovicente Yes returning expectation is a common pattern and was used earlier in all tests. But while mixing up with `async` cases using `await` is also the right pattern. And where ever we touch the test code we are using that pattern.

framework/test/mocha/unit/components/storage/entities/transaction.js
nazarhussain: I think it will not detach the error. If it would then this test case would have been failed.

pablitovicente:
> @pablitovicente Yes returning expectation is a common pattern and was used earlier in all tests. But while mixing up with `async` cases using `await` is also the right pattern. And where ever we touch the test code we are using that pattern.

ok could you double check the other occurrences - like the sample in blocks I mentioned earlier - it looked to me like they were the same case.

PR #4912, opened 2020-03-06
Implement addTransaction of TxPool

### LINKED ISSUE #4859: Create add transactions functionality in the transaction pool
### Description

The transaction pool should accept a transaction to be inserted into the pool using `addTransaction` interface.

### Motivation

In case a block is deleted or a transaction is received then these transactions should be inserted into the transaction pool.

### Acceptance Criteria

- It should check if the transaction.fee is greater than `minEntranceFee`
- It should check if the `transaction.feePriority` is greater than minimum `feePriority` in the pool in case of transaction pool size limit
- It should not add to the pool if transaction is duplicate
- It should return true if it's added
- It should set `receivedAt` property for tracking the expiry time
- validate the transaction with apply function before adding, and if it's invalid throws
- It should add to all transactions map
- It should add to the fee priority queue
- It should add to the corresponding `TransactionList` by address
- Test known ID will be rejected
- Test invalid transaction make it throw
- Test unprocessable and processable transaction does not make it throw
- Test when transaction.fee is below `minEntranceFee`
- Test transaction is added to the TransactionList
- Test duplicate transaction is not added to the TransactionList
- Test if transaction id is added to the `all` transaction map
- Test if transaction id is added to the fee priority queue
- Test when the pool is full and fee priority of the transaction is lower than min in the pool, it is rejected
- Test when the pool is full and fee priority of the transaction is higher than min in the pool, it is added

### Related issues
#4841

ishantiw:
### What was the problem?

This PR resolves #4859 

### How was it solved?

- [x] Check if the transaction.fee is greater than `minEntranceFee`
- [x] Check if the transaction.fee is greater than minimum fee in the pool in case of transaction pool size limit
- [x] Do not add to the pool if transaction is duplicate
- [x] Returns true if it's added
- [x] Sets `receivedAt` property for tracking the expiry time
- [x] Validates the transaction with apply function before adding, and if it's invalid throws
- [x] Adds to all transactions map
- [ ] Adds to the fee priority queue - #4914
- [x] Adds to the corresponding `TransactionList` by address
- [x] Implemented `getAllTransactions` of TxPool
- [x] Test Transaction Pool constructor
- [x] Test known ID will be rejected
- [x] Test invalid transaction make it throw
- [x] Test unprocessable and processable transaction does not make it throw
- [x] Test when transaction.fee is below `minEntranceFee`
- [x] Test transaction is added to the TransactionList
- [x] Test duplicate transaction is not added to the TransactionList
- [x] Test if transaction id is added to the `all` transaction map
- [ ] Test if transaction id is added to the fee priority queue - #4914
- [ ] Test when the pool is full and fee priority of the transaction is lower than min in the pool, it is rejected - #4914
- [ ] Test when the pool is full and fee priority of the transaction is higher than min in the pool, it is added - #4914

### How was it tested?

`npm t` under `elements/lisk-transaction-pool`


elements/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -109,9 +110,51 @@ export class TransactionPool {
...
+		const incomingTxAddress = getAddressFromPublicKey(
+			incomingTx.senderPublicKey,
+		);
+
+		const txResponse = await this._applyFunction([incomingTx]);
+		const txResponseErrors = txResponse[0].errors;
+
+		// If applyTransaction fails for the transaction then throw error
+		if (
+			(txResponseErrors.length === 1 &&
```
shuse2: maybe create a function `getStatus(txResonse): Status`? where Status is Processable, Unprocessable, Invalid?

elements/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -109,9 +110,51 @@ export class TransactionPool {
...
+				txResponseErrors[0].dataPath === '.nonce' &&
+				txResponseErrors[0].actual &&
+				txResponseErrors[0].expected &&
+				txResponseErrors[0].actual > txResponseErrors[0].expected) ||
```
mitsuaki-u: Instead of having so many conditions you can use the new typescript `?` feature e.g. `txResponseErrors?.[0]?.actual`

ishantiw: As discussed `txResponseErrors?.[0]?.actual` doesn't suppress ts error so keeping it like that

elements/lisk-transaction-pool/src/transaction_pool.ts
ishantiw: Added

elements/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -170,4 +167,22 @@ export class TransactionPool {
 	}
 
 	private async _reorganize(): Promise<void> {}
+
+	private _getStatus(
+		txResponse: ReadonlyArray<TransactionResponse>,
+	): TransactionStatus {
+		const txResponseErrors = txResponse[0].errors;
```
shuse2: if `errors` is 0 and status.ok, it should be processable

elements/lisk-transaction-pool/src/transaction_pool.ts
ishantiw: In AddTrx case, we will only get either `Invalid` or `unProcessable` trx, only job can promote and make a trx processable. But I've created `TransactionStatus` interface for 3 status values to be used in other places

elements/lisk-transaction-pool/src/transaction_pool.ts
ishantiw: In AddTrx case, we will only get either Invalid or unProcessable trx. Only job can promote and make a trx processable. isn't it?

PR #4946, opened 2020-03-09
Update base transaction sign to support passing multiple passphrases

### LINKED ISSUE #4906: Update base transaction sign function
### Description
Update base transaction sign function to allow the transaction to be signed by multiple pass-phrases. 

### Motivation
After the removal of `signature` and `signSignature` from the base transaction, and with the. introduction of new multi-signature LIP changes, we need to allow accounts to sign for multi-signature accounts and regular accounts with single and multiple passphrases.

### Acceptance Criteria

- `sign` function in `base_transaction` should be updated to allow signing for 
   - a regular account with one passphrase
   - a regular account with multiple pass-phrases
   - multi-signature account
- At least happy path tests cases for all transaction types are added 
### Additional Information
- Commander `sign` function needs update?

### Related Issues
#4836

pablitovicente:
### What was the problem?

This PR resolves #4906

### How was it solved?

By moving `signAll` function from multi-signature and adapting it for it to work with other transaction types

### How was it tested?

Unit tests


elements/lisk-transactions/src/base_transaction.ts
```diff
@@ -431,6 +433,52 @@ export abstract class BaseTransaction {
 		this._id = getId(this.getBytes());
 	}
 
+	public signAll(
+		networkIdentifier: string,
+		senderPassphrase?: string,
```
ManuGowda: optional? its mandatory know?

pablitovicente: It's not. It's mandatory for registration of multi-signature accounts after that the senderPassphrase does not have the same meaning. 

Imagine that key pair `C` was used for generating the registration and members where set to `[A,B,D]` in this case senderPassphrase could be either of `[A,B,D]` but will never be `C`.

if `C` would have been included as member then it could be one of the signers but it's not anymore senderPassphrase is just another passphrase.

ManuGowda: but in the case of a normal transfer transaction, for signing, we need the passphrase as mandatory know? how will that work in that case?

pablitovicente: In case of a transaction from regular account it needs to be passed. We probably need to add validation for either `senderPassphrase` or `passphrases` to be present. @shuse2 thought?

shuse2: i think this is `or` case, so we need to validation for each case.
if keys is specified, `passphrases` has to be greater than 1, if `keys` is not provided, ignore passphrases?

pablitovicente: @shuse2 need to check that proposal as `passphrases` might be 1 and be valid as it could be the case of members `[A,B,C]` and it's `C` starting the transaction

pablitovicente: @shuse2 @ManuGowda I've updated the code and the behaviour is now based on what's passed in: if `senderPassphrase` is provided it's assumed that it's an account with only one key pair and after signing it returns, on the other hand, if `senderPassphrase` is `undefined` and `passphrases` and `keys` are provided it follows the logic to sign in key order

ManuGowda: so, in this case, the commander sign should accept `senderPassphrase` as a flag or we take the first passphrase as `senderPassphrase`?

pablitovicente: Maybe it needs to respect the `signAll` (soon to be renamed `sign`) interface and pass either one or the other

elements/lisk-transactions/src/12_multisignature_transaction.ts
```diff
@@ -376,10 +376,10 @@ export class MultisignatureTransaction extends BaseTransaction {
 	public signAll(
 		networkIdentifier: string,
 		senderPassphrase: string,
+		passphrases?: ReadonlyArray<string>,
```
shuse2: shouldn't `senderPassphrase` also be optional if it has the same interface?

elements/lisk-transactions/src/base_transaction.ts
```diff
@@ -431,6 +433,69 @@ export abstract class BaseTransaction {
...
+		},
+	): void {
+		// Set network identifier if it was previously not set in the transaction
+		if (!this._networkIdentifier) {
+			this._networkIdentifier = networkIdentifier;
+		}
+
+		if (!this.signatures) {
+			this.signatures = [];
+		}
```
ManuGowda: I think this is no longer the case.. its initialized with `this.signatures = [];`, so we can remove this? I cant imagine when this case occurs `if (!this.signatures) {`

elements/lisk-transactions/src/base_transaction.ts
```diff
@@ -431,6 +433,69 @@ export abstract class BaseTransaction {
...
+			readonly numberOfSignatures: number;
+		},
+	): void {
+		// Set network identifier if it was previously not set in the transaction
+		if (!this._networkIdentifier) {
+			this._networkIdentifier = networkIdentifier;
+		}
+
+		if (!this.signatures) {
+			this.signatures = [];
```
shuse2: I think `signatures` is initialized in the constructor?

elements/lisk-transactions/test/8_transfer_transaction.spec.ts
```diff
@@ -335,4 +337,44 @@ describe('Transfer transaction class', () => {
...
+		beforeEach(async () => {
+			validTransferInstance = new TransferTransaction(transaction);
+		});
+
+		it('should have one signature for single key pair account', async () => {
+			validTransferInstance.signAll(
+				networkIdentifier,
+				account.passphrase,
+				undefined,
+				undefined,
```
ManuGowda: no need to pass the optional value?

elements/lisk-transactions/src/12_multisignature_transaction.ts
pablitovicente: Multisig registration is the exception I think as we need the sender to be in position `0` of the signatures array so this works as a kind of obligation for the account being transformed to be the first one to sign and share the transaction to other members

elements/lisk-transactions/src/base_transaction.ts
pablitovicente: Fixed

elements/lisk-transactions/src/base_transaction.ts
pablitovicente: Shu commented first and it's fixed now. Resolving.

elements/lisk-transactions/test/8_transfer_transaction.spec.ts
pablitovicente: If it's a single signature transaction there's no need as pass-phrases and key sets are not used. For regular accounts the senderPublicKey is the one that signs.

ManuGowda: what I meant was, `undefined` was not required to pass you could just send the first two arguments and it would still work.

PR #4947, opened 2020-03-09
signData, signDataWithPassphrase throws invalid error when missing passphrase - Closes #4917

### LINKED ISSUE #4917: signData, signDataWithPassphrase throws invalid error when missing passphrase
### Expected behavior
- When a required parameter(passphrase) is not provided, it should throw `pass missing passphrase parameter`
### Actual behavior
- Sign a message(buffer) without passing passphrase to function `signData`, `signDataWithPassphrase`, results in error as below
```
> sigAcc = cryptography.signData(accHash)
Thrown:
Error: Unsupported data format. Currently only Buffers or `hex` and `utf8` strings are supported.

> sigAcc = cryptography.signDataWithPassphrase(accHash)
Thrown:
Error: Unsupported data format. Currently only Buffers or `hex` and `utf8` strings are supported.
```
### Steps to reproduce
- cryptography.signData(accHash)
- cryptography.signDataWithPassphrase(accHash)
4.0.0, 3.0.0, 2.x.x

ManuGowda:
### What was the problem?
Sign a message(buffer) without passing passphrase to function signData, signDataWithPassphrase, results in an error which is misleading as below
```
> sigAcc = cryptography.signData(accHash)
Thrown:
Error: Unsupported data format. Currently only Buffers or `hex` and `utf8` strings are supported.
```

### How was it solved?
Check for required parameter and throw correct error message

### How was it tested?
- Added unit test for `signData`, `signDataWithPassphrase`


elements/lisk-cryptography/src/sign.ts
```diff
@@ -141,11 +141,15 @@ export const signDataWithPassphrase = (
 	data: Buffer,
 	passphrase: string,
 ): string => {
-	const { privateKeyBytes } = getPrivateAndPublicKeyBytesFromPassphrase(
-		passphrase,
-	);
+	if (data && passphrase) {
```
shuse2: if it requires the check, then the input type is wrong, but i think in this case, the usage of the function is wrong.
Otherwise, to make it extream, u need to check if it's Buffer or string too

ManuGowda: buffer or string is checked in the hash function.

PR #4948, opened 2020-03-09
Remove complex source handling - Closes #4945

### LINKED ISSUE #4945: Remove multiple input handling from commander
### Description

Currently passphrase can be input to the system in multiple way. (stdin, prompt, flag, file and env).
However, for multi signature and multi-signature registration we need to input multiple passphrases.

In order to make all the command consistent, it should remove `stdin`, `file` and `env`.

### Motivation

With the new multi signature transaction, there will be multiple cases ,user wants to pass the multiple-passphrase.

### Acceptance Criteria

- All of the passphrase should still support `prompt` and `flag`
- Description should say usage of `flag` is insecure.

shuse2:
### What was the problem?

This PR resolves #4945 

### How was it solved?

- Remove complex source handling for passphrase and password
- Keep the stdin reading for transaction input, and file reading for message

### How was it tested?

- All the tests are updated


commander/test/utils/reader.ts
```diff
@@ -0,0 +1,178 @@
+/*
+ * LiskHQ/lisk-commander
+ * Copyright © 2019 Lisk Foundation
```
ManuGowda: 2020

shuse2:
@ManuGowda to get the prompt multiple times, it just needs to call multiple times where it is required =)

commander/src/commands/transaction/create/delegate.ts
```diff
@@ -111,29 +109,28 @@ export default class DelegateCommand extends BaseCommand {
...
 		if (noSignature) {
-			const noSignatureResult = processFunction({
-				passphrase: undefined,
-			});
+			const noSignatureResult = processInputs(
+				nonce,
+				normalizedFee,
+				networkIdentifier,
+				username,
+				undefined,
```
ManuGowda: `undefined` is required to pass? if the passphrase is not sent it will be undefined know?

commander/src/commands/transaction/create/transfer.ts
```diff
@@ -137,31 +135,32 @@ export default class TransferCommand extends BaseCommand {
...
-		const processFunction = processInputs(
+		if (noSignature) {
+			const noSignatureResult = processInputs(
+				nonce,
+				normalizedFee,
+				networkIdentifier,
+				normalizedAmount,
+				address,
+				dataString,
+				undefined,
```
ManuGowda: same here

commander/src/utils/core/cache.ts
```diff
@@ -67,24 +67,18 @@ const stopCommand = async (
...
-		)) as PM2ProcessInstance;
-
-		if (password) {
-			return `${REDIS_CLI} -p ${redisPort} -a ${password} shutdown`;
-		}
+	const config = await getLiskConfig(installDir, network);
+	const password = config?.components?.cache?.password;
+	if (password === undefined) {
+		throw new Error('Config password is not set.');
+	}
```
ManuGowda: why do we need to throw here? let's say on alphanet the cache config is 
```
"cache": {
    "enabled": true
}
```
so this case the `password` will result in `undefined`, but that's the expected behavior 
this is the correct behaviour
```
if (password) {
		return `${REDIS_CLI} -p ${redisPort} -a ${password} shutdown`;
	}

	return `${REDIS_CLI} -p ${redisPort} shutdown`;
```

commander/src/utils/reader.ts
```diff
@@ -0,0 +1,173 @@
...
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import * as passphraseModule from '@liskhq/lisk-passphrase';
```
ManuGowda: liskPassphrase?

commander/src/utils/reader.ts
```diff
@@ -0,0 +1,173 @@
...
+
+export const readStdIn = async (): Promise<string[]> => {
+	const readFromStd = new Promise<string[]>((resolve, reject) => {
+		// tslint:disable readonly-array
+		const lines: string[] = [];
+		const rl = readline.createInterface({ input: process.stdin });
+
+		// Prevent readline hanging when command called with no input or piped
+		const id = setTimeout(() => {
+			clearTimeout(id);
```
nazarhussain: `setTimeout` is a one-time timer we don't need to `clearTimeout` inside.

commander/src/commands/transaction/create/delegate.ts
```diff
@@ -111,29 +109,28 @@ export default class DelegateCommand extends BaseCommand {
...
+			);
```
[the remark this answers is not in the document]
shuse2: this is not new function, so we can refactor this when we need to

Issue #4949, opened 2020-03-09
Typesafe StateStorePrepare


JesusTheHun:
Currently the type definition of the StateStorePrepare is as follow: 

```typescript
export interface StateStoreCache<T> {
	cache(
		filterArray: ReadonlyArray<{ readonly [key: string]: string }>,
	): Promise<ReadonlyArray<T>>;
}
```

Using type literal it is possible to provide an exhaustive list of available filters, allowing slick auto-completion and type safety, in following code : 

```typescript
class SomeTransaction {
	public async prepare(store: StateStorePrepare): Promise<void> {
		await store.account.cache([
			{
				address_in: ['1L', '2L'], // OK
				balance_in: 100, // Error
				balance_gte: 100, // OK
			},
		]);
	}
}
```

shuse2:
Closing because `prepare` method was there for technical reason , and it will be removed

PR #4950, opened 2020-03-09
StateStoreCache type safety

JesusTheHun:
Feature request presented in #4949 .

StateStoreCache currently doesn't have a precise type definition. It actually have wrong type definition since everything is a string.

See https://github.com/LiskHQ/lisk-sdk/blob/development/elements/lisk-transactions/src/base_transaction.ts#L80

I added an exhaustive list of filter keys and their expected type.


shuse2:
Thank you for the PR.
However, with the new generic serialization LIP, all the database, account system will be changed.
Therefore, I would prefer not to add further change now because it will increase amount of changes we need to handle in the near future.
Also, we will work on the conversion to TypeScript for the framework and that should improve the typing in general.

PR #4954, opened 2020-03-10
Fix to always choose distinct voted delegate list - Closes #4907

### LINKED ISSUE #4907: DPoS module unit tests are unstable
### Expected behavior
DPoS module unit tests should be stable
### Actual behavior
```
@liskhq/lisk-dpos:   ● dpos.apply() › Given block is NOT the last block of the round › should update vote weight of accounts that delegates who forged voted for
@liskhq/lisk-dpos:     expect(received).toEqual(expected) // deep equality
@liskhq/lisk-dpos:     Expected: "1230000000"
@liskhq/lisk-dpos:     Received: "2460000000"
@liskhq/lisk-dpos:       173 | 			for (const delegate of votedDelegatesFromGenerator) {
@liskhq/lisk-dpos:       174 | 				const votedDelegate = await stateStore.account.get(delegate.address);
@liskhq/lisk-dpos:     > 175 | 				expect(votedDelegate.voteWeight.toString()).toEqual(
@liskhq/lisk-dpos:           | 				                                            ^
@liskhq/lisk-dpos:       176 | 					(delegate.voteWeight + totalEarning).toString(),
@liskhq/lisk-dpos:       177 | 				);
@liskhq/lisk-dpos:       178 | 			}
@liskhq/lisk-dpos:       at Object.<anonymous> (test/unit/apply.spec.ts:175:49)
@liskhq/lisk-dpos:   ● dpos.apply() › Given block is NOT the last block of the round › should update vote weight of accounts that delegates who forged voted for
@liskhq/lisk-dpos:     expect.assertions(2)
@liskhq/lisk-dpos:     Expected two assertions to be called but received one assertion call.
@liskhq/lisk-dpos:       170 | 				generator.votedDelegatesPublicKeys.includes(delegate.publicKey),
@liskhq/lisk-dpos:       171 | 			);
@liskhq/lisk-dpos:     > 172 | 			expect.assertions(votedDelegatesFromGenerator.length);
@liskhq/lisk-dpos:           | 			       ^
@liskhq/lisk-dpos:       173 | 			for (const delegate of votedDelegatesFromGenerator) {
@liskhq/lisk-dpos:       174 | 				const votedDelegate = await stateStore.account.get(delegate.address);
@liskhq/lisk-dpos:       175 | 				expect(votedDelegate.voteWeight.toString()).toEqual(
@liskhq/lisk-dpos:       at Object.<anonymous> (test/unit/apply.spec.ts:172:11)
```
### Steps to reproduce
- These tests fails randomly, so its harder to give steps to reproduce, one of the ways to reproduce is run multiple times continuously in a loop
4.0.0

shuse2:
### What was the problem?

This PR resolves #4907 

### How was it solved?

Choose distinct voted delegate list for test.

### How was it tested?

Test should be stably passing

### DIFF
--- a/elements/lisk-dpos/test/utils/round_delegates.ts
+++ b/elements/lisk-dpos/test/utils/round_delegates.ts
@@ -53,7 +53,7 @@ export const delegateAccounts = delegatePublicKeys.map(
 		const votedDelegatesPublicKeys = [
 			votedDelegates[index].publicKey,
 			votedDelegates[index + 1].publicKey,
-			votedDelegates[index % 10].publicKey,
+			votedDelegates[index + 2].publicKey,
 		];
 		return {
 			address: getAddressFromPublicKey(pk),


elements/lisk-dpos/test/utils/round_delegates.ts
```diff
@@ -53,7 +53,7 @@ export const delegateAccounts = delegatePublicKeys.map(
 		const votedDelegatesPublicKeys = [
 			votedDelegates[index].publicKey,
 			votedDelegates[index + 1].publicKey,
-			votedDelegates[index % 10].publicKey,
```
ManuGowda: just to understand, in the first place why did we chose to do mod 10?

shuse2: it was originally like that. i think it just needs some random delegate public key, but it has to be distinct

PR #4955, opened 2020-03-10
Create command to sign transaction for multi-signature and non-multisignature account - Closes #4908

### LINKED ISSUE #4908: Create command to sign transaction for multi-signature and non-multisignature account
### Description

Create command to sign a transaction from multi-signature account.
`transaction:sign '{
    "type": 11,
    "id":"13311755314561666527",
    "senderPublicKey":"d1ba3238c3fb14c811634ec4723f9c0095dcbf8117104695584ab95f940e393f",
    "fee":"100000000",
    "signatures":[],
    "asset":{}
}' --passphrase="" --passphrase="" --mandatory-key="xxx" --mandatory-key="yyy" --optional-key="xxx" --optional-key="yyy"
`

should get the sender account information, and insert a signature to a relevant position.

### Motivation

With the new multi-signature, it needs a feature to be able to sign a transaction.

### Acceptance Criteria

- It should output a signed transaction in JSON format
- It should only allow `prompt` and `flags` options
- Test case where input transaction is invalid
- Test case where the sender is not multi-signature account

### Additional Data

This can, and probably should, use `signMultiSignatureTransaction()` in `@lisk-transactions/src/sign_multi_signature_transaction.ts`

### Related Issues
#4836

ManuGowda:
### What was the problem?

Refer #4908 

### How was it solved?

Updating `transaction sign` command

### How was it tested?
- Create transaction
`
data=$(./bin/run transaction:create:transfer 1 1 1000 5553317242494141914L - --passphrase="opinion fog place adapt clap behind write squirrel account mask earn marriage" --data='{"lisk":"zug"}')
`
- One passphrase
`
./bin/run transaction:sign $data --passphrase="opinion fog place adapt clap behind write squirrel account mask earn marriage"
`

- Multiple passphrase
`
./bin/run transaction:sign $data --passphrase="inherit moon normal relief spring bargain hobby join baby flash fog blood" --mandatory-key=eb06e0a8cbb848f81f126b538794eb122ae8035917ded1da3e5c85618602f3ba --optional-key=0b211fce4b615083701cb8a8c99407e464b2f9aa4f367095322de1b77e5fcfbe
`

- Prompt for passphrase
`
./bin/run transaction:sign $data --mandatory-key=eb06e0a8cbb848f81f126b538794eb122ae8035917ded1da3e5c85618602f3ba --optional-key=0b211fce4b615083701cb8a8c99407e464b2f9aa4f367095322de1b77e5fcfbe --number-of-passphrases=1
`


commander/src/commands/transaction/sign.ts
```diff
@@ -56,13 +70,31 @@ export default class SignCommand extends BaseCommand {
...
+		'transaction:sign \'{"id":"17528738200145418850","type":8,"senderPublicKey":"c094ebee7ec0c50ebee32918655e089f6e1a604b83bcaa760293c61e0f18ab6f","nonce":"1","fee":"100000000","asset":{"data":"{"liskhq":"zug"}","amount":"100000000000","recipientId":"5553317242494141914L"}}\' --mandatory-key=215b667a32a5cd51a94c9c2046c11fffb08c65748febec099451e3b164452bca --optional-key=922fbfdd596fa78269bbcadc67ec2a1cc15fc929a19c462169568d7a3df1a1aa --number-of-signatures=2 --number-of-passphrases=2',
+		'transaction:sign \'{"id":"17528738200145418850","type":8,"senderPublicKey":"c094ebee7ec0c50ebee32918655e089f6e1a604b83bcaa760293c61e0f18ab6f","nonce":"1","fee":"100000000","asset":{"data":"{"liskhq":"zug"}","amount":"100000000000","recipientId":"5553317242494141914L"}}\' --mandatory-key=215b667a32a5cd51a94c9c2046c11fffb08c65748febec099451e3b164452bca --optional-key=922fbfdd596fa78269bbcadc67ec2a1cc15fc929a19c462169568d7a3df1a1aa --number-of-signatures=2 --passphrase="inherit moon normal relief spring bargain hobby join baby flash fog blood" --passphrase="wear protect skill sentence lift enter wild sting lottery power floor neglect"',
```
shuse2: I think example is too big that it's hard to read. maybe we can simplify them so that the important arguments and flags are clear

commander/src/commands/transaction/sign.ts
```diff
@@ -71,14 +103,19 @@ export default class SignCommand extends BaseCommand {
...
 			},
 		} = this.parse(SignCommand);
 
 		const { transaction }: Args = args;
 		const transactionInput = transaction || (await getTransactionInput());
 		const transactionObject = parseTransactionString(transactionInput);
 		const passphrase =
-			passphraseSource ?? (await getPassphraseFromPrompt('passphrase', true));
+			passphraseSource ??
+			(await getPassphrasesFromPrompt(numberOfPassphrases as number));
```
shuse2: flag comes in always as string, so it should `parseInt`

commander/src/commands/transaction/sign.ts
```diff
@@ -88,7 +125,33 @@ export default class SignCommand extends BaseCommand {
 			...transactionObject,
 			networkIdentifier,
 		});
-		txInstance.sign(passphrase);
+
+		interface Keys {
```
shuse2: interface shouldn't exist in the function inside

commander/src/commands/transaction/sign.ts
```diff
@@ -88,7 +125,33 @@ export default class SignCommand extends BaseCommand {
...
+			readonly mandatoryKeys: Array<Readonly<string>>;
+			readonly optionalKeys: Array<Readonly<string>>;
+			readonly numberOfSignatures: number;
+		}
+
+		const keys = ({
+			mandatoryKeys,
+			optionalKeys,
+			numberOfSignatures,
+		} as unknown) as Keys;
```
shuse2: why do we need to type like this?

commander/src/commands/transaction/sign.ts
ManuGowda: in this case the flag is `flagParser.integer({` i am casting it to number as the flag can resolve to `number|undefined` so

commander/src/commands/transaction/sign.ts
ManuGowda: because the parsed flags give undefined so, any ideas not to cast?

commander/src/commands/transaction/sign.ts
shuse2: if that's the case, i think it should be a problem with the typescript as `getPassphrasesFromPrompt` has default value, and it should accept the `undefined`

commander/src/commands/transaction/sign.ts
shuse2: flag is optional, so it could be undefined too unless it's used with required.
in that sense, it's better to check if it's undefined or not

commander/src/commands/transaction/sign.ts
```diff
@@ -88,7 +128,32 @@ export default class SignCommand extends BaseCommand {
...
 			networkIdentifier,
 		});
-		txInstance.sign(passphrase);
+
+		// tslint:disable-next-line: no-object-literal-type-assertion
+		const keys = {
+			mandatoryKeys,
+			optionalKeys,
+			numberOfSignatures,
+		} as {
```
shuse2: no need to type cast here.
However `numberOfSignatures` can be undefined. 
but in the `signMultiSignatureTransaction` it seems like it's not used  although it is in argument.
maybe we can remove from the flag and the function? @pablitovicente

ManuGowda: I think we can remove.. its not that required. let me know whats your opinion pablo.

pablitovicente: it's there because the interface was designed like that but it can be removed as it only make sense during verification which in that stage the transaction will need the account to verify what was the configuration at registration time.

ManuGowda: ok then I will remove it 👍

PR #4956, opened 2020-03-10
Add fee priority queue - Closes #4914

### LINKED ISSUE #4914: Add feePriorityQueue to TxPool
### Description

We need to reject a transaction with a low fee priority when the TxPool is full. For this, we have to maintain a `feePriorityQueue` as described [LIP-0013: transaction-pool-specification](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0013.md#transaction-pool-specification)

### Motivation

We want to give more preference to high fee priority transactions when the TxPool is full.

### Acceptance Criteria

- Create a minHeap for feePriorityQueue and add it to the TxPool
- Create a function that calculates `feePriority = (trs.fee - trs.minFee) / sizeof(trs)`
- Should have a check if TxPool is full and a low fee priority transactions come then it should reject it
- Update `_minimumEntranceFee ` to `_minEntranceFeePriority`
- Update the check for entrance fee to check for `feePriority` of a trx instead of a fee.
- Test for check if TxPool is full and a low fee priority transactions come then it should reject it

### Additional Information

`getBytes` function to be passed to TxPool

ishantiw:
### What was the problem?

This PR resolves #4914 

### How was it solved?

- [x] Create a minHeap for feePriorityQueue and add it to the TxPool
- [x] Create a function that calculates `feePriority = (trs.fee - trs.minFee) / sizeof(trs)`
- [x] Should have a check if TxPool is full and a low fee priority transactions come then it should reject it
- [x] Update `_minimumEntranceFee ` to `_minEntranceFeePriority`
- [x] Update the check for entrance fee to check for `feePriority` of a trx instead of a fee.
- [x] Test for check if TxPool is full and a low fee priority transactions come then it should reject it

### How was it tested?

`npm t`


elements/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -121,11 +124,21 @@ export class TransactionPool {
...
+		incomingTx.feePriority = this._calculateFeePriority(incomingTx);
+		if (incomingTx.feePriority < this._minEntranceFeePriority) {
 			return false;
 		}
+
 		// Check if incoming transaction fee is greater than the minimum fee in the TxPool if the TxPool is full
-		// TODO: Use feePriorityQueue to reject transaction with lowest fee when txPool is full
+		const lowestFeePriorityTrx = this._feePriorityQueue.peek();
+		if (
+			Object.keys(this._allTransactions).length === this._maxTransactions &&
```
shuse2: probably safer to check with `>`

ishantiw: I'm checking `>=` as we are checking this condition before adding so it should be checking for equal as well

PR #4957, opened 2020-03-10
Add job for promoting and evicting transaction in pool - Closes #4868

### LINKED ISSUE #4868: Create a job to promote and evict transactions
### Description

When we implement a new transaction pool, processable and non-processable transactions has to be distinguished.
When adding and removing a transaction, all unknown state transactions are considered non-processable.
Therefore, the job must check all non-processable transactions that have sequential nonce if they are processable.
Also, after executing above promotion, if transaction pool is full, it has to remove the transactions from the list until it is contained within the limit.

For promote steps are,
- For each account, check for non-processable transactions which has sequential nonce
- promote all non-processable transactions by applying together with processable transactions
- Discard all invalid transactions and all subsequent transactions from the same account

For eviction steps are,
- Evict expired transactions, and remove all subsequent transactions from the same account
- If TxPool is full, remove lowest fee priority non processable transactions 
- If TxPool is full, remove the highest nonce transactions with the lowest fee priority across accounts

### Motivation

In order to return processable transactions to forger module, transaction pool need to check if the transactions are processable.
It cannot be done through `add` or `remove` public interfaces because it forces transaction pool to process transactions too frequently.

### Acceptance Criteria
- Job should be running every 500 msec
- if promoting fails when applying transactions, failed transactions and subsequent transactions have to be discarded
- if promoting transactions are valid, it should be added to `processable` in TransactionList
- If all transactions in the TransactionList are processable, it should not be selected for promotion
- Non processable transactions in the TransactionList, which are not subsequent from other transactions should not be selected for promotion
- It should not mutate the TransactionList before any async executions for potential race condition
- Test if a promoting transaction fails during apply, all subsequent promoting transactions from the same account are discarded
- Test if all non-processable transactions which were promoted are in the processable list.
- Test if all transactions in the TransactionList are processable, promotion does not execute
- Test if transactions that are not subsequent from other transactions are not checked for promotion
- Test when `applyTransaction` execution takes too long while `addTransaction` and `removeTransaction` are called, the data of TransactionList are still consistent
- Test when transaction pool is full and non processable transactions exist in the pool, lowest fee non processable transactions are removed
- Test when transaction pool is full and only processable transactions exist in the pool, lowest fee among the highest nonce processable transaction is removed
- Test all expired transactions are removed and all subsequent transactions are removed

### Additional Information
- During the job, after the `async` operation, it has to assume the `TransactionList` is mutated

### Related issues
#4841

mitsuaki-u:
### What was the problem?

This PR resolves #4868

### How was it solved?

- [x] For each account, check if it has promotable transactions exist
- [x] Promote all non-processable transactions by applying together with existing processable transactions
- [x] Discard all invalid transactions and all subsequent transactions
- [x] If TxPool size is above threshold, create heap maps for eviction
- [x] If TxPool is full, remove non-processable transactions with the lowest feePriority until tx-pool has space
- [x] If TxPool is still full, remove the lowest feePriority processable transaction within the highest nonce transaction per account until tx-pool has space
- [x] Evict all transactions which stayed more than expiry time, and update all subsequent transactions as non-processable in separate job


elements/lisk-transaction-pool/src/transaction_list.ts
```diff
@@ -54,6 +54,12 @@ export class TransactionList {
 		return this._transactions[nonce.toString()];
 	}
 
+	public getAll(): ReadonlyArray<Transaction> {
```
shuse2: with the below change, this can be removed

elements/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -232,4 +257,115 @@ export class TransactionPool {
...
+			}
+		}
+
+		if (processableFeePriorityHeap.count < 1) {
+			return false;
+		}
+
+		const evictedTransaction = processableFeePriorityHeap.pop();
+
+		return this.removeTransaction(evictedTransaction?.value as Transaction);
```
shuse2: becasue this is failing on deep scan, you could do
```
const evictedTransaction = processableFeePriorityHeap.pop();
if (!evictedTransaction) {
  return false
}
return this.removeTransaction(evictedTransaction.value);

```

elements/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -232,4 +257,115 @@ export class TransactionPool {
...
+
+	private async _expire(): Promise<void> {
+		for (const transaction of Object.values(this._allTransactions)) {
+			const timeDifference = Math.round(
+				Math.abs(
+					(transaction.receivedAt as Date).getTime() - new Date().getTime(),
+				),
+			);
+
+			if (timeDifference > DEFAULT_EXPIRY_TIME) {
```
shuse2:
```suggestion
			if (timeDifference > this._transactionExpiryTime) {
```

PR #4958, opened 2020-03-10
Add fill block functionality in forger module - Closes #4862

### LINKED ISSUE #3862: Follow common developer approach for error handling in transactions
From start we are targeting and speaking about our Lisk SDK, is the first of its kind in JavaScript that will make blockchain accessible to everyone. 

With that said, we should keep in mind that developers with different experience levels and even very beginners will also work with our SDK. So we should keep in mind the common developer approach to write program. 

Or easy to say we should not neglect basic language patterns which every one is familiar of. One of that is `if you have error throw it`. This is very common for developers and they must be familiar and doing this a lot in other projects, which is the user of `throw`. 

Current transaction processing logic does not allow you to throw error, rather we enforce a pattern to collect errors and return as array object from transaction steps `apply` and `undo`. Which could be an additional behaviour for our transaction processing, but we should not neglect the user of `throw` as its very basic language struct. 

We should refactor the processing logic to handle all of below given scenarios; 

```js
apply(store) {
  if(some_condition) {
     throw new Error('Some error in validation.') ;
  }
}
```

```js
apply(store) {
  if(some_condition) {
     return new Error('Some error in validation.') ;
  }
}
```

```js
apply(store) {
  const errors = [];
  if(some_condition) {
     errors.push(new Error('Some error in validation.')); 
  }
  return errors; 
}
```

This also relates to the fact that `enforcing a return value even empty array`, is also against the Javascript paradigm. Yes its an integral concept of functional programming, but our SDK is not here to enforce people to learn functional programming. They should be able to use the Javascript the way they are already doing.  

2.1.0

### LINKED ISSUE #4862: Add fill block functionality in forger module
### Description

When a system generates a block it should fill the block with payload(transactions) and then create/save the block.
It should create a MaxHeap for the smallest nonce transactions from each account, and pick the highest fee priority to apply.
It should repeat until payload size is close to 15kb.

### Motivation

In order to create/save a block, the system must fill the block with payload(transactions) if any transaction exists during the time of block creation.

### Acceptance Criteria

- get copy of Transactions from TxPool by calling `getProcessableTransactions`
- Pick highest fee priority transaction among the smallest nonce transactions from each account
- Apply each transaction one by one, and if it fails, it should not apply any subsequent transactions from the same account
- Validate the block for `Maximum block size: 15KB`
- Test case when no transactions are given, and empty block is created
- Test case when all transactions are valid, block with highest fees in order of nonce is created
- Test case one of the transactions is invalid, block without the invalid transaction is created
- Test case one of the transactions is invalid, and all the subsequent transactions are not included in the block

### Related issues
#4841

nazarhussain:
### What was the problem?

This PR resolves #4862  

### How was it solved?

Added forging strategy to select transactions with highest fee. 

### How was it tested?

Run all tests.


framework/src/application/node/forger/strategies.js
```diff
@@ -0,0 +1,118 @@
...
+
+		// Initialize state store which will be discarded after selection
+		const stateStore = this.chainModule.newStateStore();
+
+		// Get processable transactions from transaction pool
+		const processableTransactions = this.transactionPoolModule.getProcessableTransactions();
+
+		// Convert sender transactions to min heap by nonce
+		const transactionsBySender = Object.keys(processableTransactions).reduce(
+			(obj, senderId) => {
```
ManuGowda:
```suggestion
			(accountAccumulator, address) => {
```

framework/src/application/node/forger/strategies.js
```diff
@@ -0,0 +1,118 @@
...
+				stateStore,
+			);
+
+			// If transaction can't be processed then discard all transactions
+			// from that account as other transactions will be higher nonce
+			if (result.transactionsResponses[0].status !== TransactionStatus.OK) {
+				delete transactionsBySender[lowestNonceHighestFeeTrx.senderId];
+
+				// eslint-disable-next-line no-continue
+				continue;
```
ManuGowda: do we need this? it will continue rite?

framework/src/application/node/forger/strategies.js
```diff
@@ -0,0 +1,118 @@
...
+			}
+
+			// If transaction byte size can't fit in max payload length
+			// then discard all transactions from that account as
+			// other transactions will be higher nonce
+			const trsByteSize = BigInt(lowestNonceHighestFeeTrx.getBytes().length);
+			if (blockPayloadSize + trsByteSize > this.constants.maxPayloadLength) {
```
ManuGowda: will `maxPayloadLength` ever cross the size of the integer? I specifically didn't get why we need `BigInt` in this case?

nazarhussain: Not sure, just used it for consistency. Will move it back to integer.

framework/src/application/node/forger/strategies.js
nazarhussain: Didn't get this point. Do we need what?

framework/src/application/node/forger/index.js
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2019 Lisk Foundation
+ * Copyright © 2020 Lisk Foundation
```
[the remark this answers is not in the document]
nazarhussain: I update those for every file which is changed in PRs.

framework/src/application/node/forger/strategies.js
ManuGowda: the wrong comment, ignore, please :)

framework/src/application/node/forger/strategies.js
ManuGowda: cool.. sometimes I see in other places as well we use `BigInt` even for integer cases, so was just trying to understand the intention. 👍

framework/src/application/node/forger/index.js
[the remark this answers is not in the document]
ManuGowda: I think we should not do that.. we should only use `2020` for new files because we want the license to be associated with old code as well.. so if we change the license implication can be different.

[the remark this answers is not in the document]
nazarhussain: But any source code which is changed in 2020, should have updated license header.

framework/src/application/node/forger/strategies.js
shuse2: yes i think payload length is int32, so number is fine

elements/lisk-chain/src/chain.ts
```diff
@@ -653,6 +653,16 @@ export class Chain {
 		)(transactions, stateStore);
 	}
 
+	public async processTransactionsWithStateStore(
```
shuse2:
```suggestion
	public async applyTransactionsWithStateStore(
```

framework/src/application/node/forger/strategies.js
```diff
@@ -0,0 +1,118 @@
...
+			// Try to process transaction
+			const result = await this.chainModule.processTransactionsWithStateStore(
+				[lowestNonceHighestFeeTrx],
+				stateStore,
+			);
+
+			// If transaction can't be processed then discard all transactions
+			// from that account as other transactions will be higher nonce
+			if (result.transactionsResponses[0].status !== TransactionStatus.OK) {
+				delete transactionsBySender[lowestNonceHighestFeeTrx.senderId];
```
shuse2: in above, the state store is mutated, so it needs to createSnapshot and restore in case of failure

framework/src/application/node/forger/strategies.js
```diff
@@ -0,0 +1,118 @@
...
+			}
+
+			// If transaction byte size can't fit in max payload length
+			// then discard all transactions from that account as
+			// other transactions will be higher nonce
+			const trsByteSize = BigInt(lowestNonceHighestFeeTrx.getBytes().length);
+			if (blockPayloadSize + trsByteSize > this.constants.maxPayloadLength) {
+				delete transactionsBySender[lowestNonceHighestFeeTrx.senderId];
+				// eslint-disable-next-line no-continue
+				continue;
```
shuse2: i think in this case, you can just return all the transaction you have already

framework/src/application/node/forger/strategies.js
```diff
@@ -0,0 +1,118 @@
...
+				obj[senderId] = minHeapByNonce;
+
+				return obj;
+			},
+			{},
+		);
+
+		// Loop till we have last account exhausted to pick transactions
+		while (Object.keys(transactionsBySender).length !== 0) {
+			const feePriorityHeap = new MaxHeap();
```
shuse2: this should be out side of the loop so that you don't have to recreate the heap every time

framework/src/application/node/forger/strategies.js
```diff
@@ -0,0 +1,118 @@
...
+			if (blockPayloadSize + trsByteSize > this.constants.maxPayloadLength) {
+				delete transactionsBySender[lowestNonceHighestFeeTrx.senderId];
+				// eslint-disable-next-line no-continue
+				continue;
+			}
+
+			// Select transaction, increase block size and remove that transaction from heap
+			readyTransactions.push(lowestNonceHighestFeeTrx);
+			blockPayloadSize += trsByteSize;
+			transactionsBySender[lowestNonceHighestFeeTrx.senderId].pop();
```
shuse2: I think if you still have more transaction from the same sender, you want to put that in the heap

framework/test/jest/unit/specs/application/node/forger/forging_specs.js
```diff
@@ -0,0 +1,112 @@
+/*
```
shuse2: can we rename this file to `forging_fixtures.js`? `forging_specs` and `forging.spec` are bit confusing

framework/src/application/node/forger/strategies.js
```diff
@@ -0,0 +1,118 @@
...
+		chainModule,
+		transactionPoolModule,
+		// constants
+		maxPayloadLength,
+	}) {
+		this.chainModule = chainModule;
+		this.transactionPoolModule = transactionPoolModule;
+		this.logger = logger;
+
+		this.constants = { maxPayloadLength: BigInt(maxPayloadLength) };
```
shuse2: This also doesn't need to be bigint

framework/src/application/node/forger/strategies.js
nazarhussain: There can be any transaction from another account that can fit in the available space.

framework/src/application/node/forger/strategies.js
nazarhussain: We have to re-create this heap every time, as the lowest nonce transaction per account can change in each iteration. This heap actually contains once lowest transaction per account. We can use an array with sort, but that's another point of not using heap.

framework/src/application/node/forger/strategies.js
nazarhussain: That we do in start of the loop.

framework/test/jest/unit/specs/application/node/forger/forging_specs.js
nazarhussain: Sure.

framework/src/application/node/forger/strategies.js
nazarhussain: Ok

framework/src/application/node/forger/strategies.js
shuse2: yes, but when you choose the transaction from the account, you can just pick the next one from the same account and insert it into the heap?

framework/src/application/node/forger/strategies.js
shuse2: that's true, but i wasnt sure how feasible it is that to happen. in case we leave 1 byte, then we will go through all of the transaction to see if it fits
that's why i thought it's safer to exit here than continuing

framework/src/application/node/forger/strategies.js
nazarhussain: Yes, we can do it but to manage different scenarios, when to re-create the heap and when to put the same transaction in existing heap will make it more complex to maintain.

shuse2: from my understanding, you never need to recreate the heap.
My understanding what need to happen is this, isn't it?
```
1. create max heap
2. put the first transaction from each account
3. peek max heap
4. try apply if it's ok, pop and add next transaction from the same account. if invalid, pop and skip inserting all transactions from the same account
5. repeat 3
```

framework/src/application/node/forger/strategies.js
nazarhussain: We can add a threshold limit, but the original idea of the task was to maximize the forging fee. So I tried to fill in the block every possible way.

framework/src/application/node/forger/strategies.js
nazarhussain: It is restored inside the chain. 

https://github.com/LiskHQ/lisk-sdk/blob/71bdad9dfbc755a1a3f26e25ea3e9160f827d10f/elements/lisk-chain/src/transactions/transactions_handlers.ts#L114-L116

shuse2: ahh ok, that's nice 👍

framework/src/application/node/forger/strategies.js
```diff
@@ -0,0 +1,132 @@
...
+		let feePriorityHeap = createFeePriorityHeap(transactionsBySender);
+
+		// Loop till we have last account exhausted to pick transactions
+		while (Object.keys(transactionsBySender).length !== 0) {
+			// Get the transaction with highest fee and lowest nonce
+			const lowestNonceHighestFeeTrx = feePriorityHeap.pop().value;
+
+			// Try to process transaction
+			const result = await this.chainModule.applyTransactionsWithStateStore(
+				[lowestNonceHighestFeeTrx],
```
shuse2: I think this should be all the transaction that you collected so far and the next transaction

framework/src/application/node/forger/strategies.js
shuse2: I agree that it should maximize the fee, but it would be too time consuming to go through and apply all transactions, in the worst case scenario it would be 4k transactions

framework/src/application/node/forger/strategies.js
nazarhussain: Didn't get it. What do you mean? Its the lowest nonce and highest fee transaction in available transactions.

shuse2: I meant this should be `[... readyTransactions, lowestNonceHighestFeeTrx]`

nazarhussain: Since we are not reverting the stateStore, so don't need to process all transactions together every time.

shuse2: ah thats true good point

framework/src/application/node/forger/strategies.js
```diff
@@ -0,0 +1,130 @@
...
+				stateStore,
+			);
+
+			// If transaction can't be processed then discard all transactions
+			// from that account as other transactions will be higher nonce
+			if (result.transactionsResponses[0].status !== TransactionStatus.OK) {
+				delete transactionsBySender[lowestNonceHighestFeeTrx.senderId];
+				feePriorityHeap = createFeePriorityHeap(transactionsBySender);
```
shuse2: why are we recreating the heap here? I think you just need to continue below. (fee priority heap doesn't include any transaction from the same sender

PR #4959, opened 2020-03-11
Update lisk-transaction helpers to use new sign interface - Closes #4953

### LINKED ISSUE #4953: Update lisk-transaction helpers to use new sign interface
### Description

Change `lisk-transactions` helpers to use the new interface for `sign()`

### Motivation
After changes introduced in by [LIP-0017](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0017.md) the interface for `sign()` changed so any code calling `sign()` needs to be updated

### Acceptance Criteria

- rename `signAll()` to `sign()` in `baseTransaction` and `multisignatureTransaction`
- transfer.ts needs to be updated
- register_delegate.ts needs to be updated
- tests will probably break so they also need to be updated

### Related Issues
#4906

pablitovicente:
### What was the problem?

This PR resolves #4953

### How was it solved?

- transfer.ts was updated for supporting new `sign()` interface
- register_delegate.ts was updated for supporting new `sign()` interface
- some missing error checking was added to `sign()`

### How was it tested?

- Unit tests


elements/lisk-transactions/src/register_delegate.ts
```diff
@@ -49,21 +56,37 @@ export const registerDelegate = (
...
+		passphrases,
+		keys,
+		senderPublicKey,
+	} = inputs;
 
 	const transaction = {
 		...createBaseTransaction(inputs),
 		type: 10,
+		// For txs from multisig senderPublicKey must be set before attempting signing
+		senderPublicKey: senderPublicKey ? senderPublicKey : undefined,
```
shuse2: type `senderPublicKey` is `string | undefined` so this equation is not necessary?

elements/lisk-transactions/src/register_delegate.ts
```diff
@@ -26,6 +26,13 @@ export interface RegisterDelegateInputs {
 	readonly networkIdentifier: string;
 	readonly nonce: string;
 	readonly fee: string;
+	readonly senderPublicKey?: string;
+	readonly passphrases?: ReadonlyArray<string>;
+	readonly keys?: {
+		readonly mandatoryKeys: Array<Readonly<string>>;
+		readonly optionalKeys: Array<Readonly<string>>;
+		readonly numberOfSignatures: number;
```
ManuGowda: do we need `numberOfSignatures`?

elements/lisk-transactions/src/transfer.ts
```diff
@@ -33,9 +33,16 @@ export interface TransferInputs {
...
-	readonly passphrase?: string;
 	readonly recipientId?: string;
 	readonly recipientPublicKey?: string;
+	readonly senderPublicKey?: string;
+	readonly passphrase?: string;
+	readonly passphrases?: ReadonlyArray<string>;
+	readonly keys?: {
+		readonly mandatoryKeys: Array<Readonly<string>>;
+		readonly optionalKeys: Array<Readonly<string>>;
+		readonly numberOfSignatures: number;
```
ManuGowda: same here

elements/lisk-transactions/src/transfer.ts
```diff
@@ -140,7 +152,13 @@ export const transfer = (inputs: TransferInputs): Partial<TransactionJSON> => {
 		transactionWithSenderInfo,
 	);
 
-	transferTransaction.sign(passphrase);
+	if (passphrase) {
+		transferTransaction.sign(networkIdentifier, passphrase);
```
shuse2: shouldn't this return here? as it says `// If senderPassphrase is passed in assume only one signature required`?
or maybe add a test where it covers when `passphrase` and `passphrases` are given?

elements/lisk-transactions/src/register_delegate.ts
pablitovicente: Since this was updated recently on your PR I need to check the impact on this one.

elements/lisk-transactions/src/transfer.ts
pablitovicente: Since this was updated recently on your PR I need to check the impact on this one.

elements/lisk-transactions/src/transfer.ts
pablitovicente: Changed to return on each case

elements/lisk-transactions/src/register_delegate.ts
pablitovicente: Removed.

elements/lisk-transactions/src/register_delegate.ts
pablitovicente: Removed from sign() interface, functions and tests

elements/lisk-transactions/src/transfer.ts
pablitovicente: Removed from sign() interface, functions and tests

elements/lisk-transactions/src/transfer.ts
```diff
@@ -140,7 +151,17 @@ export const transfer = (inputs: TransferInputs): Partial<TransactionJSON> => {
...
+	}
+
+	if (passphrases && keys) {
+		transferTransaction.sign(networkIdentifier, undefined, passphrases, keys);
+
+		return transferTransaction.toJSON();
+	}
 
-	return transferTransaction.toJSON();
+	return transaction;
```
shuse2: why we don't return `transactionWithSenderInfo`?

pablitovicente: Changed to return that

PR #4970, opened 2020-03-13
Update config to be consistently camelCase - Closes #4963

### LINKED ISSUE #4963: Update config/constant to use camel case instead of capital snake case
### Description
We always use camelCase consistently, so the genesis config should also reflect the rule.
Mainly,
`framework/src/application/schema/constants_schema.js` and `framework/src/application/schema/application_config_schema.js` are affected.
Also, related config and usage should be updated, too.

### Motivation

Currently in the config we have mixture of camel case and capital snake case

### Acceptance Criteria

- There is no capital snake case from the config anymore
- There is no usages of the capital snake case for the config in the system

shuse2:
### What was the problem?

This PR resolves #4963 

### How was it solved?

- Update "constant" config values to camel case

### How was it tested?

All test should pass and application should run.

### DIFF
--- a/framework/src/application/node/node.js
+++ b/framework/src/application/node/node.js
@@ -68,10 +68,10 @@ module.exports = class Node {
 			}
 
 			if (
-				this.options.forging.waitThreshold >= this.options.constants.BLOCK_TIME
+				this.options.forging.waitThreshold >= this.options.constants.blockTime
 			) {
 				throw Error(
-					`app.node.forging.waitThreshold=${this.options.forging.waitThreshold} is greater or equal to app.genesisConfig.BLOCK_TIME=${this.options.constants.BLOCK_TIME}. It impacts the forging and propagation of blocks. Please use a smaller value for modules.chain.forging.waitThreshold`,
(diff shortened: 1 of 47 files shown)


framework/src/application/node/node.js
```diff
@@ -368,19 +368,18 @@ module.exports = class Node {
...
-			totalAmount: this.options.constants.TOTAL_AMOUNT,
-			blockSlotWindow: this.options.constants.BLOCK_SLOT_WINDOW,
-			epochTime: this.options.constants.EPOCH_TIME,
-			blockTime: this.options.constants.BLOCK_TIME,
+			maxPayloadLength: this.options.constants.maxPayloadLength,
+			maxTransactionsPerBlock: this.options.constants.maxTransactionsPerBlock,
+			activeDelegates: this.options.constants.activeDelegates,
+			rewardDistance: this.options.constants.rewards.distance,
+			rewardOffset: this.options.constants.rewards.offset,
+			rewardMileStones: this.options.constants.rewards.milestones,
```
ishantiw: `rewardMilestones: `

PR #4971, opened 2020-03-13
Integrate new transaction pool implementation and cleanup - Closes #4856

### LINKED ISSUE #4856: Integrate new transaction pool implementation and cleanup
### Description
With the new transaction pool implementation, we need to clean up the existing transaction pool implementation which is no longer needed and integrate new transaction pool implementation.

### Motivation
Maintainability is the key motivation to clean up the existing transaction pool implementation which will be outdated after the new transaction pool implementation.

### Acceptance Criteria
- Remove logic related to `validated`, `verified` and `ready` (transaction-pool)
- Remove get actions related to signature processing, validated transactions, verified transactions, adding multi-sign transactions.
- Remove all the tests related to `validated`, `verified`, `ready` and the get actions.
- Integrate new transaction pool implementation
- Consolidate `processTransactions` and `verifyTransactions` into `applyTransactions`
- Update `addTransaction` and `removeTransaction` to `add` and `remove`
- Use `AddTransactionResponse` to respond for `addTransaction`
```
interface AddTransactionResponse {
	readonly status: Status;
	readonly errors: ReadonlyArray<TransactionError>;
}
```
- Test if transactions are accepted from API and Network
- Test if valid transactions are included in the block

### Related issues
#4841

ManuGowda:
### What was the problem?

This PR resolves #4856 

### How was it solved?
- Removed old transaction pool implementation
- Integrated new transaction pool implementation
- Update test and code reference
### How was it tested?
- Run the node
- Send the transaction to network and check if its get confirmed
- Start two nodes and sync to see if the transactions are processed as expected


framework/src/application/node/node.js
```diff
@@ -387,13 +385,6 @@ module.exports = class Node {
 			// Publish to the outside
 			this.channel.publish('app:newBlock', eventData);
 
-			// Remove any transactions from the pool on new block
-			if (block.transactions.length) {
```
shuse2: this should still happen. It needs to remove transaction from the pool

framework/src/application/node/node.js
```diff
@@ -422,7 +413,10 @@ module.exports = class Node {
 				const transactions = block.transactions
 					.reverse()
 					.map(tx => this.chain.deserializeTransaction(tx));
-				this.transactionPool.onDeletedTransactions(transactions);
+
```
shuse2: instead of looping twice, it's probably good to do below. Also, insertion order doesn't matter anymore so no need to `reverse`
```
for (const transaction of block.transactions) {
  this.transactionPool.add(this.chain.deserializeTransaction(transaction);
}
```

framework/src/application/node/node.js
```diff
@@ -596,14 +584,6 @@ module.exports = class Node {
 			},
 		);
 
-		this.transactionPool.on(EVENT_UNCONFIRMED_TRANSACTION, transaction => {
-			this.logger.trace(
-				{ transactionId: transaction.id },
-				'Received EVENT_UNCONFIRMED_TRANSACTION',
-			);
-			this.transport.handleBroadcastTransaction(transaction);
```
shuse2: where do we call `handleBroadcastTransaction` now?

framework/src/application/schema/application_config_schema.js
```diff
@@ -299,13 +299,35 @@ module.exports = {
...
+									maximum: 4096,
+								},
+								maxTransactionsPerAccount: {
+									type: 'integer',
+									maximum: 64,
+								},
+								transactionExpiryTime: {
+									type: 'integer',
+									minimum: 10800000,
+									maximum: 10800000,
```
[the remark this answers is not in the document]
ManuGowda: 1 - 3 hour?

framework/src/application/schema/application_config_schema.js
```diff
@@ -299,13 +299,35 @@ module.exports = {
...
+								},
+								minEntranceFeePriority: {
+									type: 'integer',
+									minimum: 1,
+								},
+								minReplacementFeeDifference: {
+									type: 'integer',
+									minimum: 10,
```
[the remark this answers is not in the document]
ManuGowda: `minReplacementFee` starts with 10 rite?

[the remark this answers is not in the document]
shuse2: default to 10, but it can start even from `0`

framework/src/application/schema/application_config_schema.js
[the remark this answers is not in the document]
shuse2: 👍

elements/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -108,15 +123,35 @@ export class TransactionPool {
...
 		if (this._allTransactions[incomingTx.id]) {
-			return false;
+			debug('Received duplicate transaction', incomingTx.id);
+
+			// Since we receive too many duplicate transactions
+			// To avoid too many errors we are returning Status.OK
+			return { status: Status.OK, errors: [] };
 		}
+
 		// Check for minimum entrance fee to the TxPool and if its low then reject the incoming tx
```
ishantiw: `...minimum entrance fee priority to...`

elements/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -108,15 +123,35 @@ export class TransactionPool {
...
 		incomingTx.feePriority = this._calculateFeePriority(incomingTx);
 		if (incomingTx.feePriority < this._minEntranceFeePriority) {
-			return false;
+			const error = new TransactionPoolError(
+				`Rejecting transaction due to failed minimum entrance fee requirement`,
```
ishantiw: `...entrance fee priority requirement...`

elements/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -108,15 +123,35 @@ export class TransactionPool {
...
+				incomingTx.id,
+				'.fee',
+				incomingTx.feePriority.toString(),
+				this._minEntranceFeePriority.toString(),
+			);
+
+			return { status: Status.FAIL, errors: [error] };
 		}
 
 		// Check if incoming transaction fee is greater than the minimum fee in the TxPool if the TxPool is full
```
ishantiw: same here

elements/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -126,22 +161,30 @@ export class TransactionPool {
...
+
 		this._feePriorityQueue.push(incomingTx.feePriority, incomingTx.id);
 
 		const incomingTxAddress = getAddressFromPublicKey(
 			incomingTx.senderPublicKey,
 		);
 
-		const txResponse = await this._applyFunction([incomingTx]);
-		const txStatus = this._getStatus(txResponse);
+		// _applyFunction is injected from chain module applyTransaction
```
ishantiw: 👍

elements/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -164,14 +207,17 @@ export class TransactionPool {
 			this._calculateFeePriority(incomingTx),
 			incomingTx.id,
 		);
-		// Add the transaction in the _transactionList
-		return this._transactionList[incomingTxAddress].add(
+
+		// Add the transaction to _transactionList as PROCESSABLE
```
ishantiw: it can be unprocessable too

framework/src/application/schema/application_config_schema.js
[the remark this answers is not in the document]
ManuGowda: 🤔 if it starts from 0 then `minEntranceFee` is 1 correct? this doesnt make sense to me.. `minReplacementFee` must be atleast greater than `minEntranceFee`?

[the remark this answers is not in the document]
shuse2: minReplacementFeeDifference is if u have fee 10 beddow transaction, in order to replace, you need to set the fee to at least 20 beddow. So, it is not related to the entranceFee.

Actually setting 0 will not make sense because in order to replace the transaction, it has to have higher fee than original. Therefore, in this case `1` will be the correct minimum value (but it is set to 10 by default in library

framework/src/application/node/node.js
ManuGowda: 🤔 .. we missed the emit event and handling requirement... when we add a transaction to pool.. we should broadcast..

shuse2: it was already implemented 😉

framework/src/application/node/transport/transport.js
```diff
@@ -217,11 +217,14 @@ class Transport {
 
 		const { transactionIds } = data;
 		if (!transactionIds) {
+			// Get processable transactions from pool and collect transactions across accounts
+			// Limit the transactions to send based on releaseLimit
+			const transactionsBySender = this.transactionPoolModule.getProcessableTransactions();
```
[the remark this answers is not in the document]
ManuGowda: https://github.com/LiskHQ/lisk-sdk/issues/4974

framework/src/application/node/node.js
```diff
@@ -385,6 +384,13 @@ module.exports = class Node {
 			// Publish to the outside
 			this.channel.publish('app:newBlock', eventData);
 
+			// Remove any transactions from the pool on new block
+			if (block.transactions.length) {
+				block.transactions.map(tx =>
```
shuse2: same as below, `map` is not returning something so it should be `for ...of`

framework/src/application/node/node.js
```diff
@@ -410,13 +416,11 @@ module.exports = class Node {
...
 
 			if (block.transactions.length) {
-				const transactions = block.transactions
-					.reverse()
-					.map(tx => this.chain.deserializeTransaction(tx));
-
-				transactions.forEach(transaction =>
-					this.transactionPool.add(transaction),
-				);
+				for (const transaction of block.transactions.reverse()) {
```
shuse2:
```suggestion
				for (const transaction of block.transactions) {
```
No need to reverse anymore

framework/src/application/node/transport/transport.js
```diff
@@ -71,7 +71,7 @@ class Transport {
 
 	handleBroadcastTransaction(transaction) {
 		this.broadcaster.enqueueTransactionId(transaction.id);
-		this.channel.publish('app:transactions:change', transaction.toJSON());
+		this.channel.publish('app:transactions:change', transaction);
```
shuse2:
```suggestion
		this.channel.publish('app:transactions:change', transactionto.JSON());
```
this is class instance so it should be serialized

framework/src/application/schema/application_config_schema.js
```diff
@@ -572,7 +587,11 @@ module.exports = {
 					releaseLimit: 25,
 				},
 				transactions: {
-					maxTransactionsPerQueue: 1000,
+					maxTransactions: 4096,
```
shuse2: It doesn't need the default value, as it is set in the library.
If we put it here, it will be duplicate settings.

ManuGowda: wouldn't that mean every other config is set to default and the `transactions` needs to be empty?

shuse2: yes, I think all optional thing shouldn't set as `default` in the config, as it is already defined in the library. Although currently it's mixture.
p2p properties and some properties follows this, but not all

ManuGowda: ok, I will remove here then, and then maybe we can address the p2p config issue in another ticket. but it's good to have consistent approach.

shuse2: 👍 but i meant p2p config and some are following this approach, but not others =)

framework/src/application/node/node.js
```diff
@@ -389,9 +386,11 @@ module.exports = class Node {
 
 			// Remove any transactions from the pool on new block
 			if (block.transactions.length) {
-				this.transactionPool.onConfirmedTransactions(
-					block.transactions.map(tx => this.chain.deserializeTransaction(tx)),
-				);
+				for (const transaction of block.transactions) {
+					this.TransactionPool.remove(
```
shuse2:
```suggestion
					this.transactionPool.remove(
```

PR #4975, opened 2020-03-14
Update api transaction pool - Closes #4853

### LINKED ISSUE #4853: Create get all transactions from the transaction pool to provide to Actions and HTTP API
### Description

With the new dynamic fee transaction pool, the queue system will be different from the current structure.
It should return all transactions stored in the transaction pool.

### Motivation

In the custom module or HTTP API, it needs to show all the stored transactions in TxPool

### Acceptance Criteria

- All transactions from the list are returned as a copy.
- HTTP API should return result in serialized format
- HTTP API endpoint `api/node/transactions` should return all the transactions in the transaction pool
- Remove HTTP API endpoints `api/node/transactions/received|pending|ready|validated|verified`

### Related issues
#4841

shuse2:
### What was the problem?

This PR resolves #4853 

### How was it solved?

- Remove fee from constants
- Fix transactionPool.remove to delete key when empty
- Fix transactionPool.getProcessableTransactions not to return empty array
- Fix endpoint call for the getTransactionsFromPool
- Fix swagger definitions

### How was it tested?

- Manually tested `/api/node/transactions`
- Updated unit test for transaction pool


framework/src/modules/http_api/controllers/node.js
```diff
@@ -15,13 +15,65 @@
...
+				);
+			}
+			return transaction.senderPublicKey === filters.senderPublicKey;
+		});
+	}
+	if (filters.type !== null && filters.type !== undefined) {
+		filteredTransactions = filteredTransactions.filter(
+			transaction => transaction.type === filters.type,
+		);
+	}
```
ManuGowda: do we need the null and undefined check? as the swagger takes care of validation, if it's not taken care it should be done there.

framework/src/modules/http_api/controllers/node.js
```diff
@@ -15,13 +15,65 @@
...
+	if (sort) {
```
ManuGowda: I remember we discussed sorting for `nonce` and by default, it will be `asc` sorted.  are we not adding the `nonce` for sorting?

framework/src/modules/http_api/schema/swagger.yml
```diff
@@ -429,13 +417,11 @@ paths:
           required: false
           type: string
           enum:
-            - amount:asc
-            - amount:desc
             - fee:asc
```
ManuGowda: nonce?

framework/src/modules/http_api/controllers/node.js
shuse2: it is filter, where it might or might not exist, and since this is number, it should check `null` and `undefined`

ManuGowda: then why not `Number.isSafeInteger(filters.type)`?

framework/src/modules/http_api/controllers/node.js
shuse2: I think we decided not to include the nonce because if only nonce is provided, it does not provide useful information.

PR #4977, opened 2020-03-16
Fix genesis block being calculated for burnt fee - Closes #4976

### LINKED ISSUE #4976: Burnt fee is not calculated correctly for genesis block
### Expected behavior
Genesis block should have 0 burnt fee because fee is zero.

### Actual behavior
Burnt fee is recorded

### Steps to reproduce
Start application fresh and check chain_state

shuse2:
### What was the problem?

This PR resolves #4976 

### How was it solved?

- Not update burnt fee if given fee is less than 0

### How was it tested?

- Updated test
- After starting the application fresh, check chain_state table

### DIFF
--- a/elements/lisk-chain/test/unit/process.spec.ts
+++ b/elements/lisk-chain/test/unit/process.spec.ts
@@ -863,6 +863,13 @@ describe('blocks/header', () => {
 			it('should not call account update', async () => {
 				expect(storageStub.entities.Account.upsert).not.toHaveBeenCalled();
 			});
+
+			it('should not update burnt fee on chain state', async () => {
+				const genesisAccountFromStore = await stateStore.chainState.get(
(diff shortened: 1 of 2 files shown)


elements/lisk-chain/test/unit/process.spec.ts
```diff
@@ -863,6 +863,13 @@ describe('blocks/header', () => {
 			it('should not call account update', async () => {
 				expect(storageStub.entities.Account.upsert).not.toHaveBeenCalled();
 			});
+
+			it('should not update burnt fee on chain state', async () => {
```
pablitovicente: Seems this is related to the genesis block so in that case it would be nice to include it as part of the description of the case.

shuse2: the outer scope is `applyGenesis` so it should be clear enough

PR #4987, opened 2020-03-16
Fix to start/stop transaction pool in application/node - Closes #4978

### LINKED ISSUE #4978: Transaction pool should be started in the application
### Expected behavior
`transactionPool.start` and `transactionPool.stop` should be called in the application start and end

### Actual behavior
transaction pool is not started nor stopped

4.0-

shuse2:
### What was the problem?

This PR resolves #4978 

### How was it solved?

Call `transaction.start` and `transaction.stop` when bootstrapping or cleaning.

### How was it tested?

- Added test
- Add log and see reorganize job is running

### DIFF
--- a/framework/src/application/node/node.js
+++ b/framework/src/application/node/node.js
@@ -173,6 +173,7 @@ module.exports = class Node {
 			});
 
 			this.channel.subscribe('app:ready', async () => {
+				await this.transactionPool.start();
 				await this._startForging();
 			});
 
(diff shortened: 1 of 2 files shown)


framework/src/application/node/node.js
```diff
@@ -330,6 +331,7 @@ module.exports = class Node {
 	}
 
 	async cleanup(error) {
+		await this.transactionPool.stop();
```
nazarhussain: We do have a convention of `bootstrap`  and `cleanup` method in every module? Should not we follow same here instead of `start/stop`?

shuse2: I think for p2p, it's start and stop. Also, for p2p and transaction pool, it's running the interval and not just bootstrapping, so it's better to call start than bootstrap IMO

PR #4990, opened 2020-03-16
Fix data consistency with list and pool - Closes #4982

### LINKED ISSUE #4982: Account limit should be respected in transaction pool
### Expected behavior
Transaction pool should only maintain certain amount of transaction per account (by default 64)

### Actual behavior
It holds more than 64 transactions

### Steps to reproduce
Send more than 64 transactions from the same account

4.0-

shuse2:
### What was the problem?

This PR resolves #4982 

### How was it solved?

- Update `transactionList.add` and `transactionList.remove` to return what is removed
- Remove from transaction pool when something is removed

### How was it tested?

- Added test to cover this scenario
- Send transactions more than 64 from the same sender, and observe tx-pool size does not go over 64

### DIFF
--- a/elements/lisk-transaction-pool/src/transaction_pool.ts
+++ b/elements/lisk-transaction-pool/src/transaction_pool.ts
@@ -221,11 +221,6 @@ export class TransactionPool {
 			txStatus === TransactionStatus.PROCESSABLE,
 		);
 
-		if (removedID) {
-			debug('Removing from transaction pool with id', removedID);
-			delete this._allTransactions[removedID];
-		}
-
 		if (!added) {
 			return {
 				status: Status.FAIL,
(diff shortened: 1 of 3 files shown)


elements/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -217,6 +215,29 @@ export class TransactionPool {
...
+			incomingTx,
+			txStatus === TransactionStatus.PROCESSABLE,
+		);
+
+		if (removedID) {
+			debug('Removing from transaction pool with id', removedID);
+			delete this._allTransactions[removedID];
+		}
+
+		if (!added) {
```
ishantiw: Maybe we can move this condition before the above condition and early return error

PR #4991, opened 2020-03-16
Update Framework and Lisk-chain to have updated account - Closes #4924

### LINKED ISSUE #4924: Update account with new properties
### Description
Add properties in account class and in the database schema.
1. votes property is an array with objects { delegateAddress: string, amount: bigint }, and in the JSON format, amount has to be “string”
account.votes property in the database is changed to JSONB
2. unlocking property is an array with objects { delegateAddress: string, amount: bigint, unvoteHeight: number }, and in the JSON format, amount has to be “string”
account.unlocking property in the database is JSONB
3. Update voteWeight to totalVotesReceived. Property should be bigint and in the database it should be string
4. Bundle all the delegate related properties into `delegate` property for both application and database.
which will be
```
delegate: {
	lastForgedHeight: number,
	registeredHeight: number,
	totalVotesReceived: bigint,
	consecutiveMissedBlocks: number,
	isBanned: boolean,
	pomHeights: number[],
}
```

### Motivation
In order to support new DPoS system, account need to have new properties

### Acceptance Criteria
- All properties in the description is added to account class
- All properties in the description is added to mem_account table
- Test all property can be read/write from the database
- Test all property are correctly converted from JSON to instance properties

### Related issues
#4915

shuse2:
### What was the problem?

This PR resolves #4924 

### How was it solved?

Updated all the required fields for the consensus related changes

### How was it tested?

- Corresponding tests are added


elements/lisk-chain/src/account.ts
```diff
@@ -15,73 +15,116 @@ import { AccountJSON } from './types';
 
 export const accountDefaultValues = {
 	publicKey: undefined,
-	// tslint:disable-next-line:no-null-keyword
-	username: null,
-	isDelegate: 0,
 	balance: '0',
+	// tslint:disable-next-line no-null-keyword
+	username: null,
```
pablitovicente: We discussed this offline but I think it would be nice in the future to also move this one to the `delegate` object as it will make more sense there and also upon an eventual DB change it would the delegate properties will be more isolated from the rest. So maybe add a comment about this.

framework/src/application/storage/migrations/node/20200316000000_add_new_dpos_to_mem_accounts.sql
```diff
@@ -0,0 +1,18 @@
...
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+ALTER TABLE "mem_accounts" ADD COLUMN IF NOT EXISTS "totalVotesReceived" BIGINT NOT NULL DEFAULT 0;
+ALTER TABLE "mem_accounts" ADD COLUMN IF NOT EXISTS "delegate" jsonb;
+ALTER TABLE "mem_accounts" ADD COLUMN IF NOT EXISTS "votes" jsonb;
+ALTER TABLE "mem_accounts" ADD COLUMN IF NOT EXISTS "unlocking" jsonb;
```
nazarhussain: I prefer to have atomic migration for each change, not so strict, but see if you can split these to multiple migrations.

framework/src/application/storage/migrations/node/20200316000000_add_new_dpos_to_mem_accounts.sql
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+ALTER TABLE "mem_accounts" ADD COLUMN IF NOT EXISTS "totalVotesReceived" BIGINT NOT NULL DEFAULT 0;
```
nazarhussain: Jut for reference `BIGINT` in PostgreSQL and `bigint` in NodeJS has different ranges. DId we take this point in consideration?

framework/src/application/storage/entities/account.js
```diff
@@ -236,6 +244,26 @@ class ChainAccount extends AccountEntity {
...
+	static _stringifyUnlocking(data) {
+		if (data.unlocking) {
+			return {
+				...data,
+				unlocking: JSON.stringify(data.unlocking),
+			};
+		}
+		return data;
+	}
+
```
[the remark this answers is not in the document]
shuse2: I agree, and i think i made the same comment before, but this pattern was already there for long time, so for now the same pattern is kept

framework/src/application/storage/migrations/node/20200316000000_add_new_dpos_to_mem_accounts.sql
shuse2: yes, and to make the efficient order, I think it has to be BIGINT is the current option (same as voteWight now

framework/src/application/storage/migrations/node/20200316000000_add_new_dpos_to_mem_accounts.sql
shuse2: wouldn't it better if all related changes are in the one file? what's the benefit of splitting to each file?

elements/lisk-chain/src/account.ts
shuse2: With the serialization LIP, the whole thing will change, so I will keep it as it is for now

PR #5003, opened 2020-03-17
Handle block data to log properly and fix event log - Closes #4998

### LINKED ISSUE #4998: TypeError: Do not know how to serialize a BigInt
### Expected behavior

The logger should be able to log all the messages without error

### Actual behavior

```
Message: Do not know how to serialize a BigInt 
 Trace: TypeError: Do not know how to serialize a BigInt
```

also, we see `TRACE undefined: Monitor Bus Channel (module=unknown)` frequently
### Steps to reproduce

`LISK_CONSOLE_LOG_LEVEL=trace node test/test_app`

`4.0.0-alpha.0`

ishantiw:
### What was the problem?

This PR resolves #4998 

### How was it solved?

- [x] Logging only `previousBlockId` instead of all the `previousBlock` data
- [x] Event is a string so logging it as a string Monitor Bus Channel logging

### How was it tested?

`LISK_CONSOLE_LOG_LEVEL=trace node test/test_app` and check the logs

### DIFF
--- a/framework/src/application/node/processor/processor.js
+++ b/framework/src/application/node/processor/processor.js
@@ -187,7 +187,11 @@ class Processor {
 	}
 
 	async create(values) {
-		this.logger.trace({ data: values }, 'Creating block');
+		const { previousBlock, ...restOFValues } = values;
+		this.logger.trace('Creating block', {
+			restOFValues,
+			previousBlockId: previousBlock.id,
+		});
 		const highestVersion = Math.max.apply(null, Object.keys(this.processors));
 		const processor = this.processors[highestVersion];
 		return processor.create.run(values);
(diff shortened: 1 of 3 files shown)


framework/src/application/node/processor/processor.js
```diff
@@ -187,7 +187,11 @@ class Processor {
 	}
 
 	async create(values) {
-		this.logger.trace({ data: values }, 'Creating block');
+		const { previousBlock, ...restOFValues } = values;
+		this.logger.trace('Creating block', {
```
ManuGowda: the data should come first know? and then the message?

ishantiw: somehow it doesn't log if I give it as the second argument, for that, I have to stringify which can be done when using it as the second argument

PR #5013, opened 2020-03-17
Add multisignature information to account endpoint - Closes #5005

### LINKED ISSUE #5005: Add multisignature information to account endpoint
### Expected behavior
Account endpoint should have all the information related to the multi-signature. (mandatory-keys, optional-keys and number of signatures)

### Actual behavior
`/api/accounts` does not have multi-signature information

### Steps to reproduce
Call `/api/accounts`

4.0-

ManuGowda:
### What was the problem?

This PR resolves #5005 

### How was it solved?

- Added mandatory-keys, optional-keys and number of signatures
 
### How was it tested?

- Create a multi-sig account and query `api/accounts`

### DIFF
--- a/framework/src/modules/http_api/controllers/accounts.js
+++ b/framework/src/modules/http_api/controllers/accounts.js
@@ -33,6 +33,7 @@ function accountFormatter(totalSupply, account) {
 		'balance',
 		'nonce',
 		'asset',
+		'keys',
 	]);
 
 	if (account.isDelegate) {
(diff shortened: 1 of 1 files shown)


framework/src/modules/http_api/controllers/accounts.js
```diff
@@ -52,6 +53,10 @@ function accountFormatter(totalSupply, account) {
 		);
 	}
 
+	if (formattedAccount.keys.numberOfSignatures === 0) {
```
pablitovicente: Maybe better leave it empty? That way any one consuming this does not need to handle the missing key and instead decide what to do with an empty object

ManuGowda: 🤔 
if we send empty then all the account would look like 
```
{
  mandatoryKey: [
  ],
  optionalKey: [
   ],
  numberOfSignature: 0  
}
```
And this looks odd.. its like if a user is not delegate we dont send `delegate:{}` info.. so similarly if an account is not multi signature we dont send `keys:{}`.

pablitovicente: Oh ok if that's the pattern used it's fine then.

PR #5020, opened 2020-03-18
Remove transaction id validation - Closes #5009

### LINKED ISSUE #5009: Remove transaction ID validation
### Expected behavior

ID validation should no longer be required in `lisk-transactions` since  transactionId is self-generated.

### Actual behavior

Fails with "Missing required property: id"

### Steps to reproduce

Use tx elements to create a transaction without id property and broadcast to API

4.0-

mitsuaki-u:
### What was the problem?

This PR resolves #5009

### How was it solved?

Remove transactionId validation in schema check

### How was it tested?

npm run test

### DIFF
--- a/elements/lisk-transactions/src/schema.ts
+++ b/elements/lisk-transactions/src/schema.ts
@@ -64,10 +64,6 @@ export const baseTransaction = {
 	type: 'object',
 	required: ['type', 'senderPublicKey', 'fee', 'nonce', 'asset', 'signatures'],
 	properties: {
-		id: {
-			type: 'string',
-			format: 'id',
-		},
 		blockId: {
 			type: 'string',
 			format: 'id',
--- a/framework/src/application/node/transport/transport.js
+++ b/framework/src/application/node/transport/transport.js
@@ -394,8 +394,8 @@ class Transport {
 			throw err;
 		}
 
-		if (this.transactionPoolModule.contains(transactionJSON.id)) {
-			return transactionJSON.id;
+		if (this.transactionPoolModule.contains(transaction.id)) {
+			return transaction.id;
 		}
 
 		// Broadcast transaction to network if not present in pool
--- a/framework/src/application/node/transport/transport.js
+++ b/framework/src/application/node/transport/transport.js
@@ -364,10 +364,6 @@ class Transport {
 	}
 
 	async _receiveTransaction(transactionJSON) {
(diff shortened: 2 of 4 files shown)


elements/lisk-transactions/src/schema.ts
```diff
@@ -64,10 +64,6 @@ export const baseTransaction = {
 	type: 'object',
 	required: ['type', 'senderPublicKey', 'fee', 'nonce', 'asset', 'signatures'],
 	properties: {
-		id: {
```
shuse2: i think it's good to keep this, so that it wont come in wrong format if it exists

framework/src/application/node/transport/transport.js
```diff
@@ -398,6 +394,10 @@ class Transport {
 			throw err;
 		}
 
+		if (this.transactionPoolModule.contains(transactionJSON.id)) {
```
shuse2:
```suggestion
		if (this.transactionPoolModule.contains(transaction.id)) {
```

framework/src/application/node/transport/transport.js
```diff
@@ -398,6 +394,10 @@ class Transport {
...
+			return transactionJSON.id;
```
shuse2:
```suggestion
			return transaction.id;
```
since `transactionJSON` might not contain `id`, it should use `transaction`

framework/src/modules/http_api/schema/swagger.yml
```diff
@@ -1069,23 +1069,13 @@ definitions:
...
     required:
-      - id
       - type
       - senderPublicKey
       - nonce
       - fee
       - asset
       - signatures
     properties:
-      id:
```
shuse2: here also property itself can remain there

elements/lisk-transactions/src/schema.ts
mitsuaki-u: If it exists won't it just be overwritten?

shuse2: yea i guess it does...
but would it accept transaction with `id`?

mitsuaki-u: yes it would accept it and set it in constructor and then just overwrite it later.

framework/src/modules/http_api/schema/swagger.yml
shuse2: then this also does not require change

PR #5022, opened 2020-03-18
Update error message - Closes #5017

### LINKED ISSUE #5017: Invalid error message in TxPool for per account limit
### Expected behavior

When `TxPool` reaches to per account transaction limit, it should show proper message. 

```
Transaction was not added because it exceeds maximum limit per account
```

### Actual behavior

Currently its showing error: 

```
Transaction was not added because of nonce or fee
```

### Steps to reproduce
- Sent more than 64 transactions from the same account. 

4.0.0

shuse2:
### What was the problem?

This PR resolves #5017 

### How was it solved?

Update the error message

### DIFF
--- a/elements/lisk-transaction-pool/src/transaction_pool.ts
+++ b/elements/lisk-transaction-pool/src/transaction_pool.ts
@@ -226,7 +226,7 @@ export class TransactionPool {
 				status: Status.FAIL,
 				errors: [
 					new TransactionPoolError(
-						'Transaction was not added because it exceeds maximum limit per account',
+						'Transaction was not added to the pool as it exceeds maximum transaction limit per account',
 						incomingTx.id,
 					),
 				],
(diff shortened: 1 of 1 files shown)


elements/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -226,7 +226,7 @@ export class TransactionPool {
 				status: Status.FAIL,
 				errors: [
 					new TransactionPoolError(
-						'Transaction was not added because of nonce or fee',
+						'Transaction was not added because it exceeds maximum limit per account',
```
ManuGowda:
```suggestion
						`Transaction was not added to the pool as it exceeds maximum transaction limit per account`,
```

shuse2: I think `` is wrong?

ManuGowda: why?

shuse2: nvm i think i miss read

PR #5024, opened 2020-03-18
Signatures property in transation not checked for empty signature - Closes #5021

### LINKED ISSUE #5021: Signatures property in transation not checked for empty signature
### Expected behavior

Should invalidate `''` signature in `signatures`in `12_multisignature_transaction`

### Actual behavior

Not invalidated and causes error `Argument must be a valid hex string.`

4.0-

pablitovicente:
### What was the problem?

This PR resolves #5021

### How was it solved?

By adding a check for empty string in the `signatures` array for multi signature registration/

### How was it tested?

Unit test added

### DIFF
--- a/elements/lisk-transactions/src/utils/verify.ts
+++ b/elements/lisk-transactions/src/utils/verify.ts
@@ -104,7 +104,7 @@ export const validateKeysSignatures = (
 		if (signatures[i].length === 0) {
 			errors.push(
 				new TransactionError(
-					'All mandatory keys have to sign the transaction.',
+					'Invalid signatures format. signatures should not include empty string.',
 					undefined,
 					'.signatures',
 					signatures.join(','),
--- a/elements/lisk-transactions/src/utils/verify.ts
+++ b/elements/lisk-transactions/src/utils/verify.ts
@@ -98,8 +98,20 @@ export const validateKeysSignatures = (
 	transactionBytes: Buffer,
 ) => {
 	const errors = [];
+
 	// tslint:disable-next-line: prefer-for-of no-let
 	for (let i = 0; i < keys.length; i += 1) {
+		if (signatures[i].length === 0) {
+			errors.push(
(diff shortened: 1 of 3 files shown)


pablitovicente:
> We also need to change for base transaction, in that case, we should not call `validateSignature` for the empty string

This is already done in https://github.com/LiskHQ/lisk-sdk/blob/7809e36d3c1d0cc8ad31aace6b479253a091f660/elements/lisk-transactions/src/utils/verify.ts#L159

pablitovicente:
> > We also need to change for base transaction, in that case, we should not call `validateSignature` for the empty string
> 
> This is already done in
> 
> https://github.com/LiskHQ/lisk-sdk/blob/7809e36d3c1d0cc8ad31aace6b479253a091f660/elements/lisk-transactions/src/utils/verify.ts#L159

had to fix `validateKeysSignatures` so the comment was correct

elements/lisk-transactions/src/utils/verify.ts
```diff
@@ -98,8 +98,20 @@ export const validateKeysSignatures = (
 	transactionBytes: Buffer,
 ) => {
 	const errors = [];
+
 	// tslint:disable-next-line: prefer-for-of no-let
 	for (let i = 0; i < keys.length; i += 1) {
+		if (signatures[i].length === 0) {
+			errors.push(
+				new TransactionError(
+					'All mandatory keys have to sign the transaction.',
```
shuse2:
```suggestion
					'Invalid signatures format. signatures should not include empty string.',
```

shuse2: Also, since it's array, we can `continue` instead of `break`

pablitovicente: but as this are mandatory it doesn't make sense to continue as all need to be present.

Issue #1439, opened 2018-01-29
Consistent and standard error propagation


nazarhussain:
### Scenario
We have different and multi level hierarchies of modules, helpers, logic etc. And each interface layer e.g. “API”, “Websocket” is consuming those modules individually. 

Considering that some error been propagated from an underlying module and catched by “API” or “Websocket” or even on domain level catched by “process”. So we need to answer to these questions. 

- Categorically what type of error it is? 
- What need to be done with this error? 
- If interfered by HTTP Layer what status code to respond? 
- If interfered by websocket, what response need to send? 
- If logging required, what level of logging is appropriate for this error?

Currently we are dealing with all these questions on adhoc basis with redundant code. That cause instability and increase complexity in code. 

### Case Study
Suppose we this method was called 
```
logic.accounts.getAll 
```
With some parameters and in case of any error it is throwing error
```
setImmediate(cb, 'Account#getAll error');
```
Suppose this error catched on API layer, we are not sure its an SQL related error. All we can do is to add condition if(error === “Account#getAll error”) then decide to respond with 500 error code. But such if such SQL error generated by other module with different message, then we have to extend that condition. 

And if some one change the error message, all error catching logic will break. 

### Proposed Solution
So hereby I propose to introduce custom error constructors introduced and used consistently in every place of the application. Those error constructors can have different embed information on definition level, so our interface layers can catch and process those consistently. 

```js
var SQLError = function(message) {
  Error.captureStackTrace(this, this.constructor);
  this.message = message; 
  
  this.httpStatusCode = 500;
  this.log = true;
  this.logLevel = 'debug';
};
SQLError.prototype = new Error();


var ParamValidationError = function(param, message) {
  Error.captureStackTrace(this, this.constructor);
  this.param = param
  this.message = message; 
  
  this.httpStatusCode = 400;
  this.log = true;
  this.logLevel = 'info';
};
ParamValidationError.prototype = new Error();

var RecordNotFoundError = function(message) {
  Error.captureStackTrace(this, this.constructor);
  this.message = message; 
  
  this.httpStatusCode = 404;
  this.log = false;
};
ParamValidationError.prototype = new Error();

module.exports = {
  SQLError: SQLError,
  ParamValidationError: ParamValidationError,
  RecordNotFoundError: RecordNotFoundError
};
```

jondubois:
[...]

PR #5035, opened 2020-03-19
Improve error messages - Closes #5033

### LINKED ISSUE #5033: Improve error messages
### Expected behavior

Error messages display correct grammar and spelling and refer to updated properties.

### Actual behavior

Some error messages are written with incorrect grammar and/or refer to outdated transaction properties

### Steps to reproduce

- Send a transaction with nonce higher than expected and result is: ` "message": "Higher transaction nonce for account: 8765191725758661671L, Tx Nonce: 1, Account Nonce: 0",` it would be better as `"Transaction nonce for account: 8765191725758661671L is higher than expected, Tx Nonce: 1, Account Nonce: 0"`

- Send a transaction with invalid signature and result is: `"dataPath": ".signature"` should be `.signatures`

pablitovicente:
### What was the problem?

This PR resolves #5033

### How was it solved?

By fixing the error messages

### How was it tested?

- Unit test
- Send transaction with invalid signature

### DIFF
--- a/elements/lisk-transactions/src/utils/sign_and_validate.ts
+++ b/elements/lisk-transactions/src/utils/sign_and_validate.ts
@@ -36,7 +36,7 @@ export const validateSignature = (
 			? new TransactionError(
 					`Failed to validate signature ${signature}`,
 					id,
-					'.signature',
+					'.signatures',
 			  )
 			: undefined,
 	};
--- a/elements/lisk-transactions/src/utils/verify.ts
+++ b/elements/lisk-transactions/src/utils/verify.ts
@@ -72,9 +72,9 @@ export const verifyAccountNonce = (
 
 	if (nonce > account.nonce) {
 		return new TransactionError(
-			`Higher transaction nonce for account: ${
+			`Transaction nonce for account: ${
 				account.address
-			}, Tx Nonce: ${nonce.toString()}, Account Nonce: ${account.nonce.toString()}`,
+			} is higher than expected, Tx Nonce: ${nonce.toString()}, Account Nonce: ${account.nonce.toString()}`,
 			id,
 			'.nonce',
 			nonce.toString(),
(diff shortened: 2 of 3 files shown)


elements/lisk-transactions/src/utils/sign_and_validate.ts
```diff
@@ -36,7 +36,7 @@ export const validateSignature = (
 			? new TransactionError(
 					`Failed to validate signature ${signature}`,
 					id,
-					'.signature',
+					'.signatures',
```
ManuGowda: its validating signature, I know where is the confusion coming from.. maybe we can remove the whole data path.. as the error is not corresponding to `signatures` array.

pablitovicente: Hmm I think it's good to indicate to the user what property is the offending one so it can pay attention to it.

JesusTheHun: All utils should not return a `TransactionError` but something else such as `ValidationError` or even a plain `Error`. The framework who then use this package can then catch the error and throw a different one with some context, a `TransactionError` in the case of `BaseTransaction`.
So here it's not really about the message, which is fine, it's about the error handling strategy, which is not fine.
The `TransactionError` thrown in `BaseTransaction` can indeed show the path of the invalid property.
My experience in writing libs and framework shown me a good framework error message should also include any help it can to resolve the problem, such as where the developer can find the right value, or even in case of complicated issue, a link to the documentation where the problem and its solution -if any- is exposed.

shuse2: Also, i think this issue is more general one https://github.com/LiskHQ/lisk-sdk/issues/1439 although the description code itself is bit outdated.
If you have further idea, please add to this issue

Issue #5023, opened 2020-03-18
Do not log private information for blocks in the logger


ishantiw:
### Expected behavior

It should not log information like private keys in the console/logger

### Actual behavior

When creating a new block it logs some sensitive information at trace level that needs to be removed

### Steps to reproduce

Run the application and check the logs at TRACE level

### Which version(s) does this affect? (Environment, OS, etc...)
`4.0.0`

shuse2:
Closed By #5036

PR #5036, opened 2020-03-19
Update transaction pool to emit remove event and log - Closes #5034

### LINKED ISSUE #5034: Add more detailed logging for transaction pool
### Expected behavior
Transaction pool should have logs for removing the transactions asynchronously.

### Actual behavior
There is no logs for removing logs

### Steps to reproduce
Observe logs

4.0-

shuse2:
### What was the problem?

This PR resolves #5034

### How was it solved?

- Update to emit event on asynchronous removal of transaction from pool
- Add logging for the remove transaction
- #5023 is fixed, too

### How was it tested?

- Added test
- Observe addition and removal logs

### DIFF
--- a/elements/lisk-transaction-pool/src/transaction_pool.ts
+++ b/elements/lisk-transaction-pool/src/transaction_pool.ts
@@ -238,12 +238,12 @@ export class TransactionPool {
 
 		if (removedID) {
 			debug('Removing from transaction pool with id', removedID);
-			const removingTx = this._allTransactions[removedID];
+			const removedTx = this._allTransactions[removedID];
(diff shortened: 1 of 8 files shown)


elements/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -235,7 +238,14 @@ export class TransactionPool {
 
 		if (removedID) {
 			debug('Removing from transaction pool with id', removedID);
+			const removingTx = this._allTransactions[removedID];
```
mitsuaki-u: `removedTx`

PR #5043, opened 2020-03-23
Add protocol spec for unlock transaction - Closes #4919

### LINKED ISSUE #4919: Create protocol spec for unlocking transaction
### Description
We need to create protocol spec for the introduction of LIP [Introduce vote locking periods and new vote weight definition]( https://github.com/LiskHQ/lips/blob/master/proposals/lip-0023.md) 

### Motivation
- To validate the implementation of LIP [Introduce vote locking periods and new vote weight definition]( https://github.com/LiskHQ/lips/blob/master/proposals/lip-0023.md) 

### Acceptance Criteria
- Create protocol-spec for the unlocking vote transaction

### Related issues
#4915

shuse2:
### What was the problem?

This PR resolves #4919 

### How was it solved?

- Add protocol spec for unlock transaction
- Add unit test template for unlock transaction


elements/lisk-transactions/test/14_unlock_transaction.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+			it.todo('should return errors');
+		});
+
+		describe('when asset.unlockingObjects includes negative unvoteHeight', () => {
+			it.todo('should return errors');
+		});
+	});
+
+	describe('applyAsset', () => {
+		describe('given the delegate is not being punichsed', () => {
```
MaximeGagnebin:
```suggestion
		describe('given the delegate is not being punished', () => {
```

protocol-specs/generator_outputs/unlock_transaction/unlock_transaction.json
```diff
@@ -0,0 +1,235 @@
...
+	"handler": "unlock_transaction",
+	"testCases": {
+		"input": {
+			"account": {
+				"passphrase": "lava toe nuclear candy erode present guilt develop include type pluck current",
+				"publicKey": "8c3d81b1555fbe4692adfa1026ee21c043633b9369924cf2790e2e0fc6b47a66",
+				"address": "841832338348093031L"
+			},
+			"networkIdentifier": "e48feb88db5b5cf5ad71d93cdcd1d879b6d5ed187a36b0002cc34e0ef9883255",
+			"delegates": [
```
nazarhussain: For a valid unlock transaction, there must be a valid vote transaction with negative vote weight. But the state here shows `account`, `networkIdentifier` and `delegates` as input, we don't have the information of negative votes then how this unlock transaction can be valid?

protocol-specs/generator_outputs/unlock_transaction/unlock_transaction.json
```diff
@@ -0,0 +1,235 @@
...
+					"publicKey": "7cae1f08e4a4a437cffec509951ed1f30451415fff725adaa46a6d8946e95787",
+					"address": "18070133408355683425L"
+				},
+				{
+					"passphrase": "pet later deliver cave weekend shell nerve basket barely tip awful fine",
+					"publicKey": "f7b9ea443bdc180cd4116e2a86e302639b4b41659d818d5011bfff0642453c3a",
+					"address": "3640717344948993040L"
+				},
+				{
+					"passphrase": "episode topic dance ice garbage admit myself wage slim echo owner rifle",
```
nazarhussain: Why we have to store the passphrase for every delegate? I think we just need to know if unlocking for a particular delegate vote is associated with a valid delegate. For that only delegate list of public keys is enough.

protocol-specs/generator_outputs/unlock_transaction/unlock_transaction.json
```diff
@@ -0,0 +1,235 @@
...
+					"address": "12502596496028234907L"
+				},
+				{
+					"passphrase": "purse erase first gallery drama horror gloom abandon cupboard pill twist bitter",
+					"publicKey": "19528c41f749fb0acd840b5349823afea8d96d9380cf4c674a5cf522417a6946",
+					"address": "9098130216893659918L"
+				}
+			]
+		},
+		"output": {
```
nazarhussain: The output should not be the mutated state of the account, rather transaction json? Even if we want to keep this spec, but we still need the spec that how this unlock transaction will impact the account state.

shuse2: The purpose of the protocol spec now for the transaction is to generate the valid transaction that can be used in multiple tests, which is how it's done in other transaction related protocol specs.

Those scenario will be checked in the unit test using this transaction

protocol-specs/generator_outputs/unlock_transaction/unlock_transaction.json
shuse2: For unlock transaction to be valid, it needs unlocking objects in the account, but the state information will be created in the unit test

protocol-specs/generator_outputs/unlock_transaction/unlock_transaction.json
shuse2: I think passphrase is not really needed here. will remove

Issue #4918, opened 2020-03-08
Create protocol spec for vote transaction


ManuGowda:
### Description
We need to create protocol spec for the introduction of LIP [Introduce vote locking periods and new vote weight definition]( https://github.com/LiskHQ/lips/blob/master/proposals/lip-0023.md) 

### Motivation
- To validate the implementation of LIP [Introduce vote locking periods and new vote weight definition]( https://github.com/LiskHQ/lips/blob/master/proposals/lip-0023.md) 

### Acceptance Criteria
- Create protocol-spec for the vote transaction

### Related issues
#4915

PR #5048, opened 2020-03-23
Add new vote transaction - Closes #4925

### LINKED ISSUE #4925: Update the current vote transaction and a helper function to the new protocol
### Description
- Update vote transaction for new DPoS protocol, and update creation function.
- Follow https://github.com/LiskHQ/lips/blob/master/proposals/lip-0023.md#new-vote-transaction-1
- In addition to above specification,
   - Also, “castVote” function also need to be updated.
   - Commander create votes can be removed temporary along with the test, this will be handled in the different issue.
### Motivation
In order to satisfy new DPoS protocol, new vote transaction has to be created

### Acceptance Criteria
- Test case where amount is not multiple of 10 LSK
- Test case where amount is out of range of int64
- Test case where delegate address is not registered delegate
- Test case where sum of upvotes is more than sender balance
- Test case where asset.votes is more than 20
- Test case where asset.votes includes duplicate delegate address
- Test case where asset.votes has invalid sort order
- Test case where unvoting not voted delegate or unvoting more than voted amount
- Test case where unvoting to 0 amount, and account.votes entry is removed
- Test case where unvoting makes account.unlocking more than 20 entries
- Test case where account.votes becomes more than 10 entries
- Test if delegate totalVoteReceived is calculated properly on upvotes and unvotes
 - Test case where delegate goes over/below 1000 LSK totalVotesReceived
 - Test case where delegate already has 1000 LSK totalVotesReceived
- Test case where unvote creates the unlocking object

### Related issues
#4915

shuse2:
### What was the problem?

This PR resolves #4925 

### How was it solved?

- Add new vote transaction based on LIP https://github.com/LiskHQ/lips/blob/master/proposals/lip-0023.md

### How was it tested?

- All test based on protocol spec and unit test template created at #4918  are added


elements/lisk-transactions/src/13_vote_transaction.ts
```diff
@@ -0,0 +1,450 @@
...
+						senderVote => senderVote.delegateAddress !== vote.delegateAddress,
+					);
+				}
+				// Create unlocking object
+				sender.unlocking.push({
+					delegateAddress: vote.delegateAddress,
+					amount: BigInt(-1) * vote.amount,
+					unvoteHeight: store.chain.lastBlockHeader.height + 1,
+				});
+				// Sort account.unlocking
```
MaximeGagnebin: we should also account for the case when the delegate address is the same, the sorting should be lexicographical by (address, unvoteHeight, amount).

elements/lisk-transactions/src/13_vote_transaction.ts
```diff
@@ -0,0 +1,450 @@
...
+					errors.push(
+						new TransactionError(
+							'Cannot upvote which exceeds int64',
+							this.id,
+							'.asset.votes.amount',
+						),
+					);
+				}
+				sender.balance -= vote.amount;
+				if (sender.balance < BigInt(0)) {
```
MaximeGagnebin: should that be MIN_BALANCE? or are we checking for this somewhere else?

elements/lisk-transactions/src/13_vote_transaction.ts
```diff
@@ -0,0 +1,450 @@
...
+		}
+
+		return errors;
+	}
+
+	protected async undoAsset(
+		store: StateStore,
+	): Promise<ReadonlyArray<TransactionError>> {
+		const assetCopy = [...this.asset.votes];
+		// Sort by decending amount
```
MaximeGagnebin:
```suggestion
		// Sort by descending amount
```

elements/lisk-validator/src/constants.ts
```diff
@@ -17,5 +17,9 @@
 // Derived from bignum.fromBuffer(Buffer.from(new Array(8).fill(255))).
 export const MAX_EIGHT_BYTE_NUMBER = '18446744073709551615';
 export const MAX_PUBLIC_KEY_LENGTH = 32;
-export const MAX_INT64 = '9223372036854775807';
-export const MAX_UINT64 = '18446744073709551615';
+export const MAX_INT32 = 2147483647;
+export const MIN_INT32 = MAX_INT32 * -1;
+export const MAX_UINT32 = 4294967295;
+export const MAX_INT64 = BigInt('9223372036854775807');
+export const MIN_INT64 = MAX_INT64 * BigInt(-1);
```
MaximeGagnebin:
```suggestion
export const MIN_INT64 = MAX_INT64 * BigInt(-1)-BigInt(1);
```

MaximeGagnebin: or BigInt('-9223372036854775808')

elements/lisk-transactions/src/13_vote_transaction.ts
shuse2: this check is actually not necessary here. It is checked in the base transaction

elements/lisk-cryptography/src/buffer.ts
```diff
@@ -21,6 +21,7 @@ export const intToBuffer = (
 	value: number | string,
 	byteLength: number,
 	endianness: string = BIG_ENDIAN,
+	signed: boolean = false,
```
pablitovicente: **Not** for this PR but for future tech-debt tasks maybe we should consider having to functions `intToBigEndianBuffer` and `intToLittleEndianBuffer` (though I think we might be moving to only big endianness in which case this comment and any other effort will not be needed) Just for us to think about it as the function is getting too "branchy"

elements/lisk-transactions/src/13_vote_transaction.ts
```diff
@@ -0,0 +1,439 @@
...
+	public constructor(rawTransaction: unknown) {
+		super(rawTransaction);
+		const tx = (typeof rawTransaction === 'object' && rawTransaction !== null
+			? rawTransaction
+			: {}) as Partial<TransactionJSON>;
+		if (tx.asset) {
+			const rawAsset = tx.asset as RawAsset;
+			this.asset = {
+				votes: rawAsset.votes.map(vote => {
+					const amount = isNumberString(vote.amount)
```
pablitovicente: `amount` is marked as required in the schema then do we need this ternary operation? Can we just always cast it to BigInt?

elements/lisk-transactions/src/13_vote_transaction.ts
```diff
@@ -0,0 +1,439 @@
...
+				amount: vote.amount.toString(),
+			})),
+		};
+	}
+
+	protected assetToBytes(): Buffer {
+		const bufferArray = [];
+		for (const vote of this.asset.votes) {
+			const addressBuffer = intToBuffer(
+				vote.delegateAddress.slice(0, -1),
```
pablitovicente: why this slice?

elements/lisk-transactions/src/13_vote_transaction.ts
```diff
@@ -0,0 +1,439 @@
...
+				new TransactionError(
+					'Upvote can only be casted upto 10',
+					this.id,
+					'.asset.votes',
+					upvoteCount,
+					'10',
+				),
+			);
+		}
+		if (downvoteCount > MAX_VOTE) {
```
pablitovicente: don't we need the check for `upvoteCount + downvoteCount < 21`  i.e. no more than 20 votes in total?

elements/lisk-transactions/src/13_vote_transaction.ts
shuse2: This constructor is used before validation, so it should be checked for not throwing

elements/lisk-cryptography/src/buffer.ts
shuse2: yes definitely. I was thinking to refactor but it affects lot of places, and we gonna change with generic serialization anyway, so that's why i kept it for now

elements/lisk-transactions/src/13_vote_transaction.ts
shuse2: total entry check is done by schema =)

pablitovicente: Ahh the schema will catch it I think. Resolving.

elements/lisk-transactions/src/legacy_cast_votes.ts
```diff
@@ -0,0 +1,121 @@
+/*
+ * Copyright © 2019 Lisk Foundation
```
ishantiw: `2020`

elements/lisk-transactions/src/legacy_cast_votes.ts
```diff
@@ -0,0 +1,121 @@
...
+	};
+
+	if (!passphrase) {
+		return transaction;
+	}
+
+	const recipientId = getAddressFromPassphrase(passphrase);
+	const transactionWithSenderInfo = {
+		...transaction,
+		// SenderId and SenderPublicKey are expected to be exist from base transaction
```
ishantiw: `SenderId and SenderPublicKey expected to exist in the base transaction`

elements/lisk-transactions/test/legacy_cast_votes.spec.ts
```diff
@@ -0,0 +1,369 @@
+/*
+ * Copyright © 2019 Lisk Foundation
```
ishantiw: `2020`

elements/lisk-transactions/src/legacy_cast_votes.ts
```diff
@@ -0,0 +1,121 @@
...
+			...transaction.asset,
+			recipientId,
+		},
+		networkIdentifier,
+	};
+
+	const voteTransaction = new VoteTransaction(transactionWithSenderInfo);
+	voteTransaction.sign(networkIdentifier, passphrase);
+
+	return voteTransaction.toJSON();
```
ishantiw: `voteTransaction` are always going to be Partial<TransactionJSON>?  if that's the case then maybe it's better to create a new `VoteTransaction` type

elements/lisk-transactions/src/legacy_cast_votes.ts
shuse2: This is 2019 and its correct, this file will be removed once everything is implemented

elements/lisk-transactions/src/legacy_cast_votes.ts
shuse2: same as above

elements/lisk-transactions/test/legacy_cast_votes.spec.ts
shuse2: same as above =)

elements/lisk-transactions/src/legacy_cast_votes.ts
shuse2: same as above =)

PR #5049, opened 2020-03-24
Add unlock transaction - Closes #4926

### LINKED ISSUE #4926: Create an unlocking transaction and helper function
### Description
- Create unlock transaction for a new DPoS protocol, and create transaction creation function.
- Follow https://github.com/LiskHQ/lips/blob/master/proposals/lip-0023.md#new-unlock-transaction
And https://github.com/LiskHQ/lips/blob/master/proposals/lip-0024.md#update-to-validity-of-unlock-transaction
- Also, “unlockBalance” function to create transaction json also needs to be created.

### Motivation
In order to satisfy new DPoS protocol, new unlock transaction has to be created

### Acceptance Criteria
- Test case where sender does not have corresponding unlocking object
- Test case where sender is unlocking self vote, and not waited locking period
- Test case where sender is unlocking non-self vote, and not waited locking period
- Test case where sender is has multiple same unlocking object, but asset unlocking has more (this can happen if 2 or more transaction is included in the same block)
- Test case where unlocking object has more than 20 entries
- Test case where sender is unlocking self vote and currently being punished, and not waited locking period
- Test case where sender is unlocking non-self vote and the delegate is currently being punished, and not waited locking period

### Related issues
#4915

shuse2:
### What was the problem?

This PR resolves #4926 

### How was it solved?

- Add unlock transaction based on https://github.com/LiskHQ/lips/blob/master/proposals/lip-0023.md and https://github.com/LiskHQ/lips/blob/master/proposals/lip-0024.md

### How was it tested?

- Added test using protocol specs and unit test templates

### Additional Information
- This should be merged after #5048 
- It currently includes commits from #5048


elements/lisk-transactions/src/14_unlock_transaction.ts
```diff
@@ -0,0 +1,307 @@
...
+	protected async undoAsset(
+		store: StateStore,
+	): Promise<ReadonlyArray<TransactionError>> {
+		for (const unlock of this.asset.unlockingObjects) {
+			const sender = await store.account.get(this.senderId);
+
+			sender.balance -= unlock.amount;
+			sender.unlocking.push(unlock);
+			// Resort the unlocking since it's pushed to the end
+			sender.unlocking.sort((a, b) =>
```
MaximeGagnebin: to account for potentially equal addresses, we should sort by lexicographical order of (address, unvoteHeight, amout)

shuse2: Addressed

elements/lisk-transactions/src/14_unlock_transaction.ts
```diff
@@ -0,0 +1,306 @@
...
+						this.id,
+						'.asset.unlockingObjects.unvoteHeight',
+						unlock.unvoteHeight,
+					),
+				);
+			}
+			if (isPunished(sender, delegate, store.chain.lastBlockHeader.height)) {
+				errors.push(
+					new TransactionError(
+						'Delegate is currently being punished',
```
ManuGowda: 'Unlocking is not permitted as delegate is currently being punished' something like this?

elements/lisk-transactions/src/14_unlock_transaction.ts
```diff
@@ -0,0 +1,306 @@
...
+					new TransactionError(
+						'Corresponding unlocking object not found',
+						this.id,
+						'.asset.unlockingObjects',
+					),
+				);
+				continue;
+			}
+			sender.unlocking.splice(unlockIndex, 1);
+			sender.balance += unlock.amount;
```
ManuGowda: so when we unlock we update sender balance with unlock amount, what about updating the delegate vote weight? where does it happen?

elements/lisk-transactions/src/unlock_token.ts
```diff
@@ -0,0 +1,98 @@
...
+
+export const unlockToken = (
+	inputs: UnlockTokenInputs,
+): Partial<TransactionJSON> => {
+	validateInputs(inputs);
+	const { networkIdentifier, passphrase, unlockingObjects } = inputs;
+
+	const transaction = {
+		...createBaseTransaction(inputs),
+		type: 14,
```
ManuGowda:
```suggestion
		type: UnlockTransaction.TYPE,
```

elements/lisk-transactions/test/unlock_token.spec.ts
```diff
@@ -0,0 +1,151 @@
+/*
+ * Copyright © 2019 Lisk Foundation
```
ManuGowda:
```suggestion
 * Copyright © 2020 Lisk Foundation
```

elements/lisk-transactions/src/14_unlock_transaction.ts
```diff
@@ -0,0 +1,306 @@
...
+	readonly delegateAddress: string;
+	readonly amount: bigint;
+	readonly unvoteHeight: number;
+}
+
+export interface UnlockAsset {
+	readonly unlockingObjects: ReadonlyArray<Unlock>;
+}
+
+const voteAssetFormatSchema = {
```
ishantiw: shouldn't it be `unlockAssetFormatSchema`

elements/lisk-transactions/test/unlock_token.spec.ts
```diff
@@ -0,0 +1,151 @@
...
+							...validUnlockTransactionScenario.testCases.output.asset.unlockingObjects.slice(
+								0,
+								19,
+							),
+							{
+								delegateAddress: '123L',
+								amount: '-1000000000',
+								unvoteHeight: 1,
+							},
+						],
```
ManuGowda:
```suggestion
						unlockingObjects: [{
								delegateAddress: '123L',
								amount: '-1000000000',
								unvoteHeight: 1,
							}
						],
```

ManuGowda: since the expectation is against `Amount cannot be less than zero` we dont need the other values.

elements/lisk-transactions/test/14_unlock_transaction.spec.ts
```diff
@@ -13,120 +13,800 @@
...
 
 		describe('when asset.unlockingObjects does not include any unlockingObject', () => {
-			it.todo('should return errors');
+			it('should return errors', async () => {
+				(tx.asset as any).unlockingObjects = [];
+				const { errors, status } = tx.validate();
+				expect(status).toBe(Status.FAIL);
+				expect(errors).toHaveLength(1);
+				expect(errors[0].message).toInclude(
+					'should NOT have fewer than 1 items',
```
ManuGowda: the assertion is bit confusing. when there are no `unlockingObjects` the assertion should be `should have fewer than 1 items`?

elements/lisk-transactions/test/14_unlock_transaction.spec.ts
```diff
@@ -13,120 +13,800 @@
...
+					{
+						delegateAddress: '123L',
+						amount: BigInt(0),
+						unvoteHeight: 2,
+					},
+				];
+				const { errors, status } = tx.validate();
+				expect(status).toBe(Status.FAIL);
+				expect(errors).toHaveLength(1);
+				expect(errors[0].message).toInclude('Amount cannot be less than zero');
```
ManuGowda: `Amount cannot be less than or equal to zero`?

elements/lisk-transactions/test/14_unlock_transaction.spec.ts
```diff
@@ -13,120 +13,800 @@
...
+						expect(errors).toHaveLength(1);
+						expect(errors[0].message).toContain(
+							'Unlocking object has not waited locking period',
+						);
+					});
+				});
 			});
 		});
 
 		describe('given the delegate is currently being punichsed', () => {
```
ManuGowda:
```suggestion
		describe('given the delegate is currently being punished', () => {
```

elements/lisk-transactions/src/14_unlock_transaction.ts
```diff
@@ -0,0 +1,306 @@
...
+			},
+		},
+	},
+};
+
+const SIZE_UINT32 = 4;
+const SIZE_INT64 = 8;
+const SIZE_UINT64 = SIZE_INT64;
+// tslint:disable-next-line no-magic-numbers
+const TEN_UNIT = BigInt(10) * BigInt(10) ** BigInt(8);
```
ishantiw: I don't know its a bit unclear with the constant name, maybe something like `SELF_VOTE_AMOUNT_MULTIPLIER`

elements/lisk-transactions/src/14_unlock_transaction.ts
shuse2: Delegate `totalVotesReceived` is deducted when downvoting (not unlocking

elements/lisk-transactions/test/unlock_token.spec.ts
shuse2: It is slicing and creating a new unlocking object, I think it needs to set up valid object

elements/lisk-transactions/src/14_unlock_transaction.ts
shuse2: This is the `10 LSK` which all the vote needs to be multiply of but i wanted to avoid using LSK in the SDK.

ishantiw: Yes its 10 LSK, but I was thinking that this is a constant that is used in the context of self votes and maybe in future if we will have to change it then we don't have to change the constant name but just the value (although the chances are that it will never be changed)

ishantiw: But not a very important comment, if there is a better name then we can use it otherwise we can keep it as it is

elements/lisk-transactions/test/14_unlock_transaction.spec.ts
shuse2: it's expecting the error, so this is correct?

elements/lisk-transactions/src/14_unlock_transaction.ts
shuse2: I also think this name is not clear, any suggestion? This constant is used in self vote and normal vote

ishantiw: `AMOUNT_MULTIPLIER_FOR_VOTES` or `TEN_TIMES_AMOUNT_FOR_VOTES`? these are the only ones I can think of right now, but if you have any better name then feel free to use it

PR #5051, opened 2020-03-24
Add hash onion function - Closes #4930

### LINKED ISSUE #4930: Create hash onion creation command
### Description
- Create a hash onion creation function.
- `hashonion --output outputFilePath --count=1,000,000 --distance=2000`
- If output is not defined, show in console
- If count or distance are not defined, use above value as default
- Output format should be hash onion, placed in order and the first element should be the one to use at the beginning.
```
{
  count: x,
  distance: y,
  hashes: [],
}
```
Also, hash has to be in sha-256 and 128 most significant bits of the hash result.
### Motivation
With the new DPoS system, tool to output hashonion is required.

### Acceptance Criteria
- Test if correct hash onion is calculable from the hashes
- Test case when output file path is provided

### Related issues
#4915

shuse2:
### What was the problem?

This PR resolves #4930 

### How was it solved?

- Add a command to create hash function

### How was it tested?

- add unit test to check if the hash onion can be recalculated
- `./bin/run hash-onion`
- `./bin/run hash-onion --count=100000 --distance=2000 --output=./sample.json`


elements/lisk-cryptography/src/hash_onion.ts
```diff
@@ -0,0 +1,56 @@
...
+	// tslint:disable-next-line no-let
+	for (let i = 1; i <= count; i += 1) {
+		const nextHash = hash(previousHash).slice(0, HASH_SIZE);
+		if (i % distance === 0) {
+			hashes.unshift(nextHash);
+		}
+		previousHash = nextHash;
+	}
+
+	return hashes;
```
nazarhussain: Don't you think at unshifting array at every iteration is a costly operation. Compared to just keep pushing to array and reverse the array at the end.

nazarhussain: https://stackoverflow.com/questions/44031591/performance-of-array-push-vs-array-unshift

shuse2: yes we shouldn't use unshift

PR #5055, opened 2020-03-24
Fix to get last blocks with offset when deleting a block - Closes #5054

### LINKED ISSUE #5054: last block header while deleting a block should consider 2 blocks before
### Expected behavior
When deleting a block, `lastBlock`  is still the deleting block.
Therefore, it should take 2 blocks before while deleting a block.

### Actual behavior
While deleting a block, it's using the deleting block as lastBlockHeader

4.0-

shuse2:
### What was the problem?

This PR resolves #5054

### How was it solved?

- Add offset while creating the state store
- Use offset 1 while deleting a block
 
### How was it tested?

Added unit test to show which range of the blocks is used

### DIFF
--- a/elements/lisk-chain/src/chain.ts
+++ b/elements/lisk-chain/src/chain.ts
@@ -358,12 +358,12 @@ export class Chain {
 		this.dataAccess.resetBlockHeaderCache();
 	}
 
-	public async newStateStore(heightOffset: number = 0): Promise<StateStore> {
+	public async newStateStore(skipLastHeights: number = 0): Promise<StateStore> {
 		const fromHeight = Math.max(
 			1,
-			this._lastBlock.height - this.constants.stateBlockSize - heightOffset,
+			this._lastBlock.height - this.constants.stateBlockSize - skipLastHeights,
 		);
-		const toHeight = Math.max(this._lastBlock.height - heightOffset, 1);
+		const toHeight = Math.max(this._lastBlock.height - skipLastHeights, 1);
 		const lastBlockHeaders = await this.dataAccess.getBlockHeadersByHeightBetween(
 			fromHeight,
 			toHeight,
--- a/elements/lisk-chain/src/chain.ts
+++ b/elements/lisk-chain/src/chain.ts
@@ -358,12 +358,12 @@ export class Chain {
 		this.dataAccess.resetBlockHeaderCache();
 	}
 
-	public async newStateStore(): Promise<StateStore> {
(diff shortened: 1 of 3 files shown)


elements/lisk-chain/src/chain.ts
```diff
@@ -358,12 +358,12 @@ export class Chain {
 		this.dataAccess.resetBlockHeaderCache();
 	}
 
-	public async newStateStore(): Promise<StateStore> {
+	public async newStateStore(heightOffset: number = 0): Promise<StateStore> {
```
nazarhussain: Please make it named argument. Its super confusing to read the usage without it. 

```
await this.chainModule.newStateStore(1)
```
vs 
```
await this.chainModule.newStateStore({ heightOffset : 1})
```

framework/src/application/node/processor/processor.js
```diff
@@ -335,7 +335,8 @@ class Processor {
 	}
 
 	async _deleteBlock(block, processor, saveTempBlock = false) {
-		const stateStore = await this.chainModule.newStateStore();
+		// Offset must be set to 1, because lastBlock is still this deleting block
+		const stateStore = await this.chainModule.newStateStore(1);
```
nazarhussain: With this usage, it meant that you want to skip the `lastBlock` in `stateStore` cache. But its confusing, because `offset` term is used usually to offset from beginning.  So while reading,

```
await this.chainModule.newStateStore({heightOffset: 5})
```

I would assume that you want to skip the first five block and want to start stateStore from heigh 6. 

We should use some other term based on usage. May be `skipLast`, then it would be more readable. 

```
await this.chainModule.newStateStore({skipLast: 5})
```

shuse2: i will rename the argument to be `skipLastHeights`?

elements/lisk-chain/test/unit/chain.spec.ts
```diff
@@ -271,6 +271,56 @@ describe('chain', () => {
...
+			).toHaveBeenCalledWith(
+				{
+					height_gte: 1,
+					height_lte: 1,
+				},
+				{ limit: null, sort: 'height:desc' },
+			);
+		});
+
+		it('should return with the chain state from last block to last block - 309', async () => {
```
ishantiw: This statement is a bit confusing, maybe `should return the chain state from lastBlock.height to lastBlock.height - 309`

PR #5056, opened 2020-03-24
Update sorting logic to feePriorty than fee in forger strategy - Closes #5053

### LINKED ISSUE #5053: Update sorting logic to feePriorty than fee in forger strategy
### Expected behavior

The forger strategy `HighFeeForgingStrategy` should use `feePriority` to sort transactions, that will increase chances for forging more fee for delegate. 

### Actual behavior

Currently, it's sorting transaction by fee and that will always forge named fee transactions first even though it could be less profitable. 

4.0.0

nazarhussain:
### What was the problem?

This PR resolves #5053

### How was it solved?

Updated the sort property to `feePriority` than `fee`

### How was it tested?

Run unit tests for forging strategy.

### DIFF
--- a/framework/src/application/node/forger/strategies.js
+++ b/framework/src/application/node/forger/strategies.js
@@ -50,7 +50,7 @@ class HighFeeForgingStrategy {
 		const feePriorityHeap = new MaxHeap();
 		for (const senderId of Object.keys(transactionsBySender)) {
 			const lowestNonceTrx = transactionsBySender[senderId][0];
-			feePriorityHeap.push(lowestNonceTrx.fee, lowestNonceTrx);
+			feePriorityHeap.push(lowestNonceTrx.feePriority, lowestNonceTrx);
 		}
 
 		// Loop till we have last account exhausted to pick transactions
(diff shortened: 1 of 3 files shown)


elements/lisk-transaction-pool/src/types.ts
```diff
@@ -29,7 +29,7 @@ export interface TransactionObject {
 	signatures?: ReadonlyArray<string>;
 	readonly type: number;
 	verifiedOnce?: boolean;
-	feePriority?: bigint;
+	feePriority: bigint;
```
shuse2: i think this has to be optional because `feePriority` is not property of transaction. it is not assigned when inserting to the pool

nazarhussain: If we keep it optional and then assume that in response of `getProcessableTransactions` it will be available, that assumption is also wrong.

shuse2: yes i mean not read-only, is fine, but it still should be `optional`.
When inserting the transaction in to the pool the type has to match. However, the transaction class doesn't have this property, so it will not match and not build when framework become typescript

ishantiw: I think we can assume that the `feePriority` coming from `getProcessableTransactions` will always present but when call the add on pool, the incoming tx, may or maynot have this value and for that, we need to keep it `optional` and non-readonly.

PR #5060, opened 2020-03-26
Add vote weight snapshot - Closes #4938

### LINKED ISSUE #4938: Update DPoS library to snapshot totalVotesReceived to voteWeight in consensus state
### Description
Update DPoS library to snapshot totalVotesReceived instead of forgers list at the end of the round.
During the snapshot,
- banned delegate should not be included in the snapshot
- Vote weight of punished delegate should be set to 0
- While undoing, irrelevant snapshots can be removed. (ie: list which will be used more than 2 rounds ahead can be changed, so it can be deleted)
- Save the list with key: round to next round + offset(2)
- Vote weight should be capped to the self vote * 10

### Motivation
In order to calculate the correct forger list with the offset, delegate weight needs to be stored.
### Acceptance Criteria
- Test case where there are banned delegates
- Test case where there are punished delegates
- Test case for first 3 rounds, which should be created at the startup
- Test case when undo, future list is removed.
- Test case where vote weight is capped by self vote

### Related issues
#4915

shuse2:
### What was the problem?

This PR resolves #4938

### How was it solved?

- Update chain to return all the delegates in sorted order
- Create vote weight snapshot
  - If there are not enough standby delegates (less than 2), it chooses top 103
  - If there are more than 2 eligible standby delegate, all standby delegates is added

### How was it tested?
- Add unit test that all scenario ( not enough delegate, including zero vote weight, banned, and more than 2 eligible standby)
- Run the application and see genesis scenario works as expected


elements/lisk-dpos/src/delegates_list.ts
```diff
@@ -121,26 +171,166 @@ export const getForgerPublicKeysForRound = async (
...
+				continue;
+			}
+			const selfVote = account.votes.find(
+				vote => vote.delegateAddress === account.address,
+			);
+			const cappedValue =
+				(selfVote?.amount ?? BigInt(0)) * BigInt(this.voteWeightCapRate);
+			if (account.totalVotesReceived > cappedValue) {
+				account.totalVotesReceived = cappedValue;
+			}
```
MaximeGagnebin: Why do we update `totalVotesReceived` ? Logicaly, `totalVotesReceived` should not change, but `weight` is re-calculated here.

elements/lisk-dpos/src/delegates_list.ts
```diff
@@ -121,26 +171,166 @@ export const getForgerPublicKeysForRound = async (
...
+					voteWeight: account.totalVotesReceived.toString(),
+				});
+				continue;
+			}
+
+			// From here, it's below threshold
+			// Below threshold, but prepared array does not have enough slected delegate
+			if (
+				activeDelegates.length + standbyDelegates.length <
+				this.activeDelegates + this.standbyDelegates
```
MaximeGagnebin:
```suggestion
				standbyDelegates.length < this.standbyDelegates
```

elements/lisk-dpos/src/delegates_list.ts
shuse2: In this selection, account is not set to the state store, so the value itself will not persisted after this function scope.
Therefore, I thought it would be easier to mutate the totalVotesReceived here and sort, then create vote weight entries from there.
We could do it like create the voteWeight list and sort and pick too

elements/lisk-dpos/src/delegates_list.ts
MaximeGagnebin: I think that if we get here in the loop, then `activeDelegates.length = this.activeDelegates`. no?

elements/lisk-dpos/src/delegates_list.ts
MaximeGagnebin: (not very important though)

elements/lisk-dpos/test/unit/vote_weight_snapshot.spec.ts
```diff
@@ -0,0 +1,870 @@
...
+		let delegates: Account[];
+		let block: Block;
+
+		describe('when there are changes in the last block', () => {
+			let updatedDelegate: Account;
+
+			beforeEach(async () => {
+				delegates = getDelegateAccounts(200);
+				for (const delegate of delegates) {
+					delegate.totalVotesReceived = randomBigInt(900, 5000);
```
MaximeGagnebin: hmmmm, so this is the self-votes?

elements/lisk-dpos/test/unit/vote_weight_snapshot.spec.ts
```diff
@@ -0,0 +1,870 @@
...
+									amount: updatedVote,
+								},
+							],
+						},
+					],
+					{ [CONSENSUS_STATE_FORGERS_LIST_KEY]: mockedForgersList },
+				);
+			});
+
+			it('should affect the vote wieghts snapshot created', async () => {
```
MaximeGagnebin:
```suggestion
			it('should affect the vote weights snapshot created', async () => {
```

elements/lisk-dpos/test/unit/vote_weight_snapshot.spec.ts
```diff
@@ -0,0 +1,870 @@
...
+					additionalDelegates.every((delegate: Account) =>
+						voteWeights[0].delegates.find(
+							(d: Account) => d.address === delegate.address,
+						),
+					),
+				).toBeTrue();
+			});
+		});
+
+		describe('when there are non-self voting delegate within the top active delegates', () => {
```
MaximeGagnebin: I'm not sure that this should be tested here? shouldn't that be unit tested when testing the `weight` function? What I mean, is: technically there can't be any non-self voting delegates in the active delegates...

elements/lisk-dpos/test/unit/vote_weight_snapshot.spec.ts
```diff
@@ -0,0 +1,870 @@
...
+						(d: Account) => d.address === nonSelfVotedDelegate.address,
+					),
+				).toBeUndefined();
+				expect(voteWeights[0].delegates).toHaveLength(
+					delegates.length + additionalDelegates.length - 1,
+				);
+			});
+		});
+
+		describe('when there are banned delegates within the top active delegates', () => {
```
MaximeGagnebin:
```suggestion
		describe('when there are banned delegates', () => {
```

MaximeGagnebin: as above, I find the formulation "banned within the top active" a bit weird. You cannot be banned and be part ot the active... but if you're banned, then you shouldn't be selected (for this the test is fine)

elements/lisk-dpos/test/unit/vote_weight_snapshot.spec.ts
MaximeGagnebin: ok, I see that there is actually the multiplication by 10^9 in the `randomBigInt` function. (I personally find this a bit confusing)

elements/lisk-dpos/test/unit/vote_weight_snapshot.spec.ts
shuse2: what do you mean by `weight` function?
Also, isn't it possible to have non-self vote within top 101?

elements/lisk-dpos/test/unit/vote_weight_snapshot.spec.ts
shuse2: maybe ill rename this function to `randomBigIntWithPowerof8`?

MaximeGagnebin: if you want. Or have the function return a random integer and multiply by 10^8 explicitly in this line.
I don't have a strong opinion on this.

elements/lisk-dpos/test/unit/vote_weight_snapshot.spec.ts
MaximeGagnebin: I mean, you have the function
delegate weight = max (10*self-votes, totalVotes)  (or 0 if punished).

This should be tested. Then the selection mechanism is somewhat independent.

Yes, my mistake, you can have a delegate with 0 weight in the top 101.

shuse2: that function is called during the snapshot phase, so that's why it's tested here

elements/lisk-dpos/test/unit/vote_weight_snapshot.spec.ts
shuse2: same as above

elements/lisk-dpos/test/unit/vote_weight_snapshot.spec.ts
shuse2: I updated to top 101 delegates

elements/lisk-dpos/src/delegates_list.ts
```diff
@@ -121,26 +171,165 @@ export const getForgerPublicKeysForRound = async (
...
+		};
+		// Save result to the chain state with round number
+		const voteWeights = await getVoteWeights(stateStore);
+		const voteWeightsIndex = voteWeights.findIndex(vw => vw.round === round);
+		if (voteWeightsIndex > 0) {
+			voteWeights[voteWeightsIndex] = voteWeight;
+		} else {
+			voteWeights.push(voteWeight);
+		}
+		_setVoteWeights(stateStore, voteWeights);
```
ManuGowda:
```suggestion
		const extendedDelegateList = activeDelegates.concat(standbyDelegates);
		const delegateListForARound = {
			round,
			delegates: extendedDelegateList,
		};
		// Save delegateListForARound to the chain state with round number
		const voteWeights = await getVoteWeights(stateStore);
		const voteWeightsIndex = voteWeights.findIndex(vw => vw.round === round);
		if (voteWeightsIndex > 0) {
			voteWeights[voteWeightsIndex] = delegateListForARound;
		} else {
			voteWeights.push(delegateListForARound);
		}
		_setVoteWeights(stateStore, voteWeights);
```

elements/lisk-dpos/test/unit/apply.spec.ts
```diff
@@ -50,12 +50,15 @@ describe('dpos.apply()', () => {
...
 				getDelegateAccounts: jest.fn().mockResolvedValue([]),
+				getDelegates: jest.fn().mockResolvedValue([]),
 			},
 		};
 
 		dpos = new Dpos({
 			chain: chainStub,
 			activeDelegates: ACTIVE_DELEGATES,
+			// FIXME: this should be updated to consider standby delegate
+			standbyDelegates: 0,
```
ManuGowda: will this be fixed in a separate issue? can we refer the issue if so.

elements/lisk-dpos/test/unit/vote_weight_snapshot.spec.ts
```diff
@@ -0,0 +1,870 @@
+/*
+ * Copyright © 2019 Lisk Foundation
```
ManuGowda:
```suggestion
 * Copyright © 2020 Lisk Foundation
```

elements/lisk-dpos/src/delegates_list.ts
shuse2: This is not delegate list, it is the voteWeights snapshot.  calling it as delegate list will be confusing IMO

ManuGowda: it captures for a round list if delegates ordered by totalVoteWeight (active + standby), so `voteWeight` sounds more like voteWeight of the round. but internally its the list of delegates for a round order by totalVoteWeight, am i correct?

shuse2: vote weight is the effective vote weight per delegate.
Vote weight considers punishment, self vote cap, ban and standby threshold using `totalVotesReceived`.
But this list is not directly used for forging. It processes one more time to the forgers list.
List itself is called `voteWeights`, so one element is `voteWeight`, but i changed the list name to `delegateVoteWeights`

elements/lisk-dpos/src/delegates_info.ts
```diff
@@ -102,13 +107,20 @@ export class DelegatesInfo {
...
+					i - 1,
+				);
 			}
 
 			return false;
 		}
 
 		const round = this.rounds.calcRound(block.height);
 		// Now rewards and fees are distributed every block. Assuming the distrubution is already done
+		// TODO: Remove after implementing new DPoS #4951
```
pablitovicente: Maybe add these TODO items to the acceptance criteria of the referenced issue?

elements/lisk-dpos/src/delegates_list.ts
```diff
@@ -121,26 +171,163 @@ export const getForgerPublicKeysForRound = async (
...
+		debug(`Creating vote weight snapshot for round: ${round}`);
+		// This list is before executing the current block in process
+		const originalDelegates = await this.chain.dataAccess.getDelegates();
+
+		// Merge updated delegate accounts
+		const updatedAccounts = stateStore.account.getUpdated();
+		// tslint:disable-next-line readonly-keyword
+		const updatedAccountsMap: { [address: string]: Account } = {};
+		// Convert updated accounts to map for better search
+		for (const account of updatedAccounts) {
```
pablitovicente: Not possible to filter beforehand?

elements/lisk-dpos/src/delegates_list.ts
```diff
@@ -121,26 +171,163 @@ export const getForgerPublicKeysForRound = async (
...
+
+		const delegateVoteWeights = activeDelegates.concat(standbyDelegates);
+		const voteWeight = {
+			round,
+			delegates: delegateVoteWeights,
+		};
+		// Save result to the chain state with round number
+		const voteWeights = await getVoteWeights(stateStore);
+		const voteWeightsIndex = voteWeights.findIndex(vw => vw.round === round);
+		if (voteWeightsIndex > 0) {
```
shuse2:
```suggestion
		if (voteWeightsIndex > -1) {
```

PR #5062, opened 2020-03-27
Add seed reveal in the block header - Closes #4931

### LINKED ISSUE #4931: Add seedReveal to the block header
### Description
- Add seed reveal to block header.
- seedReveal should be placed right after the property of previousBlockId in getBytes function.
- hash-onion should be added to the config under each delegate information
- hash-onion should be taken from a config file, and if it does not exist in forger info, it should be saved in the forger info. It should assume multiple delegates can be enabled in a node.
- It should remember which hash is used
- When seed reveal is not provided in config, and it doesn't exist in the forger info, it should exit with proper error message
- used seed reveal should be stored in the forger info, and prune on finality

### Motivation
With the new DPoS system, in order to generate valid random seed in a decentralized way, seedReveal is required.

### Acceptance Criteria
- Test if valid byte array is produced with get bytes
- Test case when seed reveal is not provided in config, and it doesn't exist in the forger info
- Test case where seed reveal is provided and it is not in the forger info
- Test case where seed reveal is used, and the stored seed reveal is the next hash

### Related issues
#4915

shuse2:
### What was the problem?

This PR resolves #4931 

### How was it solved?

- Add hash onion config to test app
- Save used hash onion infomation in the database
- Save registered hash onion and when overwritten, cleanup all the used hash onions
- Add seed reveal in the header

### How was it tested?

- Add tests in the forging module
- Check if the seed reveal is persisted
- Add test to check if the seedReveal bytes are in the signed bytes


framework/src/application/schema/application_config_schema.js
```diff
@@ -326,6 +326,24 @@ module.exports = {
 												type: 'string',
 												format: 'publicKey',
 											},
+											hashOnion: {
```
nazarhussain: Hashes should not be added directly to the application config. Our commander generates a JOSN file, and in application config, we should just provide the path to that file. It's up to the user to generate that file with the commander or any other tool. Adding the config directly to application config JSON will make it more coupled and complex.

shuse2: Currently we are expecting the forger to put their info under `forging` (encryptedPassphrase and the public key). Also, encryptedPassphrase is generate by commander or any other tool too.
I don't think it make sense to separate them because both of the information are required when forging.
Maybe it make sense to split the forging information into other config, but let's discuss that separately

framework/src/application/node/forger/forger.js
```diff
@@ -220,7 +233,56 @@ class Forger {
 					)} is not a delegate`,
 				);
 			}
+			// Prepare hash-onion
+			const registeredHashOnion = registeredHashOnions[account.address];
```
nazarhussain: It seems that `registeredHashOnions` are just the seeds. Please rename all occurrences to `registeredOnionSeeds` or `registeredHashOnionSeeds`

framework/src/application/node/forger/forger.js
```diff
@@ -220,7 +233,56 @@ class Forger {
...
+				hashOnionConfig.hashes[hashOnionConfig.hashes.length - 1];
+			if (registeredHashOnion && registeredHashOnion !== configHashOnionSeed) {
+				this.logger.warn(
+					`Hash onion for Account ${account.address} is not the same as previous one. Overwriting with new hash onion`,
+				);
+				usedHashOnions = usedHashOnions.filter(
+					ho => ho.address !== account.address,
+				);
+			}
+			registeredHashOnions[account.address] = configHashOnionSeed;
```
nazarhussain: When overriding hash onion for a delegate, does the above line not come inside if?

framework/src/application/node/forger/forger.js
```diff
@@ -220,7 +233,56 @@ class Forger {
...
+			const highestUsedHashOnion = usedHashOnions.reduce((prev, current) => {
+				if (current.address !== account.address) {
+					return prev;
+				}
+				if (!prev || prev.count < current.count) {
+					return current;
+				}
+				return prev;
+			}, undefined);
```
nazarhussain: This part is pretty confusing. Can you explain the data structure for `usedHashOnions`? 

Why not use simply save the onion seed counter for every delegate as we already storing just the seed. This way we can just pick up the next hash from respective seed onion.

shuse2: usedHashOnion is just an array of object `{ height: number, address: string, count: number }`.
We cannot simply keep the counter because when the block is deleted, we need to revert the counter, but forger info is out of block context. Therefore, it needs to keep what it used, and clean up when finality kicks in

framework/test/jest/unit/specs/application/node/forger/forger.spec.js
```diff
@@ -878,6 +1048,200 @@ describe('forger', () => {
...
+						{
+							count: 7,
+							address: getAddressFromPublicKey(targetDelegate.publicKey),
+							height: lastBlock.height + 1,
+						},
+					]),
+				);
+			});
+
+			it('should remove all used hash onions before finality height', async () => {
```
ManuGowda: I dont see anywhere the assertion for removal, or am I missing something?

framework/src/application/node/forger/forger.js
shuse2: In case nothing is registered, it should also register the seed

framework/test/jest/unit/specs/application/node/forger/forger.spec.js
shuse2: `count 5` is removed in the expect field

PR #5063, opened 2020-03-29
Create protocol spec for calculation of random seed - Closes #4921

### LINKED ISSUE #4921: Create protocol spec for calculation of random seed
### Description
We need to create protocol spec for the introduction of LIP [Use Randao-based scheme to include standby delegates and reorder delegate list](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0022.md)

### Motivation
- To validate the implementation of LIP [Use Randao-based scheme to include standby delegates and reorder delegate list](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0022.md)

### Acceptance Criteria
- Create protocol-spec for the calculation of random seed

### Related issues
#4915

nazarhussain:
### What was the problem?

This PR resolves #4921

### How was it solved?

Created script to generate protocol specs.


protocol-specs/generators/dpos_random_seed_generation/index.js
```diff
@@ -0,0 +1,433 @@
...
+	}, {});
+
+	const seedRevealsForRandomSeed1 = [];
+
+	// From middle of current round to middle of last round
+	for (let i = h + m; i >= h - m; i -= 1) {
+		const { generatorPublicKey, seedReveal } = filteredBlocks[i];
+		const [previousForgedHeight] = forgedBlocksMap[generatorPublicKey];
+
+		// If forged only 1 block in current and last round
```
IkerAlus: this comment may be misleading.  It should be something like `If forged only 1 block in round at height i or previous round`. We have to do the seed reveal check respect to the considered block.

protocol-specs/generators/dpos_random_seed_generation/index.js
```diff
@@ -0,0 +1,433 @@
...
+				output: {
+					randomSeed1,
+					randomSeed2,
+				},
+			},
+		];
+	})(),
+});
+
+const randomSeedSecondRound = () => ({
```
IkerAlus: I would change this case so it covers the general case (when everything is ok). So it should be at a generic round r (when r > 3) and considering the corresponding seedReveal values in the past to   check if they are valid in generate the next random seeds.

protocol-specs/generators/dpos_random_seed_generation/index.js
```diff
@@ -0,0 +1,471 @@
...
+		return hexArray[0];
+	}
+
+	const buffers = hexArray.map(hexStr => Buffer.from(hexStr, 'hex'));
+	const bufferSizes = new Set(buffers.map(buffer => buffer.length));
+	if (bufferSizes.size > 1) {
+		throw new Error('All input for XOR should be same size');
+	}
+	const outputSize = [...bufferSizes][0];
+	const result = Buffer.alloc(outputSize, 0, 'hex');
```
shuse2: do we need to initialize with 0 as hex? I think it's sufficient to do alloc (it will be initialize with 0)

protocol-specs/generators/dpos_random_seed_generation/index.js
```diff
@@ -0,0 +1,471 @@
...
+		input = Buffer.from(data, 'hex');
+	} else if (data instanceof Buffer) {
+		input = data;
+	} else {
+		throw new Error('Hash input is not a valid type');
+	}
+
+	return hash(input)
+		.slice(0, 16)
+		.toString('hex');
```
shuse2: can we handle this as Buffer instead of converting to hex string?

protocol-specs/generators/dpos_random_seed_generation/index.js
```diff
@@ -0,0 +1,471 @@
...
+
+	return result.toString('hex');
+};
+
+const generateSeedOnion = (initialSeed, size) => {
+	const seeds = new Array(size);
+	seeds[0] = hash(initialSeed, 'hex');
+
+	for (let i = 1; i < size; i += 1) {
+		seeds[i] = strippedHash(seeds[i - 1], 'hex');
```
shuse2: there is no second parameter for `strippedHash`, also no need to convert it as `hex` string.
Also, this initial seeds[0] is overwritten here

protocol-specs/generators/dpos_random_seed_generation/index.js
```diff
@@ -0,0 +1,471 @@
...
+	const bufferSizes = new Set(buffers.map(buffer => buffer.length));
+	if (bufferSizes.size > 1) {
+		throw new Error('All input for XOR should be same size');
+	}
+	const outputSize = [...bufferSizes][0];
+	const result = Buffer.alloc(outputSize, 0, 'hex');
+
+	for (let i = 0; i < outputSize; i += 1) {
+		// eslint-disable-next-line no-bitwise
+		result[i] = buffers.map(b => b[i]).reduce((a, b) => a ^ b, 0);
```
shuse2: would it be correct to XOR with `0` value for the first element?

protocol-specs/generators/dpos_random_seed_generation/index.js
```diff
@@ -0,0 +1,471 @@
...
+		return {
+			generatorPublicKey: delegateList[index % numberOfDelegates].publicKey,
+			height,
+			seedReveal,
+		};
+	});
+};
+
+const calcRound = (height, blocksPerRound) =>
+	Math.floor((height - 1) / blocksPerRound) + 1;
```
shuse2:
```suggestion
	Math.ceil(height / blocksPerRound);
```

protocol-specs/generators/dpos_random_seed_generation/index.js
```diff
@@ -0,0 +1,471 @@
...
+		throw new Error(
+			`Random seed can't be calculated earlier in a round. Wait till you pass middle of round. Current height ${lastBlockHeight}`,
+		);
+	}
+
+	if (
+		lastBlockHeight >= middleBlockHeightOfCurrentRound &&
+		currentRound === 1
+	) {
+		const randomSeed1 = strippedHash(middleThreshold + 1);
```
shuse2: I think it's better to make the input consistent to be Buffer to make it clear how the number is converted to buffer

protocol-specs/generators/dpos_random_seed_generation/index.js
```diff
@@ -0,0 +1,471 @@
...
+/* eslint-disable new-cap, no-continue */
+
+const activeDelegates = 101;
+const standByDelegates = 2;
+
+const strippedHash = data => {
+	let input;
+
+	if (typeof data === 'number') {
+		input = Buffer.from(data.toString(16), 'hex');
```
shuse2: This should be
```
const buffer = Buffer.alloc(4);
buffer.writeUInt32BE(data, 0, 4)
```

protocol-specs/generators/dpos_random_seed_generation/index.js
nazarhussain: Having an explicit init value is safer than considering an implicit default value.

protocol-specs/generators/dpos_random_seed_generation/index.js
nazarhussain: We always have to use it as a string, in seed reveal or random seed. So it's convenient to use hex string output.

protocol-specs/generators/dpos_random_seed_generation/index.js
nazarhussain: The second parameter is true. Will fix. 

`seeds[0]` is not overwritten, as loop starts from 1.

protocol-specs/generators/dpos_random_seed_generation/index.js
nazarhussain: Yes, XOR with zero has no impact.

protocol-specs/generators/dpos_random_seed_generation/index.js
nazarhussain: I felt its consistent to convert the number into buffer in one place which is the function itself. Instead of doing it in multiple places. In the implementation, we can have multiple utility functions if we ant.

protocol-specs/generators/dpos_random_seed_generation/index.js
shuse2: I think this protocol spec especially, it's doing the bitwise operation. It make sense to do it consistently with buffer. `hex` is just a representation. Also, in this file, it's used as buffer mainly anyway

elements/lisk-dpos/test/unit/random_seed.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+				'should include seedReveal if its valid pre-image for block forged upto last round',
+			);
+			it.todo(
+				'should not include seedReveal if its not valid pre-image for lat forged block',
+			);
+			it.todo(
+				'should calculate hash for integer value of height for middle of current round',
+			);
+			it.todo('should perform bitwise exclusive or for above value');
+			it.todo('should return a 128-bit string');
```
shuse2: I think this should return `buffer` , anyway it needs to convert from buffer => bigint in the selection

protocol-specs/generators/dpos_random_seed_generation/index.js
```diff
@@ -0,0 +1,497 @@
...
+		return acc;
+	}, {});
+
+	if (lastBlockHeight < middleBlockHeightOfCurrentRound) {
+		throw new Error(
+			`Random seed can't be calculated earlier in a round. Wait till you pass middle of round. Current height ${lastBlockHeight}`,
+		);
+	}
+
+	if (currentRound === 1) {
```
shuse2: Any specification for first round in LIP, or this is something that came up during the discussion? @IkerAlus

protocol-specs/generators/dpos_random_seed_generation/index.js
```diff
@@ -0,0 +1,497 @@
...
+	handler: 'dpos_random_seed_generation_not_passed_middle_of_round',
+	testCases: (() => {
+		const blocksPerRound = activeDelegates + standByDelegates;
+		const blocks = generateBlocks({
+			startHeight: 1,
+			numberOfBlocks: blocksPerRound + 2,
+			delegateList: sampleDelegateList.slice(0, blocksPerRound),
+		});
+
+		const randomSeed1 = null;
```
shuse2: can we use `undefined` here? for the tslint rule, we have `no-null-keyword`

nazarhussain: I first did `undefined` but it is never get exported to JSON file.

shuse2: ah that's true then it's fine

protocol-specs/generators/dpos_random_seed_generation/index.js
nazarhussain: It came up after discussion, but it's exactly as per rules defined in LIP. If you run those rules for the first round, it will always have this particular output. So why I just excluded this condition for the first round.

nazarhussain:
@shuse2 Discussing few edge cases with @IkerAlus , will update the code firs thing tomorrow morning.

elements/lisk-dpos/test/unit/random_seed.spec.ts
shuse2: @nazarhussain Other than this comment LGTM

nazarhussain: @shuse2 Please check now.

PR #5066, opened 2020-03-30
Update forgers list to take from vote weights list - Closes #4941

### LINKED ISSUE #4941: Update DPoS library to generate forger list from voteWeight list
### Description
- Update delegate selection function in DPoS library following https://github.com/LiskHQ/lips/blob/master/proposals/lip-0022.md#round-computation
- It should use the correct voteWeight with the offset.
- In addition, the created forger list should be stored in the consensus state.

New forgersList should look like
```
[{
  round: 12,
  delegates: [address], // This list should include shuffled information with standby delegate
  standby: [address], // This is separately stored in order to check who is the standby delegate in particular round. This could be empty in some cases where we pick top 103 delegates
}]
```

### Motivation
Shuffling of the delegate must be distributed equally.
### Acceptance Criteria
- Test standby delegate selection using protocol spec if the input and output matches the result
- Test if the forger list is properly saved to the state
- Test case where there is no stand by delegates
- Forger list should include additional information standby delegate address
- Update https://github.com/LiskHQ/lisk-sdk/pull/5060#discussion_r400034777

### Related issues
#4915

shuse2:
### What was the problem?

This PR resolves #4941 

Main changes are
- elements/lisk-dpos/src/delegates_list.ts
- elements/lisk-dpos/src/delegates_info.ts
- framework/src/application/node/forger/forger.js

### How was it solved?

- Update forgers list to be calculated from vote weights

### How was it tested?

- Tested against protocol spec
- Add new unit test scenario
- Update all the tests
- Application should run as expected


elements/lisk-dpos/src/delegates_list.ts
```diff
@@ -328,67 +344,71 @@ export class DelegatesList {
...
-		// tslint:disable-next-line readonly-keyword
-		const updatedAccountsMap: { [address: string]: Account } = {};
-		// Convert updated accounts to map for better search
-		for (const account of updatedAccounts) {
-			updatedAccountsMap[account.address] = account;
+		// tslint:disable-next-line no-magic-numbers
+		if (randomSeed.length !== this.standbyDelegates) {
+			throw new Error(
+				`Exactly ${this.standbyDelegates} random seed must be provided`,
+			);
```
nazarhussain: That leaves the fact that there can be as many random seeds as many standby delegates we had. But the algorithm of calculating random seeds is restricted to two random seeds only, there is no definition of generating `n` number of random seeds.

shuse2: After discussing with research team, I remove the equality check of # of random seed and # of standby delegate
Now only checks the existence and uses modulo to loop through

elements/lisk-dpos/src/delegates_list.ts
```diff
@@ -441,14 +465,14 @@ export class DelegatesList {
...
+		const delegateAddresses = forgersList.find(fl => fl.round === round)
 			?.delegates;
 
-		if (!delegatePublicKeys) {
+		if (!delegateAddresses) {
 			throw new Error(`No delegate list found for round: ${round}`);
 		}
 
-		return shuffleDelegateListForRound(round, delegatePublicKeys);
+		return delegateAddresses;
```
ishantiw, line 447: also here, why we are not shuffling delegate list as the function clearly states that?

elements/lisk-dpos/test/unit/apply.spec.ts
```diff
@@ -271,14 +251,13 @@ describe('dpos.apply()', () => {
...
-				expect(missedBlocks).toEqual(1);
-			}
+			const { missedBlocks } = await stateStore.account.get(
+				missedDelegate.address,
+			);
+			expect(missedBlocks).toEqual(1);
 		});
 
-		it('should save next round + roundOffset active delegates list in frogers list after applying last block of round', async () => {
+		it('should save next round forgers in frogers list after applying last block of round', async () => {
```
ishantiw, line 260: Typo `frogers->forgers`

elements/lisk-dpos/src/delegates_list.ts
```diff
@@ -196,19 +169,19 @@ export const shuffleDelegateListBasedOnRandomSeed = (
...
+	const delegateAddresses = forgersList.find(fl => fl.round === round)
 		?.delegates;
 
-	if (!delegatePublicKeys) {
+	if (!delegateAddresses) {
 		throw new Error(`No delegate list found for round: ${round}`);
 	}
 
-	return shuffleDelegateListForRound(round, delegatePublicKeys);
+	return delegateAddresses;
```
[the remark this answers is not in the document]
shuse2: because when forgersList is saved, it's already shuffled

elements/lisk-dpos/src/delegates_list.ts
shuse2: I was thinking to rename this function, but it is giving shuffled delegateList so i was hesitant to change it... 🤔

ishantiw: In this case, then the function name is a bit misleading, it could be simply `getDelegateList`

PR #5067, opened 2020-03-30
Create proof-of-misbehavior transaction protocol specs - Closes #4920

### LINKED ISSUE #4920: Create protocol spec for proof of misbehavior transaction
### Description
We need to create protocol spec for the introduction of LIP [Punish BFT violations]( https://github.com/LiskHQ/lips/blob/master/proposals/lip-0024.md) 

### Motivation
- To validate the implementation of LIP [Punish BFT violations]( https://github.com/LiskHQ/lips/blob/master/proposals/lip-0024.md) 

### Acceptance Criteria
- Create protocol-spec for the proof of misbehavior transaction
- Generate all 3 cases for contradicting headers

### Related issues
#4915

mitsuaki-u:
### What was the problem?

Proof-of-misbehavior transaction protocol-specs are required to properly test punishing BFT violations.

The following scenarios must be tested:

```
- 1. b1.maxHeightPrevoted == b2.maxHeightPrevoted &&  b1.height >= b2.height

- 2. b1.height > b2.maxHeightPreviouslyForged

- 3. b1.maxHeightPrevoted > b2.maxHeightPrevoted
```

This PR resolves #4920
### How was it solved?

Given a reporting account, forging account and raw transaction, the generator will output signed pom transactions for all 3 scenarios. 

### How was it tested?

Run generator and inspect output.


protocol-specs/generators/proof_of_misbehavior_transaction/index.js
```diff
@@ -0,0 +1,531 @@
...
+
+scenario3Header1.blockSignature = sign(
+	scenario3Header1,
+	forgerKeyPair.privateKeyBytes.toString('hex'),
+);
+
+const scenario3Header2 = {
+	version: 1,
+	timestamp: 1,
+	previousBlockId: 1,
```
shuse2: `previousBlockId` length should be around `10620616195853047363`, and it should be string

protocol-specs/generators/proof_of_misbehavior_transaction/index.js
```diff
@@ -0,0 +1,531 @@
...
+};
+
+scenario3Header1.blockSignature = sign(
+	scenario3Header1,
+	forgerKeyPair.privateKeyBytes.toString('hex'),
+);
+
+const scenario3Header2 = {
+	version: 1,
+	timestamp: 1,
```
shuse2: `timestamp` probably better to be at least `height * 10` and multiple of `10`

protocol-specs/generators/proof_of_misbehavior_transaction/index.js
```diff
@@ -0,0 +1,531 @@
...
+	previousBlockId: 1,
+	seedReveal: '1',
+	height: 200000,
```
shuse2: hight for each block header should be different with some gap so that it's easier to test below condition
```
|header1.height - h| < 260,000.
|header2.height - h| < 260,000.
```

protocol-specs/generators/proof_of_misbehavior_transaction/index.js
```diff
@@ -0,0 +1,531 @@
...
+		'addb0e15a44b0fdc6ff291be28d8c98f5551d0cd9218d749e30ddb87c6e31ca9',
+};
+
+scenario3Header1.blockSignature = sign(
+	scenario3Header1,
+	forgerKeyPair.privateKeyBytes.toString('hex'),
+);
+
+const scenario3Header2 = {
+	version: 1,
```
shuse2: version should be `2` as it will be `2` for this protocol

protocol-specs/generators/proof_of_misbehavior_transaction/index.js
```diff
@@ -0,0 +1,531 @@
...
+*/
+
+const scenario3Header1 = {
+	version: 1,
+	timestamp: 1,
+	previousBlockId: 1,
+	seedReveal: '1',
+	height: 200000,
+	maxHeightPreviouslyForged: 100000,
+	maxHeightPrevoted: 100001,
```
shuse2: I think with these blocks `header1` is b2 and `header2` is b1, so `maxHeightPrevoted` on header 2 should be higher

protocol-specs/generators/proof_of_misbehavior_transaction/index.js
```diff
@@ -0,0 +1,531 @@
...
+
+	b1.maxHeightPrevoted==b2.maxHeightPrevoted &&  b1.height>=b2.height
+*/
+
+const scenario1Header1 = {
+	version: 1,
+	timestamp: 1,
+	previousBlockId: 1,
+	seedReveal: '1',
+	height: 300000,
```
shuse2: I think in this case `header1` will be b2, so the height should be opposite

mitsuaki-u: `header1` is b1 since it has a larger height.

protocol-specs/generators/proof_of_misbehavior_transaction/index.js
mitsuaki-u: it is the opposite, header1 = b1 and header2 = b2.

protocol-specs/generators/proof_of_misbehavior_transaction/index.js
mitsuaki-u: what is `10620616195853047363`?

protocol-specs/generators/proof_of_misbehavior_transaction/index.js
mitsuaki-u: does it matter what the values are for these properties? only the BFT related ones will be tested.

protocol-specs/generators/proof_of_misbehavior_transaction/index.js
mitsuaki-u: does it matter what the values are for these properties and below? only the BFT related ones will be tested.

shuse2: i think type should be correct, otherwise getBytes wont work properly (or at least in the same way), and it's better to use some realistic value IMO
`10620616195853047363` this was some random block id taken from the test

protocol-specs/generators/proof_of_misbehavior_transaction/index.js
shuse2: let scenario1Header1 b1 and scenario1Header2, and 
`(b1.maxHeightPreviouslyForged==b2.maxHeightPreviouslyForged && b1.maxHeightPrevoted==b2.maxHeightPrevoted && b1.height>b2.height)`
when this condition applies, `b1` and `b2` switches

https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#detecting-contradicting-block-headers

elements/lisk-transactions/test/15_pom_transaction.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+			'when header1.height is greater than or equal to header2.height but equal maxHeighPrevoted it should not return errors',
+		);
+
+		it.todo(
+			'when header1.height is greater than header2.maxHeightPreviouslyForged it should not return errors',
+		);
+
+		it.todo(
+			'when header1.maxHeightPrevoted is greater than header2.maxHeightPrevoted it should not return errors',
+		);
```
MaximeGagnebin, line 28: the function checking for contradiction will reorder the headers (according to the rules defined in said function). This order does not have to be respected when the POM is submitted.

What I mean: header1 in the POM might have to be b2 in the bft contradicting function. 

If I can suggest something: implement the BFT contradicting function as in the BFT LIP and call this function. This should already exist as it is used else where in BFT.

elements/lisk-transactions/test/15_pom_transaction.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+
+		it.todo('should return errors if misbehaving account is already banned');
+
+		it.todo(
+			'should return errors if misbehaving account is already punished at height h',
+		);
+
+		it.todo('should return errors if |header1.height - h >= 260000|');
+
+		it.todo('should return errors if |header2.height - h >= 260000|');
```
MaximeGagnebin, line 59:
```suggestion
		it.todo('should return errors if |header1.height - h| >= 260000');

		it.todo('should return errors if |header2.height - h| >= 260000');
```

elements/lisk-transactions/test/15_pom_transaction.spec.ts
```diff
@@ -0,0 +1,74 @@
...
+			'when header1 is not contradicting header 2 it should return errors',
+		);
+
+		it.todo(
+			'when block header 1 or 2 are not properly signed it should return errors',
+		);
+	});
+
+	describe('applyAsset', () => {
+		it.todo('should add reward to balance of the sender');
```
MaximeGagnebin, line 40: I'm not sure where this goes:
"compute reward, which is min(regular block reward at height h, misbehaving delegate balance - MIN_BALANCE)"

protocol-specs/generator_outputs/proof_of_misbehavior_transaction/proof_of_misbehavior_transaction_scenario_1.json
```diff
@@ -0,0 +1,75 @@
...
+				"delegateName": "genesis_100"
+			},
+			"networkIdentifier": "e48feb88db5b5cf5ad71d93cdcd1d879b6d5ed187a36b0002cc34e0ef9883255"
+		},
+		"output": {
+			"senderPublicKey": "c094ebee7ec0c50ebee32918655e089f6e1a604b83bcaa760293c61e0f18ab6f",
+			"nonce": "1",
+			"fee": "1500000000",
+			"type": 15,
+			"asset": {
```
MaximeGagnebin, line 33: the two blocks below are not contracticting. So this transaction should not be valid.

protocol-specs/generator_outputs/proof_of_misbehavior_transaction/proof_of_misbehavior_transaction_scenario_1.json
```diff
@@ -0,0 +1,75 @@
...
+				"header1": {
```
MaximeGagnebin, line 34: I think that these two headers are contradicting because of "scenario 2":
b1.height (=20000) > b2.maxHeightPreviouslyForged (=10000)

protocol-specs/generator_outputs/proof_of_misbehavior_transaction/proof_of_misbehavior_transaction_scenario_3.json
```diff
@@ -0,0 +1,75 @@
...
+			},
+			"networkIdentifier": "e48feb88db5b5cf5ad71d93cdcd1d879b6d5ed187a36b0002cc34e0ef9883255"
+		},
+		"output": {
+			"senderPublicKey": "c094ebee7ec0c50ebee32918655e089f6e1a604b83bcaa760293c61e0f18ab6f",
+			"nonce": "1",
+			"fee": "1500000000",
+			"type": 15,
+			"asset": {
+				"header1": {
```
MaximeGagnebin, line 34: I think that these two headers are contradicting because of "scenario 2":
b1.height (=20000) > b2.maxHeightPreviouslyForged (=10000)

mitsuaki-u:
> (didn't look at generator, only at the outcome)
> I think we only have examples for "scenario2".
> We should also be careful about the order of b1 and b2 when assessing the contradiction of header1 and header2 (they might be in the incorrect order in the transaction).

After discussion with @MaximeGagnebin, for further clarity: the order of b1 and b1 does not matter in the transaction. @shuse2

elements/lisk-transactions/test/15_pom_transaction.spec.ts
mitsuaki-u: yes it will be computed and checked in the `expect` when test is written.

elements/lisk-transactions/test/15_pom_transaction.spec.ts
mitsuaki-u: I will reword it to be general: if one is greater than the other.

PR #5075, opened 2020-03-31
Use consistent naming format for key in key-value table - Closes #5064

### LINKED ISSUE #5064: Use consistent naming format for key in key-value table
### Expected behavior
Currently for BFT, DPoS, forger info, and network_info we have key-value style of the table.
All of the key has different format, but it should be consistent.
The format should be colon separated camel case like `moduleName:keyInfo`

### Actual behavior
All of the key has different format

4.0-

ManuGowda:
### What was the problem?

This PR resolves #5064 

### How was it solved?

- Updated key value to format `modulename:keyName`

### How was it tested?

- Update tests 
- Run application and check database with key

### DIFF
--- a/elements/lisk-chain/src/constants.ts
+++ b/elements/lisk-chain/src/constants.ts
@@ -12,7 +12,7 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
-export const CHAIN_STATE_BURNT_FEE = 'Chain:burntFee';
+export const CHAIN_STATE_BURNT_FEE = 'chain:burntFee';
 
 export const DEFAULT_MIN_BLOCK_HEADER_CACHE = 309;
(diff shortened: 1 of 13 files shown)


elements/lisk-chain/src/constants.ts
```diff
@@ -12,7 +12,7 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
-export const CHAIN_STATE_KEY_BURNT_FEE = 'CHAIN_STATE_KEY_BURNT_FEE';
+export const CHAIN_STATE_BURNT_FEE = 'Chain:burntFee';
```
shuse2:
```suggestion
export const CHAIN_STATE_BURNT_FEE = 'chain:burntFee';
```
same for other places

PR #5082, opened 2020-04-01
Remove old vote transaction and exception handling - Closes #5074

### LINKED ISSUE #5074: Remove old vote transaction
### Description

Remove old vote transaction and expose the new vote transaction and `cast vote` from the transactions

### Motivation

After new vote transaction has been started to work, it should remove the old transactions

### Acceptance Criteria

- Remove old vote transaction
- Remove old castVote function
- Remove old vote transaction handling
- Update relevant tests
- Rename newCastVote, newVoteTransaction where it's used

shuse2:
### What was the problem?

This PR resolves #5074 

### How was it solved?

- Remove old vote transaction and handling of it
- Remove exception handling

### How was it tested?

- All test passes
- Application runs as expected


elements/lisk-transactions/src/11_vote_transaction.ts
```diff
@@ -1,269 +0,0 @@
-/*
```
pablitovicente, line 1: I think it would be better for the new vote transaction to take this number to avoid having a gap in the transaction numbers.

shuse2: yes I think we need to number them, but I was thinking since we have several new transaction coming in we can rename at the end?

shuse2:
if we proceed with https://research.lisk.io/t/introduce-numbering-scheme-for-transaction-types/220, it would be completely different numbering, too
I would wait for the idea to go into the LIP repo to change it
Also, i just realized, but maybe it's bit better if vote and unlock are next to each other coz they are related

pablitovicente:
> if we proceed with https://research.lisk.io/t/introduce-numbering-scheme-for-transaction-types/220, it would be completely different numbering, too
> I would wait for the idea to go into the LIP repo to change it
> Also, i just realized, but maybe it's bit better if vote and unlock are next to each other coz they are related

OK. Then maybe create the issue for the renumbering so it's in the backlog?

elements/lisk-chain/src/transactions/transactions_handlers.ts
```diff
@@ -23,32 +23,14 @@ import { DataAccess } from '../data_access';
...
 	WriteableTransactionResponse,
 } from '../types';
 
-import * as exceptionsHandlers from './exceptions_handlers';
-import * as votesWeight from './votes_weight';
-
-export const validateTransactions = (exceptions?: ExceptionOptions) => (
+export const validateTransactions = () => (
 	transactions: ReadonlyArray<BaseTransaction>,
-): ReadonlyArray<TransactionResponse> => {
```
ManuGowda, line 36: since we no longer have exceptions.. why don't we change function to??
```
export const validateTransactions = (transactions: ReadonlyArray<BaseTransaction>): ReadonlyArray<TransactionResponse> => 
```
```

shuse2: I think this was originally to make all the transaction interface the same.
Also, composeTransactions relies on this

ManuGowda: and the `composeTransactions` was designed to address this exceptions if i understand correctly? I don't have the context though. but if it doesn't make sense now, its better to clean it up now. otherwise you can ignore this comment 👍

shuse2: `composeTransactions` was just a way to handle multiple cases for transaction validation and keeping the `context` while injecting to the transaction pool.
Ill create an issue to address this

shuse2: Created: https://github.com/LiskHQ/lisk-sdk/issues/5088

Issue #4939, opened 2020-03-09
Update DPoS library to calculate random seed


shuse2:
### Description
- Update DPoS library to add function to calculate random seed following https://github.com/LiskHQ/lips/blob/master/proposals/lip-0022.md#random-seeds-computation.
- Using the last 2 rounds of blocks, 2 random seeds should be generated.
### Motivation
In order to add standby delegate, good random seed is required.
### Acceptance Criteria
- Test using protocol spec if the input and output matches the result

### Related issues
#4915

PR #5085, opened 2020-04-01
Update DPoS library to validate seedReveal value - Closes #4937

### LINKED ISSUE #4937: Update DPoS library to validate seedReveal value
### Description
- Update DPoS library to verify seedReveal value following https://github.com/LiskHQ/lips/blob/master/proposals/lip-0022.md#validating-new-block-header-property.
- `isDPoSProtocolComplient(blockHeader: BlockHeader, store: StateStore) => boolean`.
If it’s invalid seed reveal, the reward should be changed to 0 in the block_processor.
- Update `BlockHeader` to include `seedReveal`
- It should use the lastBlockHeaders in the stateStore.consensus.lastBlockHeaders

### Motivation
In order to encourage inserting correct seed reveal, the value has to be checked
### Acceptance Criteria
- Test case where seed reveal does not match with the previous one
- Test case where seed reveal does match with previous one
- Test case where delegate did not forge for last 2 rounds

### Related issues
#4915

### Additional Information
- `isBFTProtocolComplient` already exists and it should follow similar usage

ManuGowda:
### What was the problem?

This PR resolves #4937 

### How was it solved?
- Implemented `isDPoSProtocolCompliant` function in `dpos` library to check for validity of the block
- Followed the specification and implemented https://github.com/LiskHQ/lips/blob/master/proposals/lip-0022.md#validating-new-block-header-property
- Block processor to check for `isDPoSProtocolCompliant` during block creation and validation
### How was it tested?

- Implemented unit test


elements/lisk-dpos/src/dpos.ts
```diff
@@ -208,4 +208,34 @@ export class Dpos {
...
+		}
+
+		const { seedReveal: previousBlockSeedReveal } = delegateForgedBlocks.slice(
+			0,
+			1,
+		)[0];
+		const { seedReveal: newBlockSeedReveal } = blockHeader;
+
+		// Check if last block seedReveal is not a preimage of new block
+		if (previousBlockSeedReveal !== newBlockSeedReveal) {
```
shuse2, line 234:
```suggestion
		if (!hexToBuffer(previousBlockSeedReveal).equals(hash(hexToBuffer(newBlockSeedReveal)).slice(0, 16)) {
```
Where 0, and 16 is defined in #4939

elements/lisk-dpos/src/dpos.ts
```diff
@@ -208,4 +208,34 @@ export class Dpos {
 			delegateListRoundOffset,
 		});
 	}
+
+	// tslint:disable-next-line: prefer-function-over-method
+	public async isDPoSProtocolCompliant(
```
shuse2, line 213: also, can you add this usage of this function in the block_procoessor_v2 where BFTProtocolCompliant is called?

elements/lisk-dpos/src/dpos.ts
```diff
@@ -208,4 +208,34 @@ export class Dpos {
...
+		const delegateForgedBlocks = lastBlockHeaders.filter(
+			block => block.generatorPublicKey === blockHeader.generatorPublicKey,
+		);
+
+		if (!delegateForgedBlocks.length) {
+			// If the forger din't forge any block in the last three rounds
+			return true;
+		}
+
+		const { seedReveal: previousBlockSeedReveal } = delegateForgedBlocks.slice(
```
shuse2, line 227: i think you don't need to slice here?

elements/lisk-dpos/src/dpos.ts
shuse2: Also, please update the relevant test

elements/lisk-dpos/src/dpos.ts
ManuGowda: u mean a comment?

shuse2: I mean in this PR, block_processor_v2 should use this to check the dpos compliance and update/validate the reward

elements/lisk-dpos/src/dpos.ts
```diff
@@ -208,4 +208,34 @@ export class Dpos {
...
+			0,
+			1,
+		)[0];
+		const { seedReveal: newBlockSeedReveal } = blockHeader;
+
+		// Check if last block seedReveal is not a preimage of new block
```
ishantiw, line 233: I think it should be opposite if I understood it correctly,
`// Check if the new block seed reveal is not a preimage of the last block seed reveal`

ishantiw: also, I think it sounds like a correct scenario if it's not equal, maybe emphasis on what it should be,
`New block seed reveal should be a preimage of the last block seed reveal`

elements/lisk-dpos/src/dpos.ts
```diff
@@ -208,4 +209,34 @@ export class Dpos {
...
 		});
 	}
+
+	// This function is used in block_processor_v2 to check the dpos compliance and update/validate the reward
+	// tslint:disable-next-line: prefer-function-over-method
+	public async isDPoSProtocolCompliant(
+		blockHeader: BlockHeader,
+		store: StateStore,
+	): Promise<boolean> {
+		const delegateForgedBlocks = store.consensus.lastBlockHeaders.filter(
```
shuse2, line 219: this line needs to reflect
>  in the previous round or in the same round of B

from https://github.com/LiskHQ/lips/blob/master/proposals/lip-0022.md#validating-new-block-header-property

PR #5094, opened 2020-04-02
Fix variable naming with spelling mistakes - Closes #5093

### LINKED ISSUE #5093: Fix variable naming with spelling mistakes
### Expected behavior
Should resolve most of the spelling mistakes in source code.
### Actual behavior
We have spelling mistake across files
4.0.0

ManuGowda:
### What was the problem?

This PR resolves #5093 

### How was it solved?

- Updated the new block event logger to `debug`
- Fixed spelling mistakes


ManuGowda:
> LGTM although `behaviour` is not really type 😉

true, its between US english vs others. so we have at places `behaviour` so we can keep it that way.

PR #5097, opened 2020-04-03
Cleanup code related to multi-signature and in-memory channel events - Closes #5095

### LINKED ISSUE #5095: Cleanup code related to multi-signature and in-memory channel events
### Description
Cleanup code related to multi-signature and in-memory channel events

### Acceptance Criteria
- Remove `registeredHeight` property from the account
- Remove `dapps:change` event registration/usage from `application.js`
- Remove `signature:change` event registration/usage from `application.js`
- Remove 'multisignatures:change' and `multisignatures:signature:change`  event registration/usage from `application.js`
- Update event `delegates:fork` to `fork`
- Update `delegates:fork` to `fork` and fire this event from `processor.js` in case of `DOUBLE_FORGING`, `DIFFERENT_CHAIN`, `TIE_BREAK` and `DISCARD`
- Update `transaction:change` event name/usage to `newUnconfirmedTransaction` from `application.js`
- Update `processor:sync` event name/usage to `sync` from `application.js`
- Remove `loader:sync` event from `application.js`, and the usage should be listening to the above `sync` event
- Update `processor:broadcast` event name/usage to `broadcast` from `application.js`

ManuGowda:
### What was the problem?

This PR resolves #5095 

### How was it solved?

- Remove unwanted application events
- Remove registeredHeight from delegate accounts
- Update events and publish fork events
### How was it tested?

- Unit test


framework/src/application/application.js
```diff
@@ -538,10 +526,6 @@ class Application {
 				getBlockByHeight: {
 					handler: async action => this._node.actions.getBlockByHeight(action),
 				},
-				getBlocksByHeights: {
```
shuse2, line 541: why remove this?

ManuGowda: it was unused and there was no action present as well.. we have `getBlockByHeight` and `getBlockByHeightBetween` which serve the purpose. this code was a bug we left i guess.

framework/src/application/application.js
```diff
@@ -417,16 +417,11 @@ class Application {
 				'state:updated',
 				'networkEvent',
 				'networkReady',
-				'signature:change',
-				'transactions:change',
+				'newUnconfirmedTransaction',
```
nazarhussain, line 420: What is this event? And since you are refactoring this part, please change names to the pattern of lowercase with scope separated by `:`. So this event will become `transactions:new:unconfirmed`

shuse2: This event is fired when you add the transaction to the pool.
if we gonna follow `:` separated then maybe we should change it to
```
network:event
network:ready
transaction:new
block:new
block:delete
```
?

nazarhussain: Just one thing, we should stick scope to singular or plural but it would be consistent. e.g. We use `rounds`, so we should follow `blocks` and `transactions`. Or vise versa convert all to singular.

PR #5099, opened 2020-04-03
Create Proof-of-Misbheavior transaction - Closes #4927

### LINKED ISSUE #4907: DPoS module unit tests are unstable

### LINKED ISSUE #4927: Create a PoM transaction and helper function
### Description
Create proof of misbehavior transaction for supporting BFT protocol.
Follow https://github.com/LiskHQ/lips/blob/master/proposals/lip-0024.md#proof-of-misbehavior--pom

### Motivation
Refer to https://github.com/LiskHQ/lips/blob/master/proposals/lip-0024.md#motivation

### Acceptance Criteria
- Test case where one of the header is older than 260k blocks
- Test case where both of headers are older than 260k blocks
- Test case where the delegate is already banned
- Test case where the delegate is being punished
- Test case where header signature is invalid
- Test case where the delegate receive 5th PoM transaction
- Test case where delegate does not have/have enough balance

### Related issues
#4915

mitsuaki-u:
### What was the problem?

To solve the PoS “nothing at stake” problem, we introduce a transaction to demonstrate the misbehaving of a delegate as well as rules about the consequences of said misbehavior. In particular, if a delegate violates BFT rules they will be liable to be punished via a reward deduction. The reward is instead given to the sender of the PoM transaction.

This transaction contains the information necessary to prove that a delegate has breached the BFT protocol. Beside the mandatory properties of a transaction (type, senderPublicKey, etc.), the PoM needs the following properties:

    asset.header1: the header of a block.
    asset.header2: the header of a second block.

This PR resolves #4927 

### How was it solved?
- [x]   Implement POM transaction schema validation
- [x]   Implement POM transaction assetToJSON
- [x]   Implement POM transaction assetToBytes
- [x]   Implement POM transaction prepare
- [x]   Implement POM transaction validateAsset
- [x]   Implement POM transaction applyAsset
- [x]   Implement POM transaction undoAsset
- [x]   Test case where one of the header is older than 260k blocks
- [x]   Test case where both of headers are older than 260k blocks
- [x]   Test case where the delegate is already banned
- [x]   Test case where the delegate is being punished
- [x]   Test case where header signature is invalid
- [x]   Test case where the delegate receive 5th PoM transaction
- [x]   Test case where delegate does not have/have enough balance

### How was it tested?

`npm run test`


elements/lisk-transactions/src/15_proof_of_misbehavior_transaction.ts
```diff
@@ -0,0 +1,442 @@
...
+	getBlockBytesWithSignature,
+	isPunished,
+	validateSignature,
+} from './utils';
+
+const proofOfMisbehaviorAssetFormatSchema = {
+	type: 'object',
+	required: ['header1', 'header2'],
+	properties: {
+		header1: {
```
shuse2: `header2` is also needed here.
Maybe define the parameter outside?
```
const headerSchema = {
  type: 'object',
};
```
and then header1: headerSchema, etc

pablitovicente, line 36: header2 is missing?

pablitovicente: Also I suspect both header will have same schema so maybe define a separated schema for header and reference from this one?

elements/lisk-transactions/src/15_proof_of_misbehavior_transaction.ts
```diff
@@ -0,0 +1,442 @@
...
+	public readonly asset: ProofOfMisbehaviorAsset;
+	public static TYPE = 15;
+
+	public constructor(rawTransaction: unknown) {
+		super(rawTransaction);
+		const tx = (typeof rawTransaction === 'object' && rawTransaction !== null
+			? rawTransaction
+			: {}) as Partial<TransactionJSON>;
+		this.asset = (tx.asset || {}) as ProofOfMisbehaviorAsset;
+		this.asset.reward = BigInt(0);
```
shuse2, line 138: it should check if `rawTransaction` have it or not, and initialize with it if it has it, otherwise it will be always 0 when you receive from database

elements/lisk-transactions/src/15_proof_of_misbehavior_transaction.ts
```diff
@@ -0,0 +1,442 @@
...
+					'.asset.header1',
+				),
+			);
+		}
+
+		/*
+            Check for BFT violations:
+                1. Double forging
+                2. Disjointness 
+                3. Branch is not the one with largest maxHeighPrevoted
```
shuse2, line 215: maybe it's better to separate the function of `checkHeadersContradicting(blockHeader1,blockHeader2)`, and write exactly as LIP is written?

elements/lisk-transactions/src/15_proof_of_misbehavior_transaction.ts
```diff
@@ -0,0 +1,442 @@
...
+		const delegateAccount = await store.account.get(delegateId);
+		const senderAccount = await store.account.get(this.senderId);
+		const { networkIdentifier } = store.chain;
+		/*
+			|header1.height - h| < 260,000.
+			|header2.height - h| < 260,000.
+		*/
+
+		// tslint:disable-next-line no-magic-numbers
+		if (Math.abs(this.asset.header1.height - currentHeight) >= 260000) {
```
shuse2, line 272: Prefer to split `260000` as a constant

elements/lisk-transactions/src/15_proof_of_misbehavior_transaction.ts
```diff
@@ -0,0 +1,172 @@
...
+} from './base_transaction';
+import { convertToAssetError, TransactionError } from './errors';
+import { Account, BlockHeader, TransactionJSON } from './transaction_types';
+import { getBlockBytes } from './utils';
+
+const proofOfMisbehaviorAssetFormatSchema = {
+	type: 'object',
+	required: ['header1', 'header2'],
+	properties: {
+		header1: {
```
ManuGowda: maybe be extract this `header1` json schema into a variable 
ex:
const blockHeaderSchema = { key: value }
and then
```
properties: {
  header1: ...blockHeaderSchema,
  header2: ...blockHeaderSchema,
}
?

File: elements/lisk-transactions/src/15_proof_of_misbehavior_transaction.ts [hunk 6]
```diff
@@ -0,0 +1,442 @@
...
+		const blockHeader1Bytes = Buffer.concat([
+			Buffer.from(networkIdentifier, 'hex'),
+			getBlockBytes(this.asset.header1),
+		]);
+		const blockHeader2Bytes = Buffer.concat([
+			Buffer.from(networkIdentifier, 'hex'),
+			getBlockBytes(this.asset.header2),
+		]);
+
+		const { valid: validHeader1Signature } = validateSignature(
```
shuse2, line 350: this might be misleading because the 3rd argument name is `transactionBytes`. 
Maybe create a new function or Update the argument name to `bytes`?

elements/lisk-transactions/src/15_proof_of_misbehavior_transaction.ts
```diff
@@ -0,0 +1,442 @@
...
+
+		store.account.set(senderAccount.address, senderAccount);
+
+		/*
+			Update delegate account
+		*/
+		delegateAccount.delegate.pomHeights.push(currentHeight);
+
+		// tslint:disable-next-line no-magic-numbers
+		if (delegateAccount.delegate.pomHeights.length > 4) {
```
shuse2, line 403: prefer to make `4` as constant

elements/lisk-transactions/src/15_proof_of_misbehavior_transaction.ts
```diff
@@ -0,0 +1,445 @@
...
+			schemaErrors,
+		) as TransactionError[];
+
+		if (
+			this.asset.header1.generatorPublicKey !==
+			this.asset.header2.generatorPublicKey
+		) {
+			errors.push(
+				new TransactionError(
+					'GeneratorPublickey of each blockheader must be matching.',
```
pablitovicente, line 189: "GeneratorPublickey of each blockheader should be the same." or "GeneratorPublickey of each blockheader should match."

elements/lisk-transactions/src/15_proof_of_misbehavior_transaction.ts
```diff
@@ -0,0 +1,442 @@
...
+		senderAccount.balance += reward;
+		// We store the correct reward value in the asset at apply time to allow for undoing the transaction at a later point in time
+		this.asset.reward = reward;
+
+		store.account.set(senderAccount.address, senderAccount);
+
+		/*
+			Update delegate account
+		*/
+		delegateAccount.delegate.pomHeights.push(currentHeight);
```
shuse2, line 400: there is a case where delegateAccount === senderAccount, so before updating the delegate account, it should `get` again after `set` the sender account

elements/lisk-transactions/src/15_proof_of_misbehavior_transaction.ts
```diff
@@ -0,0 +1,445 @@
...
+
+		if (
+			this.asset.header1.generatorPublicKey !==
+			this.asset.header2.generatorPublicKey
+		) {
+			errors.push(
+				new TransactionError(
+					'GeneratorPublickey of each blockheader must be matching.',
+					this.id,
+					'.asset.header1',
```
ManuGowda, line 191:
```suggestion
					this.asset.header1.generatorPublicKey,
					'.asset.header1.generatorPublicKey',
```
?

elements/lisk-transactions/src/15_proof_of_misbehavior_transaction.ts
```diff
@@ -0,0 +1,442 @@
...
+	}
+
+	protected async undoAsset(
+		store: StateStore,
+	): Promise<ReadonlyArray<TransactionError>> {
+		const currentHeight = store.chain.lastBlockHeader.height + 1;
+		const delegateId = getAddressFromPublicKey(
+			this.asset.header1.generatorPublicKey,
+		);
+		const delegateAccount = await store.account.get(delegateId);
```
shuse2, line 419: same as above, there is a case delegate account and sender account is the same.
so get(delegateId) should happen after set(senderAccount)

elements/lisk-transactions/src/15_proof_of_misbehavior_transaction.ts
```diff
@@ -0,0 +1,445 @@
...
+				new TransactionError(
+					'Blockheaders are identical. No contradiction detected.',
+					this.id,
+					'.asset.header1',
+				),
+			);
+		}
+
+		/*
+            Check for BFT violations:
```
pablitovicente, line 212: Maybe move this checks to a private method as in the future might be replaced?

elements/lisk-transactions/src/15_proof_of_misbehavior_transaction.ts
```diff
@@ -0,0 +1,442 @@
...
+				},
+			},
+		},
+	},
+};
+export interface ProofOfMisbehaviorAsset {
+	readonly header1: BlockHeader;
+	readonly header2: BlockHeader;
+	// tslint:disable-next-line readonly-keyword
+	reward?: bigint;
```
shuse2, line 125: when you get the `reward` from the database, it will be in `string`, and when you return as JSON, it should be `string`

elements/lisk-transactions/src/15_proof_of_misbehavior_transaction.ts
mitsuaki-u: I don't think we are changing BFT protocol any time soon, but it can be moved to private later if it happens.

elements/lisk-transactions/src/15_proof_of_misbehavior_transaction.ts
```diff
@@ -0,0 +1,445 @@
...
+
+		return errors;
+	}
+
+	protected async applyAsset(
+		store: StateStore,
+	): Promise<ReadonlyArray<TransactionError>> {
+		const errors = [];
+		const currentHeight = store.chain.lastBlockHeader.height + 1;
+		const delegateId = getAddressFromPublicKey(
```
ManuGowda, line 260:
```suggestion
		const delegateAddress = getAddressFromPublicKey(
```

elements/lisk-transactions/test/15_pom_transaction.spec.ts
```diff
@@ -12,59 +12,459 @@
...
-		it.todo('should return errors if |header2.height - h| >= 260000');
+			const { errors, status } = await transactionWithScenario1.apply(store);
+			expect(status).toBe(Status.FAIL);
+			expect(errors).toHaveLength(3);
+			expect(errors[2].message).toInclude(
+				'Cannot apply proof-of-misbehavior. Delegate is already punished.',
+			);
+		});
+
+		it('should add reward to balance of the sender', async () => {
```
shuse2, line 298: is should test also the case where delegate doesn't have enough balance

elements/lisk-transactions/test/15_pom_transaction.spec.ts
```diff
@@ -12,59 +12,459 @@
...
+			const expectedBalance =
+				sender.balance -
+				(transactionWithScenario1.asset.reward as bigint) +
+				transactionWithScenario1.fee;
+			expect(updatedSender.balance.toString()).toEqual(
+				expectedBalance.toString(),
+			);
+		});
+
+		it('should add reward to balance of the misbehaving delegate', async () => {
```
shuse2, line 418: it should also check when delegate didn't have sufficient balance at the apply time

elements/lisk-transactions/src/15_proof_of_misbehavior_transaction.ts
```diff
@@ -0,0 +1,445 @@
...
+		*/
+
+		const blockHeader1Bytes = Buffer.concat([
+			Buffer.from(networkIdentifier, 'hex'),
+			getBlockBytes(this.asset.header1),
+		]);
+		const blockHeader2Bytes = Buffer.concat([
+			Buffer.from(networkIdentifier, 'hex'),
+			getBlockBytes(this.asset.header2),
+		]);
```
ManuGowda, line 348: maybe move this to after first block signature check? so that if its invalid we dont need to calculate these bytes

elements/lisk-transactions/src/15_proof_of_misbehavior_transaction.ts
mitsuaki-u: I feel it is unnecessary.

elements/lisk-transactions/src/15_proof_of_misbehavior_transaction.ts
```diff
@@ -0,0 +1,445 @@
...
+		store.account.set(delegateAccount.address, delegateAccount);
+
+		return errors;
+	}
+
+	protected async undoAsset(
+		store: StateStore,
+	): Promise<ReadonlyArray<TransactionError>> {
+		const currentHeight = store.chain.lastBlockHeader.height + 1;
+		const delegateId = getAddressFromPublicKey(
```
ManuGowda, line 416: same

elements/lisk-transactions/src/utils/get_block_bytes.ts
```diff
@@ -0,0 +1,120 @@
+/*
+ * Copyright © 2019 Lisk Foundation
```
ManuGowda, line 2:
```suggestion
 * Copyright © 2020 Lisk Foundation
```

elements/lisk-transactions/src/15_proof_of_misbehavior_transaction.ts
```diff
@@ -0,0 +1,445 @@
...
+		numberOfTransactions: {
+			type: 'integer',
+			minimum: 0,
+		},
+		payloadLength: {
+			type: 'integer',
+			minimum: 0,
+		},
+		previousBlockId: {
+			type: ['null', 'string'],
```
shuse2, line 85: this can be `string` because `null` is only for the genesis block case

elements/lisk-transactions/src/utils/is_punished.ts
```diff
@@ -0,0 +1,39 @@
...
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+import { Account } from '../transaction_types';
+
+const PUNISH_TIME_VOTE = 260000;
+const PUNISH_TIME_SELF_VOTE = 780000;
```
ManuGowda, line 18:
```suggestion
const VOTER_PUNISH_TIME = 260000;
const SELF_VOTE_PUNISH_TIME = 780000;
```
?

elements/lisk-transactions/src/15_proof_of_misbehavior_transaction.ts
mitsuaki-u: I will leave the type as bigint since we never return as JSON. But I will add a casting to bigint in the constructor.

elements/lisk-transactions/src/15_proof_of_misbehavior_transaction.ts
mitsuaki-u: it should remain `this.id` because it is always tx.id there. I will update the property below to add generatorPublickey.

elements/lisk-transactions/src/15_proof_of_misbehavior_transaction.ts
mitsuaki-u: You need the blockBytes to check the signature.

elements/lisk-transactions/test/15_pom_transaction.spec.ts
```diff
@@ -12,59 +12,494 @@
...
+			const { errors, status } = await transactionWithScenario1.apply(store);
+			expect(status).toBe(Status.FAIL);
+			expect(errors).toHaveLength(3);
+			expect(errors[2].message).toInclude(
+				'Cannot apply proof-of-misbehavior. Delegate is already punished.',
+			);
+		});
 
-		it.todo('should return errors if |header1.height - h| >= 260000');
+		it('should add reward to balance of the sender', async () => {
```
shuse2, line 298: there should be also test to check correct signature and id against the protocol spec

elements/lisk-transactions/src/15_proof_of_misbehavior_transaction.ts
```diff
@@ -0,0 +1,459 @@
...
+export class ProofOfMisbehaviorTransaction extends BaseTransaction {
+	public readonly asset: ProofOfMisbehaviorAsset;
+	public static TYPE = 15;
+
+	public constructor(rawTransaction: unknown) {
+		super(rawTransaction);
+		const tx = (typeof rawTransaction === 'object' && rawTransaction !== null
+			? rawTransaction
+			: {}) as Partial<TransactionJSON>;
+		this.asset = (tx.asset || {}) as ProofOfMisbehaviorAsset;
```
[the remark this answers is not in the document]
mitsuaki-u: is it necessary to convert? there is no `bigint` calculations, and `getBlockBytes` uses `intToBuffer` which converts strings to integers.

elements/lisk-transactions/test/15_pom_transaction.spec.ts
mitsuaki-u: do you mean block signature? that is already checked. the tx signature is checked in base tx tests.

elements/lisk-transactions/src/15_proof_of_misbehavior_transaction.ts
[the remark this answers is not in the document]
shuse2: I think it's ok to expect to be string, and use it as string without conversion.
but in that case, instead of type `BlockHeader`, you should create `BlockHeaderJSON`  like
```
type Modify<T, R> = Omit<T, keyof R> & R;

export type BlockHeaderJSON = Modify<BlockHeader,
	{
	readonly totalAmount: string;
	readonly totalFee: string;
	readonly reward: string;
}
>;
```
and use this type.
Also, getBytes function is using `toString()` for the bigint property but string doesn't have this method

elements/lisk-transactions/src/utils/get_block_bytes.ts
```diff
@@ -0,0 +1,120 @@
...
+	);
+
+	const totalAmountBuffer = intToBuffer(
+		block.totalAmount.toString(),
+		SIZE_INT64,
+		LITTLE_ENDIAN,
+	);
+
+	const totalFeeBuffer = intToBuffer(
+		block.totalFee.toString(),
```
shuse2, line 74: here if and `totalAmount` and `reward` is calling `toString()` but if we create `BlockHeaderJSON`, we dont need toString() anymore

elements/lisk-transactions/test/15_pom_transaction.spec.ts
shuse2: I mean the transaction signature.
Because getAssetBytes changes the signature, complete signature should be checked against the protocol spec in each transaction too

elements/lisk-transactions/test/15_pom_transaction.spec.ts
```diff
@@ -12,59 +12,494 @@
...
+		let store: StateStoreMock;
+		let sender: Account;
+		let delegate: Account;
+
+		beforeEach(async () => {
+			sender = {
+				...defaultAccount,
+				...validProofOfMisbehaviorTransactionScenario1.testCases.input
+					.reportingAccount,
+				balance: BigInt(
```
shuse2, line 106: I think sender `nonce` is giving the error too.
it should use
`nonce: validProofOfMisbehaviorTransactionScenario1.testCases.output.nonce` for checking exact error you need

elements/lisk-transactions/src/utils/get_block_bytes.ts
ManuGowda: I agree but maybe calling `toString()` on a string doesn't hurt much? the implementation will the similar, easy to modify and copy between two places in the future?

PR #5100, opened 2020-04-03
BFT to exclude standby delegate - Closes #4943

### LINKED ISSUE #4943: Update BFT library not to include standby delegates for the calculation
### Description
- Update BFT library not to include the block from standby delegate for pre-votes and pre-commits following https://github.com/LiskHQ/lips/blob/master/proposals/lip-0022.md#bft-consensus-rules-for-standby-delegates
- DPoS should have already implemented isStandby function.
### Motivation
In order to extend the finality time, it should exclude the standby delegate for calculation
### Acceptance Criteria
- Test case where delegate is standby delegate, and the calculation is skipped

### Related issues
#4915

ishantiw:
### What was the problem?

This PR resolves #4943 

### How was it solved?

- [x] Update BFT library not to include the block from standby delegate for pre-votes and pre-commits following https://github.com/LiskHQ/lips/blob/master/proposals/lip-0022.md#bft-consensus-rules-for-standby-delegates
- [x] DPoS should have already implemented isStandby function.
- [x] Test case where delegate is standby delegate, and the calculation is skipped

### How was it tested?

Run `npm run test finality_manager.spec.ts` under `elements/lisk-bft`


shuse2:
@ishantiw implementation looks good from myside, but it's not building 😢

elements/lisk-bft/src/types.ts
```diff
@@ -31,6 +31,11 @@ export interface DPoS {
 		stateStore: StateStore,
 		delegateActiveRoundLimit?: number,
 	) => Promise<number>;
+	readonly isStandbyDelegate: (
```
ManuGowda, line 34:
```suggestion
	readonly isStandByDelegate: (
```

PR #5118, opened 2020-04-07
Fix error message on delegate not found - Closes #5114

### LINKED ISSUE #5107: Incorrect error message on the unlock transaction
### Expected behavior
Unlock transaction should return correct error message.
When unlocking before the waiting time, it should return message saying that.

### Actual behavior
Error message was `Unlocking is not permitted as delegate is currently being punished`, for delegate which is not punished (empty pomHeights)

### Steps to reproduce
Send unlocking transaction before wait time

4.0-

### LINKED ISSUE #5114: Transaction is rejected during broadcast with no info
### Expected behavior

`lisk transaction:broadcast` should be able to broadcast, in case of error it should return more meaning error from the node

### Actual behavior

The below transaction is rejected but the error message doesn't give any idea about the error
```bash
./bin/run transaction:broadcast '{
        "id": "11290968189718369965",
        "type": 13,
        "senderPublicKey": "b2abc963b76fab65163b35372701049006c97cb589a1163c4d7c42c36eb581fb",
        "senderId": "7715036167861911539L",
        "nonce": "1",
        "fee": "10000000000",
        "signatures": [
                "d3844e5a900fc43464a8a41c7104a907bbd77ae1e9a46091007b5c22fbbbd723ffae09ce351eb5bde722696e6fd4004d79afb04488d738ed670434fd0ba7510d"
        ],
        "asset": {
                "votes": [
                        {
                                "delegateAddress": "16796830995384197764L",
                                "amount": "1000000000"
                        }
                ]
        }
}'

›   Error: Transaction was rejected with errors
```

on the API the response is,

```json
{
    "message": "Transaction was rejected with errors",
    "code": 409,
    "errors": {}
}
```

### Steps to reproduce

1. Create a transaction
2. Broadcast it using POST `http://localhost:4000/api/transactions` or using commander `lisk transaction:broadcast`

`4.0.0`

shuse2:
### What was the problem?

This PR resolves #5114 
This PR resolves #5107

### How was it solved?

Get the delegate by `getOrDefault` and if delegate is not registered, return error.

### How was it tested?

Send transaction described in #5114


elements/lisk-transactions/src/14_unlock_transaction.ts
```diff
@@ -218,13 +218,23 @@ export class UnlockTransaction extends BaseTransaction {
 
 		for (const unlock of this.asset.unlockingObjects) {
 			const sender = await store.account.get(this.senderId);
-			const delegate = await store.account.get(unlock.delegateAddress);
+			const delegate = await store.account.getOrDefault(unlock.delegateAddress);
+			if (!delegate.username) {
+				errors.push(
+					new TransactionError(
+						'Voted delegate is not registered',
```
ManuGowda, line 225:
```suggestion
						'Voted account is not registered as delegate',
```

PR #5119, opened 2020-04-07
Update FCS to only trigger for the same round - Closes #5115

### LINKED ISSUE #5115: Fast chain switching does not work for the future rounds
### Expected behavior
Fast chain switching should be completed without error.

### Actual behavior
~~When fast chain switching is triggered for the future rounds, it will fail as the system does not maintain forger list for the future~~ Fixed in #5119 

However, with the temporal solution #5119, if fork or out of sync happens at the end of the round, it takes longer to sync up with the other node, because fast chain switching and block synchronization mechanism does not get triggered.

### Steps to reproduce
Run a node, and prepare another node and make it out of sync for 130 blocks, then start again

4.0-

shuse2:
### What was the problem?

This PR resolves #5115 

### How was it solved?

- Update FCS to only trigger for the same round because it doesn't not know about the future rounds anymore

### How was it tested?

- Make one node behind for more than round, and check block sync is triggered
- Make one node behind for less than round, and check FCS is triggered

### DIFF
--- a/framework/test/jest/unit/specs/application/node/synchronizer/fast_chain_switching_mechanism/fast_chain_switching_mechanism.spec.js
+++ b/framework/test/jest/unit/specs/application/node/synchronizer/fast_chain_switching_mechanism/fast_chain_switching_mechanism.spec.js
@@ -139,6 +139,48 @@ describe('fast_chain_switching_mechanism', () => {
 		});
 	});
 
+	describe('isValidFor', () => {
+		beforeEach(async () => {
+			jest.spyOn(dposModule, 'getForgerAddressesForRound');
+		});
+
+		describe('when reveivedBlock is at the same round as the last block', () => {
+			it('should return true', async () => {
+				chainModule._lastBlock = { height: 340 };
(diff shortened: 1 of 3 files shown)


framework/test/jest/unit/specs/application/node/synchronizer/fast_chain_switching_mechanism/fast_chain_switching_mechanism.spec.js
```diff
@@ -139,6 +139,48 @@ describe('fast_chain_switching_mechanism', () => {
...
+		beforeEach(async () => {
+			jest.spyOn(dposModule, 'getForgerAddressesForRound');
+		});
+
+		describe('when reveivedBlock is at the same round as the last block', () => {
+			it('should return true', async () => {
+				chainModule._lastBlock = { height: 340 };
+				dposModule.getForgerAddressesForRound.mockResolvedValue([
+					'11121761073292744822L',
+				]);
```
ManuGowda, line 152: Maybe move this to beforeEach hook?

PR #5126, opened 2020-04-07
Wrong selection for top 101 delegates as active delegates - Closes #5123

### LINKED ISSUE #5123: Wrong selection for top 101 delegates as active delegates
### Expected behavior

There should always be top 101 delegates selected as active delegates, and remaining if available and eligible should be selected as standby delegates. 

### Actual behavior

Currently, in case there are 103 total delegates in the system we select those as active delegates and no standby delegates. 

### Steps to reproduce

1. Start the node with a fresh database 
2. Run the query `select value from consensus_state where key = 'dpos:forgersList';`
3. There should be 103 items in the delegates array and 2 in the standby list. 

4.0

nazarhussain:
### What was the problem?
In one edge case, we were selecting 103 as active delegates. 

This PR resolves #5123

### How was it solved?

Fixed the max slots for active delegates to 101. 

### How was it tested?

```
cd elements/lisk-dpos
yarn test
```


elements/lisk-dpos/src/delegates_list.ts
```diff
@@ -213,8 +213,11 @@ const _pickStandByDelegate = (
 ): number => {
 	const seedNumber = randomSeed.readBigUInt64BE();
 	const totalVoteWeight = _getTotalVoteWeight(delegateWeights);
+
+	// To pass voteWeight = 0, initialize threshold with negative value
 	// tslint:disable-next-line no-let
-	let threshold = seedNumber % totalVoteWeight;
+	let threshold =
+		totalVoteWeight === BigInt(0) ? BigInt(-1) : seedNumber % totalVoteWeight;
```
shuse2, line 220: This is not true because totalVoteWeight can be non zero, and we still pick top 103.
This function is specifically to select standby delegates with voteWeight, so if we pick top 103, it shouldn't come to this function

elements/lisk-dpos/src/delegates_list.ts
```diff
@@ -378,12 +381,8 @@ export class DelegatesList {
...
 		}
 		// Expect that voteWeight is stored in order of voteWeight and address
-		const hasStandbySlot =
-			voteWeight.delegates.length >
-			this.activeDelegates + this.standbyDelegates;
-		const activeDelegateSlots = hasStandbySlot
-			? this.activeDelegates
-			: this.activeDelegates + this.standbyDelegates;
+		const hasStandbySlot = voteWeight.delegates.length > this.activeDelegates;
+		const activeDelegateSlots = this.activeDelegates;
```
shuse2, line 385: this variable is not needed

elements/lisk-dpos/src/delegates_list.ts
nazarhussain: Yes `totalVoteWeight` can be non-zero. This logic is not enforcing it to be zero. 

We pick 103 delegates for forging, but not `top` 103. This `top` adjective is only for `101` delegates which we call active delegates. And rest of the 2 which we call standby delegates will be picked randomly based on a random seed.

elements/lisk-dpos/src/delegates_list.ts
```diff
@@ -213,6 +213,8 @@ const _pickStandByDelegate = (
 ): number => {
 	const seedNumber = randomSeed.readBigUInt64BE();
 	const totalVoteWeight = _getTotalVoteWeight(delegateWeights);
+
+	// Assuming that totalVoteWeight can't be zero
```
shuse2, line 217: this comment can be removed

elements/lisk-dpos/src/delegates_list.ts
```diff
@@ -378,22 +380,26 @@ export class DelegatesList {
...
 
-		// Only choose standby delegate if it exists
-		if (standbyDelegateVoteWeights.length !== 0) {
+		// If standby delegates are less or equal to what required
+		// Then don't choose based on random seed and consider those as standby
+		if (
+			standbyDelegateVoteWeights.length > 0 &&
+			standbyDelegateVoteWeights.length <= this.standbyDelegates
+		) {
+			standbyDelegateVoteWeights.map(delegate => {
```
shuse2, line 398: this is not mapping anything, so it should be `for..of`

shuse2, line 398: this is not mapping anything, so it should be for..of

PR #5130, opened 2020-04-08
Update FCS to check active delegate or not with 2 rounds setting

### LINKED ISSUE #5115: Fast chain switching does not work for the future rounds

shuse2:
### What was the problem?

This PR resolves #5115 

### How was it solved?

- Revert the changes of checking 2 rounds for the FCS
- Update FCS trigger to check if the block received is from active delegate
- Remove the block slot and state check for the FCS mechanism while querying the next blocks
- Add isActiveDelegate function to DPoS using voteWeight
- Remove validateDetached, and consolidate with `validate` in the processor
- Cleanup unused function and combine them in verify step in the chain

### How was it tested?

- Start 2 nodes, and 1 node should have longer chain, then observe
  1. First block synchronization mechanism is triggered
  2. Fast chain switching is triggered
  3. stop the node for 1.5 blocks
  4. start again and check if the fast chain switching is triggered


framework/src/application/node/block_processor_v2.js
```diff
@@ -185,6 +185,16 @@ class BlockProcessorV2 extends BaseBlockProcessor {
 		this.validate.pipe([
 			data => this._validateVersion(data),
 			data => validateSchema(data),
+			({ block }) =>
+				this.chainModule.validateBlockHeader(block, getBytes(block)),
+			({ block }) => this.bftModule.validateBlock(block),
```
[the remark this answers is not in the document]
shuse2: yea i was thinking that too, but there are several other function name that are not consistent.
Ill create an issue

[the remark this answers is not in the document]
shuse2: Created: #5133

elements/lisk-dpos/src/dpos.ts
```diff
@@ -152,6 +156,37 @@ export class Dpos {
...
+			);
+		}
+		const activeDelegateVoteWeights = voteWeight.delegates.slice(
+			0,
+			this._activeDelegates,
+		);
+
+		return (
+			activeDelegateVoteWeights.findIndex(vw => vw.address === address) > -1
+		);
```
nazarhussain, line 187: Why the logic of `isActiveDelegate` and `isStandbyDelegate` is different? It could be really simple, something which is in the forger list and not standby is an active delegate.

framework/src/application/node/block_processor_v0.js
```diff
@@ -1,258 +0,0 @@
...
-			blockSignature: signDataWithPrivateKey(
-				hash(getBytes(block)),
-				keypair.privateKey,
-			),
-		};
-	}
-}
-
-module.exports = {
-	BlockProcessorV0,
```
nazarhussain, line 257: Why we removed this file? Is it related to this issue?

framework/src/application/node/block_processor_v1.js
```diff
@@ -1,29 +0,0 @@
...
-// Block version 0 and 1 is essentially the same
-class BlockProcessorV1 extends BlockProcessorV0 {
-	// eslint-disable-next-line class-methods-use-this
-	get version() {
-		return 1;
-	}
-}
-
-module.exports = {
-	BlockProcessorV1,
```
nazarhussain, line 28: Same here.

elements/lisk-dpos/src/dpos.ts
shuse2: forgerlist can be only generated right at the round, but what we need is active delegate list in the future. Therefore, it's using the voteWeights

framework/src/application/node/block_processor_v0.js
shuse2: It was removed because it was not used in anywhere, and by changing the function name, it was going to be outdated. Also, with the new regenesis proposal, this is no longer required

elements/lisk-dpos/src/dpos.ts
nazarhussain: We can't look into the unseen future. Can we rename this function to say `isActiveDelegateInTwoFutureRounds` or something similar. Otherwise, the context of `isActiveDelegate` and `isStandbyDelgate` will always be considered the current round.

shuse2: `isActiveDelegate` and `isStandbyDelgate` are both checking as much as they can.
isActiveDelegate will check with voteWeights because it has most information, and it can check from the finalized - current round + 2, for `isStandbyDelegate` it's finalized - current round

so renaming will not describe what it is fully. Maybe ill add comments on function?

PR #5132, opened 2020-04-08
Update create vote command to use LSK - Closes #5104

### LINKED ISSUE #5104: Votes doesn't take LSK as amount
### Expected behavior
Commander votes command should take amount as LSK, not beddows

### Actual behavior
Votes command is taking amount is beddows, not LSK

### Steps to reproduce
```
/bin/run transaction:create:vote 1 25 --votes="2128442994687666000L,1000" --networkIdentifier="93d00fe5be70d90e7ae247936a2e7d83b50809c79b73fa14285f02c842348b3e" --passphrase="salad drift calm brass thought coyote impulse frown wealth purpose base elevator”
```

mitsuaki-u:
### What was the problem?

User entered amount was in beddows for create vote command.

This PR resolves #5104 

### How was it solved?

- Update commander to use LSK and convert accordingly
- Add unit test

### How was it tested?

Run commander. 
Create vote using LSK as input.

### DIFF
--- a/commander/src/commands/transaction/create/vote.ts
+++ b/commander/src/commands/transaction/create/vote.ts
@@ -157,7 +157,7 @@ export default class VoteCommand extends BaseCommand {
 
 			return {
 				delegateAddress,
-				amount: String(BigInt(normalizedAmount) * BigInt(sign)),
+				amount: (BigInt(normalizedAmount) * BigInt(sign)).toString(),
 			};
 		});
 
--- a/commander/src/commands/transaction/create/vote.ts
+++ b/commander/src/commands/transaction/create/vote.ts
@@ -135,9 +135,29 @@ export default class VoteCommand extends BaseCommand {
 				throw new Error('Delegate address must be unique.');
 			}
 
+			const numberAmount = Number(amount);
+
+			if (Number.isNaN(numberAmount)) {
+				throw new ValidationError(
+					'Enter a valid amount in number string format.',
+				);
+			}
+
+			const sign = numberAmount < 0 ? -1 : 1;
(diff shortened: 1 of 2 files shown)


commander/src/commands/transaction/create/vote.ts
```diff
@@ -135,9 +135,29 @@ export default class VoteCommand extends BaseCommand {
...
+			if (!isValidFee(normalizedAmount)) {
+				throw new ValidationError(
+					'Enter a valid vote amount in number string format.',
+				);
+			}
+
 			return {
 				delegateAddress,
-				amount,
+				amount: String(Number(normalizedAmount) * sign),
```
shuse2, line 160: normalize amount is `int64`, so it should use BigInt to multiply with sign

commander/src/commands/transaction/create/vote.ts
```diff
@@ -157,7 +157,7 @@ export default class VoteCommand extends BaseCommand {
 
 			return {
 				delegateAddress,
-				amount: String(Number(normalizedAmount) * sign),
+				amount: String(BigInt(normalizedAmount) * BigInt(sign)),
```
shuse2, line 160:
```suggestion
				amount: (BigInt(normalizedAmount) * BigInt(sign)).toString(),
```

mitsuaki-u: I think both work

shuse2: yes, `String` constructor also works, but we don't use it anywhere else, so let's keep it consistent

PR #5142, opened 2020-04-09
Fix inconsistent account api response - Closes #5141

### LINKED ISSUE #5141: Inconsistent response for delegate information on `/api/accounts` endpoint
### Expected behavior

There should be a consistent response for the information related to delegates for the API endpoint `/api/accounts`

### Actual behavior

For delegate accounts it returns:
```
            "delegate": {
                "approval": 0,
                "fees": "0",
                "missedBlocks": 5,
                "producedBlocks": 64,
                "productivity": 92.75,
                "rewards": "0"
            },
```
for non-delegate accounts it return:
```
            "delegate": {
                "consecutiveMissedBlocks": 0,
                "isBanned": false,
                "lastForgedHeight": 0,
                "pomHeights": []
            },
```

### Steps to reproduce

```
curl -XGET http://localhost:4000/api/accounts
```

4.0.0

shuse2:
### What was the problem?

This PR resolves #5141 

### How was it solved?
- Fix not to have partial formatting
- Keep approval calculation under delegate property

### How was it tested?

- Call `/api/accounts` endpoint
- Add unit test

### DIFF
--- a/framework/src/modules/http_api/controllers/accounts.js
+++ b/framework/src/modules/http_api/controllers/accounts.js
@@ -27,17 +27,13 @@ function AccountsController(scope) {
 }
 
 function accountFormatter(totalSupply, account) {
-	const formattedAccount = {
-		...account,
-		delegate: {
-			...account.delegate,
-			approval: calculateApproval(account.totalVotesReceived, totalSupply),
-		},
-	};
+	account.delegate.approval = calculateApproval(
+		account.totalVotesReceived,
+		totalSupply,
+	);
+	account.publicKey = account.publicKey || '';
 
-	formattedAccount.publicKey = formattedAccount.publicKey || '';
-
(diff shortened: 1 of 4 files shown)


framework/src/modules/http_api/controllers/accounts.js
```diff
@@ -27,42 +27,13 @@ function AccountsController(scope) {
...
-		);
-	}
-
-	// If an account is not multi signature, then do not include keys for response
-	if (formattedAccount.keys.numberOfSignatures === 0) {
-		delete formattedAccount.keys;
-	}
+	const formattedAccount = {
+		...account,
+		delegate: {
```
pablitovicente, line 32: Why not just add the approval to the original `delegate` property as the object is no longer used so mutating it does not affect anything else?

PR #5146, opened 2020-04-15
Add typescript setup to framework - Closes #5145

### LINKED ISSUE #5145: Add typescript setup for framework
### Description
Add typescript setup, tsconfig and eslint rules for framework.

### Motivation

We need to setup the typescript environment for the coming issues that we convert the framework to TypeScript

### Acceptance Criteria

- Add tsconfig
- Add eslint rules for typescript
- Allow javascript and not convert the source code to typescript

shuse2:
### What was the problem?

This PR resolves #5145 

### How was it solved?

- Add typescript setup
- Update eslint config to support typescript
- Fix critical eslint errors

### How was it tested?

- Application builds
- No eslint error


framework/src/application/application.js
```diff
@@ -477,10 +477,12 @@ class Application {
 						this._node.actions.updateForgingStatus(action),
 				},
 				getForgingStatusOfAllDelegates: {
+					// eslint-disable-next-line @typescript-eslint/require-await
```
ManuGowda, line 480: we dont need them,  `this._node.actions.getForgingStatusOfAllDelegates` is sync function, so you can remove `async`  from handler `handler: () =>`

framework/src/application/application.js
```diff
@@ -477,10 +477,12 @@ class Application {
...
 					handler: async () =>
 						this._node.actions.getForgingStatusOfAllDelegates(),
 				},
 				getTransactionsFromPool: {
+					// eslint-disable-next-line @typescript-eslint/require-await
```
ManuGowda, line 485: same here

framework/src/application/application.js
```diff
@@ -492,12 +494,15 @@ class Application {
 					handler: async action => this._node.actions.postTransaction(action),
 				},
 				getSlotNumber: {
+					// eslint-disable-next-line @typescript-eslint/require-await
```
ManuGowda, line 497: same here and other places where we have this lint issue in this file

framework/.eslintrc.js
```diff
@@ -0,0 +1,54 @@
+module.exports = {
```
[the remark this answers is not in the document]
shuse2: it's `js` and we need to take the reference to `tsconfig` with relative path, so symlink doesnt work  well here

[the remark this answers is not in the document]
shuse2: as mentioned above, it needs relative path to the tsconfig for `tsconfigRootDir: __dirname,`

Issue #4696, opened 2020-01-14
Remove tslint and use typescript-eslint


shuse2:
### Description
`tslint` is deprecated in favor of eslint.
Therefore, it should migrate to https://github.com/typescript-eslint/typescript-eslint 

- [x] @liskhq/lisk-constants #5149 
- [x] @liskhq/lisk-api-client #5150 
- [x] @liskhq/lisk-cryptograpy #5153
- [x] @liskhq/lisk-passphrase #5155
- [x] @liskhq/lisk-transaction #5161 
- [x] @liskhq/lisk-transaction-pool #5164
- [x] @liskhq/lisk-validation #5165
- [x] @liskhq/lisk-chain	#5166
- [x] @liskhq/lisk-dpos #5169
- [x] @liskhq/lisk-bft #5178
- [x] @liskhq/lisk-p2p #5179
- [x] lisk-commander #5180
- [ ] lisk-elements #5181 
- [ ] @liskhq/lisk-client #5181 

### Actual behavior
It is using tslint

PR #5153, opened 2020-04-16
Update to remove tslint from cryptography - Partially Closes #4696

shuse2:
### What was the problem?

### How was it solved?

- Remove tslint from lisk-cryptography

### How was it tested?
- All the linting should work
- no source code changes

### DIFF
--- a/.eslintrc.test.js
+++ b/.eslintrc.test.js
@@ -4,7 +4,9 @@ module.exports = {
 		'arrow-body-style': 'off',
 		'@typescript-eslint/no-magic-numbers': 'off',
 		'@typescript-eslint/no-require-imports': 'off',
-		'@typescript-eslint/no-unsafe-member-access': 'off',
 		'@typescript-eslint/no-explicit-any': 'off',
+		'@typescript-eslint/no-unsafe-assignment': 'off',
+		'@typescript-eslint/no-unsafe-member-access': 'off',
+		'@typescript-eslint/no-unsafe-call': 'off',
 	},
 };
--- a/elements/lisk-cryptography/src/buffer.ts
+++ b/elements/lisk-cryptography/src/buffer.ts
@@ -20,9 +20,9 @@ const MAX_NUMBER_BYTE_LENGTH = 6;
 export const intToBuffer = (
 	value: number | string,
 	byteLength: number,
-	endianness: string = BIG_ENDIAN,
-	signed: boolean = false,
-) => {
+	endianness = BIG_ENDIAN,
+	signed = false,
+): Buffer => {
 	if (![BIG_ENDIAN, LITTLE_ENDIAN].includes(endianness)) {
 		throw new Error(
 			`Endianness must be either ${BIG_ENDIAN} or ${LITTLE_ENDIAN}`,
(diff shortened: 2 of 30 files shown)


.eslintrc.test.js
```diff
@@ -4,7 +4,9 @@ module.exports = {
 		'arrow-body-style': 'off',
 		'@typescript-eslint/no-magic-numbers': 'off',
 		'@typescript-eslint/no-require-imports': 'off',
-		'@typescript-eslint/no-unsafe-member-access': 'off',
 		'@typescript-eslint/no-explicit-any': 'off',
+		'@typescript-eslint/no-unsafe-assignment': 'off',
+		'@typescript-eslint/no-unsafe-member-access': 'off',
```
nazarhussain, line 9: This rule was already added to same file in the previous PR.

elements/lisk-cryptography/src/buffer.ts
```diff
@@ -37,20 +37,23 @@ export const intToBuffer = (
...
 		} else {
+			// eslint-disable-next-line no-lonely-if
 			if (signed) {
 				buffer.writeBigInt64BE(BigInt(value));
 			} else {
 				buffer.writeBigUInt64BE(BigInt(value));
 			}
 		}
 	} else {
+		// eslint-disable-next-line no-lonely-if
```
nazarhussain, line 48: At what point we will update the code based on new eslint rules. I suggest for such small things we should fix them right away.

elements/lisk-cryptography/src/hash.ts
```diff
@@ -41,12 +41,13 @@ export const hash = (data: Buffer | string, format?: string): Buffer => {
 	}
 
 	throw new Error(
-		`Unsupported data:${data} and format:${format}. Currently only Buffers or hex and utf8 strings are supported.`,
+		`Unsupported data:${data} and format:${format ??
```
nazarhussain, line 44: For my knowledge what's the difference between: 

```
format ?? 'undefined' 
```

and 

```
format || 'undefined' 
```

shuse2: `??` will only check null or undefined, but || is falsy check

elements/lisk-cryptography/src/buffer.ts
shuse2: I was thinking to fix this, but it just made it hard to read, so i wasn't sure how to make it clearer

framework/src/controller/bus.js
```diff
@@ -228,7 +228,7 @@ class Bus extends EventEmitter2 {
 	}
 
 	// eslint-disable-next-line @typescript-eslint/require-await
-	async leanup() {
+	async cleanup() {
```
mitsuaki-u, line 231: 👍

.eslintrc.test.js
shuse2: target branch is up to date with the latest development branch?

elements/lisk-cryptography/src/buffer.ts
nazarhussain: We should then create a separate issue saying remove these comments and fix the code.

shuse2: this function will be removed with serialization =)

PR #5156, opened 2020-04-16
Update to allow forge block when seedreveal is used up - Closes #5154

### LINKED ISSUE #5154: Forging should not stop even when all hashOnion is used
### Description

When all hashonion is used, currently it throws an error and does not forge a block, but it should forge a block with random hash onion

### Motivation

It is better for the network to forge new blocks and receive 0 rewards, than missing the slot.

### Acceptance Criteria

- It should warn delegate to update the seed reveal when all the hashonion is used
- It should use random seed reveal when all the hash onion is used

shuse2:
### What was the problem?

This PR resolves #5154 

### How was it solved?

- When all of the seed reveal is used, it will forge a block with random seed reveal
- When all of the seed reveal is used, it will show warning

### How was it tested?

Added unit test


shuse2:
@ManuGowda created separate issue #5162

PR #5160, opened 2020-04-17
Fix build issue for the sql files - Closes #5158

### LINKED ISSUE #5158: Issues after adding typescript configuration to framework
There are following issues found after adding typescript support to framework. 

- SQL files not being move to the build folder. Since `tsc` ignores the non-ts,js files, it is not moved
- SyntaxError: Unexpected token ': (modules/http_api/init_steps/bootstrap_swagger.js). Swagger tries to load all files in the controllers directory. Before there were only `.js` files, now we have source map files as well. We just need to replace that logic by requiring each controller explicitly. 

### Steps to reproduce
npm run build and check the dist-node

5.0-

shuse2:
### What was the problem?

This PR resolves #5158 

### How was it solved?

- Separately copy sql and yml files
- Update test app to run from dist-node
- Fix the dynamic require

### How was it tested?
After building, run test app, and it should run as expected


nazarhussain:
@shuse2 Why the coverage is dropped -9.84% on this PR?

shuse2:
@nazarhussain seems like somehow `dist-node` is included in the coverage, and im trying to figure out how...

PR #5165, opened 2020-04-20
Update to remove tslint from validator - Partially Closes #4696 

shuse2:
### What was the problem?

### How was it solved?

- Remove tslint from lisk-validator

### How was it tested?
- All the linting should work
- no source code changes

### DIFF
--- a/elements/lisk-validator/src/lisk_validator.ts
+++ b/elements/lisk-validator/src/lisk_validator.ts
@@ -31,12 +31,14 @@ class LiskValidator {
 		for (const formatName of Object.keys(formats)) {
 			this.validator.addFormat(
 				formatName,
+				// eslint-disable-next-line import/namespace
 				formats[formatName as keyof typeof formats],
 			);
 		}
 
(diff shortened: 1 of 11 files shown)


elements/lisk-validator/src/lisk_validator.ts
```diff
@@ -31,12 +31,14 @@ class LiskValidator {
 		for (const formatName of Object.keys(formats)) {
 			this.validator.addFormat(
 				formatName,
+				// eslint-disable-next-line import/namespace
```
pablitovicente, line 34: isn't this rules for imports only?

shuse2: I think this is showing error because of `as keyof typeof` for typescript

pablitovicente: If possible it would be nice to disable that one then. Approved anyway.

shuse2: I think this is the only case so far, so i would like to keep it for now.
If it continues to report wrong value, we can disable them

PR #5168, opened 2020-04-20
Improve error message on fee replacement - Closes #5151

### LINKED ISSUE #5151: Invalid error from transaction pool in case of invalid account nonce
### Expected behavior

In case the transaction does not contain valid account nonce it should show the correct message that nonce is invalid. 

### Actual behavior

Currently, if the transaction nonce is invalid it shows:

```
Transaction was not added to the pool as it exceeds maximum transaction limit per account
```

This case happens only when the transaction contains valid nonce compared to persistent account state, but invalid against account nonce state in the transaction pool. 

Problem is the `TransactionList.add` method return status based on the nonce. 

https://github.com/LiskHQ/lisk-sdk/blob/5b390492a7328c59e19b771a568b66ba12c3ecd0/elements/lisk-transaction-pool/src/transaction_list.ts#L75-L76

And `TransactionPool` considers it a different error. 

https://github.com/LiskHQ/lisk-sdk/blob/5b390492a7328c59e19b771a568b66ba12c3ecd0/elements/lisk-transaction-pool/src/transaction_pool.ts#L227-L236

### Steps to reproduce

- Try to create two different transactions from the same account with the same nonce
- Broadcast those two to node 

4.0.0

shuse2:
### What was the problem?

This PR resolves #5151 

### How was it solved?

- TransactionList.add return the reason of not adding to the list

### How was it tested?

As described in #5151

### DIFF
--- a/elements/lisk-transaction-pool/src/transaction_list.ts
+++ b/elements/lisk-transaction-pool/src/transaction_list.ts
@@ -25,6 +25,10 @@ const DEFAULT_MAX_SIZE = 64;
 // tslint:disable-next-line no-magic-numbers
 export const DEFAULT_MINIMUM_REPLACEMENT_FEE_DIFFERENCE = BigInt(10);
 
+type AddResult =
+	| { added: true; removedID?: string; reason?: undefined }
+	| { added: false; removedID?: undefined; reason: string };
+
 export class TransactionList {
 	public readonly address: string;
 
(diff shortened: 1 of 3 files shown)


elements/lisk-transaction-pool/src/transaction_list.ts
```diff
@@ -25,6 +25,10 @@ const DEFAULT_MAX_SIZE = 64;
 // tslint:disable-next-line no-magic-numbers
 export const DEFAULT_MINIMUM_REPLACEMENT_FEE_DIFFERENCE = BigInt(10);
 
+type AddResult =
```
pablitovicente, line 28: This name doesn't allow to quickly figure out what it stores; is it the result of adding two values or the result of an `Add` operation? A better name would be great.

nazarhussain, line 28: May be inline return type for the function would be fine.

PR #5178, opened 2020-04-21
Update to remove tslint from bft - Partially Closes #4696

shuse2:
### What was the problem?

### How was it solved?

- Remove tslint from lisk-bft

### How was it tested?
- All the linting should work
- no source code changes

### DIFF
--- a/elements/lisk-bft/package.json
+++ b/elements/lisk-bft/package.json
@@ -24,8 +24,8 @@
 	"scripts": {
 		"clean": "./scripts/clean.sh",
 		"format": "prettier --write '**/*'",
-		"lint": "tslint --format verbose --project .",
-		"lint:fix": "npm run lint -- --fix",
+		"lint": "eslint . --ext .js,.ts",
+		"lint:fix": "eslint --fix --ext .js,.ts .",
 		"test": "jest",
 		"test:watch": "npm test -- --watch",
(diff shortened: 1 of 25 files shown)


elements/lisk-bft/package.json
```diff
@@ -24,8 +24,8 @@
 	"scripts": {
 		"clean": "./scripts/clean.sh",
 		"format": "prettier --write '**/*'",
-		"lint": "tslint --format verbose --project .",
-		"lint:fix": "npm run lint -- --fix",
+		"lint": "eslint . --ext .js,.ts",
+		"lint:fix": "eslint --fix --ext .js,.ts .",
```
nazarhussain, line 28: It's missing a `.` in command, but on other packages, we used different commands. 

```suggestion
		"lint:fix": "npm run lint -- --fix",
```

shuse2: it suppose to have the `dot` and it looks correct?

nazarhussain: Missed that dot at the end. Can we do it the way for lint command.

nazarhussain, line 28:
```suggestion
		"lint:fix": "eslint . --fix --ext .js,.ts",
```

shuse2: ill change all together in the last PR

PR #5179, opened 2020-04-21
Update to remove tslint from p2p - Partially Closes #4696

shuse2:
### What was the problem?

### How was it solved?

- Remove tslint from lisk-p2p

### How was it tested?
- All the linting should work
- no source code changes

### DIFF
--- a/elements/lisk-p2p/src/peer_book/base_list.ts
+++ b/elements/lisk-p2p/src/peer_book/base_list.ts
@@ -12,8 +12,11 @@
  * Removal or modification of this copyright notice is prohibited.
  *
  */
+// eslint-disable-next-line import/no-cycle
 import { ExistingPeerError } from '../errors';
+// eslint-disable-next-line import/no-cycle
 import { P2PEnhancedPeerInfo, P2PPeerInfo } from '../types';
+// eslint-disable-next-line import/no-cycle
 import {
 	evictPeerRandomlyFromBucket,
 	getBucketId,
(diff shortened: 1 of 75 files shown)


elements/lisk-p2p/src/peer_book/base_list.ts
```diff
@@ -12,8 +12,11 @@
  * Removal or modification of this copyright notice is prohibited.
  *
  */
+// eslint-disable-next-line import/no-cycle
 import { ExistingPeerError } from '../errors';
+// eslint-disable-next-line import/no-cycle
 import { P2PEnhancedPeerInfo, P2PPeerInfo } from '../types';
+// eslint-disable-next-line import/no-cycle
```
ManuGowda, line 19: If there are group of similar rules can be clubbed to make. it looks cleaner. if you accept this suggestion, please use the same in other places.
```suggestion
/* eslint-disable import/no-cycle */
import { ExistingPeerError } from '../errors';
import { P2PEnhancedPeerInfo, P2PPeerInfo } from '../types';
import {
	evictPeerRandomlyFromBucket,
	getBucketId,
	PEER_TYPE,
	sanitizeEnhancedPeerInfo,
} from '../utils';
/* eslint-enable */
```

shuse2: For the no-cycle, we want to fix this issue one by one, so I think it make sense not to group them

PR #5198, opened 2020-04-24
Update framework utils to be typescript

### LINKED ISSUE #5194: Update application/utils to be typescript
### Description

Update framework application/utils to be typescript

### Motivation

### Acceptance Criteria

- All files under framework/src/application/utils should be in typescript

shuse2:
### What was the problem?

This PR resolves #5194 

### How was it solved?

- Update framework utils to be typescript
- Update related test to be typescript
- Update JobQueue to be the same implementation as the on in the transaction pool, and it's duplicated since it's not exposed part of the transaction pool

### How was it tested?
- All the test should pass
- Runs the test_app as expected


framework/src/application/node/node.js
```diff
@@ -69,11 +69,7 @@ module.exports = class Node {
 				this.options.forging.waitThreshold >= this.options.constants.blockTime
 			) {
 				throw Error(
-					`forging.waitThreshold=${
-						this.options.forging.waitThreshold
-					} is greater or equal to genesisConfig.blockTime=${
-						this.options.constants.blockTime
-					}. It impacts the forging and propagation of blocks. Please use a smaller value for forging.waitThreshold`,
+					`forging.waitThreshold=${this.options.forging.waitThreshold} is greater or equal to genesisConfig.blockTime=${this.options.constants.blockTime}. It impacts the forging and propagation of blocks. Please use a smaller value for forging.waitThreshold`,
```
ManuGowda, line 72: maybe formatting issue?

shuse2: yes automatic formatting 😢

PR #5220, opened 2020-04-29
Update processor to typescript - Closes #5191

### LINKED ISSUE #5191: Update application/node/processor to be typescript
### Description

Update framework application/processor to be typescript

### Motivation

### Acceptance Criteria

- All files under framework/src/application/processor should be in typescript

shuse2:
### What was the problem?

This PR resolves #5191 

### How was it solved?

Update processor and block processor to typescript

### How was it tested?

- Check if the application runs properly with built version
- Test and build should pass


.eslintrc.js
```diff
@@ -31,7 +31,11 @@ module.exports = {
 		'@typescript-eslint/no-dynamic-delete': 'off',
 		'@typescript-eslint/no-implied-eval': 'off',
 		'@typescript-eslint/strict-boolean-expressions': 'off',
+		'@typescript-eslint/no-unsafe-member-access': 'off',
+		'@typescript-eslint/no-unsafe-assignment': 'off',
+		'@typescript-eslint/no-unsafe-return': 'off',
```
nazarhussain, line 36: I suggest we should not return `any` from a function we define.

shuse2: yea I agree, but this check was checking more than that. 
Even with setup, it's still disallow `any`

nazarhussain: Can't understand even with setup. Can you share any example which was causing problem more than function return.

shuse2: seems like i was looking at the wrong error at that time, i cannot reproduce. Reverted the change =)

PR #5231, opened 2020-04-29
Migrate node/transport to TypeScript - Closes #5193

### LINKED ISSUE #5193: Update application/node/transport to be typescript
### Description

Update framework application/node/transport to be typescript

### Motivation

### Acceptance Criteria

- All files under framework/src/application/node/transport should be in typescript

ishantiw:
### What was the problem?

This PR resolves #5193 

I have not removed `transport/transport.spec` test file as it was testing all the private functions  and I moved it to `unit/application/node/transport` as` transport_private.spec.ts` it could be removed as well

### How was it solved?

Renamed all the files inside transport folder to `.ts` extension and resolved all the TS warnings and errors.

### How was it tested?

`npm run build` and `npm run test`


elements/lisk-transaction-pool/src/index.ts
```diff
@@ -16,3 +16,4 @@ export * from './transaction_pool';
 export { MaxHeap } from './max_heap';
 export { MinHeap } from './min_heap';
 export { Job } from './job';
+export { Transaction } from './types';
```
shuse2, line 19: I think this shouldn't be exported

framework/test/jest/unit/specs/application/node/transport/transport_private.spec.ts
```diff
@@ -12,38 +12,37 @@
...
+import { BlockInstance } from '@liskhq/lisk-chain';
+import { validator } from '@liskhq/lisk-validator';
+import { Logger, Channel, Processor } from '../../../../../../../src/types';
+import { Transport } from '../../../../../../../src/application/node/transport';
+
+import { genesis } from '../../../../../../fixtures/accounts';
+import { devnetNetworkIdentifier as networkIdentifier } from '../../../../../../utils/network_identifier';
+import { Block, GenesisBlock } from '../../../../../../fixtures/blocks';
+import { InvalidTransactionError } from '../../../../../../../src/application/node/transport/errors';
+/* eslint-disable @typescript-eslint/no-unnecessary-type-assertion */
```
shuse2, line 30: why this is disabled here?

framework/src/application/node/transport/transport.ts
```diff
@@ -0,0 +1,524 @@
...
+		}
+
+		return {
+			transactions: transactionsFromQueues,
+		};
+	}
+
+	public async handleEventPostTransaction(
+		data: EventPostTransactionData,
+	): Promise<{ transactionId: string } | object> {
```
shuse2, line 347: Maybe define separate type instead of `object`

framework/src/application/node/transport/transport.ts
```diff
@@ -0,0 +1,524 @@
...
+
+		const commonBlock = await this._chainModule.getHighestCommonBlock(data.ids);
+
+		return commonBlock
+			? this._chainModule.serializeBlockHeader(commonBlock)
+			: null;
+	}
+
+	public async handleEventPostBlock(
+		data: EventPostBlockData,
```
shuse2, line 230: I can be done in other issue, but if it comes from outside of framework, probably it's better to treat it as `unknown` at the beginning, and after it's validated, cast to the actual type?

framework/src/application/node/transport/transport.ts
```diff
@@ -0,0 +1,524 @@
...
+		});
+
+		// Rate limit for certain endpoints
+		this._rateTracker = {};
+		setInterval(() => {
+			this._rateTracker = {};
+		}, DEFAULT_RATE_RESET_TIME);
+	}
+
+	public get rateTracker(): RateTracker {
```
shuse2, line 99: those are all private properties, so no need to add `getter`

framework/test/jest/unit/specs/application/node/transport/transport_private.spec.ts
ishantiw: its because most of the tests in this file does not change the expression type and check the expectation of function being called with and that's why it's throwing  `This assertion is unnecessary since it does not change the type of the expression` for more than 10 tests

framework/test/jest/unit/specs/application/node/transport/transport.spec.ts
```diff
@@ -350,37 +357,36 @@ describe('Transport', () => {
...
 
-			it('should return transaction from pool', async () => {
-				const result = await transport.handleRPCGetTransactions(
-					undefined,
-					defaultPeerId,
-				);
-				expect(result.transactions).toStrictEqual([tx]);
+			it('should throw an error when no transaction ids are provided', async () => {
+				/* eslint-disable-next-line  @typescript-eslint/no-floating-promises */
+				expect(
```
shuse2, line 375: shouldn't it be ?
```suggestion
				await expect(
```

PR #5244, opened 2020-04-30
Convert application schemas to typescript - Closes #5196

### LINKED ISSUE #5196: Update application/schema to be typescript
### Description

Update framework application/schema to be typescript

### Motivation

### Acceptance Criteria

- All files under framework/src/application/schema should be in typescript

mitsuaki-u:
### What was the problem?

This PR resolves #5196

### How was it solved?

Converted application schema to `typescript`. 

### How was it tested?

Check schema


framework/src/application/schema/application_config_schema.ts
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright © 2019 Lisk Foundation
+ * Copyright © 2020 Lisk Foundation
```
shuse2, line 2: for the update, copyright should not be changed

mitsuaki-u:
> Also, tests are failing 😢

Tests will pass once all remaining code is converted to ts

PR #5302, opened 2020-05-04
Update controller to be typescript - Closes #5188

### LINKED ISSUE #5188: Update controller to be typescript
### Description

Update framework controller to be typescript

### Motivation

### Acceptance Criteria

- All files under framework/src/controller should be in typescript

nazarhussain:
### What was the problem?

This PR resolves #5188 

### How was it solved?

- Update controller to typescript 

### How was it tested?

- Check all eslint issue is resolved for controller 
- Check all unit tests are passing


framework/src/controller/bus.ts
```diff
@@ -0,0 +1,367 @@
...
+
+		let { channel } = options;
+
+		if (options.rpcSocketPath) {
+			const rpcSocket = axon.socket('req') as ReqSocket;
+			rpcSocket.connect(options.rpcSocketPath);
+
+			// TODO: Fix this override
+			// eslint-disable-next-line
+			// @ts-ignore
```
shuse2, line 159: why this needs `ts-ignore`?

framework/src/controller/channels/base_channel.ts
```diff
@@ -0,0 +1,119 @@
...
+				undefined,
+				isPublic,
+				handler,
+			);
+			this.actions[actionName] = action;
+		}
+		this.actionsList = Object.keys(this.actions);
+	}
+
+	public isValidEventName(name: string, throwError = true): boolean | never {
```
shuse2, line 67: why does this have `never` type?

framework/src/controller/channels/base_channel.ts
```diff
@@ -0,0 +1,119 @@
...
+		const result = eventWithModuleNameReg.test(name);
+
+		if (throwError && !result) {
+			throw new Error(`[${this.moduleAlias}] Invalid event name ${name}.`);
+		}
+		return result;
+	}
+
+	public isValidActionName(name: string, throwError = true): boolean | never {
```
shuse2, line 76: same as above

framework/src/controller/bus.ts
```diff
@@ -0,0 +1,367 @@
...
+		this.on(eventName, cb);
+
+		// Communicate through unix socket
+		if (this.config.ipc.enabled) {
+			this.subSocket?.on(eventName, cb);
+		}
+	}
+
+	// eslint-disable-next-line
+	// @ts-ignore
```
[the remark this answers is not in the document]
nazarhussain: The function `once` is defined in the base class with a different signature. It accepts the event name as a string and an array of strings as well. So overriding this behavior causing TS error.

framework/src/controller/channels/child_process_channel.ts
```diff
@@ -123,40 +155,52 @@ class ChildProcessChannel extends BaseChannel {
...
-				? new Action(actionName, params, this.moduleAlias)
-				: actionName;
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	public async invoke(actionName: string, params?: object): Promise<any> {
+		const action = new Action(actionName, params, this.moduleAlias);
 
 		if (action.module === this.moduleAlias) {
-			return this.actions[action.name].handler(action);
+			// eslint-disable-next-line
+			// @ts-ignore
```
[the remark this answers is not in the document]
nazarhussain: It's because of the previous structure of the action object. The handler can be specified as an action property or directly as action definition. So it's optional here and causing this complaint. 

For sure that's need to be improved, but should we do it here in this PR or separate PR?

framework/src/controller/channels/child_process_channel.ts
```diff
@@ -165,14 +209,18 @@ class ChildProcessChannel extends BaseChannel {
 				);
 			}
 
+			// eslint-disable-next-line
+			// @ts-ignore
```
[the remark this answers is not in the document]
nazarhussain: Same as above.

framework/test/jest/config/jest.config.base.js
```diff
@@ -14,7 +14,7 @@
 module.exports = {
 	globals: {
 		'ts-jest': {
-			tsConfig: './test/tsconfig.json',
+			tsConfig: '<rootDir>/test/tsconfig.json',
```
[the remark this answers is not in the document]
nazarhussain: Yes it is, its working fine on Jenkins and with IDEs as well, same was used in other config as well. 

setupFilesAfterEnv: ['<rootDir>/test/jest/config/setup.js'],

framework/src/controller/bus.ts
```diff
@@ -0,0 +1,367 @@
...
+
+		this.pubSocket = axon.socket('pub-emitter') as PubEmitterSocket;
+		this.pubSocket.bind(this.config.socketsPath.pub);
+
+		this.subSocket = axon.socket('sub-emitter') as SubEmitterSocket;
+		this.subSocket.bind(this.config.socketsPath.sub);
+
+		this.rpcSocket = axon.socket('rep') as RepSocket;
+		this.rpcServer = new RPCServer(this.rpcSocket);
+		this.rpcSocket.bind(this.config.socketsPath.rpc);
```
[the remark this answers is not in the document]
nazarhussain: The function `bind` binds the socket client or server to a port or socket file.

framework/src/controller/bus.ts
nazarhussain: Its completely wrong logic down there... 

```
			channel = new RPCClient(rpcSocket);
```

overriding the channel object passed from options as type `BaseChannel`. Already posted a comment to fix it.

framework/src/controller/channels/base_channel.ts
nazarhussain: It throws an error if the event name is not valid.

framework/src/controller/channels/base_channel.ts
nazarhussain: It throws an error if the action name is not valid.

nazarhussain:
@shuse2 Created a sperate issue to improve and small fixes opted from JS code. https://github.com/LiskHQ/lisk-sdk/issues/5312

framework/src/controller/channels/base_channel.ts
shuse2: I think throwing is not consider as return type

framework/src/controller/bus.ts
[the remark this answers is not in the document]
shuse2: I think we have to follow typescript, so let's update the function signature

framework/src/controller/channels/base_channel.ts
nazarhussain: Throwing is not return type but `never` is. It is used to specify that function could throw an error in that case it will not return anything. https://www.typescriptlang.org/docs/handbook/basic-types.html#never

framework/src/controller/types.ts
```diff
@@ -12,14 +12,9 @@
...
-const BaseChannel = require('./base_channel.js');
-const InMemoryChannel = require('./in_memory_channel.js');
-const ChildProcessChannel = require('./child_process_channel.js');
-
-module.exports = {
-	BaseChannel,
-	InMemoryChannel,
-	ChildProcessChannel,
-};
+export interface SocketPaths {
```
ManuGowda, line 15: maybe move this to `types` or `controller`?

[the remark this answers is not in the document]
shuse2: I think this is imported from multiple places, so it should be separate as it is

framework/src/controller/bus.ts
```diff
@@ -0,0 +1,340 @@
...
+
+		Object.keys(actions).forEach(actionName => {
+			if (this.actions[`${moduleAlias}:${actionName}`] !== undefined) {
+				throw new Error(`Action "${actionName}" already registered with bus.`);
+			}
+
+			this.actions[`${moduleAlias}:${actionName}`] = actions[actionName];
+		});
+
+		let { channel } = options;
```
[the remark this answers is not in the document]
shuse2: `rpcSocketPath, type` should be declared as `const`, and it will be error

PR #5304, opened 2020-05-05
Update application/node/forger to be typescript - Closes #5190

### LINKED ISSUE #5190: Update application/node/forger to be typescript
### Description

Update framework application/forger to be typescript

### Motivation

### Acceptance Criteria

- All files under framework/src/application/forger should be in typescript

ManuGowda:
### What was the problem?

This PR resolves #5190 

### How was it solved?
- Migrated forger to typescript
- Updated unit and integration tests
### How was it tested?
- Run framework `unit and integration` tests


framework/src/application/node/forger/forger.ts
```diff
@@ -239,17 +326,22 @@ class Forger {
...
+					}
 					return prev;
-				}
-				if (!prev || prev.count < current.count) {
-					return current;
-				}
-				return prev;
-			}, undefined);
+				},
+				{ count: 0, address: '', height: 0 },
```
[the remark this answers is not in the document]
ManuGowda: the default value is of type `UsedHashOnion`? I tried keeping it `undefined` then i have to do `undefined as unknown as UsedHashOnion`, this is bit dirty. got any suggestion?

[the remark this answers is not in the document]
shuse2:
```
const highestUsedHashOnion = usedHashOnions.reduce<UsedHashOnion | undefined>(
				(prev, current) => {
					if (current.address !== account.address) {
						return prev;
					}
					if (!prev || prev.count < current.count) {
						return current;
					}
					return prev;
				},
				undefined,
			);
```
I think correct type is like this, which return `UsedHashOnion | undefined`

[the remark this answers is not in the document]
shuse2: and the undefined is checked in the next step

framework/src/application/node/forger/forger.ts
```diff
@@ -72,29 +147,31 @@ class Forger {
...
 		return Object.keys(this.keypairs).length > 0;
 	}
 
-	// eslint-disable-next-line class-methods-use-this
-	async updateForgingStatus(publicKey, password, forging) {
+	public async updateForgingStatus(
+		publicKey: string,
+		password: string,
+		forging: boolean,
+	): Promise<{ readonly publicKey: string; readonly forging: boolean }> {
```
ishantiw, line 167: can we create a return type for this?  `Promise<UpdateForgingStatus>`

framework/src/application/node/forger/forger.ts
```diff
@@ -12,28 +12,104 @@
...
+	readonly chainModule: Chain;
+	readonly maxPayloadLength: number;
+	readonly forgingDelegates?: ReadonlyArray<DelegateConfig>;
+	readonly forgingForce?: boolean;
+	readonly forgingDefaultPassword?: string;
+	readonly forgingWaitThreshold: number;
+}
+
+export class Forger {
+	private readonly logger: Logger;
```
ishantiw, line 89: can we add underscore as a prefix to private properties as some functions in the same class are following that pattern so it looks a bit inconsistent

framework/src/application/node/forger/forger.ts
```diff
@@ -86,26 +91,26 @@ interface ForgerConstructor {
...
+	private readonly _logger: Logger;
+	private readonly _storage: Storage;
+	private readonly _processorModule: ProcessorModule;
+	private readonly _dposModule: Dpos;
+	private readonly _bftModule: BFT;
+	private readonly _transactionPoolModule: TransactionPool;
+	private readonly _chainModule: Chain;
+	private readonly _keypairs: KeyPairs;
+	private readonly _config: {
 		readonly forging: {
```
ishantiw, line 103: I missed this out before but why we don't have any access modifier for this property? `public`?

ManuGowda: _config is set to private so.

ishantiw: but the property should have an access modifier for consistency? right now someone can expose it if it doesn't need to be exposed

ishantiw: Sorry I didn't see that it was nested, all good

PR #5314, opened 2020-05-07
Update lisk-chain to use lisk-db - Closes #5257

### LINKED ISSUE #5257: Update lisk-chain data_access layer
### Description

With the introduction of an `embedded` database system, the data access layer interface needs to be updated to support queries using the embedded database system's APIs.

### Acceptance Criteria
- Update `storage` related `types` in `lisk-chain`
- Update `data_access->storage` to get `EmbeddedDB` instance
- Update  `data_access->storage` public interfaces to use `EmbeddedDB` interface
- Update unit tests according to `EmbeddedDB` interface
- Add integration tests for `data_access` layer
- Use `json` encoder

### Additional Information

- Database keys should be as below, and height key should be big endian encoded number

```
blocks:height:xxxx => id
blocks:id:xxxx => BlockHeader
transactions:id:xxxx => Transaction
transactions:blockID:xxxxx => TxID[]
accounts:address:xxxx => Account
tempBlocks:height:xxx => FullBlock
```

- Some of the interface supported by EmbeddedDB

```js
// put
await db.get('blocks:height:00000000000000001', 'test')

// get
await db.get('blocks:height:0000000000000001')

// del
await db.del('blocks:height:00000000000000001')

// batch
await db.batch()
  .del('blocks:height:0000000000000001')
  .put('blocks:height:0000000000000010', 'test1')
  .put('blocks:height:0000000000000010', 'test2')
  .put('blocks:height:0000000000000010', 'test3')
  .write()

// events
db.on(<operation>, (k, v) => {})

// put
await db.get('blocks:height:00000000000000001', 'test1')

// get
await db.get('blocks:height:0000000000000001')

// del
await db.del('blocks:height:00000000000000001')

// batch
await db.batch()
  .del('blocks:height:0000000000000001')
  .put('blocks:height:0000000000000010', 'test2')
  .put('blocks:height:0000000000000010', 'test3')
  .put('blocks:height:0000000000000010', 'test4')
  .write()

// events
db.on(<operation>, (k, v) => {})
```

Related issues: #5238

### LINKED ISSUE #5357: Update lisk-db to use value as Buffer
### Description

Update `lisk-db` to manage `value` as Buffer

### Motivation

While moving to use `lisk-codec` value should not be encoded from JSON anymore

### Acceptance Criteria

- All the tests are updated
- All the usage to `Buffer`, without using lisk-codec

### Additional Information
- Related to #5264

shuse2:
### What was the problem?

This PR resolves #5257 

### How was it solved?

- Updated lisk-chain to use the lisk-db instance instead of storage
- Remove `prepare` step from transactions

### How was it tested?

- Add integration tests for the dataAccess/storage
- Update all the tests


elements/lisk-chain/src/chain.ts
```diff
@@ -378,8 +359,7 @@ export class Chain {
 	}
 
 	public async resetState(): Promise<void> {
-		await this.storage.entities.Account.resetMemTables();
-		await this.storage.entities.ConsensusState.delete();
+		await this.dataAccess.resetMemTables();
```
pablitovicente, line 362: While we are at these could we remove the `Mem` part from names? It has always seem that something was exclusively in RAM memory but the truth is that it might or not depending on the storage backend.

elements/lisk-chain/src/chain.ts
```diff
@@ -432,26 +412,27 @@ export class Chain {
 			removeFromTempTable: false,
 		},
 	): Promise<void> {
-		return this.storage.entities.Block.begin('saveBlock', async tx => {
```
pablitovicente, line 435: Could you describe how this is guaranteed with the new store?

elements/lisk-chain/src/chain.ts
```diff
@@ -432,26 +412,27 @@ export class Chain {
...
-				.getUpdated()
-				.map(anAccount => anAccount.toJSON());
+		const accounts = stateStore.account
+			.getUpdated()
+			.map(anAccount => anAccount.toJSON());
+
+		if (saveOnlyState) {
+			const batch = this._db.batch();
+			stateStore.finalize(batch);
+			await batch.write();
```
nazarhussain, line 422: Can we move this logic to `dataAccess` object?

elements/lisk-chain/src/data_access/constants.ts
```diff
@@ -0,0 +1,22 @@
...
+ */
+
+export const DB_KEY_BLOCKS_ID = 'blocks:id:';
+export const DB_KEY_BLOCKS_HEIGHT = 'blocks:height:';
+export const DB_KEY_TRANSACTIONS_BLOCK_ID = 'transactions:blockID:';
+export const DB_KEY_TRANSACTIONS_ID = 'transactions:id:';
+export const DB_KEY_TEMPBLOCKS_HEIGHT = 'tempBlocks:height:';
+export const DB_KEY_ACCOUNTS_ADDRESS = 'accounts:address:';
+export const DB_KEY_CHAIN_STATE = 'chain:';
+export const DB_KEY_CONSENSUS_STATE = 'consensus:';
```
nazarhussain, line 22: Please remove the `:` suffix from key names. Eventually, the key will be calculated with the additional suffix e.g. block id or height, so add that `:` in there. 

Or other solution would convert these to a template function which takes the object of concerning data and generate the full key.

```js
export const DB_KEY_BLOCKS_ID = ({id}) => `blocks:id:${id}`;
```

elements/lisk-chain/src/data_access/storage.ts
```diff
@@ -12,291 +12,537 @@
...
 	/*
 		Block headers
 	*/
-
-	public async getBlockHeaderByID(id: string): Promise<BlockJSON> {
-		const [block] = await this._storage.entities.Block.get({ id });
-
+	public async getBlockHeaderByID(id: string): Promise<BlockHeaderJSON> {
+		const block = await this._db.get<BlockHeaderJSON>(
+			`${DB_KEY_BLOCKS_ID}${id}`,
```
nazarhussain, line 51: Even reading here it looks better to read if we have suffix right where we use. 

`${DB_KEY_BLOCKS_ID}:${id}`

elements/lisk-chain/src/data_access/storage.ts
```diff
@@ -12,291 +12,537 @@
...
 	}
 
-	public async getBlockHeaderByHeight(height: number): Promise<BlockJSON> {
-		const [block] = await this._storage.entities.Block.get({ height });
-
-		return block;
+	public async getBlockHeaderByHeight(
+		height: number,
+	): Promise<BlockHeaderJSON> {
+		const stringHeight = formatInt(height);
```
nazarhussain, line 72: Can we have better name for this function. Its actually converting the integer to string so `intToString` or `covnertToString`

elements/lisk-chain/src/data_access/storage.ts
```diff
@@ -12,291 +12,537 @@
...
-			{},
-			{ limit: 1, sort: 'height:desc' },
-		);
+	public async getLastBlockHeader(): Promise<BlockHeaderJSON> {
+		const stream = this._db.createReadStream({
+			gte: getFirstPrefix(DB_KEY_BLOCKS_HEIGHT),
+			lte: getLastPrefix(DB_KEY_BLOCKS_HEIGHT),
+			reverse: true,
+			limit: 1,
+		});
```
nazarhussain, line 122: Can you explain a bit what this `this._db.createReadStream` parameters are actually doing. I noticed that this `getFirstPrefix` is just adding a hexadecimal `\x00` and `\xFF` so who that read stream works with these suffixes?

elements/lisk-chain/src/data_access/storage.ts
```diff
@@ -12,291 +12,537 @@
...
+
+		return blocks[0];
 	}
 
 	public async getBlocksCount(): Promise<number> {
-		const count = await this._storage.entities.Block.count({}, {});
-
-		return count;
+		const lastBlock = await this.getLastBlockHeader();
+		return lastBlock.height;
```
nazarhussain, line 161: This will be a behavioral change from the previous implementation. If the genesis block has a particular height in case of re-genesis then both of the implementations will return differen values. Do we really want that change?

elements/lisk-chain/src/data_access/storage.ts
nazarhussain: Adding a comment line above which refers to documentation of that function would be nice.

elements/lisk-chain/src/data_access/storage.ts
```diff
@@ -12,291 +12,537 @@
...
+				endHeight,
+			);
+			const blockIDs = headers.map(header => header.id);
+			const transactionIDs = [];
+			const batch = this._db.batch();
+			for (const blockID of blockIDs) {
+				try {
+					const ids = await this._db.get<string[]>(
+						`${DB_KEY_TRANSACTIONS_BLOCK_ID}${blockID}`,
+					);
```
nazarhussain, line 303: Don't we already have a fucntion `geTransactions(blockID)`?

elements/lisk-chain/src/data_access/storage.ts
```diff
@@ -12,291 +12,537 @@
...
-
-		return value;
+		try {
+			const value = await this._db.get<string>(`${DB_KEY_CHAIN_STATE}${key}`);
+
+			return value;
+		} catch (error) {
+			if (error instanceof NotFoundError) {
+				return undefined;
+			}
```
nazarhussain, line 339: Why would someone call it with unknow key? And in that case, we should throw an error. If it's for handling the case of setting the key first time then should be done outside the scope of dataAccess.

elements/lisk-chain/src/data_access/storage.ts
```diff
@@ -12,291 +12,537 @@
...
 	public async getConsensusState(key: string): Promise<string | undefined> {
-		const value = await this._storage.entities.ConsensusState.getKey(key);
-
-		return value;
+		try {
+			const value = await this._db.get<string>(
+				`${DB_KEY_CONSENSUS_STATE}${key}`,
+			);
+
+			return value;
```
nazarhussain, line 353: The same goes here.

elements/lisk-chain/src/data_access/storage.ts
```diff
@@ -12,291 +12,537 @@
...
+			batch.del(`${DB_KEY_TRANSACTIONS_BLOCK_ID}${header.id}`);
+		}
+		if (saveToTemp) {
+			batch.put(
+				`${DB_KEY_TEMPBLOCKS_HEIGHT}${formatInt(blockJSON.height)}`,
+				blockJSON,
+			);
+		}
+		stateStore.finalize(batch);
+		await batch.write();
```
nazarhussain, line 520: If batch failed writing, are we not resetting the state store snapshot? Otherwise how the `getUpdated` identify changes to accounts which change after finalization?

elements/lisk-chain/src/data_access/storage.ts
```diff
@@ -12,291 +12,537 @@
...
+			const startHeight = i * batchSize + start + i;
+			const endHeight = startHeight + batchSize - 1;
+			const headers = await this.getBlockHeadersByHeightBetween(
+				startHeight,
+				endHeight,
+			);
+			const blockIDs = headers.map(header => header.id);
+			const transactionIDs = [];
+			const batch = this._db.batch();
+			for (const blockID of blockIDs) {
```
pablitovicente, line 299: If not possible to avoid the nested loop, can we move this to a helper function? (as if this code grows the nested loop makes it harder to read and reason about)

elements/lisk-chain/src/data_access/storage.ts
```diff
@@ -12,291 +12,537 @@
...
+				batch.del(`${DB_KEY_BLOCKS_ID}${blockID}`);
+				batch.del(`${DB_KEY_TRANSACTIONS_BLOCK_ID}${blockID}`);
+			}
+			// tslint:disable-next-line no-let
+			for (let j = startHeight; j <= endHeight; j += 1) {
+				batch.del(`${DB_KEY_BLOCKS_HEIGHT}${formatInt(j)}`);
+			}
+			for (const txID of transactionIDs) {
+				batch.del(`${DB_KEY_TRANSACTIONS_ID}${txID}`);
+			}
```
pablitovicente, line 319: This will also be nicer as helper functions

elements/lisk-chain/src/data_access/storage.ts
```diff
@@ -12,291 +12,537 @@
...
+					}
+				})
+				.on('error', error => {
+					reject(error);
+				})
+				.on('end', () => {
+					resolve(accountJSONs);
+				});
+		});
+		accounts.sort((a, b) => {
```
pablitovicente, line 415: maybe there're more places where we need to sort BigInts so this could be better in a helper function?

elements/lisk-chain/src/data_access/storage.ts
```diff
@@ -12,291 +12,537 @@
...
+			}
+			return a.address.localeCompare(b.address);
+		});
 
 		return accounts;
 	}
 
-	public async resetAccountMemTables(): Promise<void> {
-		await this._storage.entities.Account.resetMemTables();
+	public async resetMemTables(): Promise<void> {
```
pablitovicente, line 429: As mentioned on different file it would be nice to remove `Mem` from here to avoid confusing hinting

elements/lisk-chain/src/state_store/account_store.ts
```diff
@@ -82,16 +60,15 @@ export class AccountStore {
 		}
 
 		// Account was not cached previously so we try to fetch it from db
-		const [elementFromDB] = await this._account.get(
-			{ [this._primaryKey]: primaryValue },
-			{ limit: null },
+		const elementFromDB = await this._dataAccess.getAccountByAddress(
```
pablitovicente, line 63: shouldn't be better to check first its existance in `this._data` and if not there then hit the storage?

elements/lisk-chain/src/state_store/account_store.ts
```diff
@@ -156,51 +132,18 @@ export class AccountStore {
...
-						// eslint-disable-next-line
-						(data as any)[key] = (updatedItem as any)[key];
-
-						return data;
-					},
-					{},
+	public finalize(batch: BatchChain): void {
+		for (const account of this._data) {
+			if (this._updatedKeys.has(account.address)) {
+				batch.put(
```
pablitovicente, line 142: is `batch.put` sync?

elements/lisk-chain/src/state_store/state_store.ts
```diff
@@ -56,11 +57,9 @@ export class StateStore {
 		this.chain.restoreSnapshot();
 	}
 
-	public async finalize(tx: StorageTransaction): Promise<void> {
-		await Promise.all([
-			this.account.finalize(tx),
-			this.chain.finalize(tx),
-			this.consensus.finalize(tx),
-		]);
+	public finalize(batch: BatchChain): void {
```
pablitovicente, line 60: Seems previous code was using a SQL transaction how does this change guaranteed that either all or none of the updates happen as to maintain consistency?

elements/lisk-chain/src/chain.ts
shuse2: This function will be removed in https://github.com/LiskHQ/lisk-sdk/issues/5320

elements/lisk-chain/src/chain.ts
shuse2: This will be solved by `batch` write.
when we want to atomically save the data,
```
const batch = this._db.batch();
batch.put('xx', xxx);
batch.del('xx', xxx);
await batch.write();
```
and here `put` and `del` for batch are sync

elements/lisk-chain/src/data_access/constants.ts
shuse2: we will remove the suffix `:` from the constants

elements/lisk-chain/src/data_access/storage.ts
shuse2: Although `formatInt` returning a string, it's returning binary representation of int in string.
Therefore, I prefer to keep it as it is

elements/lisk-chain/src/data_access/storage.ts
shuse2: the createReadStream will iterate through the `key-value` pair in lexicographical order.
`limit` is the limit of iteration, and `reverse` is the order of iteration between the lexicographical order
`gte` and `lte` is the key where you want to start or finish

elements/lisk-chain/src/data_access/storage.ts
shuse2: that's true, I think we don't have a way to get the total number of blocks other than iterating through the keys.
However, this will be removed in https://github.com/LiskHQ/lisk-sdk/issues/5320

elements/lisk-chain/src/data_access/storage.ts
shuse2: same as above

elements/lisk-chain/src/data_access/storage.ts
shuse2: I tried to keep it non-breaking as possible. For now, the bft, dpos and framework is using this and expecting not to throw and return `undefined` if not exists.
its doing like if undefined, then create new one

elements/lisk-chain/src/data_access/storage.ts
shuse2: If batching is failed, then everything fails, and the state store instance will be thrown away

elements/lisk-chain/src/data_access/storage.ts
shuse2: As above, this function will be removed in https://github.com/LiskHQ/lisk-sdk/issues/5320

elements/lisk-chain/src/data_access/storage.ts
shuse2: same as above

elements/lisk-chain/src/data_access/storage.ts
shuse2: I think this sorting is also done in the `dpos`, and as mentioned in this function comment, this will be removed.
Also, it's hard to generalize this `sort`, because this is very specific to the delegate sorting, and this should be only done in the `dpos` library

elements/lisk-chain/src/state_store/account_store.ts
shuse2: it's not changed, but it's checking in `this._data` above

elements/lisk-chain/src/state_store/account_store.ts
shuse2: `batch.put` is `sync` and only `batch.write` is async, and that will save all the updated things atomically

elements/lisk-chain/src/state_store/state_store.ts
shuse2: As mentioned above, `batch.write` will save to the database with consistency.

framework/test/utils/kv_store.ts
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright © 2018 Lisk Foundation
```
ishantiw, line 2: `2018` ?

PR #5316, opened 2020-05-08
Update framework root files to typescript - Closes #5195 and #5197

### LINKED ISSUE #5195: Update application/node root files to be typescript
### Description

Update framework application/node root files to be typescript

### Motivation

### Acceptance Criteria

- All files under framework/src/application/node root files should be in typescript

### LINKED ISSUE #5197: Update application root files to be typescript
### Description

Update framework application root files to be typescript

### Motivation

### Acceptance Criteria

- All files under framework/src/application root files should be in typescript

shuse2:
### What was the problem?

This PR resolves #5195  and resolves #5197 

### How was it solved?

- Update remaining root files to the typescript

### How was it tested?

- Application should run
- No linting error for typescript by `npx eslint --ext ts .`


elements/lisk-transaction-pool/src/types.ts
```diff
@@ -31,6 +31,7 @@ export interface TransactionObject {
 
 export interface TransactionFunctions {
 	getBytes: () => Buffer;
+	toJSON(): unknown;
```
nazarhussain, line 34: Don't we know all transaction types which can be used here with `|`

framework/src/application/application.ts
```diff
@@ -82,29 +83,85 @@ const registerProcessHooks = app => {
...
+		readonly delegates: DelegateConfig[];
+		readonly force?: boolean;
+		readonly defaultPassword?: string;
+	};
+	readonly rebuildUpToRound: string;
+	readonly network: NetworkConfig;
+	constants: {
+		[key: string]: {} | string | number | undefined;
+	};
+	genesisConfig: {};
```
nazarhussain, line 122: We should define each attribute of the `genesisConfig`

framework/src/application/application.ts
```diff
@@ -82,29 +83,85 @@ const registerProcessHooks = app => {
...
+	readonly forging: {
+		readonly waitThreshold: number;
+		readonly delegates: DelegateConfig[];
+		readonly force?: boolean;
+		readonly defaultPassword?: string;
+	};
+	readonly rebuildUpToRound: string;
+	readonly network: NetworkConfig;
+	constants: {
+		[key: string]: {} | string | number | undefined;
```
nazarhussain, line 120: Why would the constant value be `undefined`?

framework/src/application/application.ts
```diff
@@ -82,29 +83,85 @@ const registerProcessHooks = app => {
...
+	};
+	genesisConfig: {};
+	components: {
+		[key: string]: {} | undefined;
```
nazarhussain, line 124: The component config should always be an object.

framework/src/application/application.ts
```diff
@@ -82,29 +83,85 @@ const registerProcessHooks = app => {
...
+	public config: ApplicationConfig;
+	public constants: NodeConstants;
+
+	private _node!: Node;
+	private _network!: Network;
+	private _controller!: Controller;
+	private _applicationState!: ApplicationState;
+	private _transactions: { [key: number]: typeof BaseTransaction };
+	private _modules: { [key: string]: InstantiableModule<BaseModule> };
+	private _channel!: InMemoryChannel;
```
nazarhussain, line 143: Are we really following this `_` notation for private members throughout the code?

framework/src/application/application.ts
```diff
@@ -82,29 +83,85 @@ const registerProcessHooks = app => {
...
+
+	private readonly _genesisBlock: GenesisBlockInstance;
+	private _migrations: { [key: string]: object };
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	private readonly storage: any;
```
nazarhussain, line 148: Storage has a particular interface already defined in types, why not to use it?

framework/src/application/application.ts
```diff
@@ -82,29 +83,85 @@ const registerProcessHooks = app => {
...
+
+	public constructor(
+		genesisBlock: object,
```
nazarhussain, line 151: We already have a genesis block interface defined, why not to use it here?

framework/src/application/application.ts
shuse2: This `undefined` is to specify the key might not exist.
when we do `constants['some-key']`, it will be required to check if the constant exist or not

framework/src/application/application.ts
shuse2: same as above

framework/src/application/application.ts
shuse2: yes we do follow the `_` notation for the private properties

framework/src/application/application.ts
shuse2: This im not sure whether we should use the type of `object` here.
I think in the code below, after checking the schema, it's converting to the genesis block type.
I think either way should work, but maybe for user it's better to use the genesis block type

framework/src/application/application.ts
shuse2: The type was not enough, and needed to add more things to be compatible with all the libraries, but since we are gonna remove it, i thought `any` for now is fine

elements/lisk-transaction-pool/src/types.ts
shuse2: Transaction pool doesn't know about any particular transaction

framework/src/application/application.ts
```diff
@@ -82,29 +83,98 @@ const registerProcessHooks = app => {
...
+	public config: ApplicationConfig;
+	public constants: NodeConstants;
+
+	private _node!: Node;
+	private _network!: Network;
+	private _controller!: Controller;
+	private _applicationState!: ApplicationState;
+	private _transactions: { [key: number]: typeof BaseTransaction };
+	private _modules: { [key: string]: InstantiableModule<BaseModule> };
+	private _channel!: InMemoryChannel;
```
nazarhussain, line 156: `!` 👍

framework/src/application/node/node.ts
```diff
@@ -0,0 +1,695 @@
...
+				await this._rebuilder.rebuild(this._options.rebuildUpToRound);
+				this._logger.info(
+					{
+						rebuildUpToRound: this._options.rebuildUpToRound,
+					},
+					'Successfully rebuild the blockchain',
+				);
+				process.exit(0);
+				return;
+			}
```
ManuGowda, line 225: since this logic is only for rebuild, can we move this to top of bootstrap? that way its much clearer while reading as well.

framework/src/application/node/utils/sequence.ts
```diff
@@ -36,15 +36,21 @@ interface Task<T = unknown> {
 	done: WorkerResult<T>;
 }
 
+interface SequeceConfig {
```
ManuGowda, line 39: `SequenceConfig`

framework/src/version.ts
```diff
@@ -12,8 +12,4 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
-'use strict';
-
-const Node = require('./node');
-
-module.exports = { Node };
+export const version = '0.1.0';
```
ManuGowda, line 15: is this version number correct? not `0.6.0`?

shuse2: I think this need to be re considered as it is never used, but this is correct for the typescript conversion

framework/src/application/node/node.ts
shuse2: I think other bootstrapping needs to happen before, that's why it's here.
Also, rebuilder will be removed

framework/src/version.ts
ManuGowda: @nazarhussain do we still need this? I am not sure when we will use this and upgrade it

framework/src/application/application_state.ts
```diff
@@ -0,0 +1,98 @@
...
+		readonly version: string;
+		readonly wsPort: number;
+		readonly httpPort: number;
+		readonly protocolVersion: string;
+		readonly networkId: string;
+	};
+	readonly logger: Logger;
+}
+
+interface ApplicationStateUpdate {
```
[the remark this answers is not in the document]
shuse2: I think more correct to say `ApplicationStateUpdateInput`, but maybe too verbose?

framework/src/application/network/network.ts
```diff
@@ -45,11 +47,29 @@ const NETWORK_INFO_KEY_TRIED_PEERS = 'network:triedPeersList';
...
-	readonly options: P2PConfig;
-	readonly channel: Channel;
+	readonly options: NetworkConfig;
+	readonly channel: InMemoryChannel;
 	readonly logger: Logger;
 	readonly storage: Storage;
-	secret: string;
+}
+
+export interface NetworkConfig {
```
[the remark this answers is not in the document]
shuse2: I think this config is for the Network class, so it would be more easier if we need to edit network options?

framework/src/application/application_state.ts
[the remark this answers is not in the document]
ManuGowda: ahh i thought it was return type.. then its fine

framework/src/application/node/node.ts
ManuGowda: if the option `rebuildUpToRound` exists it doesn't need other bootstrapping info? 
yeah since we remove its ok.

PR #5319, opened 2020-05-12
Add varint reader/writer - Closes #5219 & #5221 & #5229 & #5230

### LINKED ISSUE #5219: Create VarInt writer
### Description

Create function `varint(value: number, schema) => buffer`

### Motivation

we need a function to write [VarInts](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0027.md#encoding-integers).

### Acceptance Criteria

- `varint()` should accept numbers (including bigInts)
- `varint()` returns a buffer that can be passed through the varInt reader to get the original value
- Full test coverage
- Performance stats are provided based on the writing of values of different sizes and measure the writing of 100, 500, 1000, 30000 varints

### Additional Information

**_Check if it will make sense to implement signed varints in same issue._**

Related issues: #5200

### LINKED ISSUE #5221: Create VarInt reader
### Description

Create function `varint(buffer, schema) => value`

### Motivation

`lisk-codec` will include a set of `Readers` for each data type defined in the LIP, to that end
we need a function to read [VarInts](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0027.md#encoding-integers).

### Acceptance Criteria

- `varint()` should accept a buffer representation of a varint value
- `varint()` returns either a number or a BigInt (this is controlled by the schema)
- Full test coverage
- Performance stats are provided based on the reading of values of different sizes and measure the reading of 100, 500, 1000, 30000 varints

### Additional Information

**_Check if it will make sense to implement signed varints in same issue._**

Related issues: #5200

### LINKED ISSUE #5229: Create signedVarInt writer
### Description

Create function `signedVarint(value: number, schema) => buffer`

### Motivation

we need a function to write [Signed VarInts](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0027.md#encoding-integers).

### Acceptance Criteria

- `signedVarint()` should accept signed numbers (including bigInts)
- `signedVarint()` returns a buffer that can be passed through the signedVarint reader to get the original value
- Full test coverage
- Performance stats are provided based on the writing of values of different sizes and measure the writing of 100, 500, 1000, 30000 signedVarint

### Additional Information

**_Check if it will make sense to implement along with varints in same issue._**

Related Issues: #5200

shuse2:
### What was the problem?

This PR resolves #5219 , resolves #5221, resolves #5229 and #5230 

### How was it solved?

Add  reader and writer for varint and zigzag, which are used for sint and uint

### How was it tested?

- Add test cases for each of them

```
writeVarInt#uint32 x 4,865,062 ops/sec ±13.33% (82 runs sampled)
writeVarInt#uint64 x 432,718 ops/sec ±13.25% (78 runs sampled)
writeSignedVarInt#sint32 x 6,388,955 ops/sec ±1.44% (87 runs sampled)
writeSignedVarInt#sint64 x 274,677 ops/sec ±0.58% (91 runs sampled)
readVarInt#uint32 x 3,069,689 ops/sec ±0.52% (92 runs sampled)
readVarInt#uint64 x 183,194 ops/sec ±5.88% (85 runs sampled)
readSignedVarInt#sint32 x 3,136,256 ops/sec ±0.63% (86 runs sampled)
readSignedVarInt#sint64 x 182,973 ops/sec ±0.42% (91 runs sampled)
```

### DIFF
--- a/elements/lisk-codec/src/varint.ts
+++ b/elements/lisk-codec/src/varint.ts
@@ -35,7 +35,6 @@ const writeVarIntNumber = (value: number, _schema: SchemaProperty): Buffer => {
 	}
 
 	result[index] = value;
-	index += 1;
 
 	return Buffer.from(result);
 };
@@ -50,7 +49,6 @@ const writeVarIntBigInt = (value: bigint, _schema: SchemaProperty): Buffer => {
 	}
 
 	result[Number(index)] = Number(value);
-	index += 1;
 
 	return Buffer.from(result);
 };
--- a/elements/lisk-codec/src/varint.ts
+++ b/elements/lisk-codec/src/varint.ts
@@ -25,7 +25,7 @@ interface SchemaProperty {
 	readonly dataType: string;
 }
 
-const writeVarIntNumber = (value: number, _schema: SchemaProperty): Buffer => {
+const writeVarIntNumber = (value: number): Buffer => {
 	const result: number[] = [];
 	let index = 0;
 	while (value > rest) {
@@ -39,7 +39,7 @@ const writeVarIntNumber = (value: number, _schema: SchemaProperty): Buffer => {
 	return Buffer.from(result);
 };
 
-const writeVarIntBigInt = (value: bigint, _schema: SchemaProperty): Buffer => {
+const writeVarIntBigInt = (value: bigint): Buffer => {
 	const result: number[] = [];
 	let index = 0;
 	while (value > BigInt(rest)) {
@@ -53,10 +53,8 @@ const writeVarIntBigInt = (value: bigint, _schema: SchemaProperty): Buffer => {
 	return Buffer.from(result);
 };
 
-export const writeVarInt = (value: int, schema: SchemaProperty): Buffer =>
-	isNumber(value)
-		? writeVarIntNumber(value, schema)
-		: writeVarIntBigInt(value, schema);
+export const writeVarInt = (value: int): Buffer =>
+	isNumber(value) ? writeVarIntNumber(value) : writeVarIntBigInt(value);
 
 export const writeSignedVarInt = (
 	value: int,
@@ -64,13 +62,13 @@ export const writeSignedVarInt = (
 ): Buffer => {
 	if (schema.dataType === 'sint32') {
 		const number = Number(value);
-		return writeVarIntNumber(((number << 1) ^ (number >> 31)) >>> 0, schema);
+		return writeVarIntNumber(((number << 1) ^ (number >> 31)) >>> 0);
 	}
 	const number = BigInt(value);
-	return writeVarInt((number << BigInt(1)) ^ (number >> BigInt(63)), schema);
+	return writeVarInt((number << BigInt(1)) ^ (number >> BigInt(63)));
 };
 
-const readVarIntNumber = (buffer: Buffer, _schema: SchemaProperty): number => {
+const readVarIntNumber = (buffer: Buffer): number => {
 	let result = 0;
 	let index = 0;
 	for (let shift = 0; shift < 32; shift += 7) {
@@ -87,7 +85,7 @@ const readVarIntNumber = (buffer: Buffer, _schema: SchemaProperty): number => {
 	throw new Error('Out of range');
 };
 
-const readVarIntBigInt = (buffer: Buffer, _schema: SchemaProperty): bigint => {
+const readVarIntBigInt = (buffer: Buffer): bigint => {
 	let result = BigInt(0);
 	let index = 0;
 	for (let shift = BigInt(0); shift < BigInt(64); shift += BigInt(7)) {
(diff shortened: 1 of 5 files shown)


elements/lisk-codec/src/varint.ts
```diff
@@ -0,0 +1,129 @@
...
+export const writeSignedVarInt = (
+	value: int,
+	schema: SchemaProperty,
+): Buffer => {
+	if (schema.dataType === 'sint32') {
+		const number = Number(value);
+		return writeVarIntNumber(((number << 1) ^ (number >> 31)) >>> 0);
+	}
+	const number = BigInt(value);
+	return writeVarInt((number << BigInt(1)) ^ (number >> BigInt(63)));
```
MaximeGagnebin, line 68: I think we should avoid the use of (number >> BigInt(63)) here. This limits the use of the varint encoder to the range of int64.
I think it is better to do something like
writeVarInt( (number >= 0n) ? 2n*number : -2n*number - 1n )

elements/lisk-codec/src/varint.ts
```diff
@@ -0,0 +1,129 @@
...
+};
+
+const readVarIntNumber = (buffer: Buffer): number => {
+	let result = 0;
+	let index = 0;
+	for (let shift = 0; shift < 32; shift += 7) {
```
MaximeGagnebin, line 74: I think that this loop is too short. The varint corresponding to a 32 bit number can be up to 6 bytes. This only loops through 5 bytes.

elements/lisk-codec/src/varint.ts
```diff
@@ -0,0 +1,129 @@
...
+			throw new Error('Invalid buffer length');
+		}
+		const bit = buffer[index];
+		index += 1;
+		result = (result | ((bit & rest) << shift)) >>> 0;
+		if ((bit & msg) === 0) {
+			return result;
+		}
+	}
+	throw new Error('Out of range');
```
MaximeGagnebin, line 85: this error message should be taken inside the loop and be inhahced a bit. You can have a 6 bytes varint fitting in the number range, and you have 6 bytes varint outside the number range.

elements/lisk-codec/src/varint.ts
```diff
@@ -0,0 +1,129 @@
...
+		value >>= BigInt(7);
+		index += 1;
+	}
+
+	result[Number(index)] = Number(value);
+
+	return Buffer.from(result);
+};
+
+export const writeVarInt = (value: int): Buffer =>
```
pablitovicente, line 56: shouldn't this also expect `schema` as a parameter to keep the interface the same for all writers/readers?

elements/lisk-codec/src/varint.ts
```diff
@@ -0,0 +1,129 @@
...
+			return result;
+		}
+	}
+	throw new Error('Out of range');
+};
+
+const readVarIntBigInt = (buffer: Buffer): bigint => {
+	let result = BigInt(0);
+	let index = 0;
+	for (let shift = BigInt(0); shift < BigInt(64); shift += BigInt(7)) {
```
MaximeGagnebin, line 91: Here too, I don't think we should limit our big varints to the uint64 range.

shuse2: this varint is not for general varint encoding and I think the LIP and protobuf doesn't specify number greater than uint64, so I think it's better to throw error for this

elements/lisk-codec/src/varint.ts
shuse2: I think it's correct to be outside. If the bit has terminating prefix, then it will return from inside. If it doesn't have the terminating prefix, it will come to the error, which means it's out of range

elements/lisk-codec/src/varint.ts
shuse2: Same as below

elements/lisk-codec/src/varint.ts
MaximeGagnebin: my point was that you can have the terminating prefix, but still be out of uint32 range.

elements/lisk-codec/src/varint.ts
MaximeGagnebin: I understand the use below (it somehow limits the loop length). But here, I don't see why not write the general formula.

elements/lisk-codec/test/varint.spec.ts
```diff
@@ -0,0 +1,213 @@
...
+				readVarInt(Buffer.from('ffffffff07', 'hex'), { dataType: 'uint32' }),
+			).toEqual(2147483647);
+			expect(
+				readVarInt(Buffer.from('ffffffff0f', 'hex'), { dataType: 'uint32' }),
+			).toEqual(4294967295);
+		});
+
+		it('should fail to decode uint32 when input is out of range', () => {
+			expect(() =>
+				readVarInt(Buffer.from('ffffffff7f', 'hex'), { dataType: 'uint32' }),
```
MaximeGagnebin, line 124: this is out of range (the test is good). But if we want the smallest out of range, then we could test '8080808010'. Which will throw accordingly (the 5th byte is larger than 0f)

elements/lisk-codec/test/varint.spec.ts
```diff
@@ -0,0 +1,213 @@
...
+				}),
+			).toEqual(BigInt('18446744073709551615'));
+		});
+
+		it('should fail to decode uint64 when input is out of range', () => {
+			expect(() =>
+				readVarInt(Buffer.from('ffffffffffffffffff02', 'hex'), {
+					dataType: 'uint32',
+				}),
+			).toThrow('Value out of range of uint32');
```
MaximeGagnebin, line 156: this should be uint64 ?

elements/lisk-codec/test/varint.spec.ts
```diff
@@ -0,0 +1,213 @@
...
+			expect(
+				readVarInt(Buffer.from('ffffffffffffffffff01', 'hex'), {
+					dataType: 'uint64',
+				}),
+			).toEqual(BigInt('18446744073709551615'));
+		});
+
+		it('should fail to decode uint64 when input is out of range', () => {
+			expect(() =>
+				readVarInt(Buffer.from('ffffffffffffffffff02', 'hex'), {
```
MaximeGagnebin, line 153: as above, if we want to test the smallest out of range, we can test '80808080808080808002'. Which will correctly throw.

PR #5324, opened 2020-05-14
Update network module to use node.db - Closes #5258

### LINKED ISSUE #5258: Update network module to use node.db
### Description

With the introduction of an embedded database system, a clear separation of concerns between blockchain, forger and node information is required for effective data management.

### Acceptance Criteria

- Update `framework/src/application/network/network.ts` 
  - Use `node.db` for `get` and `put` network info
- Update `framework` tests related to `network info` storage
- Update integration tests for `network`

### Additional Information

Database keys and data should be defined as below:

- `node.db` instance example

```
  - peers => peers[]
  - secret => Buffer
```

Depends on: #5241, #5242, #5243

Related issues: #5238

ishantiw:
### What was the problem?

This PR resolves #5258 

### How was it solved?

- Initialize `network.db` in application
- Pass `networkDB` to network constructor and use it

### How was it tested?

`npm t`

### DIFF
--- a/framework/src/application/network/network.ts
+++ b/framework/src/application/network/network.ts
@@ -14,7 +14,7 @@
 
 import { getRandomBytes } from '@liskhq/lisk-cryptography';
 import { KVStore } from '@liskhq/lisk-db';
-import * as liskP2p from '@liskhq/lisk-p2p';
+import * as liskP2P from '@liskhq/lisk-p2p';
 import { lookupPeersIPs } from './utils';
 import { Logger } from '../../types';
 import { InMemoryChannel } from '../../controller/channels';
@@ -39,7 +39,7 @@ const {
 		EVENT_MESSAGE_RECEIVED,
 		EVENT_BAN_PEER,
 	},
-} = liskP2p;
+} = liskP2P;
 
 const hasNamespaceReg = /:/;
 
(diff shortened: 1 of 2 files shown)


framework/src/application/network/network.ts
```diff
@@ -90,89 +91,87 @@ interface P2PRequest {
...
+		this._channel = channel;
+		this._logger = logger;
+		this._networkDB = networkDB;
+		this._secret = null;
 	}
 
 	public async bootstrap(): Promise<void> {
 		// Load peers from the database that were tried or connected the last time node was running
-		const previousPeersStr = await this.storage.entities.NetworkInfo.getKey(
+		const previousPeersStr = await this._networkDB.get<string>(
```
shuse2, line 116: This will fail with fresh database, so this should be in the `try-catch` and if it's `NotFoundError` then return `undefined`

framework/src/application/network/network.ts
```diff
@@ -90,89 +91,104 @@ interface P2PRequest {
...
 			// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
 			previousPeers = previousPeersStr ? JSON.parse(previousPeersStr) : [];
-		} catch (err) {
-			this.logger.error(
-				{ err: err as Error },
-				'Failed to parse JSON of previous peers.',
-			);
+		} catch (error) {
+			if (!(error instanceof NotFoundError)) {
+				this._logger.error(
```
shuse2, line 125: I think error is opposite. `if (!(error instanceof NotFoundError))` then it's parsing error or other database error.
In that case i think it make sense to log, but `else` case is not needed as it's expected error

PR #5326, opened 2020-05-14
Add boolean read/writer - Closes #5222 and #5223

### LINKED ISSUE #5222: Create boolean reader
### Description

Create function `boolean(buffer, schema) => value`

### Motivation

`lisk-codec` will include a set of `Readers` for each data type defined in the LIP, to that end
we need a function to read [booleans](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0027.md#encoding-booleans).

### Acceptance Criteria

- `boolean()` should accept a buffer representation of a boolean value
- `boolean()` returns a boolean value of true or false
- Full test coverage
- Performance stats are provided based on the reading of values of different sizes and measure the reading of 100, 500, 1000, 30000 booleans

### Additional Information

Related issues: #5200

### LINKED ISSUE #5223: Create boolean writer
### Description

Create function `boolean(object, schema) => buffer`

### Motivation

we need a function to write [booleans](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0027.md#encoding-booleans).

### Acceptance Criteria

- `boolean()` should accept a boolean value
- `boolean()` returns a buffer that can be passed through the boolean `reader` to get the original value
- Full test coverage
- Performance stats are provided based on the reading of values of different sizes and measure the reading of 100, 500, 1000, 30000 booleans

### Additional Information

Related issues: #5200

shuse2:
### What was the problem?

This PR resolves #5223 and resolves #5222 

### How was it solved?

- Adds boolean read/writer

### How was it tested?

- Added test and verify against other protobuf implementation


elements/lisk-codec/src/boolean.ts
```diff
@@ -0,0 +1,26 @@
...
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+/* eslint-disable no-bitwise */
+/* eslint-disable no-param-reassign */
+
+
+interface SchemaProperty {
+	readonly dataType: string;
```
ManuGowda, line 19: boolean?

shuse2: `dataType` itself is `string` and it contains `"boolean"`

elements/lisk-codec/src/boolean.ts
```diff
@@ -0,0 +1,26 @@
...
+}
+
+export const writeBoolean = (value: boolean, _schema: SchemaProperty): Buffer =>
+    value ? Buffer.from('01', 'hex') : Buffer.from('00', 'hex');
+
+export const readBoolean = (buffer: Buffer, _schema: SchemaProperty): boolean =>
+	buffer[0] === 0x01;
```
MaximeGagnebin, line 26: I think that here it is more common to have the function return `!(buf===0x00)`.
With the current formula, only `0x01` will return true.

shuse2: true! will change them

Issue #5261, opened 2020-05-02
Measure embedded database performance


ManuGowda:
### Description
With the introduction of EmbeddedDB, we need to measure the performance of `EmbeddedDB`.
Following  `EmbeddedDB` to consider for performance measure `LevelDB` and `RocksDB`.

### Motivation
- Optimized database storage size
- Better read/write and batch performance
- Ease of import/export of data

### Acceptance Criteria
- Measure the database storage size
- Random read
- Random write
- BatchWrite

### Additional Information
- Use 15kb, 50kb, 100kb and 150kb payload for measuring the read/write/batch operation
- Use digital ocean droplet with configuration
```
Ubuntu: 18.04.3 (LTS) x64
8 GB / 4 CPUs
160 GB SSD disk
```

Issue #5213, opened 2020-04-29
Simplify HTTP API module and separate from framework


shuse2:
### Description

Simplify the HTTP API so it only uses the `lisk-framework` actions and events to provide its endpoints. Separate the HTTP API so that it operates outside of `lisk-framework` and can be installed separately if the user wants.

### Motivation

- Remove the dependencies on database indexes, where it reduces blockchain performance
- Assign further responsibility for handling the HTTP API to `lisk-service`
- Change the main focus of the HTTP API from users to developers

#### Sprint 23
- Setup new plugin structure for HTTP API plugin #5248
- Update GET /api/blocks?height=x and GET /api/blocks/:id endpoints #5249
- Update GET /api/accounts/:address endpoint #5250
- Update GET /api/transactions/:id and POST /api/transactions endpoints #5251
- Add GET /api/node/info endpoint #5252
- Update GET /api/node/transactions?limit=xxx&offset=yyy endpoint #5253
- Update GET /api/peers?limit=xxx&offset=yyy&state=zzz endpoint #5254
- Update GET /api/delegates?limit=xxx&offset=yyy endpoint #5255
- Update GET /api/forgers endpoint #5256
- Define and update swagger.yml #5459
- Add functional test setup for http api plugin #5460
- Invalid BigInt check for uint32 #5554
- Add an action to expose all the delegates #5556

### Additional Information

- We keep using swagger for documentation purpose only, but we won't use swagger integration to avoid dependencies on its related libraries
- The open port should be configurable
- Allow whitelisting of IPs
- Allow making the port public
- Add flag to the core and register this plugin
```
--enable-plugin-http
--plugin-http-port (LISK_PLUGIN_HTTP_PORT)
  plugin must be enabled
  Specify open port for the http plugin
```

PR #5328, opened 2020-05-14
Remove storage and cache from components - Closes #5260

### LINKED ISSUE #5250: Update GET /api/accounts/:address endpoint
### Description

Add the API endpoint:

- `GET /api/accounts/:address`

- It should return decoded account
- Validate if address is base64 address if not `400` should be returned
- If address is not found `404` should be returned

### Motivation

The HTTP API should return accounts by address.

### Acceptance Criteria

- Add functional tests
- Register plugin to core, and check if it returns decoded data with address
- Check if invalid address returns 400 with error message

### Additional Information

Related issues: #5213

### LINKED ISSUE #5260: Remove postgresql related code and tests
### Description

With the introduction of an embedded database system, we need to remove any `postgresql` related code and tests.

### Acceptance Criteria

- Remove storage related information from `framework/src/application/storage`
- Update `framework` integration test to replace `postgresql` and use `EmbeddedDB`
  - Maintain tests for `processor` and `setup`

### Additional Information

- Replace `lisk-db` logic should be replaced with `postgresql->storage` logic 

Related issues: #5238

shuse2:
### What was the problem?

This PR resolves #5260 

### How was it solved?

- Remove cache and storage components in favor of #5261  and #5213 

### How was it tested?

All the tests should work as it is


framework/src/application/application.ts
```diff
@@ -165,8 +150,6 @@ export class Application {
 
 	private readonly _genesisBlock: GenesisBlockInstance;
 	private _migrations: { [key: string]: object };
```
ManuGowda, line 152: do we need them?

framework/src/modules/http_api/components/storage/entities/block.js
```diff
@@ -18,6 +18,7 @@ const path = require('path');
 const assert = require('assert');
 const {
 	entities: { Block: BlockEntity },
+	// eslint-disable-next-line
 } = require('../../../../../components/storage');
 
 const sqlFiles = {
```
ManuGowda, line 24: sql file?

framework/test/utils/random.js
```diff
@@ -34,7 +34,9 @@ const random = {};
 
 // Returns a random number between min (inclusive) and max (exclusive)
 random.number = function(min, max) {
```
ManuGowda, line 36: convert to TS?

framework/src/modules/http_api/components/storage/entities/block.js
shuse2: this is http_api, i just kept it until #5213

framework/test/utils/random.js
shuse2: ill do in different PR, it's already too big i think

framework/src/modules/http_api/components/storage/entities/block.js
nazarhussain: Don't get it. Since all migrations are already removed, storage object are removed. Even the Jenkins don't start the database then why to keep these migrations?

Jenkinsfile.sdk
```diff
@@ -3,45 +3,17 @@
...
-		timeout(1) {
-			waitUntil {
-				script {
-					def api_available = sh script: 'curl http://localhost:4000/api/node/constants >/dev/null', returnStatus: true
-					return (api_available == 0)
-				}
-			}
-		}
-	}
-}
```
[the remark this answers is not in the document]
shuse2: yea true, we will need functional test for the plugins, and the waiting functionality, we will probably use

Issue #5227, opened 2020-04-29
Create string reader


pablitovicente:
### Description

Create function `string(buffer, schema) => buffer`

### Motivation

`lisk-codec` will include a set of `Readers` for each data type defined in the LIP, to that end
we need a function to read [strings](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0027.md#encoding-strings-and-bytes).

### Acceptance Criteria

- `string()` should accept a chunk of bytes
- `string()` returns the original string
- Code should be optimised for speed as much as possible
- Full test coverage
- Performance stats are provided based on the reading of values of different sizes and measure the reading of 100, 500, 1000, 30000 strings

### Additional Information

Related issues: #5200

PR #5334, opened 2020-05-15
Create string reader writer in lisk-codec - Closes #5226 & #5227

### LINKED ISSUE #5226: Create string writer
### Description

Create function `string(object, schema) => buffer`

### Motivation

we need a function to write [strings](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0027.md#encoding-strings-and-bytes).

### Acceptance Criteria

- `string()` should accept a chunk of bytes
- `string()` returns a buffer that can be passed through the string `reader` to get the original value
- Full test coverage
- Performance stats are provided based on the reading of values of different sizes and measure the writing of 100, 500, 1000, 30000 strings

### Additional Information

Related issues: #5200

ishantiw:
### What was the problem?

This PR resolves #5226 and #5227

### How was it solved?

- Created function `string(object, schema) => buffer`
- Created function `string(buffer, schema) => buffer`
- Added unit tests
- Added benchmark

### How was it tested?

`npm t` inside `elements/lisk-codec`


elements/lisk-codec/src/string.ts
```diff
@@ -0,0 +1,28 @@
...
+ */
+import { writeVarInt } from './varint';
+
+interface SchemaProperty {
+	readonly dataType: string;
+}
+
+export const writeString = (value: string, _schema: SchemaProperty): Buffer =>
+	Buffer.concat([
+		Buffer.from('0a', 'hex'),
```
pablitovicente, line 22: do we need this?

elements/lisk-codec/src/string.ts
```diff
@@ -0,0 +1,28 @@
...
+		writeVarInt(value.length, { dataType: 'uint32' }),
```
pablitovicente, line 23: This needs to be the length of the buffer not the value itself (it might or not be the same)

elements/lisk-codec/src/string.ts
shuse2: I think this `key` is not used now in other writers

ishantiw: in order to produce hex as in protocol-spec, I had to add this line separator, for example, `lisk` -> `0a046c69736b` also @shuse2 mentioned it in the call to add it

shuse2: I think `0a` is the field number, and it changes based on the field number value

elements/lisk-codec/src/string.ts
```diff
@@ -0,0 +1,28 @@
...
+		Buffer.from(value, 'utf8'),
+	]);
+
+export const readString = (buffer: Buffer, _schema: SchemaProperty): string =>
+	buffer.toString('utf8', 2, buffer.length);
```
shuse2, line 28: so for this offset is `1` i think, if we dont include the field number

elements/lisk-codec/src/string.ts
pablitovicente: Oh yes the key is calculated outside then we might need to revisit the protocol-spec and not use them for now

elements/lisk-codec/src/string.ts
ishantiw: so if I use offset to 1 then `Checkout Lisk SDK!` -> `12436865636b6f7574204c69736b2053444b21` which is missing `0a`prefix as compared to protocol-spec.

shuse2: I think protocol spec is including the field number because if you read the whole schema, thats what should include, but for this implementation specifically, the field number should be ignored or skipped

elements/lisk-codec/src/string.ts
ishantiw: So is it that, `writeString = (key: string, value: string, _schema: SchemaProperty): Buffer`, it takes a key as a string and concatenates as a buffer?

elements/lisk-codec/src/string.ts
ishantiw: Okay, so the key's significance is from outside but not within the string writer. So in that case either I change the protocol spec or simply ignore the key part coming from the protocol-spec output

elements/lisk-codec/src/string.ts
shuse2: `key` will be assigned from outside, so 
```
writeString = (value: string, _schema: SchemaProperty): Buffer
```
should return the `<length specifier><actual buffer>`?

ishantiw: okay sounds good 👍

elements/lisk-codec/src/string.ts
pablitovicente: You can ignore the key part of concatenate in the test like simulating the whole encoder.  If you check the protocol-spec it's a complete message that's why the difference. So as you said you could remove the key or add it. Both work :)

shuse2: let's keep the protocol-spec as it is, so that we can test the most outer `encode` and `decode` function using the exact protocol-spec defined

ishantiw: Yeah I just pushed the changes and just ignoring the key part from protocol-spec

elements/lisk-codec/src/string.ts
```diff
@@ -0,0 +1,27 @@
...
+
+interface SchemaProperty {
+	readonly dataType: string;
+}
+
+export const writeString = (value: string, _schema: SchemaProperty): Buffer =>
+	Buffer.concat([
+		writeVarInt(Buffer.from(value, 'utf8').length, { dataType: 'uint32' }),
+		Buffer.from(value, 'utf8'),
+	]);
```
pablitovicente, line 24:
```suggestion
{
const buffer = Buffer.from(value, 'utf8');
Buffer.concat([
writeVarInt(buffer.length, {dataType: 'uint32'}),
buffer
])
return buffer
```

pablitovicente: This needs to be checked as I didn't run it but this way you only copy the string into the buffer once and then just read the length of that buffer

ishantiw: I think you meant 
```typescript
export const writeString = (value: string, _schema: SchemaProperty): Buffer => {
	const stringBuffer = Buffer.from(value, 'utf8');

	return Buffer.concat([
		writeVarInt(stringBuffer.length, { dataType: 'uint32' }),
		stringBuffer,
	]);
};
```

pablitovicente: I think yes? You're not using the extra var I think here right?

ishantiw: I'm using `stringBuffer` here to create the buffer to be returned as you suggested except that you were returning the `stringBuffer` itself in the writer, I'm returning the concatenated buffer in the end.

pablitovicente: hard to write in github comment box :)

PR #5336, opened 2020-05-15
Update logger to output to the correct path - Closes #5329

### LINKED ISSUE #5329: Logs are not saved under the rootPath
### Expected behavior
Logs should be saved under the `<rootPath>/<label>/logs`,

### Actual behavior
logs is saved in different place than the above folder

### Steps to reproduce
Start test_app with default setting and check `~/.lisk/` folder

development

### Additional Information
- Logger should be converted to typescript
- Logger should be moved to `application` and update the config

shuse2:
### What was the problem?

This PR resolves #5329 

### How was it solved?

- Update to use `dirs` defined in the application
- Move logger from `components` to `application`

### How was it tested?

- All the tests are updated
- Check logs are written in the `rootPath/label` when using the `test_app`


framework/src/modules/http_api/http_api.js
```diff
@@ -14,6 +14,7 @@
 
 'use strict';
 
+// eslint-disable-next-line
 const { createLoggerComponent } = require('../../components/logger');
```
nazarhussain, line 18: I think this logger is removed.

framework/src/application/application.ts
```diff
@@ -109,6 +109,11 @@ export interface ApplicationConfig {
 		readonly defaultPassword?: string;
 	};
 	readonly network: NetworkConfig;
+	readonly logger: {
```
ManuGowda, line 112: export `LoggerInput` and use `Partial< LoggerInput>` ?

framework/src/application/application.ts
```diff
@@ -338,10 +339,15 @@ export class Application {
...
-		await this._forgerDB.close();
-		await this._networkDB.close();
+		try {
+			await this._network.cleanup();
+			await this._node.cleanup();
+			await this._blockchainDB.close();
+			await this._forgerDB.close();
+			await this._networkDB.close();
+		} catch (error) {
+			this.logger.fatal({ err: error as Error }, 'failed to shutdown');
```
ManuGowda, line 349:
```suggestion
			this.logger.fatal({ err: error as Error }, 'Failed to shutdown');
```

framework/src/application/application.ts
ManuGowda: ?

PR #5337, opened 2020-05-15
Create base32 address format and checksum - Closes #5177

### LINKED ISSUE #5177: Support UI encoded Base32 address format
### Description

On the protocol level, the new address format uses 160 bits of the SHA-256 hash of the public key. On the user interface level, a 30-bit checksum generated by the BCH code proposed in BIP 173 is added and a customised Base32 encoding is used.

### Motivation

Refer https://github.com/LiskHQ/lips/blob/master/proposals/lip-0018.md#motivation

### Acceptance Criteria

- Create new functions to generate Base32 address
- Create the functions to convert Base32 to binary address and vice versa
- Create the functions to validate Base32 address
- Use lsk prefix with the new address
- Use the Base32 checksum function provided by the research team
- Test case to validate the v0 address format
- Test case to validate the Base32 address format of length 41.
- Test case to validate the substring of the first 3 characters of Base32 address equals `lsk`
- Test case to verify that every character in Base32 address substring in last 38 characters is an element of the used Base32 alphabet. Uppercase letters are not allowed.
- Test case to verify verifyChecksum

### Additional information

- Keep the old address generation function and tests for backward compatibility

Related issues: #5167

mitsuaki-u:
### What was the problem?

To increase collision resistance of addresses, provide checksum error detection and discontinue the necessity of address registration transaction, we use 160-bit hash of the public key as a binary address. 

On the user interface level a Base32 encoding is used since the overall bit length of the hash value plus checksum is large.

This PR resolves #5177


elements/lisk-cryptography/src/keys.ts
```diff
@@ -81,3 +85,93 @@ export const getAddressFromPrivateKey = (privateKey: string): string => {
...
+export const validateBase32Address = (address: string): boolean => {
+	if (address.length !== BASE32_ADDRESS_LENGTH) {
+		throw new Error(
+			'Address length does not match requirements. Expected 41 characters.',
+		);
+	}
+
+	const prefix = address.substring(0, 3);
+
+	if (prefix !== 'lsk') {
```
shuse2, line 157: default can be `lsk`, but to use it in other cases, prefix can be passed from the argument as done in above function.
`export const validateBase32Address = (address: string, prefix = 'lsk'): boolean`

elements/lisk-cryptography/src/convert.ts
```diff
@@ -23,6 +23,43 @@ import { hash } from './hash';
...
+	fromBits: number,
+	toBits: number,
+): number[] => {
+	// eslint-disable-next-line no-bitwise
+	const maxValue = (1 << toBits) - 1;
+	let accumulator = 0;
+	let bits = 0;
+	const result = [];
+	// eslint-disable-next-line
+	for (let p = 0; p < uintArray.length; ++p) {
```
ManuGowda, line 39:
```suggestion
	for (let p = 0; p < uintArray.length; p += 1) {
```
and remove the above `// eslint-disable-next-line`

elements/lisk-cryptography/src/keys.ts
```diff
@@ -81,3 +85,93 @@ export const getAddressFromPrivateKey = (privateKey: string): string => {
 
 	return getAddressFromPublicKey(publicKey);
 };
+
+const GENERATOR = [0x3b6a57b2, 0x26508e6d, 0x1ea119fa, 0x3d4233dd, 0x2a1462b3];
+
+const polymod = (uint5Array: number[]): number => {
```
ManuGowda, line 91: can we add this link for reference, otherwise its hard to follow without comments. https://github.com/LiskHQ/lips/blob/master/proposals/lip-0018.md#creating-checksum

elements/lisk-validator/src/validation.ts
```diff
@@ -190,18 +190,19 @@ export const validatePublicKeys = (
 	publicKeys.every(validatePublicKey) &&
 	validatePublicKeysForDuplicates(publicKeys);
 
-const ADDRESS_BYTE_LENGTH = 20;
-export const validateAddress = (address: string): boolean => {
-	if (!isHexString(address)) {
-		throw new Error('Address is not in hex format');
-	}
+const ADDRESS_LENGTH = 40;
```
ManuGowda, line 193: address length must be 41?

mitsuaki-u: `41` is for base32 address. this is `binaryAddress`

elements/lisk-cryptography/src/convert.ts
mitsuaki-u: I can change to `p += 1` but there is another `lint` issue, so the disable-lint will remain regardless.

ManuGowda: actually you could change this if you want `for (const p of uintArray) {` as your doing the same in `polymod`

mitsuaki-u: There is a subtle difference between `for of` and the original code. If I change to `for of` the resulting bytes are not the same and the function will not work. That is the reason I am keeping it.

elements/lisk-cryptography/src/keys.ts
```diff
@@ -81,3 +85,98 @@ export const getAddressFromPrivateKey = (privateKey: string): string => {
...
 };
+
+const GENERATOR = [0x3b6a57b2, 0x26508e6d, 0x1ea119fa, 0x3d4233dd, 0x2a1462b3];
+
+// See for details: https://github.com/LiskHQ/lips/blob/master/proposals/lip-0018.md
+const polymod = (uint5Array: number[]): number => {
+	let chk = 1;
+	for (const value of uint5Array) {
+		// eslint-disable-next-line no-bitwise
+		const top = chk >> 25;
```
nazarhussain, line 96: Please add comments explaining the statement, will be very helpful when we read the code in a month or two. 

> Get 25 bits from the right side

elements/lisk-cryptography/src/keys.ts
```diff
@@ -81,3 +85,98 @@ export const getAddressFromPrivateKey = (privateKey: string): string => {
...
+		// eslint-disable-next-line no-bitwise
+		chk = ((chk & 0x1ffffff) << 5) ^ value;
```
nazarhussain, line 98: Here as well, one-line comment explaining this expression...

elements/lisk-cryptography/src/keys.ts
```diff
@@ -81,3 +85,98 @@ export const getAddressFromPrivateKey = (privateKey: string): string => {
...
+			}
+		}
+	}
+
+	return chk;
+};
+
+export const getBinaryAddressFromPublicKey = (publicKey: string): Buffer => {
+	const publicKeyBuffer = Buffer.from(publicKey, 'hex');
+	return hash(publicKeyBuffer).slice(0, 20);
```
nazarhussain, line 113: Please move this constant `20` as `BINARY_ADDRESS_LENGTH` to constants.

elements/lisk-cryptography/src/keys.ts
mitsuaki-u: I feel that the above comment link is sufficient.

Issue #5335, opened 2020-05-15
Update all protocol-specs to follow new address format


nazarhussain:
### Expected behavior

All protocol-specs should use all address related attributes as 20 bytes hex format string. 

### Actual behavior

Currently, protocol-specs are using old address format `***L` which is now removed from the SDK.  

### Acceptance Crieteria 

- There should be no `***L` format address in any of output JSON files. 
- Remove unused protocol-specs
  - block_processing_xxx
  - transaction_signing

### Additional Information
- Use `lisk-codec` for creating protocol specs

nazarhussain:
@shuse2 Assigning this task to you, as you can do it one go while updating the specs.

PR #5338, opened 2020-05-18
Update delegate registration to save and use username and address in blockchain.db - Closes #5259

### LINKED ISSUE #5259: Update delegate registration to save and use username and address in blockchain.db
### Description

Update delegate registration to save the username and address in chain state, and use the information from the `lisk-dpos` library.

### Acceptance Criteria

- Update `lisk-transactions->delegate_transactions` registration
  - Save `username` and `address` in `blockchain.db`
- Update `DPoS` library to use `usernames` and `address` from `chain state` info
- Update unit tests in `lisk-transactions` and `lisk-dpos` library

### Additional Information

- Saving username and address example

```
chain:usernames => { registeredDelegates: { username, address }[] }
```

Related issues: #5238

nazarhussain:
### What was the problem?

This PR resolves #5259

### How was it solved?

Updated delegate registration transaction to add `chain:usernames` keys to `blockchain.db`. Also updated `lisk-dpos` to use it to fetch delegates. 

### How was it tested?

- Added unit tests for lisk-dpos

**NOTE: Can't update unit test for transaction due to https://github.com/LiskHQ/lisk-sdk/issues/5335**

### DIFF
--- a/elements/lisk-chain/src/data_access/storage.ts
+++ b/elements/lisk-chain/src/data_access/storage.ts
@@ -22,7 +22,12 @@ import {
 import { TransactionJSON } from '@liskhq/lisk-transactions';
 import { getAddressFromPublicKey } from '@liskhq/lisk-cryptography';
 
-import { AccountJSON, BlockJSON, BlockHeaderJSON } from '../types';
+import {
+	AccountJSON,
+	BlockJSON,
+	BlockHeaderJSON,
+	ChainUsernames,
+} from '../types';
 
 import {
 	DB_KEY_BLOCKS_ID,
@@ -33,6 +38,7 @@ import {
 	DB_KEY_ACCOUNTS_ADDRESS,
 	DB_KEY_CHAIN_STATE,
 	DB_KEY_CONSENSUS_STATE,
+	DB_KEY_CHAIN_USERNAMES,
 } from './constants';
 import { StateStore } from '../state_store';
 
@@ -341,28 +347,15 @@ export class Storage {
 		return accounts;
 	}
 
-	// TODO: Remove this with issue #5259
 	public async getDelegates(): Promise<AccountJSON[]> {
-		const stream = this._db.createReadStream({
-			gte: getFirstPrefix(DB_KEY_ACCOUNTS_ADDRESS),
-			lte: getLastPrefix(DB_KEY_ACCOUNTS_ADDRESS),
-		});
-		const accounts = await new Promise<AccountJSON[]>((resolve, reject) => {
-			const accountJSONs: AccountJSON[] = [];
-			stream
-				.on('data', ({ value }) => {
-					const { username } = value as AccountJSON;
-					if (username) {
-						accountJSONs.push(value);
-					}
-				})
-				.on('error', error => {
-					reject(error);
-				})
-				.on('end', () => {
-					resolve(accountJSONs);
-				});
-		});
+		// Data format for the registered delegates
+		// chain:usernames => { registeredDelegates: { username, address }[] }
+		const { registeredDelegates }: ChainUsernames = await this._db.get(
+			DB_KEY_CHAIN_USERNAMES,
+		);
+		const accounts = await this.getAccountsByAddress(
+			registeredDelegates.map(delegate => delegate.address),
+		);
 		accounts.sort((a, b) => {
 			const diff = BigInt(b.totalVotesReceived) - BigInt(a.totalVotesReceived);
 			if (diff > BigInt(0)) {
--- a/elements/lisk-chain/src/data_access/storage.ts
+++ b/elements/lisk-chain/src/data_access/storage.ts
@@ -22,12 +22,7 @@ import {
 import { TransactionJSON } from '@liskhq/lisk-transactions';
 import { getAddressFromPublicKey } from '@liskhq/lisk-cryptography';
 
-import {
-	AccountJSON,
-	BlockJSON,
-	BlockHeaderJSON,
-	ChainUsernames,
-} from '../types';
+import { AccountJSON, BlockJSON, BlockHeaderJSON } from '../types';
 
 import {
 	DB_KEY_BLOCKS_ID,
@@ -38,7 +33,6 @@ import {
 	DB_KEY_ACCOUNTS_ADDRESS,
 	DB_KEY_CHAIN_STATE,
 	DB_KEY_CONSENSUS_STATE,
-	DB_KEY_CHAIN_USERNAMES,
 } from './constants';
 import { StateStore } from '../state_store';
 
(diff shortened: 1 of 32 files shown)


elements/lisk-transactions/src/10_delegate_transaction.ts
```diff
@@ -92,8 +100,31 @@ export class DelegateTransaction extends BaseTransaction {
...
+			DELEGATE_USERNAME_CHAIN_KEY,
+		);
+		const registeredDelegates = registeredDelegatesStr
+			? (JSON.parse(registeredDelegatesStr) as ChainUsernames)
+			: { registeredDelegates: [] };
+		const usernameExists = registeredDelegates.registeredDelegates.find(
+			delegate => delegate.username === this.asset.username,
+		);
+		const updatedRegisteredDelegates = usernameExists
+			? { ...registeredDelegates }
```
shuse2, line 115: destructuring not needed here as it was not changed

elements/lisk-transactions/src/10_delegate_transaction.ts
```diff
@@ -92,8 +100,31 @@ export class DelegateTransaction extends BaseTransaction {
...
+			: {
+					registeredDelegates: [
+						...registeredDelegates.registeredDelegates,
```
shuse2, line 118: instead of spreading, push should be fine.
Also, it's better to be sorted by `username` or `address` to make it deterministic

elements/lisk-chain/src/data_access/storage.ts
```diff
@@ -341,28 +347,15 @@ export class Storage {
...
-					}
-				})
-				.on('error', error => {
-					reject(error);
-				})
-				.on('end', () => {
-					resolve(accountJSONs);
-				});
-		});
+		// Data format for the registered delegates
```
shuse2, line 351: this `getDelegates` should be moved to `dpos` where it's used.
it has temporal logic 
```
// This list is before executing the current block in process
const originalDelegates = await this.chain.dataAccess.getDelegates();
```
this should be changed to below function, using state store instead.
then this `This list is before executing the current block in process` is no longer true, that it always get the consistent information in the state store

nazarhussain: But we don't have `db` access in the dpos, we access it through `chain` object. Don't get how can we move that function outside `lisk-chain`

shuse2: we have access to `stateStore.chain.get` and `stateStore.account.get`, so it doesn't need the direct access to `db` object

elements/lisk-dpos/src/delegates_list.ts
```diff
@@ -253,26 +255,21 @@ export class DelegatesList {
...
-		const delegates = [...Object.values(updatedAccountsMap)];
+
+		// Data format for the registered delegates
+		// chain:usernames => { registeredDelegates: { username, address }[] }
+		const usernamesStr = await stateStore.chain.get(
+			CHAIN_STATE_DELEGATE_USERNAMES,
+		);
+		const usernames = usernamesStr
+			? (JSON.parse(usernamesStr) as ChainStateUsernames)
+			: { registeredDelegates: [] };
```
ishantiw, line 266: I don't know if it's really necessary but maybe we can wrap this with try/catch and throw a better error for logging at the framework level

elements/lisk-transactions/src/10_delegate_transaction.ts
```diff
@@ -92,10 +100,27 @@ export class DelegateTransaction extends BaseTransaction {
...
-		const usernameExists = store.account.find(
-			(account: Account) => account.username === this.asset.username,
+		// Data format for the registered delegates
+		// chain:usernames => { registeredDelegates: { username, address }[] }
+		const usernamesStr = await store.chain.get(DELEGATE_USERNAME_CHAIN_KEY);
+		const usernames = usernamesStr
+			? (JSON.parse(usernamesStr) as ChainUsernames)
+			: { registeredDelegates: [] };
+		const usernameExists = usernames.registeredDelegates.find(
+			delegate => delegate.username === this.asset.username,
```
ishantiw, line 110: also here, try/catch

elements/lisk-dpos/src/delegates_list.ts
nazarhussain: We don't throw errors normally in the transactions. And its just initializing default value for the attribute, which we do a similar way across the application.

elements/lisk-transactions/src/10_delegate_transaction.ts
nazarhussain: I don't understand why we need a try/catch for a simple searcing in the array. This code is just checking if a particular array element exists. The finder function itself doesn't throw an error if the object not found.

elements/lisk-dpos/src/delegates_list.ts
ishantiw: I get it for transactions but I see that we are throwing error while getting a delegate list or verify forger and other places in this file. But just highlighting that if JSON.parse will fail then it will be hard to debug

nazarhussain: The function itself is `async`. I think if there will be any synchronous error it will reject the promise and top-level code will catch the error.

elements/lisk-transactions/src/10_delegate_transaction.ts
ishantiw: Oh I was referring to `JSON.parse`, not the find function

elements/lisk-dpos/src/delegates_list.ts
ishantiw: also here I was highlighting `JSON.parse` instead of async task

elements/lisk-transactions/src/10_delegate_transaction.ts
nazarhussain: I folllowe the pattern we used on other places as well: 

https://github.com/LiskHQ/lisk-sdk/blob/359700b09e0305f811e007b9281e7ff76ed0a23f/elements/lisk-dpos/src/delegates_list.ts#L68

https://github.com/LiskHQ/lisk-sdk/blob/359700b09e0305f811e007b9281e7ff76ed0a23f/elements/lisk-dpos/src/delegates_list.ts#L89

These JSON objects are not user-provided, so we trust that what we generated previously will be parsed. And in general, we rely on top-level code (framework) above libraries that will catch any error if occurred and will log.

elements/lisk-dpos/src/delegates_list.ts
ishantiw: Almost everywhere in our code, I see `JSON.parse` wrapped around `try/catch` but if you think it will be fine and will be handled properly in case of parsing error then you can ignore it. I just wanted to highlight it.

nazarhussain: @ishantiw If there is any JSON which is populated by user input then we should try/catch it but in general we don't. See my comment. 

https://github.com/LiskHQ/lisk-sdk/pull/5338#discussion_r427166427

ishantiw: I see your point. Maybe here its safe to assume that the string will be parsed without error

Issue #5225, opened 2020-04-29
Create bytes reader


pablitovicente:
### Description

Create function `bytes(buffer, schema) => buffer`

### Motivation

`lisk-codec` will include a set of `Readers` for each data type defined in the LIP, to that end
we need a function to read [bytes](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0027.md#encoding-strings-and-bytes).

### Acceptance Criteria

- `bytes()` should accept a chunk of bytes
- `bytes()` returns the original bytes
- Code should be optimized for speed as much as possible
- Full test coverage
- Performance stats are provided based on the reading of values of different sizes and measure the reading of 100, 500, 1000, 30000 bytes

### Additional Information

Related issues: #5200

PR #5339, opened 2020-05-18
Create bytes reader writer - Closes #5224 & #5225

### LINKED ISSUE #5224: Create bytes writer
### Description

Create function `bytes(object, schema) => buffer`

### Motivation

we need a function to write [bytes](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0027.md#encoding-strings-and-bytes).

### Acceptance Criteria

- `bytes()` should accept a chunk of bytes
- `bytes()` returns a buffer that can be passed through the bytes `reader` to get the original value
- Full test coverage
- Performance stats are provided based on the reading of values of different sizes and measure the writing of 100, 500, 1000, 30000 bytes

### Additional Information

Related issues: #5200

ishantiw:
### What was the problem?

This PR resolves #5224 & #5225 

### How was it solved?

- Creater writer `bytesWriter = (buffer, schema) => buffer`
- Created reader `bytesReader = (buffer, schema) => buffer`
- Added unit tests
- Added benchmark

### How was it tested?

Run `npm t` under `elements/lisk-codec`

### DIFF
--- a/elements/lisk-codec/src/bytes.ts
+++ b/elements/lisk-codec/src/bytes.ts
@@ -18,10 +18,7 @@ interface SchemaProperty {
 }
 
 export const writeBytes = (bytes: Buffer, _schema: SchemaProperty): Buffer =>
-	Buffer.concat([
-		writeVarInt(Buffer.from(bytes).length, { dataType: 'uint32' }),
-		bytes,
-	]);
+	Buffer.concat([writeVarInt(bytes.length, { dataType: 'uint32' }), bytes]);
 
 export const readBytes = (buffer: Buffer, _schema: SchemaProperty): Buffer =>
 	buffer.slice(1, buffer.length);
--- a/elements/lisk-codec/src/bytes.ts
+++ b/elements/lisk-codec/src/bytes.ts
@@ -1,0 +1,27 @@
+/*
+ * Copyright © 2020 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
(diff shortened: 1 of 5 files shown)


elements/lisk-codec/src/bytes.ts
```diff
@@ -0,0 +1,27 @@
...
+ */
+import { writeVarInt } from './varint';
+
+interface SchemaProperty {
+	readonly dataType: string;
+}
+
+export const writeBytes = (bytes: Buffer, _schema: SchemaProperty): Buffer =>
+	Buffer.concat([
+		writeVarInt(Buffer.from(bytes).length, { dataType: 'uint32' }),
```
pablitovicente, line 22: Bytes is already a buffer so no need for this

elements/lisk-codec/src/bytes.ts
```diff
@@ -0,0 +1,24 @@
...
+
+interface SchemaProperty {
+	readonly dataType: string;
+}
+
+export const writeBytes = (bytes: Buffer, _schema: SchemaProperty): Buffer =>
+	Buffer.concat([writeVarInt(bytes.length, { dataType: 'uint32' }), bytes]);
+
+export const readBytes = (buffer: Buffer, _schema: SchemaProperty): Buffer =>
+	buffer.slice(1, buffer.length);
```
pablitovicente, line 24: This returns the whole buffer but I understand we're gonna to change the interface for readers. So approving.

Issue #5303, opened 2020-05-04
Benchmark encode/decode performance


ManuGowda:
### Description

Measure the performance of encoding and decoding using `lisk-codec`

### Motivation

- Block processing optimization

### Acceptance Criteria

- Performance stats are provided based on the encoding of the smallest and biggest objects possible in the Lisk network
  - Use transactions (transfer) and full block for measuring the performance
- Performance stats are provided based on the decoding of the smallest and biggest binary messages possible in the Lisk network
- Measure `Buffer.concat` for encoding with and without `totalLength`

PR #5341, opened 2020-05-18
Create function to read keys - Closes #5233

### LINKED ISSUE #5233: Create function to read keys
### Description

Create `readKey(bytes, schema) => object`

### Motivation

Messages are encoded as key-value pairs we need a function to read keys from a binary message based on the [LIP](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0027.md#keys).

### Acceptance Criteria

- `readKey()` is able to read keys to provide information on how to read a binary message as per the LIP
- Full tests

### Additional Information

Related issues: #5200

ManuGowda:
### What was the problem?

This PR resolves #5233 

### How was it solved?

- Implemented `readKey` function to get `fieldNumber, wireType`
### How was it tested?

- Added unit test
- Added benchmark test for performace


elements/lisk-codec/benchmark/buffer_concat.js
```diff
@@ -0,0 +1,29 @@
...
+
+const { Suite } = require('benchmark');
+const { randomBytes } = require('crypto');
+const suite = new Suite();
+const buf1 = randomBytes(1024);
+const buf2 = randomBytes(1024);
+
+suite
+  .add('Buffer.concat', () => {
+    Buffer.concat([buf1, buf2]);
```
shuse2, line 24: I think this is not needed? if we gonna keep this, probably nice to test with more elements and also with totalLength and without totalLength for the second parameter on the `Buffer.concat`

ManuGowda: ok should i remove this now?

shuse2: Either way it's fine

ManuGowda:
> `keys.ts` and `readkey.spec.ts` look good to me.
> I have to admit, I didn't really understand some of the other files (like `buffer_concat.js `)

the `buffer_concat` came out of discussion we were having about performance impacts, its not relevant to this PR, so i removed it from this PR and added to #5303

PR #5344, opened 2020-05-19
Update scalar reader and writer

### LINKED ISSUE #5343: Update reader/writer interface to support the decode
### Description

- Remove `schema` input from scalar writers, and the input should be simply value
- Add `offset` parameter to the scalar readers
- Return `value` and `size` as tuple from scalar readers

### Motivation

- Decode needs the size to be return to see how many bytes are read
- Decode needs offset input to decide where to read from
- Encode does decides which scalar writer to use, so schema is unnecessary input

### Acceptance Criteria

- All tests needs to be updated

shuse2:
### What was the problem?

This PR resolves #5343 

### How was it solved?

- Updated the interface
- Clean up the name for varint to follow other interfaces
- Add size and offset for reader

### How was it tested?

- Update tests to include offset
- Updated unit test to follow the interface

### DIFF
--- a/elements/lisk-codec/src/bytes.ts
+++ b/elements/lisk-codec/src/bytes.ts
@@ -11,14 +11,12 @@
  *
  * Removal or modification of this copyright notice is prohibited.
  */
-import { writeVarInt } from './varint';
+import { writeUInt32, readUInt32 } from './varint';
 
-interface SchemaProperty {
-	readonly dataType: string;
+export const writeBytes = (bytes: Buffer): Buffer =>
+	Buffer.concat([writeUInt32(bytes.length), bytes]);
+
+export const readBytes = (buffer: Buffer, offset: number): [Buffer, number] => {
+	const [byteLength, keySize] = readUInt32(buffer, offset);
+	return [buffer.subarray(offset + keySize, offset + keySize + byteLength), byteLength + keySize];
 }
-
-export const writeBytes = (bytes: Buffer, _schema: SchemaProperty): Buffer =>
-	Buffer.concat([writeVarInt(bytes.length, { dataType: 'uint32' }), bytes]);
-
-export const readBytes = (buffer: Buffer, _schema: SchemaProperty): Buffer =>
-	buffer.slice(1, buffer.length);
--- a/elements/lisk-codec/src/string.ts
+++ b/elements/lisk-codec/src/string.ts
@@ -11,20 +11,15 @@
  *
  * Removal or modification of this copyright notice is prohibited.
  */
-import { writeVarInt } from './varint';
+import { readBytes, writeBytes } from './bytes';
 
-interface SchemaProperty {
-	readonly dataType: string;
-}
-
-export const writeString = (value: string, _schema: SchemaProperty): Buffer => {
+export const writeString = (value: string): Buffer => {
 	const stringBuffer = Buffer.from(value, 'utf8');
 
-	return Buffer.concat([
-		writeVarInt(stringBuffer.length, { dataType: 'uint32' }),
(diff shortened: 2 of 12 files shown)


elements/lisk-codec/src/bytes.ts
```diff
@@ -11,14 +11,12 @@
...
+import { writeUInt32, readUInt32 } from './varint';
 
-interface SchemaProperty {
-	readonly dataType: string;
-}
-
-export const writeBytes = (bytes: Buffer, _schema: SchemaProperty): Buffer =>
-	Buffer.concat([writeVarInt(bytes.length, { dataType: 'uint32' }), bytes]);
+export const writeBytes = (bytes: Buffer): Buffer =>
+	Buffer.concat([writeUInt32(bytes.length), bytes]);
```
pablitovicente, line 17: this function should hint at it being a varint without having to look at the import guess it's ok in this small file

elements/lisk-codec/src/bytes.ts
```diff
@@ -11,14 +11,12 @@
...
 
-export const readBytes = (buffer: Buffer, _schema: SchemaProperty): Buffer =>
-	buffer.slice(1, buffer.length);
+export const readBytes = (buffer: Buffer, offset: number): [Buffer, number] => {
+	const [byteLength, keySize] = readUInt32(buffer, offset);
```
pablitovicente, line 20: why you read all the bytes as varints? bytes are encoded as bytes and no transformation should happen to them

elements/lisk-codec/src/string.ts
```diff
@@ -11,20 +11,15 @@
...
-		writeVarInt(stringBuffer.length, { dataType: 'uint32' }),
-		stringBuffer,
-	]);
+	return writeBytes(stringBuffer);
 };
 
-export const readString = (buffer: Buffer, _schema: SchemaProperty): string =>
-	buffer.toString('utf8', 1, buffer.length);
+export const readString = (buffer: Buffer, offset: number): [string, number] => {
+	const [value, size] = readBytes(buffer, offset);
```
pablitovicente, line 23: why you need readBytes here if you can just directly read it with Buffer native functions?

elements/lisk-codec/src/bytes.ts
shuse2: bytes is wire type 2, so the first X bytes are the length of the bytes that comes after.
Therefore, it needs to read the varint first, then take out the bytes that was defined in the bytes

elements/lisk-codec/src/string.ts
shuse2: same as above, it needs to read the first X bytes to get the length, and then value is obtained, which is same as bytes. Only difference with bytes is how the result Buffer is read

elements/lisk-codec/src/bytes.ts
pablitovicente: Oh so then this is both reading a chunk and also decoding the key? If that's the case maybe we should rename the "readers" to something more meaningful as they will no longer be reading (extracting) a chunk from the main buffer but instead they would be decoding key and slicing the buffer.

elements/lisk-codec/src/string.ts
pablitovicente: I think having this separated would be easier for other readers but guess it's ok like this too.

elements/lisk-codec/src/bytes.ts
ishantiw: maybe we could add some comments on top of this read function as the name itself is a bit abstract and hard to follow whats going on inside

PR #5347, opened 2020-05-19
Add merkle tree root calculation - Closes #5340

### LINKED ISSUE #5340: Implement constructor for Merkle Tree Class
### Description

Given a genesis block, constructor of Merkle Tree class should populate the tree. 

### Motivation

In order to populate the Merkle Tree to calculate `transactionRoot`, it must be given a genesis block. 

### Acceptance Criteria

- Add all test cases for `transactionRoot` validation

shuse2:
### What was the problem?

This PR resolves #5340 

### How was it solved?

- Add merkle root calculation in constructor

### How was it tested?

- Add test using fixture created using the function written by research team


elements/lisk-tree/src/constant.ts
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+export const EMPTY_HASH = Buffer.alloc(0);
```
ricott1, line 15: Doesn't buffer.alloc(0) just create an empty buffer? Because empty_hash should be the hash of that. `EMPTYHASH=SHA256()`. We could either import `hash` from lisk-cryptography or hard code the value (with a comment).
```suggestion
export const EMPTY_HASH = Buffer.from(e3b0c44298fc1c149afbf4c8996fb92427ae41e4649b934ca495991b7852b855, "hex");
```

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -11,21 +11,63 @@
...
+		}
+		const leafHashes = new Array<Buffer>(initValues.length);
+		for (let i = 0; i < initValues.length; i += 1) {
+			const value = Buffer.concat(
+				[LEAF_PREFIX, initValues[i]],
+				LEAF_PREFIX.length + initValues[i].length,
+			);
+			const key = hash(value);
+			leafHashes[i] = key;
+			this._data[key.toString('binary')] = value;
```
ricott1, line 40: I see that we are storing the data with the prefix included. Is there a specific reason to do this? I think is fine, but you store an extra byte and maybe is not necessary (not big deal anyway)

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -11,21 +11,63 @@
...
+			const pairBranches = layer.length >>> 1;
+			const remainder = layer.length & 1;
+			for (let j = 0; j < pairBranches; j += 1) {
+				const index = j * 2;
+				const value = Buffer.concat(
+					[BRANCH_PREFIX, layer[index], layer[index + 1]],
+					BRANCH_PREFIX.length + layer[index].length + layer[index + 1].length,
+				);
+				const key = hash(value);
+				this._data[key.toString('binary')] = value;
```
ricott1, line 55: Same here, adding the prefix to the value tells you if it's a branch or leaf node, but I'm not sure how useful that is

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -11,21 +11,63 @@
...
+	private _data: { [key: string]: Buffer } = {};
+
+	public constructor(initValues: Buffer[] = []) {
+		this._dataLength = initValues.length;
+		if (this._dataLength === 0) {
+			this._root = EMPTY_HASH;
+			return;
+		}
+		const leafHashes = new Array<Buffer>(initValues.length);
+		for (let i = 0; i < initValues.length; i += 1) {
```
ManuGowda, line 33:
```suggestion
		for (let i = 0; i < this._dataLength; i += 1) {
```

elements/lisk-tree/src/constant.ts
nazarhussain: BTW I suggest not to use magic values. Instead, have `hash(Buffer.alloc(0))

shuse2:
@ricott1 
I think as you mentioned, position of the path make more sense, but it's probably more related to the proof generation.
So this prefix will be updated or removed in that PR as discussed

elements/lisk-tree/benchmark/merkle_root.js
```diff
@@ -0,0 +1,36 @@
...
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
```
ManuGowda, line 14: maybe add the stats here as comment for reference?

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -11,21 +11,63 @@
...
+		if (this._dataLength === 0) {
+			this._root = EMPTY_HASH;
+			return;
+		}
+		const leafHashes = new Array<Buffer>(this._dataLength);
+		for (let i = 0; i < this._dataLength; i += 1) {
+			const value = Buffer.concat(
+				[LEAF_PREFIX, initValues[i]],
+				LEAF_PREFIX.length + initValues[i].length,
+			);
```
nazarhussain, line 37: @shuse2 Does not it be: 

```
const value = Buffer.concat(
   LEAF_PREFIX
   initValues[i],
   LEAF_PREFIX.length + initValues[i].length,
);
```

shuse2: Buffer.concat is `concat(value: Buffer[], totalLength: number)`
so the above is correct, and total length is just there to make it efficient

nazarhussain: Ok got it. But I don't think providing length which is an optional parameter makes it efficient. It's just there if one wants to truncate the result buffer for a particular length.

shuse2: > If totalLength is not provided, it is calculated from the Buffer instances in list by adding their lengths.
It's there not to recalculate the total size of buffer while concatenating, so the performance will be better if you know the size ahead

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -11,21 +11,63 @@
...
+			const key = hash(value);
+			leafHashes[i] = key;
+			this._data[key.toString('binary')] = value;
+		}
+		const height = this._height;
```
nazarhussain, line 42: @shuse2 I don't see any declaration of private data member `_height`

shuse2: `height` is the getter in below =) it's calculated value.

nazarhussain: Then I would suggest not have a private getter which is actually calculating some value. Getters are mostly used to return private data members, which one not to expose to the public interface. 

Better to know `calcHeight`, so while reading code we know this will calculate rather returning some static value. 

Or much better option would have a particular setter for `dataLength` which also computes the height and save its static value.

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -11,21 +11,63 @@
...
+		let layer = leafHashes;
+		for (let i = 0; i < height - 1; i += 1) {
+			const currentLayerHashes: Buffer[] = [];
+			const pairBranches = layer.length >>> 1;
```
nazarhussain, line 46: Is this `layer.length >>> 1` not same as `layer.length / 2`? Then use simple version which is more readable.

shuse2: it's integer calculation of `/ 2`, but since it's float it will have different value unless we use `floor`.
But in general, it's faster and accurate to do the bit shifting, because it's integer calculation.

PR #5351, opened 2020-05-22
Add writing array - Closes #5237

### LINKED ISSUE #5237: Create function to write array
### Description

Create function `array(value: array, schema) => buffer`

### Motivation

we need a function to write [arrays](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0027.md#encoding-arrays).

### Acceptance Criteria

- `array()` can write valid buffers for arrays of varints
- `array()` can write valid buffers for arrays of booleans
- `array()` can write valid buffers for arrays of strings
- `array()` can write valid buffers for arrays of bytes
- `array()` encodes any object recursively 
- Full tests
- Performance stats are provided based on the encoding of the smallest and biggest arrays possible in any object that could exist in the Lisk network

### Additional Information

- Check if this could be done as part of #5217

Related issues: #5200

shuse2:
### What was the problem?

This PR resolves #5237 

### How was it solved?

Added array handling for the encoder

### How was it tested?

Tested against the array protocol-spec, but not added to this PR, as the target branch is still changing the protocol specs, and integrating them

### DIFF
--- a/elements/lisk-codec/src/collection.ts
+++ b/elements/lisk-codec/src/collection.ts
@@ -13,7 +13,7 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
-import { GenericObject, CompiledSchemasArray } from './types';
+import { GenericObject, CompiledSchemasArray, CompiledSchema } from './types';
 
 import { writeSInt32, writeSInt64, writeUInt32, writeUInt64 } from './varint';
 import { writeString } from './string';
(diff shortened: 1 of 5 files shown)


protocol-specs/package-lock.json
```diff
@@ -5,22 +5,28 @@
 	"requires": true,
```
pablitovicente, line 5: Why did this file changed but no package.json? Was this outdated vs package.json?

shuse2: yes i think protocol-spec package json was outdated

PR #5352, opened 2020-05-22
Implement diff function - Closes #5275

### LINKED ISSUE #5275: Implement diff calculation function in the lisk-chain
### Description

- Implement `diff(initial: Buffer, final: Buffer) => Diff` function using `Myers algorithm` where Diff is `Array<[string, number]>`
- Implement `undo(value: Buffer, diff: Diff) => Buffer` function
- Benchmark using the account state

### Motivation

In order to provide automatic undo functionality, diff should be calculated.

### Acceptance Criteria

- Benchmark is created in the `benchmark` folder in root
- Add unit tests

### Additional Information

Related issues: #5232

ishantiw:
### What was the problem?

This PR resolves #5275

### How was it solved?

- Implement diff function based on Myers algorithm
- Test cases for diff using Account states
- Benchmark for `calculateDiff` and `undo` function

### How was it tested?

Run `npm run test diff.spec` under `elements/lisk-chain`


elements/lisk-chain/src/diff.ts
```diff
@@ -0,0 +1,139 @@
...
+type HistoryType = [string?, number?];
+
+interface Frontier {
+	[key: string]: {
+		x: number;
+		history: HistoryType[];
+	};
+}
+
+const _diff = (initialBuffer: Buffer, finalBuffer: Buffer): HistoryType[] => {
```
shuse2, line 24: this is not method, so i think we dont need `_`?

elements/lisk-chain/src/diff.ts
```diff
@@ -0,0 +1,139 @@
...
+	const one = (idx: number): number => idx - 1;
+	const initial = Array.from(initialBuffer);
```
shuse2, line 26: do we need to convert this to array? same for below

elements/lisk-chain/src/diff.ts
```diff
@@ -0,0 +1,139 @@
...
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+type HistoryType = [string?, number?];
```
shuse2, line 15: is it optional for string and number?

elements/lisk-chain/src/diff.ts
```diff
@@ -0,0 +1,139 @@
...
+			let x: number;
+			if (goDown) {
+				history = frontier[k + 1].history;
+				x = frontier[k + 1].x;
+			} else {
+				history = frontier[k - 1].history;
+				x = frontier[k - 1].x + 1;
+			}
+
+			history = [...history];
```
shuse2, line 55: what is this doing?

elements/lisk-chain/src/diff.ts
```diff
@@ -0,0 +1,139 @@
...
+				y += 1;
+				history.push(['=', initial[one(x)]]);
+			}
+
+			if (x >= initialBytesLength && y >= finalBytesLength) {
+				return history.splice(1);
+			}
+
+			if (x >= initialBytesLength && x + y === d + 2 * initialBytesLength) {
+				const fDiff = _diff(Buffer.from([]), Buffer.from(final.splice(y)));
```
shuse2, line 79: do you need to pass in the new instance here? if not `final.subarray` would be better.

elements/lisk-chain/src/diff.ts
```diff
@@ -0,0 +1,139 @@
...
+
+				return history.splice(1).concat(fDiff);
+			}
+
+			if (y >= finalBytesLength && x + y === d + 2 * finalBytesLength) {
+				const fDiff = _diff(Buffer.from(initial.splice(x)), Buffer.from([]));
```
shuse2, line 85: same as above

elements/lisk-chain/src/diff.ts
```diff
@@ -0,0 +1,139 @@
...
+		}
+	}
+	if (count > 0) {
+		reducedDiff.push(['=', count]);
+	}
+	return reducedDiff as HistoryType[];
+};
+
+export const undo = (finalBuffer: Buffer, diff: HistoryType[]): Buffer => {
+	const finalBytes = [...Array.from(finalBuffer)];
```
shuse2, line 121: do we need to change this to `array`?

elements/lisk-chain/benchmark/calculateDiff.js
```diff
@@ -0,0 +1,63 @@
...
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+const { Suite } = require('benchmark');
+const { calculateDiff, undo } = require('../dist-node/diff');
+
+const suite = new Suite();
+
```
ManuGowda, line 19: rename file to follow underscore pattern, `calculate_diff.js`

elements/lisk-chain/src/diff.ts
```diff
@@ -0,0 +1,139 @@
...
+	const res = [];
+
+	for (const d of diff.reverse()) {
```
shuse2, line 124:
```suggestion
	for (const [op, diff] of diff.reverse()) {
```

elements/lisk-chain/benchmark/calculateDiff.js
```diff
@@ -0,0 +1,63 @@
...
+const diff = calculateDiff(buffer1, buffer2);
+
+/**
+ * calculateDiff x 2,759 ops/sec ±0.74% (87 runs sampled)
+ * undo x 6,376 ops/sec ±0.49% (91 runs sampled)
+ */
+suite
+    .add('calculateDiff', () => {
+        calculateDiff(buffer1, buffer2);
```
ManuGowda, line 55: to get better benchmarking result use multi-sig account or others which can give better insight?

elements/lisk-chain/src/diff.ts
```diff
@@ -0,0 +1,139 @@
...
+		if (d[0] === '=') {
+			for (let i = 0; i < (d[1] as number); i += 1) {
+				res.unshift(finalBytes.pop() as number);
```
shuse2, line 127: `unshift` is very slow, so prefer to do in different way. (it re-indexes the array)

elements/lisk-chain/src/diff.ts
ManuGowda, line 79: maybe define `const emptyBuffer = Buffer.from([])` and use it here ?

elements/lisk-chain/src/diff.ts
```diff
@@ -0,0 +1,139 @@
...
+
+	if (initialBytesLength === 0) {
+		return final.map(b => ['+', b]);
+	}
+
+	if (finalBytesLength === 0) {
+		return initial.map(b => ['-', b]);
+	}
+
+	const frontier: Frontier = { 1: { x: 0, history: [] } };
```
ManuGowda, line 39: add some comments here.. its hard to follow

elements/lisk-chain/src/diff.ts
ManuGowda, line 85: replace with `emptyBuffer`

elements/lisk-chain/src/diff.ts
ishantiw: this is optional because when initialize here we have a blank history
`const frontier: Frontier = { 1: { x: 0, history: [] } };`

elements/lisk-chain/src/diff.ts
```diff
@@ -0,0 +1,139 @@
...
+		if (d[0] === '=') {
+			for (let i = 0; i < (d[1] as number); i += 1) {
+				res.unshift(finalBytes.pop() as number);
+			}
+		} else if (d[0] === '-') {
+			res.unshift(d[1] as number);
+		} else if (d[0] === '+') {
+			finalBytes.pop();
+		} else {
+			throw new Error('Diff contains non expected symbol');
```
ManuGowda, line 134:
```suggestion
			throw new Error('Diff contains unexpected symbol');
```

elements/lisk-chain/src/diff.ts
shuse2: I think empty array should be fine even without the option?

elements/lisk-chain/src/diff.ts
ishantiw: we want to preserve the old history because it might be used by other steps below

elements/lisk-chain/src/diff.ts
ishantiw: strange, earlier it was giving me lint error and now it doesn't, I'll remove optional

elements/lisk-chain/src/diff.ts
shuse2: but u are assigning to the same variable, so anyway it's overwritten...?😕

ishantiw: So the history is a an array of arrays and that is stored in `frontier` object, so I think when we are looping through we don't want to change `frontier[k].history`, if we directly use `history = frontier[k].history`, and modify history then it will also change arrays in `frontier[k].history`.
To simply we could also do,

```typescript
if (goDown) {
				history = [...frontier[k + 1].history];
				x = frontier[k + 1].x;
			} else {
				history = [...frontier[k - 1].history];
				x = frontier[k - 1].x + 1;
			}
```

elements/lisk-chain/src/diff.ts
ishantiw: yeah,  we don't need it, earlier there were some operations related to arrays so I didn't change, but still I have to make below changes,

```typescript
if (initialBytesLength === 0) {
		return (Array.from(final)).map(b => ['+', b]);
	}

	if (finalBytesLength === 0) {
		return Array.from(initial).map(b => ['-', b]);
	}
```

elements/lisk-chain/src/diff.ts
MaximeGagnebin: is it better to move the whole chunk at the same time? 
something like:
`res = concat( finalBytes.splice( -d[1] ), res )`

and no loop on `i`.

elements/lisk-chain/src/diff.ts
shuse2: if `b` is number, you can just use map directly on Buffer. the iterator will return number (uint8)

elements/lisk-chain/src/diff.ts
shuse2: make sense! yea i think it would be more clear if you do it on spot

elements/lisk-chain/src/diff.ts
ishantiw: yes, I removed that loop and took the whole chunk and it works fine

elements/lisk-chain/src/diff.ts
ishantiw: we want to return Array type and when I use the map on buffer it tries to modify the value of each element by something like `['+', 2]` and outputs a Buffer `Buffer <00, 00, 00, 00, 00>`

elements/lisk-chain/benchmark/calculate_diff.js
```diff
@@ -0,0 +1,63 @@
...
+const buffer2 = Buffer.concat([
+    randomBytes4,
+    randomBytes1,
+    randomBytes2,
+]);
+
+const diff = calculateDiff(buffer1, buffer2);
+
+/**
+ * calculateDiff x 2,759 ops/sec ±0.74% (87 runs sampled)
```
shuse2, line 50: On my local, it seems like this is the result now =)
```
calculateDiff x 4,602 ops/sec ±0.84% (87 runs sampled)
undo x 128,998 ops/sec ±1.15% (85 runs sampled)
```

Let's create another issue to optimise `calculateDiff` function. It will be called 150 * 2 +- 10 for the full block, so it will be good if we can optimise them.
Also, when i tried with around 500 bytes for the buffer, the result degrades pretty fast.
```
calculateDiff x 13.54 ops/sec ±5.34% (38 runs sampled)
```

ishantiw: Created issue #5364 regarding this.

PR #5355, opened 2020-05-22
Implement Merkle Tree append method - Closes #5205

### LINKED ISSUE #5205: Implement Merkle Tree append method
### Description

Adding new data blocks to the merkle tree updates the value of at most Log(N) branch nodes. We define the append path to be an array of bytes containing the branch-node values necessary to compute the updated tree. Adding new data blocks updates the append path as well. Each layer with an odd number of nodes contains a node which is part of the append path. Therefore, the append path contains at most Log(N) elements, and updating the tree requires to calculate at most Log(N) hashes.

### Motivation

We append elements to the Merkle Tree as new data sets arrive.

### Acceptance Criteria

To append a new element we need to do the following:

- Add the leaf hash of the element to the base layer
- Update the affected branch nodes in the tree
- Update the append path

### Additional information
Related issues: #5204

nazarhussain:
### What was the problem?

To append a new element we need to do the following:

1. Add the leaf hash of the element to the base layer;
2. Build the appendPath from the root downwards
3. Concatenate hashes of appendPath to form new branch nodes and new root

This PR resolves #5205

### How was it solved?

- [x] Updated `getNode` to return full node data
- [x] Generate `appendPath` traversing tree
- [x] Generate new `root`

### How was it tested?

Run unit tests for merkle tree.


nazarhussain:
@shuse2 @mitsuaki-u I did refactor constructor of the merkle tree to make it more readable. The important changes did:

1. Renamed `dataLength` to `width`, which is a more appropriate name for what it presents, already discussed with Alessandro on it. 
2. Added `layer index` and `node index` to the tree node value to identify node position in tree which is also discussed earlier. 
3. Also prefixed `branch prefix` otherwise it was getting difficult to identify leaf node compared to a branch node on branch `0`
4. Added `toString()` method to the tree class, which is helpfull to see the full tree structure. 

After refactoring the tests all tests passing against protocol-specs. I hope you will the refactoring more readable. 

Also added the `append` feature, which traverses the tree from `root` and put required nodes to a `stack` and then `pop` them back to create a new node and update all nodes upto root. 

The append is also working against the data fixtures we had. There is one edge case that is failing if we have `6` nodes in the tree and we append `7th`, that I am still debugging.

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -18,62 +18,155 @@ import { hash } from '@liskhq/lisk-cryptography';
...
+		let type: NodeType;
+		if (this._root.compare(nodeHash) === 0) {
+			type = NodeType.ROOT;
+		} else if (isLeaf(value)) {
+			type = NodeType.LEAF;
+		} else {
+			type = NodeType.BRANCH;
+		}
+		const layerIndex = type === NodeType.LEAF ? 0 : value.readInt8(1);
+		const nodeIndex =
```
ricott1, line 87: I'm confused, doesn't this return 0 for leaf nodes? If that the case i think it is wrong.
As a general comment, why do we treat leaf and branches differently? (see also previous comment for layer index)

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -18,62 +18,155 @@ import { hash } from '@liskhq/lisk-cryptography';
...
+			type === NodeType.BRANCH
+				? value.readBigInt64BE(LAYER_INDEX_SIZE + 1)
+				: BigInt(0);
+		const rightHash =
```
ricott1, line 91: I think a root can also have no children, in which case this should be wrong. I would eliminate the root nodetype and treat the root as either leaf or branch (because it could be either of the two)

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -18,62 +18,155 @@ import { hash } from '@liskhq/lisk-cryptography';
...
+			type !== NodeType.LEAF
+				? value.slice(-1 * NODE_HASH_SIZE)
+				: Buffer.alloc(0);
+		const leftHash =
```
ricott1, line 95: same as previous comment

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -18,62 +18,155 @@ import { hash } from '@liskhq/lisk-cryptography';
...
+		}
+
+		const appendPath: NodeInfo[] = [];
+		let currentNode = this.getNode(this._root);
+
+		// Create the appendPath:
+		// We start from the root layer and traverse each layer down the tree on the right side
+		// eslint-disable-next-line
+		while (true) {
+			if (this._width === 2 ** (this._getHeight() - 1)) {
```
ricott1, line 126: I would move this check to outside the while loop, like to line 121

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -18,62 +18,155 @@ import { hash } from '@liskhq/lisk-cryptography';
...
+		const appendData = this._generateLeaf(value);
+		const appendNode = this.getNode(appendData.hash);
+		appendPath.push(this.getNode(appendNode.hash));
+		// Loop through appendPath from the base layer
+		// Generate new branch nodes and push to appendPath
+		// Last element remaining is new root
+		while (appendPath.length > 1) {
+			const rightNodeInfo = appendPath.pop();
+			const leftNodeInfo = appendPath.pop();
+			// FIXME: Add correct nodeIndex:  get left node index + 1
```
ricott1, line 159: is there still something to be fixed here?

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -18,62 +18,155 @@ import { hash } from '@liskhq/lisk-cryptography';
...
-	private _data: { [key: string]: Buffer } = {};
+	private _width = 0;
+
+	// Object holds data in format { [hash]: value }
+	private _hashToBuffer: { [key: string]: Buffer } = {};
 
 	public constructor(initValues: Buffer[] = []) {
-		this._dataLength = initValues.length;
-		if (this._dataLength === 0) {
+		if (initValues.length === 0) {
```
ricott1, line 57: Even for an empty tree, I would insert the emptyHash in `_hashToBuffer`, for consistency

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -83,12 +176,159 @@ export class MerkleTree {
 	}
 
 	public clear(): void {
-		this._dataLength = 0;
-		this._data = {};
+		this._width = 0;
+		this._hashToBuffer = {};
```
ricott1, line 180: Even for an empty tree, I would insert the emptyHash in `_hashToBuffer`, for consistency

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -83,12 +176,159 @@ export class MerkleTree {
...
+		this._hashToBuffer[leafHash.toString('binary')] = leafValue;
+		this._width += 1;
+
+		return {
+			value: leafValue,
+			hash: hash(leafValue),
+		};
+	}
+
+	private _generateNode(
```
ricott1, line 216: Why not `generateBranch`?

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -18,62 +18,155 @@ import { hash } from '@liskhq/lisk-cryptography';
...
+	readonly hash: Buffer;
+	readonly value: Buffer;
+	readonly leftHash: Buffer;
+	readonly rightHash: Buffer;
+	readonly layerIndex: number;
+	readonly nodeIndex: bigint;
+}
+
+const isLeaf = (value: Buffer): boolean =>
+	value[0] === LEAF_PREFIX[0] && value.compare(Buffer.alloc(0)) !== 0;
```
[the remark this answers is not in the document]
mitsuaki-u: `value[0]` would not give an error for an empty buffer since it would be zero in both cases, but I can switch the order.

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -18,62 +18,155 @@ import { hash } from '@liskhq/lisk-cryptography';
...
+		// eslint-disable-next-line
+		if (!value) {
+			throw new Error(
+				`Hash does not exist in merkle tree: ${nodeHash.toString('binary')}`,
+			);
+		}
+
+		let type: NodeType;
+		if (this._root.compare(nodeHash) === 0) {
+			type = NodeType.ROOT;
```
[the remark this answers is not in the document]
mitsuaki-u: It is not required but for informational purposes when using `getData` function it would be nice to see which node is the root easily.

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -18,62 +18,155 @@ import { hash } from '@liskhq/lisk-cryptography';
...
+		} else if (isLeaf(value)) {
+			type = NodeType.LEAF;
+		} else {
+			type = NodeType.BRANCH;
+		}
+		const layerIndex = type === NodeType.LEAF ? 0 : value.readInt8(1);
```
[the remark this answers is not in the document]
mitsuaki-u: this check must remain otherwise the `layerIndex` is not assigned properly.

elements/lisk-tree/src/merkle_tree.ts
mitsuaki-u: You are correct it returns `0` for nodes. There is no way to determine nodeIndex from within this function with just the value Buffer, since it is not stored for leaf nodes. Therefore for now I opted to simply store 0 for now.

mitsuaki-u: Currently leaf node doesn't store nodeIndex, therefore you cannot determine it from within this function. The nodeIndex will properly be implemented in the next issue which tackles verification.

elements/lisk-tree/src/merkle_tree.ts
mitsuaki-u: Do you mean the case in which the tree is just one node? In that case this function will never be called.

elements/lisk-tree/src/merkle_tree.ts
mitsuaki-u: I thought of naming it `generateBranch`, but it also generates possible root node. It seems semantically awkward to call a root node a branch.

nazarhussain:
@mitsuaki-u One more thing. Do we really need to maintain `_build` now? as we can just use `append`. Its around 40-45 lines of logic merely duplicated, so will be harder to maintain. 

It may be little performance overhead while using `append` instead of `_build`, what's the use of case in our application to initialise the tree with larger dataset?

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -18,62 +18,155 @@ import { hash } from '@liskhq/lisk-cryptography';
...
+
+	// Object holds data in format { [hash]: value }
+	private _hashToBuffer: { [key: string]: Buffer } = {};
 
 	public constructor(initValues: Buffer[] = []) {
-		this._dataLength = initValues.length;
-		if (this._dataLength === 0) {
+		if (initValues.length === 0) {
 			this._root = EMPTY_HASH;
+			this._hashToBuffer[this._root.toString('binary')] = Buffer.alloc(0);
```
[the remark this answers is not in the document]
mitsuaki-u: To me it seems more clear what is happening if we keep it shown explicitly.

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -18,62 +18,155 @@ import { hash } from '@liskhq/lisk-cryptography';
...
+		const nodeIndex =
+			type === NodeType.BRANCH
+				? value.readBigInt64BE(LAYER_INDEX_SIZE + 1)
+				: BigInt(0);
+		const rightHash =
+			type !== NodeType.LEAF
+				? value.slice(-1 * NODE_HASH_SIZE)
+				: Buffer.alloc(0);
+		const leftHash =
+			type !== NodeType.LEAF
```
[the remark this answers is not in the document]
mitsuaki-u: Because the node can also be `NodeType.Root`. It is shorter to check if it's not `NodeType.LEAF`.

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -83,12 +176,159 @@ export class MerkleTree {
...
+		this._width = 0;
 		this._root = EMPTY_HASH;
+		this._hashToBuffer = { [this._root.toString('2')]: Buffer.alloc(0) };
+	}
+
+	public getData(): object[] {
+		return Object.keys(this._hashToBuffer).map(key =>
+			this.getNode(Buffer.from(key, 'binary')),
+		);
+	}
```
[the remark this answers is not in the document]
mitsuaki-u: I would prefer to keep it. It is very useful to see the contents of the tree for debugging purposes.

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -18,62 +18,155 @@ import { hash } from '@liskhq/lisk-cryptography';
...
+
+		return {
+			type,
+			hash: nodeHash,
+			value,
+			layerIndex,
+			nodeIndex,
+			rightHash,
+			leftHash,
+		};
```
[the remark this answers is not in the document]
mitsuaki-u: The suggestion requires more lines of code since it has multiple returns. I think it is better to return once at the end.

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -18,62 +18,153 @@ import { hash } from '@liskhq/lisk-cryptography';
...
+				// if layer below is odd numbered, push left child
+				d = this._width >> (currentLayer - 1);
+				if (d % 2 === 1) {
+					const leftNode = this.getNode(currentNode.leftHash);
+					appendPath.push(leftNode);
+				}
+
+				// go to right child
+				currentNode = this.getNode(currentNode.rightHash);
+			}
```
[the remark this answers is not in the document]
mitsuaki-u: This is also a good way to do it. However I feel both approaches near in readability. @shuse2 ?

mitsuaki-u:
> @mitsuaki-u One more thing. Do we really need to maintain `_build` now? as we can just use `append`. Its around 40-45 lines of logic merely duplicated, so will be harder to maintain.
> 
> It may be little performance overhead while using `append` instead of `_build`, what's the use of case in our application to initialise the tree with larger dataset?

After discussion it was decided to keep both `append` and `build` since both have separate use-cases.

elements/lisk-tree/src/merkle_tree.ts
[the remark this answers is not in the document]
shuse2: I would prefer not to use the string representation of binary. Also, not array functions as much as possible in the low level library like this. I feel the current implementation as readable as this.
and layer binary is incorrect in some case that needs prepending zeros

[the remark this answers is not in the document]
nazarhussain: If layer binary is incorrect then its an issue with the LIP algorithm, I don't think it's true. And difference is just do this tree width to binary conversion once and not repeat in the loop.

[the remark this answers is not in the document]
mitsuaki-u: I prefer to keep the approaches which clearly shows # of nodes per layer, although layer binary also works.

elements/lisk-tree/src/merkle_tree.ts
mitsuaki-u: It can?

mitsuaki-u:
> I would remove the nodetype Root and assign a root either as a leaf or branch, because it could be either. 

I prefer to keep it as the benefit of seeing the full root node info and not just the hash can be useful.  The suggested change doesn't seem to simplify by any meaningful degree, you would just switch from LEAF to BRANCH. 
 
>this code will have to change once verification is implmented

If you are referring to nodeIndex yes that issue will be tackled with verification PR.

elements/lisk-tree/src/merkle_tree.ts
mitsuaki-u: `generateLeaf` will only be called during initial `build` or otherwise when `append` is called.

elements/lisk-tree/src/merkle_tree.ts
mitsuaki-u: Ah sorry, yes it would be called if you wanted to append to a one node tree.

elements/lisk-tree/src/merkle_tree.ts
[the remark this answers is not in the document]
ricott1: I know it works both ways the same, but maybe we should either check `type === NodeType.BRANCH` (like for the right child) or  `type !== NodeType.LEAF` like here

PR #5359, opened 2020-05-25
Update block schema to include transactionRoot - Closes #5210

### LINKED ISSUE #5210: Update block schema to include transactionRoot
### Description

The block schema must include a new `transactionRoot` property.

### Motivation

In order for blocks to be properly validated, the schema must be updated. 

### Acceptance Criteria

- Update block schema
- Replace the usage of `payloadHash` with `transactionRoot`
- Fix all related tests

### Additional information

Related issues: #5204

ManuGowda:
### What was the problem?

This PR resolves #5210 

### How was it solved?
- Replace the usage of payloadHash with transactionRoot
- Update block schema
- Fix all related tests

### How was it tested?

-  Updated all related tests


elements/lisk-chain/src/chain.ts
```diff
@@ -347,7 +347,7 @@ export class Chain {
 			throw invalidTransactionResponse.errors;
 		}
 
-		validatePayload(block, this.constants.maxPayloadLength);
+		validateTransactionRoot(block, this.constants.maxPayloadLength);
```
shuse2, line 350: method name and the input doesn't match here

ManuGowda: `validateBlockProperties`?

shuse2: validatePayload was accurate? i think `BlockProperties` includes other things

PR #5361, opened 2020-05-26
Update db interface - Closes #5357

### LINKED ISSUE #5357: Update lisk-db to use value as Buffer

shuse2:
### What was the problem?

This PR resolves #5357 

### How was it solved?

- Update DB interface to be (key: string, value: Buffer)
- Update the usage

### How was it tested?

- All the usages are updated and integration tests are passing
- Run application and it should run (after resetting the database

### DIFF
--- a/elements/lisk-chain/src/block_reward.ts
+++ b/elements/lisk-chain/src/block_reward.ts
@@ -160,7 +160,7 @@ export const applyFeeAndRewards = async (
 	generator.balance += givenFee > 0 ? givenFee : BigInt(0);
 	const totalFeeBurntBuffer = await stateStore.chain.get(CHAIN_STATE_BURNT_FEE);
 	let totalFeeBurnt = BigInt(
-		totalFeeBurntBuffer ? JSON.parse(totalFeeBurntBuffer.toString('utf-8')) : 0,
+		totalFeeBurntBuffer ? JSON.parse(totalFeeBurntBuffer.toString('utf8')) : 0,
 	);
 	totalFeeBurnt += givenFee > 0 ? totalMinFee : BigInt(0);
 
@@ -168,7 +168,7 @@ export const applyFeeAndRewards = async (
 	stateStore.account.set(generatorAddress, generator);
 	stateStore.chain.set(
 		CHAIN_STATE_BURNT_FEE,
-		Buffer.from(JSON.stringify(totalFeeBurnt.toString()), 'utf-8'),
+		Buffer.from(JSON.stringify(totalFeeBurnt.toString()), 'utf8'),
 	);
 };
 
@@ -193,7 +193,7 @@ export const undoFeeAndRewards = async (
 	generator.balance -= totalFee - totalMinFee;
 	const totalFeeBurntBuffer = await stateStore.chain.get(CHAIN_STATE_BURNT_FEE);
 	let totalFeeBurnt = BigInt(
-		totalFeeBurntBuffer ? JSON.parse(totalFeeBurntBuffer.toString('utf-8')) : 0,
+		totalFeeBurntBuffer ? JSON.parse(totalFeeBurntBuffer.toString('utf8')) : 0,
 	);
 	totalFeeBurnt -= totalMinFee;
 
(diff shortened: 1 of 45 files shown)


elements/lisk-chain/src/block_reward.ts
```diff
@@ -158,13 +158,18 @@ export const applyFeeAndRewards = async (
 	// This is necessary only for genesis block case, where total fee is 0, which is invalid
 	// Also, genesis block cannot be reverted
 	generator.balance += givenFee > 0 ? givenFee : BigInt(0);
-	const totalFeeBurntStr = await stateStore.chain.get(CHAIN_STATE_BURNT_FEE);
-	let totalFeeBurnt = BigInt(totalFeeBurntStr ?? 0);
+	const totalFeeBurntBuffer = await stateStore.chain.get(CHAIN_STATE_BURNT_FEE);
+	let totalFeeBurnt = BigInt(
+		totalFeeBurntBuffer ? JSON.parse(totalFeeBurntBuffer.toString('utf-8')) : 0,
```
nazarhussain, line 163: I am not sure if `utf-8` is the correct value, as the documentation suggests `utf8` always. 

https://nodejs.org/api/buffer.html#buffer_buf_tostring_encoding_start_end

```suggestion
		totalFeeBurntBuffer ? JSON.parse(totalFeeBurntBuffer.toString('utf8')) : 0,
```

elements/lisk-chain/src/block_reward.ts
```diff
@@ -158,13 +158,18 @@ export const applyFeeAndRewards = async (
...
+	);
 	totalFeeBurnt += givenFee > 0 ? totalMinFee : BigInt(0);
 
 	// Update state store
 	stateStore.account.set(generatorAddress, generator);
-	stateStore.chain.set(CHAIN_STATE_BURNT_FEE, totalFeeBurnt.toString());
+	stateStore.chain.set(
+		CHAIN_STATE_BURNT_FEE,
+		Buffer.from(JSON.stringify(totalFeeBurnt.toString()), 'utf-8'),
```
nazarhussain, line 171:
```suggestion
		Buffer.from(JSON.stringify(totalFeeBurnt.toString()), 'utf8'),
```

elements/lisk-chain/src/block_reward.ts
```diff
@@ -158,13 +158,18 @@ export const applyFeeAndRewards = async (
...
+	);
 	totalFeeBurnt += givenFee > 0 ? totalMinFee : BigInt(0);
 
 	// Update state store
 	stateStore.account.set(generatorAddress, generator);
-	stateStore.chain.set(CHAIN_STATE_BURNT_FEE, totalFeeBurnt.toString());
+	stateStore.chain.set(
+		CHAIN_STATE_BURNT_FEE,
+		Buffer.from(JSON.stringify(totalFeeBurnt.toString()), 'utf8'),
+	);
```
nazarhussain, line 172: Can we use some utility function to convert back and forth from the buffer, rather repeating the same code.

elements/lisk-chain/src/data_access/storage.ts
```diff
@@ -47,21 +52,17 @@ export class Storage {
...
 		arrayOfBlockIds: ReadonlyArray<string>,
 	): Promise<BlockHeaderJSON[]> {
 		const blocks = [];
 		for (const id of arrayOfBlockIds) {
-			const block = await this._db.get<BlockHeaderJSON>(
-				`${DB_KEY_BLOCKS_ID}:${id}`,
-			);
-			blocks.push(block);
+			const block = await this._db.get(`${DB_KEY_BLOCKS_ID}:${id}`);
+			blocks.push(JSON.parse(block.toString('utf8')) as BlockHeaderJSON);
```
nazarhussain, line 65: I am concerned about the performance of `JSON.parse` on such an important function here. If such calls always require to convert the buffer to string and then parse the string. Then it does not make sense storing the buffer in the first place.

elements/lisk-chain/src/block_reward.ts
shuse2: This needs to be updated using the lisk-codec

elements/lisk-chain/src/data_access/storage.ts
shuse2: Same as above, this will be replaced by the lisk-codec

PR #5362, opened 2020-05-26
Add extra protocol-spec and tests for lisk-codec - Closes #5360

### LINKED ISSUE #5360: Add extra protocol-spec and tests for lisk-codec
### Description

Add more protocol-specs and tests using them to lisk-codec

### Motivation

Although we had protocol-specs for lisk-codec before implementing it we need to add more cases to make sure we cover several edge cases

### Acceptance Criteria

- Add protocol-spec/tests for objects with several nested objects
- Add protocol-spec/tests for objects with arrays with complex nested objects

pablitovicente:
### What was the problem?

This PR resolves #5360

### How was it solved?

- By refactoring protocol specs
- By adding new protocol specs
- By adding new tests to lisk-codec

### How was it tested?

Unit Tests


pablitovicente:
@ManuGowda feel free to suggest more test scenarios and could you select the rest of reviewers please?

elements/lisk-codec/src/collection.ts
```diff
@@ -177,76 +175,57 @@ export const readArray = (message: Buffer, offset: number, compiledSchema: Compi
...
-					result.push({});
 					// Add default value
+					result.push({});
 					continue;
 				}
 				// readObject returns Next offset, not index used
 				const [res, nextOffset] = readObject(message, index, typeSchema);
 				result.push(res);
-				index += nextOffset;
+				index = nextOffset;
```
ishantiw, line 208: Not sure I completely understand but I hope this is an intentional change. `index` will always be `nextOffset`

shuse2: This is intentional.
`readObject` and `readArray` return the nextOffset, but the scalar value returns the byte size it used.
Therefore it's correct to assign to the `index`

PR #5363, opened 2020-05-26
Update cryptography to use buffer instead of string - Closes #5358

### LINKED ISSUE #5358: Update lisk-cryptography to treat address, public/private key as buffer
### Description

Remove the usage of `address`, public key` and `private key` as string, and treat consistently as Buffer everywhere

### Motivation

With https://github.com/LiskHQ/lips/blob/master/proposals/lip-0030.md, address, and public key are treated as `bytes` in our case corresponds to `Buffer`

### Acceptance Criteria

All test needs to be updated

### Additional Information
- Related to #5264

shuse2:
### What was the problem?

This PR resolves #5358 

### How was it solved?

- Update public key, private key, address to be buffer

### How was it tested?

- All the lisk-cryptography tests are updated

### DIFF
--- a/elements/lisk-cryptography/src/types.ts
+++ b/elements/lisk-cryptography/src/types.ts
@@ -1,0 +1,19 @@
+/*
+ * Copyright © 2019 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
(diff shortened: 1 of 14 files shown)


elements/lisk-cryptography/src/types.ts
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright © 2019 Lisk Foundation
```
ManuGowda, line 2:
```suggestion
 * Copyright © 2020 Lisk Foundation
```

PR #5365, opened 2020-05-27
Optimise calculateDiff performance for diff calculation - Closes #5364

### LINKED ISSUE #5364: Improve diff function performance
### Expected behavior

The diff function to calculate the difference between states should be optimized

### Actual behavior

Below we have current performance of `calculateDiff` and `undo` function
```
calculateDiff x 5,179 ops/sec ±0.56% (89 runs sampled)
undo x 137,778 ops/sec ±1.13% (84 runs sampled)
```
### Steps to reproduce

`node elements/lisk-chain/benchmark/calculate_diff.js`

`5.0`

ishantiw:
### What was the problem?

This PR resolves #5364 

### How was it solved?

- Added `diffCommonPrefix` and `diffCommonSuffix` functions to calculate prefix and suffix of two buffers
- Modified `calculateDiff` to use `diffCommonPrefix` and `diffCommonSuffix` to optimize diff calculation
- Improved diff benchmark to `calculateDiff x 102,134 ops/sec ±0.99% (84 runs sampled)`

### How was it tested?

`node elements/lisk-chain/benchmark/calculate_diff.js`


elements/lisk-chain/src/diff.ts
```diff
@@ -130,15 +130,104 @@ const diffAlgo = (initial: Buffer, final: Buffer): HistoryType[] => {
...
+
+	return pointerMid;
+};
+
+/**
+ * This function returns the length of common suffix between two buffers
+ */
+const diffCommonSuffix = (buffer1: Buffer, buffer2: Buffer): number => {
+	// Quick check for common null cases.
+	if (!buffer1.slice(-1).equals(buffer2.slice(-1))) {
```
shuse2, line 170:
```suggestion
	if (buffer1[buffer1.length -1] !== buffer2[buffer2.length -1]) {
```

elements/lisk-chain/src/diff.ts
```diff
@@ -130,15 +130,104 @@ const diffAlgo = (initial: Buffer, final: Buffer): HistoryType[] => {
...
+	// Binary search.
+	// Performance analysis: http://neil.fraser.name/news/2007/10/09/
+	let pointerMin = 0;
+	let pointerMax = Math.min(buffer1.length, buffer2.length);
+	let pointerMid = pointerMax;
+	let pointerEnd = 0;
+
+	while (pointerMin < pointerMid) {
+		if (
+			buffer1
```
shuse2, line 182: might be good to use compare?
https://nodejs.org/api/buffer.html#buffer_buf_compare_target_targetstart_targetend_sourcestart_sourceend
i didnt check performance tho

ishantiw: It slightly degrades the performance 😕

elements/lisk-chain/src/diff.ts
```diff
@@ -130,15 +130,104 @@ const diffAlgo = (initial: Buffer, final: Buffer): HistoryType[] => {
...
+	const strippedPrefixFinal = final.slice(commonPrefix, final.length);
+
+	const strippedInitial = strippedPrefixInitial.slice(
+		0,
+		strippedPrefixInitial.length - commonSuffix,
+	);
+	const strippedFinal = strippedPrefixFinal.slice(
+		0,
+		strippedPrefixFinal.length - commonSuffix,
+	);
```
MaximeGagnebin, line 226: Here I think it might be better to:

const commonPrefix = diffCommonPrefix(initial, final);
const strippedPrefixInitial = initial.slice(commonPrefix, initial.length);
const strippedPrefixFinal = final.slice(commonPrefix, final.length);

const commonSuffix = diffCommonSuffix(strippedPrefixInitial, strippedPrefixFinal);
const strippedInitial = strippedPrefixInitial.slice(
	0,
	strippedPrefixInitial.length - commonSuffix,
);
const strippedFinal = strippedPrefixFinal.slice(
	0,
	strippedPrefixFinal.length - commonSuffix,
);

This way, computing the common suffix is done on the shorter buffers (always good) and I think we avoid some corner cases where the prefix and suffix overlap in the shorter buffer.

elements/lisk-chain/src/diff.ts
```diff
@@ -130,15 +130,104 @@ const diffAlgo = (initial: Buffer, final: Buffer): HistoryType[] => {
...
+ */
+const diffCommonPrefix = (buffer1: Buffer, buffer2: Buffer): number => {
+	// Quick check for common null cases.
+	if (buffer1[0] !== buffer2[0]) {
+		return 0;
+	}
+	// Binary search.
+	// Performance analysis: http://neil.fraser.name/news/2007/10/09/
+	let pointerMin = 0;
+	let pointerMax = Math.min(buffer1.length, buffer2.length);
```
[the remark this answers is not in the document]
ishantiw: The maximum length it should traverse to find the common prefix/suffix is the buffer with min length. Ex: between 'abc' and 'ab', max possible common string will be 2, i.e., `min('abc', 'ab')`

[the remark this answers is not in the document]
ManuGowda: ahh ok, the variable name made it confusing, 👍

PR #5366, opened 2020-05-27
Update lisk-transactions to use encoded data - Closes #5270

### LINKED ISSUE #5270: Update lisk-transactions to use encoded data
### Description

- Define asset schema in BaseTransaction and dependent transactions
- Expose schema property as static field
- Update getBytes, getBasicBytes(rename to getSigningBytes)
- Remove `assetToBytes`, `assetToJSON`, `toJSON`, `stringify` 
- Remove properties not defined in schema
- Remove unnecessary type check in constructor
- Update `lisk-chain` -> `TransactionInterfaceAdapter` to encode and decode using `lisk-codec`

### Motivation

`lisk-codec` should be used consistently across the `lisk-sdk`

### Acceptance Criteria

- Update unit tests `lisk-transactions`, `lisk-chain`, `framework`

### Additional Information

Related issues: #5200

ManuGowda:
### What was the problem?

This PR resolves #5270 

### How was it solved?

- Introduce new transaction schema
- Remove toJSON, assetToJSON, stringify
- Expose schema from transactions
- Update transaction interface adapter to support decoding
- Update base transaction and custom transactions to support encoding and decoding

### How was it tested?
- Unrelated tests were removed
- Remaining tests to be updated after chaining protocol spec https://github.com/LiskHQ/lisk-sdk/issues/5368


elements/lisk-chain/src/data_access/transaction_interface_adapter.ts
```diff
@@ -24,25 +26,52 @@ export class TransactionInterfaceAdapter {
...
+			codec.addSchema(ASSET_SCHEMA);
 		});
 	}
 
-	public fromJSON(rawTx: TransactionJSON): BaseTransaction {
+	// First encode message asset and then encode base message
+	public encode(message: BaseTransaction): Buffer {
 		// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-		const TransactionClass = this._transactionClassMap.get(rawTx.type);
+		const TransactionClass = this._transactionClassMap.get(message.type);
```
shuse2, line 45: I think it's good to have this function, but for encode, isnt is just calling `getBytes` of the message?

elements/lisk-transactions/src/12_multisignature_transaction.ts
```diff
@@ -124,7 +118,7 @@ export class MultisignatureTransaction extends BaseTransaction {
 		}
 
 		// Check if empty signatures are present
-		if (this.signatures.includes('')) {
+		if (!this.signatures.length) {
```
shuse2, line 121: i think this logic is different? it needs to check if each signature has some value i think.
`!this.signatures.every(signature => signature.length > 0)`

elements/lisk-transactions/src/base_transaction.ts
```diff
@@ -76,117 +63,97 @@ export interface StateStore {
...
+	public constructor(transaction: BaseTransactionInput<T>) {
+		this.id = transaction.id;
+		this.type = transaction.type;
+		this.asset = transaction.asset;
+		this.nonce = transaction.nonce;
+		this.fee = transaction.fee;
+		this.senderPublicKey = transaction.senderPublicKey;
+		this.signatures = transaction.signatures;
+		this._id = bufferToHex(this.id);
+		this._senderPublicKey = bufferToHex(this.senderPublicKey);
```
shuse2, line 98:
```suggestion
		this._senderPublicKeyStr = bufferToHex(this.senderPublicKey);
```

elements/lisk-transactions/src/base_transaction.ts
```diff
@@ -76,117 +63,97 @@ export interface StateStore {
...
+		this._senderId = getAddressFromPublicKey(this.senderPublicKey);
```
shuse2, line 99:
```suggestion
		this._senderIdStr = getAddressFromPublicKey(this.senderPublicKey);
```

elements/lisk-transactions/src/base_transaction.ts
```diff
@@ -76,117 +63,97 @@ export interface StateStore {
...
+	public get senderPublicKeyStr(): string {
+		return this._senderPublicKey;
+	}
+
+	public get senderIdStr(): string {
+		return bufferToHex(this.senderId);
+	}
+	/* End Getters */
+
+	public getBasicBytes(): Buffer {
```
shuse2, line 132:
```suggestion
	protected getSigningBytes(): Buffer {
```

elements/lisk-transactions/src/base_transaction.ts
```diff
@@ -400,87 +361,51 @@ export abstract class BaseTransaction {
...
-					const { passphrase } = keysAndPassphrases[aKey];
+				if (keysAndPassphrases[publicKeyStr]) {
+					const { passphrase } = keysAndPassphrases[publicKeyStr];
 					this.signatures.push(
 						signData(transactionWithNetworkIdentifierBytes, passphrase),
 					);
 				} else {
 					// Push an empty signature if a passphrase is missing
-					this.signatures.push('');
+					this.signatures.push(Buffer.from(''));
```
shuse2, line 373:
```suggestion
					this.signatures.push(Buffer.alloc(0));
```

elements/lisk-transactions/src/base_transaction.ts
```diff
@@ -400,87 +361,51 @@ export abstract class BaseTransaction {
...
 				}
 			}
-			this._id = getId(this.getBytes());
```
shuse2, line 414: Since sign will change the id, probably it's good to have this one, but u can do simply
```
this._id = hash(this.getBytes());	
```

elements/lisk-transactions/src/utils/sort.ts
```diff
@@ -15,22 +15,17 @@
...
-			return 1;
-		}
-		if (publicKeyA < publicKeyB) {
-			return -1;
-		}
-
-		return 0;
-	});
+export const sortKeysAscending = (
+	publicKeys: Array<Readonly<Buffer>>,
```
shuse2, line 19:
```suggestion
	publicKeys: Buffer[],
```

elements/lisk-transactions/src/utils/sort.ts
```diff
@@ -15,22 +15,17 @@
...
+): Array<Readonly<Buffer>> =>
+	publicKeys.sort((publicKeyA, publicKeyB) =>
+		(publicKeyA as Buffer).compare(publicKeyB as Buffer),
```
shuse2, line 22: if above, i think there is no need to cast them?

elements/lisk-transactions/src/utils/verify.ts
```diff
@@ -54,14 +59,14 @@ export const verifyMinRemainingBalance = (
...
-	id: string,
+	id: Buffer,
 	account: Account,
 	nonce: bigint,
 ): TransactionError | undefined => {
 	if (nonce < account.nonce) {
 		return new TransactionError(
 			`Incompatible transaction nonce for account: ${
-				account.address
+			bufferToHex(account.address)
```
shuse2, line 69: isnt it just account.address.toString('hex')

elements/lisk-transactions/src/utils/sort.ts
ManuGowda: keys are `Array<Readonly<Buffer>>` so it wont work as i have to cast them and send, so i made this change to support

elements/lisk-transactions/src/utils/verify.ts
ManuGowda: could do that too.. just used to utility, should i change?

shuse2: not necessary to change =)

elements/lisk-chain/src/data_access/data_access.ts
```diff
@@ -343,7 +343,7 @@ export class DataAccess {
 		const transactions =
 			// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
 			blockJSON.transactions?.map(transaction =>
-				this._transactionAdapter.fromJSON(transaction),
+				this._transactionAdapter.decode(transaction),
```
pablitovicente, line 346: Maybe rename `transaction` to `transactionBuffer` so as it to be easy to tell it's a binary buffer?

elements/lisk-chain/src/data_access/data_access.ts
```diff
@@ -384,7 +384,7 @@ export class DataAccess {
 	public deserializeTransaction(
 		transactionJSON: TransactionJSON,
 	): BaseTransaction {
-		return this._transactionAdapter.fromJSON(transactionJSON);
+		return this._transactionAdapter.decode(transactionJSON);
```
pablitovicente, line 387: Maybe rename `transactionJSON` to `transactionBuffer` as `decode()` deals with binary data and not JSON?

elements/lisk-chain/src/data_access/transaction_interface_adapter.ts
```diff
@@ -24,25 +26,52 @@ export class TransactionInterfaceAdapter {
...
+		// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+		const binaryAsset = codec.encode(TransactionClass.ASSET_SCHEMA, message.asset as unknown as GenericObject);
+		const binaryAssetWithMessage = { ...message, asset: binaryAsset };
+
+		const binaryMessage = codec.encode(BaseTransaction.BASE_SCHEMA, binaryAssetWithMessage as unknown as GenericObject);
+
+		return binaryMessage;
+	}
+
+	// First decode base message and then decode asset
```
pablitovicente, line 60: Why is this done in two steps instead of doing it in one `decode()` call? I think it might be faster in general to call decode only once if possible

elements/lisk-transactions/src/10_delegate_transaction.ts
```diff
@@ -31,67 +30,30 @@ export interface DelegateAsset {
 	readonly username: string;
 }
 
-export const delegateAssetFormatSchema = {
+export const delegateRegistrationAssetSchema = {
+	$id: 'lisk/delegate-registration-transaction',
```
pablitovicente, line 34: Not to do in this PR but should I create an issue to hash $id from schemas to avoid having this as the key of the object literal holding the schema cache? (it will work like this but just in case someone uses something different here in the future or in customapps)

elements/lisk-chain/src/data_access/data_access.ts
ManuGowda: this piece of code is being refactored by shu in chain PR, so to avoid conflicting changes, I kept like this.

elements/lisk-chain/src/data_access/data_access.ts
ManuGowda: same

elements/lisk-chain/src/data_access/transaction_interface_adapter.ts
ManuGowda: you mean, create the whole schema and decode? I followed the LIP, according to LIP we need to decode root message first and then decode asset https://github.com/LiskHQ/lips/blob/master/proposals/lip-0028.md#deserialization

elements/lisk-chain/src/data_access/transaction_interface_adapter.ts
ManuGowda: 👍

PR #5371, opened 2020-05-29
Benchmark encode/decode performance - Closes #5303

### LINKED ISSUE #5303: Benchmark encode/decode performance

pablitovicente:
### What was the problem?

This PR resolves #5303

### How was it solved?

By adding new benchmarks in lisk-codec/benchmarks

### How was it tested?

By running new benchmarks


elements/lisk-codec/benchmark/encode_decode_biggest_lisk_transaction.js
```diff
@@ -0,0 +1,613 @@
...
+	senderPublicKey: Buffer.from(
+		'0b211fce4b615083701cb8a8c99407e464b2f9aa4f367095322de1b77e5fcfbe',
+		'hex',
+	),
+	nonce: 1,
+	fee: BigInt(1500000000),
+	type: 12,
+	asset: {
+		mandatoryKeys: [
+			Buffer.from(
```
ManuGowda, line 31: maybe u can do this to avoid this `[ ...Array(64).keys() ].map(() => cr.buffcr.getRandomBytes(32))`

pablitovicente: Better to avoid random so input is always exactly the same? (it shouldn't matter but just to always keep exactly same input I would keep it fixed)

ManuGowda: true.. but since the length, type and input size remain constant, it shouldn't matter. just a suggestion for readability and less code to follow.

PR #5374, opened 2020-05-29
Update transaction pool to use new schema - Closes #5373

### LINKED ISSUE #5373: Update transaction pool to adapt to the new schema
### Description

Update senderPublicKey and Transaction id to be buffer.

### Motivation

- Transaction pool needs to be updated to use new transaction schema

### Acceptance Criteria

- All test should pass
- All type of the transaction needs to correspond to the latest schema

shuse2:
### What was the problem?

This PR resolves #5373 

### How was it solved?

- Update map to be Buffer map
- Return as Buffer map

### How was it tested?

- All test are updated


shuse2:
@mitsuaki-u it's to the feature branch, and until other issue is resolved, it will fail on jenkins

Issue #5266, opened 2020-05-04
Update lisk-dpos to use encoded data


shuse2:
### Description

- Define schema for voteWeights and forgersList
- Store encoded value and used decoded value using stateStore
- Update CHAIN_STATE_DELEGATE_USERNAMES to use binary data
- update delegate transaction (registration) based on changes

### Motivation

`lisk-codec` should be used consistently across the `lisk-sdk`

### Acceptance Criteria

- All data stored in dpos has schema defined
- Update unit tests

### Additional Information

Related issues: #5200

PR #5379, opened 2020-06-02
Update dpos to use new block header - Closes #5375

### LINKED ISSUE #5375: Update DPoS library to use new block header schema
### Description

Update DPoS schema to use updated block header schema

### Motivation

Block header schema is updated, so it needs to be adjusted.

### Acceptance Criteria

All test should pass

shuse2:
### What was the problem?

This PR resolves #5375 

### How was it solved?

- Update the types, and fix the tests

### How was it tested?

- All the tests are enabled and tested


elements/lisk-dpos/src/delegates_list.ts
```diff
@@ -51,10 +47,23 @@ interface DelegatesListConstructor {
...
 }
 
+export const decodeForgerList = (buffer: Buffer): ForgersList => {
+	// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+	const parsedForgersList = JSON.parse(buffer.toString('utf8'));
+	// eslint-disable-next-line
+	return parsedForgersList.map(
+		(fl: { round: number; delegates?: string[]; standby?: string[] }) => ({
+			round: fl.round,
+			delegates: fl.delegates?.map(d => Buffer.from(d, 'binary')) ?? [],
```
nazarhussain, line 61: As discussed everywhere we present `Buffer -> String` we will use Base64. Some should use that consistent approach everywhere...

elements/lisk-dpos/src/delegates_list.ts
```diff
@@ -64,21 +73,46 @@ export const getForgersList = async (
...
+
+	// eslint-disable-next-line
+	return parsedVoteWeights.map(
+		(vw: {
+			round: number;
+			delegates: { address: string; voteWeight: string }[];
+		}) => ({
+			round: vw.round,
+			delegates: vw.delegates.map(d => ({
+				address: Buffer.from(d.address, 'binary'),
```
nazarhussain, line 109: Same here..

elements/lisk-dpos/src/delegates_list.ts
shuse2: The Buffer usage will be update with https://github.com/LiskHQ/lisk-sdk/issues/5266 issue

nazarhussain:
@shuse2 The changes looks good overall, some listing issues probably buffer used as keys of objects.

PR #5380, opened 2020-06-02
Update bft to support new block header - Closes #5376

### LINKED ISSUE #5376: Update BFT to use updated block header schema
### Description

BFT should be adjusted to use the latest block header schema

### Motivation

- Block header schema is updated with the codec LIPs

### Acceptance Criteria

- All test should pass
- All block header and account type are updated

shuse2:
### What was the problem?

This PR resolves #5376 

### How was it solved?

- Update type definition
- Update protocol specs

### How was it tested?

- All the tests are enabled and bft tests are passing


elements/lisk-bft/src/bft.ts
```diff
@@ -139,8 +126,8 @@ export class BFT extends EventEmitter {
 	}
 
 	public forkChoice(
-		block: BlockHeaderWithID,
-		lastBlock: BlockHeaderWithID,
+		block: BlockHeader,
+		lastBlock: BlockHeader,
```
nazarhussain, line 130: Can we also change the name here as well... As we did in other PR recently...  

```
header: BlockHeader,
lastBlockHeader: BlockHeader,
```

PR #5381, opened 2020-06-02
Update lisk-bft to use encoded data - Closes #5267

### LINKED ISSUE #5267: Update lisk-bft to use encoded data
### Description

- Define schema for finalizedHeight
- Use schema and `lisk-codec` to encode and decode instead of `JSON.stringify` and `JSON.parse`
- Stored values using stateStore which is already done but as string values 

### Motivation

`lisk-codec` should be used consistently across the `lisk-sdk`

### Acceptance Criteria

- All data used in bft has schema defined
- Update unit tests

### Additional Information

Related issues: #5200

nazarhussain:
### What was the problem?

This PR resolves #5267

### How was it solved?

Used `lisk-codec` instead of `JSON.stringify` and `JSON.parse`

### How was it tested?

Run unit tests.

### DIFF
--- a/elements/lisk-bft/src/bft.ts
+++ b/elements/lisk-bft/src/bft.ts
@@ -44,6 +44,7 @@ export const BFTFinalizedHeightCodecSchema = {
 			fieldNumber: 1,
 		},
 	},
+	required: ['finalizedHeight'],
 };
 
 codec.addSchema(BFTFinalizedHeightCodecSchema);
--- a/elements/lisk-bft/src/bft.ts
+++ b/elements/lisk-bft/src/bft.ts
@@ -34,10 +34,10 @@ import {
 export const CONSENSUS_STATE_FINALIZED_HEIGHT_KEY = 'bft:finalizedHeight';
 export const EVENT_BFT_BLOCK_FINALIZED = 'EVENT_BFT_BLOCK_FINALIZED';
 
-export const BFTPersistedValuesSchema = {
+export const BFTFinalizedHeightCodecSchema = {
 	type: 'object',
-	$id: '/BFT',
-	title: 'LiskBFT',
+	$id: '/BFT/FinalizedHeight',
+	title: 'Lisk BFT Finalized Height',
 	properties: {
 		finalizedHeight: {
 			dataType: 'uint32',
@@ -46,7 +46,7 @@ export const BFTPersistedValuesSchema = {
 	},
 };
 
-codec.addSchema(BFTPersistedValuesSchema);
+codec.addSchema(BFTFinalizedHeightCodecSchema);
 
 /**
  * BFT class responsible to hold integration logic for finality manager with the framework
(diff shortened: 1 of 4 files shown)


elements/lisk-bft/src/bft.ts
```diff
@@ -32,6 +34,20 @@ import {
 export const CONSENSUS_STATE_FINALIZED_HEIGHT_KEY = 'bft:finalizedHeight';
 export const EVENT_BFT_BLOCK_FINALIZED = 'EVENT_BFT_BLOCK_FINALIZED';
 
+export const BFTPersistedValuesSchema = {
+	type: 'object',
+	$id: '/BFT',
```
shuse2, line 39: there are some other issues to store more value for the BFT, so it's better to be bit more precise for the id

elements/lisk-bft/src/bft.ts
```diff
@@ -32,6 +34,20 @@ import {
...
+	title: 'LiskBFT',
+	properties: {
+		finalizedHeight: {
+			dataType: 'uint32',
+			fieldNumber: 1,
+		},
+	},
+};
+
+codec.addSchema(BFTPersistedValuesSchema);
```
shuse2, line 49: This we should put it in the constructor so that we see the error

nazarhussain: For this one I am not sure if we do it. `codec` Is a module level object and if we want to add/modify anything to it we should do it on module level. What do you think?

shuse2: If the schema is invalid, does it properly throw the error?

nazarhussain: Not in this branch as we don't have any validation. But in development it will throw error when requiring the file.

elements/lisk-bft/src/bft.ts
```diff
@@ -32,6 +34,20 @@ import {
 export const CONSENSUS_STATE_FINALIZED_HEIGHT_KEY = 'bft:finalizedHeight';
 export const EVENT_BFT_BLOCK_FINALIZED = 'EVENT_BFT_BLOCK_FINALIZED';
 
+export const BFTFinalizedHeightCodecSchema = {
+	type: 'object',
+	$id: '/BFT/FinalizedHeight',
```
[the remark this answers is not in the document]
nazarhussain: `$id` is used as url format, we can better organise more with that. For title we have more descriptive name already.

elements/lisk-bft/src/bft.ts
```diff
@@ -32,6 +34,20 @@ import {
...
+	title: 'Lisk BFT Finalized Height',
+	properties: {
+		finalizedHeight: {
+			dataType: 'uint32',
+			fieldNumber: 1,
+		},
+	},
```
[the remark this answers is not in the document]
nazarhussain: All properties which have fieldNumber are required. But on safer side let's add required as well.

Issue #5234, opened 2020-04-30
Network Longevity - Define new block schema and processing for genesis block


shuse2:
### Description

Update genesis block schema to allow the specification of an initial state of accounts and an initial delegate list valid for a certain bootstrapping period.

### Motivation

Refer https://github.com/LiskHQ/lips/blob/master/proposals/lip-0034.md#motivation

#### Sprint 21

- Create lisk-genesis package in elements #5276
- Implement genesis block creation function in lisk-genesis #5277
- Replace epochTime with genesis block timestamp #5280
- Implement genesis block validation function in lisk-genesis #5386
- Update genesis block for devnet #5283
#### Sprint 22
- Move communityIdentifier from genesis block to configuration #5447
- Create BlockProcessorV0 which handles the genesis block #5278
- Remove current genesis block specific handling from the framework and lisk-chain #5279
- Update lisk-dpos to handle bootstrap period #5281
- Update lisk-bft to handle bootstrap period #5282

PR #5382, opened 2020-06-03
Update framework use the new schema - Closes #5378

### LINKED ISSUE #5378: Update framework to use updated schema
### Description

- Update framework to use the latest libraries
- Update framework to use updated schema and library interfaces
- Update all the test

### Motivation

- All the block, transaction and account schema are updated for codec LIPs

### Acceptance Criteria

- All the test and build should pass for the whole repository

### Additional Information
- This issue should be done at the end of the schema updating issues

shuse2:
### What was the problem?

This PR resolves #5378

### How was it solved?

- Update all the usages of the old block and account format
- Update to use base64 for external communication and logs
- Integrate all the library changes

### How was it tested?

- All the test cases are updated
- Application should run after restarting the database


framework/src/application/logger/logger.ts
```diff
@@ -86,7 +86,31 @@ class ConsoleLog {
...
+								for (const k of Object.keys(val)) {
+									// eslint-disable-next-line
+									if (typeof val[k] === 'bigint') {
+										// eslint-disable-next-line
+										val[k] = val[k].toString();
+									}
+									// eslint-disable-next-line
+									if (Buffer.isBuffer(val[k])) {
+										// eslint-disable-next-line
+										val[k] = val[k].toString('base64');
```
nazarhussain, line 104: I am not a big fan of that approach but until NodeJS natively support we can do something like this: 

```
Buffer.prototype.toJSON = function () { return this.toString('base64'); }
BigInt.prototype.toJSON = function () { return this.toString(); }
```

This way it will will be done automatically when ever we use `JSON.stringify`. 

https://github.com/cyberphone/es6-bigint-json-support#21-rfc-mode-serialization

framework/src/application/node/account.ts
```diff
@@ -0,0 +1,119 @@
...
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+export const accountAssetSchema = {
```
nazarhussain, line 15: This file contains all the types, would be nice to move it under `types` directory?

framework/src/application/node/block_processor_v2.ts
```diff
@@ -48,185 +44,62 @@ interface BlockProcessorInput {
...
-	const idsAsBuffers = ids.map(id => stringToBuffer(id));
-	const tree = new MerkleTree(idsAsBuffers);
-
-	return bufferToHex(tree.root);
+	return tree.root;
 };
 
 export class BlockProcessorV2 extends BaseBlockProcessor {
+	public static readonly schema = {
+		$id: 'test/defaultBlockHeaderAssetSchema',
```
nazarhussain, line 84: Why `$id` contains `test`? I would suggest `/block/header/v2`

elements/lisk-chain/src/utils/buffer_map.ts
```diff
@@ -42,6 +42,13 @@ export class BufferMap<V> {
 		return new BufferMap(cloneDeep(this._data));
 	}
 
+	public entries(): [Buffer, V][] {
+		return Object.entries(this._data).map(([key, value]) => [
+			Buffer.from(key, 'binary'),
+			value,
+		]) as [Buffer, V][];
+	}
```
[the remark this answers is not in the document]
shuse2: it's because if you do like `this._data[key]` u dont know if it exists or not, and to say that it seems u need to have undefined

elements/lisk-transactions/src/types.ts
```diff
@@ -99,13 +99,13 @@ export interface TransactionJSON {
...
-	readonly id: Buffer;
-	readonly type: number;
+	readonly id?: Buffer;
+	readonly type?: number;
 	readonly senderPublicKey: Buffer;
 	readonly nonce: bigint;
 	readonly fee: bigint;
 	readonly asset: T;
-	readonly signatures: Array<Readonly<Buffer>>;
+	readonly signatures?: Array<Readonly<Buffer>>;
```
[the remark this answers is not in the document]
shuse2: It was like this, and we dont want to mutate the buffer, but we need to mutate the array

elements/lisk-transactions/src/types.ts
```diff
@@ -99,13 +99,13 @@ export interface TransactionJSON {
 }
 
 export interface BaseTransactionInput<T = object> {
-	readonly id: Buffer;
-	readonly type: number;
+	readonly id?: Buffer;
+	readonly type?: number;
```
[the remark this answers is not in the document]
shuse2: type is assigned in the constructor. In this way, when u instantiate the class, u dont need to specify type

elements/lisk-chain/src/chain.ts
```diff
@@ -338,7 +346,12 @@ export class Chain {
 	}
 
 	public validateBlockHeader(block: Block): void {
-		validatePreviousBlockProperty(block, this.genesisBlock);
+		// Validate block header
+		const errors = validator.validate(blockHeaderSchema, block.header);
+		if (errors.length) {
+			throw new Error(errors[0].message);
```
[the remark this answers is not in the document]
shuse2: it is returning the error array, so i dont think this works

framework/src/application/node/account.ts
shuse2: types is standard to define typescript types, and schema is i think different

framework/src/application/logger/logger.ts
shuse2: I would prefer not to overwrite the prototype, and handle them separately.
Especially Buffer does have toJSON implementation

framework/src/application/node/synchronizer/base_synchronizer.ts
```diff
@@ -12,43 +12,123 @@
...
+			peerId,
+			data: {
+				blockId: fromID.toString('base64'),
+			},
+		}); // Note that the block matching lastFetchedID is not returned but only higher blocks.
+		if (!data || !data.length) {
+			throw new Error('Peer did not respond with block');
+		}
+		return data.map(d =>
+			this._chain.dataAccess.decode<BlockHeaderAsset>(Buffer.from(d, 'base64')),
```
nazarhussain, line 122: When we are specifying return type here `<BlockHeaderAsset>`, then how can it don't require schema for that type?

shuse2: This is more like a type helper. the asset mapping happen by the version number registered to the chain library.
There was no better way to enable type and registering

elements/lisk-chain/src/chain.ts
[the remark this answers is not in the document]
shuse2: ah you mean parse it to the error, but I think if it just return the particular error instead of asking it to handle here, it would be better.
but I do agree we should use the single error everywhere and we should fix it.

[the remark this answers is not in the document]
nazarhussain: Just for reference can you create an issue to remove all usages of throwing errors like `throw new Error(errors[0].message)`

framework/src/application/node/forger/forger.ts
```diff
@@ -354,11 +339,15 @@ export class Forger {
 			}
 			// If all hash onion is used, throw an error
 			if (highestCount >= hashOnionConfig.count) {
-				throw new Error(`All of the hash onion is used for ${account.address}`);
+				throw new Error(
+					`All of the hash onion is used for ${account.address.toString(
+						'hex',
```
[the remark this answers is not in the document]
ManuGowda: so for errors its hex and for logging its base64?

framework/src/application/node/node.ts
```diff
@@ -50,16 +52,135 @@ import { EventPostTransactionData } from '../../types';
...
+interface Payload {
+	id: string;
+	senderPublicKey: string;
+	nonce: string;
+	fee: string;
+	signatures: string[];
+}
+
+interface TransferPayload extends Payload {
+	type: 8;
```
ManuGowda, line 76: can we required the transactions and use `TransferTransaction.Type`?

framework/src/application/node/node.ts
```diff
@@ -50,16 +52,135 @@ import { EventPostTransactionData } from '../../types';
...
+			seedReveal: string;
+			maxHeightPreviouslyForged: number;
+			maxHeightPrevoted: number;
+		};
+		readonly signature: string;
+	};
+	readonly payload: Array<TransferPayload | DelegatePayload | VotePayload>;
 }
 
+const convertGenesisBlock = (genesis: GenesisBlockJSON): Block => {
```
ManuGowda, line 121: maybe `genesisBlockToJSON`? or `blockToJSON`?

framework/src/application/node/node.ts
```diff
@@ -50,16 +52,135 @@ import { EventPostTransactionData } from '../../types';
...
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any
+		const txHeader = {
+			id: Buffer.from(tx.id, 'hex'),
+			type: tx.type,
+			senderPublicKey: Buffer.from(tx.senderPublicKey, 'hex'),
+			nonce: BigInt(tx.nonce),
+			fee: BigInt(tx.fee),
+			signatures: tx.signatures.map(s => Buffer.from(s, 'hex')),
+		};
+		if (tx.type === 8) {
```
ManuGowda, line 145: same here `TransferTransaction.TYPE`?

framework/src/application/node/node.ts
```diff
@@ -565,6 +733,30 @@ export class Node {
 			mechanisms: [blockSyncMechanism, fastChainSwitchMechanism],
 		});
 
+		blockSyncMechanism.events.on(
+			EVENT_SYNCHRONIZER_SYNC_RQUIRED,
```
ManuGowda, line 737: typo `EVENT_SYNCHRONIZER_SYNC_REQUIRED`.

framework/src/application/node/node.ts
```diff
@@ -565,6 +733,30 @@ export class Node {
...
+					this._logger.error(
+						{ err: err as Error },
+						'Error occurred during synchronization.',
+					);
+				});
+			},
+		);
+
+		fastChainSwitchMechanism.events.on(
+			EVENT_SYNCHRONIZER_SYNC_RQUIRED,
```
ManuGowda, line 749: same event name for both type of sync?

framework/src/application/node/node.ts
```diff
@@ -565,6 +733,30 @@ export class Node {
...
+			({ block, peerId }) => {
+				this._synchronizer.run(block, peerId).catch(err => {
+					this._logger.error(
+						{ err: err as Error },
+						'Error occurred during synchronization.',
```
ManuGowda, line 754:
```suggestion
						'Error occurred during fast chain synchronization mechanism.',
```

framework/src/application/node/node.ts
```diff
@@ -627,26 +820,18 @@ export class Node {
 	}
 
 	private _subscribeToEvents(): void {
-		this._channel.subscribe(
-			'app:block:broadcast',
+		// FIXME: this event is using instance, it should be replaced by event emitter
+		this._processor.events.on(
+			EVENT_PROCESSOR_BRADCASRT_BLOCK,
```
ManuGowda, line 825:
```suggestion
			EVENT_PROCESSOR_BROADCAST_BLOCK,
```

framework/src/application/node/processor/processor.ts
```diff
@@ -30,6 +31,10 @@ const forkStatusList = [
 	ForkStatus.DISCARD,
 ];
 
+export const EVENT_PROCESSOR_SYNC_REQUIRED = 'EVENT_PROCESSOR_SYNC_REQUIRED';
```
ManuGowda, line 34: `PROCESSOR_SYNC_REQUIRED`, at places we use the value without `event_`, maybe we should use one format.

framework/src/application/node/processor/processor.ts
```diff
@@ -327,18 +335,18 @@ export class Processor {
...
 			lastBlock,
 			stateStore,
 		});
 
-		const blockJSON = await this.serialize(block);
 		if (!skipBroadcast) {
-			this.channel.publish('app:block:broadcast', {
-				block: blockJSON,
+			// FIXME: this is using instance, use event emitter instead
+			this.events.emit(EVENT_PROCESSOR_BRADCASRT_BLOCK, {
```
ManuGowda, line 348: typo

framework/src/application/node/synchronizer/base_synchronizer.ts
```diff
@@ -12,43 +12,123 @@
...
+import { EventEmitter } from 'events';
 import { Logger } from '../../logger';
 import { InMemoryChannel } from '../../../controller/channels';
+import { BlockHeaderAsset } from '../block_processor_v2';
+import {
+	ApplyPenaltyAndRestartError,
+	ApplyPenaltyAndAbortError,
+} from './errors';
+
+export const EVENT_SYNCHRONIZER_SYNC_RQUIRED =
```
ManuGowda, line 25: typo

framework/src/application/node/synchronizer/base_synchronizer.ts
```diff
@@ -12,43 +12,123 @@
...
+			data: string[] | undefined;
+		}>('requestFromPeer', {
+			procedure: 'getBlocksFromId',
+			peerId,
+			data: {
+				blockId: fromID.toString('base64'),
+			},
+		}); // Note that the block matching lastFetchedID is not returned but only higher blocks.
+		if (!data || !data.length) {
+			throw new Error('Peer did not respond with block');
```
ManuGowda, line 119: maybe use the same format above, din't => `did not`? did not is much better

framework/src/application/node/node.ts
shuse2: we can, but this will be removed next week. this is temporal solution

framework/src/application/node/node.ts
shuse2: Same as above, this is temporal solution until #5234

Issue #5209, opened 2020-04-28
Update PoM protocol-specs to include transactionRoot


mitsuaki-u:
### Description

Existing proof-of-misbehavior protocol specifications must be updated to include the new `transactionsRoot` property. 

### Motivation

Protocol-specs must be kept up to date.

### Acceptance Criteria

Generator outputs correct block data.

### Additional information

Related issues: #5204

PR #5383, opened 2020-06-03
Update transaction test using updated protocol-spec - Closes #5368, #5335 and #5209

### LINKED ISSUE #5335: Update all protocol-specs to follow new address format

### LINKED ISSUE #5368: Update all transaction related protocol spec
### Description

After integrating `lisk-codec` in `lisk-transactions` all the tests are broken as the input and out is modified due to introduction of `lisk-codec`

### Acceptance Criteria

- Update `protocol-spec` all transactions related specs
- Remove unwanted/unused protocol specs
- Update all the tests using modified protocol specs

### Additional Information
- Update the protocol spec after changing accounts and blocks related changes for `lisk-codec` integration

shuse2:
### What was the problem?

This PR resolves #5368, resolves #5335 and #5209 

### How was it solved?

- Updated all protocol spec to use new address system and codec
- Update transaction tests according to the protocol spec

### How was it tested?

- All transaction library test should pass


protocol-specs/generators/multisignature_registration/index.js
```diff
@@ -14,164 +14,206 @@
...
 
-const {
-	hexToBuffer,
-	intToBuffer,
-	hash,
-	bufferToHex,
-	signData,
-} = require('@liskhq/lisk-cryptography');
+const { signData } = require('@liskhq/lisk-cryptography');
+const { Codec } = require('@liskhq/lisk-codec');
```
nazarhussain, line 18: Why to use class and not the instance `codec`?

protocol-specs/generators/multisignature_registration/index.js
```diff
@@ -14,164 +14,206 @@
...
 const BaseGenerator = require('../base_generator');
 
-const networkIdentifier =
-	'e48feb88db5b5cf5ad71d93cdcd1d879b6d5ed187a36b0002cc34e0ef9883255';
+const codec = new Codec();
```
nazarhussain, line 21: We have a default instance in the package already, should be used.

protocol-specs/generators/multisignature_registration/index.js
```diff
@@ -14,164 +14,206 @@
...
+
+const networkIdentifier = Buffer.from(
+	'e48feb88db5b5cf5ad71d93cdcd1d879b6d5ed187a36b0002cc34e0ef9883255',
+	'hex',
+);
```
nazarhussain, line 26: This `networkIdentifier` is shared across almost all of the generators. Can we move such shared stuff to a common file to reduce the code footprint so we can review and manage it easily.

package.json
```diff
@@ -26,7 +26,8 @@
 			"framework",
 			"elements/*",
 			"commander",
-			"sdk"
+			"sdk",
+			"protocol-specs"
```
nazarhussain, line 30: With this change, now all elements used in `protocol-specs` will be linked to local packages. Previously only published packages were used in here. Does that will impact any thing?

protocol-specs/generators/multisignature_registration/index.js
```diff
@@ -14,164 +14,206 @@
...
+			fieldNumber: 5,
+		},
+		signatures: {
+			type: 'array',
+			items: {
+				dataType: 'bytes',
+			},
+			fieldNumber: 6,
+		},
+	},
```
nazarhussain, line 135: Same here, this base schema will be used among many places, can we move it to a shared source file?

protocol-specs/generators/multisignature_registration/index.js
```diff
@@ -14,164 +14,206 @@
...
+const getAssetBytes = asset => codec.encode(multisigRegAsset, asset);
 
-	return buf;
+const getSignBytes = tx => {
+	const assetBytes = getAssetBytes(tx.asset);
+	const signingTx = {
+		...tx,
+		asset: assetBytes,
+		signatures: [],
+	};
```
nazarhussain, line 164: Why can't get have single schema and single encode call?

shuse2: In base schema, asset is defined as bytes, and when we encode it has to be done in 2 steps, so it's also done in 2 steps as defined in LIP

protocol-specs/generators/multisignature_registration/index.js
shuse2: I just wanted to make sure there is no duplicate registration of schema

protocol-specs/generators/multisignature_registration/index.js
shuse2: same as above

protocol-specs/generators/multisignature_registration/index.js
shuse2: yes but with the current structure, all of the generator is independent, and even if we change the network identifier and it still works

package.json
shuse2: With only lerna, that was the case, but it seems like if we specify the old version, it is fetching the old version. so unless we increase the dependency version it's ok

elements/lisk-transactions/src/unlock_token.ts
```diff
@@ -85,7 +88,9 @@ export const unlockToken = (
 		return baseTransactionToJSON(transaction as UnlockTransaction);
 	}
 
-	const unlockTransaction = new UnlockTransaction(transaction as UnlockTransaction);
+	const unlockTransaction = new UnlockTransaction(
+		transaction as UnlockTransaction,
```
ishantiw, line 92: I'm getting TypeError `Conversion of type '{ type: number; asset: { unlockObjects: readonly Unlock[]; }; nonce: bigint; fee: bigint; senderPublicKey: string | undefined; }' to type 'UnlockTransaction' may be a mistake because neither type sufficiently overlaps with the other`

elements/lisk-transactions/test/unlock_token.spec.ts
```diff
@@ -12,142 +12,145 @@
...
+// import * as validUnlockTransactionScenario from '../fixtures/unlock_transaction/unlock_transaction.json';
 
-import { unlockToken } from '../src/unlock_token';
-import { TransactionJSON } from '../src/types';
+// import { unlockToken } from '../src/unlock_token';
+// import { TransactionJSON } from '../src/types';
 
 describe('#unlockToken transaction', () => {
-	let unlockTokenTransaction: Partial<TransactionJSON>;
+	it('should be updated or deleted once unlockToken design is decided', () => {
```
ishantiw, line 21: maybe it's better to use multi-line comment as its easier to follow the changes 😅

elements/lisk-transactions/src/unlock_token.ts
ishantiw: it works now

PR #5385, opened 2020-06-03
Update lisk validator datatype - Closes #5263

### LINKED ISSUE #5263: Update lisk-validator to check against dataType
### Description

With the new JSON schema definition used in `lisk-codec`

- Add validation for the `dataType` field, `string`, `bytes`, `sint32`, `sint64`, `uint32`, `uint64`
- Remove `formats` which are replaced by the above, and the current formats are not usable
- Remove unnecessary validation function
  - TransferData (check if transfer transaction check this)
  - Fee, Nonce, Transfer amount checks
 - Value checking greater or less than 0 (specifying `uint64`, should be sufficient for the use-case)

### Motivation

`lisk-validator` should use the same schema used for `lisk-codec`

### Acceptance Criteria

- Add unit test to check all dataTypes are correctly validated 
- Remove all the above functions and related tests

### Additional Information

Whats a `dataType`? refer section https://github.com/LiskHQ/lips/blob/master/proposals/lip-0027.md#data-types

Related issues: #5200

ishantiw:
### What was the problem?

This PR resolves #5263 

### How was it solved?

- Updated data types according to [lip-0027 - data-types](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0027.md#data-types)
- Removed dataTypes and formats no longer needed
- Fix or remove related tests

### How was it tested?

`npm run test` under `elements/lisk-validator`


elements/lisk-validator/src/lisk_meta_schema.ts
```diff
@@ -38,7 +38,6 @@ export const liskMetaSchema = {
 			type: 'object',
 			propertyNames: {
 				type: 'string',
-				format: 'camelCase',
```
shuse2, line 41: I think this is newly added, and we want to keep it for meta schema check

elements/lisk-validator/src/formats.ts
```diff
@@ -103,24 +92,5 @@ export const signedPublicKey = (data: string): boolean => {
...
-export const uint64 = (data: string): boolean =>
-	isNumberString(data) && isUint64(BigInt(data));
-
-export const uint32 = (data: string): boolean =>
-	isNumberString(data) && isUint32(BigInt(data));
-
-export const int32 = (data: string): boolean =>
-	isNumberString(data) && isInt32(BigInt(data));
-
-const camelCaseRegex = /^([a-z]+[A-Z]*)+$/;
```
shuse2, line 123: same as below

elements/lisk-validator/src/validation.ts
```diff
@@ -236,24 +214,30 @@ export const validateNetworkIdentifier = (
...
 };
 
-export const isInt32 = (num: bigint | number): boolean => {
-	if (typeof num === 'number') {
-		return num <= MAX_INT32 && num >= MIN_INT32;
+export const isString = (data: unknown): boolean => typeof data === 'string';
+
+export const isBoolean = (data: unknown): boolean => typeof data === 'boolean';
+
+export const isInt32 = (data: unknown): boolean => {
```
shuse2, line 221:
```suggestion
export const isSInt32 = (data: unknown): boolean => {
```

elements/lisk-validator/src/validation.ts
```diff
@@ -236,24 +214,30 @@ export const validateNetworkIdentifier = (
...
+		return data <= MAX_UINT32 && data >= 0;
 	}
 
-	return num <= BigInt(MAX_UINT32) && num >= BigInt(0);
+	return false;
 };
 
-export const isInt64 = (num: bigint): boolean =>
-	num <= MAX_INT64 && num >= MIN_INT64;
+export const isInt64 = (data: unknown): boolean =>
```
shuse2, line 237:
```suggestion
export const isSInt64 = (data: unknown): boolean =>
```

elements/lisk-validator/src/constants.ts
```diff
@@ -20,6 +20,6 @@ export const MAX_PUBLIC_KEY_LENGTH = 32;
 export const MAX_INT32 = 2147483647;
 export const MIN_INT32 = MAX_INT32 * -1;
 export const MAX_UINT32 = 4294967295;
-export const MAX_INT64 = BigInt('9223372036854775807');
-export const MIN_INT64 = MAX_INT64 * BigInt(-1) - BigInt(1);
+export const MAX_SINT64 = BigInt('9223372036854775807');
```
nazarhussain, line 23: For such larger numbers its easy to get confused. So better to add a comment here for reviewer as well. I appreciate If you add similar comment for other values as well. 

```suggestion
export const MAX_SINT64 = BigInt('9223372036854775807');  // 2 ^ (64 - 1) - 1
```

elements/lisk-validator/src/formats.ts
```diff
@@ -103,23 +92,9 @@ export const signedPublicKey = (data: string): boolean => {
...
-
-export const uint64 = (data: string): boolean =>
-	isNumberString(data) && isUint64(BigInt(data));
-
-export const uint32 = (data: string): boolean =>
-	isNumberString(data) && isUint32(BigInt(data));
-
-export const int32 = (data: string): boolean =>
-	isNumberString(data) && isInt32(BigInt(data));
-
```
nazarhussain, line 122: Why to remove these formats? With this change we can't specify these values as  `format` in the JSON schema.

elements/lisk-validator/src/keywords/data_type.ts
```diff
@@ -60,19 +69,27 @@ const compile = (
...
+		_data: Buffer | bigint | string | number,
 		_dataPath?: string,
 		_parentData?: object,
 		_parentDataProperty?: string | number,
 		_rootData?: object,
-	): boolean =>
+	): boolean => {
+		if (value === 'boolean') return isBoolean(_data);
+		if (value === 'bytes') return isBytes(_data as Buffer);
+		if (value === 'string') return isString(value);
```
nazarhussain, line 80:
```suggestion
		if (value === 'string') return isString(_data);
```

elements/lisk-validator/src/keywords/data_type.ts
```diff
@@ -60,19 +69,27 @@ const compile = (
 	}
 
 	return (
-		_data: string,
+		_data: Buffer | bigint | string | number,
```
nazarhussain, line 72: The name `_data` was used to by pass the limiting error of unused variable. You can rename it back to `data`.

```suggestion
		data: Buffer | bigint | string | number,
```

elements/lisk-validator/src/keywords/data_type.ts
```diff
@@ -60,19 +69,27 @@ const compile = (
...
+		if (value === 'bytes') return isBytes(_data as Buffer);
+		if (value === 'string') return isString(value);
+		if (value === 'uint32') return isUInt32(_data);
+		if (value === 'sint32') return isSInt32(_data);
+		if (value === 'uint64') return isSInt64(_data);
+		if (value === 'uint64') return isUInt64(_data);
+
 		// Either "dataType" or "type" can be presented in schema
-		!typePropertyPresent;
+		return !typePropertyPresent;
```
nazarhussain, line 87: You can remove this line as well, earlier this line had no impact as we had fixed validation value. Now since we had some business logic in validation function this line is no more required.

elements/lisk-validator/src/validation.ts
```diff
@@ -236,24 +214,30 @@ export const validateNetworkIdentifier = (
...
+export const isBoolean = (data: unknown): boolean => typeof data === 'boolean';
+
+export const isSInt32 = (data: unknown): boolean => {
+	if (typeof data === 'number' && Number.isInteger(data)) {
+		return data <= MAX_INT32 && data >= MIN_INT32;
 	}
 
-	return num <= BigInt(MAX_INT32) && num >= BigInt(MIN_INT32);
+	return false;
 };
```
nazarhussain, line 227: As now its `unknown`, should not we consider logic for `string` as well?

elements/lisk-validator/src/keywords/data_type.ts
nazarhussain: @ishantiw This was logically wrong. Pleas confirm you have unit test for this type, otherwise this should not pass.

elements/lisk-validator/src/formats.ts
ishantiw: I thought `format` is now only for strings? also, issue says, `Remove formats which are replaced by the new validation for these dataTypes` @shuse2 ?

elements/lisk-validator/src/validation.ts
ishantiw: you mean if data is `string`? that is being checked above `typeof data === 'number'`

elements/lisk-validator/src/keywords/data_type.ts
ishantiw: specifically this can pass because both `value` and `_data` are string :P so that's why it passed unit tests

elements/lisk-validator/src/formats.ts
nazarhussain: @ishantiw Not its just for strings can be used for numbers as well. @shuse2 The HTTP Plugin we will develop will be relying on JSON data and if we need to validate any input on earlier stage we need to keep these formats.

elements/lisk-validator/src/validation.ts
nazarhussain: Yes I meant if number can be in string format.

Issue #5211, opened 2020-04-28
Update commander to be network agnostic


shuse2:
### Description

Remove the functionalities that are specific to `lisk-core`.

### Motivation

Lisk Commander is part of the `lisk-sdk`, therefore it should be network agnostic and available as a tool for developers to use for their own blockchain applications.

### Sprint 21

- Improve validator in framework to separate default and validation #3815
- Replace controller/validator with lisk-validator #4610
- Update configurator not to use stdin and stdout #5330

### Sprint 24

- Add interface startAndListen to IPCChannel #5585
- Remove lisk-core specific commands and move to core #5245 

### Sprint 26

- ~Update `core` command to `node` command #5247~
- ~Add plugin feature to install custom application including `lisk-core` #5246~

### Additional Information

Related issues: https://github.com/LiskHQ/lisk-core/issues/254

PR #5388, opened 2020-06-04
Update commander to use new schema - Closes #5377

### LINKED ISSUE #5377: Update commander to use updated cryptography and block header schema
### Description

Update commander to use updated transaction schema and class functions

### Motivation

- Transaction schema and class were updated for the codec LIPs

### Acceptance Criteria

- All test should pass
- Input to the commander should be kept as string (not using buffer)

shuse2:
### What was the problem?

This PR resolves #5377 

### How was it solved?

- Temporally disable creation function, which will be move to core with #5211 
- Fix the usage of cryptography

### How was it tested?

- All test and linting should pass

### DIFF
--- a/commander/src/commands/account/show.ts
+++ b/commander/src/commands/account/show.ts
@@ -13,7 +13,11 @@
  * Removal or modification of this copyright notice is prohibited.
  *
  */
-import { getAddressFromPublicKey, getKeys } from '@liskhq/lisk-cryptography';
+import {
+	getAddressFromPublicKey,
+	getKeys,
+	getBase32AddressFromPublicKey,
+} from '@liskhq/lisk-cryptography';
 import { flags as flagParser } from '@oclif/command';
 
 import BaseCommand from '../../base';
@@ -22,14 +26,24 @@ import { getPassphraseFromPrompt } from '../../utils/reader';
 
 const processInput = (
 	passphrase: string,
-): { privateKey: string; publicKey: string; address: string } => {
+): {
+	privateKey: string;
+	publicKey: string;
+	address: string;
+	binaryAddress: string;
+} => {
 	const { privateKey, publicKey } = getKeys(passphrase);
-	const address = getAddressFromPublicKey(publicKey);
+	const binaryAddress = getAddressFromPublicKey(publicKey);
+	const address = getBase32AddressFromPublicKey(
+		publicKey.toString('hex'),
+		'lsk',
+	);
 
 	return {
-		privateKey,
-		publicKey,
+		privateKey: privateKey.toString('base64'),
+		publicKey: publicKey.toString('base64'),
 		address,
+		binaryAddress: binaryAddress.toString('base64'),
 	};
 };
 
--- a/commander/src/commands/transaction/create/delegate.ts
+++ b/commander/src/commands/transaction/create/delegate.ts
@@ -20,6 +20,7 @@ import {
 import { isValidFee, isValidNonce } from '@liskhq/lisk-validator';
 import { flags as flagParser } from '@oclif/command';
 
+import { getAddressAndPublicKeyFromPassphrase } from '@liskhq/lisk-cryptography';
 import BaseCommand from '../../../base';
 import { ValidationError } from '../../../utils/error';
 import { flags as commonFlags } from '../../../utils/flags';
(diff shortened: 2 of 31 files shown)


commander/src/commands/account/show.ts
```diff
@@ -22,14 +26,24 @@ import { getPassphraseFromPrompt } from '../../utils/reader';
...
+		'lsk',
+	);
 
 	return {
-		privateKey,
-		publicKey,
+		privateKey: privateKey.toString('base64'),
+		publicKey: publicKey.toString('base64'),
 		address,
+		binaryAddress: binaryAddress.toString('base64'),
```
nazarhussain, line 46: Why do we need binary address in the output. And binary address in `base64` encoding makes it confusing. 

That' the output of the `account:create` command. I suggest only `address` is good enough to show there. 

```
[{
"passphrase":"because hunt remember blue ship close frog reveal also jaguar fiscal kidney",
"privateKey":"03SbeIOUEYOqPlJyNuMoHi4mXPrI3vYmm5xXUKuGrwlWHKCElmGwM+4syf+OaUgS5CgyswOzUhCpBnGpRuRpXA==",
"publicKey":"VhyghJZhsDPuLMn/jmlIEuQoMrMDs1IQqQZxqUbkaVw=",
"binaryAddress":"ScjH0f/rz4MOAB16P7CvYyvAB7Y=",
"address":"lsk4npo8k8gh8n2bcuzcqhcgom8ooqjzxhe3m9mj3"}]
```

commander/src/commands/transaction/create/delegate.ts
```diff
@@ -40,6 +41,11 @@ const processInputs = (
 	passphrase?: string,
 ) =>
 	registerDelegate({
+		senderPublicKey: passphrase
+			? getAddressAndPublicKeyFromPassphrase(passphrase).publicKey.toString(
+					'hex',
+			  )
+			: '',
```
nazarhussain, line 48: Should not we change definition of `registerDelegate` to accept public key as Buffer?

commander/src/commands/transaction/create/transfer.ts
```diff
@@ -55,9 +56,14 @@ const processInputs = (
 		nonce,
 		fee,
 		networkIdentifier,
-		recipientId: address,
+		senderPublicKey: passphrase
+			? getAddressAndPublicKeyFromPassphrase(passphrase).publicKey.toString(
+					'hex',
+			  )
+			: '',
```
nazarhussain, line 63: Same here

commander/src/commands/transaction/sign.ts
```diff
@@ -157,6 +158,6 @@ export default class SignCommand extends BaseCommand {
 			throw errors;
 		}
 
-		this.print(removeUndefinedValues(txInstance.toJSON() as object));
+		this.print(removeUndefinedValues({} as object));
```
nazarhussain, line 161: Remove undefined values from which object?

commander/src/utils/cryptography.ts
```diff
@@ -94,7 +94,9 @@ export const { getKeys } = cryptography;
 export const getAddressFromPublicKey = (
 	publicKey: string,
 ): { readonly address: string } => ({
-	address: cryptography.getAddressFromPublicKey(publicKey),
+	address: cryptography
+		.getAddressFromPublicKey(Buffer.from(publicKey, 'hex'))
+		.toString('hex'),
```
nazarhussain, line 99: Why to return `string` for address? This function can be redefined to use buffer as input/output.

commander/src/commands/account/show.ts
shuse2: If all the endpoint from the framework uses base64, wouldn't it be useful to know the binary address in base64 format?
At least while fixing the test, it was useful

nazarhussain: I would suggest, outside from framework/application using any interface http/api we use only base64 address not base64 encoding of binary address. Binary address should stay under layers of libraries and framework.

commander/src/commands/transaction/create/delegate.ts
shuse2: It should accept, but it's commented on the other side, but this will replaced by https://github.com/LiskHQ/lisk-core/issues/271, so didnt want to change it now.
Also, all the transaction creation functions was for frontend, but since we don't use JSON format, that needs to be redesigned or discontinued when lisk-desktop and mobile start to migrate to 5.0

commander/src/commands/transaction/sign.ts
shuse2: same as above, currently transaction cannot be signed by the JSON, and with migration, this needs to be re-written.
Also, transaction instance does not accept JSON value, so it cannot instantiate now 😢

commander/src/utils/cryptography.ts
shuse2: Same as above, this usage needs to be removed when migrating to core

PR #5401, opened 2020-06-05
Adjust codec interface to allow broader type - Closes #5395

### LINKED ISSUE #5395: Update lisk-codec interface type
### Expected behavior
encode, decode and addSchema should not show typescript error for the valid schema

### Actual behavior
Some of the valid schema is showing typescript error.
Also, the encode interface value should expect any object since it unknown object, as it requires property signature in the current way.

### Steps to reproduce
Add below schema to the interface
```
{
	title: 'Used hash onion',
	$id: '/node/forger/used_hash_onion',
	type: 'object',
	required: ['usedHashOnions'],
	properties: {
		usedHashOnions: {
			type: 'array',
			fieldNumber: 1,
			items: {
				type: 'object',
				required: ['address', 'count', 'height'],
				properties: {
					address: {
						dataType: 'bytes',
						fieldNumber: 1,
					},
					count: {
						dataType: 'uint32',
						fieldNumber: 2,
					},
					height: {
						dataType: 'uint32',
						fieldNumber: 3,
					},
				}
			},
		},
	},
}
```

shuse2:
### What was the problem?

This PR resolves #5395 

### How was it solved?

Made codec interface to accept generic object as it will be validate by validator

### How was it tested?

### DIFF
--- a/elements/lisk-chain/src/chain.ts
+++ b/elements/lisk-chain/src/chain.ts
@@ -12,7 +12,7 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
-import { codec, MinimalSchema } from '@liskhq/lisk-codec';
+import { codec, Schema } from '@liskhq/lisk-codec';
 import { KVStore } from '@liskhq/lisk-db';
 import {
 	BaseTransaction,
@@ -85,7 +85,7 @@ interface ChainConstructor {
 		readonly [key: number]: typeof BaseTransaction;
 	};
 	readonly registeredBlocks: {
-		readonly [key: number]: MinimalSchema;
+		readonly [key: number]: Schema;
 	};
 	readonly accountAsset: {
 		schema: object;
--- a/elements/lisk-chain/src/chain.ts
+++ b/elements/lisk-chain/src/chain.ts
@@ -12,7 +12,7 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
-import { codec, Schema } from '@liskhq/lisk-codec';
+import { codec, MinimalSchema } from '@liskhq/lisk-codec';
 import { KVStore } from '@liskhq/lisk-db';
 import {
 	BaseTransaction,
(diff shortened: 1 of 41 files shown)


elements/lisk-chain/src/chain.ts
```diff
@@ -12,7 +12,7 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
-import { codec, Schema } from '@liskhq/lisk-codec';
+import { codec, MinimalSchema } from '@liskhq/lisk-codec';
```
pablitovicente, line 15: Maybe `GenericSchema` or `BasicSchema` or `RootSchema` as I think this always represents a full schema? No strong opinion and can be left as is.

elements/lisk-codec/src/codec.ts
```diff
@@ -79,26 +80,30 @@ export const validateSchema = (schema: {
...
 	private readonly _compileSchemas: CompiledSchemas = {};
 
-	public addSchema(schema: Schema): boolean {
+	public addSchema(schema: MinimalSchema): boolean {
 		validateSchema(schema);
 
 		const schemaName = schema.$id;
-		this._compileSchemas[schemaName] = this._compileSchema(schema, [], []);
+		this._compileSchemas[schemaName] = this._compileSchema(
+			schema as Schema,
```
pablitovicente, line 88: At this point you can specify to `Schema` already or should also be `MinimalSchema`?

shuse2: I think it's already validated, so it should be `Schema` here

elements/lisk-codec/src/types.ts
```diff
@@ -21,13 +21,21 @@ export interface GenericObject {
...
+
 export interface Schema {
+	readonly $id: string;
+	readonly type: string;
+	readonly properties: object;
+}
+
+export interface ValidatedSchema {
 	readonly $id: string;
 	readonly $schema?: string;
```
nazarhussain, line 33: Logically validated schema must have `$schema`, otherwise its can't be validated against the lisk meta schema and fall back to Draft-07 of JSON-Schema, which will allow any type to be a root level type.

PR #5409, opened 2020-06-08
Add minLength and maxLength for bytes - Closes #5394

### LINKED ISSUE #5394: Add validation keyword for data type
### Description

for dataType string, add `maxLength/minLength`
for dataType bytes add `maxLength/minLength`

### Motivation

In address or string we want to add certain keyword to extend the validation

### Acceptance Criteria

Add unit test in lisk-validator to check if those check works

shuse2:
### What was the problem?

This PR resolves #5394 

### How was it solved?

- Update to include minLength and maxLength check for bytes
- string works without modification

### How was it tested?

- Added tests


elements/lisk-validator/test/lisk_validator_keywords.spec.ts
```diff
@@ -85,6 +85,320 @@ describe('validator keywords', () => {
 				]),
 			);
 		});
+
+		describe('dataType value validation', () => {
+			describe('string', () => {
+				it('should be return empty error if valid', () => {
```
ishantiw, line 91: "should ~~be~~ return an empty error if valid"

elements/lisk-validator/test/lisk_validator_keywords.spec.ts
```diff
@@ -85,6 +85,320 @@ describe('validator keywords', () => {
...
+							},
+						},
+						{ myProp: 32 },
+					);
+					expect(result).toHaveLength(1);
+				});
+			});
+
+			describe('bytes', () => {
+				it('should be return empty error if valid', () => {
```
ishantiw, line 143: "should ~~be~~ return an empty error if valid" and also change it in the first test of each test suite

elements/lisk-validator/test/lisk_validator_keywords.spec.ts
```diff
@@ -85,6 +85,320 @@ describe('validator keywords', () => {
...
+						{
+							...validSchema,
+							properties: { myProp: { dataType: 'sint64', fieldNumber: 1 } },
+						},
+						{ myProp: 32 },
+					);
+					expect(result).toHaveLength(1);
+				});
+
+				it('should be invalid if above uint64 range', () => {
```
ishantiw, line 390: "should be invalid if above **sint64** range"

elements/lisk-validator/src/keywords/data_type.ts
```diff
@@ -75,13 +80,40 @@ const compile = (
...
+		}
+		if (value === 'uint64') {
+			return isUInt64(data);
+		}
+		if (value === 'sint32') {
+			return isSInt32(data);
+		}
+		if (value === 'sint64') {
+			return isSInt64(data);
+		}
```
nazarhussain, line 116: This redundant code is one of the reasons I was not convinced of introducing a custom keyword instead of using standard `type` with custom formats.

elements/lisk-validator/src/keywords/data_type.ts
```diff
@@ -75,13 +80,40 @@ const compile = (
...
-		if (value === 'sint64') return isSInt64(data);
+		if (value === 'boolean') {
+			return isBoolean(data);
+		}
+		if (value === 'bytes') {
+			if (!isBytes(data as Buffer)) {
+				return false;
+			}
+			const parent = parentSchema as KVPair;
+			if (typeof parent.minLength === 'number') {
```
nazarhussain, line 91: When you can validate a keyword value with another keyword then its recommended to use `macro` instead of `compile`. But here is the mixup of both, never seen such a case, so not sure which is Ajv recomendation to handle this case. So we can keep it same for now.

elements/lisk-validator/src/keywords/data_type.ts
```diff
@@ -75,13 +80,40 @@ const compile = (
...
+				if ((data as Buffer).length < parent.minLength) {
+					return false;
```
nazarhussain, line 93: We have to check what error does this validation generates? A generic error like `dataType keyword error` will not guide user to troubleshoot what's wrong.

nazarhussain: Please add a test case for it as well.

shuse2: Test is here, but the return type expects boolean. How do we change the error message here?
https://github.com/LiskHQ/lisk-sdk/pull/5409/files#diff-712c76f4454b085e696bad5b60ddf1f8R175

nazarhussain: Change `errors: 'full'` and then you can return any custom error message. 

https://github.com/LiskHQ/lisk-sdk/blob/e13591fa7dcc18a93709eac8f9f60c0c6d0aae5f/elements/lisk-validator/src/keywords/data_type.ts#L125

PR #5412, opened 2020-06-09
Remove time offset for the slot calculation and the config setting - Closes #5280

### LINKED ISSUE #5280: Replace epochTime with genesis block timestamp
### Description

- Remove the `epoch time` from the config
- Update the slot calculation in `lisk-chain` to use `genesis block timestamp` instead of epoch time. 

The slot number will be calculated as: 

```
Math.floor((slotTime - genesisBlock.timestamp) / this._interval);
```

Where `slotTime` can be any unix time and interval would be the block time which is default to 10s. 

### Motivation

- With the new [LIP](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0034.md), the epoch time difference is removed.

### Acceptance Criteria

- All tests are updated not to use the epoch time

### Additional Information

The related information can be found on https://github.com/LiskHQ/lips/blob/master/proposals/lip-0034.md#block-processing

> This way, the genesis block timestamp defines the block slots of the blockchain as `[b.header.timestamp+BLOCK_SLOT_LENGTH, b.header.timestamp + 2*BLOCK_SLOT_LENGTH)`, `[b.header.timestamp + 2*BLOCK_SLOT_LENGTH, b.header.timestamp + 3*BLOCK_SLOT_LENGTH)`, and so on.

Related issues #5234 

- lisk-chain/src/slots.ts contains the main logic of the epochTime usage
- epochTime needs to be removed from framework/src/application/schema/application_config_schema.ts, and all other usages

pablitovicente:
### What was the problem?

This PR resolves #5280

### How was it solved?

By removing epochTime usages from all libraries and framework

### How was it tested?

Updated Unit Tests

### DIFF
--- a/elements/lisk-chain/src/slots.ts
+++ b/elements/lisk-chain/src/slots.ts
@@ -35,9 +35,9 @@ export class Slots {
 	}
 
 	public getSlotNumber(time?: number): number {
-		const parsedEpochTime =
+		const slotTime =
 			time === undefined ? Math.floor(Date.now() / SEC_IN_MS) : time;
-		return Math.floor(parsedEpochTime / this._interval);
+		return Math.floor(slotTime / this._interval);
 	}
 
 	public getSlotTime(slot: number): number {
--- a/elements/lisk-chain/src/slots.ts
+++ b/elements/lisk-chain/src/slots.ts
@@ -13,36 +13,26 @@
  */
 
 interface SlotsInput {
-	readonly epochTime: string;
 	readonly interval: number;
 }
 
 const SEC_IN_MS = 1000;
 
 export class Slots {
-	private readonly _epochTime: Date;
 	private readonly _interval: number;
 
-	public constructor({ epochTime, interval }: SlotsInput) {
-		this._epochTime = new Date(epochTime);
+	public constructor({ interval }: SlotsInput) {
 		this._interval = interval;
 	}
 
-	public getEpochTime(): number {
-		return Math.floor((Date.now() - this._epochTime.getTime()) / SEC_IN_MS);
-	}
-
+	// eslint-disable-next-line class-methods-use-this
 	public getRealTime(time: number): number {
-		return (
-			Math.floor(this._epochTime.getTime() / SEC_IN_MS) * SEC_IN_MS +
-			time * SEC_IN_MS
-		);
+		return time * SEC_IN_MS;
 	}
 
-	public getSlotNumber(epochTime?: number): number {
+	public getSlotNumber(time?: number): number {
 		const parsedEpochTime =
-			epochTime === undefined ? this.getEpochTime() : epochTime;
-
+			time === undefined ? Math.floor(Date.now() / SEC_IN_MS) : time;
 		return Math.floor(parsedEpochTime / this._interval);
 	}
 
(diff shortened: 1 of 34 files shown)


elements/lisk-chain/src/slots.ts
```diff
@@ -13,40 +13,30 @@
...
-	public getEpochTime(): number {
-		return Math.floor((Date.now() - this._epochTime.getTime()) / SEC_IN_MS);
-	}
-
 	public blockTime(): number {
 		return this._interval;
 	}
 
+	// eslint-disable-next-line class-methods-use-this
 	public getRealTime(time: number): number {
```
shuse2, line 33: I think we can remove this function as it doesn't make sense anymore

elements/lisk-chain/src/slots.ts
```diff
@@ -13,40 +13,30 @@
...
-		return (
-			Math.floor(this._epochTime.getTime() / SEC_IN_MS) * SEC_IN_MS +
-			time * SEC_IN_MS
-		);
+		return time * SEC_IN_MS;
 	}
 
-	public getSlotNumber(epochTime?: number): number {
+	public getSlotNumber(time?: number): number {
 		const parsedEpochTime =
```
shuse2, line 38: Maybe rename this variable name to be `slotTime`?

pablitovicente: Done

elements/lisk-chain/src/slots.ts
pablitovicente: It's only used in one place so far but I thought of leaving it to avoid making a different calculation if needs to be used somewhere else in the future but ok to remove. So should I keep it with a better name in case is used somewhere else in the future or just remove directly and when and if the time comes we create a new method in the future? @shuse2

shuse2: I think since we use unixtime, and the time is just time in sec, we don't need to keep this function.
Also, the only usage in the forger.ts can be updated to just multiply by 1000.

pablitovicente: Yes I was just considering if we might need to use it somewhere else even although is trivial but will remove.

elements/lisk-chain/src/slots.ts
```diff
@@ -13,41 +13,31 @@
...
 
-	public getSlotNumber(epochTime?: number): number {
-		const parsedEpochTime =
-			epochTime === undefined ? this.getEpochTime() : epochTime;
-
-		return Math.floor(parsedEpochTime / this._interval);
+	public getSlotNumber(time?: number): number {
+		const slotTime =
+			time === undefined ? Math.floor(Date.now() / SEC_IN_MS) : time;
+		return Math.floor(slotTime / this._interval);
```
nazarhussain, line 40: A similar change has to be applied in another method `getSlotTime`

elements/lisk-chain/src/slots.ts
```diff
@@ -13,41 +13,31 @@
...
 
 const SEC_IN_MS = 1000;
 
 export class Slots {
-	private readonly _epochTime: Date;
 	private readonly _interval: number;
 
-	public constructor({ epochTime, interval }: SlotsInput) {
-		this._epochTime = new Date(epochTime);
+	public constructor({ interval }: SlotsInput) {
```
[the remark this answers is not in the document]
pablitovicente: I was discussing this previous days but I was told this was not required. I'm a bit confused about how this should work now :/

pablitovicente:
Closing this as original issue didn't take into account some LIP details. Will start a new one.

PR #5414, opened 2020-06-09
Remove configurator and validator - Closes #5330

### LINKED ISSUE #3815: Improve validator in framework to separate default and validation
### Expected behavior
TBD

### LINKED ISSUE #4610: Replace controller/validator with lisk-validator
### Expected behavior
We should use "lisk-validator" library consistently. We should also investigate the tests that are using validator and see if the stubbing or spying were done correctly.
## Actual behavior
During fixing issue #4026 I realized the `framework/test/jest/unit/specs/controller/helpers/validator/validator.spec.js` was giving false-positive results. 

For example, since Ajv class is being mocked and checked the beforeEach hook should like below:
image

After introducing the change above we realized the tests were not giving meaningful results since ajv itself was mocked.

Also, https://github.com/LiskHQ/lisk-sdk/pull/4603/files#diff-9d26ac4e7d17088447923c52897aa720R150 here the `validator.validate` function was stubbed, therefore try/catch statements were not throwing any error and giving false-positive results.

3.0.0+

### LINKED ISSUE #5330: Update configurator not to use stdin and stdout
### Description

- Update to TypeScript
- Remove use of `env` and `arg`
- Remove reading of the `stdin` and writing to `stdout`

### Motivation

Configurator needs an update to support https://github.com/LiskHQ/lisk-core/issues/254
- Remove `env` and `arg` because they are read from the core commands
- Remove use of `stdin` and `stdout` because they are used by core commands

### Acceptance Criteria

- When registering plugin, schema is registered to the configurator
- `yargs` is not used
- `lodash` is not used

shuse2:
### What was the problem?

This PR resolves #5330, resolves #4610 and resolves #3815 

### How was it solved?

- Remove validator and use `lisk-validator` consistently
- Add mergeDeep to handle default injection, which was handled in the validator
- Remove yarg, which is taking over stdin that needs to be handled in core

### How was it tested?

- All test is updated
- cd sdk; yarn link; cd framework; yarn link
- TBD


commander/tsconfig.json
```diff
@@ -2,7 +2,6 @@
 	"extends": "../tsconfig.json",
 	"compilerOptions": {
 		"composite": true,
-		"declaration": true,
```
nazarhussain, line 5: Why we had to remove this option?

elements/lisk-cryptography/tsconfig.json
```diff
@@ -1,7 +1,6 @@
 {
 	"extends": "../../tsconfig",
 	"compilerOptions": {
-		"declaration": true,
```
nazarhussain, line 4: Same here.

elements/lisk-validator/src/formats.ts
```diff
@@ -125,3 +127,41 @@ const camelCaseRegex = /^[a-z]+((\d)|([A-Z0-9][a-zA-Z0-9]+))*([a-z0-9A-Z])?$/;
 
 export const camelCase = (data: string): boolean =>
 	camelCaseRegex.exec(data) !== null;
+
+export const version = (data: string): boolean =>
+	/^([0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3})(-(alpha|beta|rc)\.[0-9]{1,3}(\.[0-9]{1,3})?)?$/.test(
```
nazarhussain, line 132: I suggest this regular expression should be flexible and just follow the semantic versioning standard. As the version of the application for any custom application can be different and they can have any prefix than alpha, beta, rc.

elements/lisk-validator/src/formats.ts
```diff
@@ -125,3 +127,41 @@ const camelCaseRegex = /^[a-z]+((\d)|([A-Z0-9][a-zA-Z0-9]+))*([a-z0-9A-Z])?$/;
...
+		data,
+	);
+
+export const protocolVersion = (data: string): boolean =>
+	/^(\d|[1-9]\d{1,2})\.(\d|[1-9]\d{1,2})$/.test(data);
+
+export const path = (data: string): boolean => {
+	const pathRegExp = new RegExp('^(.?)(/[^/]+)+$');
```
nazarhussain, line 140: Unescaped forward slash in this regex. 

image

This does not work with paths having a forward slash at the end, which is quite common for a directory path. 

```
/nazar/alpha/beta/
```

elements/lisk-validator/src/formats.ts
```diff
@@ -125,3 +127,41 @@ const camelCaseRegex = /^[a-z]+((\d)|([A-Z0-9][a-zA-Z0-9]+))*([a-z0-9A-Z])?$/;
...
 export const camelCase = (data: string): boolean =>
 	camelCaseRegex.exec(data) !== null;
+
+export const version = (data: string): boolean =>
+	/^([0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3})(-(alpha|beta|rc)\.[0-9]{1,3}(\.[0-9]{1,3})?)?$/.test(
+		data,
+	);
+
+export const protocolVersion = (data: string): boolean =>
+	/^(\d|[1-9]\d{1,2})\.(\d|[1-9]\d{1,2})$/.test(data);
```
nazarhussain, line 137: I think you want to it to be 2 digit protocol versions as `00.00`, but this expression will match 3 digits in each group.

elements/lisk-validator/src/formats.ts
```diff
@@ -125,3 +127,41 @@ const camelCaseRegex = /^[a-z]+((\d)|([A-Z0-9][a-zA-Z0-9]+))*([a-z0-9A-Z])?$/;
...
+	const keyRegExp = /[a-zA-Z0-9]{2,15}/;
+	const valueRegExp = /[a-f0-9]{1,256}/;
+	const keyValueRegExp = new RegExp(
+		`${keyRegExp.source}=${valueRegExp.source}`,
+	);
+	const encryptedPassphraseRegExp = new RegExp(
+		`^(${keyValueRegExp.source})(?:&(${keyValueRegExp.source})){0,10}$`,
+	);
+	return encryptedPassphraseRegExp.test(data);
+};
```
nazarhussain, line 160: This does not match with required dataset. 

image


Try this one

```
[a-z0-9A-z]*=[^&]+&*
```
image

framework/src/application/application.ts
```diff
@@ -30,21 +30,24 @@ import {
...
+	applicationConfigSchema,
+} from './schema';
 import { ApplicationState } from './application_state';
 import { Network } from './network';
 import { Node } from './node';
 import { InMemoryChannel } from '../controller/channels';
 import { Logger, createLogger } from './logger';
 
-import { DuplicateAppInstanceError } from '../errors';
+import { DuplicateAppInstanceError, SchemaValidationError } from '../errors';
```
nazarhussain, line 50: Can we replace this `SchemaValidationError` with `lisk-validator` error `LiskValidationError`?

framework/src/application/application.ts
```diff
@@ -30,21 +30,24 @@ import {
...
 import { version } from '../version';
-import * as validator from './validator';
-import * as configurator from './default_configurator';
-import { genesisBlockSchema, constantsSchema } from './schema';
+import {
+	genesisBlockSchema,
+	constantsSchema,
+	mergeDeep,
+	applicationConfigSchema,
+} from './schema';
```
nazarhussain, line 43: `mergeDeep` should not be in the `schema` file.

framework/src/application/application.ts
```diff
@@ -102,34 +105,34 @@ export interface ApplicationConfig {
 		enabled: boolean;
 	};
 	rootPath: string;
-	readonly forging: {
-		readonly waitThreshold: number;
-		readonly delegates: DelegateConfig[];
-		readonly force?: boolean;
-		readonly defaultPassword?: string;
+	forging: {
```
nazarhussain, line 108: Can we move this `interface ApplicationConfig` to `types.ts`?

framework/src/application/application.ts
```diff
@@ -156,38 +159,45 @@ export class Application {
...
 
 		appConfig.label =
-			appConfig.label ??
+			config.label ??
 			`lisk-${this._genesisBlock.header.transactionRoot.slice(0, 7)}`;
 
-		appConfig = configurator.getConfig(appConfig, {
-			failOnInvalidArg: process.env.NODE_ENV !== 'test',
-		}) as ApplicationConfig;
+		mergeDeep(appConfig, config);
```
nazarhussain, line 176: If its lodahs merge, I would prefer to reassign values. 

```suggestion
appConfig = mergeDeep({}, appConfig, config);
```

framework/src/application/application.ts
```diff
@@ -156,38 +159,45 @@ export class Application {
...
+		if (applicationConfigErrors.length) {
+			throw new Error(applicationConfigErrors.map(e => e.message).join(','));
+		}
 
-		// These constants are readonly we are loading up their default values
-		// In additional validating those values so any wrongly changed value
-		// by us can be catch on application startup
-		// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-		const constants = validator.parseEnvArgAndValidate(constantsSchema, {});
+		const constantError = validator.validate(
```
nazarhussain, line 186:
```suggestion
		const constantErrors = validator.validate(
```

framework/src/application/application.ts
```diff
@@ -156,38 +159,45 @@ export class Application {
...
+			constantsSchema,
+			constantsSchema.default,
+		);
+		if (constantError.length) {
```
nazarhussain, line 190:
```suggestion
		if (constantErrors.length) {
```

framework/src/application/application.ts
```diff
@@ -156,38 +159,45 @@ export class Application {
...
+		}
 
-		// These constants are readonly we are loading up their default values
-		// In additional validating those values so any wrongly changed value
-		// by us can be catch on application startup
-		// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-		const constants = validator.parseEnvArgAndValidate(constantsSchema, {});
+		const constantError = validator.validate(
+			constantsSchema,
+			constantsSchema.default,
```
nazarhussain, line 188: Why there will be an error for default values?

framework/src/application/application.ts
```diff
@@ -256,7 +266,13 @@ export class Application {
 			`A transaction type "${Transaction.TYPE}" is already registered.`,
 		);
 
-		validator.validate(transactionInterface, Transaction.prototype);
+		const schemaError = validator.validate(
```
nazarhussain, line 269:
```suggestion
		const transactionSchemaErrors = validator.validate(
```

framework/src/application/application.ts
```diff
@@ -256,7 +266,13 @@ export class Application {
...
+			transactionInterface,
+			Transaction.prototype,
+		);
+		if (schemaError.length) {
```
nazarhussain, line 273:
```suggestion
		if (transactionSchemaErrors.length) {
```

framework/src/application/application.ts
```diff
@@ -256,7 +266,13 @@ export class Application {
...
+			throw new SchemaValidationError((schemaError as unknown) as Error[]);
```
nazarhussain, line 274:
```suggestion
			throw new SchemaValidationError((transactionSchemaErrors as unknown) as Error[]);
```

framework/src/application/schema/application_config_schema.ts
```diff
@@ -79,16 +79,13 @@ export const applicationConfigSchema = {
 				fileLogLevel: {
 					type: 'string',
 					enum: ['trace', 'debug', 'info', 'warn', 'error', 'fatal', 'none'],
-					env: 'LISK_FILE_LOG_LEVEL',
-					arg: '--log,-l',
```
nazarhussain, line 83: Can you refer to the issue which will replace this feature? meant passing command-line options by argument or env variables in lisk-core.

framework/src/application/schema/constants_schema.ts
```diff
@@ -23,8 +23,7 @@ export const constantsSchema = {
 	],
 	properties: {
 		activeDelegates: {
-			type: 'number',
-			format: 'oddInteger',
```
nazarhussain, line 27: Why are we removing this behavior?

framework/src/application/schema/index.ts
```diff
@@ -16,3 +16,4 @@
 export { applicationConfigSchema } from './application_config_schema';
 export { constantsSchema } from './constants_schema';
 export { genesisBlockSchema } from './genesis_block_schema';
+export * from './merge_deep';
```
nazarhussain, line 19: Please move this to some utility file.

framework/src/application/schema/merge_deep.ts
```diff
@@ -0,0 +1,40 @@
...
+
+interface KVPair {
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	[key: string]: any;
+}
+
+const isObject = (item: unknown): item is KVPair =>
+	typeof item === 'object' && item !== null && !Array.isArray(item);
+
+export const mergeDeep = (dest: KVPair, src: KVPair): void => {
```
nazarhussain, line 23: I would suggest using `lodash.merge` package. We already using similar for `lodash.cloneDeep`. I think a merge is not that simple otherwise lodash would not have hundred of lines for this.

framework/src/controller/bus.ts
```diff
@@ -250,6 +248,7 @@ export class Bus extends EventEmitter2 {
 		if (this.config.ipc.enabled) {
 			this.subSocket?.on(eventName, cb);
 		}
+		return this;
```
nazarhussain, line 251: Do we want function chainning on this function?

framework/src/index.ts
```diff
@@ -12,10 +12,8 @@
...
 
-import * as configurator from './application/default_configurator';
-
-export { Application } from './application/application';
+export { Application, ApplicationConfig } from './application/application';
 export { version } from './version';
 export { systemDirs } from './application/system_dirs';
 export { BaseModule } from './modules/base_module';
-export { configurator };
+export { GenesisBlockJSON } from './application/node/node';
```
nazarhussain, line 19: Can we report this change to the documentation team @Tschakki, I think they may have to update examples in the documentation.

framework/test/test_app/app.js
```diff
@@ -28,44 +25,28 @@ const dummyBuildVersion = '#buildVersion';
...
-
-	// Support for PROTOCOL_VERSION only for tests
-	if (process.env.NODE_ENV === 'test' && process.env.PROTOCOL_VERSION) {
-		config.protocolVersion = process.env.PROTOCOL_VERSION;
-	}
-
+	const mergedConfig = {
+		...appConfig,
+		...config,
+	};
```
nazarhussain, line 47: Object spread does not work with deep merge, you have to use some utility function for it.

commander/tsconfig.json
shuse2: It is defined in `"extends": "../tsconfig.json",` now, as framework was not completely typescript, it was defined in each one

framework/src/application/application.ts
shuse2: That's what it was checking in the current implementation

elements/lisk-validator/src/formats.ts
shuse2: these regex was there in the application validator, and just moved to the common validator.
Let's create another issue to update the regex

framework/src/application/schema/application_config_schema.ts
shuse2: It's in LiskHQ/lisk-core#278

framework/src/controller/bus.ts
shuse2: this was required because it's extending the EventEmitter2, and signature has to be exactly the same

framework/test/test_app/app.js
shuse2: in this case, it will work because appConfig only have first layer, and config will overwrite everything else

framework/src/application/schema/merge_deep.ts
shuse2: I would prefer not to include the extra package if the implementation is as simple like this

elements/lisk-validator/src/formats.ts
nazarhussain: https://github.com/LiskHQ/lisk-sdk/issues/5416

framework/src/application/application.ts
nazarhussain: In old implementation, it was used `parseEnvArgAndValidate` which can change the values depending on defined env variables or command line arguments. So why we were doing validation on top of it. 

Here I don't thin its required anymore, as these are just hardcoded values we already know.

PR #5415, opened 2020-06-09
Take the diff from the StateStore of lisk-chain and store them in the database - Closes #5271 #5428

### LINKED ISSUE #5271: Take the diff from the StateStore of lisk-chain and store them in the database
### Description

- Update StateStore to save the initial value
- While finalizing, diff should be taken against the initial value
- If the value is the same, it should not be in the diff
- Define the schema of the diff as proposed below
- Database key should be `diff:${height}`

```js
{
  updated: { key: string, value: Diff }[],
  created: string[], // store the key
}
```

### Motivation

In order to provide automatic undo functionality, diffs from the StateStore of `lisk-chain` should be stored in the database.

### Acceptance Criteria

- Add unit test to cover scenario to check calculated diff can be used to go back to the previous state using `undo` function
- Check the case of newly created value
- Check the case of the same value
- Add integration test for the database

### Additional Information

Related issues: #5232

ManuGowda:
### What was the problem?

This PR resolves #5271 

### How was it solved?

- Updated StateStore to save the initial value
- Take diff against the initial value
- Do not save if no changes
- Use schema for encoding and decoding
- Use database key `diff:${height}`

### How was it tested?

- Added unit and integration test

### DIFF
--- a/elements/lisk-chain/src/chain.ts
+++ b/elements/lisk-chain/src/chain.ts
@@ -67,6 +67,7 @@ import {
 	signingBlockHeaderSchema,
 	baseAccountSchema,
 	blockHeaderSchema,
+	stateDiffSchema,
 } from './schema';
 
 interface ChainConstructor {
@@ -167,6 +168,7 @@ export class Chain {
 			},
 		};
 		codec.addSchema(accountSchema as Schema);
+		codec.addSchema(stateDiffSchema);
 		this._defaultAccountAsset = accountAsset.default;
 
 		this.dataAccess = new DataAccess({
--- a/elements/lisk-chain/src/chain.ts
+++ b/elements/lisk-chain/src/chain.ts
@@ -167,7 +167,7 @@ export class Chain {
 				},
 			},
 		};
-		codec.addSchema(accountSchema as Schema);
+		codec.addSchema(accountSchema);
 		codec.addSchema(stateDiffSchema);
 		this._defaultAccountAsset = accountAsset.default;
 
@@ -175,7 +175,7 @@ export class Chain {
 			db,
 			registeredBlockHeaders: registeredBlocks,
 			registeredTransactions,
-			accountSchema: accountSchema as Schema,
+			accountSchema,
 			minBlockHeaderCache,
 			maxBlockHeaderCache,
 		});
--- a/elements/lisk-chain/src/state_store/account_store.ts
+++ b/elements/lisk-chain/src/state_store/account_store.ts
@@ -38,7 +38,7 @@ export class AccountStore {
 	private readonly _defualtAsset: object;
 	private readonly _primaryKey = 'address';
 	private readonly _name = 'Account';
-	private _initialAccountValue = {} as Account;
+	private readonly _initialAccountValue: BufferMap<Account>;
 
 	public constructor(
 		dataAccess: DataAccess,
@@ -52,6 +52,7 @@ export class AccountStore {
 		this._originalData = new BufferMap();
 		this._originalUpdatedKeys = new BufferSet();
 		this._defualtAsset = additionalInformation.defaultAsset;
+		this._initialAccountValue = new BufferMap<Account>();
 	}
 
 	public createSnapshot(): void {
@@ -84,7 +85,10 @@ export class AccountStore {
 		// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
 		if (elementFromDB) {
 			this._data.set(primaryValue, elementFromDB as Account);
-			this._initialAccountValue = elementFromDB as Account;
+			this._initialAccountValue.set(
+				elementFromDB.address,
+				elementFromDB as Account,
+			);
 
 			return (new Account(elementFromDB) as unknown) as Account<T>;
 		}
@@ -112,7 +116,10 @@ export class AccountStore {
 				primaryValue,
 			);
 			this._data.set(primaryValue, elementFromDB as Account);
-			this._initialAccountValue = elementFromDB as Account;
+			this._initialAccountValue.set(
+				elementFromDB.address,
+				elementFromDB as Account,
+			);
 
 			return (new Account(elementFromDB as Account) as unknown) as Account<T>;
 		} catch (error) {
@@ -149,21 +156,26 @@ export class AccountStore {
 		for (const account of this._data.values()) {
 			if (this._updatedKeys.has(account.address)) {
 				const encodedFinalAccount = this._dataAccess.encodeAccount(account);
-				const dbKey = `${DB_KEY_ACCOUNTS_ADDRESS}:${keyString(account.address)}`;
-				batch.put(
-					dbKey,
-					encodedFinalAccount,
-				);
+				const dbKey = `${DB_KEY_ACCOUNTS_ADDRESS}:${keyString(
+					account.address,
+				)}`;
+				batch.put(dbKey, encodedFinalAccount);
 
-				if (Object.entries(this._initialAccountValue).length) {
-					const encodedInitialAccount = this._dataAccess.encodeAccount(this._initialAccountValue);
-					const diff = calculateDiff(encodedInitialAccount, encodedFinalAccount);
+				if (this._initialAccountValue.has(account.address)) {
+					const initialAccount = this._initialAccountValue.get(account.address);
+					const encodedInitialAccount = this._dataAccess.encodeAccount(
+						initialAccount as Account,
+					);
+					const diff = calculateDiff(
+						encodedInitialAccount,
+						encodedFinalAccount,
+					);
 					stateDiff.updated.push({
 						key: dbKey,
 						value: diff,
 					});
 				} else {
-					stateDiff.created.push(dbKey)
+					stateDiff.created.push(dbKey);
 				}
 			}
 		}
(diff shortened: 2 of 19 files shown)


elements/lisk-chain/src/state_store/account_store.ts
```diff
@@ -36,6 +38,7 @@ export class AccountStore {
 	private readonly _defualtAsset: object;
 	private readonly _primaryKey = 'address';
 	private readonly _name = 'Account';
+	private _initialAccountValue = {} as Account;
```
shuse2, line 41: This only stores 1 account. It should be `BufferMap`

elements/lisk-chain/src/state_store/chain_state_store.ts
```diff
@@ -37,6 +38,7 @@ export class ChainStateStore {
 	private readonly _lastBlockHeader: BlockHeader;
 	private readonly _networkIdentifier: Buffer;
 	private readonly _lastBlockReward: bigint;
+	private _initialValue = Buffer.alloc(0);
```
shuse2, line 41: same as above, this should be `BufferMap`

elements/lisk-chain/src/state_store/account_store.ts
```diff
@@ -137,15 +143,31 @@ export class AccountStore {
...
+				const dbKey = `${DB_KEY_ACCOUNTS_ADDRESS}:${keyString(account.address)}`;
 				batch.put(
-					`${DB_KEY_ACCOUNTS_ADDRESS}:${keyString(account.address)}`,
-					encodedAccount,
+					dbKey,
+					encodedFinalAccount,
 				);
+
+				if (Object.entries(this._initialAccountValue).length) {
+					const encodedInitialAccount = this._dataAccess.encodeAccount(this._initialAccountValue);
```
shuse2, line 159: Not specifically related this this PR, but probably it wouldve been better if the storage returns raw Buffer, and decoding was done in this file, so that this `encode` doesnt need to happen

elements/lisk-chain/src/state_store/state_store.ts
```diff
@@ -60,9 +63,34 @@ export class StateStore {
...
+		const diffToEncode = stateDiffs.reduce(
+			(acc, val) => {
+				acc.updated.push(...val.updated);
+				acc.created.push(...val.created);
+				return acc;
+			},
+			{ updated: [], created: [] },
+		);
+
+		codec.addSchema(stateDiffSchema);
```
shuse2, line 92: this `addSchema` should only happen once. It should be moved to somewhere that is doing other schema

elements/lisk-chain/src/state_store/chain_state_store.ts
```diff
@@ -48,6 +50,7 @@ export class ChainStateStore {
 		this._lastBlockReward = additionalInformation.lastBlockReward;
 		this._data = {};
 		this._originalData = {};
+		this._initialValue = {};
```
shuse2, line 53: This also should be BufferMap?

elements/lisk-chain/src/state_store/consensus_state_store.ts
```diff
@@ -42,6 +44,7 @@ export class ConsensusStateStore {
 		this._lastBlockHeaders = additionalInformation.lastBlockHeaders;
 		this._data = {};
 		this._originalData = {};
+		this._initialValue = {};
```
shuse2, line 47: same as above

elements/lisk-chain/src/state_store/account_store.ts
```diff
@@ -64,65 +68,75 @@ export class AccountStore {
...
+		const accountFromDB = await this._dataAccess.getAccountByAddress(
+			address,
 		);
 
 		// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
-		if (elementFromDB) {
-			this._data.set(primaryValue, elementFromDB as Account);
+		if (accountFromDB) {
+			this._data.set(address, accountFromDB as Account);
+			this._initialAccountValue.set(
```
shuse2, line 88: can you create an issue to add getAccount without decoding, and decode it here?

elements/lisk-chain/test/integration/state_store/save_diff.spec.ts
```diff
@@ -0,0 +1,326 @@
...
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+/* eslint-disable @typescript-eslint/restrict-template-expressions */
+import * as path from 'path';
```
shuse2, line 15: does this need to be integration test?

ManuGowda: for db we need the file path know.

shuse2: i meant this whole test suite. does it need to be tested with compatibility of database?
I thought it is enough to mock the DB and test the behavior

shuse2: As discussed, no other test covers diff saving into the database, so it is valid to be in integration

elements/lisk-chain/src/state_store/account_store.ts
ManuGowda: https://github.com/LiskHQ/lisk-sdk/issues/5428

elements/lisk-chain/src/state_store/account_store.ts
```diff
@@ -63,66 +67,69 @@ export class AccountStore {
...
-		if (element) {
-			return (new Account(element) as unknown) as Account<T>;
+		if (cachedAccount) {
+			return (new Account(cachedAccount) as unknown) as Account<T>;
 		}
 
 		// Account was not cached previously so we try to fetch it from db
-		const elementFromDB = await this._dataAccess.getAccountByAddress(
-			primaryValue,
+		const encodedAccount = await this._dataAccess.getEncodedAccountByAddress(
```
shuse2, line 79: This throws if the account does not exist, so no need to check below

elements/lisk-chain/src/state_store/account_store.ts
```diff
@@ -137,15 +144,38 @@ export class AccountStore {
...
+				} else {
+					stateDiff.created.push(dbKey);
+				}
 			}
 		}
+
+		return stateDiff;
+	}
+
+	private _getAccountInstance<T>(encodedAccount: Buffer): Account {
```
shuse2, line 177:
```suggestion
	private _getAccountInstance<T>(encodedAccount: Buffer): Account<T> {
```

elements/lisk-chain/src/chain.ts
```diff
@@ -167,6 +168,7 @@ export class Chain {
 			},
 		};
 		codec.addSchema(accountSchema as Schema);
```
ishantiw, line 170: I think you can remove this typecasting `as Schema`

ishantiw: also at line https://github.com/LiskHQ/lisk-sdk/pull/5415/files#diff-d84e8ad8c9682409e75b789bc2b45f95R178

elements/lisk-chain/src/state_store/account_store.ts
```diff
@@ -44,11 +45,10 @@ export class AccountStore {
 		this._dataAccess = dataAccess;
 		this._data = new BufferMap<Account>();
 		this._updatedKeys = new BufferSet();
-		this._primaryKey = 'address';
-		this._name = 'Account';
 		this._originalData = new BufferMap();
 		this._originalUpdatedKeys = new BufferSet();
 		this._defualtAsset = additionalInformation.defaultAsset;
```
ishantiw, line 50: `_defualtAsset ` -> `_defaultAsset `

elements/lisk-chain/src/state_store/state_store.ts
```diff
@@ -60,9 +63,35 @@ export class StateStore {
...
+		const consensusStateDiff = this.consensus.finalize(batch);
+		this._saveDiff(
+			height,
+			[accountStateDiff, chainStateDiff, consensusStateDiff],
+			batch,
+		);
+	}
+
+	// eslint-disable-next-line class-methods-use-this
+	private _saveDiff(
```
ishantiw, line 78: why not make it a function within this file?

elements/lisk-chain/test/integration/state_store/save_diff.spec.ts
```diff
@@ -0,0 +1,256 @@
...
+			const batch = db.batch();
+
+			// Act
+			stateStore.finalize(fakeHeight, batch);
+			await batch.write();
+			const diff = await db.get(`${DB_KEY_DIFF_STATE}:${fakeHeight}`);
+			const decodedDiff = codec.decode(stateDiffSchema, diff);
+
+			// Assert
+			expect(decodedDiff).toStrictEqual({
```
ishantiw, line 118: Can we also include `undo` in this test, we haven't checked whether the diff generated can be used to rollback to previous state. so can we check `previousValue = undo(currentValue, decodedDiff)` if its too much else we could cover in another PR. Same for chain and consensus state tests

PR #5417, opened 2020-06-10
Add buffer library and inject into client package - Closes #5305

### LINKED ISSUE #5305: buffer.readBigUint64BE is not a function
### Expected behavior
Transaction and Cryptography library should be able to use in the front-end application

### Actual behavior
buffer.readBigUint64BE is used to convert buffer into bigint in `lisk-cryptography:bufferToIntAsString`. 
which is used to calculate transactionId in the `lisk-transactions` and `getAddressFromPublicKey` in `lisk-cryptography`.

Currently Frontend is using https://www.npmjs.com/package/buffer, but this library does not support the new `readBigUint64BE` and `writeBigUint64BE` functions

### Steps to reproduce
Try to create transaction in the frontend

4.0-

shuse2:
### What was the problem?

This PR resolves #5305 

### How was it solved?

Add browser version of Buffer which have bigint implementation temporally 

### How was it tested?

Open index.html in `elements/lisk-client/index.html` after build, and check function that uses bigint works in browser
```
lisk.cryptography.getAddressFromPassphrase('123')
```

### DIFF
--- a/elements/lisk-client/index.html
+++ b/elements/lisk-client/index.html
@@ -8,6 +8,6 @@
 		Open the console to get started with Lisk Client.
 		<div id="output"></div>
 
-		<script src="dist-browser/index.min.js"></script>
+		<script src="dist-browser/index.js"></script>
 	</body>
 </html>
--- a/elements/lisk-client/src/index.ts
+++ b/elements/lisk-client/src/index.ts
@@ -31,7 +31,7 @@ export const constants = constantsModule;
 export const cryptography = cryptographyModule;
 export const passphrase = passphraseModule;
 export const transactions = transactionsModule;
-// Also export as `transacation` for backward compatibility.
+// Also export as `transaction` for backward compatibility.
 // See https://github.com/LiskHQ/lisk-sdk/issues/3925#issuecomment-508664703
 export const transaction = transactionsModule;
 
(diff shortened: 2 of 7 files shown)


elements/lisk-client/src/index.ts
```diff
@@ -35,6 +42,7 @@ export default {
 	cryptography,
 	passphrase,
 	transactions,
+	Buffer,
 	// Also export as `transacation` for backward compatibility.
```
Incede, line 46: typo -> 'transacation'

reyraa (approved):
Looks good. Thanks

elements/lisk-client/index.html
```diff
@@ -8,6 +8,6 @@
 		Open the console to get started with Lisk Client.
 		<div id="output"></div>
 
-		<script src="dist-browser/index.min.js"></script>
+		<script src="dist-browser/index.js"></script>
```
ManuGowda, line 11: any specific reason for not using `min`?

ManuGowda (approved):

elements/lisk-client/index.html
shuse2: This file is for testing and i wanted to use unuglify version for test

Incede (approved):

PR #5424, opened 2020-06-10
Refactor chain library - Closes #5088

### LINKED ISSUE #5088: Update transactions to be first order function
### Description

We currently have 2nd order function for the transaction handling, but as it is encapsulated in the lisk-chain, and maintaining the context in different way, we longer need this function.

Therefore, composeTransaction should be removed, and handle the transaction with simple function

### Acceptance Criteria

- Remove `composeTransactionStep` and relative tests
- Update transaction handling to be first order function
- Check if the functions that are using `composeTransactionStep` are covered with test properly

shuse2:
### What was the problem?

This PR resolves #5088 

### How was it solved?

- Update transaction handlers to be first degree functions
- Remove unnecessary ID checks
- Remove unnecessary separation of functions

### How was it tested?

- Updated all the unit and integration tests
- Connect to lisk-core and run the application


elements/lisk-chain/src/chain.ts
```diff
@@ -399,45 +336,39 @@ export class Chain {
...
+
+	// eslint-disable-next-line class-methods-use-this
+	public async apply(block: Block, stateStore: StateStore): Promise<void> {
+		if (block.payload.length > 0) {
+			const transactionsResponses = await applyTransactions(
+				block.payload,
+				stateStore,
+			);
+
+			const unappliableTransactionsResponse = transactionsResponses.filter(
```
ishantiw, line 353: `unappliable` is not a valid word, maybe `inapplicableTransactionsResponse` or `invalidTransactionsResponse` `CannotBeAppliedTransactions` 🤔

PR #5429, opened 2020-06-11
Implement generateProof for Merkle Tree  - Closes #5206 

### LINKED ISSUE #5206: Implement generateProofOfInclusion for Merkle Tree
### Description

Using the Merkle Tree, a `Prover` provides the `Verifier` with a proof-of-inclusion to prove that a specific data block in included in an authenticated data set. 

The elements constituting a proof-of-inclusion are:
1. `dataLength`: The total number of data blocks in the dataset (the length of data);
2. `idxs`: A set of indices corresponding to the position in the tree of the data in queryData;
3. `path`: The audit path, an array of bytes necessary to compute the Merkle root starting from queryData.

### Motivation

The proof-of-inclusion procedure can be used by a `light-client` node in which the the `light-client` is not required to possess the entire blockchain and can verify transactions by communicating with a full node. 

The procedure for a proof-of-inclusion is as follows:

- Verifier: The Verifier knows merkleRoot and sends the Prover a set of data blocks queryData for which they wish to obtain a proof-of-inclusion;
- Prover: The Prover finds the elements of queryData in the Merkle tree and generates the set of corresponding indices idxs;
- The Prover generates path and transmits the complete proof to the Verifier; proof contains dataLength, idxs and path;
- Verifier: The Verifier uses proof to recalculate the Merkle root using elements from queryData and checks that it equals merkleRoot.

### Acceptance Criteria

- Add test cases

### Additional information

- Generate `idxs` by assigning an index to the data block according to the following rules:

  1. If the data block is not part of the tree, it is assigned a special index and the data block is not taken into consideration when constructing the rest of the proof;
  2. The index of a data block present in the tree equals its position in the layer converted to binary, with a fixed length equal to the tree height minus the layer number, with 0s prepended as necessary. Each index starts with a 0, and the layer the node belongs to can be inferred from the length of the index and dataLength;
  3. The elements of idxs are placed in the same order as the corresponding data in queryData.

- Generate the path by adding the necessary hashes in a bottom-up/left-to-right order. This order reflects the order in which the path is consumed by the Verifier during the verification.

Related issues: #5204

mitsuaki-u:
### What was the problem?

Using the Merkle Tree, a Prover provides the Verifier with a proof-of-inclusion to prove that a specific data block in included in an authenticated data set.

This PR resolves #5206

### How was it solved?

Given a set of query data blocks, a Prover can generate a `proof-of-inclusion`: an array of objects which contain a `hash: Buffer` and a `direction: number<0|1>`. 

The Verifier in turn can compute a `merkle root` from the given `proof-of-inclusion` by hashing together the given hash values according to each corresponding given `direction`. 

The computed `merkle root` is then compared with the Verifier's own merkle root and checked if they are equal. 

### How was it tested?

An easily hand-verifiable test case was written: a merkle tree with two leaves. Once `verifyProof` is implemented, there will be more robust test cases included.


elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -163,10 +154,219 @@ export class MerkleTree {
...
+		}
+		let currentNode: NodeInfo | undefined;
+		// Iterate through query nodes in order
+		for (let j = 0; j < queryNodes.length; j += 1) {
+			currentNode = queryNodes[j];
+			// Query node does not exist in tree
+			if (!currentNode) {
+				// Insert flag for unverified node
+				path.push(undefined);
+				// eslint-disable-next-line
```
shuse2, line 200: Let's add 
```
rules: {
		'no-continue': 'off',
	}
```
in the .eslintrc.js

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -163,10 +154,219 @@ export class MerkleTree {
...
+									lastNodeOfLowerOddLayer.hash.length,
+							),
+						);
+						currentNode = this.getNode(parentNodeHash);
+					}
+					// If there is more than one node and the current node is not the last node in the layer
+					else if (
+						currentNode.hash.compare(
+							currentLayer[currentLayer.length - 1].hash,
+						) !== 0
```
shuse2, line 275:
```suggestion
						!currentNode.hash.equals(
							currentLayer[currentLayer.length - 1].hash,
						)
```

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -163,10 +154,219 @@ export class MerkleTree {
...
+						};
+						const parentNodeHash = hash(
+							Buffer.concat(
+								[BRANCH_PREFIX, pairInfo.hash, currentNode.hash],
+								BRANCH_PREFIX.length +
+									pairInfo.hash.length +
+									currentNode.hash.length,
+							),
+						);
+						if (parentNodeHash.toString('binary') in this._hashToValueMap) {
```
shuse2, line 330:
```suggestion
						if (this._hashToValueMap[parentNodeHash.toString('binary')] !== undefined) {
```
Im not sure if this is better, but it is done in this way in other places

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -163,10 +154,219 @@ export class MerkleTree {
...
+					0,
+					currentNode,
+				);
+			}
+		}
+
+		return structure;
+	}
+
+	public generatePath(_queryData: ReadonlyArray<Buffer>): Path {
```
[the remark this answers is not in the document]
mitsuaki-u: This only generates a path.

elements/lisk-tree/src/merkle_tree.ts
mitsuaki-u: To me the first way looks cleaner

PR #5430, opened 2020-06-12
Replace epochTime with genesis block timestamp - Closes #5280

### LINKED ISSUE #5280: Replace epochTime with genesis block timestamp

pablitovicente:
### What was the problem?

This PR resolves #5280

### How was it solved?

- By using a date in the genesis block
- By adjusting the calculations on slots.ts

### How was it tested?

Unit Tests


elements/lisk-dpos/test/unit/apply.spec.ts
```diff
@@ -40,6 +39,10 @@ import {
 } from '../../src/constants';
 import { StateStoreMock } from '../utils/state_store_mock';
 
+const MS_IN_A_SEC = 1000;
+const GENESIS_BLOCK_TIMESTAMP =
+	new Date(Date.UTC(2020, 5, 15, 0, 0, 0, 0)).getTime() / MS_IN_A_SEC;
+
```
nazarhussain, line 45: I saw these multiple times in `lisk-dpos` tests. Can we define it somewhere shared?

elements/lisk-chain/src/slots.ts
```diff
@@ -13,45 +13,47 @@
...
-	readonly epochTime: string;
+	readonly genesisBlockTimestamp: number;
 	readonly interval: number;
 }
 
 const SEC_IN_MS = 1000;
 
 export class Slots {
-	private readonly _epochTime: Date;
+	private readonly _genesisTime: Date;
```
shuse2, line 23: Not important, but I think it was originally `Date` because the epochTime input was ISO date.
It seems it will be a lot simpler if we use number (timestamp) here

framework/src/application/node/forger/forger.ts
```diff
@@ -343,12 +345,12 @@ export class Forger {
...
+		const MS_IN_A_SEC = 1000;
 		const currentSlot = this._chainModule.slots.getSlotNumber();
-		const currentSlotTime = this._chainModule.slots.getRealTime(
-			this._chainModule.slots.getSlotTime(currentSlot),
-		);
 
-		const currentTime = new Date().getTime();
+		const currentSlotTime = this._chainModule.slots.getSlotTime(currentSlot);
+
+		const currentTime = Math.floor(new Date().getTime() / MS_IN_A_SEC);
```
nazarhussain, line 353: Output of `getSlotTime(currentSlot)` is in Unix timestamp (meant multiplied by 1000). So you don't need to divide `currentTime` here.

shuse2, line 353: on the previous implementation, currentTime was in `ms`, and in below it's comparing with `currentTime <= currentSlotTime + waitThreshold`.

I think unit should be
currentSlotTime => ms
currentTIme => ms
waitThreshold =>ms

elements/lisk-chain/src/slots.ts
```diff
@@ -13,45 +13,47 @@
...
-		const parsedEpochTime =
-			epochTime === undefined ? this.getEpochTime() : epochTime;
 
-		return Math.floor(parsedEpochTime / this._interval);
+		return Math.floor(elapsedTime / this._interval);
 	}
 
 	public getSlotTime(slot: number): number {
-		return slot * this._interval;
+		const slotGensisTimeOffset = slot * this._interval;
```
shuse2, line 50: this is returning in `ms`, but `timeSinceGenesis` is returning in `s`, I think it's better to make it consistent

framework/src/application/node/forger/forger.ts
pablitovicente: This will be solved by using only seconds in Slots.ts so resolving.

elements/lisk-chain/src/slots.ts
pablitovicente: Changed.

elements/lisk-chain/src/slots.ts
pablitovicente: Changed.

nazarhussain:
@pablitovicente LGTM, apart from failing p2p tests. Not sure if it's because of change in this PR.

pablitovicente:
> @pablitovicente LGTM, apart from failing p2p tests. Not sure if it's because of change in this PR.

Thanks @nazarhussain p2p test failing due to timing after other changes. Another PR is fixing those issues unrelated to these changes.

PR #5431, opened 2020-06-12
Update hex to base64 - Closes #5390

### LINKED ISSUE #5390: Update all config and logs to use Base64 instead of hex
### Description

Update config, logs to output base64 format when it converts from/to Buffer.
Also, remove usage of publicKey and network identifier as string.

### Motivation

All the communication of encoded values will be communicated with base64.
String representation of bytes/Buffer is base64

### Acceptance Criteria

- Update schema validation
- Update public key usages in string and tests (forger public key and hashOnions)
- Update network identifier usages in string and tests
- Update config and genesis block to use base64 (genesis block)

shuse2:
### What was the problem?

This PR resolves #5390 

### How was it solved?

- Update all the usage of hex to base64
- Update configs to be parsed at application level and provide to lower level instances

### How was it tested?

- All tests are updated


commander/src/utils/constants.ts
```diff
@@ -49,9 +47,16 @@ export const API_PROTOCOLS: ReadonlyArray<string> = ['http:', 'https:'];
...
-	main: MAINNET_NETHASH,
-	test: TESTNET_NETHASH,
+	main: Buffer.from(
+		'ed14889723f24ecc54871d058d98ce91ff2f973192075c0155ba2b7b70ad2511',
+		'hex',
+	).toString('base64'),
+	test: Buffer.from(
+		'da3ed6a45429278bac2666961289ca17ad86595d33b31037615d4b8e8f158bba',
+		'hex',
+	).toString('base64'),
```
nazarhussain, line 59: This seems interesting. Why no we just use `base64` strings and try to remember those onward. May be to recall better, just add hex string as a comment on top of those.

commander/test/utils/api.ts
```diff
@@ -18,10 +18,8 @@ import { APIClient } from '@liskhq/lisk-api-client';
 import { getAPIClient } from '../../src/utils/api';
 
 describe('api utils', () => {
-	const mainnetNethash =
-		'ed14889723f24ecc54871d058d98ce91ff2f973192075c0155ba2b7b70ad2511';
-	const testnetNethash =
-		'da3ed6a45429278bac2666961289ca17ad86595d33b31037615d4b8e8f158bba';
+	const mainnetNethash = '7RSIlyPyTsxUhx0FjZjOkf8vlzGSB1wBVbore3CtJRE=';
+	const testnetNethash = '2j7WpFQpJ4usJmaWEonKF62GWV0zsxA3YV1Ljo8Vi7o=';
```
nazarhussain, line 22: Related to the above, same as we used here.

commander/test/utils/network_identifier.ts
```diff
@@ -18,12 +18,18 @@ import { getNetworkIdentifierWithInput } from '../../src/utils/network_identifie
...
+			'hex',
+		).toString('base64');
+		const testnetNetworkIdentifier = Buffer.from(
+			'10f236f6d00a8f565bbe43c4ef0e3818f488bf4abefbe041155a3d019ef9a947',
+			'hex',
+		).toString('base64');
+		const defaultNetworkIdentifier = Buffer.from(
+			'7777777777777777777777777777777777777777777777777777777777777777',
+			'hex',
+		).toString('base64');
```
nazarhussain, line 32: Same here, if we are gonna use `base64` and we don't need to double convert. Just convert those once and use it everywhere as base64.

framework/src/application/genesis_block.ts
```diff
@@ -0,0 +1,142 @@
...
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+// TODO: Remove this file completely after #5234
+
+import { Block } from '@liskhq/lisk-chain';
```
nazarhussain, line 17: Why we needed this at the moment? It will be removed completely with new genesis bock. Should not we just use `any` or `object` somewhere for now.

shuse2: This function was there already to convert genesis block to actual instance.
but with the new genesis block, this whole file should be removed, not only the type

commander/src/utils/constants.ts
shuse2: This nethash, network identifier will need to be updated, so that's why I kept it as it is for now.
Nethash will not be used in the network identifier

commander/test/utils/api.ts
shuse2: Same as above

commander/test/utils/network_identifier.ts
shuse2: Same as above

PR #5434, opened 2020-06-12
Allow custom schema in P2P constructor - Partially Closes #5419

### LINKED ISSUE #5419: Redefine P2P communication and message formats after using lisk-codec
### Description
Update the content of each endpoints to
#### getPeers
```
{
  peers: {
     ipAddress: string;
     port: uint32;
  }[]
}
```
#### getNodeInfo
```
{
  advertiseAddress: boolean;
  networkID: string;
  protocolVersion: string;
  nonce: string;
  // custom field injected from outside
  options: {
    height: uint32;
    maxHeightPrevoted: uint32;
    blockVersion: uint32;
  },
}
```
#### updateNodeInfo
```
// Only accept custom field
{
  height: uint32;
  maxHeightPrevoted: uint32;
  blockVersion: uint32;
}
```
#### Handshake
It should only require below information
```
{
  advertiseAddress: boolean;
  port: uint32;
  networkId: string;
  protocolVersion: string;
  nonce: string;
}
```

### Expected behavior

- P2P layer should communicate messages en/decoded using lisk-codec
- Should allow custom fields for custom modules by accepting schema from the user - Solved in https://github.com/LiskHQ/lisk-sdk/pull/5434

### Actual behavior

- Some messages are not well defined for p2p communicated and we need to finalize which fields will be required in the schema for encoding.

### Steps to reproduce

N/A

`5.0`

ishantiw:
### What was the problem?

This PR resolves #5419 

### How was it solved?

- New property to accept custom schema for `peerInfo` and `nodeInfo`
- Construct custom schema for `nodeInfo` and `peerinfo` in `framework/network`

### How was it tested?

Run `node framework/test/test_app`


elements/lisk-p2p/src/p2p.ts
```diff
@@ -257,8 +263,13 @@ export class P2P extends EventEmitter {
...
 			secret: this._secret,
 		});
-		codec.addSchema(peerInfoSchema);
-		codec.addSchema(nodeInfoSchema);
+
+		this._rpcSchema = {
+			peerInfo: config.customSchema?.peerInfo ?? peerInfoSchema,
+			nodeInfo: config.customSchema?.nodeInfo ?? nodeInfoSchema,
+		};
+		codec.addSchema(this._rpcSchema.peerInfo);
```
pablitovicente, line 271: It would `_handlePeerPoolRPC` is a callback that's registered for actions if that's the case maybe this `addSchema` calls should happen somewhere else as schema caching is expensive so we want to use it as little as possible.

framework/src/application/network/schema.ts
```diff
@@ -0,0 +1,49 @@
...
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+import { p2pSchemas } from '@liskhq/lisk-p2p';
+
+export const customNodeInfoSchema = {
```
shuse2, line 17: Instead of giving full access to the schema itself, isn't it safer to ask user to provide additional schema, and combine them in p2p constructor?
Otherwise, minimally required NodeInfo will be overwritten

elements/lisk-p2p/src/p2p.ts
ishantiw: This `addSchema` is in the constructor of P2P(not inside `_handlePeerPoolRPC`) and only called once on startup, if this is what you meant

pablitovicente: Oh my parser failed to get the correct scope. Sorry!

framework/src/application/network/schema.ts
```diff
@@ -32,7 +31,6 @@ export const customNodeInfoSchema = {
 };
 
 export const customPeerInfoSchema = {
-	...p2pSchemas.peerInfoSchema,
 	$id: '/peerInfo/custom',
 	type: 'object',
 	properties: {
```
shuse2, line 36: shouldn't the line below also suppose to be deleted? 
```
...p2pSchemas.peerInfoSchema.properties,
```

ishantiw: true, i missed it

ishantiw: addressed

PR #5436, opened 2020-06-15
Add last block forged parameter to account - Closes #4935

### LINKED ISSUE #4935: Update DPoS library to calculate last forged height
### Description

- Update `lastForgedHeight` property of delegate in the DPoS library to be calculated every block by following https://github.com/LiskHQ/lips/blob/master/proposals/lip-0023.md#delegate-productivity-1
- Initial value of lastForgedHeight should be the height which delegate is registered

### Motivation

For the safe fail mechanism for the new DPoS system, `lastForgedHeight ` must be tracked.

### Acceptance Criteria

- Test case when applying delegate lastForgedHeight increments

### Related Issues

#5058

shuse2:
### What was the problem?

This PR resolves #4935 

### How was it solved?

- Initialize last forged height from delegate registration tx to the last block height + 1
- Set last forged height to the block header height in dpos every block

### How was it tested?

- Add unit test to check if those values are set

### DIFF
--- a/elements/lisk-dpos/src/delegates_info.ts
+++ b/elements/lisk-dpos/src/delegates_info.ts
@@ -112,7 +112,7 @@ export class DelegatesInfo {
 
 		const round = this.rounds.calcRound(block.height);
 		// Safety measure is calculated every block
-		await this._updateSafetyMeasure(block, stateStore);
+		await this._updateProductivity(block, stateStore);
 
 		// Below event should only happen at the end of the round
 		if (!this._isLastBlockOfTheRound(block)) {
@@ -160,7 +160,7 @@ export class DelegatesInfo {
 		return true;
 	}
 
-	private async _updateSafetyMeasure(
+	private async _updateProductivity(
 		blockHeader: BlockHeader,
 		stateStore: StateStore,
 	): Promise<void> {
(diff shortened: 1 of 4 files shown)


elements/lisk-dpos/src/delegates_info.ts
```diff
@@ -160,7 +160,7 @@ export class DelegatesInfo {
 		return true;
 	}
 
-	private async _updateMissedBlocks(
+	private async _updateSafetyMeasure(
```
pablitovicente, line 163: Why this name? I don't see a reference in the LIP but it would be nice if possible to have a name that hints at what this does. It would seem the original name `_updateMissedBlocks` was better.

elements/lisk-dpos/test/unit/apply.spec.ts
```diff
@@ -205,7 +205,7 @@ describe('dpos.apply()', () => {
 			);
 		});
 
-		describe('consecutiveMissedBlock', () => {
+		describe('safety measure', () => {
```
pablitovicente, line 208: Same here regarding naming maybe something more descriptive can be used here?

elements/lisk-dpos/src/delegates_info.ts
shuse2: I updated the name because it's no longer only updating missed blocks.
Maybe `_updateProductivity` is better?

shuse2:
@pablitovicente the banning part is done in separate issue https://github.com/LiskHQ/lisk-sdk/pull/5438

elements/lisk-dpos/src/delegates_info.ts
pablitovicente: Oh yes `_updateProductivity` is better

PR #5438, opened 2020-06-15
Add banning for the missedBlock, and forged height diff conditions - Closes #4942

### LINKED ISSUE #4942: Update DPoS library to ban when conditions are met
### Description

- Update DPoS library to ban delegates when conditions are met by following https://github.com/LiskHQ/lips/blob/master/proposals/lip-0023.md#delegate-productivity-1
- The `consecutiveMissedBlocks` and `lastForgedHeight` properties should be already calculated 
- Add the banned list into the consensus state

### Motivation

For the safe fail mechanism for the new DPoS system, when the conditions are met, delegates should be banned.

### Acceptance Criteria

- Test case when only one condition is met
- Test case when both conditions are met

### Related Issues

#5058

shuse2:
### What was the problem?

This PR resolves #4942 

### How was it solved?

Add condition to ban the delegate based on the consecutiveMissedBlock and last forged height

### How was it tested?

Added tests for all condition
- only pass the consecutive missed block condition
- only pass the forged height diff condition
- pass both consecutive missed block and forged height diff conditions

### DIFF
--- a/elements/lisk-dpos/src/delegates_info.ts
+++ b/elements/lisk-dpos/src/delegates_info.ts
@@ -38,6 +38,8 @@ interface DelegatesInfoConstructor {
 
 const _isGenesisBlock = (header: BlockHeader): boolean => header.height === 1;
 const zeroRandomSeed = Buffer.from('00000000000000000000000000000000', 'hex');
+const maxConsecutiveMissedBlocks = 50;
+const maxLastForgedHeightDiff = 260000;
 
 export class DelegatesInfo {
 	private readonly chain: Chain;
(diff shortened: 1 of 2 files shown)


elements/lisk-dpos/src/delegates_info.ts
```diff
@@ -38,6 +38,8 @@ interface DelegatesInfoConstructor {
 
 const _isGenesisBlock = (header: BlockHeader): boolean => header.height === 1;
 const zeroRandomSeed = Buffer.from('00000000000000000000000000000000', 'hex');
+const maxConsecutiveMissedBlocks = 50;
+const maxLastForgedHeightDiff = 260000;
```
pablitovicente, line 42: Even though this numbers are in the LIP maybe these values could be config?

shuse2: It might be, but I want to minimize what is exposed at the beginning, and decide to expose the feature if there is good usecase to it

Issue #5425, opened 2020-06-11
Getting multiple data using data access should not throw


shuse2:
### Expected behavior
In lisk-chain, dataAccess.getXXX which returns array of data should not throw NotFoundError.
In framework, there are multiple places relying on this behavior.

If not matching it should return empty array

### Actual behavior
the functions throws if one of the searching id does not exist in the database

### Steps to reproduce
Try call dataAccess.get with some unknown id.

### Which version(s) does this affect? (Environment, OS, etc...)
development

### Additional Information
https://github.com/LiskHQ/lisk-sdk/pull/5439#discussion_r440773301
This should be addressed and replaced by multiple obtain function once it's fixed

Issue #5354, opened 2020-05-22
Add JSON decode and encode for the lisk-codec


shuse2:
### Description

For the JSON decode and encode,
`bytes` should be translated from/to base64 string.
`uint64/sint64` should be translated from/to number string.

interface should be same as encode, decode.

Add 
```
decodeJSON(schema: Schema, message: Buffer): JSONObject
encodeJSON(schema: Schema, message:JSONObject): Buffer
```

### Motivation

Although it will not be used in the protocol, it will be beneficial to have json decode and encode for debugging, using in HTTP plugin, or from the front end.

### Acceptance Criteria

- Test should be written using the protocol spec
- It should not be used in the framework

pablitovicente:
Values should be mapped like:

```
CODED -> JSON
Uint64 -> string representation
Sint64 -> string representation
Int32 ->  number
Sint32 -> number
String -> string
Buffer -> Base64 string
Boolean -> boolean
```

PR #5439, opened 2020-06-15
Add basic functional tests - Closes #4894

### LINKED ISSUE #4894: Add tests for application actions and events
### Description
All the exposed `app` actions and events should have integration tests.
The list doesn't contain the actions that are exposed to the network, as this should be tested in the functional tests

### Motivation

Application actions and events are ultimate endpoints that user can use in the custom modules.
It should have the integration test for these functions.

### Acceptance Criteria
Below actions should be tested
- postTransaction
- getTransactionsFromPool
- getAccount
- getAccounts
- getBlockByID
- getBlocksByIDs
- getBlockByHeight
- getBlocksByHeights
- getBlocksByHeightBetween
- getTransactionByID
- getTransactionsByIDs

for Event
- networkEvent with `postBlock`
- networkEvent with `postTransactionsAnnouncement` 

### Additional Information
Continuous Issue created #5440

shuse2:
### What was the problem?

This PR resolves #4894 

### How was it solved?

- Add functional tests for actions, events and open p2p endpoints

### How was it tested?

- Run in framework, `npm run test:functional`

### DIFF
--- a/framework/src/application/application.ts
+++ b/framework/src/application/application.ts
@@ -317,8 +317,6 @@ export class Application {
 		} catch (error) {
 			this.logger.fatal({ err: error as Error }, 'failed to shutdown');
 		}
-
-		process.exit(errorCode);
 	}
 
 	// --------------------------------------
--- a/framework/src/application/application.ts
+++ b/framework/src/application/application.ts
@@ -317,6 +317,8 @@ export class Application {
 		} catch (error) {
 			this.logger.fatal({ err: error as Error }, 'failed to shutdown');
 		}
+
+		process.exit(errorCode);
 	}
 
 	// --------------------------------------
--- a/framework/src/application/node/transport/transport.ts
+++ b/framework/src/application/node/transport/transport.ts
@@ -343,9 +343,17 @@ export class Transport {
 
 		if (idsNotInPool.length) {
 			// Check if any transaction that was not in the queues, is in the database instead.
-			const transactionsFromDatabase = await this._chainModule.dataAccess.getTransactionsByIDs(
-				idsNotInPool,
-			);
+			const transactionsFromDatabase: BaseTransaction[] = [];
+			for (const id of idsNotInPool) {
+				try {
+					const tx = await this._chainModule.dataAccess.getTransactionByID(id);
+					transactionsFromDatabase.push(tx);
+				} catch (error) {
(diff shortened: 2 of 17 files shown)


framework/src/application/application.ts
```diff
@@ -317,8 +317,6 @@ export class Application {
 		} catch (error) {
 			this.logger.fatal({ err: error as Error }, 'failed to shutdown');
 		}
-
-		process.exit(errorCode);
```
nazarhussain, line 321: We had handled the error up and we are not throwing it upward, and even we removed this `process.exit` here. How the main process will exit in case of error?

shuse2: if all the open handle was properly closed, it should be able to exit the process, but it was not the case. I will revert the the change here

framework/src/application/node/transport/transport.ts
```diff
@@ -342,9 +343,17 @@ export class Transport {
...
+			for (const id of idsNotInPool) {
+				try {
+					const tx = await this._chainModule.dataAccess.getTransactionByID(id);
+					transactionsFromDatabase.push(tx);
+				} catch (error) {
+					if (!(error instanceof NotFoundError)) {
+						throw error;
+					}
+				}
+			}
```
ManuGowda, line 356: since the code repeats, how about a private function?

framework/test/functional/setup.js
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright © 2019 Lisk Foundation
```
ManuGowda, line 2: 2020

framework/src/application/application.ts
nazarhussain: I think open handles also count for the files. And we have logging file open always, that may be the cause of it.

framework/src/application/node/transport/transport.ts
shuse2: I think https://github.com/LiskHQ/lisk-sdk/issues/5425 will cover this.

shuse2:
@ManuGowda I think in the internal code, we shouldn't need this conversion because it will be handled by coded directly.
Therefore, i dont think we need to add it to cryptography. 
Also, it's kind of related to this issue, because in the test, it's just reading from the JSON fixture
https://github.com/LiskHQ/lisk-sdk/issues/5354

PR #5441, opened 2020-06-16
Cleanup validator - Closes #3825

### LINKED ISSUE #3825: Cleanup lisk-validator
### Expected behavior

usage of inline vs imported callbacks is consistent.
Reverify if all format type are relevant.

### Actual behavior

Some formats are defined as functions in ./validation, some as inline functions in this file. We should use a consistent approach.

shuse2:
### What was the problem?

This PR resolves #3825 

### How was it solved?

Remove outdated formats that are not been used or unnecessary.

The address check in commander should be removed and use base32 address in the future.
Also, for transaction creation done in transactions or commander, it should be sufficient to do `transaction.validate()` in the future. 
It should be addressed in https://github.com/LiskHQ/lisk-core/issues/271

### How was it tested?

Updated all the tests


elements/lisk-transactions/src/register_delegate.ts
```diff
@@ -54,7 +53,9 @@ const validateInputs = ({
 		);
 	}
 
-	validateNetworkIdentifier(networkIdentifier);
+	if (hexToBuffer(networkIdentifier).length !== 32) {
+		throw new Error('Invalid network identifier length');
```
pablitovicente, line 57: Maybe add the expected length to the message so it is more informative.

shuse2: I think having valid length won't help much to the user in this case.
Also, this function and in commander, it needs to be updated when we design for the UI.

I will keep the current code for this PR, but let's discuss the validation when we rewrite for the UI

PR #5442, opened 2020-06-17
Save the calculated BFT properties on the consensus state - Closes #5273, #5453

### LINKED ISSUE #5273: Save the calculated BFT properties on the consensus state
### Description

- Define the schema for the saving values
- Save the preVotes, preCommits and state to the consensus store
- Remove undo recompute step
- Remove usage of `getBlockHeadersByHeightBetween` and `getLastBlockHeader`, instead it should use `stateStore.consensus.lastBlockHeaders`

### Motivation

In order to remove undo steps from everywhere, calculated BFT properties on the consensus state need be saved.

### Acceptance Criteria

- Finalized height should not be stored as part of this diff because it cannot be undo
- Update unit tests
- Update delete block integration test to check the BFT property

### Additional Information

Related issues: #5232

ManuGowda:
### What was the problem?

This PR resolves #5273 

### How was it solved?

- Defined the schema for the saving values
- Saved the preVotes, preCommits and state to the consensus store
- Removed undo recompute step

### How was it tested?

- Updated unit tests


framework/src/application/node/processor/processor.ts
```diff
@@ -392,6 +401,10 @@ export class Processor {
 	): Promise<void> {
 		// Offset must be set to 1, because lastBlock is still this deleting block
 		const stateStore = await this.chainModule.newStateStore(1);
+		assert(
```
shuse2, line 404: probably
```
if (block.header.height <= this.bftModule.finalityManager.finalizedHeight) {
  throw new Error(xxx)
}
```
is fine?

PR #5443, opened 2020-06-17
Getting multiple data using data access should not throw - Closes #5425

### LINKED ISSUE #5425: Getting multiple data using data access should not throw

pablitovicente:
### What was the problem?

This PR resolves #5425

### How was it solved?

- By catching instances of `NotFoundError` thrown by the db module and not letting them bubble up.
- By removing transport temp fix for this issue and replacing with correct data access method

### How was it tested?

- added integration tests
- added missing integration tests for some methods that were missing

### DIFF
--- a/elements/lisk-chain/src/data_access/storage.ts
+++ b/elements/lisk-chain/src/data_access/storage.ts
@@ -331,8 +331,15 @@ export class Storage {
 	): Promise<Buffer[]> {
 		const accounts = [];
 		for (const address of arrayOfAddresses) {
-			const account = await this.getAccountByAddress(address);
-			accounts.push(account);
+			try {
+				const account = await this.getAccountByAddress(address);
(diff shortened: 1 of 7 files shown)


pablitovicente:
Reviewers please note that Jenkins is failing to build commander but it's unrelated to these changes.

PR #5448, opened 2020-06-18
Update test settings - Closes #5446

### LINKED ISSUE #5446: Update jest coverage setting
### Description

- Update with default test to be simplified without coverage
- Add `test:coverage` script to show the coverage in text format
- Add `test:ci` script for coverage collection

### Motivation

- Using coverage in the local test breaks the source map
- To check the coverage locally, it's nice to have text format
- For the CI, it should be as it is.

shuse2:
### What was the problem?

This PR resolves #5446 

### How was it solved?

Updated to have multiple different test commands


shuse2:
After discussion, the `framework/README.md` is still valid and all other package does not mention `npm t`, we will keep it as it is

PR #5452, opened 2020-06-18
Improve performance of Lisk-codec - Partially Solves #5353

### LINKED ISSUE #5353: Improve performance of Lisk-codec
### Description

Ideally, Lisk-codec performance improvement.

2 main points of improvement are
- compiled schema format, so that it can be process faster
- Each writing and reading of Buffer

### Motivation

With current block content processing, it has about 600 encode and 600 decode processes, which takes about 10ms in total.

### Acceptance Criteria

- Check the performance against the previous implementation
- All protocol test should pass and remain the same
- Interface should remain the same.

pablitovicente:
### What was the problem?

This PR partially resolves #5353

### How was it solved?

- By removing buffer.concat from bytes/string writers and replacing with regular push to array

Benchmark before change
```
node benchmark/encode
encode x 130,545 ops/sec ±0.74% (91 runs sampled)
```
Benchmark after change
```
node benchmark/encode
encode x 162,285 ops/sec ±0.45% (90 runs sampled)
```
So this change would yield about 32K more encodings per second

### How was it tested?

- Need to update tests 
- Need to check real impact in string/bytes readers

### DIFF
--- a/elements/lisk-codec/src/bytes.ts
+++ b/elements/lisk-codec/src/bytes.ts
@@ -11,10 +11,9 @@
  *
  * Removal or modification of this copyright notice is prohibited.
  */
-import { writeUInt32, readUInt32 } from './varint';
+import { readUInt32 } from './varint';
 
-export const writeBytes = (bytes: Buffer): Buffer =>
-	Buffer.concat([writeUInt32(bytes.length), bytes]);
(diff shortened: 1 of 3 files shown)


pablitovicente:
Closing for now. Code is good to go but needs bytes/string writers/readers tests to be updated. Keeping the branch for when we can take this issue into a milestone/

PR #5456, opened 2020-06-19
Fix nested object with array issue - Closes #5454

### LINKED ISSUE #5454: Codec should be able to encode/decode genesis block asset
### Expected behavior
Codec should be able to encode and decode genesis block asset field
```
const assetSchema = {
    $id: '/my/asset/schema',
    type: 'object',
    required: ['accounts', 'initDelegates', 'initRounds'],
    properties: {
      accounts: {
        type: 'array',
        items: {
          $id: '/account/base',
          type: 'object',
          properties: {
            address: { dataType: 'bytes', fieldNumber: 1 },
            balance: { dataType: 'uint64', fieldNumber: 2 },
            publicKey: { dataType: 'bytes', fieldNumber: 3 },
            nonce: { dataType: 'uint64', fieldNumber: 4 },
            keys: {
              fieldNumber: 5,
              type: 'object',
              properties: {
                numberOfSignatures: { dataType: 'uint32', fieldNumber: 1 },
                mandatoryKeys: {
                  type: 'array',
                  items: { dataType: 'bytes' },
                  fieldNumber: 2,
                },
                optionalKeys: {
                  type: 'array',
                  items: { dataType: 'bytes' },
                  fieldNumber: 3,
                },
              },
              required: ['numberOfSignatures', 'mandatoryKeys', 'optionalKeys'],
            },
            asset: {
              fieldNumber: 6,
              type: 'object',
              $id: '/genesis_block/account/default',
              properties: {
                delegate: {
                  type: 'object',
                  fieldNumber: 1,
                  properties: {
                    username: { dataType: 'string', fieldNumber: 1 },
                    pomHeights: {
                      type: 'array',
                      items: { dataType: 'uint32' },
                      fieldNumber: 2,
                    },
                    consecutiveMissedBlocks: {
                      dataType: 'uint32',
                      fieldNumber: 3,
                    },
                    lastForgedHeight: { dataType: 'uint32', fieldNumber: 4 },
                    isBanned: { dataType: 'boolean', fieldNumber: 5 },
                    totalVotesReceived: { dataType: 'uint64', fieldNumber: 6 },
                  },
                  required: [
                    'username',
                    'pomHeights',
                    'consecutiveMissedBlocks',
                    'lastForgedHeight',
                    'isBanned',
                    'totalVotesReceived',
                  ],
                },
                sentVotes: {
                  type: 'array',
                  fieldNumber: 2,
                  items: {
                    type: 'object',
                    properties: {
                      delegateAddress: { dataType: 'bytes', fieldNumber: 1 },
                      amount: { dataType: 'uint64', fieldNumber: 2 },
                    },
                    required: ['delegateAddress', 'amount'],
                  },
                },
                unlocking: {
                  type: 'array',
                  fieldNumber: 3,
                  items: {
                    type: 'object',
                    properties: {
                      delegateAddress: { dataType: 'bytes', fieldNumber: 1 },
                      amount: { dataType: 'uint64', fieldNumber: 2 },
                      unvoteHeight: { dataType: 'uint32', fieldNumber: 3 },
                    },
                    required: ['delegateAddress', 'amount', 'unvoteHeight'],
                  },
                },
              },
            },
          },
          required: [
            'address',
            'balance',
            'publicKey',
            'nonce',
            'keys',
            'asset',
          ],
        },
        fieldNumber: 1,
      },
      initDelegates: {
        type: 'array',
        items: { dataType: 'bytes' },
        fieldNumber: 2,
        minItems: 1,
      },
      initRounds: { dataType: 'uint32', fieldNumber: 3, minimum: 3 },
    },
  };
```

### Actual behavior
with the above setting, it fails to decode, but if the filedNumber is changed to
`initrounds: 1, initDelegates: 2, accounts: 3` it works

### Steps to reproduce
encode and decode devnet genesis block with above schema

5.0-

shuse2:
### What was the problem?

This PR resolves #5454 

### How was it solved?

- Fix the termination index input
- Main changes
 - https://github.com/LiskHQ/lisk-sdk/pull/5456/files#diff-70eb309de7b22123b6fb24df26dbbf42R134-R137
- https://github.com/LiskHQ/lisk-sdk/pull/5456/files#diff-70eb309de7b22123b6fb24df26dbbf42R151-R158

### How was it tested?

- Added protocol spec to reproduce the issue


elements/lisk-codec/src/codec.ts
```diff
@@ -108,7 +108,7 @@ export class Codec {
 			this.addSchema(schema);
 		}
 		const compiledSchema = this._compileSchemas[schema.$id];
-		const [res] = readObject(message, 0, compiledSchema);
+		const [res] = readObject(message, 0, compiledSchema, message.length);
```
nazarhussain, line 111: Would not it be always `message.lenth` if `terminateIndex` is not provided? What other cases there could be to not provide  number of bytes to read from the message and you also don't want to read the full message. 

May be I am missing some context of the codec here, but it seems if we moved this logic inside the `readObject` that if `terminateIndex` is not provided it would always be `message.length - offset` that will a safe option.

shuse2: root level terminating index is always the size of the message, but each internal object property will have it's own terminating length.
Also, this is the entry point of the recursive function

nazarhussain: Yes, I understand that and that's why I suggested if `terminateIndex` is not provided, for internal object properties we already have terminating length provided.

PR #5461, opened 2020-06-23
Update all regexpr to be valid and match valid data - Closes #5418

### LINKED ISSUE #5418: Update all regexpr to be valid and match valid data
### Expected behavior

All regular expressions used in `formats` the `lisk-validator` should be verified and tested properly. 

### Actual behavior

1.  Path regex has an Unescaped forward slash in this regex.
2. The path does not match forward slash at the end. e.g. `/nazar/alpha/beta/` which is quite common to use in paths

https://github.com/LiskHQ/lisk-sdk/blob/551ec555b1d9950845a004637be4255b52647173/elements/lisk-validator/src/formats.ts#L139-L142

3. Encrypted passphrase regex does not match values 
84251684-fb769a80-ab0d-11ea-9155-a549d09f643c

https://github.com/LiskHQ/lisk-sdk/blob/551ec555b1d9950845a004637be4255b52647173/elements/lisk-validator/src/formats.ts#L151-L159

5.0

ManuGowda:
### What was the problem?

This PR resolves #5418

### How was it solved?

- Updated regex

### How was it tested?

- Added unit tests

### DIFF
--- a/elements/lisk-validator/src/formats.ts
+++ b/elements/lisk-validator/src/formats.ts
@@ -51,10 +51,8 @@ export const version = isSemVer;
 export const protocolVersion = (data: string): boolean =>
 	/^(\d|[1-9]\d{1,2})\.(\d|[1-9]\d{1,2})$/.test(data);
 
-export const path = (data: string): boolean => {
-	const pathRegExp = new RegExp('^(.?)(/[^/]+)+$');
-	return pathRegExp.test(data);
-};
+export const path = (data: string): boolean =>
+	/^(.?)(\/[^/]+)+(\/?)$/.test(data);
 
 export const encryptedPassphrase = isEncryptedPassphrase;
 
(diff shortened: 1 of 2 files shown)


elements/lisk-validator/src/validation.ts
```diff
@@ -76,7 +77,8 @@ export const isEncryptedPassphrase = (data: string): boolean => {
 	return encryptedPassphraseRegExp.test(data);
 };
 
-export const isSemVer = (version: string): boolean => !!isValidVersion(version);
+export const isSemVer = (version: string): boolean =>
+	!!isValidVersion(coerce(version));
```
nazarhussain, line 81: I think it's not what we should be doing. The `coerce` tries to translate any version string. e.g. If one specifies `v2` it will translate it to `2.0.0` Then this will be validated. But we should not allow people to define application version as `v2`, which is not a valid semantic version. Just use `semver.valid` instead.

PR #5465, opened 2020-06-24
Enable eslint rules where disabled in controller - Closes #5312

### LINKED ISSUE #5312: Improve framework/controller code base to better adopt typescript
### Expected behavior

There should be no `@ts-ignore` and minimum `eslint-disable` usage in the controller code base. 

### Actual behavior

Currently, as we ported old code to typescript there are a few `@ts-ignroe` and many `eslint-disable` used. Which can be optimized further by: 

1. Moving class private functions as independent functions that don't use class instance. 
2. Update the `Action` definition and usage so that it doesn't have the `handler` as an optional attribute. 
3. Consistently use `event.key()` and `action.key()` where ever required. 

5.0.0

ishantiw:
### What was the problem?

This PR resolves #5312 

### How was it solved?

- Removed eslint-disable comments wherever possible 
- Fix the `eslint` warnings and errors

### How was it tested?

`npm run lint`

### DIFF
--- a/framework/src/controller/bus.ts
+++ b/framework/src/controller/bus.ts
@@ -266,17 +266,21 @@ export class Bus extends EventEmitter2 {
 		return Object.keys(this.events);
 	}
 
-	// eslint-disable-next-line @typescript-eslint/require-await
 	public async cleanup(): Promise<void> {
-		if (this.pubSocket) {
-			this.pubSocket.close();
-		}
-		if (this.subSocket) {
-			this.subSocket.close();
-		}
-		if (this.rpcSocket) {
-			this.rpcSocket.close();
-		}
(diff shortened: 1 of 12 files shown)


framework/src/controller/bus.ts
```diff
@@ -259,17 +266,21 @@ export class Bus extends EventEmitter2 {
...
-		if (this.pubSocket) {
-			this.pubSocket.close();
-		}
-		if (this.subSocket) {
-			this.subSocket.close();
-		}
-		if (this.rpcSocket) {
-			this.rpcSocket.close();
-		}
+		const closeSockets = async (): Promise<void> =>
```
shuse2, line 270: same as others

PR #5467, opened 2020-06-25
Verify proof for Merkle Tree - Closes #5207

### LINKED ISSUE #5207: Implement verifyProofOfInclusion for Merkle Tree
### Description

`verifyProofOfInclusion(proof, merkleRoot) => boolean`

### Motivation

In a proof-of-inclusion procedure, the verifier will receive a proof-of-inclusion from the prover and verify that a specific data block has been included in an authenticated data set.

### Acceptance Criteria

- Add test cases

### Additional information

- Using dataLength from the proof, calculate the structure of the tree from dataLength;
  1. There are a total number of layers equal to Height=ceiling(Log(dataLength))+1;
  2. The base layer contains a number of nodes equal to dataLength;
  3. Each other layer contains half the number of nodes of the layer below. If this is an odd number, we round either down and up.
- Identify the position in the tree of each element of queryData;
- Drop all elements that are not part of the tree using the special flag;
- Consume the path in the same order in which it was constructed to calculate the Merkle root;
- Check that the resulting Merkle root equals merkleRoot to complete the verification.

Related issues: #5204

mitsuaki-u:
### What was the problem?

In a proof-of-inclusion procedure using merkle tree, the `verifier` will receive a `proof-of-inclusion` from the `prover` and verify that a specific data block has been included in an authenticated data set.

This procedure must also allow for querying multiple data blocks. 

This PR resolves #5207 

### How was it solved?

Given a queryData set, the `prover` provides the `verifier` with a proof, an object with the following properties: 
```
 {     
     path, // An array containing the hashes of the nodes required to produce the merkle root
     indexes, // An array of objects containing the layer and node index for each data block in queryData
     dataLength, // The number of data blocks e.g. leaves in the tree
  }
```

With the given `proof` and the verifier's own merkle root and queryData, the verifier checks that the merkle root can be computed accurately. 

### How was it tested?

Given a fixed data set, a random number of data blocks from fixtures are chosen in random order and a proof is generated and verified for each set. 

Merkle trees with the following `dataLength` are currently tested: [2, 3, 4, 5, 6, 7, 8, 9, 10, 50, 100, 150, 300]


elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -148,206 +144,73 @@ export class MerkleTree {
...
 	}
 
-	public generatePath(_queryData: ReadonlyArray<Buffer>): Path {
+	public generateProof(queryData: ReadonlyArray<Buffer>): Proof | undefined {
 		if (this._width === 1) {
-			return [];
+			return undefined;
 		}
-		const treeStructure = this._getStructure();
+		const treeStructure = this._getPopulatedStructure();
```
shuse2, line 159: this treeStucture is used in only one place and I think the same functionality can be achieved by just looking up `_hashToValueMap`.
Prefer to remove `_getPopulatedStructure` since it loops O(n) where n is 2 ** (height + 1) + 1

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -522,15 +382,15 @@ export class MerkleTree {
 		const nodeValue = this._hashToValueMap[hashValue.toString('binary')];
 
 		if (isLeaf(nodeValue)) {
-			return nodeValue.toString('base64');
+			return hashValue.toString('hex');
```
shuse2, line 387: please use `base64` for binary string format for consistency

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -522,15 +382,15 @@ export class MerkleTree {
...
 		}
 
 		const node = this.getNode(hashValue);
 
 		return [
-			hashValue.toString('base64'),
-			`├${'─'.repeat(level)} ${this._printNode(node.leftHash, level + 1)}`,
-			`├${'─'.repeat(level)} ${this._printNode(node.rightHash, level + 1)}`,
+			hashValue.toString('hex'),
```
shuse2, line 393: same as above

elements/lisk-tree/src/utils.ts
```diff
@@ -0,0 +1,120 @@
...
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+import { hash } from '@liskhq/lisk-cryptography';
+import { LEAF_PREFIX } from './constants';
+import { NodeLocation, NodeSide } from './types';
+
+export const isLeaf = (value: Buffer): boolean =>
+	value.compare(Buffer.alloc(0)) !== 0 && value[0] === LEAF_PREFIX[0];
```
shuse2, line 20:
```suggestion
	value[0] === LEAF_PREFIX[0];
```
I think first part of condition is not needed?

elements/lisk-tree/src/utils.ts
```diff
@@ -0,0 +1,120 @@
...
+		Buffer.concat(
+			[prefix, leftHash, rightHash],
+			prefix.length + leftHash.length + rightHash.length,
+		),
+	);
+
+export const getMaxIdxAtLayer = (layer: number, datalength: number): number => {
+	let [max, r] = [datalength, 0];
+	for (let i = 0; i < layer; i += 1) {
+		// eslint-disable-next-line
```
shuse2, line 37: prefer to specify what is disabled here

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -154,200 +150,65 @@ export class MerkleTree {
 		return this.root;
 	}
 
-	public generatePath(_queryData: ReadonlyArray<Buffer>): Path {
+	public generateProof(queryData: ReadonlyArray<Buffer>): Proof | undefined {
 		if (this._width === 1) {
-			return [];
+			return undefined;
```
shuse2, line 157: should this return root as Proof or undefined is fine?

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -154,200 +150,65 @@ export class MerkleTree {
...
-				const queryNode = this.getNode(_queryData[i]);
-				queryNodes.push(queryNode);
-			} catch (err) {
-				queryNodes.push(undefined);
+		const addedPath = new Set();
+		const indexes = [];
+		let queryNode: NodeInfo | undefined;
+
+		for (let i = 0; i < queryData.length; i += 1) {
+			queryNode = this.getNode(queryData[i]);
```
shuse2, line 166: what happen if the non existent queryData is provided?

elements/lisk-tree/src/merkle_tree.ts
```diff
@@ -454,6 +295,25 @@ export class MerkleTree {
 		};
 	}
 
+	private _getPopulatedStructure(): TreeStructure {
+		const structure: { [key: number]: NodeInfo[] } = {};
+		const allNodes = this.getData();
```
shuse2, line 302: with removal of this function `getData()` also removed, or changed to private

elements/lisk-tree/src/verify_proof.ts
```diff
@@ -0,0 +1,69 @@
...
+
+export const verifyProof = (options: {
+	queryData: ReadonlyArray<Buffer>;
+	proof: Proof;
+	rootHash: Buffer;
+}): boolean => {
+	const { path, indexes, dataLength } = options.proof;
+	// Create a map for efficient lookup
+	const locationToPathMap: { [key: string]: Buffer } = {};
+	path.forEach(p => {
```
shuse2, line 26: prefer `for ... of`

elements/lisk-tree/test/merkle_tree.spec.ts
```diff
@@ -51,61 +52,29 @@ describe('MerkleTree', () => {
...
-					(merkleTree.generatePath(queryData)[0] as any).hash.compare(
-						expectedProofHash,
-					),
-				).toEqual(0);
-			});
-
-			it('should generate the expected path hash direction', () => {
-				const expectedProofHashDirection = 1;
+	describe('generateProof and verifyProof', () => {
+		for (const test of fixture.testCases.slice(2)) {
```
shuse2, line 56: why is it slicing from 2?

elements/lisk-tree/test/merkle_tree.spec.ts
```diff
@@ -51,61 +52,29 @@ describe('MerkleTree', () => {
...
-				expect(
-					(merkleTree.generatePath(queryData)[0] as any).hash.compare(
-						expectedProofHash,
-					),
-				).toEqual(0);
-			});
-
-			it('should generate the expected path hash direction', () => {
-				const expectedProofHashDirection = 1;
+	describe('generateProof and verifyProof', () => {
```
shuse2, line 55: maybe add test case where the verify proof is false, and query data not existing?

elements/lisk-tree/src/utils.ts
```diff
@@ -0,0 +1,120 @@
...
+	const structure = [];
+	for (let i = 0; i <= Math.ceil(Math.log2(datalength)); i += 1) {
+		structure.push(getMaxIdxAtLayer(i, datalength));
+	}
+
+	return structure;
+};
+
+export const getBinary = (num: number, length: number): number[] => {
+	if (length === 0) return [];
```
shuse2, line 53:
```suggestion
	if (length === 0) {
	  return [];
	}
```

elements/lisk-tree/src/merkle_tree.ts
mitsuaki-u: `getPairLocation` only returns the coordinates of the pair node. This is done intentionally to keep the function a pure function since it is also used in `verifyProof`. Since we only have the coordinates of the pair node, it is necessary to maintain the `treeStructure` to get the pair node hash.

elements/lisk-tree/src/merkle_tree.ts
mitsuaki-u: It was actually inconsistent before, since it was printing the branches as hex and only the leaves as base64. This way it is consistent and useful for debugging.

elements/lisk-tree/src/utils.ts
mitsuaki-u: This is to avoid detecting an empty buffer as a leaf node.

elements/lisk-tree/src/merkle_tree.ts
mitsuaki-u: It will throw an error which can be caught in the outer scope.

elements/lisk-tree/src/utils.ts
shuse2: `Buffer.alloc(0)[0] === undefined`, and `undefined === LIEAF_PREFIX[0]` is false so i think not needed?

elements/lisk-tree/src/merkle_tree.ts
ricott1: I think it was left unspecified in the LIP, I would return a `Proof` with path=[], indexes=[], dataLength=1, because we are assuming the root is known

elements/lisk-tree/src/merkle_tree.ts
ricott1: In the LIP we assign a special flag index (0) to any queryData that is not part of the tree, so that the verifier can discard them

PR #5468, opened 2020-06-25
Add JSON decode and encode for the lisk-codec - Close #5354

### LINKED ISSUE #5354: Add JSON decode and encode for the lisk-codec

pablitovicente:
### What was the problem?

This PR resolves #5354

### How was it solved?

- by adding `encodeJSON`
- by adding `decodeJSON`
- by adding required type transformation maps
- by adding recursive methods to figure out the proper types

### How was it tested?

Full unit tests for both `encodeJSON` and `decodeJSON`


elements/lisk-codec/src/codec.ts
```diff
@@ -113,6 +120,42 @@ export class Codec {
 		return (res as unknown) as T;
 	}
 
+	// For performance applications use decode() instead!
+	public decodeJSON = (schema: Schema, message: Buffer): GenericObject => {
```
shuse2, line 124:
```suggestion
	public decodeJSON<T>(schema: Schema, message: Buffer): T {
```

elements/lisk-codec/src/codec.ts
```diff
@@ -113,6 +120,42 @@ export class Codec {
...
+			'toJSON',
+			decodedMessageCopy,
+			(schema as unknown) as SchemaProps,
+			[],
+		);
+		return decodedMessageCopy as GenericObject;
+	};
+
+	// For performance applications use encode() instead!
+	public encodeJSON = (
```
shuse2, line 142:
```
public encode(schema: Schema, message: object): Buffer
```

elements/lisk-codec/src/codec.ts
pablitovicente: Had to make another adjustment so I committed from local repo. Done.

elements/lisk-codec/src/codec.ts
pablitovicente: I think you meant the type for `message` which I changed. If it was not that please comment again :)

pablitovicente:
> LGTM overall, just function signature should be the same as standard encode/decode.
> Also, it could be different issue but it would be useful to have conversion function also exposed.
> 
> ```
> public toJSON<T = object>(schema: Schema, message: object): T {}
> public toObject<T = object>(schema: Schema, message: object): T {}
> ```

Will split in different issue and start it in parallel to this PR

elements/lisk-codec/src/codec.ts
```diff
@@ -113,6 +120,39 @@ export class Codec {
 		return (res as unknown) as T;
 	}
 
+	// For performance applications use decode() instead!
```
nazarhussain, line 123: I don't get this comment. `decodeJSON` and `decode` have very different use cases. Can't replace each other for performance reasons.

elements/lisk-codec/src/json_wrapper.ts
```diff
@@ -0,0 +1,180 @@
...
+	GenericObject,
+	SchemaPair,
+	SchemaProps,
+	SchemaScalarItem,
+} from './types';
+
+export interface IteratableGenericObject extends GenericObject {
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	[Symbol.iterator](): Iterator<{ key: string; value: any }>;
+}
```
nazarhussain, line 27: Should be moved to `types`

elements/lisk-codec/src/codec.ts
```diff
@@ -113,6 +120,39 @@ export class Codec {
...
+			schema,
+			message,
+		);
+		const decodedMessageCopy = objectUtils.cloneDeep(decodedMessage);
+		decodedMessageCopy[Symbol.iterator] = iterator;
+
+		recursiveTypeCast(
+			'toJSON',
+			decodedMessageCopy,
+			(schema as unknown) as SchemaProps,
```
nazarhussain, line 135: Why not pass `scehma` as it is, and treat it as you want inside the function?

elements/lisk-codec/src/json_wrapper.ts
```diff
@@ -0,0 +1,180 @@
...
+		done: boolean;
+		value: { value: IteratableGenericObject; key: string };
+	};
+} {
+	let index = 0;
+	const properties = Object.keys(this);
+	let Done = false;
+	return {
+		// eslint-disable-next-line @typescript-eslint/no-explicit-any
+		next: (): { done: boolean; value: { value: any; key: string } } => {
```
nazarhussain, line 91: The return type for `value` in both places should match.

elements/lisk-codec/src/codec.ts
```diff
@@ -113,6 +120,39 @@ export class Codec {
...
+
+	// For performance applications use encode() instead!
+	public encodeJSON(schema: Schema, message: object): Buffer {
+		const messageCopy = objectUtils.cloneDeep(message);
+		(messageCopy as IteratableGenericObject)[Symbol.iterator] = iterator;
+
+		recursiveTypeCast(
+			'fromJSON',
+			messageCopy as IteratableGenericObject,
+			(schema as unknown) as SchemaProps,
```
nazarhussain, line 149: Can we use compiled schema here? If it's efficient for `encode` we may have its benefit here as well.

elements/lisk-codec/src/json_wrapper.ts
```diff
@@ -0,0 +1,180 @@
...
+						throw new Error(
+							`Invalid schema property found. Path: ${dataPath.join(',')}`,
+						);
+					}
+					if (mode === 'toJSON') {
+						// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access, @typescript-eslint/no-explicit-any
+						(object[key] as any)[i] = _liskMessageValueToJSONValue[
+							(schemaProp.items as SchemaScalarItem).dataType
+						](value[i]);
+					} else {
```
nazarhussain, line 146: This if/else inside loop and recursive calls does not look good and would be less performant. It would be nicer if we split the implementation into two functions.

elements/lisk-codec/src/json_wrapper.ts
```diff
@@ -0,0 +1,180 @@
...
+};
+
+export const recursiveTypeCast = (
+	mode: 'toJSON' | 'fromJSON',
+	object: IteratableGenericObject,
+	schema: SchemaProps,
+	dataPath: string[],
+): void => {
+	for (const { key, value } of object) {
+		if (isObject(value)) {
```
nazarhussain, line 111: Instead of this native check can' we use schema defined `type: object`

elements/lisk-codec/src/json_wrapper.ts
```diff
@@ -0,0 +1,180 @@
...
+
+			(value as IteratableGenericObject)[Symbol.iterator] = iterator;
+			recursiveTypeCast(mode, value, schema, dataPath);
+			dataPath.pop();
+
+			delete (value as IteratableGenericObject)[
+				(Symbol.iterator as unknown) as string
+			];
+		} else if (Array.isArray(value)) {
+			dataPath.push(key);
```
nazarhussain, line 122: Same here.

elements/lisk-codec/src/codec.ts
pablitovicente: It's just in case someone thinks it can just use it directly. Imagine someone is working in a custom plugin and thinks "hey I can just use my object directly with this methods decodeJSON/encodeJSON" that might be or not what they want. So it's there as a warning to call the attention of the user.

elements/lisk-codec/src/codec.ts
pablitovicente: Changing to compiled schema will mean a rewrite of this whole PR and was discussed with Shu for now not to do it. We can refactor in the future to use compiled schema.

elements/lisk-codec/src/codec.ts
pablitovicente: Sorry @nazarhussain I do not understand this question could you rephrase/expand please?

elements/lisk-codec/src/json_wrapper.ts
pablitovicente: This is a check to go into recursion if value is an object literal; I think the TS type will not work for this?

elements/lisk-codec/src/json_wrapper.ts
pablitovicente: Same question as for the one there.

elements/lisk-codec/src/json_wrapper.ts
pablitovicente: I get the point but if we split we end up with two functions that are exactly the same but in 1 line. Will check if possible to switch the map dynamically based on the parameter with little changes.

elements/lisk-codec/src/codec.ts
nazarhussain: I think it depends upon the use case if one needs JSON, it will use it, others don't.

elements/lisk-codec/src/json_wrapper.ts
pablitovicente: OK got clarification about this. I prefer to keep the check by observing the object itself.

elements/lisk-codec/src/codec.ts
nazarhussain: You are typecasting `schema` to `SchemaProps`, looks like both are same otherwise would not work.  So why not pass the schema without `as SchemaProps`

elements/lisk-codec/src/json_wrapper.ts
nazarhussain: But its inside loop and recursion. You yourself are emphasizing on the fact the even slightest performance really matters for `lisk-codec` then why not here.

elements/lisk-codec/src/json_wrapper.ts
nazarhussain: No not TS type. If its an object, then relevant schema definition must have `type: object` for it. So why not use that?

elements/lisk-codec/src/json_wrapper.ts
pablitovicente: @nazarhussain I emphasise it for `encode()` and `decode()` this methods are extras.

elements/lisk-codec/src/json_wrapper.ts
pablitovicente: Moved.

elements/lisk-codec/src/json_wrapper.ts
pablitovicente: Fixed

elements/lisk-codec/src/codec.ts
pablitovicente: I think the comment does not harm. When documentation is written for lisk-codec maybe this comment can be removed. I think it doesn't hurt to leave it for now.

elements/lisk-codec/src/json_wrapper.ts
pablitovicente: Changed to a map to make it constant time but still sorting out the type for the new map.

elements/lisk-codec/src/json_wrapper.ts
```diff
@@ -15,15 +15,16 @@
...
 	BaseTypes,
-	GenericObject,
+	IteratableGenericObject,
 	SchemaPair,
 	SchemaProps,
 	SchemaScalarItem,
 } from './types';
 
-export interface IteratableGenericObject extends GenericObject {
+interface iteratorReturnValue {
```
nazarhussain, line 24: Types always capitalize. 

```suggestion
interface IteratorReturnValue {
```

elements/lisk-codec/src/types.ts
```diff
@@ -71,3 +71,8 @@ export interface CompiledSchemas {
 export interface Validator {
 	addMetaSchema: (schema: object, key?: string) => {};
 }
+
+export interface IteratableGenericObject extends GenericObject {
+	// eslint-disable-next-line @typescript-eslint/no-explicit-any
+	[Symbol.iterator](): Iterator<{ key: string; value: any }>;
```
nazarhussain, line 77:
```suggestion
	[Symbol.iterator](): Iterator<IteratorReturnValue>;
```

elements/lisk-codec/src/json_wrapper.ts
pablitovicente: Done. Resolving.

elements/lisk-codec/src/codec.ts
pablitovicente: Let's leave as is. Maybe let's discuss complete type refactor for lisk-codec if deemed necessary as this abstract structures structures might need better type abastractions.

elements/lisk-codec/src/types.ts
pablitovicente: Had to do it locally as I had to move the IteratorReturnValue to the types file.

PR #5469, opened 2020-06-25
Add genesis block and bootstrap handling for dpos and bft - Closes #5281 and #5282

### LINKED ISSUE #5281: Update lisk-dpos to handle bootstrap period
### Description

- Save forger list using `initDelegate` and `initRound`
- During the bootstrap period defined in the genesis block, forger list should not be updated
- During the bootstrap period - 2 rounds, voteWeight should not be calculated

### Motivation

- With the new genesis block handling, the above rules need to be handled

### Acceptance Criteria

- Add/Update unit test to check if initDelegates used as round/robin if they are lesser than required forger list length
- Add unit test to check forger list is not updated during the bootstrap period
- Add unit test to check voteWeight is not calculated during the bootstrap period
- Add unit test to check forger list are saved for the bootstrap period

### Additional Information

Related issues: #5234

### LINKED ISSUE #5282: Update lisk-bft to handle bootstrap period
### Description

- During bootstrap period vote weight for forgers will be zero so they will be ignored for voting and pre-committing blocks 
- Initialize `chainMaxHeightPrevoted` and `chainMaxHeightFinalized` with the genesis block height

### Motivation

- With the new genesis block handling, the above rule needs to be handled

### Acceptance Criteria

- Add unit test to check bft properties are not updated during the bootstrap period
- Add unit test to check `chainMaxHeightPrevoted` and `chainMaxHeightFinalized` are initialized

### Additional Information

Related issues: #5234

shuse2:
### What was the problem?

This PR resolves #5281 and resolves #5282 

Main changes are
- https://github.com/LiskHQ/lisk-sdk/pull/5469/files#diff-63c89eb28d48e7ef3b1777aa44eed949L236-L243
- https://github.com/LiskHQ/lisk-sdk/pull/5469/files#diff-495dc95b650f88a0041c61d506a06fbcR208-R213
- https://github.com/LiskHQ/lisk-sdk/pull/5469/files#diff-828b5c200879926dc1024e828e5e5e61R80-R146
- https://github.com/LiskHQ/lisk-sdk/pull/5469/files#diff-bc24d9e6f40a8d9e205b1832549d23afR166-R169
- https://github.com/LiskHQ/lisk-sdk/pull/5469/files#diff-5146b5fc71f82a08a53ad0ff0cc1167bR36-R41

### How was it solved?

- Update dpos to calculate bootstrap period
- Update dpos to handle genesis block
- Update dpos not to calculate forgerslist while bootstrap period
- Update minActiveHeight to return lastBootstrapHeight + 1 the result is lower
- Update bft not to calculate properties during bootstrap period

### How was it tested?

- Update all the test
- Connect to core and run fresh application


elements/lisk-dpos/src/dpos.ts
```diff
@@ -41,13 +42,15 @@ import {
...
-	DecodedUsernames,
 } from './types';
 
 interface DposConstructor {
 	readonly chain: Chain;
+	readonly initDelegates: ReadonlyArray<Buffer>;
 	readonly activeDelegates?: number;
 	readonly standbyDelegates?: number;
+	readonly genesisHeight?: number;
+	readonly initRound?: number;
```
nazarhussain, line 53:
```suggestion
	readonly initRounds: number;
```
It should not be optional, must be provided to initiate DPOS.

elements/lisk-dpos/src/dpos.ts
```diff
@@ -41,13 +42,15 @@ import {
...
 	DecodedVoteWeights,
-	DecodedUsernames,
 } from './types';
 
 interface DposConstructor {
 	readonly chain: Chain;
+	readonly initDelegates: ReadonlyArray<Buffer>;
 	readonly activeDelegates?: number;
 	readonly standbyDelegates?: number;
+	readonly genesisHeight?: number;
```
nazarhussain, line 52:
```suggestion
	readonly genesisBlockHeight: number;
```

It should not be optional for DPOS, must be provided from the genesis block. Its already optional `createGenesisBlock`, so should not be redundant logic here.

elements/lisk-dpos/src/dpos.ts
```diff
@@ -70,6 +73,9 @@ export class Dpos {
 
 	public constructor({
 		chain,
+		initDelegates,
+		genesisHeight = DEFAULT_GENESIS_HEIGHT,
+		initRound = DEFAULT_INIT_ROUND,
```
nazarhussain, line 78: Same as above. These should not have default values in `dpos`. Having such default values on once place will make it easy to maintain.

elements/lisk-dpos/src/dpos.ts
```diff
@@ -84,6 +90,8 @@ export class Dpos {
 		this.delegateActiveRoundLimit = 3;
 		this.chain = chain;
 		this.rounds = new Rounds({
+			genesisHeight,
+			initRound,
 			blocksPerRound: this._delegatesPerRound,
```
nazarhussain, line 95: I notice or used somewhere else as `roundLength` we should be consistent. If required can create an issue to make these consistent.

elements/lisk-dpos/src/rounds.ts
```diff
@@ -13,14 +13,32 @@
...
+		genesisHeight,
+		initRound,
+	}: RoundConstructor) {
 		this.blocksPerRound = blocksPerRound;
+		this._genesisHeight = genesisHeight;
+		this.initRound = initRound;
+	}
+
+	public isBootstrapPeriod(height: number): boolean {
+		return height > this._genesisHeight && height <= this.lastHeightBootstrap();
```
nazarhussain, line 37:
```suggestion
		return height >= this._genesisHeight && height <= this.lastHeightBootstrap();
```

janhack: I don't know if this function is every called with the height of a genesis block, but note that the genesis block itself does not belong to a round (it defines an initial state). The first child block of the genesis block is the first block of a round.

elements/lisk-dpos/test/unit/apply.spec.ts
```diff
@@ -54,6 +59,7 @@ describe('dpos.apply()', () => {
 	let stateStore: StateStoreMock;
 	const randomSeed1 = Buffer.from('283f543e68fea3c08e976ef66acd3586');
 	const randomSeed2 = Buffer.from('354c87fa7674a8061920b9daafce92af');
+	const initDelegates = delegatePublicKeys.map(pk => Buffer.from(pk, 'hex'));
```
janhack, line 62: Would it be not more consistent to use addresses as identifiers? If you have a block with generatorPublicKey property, it is super fast to compute the address from it.

janhack: We changed to addresses as identifiers a few month backs, also the shuffling and selection of standby delegates uses addresses and not public keys any more. Note that public keys will also be removed from accounts.

elements/lisk-dpos/src/rounds.ts
shuse2: genesis height will not be the bootstrap period, also this equation was defined in the LIP

elements/lisk-dpos/test/unit/apply.spec.ts
shuse2: I agree we should change it to the address, but currently genesis block holds the public key, so it will handle in the removing address issue #5471

shuse2: My misunderstanding, actually this should be already address, will be fix in this PR

elements/lisk-bft/src/bft.ts
```diff
@@ -233,14 +232,11 @@ export class BFT extends EventEmitter {
...
-		/* Check BFT migration height
-		 https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#backwards-compatibility */
-		const bftMigrationHeight =
-			this.constants.startingHeight -
-			this.constants.activeDelegates * BFT_MIGRATION_ROUND_OFFSET;
-
-		// Choose max between stored finalized height or migration height
-		const finalizedHeight = Math.max(finalizedHeightStored, bftMigrationHeight);
+		// Choose max between stored finalized height or genesis height
+		const finalizedHeight = Math.max(
```
janhack, line 236: Initially, the finalized height is the height of the genesis block. Why are we taking the maximum here?

shuse2: when the application restarts, it takes from the stored value and compare if it should be genesis height or the stored one. In most of the cases it should be the stored value.
ie)
1. running fresh it will choose genesis height
2. stop and run again, it will choose the genesis height

but maybe i can just initialize with genesis height if not exist

elements/lisk-dpos/src/dpos.ts
```diff
@@ -196,7 +163,10 @@ export class Dpos {
 			delegateActiveRoundLimit,
 		);
 
-		return this.rounds.calcRoundStartHeight(activeRounds);
+		const minBootstrapHeight = this.rounds.lastHeightBootstrap() + 1;
```
janhack, line 166: `minBootstrapHeight` is a bit misleading, because it is the first height that is not in the bootstrapping phase

elements/lisk-dpos/src/dpos.ts
```diff
@@ -196,7 +163,10 @@ export class Dpos {
...
+		const minActiveHeight = this.rounds.calcRoundStartHeight(activeRounds);
```
janhack, line 167: where does activeRounds come from and how it is computed?

elements/lisk-bft/src/bft.ts
janhack: But the genesis block should be processed only once, right? This block cannot be reverted and the finalized height can only be increased after processing blocks building on the genesis block.

elements/lisk-dpos/src/dpos.ts
shuse2: active rounds comes from the above function `_findEarliestActiveListRound` which was not changed from the previous implementation.

elements/lisk-bft/src/bft.ts
shuse2: yes, so once the finalized height exist in storage, it will never go back. 
I will change it to to initialize when not existing in the storage

janhack: Okay. Can be resolved then.

elements/lisk-dpos/src/rounds.ts
janhack: I also think the current code should work fine.

elements/lisk-bft/test/unit/bft.spec.ts
```diff
@@ -91,14 +92,15 @@ describe('bft', () => {
 			dposStub = {
 				getMinActiveHeight: jest.fn(),
 				isStandbyDelegate: jest.fn(),
+				isBootstrapPeriod: jest.fn().mockResolvedValue(false),
```
nazarhussain, line 95:
```suggestion
				isBootstrapPeriod: jest.fn().mockReturnValue(false),
```

elements/lisk-bft/test/unit/finality_manager.spec.ts
```diff
@@ -301,6 +303,37 @@ describe('finality_manager', () => {
...
 
+			it('should not update prevotes and precommits in case of within bootstrap period', async () => {
+				const header1 = createFakeBlockHeader({
+					height: 2,
+					asset: {
+						maxHeightPreviouslyForged: 0,
+					},
+				});
+
+				dposStub.isBootstrapPeriod.mockResolvedValue(true);
```
nazarhussain, line 314:
```suggestion
				dposStub.isBootstrapPeriod.mockReturnValue(true);
```

elements/lisk-dpos/test/unit/is_active_delegate.spec.ts
```diff
@@ -48,9 +48,15 @@ describe('dpos.isActiveDelegate', () => {
 				getConsensusState: jest.fn(),
 			},
 		};
+		const initDelegates = delegatePublicKeys.map(pk =>
+			getAddressFromPublicKey(Buffer.from(pk, 'hex')),
+		);
```
nazarhussain, line 53: Why should not we replace the fixture with delegate addresses?

elements/lisk-dpos/test/unit/is_standby_delegate.spec.ts
```diff
@@ -61,9 +61,15 @@ describe('dpos.isStandbyDelegate', () => {
 			slots,
 		};
 
+		const initDelegates = delegatePublicKeys.map(pk =>
+			getAddressFromPublicKey(Buffer.from(pk, 'hex')),
+		);
```
nazarhussain, line 66: Same here

PR #5474, opened 2020-06-26
Create and export toJSON and toObject from lisk-codec - Closes #5473

### LINKED ISSUE #5473: Create and export toJSON and toObject from lisk-codec
### Description

public toJSON<T = object>(schema: Schema, message: object): T {}
public toObject<T = object>(schema: Schema, message: object): T {}

### Motivation

See notes in https://github.com/LiskHQ/lisk-sdk/pull/5468

### Acceptance Criteria

- toJSON, toObject exported from lisk-codec
- minimal test for format output as behaviour is already tested in the methods from the referenced PR

pablitovicente:
### What was the problem?

This PR resolves #5473

### How was it solved?

By refactoring `decodeJSON` and `encodeJSON` to use the newly exported methods `toJSON` and `toObject`

### How was it tested?

Unit Tests

### DIFF
--- a/elements/lisk-codec/src/codec.ts
+++ b/elements/lisk-codec/src/codec.ts
@@ -123,20 +123,33 @@ export class Codec {
 			schema,
 			message,
 		);
-		const decodedMessageCopy = objectUtils.cloneDeep(decodedMessage);
-		decodedMessageCopy[Symbol.iterator] = iterator;
 
-		recursiveTypeCast(
-			'toJSON',
-			decodedMessageCopy,
-			(schema as unknown) as SchemaProps,
-			[],
-		);
-		return (decodedMessageCopy as unknown) as T;
+		const jsonMessageAsObject = this.toJSON(schema, decodedMessage);
+		return (jsonMessageAsObject as unknown) as T;
 	}
 
 	// For performance applications use encode() instead!
 	public encodeJSON(schema: Schema, message: object): Buffer {
+		const objectFromJson = this.toObject(schema, message);
+		return this.encode(schema, objectFromJson);
+	}
+
+	// eslint-disable-next-line class-methods-use-this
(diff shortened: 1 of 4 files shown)


elements/lisk-codec/src/codec.ts
```diff
@@ -123,20 +123,33 @@ export class Codec {
...
 			[],
 		);
-		return (decodedMessageCopy as unknown) as T;
+		return (messageCopy as unknown) as T;
 	}
 
-	// For performance applications use encode() instead!
-	public encodeJSON(schema: Schema, message: object): Buffer {
+	// eslint-disable-next-line class-methods-use-this
+	public toObject<T = object>(schema: Schema, message: object): T {
```
nazarhussain, line 152: `toObject` and passing object as a parameter does not feel good. 

I would prefer it to rather `toJSON` and `fromJSON`, What do you think @shuse2

shuse2: that is also fine with me =)

pablitovicente: @nazarhussain I didn't pick the names but actually toObject makes sense as it returns an object with Buffer instances and BigInts that are not compatible with JSON.stringify() or well not completely. But I'm ok with whatever name you think it's ok.

nazarhussain: Other than primitives everything is object to JS. And `toObject` lost the context of JSON, So would be better to associate both of co-related functions with the JSON which their intentions are.

nazarhussain: Also, it follows the context of NodeJS libraries. e.g. You will find `Buffer.from(string)`, `Buffer.from(array)` etc.

pablitovicente: Yes. I was just following what the original author for this requested. It's changed now.

PR #5475, opened 2020-06-26
Update community identifier to be in the genesis config - Closes #5447

### LINKED ISSUE #5447: Move communityIdentifier from genesis block to configuration
### Description

Currently, the attribute `communityIdentifier` is available in the genesis block.

https://github.com/LiskHQ/lisk-sdk/blob/54ef705c99292d0c30ec71e08a881eb363bfbf40/framework/test/fixtures/config/devnet/genesis_block.json#L2

With the latest [LIP0034](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0034.md), as we defined new schema for genesis block, this attribute doe snot belongs there anymore. So we need to move this attribute to application configuration. 

### Motivation

It's a configurable attribute, which does not deal with genesis state of the blockchain, so we moved it out from the genesis block. 

### Acceptance Criteria

- Genesis block JSON files should not have `communityIdentifier` property 
- Genesis block schema should not have `communityIdentifier` property 
- Configurations JSON files should not have `communityIdentifier` property 
- Configurations schema should not have `communityIdentifier` property 

### Additional Information
Related issues #5234

shuse2:
### What was the problem?

This PR resolves #5447 

### How was it solved?

Update community identifier to be in the genesis config, and use from there

### How was it tested?

All functional tests are working where it is using the community identifier to generate network identifier


framework/src/application/schema/application_config_schema.ts
```diff
@@ -102,6 +107,11 @@ export const applicationConfigSchema = {
 					minimum: 2,
 					description: 'Slot time interval in seconds',
 				},
+				communityIdentifier: {
+					type: 'string',
+					description:
+						'Community identifier used to generate network identifier',
```
ManuGowda, line 113:
```suggestion
						'The unique name of the relevant community as a string encoded in UTF-8 format',
```

framework/test/unit/specs/application/schema/__snapshots__/application_schema.spec.ts.snap
```diff
@@ -122,6 +122,10 @@ Object {
             "minimum": 2,
             "type": "number",
           },
+          "communityIdentifier": Object {
+            "description": "Community identifier used to generate network identifier",
```
ManuGowda, line 126: same here

PR #5483, opened 2020-06-30
Improved and more readable validation error format from Lisk SDK - Closes #5432

### LINKED ISSUE #5432: Improved and more readable validation error format from Lisk SDK
### Description

The validation errors should be more human-readable instead of structured JSON Objects. 

### Motivation

Lisk SDK relies heavily on `Lisk-Validator`, and since the inclusion `Lisk Schema` for generic serialization, such validations happen quite often. The error reporting from such validation is good in detail but not good in readable. 

Currently, the errors are reported like that...

```
{
  keyword: 'additionalProperties',
  dataPath: '',
  schemaPath: '#/additionalProperties',
  params: { additionalProperty: 'version' },
  message: 'should NOT have additional properties'
},
{
  keyword: 'additionalProperties',
  dataPath: '',
  schemaPath: '#/additionalProperties',
  params: { additionalProperty: 'totalAmount' },
  message: 'should NOT have additional properties'
},
{
  keyword: 'required',
  dataPath: '',
  schemaPath: '#/required',
  params: { missingProperty: 'payload' },
  message: "should have required property 'payload'"
}
```

Normally we log and look just for error message instead of full object, for above example these become. 

```
- should NOT have additional properties
- should NOT have additional properties
- should have required property 'payload'
```

As you notice some of the error messages are confusing if you don't know the full error object. 

We have to improve those error messages to provide more and more details in just the message. That should be encaluslated as part of the `LiskValidationError` class. 

https://github.com/LiskHQ/lisk-sdk/blob/27635b03a29779797079d43d9c9dcc53b2c32ddd/elements/lisk-validator/src/errors.ts#L34-L41

### Acceptance Criteria

- If we catch any `LiskValidationError` error anywhere and log it with `console.error` or logger.error` or any other logging technique we should get more information possible.

pablitovicente:
### What was the problem?

This PR resolves #5432

### How was it solved?

By adding formatting functions for specific error kinds.

### How was it tested?

By adding unit tests


elements/lisk-validator/src/errors.ts
```diff
@@ -31,11 +30,85 @@ export interface ErrorObject {
...
+		paramAsString = '';
+	} else {
+		paramAsString = param as string;
+	}
+	return paramAsString;
+};
+
+const errorFormatter = (error: ErrorObject): string => {
+	let errorMessage = '';
+	switch (error.keyword) {
```
nazarhussain, line 55: All looks good, except I don't feel comfortable with `switch` here, it will be called here a lot of times and conditional control flow are not good if we can achieve the same with a hash map. 

If you used it because to the `default` case, that we can also achieve via. 

```js
const defaultErrorFormater = error => (error.message ?? 'unspecified validator error');
const errorFormatMaps = {
   type: error => `Property '${
 				error.dataPath ?? ''
 			}' should be of type '${errorParamToString(error.params.type)}'`,

   additionalProperties: error => `Property '${
 				error.dataPath ?? ''
 			}' has extraneous property '${errorParamToString(
 				error.params.additionalProperty,
 			)}'`
}

const errorFormatter = (error: ErrorObject): string => (errorFormatMaps[error.keyword] | defaultErrorFormater)(error);
```

elements/lisk-validator/src/errors.ts
```diff
@@ -31,11 +30,85 @@ export interface ErrorObject {
...
+		case 'required':
+			errorMessage = `Missing property, ${errorParamToString(error.message)}`;
+			break;
+		case 'dataType':
+			errorMessage = `Property '${error.dataPath ?? ''}' ${errorParamToString(
+				error.message,
+			)}`;
+			break;
+		default:
+			errorMessage = error.message ?? 'unspecified validator error';
```
nazarhussain, line 92: Use capitalize words as in other messages. 

```suggestion
			errorMessage = error.message ?? 'Unspecified validator error';
```

Issue #5484, opened 2020-06-30
Peer should not be banned by `getTransactions` endpoint


shuse2:
### Expected behavior
Honest node should be able to start syncing with another node.

### Actual behavior
The syncing node gets immediately banned by the first node with message
```
16:35:39 ERROR lisk-framework: EVENT_MESSAGE_RECEIVED: Peer has been banned temporarily (module=lisk:app)
{
 "peerId": "127.0.0.1:3333"
}
16:35:39 ERROR lisk-framework: Peer request not fulfilled event: Could not respond to peer request (module=lisk:app)
Message: undefined
 Trace: undefined
{
 "procedure": "getTransactions"
}
```

### Steps to reproduce
While one node is running with force forging, start another node which connects with the first node.

### Which version(s) does this affect? (Environment, OS, etc...)
5.0-

PR #5487, opened 2020-06-30
Add empty previous block id for fork status check - Closes #5486

### LINKED ISSUE #5486: Synchronization should end successfully
### Expected behavior
Block synchronization should end successfully

### Actual behavior
```
Message: The "otherBuffer" argument must be an instance of Buffer or Uint8Array. Received undefined
 Trace: TypeError [ERR_INVALID_ARG_TYPE]: The "otherBuffer" argument must be an instance of Buffer or Uint8Array. Received undefined
    at Uint8Array.equals (buffer.js:809:11)
    at Object.exports.isValidBlock (/lisk-sdk/elements/lisk-bft/src/fork_choice_rule.ts:46:15)
    at BFT.forkChoice (/lisk-sdk/elements/lisk-bft/src/bft.ts:136:22)
    at /lisk-sdk/framework/src/application/node/block_processor_v2.ts:140:20
    at Pipeline.run (/lisk-sdk/framework/src/application/node/processor/pipeline.ts:64:25)
    at Processor.forkStatus (/lisk-sdk/framework/src/application/node/processor/processor.ts:251:36)
    at BlockSynchronizationMechanism._computeBestPeer (/lisk-sdk/framework/src/application/node/synchronizer/block_synchronization_mechanism.ts:579:49)
    at BlockSynchronizationMechanism.run (/lisk-sdk/framework/src/application/node/synchronizer/block_synchronization_mechanism.ts:86:32)
    at Synchronizer.run (/lisk-sdk/framework/src/application/node/synchronizer/synchronizer.ts:138:25)
    at runMicrotasks (<anonymous>)
```

### Steps to reproduce
- One node forging all blocks
- Start another node from scratch wait for sync. 
- Stop the syncing node and wait for a while and start the node again

shuse2:
### What was the problem?

This PR resolves #5486 

### How was it solved?

- Original implementation expected to be not equal with string comparison, but now since it's buffer, it should be buffer.

### How was it tested?

- Link with Lisk-core
- Make a node one block behind from another then start syncing again
- it should not fail syncing


framework/src/application/node/transport/schemas.ts
```diff
@@ -34,7 +34,6 @@ export const schemas = {
 					type: 'string',
 					format: 'base64',
 				},
-				minItems: 1,
```
pablitovicente, line 37: shouldn't this be kept if it's expecting an array of transactionIds?

shuse2: ah sorry this was part of #5484 (wrong target branch)

framework/src/application/node/synchronizer/block_synchronization_mechanism.ts
```diff
@@ -571,6 +571,7 @@ export class BlockSynchronizationMechanism extends BaseSynchronizer {
 			id: Buffer.alloc(0),
 			height: selectedPeers[randomPeerIndex].height,
 			version: selectedPeers[randomPeerIndex].blockVersion,
+			previousBlockID: Buffer.alloc(0),
```
ManuGowda, line 574: if we always send empty `previousBlockID`. why not ignore this if its not required?

shuse2: It is comparing previousBlockID in the fork choice rule to decide wether to sync or not.
Previously it was string, so comparing with `undefined` was fine, but now it should be at least Buffer to use `buf.equal`

PR #5492, opened 2020-07-01
Fix IPC communication through plugins loaded as child process - Closes #3674, #3118

### LINKED ISSUE #3674: ChildProcessChannel can not subscribe
### Expected behavior

ChildProcessChannel should be able to listen events coming from other childProcessChannels.

### Actual behavior

ChildProcessChannels can not subscribe events that coming from other child process channels.

Blocking #3118 

v.2.2.0

nazarhussain:
### What was the problem?

This PR resolves #3674, #3118

### How was it solved?

Restructured the IPC communication from the ground up and added the integration tests. 

### How was it tested?

Run all integration tests


framework/src/controller/ipc/ipc_socket.ts
```diff
@@ -0,0 +1,79 @@
...
+		this.rpcServer.sock.close();
+	}
+
+	public async start(): Promise<void> {
+		await new Promise((resolve, reject) => {
+			this.rpcServer.sock.on('bind', resolve);
+			this.rpcServer.sock.on('error', reject);
+
+			this.rpcServer.sock.bind(this._actionRpcSeverSocketPath);
+		}).finally(() => {
```
shuse2, line 74: why removing all listener here? isn't it suppose to be done in the `stop`?

framework/src/controller/ipc/ipc_server.ts
```diff
@@ -0,0 +1,57 @@
...
+
+	public async start(): Promise<void> {
+		await super.start();
+
+		await new Promise((resolve, reject) => {
+			this.pubSocket.on('bind', resolve);
+			this.pubSocket.on('error', reject);
+
+			this.pubSocket.bind(this._eventPubSocketPath);
+		}).finally(() => {
```
shuse2, line 40: why removing all listener here? isn't it suppose to be done in the stop?

framework/src/controller/channels/child_process_channel.ts
```diff
@@ -46,62 +42,86 @@ export const rejectWhenTimeout = async (timeInMillis: number): Promise<void> =>
...
 		super(moduleAlias, events, actions, options);
-		this.localBus = new EventEmitter2();
+
+		this._ipcClient = new IPCClient({
+			socketsDir: options.socketsPath.root,
+			name: moduleAlias,
+			rpcServerSocketPath: `unix://${options.socketsPath.root}/bus_rpc_socket.sock`,
+		});
+
+		this._emitter = new EventEmitter2({
```
shuse2, line 60: do we still need to use EventEmitter2?

framework/src/controller/bus.ts
```diff
@@ -112,24 +116,28 @@ export class Bus extends EventEmitter2 {
...
+				.then(data => {
+					cb(null, data);
+				})
+				.catch(error => {
+					cb(error);
+				});
 		});
 
-		this.rpcServer.expose('invokePublic', (action, cb: NodeCallback) => {
+		this._rpcServer.expose('invokePublic', (action, cb: NodeCallback) => {
```
shuse2, line 129: This can be removed as no longer used

framework/src/controller/bus.ts
```diff
@@ -231,30 +262,24 @@ export class Bus extends EventEmitter2 {
...
 		// Communicate through event emitter
-		this.on(eventName, cb);
+		this._emitter.on(eventName, cb);
 
-		// Communicate through unix socket
-		if (this.config.ipc.enabled) {
-			this.subSocket?.on(eventName, cb);
-		}
+		// // Communicate through unix socket
+		// if (this.config.ipc.enabled) {
```
shuse2, line 268: Originally there was option to allow communication directly through socket or not, but that's not the case anymore?

framework/test/integration/setup.js
```diff
@@ -1,16 +0,0 @@
-/*
- * Copyright © 2019 Lisk Foundation
```
shuse2, line 2: why removing setup?

framework/test/integration/specs/controller/ipc/ipc_client.spec.ts
```diff
@@ -0,0 +1,182 @@
...
+				await new Promise(resolve => {
+					client2.subSocket.on('message', data => {
+						expect(data).toEqual('myData');
+						resolve();
+					});
+				}),
+			]);
+
+			// Cleanup
+			client2.stop();
```
shuse2, line 111: if test fails, it wont get cleaned up? same for others

framework/src/controller/ipc/ipc_socket.ts
nazarhussain: No these are listeners just for the confirmation if server bound or started. Once it started we can remove the listeners. Ongoing we have `message` listener which is removed in `stop` .

framework/src/controller/ipc/ipc_server.ts
nazarhussain: Same as above.

framework/src/controller/bus.ts
nazarhussain: Not sure if is used or not, but it was there and have the test coverage as well. Maybe we can remove it afterward and clean all references of `isPublic`

framework/src/controller/bus.ts
nazarhussain: Now this check is only in one place inside `setup`, rest is handled through local emitter.

framework/test/integration/specs/controller/ipc/ipc_client.spec.ts
nazarhussain: But if the test fails we fix those... 😄

shuse2: but that will make the test hangs or put it in unstable situation?
I think it's better to put it in after/All?

framework/src/controller/bus.ts
shuse2: Please create an issue for this

framework/src/controller/channels/child_process_channel.ts
nazarhussain: Yes, we need a local emitter. We can use native `EventEmitter` but it does not support wildcard. If one wants to listen `channel.subscribe('ap:*')` it can't be done with native `EventEmitter`. If we don't want that support anymore then we can remove `EventEmitter2`

framework/src/controller/bus.ts
nazarhussain: https://github.com/LiskHQ/lisk-sdk/issues/5509

framework/src/controller/channels/child_process_channel.ts
shuse2: Change to push/pull will not affect the wildcard feature, then no need to remove them

framework/src/controller/ipc/ipc_server.ts
```diff
@@ -0,0 +1,57 @@
...
+export class IPCServer extends IPCSocket {
+	public constructor(options: { socketsDir: string; name: string }) {
+		super(options);
+
+		this.pubSocket = axon.socket('pub', {}) as PubSocket;
+		this.subSocket = axon.socket('pull', {}) as PullSocket;
+	}
+
+	public async start(): Promise<void> {
+		await super.start();
```
ishantiw, line 33: I'm not sure but if this server is running, should we handle errors or listen to error event on the server and handle or log it?

nazarhussain: I agree that we should add more logging into it at the debug level. I will create another issue an refer it here.

nazarhussain: @ishantiw I had set `CONNECTION_TIME_OUT=2000` which is the total time for connection including all tries. Each try time out default is 100ms and max try timeout is `5000`, but with our enforcement, it will reject in 2000. 

This connection timeout is for the clients when trying to connect to the server. But for server, its feasible as server don't need to connect to any. If it faces any issues opening the port or the socket, it will fail immediately. Suggest if we need to improve it further.

ishantiw: okay, I think as of now `CONNECTION_TIME_OUT=2000` is good enough including total tries, I thought the socket only has a chance within 100ms to connect which can fail sometimes on Jenkins that's why I thought it should be increased. If there will be some failing tests later then we can rethink to change `2000` but at the moment it's good 👍

PR #5500, opened 2020-07-01
Block synchronization mechanism error - Closes #5496

### LINKED ISSUE #5496: Block synchronization mechanism error
### Expected behavior
A node should be able to sync with other nodes who share common block with each other.

### Actual behavior
```
10:57:57 INFO lisk-framework: Applying penalty to peer and restarting synchronizer (module=lisk:app)
{
 "peerId": "127.0.0.1:5000",
 "reason": "No common block has been found between the chain and the targeted peer"
}
10:57:57 ERROR lisk-framework: Error occurred during block synchronization mechanism. (module=lisk:app)
Message: Synchronizer is already running
 Trace: Error: Synchronizer is already running
```

### Steps to reproduce
- Run a node `devnet` enable foring and let it forge for few blocks
- Run another node `devnet2` and disable forging
- The `devnet2` fails to sync with `devnet`

development, v5

ManuGowda:
### What was the problem?

This PR resolves #5496

### How was it solved?

- Handle startHeight calculation for round `0`
- Update condition check for genesis block height

### How was it tested?

- Added unit test

### DIFF
--- a/framework/src/application/node/synchronizer/utils.ts
+++ b/framework/src/application/node/synchronizer/utils.ts
@@ -117,7 +117,7 @@ export const computeBlockHeightsList = (
 	listSizeLimit: number,
 	currentRound: number,
 ): number[] => {
-	const startingHeight = (currentRound - 1) * activeDelegates;
+	const startingHeight = Math.max((currentRound - 1) * activeDelegates, 0);
 	const heightList = new Array(listSizeLimit)
 		.fill(0)
 		.map((_, i) => startingHeight - i * activeDelegates)
--- a/framework/test/unit/specs/application/node/synchronizer/block_synchronization_mechanism/block_synchronization_mechanism.spec.ts
+++ b/framework/test/unit/specs/application/node/synchronizer/block_synchronization_mechanism/block_synchronization_mechanism.spec.ts
@@ -1263,6 +1263,22 @@ describe('block_synchronization_mechanism', () => {
 				});
 			});
 		});
+
+		describe('computeBlockHeightsList', () => {
+			it('should return height list for round 0', () => {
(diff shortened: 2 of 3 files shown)


framework/test/unit/specs/application/node/synchronizer/block_synchronization_mechanism/block_synchronization_mechanism.spec.ts
```diff
@@ -1263,6 +1263,22 @@ describe('block_synchronization_mechanism', () => {
 				});
 			});
 		});
+
+		describe('computeBlockHeightsList', () => {
+			it('should return height list for round 0', () => {
```
shuse2, line 1268: Since this is util function, maybe move the test to utils?
Also, maybe add more detailed expectation as it is pure function

framework/src/application/node/synchronizer/utils.ts
```diff
@@ -117,7 +117,7 @@ export const computeBlockHeightsList = (
 	listSizeLimit: number,
 	currentRound: number,
 ): number[] => {
-	const startingHeight = (currentRound - 1) * activeDelegates;
+	const startingHeight = Math.max((currentRound - 1) * activeDelegates, 0);
```
nazarhussain, line 120: In earlier review, it was like that, but we discussed that this value will never gonna be below zero. So which triggers to put back this logic?

ManuGowda: when `currentRound` is 0, then `0-1 * 103` will be `-103` and this happens due to `this.dpos.rounds.calcRound(
			this._chain.lastBlock.header.height,
		)` for height 0

PR #5515, opened 2020-07-03
Fix plugin loading in child process and prevent application from crash if child process crashes - Closes #3142

### LINKED ISSUE #3142: Fix plugin loading in child process and prevent application from crash if child process crashes

nazarhussain:
### What was the problem?

This PR resolves #3142

### How was it solved?

Fixed loading of child process plugins and prevent application from crashing. 

### How was it tested?

1. Run the test app with `node framework/test/test_app/index.js`
2. Check process tree with `htop` or any other utility 
3. You will notice a process named `devnet` and a child process running under as http-api-plugin
4. Kill that child process for http-api-plugin
5. Check if app shows proper log for plugin to exit
6. Check if app still keep running


framework/src/application/application.ts
```diff
@@ -176,7 +176,9 @@ export class Application {
 
 	public registerPlugin(
 		pluginKlass: typeof BasePlugin,
-		options = {},
+		options: { [key: string]: any; loadAsChildProcess: boolean } = {
```
ManuGowda, line 179: maybe extract this to type.. in future if we expand it might grow and in general we can reuse if required?
`PluginOptions`?

framework/src/application/application.ts
```diff
@@ -199,7 +201,7 @@ export class Application {
 		this._plugins[pluginAlias] = pluginKlass as InstantiablePlugin<BasePlugin>;
 	}
 
-	public overrideModuleOptions(alias: string, options?: object): void {
+	public overridePluginOptions(alias: string, options?: object): void {
```
ManuGowda, line 204: after defining the above type, u could use the type instead of `object`?

framework/src/application/node/node.ts
```diff
@@ -209,7 +209,7 @@ export class Node {
 				Object.assign(this._applicationState, event.data);
 			});
 
-			this._logger.info('Modules ready and launched');
+			this._logger.info('Node ready and launched');
```
ManuGowda, line 212: since process, sync initiated below, is this log line correct? maybe it should below that?

framework/src/controller/child_process_loader.ts
```diff
@@ -19,40 +19,43 @@ import { InstantiablePlugin, BasePlugin } from '../plugins/base_plugin';
...
+];
 
-const _loadModule = async (
+const _loadPlugin = async (
 	config: {
 		[key: string]: unknown;
 		socketsPath: SocketPaths;
 	},
-	moduleOptions: object,
+	pluginOptions: object,
```
ManuGowda, line 33: same here

framework/src/application/application.ts
nazarhussain: We are not sure what configuration we get fro any plugin.

framework/src/application/node/node.ts
nazarhussain: I think its correct here, Node is ready and launched. Rest is its internal processing.

PR #5516, opened 2020-07-03
Add getSchema action on framework - Closes #5497

### LINKED ISSUE #5497: Add getSchema action on framework
### Description

- Add `getSchema` action which returns schema below
```
getSchema: () => {
  account: {},
  blockHeader: {},
  blockHeadersAssets: { [version: number]: schema },
  baseTransaction: {},
  transactionsAssets: { [type: number}: schema },
}
```
Obtain block and account schema from the chain and transactions schema from registered transactions

- Update base plugin to register to app:ready event and call getSchema action there to cache schemas in the base plugin.

### Motivation

When plugin receives encoded data through event or action, it needs to be able to decode it back in order to understand the content

### Acceptance Criteria

- Add functional test for action and check if it returns the schemas

pablitovicente:
### What was the problem?

This PR resolves #5497

### How was it solved?

By adding new action getSchema()
By exposing new action getSchema() from within applcation
By adding caching of schemas in base plugin

### How was it tested?

Functional test added for application's getSchema()


framework/src/application/node/node.ts
```diff
@@ -474,6 +480,16 @@ export class Node {
 					params.data,
 					params.peerId,
 				),
+			getSchema: () => ({
+				account: this._chain.accountSchema,
+				baseBlockHeader: blockSchema,
+				blockHeaders: {
```
nazarhussain, line 486:
```suggestion
				blockHeadersAssets: {
```

framework/src/application/node/node.ts
```diff
@@ -474,6 +480,16 @@ export class Node {
 					params.data,
 					params.peerId,
 				),
+			getSchema: () => ({
+				account: this._chain.accountSchema,
+				baseBlockHeader: blockSchema,
```
nazarhussain, line 485:
```suggestion
				blockHeader: blockSchema,
```

framework/src/application/node/node.ts
```diff
@@ -474,6 +480,16 @@ export class Node {
...
+				blockHeaders: {
+					0: BlockProcessorV0.schema,
+					2: BlockProcessorV2.schema,
+				},
+				baseTransaction: BaseTransaction.BASE_SCHEMA,
```
nazarhussain, line 490:
```suggestion
				transaction: BaseTransaction.BASE_SCHEMA,
```

framework/src/application/node/node.ts
```diff
@@ -474,6 +480,16 @@ export class Node {
...
+				transactions: this._getRegisteredTransactionSchemas(),
```
nazarhussain, line 491:
```suggestion
				transactionsAssets: this._getRegisteredTransactionSchemas(),
```

pablitovicente: The request in the issue was to use transactions @shuse2 ?

framework/src/application/node/node.ts
pablitovicente: I think base transaction is ok here why change as it's literally base transaction schema?

framework/src/application/node/node.ts
pablitovicente: The request in the issue was to use baseBlockHeader @shuse2 ?

framework/src/application/node/node.ts
pablitovicente: The request in the issue was to use blockHeaders @shuse2 ?

framework/src/application/node/node.ts
nazarhussain: Would suggest updating the issue, as for both transactions and block headers, we can specify custom assets schemas not the full transaction and block schema.

PR #5519, opened 2020-07-06
Add decode transaction support in base plugin - Partially Resolves #5251

### LINKED ISSUE #5251: Update GET /api/transactions/:id and POST /api/transactions endpoints
### Description

Add the API endpoint:
- `GET /api/transactions/:id`
- `POST /api/transactions`

- Validate if id is base64 if not `400` should be returned
- `Get` should return decoded transaction
-  It should return proper error message when `post` request transaction is invalid
- It should accept transaction in JSON format

### Motivation

The HTTP API should return transactions by id.

### Acceptance Criteria

- Add functional tests
- Check if all registered transaction is accepted
- Check if invalid transaction is rejected

### Additional Information

Related issues: #5213

### LINKED ISSUE #5512: Update child process IPC communication integration test
### Expected behavior
Child process, IPC client and IPC server integration tests should pass consistently.

### Actual behavior
On jenkins, it was hanging consistently.
https://jenkins.lisk.io/job/lisk-sdk/job/PR-5503/10/

### Steps to reproduce
Re-enable the test and run test on jenkins. 
It was not reproducible in some local environment

nazarhussain:
### What was the problem?

This PR partially resolves #5251

### How was it solved?

This will partially resolve the http api endpoint issues. 

### How was it tested?

Run all tests


framework/src/plugins/base_plugin.ts
```diff
@@ -32,12 +34,86 @@ export interface InstantiablePlugin<T, U = object> {
...
+	readonly id: string;
+	readonly asset: object;
+}
+
+const decodeTransactionToJSON = (
+	transactionBuffer: Buffer,
+	baseSchema: Schema,
+	assetsSchemas: { [key: number]: Schema },
+): TransactionJSON => {
+	const baseTransaction = codec.toJSON<BaseTransactionJSON>(
```
shuse2, line 63: `toJSON` is not the correct method to use. it should be `decodeJSON`.
toJSON is for object = JSON conversion.
Also, do we really want to convert to JSON here? if there are specific calculation done, for example `fee`, it needs to converted back to BigInt again.

framework/test/unit/specs/plugins/base_pluign.spec.ts
```diff
@@ -0,0 +1,107 @@
+/*
```
mitsuaki-u, line 1: typo in file name should be `plugin`

framework/src/plugins/base_plugin.ts
```diff
@@ -32,12 +34,93 @@ export interface InstantiablePlugin<T, U = object> {
...
+		blockHeader: Schema;
+		blockHeadersAssets: {
+			[key: number]: Schema;
+		};
+		baseTransaction: Schema;
+		transactionsAssets: {
+			[key: number]: Schema;
+		};
+	};
+	public codec: object;
```
shuse2, line 99: this should be typed correctly, otherwise it would be hard to use

framework/src/controller/event.ts
```diff
@@ -22,7 +22,7 @@ export interface EventInfoObject {
 	readonly data: object;
 }
 
-export type EventCallback = (action: EventInfoObject) => void;
+export type EventCallback = (action: EventInfoObject) => void | Promise<void>;
```
shuse2, line 25: why the type here is changed?

nazarhussain: To register a call back function with async. 

```js
channel.once('app:ready', async () => {
```

PR #5524, opened 2020-07-07
Update GET /api/accounts/:address endpoint - Closes #5250

### LINKED ISSUE #5250: Update GET /api/accounts/:address endpoint

pablitovicente:
### What was the problem?

This PR resolves #5250

### How was it solved?

- By adding controller for accounts
- By adding decoding to JSON in base plugin

### How was it tested?

- run `node test/test_app/index.js`
- send request: `http://localhost:4000/api/accounts/nQFJsJYtRL%2FAip9k1a%2FOtigdf7U%3D` should return account
- send request: `http://localhost:4000/api/accounts/nQFJsJYtRL%2FAip8k1a%2FOtigdf7U%3D` and should return 404 with message
- send request: `http://localhost:4000/api/accounts/-invalid` should fail with 400 and message
- Functional tests also added

### DIFF
--- a/framework-plugins/lisk-framework-http-api-plugin/src/controllers/account.ts
+++ b/framework-plugins/lisk-framework-http-api-plugin/src/controllers/account.ts
@@ -20,7 +20,9 @@ export const getAccount = (channel: BaseChannel, codec: PluginCodec) => async (
 	res: Response,
 	next: NextFunction,
 ): Promise<void> => {
-	if (!isBase64String(req.params.address)) {
+	const accountAddres = req.params.address;
+
+	if (!isBase64String(accountAddres)) {
 		res.status(400).send({
 			errors: [{ message: 'The Address parameter should be a base64 string.' }],
 		});
@@ -28,7 +30,7 @@ export const getAccount = (channel: BaseChannel, codec: PluginCodec) => async (
 
 	try {
 		const account: Buffer = await channel.invoke('app:getAccount', {
-			address: req.params.address,
+			address: accountAddres,
 		});
 		res.status(200).send(codec.decodeAccount(account));
 	} catch (err) {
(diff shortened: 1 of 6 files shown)


framework-plugins/lisk-framework-http-api-plugin/test/functional/account.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+ */
+import { Application } from 'lisk-framework';
+import axios from 'axios';
+import {
+	createApplication,
+	closeApplication,
+	getURL,
+} from './utils/application';
+
+describe('Hello endpoint', () => {
```
ManuGowda, line 22: Accounts

pablitovicente: Fixed

framework-plugins/lisk-framework-http-api-plugin/src/controllers/index.ts
```diff
@@ -12,4 +12,7 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
+import * as accounts from './account';
```
shuse2, line 15: i think `export * as accounts from './account'` should work

framework-plugins/lisk-framework-http-api-plugin/src/controllers/account.ts
```diff
@@ -0,0 +1,50 @@
...
+		});
+		res.status(200).send(codec.decodeAccount(account));
+	} catch (err) {
+		if (
+			/^Specified key accounts:address:(.*)does not exist/.test(
+				(err as Error).message,
+			)
+		) {
+			res.status(404).send({
+				message: `Account with address '${accountAddres}' was not found`,
```
shuse2, line 44: 404 case also should follow the same error format

framework-plugins/lisk-framework-http-api-plugin/src/controllers/index.ts
pablitovicente: Done.

PR #5538, opened 2020-07-09
Add http-api for GET api/peers?limit=xxx&offset=yyy&state=zzz endpoint - Closes #5254

### LINKED ISSUE #5254: Update GET /api/peers?limit=xxx&offset=yyy&state=zzz endpoint
### Description

Add the endpoints:

- `GET /api/peers?limit=xxx&offset=yyy&state=zzz`

- It should return decoded peers
- Validate if limit and offset are valid numbers
- Validate if state is either `connected` or `disconnected`
- If query params are not specified, it should default to `limit=100, offset=0, state=connected`

### Motivation

The HTTP API should return the stored peers in the attached node.

### Acceptance Criteria

- Add unit tests
- Check if invalid query returns 400 with error message

### Additional Information

Related issues: #5213

ishantiw:
### What was the problem?

This PR resolves #5254

### How was it solved?

- Add GET `api/peers` URL and Controller
- Add functional test

### How was it tested?

Run `npm run test:functional peers.spec.ts` under `framework-plugins/lisk-framework-http-api-plugin/`

### DIFF
--- a/framework-plugins/lisk-framework-http-api-plugin/src/controllers/peers.ts
+++ b/framework-plugins/lisk-framework-http-api-plugin/src/controllers/peers.ts
@@ -1,0 +1,86 @@
+/*
+ * Copyright © 2020 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+import { Request, Response, NextFunction } from 'express';
+import { isUInt32, isString } from '@liskhq/lisk-validator';
+import { BaseChannel } from 'lisk-framework';
+
+export enum PeerState {
+	connected = 'connected',
+	disconnected = 'disconnected',
+}
+
+interface PeerInfo {
+	readonly ipAddress: string;
+	readonly port: number;
+	readonly networkId: string;
+	readonly networkVersion: string;
+	readonly nonce: string;
+	readonly options: { [key: string]: unknown };
+}
+
+const filterPeers = (
+	peers: ReadonlyArray<PeerInfo>,
+	limit: number,
+	offset: number,
+): ReadonlyArray<PeerInfo> => {
+	if (offset === 0) {
+		return peers.slice(0, Math.min(limit, peers.length));
+	}
+
+	return peers.slice(offset, Math.min(limit + offset, peers.length));
+};
+
+export const getPeers = (channel: BaseChannel) => async (
+	req: Request,
+	res: Response,
+	_next: NextFunction,
+): Promise<void> => {
+	const { limit = 100, offset = 0, state = PeerState.connected } = req.query;
+
+	if (
+		!isUInt32(Number(limit)) ||
+		!isUInt32(Number(offset)) ||
+		!isString(state) ||
+		!(state === PeerState.connected || state === PeerState.disconnected)
+	) {
+		res.status(400).send({
+			errors: [
+				{
+					message:
+						'Invalid param value(s), limit and offset should be a valid number and state can be either "connected" or "disconnected"',
+				},
+			],
+		});
+		return;
+	}
+
+	try {
+		let peers;
+		if (state === PeerState.disconnected) {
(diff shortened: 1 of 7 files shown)


framework-plugins/lisk-framework-http-api-plugin/src/controllers/peers.ts
```diff
@@ -0,0 +1,86 @@
...
+	peers: ReadonlyArray<PeerInfo>,
+	limit: number,
+	offset: number,
+): ReadonlyArray<PeerInfo> => {
+	if (offset === 0) {
+		return peers.slice(0, Math.min(limit, peers.length));
+	}
+
+	return peers.slice(offset, Math.min(limit + offset, peers.length));
+};
```
nazarhussain, line 42: Please make it a generic utility function like `paginate(data, limit, offset)` , then rest can use it as well. Its very common for other API endpoints.

framework-plugins/lisk-framework-http-api-plugin/src/controllers/peers.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export const getPeers = (channel: BaseChannel) => async (
+	req: Request,
+	res: Response,
+	_next: NextFunction,
```
nazarhussain, line 47:
```suggestion
```

framework-plugins/lisk-framework-http-api-plugin/src/controllers/peers.ts
```diff
@@ -0,0 +1,86 @@
...
+		res.status(400).send({
+			errors: [
+				{
+					message:
+						'Invalid param value(s), limit and offset should be a valid number and state can be either "connected" or "disconnected"',
+				},
+			],
+		});
+		return;
+	}
```
nazarhussain, line 66: Doing this is also fine, but would be nice if simply defined a json-schema for all API parameters and rely on validator errors. This way error messages can be consistent across the endpoints.

framework-plugins/lisk-framework-http-api-plugin/src/controllers/peers.ts
```diff
@@ -0,0 +1,86 @@
...
+		peers = filterPeers(peers, +limit, +offset);
+
+		res.status(200).send(peers);
+	} catch (err) {
+		res.status(500).send({
+			errors: [
+				{ message: `Something went wrong while fetching peers list: ${(err as Error).message}` },
+			],
+		});
+	}
```
nazarhussain, line 85: This can be handled by error_handler middleware automatically.  Any error thrown by middleware or route handlers should be catched and processed by error_handler. If it snot doing it then there would be a problem to look at.

framework-plugins/lisk-framework-http-api-plugin/src/controllers/peers.ts
```diff
@@ -0,0 +1,86 @@
...
+import { Request, Response, NextFunction } from 'express';
+import { isUInt32, isString } from '@liskhq/lisk-validator';
+import { BaseChannel } from 'lisk-framework';
+
+export enum PeerState {
+	connected = 'connected',
+	disconnected = 'disconnected',
+}
+
+interface PeerInfo {
```
mitsuaki-u, line 23: Can you import these types from `lisk-p2p`

framework-plugins/lisk-framework-http-api-plugin/src/controllers/peers.ts
```diff
@@ -0,0 +1,86 @@
...
+
+	return peers.slice(offset, Math.min(limit + offset, peers.length));
+};
+
+export const getPeers = (channel: BaseChannel) => async (
+	req: Request,
+	res: Response,
+	_next: NextFunction,
+): Promise<void> => {
+	const { limit = 100, offset = 0, state = PeerState.connected } = req.query;
```
mitsuaki-u, line 49: Current api gives a default limit of `10`

framework-plugins/lisk-framework-http-api-plugin/src/controllers/peers.ts
ishantiw: we are trying not to have a dependency on p2p lib so not importing from there

framework-plugins/lisk-framework-http-api-plugin/src/controllers/peers.ts
ishantiw: where is that limit exactly? in the issue we have default limit value as 100 for peers list

PR #5539, opened 2020-07-09
Update GET /api/blocks?height=x and GET /api/blocks/:id endpoints - Close #5249

### LINKED ISSUE #5249: Update GET /api/blocks?height=x and GET /api/blocks/:id endpoints
### Description

Add the API endpoints:

- ` GET /api/blocks?height=x`
- `GET /api/blocks/:id`

- id should be in base64 format
- Both endpoints should return a deserialized full block
- Validate if height is number if not `400` should be returned
- Validate if id is base64 formatted string if not `400` should be returned
- Validate `/blocks` endpoint has height query if not `400` should be returned
- Use actions to obtain the data from the framework

### Motivation

The HTTP API should return blocks by height or id.

### Acceptance Criteria

- Block id input must be base64
- Add functional tests
- Register plugin to core, and check if it returns decoded data with height
- Check if invalid height returns 400 with error message
- Check if returns 404 with error message if the block does not exist

### Additional Information

- Check if `/api/blocks` returns 400 with error message

Related issues #5213

pablitovicente:
### What was the problem?

This PR resolves #5249

### How was it solved?

- By adding endpoint `api/blocks/:blockID`
- By adding endpoint `api/blocks/?height=`

### How was it tested?

- Added functional tests


pablitovicente:
Will close and re-branch again.

PR #5543, opened 2020-07-09
Update genesis block to be automatically parsed from JSON - Closes #5513

### LINKED ISSUE #5513: Update genesis block and config reading to use fromJSON function
### Description
Remove the functions which manually convert JSON format to instance format.
- Add dataType definition on the application config and constants
- Remove genesis block conversion function
- Use `fromJSON` with the defined schema for those 2 cases to convert from JSON

### Motivation

With newly implemented function `toJSON` and `fromJSON` of `lisk-codec`, JSON value which has lisk schema can be converted to instance with those function.

### Acceptance Criteria

- Config and genesis schema should be converted from JSON format using `fromJSON` function fro m`codec`

shuse2:
### What was the problem?

This PR resolves #5513 

### How was it solved?

- Remove manual conversion of genesis block and use schema and fromJSON to convert to instance

### How was it tested?

- All test should work as it is

### DIFF
--- a/framework/src/application/genesis_block.ts
+++ b/framework/src/application/genesis_block.ts
@@ -13,6 +13,7 @@
  */
 import { GenesisBlock, genesisBlockHeaderAssetSchema } from '@liskhq/lisk-genesis';
 import { blockHeaderSchema, blockSchema, baseAccountSchema } from '@liskhq/lisk-chain';
+import { objects } from '@liskhq/lisk-utils';
 import { Schema, codec } from '@liskhq/lisk-codec';
 import { AccountAsset, accountAssetSchema } from './node/account';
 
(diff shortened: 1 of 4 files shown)


framework/src/application/genesis_block.ts
```diff
@@ -65,52 +63,44 @@ export interface GenesisAccountStateJSON {
...
+									asset: {
+										...baseAccountSchema.properties.asset,
+										properties: accountSchema,
+									},
+								},
+							},
+						},
+					},
+				},
+			},
```
nazarhussain, line 97: I think this will look easier to read and understand

```js
export const genesisBlockSchema = (accountSchema: object): Schema => mergeDeep({}, 
    blockSchema, 
    genesisBlockHeaderAssetSchema, {
		properties: {
			accounts: {
				items: {
					properties: {
						asset: accountSchema,
					},
				},
			},
		},
	});
```

framework/src/application/genesis_block.ts
```diff
@@ -65,52 +63,44 @@ export interface GenesisAccountStateJSON {
...
-	keys: {
-		mandatoryKeys: account.keys.mandatoryKeys.map(key => Buffer.from(key, 'base64')),
-		optionalKeys: account.keys.optionalKeys.map(key => Buffer.from(key, 'base64')),
-		numberOfSignatures: account.keys.numberOfSignatures,
-	},
-	asset: {
-		delegate: {
-			...account.asset.delegate,
-			totalVotesReceived: BigInt(account.asset.delegate.totalVotesReceived),
+export const genesisSchema = (accountSchema: object): Schema => ({
```
pablitovicente, line 66: maybe better to call this `genesisBlockSchema` as one it's de-structured it loosed the "genesis block" context.

PR #5545, opened 2020-07-10
Update GET /api/blocks?height=x and GET /api/blocks/:id endpoints - Closes #5249

### LINKED ISSUE #5249: Update GET /api/blocks?height=x and GET /api/blocks/:id endpoints

pablitovicente:
### What was the problem?

This PR resolves #5249

### How was it solved?

By adding http plugin route `/api/blocks/?height=<block height>`
By adding http plugin route `/api/blocks/<block id>`

### How was it tested?

- Functional tests added


framework-plugins/lisk-framework-http-api-plugin/test/functional/blocks.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+// import { codec } from '@liskhq/lisk-codec';
```
ManuGowda, line 14: remove if not required?

framework-plugins/lisk-framework-http-api-plugin/test/functional/blocks.spec.ts
```diff
@@ -0,0 +1,162 @@
...
+			expect(result.data).toHaveProperty('data.header.timestamp');
+			expect(result.data).toHaveProperty('data.header.height');
+			expect(result.data).toHaveProperty('data.header.previousBlockID');
+			expect(result.data).toHaveProperty('data.header.transactionRoot');
+			expect(result.data).toHaveProperty('data.header.generatorPublicKey');
+			expect(result.data).toHaveProperty('data.header.reward');
+			expect(result.data).toHaveProperty('data.header.asset');
+			expect(result.data).toHaveProperty('data.header.asset.maxHeightPreviouslyForged');
+			expect(result.data).toHaveProperty('data.header.asset.maxHeightPrevoted');
+			expect(result.data).toHaveProperty('data.header.asset.seedReveal');
```
ManuGowda, line 56: Just a suggestion.
```
expect(Object.keys(result.data)).to.equal(['header', 'payload'])
expect(Object.keys(result.data.header)).to.equal(['id', 'version']) so on 
expect(Object.keys(result.data.header.asset)).to.equal(['maxHeightPreviouslyForged']) so on
```

framework-plugins/lisk-framework-http-api-plugin/test/functional/blocks.spec.ts
```diff
@@ -0,0 +1,162 @@
...
```
ManuGowda, line 117: if you do not like the above suggestion, then wrap the expectation to function and something like, blockValidation(result). and reuse the same above.

framework-plugins/lisk-framework-http-api-plugin/test/functional/blocks.spec.ts
pablitovicente: Changed to this.

framework-plugins/lisk-framework-http-api-plugin/test/functional/blocks.spec.ts
pablitovicente: Followed the other suggestion as right now are similar but maybe in the future it changes. Also I prefer not to have tests with much helpers that hide what's going on too much.

PR #5546, opened 2020-07-10
Add http-api for GET api/node/transactions?limit=xxx&offset=yyy - Closes #5253

### LINKED ISSUE #5253: Update GET /api/node/transactions?limit=xxx&offset=yyy endpoint
### Description

Add the endpoints:

- `GET /api/node/transactions?limit=xxx&offset=xxx`

- It should return a deserialized transaction
- Validate if limit and offset are valid numbers
- If limit or offset is not defined, it should default to `limit=100`, `offset=0`

### Motivation

The HTTP API should return the transactions currently in the pool.

### Acceptance Criteria

- Add unit tests
- Check if invalid limit or offset returns 400 with error message

### Additional Information

Related issues: #5213

mitsuaki-u:
### What was the problem?

Endpoint to retrieve transactions in pool from node was needed.

This PR resolves #5253 

### How was it solved?

-  Add GET api/node/transactions URL and Controller
-  Add functional test

### How was it tested?

Run `npm run test:functional node.spec` under `framework-plugins/lisk-framework-http-api-plugin/`

### DIFF
--- a/framework-plugins/lisk-framework-http-api-plugin/src/controllers/node.ts
+++ b/framework-plugins/lisk-framework-http-api-plugin/src/controllers/node.ts
@@ -34,8 +34,9 @@ export const getTransactions = (channel: BaseChannel, codec: PluginCodec) => asy
 	next: NextFunction,
 ): Promise<void> => {
 	const { limit, offset } = req.query;
-	let limitNumber = 0;
+	let limitNumber = 10;
 	let offsetNumber = 0;
+
 	if (limit) {
 		if (!isNumberString(limit)) {
 			res.status(400).send({
@@ -70,19 +71,12 @@ export const getTransactions = (channel: BaseChannel, codec: PluginCodec) => asy
 		decodedTransactions.push(codec.decodeTransaction(transaction));
 	}
 
-	let data = decodedTransactions;
-
-	if (offset) {
-		data = decodedTransactions.slice(offsetNumber, decodedTransactions.length);
-	}
-
-	if (limit) {
-		data = data.slice(0, limitNumber);
-	}
-
 	// 200 - Response
 	res.status(200).json({
-		data,
+		data: decodedTransactions.slice(
+			offsetNumber || 0,
+			Math.min(limitNumber + offsetNumber, decodedTransactions.length),
+		),
 		meta: { limit: limitNumber, offset: offsetNumber, total: decodedTransactions.length },
 	});
 };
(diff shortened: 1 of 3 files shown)


framework-plugins/lisk-framework-http-api-plugin/src/controllers/node.ts
```diff
@@ -26,3 +27,56 @@ export const getNodeInfo = (channel: BaseChannel) => async (
...
+
+	if (limit) {
+		if (!isNumberString(limit)) {
+			res.status(400).send({
+				errors: [{ message: 'The limit query parameter should be a number.' }],
+			});
+			return;
+		}
+		limitNumber = Number(limit);
+	}
```
ManuGowda, line 48:
```suggestion
		if (!limit || !isNumberString(limit)) {
			res.status(400).send({
				errors: [{ message: 'The limit query parameter should be a number.' }],
			});
			return;
		}
		limitNumber = Number(limit);
```

framework-plugins/lisk-framework-http-api-plugin/src/controllers/node.ts
```diff
@@ -26,3 +27,56 @@ export const getNodeInfo = (channel: BaseChannel) => async (
...
+
+	if (offset) {
+		if (!isNumberString(offset)) {
```
ManuGowda, line 51: same here

framework-plugins/lisk-framework-http-api-plugin/src/controllers/node.ts
```diff
@@ -26,3 +27,56 @@ export const getNodeInfo = (channel: BaseChannel) => async (
...
+			});
+			return;
+		}
+		offsetNumber = Number(offset);
+	}
+
+	let transactionsInPool;
+
+	try {
+		transactionsInPool = await channel.invoke<string>('app:getTransactionsFromPool');
```
ManuGowda, line 63: `channel.invoke<string>` looks suspicious, the result will be collection rite? so `<string>` is invalid?

framework-plugins/lisk-framework-http-api-plugin/src/controllers/node.ts
```diff
@@ -26,3 +27,56 @@ export const getNodeInfo = (channel: BaseChannel) => async (
...
+	for (const transaction of transactionsInPool) {
+		decodedTransactions.push(codec.decodeTransaction(transaction));
+	}
+
+	// 200 - Response
+	res.status(200).json({
+		data: decodedTransactions.slice(
+			offsetNumber || 0,
+			Math.min(limitNumber + offsetNumber, decodedTransactions.length),
+		),
```
ManuGowda, line 79: maybe slice the `transactionsInPool` collection, that way we could avoid extensive decoding and less iterations too?

PR #5547, opened 2020-07-10
Update lastBlockId to be bytes and update protocolVersion to networkVersion

### LINKED ISSUE #5528: Update application state to have only fields that are required in framework
### Expected behavior

Only necessary fields should be present in application state that is required in other places in the framework.

### Actual behavior

Right now state gives out below fields in which some of them are not required by others
```typescript
{
	                os,
			version,
			wsPort,
			protocolVersion,
			height,
			blockVersion,
			maxHeightPrevoted,
			networkId,
}
```

### Steps to reproduce

check `framework/application/application_state.ts`

`5.0`

### LINKED ISSUE #5544: lastBlockId should be type bytes
### Expected behavior
`lastBlockId` should be named `lastBlockID` with type `bytes` and handled as bytes

### Actual behavior
Currently `lastBlockId` propagates as node info through P2P is uint32, but it should be bytes.
Also, in the code it is treated as string.

### Steps to reproduce
N/A

5.0-

shuse2:
### What was the problem?

This PR resolves #5544 and resolves #5528 

### How was it solved?

- Remove application state as it was just proxying the information between node and network
- Update lastBlockId => lastBlockID and change the type to `bytes`
- Update all remaining protocolVersion to networkVersion

### How was it tested?

- Update all the test
- Connect 2 nodes and see if they have valid information


ManuGowda:
@shuse2 tests are failing, maybe protocol spec needs to update to reflect networkVersion changes.

PR #5549, opened 2020-07-13
Remove all references of public actions feature from controller - Closes #5509

### LINKED ISSUE #5509: Remove all references of public actions feature from controller
### Expected behavior

There should be no reference, usage or code related to public actions any more in the controller layer. 

### Actual behavior

Earlier we had the support that any plugins can define or white list any action by specifying `isPublic: true` in action definition. Anyone can invoke those actions remotely from other nodes. This support is not more available on the `P2P` layer. So all actions for any plugin are now just private actions, can only be invoked by plugins running on same node. 

Any reference to `isPublic` or `invokePublic` should be removed from the controller. 

5.0

nazarhussain:
### What was the problem?

This PR resolves #5509

### How was it solved?

- Removed all references and tests for `isPublic` 
- Removed all references and tests for `invokePublic` 
- Added hard coded white list of actions for network 
  -  app:getTransactions
  -  app:getLastBlock
  -  app:getBlocksFromId
  -  app:getHighestCommonBlock
- Add penalty to peer if peer is invoked any other action

### How was it tested?

Run all tests


framework/src/application/network/network.ts
```diff
@@ -308,13 +313,25 @@ export class Network {
...
+				// Send an error back to the peer.
+				request.error(error);
+				return;
+			}
+
 			try {
-				const result = await this._channel.invokePublic<liskP2P.p2pTypes.P2PNodeInfo>(
-					sanitizedProcedure,
+				const result = await this._channel.invoke<liskP2P.p2pTypes.P2PNodeInfo>(
+					`app:${request.procedure}`,
```
ishantiw, line 334: we can also remove `app:` prefix from here?

nazarhussain: No its necessary here to route right action to the Bus. As of now `Bus` and `Channel` are all have a particular prefix, if we remove it will throw error to not able to identify action.  @ishantiw @shuse2

PR #5558, opened 2020-07-14
Fix unlock transaction validate asset error path - Closes #5557

### LINKED ISSUE #5557: Fix unlock transaction validate asset error path
### Expected behavior
Error path should be valid path, `.asset.unlockObjects.amount`.

### Actual behavior
Amount path is invalidd `.asset.votes.amount`

### Steps to reproduce
https://github.com/LiskHQ/lisk-sdk/blob/bd3ef1c75bdaf03236576716eeb18390cbd1aa3c/elements/lisk-transactions/src/14_unlock_transaction.ts#L103

development

ManuGowda:
### What was the problem?

This PR resolves #5557 

### How was it solved?

- Updated error path

### How was it tested?

- Updated unit test

### DIFF
--- a/elements/lisk-transactions/src/14_unlock_transaction.ts
+++ b/elements/lisk-transactions/src/14_unlock_transaction.ts
@@ -100,7 +100,7 @@ export class UnlockTransaction extends BaseTransaction {
 					new TransactionError(
 						'Amount cannot be less than or equal to zero',
 						this.id,
-						'.asset.votes.amount',
+						'.asset.unlockObjects.amount',
 						unlock.amount.toString(),
 					),
 				);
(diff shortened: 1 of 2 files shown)


elements/lisk-transactions/src/14_unlock_transaction.ts
```diff
@@ -95,12 +95,12 @@ export class UnlockTransaction extends BaseTransaction {
 		const errors = [];
 
 		for (const unlock of this.asset.unlockObjects) {
-			if (unlock.amount <= BigInt(0)) {
```
shuse2, line 98: original logic is correct amount cannot be less than or equal to `0`

ManuGowda: yes, but validator checks for amount to be not negative.. so its not required?

pablitovicente: Even if validator checks for negative it would be nice for this part of the business logic to be maintained in the transaction itself so as to it to be more clear on its working.

Issue #5556, opened 2020-07-13
Add an action to expose all the delegates


ishantiw:
### Expected behavior

There should be an action to return all the delegates

### Actual behavior

Currently, there is no action that returns all the delegates

### Steps to reproduce

Try to fetch all the delegates from the application through actions

### Which version(s) does this affect? (Environment, OS, etc...)

`5.0`

Issue #5554, opened 2020-07-13
Invalid BigInt check for uint32


ishantiw:
### Expected behavior

For `uint32` format we should pass a number type for validation

### Actual behavior

we are passing `BigInt` type when validating `uint32` type

### Steps to reproduce

Validate any schema with `uint32` type

### Which version(s) does this affect? (Environment, OS, etc...)

`5.0`

PR #5561, opened 2020-07-14
Get delegates api - Closes #5255 & #5554 & #5556

### LINKED ISSUE #5255: Update GET /api/delegates?limit=xxx&offset=yyy endpoint
### Description

Add the endpoints:

- `GET /api/delegates?limit=xxx&offset=yyy`

- It should return decoded accounts
- Validate if limit and offset is valid number if not return 400 with error

### Motivation

The HTTP API should return a limited list of registered delegates.

### Acceptance Criteria

- Add unit tests
- Check if invalid query returns 400 with error message

### Additional Information

Related issues: #5213

ishantiw:
### What was the problem?

This PR resolves #5255, #5554 & #5556

### How was it solved?

- Exposed `getAllDelegates` from dpos #5556
- Exposed `getAllDelegates` from action in node and Application #5556
- Added handler `api/delegates?limit=xx&offset=xx` endpoint to controller #5255
- Added unit tests #5255
- Fixed bug related `uint32` and `sint32` format in `lisk-validator` #5554

### How was it tested?

Run `npm run test:functional delegates` under `framework-plugins/lisk-framework-http-api-plugin`

### DIFF
--- a/elements/lisk-dpos/src/dpos.ts
+++ b/elements/lisk-dpos/src/dpos.ts
@@ -15,7 +15,7 @@ import { hash } from '@liskhq/lisk-cryptography';
 import { codec } from '@liskhq/lisk-codec';
 import * as Debug from 'debug';
 import { EventEmitter } from 'events';
-import { voteWeightsSchema } from './schemas';
+import { voteWeightsSchema, delegatesUserNamesSchema } from './schemas';
 
 import {
 	CONSENSUS_STATE_DELEGATE_VOTE_WEIGHTS,
@@ -24,6 +24,7 @@ import {
 	DEFAULT_STANDBY_DELEGATE,
 	DEFAULT_STANDBY_THRESHOLD,
 	DEFAULT_VOTE_WEIGHT_CAP_RATE,
+	CHAIN_STATE_DELEGATE_USERNAMES,
 } from './constants';
 import { DelegatesInfo } from './delegates_info';
 import {
@@ -40,6 +41,7 @@ import {
 	ForgersList,
 	StateStore,
 	DecodedVoteWeights,
+	DecodedUsernames,
 } from './types';
 
 interface DposConstructor {
@@ -271,6 +273,21 @@ export class Dpos {
 		});
 
 		return false;
+	}
+
+	public async getAllDelegates(): Promise<DecodedUsernames | undefined> {
+		const usernamesBuffer = await this.chain.dataAccess.getChainState(
+			CHAIN_STATE_DELEGATE_USERNAMES,
+		);
+		if (usernamesBuffer) {
+			const parsedUsernames = codec.decode<DecodedUsernames>(
+				delegatesUserNamesSchema,
+				usernamesBuffer,
+			);
+			return parsedUsernames;
+		}
+
+		return undefined;
 	}
 
 	/**
--- a/elements/lisk-dpos/src/dpos.ts
+++ b/elements/lisk-dpos/src/dpos.ts
@@ -275,10 +275,11 @@ export class Dpos {
 		return false;
 	}
 
-	public async getAllUsernames(): Promise<DecodedUsernames | undefined> {
+	public async getAllUsernames(): Promise<DecodedUsernames | []> {
 		const usernamesBuffer = await this.chain.dataAccess.getChainState(
 			CHAIN_STATE_DELEGATE_USERNAMES,
 		);
+
 		if (usernamesBuffer) {
 			const parsedUsernames = codec.decode<DecodedUsernames>(
 				delegatesUserNamesSchema,
@@ -287,7 +288,7 @@ export class Dpos {
 			return parsedUsernames;
 		}
 
-		return undefined;
+		return [];
 	}
 
 	/**
--- a/elements/lisk-dpos/src/dpos.ts
+++ b/elements/lisk-dpos/src/dpos.ts
@@ -275,7 +275,7 @@ export class Dpos {
 		return false;
 	}
 
-	public async getAllDelegates(): Promise<DecodedUsernames | undefined> {
+	public async getAllUsernames(): Promise<DecodedUsernames | undefined> {
 		const usernamesBuffer = await this.chain.dataAccess.getChainState(
 			CHAIN_STATE_DELEGATE_USERNAMES,
 		);
--- a/elements/lisk-dpos/src/dpos.ts
+++ b/elements/lisk-dpos/src/dpos.ts
@@ -42,6 +42,7 @@ import {
 	StateStore,
 	DecodedVoteWeights,
 	DecodedUsernames,
+	RegisteredDelegate,
 } from './types';
 
 interface DposConstructor {
@@ -275,7 +276,7 @@ export class Dpos {
 		return false;
 	}
 
-	public async getAllUsernames(): Promise<DecodedUsernames | []> {
+	public async getAllUsernames(): Promise<RegisteredDelegate[]> {
 		const usernamesBuffer = await this.chain.dataAccess.getChainState(
 			CHAIN_STATE_DELEGATE_USERNAMES,
 		);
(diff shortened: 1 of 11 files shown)


framework-plugins/lisk-framework-http-api-plugin/src/controllers/delegates.ts
```diff
@@ -0,0 +1,63 @@
...
+			format: 'uint32',
+			description: 'Number of delegates to be returned',
+		},
+		offset: {
+			type: 'string',
+			format: 'uint32',
+			description: 'Offset to get delegates after a specific length in a delegates list',
+		},
+	},
+	default: {
```
shuse2, line 33: I dont think default will be injected here

framework-plugins/lisk-framework-http-api-plugin/src/controllers/delegates.ts
```diff
@@ -0,0 +1,63 @@
...
+	type: 'object',
+	properties: {
+		limit: {
+			type: 'string',
+			format: 'uint32',
+			description: 'Number of delegates to be returned',
+		},
+		offset: {
+			type: 'string',
+			format: 'uint32',
```
shuse2, line 29: if uint32, type: 'integer' is fine here, and you could add min and max

framework-plugins/lisk-framework-http-api-plugin/src/controllers/delegates.ts
```diff
@@ -0,0 +1,63 @@
...
+		return;
+	}
+
+	const { limit = 100, offset = 0 } = req.query;
+
+	try {
+		const encodedDelegates: string[] = await channel.invoke('app:getAllDelegates');
+		const decodedDelegates = encodedDelegates.map(delegate => codec.decodeAccount(delegate));
+
+		res.status(200).send({ data: paginateList(decodedDelegates, +limit, +offset) });
```
shuse2, line 59: meta is missing?
Also try catch is not needed if only calling next. it should be handled in the middleware

framework-plugins/lisk-framework-http-api-plugin/test/functional/delegates.spec.ts
```diff
@@ -0,0 +1,111 @@
...
+			const { response, status } = await callNetwork(
+				axios.get(getURL('/api/delegates?limit=100&offset=100')),
+			);
+			expect(response.data).toHaveLength(3);
+			expect(status).toBe(200);
+		});
+
+		it('should respond with blank array when no delegates are found', async () => {
+			app['_channel'].invoke = jest.fn();
+			when(app['_channel'].invoke)
```
shuse2, line 71: for functional, prefer not to be mocked

elements/lisk-dpos/src/dpos.ts
```diff
@@ -273,6 +275,21 @@ export class Dpos {
...
+		);
+		if (usernamesBuffer) {
+			const parsedUsernames = codec.decode<DecodedUsernames>(
+				delegatesUserNamesSchema,
+				usernamesBuffer,
+			);
+			return parsedUsernames;
+		}
+
+		return undefined;
```
shuse2, line 290: maybe returning empty array here is fine?

framework-plugins/lisk-framework-http-api-plugin/src/controllers/delegates.ts
ishantiw: when `const errors = validator.validate(getDelegatesQuerySchema, req.query);` it will throw error for integers as everything in query is of string type

framework-plugins/lisk-framework-http-api-plugin/src/controllers/delegates.ts
ishantiw: adding meta. For `next` function, async errors will not be handled on its own until express `5.0`

elements/lisk-dpos/src/types.ts
```diff
@@ -140,11 +141,11 @@ export interface DecodedVoteWeights {
...
+}
+
 export interface DecodedUsernames {
-	registeredDelegates: [
-		{
-			username: string;
-			address: Buffer;
-		},
-	];
+	registeredDelegates: [RegisteredDelegate];
```
shuse2, line 150:
```suggestion
	registeredDelegates: RegisteredDelegate[];
```

elements/lisk-validator/src/formats.ts
```diff
@@ -33,9 +33,9 @@ export const int64 = (data: string): boolean => isNumberString(data) && isSInt64
 
 export const uint64 = (data: string): boolean => isNumberString(data) && isUInt64(BigInt(data));
 
-export const uint32 = (data: string): boolean => isNumberString(data) && isUInt32(BigInt(data));
+export const uint32 = (data: string): boolean => isNumberString(data) && isUInt32(Number(data));
```
nazarhussain, line 36: Good catch.

framework-plugins/lisk-framework-http-api-plugin/src/controllers/delegates.ts
```diff
@@ -0,0 +1,61 @@
...
+	const errors = validator.validate(getDelegatesQuerySchema, req.query);
+	// 400 - Malformed query or parameters
+	if (errors.length) {
+		res.status(400).send({
+			errors: [{ message: new LiskValidationError([...errors]).message }],
+		});
+		return;
+	}
+
+	const { limit = 100, offset = 0 } = req.query;
```
nazarhussain, line 49: I never new we can set default values in object destruct. 👍

framework-plugins/lisk-framework-http-api-plugin/src/controllers/delegates.ts
```diff
@@ -0,0 +1,61 @@
...
+	try {
+		const encodedDelegates: string[] = await channel.invoke('app:getAllDelegates');
+		const decodedDelegates = encodedDelegates.map(delegate => codec.decodeAccount(delegate));
+
+		res.status(200).send({
```
nazarhussain, line 54: Always use `res.json` unless have clear need of `res.send`. The difference is `res.send` tries to detect `Content-Type` header value from the parameter. And based on different type of parameter its value would be different. So be rather explicit to mention that your are sending JSON which means `Content-Type: application/json`

Also `res.send` tries to detect and set `Content-Length` header and few more things.

framework-plugins/lisk-framework-http-api-plugin/src/controllers/peers.ts
```diff
@@ -79,9 +74,10 @@ export const getPeers = (channel: BaseChannel) => async (
 			peers = await channel.invoke<ReadonlyArray<PeerInfo>>('app:getConnectedPeers');
 		}
 
-		peers = paginateList(peers, +limit, +offset);
-
-		res.status(200).send(peers);
+		res.status(200).send({
```
nazarhussain, line 77: Same here.

framework-plugins/lisk-framework-http-api-plugin/test/functional/delegates.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+	beforeAll(async () => {
+		app = await createApplication('delegates_http_functional');
+	});
+
+	afterAll(async () => {
+		await closeApplication(app);
+	});
+
+	describe('/api/delegates', () => {
+		it('should respond with all the delegates', async () => {
```
nazarhussain, line 53: A suggestion, I prefer to group tests in terms of response codes, which match the right expectations and structure if you see in the swagger file. 

```yaml
responses:
        200:
          description: Registered delegate accounts.
          schema:
            $ref: '#/definitions/DelegateResponse'
        400:
          description: Malformed query or parameters.
          schema:
            $ref: '#/definitions/ErrorResponse'
        429:
          description: Too many requests, exceeded rate limit.
          schema:
            $ref: '#/definitions/ErrorResponse'
        500:
          description: Unexpected error.
          schema:
            $ref: '#/definitions/ErrorResponse'
```

So if we write expectations also grouped in same way that looks easy to read and track.

Issue #5566, opened 2020-07-15
Remove duplicate lisk-utils functionalities


shuse2:
### Description

Remove duplicate usage of functionalities which are defined in the lisk-utils and replace with lisk-utils.

### Motivation

We have created lisk-utils but not fully removed the duplication

### Acceptance Criteria

- there should be no duplicate code for functions that are defined in lisk-utils

PR #5568, opened 2020-07-15
Remove usage of lodash - Closes #5563

### LINKED ISSUE #5563: Remove lodash from framework
### Expected behavior
Usage of lodash should be removed. If necessary, it should install per function.

### Actual behavior
Framework has lodash as dependency

shuse2:
### What was the problem?

This PR resolves #5563 

### How was it solved?

- Replace cloneDeep with lisk-utils
- Replace maxBy function with simple map implementation

### How was it tested?

- All test should not be affected
- Check removal from package.json


elements/lisk-chain/package.json
```diff
@@ -44,8 +44,7 @@
 		"@liskhq/lisk-validator": "0.5.0-alpha.0",
 		"@types/node": "12.12.11",
 		"debug": "4.1.1",
-		"lodash.clonedeep": "4.5.0",
-		"lodash.isequal": "4.5.0"
+		"lodash.clonedeep": "4.5.0"
```
nazarhussain, line 47: Why not to remove this package?

shuse2: Since clonedeep will remain as dependency anyway, and there is another issue to replace with `lisk-utils` in general with #5566

PR #5573, opened 2020-07-16
Send notification on missedBlock, createdBlock and node start and node stop - Closes #5290

### LINKED ISSUE #5290: Send notification on missedBlock, createdBlock and node start and node stop
### Description

- Create hook for:
  - registered forger missing block
  - registered forger creating block
  - node start
  - node stop
- Call webhook with the below schema:

```js
// Create block
{ event: “forger:block:created”, time: 134, payload: { height: 1, reward: 10, address: '' } }
// Missed block
{ event: “forger:block:missed”, time: 134, payload: { height: “”, address: '' } }
// Node action
{ event: “forger:node:start”, time: 134, payload: { reason: “” } }
{ event: “forger:node:stop”, time: 134, payload: { reason: “”  } }
```

User-Agent should be in format of
```
${plugin-name}/${plugin-version} ${${os.platform()} ${os.release()}; ${os.arch()}${locale}} lisk-framework/${lisk-framework-version}

lisk-framework-forger-plugin/0.1.0 (darwin 19.5.0; x64 en_US.UTF-8) lisk-framework/0.5.0-alpha.0

```

### Motivation

This event is required in order to notify external services about forging related changes.

### Acceptance Criteria

- Add unit tests to check if the hook is fired by each of the registered forgers
- Check if the webhook is called by registering to the framework
- Emit all events to webhook if `events` is empty and filter and emit specific `events` when specified, ex: `events: ['forger:block:missed']`
### Additional Information

Related issues: #5215

pablitovicente:
### What was the problem?

This PR resolves #5290

### How was it solved?

By adding a webhook class to send notifications based on the events configured

### How was it tested?

TBD


framework-plugins/lisk-framework-forger-plugin/src/forger_plugin.ts
```diff
@@ -114,27 +122,57 @@ export class ForgerPlugin extends BasePlugin {
...
+			options.webhook,
+		);
 		this._forgerPluginDB = await getDBInstance(options.dataPath);
 
 		// eslint-disable-next-line @typescript-eslint/no-misused-promises
 		this._channel.once('app:ready', async () => {
 			this._app = initApi(options, this._channel, this.codec);
 
+			// eslint-disable-next-line no-void
+			void this._webhooks.handleEvent({
```
ManuGowda, line 143: maybe you could remove `void`?

framework-plugins/lisk-framework-forger-plugin/src/forger_plugin.ts
```diff
@@ -114,27 +122,57 @@ export class ForgerPlugin extends BasePlugin {
...
 			this._server = this._app.listen(options.port, '0.0.0.0');
 		});
+
+		// @TODO Fix me! due to the way unload works this event is never fired in time.
+		this._channel.once('app:shutdown', () => {
+			// eslint-disable-next-line no-void
+			void this._webhooks.handleEvent({
+				event: 'app:shutdown',
+				time: Date.now(),
+				payload: { reason: 'node shutdown' },
```
ManuGowda, line 173:
```suggestion
				payload: { reason: 'Node shutdown' },
```

framework-plugins/lisk-framework-forger-plugin/src/webhooks.ts
```diff
@@ -0,0 +1,81 @@
...
+
+interface blockMissed {
+	readonly height: number;
+	readonly missedBlocksByAddress: { [key: string]: number };
+}
+
+interface nodeStarted {
+	readonly reason: string;
+}
+interface nodeStopped {
```
ManuGowda, line 39: type NodeStopped = NodeStarted?
and all interface to be uppercase?

framework-plugins/lisk-framework-forger-plugin/src/forger_plugin.ts
pablitovicente: It's not possible as we discussed with @shuse2 not to wait of the resolution to avoid blocking the rest of the plugin working. In the future if this hooks are moved to its own plugin probably we should in that context wait for it.

framework-plugins/lisk-framework-forger-plugin/src/forger_plugin.ts
```diff
@@ -66,6 +68,10 @@ interface NodeInfo {
 	};
 }
 
+interface missedBlocksByAddress {
```
ManuGowda, line 71: uppercase bitte

framework-plugins/lisk-framework-forger-plugin/src/forger_plugin.ts
```diff
@@ -114,27 +122,57 @@ export class ForgerPlugin extends BasePlugin {
...
+			this._syncingWithNode = false;
 
 			// Listen to new block and delete block events
 			this._subscribeToChannel();
 
 			// Start http server
 			this._server = this._app.listen(options.port, '0.0.0.0');
 		});
+
+		// @TODO Fix me! due to the way unload works this event is never fired in time.
```
ManuGowda, line 167: is this related to what @shuse2 is working on, if so maybe link his issue number here.. so that TODO can be fixed

framework-plugins/lisk-framework-forger-plugin/src/webhooks.ts
```diff
@@ -0,0 +1,78 @@
...
+interface BlockMissed {
+	readonly height: number;
+	readonly missedBlocksByAddress: { [key: string]: number };
+}
+
+interface NodeStatusChange {
+	readonly reason: string;
+}
+
+interface webHookPayload {
```
ManuGowda, line 40: uppercase

pablitovicente: Fixed. Is there an eslint rule for this maybe?

framework-plugins/lisk-framework-forger-plugin/src/forger_plugin.ts
pablitovicente: Not sure if related. The problem I think it's related to the fact that unload does not have a handling of any potential pending events on any of the subscribers. For this to work (and report the actual node stopping instead of the plugin being unloaded) we should be able to implement some blocking mechanism based on any pending received event I think. But maybe we can discuss about later.

framework-plugins/lisk-framework-forger-plugin/src/webhooks.ts
```diff
@@ -0,0 +1,78 @@
...
+		for (const aRegisteredEvent of this.registeredEvents) {
+			if (aRegisteredEvent.events.includes(data.event)) {
+				requiredEvents.push({ url: aRegisteredEvent.url, data });
+			}
+		}
+
+		try {
+			for (const anEvent of requiredEvents) {
+				// eslint-disable-next-line no-void
+				await this.execute(anEvent.data, anEvent.url);
```
shuse2, line 70: if one execution fails, all the later one will fail, shouldn't it try-catch per execute?

framework-plugins/lisk-framework-forger-plugin/src/webhooks.ts
```diff
@@ -0,0 +1,78 @@
...
+			}
+		} catch (err) {
+			debug('Error during webhook processing');
```
shuse2, line 73: prefer to have 1 debug line

pablitovicente: The problem is that Axios adds special properties to the err object outside of message so if the error is caused by a successful http interaction with the receiving end a single line will not show it. Or maybe you meant to show both variables using one call?

framework-plugins/lisk-framework-forger-plugin/src/webhooks.ts
pablitovicente: Will try to change to Promise.allSettle

framework-plugins/lisk-framework-forger-plugin/src/webhooks.ts
shuse2: if it's one debug, call then showing it is fine

framework-plugins/lisk-framework-forger-plugin/src/webhooks.ts
pablitovicente: Promise.allSettle is not final so changed to handle inside of loop.

framework-plugins/lisk-framework-forger-plugin/src/webhooks.ts
pablitovicente: Done

framework-plugins/lisk-framework-forger-plugin/src/webhooks.ts
```diff
@@ -0,0 +1,75 @@
...
+	private readonly headers: HttpHeaders;
+	private readonly registeredEvents: readonly Webhook[];
+
+	public constructor(defaultHeaders: HttpHeaders, configuredEvents: readonly Webhook[]) {
+		this.headers = defaultHeaders;
+		this.registeredEvents = configuredEvents;
+	}
+
+	public async execute(eventData: WebHookPayload, targetURL: string): Promise<object> {
+		return axios.post(targetURL, eventData, { headers: this.headers });
```
nazarhussain, line 56: We should document it somewhere that our webhook implementation does not support JSONP, so `targetURL` should have Cross-origin network access.

PR #5577, opened 2020-07-16
Create PATCH /api/forging endpoint Closes - #5288

### LINKED ISSUE #5288: Create PATCH /api/forging endpoint
### Description

- Add `PATCH /api/forging` which requests and responds with the below JSON:

```js
// request
{
  address: '',
  password: '',
}
// response
{
  address: '',
  receivedFees: '',
  receivedRewards: '',
  producedBlocks: 20,
  missedBlocks: 1,
  enabled: true,
}
```

### Motivation

The forging helper should be able to enable delegates using the HTTP API endpoint.

### Acceptance Criteria

- Add unit test to check if the return value is as described above
- Add unit test to check delegate is enabled
- Check if the delegate is enabled/disabled as expected by registering to the framework

### Additional Information

Related issues: #5215

ishantiw:
### What was the problem?

This PR resolves #5288

### How was it solved?

- Add `PATCH /api/forging` endpoint controller
- Add it to API router
- Add functional test related to it
- Add a swagger definition for `/forging`
- Remove `/hello` controller and its tests

### How was it tested?

`npm run test:functional forging`


framework-plugins/lisk-framework-forger-plugin/src/controllers/forging.ts
```diff
@@ -0,0 +1,97 @@
...
+): Promise<void> => {
+	const errors = validator.validate(updateForgingParams, req.body);
+	// 400 - Malformed query or parameters
+	if (errors.length) {
+		res.status(400).send({
+			errors: [{ message: new LiskValidationError([...errors]).message }],
+		});
+		return;
+	}
+	const { address, password, forging } = req.body as {
```
mitsuaki-u, line 51: Optional: We can have an interface for this type

PR #5579, opened 2020-07-16
Create GET /api/forging/info endpoint - Closes #5287

### LINKED ISSUE #5287: Create GET /api/forging/info endpoint
### Description

- Add `/api/forging/info` which returns the below JSON:

```js
[{
  address: '',
  username: '',
  totalVotesReceived: '',
  totalReceivedFees: '',
  totalReceivedRewards: '',
  totalProducedBlocks: 20,
  consecutiveMissedBlocks: 1,
  forging: true,
}]
```
- Use data collected from #5285 
- Return all the result for registered forgers, which can be obtained from the framework config
- Check if framework action provides an endpoint to get the config, if not add them

### Motivation

The forging helper should return the forging info for registered forgers.

### Acceptance Criteria

- Add unit test to check if the return value is as described above
- Check if the response is return as expected by registering to the framework

### Additional Information

Related issues: #5215

ManuGowda:
### What was the problem?

This PR resolves #5287

### How was it solved?

- Introduce `api/forging/info`

### How was it tested?

- Added functional test


framework-plugins/lisk-framework-forger-plugin/src/controllers/forging_info.ts
```diff
@@ -0,0 +1,61 @@
...
+export const getForgingInfo = (channel: BaseChannel, codec: PluginCodec, db: KVStore) => async (
+	_req: Request,
+	res: Response,
+	next: NextFunction,
+): Promise<void> => {
+	try {
+		const forgingDelegates = await channel.invoke<ReadonlyArray<Forger>>(
+			'app:getForgingStatusOfAllDelegates',
+		);
+		const forgerAccounts = await channel.invoke<Buffer[]>('app:getAccounts', {
```
shuse2, line 30: i dont think this action returns `Buffer[]`

framework-plugins/lisk-framework-forger-plugin/test/functional/forging_info.spec.ts
```diff
@@ -0,0 +1,69 @@
...
+				'address',
+				'forging',
+				'username',
+				'totalReceivedFees',
+				'totalReceivedRewards',
+				'totalProducedBlocks',
+				'totalVotesReceived',
+				'consecutiveMissedBlocks',
+			]);
+			expect(resultData.data).toMatchSnapshot();
```
shuse2, line 61: if we have above key comparison, do we need `toMatchSnapshot`?

ManuGowda: true.. ill remove keys maybe... snapshot is better?

PR #5581, opened 2020-07-20
Wrong calculation for votesReceived forging info - Closes #5574

### LINKED ISSUE #5574: Wrong calculation for votesReceived forging info
### Expected behavior

The functions `_incrementForgerInfo` and `_decrementForgerInfo` should calculate `votesReceived` for corresponding delegate who received the votes. 

### Actual behavior

Currently both of the functions are adding/updating vote transaction details to the delegate who is forging the block, which logically does not relate to the votes. 

### Steps to reproduce
- Send vote to delegate `A`
- A delegate `B` forged a block and included that transaction
- If you check forging info you see delegate `B` will have that in `votesReceived` 
- Where as it belongs to delegate `A`

5.0.*

ManuGowda:
### What was the problem?

This PR resolves #5574 

### How was it solved?

- If delegate a receives votes from b and c then forgers info should save voters(b,c) address and amount, also separated add and revert voter logic
- We have consecutiveMissedBlocks details available for a given forging account removed associated missed block test and updated schema

### How was it tested?

- Unskipped voters test
-  Updated other functional tests

### DIFF
--- a/framework-plugins/lisk-framework-forger-plugin/src/forger_plugin.ts
+++ b/framework-plugins/lisk-framework-forger-plugin/src/forger_plugin.ts
@@ -27,7 +27,7 @@ import {
 	BlockHeaderJSON,
 } from 'lisk-framework';
 import { VoteTransaction } from '@liskhq/lisk-transactions';
-import { objects, dataStructures as ds } from '@liskhq/lisk-utils';
+import { objects, dataStructures } from '@liskhq/lisk-utils';
 import type { Express } from 'express';
 import { initApi } from './api';
 import {
@@ -88,7 +88,7 @@ export class ForgerPlugin extends BasePlugin {
 	private _server!: Server;
 	private _app!: Express;
 	private _channel!: BaseChannel;
-	private _forgersList!: ds.BufferMap<boolean>;
+	private _forgersList!: dataStructures.BufferMap<boolean>;
 	private _transactionFees!: TransactionFees;
 	private _webhooks!: Webhooks;
 	private _syncingWithNode!: boolean;
@@ -201,7 +201,7 @@ export class ForgerPlugin extends BasePlugin {
 	}
 
 	private async _setForgersList(): Promise<void> {
-		this._forgersList = new ds.BufferMap<boolean>();
+		this._forgersList = new dataStructures.BufferMap<boolean>();
 		const forgersList = await this._channel.invoke<Forger[]>('app:getForgingStatusOfAllDelegates');
 		for (const { address, forging } of forgersList) {
 			this._forgersList.set(Buffer.from(address, 'base64'), forging);
@@ -398,6 +398,10 @@ export class ForgerPlugin extends BasePlugin {
 				forgerInfo.votesReceived.push(votesReceived);
 			} else {
 				forgerInfo.votesReceived[voterIndex].amount += votesReceived.amount;
+				// Remove voter when amount becomes zero
+				if (forgerInfo.votesReceived[voterIndex].amount === BigInt(0)) {
+					forgerInfo.votesReceived.splice(voterIndex, 1);
+				}
 			}
 			await setForgerInfo(this._forgerPluginDB, getBinaryAddress(delegateAddress), forgerInfo);
 		}
@@ -415,11 +419,14 @@ export class ForgerPlugin extends BasePlugin {
 				aVote.address.equals(votesReceived.address),
 			);
 
-			forgerInfo.votesReceived[voterIndex].amount -= BigInt(votesReceived.amount);
-			if (forgerInfo.votesReceived[voterIndex].amount === BigInt(0)) {
-				forgerInfo.votesReceived.splice(voterIndex, 1);
+			if (voterIndex !== -1) {
+				forgerInfo.votesReceived[voterIndex].amount -= BigInt(votesReceived.amount);
+				// Remove voter when amount becomes zero
+				if (forgerInfo.votesReceived[voterIndex].amount === BigInt(0)) {
+					forgerInfo.votesReceived.splice(voterIndex, 1);
+				}
+				await setForgerInfo(this._forgerPluginDB, getBinaryAddress(delegateAddress), forgerInfo);
 			}
-			await setForgerInfo(this._forgerPluginDB, getBinaryAddress(delegateAddress), forgerInfo);
 		}
 	}
 
--- a/framework-plugins/lisk-framework-forger-plugin/src/forger_plugin.ts
+++ b/framework-plugins/lisk-framework-forger-plugin/src/forger_plugin.ts
@@ -38,7 +38,7 @@ import {
 	setForgerSyncInfo,
 } from './db';
 import * as config from './defaults';
-import { Forger, Options, TransactionFees } from './types';
+import { Forger, Options, TransactionFees, Voters } from './types';
 import { Webhooks } from './webhooks';
 
 const BLOCKS_BATCH_TO_SYNC = 1000;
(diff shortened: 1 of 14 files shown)


framework-plugins/lisk-framework-forger-plugin/src/forger_plugin.ts
```diff
@@ -288,98 +289,116 @@ export class ForgerPlugin extends BasePlugin {
...
 
 		if (this._forgersList.find(forger => forger.address === forgerAddress)) {
 			forgerInfo.totalProducedBlocks += 1;
 			forgerInfo.totalReceivedRewards += BigInt(reward);
 			forgerInfo.totalReceivedFees += this._getFee(payload, encodedBlock);
-			isUpdated = true;
-
-			// eslint-disable-next-line @typescript-eslint/no-floating-promises
-			this._webhooks.handleEvent({
+			await this._webhooks.handleEvent({
```
shuse2, line 305: this should not be awaited, it was intended to be floating-promises

framework-plugins/lisk-framework-forger-plugin/src/forger_plugin.ts
```diff
@@ -288,98 +289,116 @@ export class ForgerPlugin extends BasePlugin {
...
-						aVote.address.equals(Buffer.from(vote.delegateAddress, 'base64')),
+					const registeredDelegateIndex = this._forgersList.findIndex(
+						forger => forger.address === vote.delegateAddress,
 					);
-					if (delegateVoteIndex !== -1) {
-						forgerInfo.votesReceived[delegateVoteIndex].amount -= BigInt(vote.amount);
-						if (forgerInfo.votesReceived[delegateVoteIndex].amount === BigInt(0)) {
-							forgerInfo.votesReceived.splice(delegateVoteIndex, 1);
+					if (registeredDelegateIndex !== -1) {
+						const affectedForgerInfo = await getForgerInfo(
```
shuse2, line 346: This should be done outside of loop?
Also, if there are multiple transaction from the same sender in the same block
- if first transaction votes, and second transaction unvotes, it will have invalid result
- multiple get runs which is bit inefficient

framework-plugins/lisk-framework-forger-plugin/src/forger_plugin.ts
```diff
@@ -288,97 +293,113 @@ export class ForgerPlugin extends BasePlugin {
...
-					if (delegateVoteIndex !== -1) {
-						forgerInfo.votesReceived[delegateVoteIndex].amount -= BigInt(vote.amount);
-						if (forgerInfo.votesReceived[delegateVoteIndex].amount === BigInt(0)) {
-							forgerInfo.votesReceived.splice(delegateVoteIndex, 1);
-						}
-						isUpdated = true;
+					if (registeredDelegateIndex !== -1) {
+						acc[curr.delegateAddress] = {
+							address: senderAddress,
+							amount: BigInt(acc[curr.delegateAddress] || 0) + BigInt(curr.amount),
```
nazarhussain, line 354:
```suggestion
							amount: BigInt(acc[curr.delegateAddress].amount || 0) + BigInt(curr.amount),
```

framework-plugins/lisk-framework-forger-plugin/src/forger_plugin.ts
```diff
@@ -288,97 +293,113 @@ export class ForgerPlugin extends BasePlugin {
...
 		for (const trx of payload) {
 			if (trx.type === VoteTransaction.TYPE) {
-				for (const vote of (trx.asset as Asset).votes) {
-					const delegateVoteIndex = forgerInfo.votesReceived.findIndex(aVote =>
-						aVote.address.equals(Buffer.from(vote.delegateAddress, 'base64')),
+				const senderAddress = getAddressFromPublicKey(Buffer.from(trx.senderPublicKey, 'base64'));
+				(trx.asset as Asset).votes.reduce((acc: ForgerReceivedVotes, curr) => {
+					const registeredDelegateIndex = this._forgersList.findIndex(
+						forger => forger.address === curr.delegateAddress,
 					);
```
nazarhussain, line 350: The `_forgersList` is loaded once on start and does not change. If we can convert it to `hash map` with address as key, we can avoid few of such loops in different places and code will look more readable.

framework-plugins/lisk-framework-forger-plugin/src/forger_plugin.ts
```diff
@@ -288,97 +293,113 @@ export class ForgerPlugin extends BasePlugin {
...
-					if (delegateVoteIndex !== -1) {
-						forgerInfo.votesReceived[delegateVoteIndex].amount -= BigInt(vote.amount);
-						if (forgerInfo.votesReceived[delegateVoteIndex].amount === BigInt(0)) {
-							forgerInfo.votesReceived.splice(delegateVoteIndex, 1);
-						}
-						isUpdated = true;
+					if (registeredDelegateIndex !== -1) {
+						acc[curr.delegateAddress] = {
```
nazarhussain, line 352: If a delegate received votes from multiple different accounts, does this line will not override earlier values?

framework-plugins/lisk-framework-forger-plugin/src/forger_plugin.ts
```diff
@@ -72,15 +73,22 @@ interface MissedBlocksByAddress {
...
+	Buffer.from(base64AddressStr, 'base64').toString('binary');
+const getAddressBuffer = (base64AddressStr: string) => Buffer.from(base64AddressStr, 'base64');
 
 export class ForgerPlugin extends BasePlugin {
 	private _forgerPluginDB!: KVStore;
 	private _server!: Server;
 	private _app!: Express;
 	private _channel!: BaseChannel;
-	private _forgersList!: ReadonlyArray<Forger>;
+	private _forgersList!: ds.BufferMap<boolean>;
```
shuse2, line 91:
```suggestion
	private _forgersList!: dataStructures.BufferMap<boolean>;
```
prefer not to use abbreviation.

framework-plugins/lisk-framework-forger-plugin/src/forger_plugin.ts
```diff
@@ -244,141 +282,144 @@ export class ForgerPlugin extends BasePlugin {
...
+				getBinaryAddress(delegateAddress),
+			);
+
+			const voterIndex = forgerInfo.votesReceived.findIndex(aVote =>
+				aVote.address.equals(votesReceived.address),
+			);
+			if (voterIndex === -1) {
+				forgerInfo.votesReceived.push(votesReceived);
+			} else {
+				forgerInfo.votesReceived[voterIndex].amount += votesReceived.amount;
```
shuse2, line 400: when vote becomes 0, it should be removed?

framework-plugins/lisk-framework-forger-plugin/test/functional/forger_info/event_track.spec.ts
```diff
@@ -72,36 +71,157 @@ describe('Forger Info', () => {
...
-		it('should save forger info with votes received in new block', async () => {
-			// Arrange
-			const forgerPluginInstance = app['_controller'].plugins[ForgerPlugin.alias];
-			const forgingDelegateAddress = forgerPluginInstance['_forgersList'][0].address;
-			const transaction = createVoteTransaction({
-				amount: '10',
-				recipientAddress: forgingDelegateAddress,
-				fee: '0.3',
-				nonce: accountNonce,
+		describe('Vote transactions', () => {
```
shuse2, line 80: maybe add a case to check when the amount becomes `0`?

PR #5600, opened 2020-07-31
HTTP Plugin error handling incorrect for POST /api/transactions - Closes #5594

### LINKED ISSUE #5594: HTTP Plugin error handling incorrect for POST /api/transactions
### Expected behavior

When a request is sent to the HTTP API it should respond with errors when payload is invalid

### Actual behavior

- When a request is sent to the HTTP with an invalid schema the endpoint never responds
- There's an unhandled rejections shown in the logs.

```
{
 "id": "x5njcWAdj35KIEztqD5fFTBEkA8gJxvCwqgljJj63dE=",
 "generatorAddress": "T9jMTiejSJtX7Zhu/j0yfT3kDZI=",
 "seedReveal": "cE/686ZD8tiDA+Xg2Jo0Tg==",
 "height": 84,
 "round": 1,
 "slot": 355379,
 "reward": "0"
}
(node:81230) UnhandledPromiseRejectionWarning: Error: Invalid schema property found. Path: _id
    at Object.exports.recursiveTypeCast (/Users/pablo/Documents/LISK/Repos/lisk-sdk/elements/lisk-codec/dist-node/json_wrapper.js:91:23)
    at Codec.fromJSON (/Users/pablo/Documents/LISK/Repos/lisk-sdk/elements/lisk-codec/dist-node/codec.js:71:24)
    at encodeTransactionFromJSON (/Users/pablo/Documents/LISK/Repos/lisk-sdk/framework/dist-node/plugins/base_plugin.js:25:88)
    at Object.encodeTransaction (/Users/pablo/Documents/LISK/Repos/lisk-sdk/framework/dist-node/plugins/base_plugin.js:72:49)
    at /Users/pablo/Documents/LISK/Repos/lisk-sdk/framework-plugins/lisk-framework-http-api-plugin/dist-node/controllers/transactions.js:71:38
    at Layer.handle [as handle_request] (/Users/pablo/Documents/LISK/Repos/lisk-sdk/node_modules/express/lib/router/layer.js:95:5)
    at next (/Users/pablo/Documents/LISK/Repos/lisk-sdk/node_modules/express/lib/router/route.js:137:13)
    at Route.dispatch (/Users/pablo/Documents/LISK/Repos/lisk-sdk/node_modules/express/lib/router/route.js:112:3)
    at Layer.handle [as handle_request] (/Users/pablo/Documents/LISK/Repos/lisk-sdk/node_modules/express/lib/router/layer.js:95:5)
    at /Users/pablo/Documents/LISK/Repos/lisk-sdk/node_modules/express/lib/router/index.js:281:22
(node:81230) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
(node:81230) [DEP0018] DeprecationWarning: Unhandled promise rejections are deprecated. In the future, promise rejections that are not handled will terminate the Node.js process with a non-zero exit code.
16:10:06 INFO lisk-framework: New block added to the chain (module=lisk:app)
```

### Steps to reproduce

Send a transaction using HTTP Plugin (POST http://localhost:4000/api/transactions ) with an invalid schema and observe the behaviour of the client being used and the logs from sdk

5.0

pablitovicente:
### What was the problem?

This PR resolves #5594

### How was it solved?

By adding try/catch block in the controller for POST transactions/ to avoid unhandled rejections

### How was it tested?

- Updated Functional Test (before the fix to the code `test/functional/transactions.spec.ts` was crashing now it asserts correctly )

### Additional Notes

This PR also removes the `hello` sample route and supporting files as it was only a placeholder.
This PR also fixes usage of hex string instead of base64 for validateSignature method


framework-plugins/lisk-framework-http-api-plugin/src/controllers/transactions.ts
```diff
@@ -109,24 +109,21 @@ export const postTransaction = (channel: BaseChannel, codec: PluginCodec) => asy
...
+			transaction: encodedTransaction,
+		});
 
-		res.status(409).json({ errors: processingErrors });
-		return;
+		res.status(200).json({ data: result, meta: {} });
+	} catch (err) {
+		res
+			.status(409)
+			.json({ error: (err as Error).message.replace('Lisk validator found 1 error[s]:\n', '') });
```
shuse2, line 127: I think format should be `{ errors: [] }`, and why is it replacing?

framework-plugins/lisk-framework-http-api-plugin/test/functional/transactions.spec.ts
```diff
@@ -239,11 +239,7 @@ describe('Hello endpoint', () => {
 				// Assert
 				expect(status).toEqual(409);
 				expect(response).toEqual({
-					errors: [
-						{
-							message: `Failed to validate signature ${newSignature.toString('hex')}`,
-						},
-					],
+					error: `Failed to validate signature ${newSignature.toString('hex')}`,
```
shuse2, line 242: not related to this PR, but why `${newSignature.toString('hex')}` is `hex`?

framework-plugins/lisk-framework-http-api-plugin/src/controllers/transactions.ts
pablitovicente: As the method being invoked returns errors using lisk-validator errors it seems this method returns only an error and with the lisk validator message before the actual error. Maybe we should create a different error handler logic for the action?

framework-plugins/lisk-framework-http-api-plugin/test/functional/transactions.spec.ts
pablitovicente: Not sure; why the original author used hex?

pablitovicente: As discovered hexToBuffer is being use will fix and send in this PR or separated PR based on number of test fixes required

framework-plugins/lisk-framework-http-api-plugin/src/controllers/transactions.ts
pablitovicente: I updated the format. Kept the replace as in some cases as the action `handleEventPostTransaction` is using `LiskValidationError` maybe we should consider add an extra error handler different to this that just process and returns the errors with no additional text?

framework-plugins/lisk-framework-http-api-plugin/test/functional/transactions.spec.ts
pablitovicente: Fixed and updated some unit tests and functional http.

framework-plugins/lisk-framework-http-api-plugin/src/controllers/transactions.ts
```diff
@@ -109,24 +109,23 @@ export const postTransaction = (channel: BaseChannel, codec: PluginCodec) => asy
...
+			transaction: encodedTransaction,
+		});
 
-		res.status(409).json({ errors: processingErrors });
-		return;
+		res.status(200).json({ data: result, meta: {} });
+	} catch (err) {
+		res.status(409).json({
+			errors: [
+				{ message: (err as Error).message.replace('Lisk validator found 1 error[s]:\n', '') },
```
shuse2, line 127: I think we don't need to add extra logic to filter the error message here.
it can be `[ err ] ` IMO. If the output is wrong, we should fix in the framework side

pablitovicente: To fix in framework side we need to write a new handler there I think

pablitovicente: As discussed the lisk-validator handling will be checked as a separate issue.

PR #5620, opened 2020-08-05
Add base interfaces - Closes #5604

### LINKED ISSUE #5604: Create BaseModule and BaseAsset with module skeletons
### Description
*   Add BaseModule in lisk-framework with the defined hooks in the epic
*   Add BaseAsset in lisk-framework with the defined hooks in the epic
*   Add skeleton modules of token, dpos, sequence and keys to be implemented later
*   Define errors which can be used in the modules
    *   Error should have code
    *   code=” nonce_not_ready” to handle error in transaction

Move
*   bftThreshold
to genesis config.

Add dynamically
*   baseFees: { [type: string]: bigint } to the genesis config

Genesis Config should be allowed to add additional properties, and module specific properties should not be added to the schema.

### Motivation

Base module and base asset are required to be extended for all the modules.
Also, module skeleton will be helpful to parallelize the work for each module.

### Acceptance criteria

*   BaseModule and BaseAsset should be defined as abstract class
*   Related types, Validators and ModuleAccess should be defined.

### Related Issues
#5603

shuse2:
### What was the problem?

This PR resolves #5604 

### How was it solved?

- Add BaseAsset
- Add BaseModule
- Update genesisConig type
- Add skeleton modules for src and test

### How was it tested?

N/A

### DIFF
--- a/framework/src/application/schema/application_config_schema.ts
+++ b/framework/src/application/schema/application_config_schema.ts
@@ -98,6 +98,11 @@ export const applicationConfigSchema = {
 					description:
 						'The unique name of the relevant community as a string encoded in UTF-8 format',
 				},
+				bftThreshold: {
+					type: 'number',
+					minimum: 1,
+					description: 'Number of validator required to set finality',
+				},
 				// NOTICE: blockTime and maxPayloadLength are related and it's values
 				// need to be changed together as per recommendations noted in https://github.com/LiskHQ/lisk-sdk/issues/3151
 				// TODO this recommendations need to be updated now that we changed to a byte size block
@@ -137,7 +142,7 @@ export const applicationConfigSchema = {
 					additionalProperties: false,
 				},
 			},
-			additionalProperties: false,
+			additionalProperties: true,
 		},
 		forging: {
 			type: 'object',
--- a/framework/src/application/schema/application_config_schema.ts
+++ b/framework/src/application/schema/application_config_schema.ts
@@ -101,7 +101,7 @@ export const applicationConfigSchema = {
 				bftThreshold: {
 					type: 'number',
 					minimum: 1,
-					description: 'Number of validator required to set block finality',
+					description: 'Number of validators required to set block finality',
 				},
 				// NOTICE: blockTime and maxPayloadLength are related and it's values
 				// need to be changed together as per recommendations noted in https://github.com/LiskHQ/lisk-sdk/issues/3151
--- a/framework/src/application/schema/application_config_schema.ts
+++ b/framework/src/application/schema/application_config_schema.ts
@@ -101,7 +101,7 @@ export const applicationConfigSchema = {
 				bftThreshold: {
 					type: 'number',
 					minimum: 1,
-					description: 'Number of validator required to set finality',
+					description: 'Number of validator required to set block finality',
 				},
 				// NOTICE: blockTime and maxPayloadLength are related and it's values
 				// need to be changed together as per recommendations noted in https://github.com/LiskHQ/lisk-sdk/issues/3151
--- a/framework/src/errors.ts
+++ b/framework/src/errors.ts
@@ -24,8 +24,8 @@ export class FrameworkError extends Error {
 	}
 }
 
-export class NonceNotReadyError extends FrameworkError {
-	public code = 'ERR_NONCE_NOT_READY';
+export class NonceOutOfBoundsError extends FrameworkError {
+	public code = 'ERR_NONCE_OUT_OF_BOUNDS';
 	public actual: string;
 	public expected: string;
 	public constructor(message: string, actual: bigint, expected: bigint) {
--- a/framework/src/errors.ts
+++ b/framework/src/errors.ts
@@ -15,7 +15,7 @@
 
 export class FrameworkError extends Error {
 	public name: string;
-	public code = 'framework';
+	public code = 'ERR_FRAMEWORK';
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any
 	public constructor(...args: any[]) {
 		super(...args);
@@ -25,7 +25,7 @@ export class FrameworkError extends Error {
 }
 
 export class NonceNotReadyError extends FrameworkError {
-	public code = 'nonce_not_ready';
+	public code = 'ERR_NONCE_NOT_READY';
 	public actual: string;
 	public expected: string;
 	public constructor(message: string, actual: bigint, expected: bigint) {
@@ -37,7 +37,7 @@ export class NonceNotReadyError extends FrameworkError {
 
 export class SchemaValidationError extends FrameworkError {
 	public errors: Error[];
-	public code = 'schema_validation';
+	public code = 'ERR_SCHEMA_VALIDATION';
 	public constructor(errors: Error[]) {
 		super(JSON.stringify(errors, null, 2));
 		this.errors = errors;
@@ -47,7 +47,7 @@ export class SchemaValidationError extends FrameworkError {
 export class DuplicateAppInstanceError extends FrameworkError {
 	public appLabel: string;
 	public pidPath: string;
-	public code = 'duplicate_app_instance';
+	public code = 'ERR_DUPLICATE_APP_INSTANCE';
 	public constructor(appLabel: string, pidPath: string) {
 		super(`Duplicate app instance for "${appLabel}"`);
 		this.appLabel = appLabel;
@@ -56,7 +56,7 @@ export class DuplicateAppInstanceError extends FrameworkError {
 }
 
 export class ImplementationMissingError extends FrameworkError {
-	public code = 'implementation_missing';
+	public code = 'ERR_IMPLEMENTATION_MISSING';
 	public constructor() {
 		super('Implementation missing error');
 	}
--- a/framework/src/errors.ts
+++ b/framework/src/errors.ts
@@ -15,6 +15,7 @@
 
 export class FrameworkError extends Error {
 	public name: string;
+	public code = 'framework';
 	// eslint-disable-next-line @typescript-eslint/no-explicit-any
 	public constructor(...args: any[]) {
 		super(...args);
@@ -23,8 +24,20 @@ export class FrameworkError extends Error {
 	}
 }
 
+export class NonceNotReadyError extends FrameworkError {
+	public code = 'nonce_not_ready';
+	public actual: string;
+	public expected: string;
+	public constructor(message: string, actual: bigint, expected: bigint) {
+		super(message);
+		this.actual = actual.toString();
+		this.expected = expected.toString();
+	}
+}
+
 export class SchemaValidationError extends FrameworkError {
 	public errors: Error[];
+	public code = 'schema_validation';
 	public constructor(errors: Error[]) {
 		super(JSON.stringify(errors, null, 2));
 		this.errors = errors;
@@ -34,6 +47,7 @@ export class SchemaValidationError extends FrameworkError {
 export class DuplicateAppInstanceError extends FrameworkError {
 	public appLabel: string;
 	public pidPath: string;
+	public code = 'duplicate_app_instance';
 	public constructor(appLabel: string, pidPath: string) {
 		super(`Duplicate app instance for "${appLabel}"`);
 		this.appLabel = appLabel;
@@ -42,6 +56,7 @@ export class DuplicateAppInstanceError extends FrameworkError {
 }
 
 export class ImplementationMissingError extends FrameworkError {
+	public code = 'implementation_missing';
 	public constructor() {
 		super('Implementation missing error');
 	}
--- a/framework/src/index.ts
+++ b/framework/src/index.ts
@@ -12,10 +12,7 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
-import { BaseAsset } from '@liskhq/lisk-chain';
-
-export { BaseAsset };
-export { BaseModule } from './modules';
+export { BaseModule, BaseAsset } from './modules';
 export { Application } from './application';
 export { version } from './version';
 export { systemDirs } from './application/system_dirs';
--- a/framework/src/index.ts
+++ b/framework/src/index.ts
@@ -12,6 +12,10 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
+import { BaseAsset } from '@liskhq/lisk-chain';
+
+export { BaseAsset };
+export { BaseModule } from './modules';
 export { Application } from './application';
 export { version } from './version';
 export { systemDirs } from './application/system_dirs';
--- a/framework/src/modules/base_asset.ts
+++ b/framework/src/modules/base_asset.ts
@@ -26,6 +26,13 @@ export interface ReducerHandler {
 	invoke: (params: Record<string, unknown>) => Promise<unknown>;
 }
 
+export interface ApplyAssetInput<T> {
+	senderID: Buffer;
+	asset: T;
+	stateStore: StateStore;
+	reducerHandler: ReducerHandler;
+}
+
 export abstract class BaseAsset<T = unknown> {
 	public baseFee = BigInt(0);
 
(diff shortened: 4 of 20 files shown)


framework/src/errors.ts
```diff
@@ -23,8 +24,20 @@ export class FrameworkError extends Error {
 	}
 }
 
+export class NonceNotReadyError extends FrameworkError {
```
pablitovicente, line 27: Just to understand what's a "Nonce Not Ready" ?

shuse2: it suppose to be `nonce_too_high`, will update

shuse2: Ah but it should be bit more generic than too high, so that's why i put `nonce_not_ready` to be processed/included in a block

framework/src/application/schema/application_config_schema.ts
```diff
@@ -98,6 +98,11 @@ export const applicationConfigSchema = {
 					description:
 						'The unique name of the relevant community as a string encoded in UTF-8 format',
 				},
+				bftThreshold: {
+					type: 'number',
+					minimum: 1,
+					description: 'Number of validator required to set finality',
```
ManuGowda, line 104:
```suggestion
					description: 'Number of validator required to set block finality',
```

framework/src/errors.ts
```diff
@@ -15,6 +15,7 @@
 
 export class FrameworkError extends Error {
 	public name: string;
+	public code = 'framework';
```
ManuGowda, line 18: should we use error code like
framework: 1000-2000(range)
so that user could handle it better? would be nice to standardise code

framework/src/modules/base_module.ts
```diff
@@ -0,0 +1,82 @@
...
+ * Removal or modification of this copyright notice is prohibited.
+ */
+/* eslint-disable class-methods-use-this */
+
+import { Block } from '@liskhq/lisk-chain';
+import { GenesisBlock } from '@liskhq/lisk-genesis';
+import { GenesisConfig, Consensus, AccountSchema } from '../types';
+import { BaseAsset, StateStore, ReducerHandler } from './base_asset';
+
+interface Reducer {
```
pablitovicente, line 21: Should this be named to match the class property maybe? Like `Reducers` as it would seem it works as a dictionary of reducers. Also should this be Array instead of object to ensure order?

framework/src/modules/base_module.ts
```diff
@@ -0,0 +1,82 @@
...
+	[key: string]: (params: Record<string, unknown>, statestore: StateStore) => Promise<unknown>;
+}
+
+interface Action {
+	[key: string]: (params: Record<string, unknown>) => Promise<unknown>;
+}
+
+// TODO: Replace after "Update lisk-chain to support the on-chain architecture"
```
ManuGowda, line 29: we need a todo issue to fix major todos's 😄 
meanwhile could you add issue number here?

framework/src/modules/index.ts
```diff
@@ -0,0 +1,19 @@
+/*
+ * Copyright © 2019 Lisk Foundation
```
ManuGowda, line 2:
```suggestion
 * Copyright © 2020 Lisk Foundation
```

framework/src/modules/dpos/index.ts
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright © 2019 Lisk Foundation
```
ManuGowda, line 2:
```suggestion
 * Copyright © 2020 Lisk Foundation
```

framework/src/modules/keys/index.ts
```diff
@@ -0,0 +1,15 @@
+/*
+ * Copyright © 2019 Lisk Foundation
```
ManuGowda, line 2: maybe its there in most places,ill stop commenting now.

framework/src/modules/token/transfer_asset.ts
```diff
@@ -0,0 +1,59 @@
...
+	readonly amount: bigint;
+	readonly recipientAddress: Buffer;
+	readonly data: string;
+}
+
+export class TransferAsset extends BaseAsset {
+	public name = 'transfer';
+	public type = 0;
+	public assetSchema = {
+		$id: 'lisk/transfer-transaction',
```
ManuGowda, line 28:
```suggestion
		$id: 'lisk/transfer-asset',
```

framework/src/errors.ts
pablitovicente: maybe `NonceNotWithinLimits` (either lower or too high)? But ok to keep I guess I only have problem with the Ready word as it would seem nonce is an active component an it's only a prop. But as I said ok to keep.

framework/src/errors.ts
pablitovicente: I thought of this as well but then we need to have a translation from number code to meaningful

shuse2: yea since we will need translation, i just made it to string so that without any reference, at least it has some meaning

nazarhussain, line 18: I will suggest to use standard error pattern as `NodeJS` itself. Please see following list of error codes which are available in NodeJS and we can use similar pattern in our case. 

e.g. For this one particular it will become `ERR_FRAMEWORK`

https://nodejs.org/dist/latest/docs/api/errors.html#errors_node_js_error_codes

framework/src/errors.ts
```diff
@@ -15,6 +15,7 @@
 
 export class FrameworkError extends Error {
 	public name: string;
```
nazarhussain, line 17: We didn't set this value in the constructor, super will set it to `Error`. Need to explicitly set `this.name = 'FrameworkError';` in the constructor.

framework/src/errors.ts
```diff
@@ -23,8 +24,20 @@ export class FrameworkError extends Error {
...
 }
 
+export class NonceNotReadyError extends FrameworkError {
+	public code = 'nonce_not_ready';
+	public actual: string;
+	public expected: string;
+	public constructor(message: string, actual: bigint, expected: bigint) {
+		super(message);
+		this.actual = actual.toString();
+		this.expected = expected.toString();
```
nazarhussain, line 34: Here again, you have to set the name of error explicitly for every custom error class. 

```bash
❯ node
Welcome to Node.js v12.18.3.
Type ".help" for more information.
> class MyError extends Error {}
undefined
> const e = new MyError("message")
undefined
> e.name
'Error'
```

framework/src/index.ts
```diff
@@ -12,6 +12,10 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
+import { BaseAsset } from '@liskhq/lisk-chain';
+
+export { BaseAsset };
```
nazarhussain, line 17:
```suggestion
export { BaseAsset } from '@liskhq/lisk-chain';
```

framework/src/modules/base_asset.ts
```diff
@@ -0,0 +1,46 @@
...
+ */
+/* eslint-disable class-methods-use-this */
+
+import { Schema } from '@liskhq/lisk-codec';
+import { StateStore as ChainStateStore, BlockHeader } from '@liskhq/lisk-chain';
+
+// Limit the scope of state store to which module can access
+export type StateStore = Omit<
+	ChainStateStore,
+	'consensus' | 'finalize' | 'createSnapshot' | 'restoreSnapshot'
```
nazarhussain, line 22: Why `consensus` is omitted here? Its also a state store, what if any module want to access the consensus information during transaction asset apply?

framework/src/modules/base_asset.ts
```diff
@@ -0,0 +1,46 @@
...
+> & { chain: { lastBlockHeaders: BlockHeader[] } };
+
+export interface ReducerHandler {
+	invoke: (params: Record<string, unknown>) => Promise<unknown>;
```
nazarhussain, line 26: Reducers are supposed to invoke/call/execute functions of other modules with some input. Should not the signature of `invoke` be having more arguments to specify that module name and that function reference?

framework/src/modules/base_module.ts
```diff
@@ -0,0 +1,82 @@
...
+	reducerHandler: ReducerHandler;
+}
+
+export interface AfterBlockApplyInput extends BeforeBlockApplyInput {
+	consensus: Consensus;
+}
+
+export abstract class BaseModule<T = unknown> {
+	public readonly config: GenesisConfig;
+	public readonly assets: BaseAsset[] = [];
```
nazarhussain, line 64: What if we name it as `transactionAssets`? As per spec document that's the definition of this attribute.

framework/src/modules/dpos/dpos_module.ts
```diff
@@ -0,0 +1,20 @@
...
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+import { BaseModule } from '../base_module';
+
+export class DPoSModule extends BaseModule {
+	public name = 'dpos';
+	public accountSchema = {};
```
nazarhussain, line 19: `type` is also an abstract property of `BaseModule` but not specified here. Same in other modules also.

framework/src/modules/token/token_module.ts
```diff
@@ -0,0 +1,35 @@
...
+		type: 'object',
+		properties: {
+			balance: {
+				dataType: 'uint64',
+			},
+		},
+		default: {
+			balance: BigInt(0),
+		},
+	};
```
nazarhussain, line 34: As per my understanding, this schema will refer to the object in account with key `token`, should not we specify nested `fieldNumber` inside the schema here? As its specified in the `transfer` module.

framework/src/modules/token/token_module.ts
```diff
@@ -0,0 +1,35 @@
...
+	public accountSchema = {
+		type: 'object',
+		properties: {
+			balance: {
+				dataType: 'uint64',
+			},
+		},
+		default: {
+			balance: BigInt(0),
+		},
```
nazarhussain, line 33: We need to decide on this as a pattern, if we support `default` as part of the schema then we have to update the configuration here. 

https://github.com/LiskHQ/lisk-sdk/blob/213506f490002f45eec5caef143ef764f7ddeca0/elements/lisk-validator/src/lisk_validator.ts#L33

framework/src/types.ts
```diff
@@ -91,7 +91,8 @@ export interface NetworkConfig {
 }
 
 export interface GenesisConfig {
-	epochTime: string;
+	[key: string]: unknown;
```
nazarhussain, line 94: I believe `GenesisConfig` should be strict to given definition, not open to any extra runtime attributes.

framework/src/types.ts
```diff
@@ -153,3 +160,21 @@ export interface TransactionJSON {
...
+// TODO: replace definition from lisk-chain after #5609 "Update lisk-chain to support the on-chain architecture"
+export interface Validator {
+	address: Buffer;
+	canVote: boolean;
+}
+
+// TODO: replace definition from lisk-chain after #5609 "Update lisk-chain to support the on-chain architecture"
+export interface AccountSchema {
+	type: string;
+	properties: Record<string, unknown>;
```
nazarhussain, line 173: We should use use `Schema` type from `lisk-codec` here and not keep it open to any attributes.

framework/src/modules/base_asset.ts
shuse2: consensus is only used for the internal information (like validators, finality), which should not be changed outside of framework.
That's why it is omitted, account/chain state should have enough information to be used in module/asset

framework/src/modules/token/token_module.ts
shuse2: yes, it does need the field number for the balance

framework/src/modules/token/token_module.ts
shuse2: it is not for the validation, but chain needs to know all the default value to create the default account by combining all the schema.

framework/src/types.ts
shuse2: Module will be instantiated with the genesis config. Other than the required properties from the framework, it will be basically schemaless and each module needs to validate and use it.
Otherwise, module specific config (like # of standby delegates for dpos), will not be able to specify without requiring

framework/src/types.ts
shuse2: properties is also open for the lisk-codec `Schema` but somehow what is important is there is required property `default` here

framework/src/errors.ts
shuse2: it is actually already set in the constructor

framework/src/application/schema/application_config_schema.ts
```diff
@@ -98,6 +98,11 @@ export const applicationConfigSchema = {
 					description:
 						'The unique name of the relevant community as a string encoded in UTF-8 format',
 				},
+				bftThreshold: {
+					type: 'number',
+					minimum: 1,
+					description: 'Number of validator required to set block finality',
```
mitsuaki-u, line 104:
```suggestion
					description: 'Number of validators required to set block finality',
```

framework/src/errors.ts
mitsuaki-u: `NONCE_OUT_OF_BOUNDS`?

ishantiw: yeah it doesnt sound assuring, maybe `nonSequentialNonceError` ?

framework/src/modules/base_asset.ts
```diff
@@ -0,0 +1,46 @@
...
+	ChainStateStore,
+	'consensus' | 'finalize' | 'createSnapshot' | 'restoreSnapshot'
+> & { chain: { lastBlockHeaders: BlockHeader[] } };
+
+export interface ReducerHandler {
+	invoke: (name: string, params: Record<string, unknown>) => Promise<unknown>;
+}
+
+export interface ApplyAssetInput<T> {
+	senderID: Buffer;
```
mitsuaki-u, line 30: In other parts of the code we are using camel-case like `senderId`

framework/src/modules/base_module.ts
```diff
@@ -0,0 +1,82 @@
...
+	public abstract name: string;
+	public abstract type: number;
+
+	public constructor(config: GenesisConfig) {
+		this.config = config;
+	}
+
+	public beforeTransactionApply?(input: TransactionApplyInput): Promise<void>;
+	public afterTransactionApply?(input: TransactionApplyInput): Promise<void>;
+	public afterGenesisBlockApply?(input: AfterGenesisBlockApplyInput<T>): Promise<void>;
```
mitsuaki-u, line 79: Maybe for readability put this on the top, since it will be applied first and order is correct

framework/src/types.ts
```diff
@@ -153,3 +160,21 @@ export interface TransactionJSON {
 	readonly id: string;
 	readonly asset: object;
 }
+
+// TODO: replace definition from lisk-chain after #5609 "Update lisk-chain to support the on-chain architecture"
+export interface Validator {
```
ishantiw, line 165: We have `LiskValidator` in validator lib that is used across framework serving completely different purpose to what this `Validator` means, maybe for more clarity and distinction we can rename it to `BlockValidator`?

framework/src/errors.ts
shuse2: SequentialNonce is specific to the sequence module rule, so i wanted to have generic name where nonce is not ready to be included in a block, but allowed in the pool

framework/src/modules/base_module.ts
shuse2: I think this order is enforced by the eslint

framework/src/modules/base_asset.ts
shuse2: I think we are converting `Id` to `ID` in general, so im following that

framework/src/modules/base_asset.ts
nazarhussain: We can go with it, but I believe if someone develop a custom module working on different consensus logic, then he may need this. Not sure what use case it could be, just a thought.

framework/src/modules/token/token_module.ts
nazarhussain: I do understand our use case, but if we have `default` key inside a schema, developers may get it confused with validator default values.

shuse2: The config has the same usage of default, and to keep it consistent within framework we will go with this approach

PR #5629, opened 2020-08-07
Update BFT interfaces to support onchain architecture - Closes #5613

### LINKED ISSUE #5613: Update lisk-bft to support on-chain architecture
### Description
*   Update BFT threshold to be constant from the constructor
*   Remove dependency of DPoS
    *   isStandbyDelegate should be replaced by canVote filter from getValidators
    *   isBootstrapPeriod should be replaced by canVote filter from getValidators
    *   minActiveHeight should be given by the chain library
*   Constant activeDelegates should be replaced by weight 1 filter from getValidators
*   Rename addNewBlock to applyBlockHeader
*   Rename verifyNewBlock to verifyBlockHeader
* Update chain to be dependency

### Motivation
All BFT related code needs to be updated to follow on-chain architecture concepts.

### Acceptance criteria
*   DPoS dependency does not exist
*   Update all unit tests

### Related Issues
#5603

shuse2:
### What was the problem?

This PR resolves #5613 

### How was it solved?

- Update interface to be consistent with other libraries
- Update to use validators given from lisk-chain
- Update to use bftThreshold 

### How was it tested?

- Tests are updated without changing the internal logic


elements/lisk-bft/src/utils/validators.ts
```diff
@@ -0,0 +1,27 @@
...
+
+export const getValidators = async (stateStore: StateStore): Promise<Validator[]> => {
+	const validatorsBuffer = await stateStore.consensus.get(CONSENSUS_STATE_VALIDATORS_KEY);
+		if (!validatorsBuffer) {
+			throw new Error('Validator set must exist');
+		}
+		const { validators } = codec.decode<{ validators: Validator[]}>(validatorsSchema, validatorsBuffer);
+
+		return validators;
+};
```
nazarhussain, line 27: Can we move this as pure function to the `lisk-chain` and export from there? Then we don't need to deal with this complexity only on once place.

elements/lisk-bft/src/bft.ts
```diff
@@ -146,18 +143,27 @@ export class BFT extends EventEmitter {
...
+		blockHeader: BlockHeader,
+		stateStore: StateStore,
+	): Promise<boolean> {
 		assert(blockHeader, 'No block was provided to be verified');
 
 		const roundsThreshold = 3;
-		const heightThreshold = this.constants.activeDelegates * roundsThreshold;
+		const validators = await getValidators(stateStore);
+		const numberOfVotingValidators = validators.filter(
+			validator => validator.isConsensusParticipant,
```
nazarhussain, line 156: I believe that's the property we discussed earlier related to boolean weight value.

shuse2: yes! with the discussion with research team, i updated

PR #5636, opened 2020-08-10
Update node to support on-chain architecture change - Closes #5612

### LINKED ISSUE #5612: Update framework/node to support on-chain architecture
### Description

* Update synchronizer to remove DPoS dependency
  * isActiveDelegate should be replaced by “getValidators” from consensus state of chain library, and filter by weight greater than 0
  * Replace delegatePerRound by numberOfValidators from chain
  * Update Current round calculation to use the numberOfValidators instead
  * FCS should only consider voting validator
    - In `isValidFor`, it is checking `isActiveDelegate`, but this should use `getValidators` from chain and check if the generator is included in the validator set and has voting power
* Update forge to remove DPoS dependency
  * Use getValidators from consensus or chain library instead of DPoS
* Update actions
  * getForgingStatusOfAllDelegates => Rename to getValidators
  * getTransactionsFees => Update response to
  ```js
  [{
    minFeePerBytes: number,
    moduleType: number,
    assetType: number,
    baseFee: number,
  }]
  ```
  * getAllDelegates => Remove, it should be handled in the DPoS module
  * getSchema => Update to use schema from lisk-chain and registered modules

### Motivation

All node related code needs to be updated to follow on-chain architecture concepts.

### Acceptance criteria

* Syncrhonizer and forger should not have any dependencies to the DPoS library
* All unit tests are updated
* Related Integration test are updated

### Related Issues

#5603

shuse2:
### What was the problem?

This PR resolves #5612 

### How was it solved?

- Update synchronizer to not to depend on DPoS
- Update transport to use the processor function instead of chain function
- Update node to use updated types

### How was it tested?

- Update syncrhonizer tests
- Update transport tests
- Update node tests
- Update test utils


elements/lisk-chain/src/data_access/data_access.ts
```diff
@@ -334,7 +334,7 @@ export class DataAccess {
 		};
 	}
 
-	public encode(block: Block): Buffer {
+	public encode(block: Block<unknown>): Buffer {
```
ManuGowda, line 337:
```suggestion
	public encode(block: Block<Record<string,unknown>>): Buffer {
```

framework/src/application/node/node.ts
```diff
@@ -264,51 +248,35 @@ export class Node {
 		}
 	}
 
+	public get networkIdentifier(): Buffer {
+		return this._networkIdentifier;
+	}
+
 	// eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types,@typescript-eslint/explicit-function-return-type
 	public get actions() {
 		return {
```
ManuGowda, line 257: what about the modules actions?

shuse2: Ill create another issue for this. i havent figured out how exactly it should be setup

elements/lisk-chain/src/utils/genesis_block.ts
```diff
@@ -14,8 +14,14 @@
...
 import { GenesisBlock, AccountSchema } from '../types';
-import { baseAccountSchema, getGenesisBlockHeaderAssetSchema, blockSchema, blockHeaderSchema } from '../schema';
+import {
+	baseAccountSchema,
+	getGenesisBlockHeaderAssetSchema,
+	blockSchema,
+	blockHeaderSchema,
+} from '../schema';
 
 export const readGenesisBlockJSON = (
```
nazarhussain, line 26: This function becomes too complex, we should refactor it later and build some easier pattern to mixin schemas together which is more readable.

elements/lisk-chain/test/utils/block.ts
```diff
@@ -61,10 +62,10 @@ export const genesisBlock: GenesisBlock = {
 			accounts: genesis.header.asset.accounts.map(account => ({
 				address: Buffer.from(account.address, 'base64'),
 				token: {
-					balance: BigInt(account.balance),
+					balance: BigInt(account.token?.balance),
```
nazarhussain, line 65: Why `token?` is used as optional but `sequence` is not?

framework/test/fixtures/blocks.ts
```diff
@@ -46,7 +46,7 @@ const getKeyPair = (): { publicKey: Buffer; privateKey: Buffer } => {
 
 export const encodeValidBlockHeader = (header: BlockHeader): Buffer => {
 	const chain = new Chain({
-		accountAsset: {},
+		accounts: defaultAccountSchema,
```
nazarhussain, line 49: If its account schema, can we rename the param name as well?

framework/test/fixtures/node.ts
```diff
@@ -12,24 +12,33 @@
...
+		delegates: [],
 	},
-	constants: {
+	genesisConfig: {
 		activeDelegates: 101,
+		standbyDelegates: 2,
 		maxPayloadLength: 15 * 1024,
+		bftThreshold: 68,
+		baseFees: [],
+		blockTime: 10,
```
nazarhussain, line 39:
```suggestion
		blockTime: 10, // 10 seconds
```

framework/test/fixtures/node.ts
```diff
@@ -12,24 +12,33 @@
...
+		minFeePerByte: 10000,
```
nazarhussain, line 40:
```suggestion
		minFeePerByte: 10000, // 10k beddows or 0.00001 LSK
```

framework/test/fixtures/node.ts
```diff
@@ -12,24 +12,33 @@
...
 	forging: {
 		waitThreshold: 2,
+		delegates: [],
 	},
-	constants: {
+	genesisConfig: {
 		activeDelegates: 101,
+		standbyDelegates: 2,
 		maxPayloadLength: 15 * 1024,
+		bftThreshold: 68,
```
nazarhussain, line 37:
```suggestion
		bftThreshold: 68, // Two third of active delegates Math.ceil(activeDelegates * 2 / 3)
```

framework/test/fixtures/node.ts
```diff
@@ -12,24 +12,33 @@
...
 	},
 	forging: {
 		waitThreshold: 2,
+		delegates: [],
 	},
-	constants: {
+	genesisConfig: {
 		activeDelegates: 101,
+		standbyDelegates: 2,
 		maxPayloadLength: 15 * 1024,
```
nazarhussain, line 36:
```suggestion
		maxPayloadLength: 15 * 1024, // 15kb
```

framework/src/application/node/node.ts
shuse2: https://github.com/LiskHQ/lisk-sdk/issues/5639

framework/test/fixtures/blocks.ts
shuse2: Since constractor affects subsequent PRs, I added the change on https://github.com/LiskHQ/lisk-sdk/issues/5639

Issue #5639, opened 2020-08-11
Update variable names of module interface


shuse2:
### Description

- Update transaction properties from `assetType` => `assetID` and `moduleType` => `moduleID`
- Update Base module `type` to `id`
- Update Base asset `applyAsset` to `apply`, `assetSchema` to `schema`, `validateAsset` to `validate` and `type` to `id`
- Update lisk-chain constructor param `accounts` to `accountSchemas`
- Update hook interface variable name `input` to `context`

### Motivation

Choosing names that are not redundant and and explain the behavior is critical for external interfaces.

### Acceptance Criteria

- All the names are update across the repository including protocol specs
- All test should be passing

PR #5641, opened 2020-08-11
Update lisk-genesis to support on-chain architecture #5640

### LINKED ISSUE #5640: Update lisk-genesis to support on-chain architecture
### Description

- validation happens at the module and chain, so it does not need to be done in lisk-genesis.
- lisk-genesis should accept the account schemas defined in the modules, and create use the schema

### Motivation

- Account schemas are defined in the modules, so lisk-genesis should not depends on it internally. It should obtain from the input.

### Acceptance Criteria
- Specific account properties is not defined
- Build and test for lisk-genesis is passing

shuse2:
### What was the problem?

This PR resolves #5640 

### How was it solved?

- Remove validation logic since it is written in the module.
- Update create function input to expect accountSchema defined in modules
- Update to use schema defined in lisk-chain

### How was it tested?

- Update all the tests
- Use sample schema which is same as in module to test the basic feature


elements/lisk-chain/src/utils/account.ts
```diff
@@ -0,0 +1,35 @@
...
+	for (const [name, schema] of Object.entries(accountSchemas)) {
+		const { default: defaultValue, ...schemaWithoutDefault } = schema;
+		accountSchema.properties[name] = schemaWithoutDefault;
+		(accountSchema.required as string[]).push(name);
+		defaultAccount[name] = defaultValue;
+	}
+	return {
+		schema: accountSchema,
+		defaultAccount,
+	};
```
nazarhussain, line 34: Suggest to use the pattern we are using on other places. `default` should be the key under schema object containing default account.

elements/lisk-chain/src/chain.ts
```diff
@@ -124,14 +123,9 @@ export class Chain {
...
 
-		this._accountSchema = objects.cloneDeep(baseAccountSchema);
-		this._defaultAccount = {};
-		for (const [name, schema] of Object.entries(accounts)) {
-			const { default: defaultValue, ...schemaWithoutDefault } = schema;
-			this._accountSchema.properties[name] = schemaWithoutDefault;
-			(this._accountSchema.required as string[]).push(name);
-			this._defaultAccount[name] = defaultValue;
-		}
+		const { defaultAccount, schema } = getAccountSchemaAndDefault(accounts);
```
nazarhussain, line 126: This argument here looks like accounts objects while the function accepts the account schema. This is related to earlier comment on other PR to rename `Chain` constructor arguments name `accounts` to `accountSchema`

elements/lisk-chain/src/utils/account.ts
nazarhussain, line 34: Should not we use the same pattern we use on other places. `default` should be a key under schema object, containing default account.

elements/lisk-chain/src/chain.ts
shuse2: For the constructor param I added a point on https://github.com/LiskHQ/lisk-sdk/issues/5639

PR #5651, opened 2020-08-12
Update tx-pool to support on-chain architecture - Closes #5619

### LINKED ISSUE #5619: Update transaction-pool to support on-chain architecture
### Description

- Update constructor to have baseFee and transaction per byte in format of
```
{
  minFeePerBytes: number,
  baseFees: [{
    moduleType: number,
    assetType: number,
    baseFee: bigint,
  }]
}
```
- Remove minFee from transaction interface, and calculate using `baseFee` and `transactionPerByte`
- Update `ApplyFunction` interface return type to be `void` and expect it to throw.
  - nonce higher error should be checked with `error.code === 'nonce_not_ready'`

### Motivation

With on chain architecture, there are changes in transaction and transaction response type.

### Acceptance Criteria

- All transaction pool unit tests are updated

ishantiw:
### What was the problem?

This PR resolves #5619 

### How was it solved?

- Accept `minFeePerByte` and `baseFees` in the constructor
- Update `applyTransactions` interface to return `Promise<void>`
- Catch errors from `applyTransactions` and update `getStatus`
- Remove `TransactionError` and update `TransactionObject` interface
- `TransactionApplyError` to capture errors coming from different hooks and the trx id
- `verifyTransactions` uses `TransactionApplyError` while iterating over transactions
- Update transactionPool tests
- Pass `baseFees` and `minFeePerByte` from `genesisConfig` to txPool constructor in node 
- Define type of `applyFunction` in node

### How was it tested?

Run `npm t` under `elements/lisk-transaction-pool`


elements/lisk-transaction-pool/src/transaction_pool.ts
```diff
@@ -310,19 +308,23 @@ export class TransactionPool {
...
-		return (trx.fee - trx.minFee) / BigInt(trx.getBytes().length);
+		return (trx.fee - this._calculateMinFee(trx)) / BigInt(trx.getBytes().length);
 	}
 
-	// eslint-disable-next-line class-methods-use-this
-	private _getStatus(txResponse: TransactionResponse): TransactionStatus {
-		if (txResponse.status === Status.OK) {
-			debug('Received PROCESSABLE transaction');
+	private _calculateMinFee(trx: Transaction): bigint {
+		const baseFee = this._baseFees.filter(
```
shuse2, line 315: `find` and if `undefined` set to `BigInt(0)` would be better in case baseFee will not be  present

framework/src/application/node/node.ts
```diff
@@ -541,9 +543,15 @@ export class Node {
...
 		this._transactionPool = new TransactionPool({
-			// FIXME: #5619 any should be removed
+			baseFees: this._options.genesisConfig.baseFees.map(fees => ({
+				...fees,
+				baseFee: BigInt(fees.baseFee),
+			})),
+			minFeePerByte: this._options.genesisConfig.minFeePerByte,
 			// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
-			applyTransactions: this._processor.verifyTransactions.bind(this._processor) as any,
+			applyTransactions: this._processor.verifyTransactions.bind(
```
shuse2, line 552: i think verifyTransactions takes 2 inputs `transactions` and `stateStore`, so it should be
```
applyTransactions: async (transactions) => {
  const stateStore = this._chain.newStateStore();
  return this._processor(verifyTransactions(transactions, stateStore);
}
```

PR #5658, opened 2020-08-13
Plugins data should be associated to the data path of the application - Closes #5598

### LINKED ISSUE #5598: Plugins data should be associated to the data path of the application
### Expected behavior

All data associated with a particular application should reside in one `data path`. Either its databases shipped with SDK e.g. `blockchain.db` or any plugins data e.g. `lisk-framework-forger-plugin.db`. 

If the app is stared with following data path: 

```
~/.lisk/default
```

Then all the data related to plugins should be stored inside `${rootPath}/${label}/plugins/data/`, which is `~/.lisk/devnet/plugins/data/lisk-framework-forger-plugin.db` for `devnet`

### Actual behavior

Currently the forger plugin data is stored in `.lisk/plugins/data/` which will cause overriding data for different applications using same plugin. 

5.0.*

ManuGowda:
### What was the problem?

This PR resolves #5598 

### How was it solved?

- updated plugins data path to `rootPath/plugins/data/.db`

### How was it tested?

- Added unit test


framework-plugins/lisk-framework-forger-plugin/src/types.ts
```diff
@@ -14,7 +14,8 @@
 
 export interface Options {
 	version: string;
-	readonly dataPath: string;
+	readonly rootPath: string;
+	readonly dataPath?: string;
```
shuse2, line 18: should it just use the same variable `dataPath`?
Also, rootPath variable name is confusing since it is used in different way in the framework

framework-plugins/lisk-framework-forger-plugin/src/defaults/default_config.ts
```diff
@@ -29,15 +29,6 @@ export const defaultConfig = {
 			description:
 				'Explicitly allowing some identified entities access to forger plugin endpoints.',
 		},
-		dataPath: {
```
shuse2, line 32: parameter itself should exist as optional

framework/src/controller/controller.ts
```diff
@@ -119,7 +119,7 @@ export class Controller {
 
 		for (const alias of Object.keys(plugins)) {
 			const klass = plugins[alias];
-			const options = pluginOptions[alias];
+			const options = { dataPath: this.config.rootPath, ...pluginOptions[alias] };
```
shuse2, line 122: I think `this.config.rootPath` is not equivalent to `dataPath`.
if we add plugin options, maybe it's better to just pass all the root config and plugin config like below?
```
nodeConfig: {},
plugin: {},
```

PR #5668, opened 2020-08-18
Update rootPath in controller config to dataPath - Closes #5667

### LINKED ISSUE #5667: Update rootPath in controller config to dataPath
### Description

Update rootPath in controller config to dataPath

### Motivation

`~/.lisk/default` is the default data path and depending on network and label it might change, root path is `~/.lisk`, which could lead to confusion.

### Acceptance Criteria

- Update `rootPath` to `dataPath`
- Update corresponding code and tests

ManuGowda:
### What was the problem?

This PR resolves #5667 

### How was it solved?

- Updated `rootPath` to `dataPath`

### How was it tested?

- Updated unit test

### DIFF
--- a/framework/src/controller/controller.ts
+++ b/framework/src/controller/controller.ts
@@ -39,7 +39,7 @@ export interface ControllerOptions {
 }
 
 interface ControllerConfig {
-	readonly rootPath: string;
+	readonly dataPath: string;
 	readonly socketsPath: SocketPaths;
 	readonly dirs: {
 		readonly root: string;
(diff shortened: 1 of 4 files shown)


framework/src/controller/controller.ts
```diff
@@ -88,7 +88,7 @@ export class Controller {
 
 		const dirs = systemDirs(this.appLabel, options.config.rootPath);
```
nazarhussain, line 89: Please change same in the `systemDirs` helper as well.

nazarhussain (approved):

PR #5669, opened 2020-08-18
Add account del function in state store - Closes #5631

### LINKED ISSUE #5631: Add del on state store account
### Description

account state store should have `del` function.
- del should check if exist in state store memory, and delete from there
- del should delete from the database using batch
- Add schema deleted: [{ key: string, value: Buffer }] to diff schema
- del should save the complete diff into `deleted` of the diff schema if existed in DB.
- When deleting block, it should restore the value to the database

### Motivation

From the reclaim transaction, it requires the deleting of the account.

### Acceptance Criteria

- All the scenario are covered in the unit test

shuse2:
### What was the problem?

This PR resolves #5631 

### How was it solved?

- Add del function in the account state store
- Add deleted property in the diff

### How was it tested?

- Add unit test
- Add integration test


elements/lisk-chain/src/state_store/account_store.ts
```diff
@@ -132,6 +158,18 @@ export class AccountStore {
 				}
 			}
 		}
+		for (const deletedAddress of this._deletedKeys) {
+			const initialAccount = this._initialAccountValue.get(deletedAddress);
```
nazarhussain, line 162: User may have modified the account before deleting it, logically it seems `_initialAccountValue` contains only initial account fetched from db, any modification to it will not be persisted in it. If that's true, then user may lost the changes in the account state after deleting.

nazarhussain (changes requested):

elements/lisk-chain/src/state_store/account_store.ts
nazarhussain: We may add some test case to cover the scenario, if user modified account before deleting, then the state diff must contains the modified account in deleted buffer.

elements/lisk-chain/test/unit/state_store/account.spec.ts
```diff
@@ -166,6 +170,49 @@ describe('state store / account', () => {
 		});
 	});
 
+	describe('del', () => {
+		it('should throw an error if address does not exist nither in momory or in database', async () => {
```
ishantiw, line 174:
```suggestion
		it('should throw an error if the address exists neither in memory nor database', async () => {
```

elements/lisk-chain/src/state_store/account_store.ts
shuse2: yes, that's how it should work. if it's deleted, any changes happened before saving should be ignored and keep the original state in the diff.
Otherwise, when block is deleted it will become invalid state

ishantiw (approved):

elements/lisk-chain/src/state_store/account_store.ts
shuse2: And the case is covered https://github.com/LiskHQ/lisk-sdk/pull/5669/files#diff-5fe7163bd6761e4e768a4332ad906cc5R135

nazarhussain (approved):

PR #5674, opened 2020-08-19
Create a new event VALIDATOR_CHANGED in lisk-chain - Closes #5656

### LINKED ISSUE #5656: Create a new event VALIDATOR_CHANGED in lisk-chain
### Description
- Add `EVENT_VALIDATORS_UPDATED` in lisk-chain  when setValidators is called. the payload should be `{ validators: Validator[] }` which is set.
- Emit channel event `chain:validators:update` by listening to the above event. The payload must be in JSON format. (ie: address Buffer must be converted to Base64 string)

### Motivation

In `lisk-dpos` we had an event emitter implemeation for `EVENT_ROUND_CHANGED`. This does not seems to be relevant any more for the new dpos module. We rather have a different event `EVENT_VALIDATOR_CHANGED`. 

https://github.com/LiskHQ/lisk-sdk/blob/0bd21816e1e1e42a7abaf3619eb37a3f8a915e69/elements/lisk-dpos/src/delegates_info.ts#L144-L147

Round information is no more relevant to on-chain architecture so any of such event should be removed or replaced. 

### Acceptance Criteria
- Add unit test to see if the event is emitted on lisk-chain

pablitovicente:
### What was the problem?

This PR resolves #5656

### How was it solved?

- by emitting an event when `setValidators()` is executed

### How was it tested?

- Unit Test
- To test manually change the block time to 4 (as you need to wait 103 blocks for this to fire so lowering the time makes it a little shorter wait) add a console log or subscribe to the new event and you should get the date (if you start a clean node k/v store you need to additionally wait 3 full rounds so in total 4 rounds or around 412 blocks)


framework/src/application/node/node.ts
```diff
@@ -669,6 +669,31 @@ export class Node {
 				);
 			},
 		);
+
+		// eslint-disable-next-line @typescript-eslint/no-misused-promises
+		this._chain.events.on(
+			EVENT_VALIDATORS_CHANGED,
+			// eslint-disable-next-line @typescript-eslint/no-misused-promises
+			async (
```
shuse2, line 677: this `async` and the eslint-disable is not needed

framework/src/application/node/node.ts
```diff
@@ -669,6 +669,31 @@ export class Node {
...
+					validators: [
+						{
+							address: Buffer;
+							isConsensusParticipant: boolean;
+							minActiveHeight: number;
+						},
+					];
+				},
+				// eslint-disable-next-line @typescript-eslint/require-await
+			): Promise<void> => {
```
shuse2, line 688: also return type should be `void`

shuse2 (changes requested):

framework/src/application/node/node.ts
pablitovicente: Thought we were making all these async as a pattern in the past. Will remove.

ManuGowda (approved):

framework/src/application/node/node.ts
pablitovicente: Changed to sync. removed all eslint comments

framework/src/application/node/node.ts
pablitovicente: changed to void after making async.

shuse2 (approved):

PR #5681, opened 2020-08-20
Add support to export JSON object for genesis block in lisk-genesis - Closes #5680

### LINKED ISSUE #5680: Add support to export JSON object for genesis block in lisk-genesis
### Description

We need to add functionality to export plain JSON object for a genesis block inside `lisk-genesis`

### Motivation

The main usage of `lisk-genesis` is to create a genesis block and then use it in the app. The medium between the app and genesis block would mostly be a JSON file. So functionality to convert GenesisBlock to a JSON file would be handy for `lisk-genesis` 

### Acceptance Criteria

- Any genesis block object can be convert to JSON by providing account schema 
- It should return a JSON object not the json string

nazarhussain:
### What was the problem?

This PR resolves #5680

### How was it solved?

Added method `getGenesisBlockJSON` to `lisk-genesis`

### How was it tested?

Run all unit tests.


pablitovicente (approved):

shuse2 (commented):
One minor comment but LGTM

PR #5686, opened 2020-08-21
Keys module validation errors for genesis block provide wrong account index - Closes #5685

### LINKED ISSUE #5685: Keys module validation errors provide fixed account index for genesis block accounts
### Expected behavior

The error messages must be descriptive and well narrative, to help and pin point the exact source of data. 

### Actual behavior

Currently the keys modules is providing error messages for the genesis block account validations, which shows fixed index `0` for every account. In case of error, user can't figure out the actual account. 

https://github.com/LiskHQ/lisk-sdk/blob/cd1ad4e958a43829a20cba2b59d958b147b6a9f6/framework/src/modules/keys/keys_module.ts#L151
https://github.com/LiskHQ/lisk-sdk/blob/cd1ad4e958a43829a20cba2b59d958b147b6a9f6/framework/src/modules/keys/keys_module.ts#L172

### Steps to reproduce
- Create a genesis block with a multi-sig account which should not have valid keys 
- Link the SDK to Lisk Core
- Load up the node with that genesis block 

5.0

nazarhussain:
…g for keys

### What was the problem?

This PR resolves #5685

### How was it solved?

Improved the error messages by adding keys and account address along with account index. 

### How was it tested?

Run unit tests

### DIFF
--- a/framework/src/modules/keys/keys_module.ts
+++ b/framework/src/modules/keys/keys_module.ts
@@ -143,23 +143,23 @@ export class KeysModule extends BaseModule {
 		genesisBlock,
 	}: AfterGenesisBlockApplyContext<AccountKeys>): Promise<void> {
 		const errors = [];
-		for (const account of genesisBlock.header.asset.accounts) {
+		for (const [index, account] of genesisBlock.header.asset.accounts.entries()) {
 			if (!bufferArrayOrderByLex(account.keys.mandatoryKeys)) {
 				errors.push({
 					message: 'should be lexicographically ordered',
 					keyword: 'mandatoryKeys',
-					dataPath: '.accounts[0].keys.mandatoryKeys',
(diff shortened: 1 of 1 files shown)


framework/src/modules/keys/keys_module.ts
```diff
@@ -143,23 +143,23 @@ export class KeysModule extends BaseModule {
 		genesisBlock,
 	}: AfterGenesisBlockApplyContext<AccountKeys>): Promise<void> {
 		const errors = [];
-		for (const account of genesisBlock.header.asset.accounts) {
+		for (const [index, account] of genesisBlock.header.asset.accounts.entries()) {
```
shuse2, line 146: Maybe related to other issue, but since this loop executes a lot, probably better to use
```
for (let index = 0; index < genesisBlock.header.asset.accounts.length; i += 1) {
  const account = genesisBlock.header.asset.accounts[i];
```

PR #5696, opened 2020-08-24
Move state store mock to chain - Closes #5672

### LINKED ISSUE #5672: StateStoreMock should be defined in lisk-chain for testing
### Description

Move StateStoreMock to lisk-chain and expose as
```
testings: {
  mocks: {
    StateStoreMock,
  }
}
```

### Motivation

StateStore implementation is defined in the lisk-chain, but the mock utils is defined in `lisk-utils`.
Any changes in state store introduces change in the StateStoreMock, so it should be defined in the same package.

### Acceptance Criteria

- StateStoreMock should be removed from lisk-utils
- Any usage of lisk-utils StateStoreMock should be replaced by lisk-chain state store mock

shuse2:
### What was the problem?

This PR resolves #5672 

### How was it solved?

- Move state store mock to chain library and update test import path

### How was it tested?

- All test should pass


elements/lisk-chain/src/index.ts
```diff
@@ -42,5 +42,6 @@ export type {
 } from './types';
 export { Slots } from './slots';
 export { readGenesisBlockJSON, getValidators, getAccountSchemaWithDefault } from './utils';
+export * as testing from './testing';
```
nazarhussain, line 45: I am not sure if that also applied for a node module, but what if we put this mock into `__mocks__` directory as per Jest specification for manual mocks. We can still export if we want.

nazarhussain (changes requested):

elements/lisk-chain/src/index.ts
shuse2: Im not sure if it works with `__mocks__` within `node_module`, but we can add it later if it works?
But anyway, i think we should expose this as testing

nazarhussain (approved):

ishantiw (approved):

PR #5703, opened 2020-08-25
Add action event for module - Closes #5645

### LINKED ISSUE #5645: Add support for module action and events
### Description

- Register all module actions to the channel
- Register all module events to the channel
- Define a way to provide data access in limited manner to the action.
- Add action 'getRegisteredDelegates` in the dpos module, which returns all delegates with address and username
- Update /api/delegates endpoint to use new action

Given channel to BaseModule should have interface of 
```
interface Channel {
  emit(name: string, data?: object): void
}
```
and dataAccess should have
```
interface DataAccess {
  getChainState(key: string): Promise<Buffer | undefined>
  getAccount(address: Buffer): Promise<Account>
}
```

### Motivation

Defined actions and events should be callable through channel.

### Acceptance Criteria

- Test if the module defined action is callable from plugin
- Test if module defined events can be listened from plugin
- Enable /api/delegates functional tests

### Additional Information
- Modules are instantiated in application.ts

ishantiw:
### What was the problem?

This PR resolves #5645

### How was it solved?

- Register all module actions to the channel
- Register all module events to the channel
- Add type definition for DataAccess for modules 
- Add Handlers for setting channel and `dataAccess`
- Add `registerChannel` to register channel for publish event
- Add `setActions` to define actions for a module
- Add `setDataAccess` method to set `dataAccess` for modules
- Update /api/delegates endpoint to use dpos:getAllDelegates

### How was it tested?

Run `npm run test:unit` under `/framework` and `npm run test:functional` under `/framework-plugins/lisk-framework-http-api-plugin delegates.spec`


framework/src/modules/base_module.ts
```diff
@@ -22,24 +22,37 @@ import {
...
 export abstract class BaseModule {
 	public readonly config: GenesisConfig;
 	public transactionAssets: BaseAsset[] = [];
 	public reducers: Reducers = {};
 	public actions: Actions = {};
 	public events: string[] = [];
 	public accountSchema?: AccountSchema;
-
+	public channel!: Channel;
+	public dataAccess!: BaseModuleDataAccess;
```
shuse2, line 41: these 2 properties can be private?

framework/src/modules/base_module.ts
```diff
@@ -22,24 +22,37 @@ import {
...
-	public constructor(config: GenesisConfig) {
-		this.config = config;
+	public constructor(genesisConfig: GenesisConfig) {
+		this.config = genesisConfig;
 	}
 
+	public registerChannel = (channel: Channel): void => {
+		this.channel = channel;
+	};
+	public setDataAccess = (dataAccess: BaseModuleDataAccess): void => {
```
shuse2, line 52: maybe `registerDataAccess`?

framework/src/modules/dpos/dpos_module.ts
```diff
@@ -62,6 +64,20 @@ export class DPoSModule extends BaseModule {
 	public constructor(config: GenesisConfig) {
 		super(config);
 
+		// Set actions
+		this.actions = {
+			getAllDelegates: async _ => {
+				const validatorsBuffer = await this.dataAccess.getChainState(
```
shuse2, line 70: it should check if `validatorsBuffer`  is undefined or not

shuse2 (changes requested):

framework/src/modules/base_module.ts
ishantiw: I think you mean `protected` as these props will be used in the extended classes

ishantiw: or make a protected getter for both

shuse2 (approved):

PR #5711, opened 2020-08-26
Revert to display hex string - Closes #5706

### LINKED ISSUE #5706: Update to use hex for representation of buffer in string
### Description

Replace all base64 usage to hex

### Motivation

Currently we use base64 as string representation of buffer, but with below 2 points we decided to move back to hex.
- When ID or address is used in URL, it requries to be URIEncoded
- Consistency with the UI products
- General industry uses hex

We acknowledge the below 2 points.
- payload size from actions/events of framework will increase
- JSON format will not follow protobuf definition

### Acceptance Criteria

- All base64 usage should be converted back to hex, which include plugins (httpAPI, forger), json files and logs

shuse2:
### What was the problem?

This PR resolves #5706 

### How was it solved?

- Replace all base64 usage to hex

### How was it tested?

All build and test should be passing and no base64 should be left in repo


ManuGowda (commented):
can we remove `isBase64String` from validator if not required?

ManuGowda (approved):

pablitovicente (approved):

PR #5716, opened 2020-08-27
Add getRegisteredModules and getSchema actions and function

### LINKED ISSUE #5671: Add action to return registered module and assets along with name
### Description

- Action `getSchema().transactionsAssets` should also return moduleName and assetName.
- Action `getRegisteredModules()` should return the registered module information.
- Action `getNodeInfo` should return additional property `registeredModules` with below interface
- Add application function `getSchema` and `getRegisteredModule`
- Remove `getTransactionsFees` and replace the usage with `nodeInfo`
```
getSchema: () => RegisteredSchema;
getRegisteredModules: () => RegisteredModule[];
```
where actions, events and reducer should only return the key registered.

interface should be below
```
export interface RegisteredModule {
	id: number;
	name: string;
	actions: string[];
	events: string[];
	reducers: string[];
	transactionAssets: {
		id: number;
		name: string;
	}[];
}

export interface RegisteredSchema {
	account: Schema;
	block: Schema;
	blockHeader: Schema;
	blockHeadersAssets: { [version: number]: Schema };
	transaction: Schema;
	transactionsAssets: {
		moduleID: number;
		moduleName: string;
		assetID: number;
		assetName: string;
		schema: Schema;
	}[];
}
```
- Possibly remove `setDataAccess` and `registerChannel` from base module (consider making account schema to be static
  (name, id for both base module and asset)

### Motivation

For external purpose, it should know what modules are registered and how to convert an ID to a name.

### Acceptance Criteria

- Update functional test for the nodeInfo
- Update functional test for getSchema if it contains the name

### Additional information
- Property of RegisteredSchema is slightly updated to make it consistent
- https://github.com/LiskHQ/lisk-core/issues/318 needs a way to get the same information through applicaton instance

shuse2:
### What was the problem?

This PR resolves #5671 

### How was it solved?

- Fix the initialization order
- Add 2 actions to handle getRegisteredModule and getSchema
- Expose the function same as the action above
- remove transactionFee action and use nodeInfo instead

### How was it tested?

- Add functional test to check if the results are returned properly


ManuGowda (approved):

framework/src/application.ts
```diff
@@ -183,16 +187,24 @@ export class Application {
 
 	public registerModule(Module: typeof BaseModule): void {
 		assert(Module, 'Module implementation is required');
+		const InstanciabltModule = Module as InstantiableBaseModule;
+		const moduleInstance = new InstanciabltModule(this.config.genesisConfig);
+		this._node.registerModule(moduleInstance);
+	}
```
ishantiw, line 193:
```suggestion
		const InstantiableModule = Module as InstantiableBaseModule;
		const moduleInstance = new InstantiableModule(this.config.genesisConfig);
		this._node.registerModule(moduleInstance);
	}
```

ishantiw (approved):

PR #5719, opened 2020-08-28
Add nonce sort on the node/transactions

### LINKED ISSUE #5710: API /transactions and /node/transactions sort by nonce
### Description

Allow sorting of transactions for `/transactions` and `/node/transactions` by nonce for a given account.
 
### Motivation

With the removal of timestamp from `/transactions` and `/node/transactions` we do not have a mechanism to sort transactions. 

### Acceptance Criteria

- Should allow sorting by nonce for endpoints `/transactions` and `/node/transactions`
- Remove `receivedAt` from swagger  https://github.com/LiskHQ/lisk-sdk/blob/65766520f5744e34c6751651206d1aeb5bdf7a5d/framework/src/modules/http_api/schema/swagger.yml#L1481
- Add functional test for endpoints `/transactions` and `/node/transactions` to sort by nonce.

### Additional Information
- Required for Core Version `3.0.0-beta.0`

shuse2:
### What was the problem?

This PR resolves #5710 

### How was it solved?

- Add nonce sort in swagger
- Add handler for the sort

### How was it tested?

- Add a simple functional test

### DIFF
--- a/framework/src/modules/http_api/controllers/node.js
+++ b/framework/src/modules/http_api/controllers/node.js
@@ -71,6 +71,24 @@ function _filterTransactions(transactions, filters) {
 				return left.type - right.type;
 			});
 		}
+		if (key === 'nonce') {
+			filteredTransactions.sort((a, b) => {
+				let left = a;
+				let right = b;
+				if (order === 'desc') {
+					left = b;
+					right = a;
+				}
+				const diff = BigInt(left.nonce) - BigInt(right.nonce);
+				if (diff > BigInt(0)) {
+					return 1;
+				}
+				if (diff < BigInt(0)) {
+					return -1;
+				}
+				return 0;
+			});
+		}
 	}
 	return filteredTransactions.slice(offset, limit + offset);
 }
--- a/framework/test/mocha/functional/http/get/node/transactions_ready.js
+++ b/framework/test/mocha/functional/http/get/node/transactions_ready.js
@@ -318,6 +318,24 @@ describe('GET /api/node', () => {
 					});
 				});
 
+				describe('nonce', () => {
+					it('sorted by nonce:asc should be ok', async () => {
+						return ReadyEndpoint.makeRequest({ sort: 'nonce:asc' }, 200).then(
+							res => {
+								expect(res.body).to.not.be.empty;
+							},
+						);
+					});
+
+					it('sorted by nonce:desc should be ok', async () => {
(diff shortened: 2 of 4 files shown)


framework/test/mocha/functional/http/get/node/transactions_ready.js
```diff
@@ -318,6 +318,24 @@ describe('GET /api/node', () => {
 					});
 				});
 
+				describe('nonce', () => {
+					it('sorted by nonce:asc should be ok', async () => {
+						return ReadyEndpoint.makeRequest({ sort: 'nonce:asc' }, 200).then(
```
ManuGowda, line 323: return ReadyEndpoint
						.makeRequest({ sort: 'nonce:asc' }, 200)
						.then(res => {
							const values = _.map(res.body.data, 'nonce').map(value => {
								return parseInt(value, 10);
							});

							expect(_(_.clone(values)).sortNumbers('asc')).to.be.eql(values);
						});
And same for below test?

shuse2: It seems like above tests similar to this is written in the same way

framework/src/modules/http_api/controllers/node.js
```diff
@@ -71,6 +71,24 @@ function _filterTransactions(transactions, filters) {
...
+				const diff = BigInt(left.nonce) - BigInt(right.nonce);
+				if (diff > BigInt(0)) {
+					return 1;
+				}
+				if (diff < BigInt(0)) {
+					return -1;
+				}
+				return 0;
+			});
+		}
```
nazarhussain, line 91: Should not we make this sorting logic more generic so we can reuse it. Here its the exactly same logic compared to sorting with `fee`.

nazarhussain (changes requested):

framework/src/modules/http_api/controllers/node.js
shuse2: true. but this is already removed logic in v5. I think it doesn't have so much benefit in making it generic?

framework/test/mocha/functional/http/get/node/transactions_ready.js
ManuGowda: ya it was not adding much value..

ManuGowda (approved):

nazarhussain (approved):

PR #5733, opened 2020-08-31
Fix transaction validation process - Closes #5712

### LINKED ISSUE #5712: Submitting a transaction with below minFee requirement gives txPool feePriority error
### Expected behavior

When submitting a tx with insufficient fee, `Insufficient transaction fee. Minimum required fee is: ${minimumRequiredFee.toString()}` should be given.

### Actual behavior

Gets error: `Rejecting transaction due to failed minimum entrance fee priority requirement` from txPool.

This is due to txPool check happening before token module check.

### Steps to reproduce

Send tx with below min fee requirement. 

5.0-

shuse2:
### What was the problem?

This PR resolves #5712 

### How was it solved?

- Move minFee validation to lisk-chain
- Add additional static check for the transaction

### How was it tested?

- Send a transaction less than min fee and check the error message

### DIFF
--- a/elements/lisk-chain/src/chain.ts
+++ b/elements/lisk-chain/src/chain.ts
@@ -59,6 +59,7 @@ import {
 	getRegisteredBlockAssetSchema,
 	validatorsSchema,
 } from './schema';
+import { Transaction, calculateMinFee } from './transaction';
 
 interface ChainConstructor {
 	readonly db: KVStore;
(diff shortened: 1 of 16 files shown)


ishantiw (approved):

mitsuaki-u (changes requested):
Prefer to move validation logic to `transaction.validate` directly instead of in `lisk-chain` by passing in required constants.

mitsuaki-u (approved):

PR #5735, opened 2020-08-31
Update balance in beforeTransactionApply - Closes #5704

### LINKED ISSUE #5704: Fee not collected from sender for any transaction
### Expected behavior

When registering a multi signature account the registration fee should be deducted from the senderPublicKey account.

### Actual behavior

The fee is not deducted.

### Steps to reproduce

- fund an account with enough balance for a multi-signature registration with 4 members (25 LSK should be ok)
- Send a registration for that account where the account is not member (might or might not be related to the account not being member but this needs to be investigated as part of this issue)
- Check the account after registration and the balance will be the same as when funded

5.0

mitsuaki-u:
### What was the problem?

This PR resolves #5704 

### How was it solved?

Deduct fee from sender in `beforeTransactionApply` stage. Added unit test.

### How was it tested?

Unit test in token module.

### DIFF
--- a/framework/src/modules/token/token_module.ts
+++ b/framework/src/modules/token/token_module.ts
@@ -101,8 +101,11 @@ export class TokenModule extends BaseModule {
 	}
 
 	// eslint-disable-next-line @typescript-eslint/require-await
-	public async beforeTransactionApply({ transaction }: TransactionApplyContext): Promise<void> {
-		// Throw error if fee is lower than minimum fee (minFeePerBytes + baseFee)
+	public async beforeTransactionApply({
+		transaction,
+		stateStore,
+	}: TransactionApplyContext): Promise<void> {
+		// Check if transaction fee is lower than the minimum required fee (minFeePerBytes + baseFee)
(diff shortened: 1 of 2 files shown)


framework/src/modules/token/token_module.ts
```diff
@@ -114,6 +117,12 @@ export class TokenModule extends BaseModule {
 				`Insufficient transaction fee. Minimum required fee is: ${minimumRequiredFee.toString()}`,
 			);
 		}
+
+		// Deduct transaction fee from sender balance
+		const senderAddress = transaction.senderID;
+		const sender = await stateStore.account.getOrDefault<TokenAccount>(senderAddress);
```
shuse2, line 123: sender must always exist so prefer `get`
```suggestion
		const sender = await stateStore.account.get<TokenAccount>(senderAddress);
```

shuse2 (changes requested):

pablitovicente (approved):
Other than Shu's comment looks good.

shuse2 (approved):

PR #5740, opened 2020-08-31
Extend Keys module test coverage - Closes #5720

### LINKED ISSUE #5720: Extend Keys module test coverage
### Description

We need to extend coverage of the Keys module to more edge cases

### Motivation

Assuring we are covering all bases

### Acceptance Criteria

The following tests are added:

- [x] ~~multisig registration: check the existing protocol specs and use all available cases that are not already used in tests~~ (already being tested by modifying base case)
- [x] transaction from multisig account: account with only optional and numberOfSignatures < optionalKeys.length
- [x] afterGenesisBlockApply: valid keys property (should be ok)
- [x] afterGenesisBlockApply: key repeated in mandatory and optional key sets (should fail)
- [x] afterGenesisBlockApply: 64 keys distributed among mandatory and optional key sets (should be ok)
- [x] afterGenesisBlockApply: numberOfSignatures is bigger than count of optional plus mandatory keys (should fail)
- [x] afterGenesisBlockApply: test that PK is member of either mandatory or optional (should be ok)
- [x] afterGenesisBlockApply: test that PK is not member of either mandatory or optional (should be ok)
- [x] afterGenesisBlockApply: 64 mandatory keys and numberOfSignatures < 64 (should fail)
- [x] afterGenesisBlockApply: 64 optional keys and numberOfSignatures < 64 (should be ok)

pablitovicente:
### What was the problem?

This PR resolves #5720

### How was it solved?

- by extending test coverage

### How was it tested?

added tests


shuse2 (approved):

framework/test/unit/modules/keys/keys_module.spec.ts
```diff
@@ -944,6 +972,143 @@ describe('keys module', () => {
...
+
+		it('should not fail for a maximum of 64 optioonal keys and number of signatures smaller than 64', async () => {
+			// Arrange
+			const accounts = cloneDeep(genesisBlock.header.asset.accounts);
+
+			const optionalKeys = [...Array(64).keys()]
+				.map(() => getRandomBytes(32))
+				.sort((a, b) => a.compare(b));
+
+			accounts[0].keys.numberOfSignatures = 60;
```
ManuGowda, line 1094: why 60? they are all optional rite?

ManuGowda (approved):

framework/test/unit/modules/keys/keys_module.spec.ts
pablitovicente: They are all optional but you need to configure a `numberOfSignatures` >= 1 and I randomly choose 60

Issue #5726, opened 2020-08-28
Extend DPoS module tests coverage


nazarhussain:
### Description

During the QA process identified that some of the test scenarios are not covered by unit tests. So we need to extend the tests and add coverage for those. 

### Motivation

Make sure all scenarios listed in QA process are covered by unit tests. 

### Acceptance Criteria

- Register username over maximum length 20
- Register username as empty string
- Register username which have special chracters !@$&_.
- Register username which have only integers
- Register username with less than name fee
- Vote delegate with more than balance
- Downvote delegate who is not voted
- Downvote delegate with amount not multiple of 10LSK
- Unlock with an zero amount
- Unlock with zero unlock height
- PoM with generatorPublicKey are not same 
- PoM with identical block headers

PR #5747, opened 2020-09-01
Delegate already registered error for non-delegate accounts - Closes #5734

### LINKED ISSUE #5734: Delegate already registered error for non-delegate accounts
### Expected behavior

If we register a non-delegate account as delegate, it should not fail. 

### Actual behavior

A valid delegate registration transaction failed with error "Account is already a delegate" 

### Steps to reproduce

1. Send some balance to an account
2. Try to register delegate for that account

5.0.0

nazarhussain:
### What was the problem?

This PR resolves #5734

### How was it solved?

Updated the checks for username name validation. 

### How was it tested?

Unit tests will be added as part of  https://github.com/LiskHQ/lisk-sdk/issues/5726


shuse2 (approved):
do we need to fix the test too?

nazarhussain:
@shuse2 I mentioned in the description, will fix the test as part of the other issue. https://github.com/LiskHQ/lisk-sdk/issues/5726

ManuGowda (approved):

PR #5752, opened 2020-09-02
Fix state store scope - Closes #5751

### LINKED ISSUE #5751: Vote transaction does not affect the voter balance
### Expected behavior

After sending a vote transaction, voter balance should reflect the correct balance, with following rules: 

1. Vote transaction fee should be subtracted
2. Amount for every up-vote should be subtracted 

### Actual behavior

- Currently the vote transaction subtract the vote fee but not the up-vote amount 

### Steps to reproduce
- Need to run the QA scenario for voting. 

5.0.0

### Additional Information 

- This is because of the behaviour of the state store, as it override properties of `token` module when doing `store.accont.update` in `dpos` module. 
- So rather than changing implementation of vote transaction asset we should update the state store to only change properties of a particular module in which it was invoked.

shuse2:
### What was the problem?

This PR resolves #5751 

### How was it solved?

- pass scoped state store, which only have access to each module scope, to the module hooks

### How was it tested?

- All functional and integration tests should work as it is
- Add integration test to check the behavior
- FIx unit test


ishantiw (approved):
LGTM
Just a note, if we make all the hook functions as blank functions returning void in `base_modules` instead of abstract then we can avoid checking if it exists every time we call them. The custom modules can override these functions

framework/src/node/processor/processor.ts
```diff
@@ -469,6 +455,44 @@ export class Processor {
 		};
 	}
 
+	// eslint-disable-next-line class-methods-use-this
+	private _createScopedStateStore(stateStore: StateStore, moduleName: string): ModuleStateStore {
+		return {
+			account: {
+				get: async <T>(key: Buffer): Promise<Account<T>> => {
```
nazarhussain, line 462: Not in this PR, but in general should we use more descriptive names for generics. While looking at the code at this particular deep level, we miss the context that this generic type is.

nazarhussain (approved):

shuse2:
@ishantiw I think i was thinking 2 things
- as interface definition for user it was more clear to indicate these functions are optional
- If statement is faster than executing empty function especially it was only doing `if` once when initializing

but since it's changed now, the above doesn't make much sense. Feel free to make an issue and discuss with team =)

PR #5753, opened 2020-09-02
Update get block by height api error catch - Closes #5684

### LINKED ISSUE #5684: Endpoint /api/blocks returns invalid error message
### Expected behavior
```
{"errors":[{"message":"Block with height '10' was not found"}]}
```
### Actual behavior
http://127.0.0.1:4000/api/blocks?height=10
```
{"errors":[{"message":"Specified key blocks:height:\u0000\u0000\u0000\n does not exist"}]}
```
### Steps to reproduce
- Start a node with http api enabled
- Query endpoint `http://127.0.0.1:4000/api/blocks?height=10` before chain is grown beyond height 10

5.0

mitsuaki-u:
### What was the problem?

This PR resolves #5684 

### How was it solved?

Updated the regex match for error catching the endpoint.

### How was it tested?

Run local node and query `/api/blocks?height=10` before chain reaches height 10.


pablitovicente (approved):

ManuGowda (approved):
Test seems to fail, otherwise LGTM

PR #5769, opened 2020-09-04
Fix genesis account sorting - Closes #5768

### LINKED ISSUE #5768: chain library should sort the genesis account considering the length
### Expected behavior
While checking the order of the account, accounts which has shorter address should come in front

### Actual behavior
Currently the genesis block account is validated for order regardless of the account length

### Steps to reproduce
Add uninitialized account in genesis block account and validate

5.0-

shuse2:
### What was the problem?

This PR resolves #5768 

### How was it solved?

Sort with address length also when validating the genesis block account order

### How was it tested?

- Add unit test

### DIFF
--- a/elements/lisk-chain/src/validate.ts
+++ b/elements/lisk-chain/src/validate.ts
@@ -186,10 +186,20 @@ export const validateGenesisBlockHeader = (block: GenesisBlock, accountSchema: S
 	}
 
 	const accountAddresses = header.asset.accounts.map(a => a.address);
+	const copiedAddresses = [...accountAddresses];
+	copiedAddresses.sort((a, b) => {
+		if (a.length > b.length) {
+			return 1;
+		}
+		if (a.length < b.length) {
+			return -1;
+		}
+		return a.compare(b);
+	});
 
-	if (!objects.bufferArrayOrderByLex(accountAddresses)) {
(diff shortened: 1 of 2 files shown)


elements/lisk-chain/src/validate.ts
```diff
@@ -186,10 +186,20 @@ export const validateGenesisBlockHeader = (block: GenesisBlock, accountSchema: S
...
+		if (a.length < b.length) {
+			return -1;
+		}
+		return a.compare(b);
+	});
+
+	if (!objects.bufferArrayEqual(accountAddresses, copiedAddresses)) {
 		errors.push({
-			message: 'should be lexicographically ordered',
+			message: 'should be length and lexicographically ordered',
```
ManuGowda, line 202:
```suggestion
			message: 'should be ordered lexicographically and by length',
```

ManuGowda (approved):

elements/lisk-chain/src/validate.ts
shuse2: I put length first because length takes priority

nazarhussain: Then it should be 

> should be ordered by length and lexicographically

nazarhussain (approved):

PR #5772, opened 2020-09-04
Fix file log to write bigint and buffer as string - Closes #5707

### LINKED ISSUE #5707: Buffer JSON format is written in file log
### Expected behavior
When logging to a file, `Buffer` must be converted to string

### Actual behavior
It is writing in `{ type: 'Buffer',  data: [] }` format

### Steps to reproduce
Look at logs/lisk.log

5.0

shuse2:
### What was the problem?

This PR resolves #5707 

### How was it solved?

- Add custom file stream to parse the log before writing
- Convert Buffer and bigint to string before calling JSON stringify

### How was it tested?

- Run application with file log and observe the file log (all bigint and Buffer should be string)

### DIFF
--- a/framework/src/logger/logger.ts
+++ b/framework/src/logger/logger.ts
@@ -11,7 +11,7 @@
  *
  * Removal or modification of this copyright notice is prohibited.
  */
-
+/* eslint-disable max-classes-per-file */
 import * as path from 'path';
 import * as fs from 'fs';
 import * as bunyan from 'bunyan';
@@ -50,7 +50,7 @@ const levelToName = {
 } as { [key: number]: string };
 
 interface LogObject {
-	[key: string]: string | number | Error | undefined;
+	[key: string]: string | number | Error | undefined | bigint | Buffer;
 }
 
 class ConsoleLog {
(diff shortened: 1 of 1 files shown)


pablitovicente (approved):

framework/src/logger/logger.ts
```diff
@@ -107,6 +107,66 @@ class ConsoleLog {
...
+			level,
+			name,
+			msg,
+			hostname,
+			pid,
+			module: moduleName ?? 'unknown',
+		};
+		const otherKeys = Object.entries(others);
+		if (otherKeys.length > 0) {
+			const meta = otherKeys.reduce<Record<string, unknown>>((prev, [key, value]) => {
```
nazarhussain, line 142: This does not cover all possible cases. 

1. Nested objects with `buffer` or `bigint` values. 
2. Arrays on first level or deep level

I am not sure if such nested objects are not used within SDK. But on a higher perspective we are exposing the logger from the `Application` instance as `app.logger` so SDK user can log anything with. So our logger implementation should be very generic to handle all such cases.

nazarhussain (changes requested):

framework/src/logger/logger.ts
shuse2: I think there is a trade off of how nested that logs needs to go vs performance.
For the console log also it only checks the first level. This is focusing on the fixing the bug that it logs the Buffer JSON on our usage.

If we want to support more nested level, we should create new issue as a feature issue

nazarhussain: https://github.com/LiskHQ/lisk-sdk/issues/5778

nazarhussain (approved):

PR #5781, opened 2020-09-07
Fix PoM not to deduct amount more than minimum remaining balance - Closes #5758

### LINKED ISSUE #5758: PoM transaction for delegate with balance lower than block reward is not accepted
### Expected behavior

As per the specification mentioned on LIP https://github.com/LiskHQ/lips/blob/335a9ee963f65884e2ed4d44aa8ddb2dca823488/proposals/lip-0024.md#reward-of-pom

```
let A be the account of the misbehaving delegate

reward = min { A.balance , regular block reward }

reward is added to the balance of the sender of the PoM
reward is subtracted from A.balance
```

If delegate balance is lower than the block reward, it should transfer full remaining balance to the PoM sender. 

### Actual behavior

Due to contradicting requirement in `token` module any account must satisfy a minimum balance requirement. So any of such PoM transactions transferring full balance is rejected with following error. 

```
{
      errors: [ { message: 'Remaining balance must be greater than 5000000' } ]
}
```

### Steps to reproduce

1. Transfer 100LSK to a genesis delegate account
2. Check the last block reward which would e.g. 5LSK 
3. Transfer all amount in delegate account to some random account and keep only 4.5LSK in delegate account
3. Create a contradicting block for that genesis account
4. Send PoM Transaction for those contradicting blocks 

5.0.0

shuse2:
### What was the problem?

This PR resolves #5758 

### How was it solved?

- Add `getMinRemainingBalance` to the token module
- Use the `getMinRemainingBalance` to calculate subtractable balance from punishing delegate

### How was it tested?

- Add unit test case

### DIFF
--- a/framework/src/modules/dpos/transaction_assets/pom_transaction_asset.ts
+++ b/framework/src/modules/dpos/transaction_assets/pom_transaction_asset.ts
@@ -230,10 +230,19 @@ export class PomTransactionAsset extends BaseAsset<PomTransactionAssetContext> {
 		const delegateAccountBalance = await reducerHandler.invoke<bigint>('token:getBalance', {
 			address: delegateAccount.address,
 		});
+		const minRemainingBalance = await reducerHandler.invoke<bigint>(
+			'token:getMinRemainingBalance',
+			{},
+		);
+
+		const delegateSubtractableBalance =
+			delegateAccountBalance - minRemainingBalance > BigInt(0)
+				? delegateAccountBalance - minRemainingBalance
+				: BigInt(0);
 
 		const reward =
-			store.chain.lastBlockReward > delegateAccountBalance
-				? delegateAccountBalance
+			store.chain.lastBlockReward > delegateSubtractableBalance
+				? delegateSubtractableBalance
 				: store.chain.lastBlockReward;
 
 		await reducerHandler.invoke('token:credit', { address: senderAddress, amount: reward });
--- a/framework/src/modules/dpos/transaction_assets/pom_transaction_asset.ts
+++ b/framework/src/modules/dpos/transaction_assets/pom_transaction_asset.ts
@@ -230,10 +230,7 @@ export class PomTransactionAsset extends BaseAsset<PomTransactionAssetContext> {
 		const delegateAccountBalance = await reducerHandler.invoke<bigint>('token:getBalance', {
 			address: delegateAccount.address,
 		});
-		const minRemainingBalance = await reducerHandler.invoke<bigint>(
-			'token:getMinRemainingBalance',
-			{},
-		);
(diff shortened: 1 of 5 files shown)


framework/src/modules/dpos/transaction_assets/pom_transaction_asset.ts
```diff
@@ -230,10 +230,19 @@ export class PomTransactionAsset extends BaseAsset<PomTransactionAssetContext> {
 		const delegateAccountBalance = await reducerHandler.invoke<bigint>('token:getBalance', {
 			address: delegateAccount.address,
 		});
+		const minRemainingBalance = await reducerHandler.invoke<bigint>(
+			'token:getMinRemainingBalance',
+			{},
```
nazarhussain, line 235: Why its necessary to pass empty object? As we got a use case we should make second argument optional.

nazarhussain (changes requested):

framework/test/unit/modules/dpos/transaction_assets/pom_transaction_asset.spec.ts
```diff
@@ -293,18 +293,60 @@ describe('PomTransactionAsset', () => {
 			);
 		});
 
-		it('should add remaining balance of delegate to balance of the sender if delegate balance is less than last block reward', async () => {
+		it('should reward the sender with last block reward if delegate have emough balance', async () => {
```
mitsuaki-u, line 296:
```suggestion
		it('should reward the sender with last block reward if delegate have enough balance', async () => {
```

nazarhussain (approved):

framework/src/modules/dpos/transaction_assets/pom_transaction_asset.ts
```diff
@@ -230,10 +230,16 @@ export class PomTransactionAsset extends BaseAsset<PomTransactionAssetContext> {
 		const delegateAccountBalance = await reducerHandler.invoke<bigint>('token:getBalance', {
 			address: delegateAccount.address,
 		});
+		const minRemainingBalance = await reducerHandler.invoke<bigint>('token:getMinRemainingBalance');
+
+		const delegateSubtractableBalance =
+			delegateAccountBalance - minRemainingBalance > BigInt(0)
```
pablitovicente, line 236: Perhaps this could be better as a regular if as there's some calculation going on instead of straight selection of values based on input. But is just opinion so ok to keep.

pablitovicente (approved):

framework/src/modules/dpos/transaction_assets/pom_transaction_asset.ts
shuse2: It might be slightly better and it looks good, but ill just keep it as it is since it doesnt change significantly in both way

PR #5782, opened 2020-09-07
Remove special handling for bootstrap period - Closes #5780

### LINKED ISSUE #5780: DPoS should not ban or accept punishment during bootstrap period
### Description
- Add logic in `lisk-chain/setValidators` to not to update the validator during the bootstrap period.
- Remove bootstrap specific logic from dpos afterBlockApply (it should calculate productivity during bootstrap period)

### Expected behavior
- During the bootstrap period, all productivity calculation should happen
- Validator should not be changed during the bootstrap period

### Actual behavior
- DPoS is not calculating the productivity during the bootstrap period
- lisk-chain does not protect the validator change in function

5.0-

shuse2:
### What was the problem?

This PR resolves #5780 

### How was it solved?

- Add logic in setValidator not to update validator during bootstrap period
- Remove special handling of bootstrap period from dpos module

### How was it tested?

- Add unit test case to check if validator is not updated during bootstrap period

### DIFF
--- a/elements/lisk-chain/src/chain.ts
+++ b/elements/lisk-chain/src/chain.ts
@@ -198,13 +198,6 @@ export class Chain {
 		return this._genesisBlock;
 	}
 
-	public get lastBootstrapHeight(): number {
-		return (
-			this._numberOfValidators * this._genesisBlock.header.asset.initRounds +
-			this._genesisBlock.header.height
-		);
-	}
-
 	public get accountSchema(): Schema {
 		return this._accountSchema;
 	}
@@ -433,7 +426,16 @@ export class Chain {
 	public async setValidators(
 		validators: { address: Buffer; isConsensusParticipant: boolean }[],
 		stateStore: StateStore,
+		blockHeader: BlockHeader,
 	): Promise<void> {
+		if (this._getLastBootstrapHeight() > blockHeader.height) {
+			debug(
+				`Skipping updating validator since current height ${
+					blockHeader.height
+				} is lower than last bootstrap height ${this._getLastBootstrapHeight()}`,
+			);
+			return;
+		}
 		const validatorsBuffer = await stateStore.consensus.get(CONSENSUS_STATE_VALIDATORS_KEY);
 		if (!validatorsBuffer) {
 			throw new Error('Previous validator set must exist');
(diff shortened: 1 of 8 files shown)


MaximeGagnebin (approved):
Looks good.
I have to say that this PR is a bit outside my comfort zone for review.

elements/lisk-chain/src/chain.ts
```diff
@@ -433,7 +426,16 @@ export class Chain {
 	public async setValidators(
 		validators: { address: Buffer; isConsensusParticipant: boolean }[],
 		stateStore: StateStore,
+		blockHeader: BlockHeader,
```
nazarhussain, line 429: I believe the chain should be aware o the last block header and validators should be always updated based on last block. So need to pass extra block header param.

framework/src/modules/dpos/dpos_module.ts
```diff
@@ -175,6 +167,23 @@ export class DPoSModule extends BaseModule {
 		}
 
 		await setRegisteredDelegates(context.stateStore, { registeredDelegates: delegateUsernames });
+
+		const initialRound = this.rounds.calcRound(context.genesisBlock.header.height) + 1;
```
nazarhussain, line 171: Suggest to use some name other than `initialRound`, its got mixed up with the `asset.initRounds` property.

nazarhussain (changes requested):

elements/lisk-chain/src/chain.ts
shuse2: This `blockHeader` is not the `lastBlock` that chain holds, but is currently being processed block.
Therefore, it needs to be passed to this function.

ex)
this.lastBlock.height === 2 then blockHeader.height will be 3

nazarhussain: If that's the case always, can't we use that offset ourself in chain? to reduce an extra parameter and give user flexibility to update validators in custom module without need to worry about block header. What do you think?

shuse2: Actually, from the custom module, this parameter is not exposed.
https://github.com/LiskHQ/lisk-sdk/pull/5782/files#diff-b543eea0d99418fe1fb34ff986a2e848R450-R452
Custom module will not need to pass the block header manually

nazarhussain (approved):

PR #5789, opened 2020-09-08
Add missing banning scenarios - Closes #5761

### LINKED ISSUE #5761: Missing peer banning for a few scenarios
### Expected behavior

- `postNodeInfo` with >4 send a status message in a 10 s time window - `Should apply penalty of 10. The number of postNodeInfo messages by an honest peer in a 10 s time window is throttled at 3.`

- `getNodeInfo` with More than one node header message for one connection - `Should apply penalty of 10. There should only be one getNodeInfo message when the connection is established.`

- `getPeers` with More than one message after establishing a connection - Should apply penalty of 10. Honest peers should send a getPeers message only once directly after establishing a connection.

### Actual behavior

Missing banning in the above cases

### Steps to reproduce

Request `getPeers` 2 times after connecting or post nodeInfo 4 times in 10 seconds

`5.0`

ishantiw:
### What was the problem?

This PR resolves #5761

### How was it solved?

- Add a constant and config for peer node status frequency per 10 seconds
- Add RPC requests counter for `getPeers` and `getNodeInfo`, `count > 1 => applyPenalty(10)`
- Add interval and banning for `postNodeInfo` messages > 4 per 10 seconds
- Functional tests to check if a peer is getting banned when sending `postNodeInfo > 4 within 10 secs`
- Functional tests to check if peer is getting banned for requesting more than 10 times getPeers and getNodeInfo RPCs

### How was it tested?

Run tests under `elements/lisk-p2p`
`npm run test:functional`
`npm run test:unit`
`npm run test:integration`


elements/lisk-p2p/src/peer/base.ts
```diff
@@ -174,6 +191,22 @@ export class Peer extends EventEmitter {
 				return;
 			}
 
+			// Apply penalty when you receive getNodeInfo RPC more than once
+			if (rawRequest.procedure === REMOTE_EVENT_RPC_GET_NODE_INFO) {
+				if (this._discoveryMessageCounter.getNodeInfo >= 1) {
```
shuse2, line 196: I think this should compare with the rate (ie until 4th time, it shouldnt apply penalty?
```
				if (this._discoveryMessageCounter.getNodeInfo > this._peerStatusMessageRate) {
```

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -174,6 +191,22 @@ export class Peer extends EventEmitter {
...
+					this.applyPenalty(10);
+				}
+				this._discoveryMessageCounter.getNodeInfo += 1;
+			}
+
+			// Apply penalty when you receive getPeers RPC more than once
+			if (rawRequest.procedure === REMOTE_EVENT_RPC_GET_PEERS_LIST) {
+				if (this._discoveryMessageCounter.getPeers >= 1) {
```
shuse2, line 204: same as above.
Also, addition should come first?

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -230,6 +263,10 @@ export class Peer extends EventEmitter {
 			};
 
 			if (message.event === REMOTE_EVENT_POST_NODE_INFO) {
+				this._peerStatusMessageRate += 1;
```
shuse2, line 266: this should not be incremented?

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -489,6 +527,10 @@ export class Peer extends EventEmitter {
 			this._peerInfo.internalState.productivity = { ...DEFAULT_PRODUCTIVITY };
 		}
 	}
+	private _resetStatusMessageRate(): void {
+		this._peerStatusMessageRate = 0;
```
shuse2, line 531: This should reset the counter?

elements/lisk-p2p/src/peer/base.ts
```diff
@@ -230,6 +263,10 @@ export class Peer extends EventEmitter {
 			};
 
 			if (message.event === REMOTE_EVENT_POST_NODE_INFO) {
+				this._peerStatusMessageRate += 1;
+				if (this._peerStatusMessageRate > 4) {
```
shuse2, line 267: this should check the counter against the `_peerStatusMessageRate`?

shuse2 (changes requested):

elements/lisk-p2p/src/peer/base.ts
ishantiw: _peerStatusMessageRate is for `postNodeInfo` i.e., only for status messages. whereas for RPC requests a peer is only allowed to request nodeInfo or peerlist once when it makes a connection.
`Add RPC requests counter for getPeers and getNodeInfo, count > 1 => applyPenalty(10)`

elements/lisk-p2p/src/peer/base.ts
ishantiw: oh I missed a separate counter for it and incrementing the config constant, I wonder how tests are passing 🤔 . I'll fix it

elements/lisk-p2p/src/peer/base.ts
ishantiw: I'll add a new counter for it

shuse2 (approved):

elements/lisk-p2p/test/functional/message_rate.ts
```diff
@@ -224,4 +225,127 @@ describe('Message rate limit', () => {
...
+		it('should ban a peer when it requests RPC getNodeInfo explicitly for more than 10 times', async () => {
+			// Arrange
+			const attackerNode = p2pNodeList[0];
+			const targetNode = p2pNodeList[1];
+
+			const targetPeerId = `127.0.0.1:${targetNode.config.port}`;
+
+			// Act
+			// With every getNodeInfo RPC request apply penalty of 10, 10 * 10 times will result in banning
+			for (let i = 0; i < 10; i += 1) {
```
mitsuaki-u, line 327: If request is allowed one time without penalty, should this be `i < 11`?

mitsuaki-u (approved):
One minor question but otherwise looks good!

elements/lisk-p2p/test/functional/message_rate.ts
ishantiw: when a peer connects, it asks for status and peer list immediately, so, one request is already done while connecting and when we do 10 other requests it gets banned

PR #5790, opened 2020-09-08
Update dev dependencies - Closes #5449

### LINKED ISSUE #5449: Update dev-dependencies for all the packages
### Description

There are several dev-dependencies that are outdated across packages, and it should be updated.

`yarn audit` on development branch reported 
```
8 vulnerabilities found - Packages audited: 1333
Severity: 8 Low
```

shuse2:
### What was the problem?

This PR resolves #5449 

### How was it solved?

- Update outdated dev dependencies
- Fix linting errors

### How was it tested?


.eslintrc.test.js
```diff
@@ -3,6 +3,8 @@ module.exports = {
 	rules: {
 		'arrow-body-style': 'off',
 		'dot-notation': 'off',
+		'jest/no-conditional-expect': 'off',
+		'jest/valid-title': 'off',
```
pablitovicente, line 7: why are you turning this off?

package.json
```diff
@@ -45,7 +45,7 @@
 		"lint": "lerna run lint",
 		"lint:fix": "lerna run lint:fix",
 		"test": "lerna run test",
-		"test:elements": "lerna run test --ignore=lisk-commander --ignore=lisk-framework --ignore=lisk-sdk --ignore=@liskhq/lisk-framework*",
+		"test:elements": "lerna run test --ignore=lisk-commander --ignore=lisk-framework --ignore=lisk-sdk --ignore=@liskhq/lisk-framework* -- -u",
```
pablitovicente, line 48: Are we sure we want to update snapshot on each run? Maybe it would be better not to in case the need to update snapshots is an unwanted side effect?

pablitovicente (commented):
Just two questions rest LGTM

package.json
shuse2: i forgot to remove 😓

.eslintrc.test.js
shuse2: it's a new rule, but we are looping the test using protocol-spec in a lot of places

pablitovicente (approved):

ishantiw (approved):

PR #5817, opened 2020-09-15
Add banning on calling non existing events - Closes #5801

### LINKED ISSUE #5801: Non existing event from peers are not banned
### Expected behavior
If non existing request or event are received from a peer, it should ban the peer otherwise it will be an open point for an attack.

### Actual behavior
It is ignoring the event, while it is banning the random request

### Steps to reproduce
```
for (let i = 0; i < 100; i += 1) {
	p2pNode.sendToPeer({ event: 'randomEvent', data: { transactionIds: [randomBytes(32).toString('base64')] } }, '127.0.0.1:5000');
}
```

5.0.0-alpha.0

shuse2:
### What was the problem?

This PR resolves #5801 

### How was it solved?

- Add whitelist for event as with the actions.

### How was it tested?

- Add functional test

### DIFF
--- a/framework/src/node/network/network.ts
+++ b/framework/src/node/network/network.ts
@@ -353,7 +353,7 @@ export class Network {
 					const error = new Error(`Sent event "${packet.event}" is not permitted.`);
 					this._logger.error(
 						{ err: error, event: packet.event },
-						'Peer request not fulfilled event: Sent event is not permitted.',
+						'Peer request not fulfilled. Sent event is not permitted.',
 					);
 					// Ban peer on if non-permitted procedure is requested
 					this._p2p.applyPenalty({ peerId: packet.peerId, penalty: 100 });
--- a/framework/src/node/network/network.ts
+++ b/framework/src/node/network/network.ts
@@ -358,7 +358,6 @@ export class Network {
 					// Ban peer on if non-permitted procedure is requested
 					this._p2p.applyPenalty({ peerId: packet.peerId, penalty: 100 });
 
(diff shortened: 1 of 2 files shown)


framework/src/node/network/network.ts
```diff
@@ -347,6 +349,18 @@ export class Network {
 		this._p2p.on(
 			EVENT_MESSAGE_RECEIVED,
 			(packet: { readonly peerId: string; readonly event: string }) => {
+				if (!REMOTE_EVENTS_WHITE_LIST.includes(packet.event)) {
+					const error = new Error(`Sent event "${packet.event}" is not permitted.`);
+					this._logger.error(
+						{ err: error, event: packet.event },
+						'Peer request not fulfilled event: Sent event is not permitted.',
```
mitsuaki-u, line 356:
```suggestion
						'Peer request not fulfilled. Sent event is not permitted.',
```

framework/src/node/network/network.ts
```diff
@@ -347,6 +349,18 @@ export class Network {
...
+					);
+					// Ban peer on if non-permitted procedure is requested
+					this._p2p.applyPenalty({ peerId: packet.peerId, penalty: 100 });
+
+					// Send an error back to the peer.
```
mitsuaki-u, line 361: Comment says so, but it seems to just return?

ishantiw (approved):

mitsuaki-u (approved):
LGTM, minor comment.

PR #5818, opened 2020-09-15
Add hash onion config validation - Closes #5813

### LINKED ISSUE #5813: Hash-onion config should be validated
### Expected behavior
When `foring.delegates` of the config exist, `hashOnion` and the content of `hashOnion` should be required, and `hashes` should be at least length of 1.

### Actual behavior
Currently there is no validation of `hashOnion`

### Steps to reproduce
Add empty hashes to the config and start the application

5.0.0-alpha.0

shuse2:
### What was the problem?

This PR resolves #5813 

### How was it solved?

- Add schema validation for the hash onion in config

### How was it tested?

- Modify config and see if empty hash onion is allowed

### DIFF
--- a/framework/test/unit/application.spec.ts
+++ b/framework/test/unit/application.spec.ts
@@ -166,6 +166,31 @@ describe('Application', () => {
 				"Lisk validator found 1 error[s]:\nMissing property, should have required property 'communityIdentifier'",
 			);
 		});
+
+		it('should throw if invalid forger is provided', () => {
+			// Arrange
+			const invalidConfig = objects.mergeDeep({}, config, {
+				forging: {
+					delegates: [
+						{
+							encryptedPassphrase:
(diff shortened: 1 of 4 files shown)


framework/test/unit/application.spec.ts
```diff
@@ -166,6 +166,31 @@ describe('Application', () => {
...
+								hashes: [],
+							},
+						},
+					],
+				},
+			});
+
+			// Act & Assert
+			expect(() => Application.defaultApplication(genesisBlockJSON, invalidConfig)).toThrow(
+				'Lisk validator found 4 error[s]',
```
nazarhussain, line 191: Please add exception for a particular error. Above 4 errors can be anything else non-related as well.

nazarhussain (changes requested):

nazarhussain (approved):

Incede (commented):
The validator error message needs to be improved in general.

Incede (approved):

PR #5819, opened 2020-09-16
Add module id validation - Closes #5815

### LINKED ISSUE #5815: Validate module id when registration
### Expected behavior
* moduleID less than 2 are forbidden to be registered ( moduleID < 2)
* moduleID less than 1000 are forbidden to be registered for non default modules (moduleID < 1000)

### Actual behavior
Currently there are no validations

### Steps to reproduce
Register module with moduleID less than 1 or custom module with less than 1000

5.0.0-alpha.0

shuse2:
### What was the problem?

This PR resolves #5815 

### How was it solved?

- Add validation for all the module which must be greater than or equal to 2
- Add validation for all external module which must be greater than or equal to 1000

### How was it tested?

- Add unit test

### DIFF
--- a/framework/src/application.ts
+++ b/framework/src/application.ts
@@ -146,10 +146,10 @@ export class Application {
 		config: Partial<ApplicationConfig> = {},
 	): Application {
 		const application = new Application(genesisBlock, config);
-		application._registerDefaultModule(TokenModule);
-		application._registerDefaultModule(SequenceModule);
-		application._registerDefaultModule(KeysModule);
-		application._registerDefaultModule(DPoSModule);
+		application._registerModule(TokenModule);
+		application._registerModule(SequenceModule);
+		application._registerModule(KeysModule);
+		application._registerModule(DPoSModule);
 
 		return application;
 	}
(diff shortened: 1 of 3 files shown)


pablitovicente (approved):

framework/src/application.ts
```diff
@@ -285,6 +289,14 @@ export class Application {
 	// --------------------------------------
 	// Private
 	// --------------------------------------
+
+	private _registerDefaultModule(Module: typeof BaseModule): void {
+		assert(Module, 'Module implementation is required');
+		const InstantiableModule = Module as InstantiableBaseModule;
+		const moduleInstance = new InstantiableModule(this.config.genesisConfig);
+		this._node.registerModule(moduleInstance);
+	}
```
nazarhussain, line 298: Can we just call it `_registerModule` and have extra parameter `validate id: boolean` or `default: boolean` . Then we can use it inside the `registerModule` as well within the class, to reuse the logic. This will help in future if there is nay update in that logic.

nazarhussain (changes requested):

nazarhussain (approved):

PR #5826, opened 2020-09-18
RateLimit counter for getPeers should be reset - Closes #5804 

### LINKED ISSUE #5804: RateLimit counter for getPeers should be reset
### Expected behavior
Even when running a small network, peer should not get banned if not misbehaving.
To achieve that
* the counter for getPeers should be reset every discovery period (30s)
* Returned length of peer list should be lowered to `200` from `1000`

### Actual behavior
When connecting 2 nodes, since one node constantly ask for peers (since it does not have enough outbound peer) from seed node, it will reach threshold of banning after a while.
Every 30s it will increase penalty by 10, so after 300s it will get banned

### Steps to reproduce
Connect 2 local nodes and wait for more than 5 mins to see if non-seed node will get banned

5.0.0-alpha.0

### Additional Information
Resetting the counter every 10s on `peer/base.ts` should mitigate this problem

ishantiw:
### What was the problem?

This PR resolves #5804 

### How was it solved?

- Discovery is triggered with seed peers every 30 secs if outbound slots are not filled. Since `getPeers` and `getNodeInfo` RPC requests are only allowed once after connection it can lead to banning if a node is not able to fulfill its slot and keep on requesting for peers from seed nodes. To avoid banning and keeping getPeer request limited, `getPeers` request will be allowed once every 10 secs, else a peer will get 10 penalty score.

- `DEFAULT_MAX_PEER_DISCOVERY_RESPONSE_LENGTH` setting it from 1000 to 200, lowering the number of peers being shared to lower the payload size as we are allowing multiple getPeer requests. Also, this value is configurable in the library.

### How was it tested?

`npm run test:unit`
`npm run test:integration`
`npm run test:functional`


elements/lisk-p2p/test/unit/peer/base.ts
```diff
@@ -785,6 +785,31 @@ describe('peer/base', () => {
 				);
 			});
 
+			it('should not apply any penalty for second getPeers after 10 secs', () => {
+				// Arrange
+				const rawMessageRCP = {
```
mitsuaki-u, line 790:
```suggestion
				const rawMessageRPC = {
```
same for below

mitsuaki-u (approved):
LGTM one minor comment

shuse2 (approved):

PR #5862, opened 2020-10-08
Reject outgoing connection on receiving unexpected messages 

### LINKED ISSUE #5861: Reject outbound connection if it sends unexpected messages
### Expected behaviour

Unexpected messages should not be allowed from any type of connection

### Actual behaviour

On outbound connection its possible to send control messages which is not required by the protocol.

`5.0.0`

ishantiw:
### What was the problem?

This PR resolves #5861 

### How was it solved?

- Ban the peer on receiving control frames on outgoing connection
- Add tests for outgoing/incoming connections for control frames

### How was it tested?

Connect to or receive connection from a peer and send control frames to see if the peer bans the node

### DIFF
--- a/elements/lisk-p2p/src/peer/outbound.ts
+++ b/elements/lisk-p2p/src/peer/outbound.ts
@@ -128,7 +128,7 @@ export class OutboundPeer extends Peer {
 			connectTimeout,
 			ackTimeout,
 			multiplex: false,
-			autoConnect: false,
+			autoConnect: true,
 			autoReconnect: false,
 			maxPayload: this._peerConfig.wsMaxPayload,
 		};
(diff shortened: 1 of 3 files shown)


elements/lisk-p2p/src/peer/outbound.ts
```diff
@@ -194,6 +194,15 @@ export class OutboundPeer extends Peer {
 		// Bind RPC and remote event handlers
 		outboundSocket.on(REMOTE_SC_EVENT_RPC_REQUEST, this._handleRawRPC);
 		outboundSocket.on(REMOTE_SC_EVENT_MESSAGE, this._handleRawMessage);
+		// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access,@typescript-eslint/no-unsafe-assignment
+		const transportSocket = (outboundSocket as any).transport;
+		// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access
+		if (transportSocket?.socket) {
+			// eslint-disable-next-line @typescript-eslint/no-unsafe-member-access,@typescript-eslint/no-unsafe-call
+			transportSocket.socket.on(REMOTE_EVENT_PING, () => this.applyPenalty(100));
```
shuse2, line 202: dont we need to unbound them?

shuse2 (approved):

mitsuaki-u (approved):

elements/lisk-p2p/src/peer/outbound.ts
ishantiw: `transportSocket.socket` is cleaned up by socketcluster internally so we don't need it

PR #5867, opened 2020-10-13
Include meta property in HTTP api response - Closes #5864

### LINKED ISSUE #5846: Application should start with minimum configuration possible
### Expected behavior

The application should start with minimum possible configuration. Best case scenario by only providing genesis block. e.g. 

```
const app = Application.defaultApplication(genesisBlock)
```

If not we should revisit the configuration which are ultimately required to start the app and can't have a default value. e.g. CommunityIdentifier seems required, but we can skip it by using app label as identifier. 

Similarly for every default module we should use default configuration values if not provided, unless its necessary to get a user specified value. 

- Update the usage on readme

### Actual behavior

Now to start the application there are plenty of configuration which user has to provide by himself. 

5.0.0

### LINKED ISSUE #5864: Include meta property in HTTP api response
### Expected behavior

- Add `meta` info to
   - `api/blocks`
   - `api/accounts`
   - `api/node/info`

### Actual behavior

- `meta` property is not present in 
   - `api/blocks`
   - `api/accounts`
   - `api/node/info`

### Steps to reproduce

http://localhost:4000/api/blocks?height=1

5.0.0

ManuGowda:
### What was the problem?

This PR resolves #5864 

### How was it solved?

- Add `meta` info to response

### How was it tested?

- Updated unit tests

### DIFF
--- a/config/devnet/genesis_block.json
+++ b/config/devnet/genesis_block.json
@@ -118,58 +118,6 @@
 			],
 			"accounts": [
 				{
-					"address": "8789de4316d79d22",
-					"token": {
-						"balance": "12300000000"
-					},
-					"sequence": {
-						"nonce": "0"
-					},
-					"keys": {
-						"mandatoryKeys": [],
-						"optionalKeys": [],
-						"numberOfSignatures": 0
-					},
-					"dpos": {
-						"delegate": {
-							"username": "",
-							"pomHeights": [],
-							"consecutiveMissedBlocks": 0,
-							"lastForgedHeight": 0,
-							"isBanned": false,
-							"totalVotesReceived": "0"
-						},
-						"sentVotes": [],
-						"unlocking": []
-					}
-				},
-				{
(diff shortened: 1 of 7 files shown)


nazarhussain (approved):

config/devnet/genesis_block.json
```diff
@@ -2,11 +2,11 @@
 	"header": {
```
pablitovicente, line 2: Why are there changes to this file if the change is about HTTP responses?

pablitovicente (commented):
Looks good I think though if we're adding the `meta` property to match endpoints that do support pagination shouldn't we also add the children properties to this added `meta` prop? I imagine the idea behind the issue is to simplify handling to consumers of the API if that's the case there will be a good chance that any consumer assuming an always present `meta` property will also look at its children properties.

pablitovicente (approved):

PR #5871, opened 2020-10-14
Update updateForgerStatus to be more safe - Closes #5839

### LINKED ISSUE #5839: Update updateForgerStatus to be more safe
### Description

Update application actions
* getForgingStatus to return corresponding forger info
```
{
	height: number;
	maxHeightPrevoted: number;
	maxHeightPreviouslyForged: number;
}
```
* updateForgingStatus to include properties `overwrite` as optional and `maxHeightPreviouslyForged` and `maxHeightPrevoted`, `height` as required
  - when overwrite is false and maxHeightPreviouslyForged, maxHeightPrevoted, height does not match, it should fail to enable forging
  - when overwrite is false and maxHeightPreviouslyForged, maxHeightPrevoted, height does match, it should enable forging
  - when overwrite is true, it should overwrite the maxHeightPreviouslyForged and maxHeightPrevoted, height to the input value and enable forging
* Update forging plugin to accept `maxHeightPreviouslyForged`, `maxHeightPrevoted`, `height` and `overwrite` parameters

#### Note
Only enable forging if the node is synced up to the tip of the chain where the provided forger info matches below condition. 

`(forgingInfo.heightPrevoted < B.heightPrevoted ) || ( forgingInfo.heightPrevoted == B.heightPrevoted && forgingInfo.height < B.height)`

where `B` is the block currently at the tip of the chain of the node.

### Motivation

Currently when enabling forging, it is enabling without any check on the forger info.
However, this can lead to misbehavior of BFT protocol.

https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#migrating-to-a-new-forging-node-with-data

### Acceptance Criteria

- Add unit test and functional test
  - when overwrite false and not matched maxHeightPreviouslyForged, maxHeightPrevoted, height
  - when overwrite false and matched maxHeightPreviouslyForged, maxHeightPrevoted, height
  - when overwrite false and data does not exist
  - when overwrite true

### Additional Info
- Update core command to support this change

ManuGowda:
### What was the problem?

This PR resolves #5839 

### How was it solved?

- Update endpoints `api/forging` and `api/forging/info`

### How was it tested?

- Added new functional test
- For manual testing 
   - Forger info `http://localhost:4001/api/forging/info`
   - Enable forging using `http://localhost:4001/api/forging`


framework/src/node/forger/forger.ts
```diff
@@ -194,6 +207,14 @@ export class Forger {
 		const account = await this._chainModule.dataAccess.getAccountByAddress(forgerAddress);
 
 		if (forging) {
+			if (force) {
+				previouslyForgedMap.set(forgerAddress, {
+					height: maxHeightPreviouslyForged,
+					maxHeightPrevoted: 0,
+					maxHeightPreviouslyForged,
+				});
+				await savePreviouslyForgedMap(this._db, previouslyForgedMap);
```
shuse2, line 216: maybe log `info` here that the information is updated?

framework/test/unit/node/forger/forger.spec.ts
```diff
@@ -126,12 +126,21 @@ describe('forger', () => {
 
 	describe('Forger', () => {
 		describe('updateForgingStatus', () => {
+			beforeEach(() => {
```
shuse2, line 129: it should have the tests for cases below?
- when force false and not matched maxHeightPreviouslyForged
- when force false and matched maxHeightPreviouslyForged
- when force false and data does not exist
- when force true

shuse2 (changes requested):

framework/src/node/forger/forger.ts
```diff
@@ -191,15 +223,52 @@ export class Forger {
...
+			throw new Error('Failed to enable forging due to missing forger info.');
+		}
+
+		if (isContradictingForgerInfo(previouslyForgedMap, forgerAddress, forgingInput, overwrite)) {
+			const expectedForgerInfo = previouslyForgedMap.get(forgerAddress) ?? {
+				height: 0,
+				maxHeightPreviouslyForged: 0,
+				maxHeightPrevoted: 0,
+			};
+			assert(
```
janhack, line 248: If I understand correctly, instead of throwing the error explicitly I expect the assert to throw the error because of what I check in the if condition is the same as the first condition of the assert? If this is true, then I find it quite convoluted.

ManuGowda: mmm since throwing the error will the conflicting information was too much, so i added assert. but the assertion is only against height, maxHeightPreviouslyForged and maxHeightPrevoted. I can refactor to throw an error if its twisted

framework/src/node/forger/forger.ts
```diff
@@ -191,15 +223,52 @@ export class Forger {
...
+			this._logger.info(`Forging disabled on account: ${forgerAddress.toString('hex')}`);
+			return {
+				address: forgerAddress,
+				forging,
+			};
+		}
+
+		const previouslyForgedMap = await getPreviouslyForgedMap(this._db);
+		const forgingInput = { height, maxHeightPreviouslyForged, maxHeightPrevoted };
+		if (forgerInfoMissing(previouslyForgedMap, forgerAddress, forgingInput, overwrite)) {
```
janhack, line 238: This case and the next if condition all treat the case the we do not overwrite. In this case the previous forger info has to exist and equal the one given as input. I think the code is correct, but it is a bit hard to see that all cases are covered. We also do a few redundant checks. Maybe the structure could be

if not synced:
   throw error
if not overwrite
  check that forger info exists
  check that forger info is equal
else
  set forger info as given in input

janhack: Maybe the check could also be moved back in place to make it a bit easier to read. They are also not re-used anywhere else.

framework/src/node/forger/forger.ts
```diff
@@ -91,6 +90,35 @@ interface CreateBlockInput {
...
 const BLOCK_VERSION = 2;
+const forgerInfoMissing = (
+	previouslyForgedMap: dataStructures.BufferMap<ForgedInfo>,
+	forgerAddress: Buffer,
+	{ height, maxHeightPreviouslyForged, maxHeightPrevoted }: ForgedInfo,
+	overwrite?: boolean,
+) =>
+	!overwrite &&
+	!previouslyForgedMap.has(forgerAddress) &&
+	maxHeightPreviouslyForged !== 0 &&
```
janhack, line 101: I don't understand why we check that the inputs are non-zero here (this check and the following two checks).

framework-plugins/lisk-framework-forger-plugin/src/controllers/forging.ts
```diff
@@ -42,8 +58,14 @@ interface ForgingResponseData {
 
 interface ForgingRequestData extends ForgingResponseData {
 	readonly password: string;
+	readonly height: number;
+	readonly maxHeightPreviouslyForged: number;
+	readonly maxHeightPrevoted: number;
+	readonly overwrite?: boolean;
 }
 
+const isGreaterThanZero = (value: number) => value === null || value === undefined || value < 0;
```
shuse2, line 67: the input should be either `unknown`, `value? number`, or `value: number | null | undefined`?
Also, function name sounds opposite to me

framework-plugins/lisk-framework-forger-plugin/src/controllers/forging.ts
```diff
@@ -57,11 +79,35 @@ export const updateForging = (channel: BaseChannel, db: KVStore) => async (
...
 	if (!isHexString(address)) {
 		res.status(400).send({
-			errors: [{ message: 'The Address parameter should be a hex string.' }],
+			errors: [{ message: 'The address parameter should be a hex string.' }],
+		});
+		return;
+	}
+
+	if (
+		isGreaterThanZero(maxHeightPreviouslyForged) &&
```
shuse2, line 100: shouldnt this be  `||`?

framework-plugins/lisk-framework-forger-plugin/src/controllers/forging.ts
```diff
@@ -89,6 +139,8 @@ export const updateForging = (channel: BaseChannel, db: KVStore) => async (
 				votesReceived,
 				totalReceivedFees: totalReceivedFees.toString(),
 				totalReceivedRewards: totalReceivedRewards.toString(),
+				maxHeightPreviouslyForged,
```
shuse2, line 142: should it return height too?

framework-plugins/lisk-framework-forger-plugin/src/controllers/forging.ts
```diff
@@ -42,8 +58,14 @@ interface ForgingResponseData {
...
 interface ForgingRequestData extends ForgingResponseData {
 	readonly password: string;
+	readonly height: number;
+	readonly maxHeightPreviouslyForged: number;
+	readonly maxHeightPrevoted: number;
+	readonly overwrite?: boolean;
 }
 
+const isGreaterThanZero = (value: number) => value === null || value === undefined || value < 0;
+
```
shuse2, line 68: missing update on getting forger info?

framework/src/node/forger/forger.ts
shuse2: I also dont understand the `assert` here, i think it's better to simply throw

framework/src/node/forger/forger.ts
```diff
@@ -91,6 +90,35 @@ interface CreateBlockInput {
 }
 
 const BLOCK_VERSION = 2;
+const forgerInfoMissing = (
+	previouslyForgedMap: dataStructures.BufferMap<ForgedInfo>,
+	forgerAddress: Buffer,
+	{ height, maxHeightPreviouslyForged, maxHeightPrevoted }: ForgedInfo,
+	overwrite?: boolean,
+) =>
+	!overwrite &&
```
shuse2, line 99: maybe `overwrite` should be checked outside? same as below

shuse2 (changes requested):

framework/src/node/forger/forger.ts
```diff
@@ -191,16 +198,52 @@ export class Forger {
...
+			}
+		}
+
+		previouslyForgedMap.set(forgerAddress, {
+			height,
+			maxHeightPrevoted,
+			maxHeightPreviouslyForged,
+		});
+		await setPreviouslyForgedMap(this._db, previouslyForgedMap);
+		this._logger.info(forgingInput, 'Updated forgerInfo');
```
janhack, line 241: Isn't it a bit misleading if we log this even if `overwrite==False` and the forgerInfo actually stays the same?

In theory, line 235-241 could be in the else branch for the case `overwrite==True`, right?

framework/src/node/forger/forger.ts
```diff
@@ -191,16 +198,52 @@ export class Forger {
...
+		const previouslyForgedMap = await getPreviouslyForgedMap(this._db);
+		const forgingInput = { height, maxHeightPreviouslyForged, maxHeightPrevoted };
+
+		if (!isSyncedWithNetwork(lastBlockHeader, forgingInput)) {
+			throw new Error('Failed to enable forging as the node is not synced to the network.');
+		}
+
+		if (!overwrite) {
+			// check if forger info exists
+			if (!previouslyForgedMap.has(forgerAddress)) {
```
shuse2, line 221: if delegate is forging for a first time, this will always fail. I think originally it was checking if the values are all zeros to see if it's first time or not?

framework/src/node/forger/forger.ts
```diff
@@ -161,17 +164,21 @@ export class Forger {
...
-		const encryptedList = this._config.forging.delegates;
-		const encryptedItem = encryptedList?.find(item => item.address.equals(forgerAddress));
+		const encryptedForgers = this._config.forging.delegates;
+		const encryptedForger = encryptedForgers?.find(item => item.address.equals(forgerAddress));
 
 		let keypair: Keypair;
 		let passphrase: string;
 
-		if (encryptedItem) {
+		if (encryptedForger) {
```
shuse2, line 178: I think checking other way around is more readable.
```
if (!encryptedForger) {
  throw new Error(`Delegate with address: ${forgerAddress.toString('hex')} not found`);
}
```

janhack (approved):
Looks all good and more readable now.

framework/src/node/forger/forger.ts
janhack: So with the current code the node operator has to call `updateForgingStatus` with `overwrite==True` and `height`, `maxHeightPreviouslyForged`, `maxHeightPrevoted` set to `0`.

shuse2: yes, i think it would be better if `overwrite = true` should only be used in the exception case. in the normal scenario, so it should never be specified.

shuse2 (approved):

janhack (approved):

PR #5875, opened 2020-10-16
Update registerModule to validate asset - Closes #5848

### LINKED ISSUE #5484: Peer should not be banned by `getTransactions` endpoint

### LINKED ISSUE #5848: Validate module asset at the time of registration
### Expected behavior

If any module asset is not fully compliant with the interface and not have specified all required attributes then it should fail at the time of registration if not specified `id`, `name`, `apply`, and `schema` properties, as well as check if it extends `BaseAsset` class.

### Actual behavior

If we register any asset which is not complaint to interface it does not throw error. It does not even validate the instance level check. 

```
// This should fail as the assets are not instance instead class references. 
transactionAssets = [MyAsset(), MyAsset2(), PurchaseNFT()];
```

5.0.0

mitsuaki-u:
### What was the problem?

This PR resolves #5848 

### How was it solved?

Add validation logic to `registerModule` since we cannot change interface to asset class. 

### How was it tested?

Added relevant unit tests.

### DIFF
--- a/framework/src/node/node.ts
+++ b/framework/src/node/node.ts
@@ -60,7 +60,7 @@ import {
 } from './processor/processor';
 import { EVENT_SYNCHRONIZER_SYNC_REQUIRED } from './synchronizer/base_synchronizer';
 import { Network } from './network';
-import { BaseModule } from '../modules';
+import { BaseAsset, BaseModule } from '../modules';
 import { Bus } from '../controller/bus';
 
 const forgeInterval = 1000;
@@ -162,7 +162,7 @@ export class Node {
 	public registerModule(customModule: BaseModule): void {
 		const exist = this._registeredModules.find(rm => rm.id === customModule.id);
 		if (exist) {
-			throw new Error(`Custom module with id ${customModule.id} already exists`);
+			throw new Error(`Custom module with id ${customModule.id} already exists.`);
 		}
 
 		if (!customModule.name || !customModule.id) {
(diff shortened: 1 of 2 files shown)


framework/src/node/node.ts
```diff
@@ -172,14 +172,38 @@ export class Node {
...
+			throw new Error(`Custom module must have id greater than ${MINIMUM_MODULE_ID}.`);
 		}
 		if (customModule.accountSchema) {
 			this._registeredAccountSchemas[customModule.name] = {
 				...customModule.accountSchema,
 				fieldNumber: customModule.id,
 			};
 		}
+
+		if (customModule.transactionAssets.length > 0) {
```
shuse2, line 184: I think this `if` is not needed, since when `length === 0` the `for..of` will not execute

shuse2 (commented):
LGTM

pablitovicente (approved):
LGTM. Agree with Shu's comment

shuse2 (approved):

PR #5889, opened 2020-10-19
Add default app config - Closes #5846

### LINKED ISSUE #5846: Application should start with minimum configuration possible

Incede:
### What was the problem?

This PR resolves #5846 

### How was it solved?

* Initialize default config parameters in DPos & Token modules and in `application_config_schema.ts`

### How was it tested?

* Add unit test
* Test manually like `app = Application.defaultApplication(genesisBlock)` by linking to local `lisk-sdk/sdk`


framework/test/unit/application.spec.ts
```diff
@@ -58,6 +58,12 @@ describe('Application', () => {
 	});
 
 	describe('#constructor', () => {
+		it('should be able to start the application with default parameters if config is not provided', () => {
+			const app = Application.defaultApplication(genesisBlockJSON);
+
+			expect(app.config).toBeDefined();
```
shuse2, line 64: i think it would be nice to check if the optional configs (like minRemainingBalance) is set

framework/test/unit/application.spec.ts
```diff
@@ -151,7 +157,7 @@ describe('Application', () => {
 			// Assert
 			expect(app.logger).toBeUndefined();
 		});
-
+		/*
```
shuse2, line 160: we can delete the test if no longer valid, same for below

framework/src/modules/token/token_module.ts
```diff
@@ -94,7 +94,9 @@ export class TokenModule extends BaseModule {
 
 	public constructor(genesisConfig: GenesisConfig) {
 		super(genesisConfig);
-		const { minRemainingBalance } = this.config;
+		const minRemainingBalance = this.config.minRemainingBalance
+			? this.config.minRemainingBalance
+			: '5000000';
```
shuse2, line 99: prefer to define constant

shuse2 (changes requested):

shuse2 (approved):

mitsuaki-u (approved):
LGTM

PR #5891, opened 2020-10-19
ReadError: Database is not open - Closes #5851

### LINKED ISSUE #5851: ReadError: Database is not open
### Expected behavior

In case of any error in loading or execution of any logic, the application must cleanup and exit successfully. 

### Actual behavior

If we have any issue in the plugin loading logic then application crashes with above error. 

### Steps to reproduce

```
class MyPlugin extends BasePlugin {
    myAttribute = undefined. 

    // all required interfaces and attributes 


   async load(channel) {
       channel.subscribe('app:ready', () => { 
           this.myAttribute = options.myAttribute; // options does not exists. 
       });
   }
}
```

5.0.0

nazarhussain:
### What was the problem?

This PR resolves #5851

### How was it solved?

1. Changed the order of initialization of application 
2. Now plugins are loaded at very end when Node is ready 
3. Started the transaction pool synchronously inside `node.init`
3. Started forging synchronously inside `node.init`
4. Added the graceful shutdown to application 
5. Application would wait before shutdown if its still initializing 

### How was it tested?

Add/Updated unit tests.


framework/src/application.ts
```diff
@@ -230,6 +232,11 @@ export class Application {
 		this._nodeDB = this._getDBInstance(this.config, 'node.db');
 
 		// Initialize all objects
+		this._loadingProcess = new jobHandlers.Defer<boolean>(
```
shuse2, line 235: instead of using defer, u can use `mutex` from the util library

shuse2 (changes requested):

framework/src/application.ts
```diff
@@ -230,6 +232,11 @@ export class Application {
...
+			RUN_WAIT_TIMEOUT,
+			`Application could not started in ${RUN_WAIT_TIMEOUT}ms`,
```
pablitovicente, line 237:
```suggestion
			`Application could not be started in ${RUN_WAIT_TIMEOUT}ms`,
```

framework/src/application.ts
```diff
@@ -255,6 +264,11 @@ export class Application {
 	public async shutdown(errorCode = 0, message = ''): Promise<void> {
 		this.logger.info({ errorCode, message }, 'Application shutdown started');
 
+		// Wait if the loading process still in progress
+		if (this._loadingProcess) {
```
pablitovicente, line 268: it's better to use verbs/actions to make the code more readable though probably this is going to look better once it is replaced by the Jobs as suggested by Shu.

framework/src/application.ts
```diff
@@ -39,6 +39,7 @@ import {
 } from './types';
 import { BaseModule, TokenModule, SequenceModule, KeysModule, DPoSModule } from './modules';
 
+const RUN_WAIT_TIMEOUT = 3000;
```
pablitovicente, line 42: Maybe better to name this `BOOTSTRAP_TIMEOUT` as it conveys more meaning

framework/src/application.ts
```diff
@@ -229,6 +233,21 @@ export class Application {
 		this._blockchainDB = this._getDBInstance(this.config, 'blockchain.db');
 		this._nodeDB = this._getDBInstance(this.config, 'node.db');
 
+		// Acquire mutex for loading process
+		this._runMutexReleaseHandler = await this._mutex.acquire();
```
shuse2, line 237: i think no need to assign it to the instance variable here.
also, it's probably safer to run runExecutive all below code so that if something failed it will release immediately and run the cleanup.
Also, in that way, no need to think about the timeout. (ie: depending on the genesis block size, the node init might vary on time

elements/lisk-utils/src/job_handlers/mutex.ts
```diff
@@ -40,6 +40,29 @@ export class Mutex {
...
+		worker: () => Promise<T>,
+		timeout: number,
+		timeoutMessage?: string,
+	): Promise<T> {
+		const release = await this.acquire();
+
+		const timeoutPromise = new Promise<T>((_, reject) => {
+			setTimeout(() => {
+				if (this.isLocked()) {
+					release();
```
shuse2, line 53: it is possible that this mutex is locked from other acquire, but release from the closure it locked

framework/src/application.ts
```diff
@@ -233,52 +232,39 @@ export class Application {
...
-		this._runMutexReleaseHandler();
-
-		await this._controller.loadPlugins(this._plugins, this.config.plugins);
-		this.logger.debug(this._controller.bus.getEvents(), 'Application listening to events');
-		this.logger.debug(this._controller.bus.getActions(), 'Application ready for actions');
-
-		this._channel.publish('app:ready');
+				this._channel.publish('app:ready');
+			},
+			RUN_TIMEOUT,
```
shuse2, line 259: as we don't know the initalization time required for  mainnet, or other network, i think it's safer to not to include the timeout here.
Also, even if it fails to start, it shouldnt be a problem if killed

elements/lisk-utils/src/job_handlers/mutex.ts
```diff
@@ -40,6 +40,29 @@ export class Mutex {
...
+					reject(Error(timeoutMessage ?? 'Mutex run exclusive timeout'));
+				}
+			}, timeout);
+		});
+
+		try {
+			return await Promise.race([worker(), timeoutPromise]);
```
shuse2, line 60: probably my mistake, but if timeoutPromise resolve first, that does not mean worker is stopped.
Therefore, it actually defeats the purpose of the mutex here

pablitovicente (approved):

shuse2 (approved):

PR #5895, opened 2020-10-20
Fix maxHeightPrevoted to be initialized with the genesis height - Closes #5893

### LINKED ISSUE #5893: BFT should initialize maxHeightPrevoted to genesis block height
### Expected behavior
`maxHeightPrevoted` should be initialized with genesis block height, which might not be 0

### Actual behavior
Currently it is initializing with `0`

5.0.0-alpha.2

shuse2:
### What was the problem?

This PR resolves #5893 

### How was it solved?

- Update to use finzlied height when prevote does not exist. When prevote does not exist, finalizedHeight should be the height at the genesis block. Therefore, indirectly it is setting to the genesis block height

### How was it tested?

- Add unit test to see if it will be initialized

### DIFF
--- a/elements/lisk-bft/src/finality_manager.ts
+++ b/elements/lisk-bft/src/finality_manager.ts
@@ -66,17 +66,17 @@ export const BFTVotingLedgerSchema = {
 			fieldNumber: 2,
 			items: {
 				type: 'object',
-				required: ['height', 'preVotes', 'preCommits'],
+				required: ['height', 'prevotes', 'precommits'],
 				properties: {
 					height: {
 						dataType: 'uint32',
 						fieldNumber: 1,
 					},
-					preVotes: {
+					prevotes: {
 						dataType: 'uint32',
 						fieldNumber: 2,
 					},
-					preCommits: {
+					precommits: {
 						dataType: 'uint32',
 						fieldNumber: 3,
 					},
@@ -96,12 +96,12 @@ interface ValidatorsState {
 
 interface LedgerState {
 	height: number;
-	preVotes: number;
-	preCommits: number;
+	prevotes: number;
+	precommits: number;
 }
 
 interface LedgerMap {
-	[key: string]: { preVotes: number; preCommits: number };
+	[key: string]: { prevotes: number; precommits: number };
 }
 
 interface ValidatorState {
@@ -174,7 +174,7 @@ export class FinalityManager extends EventEmitter {
 		await this.verifyBlockHeaders(blockHeader, stateStore);
 
 		// Update the pre-votes and pre-commits
-		await this.updatePreVotesPreCommits(blockHeader, stateStore, lastBlockHeaders);
+		await this.updatePrevotesPrecommits(blockHeader, stateStore, lastBlockHeaders);
 
 		// Update the pre-voted confirmed and finalized height
 		await this.updateFinalizedHeight(stateStore);
@@ -186,12 +186,12 @@ export class FinalityManager extends EventEmitter {
 		return this;
 	}
 
-	public async updatePreVotesPreCommits(
+	public async updatePrevotesPrecommits(
 		header: BlockHeader,
 		stateStore: StateStore,
 		bftBlockHeaders: ReadonlyArray<BlockHeader>,
 	): Promise<boolean> {
-		debug('updatePreVotesPreCommits invoked');
+		debug('updatePrevotesPrecommits invoked');
 		// If validator forged a block with higher or same height previously
 		// That means he is forging on other chain and we don't count any
 		// Pre-votes and pre-commits from him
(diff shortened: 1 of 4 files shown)


elements/lisk-bft/src/finality_manager.ts
```diff
@@ -411,27 +411,17 @@ export class FinalityManager extends EventEmitter {
...
-		const preVoted = highestHeightPreVoted ? parseInt(highestHeightPreVoted, 10) : 0;
-
-		const highestHeightPreCommitted = Object.keys(ledger)
-			.reverse()
-			.find(key => ledger[key].preCommits >= this.preCommitThreshold);
-
-		const preCommitted = highestHeightPreCommitted ? parseInt(highestHeightPreCommitted, 10) : 0;
-
-		return { preVoted, preCommitted };
+		return highestHeightPreVoted ? parseInt(highestHeightPreVoted, 10) : this.finalizedHeight;
```
ManuGowda, line 424:
```suggestion
		return chainMaxHeightPreVoted ? parseInt(chainMaxHeightPreVoted, 10) : this.finalizedHeight;
```

elements/lisk-bft/src/finality_manager.ts
```diff
@@ -411,27 +411,17 @@ export class FinalityManager extends EventEmitter {
...
-
-		return preVoted;
+		return this._calculateMaxHeightPrevoted(ledger);
 	}
 
-	private _getChainMaxHeightStatus(ledger: LedgerMap): { preVoted: number; preCommitted: number } {
+	private _calculateMaxHeightPrevoted(ledger: LedgerMap): number {
 		debug('updatePreVotedAndFinalizedHeight invoked');
 
 		const highestHeightPreVoted = Object.keys(ledger)
```
ManuGowda, line 420:
```suggestion
		const chainMaxHeightPreVoted = Object.keys(ledger)
```

elements/lisk-bft/src/finality_manager.ts
shuse2: Updated to `maxHeightPrevoted` since thats the name we use everywhere else

elements/lisk-bft/src/finality_manager.ts
```diff
@@ -411,27 +411,17 @@ export class FinalityManager extends EventEmitter {
...
-		const preVoted = highestHeightPreVoted ? parseInt(highestHeightPreVoted, 10) : 0;
-
-		const highestHeightPreCommitted = Object.keys(ledger)
-			.reverse()
-			.find(key => ledger[key].preCommits >= this.preCommitThreshold);
-
-		const preCommitted = highestHeightPreCommitted ? parseInt(highestHeightPreCommitted, 10) : 0;
-
-		return { preVoted, preCommitted };
+		return maxHeightPreVoted ? parseInt(maxHeightPreVoted, 10) : this.finalizedHeight;
```
janhack, line 424: I don't fully understand the code:
- Why is `maxHeightPreVoted` a string and needs to be pared and not a number?
- Do I return this.finalizedHeight if maxHeightPreVoted is undefined?
- The `finalizedHeight` is part of the state and initialized with the height of the genesis block, right? The `maxHeightPrevoted` is not part of the state and always computed from scratch?
- In the prototype implementation `maxHeightPrevoted` together with `finalizedHeight` is maintained as part of the state. The update works as follows: Go over the last 303 block headers and checking and if for any of them the prevotes/precommits and update if the height is larger. If you only store the `finalizedHeight` explicitly, you will have to go over all block headers from `finalizedHeight` onwards (this number can be arbitrarily large, e.g., 10,000). Does this happen here? 
-

janhack: Cosmetic remark: The camelCase is a bit inconsistent, sometimes prevote and sometimes preVote. Note that the LIP uses prevote and precommit as one work in lower case.

janhack: Another remark: Why are there two methods `getMaxHeightPrevoted` and `_calculateMaxHeightPrevoted`? The second method is called directly as well, but before the call it seems the same things happens as in  `getMaxHeightPrevoted`.

shuse2: > Why is maxHeightPreVoted a string and needs to be parsed and not a number?

Node.js object key is always a string even though we assign a number. Therefore, it needs to be parsed.

> Do I return this.finalizedHeight only if maxHeightPreVoted is undefined? Does this happen only if there is no block which reaches the prevoteThreshold?

yes. maxHeightPrevoted is undefined only if there is no block height which is above prevoteThreshold.

> The finalizedHeight is part of the state and initialized with the height of the genesis block, right? The maxHeightPrevoted is not part of the state and always computed from scratch?

> In the prototype implementation maxHeightPrevoted together with finalizedHeight is maintained as part of the state. The update works as follows: Go over the last 303 block headers, check if for any of them the prevotes/precommits threshold is reach and update if the height is larger. If you only store the finalizedHeight explicitly, you will have to go over all block headers from finalizedHeight onwards (this number can be arbitrarily large, e.g., 10,000). Does this happen here?

prevotes/precommits are also stored as state until finalized height. it's called `ledger` in the code. the structure is like
```
{
  ${height}: {
    preVotes: number,
    preCommits: number,
  }
}
```
so moving back and forward is managed along with account state.

shuse2: `_calculateMaxHeightPrevoted` is used in 2 different places, so that's why it is extracted to private method

janhack (approved):
Looks good.

ManuGowda (approved):

PR #5897, opened 2020-10-20
Update to have headers contradicting check as pure function - Closes #5894 

### LINKED ISSUE #5894: Refactor BFT to expose a function to check contradicting block header
### Description

This contradiction check should be extracted to a pure function and exposed from the library.

https://github.com/LiskHQ/lisk-sdk/blob/development/elements/lisk-bft/src/finality_manager.ts#L368-L404

Also, update `reportDelegateMisbehavior` to use the function extracted above.

### Motivation

The contradiction check will be used in plugin, and already used in the reportDelegateMisbehavior asset.

### Acceptance Criteria

All test should remain the same, and it should not break existing tests.

shuse2:
### What was the problem? 

This PR resolves #5894 

### How was it solved?

- Refactor to have pure function to check contradicting block headers
- Update to use the pure function in validation and in `reportDelegateMisbehavior` asset

### How was it tested?

- All test should pass and behavior except error type should remain

### DIFF
--- a/elements/lisk-bft/src/finality_manager.ts
+++ b/elements/lisk-bft/src/finality_manager.ts
@@ -20,12 +20,8 @@ import * as Debug from 'debug';
 import { EventEmitter } from 'events';
 import { dataStructures } from '@liskhq/lisk-utils';
 import { BFT_ROUND_THRESHOLD } from './constant';
-import {
-	BFTChainDisjointError,
-	BFTForkChoiceRuleError,
-	BFTInvalidAttributeError,
-	BFTLowerChainBranchError,
-} from './types';
+import { BFTInvalidAttributeError, BFTError } from './types';
+import { areHeadersContradicting } from './header_contradicting';
 
 // eslint-disable-next-line new-cap
 const debug = Debug('lisk:bft:consensus_manager');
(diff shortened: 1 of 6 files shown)


elements/lisk-bft/src/finality_manager.ts
```diff
@@ -365,42 +361,8 @@ export class FinalityManager extends EventEmitter {
...
-		}
-
-		if (earlierBlock.height > laterBlock.asset.maxHeightPreviouslyForged) {
-			throw new BFTChainDisjointError();
-		}
-
-		if (earlierBlock.asset.maxHeightPrevoted > laterBlock.asset.maxHeightPrevoted) {
-			throw new BFTLowerChainBranchError();
+		if (areHeadersContradicting(validatorLastBlock, blockHeader)) {
+			throw new BFTError();
```
nazarhussain, line 365: Seems good, just one concern here. Having particular errors e.g. `BFTChainDisjointError`, `BFTForkChoiceRuleError` were useful in debugging. If we can still keep those and throw those particular errors from within `areHeadersContradicting` that would nice.

nazarhussain (changes requested):

framework/src/modules/dpos/transaction_assets/pom_transaction_asset.ts
```diff
@@ -109,34 +110,8 @@ export class PomTransactionAsset extends BaseAsset<PomTransactionAssetContext> {
...
-			b2 = asset.header1;
-		}
-
-		if (
-			!(b1.asset.maxHeightPrevoted === b2.asset.maxHeightPrevoted && b1.height >= b2.height) &&
-			!(b1.height > b2.asset.maxHeightPreviouslyForged) &&
-			!(b1.asset.maxHeightPrevoted > b2.asset.maxHeightPrevoted)
-		) {
+		// Check for BFT violations:
+		if (!areHeadersContradicting(asset.header1, asset.header2)) {
```
mitsuaki-u, line 114: do we first need to check if the blockheaders are ordered correctly?

shuse2: ordering is taken care in the function https://github.com/LiskHQ/lisk-sdk/pull/5897/files#diff-9b023404122bb142e9cfbdaeb19057ed4d81383f2f6d253d0a78f2b78722fa9cR18-R34

elements/lisk-bft/src/finality_manager.ts
shuse2: this function will be used in both contradicting and non contradicting case, so it would be nice to have boolean check IMO.
I cannot think of a way to keep the boolean return value and not throwing error in a nice way here

mitsuaki-u (approved):

nazarhussain (approved):

PR #5906, opened 2020-10-22
Add a check of contradicting blocks after receiving each - Closes #5878 

### LINKED ISSUE #5878: Add functionality to find conflicting block for new blocks in PoM plugin
### Description

For every new block detected in the PoM Plugin:

1. Check if it already exists in db.
2. Create pure function to find conflicting blockheaders from BFT 
3. Decode the blockheader returned from event and get the generatorPublicKey.
4. Search the last 260,000 blockheaders generated by this generatorPublicKey on all chains, forks included for the conflicting blockheader.

### Motivation

In order to submit a PoM transaction, the conflicting blockheaders `b1` and `b2` must be located. 

### Acceptance Criteria

- Plugin must detect conflicting blockheaders stored 
- Write relevant tests

shuse2:
### What was the problem?

This PR resolves #5878 

### How was it solved?

- Execute block contradiction check when new block is saved
- return contradicting block in DB if exist

### How was it tested?

- Add integration tests


pablitovicente (approved):

framework-plugins/lisk-framework-report-misbehavior-plugin/src/db.ts
```diff
@@ -83,3 +102,33 @@ export const saveBlockHeaders = async (
...
+		});
+		let conflictingHeader: BlockHeader | undefined;
+		stream
+			.on('data', ({ value }: { value: Buffer }) => {
+				const { blockHeaders } = codec.decode<BlockHeaders>(blockHeadersSchema, value);
+				for (const encodedHeader of blockHeaders) {
+					const decodedBlockHeader = decodeBlockHeader(encodedHeader, schemas);
+					// If there are multiple contradicting blocks, it will report the one with highest height
+					if (areHeadersContradicting(blockHeader, decodedBlockHeader)) {
+						conflictingHeader = decodedBlockHeader;
```
nazarhussain, line 124: Should not we close the stream here? Why to parse the tons of blocks if we already found on conflicting?

nazarhussain (changes requested):

framework-plugins/lisk-framework-report-misbehavior-plugin/src/db.ts
shuse2: yes, i wanted to close the stream here, but there are no interface to close in the middle

nazarhussain: The combination of these two may help. 

https://nodejs.org/api/stream.html#stream_readable_pause
https://nodejs.org/api/stream.html#stream_readable_destroy_error

So you pause the stream that will immediately stop emitting new `data` events and then you can destroy that will eventually release all resources and close.

nazarhussain (approved):

PR #5908, opened 2020-10-23
GET /api/stats/transactions propagation endpoint - Closes #5884

### LINKED ISSUE #5884: GET  /api/stats/transactions propagation endpoint
### Description

Transaction propagation `(api/stats/transactions)`

- Subscribe to `app:network:event` event and check for `postTransactionsAnnouncement` event, increment for every transactionIds `[transactionIds] +=1`, in `{ 'id1': <count of id1 received>, 'id2': <count of id2 received> ... }`. It will always take the timeReceived of the last transaction id received.
- Job to clean up trx older than 10 mins. We can trigger this job on receiving a **new** `transactionID`, we can check for all the transactionIDs whose timeReceived is greater than 10 mins and delete it.

**Response:**

```typescript
{ // Delete all the trxIds older than 10 minute
    transactions: { <transactionId>: {count: <count of ids>, timeReceived: Time },
    averageReceivedTransactions: totalCount/numOfTransactionIds, // moving average
    connectedPeers: channel.invoke(app:getConnectedPeers).count,
}
```

### Motivation

The monitor plugin requires this endpoint to expose how many times a transaction is received on each node.

### Acceptance Criteria

- Add unit test to check if the above values are exposed
- Check if the above values are responded by registering to the framework

### Additional Information

Related issues: #5216

pablitovicente:
### What was the problem?

This PR resolves #5884

### How was it solved?

- Added event subscription method
- Added subscription handler for transactions announcement
- Added controllers folder and transaction stats controller
- Added shared state cleanup

### How was it tested?

- Unit tested added


framework-plugins/lisk-framework-monitor-plugin/src/controllers/transactions.ts
```diff
@@ -0,0 +1,57 @@
...
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+import { Request, Response, NextFunction } from 'express';
+import { BaseChannel } from 'lisk-framework';
+import { SharedState } from '../types';
+
+interface PeerInfo {
```
mitsuaki-u, line 18: maybe move this interface to a `types` file since it is also be used in blocks endpoint

pablitovicente: moved and updated usages.

mitsuaki-u (approved):

ishantiw (approved):

PR #5912, opened 2020-10-26
Loading a plugin as child process failed which is not exported from an npm package - Closes #5888

### LINKED ISSUE #5888: Loading a plugin as child process failed which is not exported from an npm package
Currently off-chain architecture logic for loading plugin as child process is based on assumption that the plugin itself be exported from an npm package. Which implies that `info.name` property of the plugin should be a valid npm package name. 

### Expected behavior

Based on the approach we take there can be three expected behaviours:

1. The plugin which is required to load as child plugin is not exported from an npm package, it should show a warming and fallback to loading it as in-memory plugin. 
2. At the time of registration of the plugin check if the plugin name is a valid npm package, otherwise fail with the error 
3. Develop some new loader which somehow make it possible to load any plugin as child process. 

### Actual behavior

Currently at the time of starting of the app if plugin is not exported from an npm package then it shows following error in logs, unload/crash the plugin loader and keep running the application. 

```
13:19:22 INFO myPlugin: Loading child-process plugin (module=lisk:app)
{
 "version": "1.0.0",
 "alias": "myPlugin"
} 
Error: Cannot find module 'myPlugin'
Require stack:
- /lisk-sdk/framework/dist-node/controller/child_process_loader.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:831:15)
    at Function.Module._load (internal/modules/cjs/loader.js:687:27)
    at Module.require (internal/modules/cjs/loader.js:903:19)
    at require (internal/modules/cjs/helpers.js:74:18)
    at Object.<anonymous> (/lisk-sdk/framework/dist-node/controller/child_process_loader.js:7:15)
    at Module._compile (internal/modules/cjs/loader.js:1015:30)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1035:10)
    at Module.load (internal/modules/cjs/loader.js:879:32)
    at Function.Module._load (internal/modules/cjs/loader.js:724:14)
    at Function.executeUserEntryPoint [as runMain] (internal/modules/run_main.js:60:12) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/lisk-sdk/framework/dist-node/controller/child_process_loader.js'
  ]
}
13:19:22 ERROR myPlugin: Child process plugin exited (module=lisk:app)
{
 "version": "1.0.0",
 "pluginAlias": "myPlugin",
 "code": 1,
 "signal": ""
} 
13:19:22 INFO lisk-framework: New block added to the chain (module=lisk:app)
{
 "id": "07bb4baa9f379a02cf011e5267d41ec4dcf999ba22ba937528c28d156b0d16a5",
 "height": 16,
 "numberOfTransactions": 0
} 
```

### Steps to reproduce

You can use following script to reproduce the error. 

```js
const {
	Application,
	genesisBlockDevnet,
	configDevnet,
	HTTPAPIPlugin,
	ForgerPlugin,
	BasePlugin,
} = require('./sdk/dist-node');

const app = Application.defaultApplication(genesisBlockDevnet, {
	...configDevnet,
	ipc: { enabled: true },
});

app.registerPlugin(HTTPAPIPlugin, {loadAsChildProcess: true});
app.registerPlugin(ForgerPlugin, {loadAsChildProcess: true});

class MyPlugin extends BasePlugin {
	static alias = 'myPlugin';
	static info = {
		name: 'myPlugin',
		author: 'nazar',
		version: '1.0.0',
	};
	events = [];
	actions = {};

	// all required interfaces and attributes

	async load(channel) {
		channel.subscribe('app:ready', () => {
			console.log('my app is ready....');
		});
	}

	async unload() {}
}

app.registerPlugin(MyPlugin, { loadAsChildProcess: true });

app
	.run()
	.then(() => console.info('App is running'))
	.catch(console.error);

```

5.0.0

nazarhussain:
### What was the problem?

This PR resolves #5888

### How was it solved?

- Used the fail early strategy to validate the plugin info
- To load the plugin as child process, one must follow any of two conditions. 
- 1. Plugin must be exported from an npm package and name can be defined as `info.name`
- 2. Plugin must be exported from an export path defined by `info.exportPath`

### How was it tested?

- Added unit tests


framework/src/plugins/base_plugin.ts
```diff
@@ -252,3 +254,61 @@ export abstract class BasePlugin {
...
+	strict = true,
+): string | undefined => {
+	let nodeModule: Record<string, unknown> | undefined;
+	let nodeModulePath: string | undefined;
+
+	try {
+		// Check if plugin name is an npm package
+		// eslint-disable-next-line global-require, import/no-dynamic-require, @typescript-eslint/no-var-requires, @typescript-eslint/no-unsafe-assignment
+		nodeModule = require(pluginKlass.info.name);
+		nodeModulePath = pluginKlass.info.name;
```
shuse2, line 271: in this case, it should be already returned?

framework/src/plugins/base_plugin.ts
```diff
@@ -252,3 +254,61 @@ export abstract class BasePlugin {
...
+	} catch (error) {
+		/* Plugin info.name is not an npm package */
+	}
+
+	if (!nodeModule && pluginKlass.info.exportPath) {
+		try {
+			// Check if plugin name is an npm package
+			// eslint-disable-next-line global-require, import/no-dynamic-require, @typescript-eslint/no-var-requires, @typescript-eslint/no-unsafe-assignment
+			nodeModule = require(pluginKlass.info.exportPath);
+			nodeModulePath = pluginKlass.info.exportPath;
```
shuse2, line 281: same as above, and below condition should be simplified by that

framework/src/application.ts
```diff
@@ -156,27 +162,34 @@ export class Application {
...
+		pluginKlass: InstantiablePlugin<BasePlugin>,
+		options: PluginOptions = { loadAsChildProcess: false },
 	): void {
 		assert(pluginKlass, 'Plugin implementation is required');
 		assert(typeof options === 'object', 'Plugin options must be provided or set to empty object.');
-		assert(alias ?? pluginKlass.alias, 'Plugin alias must be provided.');
-		const pluginAlias = alias ?? pluginKlass.alias;
+		validatePluginSpec(pluginKlass);
+
+		const pluginAlias = options?.alias ?? pluginKlass.alias;
```
shuse2, line 172: why `alias` needs to be overwritten?

shuse2 (changes requested):

framework/src/plugins/base_plugin.ts
nazarhussain: This just check if the name is a valid package name, the other condition wether the Plugin is exported from this package or not is tested below.

framework/src/plugins/base_plugin.ts
nazarhussain: Same as above.

framework/src/application.ts
nazarhussain: It was already there to register a plugin with different alias. Say you imported a plugin and want to use it with own prefix alias to avoid any future conflict.

shuse2 (approved):

ishantiw (approved):

PR #5916, opened 2020-10-27
Create PoM transaction with conflicting blockheaders in PoM plugin - Closes #5879

### LINKED ISSUE #5879: Create PoM transaction with conflicting blockheaders in PoM plugin
### Description

1. Create the proof-of-misbehavior transaction with the incoming blockheader from the event as b1 and the retrieved conflicting blockheader from storage as b2. (Refer to DPoS module PoM Asset)

2. Sign transaction

3. Send the created transaction to network through channel, and log the result. 

### Motivation

In order to submit a PoM transaction and secure the network, it must be created. 

### Acceptance Criteria

- Created transaction must be valid
- Write relevant tests

ManuGowda:
### What was the problem?

This PR resolves #5879

### How was it solved?

- Added a function to create PoM transaction
- Send PoM transaction using channel

### How was it tested?

- Added unit tests


framework-plugins/lisk-framework-report-misbehavior-plugin/src/report_misbehavior_plugin.ts
```diff
@@ -43,7 +48,7 @@ import * as controllers from './controllers';
 
 // eslint-disable-next-line
 const packageJSON = require('../package.json');
-const logger = debug('plugin:resport-misbehavior');
+const logger = debug('plugin:report-misbehavior');
```
nazarhussain, line 51: Logger makes it confusing with the file logger. Once see it in the code may look for the relevant logs in the console or log file. Better to rename it to `debug` which is convention we used other places as well.

nazarhussain (approved):

mitsuaki-u (changes requested):
LGTM

mitsuaki-u (approved):

PR #5930, opened 2020-10-28
Add utility function calculateMinFee to lisk-transactions library - Closes #5673

### LINKED ISSUE #5673: Add utility functions on lisk-transactions
### Description

- Add additional utility functions to lisk-transactions for UI development (Discussion required)

### Motivation

Currently, lisk-transactions has minimal util functions that can be used in any transactions.

### Acceptance Criteria

TBD

ManuGowda:
### What was the problem?

This PR resolves #5673

### How was it solved?

- compute min fee from transaction without fee and signatures
- return minFee for given transaction object

### How was it tested?

- Added unit tests


elements/lisk-transactions/src/fee.ts
```diff
@@ -0,0 +1,67 @@
...
+	return BigInt(size * (options?.minFeePerByte ?? DEFAULT_MIN_FEE_PER_BYTE)) + BigInt(baseFee);
+};
+
+export const getMinFee = (
+	assetSchema: object,
+	trx: Record<string, unknown>,
+	options?: Options,
+): bigint => {
+	// eslint-disable-next-line no-param-reassign
+	trx.fee = BigInt(0);
```
shuse2, line 58: I think it would be nice that if it's a function that return `minFee`, it doesn't mutate the input

elements/lisk-transactions/test/fee.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+
+import { getAddressAndPublicKeyFromPassphrase } from '@liskhq/lisk-cryptography';
+import * as Fee from '../src/fee';
+
+const validAssetSchema = {
```
shuse2, line 19: all variable should be within `describe`

elements/lisk-transactions/test/fee.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+
+import { getAddressAndPublicKeyFromPassphrase } from '@liskhq/lisk-cryptography';
+import * as Fee from '../src/fee';
```
shuse2, line 17: `Fee` is not constructor, so prefer to use camel case

elements/lisk-transactions/test/fee.spec.ts
```diff
@@ -0,0 +1,173 @@
...
+	{
+		moduleID: 3,
+		assetID: 0,
+		baseFee: '1',
+	},
+];
+
+describe('fee', () => {
+	beforeEach(() => {
+		jest.spyOn(Fee, 'calculateMinFee');
```
shuse2, line 76: since it pure function, no need to spy?

shuse2 (changes requested):

elements/lisk-transactions/test/fee.spec.ts
ManuGowda: not necessarily? https://jestjs.io/docs/en/api#describename-fn, and we do use mix of both approach in SDK, so we can keep it this way

shuse2 (approved):

janhack (approved):
Looks good to me.

The only two minor comments: 
- The names of both functions could be improved so that it is more clear what they do and what the difference is.
- I don't understand why `calculateMinFee` is exported and in which cases you would want to use that function somewhere ellse. If I already have a transaction with a fee, then I would only want to check if that transaction satisfies the minimum fee requirement. If I have a transaction without fee, then I would want to use the `getMinFee` function.

PR #5931, opened 2020-10-29
Add getUnlockings action on dpos module - Closes #5843

### LINKED ISSUE #5843: Add getUnlockings action on dpos module
### Description

- add an action below to return all unlocking objects with additional information `minUnlockHeight`, which shows the minimum height when the object can be unlocked.
```
getUnlockings({ address: string }) => {
  delegateAddress: string;
  amount: string;
  unvoteHeight: number;
  minUnlockHeight: number;
}[]
```

### Motivation

- User needs to know when the locked token can be unlocked with the `unlockToken` transaction

### Acceptance Criteria

- Add unit test to check the `minUnlockHeight` result for cases
 1. self vote
 2. vote delegate
 3. self vote with punishment
 4. vote delegate who is punished

nazarhussain:
### What was the problem?

This PR resolves #5843

### How was it solved?

- Updated BaesModule._dataAccess to have new interface getAccountByAddress
- Updated BaesModule._dataAccess to have new interface getLastBlockHeader
- Deprecated BaesModule._dataAccess.getAccount
- Add getUnlockings actions to dpos module 

### How was it tested?

- Added relevant unit tests

### DIFF
--- a/framework/src/modules/base_module.ts
+++ b/framework/src/modules/base_module.ts
@@ -26,7 +26,7 @@ import {
 } from '../types';
 import { BaseAsset } from './base_asset';
 
-export interface BaseModuleChannel {
+interface BaseModuleChannel {
 	publish(name: string, data?: object): void;
 }
 
--- a/framework/src/modules/base_module.ts
+++ b/framework/src/modules/base_module.ts
@@ -26,7 +26,7 @@ import {
 } from '../types';
 import { BaseAsset } from './base_asset';
 
-interface Channel {
+export interface BaseModuleChannel {
 	publish(name: string, data?: object): void;
 }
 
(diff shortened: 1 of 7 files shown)


framework/src/modules/base_module.ts
```diff
@@ -26,7 +26,7 @@ import {
 } from '../types';
 import { BaseAsset } from './base_asset';
 
-interface Channel {
+export interface BaseModuleChannel {
```
shuse2, line 29: why exporting this type and changing the name here?

shuse2 (changes requested):

framework/src/modules/base_module.ts
nazarhussain: It was needed to reference in a typescript file where I was creating some custom module. If you think we can remove the export. 

For name it was confusing when compared along with the second parameter of `BaseModule.init`

shuse2: yea let's remove `export` if not needed

ishantiw (approved):

PR #5936, opened 2020-10-30
Create job to periodically clean blockHeaders in PoM plugin - Closes #5881 

### LINKED ISSUE #5881: Create periodic job to clean up old blockheaders in PoM plugin
### Description

For the PoM plugin, we require a job running every X time units, which will delete the last X blocks with db.delete(). 

The time interval should be configurable, with a default value set. 

The optimal default time units for the job should be tested and set in this issue. 

### Motivation

In storage we will have to iterate through every `generatorPublicKey` which could potentially be many operations. There a periodic job is required to clean up old blocks. 

### Acceptance Criteria

- Define default time interval for job
- Check resource usage of job while running plugin
- Write relevant tests.

mitsuaki-u:
### What was the problem?

This PR resolves #5881 

### How was it solved?

- Implement clearBlockHeaders functionality
- Add job to clear old blockheaders in PoM
- Add integration test for cleaning blockheaders in PoM

### How was it tested?

Added integration test and checked contents of db after job operations.


shuse2 (approved):

framework-plugins/lisk-framework-report-misbehavior-plugin/src/report_misbehavior_plugin.ts
```diff
@@ -177,6 +187,10 @@ export class ReportMisbehaviorPlugin extends BasePlugin {
 							transaction: encodedTransaction,
 						});
 
+						// Set new currentHeight
+						if (decodedBlockHeader.height > this._state.currentHeight) {
```
ManuGowda, line 191: so this will be executed only when `contradictingBlock` is true.. a scenario where upto x million blocks, no contradicting blocks found the `this._state.currentHeight` will be zero? so the `clearBlockHeaders` job will not clear anything?
- from height 0 to 3million blocks not contradiction found so, `this._state.currentHeight` will be zero?
- when `this._clearBlockHeadersInterval` time is triggered the `this._state.currentHeight` will be zero so nothing happens?

framework-plugins/lisk-framework-report-misbehavior-plugin/test/integration/cleanup_job.spec.ts
```diff
@@ -0,0 +1,130 @@
...
+			publicKey: getAddressAndPublicKeyFromPassphrase(defaultAccount.passphrase).publicKey,
+			currentHeight: 1000000000,
+		};
+	});
+
+	afterEach(async () => {
+		await reportMisbehaviorPlugin.unload();
+	});
+
+	it('should clear old blockheaders', async () => {
```
ManuGowda, line 117: can we have a test where `_pluginDB.get` is called to clear the blocks?

mitsuaki-u: It is already tested that ` _pluginDB.get` in the expect below.

ManuGowda (approved):

PR #5948, opened 2020-11-04
Update actions/events communication format to JSON-RPC - Closes #5918

### LINKED ISSUE #5918: Update actions/events communication format to JSON-RPC
### Description

Update communication schema of actions and events to be in the format of JSON-RPC v2 referencing in `action.ts` and `event.ts`

### Motivation

Enable message communication to be transport agnostic by introducing JSON-RPC.
JSON-RPC node communication is a standard in the industry, and as actions and events are going to be exposed for API client, it should follow the standard.

Why update only the action and events format to JSONRPC?
Why the new APIClients IPC/WS does not respond in JSONRPC format?

### Acceptance Criteria

- `invoke`, `subscribe` function interface should not be changed.

### Additional Information
Related issue #5919 
- Reference https://www.jsonrpc.org/specification

ManuGowda:
### What was the problem?

This PR resolves #5918

### How was it solved?

- Update action, events, bus, channels to
  - requests jsonrpc
  - respond jsonrpc
  - subscribe to events for jsonrpc notifications

### How was it tested?

- Updated tests


framework/src/controller/jsonrpc/index.ts
```diff
@@ -0,0 +1,106 @@
+/*
+ * Copyright © 2020 Lisk Foundation
```
shuse2, line 2: this file should be called `types`?

framework/src/controller/action.ts
```diff
@@ -32,56 +33,74 @@ export interface ActionsObject {
...
+		this.method = method;
+		this.params = params ?? {};
 		this.handler = handler;
 	}
 
-	public static deserialize(data: ActionInfoObject | string): Action {
-		const parsedAction: ActionInfoObject =
-			typeof data === 'string' ? (JSON.parse(data) as ActionInfoObject) : data;
+	public static fromJSONRPC(data: JSONRPC.RequestObject | string): Action {
+		// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
```
shuse2, line 69: why this is disabled?

framework/src/controller/action.ts
```diff
@@ -32,56 +33,72 @@ export interface ActionsObject {
...
+	public id: ID;
+	public method: string;
 	public params: object;
+	public source?: string;
+	public handler?: (action: ActionInfoObject) => unknown;
 
 	public constructor(
-		name: string,
+		id: ID,
+		method: string,
```
nazarhussain, line 48: I don't think changing `name` to `method` or removing `module` attribute is appropriate here. In framework we still call and should call it `action name` and `action.module`. The change should be translated in `toJsonRPC` function where ``method: `${module}:${name}` ``

framework/src/controller/action.ts
```diff
@@ -32,56 +33,72 @@ export interface ActionsObject {
...
 		);
-		[this.module, this.name] = name.split(':');
-		this.params = params ?? {};
-
 		if (source) {
 			assert(moduleNameReg.test(source), `Source name "${source}" must be a valid module name.`);
 			this.source = source;
 		}
 
+		this.id = id;
```
nazarhussain, line 62: This attribute seems required but not asserted neither any default value is assigned.

framework/src/controller/action.ts
```diff
@@ -32,56 +33,72 @@ export interface ActionsObject {
...
-		return `${this.source ?? 'undefined'} -> ${this.module}:${this.name}`;
+	public get module(): string {
+		const [moduleName] = this.method.split(':');
+		return moduleName;
+	}
+
+	public get name(): string {
+		const [, eventName] = this.method.split(':');
+
+		return eventName;
```
nazarhussain, line 101: I don't see why these are required here. If its a needed, you can add those in initializer and assign to private variables.

framework/src/controller/action.ts
```diff
@@ -32,56 +33,72 @@ export interface ActionsObject {
...
+		this.id = id;
+		this.method = method;
+		this.params = params ?? {};
 		this.handler = handler;
 	}
 
-	public static deserialize(data: ActionInfoObject | string): Action {
-		const parsedAction: ActionInfoObject =
-			typeof data === 'string' ? (JSON.parse(data) as ActionInfoObject) : data;
+	public static fromJSONRPC(data: RequestObject | string): Action {
```
nazarhussain, line 68: I would suggest for `fromJsonRPC`

nazarhussain (changes requested):

nazarhussain (approved):

shuse2 (approved):

PR #5956, opened 2020-11-06
Update bus and channel models to only support JSON-RPC format on bus level - Partially Closes #5921

### LINKED ISSUE #5921: Integrate WS server with controller and enable channel communication
### Description

- Add support for WS server to expose all actions and events in controller folder.

### Motivation

WS server must be integrated seamlessly to the channel in controller folder, so that all actions and events registered are accessible through websocket 

### Acceptance Criteria

- Add test to check if module actions and events are accessible from the ws server
- Add test to check if plugin actions and events are accessible from the ws server
- Add test to check if framework actions and events are accesible from ws server

### Additional Information
Related issue #5919

nazarhussain:
### What was the problem?

This PR partially resolves #5921

### How was it solved?

- Updated the bus to only support the JSON-RPC format on bus level 
- Updated actions and events to adjust as per bus changes  

### How was it tested?

- Updated relevant integration tests
- Updated relevant unit tests

### DIFF
--- a/framework/src/controller/action.ts
+++ b/framework/src/controller/action.ts
@@ -15,7 +15,7 @@
 import { strict as assert } from 'assert';
 import { actionWithModuleNameReg, moduleNameReg } from '../constants';
 import { RequestObject, VERSION } from './jsonrpc';
-import { ResponseObject, JsonRpcResult, ID, JsonRpcError } from './jsonrpc/types';
+import { ResponseObject, JSONRPCResult, ID, JSONRPCError } from './jsonrpc/types';
 
 export interface ActionInfoObject {
 	readonly module: string;
@@ -86,11 +86,11 @@ export class Action {
 		};
 	}
 
-	public buildJSONRPCResponse<T = JsonRpcResult>({
+	public buildJSONRPCResponse<T = JSONRPCResult>({
 		error,
 		result,
 	}: {
-		error?: JsonRpcError;
+		error?: JSONRPCError;
 		result?: T;
 	}): ResponseObject<T> {
 		if (error) {
--- a/framework/src/controller/action.ts
+++ b/framework/src/controller/action.ts
@@ -15,12 +15,13 @@
 import { strict as assert } from 'assert';
 import { actionWithModuleNameReg, moduleNameReg } from '../constants';
 import { RequestObject, VERSION } from './jsonrpc';
+import { ResponseObject, JsonRpcResult, ID, JsonRpcError } from './jsonrpc/types';
 
 export interface ActionInfoObject {
 	readonly module: string;
 	readonly name: string;
 	readonly source?: string;
-	readonly params: object;
+	readonly params?: object;
 }
 
 export type ActionHandler = (action: ActionInfoObject) => unknown;
@@ -33,28 +34,24 @@ export interface ActionsObject {
 	[key: string]: Action;
 }
 
-type ID = string | number | null;
-
 export class Action {
-	public jsonrpc = VERSION;
-	public id: ID;
-	public method: string;
-	public params: object;
-	public module: string;
-	public name: string;
-	public source?: string;
+	public readonly id: ID;
+	public readonly module: string;
+	public readonly name: string;
+	public readonly source?: string;
+	public readonly params?: object;
 	public handler?: (action: ActionInfoObject) => unknown;
 
 	public constructor(
 		id: ID,
-		method: string,
+		name: string,
 		params?: object,
 		source?: string,
 		handler?: (action: ActionInfoObject) => unknown,
 	) {
 		assert(
-			actionWithModuleNameReg.test(method),
-			`Action method "${method}" must be a valid method with module name and action name.`,
+			actionWithModuleNameReg.test(name),
+			`Action name "${name}" must be a valid name with module name and action name.`,
 		);
 		if (source) {
 			assert(moduleNameReg.test(source), `Source name "${source}" must be a valid module name.`);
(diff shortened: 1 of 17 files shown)


framework/src/controller/bus.ts
```diff
@@ -225,18 +236,29 @@ export class Bus {
...
 	}
 
-	public publish(eventName: string, eventValue?: object): void {
+	public publish(eventData: string | JSONRPC.NotificationRequest): void {
+		const parsedEvent = Event.fromJSONRPCNotification(eventData);
+
+		try {
+			JSONRPC.validateJSONRPCRequest(parsedEvent.toJSONRPCNotification() as never);
+		} catch (error) {
+			throw JSONRPC.errorResponse(null, JSONRPC.invalidRequest());
```
shuse2, line 245: i just realized that this is not `Error` instance. when it throws it should be always error instance.

shuse2 (changes requested):

framework/src/controller/bus.ts
nazarhussain: I missed it as it was not highlighted by eslint.

framework/src/controller/action.ts
```diff
@@ -15,12 +15,13 @@
 import { strict as assert } from 'assert';
 import { actionWithModuleNameReg, moduleNameReg } from '../constants';
 import { RequestObject, VERSION } from './jsonrpc';
+import { ResponseObject, JsonRpcResult, ID, JsonRpcError } from './jsonrpc/types';
```
ManuGowda: you could import from above definition.
`import { ID, JsonRpcError, RequestObject, ResponseObject, JsonRpcResult, VERSION } from './jsonrpc';`

framework/src/controller/jsonrpc/types.ts
```diff
@@ -13,30 +13,37 @@
...
-	jsonrpc: string;
-	id: ID;
-	result: Result;
-}
-export interface NotificationObject {
-	jsonrpc: string;
-	method: string;
-	result?: Result;
-}
+export type JsonRpcResult = string | number | boolean | object;
```
ManuGowda:
```suggestion
export type JSONRPCResult = string | number | boolean | object;
```

framework/src/controller/bus.ts
```diff
@@ -17,13 +17,14 @@ import { PubSocket, PullSocket, PushSocket, ReqSocket, SubSocket } from 'pm2-axo
...
 import { Action, ActionsObject } from './action';
+import { Event, EventsArray } from './event';
 import * as JSONRPC from './jsonrpc';
 import { Logger } from '../logger';
 import { BaseChannel } from './channels/base_channel';
-import { EventInfoObject, EventsArray } from './event';
 import { IPCServer } from './ipc/ipc_server';
 import { ActionInfoForBus, SocketPaths } from '../types';
 import { WSServer } from './ws/ws_server';
+import { ResponseObjectWithError, ResponseObjectWithResult } from './jsonrpc';
```
ManuGowda, line 27: you could use the `JSONRPC. ResponseObjectWithError` ?

framework/src/controller/bus.ts
```diff
@@ -195,26 +191,45 @@ export class Bus {
...
+	public async invoke<T>(
+		actionData: string | JSONRPC.RequestObject,
+	): Promise<JSONRPC.ResponseObjectWithResult<T>> {
+		const parsedAction = Action.fromJSONRPCRequest(actionData);
+
+		try {
+			JSONRPC.validateJSONRPCRequest(parsedAction.toJSONRPCRequest() as never);
+		} catch (error) {
+			this.logger.error(
+				JSONRPC.errorResponse(parsedAction.id, JSONRPC.invalidRequest()),
```
ManuGowda, line 203: the `JSONRPC.invalidRequest` already convey its invalid request. so maybe the data can be about the validation error which tells about what property of json rpc request is incorrect?

framework/src/controller/bus.ts
```diff
@@ -225,18 +240,33 @@ export class Bus {
...
-	public publish(eventName: string, eventValue?: object): void {
+	public publish(eventData: string | JSONRPC.NotificationRequest): void {
+		const parsedEvent = Event.fromJSONRPCNotification(eventData);
+
+		try {
+			JSONRPC.validateJSONRPCRequest(parsedEvent.toJSONRPCNotification() as never);
+		} catch (error) {
+			this.logger.error(
+				JSONRPC.errorResponse(null, JSONRPC.invalidRequest()),
+				'Invalid publish request',
```
ManuGowda, line 251: same here

framework/src/controller/controller.ts
```diff
@@ -285,7 +285,7 @@ export class Controller {
 			new Promise((_, reject) => {
 				setTimeout(() => {
 					reject(new Error('Child process plugin loading timeout'));
-				}, 2000);
+				}, 3000);
```
ManuGowda, line 288: why did we change this value? could add comment if there is specific reason

shuse2 (approved):

framework/src/controller/jsonrpc/types.ts
```diff
@@ -13,30 +13,37 @@
...
+	data?: JsonRpcResult;
 }
+
 export interface RequestObject {
-	readonly id: string | number | null;
+	readonly id: ID;
 	readonly jsonrpc: string;
 	readonly method: string;
-	readonly params?: object;
+	readonly params?: object & { source?: string };
```
ManuGowda, line 28: why does json rpc request has source? maybe if you want to reuse we can extend the type and omit or add?

framework/src/controller/jsonrpc/utils.ts
```diff
@@ -13,14 +13,21 @@
...
+	ResponseObjectWithResult,
+	ResponseObjectWithError,
+} from './types';
 
 export const VERSION = '2.0';
 
 const RequestSchema = {
 	id: 'jsonRPCRequestSchema',
 	type: 'object',
-	required: ['jsonrpc', 'method', 'id'],
```
ManuGowda, line 23: i think for a valid jsonrpc request the property id is required.. but the value can be string, number, null?
```
id
An identifier established by the Client that **MUST contain a String, Number, or NULL** value if included. If it is not included it is assumed to be a notification. The value SHOULD normally not be Null [1] and Numbers SHOULD NOT contain fractional parts [2]
```

framework/src/controller/bus.ts
nazarhussain: The error handling is improved in upcoming PR.

framework/src/controller/bus.ts
nazarhussain: The error handling is improved in upcoming PR.

framework/src/controller/jsonrpc/types.ts
nazarhussain: When we build action from request or build request from action, no information should be lost. That's why had to add this extra params source in the params.

framework/src/controller/jsonrpc/utils.ts
nazarhussain: There is difference between request and notification. In next PR both had been segregated into separate schemas and this point is covered.

framework/src/controller/jsonrpc/types.ts
ManuGowda: the request can be json rpc or ipc/in-memory action, so what i meant was source was not required for ipc/in-memory action. we could have two separate types. since its optional you are allowing it to coexists.

ManuGowda (approved):

framework/src/controller/jsonrpc/types.ts
nazarhussain: Request are always `json-rpc` no matter invoked from IPC or In-Memory. Bus only communicate in `JSON-RPC` format.

PR #5962, opened 2020-11-09
Implement transaction namespace for APIClient - Closes #5924

### LINKED ISSUE #5924: Implement transaction namespace for APIClient
### Description

Add functions for APIClient with name space transaction as below
```
{
	get: (id: Buffer) => Promise<Record<string, unknown>>;
	getFromPool: () => Promise<Record<string, unknown>[]>;
	create: (
		input: {
			moduleID?: number; // id takes priority
			moduleName?: number;
			assetID?: number; // id takes priority
			assetName?: number;
			fee: bigint;
			nonce: bigint;
			senderPublicKey: Buffer;
			asset: Record<string, unknown>;
			signatures?: Buffer[];
		},
		passphrase: string,
		options?: {
			nonce?: bigint;
			includeSenderSignature?: boolean;
			multisignatureKeys?: {
				mandatoryKeys: Buffer[];
				optionalKeys: Buffer[];
			};
		},
	) => Promise<Record<string, unknown>>;
	sign: (transaction: Record<string, unknown>, passphrases: string[], options?: {
		includeSenderSignature?: boolean;
		multisignatureKeys?: {
			mandatoryKeys: Buffer[];
			optionalKeys: Buffer[];
		};
	}) => Promise<Record<string, unknown>>;
	send: (input: Record<string, unknown>) => Promise<void>;
	calculateMinFee: (tx: Record<string, unknown>) => bigint;
	encode: (input: Record<string, unknown>) => Buffer;
	decode: (input: Buffer) => Record<string, unknown>;
}
```

### Motivation

- Add utility functions to support basic operation for transactions

### Acceptance Criteria

- All data should be JavaScript Object format
- Add unit tests

### Additional Information
Related issue #5919
- Reference https://github.com/LiskHQ/lisk-sdk/blob/api_client_idea/elements/lisk-api-client/src/transaction.ts

Incede:
### What was the problem?

This PR resolves #5924 

### How was it solved?

* Define namespace & utility functions for transactions in API client
* Initialize transaction namespace in API client
* Add logic to decode & encode account for API client 

### How was it tested?

* Add unit tests


elements/lisk-api-client/src/codec.ts
```diff
@@ -46,7 +51,7 @@ export const decodeTransaction = (
 	}>(registeredSchema.transaction, encodedTransaction);
 	const assetSchema = getTransactionAssetSchema(transaction, registeredSchema);
 	const asset = codec.decode(assetSchema, transaction.asset);
-	const id = hash(encodedTransaction);
+	const id = hash(encodedTransaction).toString('hex');
```
shuse2, line 54: id should be `buffer` in the JS object format

elements/lisk-api-client/src/transaction.ts
```diff
@@ -0,0 +1,178 @@
...
+		const networkIdentifier = Buffer.from(this._nodeInfo.networkIdentifier, 'hex');
+		const { publicKey, address } = getAddressAndPublicKeyFromPassphrase(passphrase);
+		const accountHex = await this._channel.invoke<string>('app:getAccount', {
+			address: address.toString('hex'),
+		});
+		const account = decodeAccount(Buffer.from(accountHex, 'hex'), this._schema);
+		if (!txInput.moduleID) {
+			if (!txInput.moduleName) {
+				throw new Error('Missing moduleId and moduleName');
+			} else {
```
shuse2, line 88: else clause is not required, since it's thrown. Same for below

elements/lisk-api-client/src/transaction.ts
```diff
@@ -0,0 +1,178 @@
...
+				const registeredModule = this._nodeInfo.registeredModules.find(
+					module => module.name === input.moduleName,
+				);
+				txInput.moduleID = registeredModule?.id ? registeredModule.id : txInput.moduleID;
+			}
+		}
+		if (txInput.assetID === undefined || null) {
```
shuse2, line 95:
```suggestion
		if (txInput.assetID === undefined || txInput.assetID === null) {
```
or u could check `typeof txInput.assetID !== 'number'`

elements/lisk-api-client/src/transaction.ts
```diff
@@ -0,0 +1,178 @@
...
+
+	public async get(id: Buffer): Promise<Record<string, unknown>> {
+		const transactionHex = await this._channel.invoke<string>('app:getTransactionByID', {
+			id: id.toString('hex'),
+		});
+		return decodeTransaction(Buffer.from(transactionHex, 'hex'), this._schema);
+	}
+
+	public async getFromPool(): Promise<Record<string, unknown>[]> {
+		return this._channel.invoke('app:getTransactionsFromPool');
```
shuse2, line 149: I think this also returns encoded transactions. return type is `string[]`

elements/lisk-api-client/src/transaction.ts
```diff
@@ -0,0 +1,178 @@
...
+	}
+
+	public sign(
+		transaction: Record<string, unknown>,
+		passphrases: string[],
+	): Record<string, unknown> {
+		const assetSchema = getTransactionAssetSchema(transaction, this._schema);
+		const networkIdentifier = Buffer.from(this._nodeInfo.networkIdentifier, 'hex');
+		return signTransaction(assetSchema, transaction, networkIdentifier, passphrases[0]);
```
shuse2, line 158: this should also check if the sender is multi-signature account, or optionally keys are provided, and use `signMultiSignatureTransaction` for each passphrase or `signTransaction` with the first element

elements/lisk-api-client/src/transaction.ts
```diff
@@ -0,0 +1,178 @@
...
+
+	public decode(transaction: Buffer): Record<string, unknown> {
+		return decodeTransaction(transaction, this._schema);
+	}
+
+	public encode(transaction: Record<string, unknown>): Buffer {
+		return encodeTransaction(transaction, this._schema);
+	}
+
+	public getMinFee(transaction: Record<string, unknown>): bigint {
```
shuse2, line 174:
```suggestion
	public computeMinFee(transaction: Record<string, unknown>): bigint {
```

shuse2 (changes requested):

elements/lisk-api-client/src/transaction.ts
```diff
@@ -0,0 +1,178 @@
...
+				);
+				txInput.assetID = registeredAsset?.id ? registeredAsset.id : txInput.assetID;
+			}
+		}
+		if (!options?.nonce && !txInput.nonce) {
+			if (
+				typeof account.sequence !== 'object' ||
+				!(account.sequence as Record<string, unknown>).nonce
+			) {
+				throw new Error('Unspported account type.');
```
ishantiw, line 110:
```suggestion
				throw new Error('Unsupported account type.');
```

elements/lisk-api-client/test/unit/transaction.spec.ts
```diff
@@ -0,0 +1,716 @@
...
+		fee: BigInt('10000000'),
+		senderPublicKey: publicKey1,
+		asset: {
+			recipientAddress: Buffer.from('3a971fd02b4a07fc20aad1936d3cb1d263b96e0f', 'hex'),
+			amount: BigInt('4008489300000000'),
+			data: '',
+		},
+	};
+	const txId = Buffer.from(tx.id, 'hex');
+	const schema = {
```
ishantiw, line 139: can we move schema to utils folder as its making the setup part too big in this test file? maybe also nodeInfo and some other variables that are using huge space and can be reused for future tests?

elements/lisk-api-client/src/transaction.ts
```diff
@@ -0,0 +1,178 @@
...
+import { getAddressAndPublicKeyFromPassphrase } from '@liskhq/lisk-cryptography';
+import {
+	decodeTransaction,
+	encodeTransaction,
+	getTransactionAssetSchema,
+	decodeAccount,
+} from './codec';
+import { Channel, RegisteredSchemas, NodeInfo } from './types';
+
+export interface CreateTransactionOptions {
```
ishantiw, line 30: if we export it then we can move it to `types.ts`?

elements/lisk-api-client/src/transaction.ts
```diff
@@ -0,0 +1,178 @@
...
+	nonce?: bigint;
+	includeSenderSignature?: boolean;
+	multisignatureKeys?: {
+		mandatoryKeys: Buffer[];
+		optionalKeys: Buffer[];
+	};
+}
+
+export interface MultiSignatureKeys {
```
ishantiw, line 39: same here, if we export it then we can move it to `types.ts`?

elements/lisk-api-client/test/unit/transaction.spec.ts
Incede: yep, what should I add them under in utils ?

elements/lisk-api-client/src/transaction.ts
Incede: but it's only being used in `transaction.ts` so it shouldn't go to `types.ts` imo

ishantiw: we are keeping all the exported types in `types.ts`, if not used elsewhere we can simply remove export keyword

elements/lisk-api-client/test/unit/transaction.spec.ts
ishantiw: for ex, you can create `test/utils/transactions.ts` and add `export const transactionSchema = {...`, same goes for nodeInfo and other variables if they are too big and generic to be in the test

elements/lisk-api-client/src/transaction.ts
```diff
@@ -0,0 +1,220 @@
...
+	): Promise<Record<string, unknown>> {
+		const assetSchema = getTransactionAssetSchema(transaction, this._schema);
+		const networkIdentifier = Buffer.from(this._nodeInfo.networkIdentifier, 'hex');
+		const address = getAddressFromPublicKey(transaction.senderPublicKey as Buffer);
+		const accountHex = await this._channel.invoke<string>('app:getAccount', {
+			address: address.toString('hex'),
+		});
+		const account = decodeAccount(Buffer.from(accountHex, 'hex'), this._schema);
+		if (account.keys && (account.keys as MultiSignatureKeys).numberOfSignatures > 0) {
+			passphrases.forEach(passphrase =>
```
shuse2, line 177: prefer `for...of`.
Also it does mutate the `signatures` right?

elements/lisk-api-client/src/transaction.ts
```diff
@@ -0,0 +1,220 @@
...
+					transaction,
+					networkIdentifier,
+					passphrase,
+					account.keys as MultiSignatureKeys,
+					options?.includeSenderSignature,
+				),
+			);
+		}
+		if (options?.multisignatureKeys && options?.includeSenderSignature) {
+			passphrases.forEach(passphrase =>
```
shuse2, line 189: same as above

elements/lisk-api-client/src/transaction.ts
```diff
@@ -0,0 +1,220 @@
...
+				signMultiSignatureTransaction(
+					assetSchema,
+					transaction,
+					networkIdentifier,
+					passphrase,
+					account.keys as MultiSignatureKeys,
+					options?.includeSenderSignature,
+				),
+			);
```
shuse2, line 186: it should return at this point, otherwise, it will sign again in below

elements/lisk-api-client/src/transaction.ts
```diff
@@ -0,0 +1,220 @@
...
+			passphrases.forEach(passphrase =>
+				signMultiSignatureTransaction(
+					assetSchema,
+					transaction,
+					networkIdentifier,
+					passphrase,
+					options.multisignatureKeys as MultiSignatureKeys,
+					options.includeSenderSignature,
+				),
+			);
```
shuse2, line 198: same as above

elements/lisk-api-client/src/transaction.ts
```diff
@@ -0,0 +1,220 @@
...
+		const transactionHex = await this._channel.invoke<string>('app:getTransactionByID', {
+			id: id.toString('hex'),
+		});
+		return decodeTransaction(Buffer.from(transactionHex, 'hex'), this._schema);
+	}
+
+	public async getFromPool(): Promise<Record<string, unknown>[]> {
+		const transactionsHex = await this._channel.invoke<string[]>('app:getTransactionsFromPool');
+		const decodedTransactions: Record<string, unknown>[] = [];
+		transactionsHex.forEach(transactionHex => {
```
shuse2, line 152: prefer `for..of`

elements/lisk-api-client/src/transaction.ts
```diff
@@ -0,0 +1,220 @@
...
+				throw new Error('Missing moduleId and moduleName');
+			}
+			const registeredModule = this._nodeInfo.registeredModules.find(
+				module => module.name === input.moduleName,
+			);
+			txInput.moduleID = registeredModule?.id ? registeredModule.id : txInput.moduleID;
+		}
+		if (typeof txInput.assetID !== 'number') {
+			if (!txInput.assetName) {
+				throw new Error('Missing assetId and assetName');
```
shuse2, line 99:
```suggestion
				throw new Error('Missing assetID and assetName.');
```

elements/lisk-api-client/src/transaction.ts
```diff
@@ -0,0 +1,220 @@
...
+		const txInput = input;
+		const networkIdentifier = Buffer.from(this._nodeInfo.networkIdentifier, 'hex');
+		const { publicKey, address } = getAddressAndPublicKeyFromPassphrase(passphrase);
+		const accountHex = await this._channel.invoke<string>('app:getAccount', {
+			address: address.toString('hex'),
+		});
+		const account = decodeAccount(Buffer.from(accountHex, 'hex'), this._schema);
+		if (!txInput.moduleID) {
+			if (!txInput.moduleName) {
+				throw new Error('Missing moduleId and moduleName');
```
shuse2, line 90:
```suggestion
				throw new Error('Missing moduleID and moduleName.');
```

elements/lisk-api-client/test/unit/transaction.spec.ts
```diff
@@ -0,0 +1,716 @@
...
+				// Act
+				await transaction.getFromPool();
+
+				// Assert
+				expect(channelMock.invoke).toHaveBeenCalledTimes(1);
+				expect(channelMock.invoke).toHaveBeenCalledWith('app:getTransactionsFromPool');
+			});
+		});
+
+		describe('create', () => {
```
shuse2, line 652: `create` and `sign` has a lot of condtions in the logic.
I think each of those cases should be tested.
In general we do like
```
describe('create',  ...
  describe('when XXX happens',
    it('should be ...
  describe('when YYY happens',
    it('should be ...
```

shuse2 (changes requested):

elements/lisk-api-client/test/utils/transaction.ts
```diff
@@ -0,0 +1,554 @@
+export const nodeInfo = {
```
ishantiw, line 1: Add License on top

ishantiw (changes requested):

elements/lisk-api-client/test/unit/transaction.spec.ts
```diff
@@ -0,0 +1,205 @@
...
+				it('should return created tx', async () => {
+					const returnedTx = await transaction.create(validTransaction, passphrase1);
+					expect(returnedTx.signatures).toHaveLength(1);
+					expect(returnedTx.signatures).toMatchSnapshot();
+				});
+			});
+
+			describe('when called without module id and module name in input', () => {
+				it('should throw error', async () => {
+				try {
```
shuse2, line 108: I think u would have linting error with this. try
```
await expect(transaction.create({ ...validTransaction, moduleID: undefined }, passphrase1)).rejects.toThrow('Missing moduleID and moduleName');
```

shuse2 (approved):

ishantiw (approved):

PR #5964, opened 2020-11-10
Implement WSClient connection for APIClient - Closes #5928

### LINKED ISSUE #5928: Implement WSClient connection for APIClient
### Description

- Add WS Channel which has below interface
```
public async connect(): Promise<void>
public async disconnect(): Promise<void>
public async invoke<T>(actionName: string, params?: Record<string, unknown>): Promise<T>
public subscribe(eventName: string, cb: EventCallback): void
```

- Use library  https://github.com/heineiuo/isomorphic-ws

### Motivation

- APIClient needs to be able to communicate using IPC socket.

### Acceptance Criteria

- Test WS client can be used against core
- Acknowledgment Timeout to be defined by 2 seconds
- Connection Timeout to be defined by 2 seconds
- Add unit tests

### Additional Information
- Related issue #5919
- Reference https://github.com/LiskHQ/lisk-sdk/blob/api_client_idea/elements/lisk-api-client/src/channels/ipc.ts

nazarhussain:
### What was the problem?

This PR resolves #5928

### How was it solved?

- Created a WSClient class with required interfaces 

### How was it tested?

- Added unit tests


elements/lisk-api-client/package.json
```diff
@@ -38,8 +38,10 @@
 	"dependencies": {
 		"@liskhq/lisk-codec": "^0.1.0-alpha.0",
 		"@liskhq/lisk-cryptography": "^3.0.0-alpha.0",
+		"isomorphic-ws": "4.0.1",
 		"pm2-axon": "4.0.0",
-		"pm2-axon-rpc": "0.6.0"
+		"pm2-axon-rpc": "0.6.0",
+		"ws": "7.4.0"
```
shuse2, line 44: `ws` seems to be used only in test, so `devDependencies`?

elements/lisk-api-client/src/create_ws_client.ts
```diff
@@ -0,0 +1,26 @@
...
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+
+import { WSClient } from './ws_client';
+
+export const createWSClient = async (url: string, autoConnect = true): Promise<WSClient> => {
+	const client = new WSClient(url);
+
+	if (autoConnect) {
```
shuse2, line 21: it should always connect since other type of API client is always connected

elements/lisk-api-client/src/ws_client.ts
```diff
@@ -13,15 +13,183 @@
...
+
+	const promise = new Promise<T>((_resolve, _reject) => {
+		resolve = _resolve;
+		reject = _reject;
+	});
+
+	return { promise, resolve, reject };
 };
+
+export class WSClient {
```
shuse2, line 59: Probably `WSChannel` is more consistent with other implementation

elements/lisk-api-client/src/ws_client.ts
```diff
@@ -13,15 +13,183 @@
...
+			this._handleMessage(data as string);
+		});
+	}
+
+	public async disconnect(): Promise<void> {
+		this._requestCounter = 0;
+		this._pendingRequests = {};
+
+		if (!this._ws) {
+			return Promise.resolve();
```
shuse2, line 111: if `async`, no need to specify `Proise.resolve()`. u can just say `return`?

shuse2 (changes requested):

elements/lisk-api-client/package.json
nazarhussain: The `lisk-api-client` can be used in Node environment as well so it do require `ws` to in dependencies. The documentation of `isomorphic-ws` also suggest to install both packages. 

https://github.com/heineiuo/isomorphic-ws#usage

elements/lisk-api-client/src/ws_client.ts
nazarhussain: es-lint complaint about consistent return, so had to add this.

elements/lisk-api-client/package.json
shuse2: ah ok make sense!

shuse2 (approved):

mitsuaki-u (approved):

PR #5969, opened 2020-11-12
Add JSON format conversion function to APIClient - closes #5955

### LINKED ISSUE #5955: Add JSON format conversion function to APIClient
### Description

for account, block and transaction namespaces, it should add `toJSON` and `fromJSON` functions which converts between JSON fromat and JS object format.

### Motivation

When displaying to user, often JS object needs to be parsed to JSON format.

### Acceptance Criteria

- Add unit tests

pablitovicente:
### What was the problem?

This PR resolves #5955

### How was it solved?

- By adding `toJSON`/`fromJSON` to account, transaction and block namespaces

### How was it tested?

- added unit tests


elements/lisk-api-client/src/block.ts
```diff
@@ -12,40 +12,135 @@
...
+		for (const tx of block.payload) {
+			const { asset: txAsset, ...txRoot } = tx;
+			delete txRoot.id;
+
+			const schemaAsset = getTransactionAssetSchema(tx, this._schemas);
+			const jsonTxAsset = codec.toJSON(schemaAsset, txAsset as object);
+			const jsonTxRoot = codec.toJSON(this._schemas.transaction, txRoot);
+
+			const jsonTx = {
+				...jsonTxRoot,
```
shuse2, line 90: `id` should be inserted back as hex string

elements/lisk-api-client/src/block.ts
```diff
@@ -12,40 +12,135 @@
...
+				asset: jsonTxAsset,
+			};
+
+			json.payload.push(jsonTx as never);
```
shuse2, line 94: probably can define the payload as below, so that no need to cast it as nver
```
const payload: Record<string, unknown>[] = [];
```

elements/lisk-api-client/src/block.ts
```diff
@@ -12,40 +12,135 @@
...
+			},
+			payload: [],
+		};
+		const { asset, ...headerRoot } = block.header;
+
+		// We need to remove this 'cos schema does not have id
+		delete headerRoot.id;
+
+		// decode header
+		json.header = { ...codec.toJSON(this._schemas.blockHeader, headerRoot), asset: {} };
```
shuse2, line 71: if u reassign whole object, maybe it would be better to just create new variable here?
like
```
const header = { ...codec.toJSON(this._schemas.blockHeader, headerRoot), asset: {} };
```

elements/lisk-api-client/src/block.ts
```diff
@@ -12,40 +12,135 @@
...
+		// decode transactions
+		for (const tx of block.payload) {
+			const { asset: txAsset, ...txRoot } = tx;
+			delete txRoot.id;
+
+			const schemaAsset = getTransactionAssetSchema(tx, this._schemas);
+			const txAssetObject = codec.fromJSON(schemaAsset, txAsset as object);
+			const txRootObject = codec.fromJSON(this._schemas.transaction, txRoot);
+
+			const txObject = {
```
shuse2, line 136: id needs to be re inserted as Buffer, same for other deleted `id`

shuse2 (changes requested):

elements/lisk-api-client/src/transaction.ts
```diff
@@ -212,4 +213,36 @@ export class Transaction {
...
+	public toJSON(transaction: Record<string, unknown>): Record<string, unknown> {
+		const { asset: txAsset, ...txRoot } = transaction;
+		delete txRoot.id;
+
+		const schemaAsset = getTransactionAssetSchema(txRoot, this._schema);
+		const jsonTxAsset = codec.toJSON(schemaAsset, txAsset as object);
+		const jsonTxRoot = codec.toJSON(this._schema.transaction, txRoot);
+
+		const jsonTx = {
+			...jsonTxRoot,
```
ManuGowda, line 226: we need id property? and same for fromJSON?

ManuGowda (commented):
LGTM

ManuGowda (approved):

shuse2 (approved):

PR #5974, opened 2020-11-12
Update `getForgers` endpoint to include full validator properties - Closes #5954 

### LINKED ISSUE #5954: Update `getForgers` endpoint to include full validator properties
### Description

Update `getForgers` action to return additional properties below
- minActiveHeight
- isConsensusParticipant

### Motivation

When getting forger, it should be able to detect who is selected from selected delegates

### Acceptance Criteria

- Update functional test to see new properties are added

Incede:
### What was the problem?

This PR resolves #5954 

### How was it solved?

* Update node action getvalidators to return additional properties

### How was it tested?

* Add functional test for action getForgers


framework/src/node/node.ts
```diff
@@ -340,10 +340,17 @@ export class Node {
...
+				ReadonlyArray<{
+					address: string;
+					nextForgingTime: number;
+					minActiveHeight: number;
+					isConsensusParticipant: boolean;
+				}>
 			> => {
 				const validators = await this._chain.getValidators();
 				const validatorAddresses = validators.map(v => v.address);
+				const validatorminActiveHeights = validators.map(v => v.minActiveHeight);
```
shuse2, line 352: instead of mapping one by one, it would be better to get the validators in the `for` loop below.
```
for (let i = slotInRound; i < slotInRound + this._chain.numberOfValidators; i += 1) {
  const validator = validators[i % validators.length];
  forgersInfo.push({
    nextForgingTime,
    ...validator,
    address: validator.address.toString('hex'),
});
```

shuse2 (changes requested):

framework/src/node/node.ts
```diff
@@ -352,9 +356,13 @@ export class Node {
...
 				const forgersInfo = [];
 				for (let i = slotInRound; i < slotInRound + this._chain.numberOfValidators; i += 1) {
+					const validator = validators[i % validators.length];
 					forgersInfo.push({
-						address: validatorAddresses[i % validatorAddresses.length].toString('hex'),
+						...validator,
+						address: validator.address.toString('hex'),
 						nextForgingTime,
+						minActiveHeight: validator.minActiveHeight,
+						isConsensusParticipant: validator.isConsensusParticipant,
```
shuse2, line 365: these are unnecessary since `...validator` will cover this part.
`address` is the special case because it needs to be converted to the `hex` string

shuse2 (changes requested):

shuse2 (approved):

nazarhussain (approved):

PR #5982, opened 2020-11-16
Test Lisk client compatibility with client-side environments - Closes #5714

### LINKED ISSUE #5714: Test Lisk client compatibility with client-side environments
### Description
Since Lisk client targets clients in various environments such a Node, Browser and React Native, it's a good practice to test the compatibility with said environments. To do so, we should define a set of e2e tests with a set up that matches real life projects.

### Motivation
- Ensure Lisk Client works out of the box in client side projects.
- Ensure introducing/updating Lisk client dependencies doesn't break users' projects.

### Acceptance Criteria
E2E tests should:
- Target basic and important features of Lisk Client.
- Run on Node, Browser and React Native.
- Be incorporated into CI/CD pipelines.

nazarhussain:
### What was the problem?

This PR resolves #5714

### How was it solved?

- Added Cypress support to lisk-client
- Added e2e tests for Lisk-Cryptography 
- Added Jenkins Pipeline

### How was it tested?

Run 

```
npx lerna run test:e2e
```


Jenkinsfile.sdk
```diff
@@ -266,6 +266,18 @@ pipeline {
 						}
 					}
 				}
+
+				stage('Lisk Client End-to-End') {
+					agent { node { label 'lisk-framework' }}
```
ManuGowda, line 271:
```suggestion
					agent { node { label 'lisk-client' }}
```

elements/lisk-client/cypress/plugins/index.js
```diff
@@ -0,0 +1,35 @@
+/// <reference types="cypress" />
```
ManuGowda, line 1: `.ts` file extension? https://docs.cypress.io/guides/tooling/typescript-support.html#Types-for-plugins

elements/lisk-client/cypress/support/commands.js
```diff
@@ -0,0 +1,40 @@
+/*
```
ManuGowda, line 1: everything is commented, do we need this file?

nazarhussain: These are autogenerated file from cypress, kept as skeleton to be useful later.

elements/lisk-client/cypress/plugins/index.js
nazarhussain: These are autogenerated file from cypress, kept as skeleton to be useful later.

nazarhussain:
@ManuGowda The auto-generated files under `cypress` are kept as it is. I also had to put in elsint ignore to avoid any check on these.

ManuGowda (approved):

nazarhussain:
@ManuGowda @mitsuaki-u The important change that should be considered is `test/tsconfig.json` file. Earlier all such files are linked to templates. Which have `sourceMap: true`, this option does not work with cypress, and to fix that only error I had to create a new tsconfig for the tests of lisk-client with `sourceMap: false`

https://github.com/LiskHQ/lisk-sdk/pull/5982/files#diff-53aaef287c11741352c0d54b1a66d8c8dd82505cbbe42236f40680339e0ff4da

elements/lisk-client/test/lisk-cryptograph/buffer.spec.ts
```diff
@@ -0,0 +1,188 @@
+/*
```
mitsuaki-u, line 1: the folder name is missing a `y` in `cryptography`

mitsuaki-u (approved):
Minor comment but LGTM

PR #5984, opened 2020-11-18
Fix nonce validation logic - Closes #5983

### LINKED ISSUE #5983: Wrong logic for nonce validation on transaction.create
### Expected behavior

When using an account with nonce === 0 transaction.create should not fail.

### Actual behavior

It's failing as a check is implemented in a way that a truthy value is possible.

### Steps to reproduce

create a transaction for an account with nonce === 0 using lisk-api-client

5.1

Incede:
### What was the problem?

This PR resolves #5983 

### How was it solved?

* Fix logic to check for undefined, null or negative nonce values in the input

### How was it tested?

* Add unit test for cases when nonce is zero, negative & undefined


elements/lisk-api-client/test/unit/transaction.spec.ts
```diff
@@ -121,14 +121,50 @@ describe('transaction', () => {
 				});
 			});
 
-			describe('when called without nonce in input', () => {
+			describe('when called without nonce in input and account is not supporting nonce', () => {
```
pablitovicente, line 124: Maybe try to improve this phrase it's confusing.

pablitovicente (approved):

shuse2 (approved):

PR #6001, opened 2020-11-23
Remove IPC config property - Closes #5965

### LINKED ISSUE #5965: Remove `ipc` property from config
### Description

`ipc` property in the config should be removed and the usage should be replaced by `rpc.enable === true && rpc.mode === 'ipc'` 

### Motivation

By adding `rpc` property, ipc property is outdated

### Acceptance Criteria

There should be no usage of `ipc` property from the config

shuse2:
### What was the problem?

This PR resolves #5965 

### How was it solved?

Remove all reference to the config

### How was it tested?

- All test should not be affected

### DIFF
--- a/config/devnet/config.json
+++ b/config/devnet/config.json
@@ -1,7 +1,4 @@
 {
-	"ipc": {
-		"enabled": false
-	},
 	"version": "2.1.0",
 	"networkVersion": "1.1",
 	"label": "devnet",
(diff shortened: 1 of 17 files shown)


shuse2:
https://github.com/TypeStrong/ts-node/issues/619#issuecomment-511366883
This solution would work for running the child process with ts-node, but still it does not initialize

ManuGowda (approved):

nazarhussain (approved):

PR #6003, opened 2020-11-23
Fix timeout for action "monitor:getForkStats" in API Client - Closes #5986

### LINKED ISSUE #5986: Timeout for action "monitor:getForkStats" in API Client when node is syncing
### Should return 
```
{ forkEventCount: 0, blockHeaders: {} }
```
or some other output value
### Actual behavior
```
Error: Response not received in 3000ms
    at Timeout.<anonymous> (/Users/user/node_modules/@liskhq/lisk-api-client/dist-node/ws_channel.js:11:16)
    at listOnTimeout (internal/timers.js:549:17)
    at processTimers (internal/timers.js:492:7)
```
If the action is called when node is not syncing it works as expected
### Steps to reproduce
Invoke action `monitor:getForkStats` from API client while node is syncing

5.1

Incede:
### What was the problem?

This PR resolves #5986 

### How was it solved?
* Return block header from action `getForkStats` as JSON instead of object

### How was it tested?

* Unit Test was updated
* Tested manually by linking locally


nazarhussain (approved):

framework-plugins/lisk-framework-monitor-plugin/src/monitor_plugin.ts
```diff
@@ -197,7 +197,7 @@ export class MonitorPlugin extends BasePlugin {
 		if (this._state.forks.blockHeaders[blockId]) {
 			this._state.forks.blockHeaders[blockId].timeReceived = Date.now();
 		} else {
-			const decodedHeader = codec.decode<RawBlockHeader>(this.schemas.blockHeader, header);
+			const decodedHeader = codec.decodeJSON<RawBlockHeader>(this.schemas.blockHeader, header);
```
ManuGowda, line 200:
```suggestion
			const decodedHeader = codec.decodeJSON<object>(this.schemas.blockHeader, header);
```

framework-plugins/lisk-framework-monitor-plugin/test/unit/forks.spec.ts
```diff
@@ -57,7 +57,7 @@ describe('_handleFork', () => {
 			monitorInstance.schemas.block,
 			Buffer.from(encodedBlock, 'hex'),
 		);
-		const expectedDecodedHeader = codec.decode<RawBlockHeader>(
+		const expectedDecodedHeader = codec.decodeJSON<RawBlockHeader>(
```
ManuGowda, line 60:
```suggestion
		const expectedDecodedHeader = codec.decodeJSON<object>(
```

ManuGowda (commented):
either u need use type `BaseBlockHeaderJSON` or `Record<string, unknown>` for decodeJSON

framework-plugins/lisk-framework-monitor-plugin/src/monitor_plugin.ts
shuse2: prefer to use `Record<string, unknown>` in case of JSON object

ManuGowda (approved):

PR #6005, opened 2020-11-23
Fix empty object array case not to throw - Closes #6000

### LINKED ISSUE #6000: Fix codec to be able to decode/encode empty object
### Expected behavior
Decoded value should be able to encode back to the binary

### Actual behavior
With below schema and value, it fails to encode back to buffer again.

### Steps to reproduce
```
const schema = {
	$id: 'testSchema',
	type: 'object',
	properties: {
		b: { fieldNumber: 2, dataType: 'string' },
		a: { fieldNumber: 1, dataType: 'string' },
		d: { fieldNumber: 4, dataType: 'bytes' },
		e: { fieldNumber: 5, dataType: 'uint32' },
		c: {
			type: 'object',
			fieldNumber: 3,
			properties: {
				cc: { fieldNumber: 3, dataType: 'string' },
				ca: { fieldNumber: 1, dataType: 'string' },
				cb: {
					type: 'object',
					fieldNumber: 2,
					properties: {
						cbb: { fieldNumber: 2, dataType: 'string' },
						cba: { fieldNumber: 1, dataType: 'string' },
						cbc: {
							type: 'object',
							fieldNumber: 3,
							properties: {
								cbcb: { fieldNumber: 3, dataType: 'string' },
								cbca: { fieldNumber: 2, dataType: 'string' },
							},
						},
						cbd: { fieldNumber: 4, dataType: 'string' },
					},
				},
			},
		},
		f: {
			type: 'array',
			fieldNumber: 6,
			items: {
				type: 'object',
				properties: {
					fc: { fieldNumber: 3, dataType: 'string' },
					fa: { fieldNumber: 1, dataType: 'string' },
					fb: {
						type: 'object',
						fieldNumber: 2,
						properties: {
							fbb: { fieldNumber: 2, dataType: 'string' },
							fba: { fieldNumber: 1, dataType: 'string' },
							fbc: {
								type: 'object',
								fieldNumber: 3,
								properties: {
									fbcb: { fieldNumber: 3, dataType: 'string' },
									fbca: { fieldNumber: 2, dataType: 'string' },
								},
							},
							fbd: { fieldNumber: 4, dataType: 'string' },
						},
					},
				},
			},
		},
	},
};
const obj = codec.decode(schema, Buffer.from('80000000008000e348328000000000804800', 'hex');
codec.encode(schema, obj);
```

shuse2:
### What was the problem?

This PR resolves #6000 

### How was it solved?

- Remove the initialization of the empty object, since all empty object should have at least some default value. Therefore, if the object size is `0`, it should not create an empty object

### How was it tested?

- Added test case

### DIFF
--- a/elements/lisk-codec/test/random_bytes.spec.ts
+++ b/elements/lisk-codec/test/random_bytes.spec.ts
@@ -1,0 +1,102 @@
+/*
+ * Copyright © 2020 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+import { Codec } from '../src/codec';
+
+describe('random bytes', () => {
+	const schema = {
+		$id: 'testSchema',
+		type: 'object',
+		properties: {
+			b: { fieldNumber: 2, dataType: 'string' },
+			a: { fieldNumber: 1, dataType: 'string' },
+			d: { fieldNumber: 4, dataType: 'bytes' },
+			e: { fieldNumber: 5, dataType: 'uint32' },
+			c: {
+				type: 'object',
+				fieldNumber: 3,
+				properties: {
+					cc: { fieldNumber: 3, dataType: 'string' },
+					ca: { fieldNumber: 1, dataType: 'string' },
+					cb: {
+						type: 'object',
+						fieldNumber: 2,
+						properties: {
+							cbb: { fieldNumber: 2, dataType: 'string' },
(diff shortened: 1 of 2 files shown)


ManuGowda (approved):

elements/lisk-codec/test/random_bytes.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+	};
+	let codec: Codec;
+
+	beforeEach(() => {
+		codec = new Codec();
+		codec.addSchema(schema);
+	});
+
+	describe('decode and encode', () => {
+		const testCases = [Buffer.from('80000000008000e348328000000000804800', 'hex')];
```
nazarhussain, line 89: Please add a decoded version of this buffer in comments, that will make it easy to understand the test scenario.

nazarhussain (approved):

elements/lisk-codec/test/random_bytes.spec.ts
shuse2: This is random bytes, so decoded version is just an empty objects as defined by schema.
ie) it doesn't corresponds to specific expectation

PR #6020, opened 2020-11-24
Add validation to check the network data before using - Closes #6019

### LINKED ISSUE #6019: when using `app:network:event`, the content must be validated
### Expected behavior
`app:network:event` can contain any information, therefore, content should be validated and used with caution.

### Actual behavior
In monitor and report misbehavior plugin, the content is not validated properly. When sending unexpected content, it will crash the plugin

### Steps to reproduce
1. Run a node with monitor and report misbehavior plugin. 
2. send random data `p2pNode.sendToPeer({ event: 'postBlock', data: null }, '127.0.0.1:5000');` like this to the node
3. observe p2p client will be banned, and also plugin is crashed

5.0-

shuse2:
### What was the problem?

This PR resolves #6019 

### How was it solved?

- Add validation before using the network data

### How was it tested?

- Add unit test
- Link with core and run reportMisbehavior and monitor plugins, and try to send invalid postBlock data to the node. Observe it will not crash the plugin

### DIFF
--- a/framework-plugins/lisk-framework-monitor-plugin/src/monitor_plugin.ts
+++ b/framework-plugins/lisk-framework-monitor-plugin/src/monitor_plugin.ts
@@ -15,6 +15,7 @@ import { Server } from 'http';
 import { RawBlock, RawBlockHeader } from '@liskhq/lisk-chain';
 import { codec } from '@liskhq/lisk-codec';
 import { hash } from '@liskhq/lisk-cryptography';
+import { validator } from '@liskhq/lisk-validator';
 import { objects } from '@liskhq/lisk-utils';
 import {
 	ActionsDefinition,
@@ -32,6 +33,7 @@ import * as middlewares from './middlewares';
 import * as config from './defaults';
 import { Options, SharedState } from './types';
 import * as controllers from './controllers';
+import { transactionAnnouncementSchema, postBlockEventSchema } from './schema';
 
 // eslint-disable-next-line
 const pJSON = require('../package.json');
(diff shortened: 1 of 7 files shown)


framework-plugins/lisk-framework-monitor-plugin/src/monitor_plugin.ts
```diff
@@ -148,10 +150,21 @@ export class MonitorPlugin extends BasePlugin {
...
+				if (errors.length > 0) {
+					return;
+				}
 				this._handlePostTransactionAnnounce(data as { transactionIds: string[] });
 			}
 
 			if (event === 'postBlock') {
+				const errors = validator.validate(postBlockEventSchema, data as Record<string, unknown>);
+				if (errors.length > 0) {
+					return;
```
mitsuaki-u, line 166: should we add a debug log here

mitsuaki-u (approved):
LGTM one minor comment

framework-plugins/lisk-framework-monitor-plugin/src/monitor_plugin.ts
nazarhussain: Anyone can spam with random data and logging all can be IO consumption. But for debug mode it would be useful.

framework-plugins/lisk-framework-monitor-plugin/src/schema.ts
```diff
@@ -0,0 +1,39 @@
...
+	properties: {
+		transactionIds: {
+			type: 'array',
+			items: {
+				type: 'string',
+				format: 'hex',
+			},
+		},
+	},
+};
```
nazarhussain, line 39: This does not match exactly with what we use in the Node. 

https://github.com/LiskHQ/lisk-sdk/blob/4409d34102b5bc6703fbc48813bd1f7557ee4a93/framework/src/node/transport/schemas.ts#L51-L65

nazarhussain (changes requested):

framework-plugins/lisk-framework-monitor-plugin/src/monitor_plugin.ts
shuse2: In this case, the framework will log and ban the peer, so i think it should be sufficient

framework-plugins/lisk-framework-monitor-plugin/src/schema.ts
shuse2: number of the list doesnt really matter here. it just needs to be hex string array to process further

nazarhussain (approved):

PR #6022, opened 2020-11-24
Fix APIClient event subscription - Closes #6021

### LINKED ISSUE #6021: APIClient event is outdated and type in incorrect
### Expected behavior
- APIClient subscribe event should be aligned with #6012 
- APIClient should be usable from plugins

### Actual behavior
-  APIClient subscribe is outdated

### Steps to reproduce
- Try to use APIClient subscribe

5.0-

shuse2:
### What was the problem?

This PR resolves #6021 

### How was it solved?

- Update event subscription type and logics

### How was it tested?

- Try to listen to event using api client


ManuGowda (approved):
The tests seems to fail, otherwise LGTM

elements/lisk-api-client/src/types.ts
```diff
@@ -33,18 +33,13 @@ export interface JSONRPCResponse<T> {
 
 export type JSONRPCMessage<T> = JSONRPCNotification<T> | JSONRPCResponse<T>;
 
-export interface EventInfoObject<T> {
-	readonly module: string;
-	readonly name: string;
-	readonly data: T;
-}
-export type EventCallback<T> = (event: EventInfoObject<T>) => void | Promise<void>;
+export type EventCallback = (event?: Record<string, unknown>) => void | Promise<void>;
```
nazarhussain, line 36: Can we make generic as it was earlier, so we can pass on the type of data the event handler is gonna receive?

nazarhussain (changes requested):

elements/lisk-api-client/src/api_client.ts
```diff
@@ -51,7 +51,7 @@ export class APIClient {
 		return this._channel.invoke(actionName, params);
 	}
 
-	public subscribe<T = Record<string, unknown>>(eventName: string, cb: EventCallback<T>): void {
+	public subscribe(eventName: string, cb: EventCallback): void {
```
nazarhussain, line 54: Should not we make this one generic as we did for the other clients? 
```suggestion
	public subscribe<T = Record<string, unknown>>(eventName: string, cb: EventCallback<T>): void {
```

nazarhussain (changes requested):

elements/lisk-api-client/src/api_client.ts
shuse2: i thought i updated, maybe it was out of commit...

nazarhussain (approved):

PR #6028, opened 2020-11-25
Fix to ignore zero amount change and add error for negative amount - Closes #6027

### LINKED ISSUE #6027: Report misbehavior fails when delegate has zero balance
### Expected behavior
when reward is zero, it should not try to change the balance.

### Actual behavior
When delegate has zero balance, reportMisbehavior reward will be 0.
it will try to deduct `0` which should have no effect. However, token has the check for the min balance requirement. Therefore, the transaction will be rejected

### Steps to reproduce
1. Start 2 nodes and enable plugin in one of the plugin (add reportMisbehavior config encrypted passphrase)
2. enable plugin with api client
3. wait till misbehavior happen
4. observe the error log

5.0-

shuse2:
### What was the problem?

This PR resolves #6027 

### How was it solved?

- Add check for debit and credit to throw error if amount is negative
- Add check for debit and credit to ignore when amount is zero

### How was it tested?

TBD


nazarhussain (approved):

framework/src/modules/token/token_module.ts
```diff
@@ -51,6 +51,9 @@ export class TokenModule extends BaseModule {
 			if (typeof amount !== 'bigint') {
 				throw new Error('Amount must be a bigint');
 			}
+			if (amount <= BigInt(0)) {
+				throw new Error('Amount must be a positive bigint');
```
ManuGowda, line 55:
```suggestion
				throw new Error('Amount must be a positive bigint.');
```

framework/src/modules/token/token_module.ts
```diff
@@ -68,6 +71,9 @@ export class TokenModule extends BaseModule {
...
```
ManuGowda, line 75:
```suggestion
				throw new Error('Amount must be a positive bigint.');
```

framework/test/unit/modules/token/token_module.spec.ts
```diff
@@ -103,14 +103,36 @@ describe('token module', () => {
...
+
+		it('should throw error if amount is less than zero', async () => {
+			senderAccount.token.balance = BigInt(0);
+
+			return expect(
+				tokenModule.reducers.credit(
+					{ address: senderAccount.address, amount: BigInt(-10) },
+					stateStore,
+				),
+			).rejects.toThrow('Amount must be a positive bigint');
```
ManuGowda, line 125:
```suggestion
			).rejects.toThrow('Amount must be a positive bigint.');
```

ManuGowda (approved):

framework/test/unit/modules/token/token_module.spec.ts
```diff
@@ -103,14 +103,36 @@ describe('token module', () => {
...
-		it('should throw error if account does not have sufficient balance', async () => {
+		it('should throw error if amount is zero', async () => {
 			senderAccount.token.balance = BigInt(0);
 
 			return expect(
 				tokenModule.reducers.credit(
 					{ address: senderAccount.address, amount: BigInt('0') },
 					stateStore,
 				),
+			).rejects.toThrow('Amount must be a positive bigint');
```
shuse2, line 114:
```suggestion
			).rejects.toThrow('Amount must be a positive bigint.');
```

PR #6029, opened 2020-11-26
Unformatted JSON-RPC request crashes the application - Closes #6026

### LINKED ISSUE #6026: Unformatted JSON-RPC request crashes the application
### Expected behavior

If application receive unformatted or invalid JSON format request as JSON-RPC request. It should validate, show the error and keep running.  

### Actual behavior

If application receives a request which is invalid JSON then it crashes with below error. 

```
15:40:35 FATAL lisk-framework: System error: unhandledRejection (module=lisk:app)
Message: Unexpected end of JSON input
 Trace: SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at Function.fromJSONRPCRequest (~/.config/yarn/global/node_modules/lisk-framework/dist-node/controller/action.js:15:72)
    at ~/.config/yarn/global/node_modules/lisk-framework/dist-node/controller/bus.js:206:54
    at runMicrotasks (<anonymous>)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
15:40:35 INFO lisk-framework: Application shutdown started (module=lisk:app)
{
 "errorCode": 1,
 "message": "Unexpected end of JSON input"
}
```
The root cause of the issue is following code. 

https://github.com/LiskHQ/lisk-sdk/blob/7c8644c8bdbfa84c2f14d1a658accb8cf46bb981/framework/src/controller/action.ts#L58

The same can be produced with an event as well. But since application don't receive external notifications so its safe. But application is accepting external action invocation calls so this issue can crash any application. 

### Steps to reproduce

- Run the lisk-core with command `lisk-core start --network devnet --api-ws`
- Or you can use any other application as well 
- Use this command to install a command line utility `brew install websocat`
- Connect to the node `websocat ws://localhost:8080/ws`
- If you see an empty prompt, means you are connected to web socket server
- The prompt accepts any input that will be sent as request to server 
- Press new line that means we are sending empty string as request to server 
- This will fail the request parsing and crashes the application

5.0.0

nazarhussain:
### What was the problem?

This PR resolves #6026

### How was it solved?

- Add extra validation on request parsing

### How was it tested?

- Run framework unit tests 
- Follow the steps mentioned on the issue to test manually

### DIFF
--- a/framework/src/controller/bus.ts
+++ b/framework/src/controller/bus.ts
@@ -167,10 +167,28 @@ export class Bus {
 	public async invoke<T>(
 		rawRequest: string | JSONRPC.RequestObject,
 	): Promise<JSONRPC.ResponseObjectWithResult<T>> {
-		const request =
+		let request!: JSONRPC.RequestObject;
+
+		// As the request can be invoked from external source, so we should validate if it exists and valid JSON object
+		if (!rawRequest) {
+			this.logger.error('Empty invoke request.');
+			throw new JSONRPC.JSONRPCError(
+				'Invalid invoke request.',
+				JSONRPC.errorResponse(null, JSONRPC.invalidRequest()),
+			);
+		}
+
+		try {
+			request =
 			typeof rawRequest === 'string'
 				? (JSON.parse(rawRequest) as JSONRPC.RequestObject)
 				: rawRequest;
+		} catch (error) {
+			throw new JSONRPC.JSONRPCError(
+				'Invalid invoke request.',
+				JSONRPC.errorResponse(null, JSONRPC.invalidRequest()),
+			);
+		}
 
 		try {
 			JSONRPC.validateJSONRPCRequest(request as never);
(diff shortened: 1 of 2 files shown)


framework/src/controller/bus.ts
```diff
@@ -167,10 +167,28 @@ export class Bus {
...
+		if (!rawRequest) {
+			this.logger.error('Empty invoke request.');
+			throw new JSONRPC.JSONRPCError(
+				'Invalid invoke request.',
+				JSONRPC.errorResponse(null, JSONRPC.invalidRequest()),
+			);
+		}
+
+		try {
+			request =
```
ManuGowda, line 182: we had request schema validation in place, we can't use that here to validate the json rpc request?

mitsuaki-u (approved):

ManuGowda (approved):

PR #6040, opened 2020-12-04
Separate action and p2p endpoint - Closes #6034

### LINKED ISSUE #6034: Application crashes when some specific actions are called over a limit
### Expected behaviour

Any action can be called as many times any module or APIClient can

### Actual behaviour

When certain actions that are handled by transport class are called over 10 or more times then it crashes the node with an error.
```ts
16:33:03 INFO lisk-framework: Application shutdown completed (module=lisk:app)
{
 "errorCode": 1,
 "message": "Peer not found: undefined"
} 
```

When an APIClient calls it then more than 10 times then it throws error in client side.

### Steps to reproduce

Run a node and using client run below,
```ts
for (let i=0;i<10;i++) {
  await client.invoke('app:getLastBlock');
}
```

`5.0`

shuse2:
### What was the problem?

This PR resolves #6034 

### How was it solved?

- Remove actions which is undocumented and meant for P2P communications from application
- Register those specific functions to the P2P communication
- Keep the `getLastBlock` action as this was used both for the action and P2P

### How was it tested?

- Functional tests should work as it is
- Call the `app:getLastBlock` actions more than 10 times in loop
- Run a node with this change and try to sync with the the node

### DIFF
--- a/framework/src/controller/bus.ts
+++ b/framework/src/controller/bus.ts
@@ -389,8 +389,8 @@ export class Bus {
 					.then(data => {
 						cb(null, data as JSONRPC.ResponseObjectWithResult);
 					})
-					.catch(error => {
-						cb(error);
+					.catch((error: JSONRPC.JSONRPCError) => {
+						cb(error, error.response);
 					});
 			},
 		);
@@ -407,8 +407,8 @@ export class Bus {
 				.then(data => {
 					socket.send(JSON.stringify(data as JSONRPC.ResponseObjectWithResult));
 				})
-				.catch(error => {
-					socket.send(JSON.stringify((error as JSONRPC.JSONRPCError).response));
+				.catch((error: JSONRPC.JSONRPCError) => {
+					socket.send(JSON.stringify(error.response));
 				});
 		});
 	}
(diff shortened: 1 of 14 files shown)


framework/src/controller/bus.ts
```diff
@@ -376,7 +376,17 @@ export class Bus {
...
+							cb(error);
+							return;
+						}
+						const parsedAction = Action.fromJSONRPCRequest(action);
+						cb(
+							new JSONRPC.JSONRPCError(
+								(error as Error).message,
+								JSONRPC.errorResponse(parsedAction.id, JSONRPC.invalidRequest()),
+							),
+						);
```
nazarhussain, line 389: Please move this logic to `Bus.invoke` to it always throw a `JSONRPCError`. This way it will be consistent across all type of channels.

nazarhussain (changes requested):

Incede (approved):

nazarhussain (approved):

nazarhussain (approved):

Incede (approved):

PR #6042, opened 2020-12-07
Resolve relative root path - Closes #6035

### LINKED ISSUE #6035: Relative rootPath keeps node from booting
### Expected behavior
When configuring a node with rootPath: './store' the node data should store in the project root/store folder and run.

### Actual behavior
When configuring a node with rootPath: './store' the node will hang on the loading controller step. It creates the store folder but the node won't start.

### Steps to reproduce
Configure an application with configuration argument rootPath: './store'

SDK V5, experienced on Mac and Debian

Incede:
### What was the problem?

This PR resolves #6035 

### How was it solved?

* Add logic to resolve root path while creating system directories

### How was it tested?

* Updated unit tests
* Tested manually 
```
config.rootPath = './store';
app = Application.defaultApplication(genesisBlock, config);
app.run();
```

### DIFF
--- a/framework/src/application.ts
+++ b/framework/src/application.ts
@@ -13,7 +13,6 @@
  */
 
 import * as fs from 'fs-extra';
-import * as os from 'os';
 import * as path from 'path';
 import * as psList from 'ps-list';
 import * as assert from 'assert';
@@ -140,7 +139,6 @@ export class Application {
 			config.label ?? `lisk-${config.genesisConfig?.communityIdentifier}`;
 
 		const mergedConfig = objects.mergeDeep({}, appConfig, config) as ApplicationConfig;
-		mergedConfig.rootPath = mergedConfig.rootPath.replace('~', os.homedir());
 		const applicationConfigErrors = validator.validate(applicationConfigSchema, mergedConfig);
 		if (applicationConfigErrors.length) {
 			throw new LiskValidationError(applicationConfigErrors);
--- a/framework/src/system_dirs.ts
+++ b/framework/src/system_dirs.ts
@@ -12,14 +12,18 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
+import * as os from 'os';
 import * as path from 'path';
 
 // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types,@typescript-eslint/explicit-function-return-type
-export const systemDirs = (appLabel: string, rootPath: string) => ({
-	dataPath: path.resolve(path.join(rootPath, appLabel)),
-	data: path.resolve(path.join(rootPath, appLabel, 'data')),
-	tmp: path.resolve(path.join(rootPath, appLabel, 'tmp')),
-	logs: path.resolve(path.join(rootPath, appLabel, 'logs')),
-	sockets: path.resolve(path.join(rootPath, appLabel, 'tmp', 'sockets')),
-	pids: path.resolve(path.join(rootPath, appLabel, 'tmp', 'pids')),
-});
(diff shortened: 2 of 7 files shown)


framework/src/system_dirs.ts
```diff
@@ -16,10 +16,10 @@ import * as path from 'path';
...
-	tmp: path.join(rootPath, appLabel, 'tmp'),
-	logs: path.join(rootPath, appLabel, 'logs'),
-	sockets: path.join(rootPath, appLabel, 'tmp', 'sockets'),
-	pids: path.join(rootPath, appLabel, 'tmp', 'pids'),
+	dataPath: path.resolve(rootPath, appLabel),
+	data: path.resolve(rootPath, appLabel, 'data'),
+	tmp: path.resolve(rootPath, appLabel, 'tmp'),
+	logs: path.resolve(rootPath, appLabel, 'logs'),
+	sockets: path.resolve(rootPath, appLabel, 'tmp', 'sockets'),
+	pids: path.resolve(rootPath, appLabel, 'tmp', 'pids'),
```
nazarhussain, line 24: Problem using this approach is that if ever `appLabel` got `/` in start it will skip the `rootPath` completely. e.g. 

```
path.resolve('/root/path', 'data', 'second/dir', '/third/dir`) 
```

The above will not result `/root/path/data/second/dir/third/dir` instead it will just return `/third/dir`.

So I would suggest to use `path.join` first and then use single value for `path.resolve` to get absolute path.

framework/test/unit/application.spec.ts
```diff
@@ -668,6 +668,7 @@ describe('Application', () => {
 		let _nodeDBSpy: jest.SpyInstance<any, unknown[]>;
 
 		beforeEach(async () => {
+			config.rootPath = '/lisk';
```
nazarhussain, line 671: May be instead of changing current ones, we should add a new tests case particularly for absolute root path.

framework/test/unit/controller/controller.spec.ts
```diff
@@ -74,7 +74,7 @@ describe('Controller Class', () => {
 		publish: jest.fn(),
 	};
 	const config = {
-		rootPath: '~/.lisk',
+		rootPath: '/lisk',
```
nazarhussain, line 77: Same here.

nazarhussain (changes requested):

framework/src/system_dirs.ts
Incede: @shuse2 any suggestions ?

shuse2: ah i also see the problem
i think we can do like
```
dataPath: path.resolve(path.join(rootPath, appLabel)),
```

framework/src/application.ts
```diff
@@ -140,7 +139,6 @@ export class Application {
 			config.label ?? `lisk-${config.genesisConfig?.communityIdentifier}`;
 
 		const mergedConfig = objects.mergeDeep({}, appConfig, config) as ApplicationConfig;
```
shuse2, line 141: i think in this file there are place that uses `rootPath` directly

nazarhussain: I believe whenever rootPath is directly passed the inner function either logger or controller uses the system dir to get the correct path. So that will work.

nazarhussain (dismissed):

shuse2 (approved):

shuse2 (approved):

shuse2 (approved):

nazarhussain (approved):

PR #6044, opened 2020-12-09
Respond status code when client IP not allowed - Closes #6037

### LINKED ISSUE #6037: HTTP status code when client IP is not allowed
### Expected behavior
If the client IP is not allowed to access an API the response HTTP status code should be `401`.

### Actual behavior
The HTTP and monitoring APIs (and possibly more) return `500`.

### Steps to reproduce
```
lisk-core start --enable-http-api-plugin --http-api-plugin-port=5000 --http-api-plugin-whitelist=1.2.3.4
curl --verbose http://127.0.0.1:5000/api/node/info
```

5.0.0 / 3.0.0-beta.2

Incede:
### What was the problem?

This PR resolves #6037

### How was it solved?

* Handle error status response

### How was it tested?
```
./bin/run sdk:link /path-to-lisk-sdk/lisk-sdk/sdk
./bin/run start --enable-http-api-plugin --http-api-plugin-port=5000 --http-api-plugin-whitelist=1.2.3.4
curl --verbose http://127.0.0.1:5000/api/node/info
```

### DIFF
--- a/framework-plugins/lisk-framework-http-api-plugin/src/middlewares/errors.ts
+++ b/framework-plugins/lisk-framework-http-api-plugin/src/middlewares/errors.ts
@@ -32,6 +32,7 @@ export const errorMiddleware = () => (
 	_next: NextFunction,
 ): void => {
 	let errors;
+	let responseCode = 500;
 
 	if (Array.isArray(err)) {
 		errors = err;
@@ -46,7 +47,11 @@ export const errorMiddleware = () => (
 		Object.defineProperty(error, 'message', { enumerable: true });
 	}
 
-	const statusCode = err instanceof ErrorWithStatus ? err.statusCode : 500;
+	if (err instanceof ErrorWithStatus) {
+		const { statusCode, ...message } = err;
+		errors = message;
+		responseCode = statusCode;
+	}
 
-	res.status(statusCode).send({ errors });
+	res.status(responseCode).send({ errors });
 };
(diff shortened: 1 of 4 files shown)


framework-plugins/lisk-framework-http-api-plugin/src/middlewares/errors.ts
```diff
@@ -38,5 +46,9 @@ export const errorMiddleware = () => (
 		Object.defineProperty(error, 'message', { enumerable: true });
 	}
 
-	res.status(500).send({ errors });
+	const statusCode = (err as ErrorWithStatus).statusCode
```
shuse2, line 49: maybe it's better to check like
```
const statusCode = err instanceof ErrorWithStatus ? err.statusCode : 500;
```

shuse2 (changes requested):
same change should be applied for monitor plugin

framework-plugins/lisk-framework-http-api-plugin/src/middlewares/errors.ts
```diff
@@ -13,6 +13,14 @@
  */
 import { Request, Response, NextFunction } from 'express';
 
+export class ErrorWithStatus extends Error {
```
nazarhussain, line 16:
```suggestion
export class HTTPError extends Error {
```

nazarhussain (changes requested):

nazarhussain (approved):

shuse2 (approved):

PR #6057, opened 2021-01-05
Resolve tilde - Closes #6017

### LINKED ISSUE #6017: Update createIPCClient to allow "~"
### Expected behavior
createIPCClient should accept relative path, absolute path and path with `~`

### Actual behavior
`~` cannot be used for the input.

### Steps to reproduce
Create IPCClient with `~/.lisk/lisk-core`

5.0-

Incede:
### What was the problem?

This PR resolves #6017 

### How was it solved?

* Resolve tilde before setting up socket directories

### How was it tested?

Tested manually by linking to the local `lisk-api-client` and `createIPCClient('~/.lisk/lisk-core')`

### DIFF
--- a/elements/lisk-api-client/src/ipc_channel.ts
+++ b/elements/lisk-api-client/src/ipc_channel.ts
@@ -19,6 +19,7 @@
 /// <reference path="../external_types/pm2-axon-rpc/index.d.ts" />
 import * as path from 'path';
 import * as axon from 'pm2-axon';
+import { homedir } from 'os';
 import { PubSocket, PullSocket, PushSocket, SubSocket, ReqSocket } from 'pm2-axon';
 import { Client as RPCClient } from 'pm2-axon-rpc';
 import { EventEmitter } from 'events';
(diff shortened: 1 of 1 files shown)


elements/lisk-api-client/src/ipc_channel.ts
```diff
@@ -34,7 +35,8 @@ import { convertRPCError } from './utils';
 const CONNECTION_TIME_OUT = 2000;
 
 const getSocketsPath = (dataPath: string) => {
-	const socketDir = path.join(path.resolve(dataPath), 'tmp', 'sockets');
+	const dataPathWithoutTilde = dataPath.replace('~', homedir());
+	const socketDir = path.join(path.resolve(dataPathWithoutTilde), 'tmp', 'sockets');
```
ManuGowda, line 39:
```suggestion
	const socketDir = path.join(dataPath.replace('~', homedir()), 'tmp', 'sockets');
```

Incede: I think this will not work for relative data paths like `./myDir`.
Although can be reformatted as 
`const socketDir = path.join(path.resolve(dataPath.replace('~', homedir())), 'tmp', 'sockets');`

ManuGowda (approved):

mitsuaki-u (approved):
LGTM but jenkins still failling

PR #6064, opened 2021-01-06
Enable child process IPC plugins - Closes #6010

### LINKED ISSUE #6010: Update IPC server to respect the enable flag
### Description

IPC server should be accessible only if the rpc is enabled.

#### Idea
- split the ipc server for external communication and child process communication, and always limit the access to child process communication ipc, and only create IPC server when `rpc` enable and mode `ipc`

### Motivation

External access through IPC should be avoided when rpc enable is `false`

### Acceptance Criteria
When rpc enable is `false`, there shouldn't be a way to connect through IPC

mitsuaki-u:
### What was the problem?

This PR resolves #6010 

### How was it solved?

Add checks in controller and bus for rpc / ipc configurations.

### How was it tested?

- Add unit tests
- Update functional tests


framework/src/controller/bus.ts
```diff
@@ -119,8 +121,9 @@ export class Bus {
 	}
 
 	public async setup(): Promise<boolean> {
-		await this._setupIPCServer();
-
+		if (this.config.rpc.enable && this.config.rpc.mode === 'ipc') {
```
shuse2, line 124:
```suggestion
		if (this.config.rpc.enable) {
```

shuse2 (changes requested):
LGTM except one comment =)

nazarhussain (approved):

shuse2 (approved):

PR #6077, opened 2021-01-13
Namespace external types - Closes #6045

### LINKED ISSUE #6045: All framework plugin must use  the same framework version
### Expected behavior
plugin should be independent on each other, and if it's patch release, it should be possible to install different version.

### Actual behavior
framework contains `external_types` for the `axon`, and these types will be defined twice, and it will not be able to be built.

### Steps to reproduce
Try to install 2 packages which use different framework version

### Additional information
possible mitigation will be to remove the usage of `axon` and use the raw IPC, which will not require types.
Similar implementation already exists in `ws`.

5.0.0-

Incede:
### What was the problem?

This PR resolves #6045 

### How was it solved?

* Namespace external types

### How was it tested?

Tested manually by successfully building sdk with 2 plugins using different framework version

### DIFF
--- a/framework/src/controller/bus.ts
+++ b/framework/src/controller/bus.ts
@@ -14,9 +14,8 @@
 
 import { LiskValidationError } from '@liskhq/lisk-validator';
 import { EventEmitter2, Listener } from 'eventemitter2';
-import * as axon from 'pm2-axon';
-import { ReqSocket } from 'pm2-axon';
-import { Client as RPCClient } from 'pm2-axon-rpc';
+import { Axon } from 'pm2-axon';
+import { AxonRpc } from 'pm2-axon-rpc';
 import { Logger } from '../logger';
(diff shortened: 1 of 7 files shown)


framework/src/controller/bus.ts
```diff
@@ -14,9 +14,8 @@
 
 import { LiskValidationError } from '@liskhq/lisk-validator';
 import { EventEmitter2, Listener } from 'eventemitter2';
-import * as axon from 'pm2-axon';
-import { ReqSocket } from 'pm2-axon';
-import { Client as RPCClient } from 'pm2-axon-rpc';
+import { Axon } from 'pm2-axon';
+import { AxonRpc } from 'pm2-axon-rpc';
```
shuse2, line 18: I think these import should not be changed. This shouldnt be affected by types?

PR #6085, opened 2021-01-19
Create oclif commands skeleton in the template - Closes #6075

### LINKED ISSUE #6075: Create oclif commands skeleton in the template
### Description

The template we develop should export `oclif` commands which are available in `lisk-core`. In implementation all the command should be developed in a utility package and then imported to the template. 

### Motivation

This will able user to run those commands e.g. `/run start` with all possible flags, options we use in `lisk-core`. This will also enable use to fix any bug in the utility package to cover up any issue in such commands. User will also be able to just remove the default code and make their own implementation of the command. 

### Acceptance Criteria

- Every command added to template should just be an import and export not a real implementation 
- The test skeleton should also be generated for `account:create`
- All the bootstrap commands to be exported as abstract class

### Additional Information

- For reference we can use the `lisk-core` commands and use `lisk-commander` as library for their reference

ManuGowda:
### What was the problem?

This PR resolves #6075 

### How was it solved?

- Add base commands to commander
- Add commands to `lisk-ts` template

### How was it tested?

1. `cd lisk-sdk`; `yarn; yarn build`
2. `cd commander; yarn link`
3. `mkdir /tmp/test_lisk; cd /tmp/test_lisk`
4. `../lisk-sdk/commander/bin/run init`
5. `yarn link lisk-commander`
6. `./bin/run`
7, `npm run test`

### DIFF
--- a/commander/package.json
+++ b/commander/package.json
@@ -43,6 +43,9 @@
 		"prepack": "oclif-dev manifest && npm shrinkwrap && npm prune --production && npm shrinkwrap",
 		"prepublishOnly": "rm -r ./node_modules && npm install && npm run lint && npm run build"
 	},
+	"lisk": {
+		"addressPrefix": "lsk"
+	},
 	"oclif": {
 		"commands": "./dist/commands",
 		"bin": "lisk",
@@ -82,6 +85,7 @@
 		"/docs"
 	],
 	"dependencies": {
+		"lisk-sdk": "5.0.3",
 		"@liskhq/lisk-cryptography": "^3.0.0",
 		"@liskhq/lisk-passphrase": "^3.0.1",
 		"@liskhq/lisk-validator": "^0.5.0",
@@ -90,6 +94,7 @@
 		"@oclif/errors": "1.2.2",
 		"@oclif/plugin-autocomplete": "0.2.1",
 		"@oclif/plugin-help": "2.2.3",
+		"axios": "0.21.1",
 		"bip39": "3.0.2",
 		"chalk": "3.0.0",
 		"cli-table3": "0.5.1",
--- a/commander/package.json
+++ b/commander/package.json
@@ -87,6 +87,7 @@
 		"@liskhq/lisk-passphrase": "^3.0.1",
 		"@liskhq/lisk-validator": "^0.5.0",
 		"@oclif/command": "1.5.19",
+		"@oclif/parser": "3.8.5",
 		"@oclif/config": "1.14.0",
 		"@oclif/errors": "1.2.2",
 		"@oclif/plugin-autocomplete": "0.2.1",
--- a/commander/package.json
+++ b/commander/package.json
@@ -82,10 +82,15 @@
 		"/docs"
 	],
 	"dependencies": {
-		"lisk-sdk": "5.0.3",
+		"lisk-framework": "0.7.3",
+		"@liskhq/lisk-api-client": "5.0.2",
+		"@liskhq/lisk-codec": "0.1.0",
 		"@liskhq/lisk-cryptography": "^3.0.0",
+		"@liskhq/lisk-db": "0.1.0",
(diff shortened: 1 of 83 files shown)


commander/src/bootstrapping/templates/lisk-template-ts/templates/app/src/commands/account/show.ts
```diff
@@ -0,0 +1 @@
+export { AccountShowCommand } from 'lisk-commander';
```
shuse2, line 1: For all the commands, it should be either consistently expose as `default` or as object.
ie) some file exposes as `default class` but some as object

commander/src/bootstrapping/templates/lisk-template-ts/templates/app/src/commands/start.ts
```diff
@@ -0,0 +1,132 @@
...
+		// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+		config.plugins[MonitorPlugin.alias] = config.plugins[MonitorPlugin.alias] ?? {};
+		config.plugins[MonitorPlugin.alias].whiteList = flags['monitor-plugin-whitelist']
+			.split(',')
+			.filter(Boolean);
+	}
+};
+
+export class StartCommand extends BaseStartCommand {
+	static flags = {
```
shuse2, line 55: Failing on build here

shuse2 (changes requested):
generally LGTM

commander/src/bootstrapping/commands/blockchain/hash.ts
```diff
@@ -0,0 +1,80 @@
...
+
+	static examples = ['blockchain:hash', 'blockchain:hash --data-path ./data'];
+
+	static flags = {
+		'data-path': flagParser.string({
+			char: 'd',
+			description:
+				'Directory path to specify where node data is stored. Environment variable "LISK_DATA_PATH" can also be used.',
+			env: 'LISK_DATA_PATH',
+		}),
```
nazarhussain, line 33: There is a `data-path` in every other command. Would not it be nice to export such flags independently and reuse. Something like below. 

import commonFlags from '../common_flags`

```
static flags = {
     commonFlags.dataPath
}
```

commander/src/bootstrapping/commands/start.ts
```diff
@@ -0,0 +1,235 @@
...
+			description:
+				'Seed peers to initially connect to in format of comma separated "ip:port". IP can be DNS name or IPV4 format. Environment variable "LISK_SEED_PEERS" can also be used.',
+		}),
+	};
+
+	// eslint-disable-next-line @typescript-eslint/require-await
+	async run(): Promise<void> {
+		const { flags } = this.parse(StartCommand);
+		const dataPath = flags['data-path'] ? flags['data-path'] : getDefaultPath();
+		this.log(`Starting Lisk Core at ${getFullPath(dataPath)}.`);
```
nazarhussain, line 114:
```suggestion
		this.log(`Starting Lisk application at ${getFullPath(dataPath)}.`);
```

nazarhussain (changes requested):
The `dataPath` logic in all the commands is not compatible with custom app. It always point to `~/.lisk/lisk-core` which is wrong in reference to custom app. Hence every command using `dataPath` in one or other way is failing.

ManuGowda:
@nazarhussain i have created a separate issue to address the improvements https://github.com/LiskHQ/lisk-sdk/issues/6087

commander/package.json
```diff
@@ -82,14 +82,17 @@
 		"/docs"
 	],
 	"dependencies": {
+		"lisk-sdk": "5.0.3",
```
shuse2, line 85: prefer to use each element or framework package from `lisk-commander`

shuse2 (approved):

nazarhussain (approved):

PR #6090, opened 2021-01-25
Create module bootstrapping - Closes #6070

### LINKED ISSUE #6070: Create module bootstrapping generator
### Description

This generator will bootstrap a module and all its dependencies. The interface of the generator should be: 

```
lisk generate:module <name> <id> 
```

### Motivation

Use should be able to generate a skeleton of the module quickly. So he can explore the features which are provided with on-chain architecture. 

### Acceptance Criteria

- The generated module should be under `modules` directory. 
- The generated module should also be registered to the application 
- The module directory skeleton should be following: 

```
   src/
      modules.ts  <-- require modules and register 
      modules/
        module_name/
          module_name.ts  <-- Module class default export 

```
- It should also generate the relevant test skeleton
- It should detected existing module with same name and id warn user before overwriting

### Additional Information

- The module source code should include every life cycle hooks along with sample code but everything should be commented. 
- It should have a sample array of events but commented 
- It should have a sample actions object but commented 
- It should have set of sample reducers but commented 
- It should have an empty array of assets, which will later be updated when new asset is generated

ishantiw:
### What was the problem?

This PR resolves #6070 

### How was it solved?

🌱 Add template for modules

🌱 Add generate:module command 

🌱 Add module generator files

✅ Add test skeleton for module

✅ Add test for `generate:module` command

### How was it tested?

Run `lisk generate:module nft 5` in the myapp directory <path> created from `lisk init <path>`


commander/src/bootstrapping/templates/lisk-template-ts/templates/app/test/unit/modules/module.ts
```diff
@@ -0,0 +1,32 @@
...
+	describe('constructor', () => {
+		it.todo('should have valid id');
+		it.todo('should have valid name');
+	});
+
+	describe('beforeBlockApply', () => {});
+	describe('afterBlockApply', () => {});
+	describe('beforeTransactionApply', () => {});
+	describe('afterTransactionApply', () => {});
+	describe('afterGenesisBlockApply', () => {});
```
nazarhussain, line 31: These empty blocks is causing linting error in the generated app. Either we add some implementation or remove empty functions in definitions.

nazarhussain (changes requested):

commander/src/commands/generate/module.ts
```diff
@@ -0,0 +1,68 @@
...
+			regexCamelCase.test(moduleName) ||
+			regexWhitespace.test(moduleName) ||
+			regexAlphabets.test(moduleName)
+		) {
+			this.error('Invalid module name');
+		}
+
+		if (Number.isNaN(Number(moduleID))) {
+			this.error('Invalid module ID');
+		}
```
nazarhussain, line 56: Also add a check that moduleID must be greater than equal to `MINIMUM_EXTERNAL_MODULE_ID` defined in SDK.

nazarhussain (changes requested):

commander/src/bootstrapping/templates/lisk-template-ts/templates/app/test/unit/modules/module.ts
ishantiw: ts-morph lib for import statement uses double quotes, `import { Nft } from "./modules/nft";` this is the only lint error which is not fixed

nazarhussain (approved):

commander/src/bootstrapping/templates/lisk-template-ts/templates/modules/test/module.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ *
+ */
+
+// import * as modules from '../../../src/app/modules'
```
mitsuaki-u, line 17: since this file is for specific module, shouldn't this import the correct module using template variable?

commander/src/commands/generate/module.ts
```diff
@@ -0,0 +1,70 @@
...
+		if (
+			regexCamelCase.test(moduleName) ||
+			regexWhitespace.test(moduleName) ||
+			regexAlphabets.test(moduleName)
+		) {
+			this.error('Invalid module name');
+		}
+
+		if (Number.isNaN(Number(moduleID)) || +moduleID < MINIMUM_EXTERNAL_MODULE_ID) {
+			this.error('Invalid module ID');
```
mitsuaki-u, line 57: Should specify a more descriptive error message here as user may not know they need to use an ID number higher than `MINIMUM_EXTERNAL_MODULE_ID`

mitsuaki-u (commented):
Some minor comments

commander/src/bootstrapping/templates/lisk-template-ts/generators/module_generator.ts
```diff
@@ -0,0 +1,109 @@
...
+
+		registerFunction.setBodyText(
+			`${registerFunction.getBodyText()} _app.registerModule(${this._moduleClass});`,
+		);
+
+		modulesFile.organizeImports();
+		await modulesFile.save();
+	}
+
+	public createModuleUnitTest() {
```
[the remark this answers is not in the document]
ishantiw: Similar to what we have for registering module, `registerModule()` we have separated writing for tests `createModuleUnitTest()`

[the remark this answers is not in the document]
ishantiw: moved it within `writing()`

mitsuaki-u (approved):

PR #6100, opened 2021-01-28
Create network config - Closes #6092

### LINKED ISSUE #6092: Update generate:network-config to be in application command
### Description

Update commander command `generate:network-config` to be in the application command.

### Motivation

- Since we moved the genesis block to be in the application command, the related config generation also should be in the application command

### Acceptance Criteria

- Add output flag to specify the location of the config stored, and default to the current working directory
- Argument should be removed, but flag should stay the same
- command name should be `myapp config:create`
- It should be added to the template

Incede:
### What was the problem?

This PR resolves #6092 

### How was it solved?

* Move config creation to application from commander

### How was it tested?

Tested manually after linking `lisk-commander` in application `./bin/run config:create`


commander/src/bootstrapping/commands/config/create.ts
```diff
@@ -17,24 +17,22 @@ import { Command, flags as flagParser } from '@oclif/command';
...
-	];
-	static args = [
-		{
-			name: 'networkName',
-			description: 'Directory where the config file is saved.',
-			required: true,
-		},
+		'generate:network-config --networkName mydir',
+		'generate:network-config --networkName mydir --label alpha-sdk-app',
+		'generate:network-config --networkName mydir --label alpha-sdk-app --community-identifier sdk',
```
ManuGowda, line 27:
```suggestion
		'generate:network-config --'network-name' mydir',
		'generate:network-config --'network-name' mydir --label alpha-sdk-app',
		'generate:network-config --'network-name' mydir --label alpha-sdk-app --community-identifier sdk',
```

commander/src/bootstrapping/commands/config/create.ts
```diff
@@ -17,24 +17,22 @@ import { Command, flags as flagParser } from '@oclif/command';
...
 	];
 
 	static flags = {
+		networkName: flagParser.string({
+			char: 'n',
+			description: 'Directory where the config file is saved',
+			default: '',
+		}),
```
ManuGowda, line 35:
```suggestion
		'network-name': flagParser.string({
			char: 'n',
			description: 'Name for the network directory where the config file is saved.',
			default: '',
		}),
```

commander/src/bootstrapping/commands/config/create.ts
```diff
@@ -49,10 +47,9 @@ export default class NetworkConfigCommand extends Command {
 
 	async run(): Promise<void> {
 		const {
-			flags: { label, 'community-identifier': communityIdentifier },
+			flags: { networkName, label, 'community-identifier': communityIdentifier },
```
ManuGowda, line 50:
```suggestion
			flags: { 'network-name': networkName, label, 'community-identifier': communityIdentifier },
```

commander/src/bootstrapping/commands/config/create.ts
```diff
@@ -17,24 +17,22 @@ import { Command, flags as flagParser } from '@oclif/command';
...
-			description: 'Directory where the config file is saved.',
-			required: true,
-		},
+		'generate:network-config --network-name mydir',
+		'generate:network-config --network-name mydir --label alpha-sdk-app',
+		'generate:network-config --network-name mydir --label alpha-sdk-app --community-identifier sdk',
 	];
 
 	static flags = {
+		'network-name': flagParser.string({
```
shuse2, line 31: i think in other places we use `network`

commander/src/bootstrapping/commands/config/create.ts
```diff
@@ -17,24 +17,22 @@ import { Command, flags as flagParser } from '@oclif/command';
...
-	];
-	static args = [
-		{
-			name: 'networkName',
-			description: 'Directory where the config file is saved.',
-			required: true,
-		},
+		'generate:network-config --network-name mydir',
+		'generate:network-config --network-name mydir --label alpha-sdk-app',
+		'generate:network-config --network-name mydir --label alpha-sdk-app --community-identifier sdk',
```
nazarhussain, line 27: Officially we are in `beta` stage now. Suggest to not add `alpha-` prefix anywhere.

commander/src/bootstrapping/commands/config/create.ts
nazarhussain, line 31: Having a required flag is not a good practice. Suggest to use use arguments instead.

commander/src/bootstrapping/commands/config/create.ts
```diff
@@ -17,24 +17,22 @@ import { Command, flags as flagParser } from '@oclif/command';
...
 
 	static flags = {
+		'network-name': flagParser.string({
+			char: 'n',
+			description: 'Name of the network directory where the config file is saved',
+			default: '',
+		}),
 		label: flagParser.string({
 			char: 'l',
 			description: 'App Label',
```
nazarhussain, line 38: When we bootstrap app the app-lable default value is the directory name. Later that also been added to `package.json`. Is there a way to check if `cwd` contains a `package.json` then use its package name as default value.

commander/src/bootstrapping/commands/config/create.ts
```diff
@@ -78,12 +75,11 @@ export default class NetworkConfigCommand extends Command {
...
 				this.error('Operation cancelled, config file already present at the desired location');
 			} else {
-				fs.writeJSONSync(resolve(configPath, 'config.json'), JSON.stringify(defaultConfig));
+				fs.writeJSONSync(resolve(configPath, 'config.json'), defaultConfig, { spaces: '' });
 			}
 		} else {
 			fs.mkdirSync(configPath, { recursive: true });
-			// write config in config.json at the proper path
-			fs.writeJSONSync(resolve(configPath, 'config.json'), JSON.stringify(defaultConfig));
+			fs.writeJSONSync(resolve(configPath, 'config.json'), defaultConfig, { spaces: '' });
```
nazarhussain, line 82: Try sing `'\t'` as spaces, that will result in formatted in more readable and formatted file not the one-liner json.

nazarhussain (changes requested):

commander/src/bootstrapping/commands/config/create.ts
nazarhussain, line 31: Should not we use `--output`, `--output-dir` rather network name here? as its just pointing to any directory, which have no link with term network.

nazarhussain (changes requested):

commander/src/bootstrapping/commands/config/create.ts
Incede: I think @shuse2 mentioned in the issue to use as flag not argument

ManuGowda (approved):

nazarhussain (approved):

PR #6104, opened 2021-02-02
Update API client functions - Closes #6099

### LINKED ISSUE #6099: Update api client functions to accept hex string
### Description

Currently, most of the function that requires binary input in APIClient accepts only Buffer, but it should also accept hex string

```
client.block.get
client.block.decode
client.account.get
client.account.decode
client.transaction.get
client.transaction.decode
```

### Motivation

- There were misuse of function reported frequently
- when `subscribe` to event, binary data is returned as hex string

### Acceptance Criteria

- Unit test added
- If input is neither Buffer nor hex string, it should throw an error

Incede:
### What was the problem?

This PR resolves #6099 

### How was it solved?

* Accept hex string as input for `get` & `decode` functions

### How was it tested?

* Added unit tests


elements/lisk-api-client/test/unit/account.spec.ts
```diff
@@ -65,7 +65,7 @@ describe('account', () => {
 			});
 		});
 
-		describe('get', () => {
+		describe('get from address as buffer', () => {
```
ManuGowda, line 68: probably we can structure it like
```
describe('get', () => {
   describe('account by address as string', () => {
      it('should invoke app:getAccount', async () => {});
   });
   
   describe('account by address as buffer', () => {
      it('should invoke app:getAccount', async () => {});
   });
});
```
?
if yes, then same comment for below tests as well.

ManuGowda (commented):
LGTM, some minor suggestion

elements/lisk-api-client/test/unit/account.spec.ts
```diff
@@ -66,21 +66,42 @@ describe('account', () => {
...
-			it('should invoke app:getAccount', async () => {
-				// Arrange
-				when(channel.invoke as any)
-					.calledWith('app:getAccount', { address: '9d0149b0962d44bfc08a9f64d5afceb6281d7fb5' })
-					.mockResolvedValue(encodedAccountHex as never);
+			describe('account by address as buffer', () => {
+				it('should invoke app:getAccount', async () => {
+					// Arrange
+					when(channel.invoke as any)
+						.calledWith('app:getAccount', { address: '9d0149b0962d44bfc08a9f64d5afceb6281d7fb5' })
```
ishantiw, line 73: shouldn't address be passed as a buffer in this test case?

elements/lisk-api-client/test/unit/block.spec.ts
```diff
@@ -130,14 +130,29 @@ describe('block', () => {
...
+					// Act
+					await block.get(blockId);
 
-				// Assert
-				expect(channel.invoke).toHaveBeenCalledTimes(1);
-				expect(channel.invoke).toHaveBeenCalledWith('app:getBlockByID', {
-					id: blockId.toString('hex'),
+					// Assert
+					expect(channel.invoke).toHaveBeenCalledTimes(1);
+					expect(channel.invoke).toHaveBeenCalledWith('app:getBlockByID', {
```
ishantiw, line 140: same here, `id` should be a buffer?

ishantiw (commented):
A few comments and Lisk Framework Plugins Functional tests are failing

ishantiw (commented):
oh ignore my comments, I see that its always converted to hex before calling the channel, only tests are not passing

elements/lisk-api-client/test/unit/account.spec.ts
ishantiw: channel is always called with address hex string so ignore the comment

ManuGowda (approved):

ishantiw (approved):

PR #6110, opened 2021-02-03
Generate initial genesis block and config with init command - Closes #6095

### LINKED ISSUE #6095: Generate initial genesis block and config with init command
### Description
- When creating a new application, there should be `config` folder and by default `mainnet`
- `mainnet` should contain `genesis_block.json` and `config.json`
- all the auxiliary files should be output into `myapp/.secrets/mainnet(network)/xxx`

### Motivation

After using init command, the application should be runnable without modification

### Acceptance Criteria

- After calling `./bin/run init ~/lisk-sample`, `./bin/run (myapp) start` should work

### Additional Information
- `init` can assume the application has only `default` modules are registered

ishantiw:
### What was the problem?

This PR resolves #6095 

### How was it solved?

- ♻️ Extract genesis creation logic out of create command 80b8447
- ♻️ Generate genesis block and config on running init 362aa80

### How was it tested?

- Generate your app using `./bin/run init /User/xyz/myapp`
- `cd /User/xyz/myapp`
- `<path_to_lisk_commander>/bin/run init`

**Note:** In order to link local repos, add below code to `init_generator.ts` in beginning of `end()` method,

```
this.spawnCommandSync(`yarn`, [
			'link',
			'lisk-commander',
		]);
this.spawnCommandSync(`yarn`, [
			'link',
			'lisk-framework',
		]);
```

### DIFF
--- a/commander/src/bootstrapping/templates/lisk-template-ts/generators/init_generator.ts
+++ b/commander/src/bootstrapping/templates/lisk-template-ts/generators/init_generator.ts
@@ -16,6 +16,7 @@
 
 import { userInfo } from 'os';
 import { basename, join } from 'path';
+import fs from 'fs';
 
 import * as Generator from 'yeoman-generator';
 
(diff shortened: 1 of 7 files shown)


shuse2 (approved):

commander/src/bootstrapping/templates/lisk-template-ts/generators/init_generator.ts
```diff
@@ -73,4 +74,33 @@ export default class InitGenerator extends Generator {
...
+
+		fs.renameSync(
+			`${this.destinationRoot()}/config/mainnet/accounts.json`,
+			`${this.destinationRoot()}/.secrets/mainnet/accounts.json`,
+		);
+
+		fs.renameSync(
+			`${this.destinationRoot()}/config/mainnet/forging_info.json`,
+			`${this.destinationRoot()}/.secrets/mainnet/forging_info.json`,
+		);
```
nazarhussain, line 101: Use `this.destinationPath` instead of `this.destinationRoot` with string concatenation. e.g. `this. destinationPath('bin/run')` is equal to `'${this.destinationRoot()}/bin/run'`

nazarhussain (changes requested):

Incede (approved):
Maybe indicate about `.secrets` in the log.
Everything else LGTM!

PR #6113, opened 2021-02-05
Add access to logger to module and plugins - Closes #6097

### LINKED ISSUE #6097: Add access to logger to module and plugins
### Description

Add logger in the modules and plugins.

### Motivation

When writing a module or plugin, often required to add some logs. However, currently it is not provided.

### Acceptance Criteria

- Logger is usable in module (thus, asset have also access if needed)
- Logger is usable in plugin which runs on memory
- Logger is usable in plugin which runs on child process
- Add unit tests to check if the logger is passed correctly

### Additional Information
- Plugin must work in both in memory and in child process, therefore, the logger must take that in account.
- Logger can be passed as part of the context, but since it does not change, it will probably better to pass in as parameter of constructor

nazarhussain:
### What was the problem?

This PR resolves #6097

### How was it solved?

- Added logger support for module and plugin as protected member. 
- Logger will be available for both after `init` process 

### How was it tested?

- Unit tests for framework

### DIFF
--- a/framework-plugins/lisk-framework-report-misbehavior-plugin/src/report_misbehavior_plugin.ts
+++ b/framework-plugins/lisk-framework-report-misbehavior-plugin/src/report_misbehavior_plugin.ts
@@ -30,7 +30,6 @@ import {
 	PluginInfo,
 } from 'lisk-framework';
 import { objects } from '@liskhq/lisk-utils';
-import * as createDebug from 'debug';
 import {
 	getDBInstance,
 	saveBlockHeaders,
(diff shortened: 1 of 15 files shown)


ManuGowda (commented):
Some of the logs in `report_misbehavior_plugin.ts` still use debug, shouldn't they become info and error accordingly?

shuse2 (approved):

nazarhussain:
@ManuGowda I am sure we can refactor those well. But those were all initially done by `debug` package used for debugging, so I kept those as debug level. If you suggest I can update those in this PR.

ManuGowda (approved):

PR #6144, opened 2021-02-09
Update commander start command - Closes #6143

### LINKED ISSUE #6143: Custom "rpc" config values are not respected with commander start command
### Expected behavior
`lisk start --config /path/to/custom_config` should respect `rpc` properties defined in custom config

### Actual behavior
The values defined for `rpc` properties in custom config are ignored.

### Steps to reproduce
Define config.json as
```
{
"rpc": {
    "port": "5678"
    "host": "127.0.0.3"
}
```
Start node with WS server using this config.json `lisk start --api-ws --config /path/to/custom_config`
Check on which port & from which IP addresses the WS server can accept requests & send responses.
5.*

Incede:
### What was the problem?

This PR resolves #6143 

### How was it solved?
* Use rpc properties from config file in start command

### How was it tested?

* Tested manually by using custom rpc config values to run a node & sending requests to it from the custom port & address.


ManuGowda (approved):

commander/src/bootstrapping/commands/start.ts
```diff
@@ -170,10 +170,17 @@ export abstract class StartCommand extends Command {
...
+				enable: flags['api-ipc'],
+				mode: 'ipc',
+			});
 		}
 		if (flags['api-ws']) {
-			config.rpc = { enable: flags['api-ws'], mode: 'ws', port: flags['api-ws-port'] };
+			config.rpc = utils.objects.mergeDeep({}, config.rpc, {
+				enable: flags['api-ws'],
+				mode: 'ws',
+				port: flags['api-ws-port'],
```
nazarhussain, line 182: What if user only provided `--api-ws` flag and not `--api-ws-port`? I am not sure what would be behaviour of the config if `port` is specified as undefined. So may may avoid adding port to config unless if specified.

nazarhussain (changes requested):

commander/src/bootstrapping/commands/start.ts
shuse2: In that case, `flags['api-ws-port']` will be undefined and use the default value?

nazarhussain: I tested `lisk-utils` behaviour for `mergeDeep` using `undefined` and `null` values and it works.

nazarhussain (approved):

PR #6165, opened 2021-02-11
Remove blockchain download command - Closes #6164

### LINKED ISSUE #6164: Unintended blockchain:download command for bootstrapped application
### Expected behavior

Being Lisk platform, we are not providing snapshot hosting service for custom applications. So as we bootstrapped application it should have either of two behaviours: 

1. It should have not have a `blockchain:download` command completely 
2. Or it should have a configureable endpoint from where to download 

And without setting a custom endpoints to download the snapshot the default command should throw error. 

### Actual behavior

For any custom bootstrapped application it downloads the snapshot from the Lisk Mainnet. 

```
> ./bin/run blockchain:download
Downloading snapshot from https://downloads.lisk.io/lisk/mainnet/blockchain.db.tar.gz to ~/lisk-app
    Error: Request failed with status code 404
```

5.0.x

shuse2:
### What was the problem?

This PR resolves #6164 

### How was it solved?

- Remove download command since it's specific to Lisk-core

### How was it tested?

n/a


ManuGowda (commented):
you need to remove the index.js file and extended command in template

nazarhussain (changes requested):
LGTM other than what @ManuGowda  highlighted.

ManuGowda (approved):
maybe you need to remove from here too `commander/src/index.ts`

nazarhussain (approved):

PR #6168, opened 2021-02-12
Fix generate plugin initial setting - Closes #6154

### LINKED ISSUE #6154: Lisk Generate Plugin prompts for unwanted options with no defaults
### Expected behavior

The plugin generator should only ask for the `alias` and use `package.json` to fill-in rest of the data. Additional information needed for a plugin needed is: 

1. Author
2. Version 
3. Name of the plugin 

When new plugin is generated its safe to just populate version with `0.1.0` version. Name of the plugin can be simply the alias itself and author name can be picked from `package.json. 

### Actual behavior

Current generator asks for these three options and two of these without any default. 

```
> /lisk-sdk/commander/bin/run generate:plugin MyPlugin 
Using template "lisk-ts"
Generating plugin skeleton
? Author of plugin 
? Version of plugin 0.1.0
? Name of plugin 
Registering plugin...
identical .liskrc.json
   create src/app/plugins/MyPlugin/MyPlugin.ts
   create test/unit/plugins/MyPlugin/MyPlugin.spec.ts
```

The result of which end up having: 

```js
       public static get info(): PluginInfo {
		return {
			author: '',
			version: '0.1.0',
			name: '',
		};
	}
```
This is invalid for a plugin and when we start the app it fails with following error. 

```
Starting Lisk lisk-app at /Users/nazar/.lisk/lisk-app.
    AssertionError: Plugin name is required.
    Code: ERR_ASSERTION
```

So either we remove all prompts (recommended) and use default values as described in the expected behaviour. Or use default values for all prompts so plugin would not have empty string values. 

### Steps to reproduce

```
> /lisk-sdk/commander/bin/run generate:plugin MyPlugin 
```

5.0.x

shuse2:
### What was the problem?

This PR resolves #6154

### How was it solved?

- Fix path to the package.json
- Update default name to be alias

### How was it tested?

- Call `commander/bin/run generate:plugin http`

### DIFF
--- a/commander/src/bootstrapping/templates/lisk-template-ts/generators/plugin_generator.ts
+++ b/commander/src/bootstrapping/templates/lisk-template-ts/generators/plugin_generator.ts
@@ -46,7 +46,7 @@ export default class PluginGenerator extends Generator {
 		// Check for existing package.json in root directory to use existing info
 		try {
 			// eslint-disable-next-line
-			this._packageJSON = await import(`${this.destinationRoot()}/package.json`);
+			this._packageJSON = await import(this.destinationPath('package.json'));
 		} catch (err) {
 			this._packageJSON = undefined;
 		}
--- a/commander/src/bootstrapping/templates/lisk-template-ts/generators/plugin_generator.ts
+++ b/commander/src/bootstrapping/templates/lisk-template-ts/generators/plugin_generator.ts
@@ -46,7 +46,7 @@ export default class PluginGenerator extends Generator {
 		// Check for existing package.json in root directory to use existing info
 		try {
 			// eslint-disable-next-line
-			this._packageJSON = await import(`${this.destinationRoot()}/src/app/package.json`);
+			this._packageJSON = await import(`${this.destinationRoot()}/package.json`);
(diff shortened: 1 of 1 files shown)


nazarhussain (approved):

commander/src/bootstrapping/templates/lisk-template-ts/generators/plugin_generator.ts
```diff
@@ -46,7 +46,7 @@ export default class PluginGenerator extends Generator {
 		// Check for existing package.json in root directory to use existing info
 		try {
 			// eslint-disable-next-line
-			this._packageJSON = await import(`${this.destinationRoot()}/src/app/package.json`);
+			this._packageJSON = await import(`${this.destinationRoot()}/package.json`);
```
nazarhussain, line 49: This can be shorthanded by: 

```suggestion
			this._packageJSON = await import(this.destinationPath('package.json'));
```

nazarhussain (changes requested):

nazarhussain (approved):

mitsuaki-u (approved):

Issue #6162, opened 2021-02-11
Transaction create command --network flag not failing on unknown network


nazarhussain:
### Expected behavior

Transaction create command in bootstrap application shows a supported flag `--network` which is also supported by `LISK_NETWORK` 

```
> ./bin/run transaction:create --help 
Create transaction which can be broadcasted to the network. Note: fee and amount should be in Beddows!!

USAGE
  $ lisk-app transaction:create MODULEID ASSETID FEE

OPTIONS

  -n, --network=network
      [default: mainnet] Default network config to use. Environment variable "LISK_NETWORK" can 
      also be used.
```

As per description when `--network` flag is specified use a particular network configuration.  

### Actual behavior

With `--network` flag specified or `LISK_NETWORK` it never checks it value. Means providing an invalid network does not fail the command and it tries to connect to the `mainnet` always. 

### Steps to reproduce

Simplest way to reproduce. 

```
 > ./bin/run transaction:create 2 0 1000000 --network unkown
```

### Which version(s) does this affect? (Environment, OS, etc...)
5.0.x

PR #6169, opened 2021-02-12
Transaction create command with no signature failed - Closes #6160, #6161, #6162

### LINKED ISSUE #6160: Transaction create with offline mode is not working
### Expected behavior

Transaction create command in bootstrap application shows a supported flag `--offline` 

```
> ./bin/run transaction:create --help 
Create transaction which can be broadcasted to the network. Note: fee and amount should be in Beddows!!

USAGE
  $ lisk-app transaction:create MODULEID ASSETID FEE

OPTIONS

  --offline
      Specify whether to connect to a local node or not.
```

As per description when `--offline` flag is specified it should not connect to the application. 

### Actual behavior

With with `--offline` flag it tries to connect to node and failed. 

```
    Error: Application at data path ~/.lisk/lisk-app is not running.
```

### Steps to reproduce

Simplest way to reproduce. 

```
> ./bin/run transaction:create 2 0 1000000 --offline
```

5.0.x

### LINKED ISSUE #6161: Transaction create command with no signature failed
### Expected behavior

Transaction create command in bootstrap application shows a supported flag ` --no-signature`

```
> ./bin/run transaction:create --help 
Create transaction which can be broadcasted to the network. Note: fee and amount should be in Beddows!!

USAGE
  $ lisk-app transaction:create MODULEID ASSETID FEE

OPTIONS

  --no-signature
      Creates the transaction without a signature. Your passphrase will therefore not be 
      required
```

As per description when ` --no-signature` flag is specified it should ask for the signtaure.

### Actual behavior

When this flag is specified it failed with error. 

```
 ›   Error: Cannot read property 'toString' of undefined
```

### Steps to reproduce

```
> ./bin/run transaction:create 2 0 1000000 --no-signature 
```

5.0.x

nazarhussain:
### What was the problem?

This PR resolves  #6160, #6161, #6162

### How was it solved?

- Refactored transaction creation to use oclif flag dependencies 
- When no-signature is specified use the public key to get the address

### How was it tested?

- Run unit tests
- Follow the steps defined in the issue

### DIFF
--- a/commander/src/bootstrapping/commands/transaction/create.ts
+++ b/commander/src/bootstrapping/commands/transaction/create.ts
@@ -31,7 +31,7 @@ import {
 	getAssetSchema,
 	transactionToJSON,
 } from '../../../utils/transaction';
-import { getDefaultPath, getGenesisBlockAndConfig } from '../../../utils/path';
+import { getDefaultPath } from '../../../utils/path';
 import { isApplicationRunning } from '../../../utils/application';
 import { PromiseResolvedType } from '../../../types';
 
@@ -42,7 +42,6 @@ interface Args {
 }
 
 interface CreateFlags {
-	network: string;
 	'network-identifier'?: string;
 	passphrase?: string;
 	asset?: string;
@@ -241,7 +240,6 @@ export abstract class CreateCommand extends Command {
 	];
 
 	static flags = {
-		network: flagsWithParser.network,
 		passphrase: flagsWithParser.passphrase,
 		asset: flagParser.string({
 			char: 'a',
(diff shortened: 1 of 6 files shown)


shuse2 (approved):

commander/src/bootstrapping/commands/transaction/create.ts
```diff
@@ -300,18 +291,23 @@ export abstract class CreateCommand extends Command {
...
 				args as Args,
 				flags,
 				this._schema,
 				incompleteTransaction,
 			);
 		} else {
-			this._client = await getApiClient(flags['data-path'] as string, this.config.pjson.name);
+			if (!isApplicationRunning(this._dataPath)) {
+				throw new Error(`Application at data path ${this._dataPath} is not running.`);
+			}
```
ManuGowda, line 306: `await getApiClient` already checks for app running, why do we need this extra check?

ManuGowda (approved):

PR #6170, opened 2021-02-15
Transaction create command failed on empty data attribute - Closes #6158

### LINKED ISSUE #6158: Transaction create command failed on empty data attribute
### Expected behavior

Transaction create command should work fine with an empty data field. 

### Actual behavior

For an empty data field the transaction creation failed. 

```
> ./bin/run transaction:create 2 0 1000000
? Please enter passphrase:  [hidden]
? Please re-enter passphrase:  [hidden]
? Please enter: amount:  10
? Please enter: recipientAddress:  7294d56fa05293e11086a86f0c1c42e5a9dce21e
? Please enter: data:  
 ›   Error: Lisk validator found 1 error[s]:
 ›   Property '.data' should pass "dataType" keyword validation
```

5.0.x

ManuGowda:
### What was the problem?

This PR resolves #6158 

### How was it solved?

- Fix transaction create asset 

### How was it tested?

- Updated unit test

### DIFF
--- a/commander/src/utils/reader.ts
+++ b/commander/src/utils/reader.ts
@@ -236,14 +236,20 @@ const getNestedPropertyTemplate = (schema: Schema): NestedPropertyTemplate => {
 };
 
 const castValue = (
-	strVal: string,
+	val: string,
 	schemaType: string,
-): number | string | Record<string, unknown> => {
+): number | bigint | string | string[] | Record<string, unknown> => {
 	if (schemaType === 'object') {
 		// eslint-disable-next-line @typescript-eslint/no-unsafe-return
-		return JSON.parse(strVal);
+		return JSON.parse(val);
 	}
-	return Number.isInteger(Number(strVal)) ? Number(strVal) : strVal;
+	if (schemaType === 'array') {
+		return val.split(',');
+	}
+	if (Number.isInteger(Number(val))) {
+		return schemaType === 'uint64' ? BigInt(val) : Number(val);
+	}
+	return val;
 };
 
 export const transformAsset = (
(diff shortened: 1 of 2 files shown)


commander/src/utils/reader.ts
```diff
@@ -236,14 +236,20 @@ const getNestedPropertyTemplate = (schema: Schema): NestedPropertyTemplate => {
...
 	if (schemaType === 'object') {
 		// eslint-disable-next-line @typescript-eslint/no-unsafe-return
-		return JSON.parse(strVal);
+		return JSON.parse(val);
 	}
-	return Number.isInteger(Number(strVal)) ? Number(strVal) : strVal;
+	if (schemaType === 'array') {
+		return val.split(',');
+	}
+	if (Number.isInteger(Number(val))) {
```
nazarhussain, line 249: What if the `val` here is a an integer string but schemaType suggest that dataType should be string? I suggest to check `schemaType` not that value is integer or not.

nazarhussain (changes requested):

commander/src/utils/reader.ts
```diff
@@ -236,14 +236,25 @@ const getNestedPropertyTemplate = (schema: Schema): NestedPropertyTemplate => {
...
+	}
+	if (
+		schemaType === 'uint64' ||
+		schemaType === 'sint64' ||
+		schemaType === 'uint32' ||
+		schemaType === 'sint32'
+	) {
+		return schemaType === 'uint64' ? BigInt(val) : Number(val);
+	}
+	return val;
```
nazarhussain, line 257: This missed the condition for `sint64` type. Would be nice to split this condition to two for for bigint and number.

nazarhussain (changes requested):

nazarhussain (approved):

shuse2 (approved):

PR #6179, opened 2021-02-17
Add createTransaction in lisk-framework - Closes #6123

### LINKED ISSUE #6123: Add createTransaction in lisk-framework
### Description

- Add createTransaction to generate transaction based on a asset definition
```
const tx = createTransaction({ 
  moduleID: number,
  assetClass: Asset,
  asset: {},
  networkIdentifier: Buffer,
  passphrase?: string,
  nonce?: bigint,
  fee?: bigint,
});
```

### Motivation

When testing block processing or hooks, it requires a transaction. Therefore, there should be an easy way to generate any custom transaction for testing

### Acceptance Criteria

- Transaction should be statically valid if passphrase is provided
- If passphrase does not exist, it should not sign the transaction
- Multi-signature should not be considered (since it's module)

mitsuaki-u:
### What was the problem?

This PR resolves #6123

### How was it solved?

- Add function `createTransaction` for framework testing
- Add related unit tests
- Added needed dependency to framework

### How was it tested?

- Run unit tests


elements/lisk-chain/src/index.ts
```diff
@@ -17,7 +17,7 @@ const events = { EVENT_DELETE_BLOCK, EVENT_NEW_BLOCK, EVENT_VALIDATORS_CHANGED }
 
 export { events };
 export { Chain } from './chain';
-export { Transaction, transactionSchema, calculateMinFee } from './transaction';
+export { Transaction, TransactionInput, transactionSchema, calculateMinFee } from './transaction';
```
ManuGowda, line 20: why do we need to export the input `TransactionInput`? i do not see it being used anywhere, can we not export if it's not used?

framework/src/testing/create_transaction.ts
```diff
@@ -0,0 +1,71 @@
...
+import { AssetClass } from './types';
+
+interface CreateTransactionInput {
+	moduleID: number;
+	assetClass: AssetClass;
+	asset: Record<string, unknown>;
+	nonce?: bigint;
+	fee?: bigint;
+	passphrase?: string;
+	networkIdentifier?: Buffer;
```
ManuGowda, line 29: networkIdentifier is required?

framework/test/unit/testing/create_transaction.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+import { TransferAsset } from '../../../src/modules/token';
+import { createTransaction } from '../../../src/testing';
+
+describe('Create Transaction', () => {
+	const asset = {
+		id: 0,
```
ManuGowda, line 20: 🤔 asset shouldn't have id?

ManuGowda (commented):
LGTM, minor comments

framework/src/testing/create_transaction.ts
mitsuaki-u: It is only required if user wants to sign transaction. Since we are making that optional, it is not required.

elements/lisk-chain/src/index.ts
mitsuaki-u: Sure I originally was using it but I updated the code to not use it. Will un-export.

framework/test/unit/testing/create_transaction.spec.ts
mitsuaki-u: All asset have an `id`

ManuGowda (approved):

framework/src/testing/create_transaction.ts
```diff
@@ -0,0 +1,71 @@
...
+import { Transaction } from '@liskhq/lisk-chain';
+import { codec } from '@liskhq/lisk-codec';
+import { getAddressAndPublicKeyFromPassphrase } from '@liskhq/lisk-cryptography';
+import { signTransaction, validateTransaction } from '@liskhq/lisk-transactions';
+import { AssetClass } from './types';
+
+interface CreateTransactionInput {
+	moduleID: number;
+	assetClass: AssetClass;
+	asset: Record<string, unknown>;
```
nazarhussain, line 25: Some transactions can be without any assets object, so make this as optional.

framework/src/testing/create_transaction.ts
```diff
@@ -0,0 +1,71 @@
...
+	}
+
+	if (!passphrase) {
+		return transaction;
+	}
+
+	return (signTransaction(
+		assetInstance.schema,
+		{ ...transaction, asset },
+		networkIdentifier ?? Buffer.alloc(1),
```
nazarhussain, line 68: Why its `Buffer.alloc(1)` and not empty buffer?

framework/test/unit/testing/create_transaction.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+describe('Create Transaction', () => {
+	const asset = {
+		amount: BigInt('1000000000000'),
+		recipientAddress: Buffer.from('ab0041a7d3f7b2c290b5b834d46bdc7b7eb85815', 'hex'),
+		data: 'moon',
+	};
+
+	it('should return a valid transaction', () => {
+		expect(
+			createTransaction({ moduleID: 2, assetClass: TransferAsset as any, asset }),
```
nazarhussain, line 27: Why we need to pass `as any` here? If types are not matching that means all the SDK users has to use `as any` or `as never` in their tests. We should avoid that and fix the type.

nazarhussain (changes requested):

framework/src/testing/create_transaction.ts
mitsuaki-u: Keeping as is since in most cases transaction will have an asset. User can provide empty object if not.

framework/src/testing/create_transaction.ts
mitsuaki-u: `signTransaction` internally checks for the length of the buffer and therefore it would not sign if used with `Buffer.alloc(0`. To avoid confusion I will simply add a check if user provided `networkIdentifier` rather than setting a default.

framework/test/unit/testing/create_transaction.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+	};
+
+	it('should return a valid transaction', () => {
+		expect(createTransaction({ moduleID: 2, assetClass: TransferAsset, asset })).toMatchSnapshot();
+	});
+
+	it('should return valid signed transaction with passphrase', () => {
+		const transaction = createTransaction({
+			moduleID: 2,
+			assetClass: TransferAsset as any,
```
nazarhussain, line 32:
```suggestion
			assetClass: TransferAsset,
```

nazarhussain (changes requested):

nazarhussain (approved):

PR #6183, opened 2021-02-22
Add waitUntilHeight in lisk-framework - Closes #6124

### LINKED ISSUE #6124: Add waitUntilBlockHeight in lisk-framework
### Description

Add waitUntilBlockHeight function as utility
```
await waitUntilBlockHeight({apiClient: APIClient, height: number, timeout?: number }) => Promise<void>;
```

### Motivation

When testing block processing with application, it is often required a way to wait until the block is processed

### Acceptance Criteria

- If `timeout` is not specified, it should wait until the block height is reached 
- If user provides a `timeout` parameter, they should calculate the estimated amount of time required to reach the specified `height`(e.g. for `height` of 2, since default `blockTime` is ~10 seconds, `timeout` parameter should be at least `20000` ms. If the chain does not reach the specified `height` within the given `timeout`, the function will reject with an error.

mitsuaki-u:
### What was the problem?

This PR resolves #6124 

### How was it solved?

- Moved `lisk-api` from dev-dependency to dependency. This is done for testing and type purposes. 
- Added util function `waitUntilHeight`

### How was it tested?

Run unit tests for `waitUntilHeight`

### DIFF
--- a/framework/test/unit/testing/utils.spec.ts
+++ b/framework/test/unit/testing/utils.spec.ts
@@ -11,6 +11,8 @@
  *
  * Removal or modification of this copyright notice is prohibited.
  */
+
+import { APIClient } from '@liskhq/lisk-api-client';
 import { SequenceModule, TokenModule } from '../../../src';
 import {
 	getAccountSchemaFromModules,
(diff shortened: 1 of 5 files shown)


framework/test/unit/testing/utils.spec.ts
```diff
@@ -12,8 +12,13 @@
...
 import { SequenceModule, TokenModule } from '../../../src';
-import { getAccountSchemaFromModules, getModuleInstance } from '../../../src/testing/utils';
+import {
+	getAccountSchemaFromModules,
+	getModuleInstance,
+	waitUntilBlockHeight,
+} from '../../../src/testing/utils';
 
 describe('utils', () => {
 	describe('getAccountSchemaFromModules()', () => {
```
ManuGowda, line 24:
```suggestion
	describe('getAccountSchemaFromModules', () => {
```

ManuGowda (approved):

Incede (approved):

PR #6189, opened 2021-03-01
Add getBlockProcessingEnv in lisk-framework - Closes #6119

### LINKED ISSUE #6119: Add getBlockProcessingEnv in lisk-framework
### Description

- Add getBlockProcessingEnv which helps block processing

```
const env = getBlockProcessingEnv(Module[], {
  genesisConfig?: GenesisConfig,
  databasePath?: string,
});
```

It should have property to have
```
process(block)
processUntilHeight(101)
getStateStore
```

### Motivation

- When testing the module logics in integrated environment, block processing is required

### Acceptance Criteria

- It should use the real block processing including database
- It should initialise the processor from framework

### Additional Information

- We have similar functionality in framework integration test
- This will be used for integration test
- `getStateStore` should return real stateStore for assertion

ManuGowda:
### What was the problem?

This PR resolves #6119 

### How was it solved?

- Added test util for `getBlockProcessingEnv`

### How was it tested?

- Added integration tests


framework/src/testing/fixtures/genesis_block.ts
```diff
@@ -35,14 +35,17 @@ export const createGenesisBlockWithAccounts = <T = AccountDefaultProps>(
 		delegate =>
 			(({
 				...delegate,
-				address: Buffer.from(delegate.address, 'hex'),
+				address: delegate.address,
 			} as unknown) as PartialAccount<T>),
 	);
+	// Set genesis block timestamp to 2 months in past relative to current date
+	const timestamp = Math.floor(new Date().setMonth(new Date().getMonth() - 2) / 1000);
```
shuse2, line 42: wouldnt 2 month too much? if i want to make a block on top of the block, it will have 2 month worth of the missed blocks?

framework/test/integration/testing/block_processing_env.spec.ts
```diff
@@ -0,0 +1,77 @@
+/*
```
shuse2, line 1: i think we forgot to create issue for upgrading the integration tests to use this new function. We should test it if all the interface are usable at least for us

shuse2 (commented):
generally LGTM

framework/test/integration/testing/block_processing_env.spec.ts
ManuGowda: https://github.com/LiskHQ/lisk-sdk/issues/6203

shuse2 (approved):

Incede (approved):

PR #6194, opened 2021-03-01
Update monitor plugin to use lisk-framework - Closes #6131

### LINKED ISSUE #6131: Update monitor plugin to use lisk-framework
### Description

Update monitor plugin tests to use created testing utilities.

### Motivation

- All testing utilities should be used in the framework to test the usability of the functions

### Acceptance Criteria

- Plugin functional tests are using the utility functions created

ishantiw:
### What was the problem?

This PR resolves #6131

### How was it solved?

♻️ Use testing in monitor plugin from framework #6131 9b10acc

### How was it tested?

`npm run test`

### DIFF
--- a/framework-plugins/lisk-framework-monitor-plugin/test/unit/forks.spec.ts
+++ b/framework-plugins/lisk-framework-monitor-plugin/test/unit/forks.spec.ts
@@ -12,33 +12,22 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
-import { blockHeaderSchema, blockSchema } from '@liskhq/lisk-chain';
+import { blockHeaderSchema, blockSchema, RawBlock } from '@liskhq/lisk-chain';
 import { codec } from '@liskhq/lisk-codec';
 import { hash } from '@liskhq/lisk-cryptography';
-import { RawBlock } from '@liskhq/lisk-chain';
+import { testing } from 'lisk-framework';
 import { MonitorPlugin } from '../../src/monitor_plugin';
 
 describe('_handleFork', () => {
 	let monitorPluginInstance: MonitorPlugin;
 	let encodedBlock: string;
-	const channelMock = {
-		registerToBus: jest.fn(),
-		once: jest.fn(),
-		publish: jest.fn(),
-		subscribe: jest.fn(),
-		isValidEventName: jest.fn(),
-		isValidActionName: jest.fn(),
-		invoke: jest.fn(),
-		eventsList: [],
-		actionsList: [],
-		actions: {},
-		moduleAlias: '',
-		options: {},
-	} as any;
+	const {
+		mocks: { channelMock },
+	} = testing;
 
 	beforeEach(async () => {
 		monitorPluginInstance = new (MonitorPlugin as any)();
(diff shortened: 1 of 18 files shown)


nazarhussain (approved):

framework-plugins/lisk-framework-monitor-plugin/test/unit/forks.spec.ts
```diff
@@ -12,33 +12,22 @@
...
-		options: {},
-	} as any;
+	const {
+		mocks: { channelMock },
+	} = testing;
 
 	beforeEach(async () => {
 		monitorPluginInstance = new (MonitorPlugin as any)();
-		await monitorPluginInstance.load(channelMock);
+		await monitorPluginInstance.load(channelMock as any);
```
shuse2, line 30: why do we need `any` here?

framework/test/utils/node/state_store_mock.ts
```diff
@@ -71,9 +71,9 @@ export class StateStoreMock {
 				const index = this.accountData.findIndex(acc => acc.address.equals(address));
 				if (index > -1) {
 					this.accountData[index] = account;
-					return;
```
shuse2, line 74: i think this `return` was correct?

framework/test/utils/node/state_store_mock.ts
```diff
@@ -85,6 +85,7 @@ export class StateStoreMock {
 			// eslint-disable-next-line @typescript-eslint/require-await
 			set: async (key: string, value: Buffer): Promise<void> => {
 				this.chainData[key] = value;
+				return Promise.resolve();
```
shuse2, line 88: if it's `async` there is no need to explicitly return resolve?

framework/test/utils/node/state_store_mock.ts
```diff
@@ -71,9 +71,9 @@ export class StateStoreMock {
 				const index = this.accountData.findIndex(acc => acc.address.equals(address));
 				if (index > -1) {
 					this.accountData[index] = account;
-					return;
 				}
 				this.accountData.push(account);
+				return Promise.resolve();
```
shuse2, line 76: same as below

framework-plugins/lisk-framework-monitor-plugin/test/unit/forks.spec.ts
ishantiw: channelMock is just an object and TS expects and throws `is not a class derived from 'BaseChannel'`. So either we can create a channel mock of a derived from BaseChannel or typecast to any

ishantiw: channelMock is just an object and TS expects and throws `is not a class derived from 'BaseChannel'`. So either we can create a channel mock of a derived from BaseChannel or typecast to any

framework/test/utils/node/state_store_mock.ts
ishantiw: It was throwing error for reason earlier after rebasing but now its not, i'll keep it to return

shuse2 (approved):

PR #6215, opened 2021-03-10
Codec validateSchema should check for the field number existence - Closes #6191

### LINKED ISSUE #6191: Codec validateSchema should check for the field number existence
### Expected behavior
Schema which do not have field number should fail to add and compile

### Actual behavior
Related to #6190 , field number existence should be checked before adding schema and compiling the schema.

### Steps to reproduce
Try to compile schema without field number

5.0.3-

Incede:
### What was the problem?

This PR resolves #6191 

### How was it solved?

* Check properties in schema to include `fieldNumber`

### How was it tested?

* Tested manually using custom schema in test cases

### DIFF
--- a/elements/lisk-codec/package.json
+++ b/elements/lisk-codec/package.json
@@ -54,6 +54,7 @@
 		"jest": "26.4.2",
 		"jest-extended": "0.11.5",
 		"jest-when": "2.7.2",
+		"lodash.clonedeep": "4.5.0",
 		"prettier": "2.0.5",
 		"source-map-support": "0.5.19",
 		"ts-jest": "26.3.0",
--- a/elements/lisk-codec/package.json
+++ b/elements/lisk-codec/package.json
@@ -54,7 +54,6 @@
 		"jest": "26.4.2",
 		"jest-extended": "0.11.5",
 		"jest-when": "2.7.2",
-		"lodash.clonedeep": "4.5.0",
 		"prettier": "2.0.5",
 		"source-map-support": "0.5.19",
 		"ts-jest": "26.3.0",
--- a/elements/lisk-codec/src/codec.ts
+++ b/elements/lisk-codec/src/codec.ts
@@ -67,15 +67,6 @@ export const validateSchema = (schema: {
 		throw new LiskValidationError([...errors]);
 	}
 
-	const { properties } = schemaToValidate as ValidatedSchema;
-	if (properties) {
-		Object.values(properties).forEach(property => {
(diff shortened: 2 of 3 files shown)


elements/lisk-codec/src/codec.ts
```diff
@@ -67,6 +67,15 @@ export const validateSchema = (schema: {
 		throw new LiskValidationError([...errors]);
 	}
 
+	const { properties } = schemaToValidate as ValidatedSchema;
+	if (properties) {
+		Object.values(properties).forEach(property => {
```
shuse2, line 72: Prefer to use `for...of` as it is slow although this only runs once so it would not make big difference

elements/lisk-codec/src/codec.ts
```diff
@@ -67,6 +67,15 @@ export const validateSchema = (schema: {
...
+			if (!('fieldNumber' in property)) {
```
shuse2, line 73: field number can exist in the nested properties, too. 
Did you consider updating `elements/lisk-validator/src/lisk_meta_schema.ts` which is executed in above L63?

shuse2 (changes requested):

elements/lisk-codec/src/codec.ts
```diff
@@ -163,13 +163,21 @@ export class Codec {
 			if (properties === undefined) {
 				throw new Error('Invalid schema. Missing "properties" property');
 			}
+			for (const property of Object.values(properties)) {
+				if (!('fieldNumber' in property)) {
```
shuse2, line 167: I think we want to check `fieldNumber` exists and the value in the range 1 to 18’999. (https://github.com/LiskHQ/lips/blob/master/proposals/lip-0027.md#field-name-and-field-number)

shuse2 (changes requested):
Please add unit tests to check if the root and nested field number check is done correctly

ishantiw (approved):

elements/lisk-codec/package.json
```diff
@@ -54,6 +54,7 @@
 		"jest": "26.4.2",
 		"jest-extended": "0.11.5",
 		"jest-when": "2.7.2",
+		"lodash.clonedeep": "4.5.0",
```
shuse2, line 57: Please use cloneDeep from `@liskhq/lisk-utils`

shuse2 (changes requested):
other than the dependency usage LGTM

shuse2 (approved):

PR #6275, opened 2021-03-24
Copiable text component - Closes #6224

### LINKED ISSUE #6224: Create react component for copiable text
### Description

Create component to show an text with copy icon with properties
- onClick: () => void
- text

It should copy the input text to the clipboard

### Acceptance Criteria

- Use Text component
- Use Icon component
- Icon should be displayed only on hover
- After click, it should show `Copied` with checkmark icon, and remove it after 1s
- Timeout cleanup should be handled properly

### Additional Information
Related Issues: #6216

ishantiw:
### What was the problem?

This PR resolves #6224 

### How was it solved?

- Use Text component
- Use Icon component
- copy to clipboard handling
- setTimeout to original text

### How was it tested?

`npm run start:web`


framework-plugins/lisk-framework-dashboard-plugin/src/ui/components/CopyText/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+ * Removal or modification of this copyright notice is prohibited.
+ */
+import * as React from 'react';
+import styles from './CopyText.module.scss';
+import Icon from '../Icon';
+import Text from '../Text';
+
+interface Props {
+	name: string;
+	size?: 's' | 'm' | 'l' | 'xl';
```
shuse2, line 21: these property i think you can export Props from Text and reference rather than redefining

framework-plugins/lisk-framework-dashboard-plugin/src/ui/components/CopyText/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+}
+
+const CopyText: React.FC<Props> = (props: Props) => {
+	const [hover, setHover] = React.useState(true);
+	const [name, setName] = React.useState(props.name);
+
+	const clipToClipboard = async (text: string) => {
+		setHover(false);
+		setName('Copied');
+		setTimeout(() => {
```
shuse2, line 33: i think the timeout id need to be cleaned up in case of the component being removed?

framework-plugins/lisk-framework-dashboard-plugin/src/ui/components/CopyText/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+	type?: 'h1' | 'h2' | 'h3' | 'th' | 'tr' | 'p';
+}
+
+const CopyText: React.FC<Props> = (props: Props) => {
+	const [hover, setHover] = React.useState(true);
+	const [name, setName] = React.useState(props.name);
+
+	const clipToClipboard = async (text: string) => {
+		setHover(false);
+		setName('Copied');
```
shuse2, line 32: while displaying the copied, icon should be hidden

framework-plugins/lisk-framework-dashboard-plugin/src/ui/components/CopyText/CopyText.module.scss
```diff
@@ -0,0 +1,11 @@
+.clickable {
+	cursor: pointer;
```
shuse2, line 2: this is not applied

framework-plugins/lisk-framework-dashboard-plugin/src/ui/components/CopyText/CopyText.module.scss
```diff
@@ -0,0 +1,11 @@
...
+}
+
+.clickableRow {
+	float: left;
```
shuse2, line 6: you can use `flex` here

shuse2 (changes requested):

framework-plugins/lisk-framework-dashboard-plugin/src/ui/components/CopyText/index.tsx
```diff
@@ -0,0 +1,61 @@
...
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+import * as React from 'react';
+import styles from './CopyText.module.scss';
+import Icon from '../Icon';
+import Text from '../Text';
+
+interface Props {
+	name: string;
```
nazarhussain, line 20: `name` prop does not looks nice for a `CopyableText`. Logically this component should just use `children` prop to render any content. If not then use some better name e.g. `text`, `content`.

framework-plugins/lisk-framework-dashboard-plugin/src/ui/components/CopyText/index.tsx
nazarhussain: Yes it should be exported from `Text/index.tsx` and then union those types here.

nazarhussain (changes requested):

framework-plugins/lisk-framework-dashboard-plugin/src/ui/components/CopyText/CopyText.module.scss
ishantiw: when using flex, Text and Icon components don't align side by side and icon goes to the next line

shuse2: the flex needs to be added to the container

nazarhussain (approved):

shuse2 (approved):

PR #6286, opened 2021-03-26
Remove unnecessary testing utils from plugins - Closes #6211

### LINKED ISSUE #6211: Remove unnecessary testing utils from plugins
### Description

- Remove the util functions in `plugin/test/utils` or `plugin/test/functional/utils` such as `createApplicationEnv` or `closeApplicationEnv`.
If the plugin needs specific setup, (for example, needs certain modules and accounts to exist), create `createDefaultApplicationEnv` which takes `ApplicationEnvConfig` in `framework/testing`, but all default modules are registered, and initial accounts (faucet account) having balance.
In this way, if developers are using `defaultApplication`, they can add their module and plugins, and use the testing function without needing to add much of configuration.

- https://github.com/LiskHQ/lisk-sdk/blob/5e0758bbf7818b1341f9c81a6a91b2c9588d817d/framework-plugins/lisk-framework-forger-plugin/test/utils/application.ts#L86 can be removed since it is mitigated in the validator now

### Motivation

After creating the testing tools, most of the application related utility functions are unnecessary.

### Acceptance Criteria

- All the necessary function should be provided by the `framework/testing` (ie: no duplicate or similar function across plugin tests)
- Only special case which are used in the specific plugin should exist in the util

ManuGowda:
### What was the problem?

This PR resolves #6211

### How was it solved?

- Created util `createDefaultApplicationEnv` for plugin functional tests
- Cleaned up plugin functional tests

### How was it tested?

- Updated tests


framework/src/testing/app_env.ts
```diff
@@ -123,3 +131,60 @@ export class ApplicationEnv {
 		return application;
 	}
 }
+
+export const createDefaultApplicationEnv = (
+	appEnvConfig: Partial<ApplicationEnvConfig>,
```
shuse2, line 136: i think extra module should be addable as optional here, otherwise it's not useful to others

framework/src/testing/app_env.ts
```diff
@@ -123,3 +131,60 @@ export class ApplicationEnv {
...
+	const config = objects.mergeDeep(
+		{},
+		{ ...defaultConfig },
+		{
+			...appEnvConfig.config,
+			logger: {
+				consoleLogLevel: 'fatal',
+				fileLogLevel: 'fatal',
+				logFileName: 'lisk.log',
+			},
```
shuse2, line 147: can we put his setting as default for the underline app env?

shuse2 (changes requested):
LGTM. just some minor points

mitsuaki-u (approved):
LGTM besides other comments

shuse2 (approved):

PR #6296, opened 2021-04-01
add wsPort to app:getNodeInfo

JesusTheHun:
### What was the problem?

Issue #6297 `Add wsPort to app:getNodeInfo`

### How was it solved?

By adding code

### How was it tested?

Updated functional test

### DIFF
--- a/framework/src/node/node.ts
+++ b/framework/src/node/node.ts
@@ -507,7 +507,9 @@ export class Node {
 					...this._options.genesisConfig,
 				},
 				registeredModules: this.getRegisteredModules(),
-				wsPort: this._options.network.port,
+				network: {
+					port: this._options.network.port,
+				}
 			}),
 			getConnectedPeers: () => this._networkModule.getConnectedPeers(),
 			getDisconnectedPeers: () => this._networkModule.getDisconnectedPeers(),
--- a/framework/src/node/node.ts
+++ b/framework/src/node/node.ts
@@ -509,7 +509,7 @@ export class Node {
 				registeredModules: this.getRegisteredModules(),
 				network: {
 					port: this._options.network.port,
-				}
+				},
 			}),
 			getConnectedPeers: () => this._networkModule.getConnectedPeers(),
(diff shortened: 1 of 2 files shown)


ManuGowda:
@JesusTheHun thank you for creating the pull request, each pull request need to be created against an issue/bug, could you please create issue with details and then associate the issue with PR? 
Thank you 🙏🏼

JesusTheHun:
@ManuGowda it's done !

framework/src/node/node.ts
```diff
@@ -507,6 +507,7 @@ export class Node {
 					...this._options.genesisConfig,
 				},
 				registeredModules: this.getRegisteredModules(),
+				wsPort: this._options.network.port,
```
shuse2, line 510: i would prefer to have 
```
{
  port: this._options.network.port
}
```
so that we can add more network information if required, and it is consistent with the config structure

shuse2 (changes requested):

framework/src/node/node.ts
JesusTheHun: Do you mean

1. 
```ts
{
  ...
  network: {
    port: port: this._options.network.port
  }
  ...
}
```

2.
```ts
{
  ...
  port: port: this._options.network.port
  ...
}
```

Pick one 😄 
I like the 1. better, it was suggested by @nazarhussain on discord. It might feel empty right now but at least there is context. One may imagine having a chain augmenting the nodeInfo that would conflict in the absence of a namespace for network props.

shuse2: ah sorry i meant option `1`!

shuse2 (approved):

nazarhussain:
@JesusTheHun Can you please check why CI server is failing this PR on formatting step. Probably a comma is missing after the last `}` you added.

nazarhussain:
@JesusTheHun The PR seems fine, please fix the failing tests so we can merge it.

nazarhussain (approved):

PR #6311, opened 2021-04-08
Create react component for info panel - Closes #6229

### LINKED ISSUE #6229: Create react component for small widget
### Description

Create component to show an small widget on header with properties
- title
- onClick?: () => void;
- color

### Acceptance Criteria

- Use SCSS constant
- Add sample to the main page for testing

### Additional Information
Related Issues: #6216

Incede:
### What was the problem?

This PR resolves #6229

### How was it solved?

* Added new React component InfoPanel
* Extended widget properties

### How was it tested?

* Added sample to main page


ManuGowda (approved):
LGTM, made few suggestions

framework-plugins/lisk-framework-dashboard-plugin/src/ui/components/InfoPanel/InfoPanel.module.scss
```diff
@@ -0,0 +1,59 @@
...
+	height: 20px;
+	margin-bottom: 10px;
+
+	/* identical to box height */
+
+	/* Platinum gray */
+	color: #dfe6f2;
+}
+
+.infoContent {
```
shuse2, line 20: the `Text` component should be used so that it doesn't require the font settings

framework-plugins/lisk-framework-dashboard-plugin/src/ui/components/InfoPanel/InfoPanel.module.scss
```diff
@@ -0,0 +1,59 @@
+@import '../../styles/constants';
+
+.infoHeading {
```
shuse2, line 3: This is also not required if the Text component is used

framework-plugins/lisk-framework-dashboard-plugin/src/ui/components/InfoPanel/index.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
+import { Widget, WidgetBody } from '../widget';
+import Icon from '../Icon';
+import styles from './InfoPanel.module.scss';
+
+interface Props {
+	onClick?: (event: React.MouseEvent | Event) => void;
+	color?: 'green' | 'pink' | 'yellow' | 'blue' | 'white' | 'gray' | 'red';
```
shuse2, line 8: you can use the interface exposed from Text

framework-plugins/lisk-framework-dashboard-plugin/src/ui/components/widget/Widget.module.scss
```diff
@@ -34,6 +33,18 @@
...
+	background: $color-background-darker;
+}
+
+.body-light {
+	background: $color-ink-blue;
+}
+
+.alpha {
+	background: red;
+}
```
shuse2, line 46: do we need this?

shuse2 (changes requested):

framework-plugins/lisk-framework-dashboard-plugin/src/ui/components/InfoPanel/InfoPanel.module.scss
Incede: The Text component does not support all the styles required here

framework-plugins/lisk-framework-dashboard-plugin/src/ui/components/InfoPanel/InfoPanel.module.scss
Incede: same here. The Text component should be extended to be used here

framework-plugins/lisk-framework-dashboard-plugin/src/ui/components/InfoPanel/index.tsx
Incede: yes but I think we still need to use `div` to support all the classes and props OR `Text` component can be extended if that's okay ?

shuse2: What i meant was you can do like https://github.com/LiskHQ/lisk-sdk/blob/feature/6216-add_dashboard-plugin/framework-plugins/lisk-framework-dashboard-plugin/src/ui/components/CopiableText/index.tsx#L21-L23
or simply import and do
```
color?: TextProps['color'],
```

framework-plugins/lisk-framework-dashboard-plugin/src/ui/components/InfoPanel/InfoPanel.module.scss
shuse2: im not sure what do u mean by extending here, but 
```
{props.title}
```
this can be just
```
<Text type="tr" color={props.color}>{props.title}</Text>
```
(type tr probably better to be renamed in other issue)

framework-plugins/lisk-framework-dashboard-plugin/src/ui/components/InfoPanel/InfoPanel.module.scss
shuse2: Actually, you are using {props.children}, so no need to apply the text style here. the user from the upper later will specify the style using `Text`

framework-plugins/lisk-framework-dashboard-plugin/src/ui/components/InfoPanel/InfoPanel.module.scss
Incede: This doesn't work because in the "InfoPanel" with `icon` "margins" are changed and it is not styled similar to other panels.

shuse2: you can keep the margins for outer layer, but the Text related styles can be just removed from here as the responsibility is in `Text` Component

shuse2 (approved):

PR #6319, opened 2021-04-13
Fix updated deepscan

shuse2:
### What was the problem?

FIx deepscan new issue


shuse2:
Integrated into #6320

PR #6330, opened 2021-04-20
Update AJV to 8.1.0 - Closes #6329

### LINKED ISSUE #6329: Update AJV to 8.1.0
### Description
Update ajv to latest 8.1.0.

There are a lot of breaking changes. Some of the changes observed was
- dataPath is now instancePath with `/` separated instead of `.`
- addKeyword have different interface
- `must` is used instead of `should` in error message
- `id` cannot be used anymore. it must be `$id`
- meta schema `format` keyword does not work

Official migration documentation is 
- https://ajv.js.org/v6-to-v8-migration.html#potential-migration-difficulties

Initial tried changes are here
- https://github.com/LiskHQ/lisk-sdk/commit/9d25bb35dd2bc1f1c7e0ba3b7c79e71390ac8758

Also,
```
addUsedSchema: false,
```
Check this property added to the Ajv is still necessary or not.

nazarhussain:
### What was the problem?

This PR resolves #6329

### How was it solved?

- Upgraded the ajv package 
- Update lisk-validator to be compatible with Ajv
- Update test expectations for newer error messages 

### How was it tested?

- Run all tests


elements/lisk-validator/src/types.ts
```diff
@@ -0,0 +1,22 @@
...
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+import { ErrorObject } from 'ajv';
+
+export { DataValidateFunction, DataValidationCxt } from 'ajv/dist/types';
```
shuse2, line 17: Prefer not to introduce types from non exported files

elements/lisk-validator/src/keywords/field_number.ts
```diff
@@ -48,26 +42,22 @@ const deepValue = (obj: object, path: string): any => {
...
-
 const compile = (
-	value: number,
-	parentSchema: object,
-	it: Partial<AjvContext>,
-): ValidateFunction => {
+	value: JSONSchemaType<number>,
+	parentSchema: AnySchemaObject,
+	it: SchemaCxt,
+): DataValidateFunction => {
```
shuse2, line 49: I think no need to define the return type if the type is not exposed

protocol-specs/generators/lisk_codec/types_generators/transaction.js
```diff
@@ -35,7 +35,7 @@ const baseTransactionSchema = {
 		asset: { dataType: 'bytes', fieldNumber: 6 },
 		signatures: { type: 'array', items: { dataType: 'bytes' }, fieldNumber: 7 },
 	},
-	required: ['type', 'nonce', 'fee', 'senderPublicKey', 'asset'],
+	required: ['moduleID', 'assetID', 'nonce', 'fee', 'senderPublicKey', 'asset', 'signatures'],
```
shuse2, line 38: why adding the requirement here?

shuse2 (changes requested):

elements/lisk-validator/src/types.ts
nazarhussain: Yes we should not, but that's what they described in their docs. 

https://ajv.js.org/json-type-definition.html

elements/lisk-validator/src/keywords/field_number.ts
nazarhussain: Tried that it was complaining about not matching type with compile function.

protocol-specs/generators/lisk_codec/types_generators/transaction.js
nazarhussain: The actual error was, the `type` property was mentioned in `required` but its definition in `properties` were missing. Also noticed few of the properties which have defined `fieldNumber` are not mentioned in `required`. So its better the validation to fail than codec.

elements/lisk-validator/src/types.ts
shuse2: ah ok! then it should be fine. They will probably manage the breaking change properly if it's in the doc

elements/lisk-validator/src/keywords/field_number.ts
shuse2: if above is type def can be used, this isn't a problem

protocol-specs/generators/lisk_codec/types_generators/transaction.js
shuse2: probably not a problem in this case but it's possible to have property with fieldNumber which is not required

shuse2 (approved):

ishantiw (approved):

PR #6331, opened 2021-04-21
Update copiable text - Closes #6291

### LINKED ISSUE #6291: Update CopiableText component to fix layout issue with longer text
### Expected behavior

The layout and text should not break when hover the copiable text which have longer text the container. 

### Actual behavior

Currently when we hover the text the text wraps down and provide space for the copy icon. Which breaks the layout a bit. 

Normal
image

Hover
image

### Additional Information 
- The existing content wrapping is not working on Safari browser. 

5.1.x

ishantiw:
### What was the problem?

This PR resolves #6291 

### How was it solved?

- Fix max width of copiable text to avoid flicker while copying

### How was it tested?

`npm run start:web`

### DIFF
--- a/framework-plugins/lisk-framework-dashboard-plugin/src/ui/components/CopiableText/CopiableText.module.scss
+++ b/framework-plugins/lisk-framework-dashboard-plugin/src/ui/components/CopiableText/CopiableText.module.scss
@@ -1,15 +1,17 @@
 .clickable {
 	cursor: pointer;
 	margin-left: 8px;
+	max-width: 5%;
 }
 
 .clickableRow {
 	padding: 0.5px;
-	max-width: 500px;
 	min-height: 20px;
+	max-width: 95%;
 }
 
 .clickableContainer {
+	min-width: 100%;
 	display: inline-flex;
 	align-items: center;
 }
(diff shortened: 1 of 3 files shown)


shuse2 (changes requested):
Behavior change looks good.
I think there are 8px space between icon and text is missing so maybe we can fix it in this PR?

framework-plugins/lisk-framework-dashboard-plugin/src/ui/components/CopiableText/CopiableText.module.scss
```diff
@@ -1,10 +1,11 @@
 .clickable {
 	cursor: pointer;
+	margin-left: 8px;
 }
 
 .clickableRow {
 	padding: 0.5px;
-	height: auto;
+	max-width: 500px;
```
nazarhussain, line 8: We can't assign max-width inside the component. It must inherit the width of the container. If container is 600px and content is also long it should just stretch to available width.

nazarhussain (changes requested):

framework-plugins/lisk-framework-dashboard-plugin/src/ui/components/CopiableText/CopiableText.module.scss
ishantiw: the problem is that if we keep width relative to parent container, it always wraps the text in a way that on hover it flickers

ishantiw: I have given max-width as 95% of the container and 5% to icon so that they have their own space within the container and also the width can vary with the width of parent container

shuse2 (approved):

nazarhussain (approved):

PR #6358, opened 2021-04-27
Forging fails when genesis block timestamp is in the future - Closes #6356

### LINKED ISSUE #6356: Forging fails when genesis block timestamp is in the future
### Expected behavior

When we migrate Lisk Core from 2.x.x to 3.x.x, the blockchain should be able to start considering genesis block timestamp in future.

More details can be found in LIP: https://github.com/LiskHQ/lips/blob/master/proposals/lip-0035.md#general-block-properties

### Actual behavior

When the genesis block timestamp is in the future(10sec, 1hr, 2hr, 1day, etc..), the application fails to forge blocks.
```
12:14:01 INFO lisk-framework: Node ready and launched (module=lisk:app)
12:14:02 ERROR lisk-framework: Cannot read property 'address' of undefined (module=lisk:app)
Message: Cannot read property 'address' of undefined
 Trace: TypeError: Cannot read property 'address' of undefined
    at Forger.forge (lisk-sdk/framework/src/node/forger/forger.ts:387:57)
    at Node._forgingTask (lisk-sdk/framework/src/node/node.ts:680:4)
    at Timeout._onTimeout (lisk-sdk/elements/lisk-utils/src/job_handlers/scheduler.ts:47:5)
```

### Steps to reproduce

- Create timestamp with(2hrs) in the future
- Update devnet genesis block `timestamp` field with the generated value
- Start the application and observe the error

5.x.x, development

ManuGowda:
### What was the problem?

This PR resolves #6356

### How was it solved?

- Check for validator existence
 
### How was it tested?

- Added unit test
- Tested manually by updating devnet genesis block timestamp to future(100sec)

### DIFF
--- a/framework/src/node/forger/forger.ts
+++ b/framework/src/node/forger/forger.ts
@@ -385,7 +385,7 @@ export class Forger {
 
 		const validator = await this._chainModule.getValidator(currentTime);
 		if (!validator) {
-			this._logger.trace(
+			this._logger.debug(
 				{ currentSlot: this._chainModule.slots.getSlotNumber() },
 				'No validator is set for current time slot',
 			);
@@ -394,7 +394,7 @@ export class Forger {
 		const validatorKeypair = this._keypairs.get(validator.address);
 
 		if (validatorKeypair === undefined) {
-			this._logger.trace(
+			this._logger.debug(
 				{ currentSlot: this._chainModule.slots.getSlotNumber() },
 				'Waiting for delegate slot',
 			);
--- a/framework/src/node/forger/forger.ts
+++ b/framework/src/node/forger/forger.ts
@@ -387,7 +387,7 @@ export class Forger {
 		if (!validator) {
 			this._logger.trace(
 				{ currentSlot: this._chainModule.slots.getSlotNumber() },
-				'Waiting for delegate slot',
+				'No validator is set for current time slot',
 			);
 			return;
 		}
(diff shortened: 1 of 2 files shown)


framework/src/node/forger/forger.ts
```diff
@@ -384,6 +384,13 @@ export class Forger {
 		}
 
 		const validator = await this._chainModule.getValidator(currentTime);
+		if (!validator) {
+			this._logger.trace(
+				{ currentSlot: this._chainModule.slots.getSlotNumber() },
+				'Waiting for delegate slot',
```
shuse2, line 390: maybe something like `No validator is set for current time slot`?

framework/test/unit/node/forger/forger.spec.ts
```diff
@@ -928,6 +928,38 @@ describe('forger', () => {
...
+				when(getSlotNumberStub)
+					.calledWith(lastBlock.header.timestamp)
+					.mockReturnValue(changedLastBlockSlot);
+
+				chainModuleStub.slots.getSlotTime.mockReturnValue(futureSlotTime);
+
+				const data = await forgeModule.forge();
+
+				expect(data).toBeUndefined();
+				expect(loggerStub.trace).toHaveBeenCalledTimes(1);
```
shuse2, line 955: I think it's better not to check the log. It's better to check if the block is not created and no error is thrown?

shuse2 (approved):

framework/src/node/forger/forger.ts
```diff
@@ -384,6 +384,13 @@ export class Forger {
 		}
 
 		const validator = await this._chainModule.getValidator(currentTime);
+		if (!validator) {
+			this._logger.trace(
```
nazarhussain, line 388: Is not this information is better at `debug` level?

nazarhussain (approved):

PR #6364, opened 2021-04-30
Fix unstable application url fetch - Closes #6362

### LINKED ISSUE #6362: Invalid application URL resolution outside development environment
### Expected behavior
Fetch to `/api/config` and access to `applicationUrl` within the `getApplicationUrl()` function in dashboard plugin utils should always return a definite value
### Actual behavior
Fetch to `/api/config` and access to `applicationUrl` within the `getApplicationUrl()` function in dashboard plugin utils doesn't return a definite value all the time due to a missing `await` of the fetch call.
### Steps to reproduce
Running dashboard plugin outside of development environment

development

mehmetegemen:
### What was the problem?

This PR resolves #6362 

### How was it solved?

- Add `await` to `.json()` call's returning promise

### How was it tested?

- Check if dashboard plugin react app works in prod env with default ws port 8080, manually
- Check if dashboard plugin react app works in dev env with ws port 5000, manually

### DIFF
--- a/framework-plugins/lisk-framework-dashboard-plugin/src/ui/utils/index.ts
+++ b/framework-plugins/lisk-framework-dashboard-plugin/src/ui/utils/index.ts
@@ -65,7 +65,7 @@ export const getApplicationUrl = async () => {
 		return 'ws://localhost:5000/ws';
 	}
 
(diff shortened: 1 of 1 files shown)


shuse2:
Superseded by #6368

PR #6368, opened 2021-05-01
Make application name configurable - Closes #6342

### LINKED ISSUE #6342: Update dashboard plugin application name to be customizable
### Expected behavior
Application name should be able set in dashboard plugin config, and the config should be reflected to the UI

### Actual behavior
Currently the application name is not set from the config

### Steps to reproduce
Check the mainpage on the dashboard

feature/6216-add_dashboard-plugin

mehmetegemen:
### What was the problem?

This PR resolves #6342 

### How was it solved?

- Created a general Config retreiver along with an interface and a default value container
- In react app we fetch config for application after connection along with other loading functions
- Loaded application name is directly passed into Logo component since it has a protection of empty prop

### How was it tested?
- `yarn start:web` with lisk-core api-ws-port 5000 to see dev env values
- Running prod lisk-core to see prod default config values
- Running prod lisk-core with extra config to see config affects the react app

### DIFF
--- a/framework-plugins/lisk-framework-dashboard-plugin/src/ui/pages/MainPage.tsx
+++ b/framework-plugins/lisk-framework-dashboard-plugin/src/ui/pages/MainPage.tsx
@@ -208,7 +208,7 @@ const MainPage: React.FC = () => {
 	// Get config as whole
 	React.useEffect(() => {
 		const initConfig = async () => {
-			setDashboard({ ...dashboard, ...await getConfig() });
+			setDashboard({ ...dashboard, ...(await getConfig()) });
 		};
 
 		initConfig().catch(console.error);
--- a/framework-plugins/lisk-framework-dashboard-plugin/src/ui/pages/MainPage.tsx
+++ b/framework-plugins/lisk-framework-dashboard-plugin/src/ui/pages/MainPage.tsx
@@ -39,7 +39,7 @@ import {
 	SendTransactionOptions,
 	CallActionOptions,
 } from '../types';
-import { getApplicationUrl, updateStatesOnNewBlock, updateStatesOnNewTransaction } from '../utils';
+import { getConfig, updateStatesOnNewBlock, updateStatesOnNewTransaction } from '../utils';
 import useRefState from '../utils/useRefState';
 import styles from './MainPage.module.scss';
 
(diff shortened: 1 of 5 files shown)


framework-plugins/lisk-framework-dashboard-plugin/src/ui/utils/index.ts
```diff
@@ -60,12 +70,14 @@ export const updateStatesOnNewTransaction = (
 	return [transaction, ...unconfirmedTransactions].slice(-1 * MAX_TRANSACTIONS) as Transaction[];
 };
 
-export const getApplicationUrl = async () => {
+export const getConfig = async (prop: keyof Config) => {
```
shuse2, line 73: instead of returning per property, i think it's better to return complete config to avoid call

shuse2 (changes requested):

framework-plugins/lisk-framework-dashboard-plugin/src/ui/utils/index.ts
mehmetegemen: Application name and URL are fetched once so 2 calls in total. I don't feel that remedying 2 calls' overhead is worth complicating MainPage's complexity and decreasing it's clarity by introducing usage of joint config calls. Are you sure of it? Or do you have something in your mind that is a very clear solution?

shuse2: If we increase the config properties, it will linearly increase the call.
Also, calling config endpoint once in the root as it was done originally does not seem to be increase of complexity

framework-plugins/lisk-framework-dashboard-plugin/src/ui/pages/MainPage.tsx
```diff
@@ -204,13 +205,13 @@ const MainPage: React.FC = () => {
 		setShowAccount(newAccount);
 	};
 
-	// Get connection string
+	// Get config as whole
 	React.useEffect(() => {
-		const initConnectionStr = async () => {
-			setDashboard({ ...dashboard, applicationUrl: await getApplicationUrl() });
+		const initConfig = async () => {
+			setDashboard({ ...dashboard, ...(await getConfig()) });
```
shuse2, line 211: can we split the line for await?

shuse2 (approved):
Minor comment but LGTM

nazarhussain (approved):

PR #6385, opened 2021-05-06
Update faucet config to use API - Closes #6384

### LINKED ISSUE #6384: Update faucet config fetch from plugin
### Expected behavior
Faucet should obtain the application config in the same way as the dashboard plugin
Also, it should have default value in case the server is not connected

### Actual behavior
- Faucet cannot be started without plugin being added to application
- It is not using API from the plugin

### Steps to reproduce
Start faucet plugin without plugin connected to application

development-

shuse2:
### What was the problem?

This PR resolves #6384 

### How was it solved?
- Update to fetch config from API in faucet
- Update default application port to 8080 (which is default in framework)

### How was it tested?

- Run `npm run start:web`
- Connect to core and run the plugin

Both should connect to the core properly

### DIFF
--- a/framework-plugins/lisk-framework-dashboard-plugin/src/ui/utils/index.ts
+++ b/framework-plugins/lisk-framework-dashboard-plugin/src/ui/utils/index.ts
@@ -20,7 +20,7 @@ interface Config {
 }
 
 const configDevEnvValues: Config = {
-	applicationUrl: 'ws://localhost:5000/ws',
+	applicationUrl: 'ws://localhost:8080/ws',
 	applicationName: 'Lisk',
 };
 
@@ -74,7 +74,7 @@ export const updateStatesOnNewTransaction = (
 };
 
 export const getConfig = async () => {
-	if (!process.env.NODE_ENV || process.env.NODE_ENV === 'development') {
+	if (process.env.NODE_ENV === 'development') {
 		return configDevEnvValues;
 	}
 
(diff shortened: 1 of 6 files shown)


framework-plugins/lisk-framework-dashboard-plugin/src/ui/utils/index.ts
```diff
@@ -20,7 +20,7 @@ interface Config {
 }
 
 const configDevEnvValues: Config = {
-	applicationUrl: 'ws://localhost:5000/ws',
+	applicationUrl: 'ws://localhost:8080/ws',
```
ManuGowda, line 23: shouldn't the port be read from config? if user change the default port and starts with different then faucet wont work?

framework-plugins/lisk-framework-faucet-plugin/src/ui/app.tsx
```diff
@@ -19,33 +19,51 @@ interface FaucetConfig {
...
 	applicationUrl: string;
 	tokenPrefix: string;
-	logoURL: string;
+	logoURL?: string;
 	captchaSitekey: string;
 }
 
+const defaultFaucetConfig: FaucetConfig = {
+	amount: '10000000000',
+	applicationUrl: 'ws://localhost:8080/ws',
```
ManuGowda, line 28: same here

framework-plugins/lisk-framework-dashboard-plugin/src/ui/utils/index.ts
shuse2: this is just default in the frontend, it will call API and overwrite it

ManuGowda (approved):

ishantiw (approved):

PR #6415, opened 2021-05-11
Add alphabet validation for plugin alias - Closes #6394

### LINKED ISSUE #6394: Non-Sanitize invalid names for bootstraping generators cause invalid code
### Expected behavior

One any of the bootstrapped generator is called the output should always be a valid source code files. 

### Actual behavior

If hyphenated names are passed it outputs the source code which is broken. 

```bash
lisk generate:plugin lns-dashboard
```

Above command end up generating code like below which is invalid js/ts code.  

```js
export class Lns-dashboardPlugin extends BasePlugin {
```

### Acceptance criteria 
- We should sanitize all the input from user to generate valid JS/TS code. 
- All the names for classes or variables should be PascalCase
- All the file paths should be underscore 

development

shuse2:
### What was the problem?

This PR resolves #6394 

### How was it solved?

- Add same validation rule as module (only alphabet restriction was missing in the plugin)

### How was it tested?

- Try generating a plugin with `generate:plugin sample-plugin` and it should fail

### DIFF
--- a/commander/src/commands/generate/plugin.ts
+++ b/commander/src/commands/generate/plugin.ts
@@ -58,7 +58,8 @@ export default class PluginCommand extends BaseBootstrapCommand {
 		// validate folder name to not include camelcase or whitespace
 		const regexWhitespace = /\s/g;
 		const regexCamelCase = /^([a-z]+)(([A-Z]([a-z]+))+)$/;
-		if (regexCamelCase.test(alias) || regexWhitespace.test(alias)) {
+		const regexAlphabets = /[^A-Za-z]/;
+		if (regexCamelCase.test(alias) || regexWhitespace.test(alias) || regexAlphabets.test(alias)) {
 			this.error('Invalid plugin alias');
 		}
 		if (standalone) {
--- a/commander/src/commands/generate/plugin.ts
+++ b/commander/src/commands/generate/plugin.ts
@@ -39,7 +39,6 @@ export default class PluginCommand extends BaseBootstrapCommand {
 		output: flagParser.string({
 			description: 'Path to create the plugin.',
 			char: 'o',
-			default: process.env.INIT_CWD ?? process.cwd(),
 			dependsOn: ['standalone'],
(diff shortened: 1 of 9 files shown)


commander/src/commands/generate/plugin.ts
```diff
@@ -41,7 +41,8 @@ export default class PluginCommand extends BaseBootstrapCommand {
 		// validate folder name to not include camelcase or whitespace
 		const regexWhitespace = /\s/g;
 		const regexCamelCase = /^([a-z]+)(([A-Z]([a-z]+))+)$/;
-		if (regexCamelCase.test(alias) || regexWhitespace.test(alias)) {
+		const regexAlphabets = /[^A-Za-z]/;
+		if (regexCamelCase.test(alias) || regexWhitespace.test(alias) || regexAlphabets.test(alias)) {
```
nazarhussain, line 45: @shuse2 I tested it with several values and it didn't work well and I think these valid values. 

```
> alias = "my-plugin"
'my-plugin'
> regexCamelCase.test(alias) || regexWhitespace.test(alias) || regexAlphabets.test(alias)
true
> alias = "my_plugin"
'my_plugin'
> regexCamelCase.test(alias) || regexWhitespace.test(alias) || regexAlphabets.test(alias)
true
```


Can we also expand the scope of the issue a bit further to also  to? 

- validate plugin class names and 
- module names and class names and
- application label

nazarhussain (changes requested):

commander/src/commands/generate/plugin.ts
shuse2: right now, what we allow in the class name seems to be with `regexCamelCase.test(alias) || regexWhitespace.test(alias) || regexAlphabets.test(alias)` rule, and i think it's better to be consistent with the module naming?

nazarhussain (approved):

mitsuaki-u (approved):

PR #6427, opened 2021-05-14
Fix JSON text area behavior - Closes #6422

### LINKED ISSUE #6422: Text area for params in send transaction widget is having problem with editing json object
### Expected behavior

Text area for params in send transaction widget should accept JSON and also allow users to edit properly.

### Actual behavior

Currently, when a user pastes a json param object in the text area it doesn't allow users to edit it properly and it shows cursor is at the wrong place while editing

### Steps to reproduce

try and paste below in Text area for params in send transaction widget 
```js
{
    "numberOfSignatures": 2,
    "optionalKeys": ["508a965871253595b362f8dc27bff6e67b39bdd466531be9c6f8c401253979c",
                    "a9a3c363a71a3089566352127cf0e6f79d3834e1d67b4132b98d35afd3b85375"],
     "mandatoryKeys": ["7720c58afffb0ab6503afe19a1caaacd6c6677df5b97d9c33a117f69860ab686"]
  }
 ```
  and try to edit or navigate through the text, it will show cursor in different places and it makes it very difficult to edit.
  For example in the picture below, I tried to add some text at the end and the cursor position is somewhere in the middle
  
image

`5.1.0`

shuse2:
### What was the problem?

This PR resolves #6422

### How was it solved?

- Remove usage of editor
- Update to use textarea and pre/code to handle manually
- Reference: https://css-tricks.com/creating-an-editable-textarea-that-supports-syntax-highlighted-code/

### How was it tested?

- Try to use the big JSON as described in the issue

### DIFF
--- a/framework-plugins/lisk-framework-dashboard-plugin/package.json
+++ b/framework-plugins/lisk-framework-dashboard-plugin/package.json
@@ -49,8 +49,7 @@
 		"react": "^17.0.1",
 		"react-dom": "^17.0.1",
 		"react-router-dom": "^5.2.0",
-		"react-select": "4.3.0",
-		"react-simple-code-editor": "0.11.0"
+		"react-select": "4.3.0"
 	},
 	"devDependencies": {
 		"@babel/core": "7.12.3",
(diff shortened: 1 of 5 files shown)


ishantiw (approved):
works well 🚀

ManuGowda (approved):

PR #6428, opened 2021-05-14
Base fee missing in api client - Closes #6424

### LINKED ISSUE #6424: Base fee is missing when sending transaction through lisk-api-client
### Expected behavior

Should consider base fee of a transaction while sending a transaction

### Actual behavior

Base fee is missing when we are sending transaction through api client.

### Steps to reproduce

Create lisk client and call `client.transaction.computeMinFee`

`5.1.0`

ishantiw:
### What was the problem?

This PR resolves #6424

### How was it solved?

Added `options` argument to `computeMinFee(assetSchema, transaction, options);` 

### How was it tested?

Run dashboard plugin, send delegate registration transaction


elements/lisk-api-client/src/transaction.ts
```diff
@@ -234,7 +246,13 @@ export class Transaction {
 
 	public computeMinFee(transaction: Record<string, unknown>): bigint {
 		const assetSchema = getTransactionAssetSchema(transaction, this._schema);
-		return computeMinFee(assetSchema, transaction);
+		const options: Options = {
+			minFeePerByte: this._nodeInfo.genesisConfig.minFeePerByte,
+			baseFees: this._nodeInfo.genesisConfig.baseFees,
+			numberOfSignatures: (transaction.asset as { numberOfSignatures: number }).numberOfSignatures || 1,
```
shuse2, line 252: shouldnt this be `transaction?.signatures.length > 0 ? transaction?.signatures.length : 1`
or accept as optional params as in the `computeMinFee`

shuse2 (approved):

ManuGowda (approved):

PR #6441, opened 2021-05-18
Fix npm install for generated app - Closes #6440

### LINKED ISSUE #6440: Generated application requires files property in the package.json
### Expected behavior
While installing, there should be no error

### Actual behavior
```
(node:333) Error Plugin: lisky8: files attribute must be specified in /home/zpt/temp/lisky8/package.json
```

### Steps to reproduce
./bin/run init ~/tmp/sample --registry https://npm.lisk.io

5.1.0-alpha.1

shuse2:
### What was the problem?

This PR resolves #6440

### How was it solved?

- Change package.json to package-template.json and rename during the generation so that it will not be used during the npm pack
- Add skipInstall to ignore the npm install by yeoman-environment

### How was it tested?

```
mkdir sample
cd sample
${abs path to commander}/bin/run init --registry https://npm.lisk.io
```

### DIFF
--- a/commander/src/bootstrapping/generators/init_generator.ts
+++ b/commander/src/bootstrapping/generators/init_generator.ts
@@ -20,6 +20,8 @@ export default class InitGenerator extends BaseGenerator {
 	public async initializing(): Promise<void> {
 		await this._loadAndValidateTemplate();
 
+		// Enable slipInstall for env so that the only generator install will run
+		this.env.options.skipInstall = true;
 		this.log('Initializing git repository');
 		this.spawnCommandSync('git', ['init', '--quiet']);
 	}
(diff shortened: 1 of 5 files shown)


commander/src/bootstrapping/generators/init_generator.ts
```diff
@@ -20,6 +20,8 @@ export default class InitGenerator extends BaseGenerator {
 	public async initializing(): Promise<void> {
 		await this._loadAndValidateTemplate();
 
+		// Enable slipInstall for env so that the only generator install will run
```
ManuGowda, line 23:
```suggestion
		// Enable skipInstall for env so that the only generator install will run
```

commander/src/bootstrapping/generators/init_plugin_generator.ts
```diff
@@ -28,6 +28,8 @@ export default class InitPluginGenerator extends BaseGenerator {
 		this._liskInitPluginArgs = {
 			alias: opts.alias,
 		};
+		// Enable slipInstall for env so that the only generator install will run
```
ManuGowda, line 31:
```suggestion
		// Enable skipInstall for env so that the only generator install will run
```

ManuGowda (approved):

ishantiw (approved):

Issue #6449, opened 2021-05-20
Lisk Commander returns an error if asset ID = 0


Tschakki:
### Expected behavior

It should be possible to generate transaction assets with ID = 0 using the `generate:asset` command of Lisk Commander.

### Actual behavior

If the asset ID is defined as 0, the following error is thrown:

```
$ lisk generate:asset heyModule heyAsset 0
 ›   Error: Invalid asset ID, only positive integers are allowed
```

### Steps to reproduce

1. Install Lisk Commander v5.1.0-alpha.4
2. Initialize an app 
3. Generate a module skeleton
4. Generate an asset skeleton for the module, with asset ID 0

```
lisk init
lisk generate:module heyModule 1001
lisk generate:asset heyModule heyAsset 0
```

### Which version(s) does this affect? (Environment, OS, etc...)
Lisk Commander v5.1.0-alpha.4

PR #6450, opened 2021-05-20
Asset file path in bootstrapping generator is not correct - Closes #6448, #6449

### LINKED ISSUE #6448: Asset file path in bootstrapping generator is not correct
### Expected behavior

When we generate an asset, the generated code should be valid without syntax or linting errors. 

### Actual behavior

Currently the asset file path on the `import ` inside `module` and `spec` is not correct. 

### Steps to reproduce

```
lisk init ~/my-app
cd ~/my-app
lisk generate:module lns 1000
lisk generate:asset lns transfer 1
```
The above code generate a file named `transfer_asset.ts`  but the import path uses `transfer.ts` 

5.1.0

nazarhussain:
### What was the problem?

This PR resolves #6448, #6449

### How was it solved?

- Fixed asset file path in the bootstrapping
- Fixed the positive value check in the asset oclif command

### How was it tested?

- Tested manually


commander/src/bootstrapping/templates/lisk-template-ts/templates/asset/src/app/modules/assets/asset.ts
```diff
@@ -13,11 +13,12 @@ export class <%= assetClass %> extends BaseAsset {
 		properties: {},
   };
 
-    public validate({ asset }: ValidateAssetContext<{}>): void {
+  public validate({ asset }: ValidateAssetContext<{}>): void {
     // Validate your asset
   }
 
-  	public async apply({ asset, transaction, stateStore }: ApplyAssetContext<{}>): Promise<void> {
+	// eslint-disable-next-line @typescript-eslint/require-await
```
ManuGowda, line 20: i don't think we need this?

commander/src/bootstrapping/templates/lisk-template-ts/templates/module/src/app/modules/module.ts
```diff
@@ -46,9 +53,9 @@ export class <%= moduleClass %> extends BaseModule {
     ];
     public id = <%= moduleID %>;
 
-    public constructor(genesisConfig: GenesisConfig) {
-        super(genesisConfig);
-    }
+    // public constructor(genesisConfig: GenesisConfig) {
```
ManuGowda, line 56: why commented?

ManuGowda (approved):

shuse2 (approved):

commander/src/bootstrapping/templates/lisk-template-ts/templates/asset/src/app/modules/assets/asset.ts
nazarhussain: In bootstrapped application its cause linting errors.

commander/src/bootstrapping/templates/lisk-template-ts/templates/module/src/app/modules/module.ts
nazarhussain: In bootstrapped application its cause linting errors, so why kept it commented as it would be useful for end user.

PR #6474, opened 2021-06-08
Add empty signature option - Closes #64670

### LINKED ISSUE #6470: Lisk Elements doesn't calculate the minFee of a multisignature transaction correctly
### Description
The signatures of a given accounts with total members (`optionalKeys.length + mandatoryKeys.length`) more than the required number of signature (`numberOfSignatures`) is an array like:
```
signatures: ['valid_signature_1', 'valid_signature_2', '', 'valid_signature_3']
```
where there are 4 members and `numberOfSignatures` equals to 3.

To get the `minFee` we do:
```
const minFee = transactions.computeMinFee(schema, {
    ...transactionWithoutFee,
    signatures: undefined,
  }, {
    baseFees: BASE_FEES,
    numberOfSignatures,
  });
```
Lisk Elements 5.1.0 doesn't take the empty string placeholder (for the third member who hasn't signed the tx) into account for calculating the transaction byte size, therefore the `minFee` calculated using said byte size is lower than Lisk Core expects. 

### Steps to reproduce
Try broadcasting a multisignature transaction of any type other than `4:0` with fee calculated by `transactions.computeMinFee` where the account has 1 mandatory member, 4 optional members and `numberOfSignatures` is 2.
The broadcast will fail because `fee` is lower than `minFee` required.

Lisk Elements 5.1.0

shuse2:
### What was the problem?

This PR resolves #6470 

### How was it solved?

- Add `numberOfEmptySignatures` option to specify number of empty signatures to include in calculation

### How was it tested?

- Add test case where signatures need to include empty signature

### DIFF
--- a/elements/lisk-transactions/src/fee.ts
+++ b/elements/lisk-transactions/src/fee.ts
@@ -38,13 +38,11 @@ const computeTransactionMinFee = (
 	trx: Record<string, unknown>,
 	options?: Options,
 ): bigint => {
-	const mockSignatures = new Array(options?.numberOfSignatures ?? DEFAULT_NUMBER_OF_SIGNATURES).fill(
-		Buffer.alloc(DEFAULT_SIGNATURE_BYTE_SIZE),
-	);
+	const mockSignatures = new Array(
+		options?.numberOfSignatures ?? DEFAULT_NUMBER_OF_SIGNATURES,
+	).fill(Buffer.alloc(DEFAULT_SIGNATURE_BYTE_SIZE));
 	if (options?.numberOfEmptySignatures) {
-		mockSignatures.push(...new Array(options.numberOfEmptySignatures).fill(
-			Buffer.alloc(0),
-		));
+		mockSignatures.push(...new Array(options.numberOfEmptySignatures).fill(Buffer.alloc(0)));
 	}
 	const size = getBytes(assetSchema, {
 		...trx,
--- a/elements/lisk-transactions/src/fee.ts
+++ b/elements/lisk-transactions/src/fee.ts
@@ -22,9 +22,10 @@ interface BaseFee {
 }
 
 interface Options {
-	readonly minFeePerByte: number;
-	readonly baseFees: BaseFee[];
-	readonly numberOfSignatures: number;
+	readonly minFeePerByte?: number;
+	readonly baseFees?: BaseFee[];
+	readonly numberOfSignatures?: number;
+	readonly numberOfEmptySignatures?: number;
 }
 
 const DEFAULT_MIN_FEE_PER_BYTE = 1000;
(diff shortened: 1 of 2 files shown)


elements/lisk-transactions/src/fee.ts
```diff
@@ -22,9 +22,10 @@ interface BaseFee {
 }
 
 interface Options {
-	readonly minFeePerByte: number;
-	readonly baseFees: BaseFee[];
-	readonly numberOfSignatures: number;
+	readonly minFeePerByte?: number;
+	readonly baseFees?: BaseFee[];
+	readonly numberOfSignatures?: number;
+	readonly numberOfEmptySignatures?: number;
```
ManuGowda, line 28: do we need this extra param? from signatures array we can know the empty buffer and take the count?

shuse2: we don't use `signatures` param in this function, so we need to know how many is required and how many signatures are empty

ManuGowda (approved):

elements/lisk-transactions/src/fee.ts
```diff
@@ -37,14 +38,18 @@ const computeTransactionMinFee = (
 	trx: Record<string, unknown>,
 	options?: Options,
 ): bigint => {
+	const mockSignatures = new Array(
```
mitsuaki-u, line 41: Is it mock? maybe name it `emptySignatures`?

elements/lisk-transactions/src/fee.ts
```diff
@@ -37,14 +38,18 @@ const computeTransactionMinFee = (
...
 	const size = getBytes(assetSchema, {
 		...trx,
-		signatures: new Array(options?.numberOfSignatures ?? DEFAULT_NUMBER_OF_SIGNATURES).fill(
-			Buffer.alloc(DEFAULT_SIGNATURE_BYTE_SIZE),
-		),
+		signatures: mockSignatures,
 	}).length;
 	const baseFee =
-		options?.baseFees.find(bf => bf.moduleID === trx.moduleID && bf.assetID === trx.assetID)
+		options?.baseFees?.find(bf => bf.moduleID === trx.moduleID && bf.assetID === trx.assetID)
```
mitsuaki-u, line 52:
```suggestion
		options?.baseFees?.find(baseFee => baseFee.moduleID === trx.moduleID && baseFee.assetID === trx.assetID)
```

mitsuaki-u (approved):
LGTM

elements/lisk-transactions/src/fee.ts
shuse2: it's both 64bytes and 0 bytes, so empty is not correct I think

ManuGowda: dummySignatures?

PR #6504, opened 2021-06-25
Update tests specs for the BLS signatures - Closes #6503 

### LINKED ISSUE #6503: Update tests specs for the BLS signatures
### Expected behavior

All the tests specs should be up-to-date with the LIP specification. 

### Actual behavior

There are some changes introduced in LIP, which is not yet been merged. So test specs are not matching with it. 

development

nazarhussain:
### What was the problem?

This PR resolves #6503 

### How was it solved?

- Added more test cases matching to LIP update. 

### How was it tested?

- Run all unit tests


elements/lisk-cryptography/test/protocol_specs/bls_specs/fast_aggregate_verify/pk_is_identity_element.yml
```diff
@@ -0,0 +1,7 @@
+input:
```
AndreasKendziorra, line 1: I find the file name slightly misleading as one could think that one public key is the identity element, but it is about the aggregate public key. So either
"sum_of_pks_is_identity_element" or
"aggregate_pk_is_identity_element"

AndreasKendziorra (changes requested):
Just some comments regarding file names for some test cases. Everything else is good!

AndreasKendziorra (approved):

shuse2 (approved):

PR #6509, opened 2021-06-29
Update tree interface - Closes #6491 and #6510

### LINKED ISSUE #6491: Update merkle tree to accept the database interface
### Description

Currently all the leaf/branch data are stored in memory. However, with the interoperability, we want to store the CCM data on the plugin with the whole merkle tree data.
Therefore, we should accept the database interface, which can be either in-memory or actual key-value DB instance.

### Specification

Update MerkleTree constructor to be 
```
constructor(initValues: Buffer[] = [], options?: { preHashedLeaf?: boolean, db?: Database })
```
where Database interface is
```
interface Database {
  get(key: Buffer): Promise<Buffer>;
  set(key: Buffer, value: Buffer): Promise<void>;
}
```
It should replace all the in-memory usage with the above database interface
db instance should default to in-memory

### Acceptance Criteria

- All in-memory usage and test should not be changed except the constructor
- RocksDB instance should be usable for the MerkleTree

shuse2:
### What was the problem?

This PR resolves #6491 and #6510

### How was it solved?

- Create database interface and expect to be optional input in constructor
- Update merkle tree initialization to be separate function because it's asynchronous task now
- Rename the variable to be consistent with SMT
- Remove unnecessary interface which cannot be supported with DB

### How was it tested?

- All test should be passing

### DIFF
--- a/elements/lisk-tree/src/merkle_tree/merkle_tree.ts
+++ b/elements/lisk-tree/src/merkle_tree/merkle_tree.ts
@@ -169,12 +169,12 @@ export class MerkleTree {
 	public async generateProof(queryData: ReadonlyArray<Buffer>): Promise<Proof> {
 		if (this._size === 0) {
 			return {
-				path: [],
+				siblingHashes: [],
 				indexes: [],
-				dataLength: 0,
+				size: 0,
 			};
 		}
-		const path = [];
+		const siblingHashes = [];
 		const addedPath = new dataStructures.BufferSet();
 		const indexes = [];
 		let queryNode: NodeInfo | undefined;
@@ -184,7 +184,7 @@ export class MerkleTree {
 			try {
 				queryNode = await this.getNode(queryData[i]);
 			} catch (err) {
-				path.push({
+				siblingHashes.push({
 					hash: queryData[i],
 					layerIndex: undefined,
 					nodeIndex: undefined,
@@ -200,7 +200,7 @@ export class MerkleTree {
 			if (this._size === 1 && this._root.equals(queryNode.hash)) {
 				if (!addedPath.has(queryNode.hash)) {
 					addedPath.add(queryNode.hash);
-					path.push({
+					siblingHashes.push({
 						hash: queryNode.hash,
 						layerIndex: 0,
 						nodeIndex: 0,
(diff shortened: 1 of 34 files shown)


nazarhussain (approved):

elements/lisk-tree/src/merkle_tree/merkle_tree.ts
```diff
@@ -24,41 +24,57 @@ import {
...
 	BRANCH_PREFIX,
 } from './constants';
-import { NodeData, NodeInfo, NodeType, NodeSide, Proof } from './types';
+import { InMemoryDB } from './inmemory_db';
+import { PrefixStore } from './prefix_store';
+import { NodeData, NodeInfo, NodeType, NodeSide, Proof, Database } from './types';
 import { generateHash, getBinaryString, isLeaf, getPairLocation } from './utils';
 
+const hashToValuePrefix = Buffer.from([0]);
+const locationToHashPrefix = Buffer.from([1]);
```
nazarhussain, line 33: Can we make these as constants and move to constants file? etc `LOCATION_TO_HASH_PREFIX`

elements/lisk-tree/src/merkle_tree/merkle_tree.ts
```diff
@@ -23,42 +23,57 @@ import {
...
+	public constructor(options?: { preHashedLeaf?: boolean; db?: Database }) {
+		this._db = options?.db ?? new InMemoryDB();
+		this._hashToValueMap = new PrefixStore(this._db, PREFIX_HASH_TO_VALUE);
+		this._locationToHashMap = new PrefixStore(this._db, PREFIX_LOC_TO_HASH);
+		this._preHashedLeaf = options?.preHashedLeaf ?? false;
+
+		this._root = EMPTY_HASH;
+	}
+
+	public async init(initValues: Buffer[]): Promise<void> {
```
mitsuaki-u, line 52: why do we need seperate `init` here? why not just in constructor?

shuse2: it's needed because now get/set is `Promise` and construtor cannot be async

mitsuaki-u (approved):

elements/lisk-tree/src/merkle_tree/merkle_tree.ts
nazarhussain: I guess because constructors can't be async. The constructor must return an object of a particular type only, means `MerkleTree` constructor must return an object of type `MerkleTree`, making it async will turn type to `Promise<MerkleTree>`, so we can't make constructors async.

PR #6512, opened 2021-06-30
Implement utility functions for sparse Merkle tree - Closes #6496 

### LINKED ISSUE #6496: Implement utility functions for sparse Merkle tree
### Description

The following utility functions are needed in `SparseMerkleTree` class :

- `sortByBitmapAndKey` : sort queries by the longest binaryBitmap, breaking ties by smaller key
- `filterQueries` : remove queries that have merged together, keep only those with a different key prefix
- `areSiblingQueries` : check if two queries correspond to nodes that are children of the same branch node
- `calculateRoot` : calculate Merkle root of tree from query response of function `generateMultiProof`
-  `binaryExpansion` : prepend leading 0s to binary value of key

### Motivation

Utils needed by other functions

### Acceptance Criteria

Other utility functions if necessary should also be implemented

### Additional Information

Other utility functions like :
- `binarySearch`
- `binaryStringToBuffer`
- `bufferToBinaryString`

nazarhussain:
### What was the problem?

This PR resolves #6496 

**There are two functions `leafNode` and `branchHash` for which fake implementation is used.**

### How was it solved?

- Added utils functions defined in issue

### How was it tested?

- Added unit tests

### DIFF
--- a/elements/lisk-tree/src/sparse_merkle_tree/types.ts
+++ b/elements/lisk-tree/src/sparse_merkle_tree/types.ts
@@ -12,20 +12,21 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
-export interface InclusionProof {
+export interface Proof {
 	readonly siblingHashes: Buffer[];
-	readonly queries: InclusionProofQuery[];
+	readonly queries: ProofQuery[];
 }
 
-export interface InclusionProofQuery {
+export interface ProofQuery {
 	readonly key: Buffer;
 	readonly value: Buffer;
-	readonly bitmap: Buffer;
+	// During calculations bitmap values can change so these are not readonly
+	bitmap: Buffer;
+	binaryBitmap: string;
 }
 
-export interface InclusionProofQueryWithHash extends InclusionProofQuery {
+export interface ProofQueryWithHash extends ProofQuery {
 	hash: Buffer;
-	binaryBitmap: string;
 }
 
 export interface Database {
--- a/elements/lisk-tree/src/sparse_merkle_tree/types.ts
+++ b/elements/lisk-tree/src/sparse_merkle_tree/types.ts
@@ -1,0 +1,24 @@
+/*
+ * Copyright © 2021 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+export interface InclusionProof {
+	readonly siblingHashes: Buffer[];
+	readonly queries: InclusionProofQueryResult[];
+}
+
+export interface InclusionProofQueryResult {
+	key: Buffer;
+	value: Buffer;
+	bitmap: Buffer;
+}
--- a/elements/lisk-tree/src/sparse_merkle_tree/types.ts
+++ b/elements/lisk-tree/src/sparse_merkle_tree/types.ts
@@ -14,19 +14,13 @@
 
 export interface Proof {
 	readonly siblingHashes: Buffer[];
-	readonly queries: ProofQuery[];
+	readonly queries: Query[];
 }
 
-export interface ProofQuery {
+export interface Query {
 	readonly key: Buffer;
 	readonly value: Buffer;
-	// During calculations bitmap values can change so these are not readonly
-	bitmap: Buffer;
-	binaryBitmap: string;
-}
-
-export interface ProofQueryWithHash extends ProofQuery {
-	hash: Buffer;
+	readonly bitmap: Buffer;
 }
 
 export interface Database {
--- a/elements/lisk-tree/src/sparse_merkle_tree/utils.ts
+++ b/elements/lisk-tree/src/sparse_merkle_tree/utils.ts
@@ -15,33 +15,35 @@
 import { hash } from '@liskhq/lisk-cryptography';
 import { objects } from '@liskhq/lisk-utils';
 import { BRANCH_HASH_PREFIX, EMPTY_HASH, LEAF_HASH_PREFIX } from './constants';
-import { InclusionProofQuery, InclusionProofQueryWithHash } from './types';
+import { ProofQuery, ProofQueryWithHash } from './types';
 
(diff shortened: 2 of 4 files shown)


elements/lisk-tree/src/sparse_merkle_tree/utils.ts
```diff
@@ -13,9 +13,180 @@
...
+	const keyPrefix2 = binaryKey2.substring(0, h - 1);
+
+	if (keyPrefix1 !== keyPrefix2) {
+		return false;
+	}
+
+	const d1 = binaryKey1[h];
+	const d2 = binaryKey2[h];
+
+	return (d1 === '0' && d2 === '1') || (d1 === '1' && d2 === '0');
```
shuse2, line 88: in the LIP, it says `return d1==0 and d2==1` because q1 and q2 are left and right respectively?

elements/lisk-tree/src/sparse_merkle_tree/types.ts
```diff
@@ -12,6 +12,22 @@
...
+export interface InclusionProofQuery {
+	readonly key: Buffer;
+	readonly value: Buffer;
+	readonly bitmap: Buffer;
+}
+
+export interface InclusionProofQueryWithHash extends InclusionProofQuery {
+	hash: Buffer;
+	binaryBitmap: string;
+}
```
Incede, line 29: Do we really need this ?

I think `hash` & `binaryBitmap` can be derived from the base properties using other util functions.

elements/lisk-tree/src/sparse_merkle_tree/utils.ts
```diff
@@ -13,9 +13,180 @@
...
+export const sortByBitmapAndKey = <T extends InclusionProofQuery>(queries: T[]): T[] =>
+	queries.sort((q1, q2) => {
+		const q1BinaryBitmap = bufferToBinaryString(q1.bitmap);
+		const q2BinaryBitmap = bufferToBinaryString(q2.bitmap);
+
+		if (q1BinaryBitmap.length === q2BinaryBitmap.length) {
+			return q1.key.byteLength - q2.key.byteLength;
+		}
+
+		return q2BinaryBitmap.length - q1BinaryBitmap.length;
```
Incede, line 31: maybe isolate the logic here as a separate function for better readability

elements/lisk-tree/src/sparse_merkle_tree/utils.ts
```diff
@@ -13,9 +13,180 @@
...
+	for (const q of queries) {
+		data.push({
+			...q,
+			binaryBitmap: bufferToBinaryString(q.bitmap),
+			hash: q.value.byteLength === 0 ? EMPTY_HASH : leafHash(q.key, q.value),
+		});
+	}
+
+	let sortedQueries = filterQueries<InclusionProofQueryWithHash>(
+		sortByBitmapAndKey<InclusionProofQueryWithHash>(data),
```
Incede, line 110: I think we can avoid passing the `data` parameter to this function.
The needed conversion(binaryBitmap, hash) can be done in the concerned function itself.

Related to my comment on `types.ts`

Incede (changes requested):

elements/lisk-tree/src/sparse_merkle_tree/utils.ts
```diff
@@ -13,9 +13,180 @@
...
+		const h = q.binaryBitmap.length;
+		const binaryKey = binaryExpansion(q.key, keyLength);
+
+		// we distinguish three cases for the sibling hash:
+		// 1. sibling is next element of sortedQueries
+		let siblingHash!: Buffer;
+
+		if (sortedQueries.length > 1 && areSiblingQueries(q, sortedQueries[1], keyLength)) {
+			siblingHash = sortedQueries[1].hash;
+			sortedQueries = sortedQueries.splice(1, 1);
```
shuse2, line 133: return value of `splice` is the removed element, but here i think you just need to remove?

elements/lisk-tree/src/sparse_merkle_tree/utils.ts
```diff
@@ -13,9 +13,180 @@
...
+		}
+		// 2. sibling is default empty node
+		else if (b === '0') {
+			siblingHash = EMPTY_HASH;
+		}
+		// 3. sibling hash comes from siblingHashes
+		else if (b === '1') {
+			// eslint-disable-next-line prefer-destructuring
+			siblingHash = siblingHashes[0];
+			siblingHashes = siblingHashes.splice(0, 1);
```
shuse2, line 143: same as above

shuse2 (changes requested):

elements/lisk-tree/src/sparse_merkle_tree/types.ts
nazarhussain: I am not sure, as per the LIP these attributes will be passed to the function.

elements/lisk-tree/src/sparse_merkle_tree/utils.ts
nazarhussain: Didn't understood. Which logic to move to separate function?

elements/lisk-tree/src/sparse_merkle_tree/utils.ts
nazarhussain: Yes it can be done, but which function it would be I am not sure. I tried to follow exact steps defined in the LIP without thinking of optimization concerns other functions.

elements/lisk-tree/src/sparse_merkle_tree/utils.ts
nazarhussain: Was discussing with @ricott1 and he suggested that we should fix this and make it commutative.

elements/lisk-tree/src/sparse_merkle_tree/utils.ts
Incede: I think `bufferToBinaryString`& `leafHash` can be used.

I agree the LIP follows this approach but imo we may consider these as optimizations or for consistency & reduced complexity across function interfaces.

elements/lisk-tree/src/sparse_merkle_tree/utils.ts
Incede: I was thinking to isolate this code block but it's not that important.
```
		if (q1BinaryBitmap.length === q2BinaryBitmap.length) {
			return q1.key.byteLength - q2.key.byteLength;
		}

		return q2BinaryBitmap.length - q1BinaryBitmap.length;
```

Incede (approved):

shuse2 (approved):

PR #6549, opened 2021-07-02
Calculate light merkle root without data - Closes #6490

### LINKED ISSUE #6490: Implement function to calculate merkle tree root without using data
### Description

Add function to calculate next merkle root with current root, append path, size and the next leaf value.

### Specification

```
const calculateMerkleRoot = (root: Buffer, appendPath: []Buffer, size: number, value: Buffer) => {
  root: Buffer,
  appendPath: []Buffer,
  size: number,
}
```
It should implement this function according to the [LIP-0031](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0031.md)

### Acceptance Criteria

- Unit tests are added to cover all the path
- Use protocol spec to check if the calculated root are correct
- calculateMerkleRoot is exposed from library

### Additional Information

Related issues : #6485

mitsuaki-u:
### What was the problem?

This PR resolves 6490

### How was it solved?

Create utility function to calculate merkle root, appendPath and size without merkle data.


elements/lisk-tree/src/merkle_tree/utils.ts
```diff
@@ -142,3 +142,54 @@ export const getPairLocation = (nodeInfo: {
...
+
+	for (subTreeIndex = 0; subTreeIndex < treeHeight; subTreeIndex += 1) {
+		if (!((size >> subTreeIndex) & 1)) {
+			break;
+		}
+	}
+
+	let newAppendPath = appendPath.slice(0);
+	currentHash = newLeafHash;
+	const splicedPath = newAppendPath.splice(subTreeIndex);
```
shuse2, line 188: why are we using `splice` here instead of `splice`?

shuse2 (changes requested):

elements/lisk-tree/src/merkle_tree/utils.ts
mitsuaki-u: It is to easily split the appendPath into two since `splice` mutates the original. We can also use slice but with splice it is a one-liner.

elements/lisk-tree/src/merkle_tree/utils.ts
```diff
@@ -142,3 +142,54 @@ export const getPairLocation = (nodeInfo: {
...
+	let subTreeIndex;
+	const treeHeight = Math.ceil(Math.log2(size)) + 1;
+
+	for (subTreeIndex = 0; subTreeIndex < treeHeight; subTreeIndex += 1) {
+		if (!((size >> subTreeIndex) & 1)) {
+			break;
+		}
+	}
+
+	let newAppendPath = appendPath.slice(0);
```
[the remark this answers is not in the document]
mitsuaki-u: The appendPath is added to later so it cannot be `const`

shuse2 (approved):

ricott1 (approved):

PR #6558, opened 2021-07-05
Implement verify function for sparse Merkle tree - Closes #6498 

### LINKED ISSUE #6498: Implement verify function for sparse Merkle tree
### Description
The `verify` function takes as input a set of `queryKeys`, the query response from the function `generateMultiProof` and a `rootHash` to check if the Merkle root of the tree obtained from the query response matches the `rootHash` hence verifying (or not) the proof provided.

### Motivation

The proof provided by the prover needs to be checked by the verifier for correctness.

### Acceptance Criteria

Add test cases

### Additional Information

Related issues : #6485 #6496

nazarhussain:
### What was the problem?

This PR resolves #6498 

### How was it solved?

- Created verify function for smt 

### How was it tested?

- Run all unit tests

### DIFF
--- a/elements/lisk-tree/src/sparse_merkle_tree/utils.ts
+++ b/elements/lisk-tree/src/sparse_merkle_tree/utils.ts
@@ -15,9 +15,56 @@
 import { hash } from '@liskhq/lisk-cryptography';
 import { objects } from '@liskhq/lisk-utils';
 import { BRANCH_HASH_PREFIX, EMPTY_HASH, LEAF_HASH_PREFIX } from './constants';
-import { Query } from './types';
+import { Query, Proof } from './types';
 
 type CalculateRootQueryObjects = Omit<Query, 'bitmap'> & { hash: Buffer; binaryBitmap: string };
+
+export const getOverlappingStr = (str1: string, str2: string) => {
+	const output = [''];
+
+	for (let i = 0; i < str1.length; i += 1) {
+		if (str1[i] !== str2[i]) {
+			return output.join('');
+		}
+
+		output.push(str1[i]);
+	}
+
+	return output.join('');
+};
+
+export const verify = (
+	queryKeys: Buffer[],
+	proof: Proof,
+	merkleRoot: Buffer,
+	keyLength: number,
+): boolean => {
+	if (queryKeys.length !== proof.queries.length) {
+		return false;
+	}
+
+	for (const [index, q] of proof.queries.entries()) {
+		const k = queryKeys[index];
+
+		// q is an inclusion proof for k or a default empty node
+		if (k.equals(q.key)) {
+			continue;
+		}
+
+		// q is an inclusion proof for another leaf node
+		const binaryResponseBitmap = bufferToBinaryString(q.bitmap);
+		const binaryResponseKey = binaryExpansion(q.key, keyLength);
+		const binaryQueryKey = binaryExpansion(k, keyLength);
+		const sharedPrefix = getOverlappingStr(binaryResponseKey, binaryQueryKey);
+
+		if (binaryResponseBitmap.length > sharedPrefix.length) {
+			// q does not give an non-inclusion proof for k
+			return false;
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


ricott1 (approved):
A minor comment, the rest looks good. I am wondering however if we should try to specify this avoiding binary strings and rather using bitwise operations on integers. It is mostly for an aesthetic reason, but it could also improve performance.

elements/lisk-tree/src/sparse_merkle_tree/utils.ts
```diff
@@ -15,10 +15,57 @@
 import { hash } from '@liskhq/lisk-cryptography';
 import { objects } from '@liskhq/lisk-utils';
 import { BRANCH_HASH_PREFIX, EMPTY_HASH, LEAF_HASH_PREFIX } from './constants';
-import { Query } from './types';
+import { Query, Proof } from './types';
 
 type CalculateRootQueryObjects = Omit<Query, 'bitmap'> & { hash: Buffer; binaryBitmap: string };
 
+export const getOverlappingStr = (str1: string, str2: string) => {
+	const output = [''];
```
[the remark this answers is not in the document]
nazarhussain: Yes we can use string concatenation. But as per earlier experience string concatenation is slower than array push/join. Didn't tested it with latest version NodeJS version though.

elements/lisk-tree/test/sparse_merkle_tree/utils.spec.ts
```diff
@@ -283,4 +285,135 @@ describe('utils', () => {
...
+			};
+			const merkleRoot = Buffer.from(
+				'21ecda9db382eff32c9ec899fc7090cf58858e8c22a2af82510cd4d9c9a42c2f',
+				'hex',
+			);
+
+			expect(verify(queryKeys, proof, merkleRoot, 1)).toBeFalse();
+		});
+
+		it('should return false if queries does not match and proof does not provide inclusion proof of one of query key', () => {
```
Incede, line 327: I think you mean to say here "proof does not provide inclusion or exclusion proof" since here the proof for the key `01100000` should be an exclusion proof.

Incede (changes requested):

elements/lisk-tree/test/sparse_merkle_tree/utils.spec.ts
```diff
@@ -283,4 +285,135 @@ describe('utils', () => {
...
+			};
+			const merkleRoot = Buffer.from(
+				'21ecda9db382eff32c9ec899fc7090cf58858e8c22a2af82510cd4d9c9a42c2f',
+				'hex',
+			);
+
+			expect(verify(queryKeys, proof, merkleRoot, 1)).toBeFalse();
+		});
+
+		it('should return true if queries does not match but proof does provide inclusion proof of that query key', () => {
```
Incede, line 359: same here - the proof for key `10101000` is an exclusion proof

elements/lisk-tree/test/sparse_merkle_tree/utils.spec.ts
nazarhussain: So the actual logic is either all `user queries keys` match to the `proof query objects keys`, or if they does not match then some of query object in the proof provide proof of the keys user asked. 

So in this test case we are checking the query objects asked by user `00110011, 01100000` are different than what we get in proof `00110011, 11100001`. In this case user asked for key `01100000` but proof contains `11100001` yet its good enough to provide inclusion of earlier key.

elements/lisk-tree/test/sparse_merkle_tree/utils.spec.ts
nazarhussain: Same as above.

elements/lisk-tree/src/sparse_merkle_tree/utils.ts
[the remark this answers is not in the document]
ricott1: ok, thanks

Incede (approved):

PR #6511, opened 2021-06-30
Create sparse Merkle tree and sub-component classes with interface & constructor - Closes #6487 #6488

Incede:
### What was the problem?

This PR resolves #6487 #6488 

### How was it solved?

* Added interface for `SparseMerkleTree` and different types of nodes
* Implemented constructor and basic functions

### How was it tested?

Added test interface

PR #6570, opened 2021-07-07
Implement update function for Sparse merkle tree - Closes #6489

### LINKED ISSUE #6489: Implement update function for sparse Merkle tree
### Description
The `update` function will take a key/value pair and add it to the existing dataset or update it's value if it's already present. It will also update the Merkle root of the sparse Merkle tree to reflect the new data. To do so it needs to update all the branch nodes present in the path to the root.

### Motivation
* To add new data/nodes to the dataset/tree
* To update existing data/node values

### Acceptance Criteria
* Add test cases to match the protocol specifications (fixtures) provided.

### Additional Information
Related issues : #6485

ishantiw:
### What was the problem?

This PR resolves #6489 

### How was it solved?

- Created `public async update(key: Buffer, value: Buffer) ` function as described in [SMT:update(k,v)](https://github.com/LiskHQ/lips-staging/blob/7520afe579d20f6e943d30de475eb0ce186b04c1/proposals/lip-Introduce-sparse-Merkle-trees.md#root-hash-calculation)

### How was it tested?

`npm t` under `lisk-tree`


elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -35,8 +43,108 @@ export class SparseMerkleTree {
...
+
+		return new Branch(leftHash, rightHash);
+	}
+
+	public async update(key: Buffer, value: Buffer) {
+		if (!value) {
+			throw new Error('Value cannot be empty');
+		}
+
+		if (key.length !== this.keyLength) {
```
nazarhussain, line 69: Would prefer to use `byteLength` instead of `length` consistently when `Buffer` type is involved. It gave more readability and consistent with NodeJS docs as well.

nazarhussain, line 69: Would prefer to use `byteLength` instead of `length` consistently when `Buffer` type is involved. It gave more readability and consistent with NodeJS docs as well.

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -35,8 +43,108 @@ export class SparseMerkleTree {
...
+		}
+
+		const newLeaf = new Leaf(key, value);
+		const ancestorNodes: TreeNode[] = [];
+		let bottomNode: TreeNode = new Empty();
+		let currentNode = await this.getNode(this.rootHash);
+		const binaryKey = binaryExpansion(key, this.keyLength);
+		let h = 0;
+
+		while (!isLeaf(currentNode.hash)) {
```
nazarhussain, line 80: This function `isLeaf` expect `value` not the `hash`. From hash we can't check the type of the node as the prefix was concatenated before hashing.

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -35,8 +43,108 @@ export class SparseMerkleTree {
...
+			// Append currentNode to ancestorNodes
+			ancestorNodes.push(currentNode);
+			const d = binaryKey[h];
+			if (+d === 0) {
```
nazarhussain, line 84: On other places in code for similar check we are doing string comparison. I don't have preferences but we should use one consistent approach. Don't you think typecasting and then comparison is unnecessary overhead?

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -35,8 +43,108 @@ export class SparseMerkleTree {
...
+				currentNode = await this.getNode(currentNode.hash);
+			} else if (+d === 1) {
+				currentNode = await this.getNode(currentNode.hash);
```
nazarhussain, line 87: How we differentiate between left and right nodes here?

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -35,8 +43,108 @@ export class SparseMerkleTree {
...
+			if (+d === 0) {
+				bottomNode = new Branch(newLeaf.hash, currentNode.hash);
+			} else if (+d === 1) {
+				bottomNode = new Branch(currentNode.hash, newLeaf.hash);
+			}
+		}
+		// Finally update all branch nodes in ancestorNodes
+		// Starting from the last
+		while (h > 0) {
+			const d = +binaryKey[h];
```
nazarhussain, line 123: The LIP refers to `h - 1` here. 

> let d be the digit at position h-1 of binaryKey

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -35,8 +43,108 @@ export class SparseMerkleTree {
...
+			const p = ancestorNodes[h - 1];
+			if (d === 0 && p instanceof Branch) {
```
nazarhussain, line 125: Is there a possibility that we have binary key value either `0` or `1` and node there in path is not a branch node?

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -35,8 +43,108 @@ export class SparseMerkleTree {
...
+			} else if (d === 1 && p instanceof Branch) {
+				// Let siblingNode be the left child node of p
+				const siblingNodeHash = p.rightHash;
+				// Update p.data to siblingNode.hash|bottomNode.hash
+				p.update(bottomNode.hash, NodeSide.RIGHT);
+				// Update p.hash to branchHash(p.data)
+				p.update(siblingNodeHash, NodeSide.LEFT);
+			}
+			bottomNode = p;
+			h += 1;
```
nazarhussain, line 141: In above loop we have `while (h > 0) {` and here we are doing `h += 1`. It seems loop will never end.

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -35,8 +43,108 @@ export class SparseMerkleTree {
...
+		}
+		// The final value of bottomNode is the root node of the tree
+		return bottomNode;
```
nazarhussain, line 144: This seems wrong the `bottomNode` can't be the rootNode. If you are traversing from bottom to top would be nice to rename this variable to `curreentNode`

nazarhussain (changes requested):

elements/lisk-tree/src/sparse_merkle_tree/branch.ts
```diff
@@ -12,35 +12,47 @@
...
-import { NodeSide } from './constants';
+import { hash } from '@liskhq/lisk-cryptography';
+import { branchDataBuffer } from './utils';
+import { NodeSide, NodeType } from './constants';
 
 export class Branch {
-	private _left: Leaf;
-	private _right: Leaf;
+	private _leftHash: Buffer;
+	private _rightHash: Buffer;
```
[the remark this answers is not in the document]
ishantiw: So the branch constructor needs left/right instance `(left: Leaf, right: Leaf)`, so first of all, we can have left or right a branch node too apart from leaf, right?. So in order to create a branch node, I need to pass left/right, so when I call `getNode`, and create a branch object, I need to instantiate the left/right leaf or branch for it, and then those instance will need left/right leaf or branch object for themselves and so on.

elements/lisk-tree/src/sparse_merkle_tree/error.ts
```diff
@@ -0,0 +1,15 @@
...
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+export class NotFoundError extends Error {}
```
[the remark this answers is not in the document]
ishantiw: I simply copied `error` and `prefix_store` files from`merkle_tree` for consistency, if we want to change it then its also for fine but we should do it for both?

elements/lisk-tree/src/sparse_merkle_tree/prefix_store.ts
```diff
@@ -0,0 +1,42 @@
...
+
+	public async get(key: Buffer): Promise<Buffer | undefined> {
+		try {
+			const value = await this._db.get(Buffer.concat([this._prefix, key]));
+			return value;
+		} catch (error) {
+			if (error instanceof NotFoundError) {
+				return undefined;
+			}
+			throw error;
```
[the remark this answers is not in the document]
ishantiw: This file is copied from `merkle_tree/prefix_store` if we agree to do that then lets change both

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
ishantiw: I did changed it but somehow I didn't push the changes 🤔

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
ishantiw: if you see ancestorNodes array we are only pushing branch (or default branch) nodes so there will not be any non branch nodes

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
ishantiw: ah good catch, its `h -=1` 😅

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
ishantiw: so the first node in `ancestorNodes` array is the root node that we insert in the beginning of this function. Then in this loop, we are assigning `bottomNode = p;` in the end of the loop where `p` is `ancestorNodes[h - 1]`, and if we are deducting `h--` then at `h = 1` we will have  =>`p = ancestorNodes[1-1]` => `bottomNode = p;` => hence `bottomNode` is a `rootNode` in the end.

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
ishantiw: it was incorrect, it should be `currentNode = await this.getNode(currentNode.rightHash);`

elements/lisk-tree/src/sparse_merkle_tree/prefix_store.ts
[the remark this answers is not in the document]
nazarhussain: We should follow same approach of either throwing error or returning undefined from state store. Better to create a separate issue for it.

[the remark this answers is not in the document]
nazarhussain: We should follow same approach of either throwing error or returning undefined from state store. Better to create a separate issue for it.

nazarhussain (approved):

elements/lisk-tree/src/sparse_merkle_tree/branch.ts
```diff
@@ -12,35 +12,43 @@
...
 	private _hash: Buffer;
-	public constructor(left: Leaf, right: Leaf) {
-		this._left = left;
-		this._right = right;
-		this._hash = branchHash(this._left.hash, this._right.hash);
+	private _data: Buffer;
```
Incede, line 23: maybe rename to `digest` or `digestBuffer` whichever you prefer, imo sounds better and more relatable to the implemented property.
similarly the function names in `utils` would have to be changed to `branchDigest` and so on
```suggestion
	private _digest: Buffer;
```

elements/lisk-tree/src/sparse_merkle_tree/branch.ts
```diff
@@ -12,35 +12,43 @@
...
+import { hash } from '@liskhq/lisk-cryptography';
+import { branchDataBuffer } from './utils';
 import { NodeSide } from './constants';
 
 export class Branch {
-	private _left: Leaf;
-	private _right: Leaf;
+	private _leftHash: Buffer;
+	private _rightHash: Buffer;
 	private _hash: Buffer;
```
Incede, line 22: I think this is no longer needed as a separate class attribute since it can always be computed from the buffer (data/digest). 
Same for `Leaf` class

elements/lisk-tree/src/sparse_merkle_tree/constants.ts
```diff
@@ -19,8 +19,16 @@ export const EMPTY_HASH = hash(EMPTY_VALUE);
 
 export const LEAF_HASH_PREFIX = Buffer.from('00', 'hex');
 export const BRANCH_HASH_PREFIX = Buffer.from('01', 'hex');
+export const PREFIX_HASH_TO_VALUE = Buffer.from([0]);
```
Incede, line 22: Is this being used anywhere ?

elements/lisk-tree/src/sparse_merkle_tree/constants.ts
```diff
@@ -19,8 +19,16 @@ export const EMPTY_HASH = hash(EMPTY_VALUE);
...
+export const NODE_HASH_SIZE = 32;
```
Incede, line 23: How did you derive this, Is it always 32 ?

elements/lisk-tree/src/sparse_merkle_tree/constants.ts
```diff
@@ -19,8 +19,16 @@ export const EMPTY_HASH = hash(EMPTY_VALUE);
...
 
 export const enum NodeSide {
 	LEFT = 0,
 	RIGHT = 1,
 }
+
+export const enum NodeType {
```
Incede, line 30: I think it's not being used anywhere

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -35,8 +40,114 @@ export class SparseMerkleTree {
 	public get db(): Database {
 		return this._db;
 	}
+
+	public async getNode(nodeHash: Buffer): Promise<TreeNode> {
+		const data = await this._db.get(nodeHash);
+
+		if (!data) {
+			throw new Error(`Hash does not exist in merkle tree: ${nodeHash.toString('hex')}`);
```
Incede, line 48:
```suggestion
			throw new Error(`Node with input hash : ${nodeHash.toString('hex')} does not exist in the tree`);
```

elements/lisk-tree/src/sparse_merkle_tree/utils.ts
```diff
@@ -182,8 +184,31 @@ export const binaryStringToBuffer = (str: string) => {
 	return buf;
 };
 
-export const leafHash = (key: Buffer, value: Buffer): Buffer =>
-	hash(Buffer.concat([LEAF_HASH_PREFIX, key, value]));
+export const getLeafData = (data: Buffer, keyLength: number): { key: Buffer; value: Buffer } => {
```
Incede, line 187: maybe rename for better readability
```suggestion
export const parseLeaf = (data: Buffer, keyLength: number): { key: Buffer; value: Buffer } => {
```

elements/lisk-tree/src/sparse_merkle_tree/utils.ts
```diff
@@ -182,8 +184,31 @@ export const binaryStringToBuffer = (str: string) => {
...
+	const key = data.slice(1, keyLength + 1);
+	// Get data
+	const value = data.slice(keyLength + 1, data.length);
+
+	return {
+		key,
+		value,
+	};
+};
+export const getBranchData = (data: Buffer): { leftHash: Buffer; rightHash: Buffer } => {
```
Incede, line 198: same here
```suggestion
export const parseBranch = (data: Buffer): { leftHash: Buffer; rightHash: Buffer } => {
```

elements/lisk-tree/src/sparse_merkle_tree/utils.ts
```diff
@@ -182,8 +184,31 @@ export const binaryStringToBuffer = (str: string) => {
...
+	// Get right hash
+	const rightHash = data.slice(-1 * NODE_HASH_SIZE);
+
+	return {
+		leftHash,
+		rightHash,
+	};
+};
+
+export const leafDataBuffer = (key: Buffer, value: Buffer): Buffer =>
```
Incede, line 210: similar suggestion
```suggestion
export const digestLeaf = (key: Buffer, value: Buffer): Buffer =>
```

elements/lisk-tree/src/sparse_merkle_tree/utils.ts
```diff
@@ -182,8 +184,31 @@ export const binaryStringToBuffer = (str: string) => {
...
+	Buffer.concat([LEAF_HASH_PREFIX, key, value]);
 
-export const branchHash = (leftHash: Buffer, rightHash: Buffer): Buffer =>
-	hash(Buffer.concat([BRANCH_HASH_PREFIX, leftHash, rightHash]));
+export const branchDataBuffer = (leftHash: Buffer, rightHash: Buffer): Buffer =>
```
Incede, line 213: same
```suggestion
export const digestBranch = (leftHash: Buffer, rightHash: Buffer): Buffer =>
```

Incede (changes requested):
Suggested name changes for better readability.
Also, not sure about the `slice` logic for branch & leaf so I'd still wait for the tests.

elements/lisk-tree/src/sparse_merkle_tree/constants.ts
ishantiw: we are using SHA256 so it should always give out 32 byte hash

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -35,8 +45,113 @@ export class SparseMerkleTree {
...
+				currentNode = await this.getNode(currentNode.rightHash);
+			}
+			h += 1;
+		}
+
+		// The currentNode is an empty node, newLeaf will replace the default empty node or currentNode will be updated to newLeaf
+		let bottomNode: TreeNode = new Empty();
+		if (currentNode instanceof Empty) {
+			// delete the empty node and update the tree, the new leaf will substitute the empty node
+			await this._db.del(currentNode.hash);
```
ricott1, line 106: aren't you deleting all empty nodes in the tree here?

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -12,20 +12,30 @@
...
 	private readonly _keyLength: number;
-	public constructor(db: Database, rootHash?: Buffer, keyLength = 36) {
-		this._db = db;
-		this._keyLength = keyLength;
-		this._rootHash = rootHash ?? EMPTY_HASH;
+	private _rootNode: TreeNode;
+
+	public constructor(options: { db?: Database; rootHash?: Buffer; keyLength?: number }) {
+		this._db = options?.db ?? new InMemoryDB();
+		this._keyLength = options?.keyLength ?? 32;
```
Incede, line 31: I think for us `keyLength` default value should be 36
https://github.com/LiskHQ/lisk-sdk/issues/6511#discussion_r662412630

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -35,8 +45,113 @@ export class SparseMerkleTree {
...
+			bottomNode = newLeaf;
+		} else {
+			// We need to create new branches in the tree to fulfill the
+			// Condition of one leaf per empty subtree
+			// Note: h is set to the last value from the previous loop
+			const currentNodeBinaryKey = bufferToBinaryString(currentNode.key);
+			while (binaryKey.charAt(h) === currentNodeBinaryKey.charAt(h)) {
+				// Create branch node with empty value
+				const newBranch = new Branch(EMPTY_HASH, EMPTY_HASH);
+				await this._db.set(newBranch.hash, newBranch.digest);
```
ricott1, line 118: you dont have to store it in the db at this point, I would store it only once it is finalized

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -12,20 +12,30 @@
...
-		this._keyLength = keyLength;
-		this._rootHash = rootHash ?? EMPTY_HASH;
+	private _rootNode: TreeNode;
+
+	public constructor(options: { db?: Database; rootHash?: Buffer; keyLength?: number }) {
+		this._db = options?.db ?? new InMemoryDB();
+		this._keyLength = options?.keyLength ?? 32;
+		this._rootNode = new Empty();
 	}
 	public get rootHash(): Buffer {
```
Incede, line 34: I think this `get rootHash` is not needed since you can get the `rootNode`. Also, the class does not have any property `rootHash` now.

ricott1 (changes requested):
some minor points, maybe the one about deleting empty nodes from the db is more important

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
ishantiw: removed it

elements/lisk-tree/src/sparse_merkle_tree/branch.ts
ishantiw: yeah I agree but still it is handy to use as a property instead of using hash function within update/delete funcs. but not any strong opinion on this

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -12,20 +12,27 @@
...
-		this._db = db;
-		this._keyLength = keyLength;
-		this._rootHash = rootHash ?? EMPTY_HASH;
+	private _rootNode: TreeNode;
+
+	public constructor(options: { db?: Database; keyLength?: number }) {
+		this._db = options?.db ?? new InMemoryDB();
+		this._keyLength = options?.keyLength ?? DEFAULT_KEY_LENGTH;
+		this._rootNode = new Empty();
 	}
```
Incede, line 33: 1. I think we still need `rootHash` property in constructor for the scenario when it's needed to instantiate a new SMT from previous state root (from previous block header). 
2. It could be tricky to default use `InMemoryDB` if database is not provided because depending upon the user's system and the size of the SMT the system could quickly run out of memory and the library can fail. So, I'd suggest to make the db parameter non-optional so that the user knows up-front what he's doing.
```suggestion
		public constructor(db: Database, rootHash?: Buffer, keyLength?: number) {
		this._db = db;
		this._keyLength = keyLength ?? DEFAULT_KEY_LENGTH;
		if (rootHash) {
			const { leftHash, rightHash } = parseBranch(rootHash);
			this._rootNode = new Branch(leftHash, rightHash);
		}
		else {
			this._rootNode = new Empty();
		}
	}
```

Incede (changes requested):

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
ishantiw: 1. Yeah I actually kept that property with that in mind but completely forgot and removed it when I saw your previous comment. I will revert it back.
2. Good point, I will make it required prop for the constructor

Incede: Okay, although for point 1, I still think we don't need `rootHash` or `get rootHash` as explicit properties of SMT since we have `rootNode`. But it should be possible to initialize using `rootHash` like in the suggestion above.

ishantiw: I'm removing `rootNode` as we start with empty root node whereas the `rootHash` can have value other than empty hash. In order to retrieve the value of rootHash we need to do async call to db in the constructor which is not possible or we create a separate `init()` function

elements/lisk-tree/src/sparse_merkle_tree/utils.ts
```diff
@@ -229,8 +231,31 @@ export const binaryStringToBuffer = (str: string) => {
...
+	};
+};
+
+export const leafDigest = (key: Buffer, value: Buffer): Buffer =>
+	Buffer.concat([LEAF_HASH_PREFIX, key, value]);
 
-export const branchHash = (leftHash: Buffer, rightHash: Buffer): Buffer =>
-	hash(Buffer.concat([BRANCH_HASH_PREFIX, leftHash, rightHash]));
+export const branchDigest = (leftHash: Buffer, rightHash: Buffer): Buffer =>
+	Buffer.concat([BRANCH_HASH_PREFIX, leftHash, rightHash]);
```
nazarhussain, line 261: I don't think these functions names are conceptually correct here. We are not creating any digest rather creating message itself to be digested. Would rather suggest to use `branchMessage` or `branchData` or `calcBranchHashMessage` 

> A message digest is a cryptographic hash function containing a string of digits created by a one-way hashing formula.

nazarhussain (approved):

ricott1 (approved):

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -35,8 +42,114 @@ export class SparseMerkleTree {
...
+
+			return new Leaf(key, value);
+		}
+
+		const { leftHash, rightHash } = parseBranch(data);
+
+		return new Branch(leftHash, rightHash);
+	}
+	// As specified in from https://github.com/LiskHQ/lips/blob/master/proposals/lip-0039.md
+	public async update(key: Buffer, value: Buffer) {
```
Incede, line 68: Is it not necessary to declare the return type ?
```suggestion
	public async update(key: Buffer, value: Buffer): Promise<TreeNode> {
```

Incede (changes requested):
Add 1 last minor comment.
Besides it'd be nice to add a test case for `keyLength = 38`

elements/lisk-tree/test/sparse_merkle_tree/utils.spec.ts
```diff
@@ -416,4 +422,113 @@ describe('utils', () => {
...
+				str:
+					'1101101100010011101111100000000100111100101000011011000011100111000100110010100000001111100110111100011000110000000101111111001010100011010001010010001110010001111010011000101100000001100010010011010111100000000010110110010110111010010101011001101100100101001001111101100010010010010100001110000111000000',
+				buf: Buffer.from(
+					'db13be013ca1b0e713280f9bc63017f2a3452391e98b018935e00b65ba559b2527d89250e1c0',
+					'hex',
+				),
+				keyLength: 38,
+			},
+			{
+				str:
```
Incede, line 460: I am assuming you have verified these fixtures.

Incede (approved):
LGTM

elements/lisk-tree/test/sparse_merkle_tree/utils.spec.ts
ishantiw: I have verified this with 2 different online tools, and it could be verified by anyone manually expanding each byte

PR #6576, opened 2021-07-12
Fix handling of genesis block - Closes #6573

### LINKED ISSUE #6573: Handling of genesis block takes too much time
### Expected behavior

Even with a large genesis block file, it should be able to sync the nodes properly.

### Actual behavior

- Procedure `getHighestCommonBlock` for given block ids returns the whole common block header, in case of testnet or mainnet genesis block header the size of genesis block header is too big (20-200MB).
- When getting the block, it tries to decode the asset which takes long.
- Genesis block are kept in memory

### Steps to reproduce

- Migrate 2.1.6 and generate genesis block (which is around 200MB)
- Run `lisk-core start -n mainnet` with the new genesis block
- Connect to running node as seed peer
- Observer the syncing node fails to request block during synchronisation

5.1.1, 3.0.0

shuse2:
### What was the problem?

This PR resolves #6573 

### How was it solved?

- Update not to hold genesis block in memory
- Update getHighestCommonBlock to getHighestCommonBlockID

### How was it tested?

Run 2 nodes with heavy genesis block (modified mainnet snapshot etc)
- Check 2 nodes will sync even after 500 blocks. (after 500 blocks, genesis block will not be on cache)
- Restart one of the node and check the starting time is significantly faster than 1st time

### DIFF
--- a/elements/lisk-chain/src/data_access/storage.ts
+++ b/elements/lisk-chain/src/data_access/storage.ts
@@ -292,6 +292,12 @@ export class Storage {
 		}
 	}
 
+	// Warning: This function should never be used. This exist only for migration purpose.
+	// Specifically, only to set genesis state between 5.1.2 => 5.1.3
+	public async setConsensusState(key: string, val: Buffer): Promise<void> {
+		await this._db.put(`${DB_KEY_CONSENSUS_STATE}:${key}`, val);
+	}
+
 	/*
 		Accounts
 	*/
--- a/elements/lisk-chain/src/data_access/storage.ts
+++ b/elements/lisk-chain/src/data_access/storage.ts
@@ -292,6 +292,7 @@ export class Storage {
 		}
 	}
 
+	// TODO: Remove in next version
 	// Warning: This function should never be used. This exist only for migration purpose.
 	// Specifically, only to set genesis state between 5.1.2 => 5.1.3
 	public async setConsensusState(key: string, val: Buffer): Promise<void> {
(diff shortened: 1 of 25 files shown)


elements/lisk-chain/src/data_access/storage.ts
```diff
@@ -292,6 +292,12 @@ export class Storage {
 		}
 	}
 
+	// Warning: This function should never be used. This exist only for migration purpose.
```
ManuGowda, line 295: Also add a TODO? to be removed in 6.0.0

elements/lisk-chain/src/types.ts
```diff
@@ -107,3 +107,9 @@ export interface Validator {
 	minActiveHeight: number;
 	isConsensusParticipant: boolean;
 }
+
+export interface GenesisInfo {
+	height: number;
+	timestamp: number;
```
ManuGowda, line 113: are we using timestamp somewhere? i only saw height and initRounds usage, if not used maybe remove?

framework/src/node/transport/transport.ts
```diff
@@ -196,10 +197,7 @@ export class Transport {
 		return codec.encode(getBlocksFromIdResponseSchema, { blocks: encodedBlocks });
 	}
 
-	public async handleRPCGetHighestCommonBlock(
-		data: unknown,
-		peerId: string,
-	): Promise<Buffer | undefined> {
+	public async handleRPCGetHighestCommonBlock(data: unknown, peerId: string): Promise<Buffer> {
```
ManuGowda, line 200:
```suggestion
	public async handleRPCGetHighestCommonBlockID(data: unknown, peerId: string): Promise<Buffer> {
```

ManuGowda (approved):

ishantiw (approved):

PR #6577, opened 2021-07-12
Update database to have key as Buffer - Closes #6562

### LINKED ISSUE #6562: Update database to have key as Buffer
### Description

Currently we use  key as `string`  in the database. However, in the new state machine, everything is handled as bytes.
Therefore, all  the key usage in elements/lisk-db/src/kv_store.ts should be updated to Buffer.
Consequently, all the usage of KVStore needs to be updated to use Buffer.

### Acceptance Criteria

All the source code and tests should be updated to use the key as Buffer and pass the CI.

### Additional Information
Related Issue #6554

nazarhussain:
### What was the problem?

This PR resolves #6562

### How was it solved?

- Updated `lisk-db` kv store to accept keys as buffers
- Updated all dependent packages 
- Updated all tests 

### How was it tested?

- Run unit tests


elements/lisk-chain/src/constants.ts
```diff
@@ -14,9 +14,9 @@
 
 import { createMessageTag, hash } from '@liskhq/lisk-cryptography';
 
-export const CHAIN_STATE_BURNT_FEE = 'burntFee';
-export const CONSENSUS_STATE_FINALIZED_HEIGHT_KEY = 'finalizedHeight';
-export const CONSENSUS_STATE_VALIDATORS_KEY = 'validators';
+export const CHAIN_STATE_BURNT_FEE = Buffer.from('burntFee', 'utf8');
+export const CONSENSUS_STATE_FINALIZED_HEIGHT_KEY = Buffer.from('finalizedHeight', 'utf8');
+export const CONSENSUS_STATE_VALIDATORS_KEY = Buffer.from('validators', 'utf8');
```
nazarhussain, line 19: @shuse2 To save storage we can also use 1 byte keys here... e.g. `0x01`, `0x02`. Since on storage level every key is just binary buffer so it will save storage and we will still have context in the code.

elements/lisk-chain/src/data_access/constants.ts
```diff
@@ -12,12 +12,12 @@
...
-export const DB_KEY_DIFF_STATE = 'diff';
+export const DB_KEY_BLOCKS_ID = Buffer.from('blocks:id', 'utf8');
+export const DB_KEY_BLOCKS_HEIGHT = Buffer.from('blocks:height', 'utf8');
+export const DB_KEY_TRANSACTIONS_BLOCK_ID = Buffer.from('transactions:blockID', 'utf8');
+export const DB_KEY_TRANSACTIONS_ID = Buffer.from('transactions:id', 'utf8');
+export const DB_KEY_TEMPBLOCKS_HEIGHT = Buffer.from('tempBlocks:height', 'utf8');
+export const DB_KEY_ACCOUNTS_ADDRESS = Buffer.from('accounts:address', 'utf8');
+export const DB_KEY_CHAIN_STATE = Buffer.from('chain', 'utf8');
+export const DB_KEY_CONSENSUS_STATE = Buffer.from('consensus', 'utf8');
+export const DB_KEY_DIFF_STATE = Buffer.from('diff', 'utf8');
```
nazarhussain, line 23: Same applies here.

elements/lisk-chain/src/constants.ts
shuse2: yes, when we apply the state model changes, it will all be bytes. for now, i think this is fine

shuse2: Created new issue https://github.com/LiskHQ/lisk-sdk/issues/6586

shuse2 (approved):
Functional test is still failing on type, but otherwise LGTM

ManuGowda (approved):

elements/lisk-chain/src/state_store/state_store.ts
```diff
@@ -79,7 +80,7 @@ export class StateStore {
 		this.chain.restoreSnapshot();
 	}
 
-	public finalize(height: string, batch: BatchChain): void {
+	public finalize(height: number | bigint, batch: BatchChain): void {
```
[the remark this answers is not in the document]
shuse2: yes, height is always number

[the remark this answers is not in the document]
nazarhussain: Yes it should not, I remember there were some mixed use of number and bigint may be for tests. Let me fix this one.

PR #6580, opened 2021-07-14
Implement remove function for sparse Merkle tree - Closes #6494

### LINKED ISSUE #6494: Implement remove function for sparse Merkle tree
### Description

The `remove` function will remove a key/value pair from the dataset by deleting the corresponding leaf node in the tree and rearranging the affected nodes in the tree. The Merkle root is updated accordingly.

### Motivation

To delete data/nodes from the dataset/tree

### Acceptance Criteria

Add test cases to match the output in the protocol specifications (fixtures) provided.

### Additional Information

Related issues : #6485

ManuGowda:
### What was the problem?

This PR resolves #6494

### How was it solved?

- Implement SMT `remove` interface according to [specification](https://github.com/LiskHQ/lips-staging/blob/7520afe579d20f6e943d30de475eb0ce186b04c1/proposals/lip-Introduce-sparse-Merkle-trees.md#removing-a-leaf-node)

### How was it tested?

- Run `npm run tests` from lisk-tree


elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -149,10 +149,99 @@ export class SparseMerkleTree {
...
+			// an exact emptyHash check is required
+			// not using EMPTY_HASH here to make sure we use correct hash from Empty class
+			const emptyHash = new Empty().hash;
+			while (h > 0) {
+				const p = ancestorNodes[h - 1] as Branch;
+
+				// if one of the children is empty then break the condition
+				if (
+					p instanceof Branch &&
+					!p.leftHash.equals(emptyHash) &&
```
ricott1, line 209: another option would be to get the node from left and right hash with `await this.getNode(p.leftHash);` and check using instanceof. This would be more in line with the checks we do before, and avoid having the accessory definition on line 202

ricott1 (approved):
I'm not checking the utils function as you mentioned. The code looks fine, just a minor comment on how to do the check against the empty node, up to you

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
ManuGowda: either way its fine with me, but it will multiple additional call to db? which can be avoided unless the condition is wrong?

ishantiw (approved):
LGTM

Incede (approved):

Issue #6497, opened 2021-06-22
Implement generateMultiProof function for sparse Merkle tree


Incede:
### Description

Given an array of keys `queryKeys`, the function `generateMultiProof(queryKeys)` will generate a query response which can be used to verify if the members of `queryKeys` are present in the sparse Merkle tree or not.

The response has the properties `siblingHashes` and `queries`.
`queries` is an array of objects, one per each key in `queryKeys` and contains the following properties : 
- `key`
- `bitmap`
- `value`

### Motivation

* To provide proof of inclusion for multiple keys
* To provide proof of exclusion for multiple keys

### Acceptance Criteria
Add test cases to match the protocol specifications

### Additional Information

Related issues : #6485 #6495 #6496

PR #6584, opened 2021-07-16
Implement proof functions for sparse Merkle tree - Closes #6495 #6497

### LINKED ISSUE #6495: Implement generateSingleProof function for sparse Merkle tree
### Description

The `generateSingleProof` function will generate the query response to a single key, including the sibling hashes and the hash of the visited nodes. This response can be used to verify if the key is present in the sparse Merkle tree or not.

The response has following properties :
- `siblingHashes`
- `ancestorHashes`
- `binaryBitmap`
- `key`
- `value` 

### Motivation

* To provide proof of inclusion for a key 
* To provide proof a exclusion for key

### Acceptance Criteria
 Add test cases to match the protocol specifications provided

### Additional Information
Related issues : #6485

Incede:
### What was the problem?

This PR resolves #6495 #6497 

### How was it solved?

* Followed [LIP-0039](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0039.md)

### How was it tested?

* Added unit tests
* `npm run test` in `elements/lisk-tree`


elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -239,9 +260,171 @@ export class SparseMerkleTree {
...
+				siblingHashes,
+				ancestorHashes,
+				binaryBitmap,
+				key: currentNode.key,
+				value: currentNode.value,
+			};
+		}
+		if (
+			currentNode instanceof Leaf &&
+			currentNode.key.toString('hex') === queryKey.toString('hex')
```
shuse2, line 330:
```suggestion
			currentNode.key.equal(queryKey)
```

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -239,9 +260,171 @@ export class SparseMerkleTree {
...
+			ancestorHashes.push(currentNode.hash); // in case the leaf is sibling to another node
+			return {
+				siblingHashes,
+				ancestorHashes,
+				binaryBitmap,
+				key: currentNode.key,
+				value: currentNode.value,
+			};
+		}
+		return {
```
shuse2, line 342: in which branch in the LIP this happens?

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -239,9 +260,171 @@ export class SparseMerkleTree {
 		return bottomNode;
 	}
 
-	/*
-		public generateSingleProof() {}
-		public generateMultiProof() {}
-		public verifyMultiProof() {}
-		*/
+	public generateSingleProof = async (queryKey: Buffer): Promise<SingleProof> => {
```
shuse2, line 263:
```suggestion
	public async generateSingleProof(queryKey: Buffer): Promise<SingleProof> {
```

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -239,9 +260,171 @@ export class SparseMerkleTree {
...
+		return {
+			key: queryKey,
+			value: EMPTY_VALUE,
+			binaryBitmap: bufferToBinaryString(EMPTY_HASH),
+			siblingHashes: [],
+			ancestorHashes: [],
+		};
+	};
+
+	public generateMultiProof = async (queryKeys: Buffer[]): Promise<Proof> => {
```
shuse2, line 351:
```suggestion
	public async generateMultiProof(queryKeys: Buffer[]): Promise<Proof> {
```

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -239,9 +260,171 @@ export class SparseMerkleTree {
...
+		const partialQueries: SingleProof[] = [];
+		for (const queryKey of queryKeys) {
+			const query = await this.generateSingleProof(queryKey);
+			partialQueries.push(query);
+		}
+
+		const queries: Query[] = [...partialQueries].map(sp => ({
```
shuse2, line 358: No need to change it now, but in general, `map` is slower, so we prefer `for` loop in the low level libraries where performance is important.

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -239,9 +260,171 @@ export class SparseMerkleTree {
...
+			binaryBitmap: sp.binaryBitmap,
+			key: sp.key,
+			value: sp.value,
+			siblingHashes: sp.siblingHashes,
+			height: sp.binaryBitmap.length,
+		}));
+		sortedQueries = sortByBitmapAndKey(sortedQueries);
+
+		while (sortedQueries.length > 0) {
+			const sp = sortedQueries.shift()!;
```
shuse2, line 375: `shift` is very slow, so prefer not to use them. it re-indexes the whole array

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -239,9 +260,171 @@ export class SparseMerkleTree {
...
+			if (sp.height === 0) {
+				continue;
+			}
+			const b = sp.binaryBitmap.charAt(sp.binaryBitmap.length - sp.height);
+			if (b === '1') {
+				const nodeHash = sp.siblingHashes.pop()!;
+				let isPresentInSiblingHashes = false;
+				let isPresentInAncestorHashes = false;
+				for (const i of siblingHashes) {
+					if (i.equals(nodeHash)) isPresentInSiblingHashes = true;
```
shuse2, line 385:
```
if (i.equals(nodeHash)) {
  isPresentInSiblingHashes = true;
  break;
}
```
same for below. but probably union logic can be optimize in other way

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -239,9 +260,171 @@ export class SparseMerkleTree {
...
+				};
+				const insertIndex = binarySearch(
+					sortedQueriesWithBinaryKey,
+					callback => treeSort(spWithBinaryKey, callback) < 0,
+				);
+				if (insertIndex === sortedQueries.length) sortedQueries.push(sp);
+				else {
+					const keyPrefix = bufferToBinaryString(sp.key).substring(0, sp.height);
+					const query = sortedQueries[insertIndex];
+					if (!bufferToBinaryString(query.key).endsWith(keyPrefix, query.height))
```
shuse2, line 420:
```suggestion
					if (!bufferToBinaryString(query.key).endsWith(keyPrefix, query.height)) {
					   sortedQueries.splice(insertIndex, 0, sp);
					}
```

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -239,9 +260,171 @@ export class SparseMerkleTree {
...
+					if (i.equals(nodeHash)) isPresentInAncestorHashes = true;
+				}
+				if (!isPresentInSiblingHashes && !isPresentInAncestorHashes) {
+					// TODO : optimize this
+					siblingHashes.push(nodeHash);
+				}
+			}
+			sp.height -= 1;
+
+			if (sortedQueries.length > 0) {
```
shuse2, line 397: where is this logic matching to in the LIP?

shuse2 (changes requested):

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
Incede: Any suggestions ?

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
Incede: It's not supposed to happen, this condition should never reach. I added it because there's no default return statement for the function. Maybe there's another way around to default return ?

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
shuse2: Maybe keep track of the index of `sortedQuery` might be better?
but maybe not so important if the sortedQueries doesn't grow much in size.

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
Incede: In the LIP it is mentioned in the beginning of the `while` loop :
```
        sortedQueries = sortByBitmapAndKey(sortedQueries)
```

ricott1 (changes requested):
I still haven't found the error in the proof generation

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -239,9 +260,170 @@ export class SparseMerkleTree {
...
+					siblingHashes: sp.siblingHashes,
+					height: sp.height,
+				};
+				const insertIndex = binarySearch(
+					sortedQueriesWithBinaryKey,
+					callback => treeSort(spWithBinaryKey, callback) < 0,
+				);
+				if (insertIndex === sortedQueries.length) sortedQueries.push(sp);
+				else {
+					const keyPrefix = bufferToBinaryString(sp.key).substring(0, sp.height);
```
ishantiw, line 417: shouldn't this be using `binaryExpansion`?
```
const keyPrefix = binaryExpansion(sp.key, this.keyLength).substring(0, sp.height);
```

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
ishantiw: here too, use `binaryExpansion` for key

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
Incede: I think, this logic does not works.

shuse2: i think i just meant the `if` format should follow the standard

ishantiw: I think still will be nice to early break the loop once the condition is met as Shu suggested. Earlier, the problem was it was breaking the outer `while` loop in the absence of parenthesis
```suggestion
					if (i.equals(nodeHash)) { 
					        isPresentInSiblingHashes = true;
					        break;
					}
```

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -239,9 +260,168 @@ export class SparseMerkleTree {
...
+			const b = sp.binaryBitmap.charAt(sp.binaryBitmap.length - sp.height);
+			if (b === '1') {
+				const nodeHash = sp.siblingHashes.pop()!;
+				let isPresentInSiblingHashes = false;
+				let isPresentInAncestorHashes = false;
+				for (const i of siblingHashes) {
+					if (i.equals(nodeHash)) isPresentInSiblingHashes = true;
+				}
+				for (const i of ancestorHashes) {
+					if (i.equals(nodeHash)) isPresentInAncestorHashes = true;
```
ishantiw, line 385: same here

Incede: nope, this doesn't work either.

ishantiw (approved):

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
Incede: https://github.com/LiskHQ/lisk-sdk/issues/6597

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
Incede: https://github.com/LiskHQ/lisk-sdk/issues/6597

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
```diff
@@ -239,9 +260,174 @@ export class SparseMerkleTree {
...
+				ancestorHashes,
+				binaryBitmap,
+				key: queryKey,
+				value: EMPTY_VALUE,
+			};
+		}
+
+		if (
+			currentNode instanceof Leaf &&
+			currentNode.key.toString('hex') !== queryKey.toString('hex')
```
shuse2, line 316:
```suggestion
			!currentNode.key.equals(queryKey)
```

elements/lisk-tree/src/sparse_merkle_tree/sparse_merkle_tree.ts
shuse2: i think logically `currentNode instanceof Leaf` is not needed because only possible case is the leaf.
Therefore, you could just check current `currentNode.key.equals(queryKey)` then return x, otherwise return y without last return

shuse2 (approved):
LGYM, just minor suggestion to organize the if statements

ricott1 (approved):
great job!

PR #6585, opened 2021-07-16
Update Leafs of Regular Merkle Tree - Closes #6493

### LINKED ISSUE #6493: Implement function to update leaf value on Merkle tree
### Description

Add method to update leaf node on the merkle tree with full data

### Specification

Add update method in the Merkle tree class with full data
```
function update(idxs, updateData):
  updateHashes = [leafHash(data) for data in updateData] 
  size = getSize()
  idxs = [idx]
  siblingHashes = getSiblingHashes(idxs)
  calculatedTree = calculatePathNodes(updateHashes, size, idxs, siblingHashes)
  store elements of calculatedTree 
  // The root node has index 0
  calculatedRoot = calculatedTree["0"]
  return calculatedRoot
```
and
```
function calculateRootFromUpdateData(updateData, proof):
  updateHashes = [leafHash(data) for data in updateData] 
  [size, idxs, siblingHashes] = proof
  calculatedTree = calculatePathNodes(updateHashes, size, idxs, siblingHashes)
  // The root node has index 0
  calculatedRoot = calculatedTree["0"]
  return calculatedRoot
```

### Acceptance Criteria

- Add unit test to check updated merkle tree is correct using the existing protocol-spec

### Additional Information

Related issues : #6485

mehmetegemen:
### What was the problem?

This PR resolves  #6493

### How was it solved?

- Added calculatePathNodes, getSiblingHashes, update and calculateRootFromUpdateData functions

### How was it tested?

- Created a merkle tree with initial values, created another merkle tree with 2 different leafs, and checked if calculated tree of first merkle with updated data equals to merkle tree with 2 different leafs


elements/lisk-tree/src/merkle_tree/calculate.ts
```diff
@@ -0,0 +1,153 @@
...
+import { LEAF_PREFIX, NODE_HASH_SIZE } from "./constants";
+import { NodeIndex, NodeInfo, NodeLocation, NodeSide, NonNullableStruct, Proof, SiblingHash } from "./types";
+import { buildBranch, createNewBranchNode, createNewLeafNode, generateNode, getBinaryString,
+    getHeight, getPairLocation, getParentLocation, isLocationEqualToSiblingHash } from "./utils";
+
+export const calculatePathNodes = (
+    queryData: ReadonlyArray<Buffer>,
+    size: number,
+    idxs: ReadonlyArray<NodeIndex>,
+    siblingHashes: ReadonlyArray<SiblingHash>) => {
```
shuse2, line 11: type of idxs and siblingHashes should be
```
idxs: ReadonlyArray<number>,
siblingHashes: ReadonlyArray<Buffer>,
```
otherwise input of https://github.com/LiskHQ/lisk-sdk/pull/6585/files#diff-ab59b447c02797885741efe9ce43a0320e12f69ab1675e638f825b52637efe57R149 will be invalid
also, it should follow https://github.com/LiskHQ/lips-staging/blob/master/proposals/lip-0031.md#proof-serialization

shuse2: We might need to change the `number` to `bigint` later, but this will apply everywhere else, so let's not change it now

elements/lisk-tree/src/index.ts
```diff
@@ -14,3 +14,4 @@
 
 export { MerkleTree } from './merkle_tree/merkle_tree';
 export { verifyProof } from './merkle_tree/verify_proof';
+export { calculatePathNodes, calculateRootFromUpdateData } from './merkle_tree/calculate';
```
shuse2, line 17: `calculatePathNodes` does not need to be exported

elements/lisk-tree/src/merkle_tree/merkle_tree.ts
```diff
@@ -267,6 +272,134 @@ export class MerkleTree {
...
+
+			if (parentBinaryIndex !== '0') {
+				indexes[parentBinaryIndex] = parentIndex;
+			}
+		}
+
+		return siblingHashes;
+	}
+
+	public async update(idxs: ReadonlyArray<NodeIndex>, updateData: ReadonlyArray<Buffer>) {
```
shuse2, line 370: same here as above. index should be array of number

elements/lisk-tree/src/merkle_tree/calculate.ts
mehmetegemen: done

elements/lisk-tree/src/index.ts
mehmetegemen: done

elements/lisk-tree/src/merkle_tree/merkle_tree.ts
```diff
@@ -267,6 +270,141 @@ export class MerkleTree {
...
+		
+		for (const updatedNode of Object.values(calculatedTree)) {
+				const existingNodeHash = await this._locationToHashMap.get(
+					getBinaryString(updatedNode.nodeIndex, this._getHeight() - updatedNode.layerIndex),
+				);
+				await this._locationToHashMap.set(
+					getBinaryString(updatedNode.nodeIndex, this._getHeight() - updatedNode.layerIndex),
+					updatedNode.hash,
+				);
+				await this._hashToValueMap.set(updatedNode.hash, updatedNode.value);
```
mehmetegemen, line 398: This should get the node value from updateData I guess

elements/lisk-tree/src/merkle_tree/types.ts
```diff
@@ -37,6 +38,12 @@ export const enum NodeSide {
 export interface TreeStructure {
 	[key: number]: NodeInfo[];
 }
+
+export interface NodeIndex {
+	layerIndex: number | undefined;
+	nodeIndex: number | undefined;
+}
+
```
mehmetegemen, line 46: Since `side` is optional in `NodeLocation`, I believe I can delete `NodeIndex` from types.

mehmetegemen: Though `NodeIndex` contains the possibility of `undefined` values whereas `NodeLocation` doesn't, hmm.

elements/lisk-tree/src/merkle_tree/merkle_tree.ts
mehmetegemen: done

elements/lisk-tree/src/merkle_tree/merkle_tree.ts
mehmetegemen: done

elements/lisk-tree/src/merkle_tree/merkle_tree.ts
```diff
@@ -267,6 +282,176 @@ export class MerkleTree {
...
+				1,
+				serializedIndexBinaryString.length,
+			);
+			const location = {
+				nodeIndex: parseInt(indexBinaryString, 2),
+				layerIndex: treeHeight - indexBinaryString.length,
+			};
+
+			sortedIndexes.push(location);
+		}
```
mehmetegemen, line 305: should use `getLocationFromIndex()` here

elements/lisk-tree/src/merkle_tree/utils.ts
```diff
@@ -193,3 +201,136 @@ export const calculateMerkleRoot = ({ value, appendPath, size }: MerkleRootInfo)
...
+
+export const getLocationFromIndex = (index: number, size: number) => {
+	const treeHeight = Math.ceil(Math.log2(size)) + 1;
+
+	const serializedIndexBinaryString = index.toString(2);
+	const indexBinaryString = serializedIndexBinaryString.substring(
+		1,
+		serializedIndexBinaryString.length,
+	);
+	const location = {
```
ricott1, line 330: I think at some point here we should check that the index is not larger than the layer size

ricott1 (changes requested):
Minor comments. Another one: we don't check anywhere that the idx in the idxs array of the proof is smaller than the size (maybe I missed it)? For instance, we should check it properly in the `getLocationFromIndex` function

elements/lisk-tree/src/merkle_tree/merkle_tree.ts
```diff
@@ -267,6 +282,176 @@ export class MerkleTree {
 		return rightWitness;
 	}
 
+	public async getSiblingHashes(idxs: ReadonlyArray<number>): Promise<Buffer[]> {
```
shuse2, line 285: this should be `private`? also private function should start from `_`

elements/lisk-tree/src/merkle_tree/merkle_tree.ts
shuse2: when we update `hashToValue` we should also update `locationToHash`

elements/lisk-tree/src/merkle_tree/merkle_tree.ts
```diff
@@ -267,6 +282,176 @@ export class MerkleTree {
...
+					updateDataOfCalculatedPathLeafs[binaryIndex],
+				);
+			} else {
+				await this._hashToValueMap.set(updatedNode.hash, updatedNode.value);
+			}
+
+			await this._hashToValueMap.del(existingNodeHash as Buffer);
+		}
+
+		const calculatedRoot = calculatedTree['0'].hash;
```
shuse2, line 449: maybe define `0` as rootBinaryIndex to avoid magic number?

elements/lisk-tree/src/merkle_tree/merkle_tree.ts
```diff
@@ -267,6 +282,176 @@ export class MerkleTree {
...
+			if (binaryIndex in updateDataOfCalculatedPathLeafs) {
+				await this._hashToValueMap.set(
+					updatedNode.hash,
+					updateDataOfCalculatedPathLeafs[binaryIndex],
+				);
+			} else {
+				await this._hashToValueMap.set(updatedNode.hash, updatedNode.value);
+			}
+
+			await this._hashToValueMap.del(existingNodeHash as Buffer);
```
shuse2, line 446: why do we need conversion here?

shuse2 (changes requested):

elements/lisk-tree/src/merkle_tree/merkle_tree.ts
```diff
@@ -267,6 +282,176 @@ export class MerkleTree {
...
+	public async getSiblingHashes(idxs: ReadonlyArray<number>): Promise<Buffer[]> {
+		if (this._size === 0) {
+			return [];
+		}
+
+		const treeHeight = Math.ceil(Math.log2(this._size)) + 1;
+
+		const sortedIndexes: NodeIndex[] = [];
+		for (const index of idxs) {
+			const serializedIndexBinaryString = index.toString(2);
```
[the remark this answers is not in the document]
ricott1: does that work on the encoded or ''normal'' index?

elements/lisk-tree/src/merkle_tree/merkle_tree.ts
```diff
@@ -267,6 +282,176 @@ export class MerkleTree {
...
+				serializedIndexBinaryString.length,
+			);
+			const location = {
+				nodeIndex: parseInt(indexBinaryString, 2),
+				layerIndex: treeHeight - indexBinaryString.length,
+			};
+
+			sortedIndexes.push(location);
+		}
+		// Remove empty indexes
```
ricott1, line 306: I would do this before nin the loop of line 293. Just check there directly the encoded index against a constant flag (00 hex), and in case it is equal, just continue

ricott1 (changes requested):

elements/lisk-tree/src/merkle_tree/calculate.ts
```diff
@@ -0,0 +1,183 @@
+import { hash } from '../../../lisk-cryptography/dist-node';
+import { LEAF_PREFIX, NODE_HASH_SIZE } from './constants';
```
mitsuaki-u, line 2: why is there an entire new `calculate` file? we can put this in `utils`

mehmetegemen: I think it was Shu's and Alessandro's idea but don't remember the details now

elements/lisk-tree/src/merkle_tree/calculate.ts
```diff
@@ -0,0 +1,47 @@
...
+		const leafValueWithoutNodeIndex = Buffer.concat(
+			[LEAF_PREFIX, data],
+			LEAF_PREFIX.length + data.length,
+		);
+		const leafHash = hash(leafValueWithoutNodeIndex);
+		updateHashes.push(leafHash);
+	}
+
+	const calculatedTree = calculatePathNodes(updateHashes, size, indexes, siblingHashes);
+	const calculatedRoot = calculatedTree.get(2); // 2 is the index for root "10"
```
ricott1, line 45: In my understanding we now use indexes that are always encoded (with the prepended 1). Should we clarify this at some point? maybe at the beginning of the file in a comment or even better by choosing a different variable name for indexes? (maybe encodedIndexes?)

elements/lisk-tree/src/merkle_tree/merkle_tree.ts
```diff
@@ -267,13 +276,100 @@ export class MerkleTree {
 		return rightWitness;
 	}
 
+	public async update(idxs: number[], updateData: ReadonlyArray<Buffer>): Promise<Buffer> {
```
ricott1, line 279: follow up on the previous comment: this is also an array of encoded indexes, maybe we could use the same variable name here (also I would at least use `indexes` here or `idxs` in `calculateRootFromUpdateData`)

ricott1: Furthermore, these idxs can only correspond to leaf nodes. If they are encoded, we can actually still check that they have the correct length. Thinking about it, I think this is important, because the real update is given by the `idxs` array, if they can correspond to internal nodes, we allow for updates of internal nodes.

elements/lisk-tree/src/merkle_tree/merkle_tree.ts
```diff
@@ -267,13 +276,100 @@ export class MerkleTree {
...
+				sortedIdxs = sortedIdxs.slice(2);
+				const parentIndex = currentIndex >> 1;
+				insertNewIndex(sortedIdxs, parentIndex);
+				continue;
+			}
+			const currentNodeLoc = getLocation(currentIndex, height);
+			if (currentNodeLoc.layerIndex === height) {
+				return siblingHashes;
+			}
+			const siblingNode = getRightSiblingInfo(
```
ricott1, line 342: Should we consider renaming the function `getRightSiblingInfo`?

ricott1 (changes requested):
I left some comments, mostly about variable naming but also an important one about the update function

elements/lisk-tree/src/merkle_tree/calculate.ts
shuse2: Ill add a comment. For variable, we use
- index == encoded index
- nodeIndex, layerIndex => decoded index

elements/lisk-tree/src/merkle_tree/merkle_tree.ts
shuse2: any suggestion? I think this gets the sibling information except from left up right? Ill add the comment as well

elements/lisk-tree/src/merkle_tree/merkle_tree.ts
shuse2: Added layer check

ricott1 (approved):
LGTM, good job!

shuse2 (approved):

PR #6591, opened 2021-07-19
Create new state store - Closes #6564

### LINKED ISSUE #6564: Create new state store
### Description

Add new state store class in chain library while keeping the old one with below interface.
```
interface StateStore {
  getStore: (moduleID: number, storePrefix: number) => SubStore;
  createSnapshot(): void;
  restoreSnapshot(): void;
}
```
where store is
```
interface SubStore {
  get(key: Buffer): Promise<Buffer>;
  getWithSchema<T>(key: Buffer, schema: Schema): Promise<T>;
  has(key: Buffer): Promise<boolean>;
  del(key: Buffer): boolean;
  set(key: Buffer, value: Buffer): void;
  setWithSchema<T>(key: Buffer, value: T, schema: Schema): void;
  iterate(input: { start: Buffer, end: Buffer, limit?: number, reverse?: boolean}): Promise<Buffer[]>;
  iterateWithSchema<T>(input: { start: Buffer, end: Buffer, limit?: number, reverse?: boolean}): Promise<T[]>;
}
```

### Acceptance Criteria
- when `getStore` is called internal data structure should not be copied
- `get` and `iterate` should return copy of the buffer
- Diff tracking should be kept
- Add unit test to cover all the cases

### Additional Information
Related issue #6554 

- state store should expect the db instance in the constructor as an interface
```
{
  get: (key: Buffer): Promise<void>;
}
// substore example
this._prefix = Buffer.concat([storagePrefix, moduleID, storePrefix]);
get(key) {
await this._db.get(Buffer.concat([this._prefix, key]);
}
```

shuse2:
### What was the problem?

This PR resolves #6564 

### How was it solved?

- Add new state store with the new interface
- Move current state store 
- Add in-memory db implementation in lisk-db for testing

### How was it tested?

- Add unit tests


mitsuaki-u (changes requested):

mitsuaki-u (approved):

elements/lisk-chain/src/state_store/state_store.ts
```diff
@@ -12,78 +12,231 @@
...
+		if (this._cache.existAny(prefixedKey)) {
+			this._cache.set(prefixedKey, value);
+			return;
+		}
+		// 3. it does not exist in cache, but it does exist in DB => cache first and update
+		const dataExist = await this._ensureCache(prefixedKey);
+		if (dataExist) {
+			this._cache.set(prefixedKey, value);
+			return;
+		}
```
Incede, line 108: I think 2 separate cache calls `cache.cache`,`cache.set` can be avoided.
```suggestion
		let dataExist: bool;
		try {
			const value = await this._db.get(prefixedKey);
			dataExist = true;
		} catch (error) {
			if (error instanceof DBNotFoundError) {
				dataExist = false;
			}
		}
		if (dataExist) {
			this._cache.cache(prefixedKey, value);
			return;
		}
```

elements/lisk-chain/src/state_store/state_store.ts
```diff
@@ -12,78 +12,231 @@
...
+import { StateDiff } from '../types';
+import { CacheDB } from './cache_db';
+import { NotFoundError } from './errors';
+import { DatabaseReader, DatabaseWriter } from './types';
+import { copyBuffer } from './utils';
+
+// TODO: Update when other db keys are defined
+export const DB_PREFIX_STATE_STORE = Buffer.from([5]);
+
+export interface IterationOptions {
```
Incede, line 26: I think this sounds better
```suggestion
export interface IterateOptions {
```

Incede (changes requested):

elements/lisk-chain/src/state_store/state_store.ts
shuse2: explicit `ensureCache` is important because if there is data in the database, we want to `cache` and then set the value.
`cache` will `cache` the value in `cacheDB` and also keep the `init` value from the database for the `diff` calculation.
so basically it's `cache`ing the original value in the database and then setting the new value from the input

elements/lisk-chain/src/state_store/state_store.ts
Incede, line 26: sounds better
```suggestion
export interface IterateOptions {
```

Incede (approved):

PR #6595, opened 2021-07-21
Fix maxHeightPrevoted calculation - Closes #6594

### LINKED ISSUE #6594: Invalid prevote calculation
### Expected behavior
`maxHeightPrevoted` should always equal or greater than the current last block `maxHeightPrevoted` as described in [LIP-0014](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0014.md#computing-prevotes-and-precommits) with `computeChainMaxHeightPrevoted`.

### Actual behavior
When there are not enough `prevote` for more than 5 rounds (we keep the prevote record for last 5 rounds), the `maxHeightPrevoted`will be initialized to the `finalizedHeight`.
This is possible to make the `maxHeightPrevoted` lower than the current `maxHeightPrevoted` and also possible to introduce non-deterministic behavior because `finalizedHeight` might be different on each nodes.

### Steps to reproduce
Have a chain which has not enough prevote for more than 5 rounds, and try to forge a block.

5.1.3

shuse2:
### What was the problem?

This PR resolves #6594

### How was it solved?

- Update to use maxHeightPrevoted from last block in case there is no prevote above threshold in the state
- In case input maxHeightPrevoted does not exist, genesis height is used

### How was it tested?

- Sync with the current testnet and check if it can be synced and generate new block
- Added unit test

### DIFF
--- a/elements/lisk-bft/src/finality_manager.ts
+++ b/elements/lisk-bft/src/finality_manager.ts
@@ -122,13 +122,16 @@ export class FinalityManager extends EventEmitter {
 
 	public finalizedHeight: number;
 	private readonly _chain: Chain;
+	private readonly _genesisHeight: number;
 
 	public constructor({
 		chain,
+		genesisHeight,
 		finalizedHeight,
 		threshold,
 	}: {
 		readonly chain: Chain;
+		readonly genesisHeight: number;
 		readonly finalizedHeight: number;
 		readonly threshold: number;
 	}) {
@@ -136,6 +139,7 @@ export class FinalityManager extends EventEmitter {
 		assert(threshold > 0, 'Must provide a positive threshold');
 
 		this._chain = chain;
+		this._genesisHeight = genesisHeight;
 
 		// Threshold to consider a block pre-voted
 		this.preVoteThreshold = threshold;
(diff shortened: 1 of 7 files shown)


elements/lisk-bft/src/finality_manager.ts
```diff
@@ -339,7 +343,12 @@ export class FinalityManager extends EventEmitter {
 
 		const bftBlockHeaders = stateStore.chain.lastBlockHeaders;
 		const { ledger } = await this._getVotingLedger(stateStore);
-		const chainMaxHeightPrevoted = this._calculateMaxHeightPrevoted(ledger);
+		// lastBlockHeaders are sorted by height desc
+		const lastMaxHeightPrevoted =
+			bftBlockHeaders.length > 0 && bftBlockHeaders[0].asset.maxHeightPrevoted
+				? bftBlockHeaders[0].asset.maxHeightPrevoted
```
janhack, line 349: `BlockHeaders[0]` is the block header at the current tip of the chain?

shuse2: yes! `bftBlockHeaders` are sorted by height desc, so the first element will be the last block

elements/lisk-bft/test/unit/finality_manager.spec.ts
```diff
@@ -94,20 +95,39 @@ describe('finality_manager', () => {
...
+							height: 12,
+							prevotes: 66,
+							precommits: 0,
+						},
+					],
+				};
+				jest
+					.spyOn(chainStub.dataAccess, 'getConsensusState')
+					.mockResolvedValue(codec.encode(BFTVotingLedgerSchema, validatorLedger));
+				const lastMaxHeightPrevoted = 30;
```
janhack, line 128: I don't complete understand these specs, as the ledger state only contains 3 blocks and the `lastMaxHeightPrevoted` is larger than the height of the current tip of the chain.

shuse2: main case i wanted to create here is that state does not have any prevote higher than threshold, and maxHeightPrevoted not in the list.
but i guess `lastMaxHeightPrevoted` should be lower than the height in the state to be realistic

janhack (approved):
Looks good.

elements/lisk-bft/test/unit/finality_manager.spec.ts
shuse2: Updated test case height to be more realistic

nazarhussain (approved):

PR #6596, opened 2021-07-21
Update forger plugin to consider genesis height - #6582

### LINKED ISSUE #6582: Forger plugin doesn't collect any data running on the testnet
### Expected behavior

Forger plugin should work and collect data

### Actual behavior

Forger plugin loads and responds to calls but doesn't collect any data. Additionally there's an error thrown in the lisk-core logs:

```
(node:1242) UnhandledPromiseRejectionWarning: Error: Specified key blocks:height:00d6c57b does not exist
    at parseError (/home/lisk/.nvm/versions/node/v12.22.1/lib/node_modules/lisk-core/node_modules/lisk-framework/dist-node/controller/bus.js:20:12)
    at Bus.invoke (/home/lisk/.nvm/versions/node/v12.22.1/lib/node_modules/lisk-core/node_modules/lisk-framework/dist-node/controller/bus.js:125:23)
(node:1242) UnhandledPromiseRejectionWarning: Unhandled promise rejection. This error originated either by throwing inside of an async function without a catch block, or by rejecting a promise which was not handled with .catch(). To terminate the node process on unhandled promise rejection, use the CLI flag `--unhandled-rejections=strict` (see https://nodejs.org/api/cli.html#cli_unhandled_rejections_mode). (rejection id: 1)
```

### Steps to reproduce

Run the lisk-core with the forger plugin loaded on the **testnet**, check if data is collected.

lisk-core 3.0.0-rc.2

shuse2:
### What was the problem?

This PR resolves #6582 

### How was it solved?

- Update lisk-chain to maintain genesis height
- Update nodeInfo to include `genesisHeight` as response
- Update forger plugin to consider genesis height as the initial point before syncing

### How was it tested?

- Run forger plugin with testnet data

### DIFF
--- a/framework-plugins/lisk-framework-forger-plugin/src/forger_plugin.ts
+++ b/framework-plugins/lisk-framework-forger-plugin/src/forger_plugin.ts
@@ -60,6 +60,7 @@ interface ForgerPayloadInfo {
 }
 
 interface NodeInfo {
+	genesisHeight: number;
 	genesisConfig: GenesisConfig;
 }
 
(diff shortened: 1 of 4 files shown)


framework-plugins/lisk-framework-forger-plugin/src/forger_plugin.ts
```diff
@@ -187,20 +188,22 @@ export class ForgerPlugin extends BasePlugin {
 			header: { height: lastBlockHeight },
 		} = this.codec.decodeBlock(await this._channel.invoke<string>('app:getLastBlock'));
 		const { syncUptoHeight } = await getForgerSyncInfo(this._forgerPluginDB);
+		const { genesisHeight } = await this._channel.invoke<NodeInfo>('app:getNodeInfo');
+		const forgerPluginSyncedHeight = syncUptoHeight === 0 ? genesisHeight : syncUptoHeight;
```
ManuGowda, line 192: in case of migration the height of regenesis is not 0?

shuse2: `syncUpToHeight === 0` means there is no data in this plugin or actually synced height `0`. therefore, replacing with genesis height should be ok

ManuGowda (approved):

nazarhussain (approved):

PR #6599, opened 2021-07-26
Create event queue - Closes #6563

### LINKED ISSUE #6563: Create event queue
### Description

Add event  queue in state machine domain with below interfaces.

```
interface EventQueue {
  add: (key: string, value: Buffer) => void;
  createSnapshot() => void;
  restoreSnapshot() => void;
  getEvents(): Event[];
}

// Framework or module add this event
interface Event {
  key: string;
  value: Buffer; // encoded as hex string when publishing
}
```

### Acceptance Criteria

- All public functions should be unit tested

### Additional Information
Related  Issue #6554 

- behavior of createSnapshot should be the same as state store when createSnapshot is called, it should copy internal events and restore when `restoreSnapshot` is called.

ManuGowda:
### What was the problem?

This PR resolves #6563

### How was it solved?

- Implemented event queue 

### How was it tested?

- Added unit tests


framework/test/unit/node/state_machine/event_queue.spec.ts
```diff
@@ -0,0 +1,87 @@
...
+
+		expect(eventQueue['_originalEvents']).toBeEmpty();
+		addedEvents.forEach((e, i) => {
+			expect(e.key).toEqual(events[i].key);
+			expect(e.value.equals(events[i].value)).toBeTrue();
+		});
+		return expect(addedEvents.length === events.length);
+	});
+
+	// eslint-disable-next-line @typescript-eslint/require-await
```
shuse2, line 76: if it's not await, let's not make it await

shuse2 (changes requested):

shuse2 (approved):

ishantiw (approved):

PR #6600, opened 2021-07-26
Update block to be a class - Closes #6581

### LINKED ISSUE #6581: Update block to be a class
### Description

Update block and block header to be a class with below interface.
This is required to support `getAsset` and `setAsset`

```
class Block {
  public static fromBytes(value: Buffer): Block;
  public getBytes(): Buffer;
  public get header() BlockHeader;
  public get payload() []Transaction;
}

class BlockHeader {
  public static fromBytes(value: Buffer): BlockHeader;
  public getBytes(): Buffer;
  public toJSON(): BlockHeaderJSON;
  public validate(): void;
  public getSigningBytes(): Buffer;
  public sign(key: Buffer);
  public getAsset(moduleID: number): Buffer;
  public setAsset(moduleID: number, value: Buffer): void;
  public get generatorAddress(): Buffer;
  public get id(): Buffer;
  public get version(): number;
  public get previousBlockID(): Buffer;
  public get timestamp(): Buffer;
  public get stateRoot(): Buffer;
  public get transactionRoot(): Buffer;
  public get signature(): Buffer;
}
```

### Acceptance Criteria
- Add unit test for the added classes

nazarhussain:
### What was the problem?

This PR resolves #6581

### How was it solved?

- Created two classes `Block` and `BlockHeader` 

### How was it tested?

- Added unit tests

### DIFF
--- a/elements/lisk-chain/src/block.ts
+++ b/elements/lisk-chain/src/block.ts
@@ -17,6 +17,11 @@ import { BlockHeader } from './block_header';
 import { blockSchema } from './schema';
 import { Transaction } from './transaction';
 
+interface BlockAttrs {
+	header: Buffer;
+	payload: Buffer[];
+}
+
 export class Block {
 	// eslint-disable-next-line no-useless-constructor
 	public constructor(public readonly header: BlockHeader, public readonly payload: Transaction[]) {
@@ -24,10 +29,7 @@ export class Block {
 	}
 
 	public static fromBytes(value: Buffer): Block {
-		const { header, payload } = codec.decode<{ header: Buffer; payload: Buffer[] }>(
-			blockSchema,
-			value,
-		);
+		const { header, payload } = codec.decode<BlockAttrs>(blockSchema, value);
 
 		return new Block(
 			BlockHeader.fromBytes(header),
--- a/elements/lisk-chain/src/block.ts
+++ b/elements/lisk-chain/src/block.ts
@@ -1,0 +1,44 @@
+/*
+ * Copyright © 2021 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+import { codec } from '@liskhq/lisk-codec';
+import { BlockHeader } from './block_header';
+import { blockSchema } from './schema';
+import { Transaction } from './transaction';
+
+export class Block {
+	// eslint-disable-next-line no-useless-constructor
+	public constructor(public readonly header: BlockHeader, public readonly payload: Transaction[]) {
+		// No body necessary
+	}
+
+	public static fromBytes(value: Buffer): Block {
+		const { header, payload } = codec.decode<{ header: Buffer; payload: Buffer[] }>(
+			blockSchema,
+			value,
+		);
+
+		return new Block(
+			BlockHeader.fromBytes(header),
+			payload.map(v => Transaction.decode(v)),
+		);
+	}
+
+	public getBytes(): Buffer {
+		return codec.encode(blockSchema, {
+			header: this.header.getBytes(),
+			payload: this.payload.map(p => p.getBytes()),
+		});
+	}
+}
--- a/elements/lisk-chain/src/block_header.ts
+++ b/elements/lisk-chain/src/block_header.ts
@@ -12,53 +12,106 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
-import { signDataWithPrivateKey } from '@liskhq/lisk-cryptography';
+import { objects } from '@liskhq/lisk-utils';
+import { signDataWithPrivateKey, hash } from '@liskhq/lisk-cryptography';
 import { codec } from '@liskhq/lisk-codec';
 import { validator, LiskValidationError } from '@liskhq/lisk-validator';
-import { blockHeaderSchema } from './schema';
 import { TAG_BLOCK_HEADER } from './constants';
 
+export const signingBlockHeaderSchema = {
+	$id: '/block/header/signing/2',
+	type: 'object',
+	properties: {
+		version: { dataType: 'uint32', fieldNumber: 1 },
+		timestamp: { dataType: 'uint32', fieldNumber: 2 },
+		height: { dataType: 'uint32', fieldNumber: 3 },
+		previousBlockID: { dataType: 'bytes', fieldNumber: 4 },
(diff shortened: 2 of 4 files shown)


elements/lisk-chain/src/block_header.ts
```diff
@@ -0,0 +1,164 @@
...
+	public sign(key: Buffer, networkIdentifier: Buffer) {
+		return signDataWithPrivateKey(TAG_BLOCK_HEADER, networkIdentifier, this.getSigningBytes(), key);
+	}
+
+	public getAsset(moduleID: number): Buffer | undefined {
+		return this._assets.find(a => a.moduleID === moduleID)?.data;
+	}
+	public setAsset(moduleID: number, value: Buffer): void {
+		const asset = this.getAsset(moduleID);
+		if (asset) {
```
shuse2, line 126: should we overwrite?

elements/lisk-chain/src/block_header.ts
```diff
@@ -0,0 +1,164 @@
...
+		this.stateRoot = stateRoot;
+		this.transactionRoot = transactionRoot;
+		this._assets = [...assets];
+
+		this._signature = signature;
+		this._id = id;
+	}
+
+	public static fromBytes(value: Buffer): BlockHeader {
+		return new BlockHeader(codec.decode<BlockHeaderAttrs>(blockHeaderSchema, value));
```
shuse2, line 81: probably `id` should be calculated here as well before assigning

elements/lisk-chain/src/block_header.ts
```diff
@@ -0,0 +1,164 @@
...
+	}
+
+	public toJSON(): BlockHeaderJSON {
+		return codec.toJSON(blockHeaderSchema, this._getBlockHeaderProps());
+	}
+
+	public validate(): void {
+		const headerWithoutAsset = {
+			...this._getBlockHeaderProps(),
+			asset: Buffer.alloc(0),
```
shuse2, line 95: we don't have asset =)

elements/lisk-chain/src/block_header.ts
```diff
@@ -0,0 +1,164 @@
...
+		};
+		// Validate block header
+		const errors = validator.validate(blockHeaderSchema, headerWithoutAsset);
```
shuse2, line 98: no need to specifically change it to without asset. we can have the schema `data` as bytes

elements/lisk-chain/src/block_header.ts
nazarhussain: Approach 1: Different modules can set asset on different logic, so we can overwrite asset. 
Approach 2: Let user handle logic on their side and set the asset once for a module. 

I believe Approach 2 is safe.

shuse2: I was thinking a case where a module want to set in `initBlock` and update in `sealBlock`, but i guess it's not necessary for now

elements/lisk-chain/src/block_header.ts
```diff
@@ -0,0 +1,233 @@
...
+		}
+	}
+
+	public getSigningBytes(): Buffer {
+		const blockHeaderBytes = codec.encode(signingBlockHeaderSchema, this._getSigningProps());
+
+		return blockHeaderBytes;
+	}
+
+	public sign(key: Buffer, networkIdentifier: Buffer) {
```
shuse2, line 164:
```suggestion
	public sign(networkIdentifier: Buffer, privateKey: Buffer) {
```

shuse2 (changes requested):
Overall LGTM, just minor things

shuse2 (approved):

elements/lisk-chain/src/block.ts
```diff
@@ -0,0 +1,46 @@
...
+import { BlockHeader } from './block_header';
+import { blockSchema } from './schema';
+import { Transaction } from './transaction';
+
+interface BlockAttrs {
+	header: Buffer;
+	payload: Buffer[];
+}
+
+export class Block {
```
ishantiw, line 25: The issue says it should have 2 more methods,

```js
public get header() BlockHeader;
public get payload() []Transaction;
```

elements/lisk-chain/test/unit/block_header.spec.ts
```diff
@@ -0,0 +1,181 @@
+/*
+ * Copyright © 2019 Lisk Foundation
```
ishantiw, line 2: if its a new file then update license
```suggestion
 * Copyright © 2021 Lisk Foundation
```

elements/lisk-chain/test/unit/block_header.spec.ts
```diff
@@ -0,0 +1,181 @@
...
+	'previousBlockID',
+	'stateRoot',
+	'transactionRoot',
+	'generatorAddress',
+	'assets',
+];
+
+describe('block_header', () => {
+	describe('signingBlockHeaderSchema', () => {
+		it('should be valid schema without signature or id', () => {
```
ishantiw, line 60:
```suggestion
		it('should be valid schema without signature and id', () => {
```

elements/lisk-chain/src/block_header.ts
```diff
@@ -0,0 +1,286 @@
...
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+import { objects } from '@liskhq/lisk-utils';
+import { signDataWithPrivateKey, hash } from '@liskhq/lisk-cryptography';
+import { codec } from '@liskhq/lisk-codec';
+import { validator, LiskValidationError } from '@liskhq/lisk-validator';
+import { TAG_BLOCK_HEADER } from './constants';
+
+export const signingBlockHeaderSchema = {
```
ishantiw, line 21: should we move these schemas to `schema.ts`?

elements/lisk-chain/src/block_header.ts
```diff
@@ -0,0 +1,286 @@
...
+	readonly previousBlockID: Buffer;
+	readonly timestamp?: number;
+	readonly stateRoot?: Buffer;
+	readonly transactionRoot?: Buffer;
+	readonly assets: ReadonlyArray<BlockHeaderAsset>;
+	signature?: Buffer;
+	id?: Buffer;
+}
+
+export interface BlockHeaderJSON {
```
ishantiw, line 95: if we are exporting then maybe move it to `types.ts`

elements/lisk-chain/src/block.ts
nazarhussain: Not required to to have getters as the `header` and `payload` both are public readonly properties.

elements/lisk-chain/src/block_header.ts
nazarhussain: Yes we should, when we integrate new block header. If done now, will break a lot of things in the framework and other places.

elements/lisk-chain/src/block_header.ts
nazarhussain: Same as above.

ishantiw (approved):

PR #6605, opened 2021-07-28
Implement generator component - Closes #6561

### LINKED ISSUE #6561: Create generator component using forger code
### Description

Add  generator component with interface  below.

```
interface Generator {
	registerModule(mod: Module): void;
	init(params: Params): Promise<void>;
	start(): Promise<void>;
	stop(): Promise<void>;
}
```
where it should expect module  with interface 
```
interface Module {
  initBlock(ctx: BlockGenerateContext): Promise<void>;
  sealBlock(ctx: BlockGenerateContext): Promise<void>;
}

interface BlockGenerateContext {
  logger: Logger;
  networkIdentifier: Buffer;
  getAPIContext(): APIContext;
  getStore: (moduleID: number, storePrefix: number) => ImmutableModuleStore;
  header: {
    version: number;
    height: number;
    previousBlockID: Buffer;
    generatorAddress: Buffer;
    assets: Buffer[];
  },
  getGeneratorStore: (moduleID: number) => GeneratorStore,
  setAsset: (moduleID: number, data: Buffer) => void;
  getAsset: (moduleID: number) => Buffer[];
}
```
Constructor should resolve the dependency such as
```
chain: Chain;
stateMachine: StateMachine;
consensus: Consensus;
p2p: P2P;
```

and initialization should expect
```
interface Params {
  blockchainDB: DB;
  generatorDB: DB;
  logger: Loggere;
  generatorConnfig: GeneratorConfig;
}
```

- transaction pool instance should be created and managed in this component
- transaction broadcasting and handling are managed in this folder

### Acceptance Criteria

- Current forging logic should be kept as it is. (hash onion etc) we will update with the  new module
- interface follows the above description
- Add unit test

### Additional Information
Related Issues: #6554

shuse2:
### What was the problem?

This PR resolves #6561

### How was it solved?

- Create generator component using forger/transport components
- Move forger logics into generator component
- Update block generation logic to use new module hooks
- Split transport into network_endpoint and endpoint

### How was it tested?

- Add unit tests

### DIFF
--- a/elements/lisk-chain/src/block_header.ts
+++ b/elements/lisk-chain/src/block_header.ts
@@ -84,7 +84,7 @@ interface BlockHeaderAttrs {
 	readonly height: number;
 	readonly generatorAddress: Buffer;
 	readonly previousBlockID: Buffer;
-	readonly timestamp?: number;
+	readonly timestamp: number;
 	readonly stateRoot?: Buffer;
 	readonly transactionRoot?: Buffer;
 	readonly assets: ReadonlyArray<BlockHeaderAsset>;
@@ -110,7 +110,7 @@ export class BlockHeader {
 	public readonly height: number;
 	public readonly generatorAddress: Buffer;
 	public readonly previousBlockID: Buffer;
-	private _timestamp?: number;
+	public readonly timestamp: number;
 	private _stateRoot?: Buffer;
 	private _transactionRoot?: Buffer;
 	private readonly _assets: BlockHeaderAsset[];
@@ -133,7 +133,7 @@ export class BlockHeader {
 		this.height = height;
 		this.generatorAddress = generatorAddress;
 		this.previousBlockID = previousBlockID;
-		this._timestamp = timestamp;
+		this.timestamp = timestamp;
 		this._stateRoot = stateRoot;
 		this._transactionRoot = transactionRoot;
 		this._assets = objects.cloneDeep<BlockHeaderAsset[]>([...assets]);
@@ -144,15 +144,6 @@ export class BlockHeader {
 
 	public static fromBytes(value: Buffer): BlockHeader {
 		return new BlockHeader(codec.decode<BlockHeaderAttrs>(blockHeaderSchema, value));
-	}
-
-	public get timestamp() {
-		return this._timestamp;
-	}
-
-	public set timestamp(val) {
-		this._timestamp = val;
-		this._resetComputedValues();
 	}
 
 	public get stateRoot() {
@@ -202,15 +193,13 @@ export class BlockHeader {
 		return blockHeaderBytes;
 	}
 
-	public sign(networkIdentifier: Buffer, privateKey: Buffer) {
+	public sign(networkIdentifier: Buffer, privateKey: Buffer): void {
 		this._signature = signDataWithPrivateKey(
 			TAG_BLOCK_HEADER,
 			networkIdentifier,
 			this.getSigningBytes(),
 			privateKey,
 		);
-
-		return this._signature;
 	}
 
 	public getAsset(moduleID: number): Buffer | undefined {
--- a/framework/src/node/consensus/consensus.ts
+++ b/framework/src/node/consensus/consensus.ts
@@ -150,6 +150,11 @@ export class Consensus {
 		this._network.registerEndpoint(NETWORK_RPC_GET_HIGHEST_COMMON_BLOCK, async ({ data, peerId }) =>
 			this._endpoint.handleRPCGetHighestCommonBlock(data, peerId),
 		);
+		this._network.registerHandler(NETWORK_EVENT_POST_BLOCK, ({ data, peerId }) => {
+			this.onBlockReceive(data, peerId).catch(err => {
+				this._logger.error({ err: err as Error, peerId }, 'Fail to handle received block');
+			});
+		});
 
 		this._logger.debug(
 			{
@@ -389,6 +394,13 @@ export class Consensus {
 			);
 			await this._validate(block);
 			await this._executeValidated(block);
+
+			this._network.applyNodeInfo({
+				height: block.header.height,
+				lastBlockID: block.header.id,
+				maxHeightPrevoted: block.header.asset.maxHeightPrevoted,
+				blockVersion: block.header.version,
+			});
 		});
 	}
 
--- a/framework/src/node/consensus/consensus.ts
+++ b/framework/src/node/consensus/consensus.ts
@@ -38,6 +38,7 @@ import {
 	EVENT_BLOCK_BROADCAST,
 	EVENT_FORK_DETECTED,
 	NETWORK_EVENT_POST_BLOCK,
+	NETWORK_EVENT_POST_NODE_INFO,
 	NETWORK_RPC_GET_BLOCKS_FROM_ID,
 	NETWORK_RPC_GET_HIGHEST_COMMON_BLOCK,
 	NETWORK_RPC_GET_LAST_BLOCK,
@@ -154,6 +155,9 @@ export class Consensus {
 			this.onBlockReceive(data, peerId).catch(err => {
 				this._logger.error({ err: err as Error, peerId }, 'Fail to handle received block');
 			});
+		});
+		this._network.registerHandler(NETWORK_EVENT_POST_NODE_INFO, ({ data, peerId }) => {
+			this._logger.debug({ peerId, data }, 'Received new node info');
 		});
 
 		this._logger.debug(
--- a/framework/src/node/generator/broadcaster.ts
+++ b/framework/src/node/generator/broadcaster.ts
@@ -1,0 +1,93 @@
+/*
+ * Copyright © 2019 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+import { codec } from '@liskhq/lisk-codec';
+import { TransactionPool } from '@liskhq/lisk-transaction-pool';
+import { EVENT_POST_TRANSACTION_ANNOUNCEMENT } from '../../constants';
+
+import { Logger } from '../../logger';
+import { Network } from '../network';
+import { postTransactionsAnnouncementSchema } from './schemas';
+
+interface BroadcasterConfig {
+	readonly releaseLimit: number;
+	readonly interval: number;
+}
+
+export interface BroadcasterConstructor extends BroadcasterConfig {
+	readonly transactionPool: TransactionPool;
+	readonly network: Network;
+}
+
+interface BroadcasterInitArgs {
+	readonly logger: Logger;
+}
+
+export class Broadcaster {
+	private readonly _transactionPool: TransactionPool;
+	private readonly _network: Network;
+	private readonly _config: BroadcasterConfig;
+	private _transactionIdQueue: Buffer[];
+
+	private _logger!: Logger;
+
+	public constructor({ transactionPool, releaseLimit, interval, network }: BroadcasterConstructor) {
+		this._transactionPool = transactionPool;
+		this._network = network;
+		this._config = {
+			releaseLimit,
+			interval,
+		};
+		this._transactionIdQueue = [];
+	}
+
+	public init(args: BroadcasterInitArgs): void {
+		this._logger = args.logger;
+		setInterval(() => {
+			try {
+				this._broadcast();
+			} catch (err) {
+				// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+				this._logger.error({ err }, 'Failed to broadcast information');
+			}
+		}, this._config.interval);
+	}
+
+	public enqueueTransactionId(transactionId: Buffer): boolean {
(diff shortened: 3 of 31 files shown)


framework/src/node/generator/endpoint.ts
```diff
@@ -0,0 +1,229 @@
...
+		this._broadcaster = args.broadcaster;
+	}
+
+	public init(args: EndpointInit) {
+		this._logger = args.logger;
+		this._generatorDB = args.generatorDB;
+		this._blockchainDB = args.blockchainDB;
+	}
+
+	public async postTransaction(ctx: EndpointContext): Promise<PostTransactionResponse> {
```
shuse2, line 99: This function is copied from transport, and input is updated to `EndpointContext`

framework/src/node/generator/endpoint.ts
```diff
@@ -0,0 +1,229 @@
...
+				senderPublicKey: transaction.senderPublicKey,
+			},
+			'Added transaction to pool',
+		);
+		return {
+			transactionId: transaction.id.toString('hex'),
+		};
+	}
+
+	public async updateForgingStatus(ctx: EndpointContext): Promise<UpdateForgingStatusResponse> {
```
shuse2, line 153: This function is copied from `forger.ts` and updated to use `EndpointContext` as input.
Also, BFT properties checks are handled in `_consensus.verifyGeneratorInfo` and `_consensus.isSynced` functions instead

framework/src/node/generator/network_endpoint.ts
```diff
@@ -0,0 +1,293 @@
...
+		this._stateMachine = args.stateMachine;
+		this._broadcaster = args.broadcaster;
+	}
+
+	public init(args: NetworkEndpointInitArgs): void {
+		this._logger = args.logger;
+		this._blockchainDB = args.blockchainDB;
+	}
+
+	public async handleRPCGetTransactions(data: unknown, peerId: string): Promise<Buffer> {
```
shuse2, line 78: this function is copied from `transport`

framework/src/node/generator/network_endpoint.ts
```diff
@@ -0,0 +1,293 @@
...
+		return codec.encode(getTransactionsResponseSchema, {
+			transactions: transactionsFromQueues,
+		});
+	}
+
+	/**
+	 * Process transactions IDs announcement. First validates, filter the known transactions
+	 * and finally ask to the emitter the ones that are unknown.
+	 */
+	public async handleEventPostTransactionsAnnouncement(
```
shuse2, line 160: this function is copied from `transport`

framework/src/node/generator/schemas.ts
```diff
@@ -0,0 +1,168 @@
+/*
```
shuse2, line 1: these definitions are copied from schema in transport folder

framework/src/node/generator/strategies.ts
```diff
@@ -0,0 +1,150 @@
...
+import { TransactionPool, PooledTransaction } from '@liskhq/lisk-transaction-pool';
+import { dataStructures } from '@liskhq/lisk-utils';
+import { Chain, Transaction } from '@liskhq/lisk-chain';
+import { getAddressFromPublicKey } from '@liskhq/lisk-cryptography';
+import { KVStore } from '@liskhq/lisk-db';
+import { StateMachine, VerifyStatus, EventQueue, TransactionContext } from '../state_machine';
+import { Logger } from '../../logger';
+import { WritableBlockHeader } from './types';
+
+export class HighFeeGenerationStrategy {
```
shuse2, line 24: this is copied from `strategies` in `forger` fodler.
Updated to use state machine instead of processor

framework/src/node/network/base_network_endpoint.ts
```diff
@@ -0,0 +1,43 @@
...
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+import { Network } from './network';
```
shuse2, line 14: created to unify rate limit functionality

framework/src/node/generator/generator.ts
```diff
@@ -0,0 +1,468 @@
...
+		if (block.payload.length) {
+			for (const transaction of block.payload) {
+				this._pool.add(transaction).catch(err => {
+					this._logger.error({ err: err as Error }, 'Failed to add transaction back to the pool');
+				});
+			}
+		}
+	}
+
+	public async _loadTransactionsFromNetwork(): Promise<void> {
```
shuse2, line 230: this function is copied from `synchronizer`

framework/src/node/generator/generator.ts
```diff
@@ -0,0 +1,468 @@
...
+			});
+			const result = await this._stateMachine.verifyTransaction(txContext);
+			if (result.status === VerifyStatus.FAIL) {
+				throw new Error('Invalid transaction');
+			}
+		}
+	}
+
+	// eslint-disable-next-line @typescript-eslint/require-await
+	private async _loadGenerators(): Promise<void> {
```
shuse2, line 265: this function is copied from `forger`

framework/src/node/generator/generator.ts
```diff
@@ -0,0 +1,468 @@
...
+				height: generatedBlock.header.height,
+				generatorAddress: generator.toString('hex'),
+			},
+			'Generated new block',
+		);
+
+		await this._consensus.execute(generatedBlock as never);
+	}
+
+	private async _generateBlock(
```
shuse2, line 403: this is new logic for block generation to use state machine and module hooks (init and seal)

framework/src/node/generator/generator.ts
```diff
@@ -0,0 +1,468 @@
...
+			const { error } = await this._pool.add(transaction);
+
+			if (error) {
+				this._logger.error({ err: error }, 'Failed to add transaction to pool.');
+				throw error;
+			}
+		}
+	}
+
+	private async _generateLoop(): Promise<void> {
```
shuse2, line 345: this logic is copied from `forger`

framework/src/node/generator/strategies.ts
```diff
@@ -0,0 +1,150 @@
...
+				eventQueue,
+				logger: this._logger,
+				networkIdentifier: this._chain.constants.networkIdentifier,
+				stateStore,
+				transaction: lowestNonceHighestFeeTrx,
+				header,
+			});
+			stateStore.createSnapshot();
+			try {
+				const result = await this._stateMachine.verifyTransaction(txContext);
```
shuse2, line 98: this is the main difference from the `forger/strategies`

elements/lisk-chain/src/block_header.ts
```diff
@@ -84,7 +84,7 @@ interface BlockHeaderAttrs {
 	readonly height: number;
 	readonly generatorAddress: Buffer;
 	readonly previousBlockID: Buffer;
-	readonly timestamp?: number;
+	readonly timestamp: number;
```
nazarhussain, line 87: I kept it optional earlier because someone can start building block header in multiple steps and want to set the time stamp in the end right before signing it.

framework/src/node/network/network.ts
```diff
@@ -364,7 +360,7 @@ export class Network {
 				readonly event: string;
 				readonly data: Buffer | undefined;
 			}) => {
-				if (!REMOTE_EVENTS_WHITE_LIST.includes(packet.event)) {
+				if (!Object.keys(this._eventHandlers).includes(packet.event)) {
```
nazarhussain, line 363: Does that means we don't have any local events any more?

framework/src/node/state_machine/block_context.ts
```diff
@@ -40,7 +40,7 @@ export class BlockContext {
 	private readonly _logger: Logger;
 	private readonly _eventQueue: EventQueue;
 	private readonly _header: BlockHeader;
-	private readonly _transactions?: ReadonlyArray<Transaction>;
+	private _transactions?: ReadonlyArray<Transaction>;
```
nazarhussain, line 43: Why we need to set the transactions after context intialization? I believe context must be always immutable object. If you really want to change the context must create new one with new transactions.

framework/src/node/state_machine/types.ts
```diff
@@ -26,24 +27,14 @@ export interface ImmutableSubStore {
...
-		reverse?: boolean;
-	}): Promise<Buffer[]>;
-	iterateWithSchema<T>(input: {
-		start: Buffer;
-		end: Buffer;
-		limit?: number;
-		reverse?: boolean;
-	}): Promise<T[]>;
+	iterate(input: IterateOptions): Promise<{ key: Buffer; value: Buffer }[]>;
+	iterateWithSchema<T>(input: IterateOptions, schema: Schema): Promise<{ key: Buffer; value: T }[]>;
```
nazarhussain, line 31: These really don't look like iterators. We should rather rename to `toArray` and `toArrayWithSchema`

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Iterators_and_Generators
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols#the_iterable_protocol

framework/src/types.ts
```diff
@@ -160,11 +160,25 @@ interface RPCConfig {
...
+	readonly encryptedPassphrase: string;
+}
+
+export interface GeneratorConfig {
+	waitThreshold: number;
+	generators: Generator[];
+	force?: boolean;
+	defaultPassword?: string;
+	modules: Record<string, unknown>;
+}
```
nazarhussain, line 174: These two look very odd to each other. If `Generator` is an account then `GeneratorConfig` relates to account config, which is not. We can change `GeneratorConfig` to `GenerationConfig` or `Generator` to `GeneratorAccount`. Not sure either.

nazarhussain (changes requested):

elements/lisk-chain/src/block_header.ts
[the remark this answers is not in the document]
shuse2: Timestamp is needed during the block execution, so it is always needed when initializing the block header

framework/src/node/consensus/consensus.ts
```diff
@@ -150,6 +150,11 @@ export class Consensus {
 		this._network.registerEndpoint(NETWORK_RPC_GET_HIGHEST_COMMON_BLOCK, async ({ data, peerId }) =>
 			this._endpoint.handleRPCGetHighestCommonBlock(data, peerId),
 		);
+		this._network.registerHandler(NETWORK_EVENT_POST_BLOCK, ({ data, peerId }) => {
+			this.onBlockReceive(data, peerId).catch(err => {
+				this._logger.error({ err: err as Error, peerId }, 'Fail to handle received block');
+			});
+		});
```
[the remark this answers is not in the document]
shuse2: event handler seems always should be synchronous. (we always have the eslint rule complaining)

framework/src/node/generator/broadcaster.ts
```diff
@@ -0,0 +1,93 @@
...
+import { codec } from '@liskhq/lisk-codec';
+import { TransactionPool } from '@liskhq/lisk-transaction-pool';
+import { EVENT_POST_TRANSACTION_ANNOUNCEMENT } from '../../constants';
+
+import { Logger } from '../../logger';
+import { Network } from '../network';
+import { postTransactionsAnnouncementSchema } from './schemas';
+
+interface BroadcasterConfig {
+	readonly releaseLimit: number;
```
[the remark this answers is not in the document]
shuse2: it was called `releaseLimit`. keep in mind this code is not new.

framework/src/node/generator/types.ts
```diff
@@ -0,0 +1,78 @@
...
+
+export interface WritableBlockHeader {
+	version: number;
+	height: number;
+	timestamp: number;
+	previousBlockID: Buffer;
+	generatorAddress: Buffer;
+	getAsset: (moduleID: number) => Buffer | undefined;
+	setAsset: (moduleID: number, value: Buffer) => void;
+}
```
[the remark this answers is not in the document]
shuse2: we have certain property which cannot be used by the block generation. ex) state root etc. That's why it's passing interface with limited capability

framework/src/node/state_machine/block_context.ts
shuse2: `transactions`  can only be used in the `afterBlockExec` or `sealBlock` step now. That's why it's assigned after they been executed.

framework/src/types.ts
shuse2: `GenerationConfig` sounds good! will update

shuse2:
> Why generator component need to handle RPC requests? Is it replacing transaction pool?
Transaction pool is managed under generator. that's why it will request to load, handle announcement and provide endpoints

ManuGowda (approved):

framework/src/node/generator/broadcaster.ts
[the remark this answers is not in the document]
nazarhussain: Ok. But as now we are changing it all, why not rename it now?

framework/src/node/generator/types.ts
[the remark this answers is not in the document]
nazarhussain: Can't we use `Omit` or `Pick` utility type so we still got a reference that from where this type is from.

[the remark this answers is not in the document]
shuse2: I would still prefer to keep it as interface, and as long as  this is satisfied, it should transpile as typescript

nazarhussain (approved):
Apart from some small open comments LGTM.

PR #6607, opened 2021-07-29
Update application config to be the new format - Closes #6569

### LINKED ISSUE #6569: Update application config to be the new format
### Description

Update application config to be more module centric as below interface.
```
applicationConfigSchema = {
  ...
  genesis: {
    communityIdentifier: ‘’,
    modules: [moduleName]: Record<string, unknown>,
  },
  generator: {
    password: "",
    modules: [moduleName]: Record<string, unknown>
  },
}
```

### Acceptance Criteria
- Configs schema and fixtures are updated
- All unit tests are fixed

### Additional Information
Related issues #6554

nazarhussain:
### What was the problem?

This PR resolves #6569

### How was it solved?

- Configs schema and fixtures are updated
- All unit tests are fixed

### How was it tested?

- Run all unit tests

### DIFF
--- a/framework/src/schema/application_config_schema.ts
+++ b/framework/src/schema/application_config_schema.ts
@@ -377,12 +377,81 @@ export const applicationConfigSchema = {
 		},
 		genesis: {
 			type: 'object',
-			required: ['communityIdentifier', 'modules'],
+			required: ['blockTime', 'communityIdentifier', 'maxPayloadLength', 'rewards', 'modules'],
 			properties: {
+				blockTime: {
+					type: 'number',
+					minimum: 2,
+					description: 'Slot time interval in seconds',
+				},
 				communityIdentifier: {
 					type: 'string',
 					description:
 						'The unique name of the relevant community as a string encoded in UTF-8 format',
+				},
+				bftThreshold: {
+					type: 'integer',
+					minimum: 1,
+					description: 'Number of validators required to set block finality',
+				},
+				minFeePerByte: {
+					type: 'integer',
+					minimum: 0,
+					description: 'Minimum fee per bytes required for a transaction to be valid',
+				},
+				baseFees: {
+					type: 'array',
+					description: 'Base fee for a transaction to be valid',
+					items: {
+						type: 'object',
+						properties: {
+							moduleID: {
+								type: 'number',
+								minimum: 2,
+							},
+							assetID: {
+								type: 'integer',
+								minimum: 0,
+							},
+							baseFee: {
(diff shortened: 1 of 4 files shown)


shuse2 (approved):

framework/src/schema/application_config_schema.ts
```diff
@@ -12,10 +12,94 @@
...
+	},
+	additionalProperties: { type: 'object' },
+};
+
+const rewardsConfigSchema = {
+	type: 'object',
+	required: ['milestones', 'offset', 'distance'],
+	description: 'Object representing LSK rewards milestone',
+	properties: {
+		milestones: {
```
ManuGowda, line 29: the naming has changed, maybe reflect the change? https://github.com/LiskHQ/lips-staging/pull/94/files#diff-dfeda675218ab1dec2be752051c6b2a96bd201a0d704ebbb438c31b32312fb25R63
@IkerAlus its `brackets` now?

framework/src/schema/application_config_schema.ts
```diff
@@ -12,10 +12,94 @@
...
+		properties: {
+			encryptedPassphrase: {
+				type: 'string',
+				format: 'encryptedPassphrase',
+			},
+			address: {
+				type: 'string',
+				format: 'hex',
+			},
+			hashOnion: {
```
ManuGowda, line 64: doesn't this go to random module specific config?

framework/src/schema/application_config_schema.ts
```diff
@@ -357,6 +382,89 @@ export const applicationConfigSchema = {
 				},
 			},
 		},
+		genesis: {
```
ManuGowda, line 385: some of the props should go into specific module?

framework/src/schema/application_config_schema.ts
nazarhussain: Yes I believe module specific config must be splitted. But I am not sure which one are those and when to do it.   @shuse2 may explain here.

framework/src/schema/application_config_schema.ts
nazarhussain: Yes these should go, but may be @shuse2  plan is to do it with the issues for each module when we start.

framework/src/schema/application_config_schema.ts
nazarhussain: When the implementation of that LIP starts, we should make that change.

ManuGowda (approved):

PR #6610, opened 2021-08-03
Fix transport to encode BigNumber to string - Closes #6609

### LINKED ISSUE #6609: Block processing failure at transport layer due to postBlock event received invalid event message
### Expected behavior

The node should process a valid BigNumber amount.

### Actual behavior

Failed to process the block when `totalAmount` is a number is greater than 15 digits.
`
{"name":"lisk-framework","hostname":"nbg1-node-01.ed148.net","pid":1178979,"level":40,"error":{"name":"BigNumber Error"},"event":"postBlock","msg":"Received invalid event message","time":"2021-08-03T09:51:50.501Z","v":0}
`

### Steps to reproduce

- Send balance transfer transaction with amount grater than 15 digits
- Observe the forger can generate a block

lisk-sdk@2.3.8, lisk-core@2.1.6

shuse2:
### What was the problem?

This PR resolves #6609

### How was it solved?

- Update transport to encode big number to string
- Lock sinon-chai to old version for building

### How was it tested?

Run network with delegate nodes and send transaction with amount `1300262865374292`

### DIFF
--- a/framework/src/modules/chain/transport/transport.js
+++ b/framework/src/modules/chain/transport/transport.js
@@ -155,15 +155,15 @@ class Transport {
 		}
 
 		if (block.totalAmount) {
-			block.totalAmount = block.totalAmount.toNumber();
+			block.totalAmount = block.totalAmount.toString();
 		}
 
 		if (block.totalFee) {
-			block.totalFee = block.totalFee.toNumber();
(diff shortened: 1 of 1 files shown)


ManuGowda (approved):
LGTM

Nazgolze (approved):

PR #6645, opened 2021-08-12
Codec fromJSON and toJSON fails on extra properties #5670

### LINKED ISSUE #5670: Codec fromJSON and toJSON fails on extra properties
### Expected behavior
codec fromJSON and toJSON should ignore extra peroperties and encode/decode without error.

### Actual behavior
codec.fromJSON or toJSON fails if the object contains extra properties.

### Steps to reproduce
```
const schema = {
	$id: '/schema',
	type: 'object',
	properties: {
		valueA: {
			dataType: 'uint64',
			fieldNumber: 1,
		},
	},
	required: ['valueA'],
};
codec.fromJSON(schema, { valueA: '123', extra: true });
codec.toJSON(schema, { valueA: BigInt(999), extra: true });
```

5.0-

nazarhussain:
### What was the problem?

This PR resolves #5670

### How was it solved?

- Fixed the issue for throwing error on extra attributes 

### How was it tested?

- Added unit tests for ignoring extra attributes on `fromJSON` and `toJSON`

### DIFF
--- a/elements/lisk-codec/src/json_wrapper.ts
+++ b/elements/lisk-codec/src/json_wrapper.ts
@@ -12,7 +12,7 @@
  *
  * Removal or modification of this copyright notice is prohibited.
  */
-
+import { objects } from '@liskhq/lisk-utils';
 import {
 	BaseTypes,
 	IteratableGenericObject,
@@ -56,7 +56,7 @@ const mappers: mappersInterface = {
 /* eslint-enable @typescript-eslint/explicit-function-return-type */
 
 const findObjectByPath = (message: SchemaProps, pathArr: string[]): SchemaProps | undefined => {
-	let result: SchemaProps = message;
+	let result: SchemaProps = objects.cloneDeep(message);
 	for (let i = 0; i < pathArr.length; i += 1) {
 		if (!result.properties && !result.items) {
 			return undefined;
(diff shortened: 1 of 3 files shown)


elements/lisk-codec/src/json_wrapper.ts
```diff
@@ -56,7 +56,7 @@ const mappers: mappersInterface = {
 /* eslint-enable @typescript-eslint/explicit-function-return-type */
 
 const findObjectByPath = (message: SchemaProps, pathArr: string[]): SchemaProps | undefined => {
-	let result: SchemaProps = message;
+	let result: SchemaProps = objects.cloneDeep(message);
```
shuse2, line 59: if we need to clone the object recursively, wouldn't it be very slow?

nazarhussain: I am not sure about exact/future usage of this function, but current implementation is kind of changing the value of the function parameter, which is an anti-pattern. 

https://github.com/LiskHQ/lisk-sdk/blob/b0811447013466171f2d63ab813125cb8e41f658/elements/lisk-codec/src/json_wrapper.ts#L68

BTW this function `findObjectByPath` is only used in the json-wrapper so would not actually have any impact on actual encoding/decoding part.

ManuGowda (approved):

shuse2 (approved):

PR #6646, opened 2021-08-12
API client node info type is not matching to returned data - Closes #6603

### LINKED ISSUE #6603: API client node info type is not matching to returned data
### Expected behavior

The `client.node.getNodeInfo()` must have matching type with the return value of `app:getNodeInfo` action. 

### Actual behavior

Currently `genesisHeight` and `network` attributes are missing from the type. 

https://github.com/LiskHQ/lisk-sdk/blob/4ced142da46c76b529b78e39ca35d2e6ac94a322/elements/lisk-api-client/src/types.ts#L97-L101

5.1.x

nazarhussain:
### What was the problem?

This PR resolves #6603

### How was it solved?

- Update types matching to following. 

https://github.com/LiskHQ/lisk-sdk/blob/1f7a4216c9d1746c6f3b61635bdf457ea35f230e/framework/src/node/node.ts#L500-L522

### How was it tested?

- No test is updated


elements/lisk-api-client/src/types.ts
```diff
@@ -100,11 +100,26 @@ export interface NodeInfo {
 	readonly networkIdentifier: string;
 	readonly lastBlockID: string;
 	readonly height: number;
+	readonly genesisHeight: number;
 	readonly finalizedHeight: number;
 	readonly syncing: boolean;
 	readonly unconfirmedTransactions: number;
 	readonly genesisConfig: GenesisConfig;
 	readonly registeredModules: RegisteredModule[];
+	readonly network: {
```
ManuGowda, line 109: maybe include this in this issue as well?

ManuGowda (approved):
LGTM

elements/lisk-api-client/src/types.ts
nazarhussain: Updated the the issue.

shuse2 (approved):

PR #6650, opened 2021-08-13
Improve lisk-codec protocol specs to be more deterministic - Closes #6589

### LINKED ISSUE #6589: Improve lisk-codec protocol specs to be more deterministic
### Description

The protocol specs of any component in the system must be deterministic, means input and output must defined explicitly for every case. And there must not be any modifications/mutations to output/input runtime to pass the tests.  

Currently we have protocol specs for lisk-codec `encoding` behaviour, but we don't have it for `decoding` behaviour.  And as per our implementation of `lisk-codec` and by specification of the LIP the encoding and decoding may not turn out be same as always. e.g. Its a viable chance that in example below `data` and `reverseData` may not match strictly because of the default values for properties which are not defined in the `data` object. 

```
const data = {a: 1, b: 2}
const binary = codec.encode(schema, data);
const reverseData = codec.decode(schema, binary);
```

To cover this part we had a lot of custom manipulation of protocol specs in test suites, which does not fulfil the purpose of having protocol specs. e.g. few of reference mentioned below. We need to improve these and protocol specs must define the decoding specs and we should avoid runtime mutation of specs input/output in our tests

https://github.com/LiskHQ/lisk-sdk/blob/2db69cdd52b2ea8461525a34c388619921481c64/elements/lisk-codec/test/decode.spec.ts#L67-L71

https://github.com/LiskHQ/lisk-sdk/blob/2db69cdd52b2ea8461525a34c388619921481c64/elements/lisk-codec/test/decode.spec.ts#L113-L117

https://github.com/LiskHQ/lisk-sdk/blob/2db69cdd52b2ea8461525a34c388619921481c64/elements/lisk-codec/test/decode.spec.ts#L161-L167

### Motivation

The protocol specs for Lisk-Codec must deterministic and cover both encoding and decoding behaviours.

nazarhussain:
### What was the problem?

This PR resolves #6589

### How was it solved?

- Created decoding specs for every existing encoding spec. 
- Standardize all output for lisk-codec to use consistent format.
- Updated all unit tests which were mutating the specs while running decoding scenarios.
- Simply `fromJSON` and `toJSON` tests, as they will simply be exactly matching with specs. 

### How was it tested?

- Run all unit tests


protocol-specs/generator_outputs/transaction_merkle_root/transaction_merkle_root.json
```diff
@@ -17,1694 +17,1694 @@
 		{
 			"description": "Given valid transaction ids: 1",
 			"input": {
-				"transactionIds": ["051f9de183c217138a91c2633e0414026b39f53b026a8c8675c9f54ab81a2f2f"]
+				"transactionIds": ["26aaec792a11985e452678996a112674064aa2b4b86212d6c1159e9985f97554"]
```
shuse2, line 20: why is this canging?

shuse2 (approved):

protocol-specs/generator_outputs/transaction_merkle_root/transaction_merkle_root.json
nazarhussain: Because its using `getRandomBytes` to generate the fixtures. I will exclude any changes to unwanted files from this PR.

ishantiw (approved):

PR #6657, opened 2021-08-17
Add JSON RPC over HTTP protocol - Closes #6618

### LINKED ISSUE #6618: Add JSON RPC over HTTP protocol
### Description

Add JSON RPC over HTTP protocol by creating a HTTP server in Controller where other communication servers are created.
#### Add Server

```ts
// Controller
const WSServer = require('ws').Server;
const server = require('http').createServer();
```
#### Add HTTP as config options 

```ts
rpc: {
	ipc: { path: ‘tmp/external/’},
 	ws: { port: 8080 },
 	http: { port: 8000 }, // Add config option for http
        modes: ['ipc', 'ws', 'http'] // Add to possible values of modes
}
```

### Acceptance Criteria

- Create http server and listen at specified `port` when mode `http` is defined
- Add unit, integration and functional tests for JSON RPC over HTTP protocol

### Additional Information

- Related #6612

ManuGowda:
### What was the problem?

This PR resolves #6618 

### How was it solved?

- Implemented JSON RPC over HTTP protocol

### How was it tested?

- Added tests

### DIFF
--- a/framework/src/controller/http/http_server.ts
+++ b/framework/src/controller/http/http_server.ts
@@ -14,43 +14,63 @@
 import * as HTTP from 'http';
 import { Logger } from '../../logger';
 
-export type HTTPRequestListener = (req: HTTP.IncomingMessage, res: HTTP.ServerResponse, message: string) => void;
+export type HTTPRequestListener = (
+	req: HTTP.IncomingMessage,
+	res: HTTP.ServerResponse,
+	message: string,
+) => void;
 
+const ALLOWED_METHODS = ['GET', 'POST'];
 export class HTTPServer {
-  public server!: HTTP.Server;
-  private readonly port: number;
-  private readonly host?: string;
-  private readonly logger: Logger;
+	public server!: HTTP.Server;
+	private readonly port: number;
+	private readonly host?: string;
+	private readonly logger: Logger;
 
-  public constructor(options: { port: number; host?: string; logger: Logger }) {
-    this.host = options.host;
-    this.port = options.port;
-    this.logger = options.logger;
-  }
+	public constructor(options: { port: number; host?: string; logger: Logger }) {
+		this.host = options.host;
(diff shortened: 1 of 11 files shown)


framework/src/controller/http/http_server.ts
```diff
@@ -0,0 +1,56 @@
...
+      }
+      const message = Buffer.concat(buffers).toString();
+      return httpRequestListener(req, res, message);
+    });
+    this.server.listen(this.port, this.host, () => {
+      this.logger.info(`HTTP Server is listening at port ${this.port}`)
+    });
+    this.server.on('error', error => {
+      this.logger.error(error);
+    });
```
mehmetegemen, line 46: If we don't do any other thing than logging maybe we can simplify it as
```
this.server.on('error', this.logger.error);
```

framework/src/controller/http/http_server.ts
```diff
@@ -14,43 +14,63 @@
...
+	}
+
+	public start(httpRequestListener: HTTPRequestListener): HTTP.Server {
+		// eslint-disable-next-line @typescript-eslint/no-misused-promises
+		this.server = HTTP.createServer(async (req, res) => {
+			const headers = {
+				'Access-Control-Allow-Origin': '*',
+				'Access-Control-Allow-Methods': ALLOWED_METHODS.join(','),
+				'Content-Type': 'application/json',
+			};
```
mehmetegemen, line 43: Can `headers` have an information from `req` or `res` in the future? Otherwise it's a constant piece of information and maybe we can move it out as `ALLOWED_METHODS`.

ManuGowda: @mehmetegemen this is the server part, i do think we will have anything for request. so the current approach should be good.

mehmetegemen (approved):

Incede (approved):

PR #6664, opened 2021-08-19
 Improve plugin class and align it with new architecture - Closes #6623 

### LINKED ISSUE #6623: Improve plugin class and align it with new architecture
#### Init
To be consistent with the module, plugins should also use the “init” function instead of constructor.
```ts
init(context: PluginInitContext): Promise<void>;
interface PluginInitContext {
  config: Record<string, unknown>; // plugin config
  channel: PluginChannel;
  options: {
    dataPath: string;
    applicationConfig: ApplicationConfig; // Omit<ApplicationConfig, ‘plugins’> should not have plugin related property
  },
}
```
#### name

Rename the `alias` to `name` public property to be consistent with the `module`. Content should be like `httpAPI`.

```ts
public abstract name: string;
```

#### info

Remove info property and use `name` property instead wherever required. 
`exportPath` is needed if a plugin is expected to run in child process and must be implemented in the extended class. Rename to `nodeModulePath` as a function defined as below.
```ts
public abstract get nodeModulePath(): string;
```

#### default

Rename it to `configSchema`, or make it consistent with the module.

#### Update Tests

Update plugin related tests to adapt to prop changes and config input changes.

### Acceptance Criteria

- All the tests under `framework` and `framework-plugin` should pass

### Additional Information

- Related #6612

Incede:
### What was the problem?

This PR resolves #6623

### How was it solved?

Changes made as per the issue

### How was it tested?

All necessary tests updated

### DIFF
--- a/framework/src/application.ts
+++ b/framework/src/application.ts
@@ -340,7 +340,7 @@ export class Application {
 	}
 
 	private async _loadPlugins(): Promise<void> {
-		const { plugins, ...rest } = this.config; 
+		const { plugins, ...rest } = this.config;
 		const appConfigForPlugin: ApplicationConfigForPlugin = rest;
 		await this._controller.loadPlugins(this._plugins, this.config.plugins, appConfigForPlugin);
 	}
--- a/framework/src/application.ts
+++ b/framework/src/application.ts
@@ -189,18 +189,18 @@ export class Application {
 	): void {
 		assert(pluginKlass, 'Plugin implementation is required');
 		assert(typeof options === 'object', 'Plugin options must be provided or set to empty object.');
-
-		const pluginAlias = options?.alias ?? pluginKlass.name;
+		const pluginObject = new pluginKlass();
+		const pluginAlias = options?.alias ?? pluginObject.name;
 
 		assert(
 			!Object.keys(this._plugins).includes(pluginAlias),
-			`A plugin with alias "${pluginAlias}" already registered.`,
+			`A plugin with name "${pluginAlias}" already registered.`,
 		);
 
 		if (options.loadAsChildProcess) {
 			if (!getPluginExportPath(pluginKlass)) {
 				throw new Error(
-					`Unable to register plugin "${pluginAlias}" to load as child process. \n -> To load plugin as child process it must be exported. \n -> You can specify npm package as "info.name". \n -> Or you can specify any static path as "info.exportPath". \n -> To fix this issue you can simply assign __filename to info.exportPath in your plugin.`,
+					`Unable to register plugin "${pluginAlias}" to load as child process. \n -> To load plugin as child process it must be exported. \n -> You can specify npm package as "name". \n -> Or you can specify any static path as "nodeModulePath". \n -> To fix this issue you can simply assign __filename to nodeModulePath in your plugin.`,
 				);
 			}
 		}
@@ -210,7 +210,7 @@ export class Application {
 			options,
 		);
 
-		validatePluginSpec(pluginKlass, this.config.plugins[pluginAlias]);
+		validatePluginSpec(pluginObject);
 
 		this._plugins[pluginAlias] = pluginKlass;
 	}
--- a/framework/src/application.ts
+++ b/framework/src/application.ts
@@ -42,14 +42,6 @@ import {
 	RegisteredModule,
 	UpdateForgingStatusInput,
 	PartialApplicationConfig,
-	ApplicationConfig,
-	GenesisConfig,
-	EventPostTransactionData,
-	PluginOptions,
-	RegisteredSchema,
-	RegisteredModule,
-	UpdateForgingStatusInput,
-	PartialApplicationConfig,
 	PluginOptionsWithApplicationConfig,
 } from './types';
 
@@ -189,8 +181,8 @@ export class Application {
 	): void {
 		assert(pluginKlass, 'Plugin implementation is required');
 		assert(typeof options === 'object', 'Plugin options must be provided or set to empty object.');
-		const pluginObject = new pluginKlass();
-		const pluginAlias = options?.alias ?? pluginObject.name;
+		const PluginObject = new pluginKlass();
+		const pluginAlias = options?.alias ?? PluginObject.name;
 
 		assert(
 			!Object.keys(this._plugins).includes(pluginAlias),
@@ -210,7 +202,7 @@ export class Application {
 			options,
 		);
 
-		validatePluginSpec(pluginObject);
+		validatePluginSpec(PluginObject);
 
 		this._plugins[pluginAlias] = pluginKlass;
 	}
--- a/framework/src/application.ts
+++ b/framework/src/application.ts
@@ -42,8 +42,15 @@ import {
 	RegisteredModule,
 	UpdateForgingStatusInput,
 	PartialApplicationConfig,
-	PluginOptionsWithAppConfig,
-	AppConfigForPlugin,
+	ApplicationConfig,
+	GenesisConfig,
+	EventPostTransactionData,
+	PluginOptions,
+	RegisteredSchema,
+	RegisteredModule,
+	UpdateForgingStatusInput,
+	PartialApplicationConfig,
+	PluginOptionsWithApplicationConfig,
 } from './types';
 
 import {
@@ -59,7 +66,6 @@ import { Node } from './node';
 import { Logger, createLogger } from './logger';
 
 import { DuplicateAppInstanceError } from './errors';
-
 import { BaseModule, TokenModule, SequenceModule, KeysModule, DPoSModule } from './modules';
 
 const MINIMUM_EXTERNAL_MODULE_ID = 1000;
@@ -184,7 +190,7 @@ export class Application {
 		assert(pluginKlass, 'Plugin implementation is required');
 		assert(typeof options === 'object', 'Plugin options must be provided or set to empty object.');
 
-		const pluginAlias = options?.alias ?? pluginKlass.alias;
+		const pluginAlias = options?.alias ?? pluginKlass.name;
 
 		assert(
 			!Object.keys(this._plugins).includes(pluginAlias),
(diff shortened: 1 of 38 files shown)


framework/src/controller/controller.ts
```diff
@@ -171,16 +171,15 @@ export class Controller {
...
 	private async _loadInMemoryPlugin(
-		alias: string,
+		name: string,
 		Klass: InstantiablePlugin,
-		options: PluginOptionsWithAppConfig,
+		options: PluginOptionsWithApplicationConfig,
 	): Promise<void> {
-		const pluginAlias = alias || Klass.alias;
-		const { name, version } = Klass.info;
+		const pluginAlias = name || Klass.name;
```
shuse2, line 178: This should use `plugin.name`

framework/src/controller/controller.ts
```diff
@@ -189,25 +188,29 @@ export class Controller {
...
 
 	private async _loadChildProcessPlugin(
-		alias: string,
+		name: string,
 		Klass: InstantiablePlugin,
 		options: PluginOptions,
 	): Promise<void> {
-		const pluginAlias = alias || Klass.alias;
-		const { name, version } = Klass.info;
+		const pluginAlias = name || Klass.name;
```
shuse2, line 211: same as above

shuse2 (changes requested):

framework/src/plugins/base_plugin.ts
```diff
@@ -195,29 +184,41 @@ export interface PluginCodec {
 	encodeTransaction: (transaction: TransactionJSON) => string;
 }
 
+interface PluginInitContext {
```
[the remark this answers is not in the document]
Incede: I think this can't be done because creates a dependency cycle due to `BaseChannel` via bus if moved to types.ts

framework/src/application.ts
```diff
@@ -183,18 +181,18 @@ export class Application {
 	): void {
 		assert(pluginKlass, 'Plugin implementation is required');
 		assert(typeof options === 'object', 'Plugin options must be provided or set to empty object.');
-
-		const pluginAlias = options?.alias ?? pluginKlass.alias;
+		const PluginObject = new pluginKlass();
+		const pluginAlias = options?.alias ?? PluginObject.name;
```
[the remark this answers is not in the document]
Incede: https://github.com/LiskHQ/lisk-sdk/blob/3b304af2b2323d3c786c642a9a91fe8931f99155/framework/test/unit/application.spec.ts#L531
In some test name is being passed through options, remove those test too ?

framework/src/plugins/base_plugin.ts
```diff
@@ -301,23 +292,25 @@ export const getPluginExportPath = (
...
 	let nodeModulePath: string | undefined;
 
+	const PluginInstance = new pluginKlass();
+
 	try {
 		// Check if plugin name is an npm package
 		// eslint-disable-next-line global-require, import/no-dynamic-require, @typescript-eslint/no-var-requires, @typescript-eslint/no-unsafe-assignment
-		nodeModule = require(pluginKlass.info.name);
-		nodeModulePath = pluginKlass.info.name;
+		nodeModule = require(PluginInstance.name);
```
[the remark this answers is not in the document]
Incede: https://github.com/LiskHQ/lisk-sdk/blob/development/framework/test/unit/plugins/base_plugin.spec.ts#L146
I think this test is based on this try-catch block, and fails upon it's removal.

framework/src/application.ts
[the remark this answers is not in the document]
shuse2: yea sounds good! (as we dont use alias anymore

framework/src/plugins/base_plugin.ts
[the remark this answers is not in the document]
shuse2: Previously `info.name` was  the package name. That's why `require` with package name make sense. Now, the below block take care both patterns, so this is not necessary. test also can be removed as well

shuse2 (commented):
LGTM in general!

framework/src/plugins/base_plugin.ts
```diff
@@ -195,29 +182,41 @@ export interface PluginCodec {
...
+	channel: BaseChannel;
+	options: {
+		readonly dataPath: string;
+		appConfig: Omit<ApplicationConfig, 'plugins'>;
+	};
+}
+
 export abstract class BasePlugin<
-	T extends PluginOptionsWithAppConfig = PluginOptionsWithAppConfig
+	T extends PluginOptionsWithApplicationConfig = PluginOptionsWithApplicationConfig
```
nazarhussain, line 195: Since we made plugin `config` and `appConfig` two separate things, so this generic type needed to be changed. It should just be plugin config. If we want to really save `options` at plugin instance level then we create another attribute in the base plugin for it.

nazarhussain (changes requested):

framework/src/plugins/base_plugin.ts
```diff
@@ -195,29 +182,41 @@ export interface PluginCodec {
...
 > {
-	public readonly options: T;
+	public options!: T;
```
nazarhussain, line 197: As said earlier, these options should be splitted as `config` and `options`

framework/src/plugins/base_plugin.ts
```diff
@@ -195,29 +182,41 @@ export interface PluginCodec {
...
+	public async init(context: PluginInitContext): Promise<void> {
+		if (this.configSchema) {
 			this.options = objects.mergeDeep(
 				{},
-				(this.defaults as SchemaWithDefault).default ?? {},
-				options,
+				(this.configSchema as SchemaWithDefault).default ?? {},
+				context.options,
+				context.config,
 			) as T;
```
nazarhussain, line 211: It should not merge the config and options together.

framework/src/plugins/base_plugin.ts
```diff
@@ -252,41 +251,31 @@ export abstract class BasePlugin<
...
-		throw new ImplementationMissingError();
-	}
-	// eslint-disable-next-line @typescript-eslint/no-empty-function
-	public static get info(): PluginInfo {
-		throw new ImplementationMissingError();
-	}
-
 	// TODO: To make non-breaking change we have to keep "object" here
-	public get defaults(): SchemaWithDefault | object | undefined {
+	public get configSchema(): SchemaWithDefault | object | undefined {
```
nazarhussain, line 272:
```suggestion
	public get configSchema(): SchemaWithDefault | undefined {
```

framework/src/plugins/base_plugin.ts
```diff
@@ -295,64 +284,58 @@ export abstract class BasePlugin<
 // TODO: Once the issue fixed we can use require.resolve to rewrite the logic
 //  https://github.com/facebook/jest/issues/9543
 export const getPluginExportPath = (
```
nazarhussain, line 286: Please change this function name as well matching to plugin property you renamed to `nodeModulePath`

framework/src/plugins/base_plugin.ts
```diff
@@ -295,64 +284,58 @@ export abstract class BasePlugin<
...
-	assert(PluginKlass.alias, 'Plugin alias is required.');
-	assert(PluginKlass.info.name, 'Plugin name is required.');
-	assert(PluginKlass.info.author, 'Plugin author is required.');
-	assert(PluginKlass.info.version, 'Plugin version is required.');
-	assert(pluginObject.events, 'Plugin events are required.');
-	assert(pluginObject.actions, 'Plugin actions are required.');
+export const validatePluginSpec = (plugin: BasePlugin): void => {
+	assert(plugin.name, 'Plugin name is required.');
+	assert(plugin.events, 'Plugin events are required.');
+	assert(plugin.actions, 'Plugin actions are required.');
```
nazarhussain, line 331: In past we discussed to not use `assert` package any more. If that's the case we should refactor it now as we are already changing this file a lot. @shuse2

framework/src/plugins/base_plugin.ts
```diff
@@ -252,41 +251,31 @@ export abstract class BasePlugin<
...
-	}
-
 	// TODO: To make non-breaking change we have to keep "object" here
-	public get defaults(): SchemaWithDefault | object | undefined {
+	public get configSchema(): SchemaWithDefault | object | undefined {
 		return undefined;
 	}
+	public abstract get nodeModulePath(): string;
 	public abstract get events(): EventsDefinition;
 	public abstract get actions(): ActionsDefinition;
```
nazarhussain, line 277: I will suggest to remove these from `abstract` and not make these required for plugin developer to define. One can just develop a plugin without defining events and actions.

nazarhussain (changes requested):

framework/src/plugins/base_plugin.ts
shuse2: we don't use assert anymore, but since this PR is blocking other PR and it's already pretty big, refactoring later is also fine. up to @Incede  =)

framework/src/plugins/base_plugin.ts
shuse2: I will create issue to make properties to be consistent

nazarhussain (approved):
We decided to proceed with this PR but following points to be covered separately. 


1. Make plugin `name` a property rather the getter.
2. Rename function overridePluginOptions to updatePluginConfig  to be consistent across naming convention. 
3. Rename all files in plugins  `defaults/config.ts`  to` cofnigSchema.ts`  or `schema/config.ts`
4. Don’t need the code `objects.mergeDeep` in every plugin load  its already done in the base plugin. 
5. Need to remove `events` and `actions` from abstract members and create properties with default values. 
6. If possible create an issue to remove dependency of declaring `nodeModulePath` in every plugin.
7. Two functions `_loadInMemoryPlugin` and `_loadChildProcessPlugin` have extra parameter `name`. which can be removed now. Earlier it was the feature to register a plugin with different alias, now we restrict everywhere to just use plugin instance name. 
Remove the use of assert around new code case

nazarhussain:
Created the issue to cover above points.

shuse2 (approved):

PR #6682, opened 2021-08-24
Husky support is broken for pre-commit hooks - Closes #6681

### LINKED ISSUE #6681: Husky support is broken for pre-commit hooks
### Expected behavior

When we commit changes to the repo, the`prettier` and `eslint` must be executed before that. Which was earlier been configured using `huky` and `lint-staged` package. Which seems to be broken. 

### Actual behavior

Currently on commit no hook is executed. 

### Steps to reproduce

1. Make changes to any file locally 
2. Run `git commit -m "test commit"
3. Notice if the hooks are executed 

development

nazarhussain:
### What was the problem?

1. The `husky` was not installed automatically anymore. 
2. The `husky` no more support JS/Package.json based configuration 

For more reference. 

https://blog.typicode.com/husky-git-hooks-javascript-config/
https://blog.typicode.com/husky-git-hooks-autoinstall/

This PR resolves #6681

### How was it solved?

- Updated the `husky` and `lint-staged` packages 
- Added `husky` file based hooks 
- Added `prepare` script for the `Yarn 1` and `npm`
- Added `postinstall` script for `Yarn 2`

### How was it tested?

- Tested manually

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,12 +51,9 @@
 		"test:framework-plugins": "lerna run test --scope=@liskhq/lisk-framework-*",
 		"test:commander": "lerna run test --scope=lisk-commander",
 		"build": "lerna run build",
-		"init": "./scripts/init.sh"
-	},
-	"husky": {
-		"hooks": {
-			"pre-commit": "lint-staged"
-		}
+		"init": "./scripts/init.sh",
+		"prepare": "husky install",
+		"postinstall": "husky install"
 	},
 	"devDependencies": {
 		"@types/node": "12.20.6",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -51,12 +51,9 @@
...
 		"build": "lerna run build",
-		"init": "./scripts/init.sh"
-	},
-	"husky": {
-		"hooks": {
-			"pre-commit": "lint-staged"
-		}
+		"init": "./scripts/init.sh",
+		"prepare": "husky install",
+		"postinstall": "husky install"
```
shuse2, line 56: `prepare` and `postinstall` both runs for me on yarn1. probably only `postintall` is fine?

nazarhussain: > Yarn 2 doesn't support `prepare`  lifecycle script, so husky needs to be installed differently (this doesn't apply to Yarn 1 though). See Yarn 2 install.

https://typicode.github.io/husky/#/

That's from their docs, I thought to add both, so when in future we upgrade `yarn` we don't miss this. Let me know if you think to remove one.

shuse2 (approved):

PR #6699, opened 2021-08-30
 Update lisk-commander with Plugin class changes - Closes #6636 

### LINKED ISSUE #6636: Update lisk-commander with Plugin class changes
### Description

Update Lisk Commander with Plugin class changes done in #6623 . Update bootstrapping commands and other places in Lisk Commander where Plugins are used.

### Acceptance Criteria

- All the tests in Lisk Commander should pass
- `lisk generate:plugin <name>` should successfully bootstrap plugin

### Additional Information

- Related #6612

Incede:
### What was the problem?

This PR resolves #6636 

### How was it solved?
* Updated necessary changes for plugin

### How was it tested?

* `yarn test` under commander
* Bootstrap a new plugin using `./bin/run generate:plugin <name>`


commander/src/commands/generate/plugin.ts
```diff
@@ -26,7 +26,7 @@ export default class PluginCommand extends BaseBootstrapCommand {
 	static args = [
 		{
 			name: 'alias',
```
shuse2, line 28: name should be changed to name here and the usage

commander/src/bootstrapping/templates/lisk-template-ts/templates/init_plugin/src/plugin.ts
```diff
@@ -1,38 +1,14 @@
...
-			// 'block:missed'
-		];
-	}
-
-	public get actions(): ActionsDefinition {
-		return {
-		// 	hello: async () => { hello: 'world' },
-		};
-	}
+	public name:'<%= name %>';
```
shuse2, line 9: maybe it would be nice to have 
```
public nodeModulePath = __filename;
```
as well for easy use, and keep `events` and `actions` so that user can start implementing easily

shuse2 (changes requested):

commander/src/commands/generate/plugin.ts
```diff
@@ -26,7 +26,7 @@ export default class PluginCommand extends BaseBootstrapCommand {
...
-			description: 'Alias of the plugin.',
+			description: 'Name of the plugin.',
 			required: true,
 		},
 	];
```
ManuGowda, line 32: The args, variable name need to be updated

ManuGowda (commented):
The tests need to be updated as well?

ManuGowda (approved):

PR #6712, opened 2021-09-03
Add reward module initials - Closes #6687

### LINKED ISSUE #6687: Create class and constants for Rewards Module
### Description
Create rewards module class and constants except a schema since the module has an empty state and has no commands.

- Define constants at `constants.ts` as specified in https://research.lisk.com/t/define-state-transitions-of-reward-module/316#constants-6
- Define an empty reward module class with necessary initial variables that base module requires
- Define the interface for the dependency modules and add dependency function in the class (Dependencies are Random module, Token module and BFT module)
- Add config schema to the class

### Acceptance Criteria
- Constants and module files should exist

### Additional Information
- `TOKEN_ID_REWARD` will be specified in config parameter of `init` method of the reward module class.

### Related Issue
#6686

mehmetegemen:
### What was the problem?

This PR resolves #6687

### How was it solved?

- Added required files for further development
- Created the class and bound dependencies to it

### How was it tested?

- Manually checked files


framework/src/modules/reward/schemas.ts
```diff
@@ -0,0 +1,32 @@
...
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+export const configSchema = {
+	$id: '/reward/config',
+	type: 'object',
+	properties: {
+		tokenIDReward: {
```
shuse2, line 19: I think it's better to have `required` field for `chainID` and `localID`

shuse2 (changes requested):
One minor thing but looks good to me

mitsuaki-u (approved):
LGTM besides @shuse2 's comment

shuse2 (approved):

PR #6741, opened 2021-09-07
Refactor controller - Closes #6707

### LINKED ISSUE #6707: Application constructor should not open handler
### Description

Framework constructor should not open any handler.
All handlers (file, DB or port etc) should be opened on `run`.
All open handles should be closed when shutdown.

### Acceptance Criteria

- No open handle remains after shutting down

shuse2:
### What was the problem?

This PR resolves #6707 

### How was it solved?

- Refactor controller to handle the plugins
- Update constructor not to use open handles (zeromq class opens handler, so moved to `start`)

### How was it tested?

- All unit test should pass
- test app should start (although it will fail on for ` forgersList is not iterable`

### DIFF
--- a/framework/src/application.ts
+++ b/framework/src/application.ts
@@ -20,20 +20,18 @@ import { Block } from '@liskhq/lisk-chain';
 import { KVStore } from '@liskhq/lisk-db';
 import { validator, LiskValidationError } from '@liskhq/lisk-validator';
 import { objects, jobHandlers } from '@liskhq/lisk-utils';
-import { APP_EVENT_SHUTDOWN, APP_EVENT_READY, APP_IDENTIFIER } from './constants';
+import { APP_EVENT_SHUTDOWN, APP_EVENT_READY } from './constants';
 import {
-	RPCConfig,
 	ApplicationConfig,
 	PluginConfig,
 	RegisteredSchema,
 	RegisteredModule,
 	PartialApplicationConfig,
-	ApplicationConfigForPlugin,
 	EndpointHandlers,
 	PluginEndpointContext,
 } from './types';
 
-import { BasePlugin, getPluginExportPath, validatePluginSpec } from './plugins/base_plugin';
+import { BasePlugin } from './plugins/base_plugin';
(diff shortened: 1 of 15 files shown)


framework/src/application.ts
```diff
@@ -243,20 +219,20 @@ export class Application {
...
-			await this._loadPlugins();
+			await this._controller.start();
 			await this._node.start();
-			this.logger.debug(this._controller.bus.getEvents(), 'Application listening to events');
-			this.logger.debug(this._controller.bus.getEndpoints(), 'Application ready for actions');
+			this.logger.debug(this._controller.getEvents(), 'Application listening to events');
+			this.logger.debug(this._controller.getEndpoints(), 'Application ready for actions');
 
-			this._channel.publish(APP_EVENT_READY);
+			this._controller.channel.publish(APP_EVENT_READY);
```
ishantiw, line 235: we can use `this.channel`

framework/src/application.ts
```diff
@@ -126,19 +122,25 @@ export class Application {
...
+			appConfig: rootConfigs,
+			pluginConfigs: plugins,
+		});
 	}
 
 	public get networkIdentifier(): Buffer {
 		return this._node.networkIdentifier;
 	}
 
+	public get channel(): InMemoryChannel {
```
ishantiw, line 140: before `app.run` if user accesses `app.channel` will it throw error? or should return be `InMemory | undefined`

framework/test/unit/controller/http/http_server.spec.ts
```diff
@@ -15,33 +15,32 @@ import * as HTTP from 'http';
 import { HTTPServer } from '../../../../src/controller/http/http_server';
 
 describe('HTTPServer', () => {
+	const logger = {
```
ishantiw, line 18: can we use channelMock from testing here?

ishantiw (commented):
LGTM, just some minor comments

Incede (approved):

ishantiw (approved):

PR #6748, opened 2021-09-08
 Create module class, constants, schemas for validators module - Closes #6669 

### LINKED ISSUE #6669: Create module class, constants, schemas for validators module
### Description

Create validators module and define constants and schema. Add validators module and constants.ts, schema.ts and types.ts under the folder as specified in https://research.lisk.com/t/introduce-validators-module/317.

- Define store prefixes and invalid bls keys in constant.ts
- Define all data type in types.ts
- Define all schema for store in schema.ts

Implement `init` hook, and validate and store `block time` from the config in memory.

### Acceptance Criteria
- All 4 files exist in the validators module folder

### Additional Information
- Empty validator module file already exist
- blockTime will be defined in `moduleConfig.blockTime`

Related issue: #6668

Incede:
### What was the problem?

This PR resolves #6669 

### How was it solved?

* Create initial files for validators module
* Rename all instances of `validator` with `validators`
* Fix failing unit tests under framework due to new changes

### How was it tested?
* `yarn test`

### DIFF
--- a/framework/src/modules/validators/constants.ts
+++ b/framework/src/modules/validators/constants.ts
@@ -1,0 +1,20 @@
+/*
+ * Copyright © 2021 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+export const MODULE_ID_VALIDATORS = 11; // TBD
+export const STORE_PREFIX_VALIDATORS_DATA = Buffer.from('0000');
+export const STORE_PREFIX_GENERATOR_LIST = Buffer.from('4000', 'hex');
+export const STORE_PREFIX_BLS_KEYS = Buffer.from('8000', 'hex');
+export const STORE_PREFIX_GENESIS_DATA = Buffer.from('c000', 'hex');
+export const INVALID_BLS_KEY = Buffer.alloc(48);
--- a/framework/src/modules/validators/constants.ts
+++ b/framework/src/modules/validators/constants.ts
@@ -13,8 +13,8 @@
  */
 
 export const MODULE_ID_VALIDATORS = 11; // TBD
-export const STORE_PREFIX_VALIDATORS_DATA = Buffer.from('0000');
-export const STORE_PREFIX_GENERATOR_LIST = Buffer.from('4000', 'hex');
-export const STORE_PREFIX_BLS_KEYS = Buffer.from('8000', 'hex');
-export const STORE_PREFIX_GENESIS_DATA = Buffer.from('c000', 'hex');
+export const STORE_PREFIX_VALIDATORS_DATA = 0x0000;
+export const STORE_PREFIX_GENERATOR_LIST = 0x4000;
+export const STORE_PREFIX_BLS_KEYS = 0x8000;
+export const STORE_PREFIX_GENESIS_DATA = 0xc000;
 export const INVALID_BLS_KEY = Buffer.alloc(48);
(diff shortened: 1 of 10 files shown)


framework/src/modules/validators/module.ts
```diff
@@ -0,0 +1,45 @@
...
+	public name = 'validators';
+	public api = new ValidatorsAPI(this.id);
+	public endpoint = new ValidatorsEndpoint(this.id);
+	private _blockTime!: number;
+	private _moduleConfig!: ModuleConfig;
+
+	// eslint-disable-next-line @typescript-eslint/require-await
+	public async init(args: ModuleInitArgs): Promise<void> {
+		const { moduleConfig } = args;
+		this._moduleConfig = (moduleConfig as unknown) as ModuleConfig;
```
shuse2, line 33: it would be better to validate the property than casting

framework/src/modules/validators/module.ts
```diff
@@ -0,0 +1,45 @@
...
+		if (this._moduleConfig.blockTime < 1) {
+			throw new Error('Block time cannot be less than 1');
+		}
+		this._blockTime = this._moduleConfig.blockTime;
```
shuse2, line 37: do we need to store both `moduleConfig` and `blockTime`?

framework/src/modules/validators/constants.ts
```diff
@@ -12,13 +12,9 @@
...
-import { ValidatorEndpoint } from './endpoint';
-
-export class ValidatorModule extends BaseModule {
-	public id = 10;
-	public name = 'validator';
-	public api = new ValidatorAPI(this.id);
-	public endpoint = new ValidatorEndpoint(this.id);
-}
+export const MODULE_ID_VALIDATORS = 11; // TBD
+export const STORE_PREFIX_VALIDATORS_DATA = Buffer.from('0000');
```
shuse2, line 16:
```suggestion
export const STORE_PREFIX_VALIDATORS_DATA = 0x0000;
```
it the implementation, we used `number` for prefix, so you could do like above. Otherwise, it needs to be converted again later

shuse2 (changes requested):
Unit test is failing

framework/src/modules/validators/module.ts
```diff
@@ -0,0 +1,43 @@
...
+	public api = new ValidatorsAPI(this.id);
+	public endpoint = new ValidatorsEndpoint(this.id);
+	private _blockTime!: number;
+
+	// eslint-disable-next-line @typescript-eslint/require-await
+	public async init(args: ModuleInitArgs): Promise<void> {
+		const { moduleConfig } = args;
+		if (((moduleConfig as unknown) as ModuleConfig).blockTime < 1) {
+			throw new Error('Block time cannot be less than 1');
+		}
```
shuse2, line 34: I think Instead of casting by expecting it exists, it would be better to validate like below?
```
const blockTime = moduleConfig.blockTime;
if (!blockTime || typeof blockTime !== 'string') {
	throw new Error('BlockTiem must be a number.');
}
if (blockTime < 1) {
	throw new Error('Block time cannot be less than 1');
}
```

shuse2 (changes requested):
minor point, otherwise LGTM

shuse2 (approved):

framework/src/modules/validators/module.ts
```diff
@@ -0,0 +1,47 @@
...
+	public endpoint = new ValidatorsEndpoint(this.id);
+	private _blockTime!: number;
+
+	// eslint-disable-next-line @typescript-eslint/require-await
+	public async init(args: ModuleInitArgs): Promise<void> {
+		const { moduleConfig } = args;
+		const { blockTime } = moduleConfig;
+
+		if (!blockTime || typeof blockTime !== 'number') {
+			throw new Error('BlockTiem must be a number.');
```
mehmetegemen, line 34: `BlockTiem` must be `BlockTime`.

framework/src/modules/validators/module.ts
```diff
@@ -0,0 +1,47 @@
...
+		}
+		if (blockTime < 1) {
+			throw new Error('Block time cannot be less than 1');
```
mehmetegemen, line 37: Dot is missing at the end of the error message.

mehmetegemen (changes requested):

mehmetegemen (approved):

PR #6760, opened 2021-09-10
Update endpoint to use `_` separator - Closes #6622 and #6666

### LINKED ISSUE #6622: Rename all the API endpoint names of actions and events to have underscore
### Description

Update separator to `"_"` from `“:”` because in JSON RPC, `method` should be using the characters allowed for method definition. Namespace is often separated with `_ (underscore)` in other projects such as `web3`.

Example, `app_getNodeStatus`, `app_getConnectedPeers`, `namespace_methodName`

### Acceptance Criteria

- All tests in the framework and framework-plugin should pass
- Modify regular expression for module and plugin name to not to have `_(underscore)`

### Additional Information

- Related #6554

### LINKED ISSUE #6666: Replace module/plugin lifecycle event with log
### Description

Replace below events with logs
```
channel.publish(`${pluginAlias}:registeredToBus`);
channel.publish(`${pluginAlias}:loading:started`);
...
```

### Motivation

Until module and plugin is initialized, RPC server are not enabled.
Also, there are no use-cases for those events in the framework as well

### Additional Information
- There are more events than above 2, and all lifecycle should be removed

shuse2:
### What was the problem?

This PR resolves #6622 and resolves #6666

### How was it solved?

- Replace `:` with `_`
- Replace `_plugin_:loadingFinished` with `loaded` `unloaded` and `unloadedWithError`.

### How was it tested?

- Updated unit tests
- Use test app to check if plugin load/unload with child process. (This needed several code changes on APIClient and plugin, but since there are lots of build issue, it required to change the built code)


framework/test/functional/rpc-api/ipc/ipc_client_in_memory.spec.ts
```diff
@@ -110,9 +110,9 @@ describe('api client ipc mode', () => {
 		});
 
 		it('should return a list of registered actions', async () => {
-			const actions = await await client.invoke('app:getRegisteredActions');
+			const actions = await await client.invoke('app_getRegisteredActions');
 			expect(actions).toBeArray();
-			expect(actions).toContain('app:getConnectedPeers');
+			expect(actions).toContain('app_getConnectedPeers');
 			expect(actions).toContain('dpos:getAllDelegates');
```
ishantiw, line 116: shouldn't we always have `namespace_<actionName>`?

ishantiw (approved):

mehmetegemen (approved):

PR #6764, opened 2021-09-10
 Implement validators endpoint - Closes #6671

### LINKED ISSUE #6671: Implement validators endpoint
### Description

Implement ValidatorsAPI in `endpoint.ts` as defined in https://research.lisk.com/t/introduce-validators-module/317.
```
getGeneratorList
validateBLSKey
```

### Acceptance Criteria

- All functions are implemented as per specification
- All functions must respond with JSON format
- All functions are fully unit tested

### Additional Information
Related issue #6668

Incede:
### What was the problem?

This PR resolves #6671 

### How was it solved?
* Add endpoints definitions as described in the [LIP](https://research.lisk.com/t/introduce-validators-module/317)

### How was it tested?

* Added unit tests


framework/src/modules/validators/endpoint.ts
```diff
@@ -12,6 +12,46 @@
...
+	STORE_PREFIX_BLS_KEYS,
+	STORE_PREFIX_GENERATOR_LIST,
+} from './constants';
 
-export class ValidatorsEndpoint extends BaseEndpoint {}
+export class ValidatorsEndpoint extends BaseEndpoint {
+	public async getGeneratorList(ctx: ModuleEndpointContext): Promise<Buffer[]> {
+		const subStore = ctx.getStore(MODULE_ID_VALIDATORS, STORE_PREFIX_GENERATOR_LIST);
+		const value = await subStore.getWithSchema<Record<string, unknown>>(
+			Buffer.from([0]),
```
shuse2, line 30: maybe it's better to define `emptyKey` and it should be `Buffer.alloc(0)`. 
`{ address: Buffer[] }` which should be defined or already defined in `types.ts`

framework/src/modules/validators/endpoint.ts
```diff
@@ -12,6 +12,46 @@
...
+import { generatorListSchema, validateBLSKeyRequest, validateBLSKeyRequestSchema } from './schemas';
+import {
+	MODULE_ID_VALIDATORS,
+	STORE_PREFIX_BLS_KEYS,
+	STORE_PREFIX_GENERATOR_LIST,
+} from './constants';
 
-export class ValidatorsEndpoint extends BaseEndpoint {}
+export class ValidatorsEndpoint extends BaseEndpoint {
+	public async getGeneratorList(ctx: ModuleEndpointContext): Promise<Buffer[]> {
```
shuse2, line 27: `Buffer[]` is not valid JSON compatible format. it should the response format should be object with only json compatible format like
```
{
  list: string[],
}
```

framework/src/modules/validators/endpoint.ts
```diff
@@ -12,6 +12,46 @@
...
+		if (reqErrors?.length) {
+			throw new LiskValidationError(reqErrors);
+		}
+
+		const req = (ctx.params as unknown) as validateBLSKeyRequest;
+		const { proofOfPosession, blsKey } = req;
+
+		const subStore = ctx.getStore(MODULE_ID_VALIDATORS, STORE_PREFIX_BLS_KEYS);
+
+		if (await subStore.get(blsKey)) {
```
shuse2, line 48: if the key does not exist, `get` will throw. it should use try-catch or `has`

framework/test/unit/modules/validators/endpoint.spec.ts
```diff
@@ -0,0 +1,55 @@
...
+describe('ValidatorsModuleEndpoint', () => {
+	const logger: Logger = fakeLogger;
+	let validatorsModule: ValidatorsModule;
+
+	beforeAll(() => {
+		validatorsModule = new ValidatorsModule();
+	});
+
+	describe('validateBLSKey', () => {
+		describe('when request data is invalid', () => {
```
shuse2, line 28: it should also have success case as well

shuse2 (changes requested):

framework/src/modules/validators/endpoint.ts
```diff
@@ -12,6 +12,46 @@
...
-export class ValidatorsEndpoint extends BaseEndpoint {}
+export class ValidatorsEndpoint extends BaseEndpoint {
+	public async getGeneratorList(ctx: ModuleEndpointContext): Promise<Buffer[]> {
+		const subStore = ctx.getStore(MODULE_ID_VALIDATORS, STORE_PREFIX_GENERATOR_LIST);
+		const value = await subStore.getWithSchema<Record<string, unknown>>(
+			Buffer.from([0]),
+			generatorListSchema,
+		);
+		const list = value.addresses as Buffer[];
+		return list;
```
ishantiw, line 34:
```suggestion
		return value.addresses as Buffer[];
```

framework/src/modules/validators/endpoint.ts
```diff
@@ -12,6 +12,46 @@
...
+		const subStore = ctx.getStore(MODULE_ID_VALIDATORS, STORE_PREFIX_BLS_KEYS);
+
+		if (await subStore.get(blsKey)) {
+			return false;
+		}
+		if (!blsPopVerify(blsKey, proofOfPosession)) {
+			return false;
+		}
+
+		return true;
```
ishantiw, line 55:
```suggestion
		return blsPopVerify(blsKey, proofOfPosession)
```

framework/src/modules/validators/schemas.ts
```diff
@@ -42,6 +46,8 @@ export const generatorListSchema = {
 };
 
 export const validatorAddressSchema = {
+	$id: '/validators/registeredBlsKeysSubStore',
+	title: 'Registered Validators Bls Keys',
```
ishantiw, line 50: this looks off, I think its ?
```suggestion
	$id: '/validators/address',
	title: 'Validators Address',
```

framework/src/modules/validators/endpoint.ts
```diff
@@ -12,6 +12,46 @@
...
+	}
+
+	public async validateBLSKey(ctx: ModuleEndpointContext): Promise<boolean> {
+		const reqErrors = validator.validate(validateBLSKeyRequestSchema, ctx.params);
+		if (reqErrors?.length) {
+			throw new LiskValidationError(reqErrors);
+		}
+
+		const req = (ctx.params as unknown) as validateBLSKeyRequest;
+		const { proofOfPosession, blsKey } = req;
```
ishantiw, line 44:
```suggestion
		const { proofOfPossession, blsKey } = req;
```

framework/src/modules/validators/endpoint.ts
ishantiw: please rename `proofOfPosession` -> `proofOfPossession` everywhere else

ishantiw (changes requested):

framework/src/modules/validators/endpoint.ts
ishantiw: please rename `proofOfPosession` -> `proofOfPossession` everywhere else

framework/src/modules/validators/endpoint.ts
Incede: changing to return 
```
{
  list: string[],
}
```
as @shuse2 suggested

framework/src/modules/validators/schemas.ts
Incede: yes, I've changed the `title` to `Validators Addresses` but kept the `$id` same to follow the substore name defined in the LIP.

framework/src/modules/validators/endpoint.ts
```diff
@@ -12,6 +12,55 @@
...
+		const subStore = ctx.getStore(MODULE_ID_VALIDATORS, STORE_PREFIX_GENERATOR_LIST);
+		const emptyKey = Buffer.alloc(0);
+		const value = await subStore.getWithSchema<GeneratorList>(emptyKey, generatorListSchema);
+
+		const addressList = value.addresses;
+
+		return { list: addressList.map(buf => buf.toString()) };
+	}
+
+	public async validateBLSKey(ctx: ModuleEndpointContext): Promise<boolean> {
```
shuse2, line 39:
```suggestion
	public async validateBLSKey(ctx: ModuleEndpointContext): Promise<{ valid: boolean}> {
```

shuse2 (commented):
other than missing test and the return type LGTM

framework/test/unit/modules/validators/endpoint.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+				await expect(
+					validatorsModule.endpoint.validateBLSKey({
+						getStore: getStore1,
+						logger,
+						params: {
+							proofOfPossession: proof,
+							blsKey: pk,
+						},
+					}),
+				).resolves.toStrictEqual({ valid: false });
```
shuse2, line 51: this should resolve `true` with the test case?
looking at logic I think there should be
- when key already exist
- when key does not exist, but invalid PoP
- when key does not exist, and PoP is valid

framework/test/unit/modules/validators/endpoint.spec.ts
```diff
@@ -0,0 +1,82 @@
...
+			});
+		});
+
+		describe('when request data is invalid', () => {
+			it('should reject with error', async () => {
```
shuse2, line 56:
```suggestion
			it('should reject with error with invalid request params', async () => {
```

framework/src/modules/validators/schemas.ts
```diff
@@ -62,3 +70,23 @@ export const genesisDataSchema = {
 	},
 	required: ['timestamp'],
 };
+
+export interface validateBLSKeyRequest {
```
shuse2, line 74:
```suggestion
export interface ValidateBLSKeyRequest {
```

framework/src/modules/validators/schemas.ts
```diff
@@ -62,3 +70,23 @@ export const genesisDataSchema = {
...
+	proofOfPossession: Buffer;
+	blsKey: Buffer;
```
shuse2, line 76: Endpoint request/response is always JSON format. (Buffer => `hex string`

shuse2 (changes requested):

framework/test/unit/modules/validators/endpoint.spec.ts
Incede: I think it should resolve `false` since the key already exists in the substore.
As per the LIP 
```
  if there exists an entry in the registered BLS keys substore with storeKey == blsKey:
        return False 
```

shuse2: ah yea that's correct. maybe test description should say that

ishantiw (approved):

shuse2 (approved):

PR #6765, opened 2021-09-13
Update block schema - Closes #6739

### LINKED ISSUE #6739: Update block schema and usages
### Description

1. Update current block header schema to be defined as defined in [LIP](https://github.com/LiskHQ/lips/blob/55ac3d48fd06ae54ee41b5ad9d7c90e7b01fd510/proposals/lip-update_block_schema_and_block_processing.md#block-header). Also, remove unused schemas and types

2. move current block header property `assets` should be moved to block property as `BlockAssets` and the class should have
* fromBytes
* fromJSON
* getBytes
* sort
* getAsset (same as the current implementation)
* setAsset (same as the current implementation)
Include BlockAssets as part of block.

3. Update storage and data access to store and get `assets` as part of `Block`. Block Assets should be stored as
```
[storageKey,blockID]: encodeBytesArray(assets),
```
where `encodeBytesArray` and `decodeBytesArray` should be defined as utils which encode/decode bytes array using schema like
```
arraySchema = {
  "type": "object",
  "required": ["list"],
  "properties": {
    "list": {
      "type": "array",
      "fieldNumber": 1,
      "items": {
        "dataType": "bytes"
      }
    },
  }
}
```

4. Update framework to use updated properties to pass build and test. (Mainly maxHeightPrevoted and maxHeightGenerated)

### Acceptance Criteria

- Build and unit test should pass
- Unit test should be added for BlockAssets
- Storage integration test on lisk-chain should be added

### Additional Information
Related ##6749

shuse2:
### What was the problem?

This PR resolves #6739

### How was it solved?

- Update block schema with new definition
- Update lisk-chain to store assets
- Update state machine context to have assets
- Update consensus to use block header property directly

### How was it tested?

- Update unit tests

### DIFF
--- a/elements/lisk-chain/src/block.ts
+++ b/elements/lisk-chain/src/block.ts
@@ -13,6 +13,7 @@
  */
 
 import { codec } from '@liskhq/lisk-codec';
+import { BlockAssets } from './block_assets';
 import { BlockHeader } from './block_header';
 import { blockSchema } from './schema';
 import { Transaction } from './transaction';
@@ -20,35 +21,45 @@ import { Transaction } from './transaction';
 interface BlockAttrs {
 	header: Buffer;
 	payload: Buffer[];
+	assets: Buffer[];
 }
 
 export class Block {
 	// eslint-disable-next-line no-useless-constructor
-	public constructor(public readonly header: BlockHeader, public readonly payload: Transaction[]) {
+	public constructor(
+		public readonly header: BlockHeader,
+		public readonly payload: Transaction[],
+		public readonly assets: BlockAssets,
+	) {
 		// No body necessary
 	}
 
 	public static fromBytes(value: Buffer): Block {
-		const { header, payload } = codec.decode<BlockAttrs>(blockSchema, value);
+		const { header, payload, assets } = codec.decode<BlockAttrs>(blockSchema, value);
 
 		return new Block(
 			BlockHeader.fromBytes(header),
 			payload.map(v => Transaction.fromBytes(v)),
+			BlockAssets.fromBytes(assets),
 		);
 	}
 
 	public static fromJSON(value: Record<string, unknown>): Block {
-		const { header, payload } = value;
+		const { header, payload, assets } = value;
 		if (typeof header !== 'object') {
 			throw new Error('Invalid block format. header must be an object.');
 		}
 		if (!Array.isArray(payload)) {
 			throw new Error('Invalid block format. payload must be an array.');
 		}
+		if (!Array.isArray(assets)) {
+			throw new Error('Invalid block format. assets must be an array.');
+		}
 
 		return new Block(
 			BlockHeader.fromJSON(value.header as Record<string, unknown>),
 			payload.map(v => Transaction.fromBytes(v)),
+			BlockAssets.fromJSON(assets),
 		);
 	}
 
@@ -56,6 +67,7 @@ export class Block {
 		return codec.encode(blockSchema, {
 			header: this.header.getBytes(),
 			payload: this.payload.map(p => p.getBytes()),
+			assets: this.assets.getBytes(),
 		});
 	}
 
--- a/elements/lisk-chain/src/block_assets.ts
+++ b/elements/lisk-chain/src/block_assets.ts
@@ -1,0 +1,61 @@
+/*
+ * Copyright © 2021 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+import { codec } from '@liskhq/lisk-codec';
+import { blockAssetSchema } from './schema';
+
+export interface BlockAsset {
+	moduleID: number;
+	data: Buffer;
+}
+
+export class BlockAssets {
+	private readonly _assets: BlockAsset[] = [];
+
+	public constructor(assets: BlockAsset[] = []) {
+		this._assets = assets;
+	}
+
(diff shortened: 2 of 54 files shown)


mitsuaki-u (changes requested):

mitsuaki-u (approved):

elements/lisk-chain/src/schema.ts
```diff
@@ -59,30 +77,50 @@ export const signingBlockHeaderSchema = {
...
+	$id: '/block/header/3',
 	required: [...blockHeaderSchema.required, 'id'],
 	properties: {
 		...blockHeaderSchema.properties,
-		id: { dataType: 'bytes', fieldNumber: 10 },
+		id: { dataType: 'bytes', fieldNumber: 14 },
+	},
+};
+
+export const blockAssetSchema = {
```
Incede, line 111: I think this would be more appropriate if we follow the naming conventions in the LIP. 
The LIP defines new block property `assets`, and since we have `blockHeaderSchema` for the property `header`, hence `blockAssetsSchema` would be inline.
```suggestion
export const blockAssetsSchema = {
```

elements/lisk-chain/test/unit/block_assets.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+
+		describe('sort', () => {
+			it('should sort the assets in ascending order by module ID', () => {
+				assets.sort();
+				expect(assets['_assets'][0].moduleID).toEqual(3);
+			});
+		});
+
+		describe('getAsset', () => {
+			it('it should return undefined if no matching asset exsit ', () => {
```
Incede, line 42:
```suggestion
			it('it should return undefined if no matching asset exists ', () => {
```

elements/lisk-chain/test/unit/block_assets.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+				expect(assets.getAsset(5)).toBeUndefined();
+			});
+
+			it('it should return asset data if matching asset exsit ', () => {
```
Incede, line 46:
```suggestion
			it('it should return asset data if matching asset exists ', () => {
```

elements/lisk-chain/test/unit/block_assets.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+				expect(assets.getAsset(3)).toBeInstanceOf(Buffer);
+			});
+		});
+
+		describe('setAsset', () => {
+			it('it not should overwrite existing asset', () => {
```
Incede, line 52:
```suggestion
			it('it should not overwrite existing asset', () => {
```

elements/lisk-chain/test/unit/block_assets.spec.ts
```diff
@@ -0,0 +1,77 @@
...
+				const data = getRandomBytes(32);
+				expect(() => assets.setAsset(3, data)).toThrow();
+			});
+
+			it('it should add asset data if matching asset does not exsit ', () => {
```
Incede, line 57:
```suggestion
			it('it should add asset data if matching asset does not exist ', () => {
```

elements/lisk-chain/test/unit/block_header.spec.ts
```diff
@@ -104,9 +114,9 @@ describe('block_header', () => {
 				expect(blockHeader.generatorAddress).toEqual(data.generatorAddress);
 				expect(blockHeader.previousBlockID).toEqual(data.previousBlockID);
 				expect(blockHeader.stateRoot).toEqual(data.stateRoot);
+				expect(blockHeader.validatorsHash).toEqual(data.validatorsHash);
+				expect(blockHeader.aggregateCommit).toEqual(data.aggregateCommit);
```
Incede, line 118: maybe expect to match other header properties as well
```suggestion
				expect(blockHeader.validatorsHash).toEqual(data.validatorsHash);
				expect(blockHeader.aggregateCommit).toEqual(data.aggregateCommit);
				expect(blockHeader.maxHeightPrevoted).toEqual(data.maxHeightPrevoted);
				expect(blockHeader.maxHeightGenerated).toEqual(data.maxHeightGenerated);
				expect(blockHeader.assetsRoot).toEqual(data.assetsRoot);
```

elements/lisk-chain/src/block.ts
```diff
@@ -13,49 +13,61 @@
...
+	assets: Buffer[];
 }
 
 export class Block {
 	// eslint-disable-next-line no-useless-constructor
-	public constructor(public readonly header: BlockHeader, public readonly payload: Transaction[]) {
+	public constructor(
+		public readonly header: BlockHeader,
+		public readonly payload: Transaction[],
+		public readonly assets: BlockAssets,
```
Incede, line 32: A bit unsure here if it can be redefined to be something like `BlockAsset[]`. In the LIP it says `assets` is an array of objects and also since payload is an array of objects too.

elements/lisk-chain/src/block_assets.ts
```diff
@@ -0,0 +1,61 @@
...
+
+import { codec } from '@liskhq/lisk-codec';
+import { blockAssetSchema } from './schema';
+
+export interface BlockAsset {
+	moduleID: number;
+	data: Buffer;
+}
+
+export class BlockAssets {
```
Incede, line 23: It's more of a question, can we drop "block" and just keep it as `Assets` ? 
thinking to align naming with other block property `payload` & `Transaction`

Incede (changes requested):

elements/lisk-chain/src/block.ts
shuse2: other properties are also class instance, and we could create `BlockAsset` class, but we want to have `blockAssets.get` and `blockAssets.set`. Therefore it's better to be `BlockAssets` class

elements/lisk-chain/src/schema.ts
shuse2: asset is not part of the block header, and we need to decode asset one by one. Notice block schema has byte array for `assets`

Incede (approved):

PR #6770, opened 2021-09-14
Bootstrap auth module with initials - Closes #6732

### LINKED ISSUE #6732: Bootstrap auth module with class definition, schema and constants
### Description

- Create auth module class
- Define constants in `constants.ts` and schema in `schema.ts`
- Define configurations that are received from the `init` hook
- Declare hooks `verifyTransaction` and `afterTransactionExecution`

### Acceptance Criteria

- All files and class exist in the auth module folder

### Additional Information

Related to #6731

mehmetegemen:
### What was the problem?

This PR resolves #6732 

### How was it solved?

- Added necessary initial files for auth module

### How was it tested?

- Manually checked for files' existence

### DIFF
--- a/framework/src/modules/auth/api.ts
+++ b/framework/src/modules/auth/api.ts
@@ -1,0 +1,17 @@
+/*
+ * Copyright © 2021 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+import { BaseAPI } from '..';
+
+export class AuthAPI extends BaseAPI {}
--- a/framework/src/modules/auth/api.ts
+++ b/framework/src/modules/auth/api.ts
@@ -12,6 +12,6 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
-import { BaseAPI } from '..';
+import { BaseAPI } from '../base_api';
 
 export class AuthAPI extends BaseAPI {}
--- a/framework/src/modules/auth/endpoint.ts
+++ b/framework/src/modules/auth/endpoint.ts
@@ -12,6 +12,6 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
-import { BaseEndpoint } from '..';
+import { BaseEndpoint } from '../base_endpoint';
 
 export class AuthEndpoint extends BaseEndpoint {}
--- a/framework/src/modules/auth/endpoint.ts
+++ b/framework/src/modules/auth/endpoint.ts
@@ -1,0 +1,17 @@
+/*
+ * Copyright © 2021 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+import { BaseEndpoint } from '..';
+
+export class AuthEndpoint extends BaseEndpoint {}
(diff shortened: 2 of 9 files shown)


framework/src/modules/auth/endpoint.ts
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+import { BaseEndpoint } from '..';
```
shuse2, line 15:
```suggestion
import { BaseEndpoint } from '../base_endpoint';
```

framework/src/modules/auth/module.ts
```diff
@@ -0,0 +1,48 @@
...
+import { AuthEndpoint } from './endpoint';
+import { configSchema } from './schemas';
+
+export class AuthModule extends BaseModule {
+	public id = MODULE_ID_AUTH;
+	public name = 'auth';
+	public api = new AuthAPI(this.id);
+	public endpoint = new AuthEndpoint(this.id);
+	public configSchema = configSchema;
+	public commands = [new RegisterMultisignitureCommand()];
```
shuse2, line 34: probably it needs to have `moduleID` as input for each command

framework/src/modules/auth/schemas.ts
```diff
@@ -0,0 +1,74 @@
...
+			items: {
+				dataType: 'bytes',
+			},
+			fieldNumber: 4,
+		},
+	},
+	required: ['nonce', 'numberOfSignatures', 'mandatoryKeys', 'optionalKeys'],
+};
+
+export const registerMultisignatureCommandSchema = {
```
shuse2, line 44:
```suggestion
export const registerMultisignatureParamsSchema = {
```

framework/src/modules/auth/schemas.ts
```diff
@@ -0,0 +1,74 @@
...
+	$id: '/auth/command/regMultisig',
+	type: 'object',
+	properties: {
+		numberOfSignatures: {
+			dataType: 'uint32',
+			fieldNumber: 1,
```
shuse2, line 50: it's better to have max,min as in https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/modules/keys/schemas.ts#L20-L47

shuse2 (changes requested):

framework/src/modules/auth/module.ts
mehmetegemen: `BaseCommand` has no `moduleID` param in its constructor like for example `BaseAPI` does, should I still put it there without a convention in the `BaseCommand` constructor?

shuse2: It's ok if we add it later, but to access the storage, it will require the moduleID. (we can consider adding to base command later if needed

mehmetegemen: then let's add it now to set an example for other modules and `BaseCommand`? Adding it now without `BaseCommand` changes wouldn't hurt as you say.

shuse2: yea sounds good!

mehmetegemen: Actually it hurts 😄 We cannot pass the `moduleID` without adding it to constructor as a parameter.

mehmetegemen: It gives the error `Expected 0 arguments but got 1`.

shuse2: yes, we need to add the constructor if it's not same as base command

framework/src/modules/auth/api.ts
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+import { BaseAPI } from '..';
```
shuse2, line 15:
```suggestion
import { BaseAPI } from '../base_api';
```

framework/src/modules/auth/module.ts
mehmetegemen: Yes, we need to add constructor because if I don't pass the `moduleID` to the command we cannot make it a working command and can't proceed with other issues of the module epic. Adding the `moduleID` and constructor asap.

mitsuaki-u (approved):
minor comment LGTM besides @shuse2 comment

shuse2 (approved):

PR #6784, opened 2021-09-17
Implement afterBlockExecute hook for the reward module - Closes #6690

### LINKED ISSUE #6690: Implement afterBlockExecute hook of Rewards Module
### Description
Implement `afterBlockExecute` hook as specified in https://research.lisk.com/t/define-state-transitions-of-reward-module/316#after-block-execution-22

### Acceptance criteria
- Possible scenarios are tested in the unit test(e.g. BFT violation, seedReveal violation etc.)

### Additional Information
- `TOKEN_ID_REWARD` will be specified in config parameter of init method of the reward module class.

### Related Issue
#6686

mehmetegemen:
### What was the problem?

This PR resolves #6690 

### How was it solved?

- Implemented afterBlockExecute hook in the reward module class

### How was it tested?

- It was tested with unit tests

### DIFF
--- a/framework/src/modules/reward/module.ts
+++ b/framework/src/modules/reward/module.ts
@@ -70,7 +70,7 @@ export class RewardModule extends BaseModule {
 		);
 
 		if (blockReward <= BigInt(0)) {
-			throw new Error('Reward amount to be minted is not valid.');
+			return;
 		}
 
 		await this._tokenAPI.mint(
--- a/framework/src/modules/reward/module.ts
+++ b/framework/src/modules/reward/module.ts
@@ -36,7 +36,7 @@ export class RewardModule extends BaseModule {
 		this._tokenAPI = tokenAPI;
 		this._randomAPI = randomAPI;
 		this._bftAPI = bftAPI;
-		this.api.addDependencies(bftAPI, randomAPI);
+		this.api.addDependencies(this._bftAPI, this._randomAPI);
 	}
 
 	// eslint-disable-next-line @typescript-eslint/require-await
--- a/framework/src/modules/reward/module.ts
+++ b/framework/src/modules/reward/module.ts
@@ -62,9 +62,7 @@ export class RewardModule extends BaseModule {
 		});
 	}
 
-	// eslint-disable-next-line @typescript-eslint/require-await
 	public async afterBlockExecute(context: BlockAfterExecuteContext): Promise<void> {
-		// eslint-disable-next-line no-console
 		const blockReward = await this.api.getBlockReward(
 			context.getAPIContext(),
 			context.header,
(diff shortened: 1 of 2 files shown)


framework/src/modules/reward/module.ts
```diff
@@ -62,9 +62,22 @@ export class RewardModule extends BaseModule {
...
-		// eslint-disable-next-line no-console
-		console.log(this._tokenAPI, this._bftAPI, this._randomAPI, this._tokenIDReward);
+	public async afterBlockExecute(context: BlockAfterExecuteContext): Promise<void> {
+		const blockReward = await this.api.getBlockReward(
+			context.getAPIContext(),
+			context.header,
+			context.assets,
+		);
+
+		if (blockReward <= BigInt(0)) {
```
shuse2, line 72: `0` reward should be valid, and it is calculating by itself, so no validation is needed?

framework/test/unit/modules/reward/reward_module.spec.ts
```diff
@@ -46,4 +48,39 @@ describe('RewardModule', () => {
...
 	});
+
+	const { brackets, offset, distance } = moduleConfig as {
+		brackets: ReadonlyArray<bigint>;
+		offset: number;
+		distance: number;
+	};
+
+	describe.each(Object.entries(brackets))(
+		'afterBlockExecute test for brackets',
```
shuse2, line 59: i don't think it's necessary to test all the brackets assuming the util is tested well

shuse2 (changes requested):

framework/src/modules/reward/module.ts
mehmetegemen: I thought minting 0 reward is useless and redundant, don't you think so? Because if I don't end the control flow there the API will mint 0 reward.

framework/test/unit/modules/reward/reward_module.spec.ts
mehmetegemen: Then you think I should test the just first bracket in a single test?

framework/src/modules/reward/module.ts
mehmetegemen: On the other hand, 0 reward is valid and not minting is `mint` functions responsibility, so probably I should remove this conditional. Still waiting for your approval.

framework/test/unit/modules/reward/reward_module.spec.ts
shuse2: yes, i think just first bracket after offset should be good

framework/src/modules/reward/module.ts
shuse2: yes, then it should be if `blockReward <= 0` return. if thrown error, block will be rejected

mehmetegemen: done

framework/test/unit/modules/reward/reward_module.spec.ts
mehmetegemen: done

shuse2 (approved):

Incede (approved):

PR #6790, opened 2021-09-20
Add BFT API - Closes #6744

### LINKED ISSUE #6744: Add API for BFT module
### Description

Implement BFTAPI in `api.ts` as defined in https://research.lisk.com/t/introduce-bft-module/321.
```
areHeadersContradicting
isHeaderContradictingChain
existBFTParameters
getBFTParameters
getBFTHeights
impliesMaximalPrevotes
getNextHeightBFTParameters
setBFTParameters

```

If API requires access to the state, the first parameter must be `apiContext: APIContext`; otherwise, the input should follow the specification.

### Acceptance Criteria

- All functions are implemented as specification
- All functions are fully unit tested

### Additional Information

Related to #6742

shuse2:
### What was the problem?

This PR resolves #6744 

### How was it solved?

- Add BFT API implementation and tests
- Fix typos

### How was it tested?

Added unit tests


framework/src/modules/bft/api.ts
```diff
@@ -12,45 +12,253 @@
...
+	}
+
+	public async getBFTHeights(context: ImmutableAPIContext): Promise<BFTHeights> {
+		const votesStore = context.getStore(this.moduleID, STORE_PREFIX_BFT_VOTES);
+		const bftVotes = await votesStore.getWithSchema<BFTVotes>(EMPTY_KEY, bftVotesSchema);
 		return {
-			maxHeightPrecommited: 0,
-			maxHeightPrevoted: 0,
+			maxHeightPrevoted: bftVotes.maxHeightPrevoted,
+			maxHeightPrecommited: bftVotes.maxHeightPrecommited,
```
shuse2, line 99: `maxHeightPrecommited` => `maxHeightPrecommitted` is done in next PR https://github.com/LiskHQ/lisk-sdk/pull/6796

mitsuaki-u (changes requested):
LGTM minor comments

mitsuaki-u (approved):

framework/src/modules/bft/api.ts
```diff
@@ -12,45 +12,253 @@
...
+	): Promise<void> {
+		let aggregateBFTWeight = BigInt(0);
+		for (const validator of validators) {
+			if (validator.bftWeight <= 0) {
+				throw new Error('Invalid BFT weight. BFT weight must be a positive integer.');
+			}
+			aggregateBFTWeight += validator.bftWeight;
+		}
+		if (
+			aggregateBFTWeight / BigInt(3) + BigInt(1) > precommitThreshold ||
```
janhack, line 167: Just to confirm, we do integer division here so the left side is equivalent to
`floor(1/3*aggregateBFTWeight)+1`, right?

framework/src/modules/bft/api.ts
```diff
@@ -12,45 +12,253 @@
...
+		return result.key.readUInt32BE(0);
+	}
+
+	public async setBFTParameters(
+		context: APIContext,
+		precommitThreshold: bigint,
+		certificateThreshold: bigint,
+		validators: Validator[],
+	): Promise<void> {
+		let aggregateBFTWeight = BigInt(0);
```
janhack, line 159: The following check is currently missing:
```
if length(validators) > LSK_BFT_BATCH_SIZE:
        fail
```

framework/src/modules/bft/api.ts
```diff
@@ -12,45 +12,253 @@
...
+				minActiveHeight: nextHeight,
+				largestHeightPrecommit: nextHeight - 1,
+			});
+		}
+		sortValidatorsByAddress(nextActiveValidators);
+		bftVotes.activeValidatorsVoteInfo = nextActiveValidators;
+		await votesStore.setWithSchema(EMPTY_KEY, bftVotes, bftVotesSchema);
+	}
+
+	public async getValidators(context: ImmutableAPIContext): Promise<Validator[]> {
```
janhack, line 225: Currently, this API is not contained in the LIP. Where is it needed? Couldn't `getBFTParameters` be used instead?

janhack (changes requested):
I added a few comments, mostly for clarification.

framework/src/modules/bft/api.ts
shuse2: it is currently used in the fast sync logic where it needs current validator.
getBFTParameters with last block height should be give the latest one. Ill create another issue to refactor since it will affect the test a lot

framework/src/modules/bft/api.ts
shuse2: yes, int calculation is `floor` as default so left one is equivalent to the `floor(1/3*aggregateBFTWeight)+1`

shuse2:
```
> BigInt(103) / BigInt(3) + BigInt(1)
35n
```

janhack: All good then :-)

framework/src/modules/bft/api.ts
```diff
@@ -156,6 +161,11 @@ export class BFTAPI extends BaseAPI {
 		certificateThreshold: bigint,
 		validators: Validator[],
 	): Promise<void> {
+		if (validators.length > this._batchSize) {
+			throw new Error(
+				`Invalid validators size. Validators must be greater than batch size ${this._batchSize}.`,
```
janhack, line 166: I think the error message is incorrect (you want validators.length <= this._batchSize). It could be instead:

> `Invalid number of validators. The number of validators can be at most the batch size ${this._batchSize}.`,

framework/src/modules/bft/api.ts
janhack: If you think it is useful, we can add the API to the LIP. Or you prefer to refactor?

shuse2: I think it would be good to add to the API in the LIP as well as current validators could be useful

janhack: Okay, I will add the API then. Should we call it maybe `getCurrentValidators` to emphasize it is the current validator information? 

Note that when this API is called for the last block of the round at height `h` and the hook `afterBlockExecute` of the BFT module is already called, then `getCurrentValidators` will return the validators active at round `h+1` (next round). But I think this is fine and stills makes sense.

janhack (approved):
Looks all good.

Minor comment/suggestion: we could rename the new API `getValidators` to `getCurrentValidators` or something similar, see related comment.

framework/src/modules/bft/api.ts
shuse2: `getCurrentValidators` sounds good, ill update everywhere!
the return value depends on the APIContext used as well. during the block execution, it will use the same context but in synchronisation, it will use different context so it will always look at the information saved to the DB

shuse2:
`getCurrentValidators` logic should be updated with #6809

PR #6791, opened 2021-09-20
Implement auth module API - Closes #6737

### LINKED ISSUE #6737: Add API for auth module
#### getAuthAccount
This function is used to retrieve information about an auth account.
##### Parameters
`address`: A 20 bytes value identifying the auth account.
##### Returns
This function returns `authAccount(address)`. If there is no entry corresponding to `address`, it returns an empty object.

### Acceptance Criteria

- Add unit test for this getAuthAccount API

### Additional Information

Related to #6731

mitsuaki-u:
### What was the problem?

This PR resolves #6737

### How was it solved?

Added API for auth module as described in #6737 

### How was it tested?

- Added unit tests

### DIFF
--- a/framework/src/modules/auth/api.ts
+++ b/framework/src/modules/auth/api.ts
@@ -12,7 +12,7 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
-import { NotFoundError } from '@liskhq/lisk-db';
+import { NotFoundError } from '@liskhq/lisk-chain';
 import { BaseAPI } from '../base_api';
 import { ImmutableAPIContext } from '../../node/state_machine';
 import { AuthData, authAccountSchema } from './schemas';
@@ -29,11 +29,11 @@ export class AuthAPI extends BaseAPI {
 
 			return authData;
 		} catch (error) {
-			if (error instanceof NotFoundError) {
-				return { nonce: BigInt(0), numberOfSignatures: 0, mandatoryKeys: [], optionalKeys: [] };
+			if (!(error instanceof NotFoundError)) {
+				throw error;
 			}
 
-			throw error;
+			return { nonce: BigInt(0), numberOfSignatures: 0, mandatoryKeys: [], optionalKeys: [] };
 		}
 	}
 }
--- a/framework/src/modules/auth/api.ts
+++ b/framework/src/modules/auth/api.ts
@@ -12,6 +12,7 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
+import { NotFoundError } from '@liskhq/lisk-db';
 import { BaseAPI } from '../base_api';
 import { ImmutableAPIContext } from '../../node/state_machine';
 import { AuthData, authAccountSchema } from './schemas';
@@ -23,13 +24,16 @@ export class AuthAPI extends BaseAPI {
 		address: Buffer,
 	): Promise<AuthData | {}> {
 		const authDataStore = apiContext.getStore(this.moduleID, STORE_PREFIX_AUTH);
-		let authData;
 		try {
-			authData = await authDataStore.getWithSchema<AuthData>(address, authAccountSchema);
-		} catch (error) {
-			authData = {};
-		}
+			const authData = await authDataStore.getWithSchema<AuthData>(address, authAccountSchema);
 
-		return authData;
+			return authData;
+		} catch (error) {
+			if (error instanceof NotFoundError) {
+				return { nonce: BigInt(0), numberOfSignatures: 0, mandatoryKeys: [], optionalKeys: [] };
+			}
+
+			throw error;
+		}
 	}
 }
--- a/framework/src/modules/auth/api.ts
+++ b/framework/src/modules/auth/api.ts
@@ -19,10 +19,7 @@ import { AuthData, authAccountSchema } from './schemas';
 import { STORE_PREFIX_AUTH } from './constants';
 
 export class AuthAPI extends BaseAPI {
-	public async getAuthAccount(
-		apiContext: ImmutableAPIContext,
-		address: Buffer,
-	): Promise<AuthData | {}> {
+	public async getAuthAccount(apiContext: ImmutableAPIContext, address: Buffer): Promise<AuthData> {
 		const authDataStore = apiContext.getStore(this.moduleID, STORE_PREFIX_AUTH);
 		try {
 			const authData = await authDataStore.getWithSchema<AuthData>(address, authAccountSchema);
--- a/framework/src/modules/auth/api.ts
+++ b/framework/src/modules/auth/api.ts
@@ -13,5 +13,23 @@
  */
 
 import { BaseAPI } from '../base_api';
+import { ImmutableAPIContext } from '../../node/state_machine';
+import { AuthData, authAccountSchema } from './schemas';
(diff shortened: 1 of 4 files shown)


framework/src/modules/auth/api.ts
```diff
@@ -13,5 +13,23 @@
...
+export class AuthAPI extends BaseAPI {
+	public async getAuthAccount(
+		apiContext: ImmutableAPIContext,
+		address: Buffer,
+	): Promise<AuthData | {}> {
+		const authDataStore = apiContext.getStore(this.moduleID, STORE_PREFIX_AUTH);
+		let authData;
+		try {
+			authData = await authDataStore.getWithSchema<AuthData>(address, authAccountSchema);
+		} catch (error) {
```
shuse2, line 29: it should check if `error` is NotFound error, and if not it should throw error

framework/src/modules/auth/api.ts
```diff
@@ -13,5 +13,23 @@
...
+			authData = {};
```
shuse2, line 30: Probably better to return zero value, instead of empty object.
```
return {
  nonce: BigInt(0),
  numberOfSignatures: 0,
  mandatoryKeys: [],
  optionalKeys: [],
}
```

framework/src/modules/auth/api.ts
```diff
@@ -13,5 +13,23 @@
...
-export class AuthAPI extends BaseAPI {}
+export class AuthAPI extends BaseAPI {
+	public async getAuthAccount(
+		apiContext: ImmutableAPIContext,
+		address: Buffer,
+	): Promise<AuthData | {}> {
+		const authDataStore = apiContext.getStore(this.moduleID, STORE_PREFIX_AUTH);
+		let authData;
+		try {
+			authData = await authDataStore.getWithSchema<AuthData>(address, authAccountSchema);
```
shuse2, line 28:
```suggestion
			const authData = await authDataStore.getWithSchema<AuthData>(address, authAccountSchema);
			return authData;
```

framework/test/unit/modules/auth/api.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+
+describe('AuthAPI', () => {
+	let authAPI: AuthAPI;
+	let context: any;
+	let authStore: any;
+	const address = Buffer.from('fa1c00809ff1b10cd269a711eef40a465ba4a9cb', 'hex');
+
+	beforeEach(async () => {
+		authAPI = new AuthAPI(1);
+		context = createTransientAPIContext({});
```
shuse2, line 28: do you need empty object here?

framework/test/unit/modules/auth/api.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+ */
+
+import { AuthAPI } from '../../../../src/modules/auth/api';
+import { STORE_PREFIX_AUTH } from '../../../../src/modules/auth/constants';
+import { authAccountSchema } from '../../../../src/modules/auth/schemas';
+import { createTransientAPIContext } from '../../../../src/testing';
+
+describe('AuthAPI', () => {
+	let authAPI: AuthAPI;
+	let context: any;
```
shuse2, line 22:
```suggestion
	let context: APIContext;
```

framework/test/unit/modules/auth/api.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+	let authStore: any;
```
shuse2, line 23:
```suggestion
	let authStore: StateStore;
```

framework/test/unit/modules/auth/api.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+			const authData = await authAPI.getAuthAccount(context, address);
+
+			expect(authData).toHaveProperty('nonce', BigInt(1));
+			expect(authData).toHaveProperty('numberOfSignatures', 1);
+			expect(authData).toHaveProperty('mandatoryKeys', []);
+			expect(authData).toHaveProperty('optionalKeys', []);
+		});
+
+		it('should return empty object given address not in db', async () => {
+			const authData = await authAPI.getAuthAccount(context, Buffer.from('invalid', 'utf8'));
```
shuse2, line 53:
```suggestion
			const authData = await authAPI.getAuthAccount(context, getRandomBytes(20));
```

framework/test/unit/modules/auth/api.spec.ts
```diff
@@ -0,0 +1,58 @@
...
+	beforeEach(async () => {
+		authAPI = new AuthAPI(1);
+		context = createTransientAPIContext({});
+		authStore = context.getStore(authAPI['moduleID'], STORE_PREFIX_AUTH);
+		await authStore.setWithSchema(
+			address,
+			{
+				nonce: 1,
+				numberOfSignatures: 1,
+				mandatoryKeys: [],
```
shuse2, line 35:
```suggestion
				mandatoryKeys: [getRandomBytes(64), getRandomBytes(64)],
```
I think it would be better to have something to make sure it is properly obtained. same for `optionalKeys`

framework/src/modules/auth/api.ts
mitsuaki-u: I was following the LIP which says: `This function returns authAccount(address). If there is no entry corresponding to address, it returns an empty object.`

framework/test/unit/modules/auth/api.spec.ts
mitsuaki-u: yes

framework/src/modules/auth/schemas.ts
```diff
@@ -12,7 +12,15 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
+export interface AuthData {
+	nonce: number;
```
mehmetegemen, line 16: `nonce`'s type is `uint64` in the schema, shouldn't it be a `bigint`? It's also `bigint` in the old `sequence` module.

framework/test/unit/modules/auth/api.spec.ts
```diff
@@ -0,0 +1,61 @@
...
+	const address = Buffer.from('fa1c00809ff1b10cd269a711eef40a465ba4a9cb', 'hex');
+	const expectedAuthData = {
+		nonce: 1,
+		numberOfSignatures: 1,
+		mandatoryKeys: [getRandomBytes(64), getRandomBytes(64)],
+		optionalKeys: [getRandomBytes(64), getRandomBytes(64)],
+	};
+
+	beforeEach(async () => {
+		authAPI = new AuthAPI(1);
```
mehmetegemen, line 36: Why don't we use `MODULE_ID_AUTH` from constants here?

mehmetegemen (changes requested):

shuse2 (approved):
Other than comments from @mehmetegemen , LGTM

framework/src/modules/auth/schemas.ts
```diff
@@ -12,7 +12,15 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
+export interface AuthData {
```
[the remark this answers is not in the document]
mitsuaki-u: In the LIP it is called `AuthData`.

shuse2 (approved):

PR #6796, opened 2021-09-22
Add BFT afterBlockExecute - Closes #6746

### LINKED ISSUE #6746: Implement BFT module afterBlockExecution hook
### Description

Implement `afterBlockExecution` as defined in https://research.lisk.com/t/introduce-bft-module/321.

### Acceptance Criteria

- Possible scenarios are tested in the unit test
- Reuse protocol spec for the previous BFT for the test

### Additional Information
Related issue #6742

shuse2:
### What was the problem?

This PR resolves #6746 

### How was it solved?

- Add all afterBlockExecute hook logic
- Recover protocol spec test from lisk-bft
- Add unit tests

### How was it tested?

- Recover protocol spec from lisk-bft
- Add unit tests


framework/src/modules/bft/bft_votes.ts
```diff
@@ -0,0 +1,173 @@
...
+export const updateMaxHeightCertified = (bftVotes: BFTVotes, header: BlockHeader): void => {
+	if (
+		header.aggregateCommit.height === 0 &&
+		header.aggregateCommit.aggregationBits.length === 0 &&
+		header.aggregateCommit.certificateSignature.length === 0
+	) {
+		return;
+	}
+	// eslint-disable-next-line no-param-reassign
+	bftVotes.maxHeightCertified = header.height;
```
janhack, line 172:
```suggestion
	bftVotes.maxHeightCertified = header.aggregateCommit.height;
```

janhack: It should be set to the height property in the aggregate commit.

framework/src/modules/bft/bft_votes.ts
```diff
@@ -0,0 +1,173 @@
...
+	}
+	const [newBlockBFTInfo] = bftVotes.blockBFTInfos;
+	// a block header only implies votes if maxHeightGenerated < height
+	if (newBlockBFTInfo.maxHeightGenerated >= newBlockBFTInfo.height) {
+		return;
+	}
+	const validatorInfo = bftVotes.activeValidatorsVoteInfo.find(v =>
+		v.address.equals(newBlockBFTInfo.generatorAddress),
+	);
+	if (!validatorInfo) {
```
janhack, line 68: If the validator is not contained in the active validators, we can just return here. This will happen, for example, for standby delegates which have BFT weight 0 and are therefore not contained in the active validators (actually, the API `setBFTParameters` only allows positive BFT weights). 

I missed this point in the BFT module LIP and will also update it there.

framework/src/modules/bft/bft_votes.ts
```diff
@@ -0,0 +1,173 @@
...
+	const heightNotPrevoted = getHeightNotPrevoted(bftVotes);
+
+	const minPrecommitHeight = Math.max(
+		validatorInfo.minActiveHeight,
+		heightNotPrevoted + 1,
+		validatorInfo.largestHeightPrecommit + 1,
+	);
+	let hasPrecommitted = false;
+
+	for (const blockBFTInfo of bftVotes.blockBFTInfos) {
```
janhack, line 80: Just to confirm as the order of the iteration is important here: The order is always `bftVotes.blockBFTInfos[0], bftVotes.blockBFTInfos[1], ...` here, right?

framework/src/modules/bft/bft_votes.ts
```diff
@@ -0,0 +1,173 @@
...
+		if (blockBFTInfo.height < minPrecommitHeight) {
+			break;
+		}
+		const params = await paramsCache.getParameters(blockBFTInfo.height);
+		if (blockBFTInfo.prevoteWeight >= params.prevoteThreshold) {
+			const bftValidator = params.validators.find(v =>
+				v.address.equals(blockBFTInfo.generatorAddress),
+			);
+			if (!bftValidator) {
```
janhack, line 89: Also here, if there is no entry, we should assume BFT weight 0 (maybe a previous active delegate was standby delegate in a previous round).

framework/src/modules/bft/bft_votes.ts
shuse2: yes, we always insert the latest block in [0] index and this loop is same as saying `for (let i = 0; i < bftVotes.blockBFTInfos.length; i++)`

framework/src/modules/bft/bft_params.ts
```diff
@@ -0,0 +1,75 @@
...
+	const results = await paramsStore.iterate({
+		start,
+		end,
+	});
+	for (const result of results) {
+		await paramsStore.del(result.key);
+	}
+};
+
+export class BFTParametersCache {
```
janhack, line 57: I think this cache is a good idea to avoid many reads. As far as I understand though, I have a separate cached value for every height. This means if in the Lisk Mainchain I iterate over 309 blocks (for instance in `updateMaxHeightPrevoted`), then I would call the the catch for 309 not yet cached values which result in 309 calls to `getBFTParameters`. This results in 309 iterations on the state store.

However, the state store only stores one key-value pair per round so with one of these iterations I could in theory add 103 items to the cache. This would reduce the number of calls to the state store to 3-4 (4 as rounds 309 blocks may cover 4 rounds).

framework/src/modules/bft/module.ts
```diff
@@ -49,6 +60,18 @@ export class BFTModule extends BaseModule {
...
+		const paramsCache = new BFTParametersCache(paramsStore);
+		const bftVotes = await votesStore.getWithSchema<BFTVotes>(EMPTY_KEY, bftVotesSchema);
+
+		insertBlockBFTInfo(bftVotes, context.header, this._maxLengthBlockBFTInfos);
+		await updatePrevotesPrecommits(bftVotes, paramsCache);
+		await updateMaxHeightPrevoted(bftVotes, paramsCache);
+		await updateMaxHeightPrecommitted(bftVotes, paramsCache);
+		updateMaxHeightCertified(bftVotes, context.header);
+		await votesStore.setWithSchema(EMPTY_KEY, bftVotes, bftVotesSchema);
+		await deleteBFTParameters(paramsStore, bftVotes.maxHeightCertified + 1);
```
janhack, line 75: I am not sure this step is correct. In the LIP it says

> Let `h` be the largest integer such that `h <= bftVotes.maxHeightCertified+1` and there is an entry in the BFT Parameters store with key `h`. Then remove any key-value pair from BFT Parameters with a key strictly smaller than `h`.

Assume `bftVotes.maxHeightCertified == 400` and `h == 310` as there is no larger entry in the BFT Parameters store at height at most 400. Wouldn't the entry at height `h` be removed here?

janhack (changes requested):
In general the implementation looks very clean and easy to read with clear correspondence to the LIP. I left a few comments regarding some unclear details.

In general, I only reviewed the state transition logic here (basically all the steps in **After Block Execution** and the related logic defined in the LIP "Add weights to Lisk-BFT consensus protocol"). I did not check any of the tests or schema definitions. I also only reviewed the code that was changed. Let me know if I should have a look add anything else.

framework/src/modules/bft/module.ts
shuse2: https://github.com/LiskHQ/lisk-sdk/pull/6796/files#diff-b0486fd5b86e4f912f1d74c0eca9757dfee97cbca56340a907839686961d99c7R41-R55

in that case, the input will be 401 and since height 401 doesn't exist, it will not delete anything with above code.
it will delete if height `401` exist in the params store,  and it will delete all the params from `0 - 400`

framework/src/modules/bft/bft_votes.ts
shuse2: Considering the standby delegate is not part of the validators as above, shouldn't this logic changed to `continue`?
`blockBFTInfos` will include the block from standby delegate?

framework/src/modules/bft/bft_votes.ts
janhack: All good then.

framework/src/modules/bft/module.ts
janhack: > it will delete all the params from 0 - 400

But this would be incorrect, as the entry with key 310 needs to be maintained.

framework/src/modules/bft/bft_params.ts
janhack: @shuse2 Do you think some further optimization as suggested above is possible and would make sense to introduce?

framework/src/modules/bft/module.ts
shuse2: > it will delete all the params from 0 - 400

will happen only if `if height 401 exist in the params store`, but that's not correct?

framework/src/modules/bft/bft_params.ts
shuse2: wouldnt it difficult to predict how frequent `setBFTParameters` will be called? (we know how frequent it will be called for DPoS, but not for other modules)

framework/src/modules/bft/bft_votes.ts
janhack: We could change the logic to use `continue` here but it should not make a difference as this case should never occur.

As I mentioned, if a delegate is a standby delegate in a round, then they are now counted as not active and therefore `validatorInfo.minActiveHeigh` is the first height of the round since when they are continuosly active delegate. So we should never check this logic for a height where `blockBFTInfo.generatorAddress` is standby delegate.

framework/src/modules/bft/bft_params.ts
janhack: As far as I understand, the cache is only used for in the `afterBlockExecute` hook. During this hook, the BFT Parameters store stays unchanged and I only use the BFT Parameters for heights that appear in `bftVotes.blockBFTInfos`.  

So what I could do is one iteration of the state store to create a cache for the height `bftVotes.blockBFTInfos[-1].height` to `bftVotes.blockBFTInfos[0].height`.

framework/src/modules/bft/module.ts
janhack: Ah, I missed the following line above
```
if (!exist) {
    return;
}
```
I still think the logic is incorrect. Assume we have the following entries in the BFT Parameters store:
104 -> ... 
207 -> ...
310 -> ..
Then according to the LIP, all entries, except the largest one that is at most 401, should be removed. If I understand the implementation correctly, no entry would be removed as there is no entry for height 401.

janhack:
I replied again to the open comments.

framework/src/modules/bft/module.ts
shuse2: i think it's doing
there is an entry in the BFT Parameters store with key `401`. Then remove any key-value pair from BFT Parameters with a key strictly smaller than `401`.
so if there is no entry for height 401, it doesn't delete anything.
so if `h` is 310`, it will delete 104 and 207?

shuse2: or when it's called with `401` and there is not entry for 401, it should delete everything except `310`?

framework/src/modules/bft/bft_params.ts
```diff
@@ -0,0 +1,105 @@
...
+};
+
+export const deleteBFTParameters = async (paramsStore: SubStore, height: number): Promise<void> => {
+	const start = intToBuffer(0, 4, BIG_ENDIAN);
+	const end = intToBuffer(height, 4, BIG_ENDIAN);
+	const results = await paramsStore.iterate({
+		start,
+		end,
+	});
+	if (results.length <= 1) {
```
janhack, line 48: Cosmetic comment: I think this check is not necessary as the loop will not be executed if `results.length == 1`.

framework/src/modules/bft/bft_params.ts
janhack: Nice solution for the cache now. I resolve this comment.

framework/src/modules/bft/bft_votes.ts
janhack: As discussed, we keep it as is.

framework/src/modules/bft/bft_params.ts
janhack: Maybe we can also add a comment below as clarification what this function does, e.g.

> delete all BFT Parameters except the one of largest height which is at most the input height

framework/src/modules/bft/module.ts
```diff
@@ -49,6 +60,26 @@ export class BFTModule extends BaseModule {
...
+		await updatePrevotesPrecommits(bftVotes, paramsCache);
+		await updateMaxHeightPrevoted(bftVotes, paramsCache);
+		await updateMaxHeightPrecommitted(bftVotes, paramsCache);
+		updateMaxHeightCertified(bftVotes, context.header);
+		await votesStore.setWithSchema(EMPTY_KEY, bftVotes, bftVotesSchema);
+		const minRemovingHeight = Math.min(
+			bftVotes.blockBFTInfos[bftVotes.blockBFTInfos.length - 1].height,
+			bftVotes.maxHeightCertified,
+		);
+		await deleteBFTParameters(paramsStore, minRemovingHeight + 1);
```
janhack, line 83: I also updated the BFT module LIP regarding this step, see https://github.com/LiskHQ/lips/pull/113/commits/27fb914c19327c0e5ed7bb9e8d32ba888beb26b6

Here it should be 
```
const minRemovingHeight = Math.min(
    bftVotes.blockBFTInfos[bftVotes.blockBFTInfos.length - 1].height,
    bftVotes.maxHeightCertified+1,
);
await deleteBFTParameters(paramsStore, minRemovingHeight);
```

I would also rename `minRemovingHeight` to `minHeightBFTParametersRequired` or something similar.

janhack (changes requested):
I think there is still a minor issue in the BFT Parameters deletion.

framework/src/modules/bft/module.ts
janhack: > or when it's called with 401 and there is not entry for 401, it should delete everything except 310?

Yep, that is correct.

janhack: I think we clarified this point so the comment can be resolved.

framework/src/modules/bft/bft_params.ts
shuse2: the equal case might be not necessary, but empty case is important, so ill keep the condition and add the comment

janhack: Note that also for empty results array the loop is not executed, but it is also correct like this.

janhack (approved):
Looks all good.

mitsuaki-u (approved):

PR #6797, opened 2021-09-22
Implement verifyTransaction hook of Auth Module - Closes #6734

### LINKED ISSUE #6734: Implement auth module verifyTransaction hook
### Description

As part of the verification of a transaction `trs`, the following checks are applied. If the checks fail the transaction is invalid and the whole block is discarded.
- Check that the transaction `nonce` is equal to the account `nonce` by returning `isValidNonce(trs)`.

```ts
isValidNonce(trs):
    let senderAddress be the address corresponding to trs.senderPublicKey
    return trs.nonce == authAccount(senderAddress).nonce
 ```
 
Notice that a transaction in the transaction pool with a `nonce` greater than the account `nonce` is considered "pending" rather than "invalid", since it could become "valid" in the future.
**Note:** Return `status` with 'pending' type if nonce is greater than account nonce

- Check the `signatures` property of the transaction [as explained in LIP 0017](https://github.com/LiskHQ/lips-staging/blob/update-LIP31/proposals/lip-0017.md#signatures-replaces-signature) using `authAccount(address)`, where `address` is the address corresponding to `trs.senderPublicKey`.

### Acceptance Criteria

- Should only pass the check if the nonce is equal to account nonce
- Should check address property belongs to the corresponding senderPublicKey
- Add unit tests checking the verify conditions

### Additional Information

- Related to #6731
- Check Sequence and Keys module for the code reference

### PR DESCRIPTION
**Pull request by ishantiw**
### What was the problem?

This PR resolves #6734 

### How was it solved?

🌱 Add verifyTransaction hook implementation 1a1ee89
♻️ Update protocol specs for multisignature registration 37441e4
✅ Add auth module tests 41e6bc0
✅ Add nonce related tests 4f349fc

### How was it tested?

`npm run test:unit auth_module.spec`

### CODE REVIEW

File: framework/src/modules/auth/module.ts [hunk 1]
```diff
@@ -33,10 +43,144 @@ export class AuthModule extends BaseModule {
...
+					transaction.nonce,
+					senderAccount.nonce,
+				),
+			};
+		}
+		const transactionBytes = transaction.getSigningBytes();
+
+		if (
+			transaction.moduleID === this.id &&
+			this.commands.find(cmd => cmd.id === transaction.commandID)
```
shuse2, line 68: I think second condition will be invalid if we add more commands. Probably better to specifically check if it's 'registerMultisignatureGroup' command

framework/src/modules/auth/module.ts
```diff
@@ -33,10 +43,144 @@ export class AuthModule extends BaseModule {
...
+				// Verify each optional key signed in order
+				validateKeysSignatures(
+					TAG_TRANSACTION,
+					networkIdentifier,
+					optionalKeys,
+					transaction.signatures.slice(mandatoryKeys.length + 1),
+					transactionBytes,
+					transaction.id,
+				);
+			} catch (error) {
```
shuse2, line 124: I think this is more explicit and better, but even if error is not caught, the error will be treated as FAIL outside of this function =)

framework/src/modules/auth/module.ts
```diff
@@ -33,10 +43,144 @@ export class AuthModule extends BaseModule {
 	public configSchema = configSchema;
 	public commands = [new RegisterMultisignatureCommand(this.id)];
 
-	// eslint-disable-next-line @typescript-eslint/require-await
-	public async verifyTransaction(_context: TransactionVerifyContext): Promise<VerificationResult> {
+	public async verifyTransaction(context: TransactionVerifyContext): Promise<VerificationResult> {
```
shuse2, line 46: it might be easier to read if we split each cases in private function?
- registerMultisignatureGroup
- multi-signature
- single-signature

framework/test/unit/modules/auth/auth_module.spec.ts
```diff
@@ -0,0 +1,922 @@
...
+			.calledWith(address, authAccountSchema)
+			.mockReturnValue({
+				mandatoryKeys: [],
+				optionalKeys: [],
+				nonce: BigInt(0),
+				numberOfSignatures: 0,
+			});
+	});
+
+	describe('beforeTransactionApply', () => {
```
shuse2, line 93:
```suggestion
	describe('verifyTransaction', () => {
```

shuse2 (changes requested):
LGTM, just some minor comments

framework/src/modules/auth/module.ts
ishantiw: ah yeah true, actually I had pending case in mind so was trying handle and reform the return statuses for failed cases too

framework/src/modules/auth/module.ts
ishantiw: extracted to util funcs

shuse2 (approved):

mehmetegemen (approved):
Minor comment but LGTM, great job! 🎉

framework/src/modules/auth/constants.ts
```diff
@@ -14,3 +14,6 @@
 
 export const MODULE_ID_AUTH = 12; // TBD
 export const STORE_PREFIX_AUTH = 0x0000;
+
+// Commands
+export const AUTH_REGISTER_MULTISIGNATURE_CMD_ID = 0;
```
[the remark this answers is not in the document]
ishantiw: yeah I think its better to be consistent with other modules, also, starting with `COMMAND` is better to group these constants

framework/src/modules/auth/constants.ts
```diff
@@ -14,3 +14,6 @@
 
 export const MODULE_ID_AUTH = 12; // TBD
 export const STORE_PREFIX_AUTH = 0x0000;
+
+// Commands
+export const COMMAND_ID_DELEGATE_REGISTRATION = 0;
```
mitsuaki-u, line 19: This should be `COMMAND_ID_MULTISIG_REGISTRATION`?

PR #6805, opened 2021-09-29
Implement auth endpoint - Closes #6738

### LINKED ISSUE #6738: Add Endpoint for auth module
#### `getAuthAccount`

This function is used to retrieve information about an auth account.
##### Parameters
`address`: A 20 bytes value identifying the auth account.
##### Returns
This function returns `authAccount(address)`. If there is no auth account corresponding to address, it returns an empty object.

#### `verifySignatures`
This function verifies the signatures of a given transaction `trs`, including transactions from multisignature accounts.
##### Parameters
`trs`: A transaction object.
##### Returns
This function returns `true` if the transaction object contains a valid signature, `false` otherwise, as specified in [LIP 0017](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0017.md#transaction-creation-and-verification).

#### `verifyNonce`
This function verifies the nonce of a given transaction trs and returns a boolean value.
##### Parameters
`trs`: A transaction object.

### Acceptance Criteria

- Add unit test for all the endpoints

### Additional Information

Related to #6731

mehmetegemen:
### What was the problem?

This PR resolves #6738 

### How was it solved?

- Added `verifySignature`, `verifyNonce` and `getAuthAccount` methods to Endpoint
- Added a `createTransientModuleEndpointContext` test utility for convenience

### How was it tested?

- Added unit tests


framework/src/modules/auth/utils.ts
```diff
@@ -190,3 +190,25 @@ export const verifyNonce = (
...
+	if (typeof transactionParameter !== 'string') {
+		throw new Error('Transaction parameter must be a string.');
+	}
+
+	let transactionParameterObject;
+	try {
+		transactionParameterObject = JSON.parse(transactionParameter) as Record<string, unknown>;
+	} catch {
+		throw new Error('Transaction parameter is not a valid JSON object.');
+	}
```
shuse2, line 208: In the current implementation, we expect transaction from endpoint to be encoded transaction in hex string.
I think for now we should keep it consistent. ie) you can do something like below
```
- check input is hex string
- Create buffer from the hex string
- Transaction.fromBytes()
```

framework/src/modules/auth/endpoint.ts
```diff
@@ -12,6 +12,113 @@
...
+
+		const { senderPublicKey, signatures } = transaction;
+
+		const accountAddress = getAddressFromPublicKey(senderPublicKey);
+
+		const store = getStore(MODULE_ID_AUTH, STORE_PREFIX_AUTH);
+		const account = await store.getWithSchema<AuthAccount>(accountAddress, authAccountSchema);
+
+		const transactionBytes = transaction.getSigningBytes();
+
```
shuse2, line 80: there is a case where command is registerMultisignatureGroup.
In that case, there is different signature check. Please refer to `beforeTransactionExecute` step

shuse2 (changes requested):

framework/src/modules/auth/utils.ts
mehmetegemen: done, please check.

framework/src/modules/auth/endpoint.ts
mehmetegemen: done, please check.

framework/test/unit/modules/auth/endpoint.spec.ts
```diff
@@ -0,0 +1,461 @@
...
+			expect(authAccount).toHaveProperty(
+				'numberOfSignatures',
+				expectedAuthAccount.numberOfSignatures,
+			);
+			expect(authAccount).toHaveProperty('mandatoryKeys', expectedAuthAccount.mandatoryKeys);
+			expect(authAccount).toHaveProperty('optionalKeys', expectedAuthAccount.optionalKeys);
+		});
+	});
+
+	describe('verifyTransaction', () => {
```
shuse2, line 149: I think fail case is missing

shuse2 (changes requested):
mostly looks good to me. Just missing test case

framework/test/unit/modules/auth/endpoint.spec.ts
shuse2: We discussed and this case is covered by the util function, so it is not needed

shuse2 (approved):

Incede (approved):

PR #6807, opened 2021-09-30
UnlockHeight is not considered on unlockToken transaction during punishment - Closes #6793

### LINKED ISSUE #6793: unlockHeight is not considered on unlockToken transaction during punishment
### Expected behavior
According to the [LIP](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0024.md#update-to-validity-of-unlock-transaction)
```
lastPomHeight < U.unvoteHeight + 2000
or 
lastPomHeight < U.unvoteHeight + 260,000
```
needs to be considered, so that if unvoted 2000 blocks before, it the token will not be locked.

### Actual behavior

https://github.com/LiskHQ/lisk-sdk/blob/main/framework/src/modules/dpos/transaction_assets/unlock_transaction_asset.ts#L108-L127

https://github.com/LiskHQ/lisk-sdk/blob/a1e9c65edee71b7d415bfed9b7cfa29f03a3d385/framework/src/modules/dpos/utils.ts#L62-L72

Currently, during the execution of unlocking, the unvote height is not considered for the condition. Therefore, if the voter unvoted the delegate but not unlocked, then delegate got punished, the token will become locked until first condition is met

Also, this feature should be abled to switched on by genesis config using height

### Steps to reproduce
1. Vote for a delegate
2. Unvote a delegate
3. Punish a delegate

5.1.4-

Incede:
### What was the problem?

This PR resolves #6793 

### How was it solved?

* Added condition to check `unvoteHeight` as per the [LIP](https://github.com/LiskHQ/lips/blob/master/proposals/lip-0024.md#update-to-validity-of-unlock-transaction)
* Refactor existing code
* Added condition to make the change configurable by hardfork height 

### How was it tested?

Added unit tests

### DIFF
--- a/framework/src/modules/dpos/transaction_assets/pom_transaction_asset.ts
+++ b/framework/src/modules/dpos/transaction_assets/pom_transaction_asset.ts
@@ -20,7 +20,7 @@ import { BaseAsset } from '../../base_asset';
 import { ApplyAssetContext, ValidateAssetContext } from '../../../types';
 import { MAX_PUNISHABLE_BLOCK_HEIGHT_DIFFERENCE, MAX_POM_HEIGHTS } from '../constants';
 import { DPOSAccountProps, PomTransactionAssetContext } from '../types';
-import { getUnlockDelayPeriod, validateSignature } from '../utils';
+import { getPunishmentPeriod, validateSignature } from '../utils';
 
 const signingBlockHeaderSchema = {
 	$id: 'lisk/dpos/signingBlockHeader',
@@ -155,7 +155,7 @@ export class PomTransactionAsset extends BaseAsset<PomTransactionAssetContext> {
 		}
 
 		if (
-			getUnlockDelayPeriod(
+			getPunishmentPeriod(
 				delegateAccount,
 				delegateAccount,
 				store.chain.lastBlockHeaders[0].height,
--- a/framework/src/modules/dpos/transaction_assets/pom_transaction_asset.ts
+++ b/framework/src/modules/dpos/transaction_assets/pom_transaction_asset.ts
@@ -20,7 +20,7 @@ import { BaseAsset } from '../../base_asset';
 import { ApplyAssetContext, ValidateAssetContext } from '../../../types';
 import { MAX_PUNISHABLE_BLOCK_HEIGHT_DIFFERENCE, MAX_POM_HEIGHTS } from '../constants';
 import { DPOSAccountProps, PomTransactionAssetContext } from '../types';
-import { getPunishmentPeriod, validateSignature } from '../utils';
+import { getUnlockDelayPeriod, validateSignature } from '../utils';
 
 const signingBlockHeaderSchema = {
 	$id: 'lisk/dpos/signingBlockHeader',
@@ -155,7 +155,7 @@ export class PomTransactionAsset extends BaseAsset<PomTransactionAssetContext> {
 		}
 
 		if (
-			getPunishmentPeriod(
+			getUnlockDelayPeriod(
 				delegateAccount,
 				delegateAccount,
 				store.chain.lastBlockHeaders[0].height,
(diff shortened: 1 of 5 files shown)


framework/src/modules/dpos/transaction_assets/pom_transaction_asset.ts
```diff
@@ -155,7 +155,7 @@ export class PomTransactionAsset extends BaseAsset<PomTransactionAssetContext> {
 		}
 
 		if (
-			getPunishmentPeriod(
+			getUnlockDelayPeriod(
```
MaximeGagnebin, line 158: Should we rename the function here too?
I think that PoM is different than unlock. Maybe it makes sense to keep the name here? I see that this is a bit conflicting as this function is used in both `unlock` and `pom`. 
With the way you made the changes, I have to say that keeping `getPunishmentPeriod` looks also good to me.

MaximeGagnebin (approved):
Nice, the logic looks all good. I only left a comment regarding the function name, so I'll approve :) .

framework/src/modules/dpos/transaction_assets/unlock_transaction_asset.ts
```diff
@@ -105,25 +112,36 @@ export class UnlockTransactionAsset extends BaseAsset<UnlockTransactionAssetCont
...
+						'Unlocking is not permitted as the delegate is currently being punished.',
+					);
+				}
+				// unlocking condition before hard fork, to be deprecated
+			} else if (currentHeight < this._unlockFixHeight) {
+				if (minUnlockDelay > currentHeight) {
+					throw new Error(
+						'Unlocking is not permitted as the delegate is currently being punished.',
+					);
+				}
```
shuse2, line 144:
```suggestion
			if (currentHeight >= this._unlockFixHeight &&
					minUnlockDelay > currentHeight &&
					lastPomHeight !== undefined &&
					lastPomHeight < minWaitingHeight
			) {
				throw new Error(
					'Unlocking is not permitted as the delegate is currently being punished.',
				);
			}
			// unlocking condition before hard fork, to be deprecated
			if (currentHeight < this._unlockFixHeight && minUnlockDelay > currentHeight) {
				throw new Error(
					'Unlocking is not permitted as the delegate is currently being punished.',
				);
			}
```

I think nested if can be combined?

framework/src/modules/dpos/transaction_assets/unlock_transaction_asset.ts
```diff
@@ -105,25 +112,36 @@ export class UnlockTransactionAsset extends BaseAsset<UnlockTransactionAssetCont
...
-				sender.address,
-				delegate.address,
-				store.chain.lastBlockHeaders[0].height,
-				unlock,
-			);
+			const currentHeight = store.chain.lastBlockHeaders[0].height + 1;
+			const minWaitingHeight = getMinWaitingHeight(sender.address, delegate.address, unlock);
 
-			if (waitingPeriod > 0) {
+			if (minWaitingHeight > currentHeight) {
```
shuse2, line 118: `minWaitingHeight == currentHeight` should be ok here?

shuse2 (commented):
Some minor comments/questions, but nicely done!

framework/src/modules/dpos/transaction_assets/unlock_transaction_asset.ts
shuse2: maybe this is question to @MaximeGagnebin

MaximeGagnebin: so, `minWaitingHeight == currentHeight` should be ok, indeed. 
Isn't that what the logic does? like if `>` then throw error?

MaximeGagnebin (approved):
Still looks good 🙌

framework/src/modules/dpos/transaction_assets/unlock_transaction_asset.ts
shuse2: yes indeed 😅

shuse2 (approved):

PR #6808, opened 2021-09-30
🌱  Bootstrap random module - Closes #6774

### LINKED ISSUE #6774: Create class, constants, schemas for random module
### Description

- Create random module class
- Create empty API and Endpoint class
- Initialize API and Endpoint in the Random class 
- Define constants (constants.ts), types (types.ts) and schema (schema.ts)
- Define configurations which are received from the init hook (length of validator reveals `MAX_LENGTH_REVEALS_MAINCHAIN`)
- Define hooks initBlock, afterGenesisBlockExecute, verifyBlock and afterBlockExecute

### Acceptance Criteria

- All files and classes exist in the `random` module folder

### Additional Information

- Related #6773

ishantiw:
### What was the problem?

This PR resolves #6774

### How was it solved?

- 🌱  Add class files for module, api and endpoint 0d3e48f
- 🌱  Add schema, types and constants files 0d3e48f
- Assign `maxLengthRevealsMainchain` constant from config in `init()` 0d3e48f

### How was it tested?

Check all the files exist under `random` folder and all the declarations are present


framework/src/modules/random/constants.ts
```diff
@@ -0,0 +1,16 @@
...
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+export const MODULE_ID_RANDOM = 13;
```
mehmetegemen, line 14: We used 13 in DPoS_v2

framework/src/modules/random/module.ts
```diff
@@ -0,0 +1,44 @@
...
+	public async initBlock(_context: BlockGenerateContext): Promise<void> {}
+
+	// eslint-disable-next-line @typescript-eslint/require-await, @typescript-eslint/no-empty-function
+	public async verifyBlock(_context: BlockVerifyContext): Promise<void> {}
+
+	// eslint-disable-next-line @typescript-eslint/require-await, @typescript-eslint/no-empty-function
+	public async afterGenesisBlockExecute?(_context: GenesisBlockExecuteContext): Promise<void> {}
+
+	// eslint-disable-next-line @typescript-eslint/require-await, @typescript-eslint/no-empty-function
+	public async afterBlockExecute?(_context: BlockAfterExecuteContext): Promise<void> {}
```
mehmetegemen, line 43: Why these two methods are optional?

mehmetegemen:
few minor comments, rest looks good to me.

framework/src/modules/random/module.ts
```diff
@@ -0,0 +1,61 @@
...
+
+	// eslint-disable-next-line @typescript-eslint/require-await
+	public async init(args: ModuleInitArgs): Promise<void> {
+		const { moduleConfig } = args;
+		const errors = validator.validate(randomModuleConfig, moduleConfig);
+		if (errors.length) {
+			throw new LiskValidationError(errors);
+		}
+		this._maxLengthRevealMainchain =
+			(moduleConfig as ModuleConfig).maxLengthRevealsMainchain ?? MAX_LENGTH_REVEALS_MAINCHAIN;
```
mehmetegemen, line 45: In other modules we set `moduleConfig` to `this._moduleConfig` for a reason I don't remember but it's the convention. And if you set it that way you don't have to type `(moduleConfig as ModuleConfig).configName` every time you want to use the config, instead you can do `this._moduleConfig = moduleConfig as ModuleConfig` and use it without casting. Also this way hooks can use the config too, otherwise config is trapped inside the init, though I don't know if hooks use the config.

framework/src/modules/random/constants.ts
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+export const MODULE_ID_RANDOM = 14;
+export const MAX_LENGTH_REVEALS_MAINCHAIN = 206;
```
shuse2, line 15: maybe better to name `DEFAULT_MAX_LENGTH_REVEALS`?

framework/src/modules/random/module.ts
```diff
@@ -0,0 +1,61 @@
...
+import { randomModuleConfig } from './schemas';
+import { ModuleConfig } from './types';
+
+export class RandomModule extends BaseModule {
+	public id = MODULE_ID_RANDOM;
+	public name = 'random';
+	public api = new RandomAPI(this.id);
+	public endpoint = new RandomEndpoint(this.id);
+
+	private _maxLengthRevealMainchain!: number;
```
shuse2, line 35:
```suggestion
	private _maxLengthReveal!: number;
```
as it is not only for mainchain

framework/src/modules/random/module.ts
shuse2: I think either way is fine. There are no strict rule to save whole config or just one value. which every make sense per module should be good IMO.
for example, if there are too many property in config, maybe keep it as config while if there is one, just keep the one value

shuse2 (changes requested):

framework/src/modules/random/module.ts
ishantiw: I saw both approaches in the new modules, maybe either of the approaches is fine, @mehmetegemen btw the config is not trapped in the init, you assign the properties of a config to the class properties so it can be used in other places too. To be consistent i'll use `this._moduleConfig = moduleConfig as ModuleConfig`

framework/src/modules/random/module.ts
```diff
@@ -41,10 +42,13 @@ export class RandomModule extends BaseModule {
 		if (errors.length) {
 			throw new LiskValidationError(errors);
 		}
-		this._maxLengthRevealMainchain =
-			(moduleConfig as ModuleConfig).maxLengthRevealsMainchain ?? MAX_LENGTH_REVEALS_MAINCHAIN;
+		this._moduleConfig = moduleConfig as ModuleConfig;
+		this._maxLengthReveal =
+			(moduleConfig as ModuleConfig).maxLengthRevealsMainchain ?? DEFAULT_MAX_LENGTH_REVEALS;
```
mehmetegemen, line 47: I think if you use `this._moduleConfig` here instead of `(moduleConfig as ModuleConfig)` then you don't repeat `(moduleConfig as ModuleConfig)` two times in init.

mehmetegemen: By doing that also you don't need the console.log below

shuse2: i guess you can keep one of them (not necessary to duplicate the data)

mehmetegemen (approved):

shuse2 (approved):

PR #6878, opened 2021-11-03
Update validators from DPoS module - Closes #6876

### LINKED ISSUE #6876: Implement updateValidators for DPoS Module
### Description

Create a private function in the DPoS module for `afterBlockExecute` hook to use it and implement the logic in an auxiliary function that is reusable and call it in that private function. Also create constants, adjust interfaces and add tests. Also adjust other auxiliary functions that this function going to use for new interfaces.

In this function the logic is about updating validator list with randomSeed that we get from randomAPI and setting BFT parameters with BFT API. See old code for the responsibility of the code and LIP for new adjustments needed to be done to use new modules.

### Acceptance Criteria
- Unit tests must pass

### Additional Information
- Code must use new module APIs and as much as the old DPoS module code
- Related to #6722

shuse2:
### What was the problem?

This PR resolves #6876

### How was it solved?

- Migrated the old utilities to select standby delegates
- Update to get random seed from random module
- Update to set validator to validator module

### How was it tested?

Migrate tests from previous dpos


framework/src/modules/dpos_v2/module.ts
```diff
@@ -42,7 +42,12 @@ import {
 	SnapshotStoreData,
 } from './types';
 import { Rounds } from './rounds';
-import { isCurrentlyPunished } from './utils';
+import {
+	isCurrentlyPunished,
+	selectStandbyDelegates,
+	shuffleDelegateList,
+	validtorsEqual,
```
mehmetegemen, line 49: typo

framework/src/modules/dpos_v2/utils.ts
```diff
@@ -147,3 +148,96 @@ export const getVoterOrDefault = async (voterStore: SubStore, address: Buffer) =
...
+	randomSeed2?: Buffer,
+): Buffer[] => {
+	const numberOfCandidates = 1 + (randomSeed2 !== undefined ? 1 : 0);
+	// if delegate weights is smaller than number selecting, select all
+	if (delegateWeights.length <= numberOfCandidates) {
+		return delegateWeights.map(c => c.delegateAddress);
+	}
+	const result: Buffer[] = [];
+	const index = pickStandByDelegate(delegateWeights, randomSeed1);
+	const [selected] = delegateWeights.splice(index, 1);
```
mehmetegemen, line 216: if the below is `secondStandby` maybe we can call this `firstStandby`?

framework/test/unit/modules/dpos_v2/module.spec.ts
```diff
@@ -425,4 +425,218 @@ describe('DPoS module', () => {
...
+				moduleConfig: defaultConfigs,
+			});
+		});
+
+		describe('given valid scenarios', () => {
+			const scenarios = [
+				forgerSelectionZeroStandbyScenario,
+				forgerSelectionOneStandbyScenario,
+				forgerSelectionTwoStandbyScenario,
+				forgerSelectionLessTHan103Scenario,
```
mehmetegemen, line 446: THan, typo

framework/test/unit/modules/dpos_v2/module.spec.ts
```diff
@@ -425,4 +425,218 @@ describe('DPoS module', () => {
...
+				forgerSelectionMoreThan2StandByScenario,
+			];
+			const defaultRound = 5;
+
+			for (const scenario of scenarios) {
```
mehmetegemen, line 451: Can't we use `describe.each` here?

mehmetegemen (commented):
Minor comments, LGTM 🎉

framework/test/unit/modules/dpos_v2/module.spec.ts
shuse2: I think we can, but  i would keep it like  this for now. I just copied from the existing test =)

mitsuaki-u (approved):

mehmetegemen (approved):

PR #6897, opened 2021-11-12
Add block transaction and assets root validation - Closes #6886

### LINKED ISSUE #6886: Add  txRoot  and assetRoot validation
### Description

Add transaction root and asset  root  validation in  Block.validate function

### Acceptance Criteria

- validate function should remain synchronous function

### Additional Information
#6885 must be implemented first

mitsuaki-u:
### What was the problem?

This PR resolves #6886

### How was it solved?

- Added txRoot and assetsRoot validation to block in `lisk-chain`
- Added unit tests

### How was it tested?

Run unit tests

### DIFF
--- a/elements/lisk-chain/test/utils/block.ts
+++ b/elements/lisk-chain/test/utils/block.ts
@@ -69,6 +69,7 @@ export const createValidDefaultBlock = async (
 	const txTree = new MerkleTree();
 	await txTree.init(payload.map(tx => tx.id));
 	const blockAssets = block?.assets ?? new BlockAssets();
+	const assetsRoot = await blockAssets.getRoot();
 
 	const blockHeader = new BlockHeader({
 		version: 2,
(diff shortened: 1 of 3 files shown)


elements/lisk-chain/test/utils/block.ts
```diff
@@ -76,7 +77,7 @@ export const createValidDefaultBlock = async (
 		timestamp: genesis.header.timestamp + 10,
 		transactionRoot: txTree.root,
 		stateRoot: getRandomBytes(32),
-		assetsRoot: getRandomBytes(32),
+		assetsRoot: await blockAssets.getRoot(),
```
shuse2, line 80: Prefer to keep the `await` outside of the assignment.
Also, probably we can update `getRoot` to be synchronous function using `calculateMerkleRootWithLeaves`. If it affects too many places, we can split it in another issue

shuse2 (approved):

elements/lisk-chain/test/utils/block.ts
mitsuaki-u: Let's do it in a separate issue since it wasn't part of the original estimation

ishantiw (approved):

PR #6901, opened 2021-11-12
Update application class - Closes #6629 and #6892

### LINKED ISSUE #6629: Update application class to follow new framework architecture
### Description

- Update default application to return an object below, so that external library can use api as dependency.
```
interface DefaultApplication {
 app: Application;
 api: {
   token: xxx,
   dpos: xxx,
  ....
 }, 
}
```
- Update `run` function to accept genesis block as instance of Block.
- 

With the new architecture, external module will need direct access to API for dependency with type. Therefore, we need to expose the default module api as well.

> Update `run` function to accept genesis block as instance of Block.

Genesis block cannot be maintained in the memory. Therefore, it should be passed in the run phase, and only temporally used. (no assignment as variable). Also, receiving as instance will increase the flexibility of the genesis block format.

### Additional Information
- this issue should be handle after #6612

### LINKED ISSUE #6892: Fix `getGeneratorsBetweenTimestamps` type and return keys
### Description

- Update the return type to be `Record<string, number>` to reflect the correct typing
- Result key should be in `binary` format than `utf8`. ie) `generatorList.addresses.map(buf => buf.toString('binary'));`  because some character in Buffer cannot be represented in `utf8` format

### Acceptance Criteria

- Update unit test
- DPoS dependency type should be updated

shuse2:
### What was the problem?

This PR resolves #6629 and resolves #6892

### How was it solved?

- Update application class to return default application with APIs
- Remove getDefaultModules since it is no longer useful
- Fix #6892 to match type for build
- Expose bftAPI and validatorAPI
- Fix expected interface name `isSeedRevealValid` from `isValidSeedReveal`
- Fix DPoS not to use config in `addDependencies`

### How was it tested?

- Update unit tests

### DIFF
--- a/framework/src/modules/token/api.ts
+++ b/framework/src/modules/token/api.ts
@@ -158,4 +158,57 @@ export class TokenAPI extends BaseAPI {
 		}
 		await userStore.setWithSchema(address, address, userStoreSchema);
 	}
+
+	public async burn(
+		apiContext: APIContext,
+		senderAddress: Buffer,
+		_id: TokenID,
+		amount: bigint,
+	): Promise<void> {
+		const userStore = apiContext.getStore(this.moduleID, STORE_PREFIX_USER);
+		const sender = await userStore.getWithSchema<UserStoreData>(senderAddress, userStoreSchema);
+		if (sender.availableBalance < amount + this._minBalance) {
+			throw new Error(
+				`Sender ${senderAddress.toString(
+					'hex',
+				)} balance ${sender.availableBalance.toString()} is not sufficient for ${(
+					amount + this._minBalance
+				).toString()}`,
+			);
+		}
+		sender.availableBalance -= amount;
+		await userStore.setWithSchema(senderAddress, sender, userStoreSchema);
+	}
+
+	public async mint(
+		apiContext: APIContext,
+		address: Buffer,
+		_id: TokenID,
+		amount: bigint,
+	): Promise<void> {
+		const userStore = apiContext.getStore(this.moduleID, STORE_PREFIX_USER);
+		let recipient: UserStoreData;
+		try {
+			recipient = await userStore.getWithSchema<UserStoreData>(address, userStoreSchema);
+		} catch (error) {
+			if (!(error instanceof NotFoundError)) {
+				throw error;
+			}
+			recipient = {
+				availableBalance: BigInt(0),
+				lockedBalances: [],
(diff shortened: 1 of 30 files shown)


framework/src/modules/token/api.ts
```diff
@@ -158,4 +158,57 @@ export class TokenAPI extends BaseAPI {
...
+		} catch (error) {
+			if (!(error instanceof NotFoundError)) {
+				throw error;
+			}
+			recipient = {
+				availableBalance: BigInt(0),
+				lockedBalances: [],
+			};
+		}
+		if (recipient.availableBalance < amount + this._minBalance) {
```
Incede, line 202: Why this check here ?
Do you mean `if (amount < this_minbalance)` ?

Incede (changes requested):

mehmetegemen (approved):
Great work Shu

framework/src/modules/token/api.ts
shuse2: good catch. it shouldve been `recipient.availableBalance + amount  < + this._minBalance` to check if the  recipient  will have more  than min balance after receiving  the  token

framework/src/modules/token/api.ts
```diff
@@ -81,7 +81,7 @@ export class TokenAPI extends BaseAPI {
 				lockedBalances: [],
 			};
 		}
-		if (recipient.availableBalance < amount + this._minBalance) {
+		if (recipient.availableBalance + amount < this._minBalance) {
 			throw new Error(
 				`Recipient ${recipientAddress.toString(
 					'hex',
```
Incede, line 87: I think the error message doesn't fit the logic checked here

framework/src/modules/token/api.ts
```diff
@@ -158,4 +158,57 @@ export class TokenAPI extends BaseAPI {
...
+			}
+			recipient = {
+				availableBalance: BigInt(0),
+				lockedBalances: [],
+			};
+		}
+		if (recipient.availableBalance + amount < this._minBalance) {
+			throw new Error(
+				`Recipient ${address.toString(
+					'hex',
```
Incede, line 205: same here, error msg needs to be updated

Incede (changes requested):

Incede (approved):

mehmetegemen (approved):

PR #6912, opened 2021-11-17
 Update genesis block static validation - Closes #6895 

### LINKED ISSUE #6895: Update genesis block static validation
### Description
* Create a function `validateGenesis` under `block.ts` in chain
* Validate block schema as per [LIP](https://github.com/LiskHQ/lips/blob/add-LIP-update_genesis_block_schema_and_processing/proposals/lip-update_genesis_block_schema_and_processing.md)
* Validate properties transaction, block id, block assets and it's schema as specified in the [LIP](https://github.com/LiskHQ/lips/blob/add-LIP-update_genesis_block_schema_and_processing/proposals/lip-update_genesis_block_schema_and_processing.md)

### Acceptance Criteria

* Add unit test for each property validation

### Additional Information
* Related Issue #6893

Incede:
### What was the problem?

This PR resolves #6895

### How was it solved?

* Created function `validateGenesis` under `block.ts` in chain
*  Created function `validateGenesis` under `block_assets.ts` in chain
* Added static validation checks as per [LIP](https://github.com/LiskHQ/lips/blob/add-LIP-update_genesis_block_schema_and_processing/proposals/lip-update_genesis_block_schema_and_processing.md)
* Refactored & updated static validation of normal block to check asset schema

### How was it tested?
* Added unit tests

### DIFF
--- a/elements/lisk-chain/src/block.ts
+++ b/elements/lisk-chain/src/block.ts
@@ -16,7 +16,6 @@ import { codec } from '@liskhq/lisk-codec';
 import { regularMerkleTree } from '@liskhq/lisk-tree';
 import { BlockAssets } from './block_assets';
 import { BlockHeader } from './block_header';
-import { MAX_ASSET_DATA_SIZE_BYTES } from './constants';
 import { blockSchema } from './schema';
 import { Transaction } from './transaction';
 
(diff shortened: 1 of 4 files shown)


elements/lisk-chain/test/unit/block.spec.ts
```diff
@@ -228,4 +228,186 @@ describe('block', () => {
...
+			it('should throw error', async () => {
+				// Arrange
+				const txs = new Array(20).fill(0).map(() => tx);
+				block = await createValidDefaultBlock({ payload: txs, assets: blockAssets });
+				// Act & assert
+				expect(() => block.validateGenesis()).toThrow('Payload length must be zero');
+			});
+		});
+
+		describe('when an asset data has size more than the limit', () => {
```
shuse2, line 272: this should only applied to the normal block?

elements/lisk-chain/test/unit/block.spec.ts
```diff
@@ -228,4 +228,186 @@ describe('block', () => {
...
+						data: getRandomBytes(64),
+					},
+				];
+				block = await createValidDefaultBlock({ assets: new BlockAssets(assets) });
+				// Act & assert
+				expect(block.validateGenesis()).toBeUndefined();
+			});
+		});
+
+		describe('when the assets are not sorted by moduleID', () => {
```
shuse2, line 308: i think block assets logic are moved to BlockAssets, so the test should also moved/added to the block assets spec

shuse2 (changes requested):

elements/lisk-chain/test/unit/block.spec.ts
Incede: I think here it is just checking that there is no limitation to the `data` property i.e. it does not throw error

elements/lisk-chain/test/unit/block.spec.ts
Incede: yea but since it was not part of the original estimation would be better to move in a separate issue. Also, there are multiple places where the unit tests were added in the parent function instead of the owner(`./validate`). Also need to remove the old tests & usages for the same.

Incede: make sense, I'll move the tests as per your suggestion.

elements/lisk-chain/src/block.ts
```diff
@@ -116,4 +94,19 @@ export class Block {
 			throw new Error('Invalid assets root');
 		}
 	}
+
+	public validateGenesis(): void {
+		// this.header.validate();
```
mitsuaki-u, line 99: This can be uncommented

mitsuaki-u (changes requested):

shuse2 (approved):

PR #6925, opened 2021-11-20
Update to hash SMT key - Closes #6924

### LINKED ISSUE #6924: SMT key should be hashed before updating
### Expected behavior
SMT key should be always 38 bytes, and first 6 bytes should be module id + storage key and the rest should be hashed.

### Actual behavior
SMT key is using the same `key` as the database.

### Steps to reproduce
Observe key `smt.update` or `smt.remove` is used with.

feature branch

shuse2:
### What was the problem?

This PR resolves #6924 

### How was it solved?

- Update SMT key to be hashed except the module id and storage key

### How was it tested?

- Update all unit test to have default key length

### DIFF
--- a/elements/lisk-chain/src/state_store/utils.ts
+++ b/elements/lisk-chain/src/state_store/utils.ts
@@ -13,14 +13,13 @@
  */
 
 import { hash } from '@liskhq/lisk-cryptography';
+import { SMT_PREFIX_SIZE } from '../constants';
 
 export const copyBuffer = (value: Buffer): Buffer => {
 	const copiedValue = Buffer.alloc(value.length);
 	value.copy(copiedValue);
 	return copiedValue;
 };
-
-const SMT_PREFIX_SIZE = 6;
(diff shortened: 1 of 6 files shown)


mitsuaki-u (approved):

elements/lisk-chain/src/state_store/utils.ts
```diff
@@ -12,8 +12,15 @@
...
 
+import { hash } from '@liskhq/lisk-cryptography';
+
 export const copyBuffer = (value: Buffer): Buffer => {
 	const copiedValue = Buffer.alloc(value.length);
 	value.copy(copiedValue);
 	return copiedValue;
 };
+
+const SMT_PREFIX_SIZE = 6;
```
Incede, line 23: probably should be moved to `constants.ts`

Incede (changes requested):

Incede (approved):

PR #6938, opened 2021-11-24
Update genesis block creation - Closes #6628 and #6903

### LINKED ISSUE #6628: Remove lisk-genesis and update framework application and testing to include genesis block creation
### Description

- remove lisk-genesis library
- create `generateGenesisBlock` method in the framework application
```
async generateGenesisBlock({
  height?: number;
  timestamp?: number;
  previousBlockID?: Buffer;
  assets: {
    schema: Schema,
    moduleID: number,
    data: Record<string, unknown>,
  }[],
}): Promise<Block>;
```
-  create  `generateBlock` method in framework application
```
async generateBlock({
	readonly height: number;
	readonly timestamp: number;
	readonly generatorAddress: Buffer;
	readonly privateKey: Buffer;
	readonly transactions?: Transaction[];
	readonly db?: GeneratorDB;
}): Promise<Block>;
```
- Recover genesis block testing tools `framework/src/testing/create_genesis_block.ts`
- testing tools should provide genesis block with default balances
- Update genesis block JSON fixture
### Motivation

In order to support new framework module architecture, genesis block handling was disabled and currently it doesn't have any genesis block handling.

### Additional Information
- This should be handled after new genesis block format is decided.
- https://github.com/LiskHQ/lisk-sdk/issues/6903 is required to do this change

### LINKED ISSUE #6903: Update genesis block to be passed at `run` function of application
### Description

Update `genesis block` to be passed at the `run` function instead of constructor.
Application class does not need  `genesis block` at the instantiation.
When running the application it requires the `genesis block`.

### Acceptance Criteria
- Update the interface
- README must be updated

shuse2:
### What was the problem?

This PR resolves #6628 and resolves #6903

### How was it solved?

- Remove lisk-genesis since genesis block cannot be created independently anymore
- Add genesis block creation function in generator and exposed through application
- Update constructor of application not to require genesis block
- Expose block creation function for consistency and test

### How was it tested?

- Create test for genesis block creation
- Update all existing tests


mitsuaki-u (approved):
🚀

elements/lisk-chain/src/block.ts
```diff
@@ -45,7 +51,7 @@ export class Block {
 		);
 	}
 
-	public static fromJSON(value: Record<string, unknown>): Block {
+	public static fromJSON(value: BlockJSON): Block {
```
mehmetegemen, line 54: I understand that we use `BlockJSON` type as the return type of `toJSON` method. I don't understand why we use a type with non-unknown properties and also have type guards in the method code. Why do we do that?

mehmetegemen (approved):

PR #6939, opened 2021-11-25
Fix module generator and plugin generator templates broken since 185d8acb - Closes #6909

### LINKED ISSUE #6909: Error when you are creating module with lisk generate:module
When you create new module with  lisk generate:module command, there is a line added:

`app.registerModule(HelloModule);`

But default code in modules.ts file:

`
import { Application } from 'lisk-sdk';
export const registerModules = (_app: Application): void => {};
`

So when you try to run your app you got a error: ReferenceError: app is not defined

Probably have to fix it either in the module that generates the application or module skeleton.

markokeeffe:
### What was the problem?

This PR resolves #6909

### How was it solved?

Commit 185d8acb232e14c44629b837f7e74121e4bc7809 changed the `app` argument to `_app` in `modules.ts` and `plugins.ts`. This updates the generator templates to use the new `_app` variable.

### How was it tested?

I ran `lisk generate:module hello 1000` and was able to start my app.

### DIFF
--- a/commander/src/bootstrapping/templates/lisk-template-ts/generators/module_generator.ts
+++ b/commander/src/bootstrapping/templates/lisk-template-ts/generators/module_generator.ts
@@ -90,7 +90,7 @@ export default class ModuleGenerator extends Generator {
 			.getInitializerIfKindOrThrow(SyntaxKind.ArrowFunction);
 
 		registerFunction.setBodyText(
-			`${registerFunction.getBodyText()}\napp.registerModule(${this._moduleClass});`,
+			`${registerFunction.getBodyText()}\n_app.registerModule(${this._moduleClass});`,
 		);
 
 		modulesFile.organizeImports();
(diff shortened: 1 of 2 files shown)


markokeeffe:
I'm new to Lisk. I was going through the [Creating a module](https://lisk.com/documentation/lisk-sdk/guides/app-development/module.html) tutorial and ran into a `ReferenceError: app is not defined` error.

It took me a good while to figure out where the error was coming from. It would be extremely useful to show a stack trace when Node errors are thrown!

shuse2:
@markokeeffe Hello. Thank you for the contribution!
the `_app` was intentional because before any module or plugin is added, typescript will complain with unused variable.
If we change it to prepend `_`, it will skip this check.

Maybe, instead of prepending `_`, we could change it to use `app` as argument and add `typescript-ignore`?

shuse2:
Closing due to inactivity. However we will resolve #6909 in the next release

Issue #6940, opened 2021-11-25
Add limit reset functionality in consensus and generator network endpoint 


mitsuaki-u:
### Expected behavior

In consensus and generator network endpoint the rate limiting functionality should have a periodic resetting job. 
Timer shouldn't start at constructor.

### Actual behavior

The reset job is missing.

### Steps to reproduce

Check `node/consensus/network_endpoint/`  and `node/generator/network_endpoint/` 
https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/node/transport/transport.ts#L121-L124

### Which version(s) does this affect? (Environment, OS, etc...)

Feature branch

PR #6941, opened 2021-11-25
Add single commit event handler in consensus network layer - Closes #6848

### LINKED ISSUE #6848: Create network endpoint in consensus to handle single commits
### Description

We need to handle incoming commits as well as define schema for network communication. Add a commit handler inside `consensus/network_endpoint`.

```ts
public async handleEventSingleCommits(data: unknown, peerId: string): Promise<Buffer> {
// Verifies incoming single commit and bans the peer if certain validation fails according to LIP:Introduce a certificate generation mechanism
}

}
Where data is an array of SingleCommits,
interface SingleCommit {
	blockID: Buffer;
	height: number;
	validatorAddress: Buffer;
	certificateSignature: Buffer;
}

Where data is an array of SingleCommits,
interface SingleCommit {
	blockID: Buffer;
	height: number;
	validatorAddress: Buffer;
	certificateSignature: Buffer;
}
```

- Use `commitPool.validateCommit(commit)` from the certificate created in https://github.com/LiskHQ/lisk-sdk/issues/6841
- Apply penalty of 100 to the peerID if the above verification of certificate fails
- Apply penalty of 100 to the peerID if the schema check fails for incoming data

### Acceptance Criteria

- Add unit test for `handleEventSingleCommits` 
  - Pass valid single commits
  - Pass invalid single commits
  - Pass invalid data object

mitsuaki-u:
### What was the problem?

This PR resolves #6848 

Note: Rate limit reset will be added in issue #6940 

### How was it solved?

- 🌱 Add single commit handler a9c5dc24103fd5c1e2787743d7f87e2622aaf037
- 🌱 Add tests a776d5fbfdb47f158a47c9d8f99ac94631302f5c

### How was it tested?

Run unit test


framework/src/node/consensus/network_endpoint.ts
```diff
@@ -167,6 +175,67 @@ export class NetworkEndpoint {
...
+	public handleEventSingleCommit(data: unknown, peerId: string): void {
+		this._addRateLimit(
+			NETWORK_RPC_GET_SINGLE_COMMIT_FROM_ID,
+			peerId,
+			DEFAULT_SINGLE_COMMIT_FROM_IDS_RATE_LIMIT_FREQUENCY,
+		);
+		let decodedData: SingleCommit;
+
+		try {
+			decodedData = codec.decode<SingleCommit>(singleCommitSchema, data as never);
```
shuse2, line 187: Other endpoints have data schema defined, so it would be better to have the surrounding schema. ex)
```
{
  singleCommit: Buffer
}
```
like in https://github.com/LiskHQ/lisk-sdk/blob/feature/6554-improve-framework-architecture/framework/src/node/consensus/schema.ts#L96-L107

mehmetegemen (approved):
Apart from Shu's comment, LGTM 🎉 One note: I take no chances when it comes to testing, so if you find appropriate you can add tests for other invalid data scenarios to cover other parts of the schema changes in the future. Other than that very good job!

mitsuaki-u:
Any further tests would be testing the `ajv validator library` functionality. I think it is not necessary as this is tested adequately already in other areas of the code as well as the library itself.

mehmetegemen:
@mitsuaki-u You are partly right about that. My concern was testing schema changes in the future, like would we break something due to a schema change. But it's your call, was just a food for thought.

mitsuaki-u:
@mehmetegemen For the invalid data value tests the schema is not likely to change as the `blockID`, `validatorAddress` and `certificateSignature` bytes length are part of the protocol. If they do change in a future these properties are thoroughly tested throughout the code already and we will make all the corresponding changes in the tests at that time.

shuse2 (approved):

PR #6954, opened 2021-12-02
Fix SMT leaf value to be hashed before update - Closes 6900

### LINKED ISSUE #6900: Update Sparse Merkle Tree leaf value to be hashed
### Description

Stored value should be hashed as https://github.com/LiskHQ/lips/blob/main/proposals/lip-0040.md#specification specifies. 

When updating the value 
- https://github.com/LiskHQ/lisk-sdk/blob/feature/6554-improve-framework-architecture/elements/lisk-chain/src/state_store/cache_db.ts#L173-L191
- https://github.com/LiskHQ/lisk-sdk/blob/feature/6554-improve-framework-architecture/elements/lisk-chain/src/data_access/storage.ts#L420-L424

value is not hashed before setting to the smt

### Acceptance Criteria

- Ensure stored value is hashed

shuse2:
### What was the problem?

This PR resolves #6900

### How was it solved?

Add hash before updating the smt value

### How was it tested?

Updated tests


elements/lisk-chain/src/data_access/storage.ts
```diff
@@ -417,11 +417,11 @@ export class Storage {
 		// Revert all deleted values
 		for (const { key, value: previousValue } of deletedStates) {
 			batch.put(key, previousValue);
-			await smt.update(toSMTKey(key), previousValue);
+			await smt.update(toSMTKey(key), hash(previousValue));
```
ricott1, line 420: SHould we rather use a util function `toSTMValue(value)` to be more consistent?

ricott1 (changes requested):
One comment about using a more generic `toSMTValue` accessory function

elements/lisk-chain/src/data_access/storage.ts
shuse2: `toSTMKey` has logic where it hashes excluding first 6 bytes, while for value it's just simple hash, so i think it is ok like this

ricott1 (approved):

Incede (approved):

PR #6955, opened 2021-12-02
Implement add, validate commits and getCommitsByHeight methods - Closes #6849

### LINKED ISSUE #6849: Implement add/validate commit functions of CommitPool
### Description

- Implement `addCommit(commit)` function where it adds a commit to `_nonGossipedCommits` data structure
- Implement `validateCommit(commit)` function where it adds or rejects or throws validation error (for banning) if it fails. Refer to steps of validation provided in [LIP:SingleCommits:Gossip](https://research.lisk.com/t/introduce-certificate-generation-mechanism/296#single-commit-p2p-gossip-37)
- Implement `getCommitsByHeight (height: number)` where it returns commits from `_nonGossipedCommits` and `_gossipedCommits` for that height.

```ts
public getCommitsByHeight (height: number) {
		return [...this._nonGossipedCommits[height], ...this._gossipedCommits[height]];
	}
```

### Acceptance Criteria

- Add unit test for `addCommit`, `getCommitsByHeight` and `validateCommit`

mehmetegemen:
### What was the problem?

This PR resolves #6849 

### How was it solved?

- Implement addCommit, validateCommit and getCommitsByHeight methods

### How was it tested?

- Not yet tested

### DIFF
--- a/framework/src/node/consensus/certificate_generation/commit_pool.ts
+++ b/framework/src/node/consensus/certificate_generation/commit_pool.ts
@@ -17,6 +17,7 @@ import { EMPTY_BUFFER } from './constants';
 import { BFTParameterNotFoundError } from '../../../modules/bft/errors';
 import { APIContext } from '../../state_machine/types';
 import { BFTAPI, ValidatorAPI } from '../types';
+import { COMMIT_RANGE_STORED } from './constants';
 import {
 	AggregateCommit,
 	Certificate,
@@ -29,6 +30,7 @@ import {
 	verifyAggregateCertificateSignature,
 	getSortedWeightsAndValidatorKeys,
 	signCertificate,
+	verifySingleCertificateSignature,
 } from './utils';
 
 export class CommitPool {
@@ -62,15 +64,103 @@ export class CommitPool {
 
 	// eslint-disable-next-line @typescript-eslint/no-empty-function
 	public async job(): Promise<void> {}
-	// eslint-disable-next-line @typescript-eslint/no-empty-function
-	public addCommit(_commit: SingleCommit, _height: number): void {}
-	// eslint-disable-next-line @typescript-eslint/no-empty-function
-	public validateCommit(_commit: SingleCommit): boolean {
+
+	public addCommit(commit: SingleCommit, height: number): void {
+		const currentCommits = this._nonGossipedCommits.get(height) ?? [];
+		const doesCommitExist = currentCommits.some(aCommit =>
+			aCommit.certificateSignature.equals(commit.certificateSignature),
+		);
+		if (!doesCommitExist) {
+			this._nonGossipedCommits.set(height, [...currentCommits, commit]);
+		}
+	}
+
+	public async validateCommit(apiContext: APIContext, commit: SingleCommit): Promise<boolean> {
+		// Validation step 4
+		const blockHeaderAtCommitHeight = await this._chain.dataAccess.getBlockHeaderByHeight(
+			commit.height,
+		);
+		if (!blockHeaderAtCommitHeight.id.equals(commit.blockID)) {
+			return false;
+		}
+
+		// Validation step 1
+		const doesCommitExistsInNonGossipedCommits = !!this._nonGossipedCommits
+			.get(commit.height)
+			?.some(
+				nonGossipedCommit =>
+					nonGossipedCommit.blockID === commit.blockID &&
+					nonGossipedCommit.validatorAddress === commit.validatorAddress,
+			);
+
+		const doesCommitExistsInGossipedCommits = !!this._gossipedCommits
+			.get(commit.height)
+			?.some(
+				gossipedCommit =>
+					gossipedCommit.blockID === commit.blockID &&
+					gossipedCommit.validatorAddress === commit.validatorAddress,
+			);
+
+		const doesCommitExist =
+			doesCommitExistsInGossipedCommits || doesCommitExistsInNonGossipedCommits;
+
+		if (doesCommitExist) {
+			return false;
+		}
+
+		// Validation Step 2
+		const maxRemovalHeight = await this._getMaxRemovalHeight();
+		if (commit.height <= maxRemovalHeight) {
+			return false;
+		}
+
+		// Validation Step 3
+		const { maxHeightPrecommitted } = await this._bftAPI.getBFTHeights(apiContext);
+		const isCommitInRange =
+			commit.height >= maxHeightPrecommitted - COMMIT_RANGE_STORED &&
+			commit.height < maxHeightPrecommitted;
+		const doesBFTParamExistForNextHeight = await this._bftAPI.existBFTParameters(
+			apiContext,
+			commit.height + 1,
+		);
+		if (!isCommitInRange && !doesBFTParamExistForNextHeight) {
+			return false;
+		}
+
+		// Validation Step 5
+		const { validators } = await this._bftAPI.getBFTParameters(apiContext, commit.height);
+		const isCommitValidatorActive = validators.find(validator =>
+			validator.address.equals(commit.validatorAddress),
+		);
+		if (!isCommitValidatorActive) {
+			throw new Error('Commit validator was not active for its height.');
+		}
+
+		// Validation Step 6
+		const certificate = computeCertificateFromBlockHeader(blockHeaderAtCommitHeight);
+		const { blsKey } = await this._validatorsAPI.getValidatorAccount(
+			apiContext,
+			commit.validatorAddress,
+		);
+		const { networkIdentifier } = this._chain;
+		const isSingleCertificateVerified = verifySingleCertificateSignature(
+			blsKey,
+			commit.certificateSignature,
+			networkIdentifier,
+			certificate,
+		);
+
+		if (!isSingleCertificateVerified) {
+			throw new Error('Certificate signature is not valid.');
+		}
+
 		return true;
 	}
-	// eslint-disable-next-line @typescript-eslint/no-empty-function
-	public getCommitsByHeight(_height: number): SingleCommit[] {
-		return [];
+
+	public getCommitsByHeight(height: number): SingleCommit[] {
+		const nonGossipedCommits = this._nonGossipedCommits.get(height) ?? [];
+		const gossipedCommits = this._gossipedCommits.get(height) ?? [];
+		return [...nonGossipedCommits, ...gossipedCommits];
 	}
 
 	public createSingleCommit(
@@ -235,5 +325,12 @@ export class CommitPool {
 			aggregationBits: EMPTY_BUFFER,
 			certificateSignature: EMPTY_BUFFER,
 		};
+	}
+
+	private async _getMaxRemovalHeight() {
+		const blockHeader = await this._chain.dataAccess.getBlockHeaderByHeight(
+			this._chain.finalizedHeight,
+		);
+		return blockHeader.aggregateCommit.height;
 	}
 }
--- a/framework/src/node/consensus/certificate_generation/commit_pool.ts
+++ b/framework/src/node/consensus/certificate_generation/commit_pool.ts
@@ -89,16 +89,16 @@ export class CommitPool {
 			.get(commit.height)
 			?.some(
 				nonGossipedCommit =>
-					nonGossipedCommit.blockID === commit.blockID &&
-					nonGossipedCommit.validatorAddress === commit.validatorAddress,
+					nonGossipedCommit.blockID.equals(commit.blockID) &&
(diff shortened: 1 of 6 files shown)


framework/src/node/consensus/certificate_generation/commit_pool.ts
```diff
@@ -62,15 +64,103 @@ export class CommitPool {
...
+		if (!blockHeaderAtCommitHeight.id.equals(commit.blockID)) {
+			return false;
+		}
+
+		// Validation step 1
+		const doesCommitExistsInNonGossipedCommits = !!this._nonGossipedCommits
+			.get(commit.height)
+			?.some(
+				nonGossipedCommit =>
+					nonGossipedCommit.blockID === commit.blockID &&
```
ishantiw, line 92: use `.equals` for buffers

framework/src/node/consensus/certificate_generation/commit_pool.ts
```diff
@@ -62,15 +64,103 @@ export class CommitPool {
...
+					nonGossipedCommit.validatorAddress === commit.validatorAddress,
```
ishantiw, line 93: use `.equals`, also in other places

framework/src/node/consensus/certificate_generation/commit_pool.ts
```diff
@@ -62,15 +64,103 @@ export class CommitPool {
...
+			commit.height,
+		);
+		if (!blockHeaderAtCommitHeight.id.equals(commit.blockID)) {
+			return false;
+		}
+
+		// Validation step 1
+		const doesCommitExistsInNonGossipedCommits = !!this._nonGossipedCommits
+			.get(commit.height)
+			?.some(
```
ishantiw, line 90: shouldn't we use `.find` as we always have unique commits for a height (we always discard duplicates while adding)

framework/src/node/consensus/certificate_generation/commit_pool.ts
```diff
@@ -62,15 +64,103 @@ export class CommitPool {
...
+		// Validation step 4
+		const blockHeaderAtCommitHeight = await this._chain.dataAccess.getBlockHeaderByHeight(
+			commit.height,
+		);
+		if (!blockHeaderAtCommitHeight.id.equals(commit.blockID)) {
+			return false;
+		}
+
+		// Validation step 1
+		const doesCommitExistsInNonGossipedCommits = !!this._nonGossipedCommits
```
ishantiw, line 88: since the func name is validateCommit we can avoid `commit` keyword, we can simplify name to `existsInNonGossiped` and `existsInGossiped`

framework/src/node/consensus/certificate_generation/commit_pool.ts
```diff
@@ -62,15 +64,103 @@ export class CommitPool {
...
+		const maxRemovalHeight = await this._getMaxRemovalHeight();
+		if (commit.height <= maxRemovalHeight) {
+			return false;
+		}
+
+		// Validation Step 3
+		const { maxHeightPrecommitted } = await this._bftAPI.getBFTHeights(apiContext);
+		const isCommitInRange =
+			commit.height >= maxHeightPrecommitted - COMMIT_RANGE_STORED &&
+			commit.height < maxHeightPrecommitted;
```
ishantiw, line 121: shouldn't this be less than equal?
```suggestion
			commit.height <= maxHeightPrecommitted;
```

framework/src/node/consensus/certificate_generation/commit_pool.ts
```diff
@@ -62,15 +64,103 @@ export class CommitPool {
...
+		const doesBFTParamExistForNextHeight = await this._bftAPI.existBFTParameters(
+			apiContext,
+			commit.height + 1,
+		);
+		if (!isCommitInRange && !doesBFTParamExistForNextHeight) {
+			return false;
+		}
+
+		// Validation Step 5
```
ishantiw, line 130: Validation step 4 is missing

framework/src/node/consensus/network_endpoint.ts
```diff
@@ -219,8 +225,12 @@ export class NetworkEndpoint {
 			throw error;
 		}
 
+		let isValidCommit: boolean;
 		try {
-			this._commitPool.validateCommit(decodedData.singleCommit);
+			isValidCommit = await this._commitPool.validateCommit(
+				this._apiContext,
+				decodedData.singleCommit,
+			);
```
ishantiw, line 233:
```suggestion
			const isValidCommit = await this._commitPool.validateCommit(
				this._apiContext,
				decodedData.singleCommit,
			);
			if (!isValidCommit) { return }
```

ishantiw: you can return within try block only

framework/test/unit/node/consensus/certificate_generation/commit_pool.spec.ts
```diff
@@ -89,13 +95,373 @@ describe('CommitPool', () => {
...
+
+		it('should return false when single commit exists in gossiped commits but not in non-gossipped commits', async () => {
+			commitPool['_gossipedCommits'].set(commit.height, [commit]);
+
+			const isCommitValid = await commitPool.validateCommit(apiContext, commit);
+
+			expect(isCommitValid).toBeFalse();
+		});
+
+		it('should return false when single commit exists in non-gossiped commits but not in gossipped commits', async () => {
```
ishantiw, line 266:
```suggestion
		it('should return false when single commit already exists in non-gossiped commits', async () => {
```

framework/test/unit/node/consensus/certificate_generation/commit_pool.spec.ts
```diff
@@ -89,13 +95,373 @@ describe('CommitPool', () => {
...
+			commitPool['_nonGossipedCommits'].set(commit.height, [commit]);
+
+			const isCommitValid = await commitPool.validateCommit(apiContext, commit);
+
+			expect(isCommitValid).toBeFalse();
+		});
+
+		it('should return false when single commit exists both in gossiped commits and in non-gossipped commits', async () => {
```
ishantiw, line 274: This should never be the case when a commit exists in both gossiped and non-gossiped as we always make this sure while adding to commit pool

ishantiw (changes requested):

framework/src/node/consensus/certificate_generation/commit_pool.ts
mehmetegemen: it's at the beginning  of the validaCommit method 😄 (As per Jan's suggestion). May you check that the combination of validation step 4 and validation step 2 in the code is valid for validation step 2 in the LIP? I am not 100% certain about it.

framework/src/node/consensus/network_endpoint.ts
mehmetegemen: We need Ishan as a Service as the world, amazing as always 😄 Will do.

framework/src/node/consensus/certificate_generation/commit_pool.ts
mehmetegemen: It's hard to understand end range in the LIP is included or not, are you sure about this?

framework/test/unit/node/consensus/certificate_generation/commit_pool.spec.ts
mehmetegemen: Will remove.

framework/test/unit/node/consensus/certificate_generation/commit_pool.spec.ts
mehmetegemen: Will redact.

framework/src/node/consensus/certificate_generation/commit_pool.ts
mehmetegemen: Great advice, will do.

framework/src/node/consensus/certificate_generation/commit_pool.ts
mehmetegemen: I wanted something to return boolean instead of the item itself. Since we cast the result to boolean by double negation, it actually doesn't matter. If you really think I should change it to `find`, I can.

framework/src/node/consensus/certificate_generation/commit_pool.ts
mehmetegemen: Will do.

framework/src/node/consensus/certificate_generation/commit_pool.ts
ishantiw: to my understanding, 
```ts
{bftModule.getBFTHeights().maxHeightPrecommitted - COMMIT_RANGE_STORED, ..., bftModule.getBFTHeights().maxHeightPrecommitted}
```
curly brackets `{ }` signify the set of values and the LIP says `Discard m if m.height is not in` the above range so I think its included. But we can clarify with @janhack

framework/src/node/consensus/certificate_generation/commit_pool.ts
ishantiw: ah missed out the validation step in the beginning, I think step 2 and step 4 can be doesn't need a combination so its fine

framework/src/node/consensus/certificate_generation/commit_pool.ts
janhack: Ishan's suggestion is correct. The notation {2, ..., 5} means the values 2, 3, 4, 5 so including the bounds.

janhack: I noticed that the formatting of the relevant section in LIP 61 was broken and fixed it in the PR https://github.com/LiskHQ/lips/pull/125.

framework/src/node/consensus/certificate_generation/commit_pool.ts
mehmetegemen: done

framework/src/node/consensus/certificate_generation/commit_pool.ts
mehmetegemen: done

framework/src/node/consensus/certificate_generation/commit_pool.ts
mehmetegemen: done

framework/src/node/consensus/certificate_generation/commit_pool.ts
mehmetegemen: done

framework/src/node/consensus/network_endpoint.ts
mehmetegemen: done

framework/test/unit/node/consensus/certificate_generation/commit_pool.spec.ts
mehmetegemen: done

framework/test/unit/node/consensus/certificate_generation/commit_pool.spec.ts
mehmetegemen: Decided it is better to leave it as it is

framework/test/unit/node/consensus/certificate_generation/commit_pool.spec.ts
```diff
@@ -89,13 +95,364 @@ describe('CommitPool', () => {
...
+			const blockID = getRandomBytes(32);
+
+			height = 1031;
+
+			nonGossipedCommits = Array.from({ length: 1 }, () => ({
+				blockID,
+				certificateSignature: getRandomBytes(96),
+				height,
+				validatorAddress: getRandomBytes(20),
+			}));
```
Incede, line 111: I think constants can be moved to `describe` scope.

framework/test/unit/node/consensus/certificate_generation/commit_pool.spec.ts
```diff
@@ -333,6 +690,39 @@ describe('CommitPool', () => {
...
+
+			blockHeader = createFakeBlockHeader({
+				height: finalizedHeight,
+				timestamp: finalizedHeight * 10,
+				aggregateCommit: {
+					aggregationBits: Buffer.alloc(0),
+					certificateSignature: Buffer.alloc(0),
+					height: finalizedHeight,
+				},
+			});
```
Incede, line 708: same, prefer to only keep values in `beforeEach` which need to be initialized independently for each test

Incede (changes requested):

Incede (approved):
Although I feel the scope definitions in tests should follow the best practices, approving from my end.

PR #6956, opened 2021-12-02
Update all modules to have default config - Closes #6953

### LINKED ISSUE #6953: Update all module config to have default value
### Description

Create default value for all of the modules config using the value which are used in the lisk mainchain.

### Motivation

Without default value set in the module, developer will need to specify a lot of config value to setup the application

### Additional Information
Module should be able to run even if no config is provided from the application

shuse2:
### What was the problem?

This PR resolves #6953

### How was it solved?

- Create `defaultConfig` on each module, and overwrite with the given config
- Remove all outdated modules

### How was it tested?

- Add unit test to check if init resolves without any config given


mitsuaki-u (approved):
Other than build fail LGTM!

PR #6957, opened 2021-12-02
Update genesis block create command - Closes #6944

### LINKED ISSUE #6944: Update genesis-block:create command
### Description

Update genesis-block:create command to follow new genesis block.

```
genesis-block:create --output mydir
genesis-block:create --output mydir --assets-file path/to/file
```
if `assets-file` is added, it should read the JSON file, validate if it contains below structure. Use the assets file to generate the genesis block, and save only the genesis block
```
{
  assets: {
    moduleID: number;
    data: Record<string, unknown>;
    schema: Schema,
  }[],
}
```
if assets file is not present, check if token/dpos module exist as it does now, and create accounts with token and validators, and return the created accounts, generator info and password

### Acceptance Criteria

- Genesis block is created and can be used for running an application
- test is updated

shuse2:
### What was the problem?

This PR resolves #6944 

### How was it solved?

Update genesis-block:create command template
- Add `network` flag to select which config to use as base
- Add `--config` flag to be able to specify the config to overwrite
- Add `--assets-file` to specify JSON file for assets to create genesis block

### How was it tested?

1. create new template by `./bin/run init ../examples/new-app` (this will fail since packages are not uploaded
2. cd ../exmaples/new-app
3. ./bin/run genesis-block:create


ishantiw (approved):
LGTM

Incede (approved):

PR #6971, opened 2022-01-05
Update generator to listen on BFT parameters for creating commits - Closes #6851

### LINKED ISSUE #6851: Update generator to listen on BFT parameters for creating commits
### Description

In generator, add logic to call function `consensus.commitPool.createSingleCommit` as specified in #6844
[LIP:Commit:create:afterBlockProcessing](https://research.lisk.com/t/introduce-certificate-generation-mechanism/296#creation-after-block-processing-40)

#### Block Creation
- For every height `h` in `{h1+1, h1+2, ..., h2}` such that `bftModule.existBFTParameters(h+1)` returns `True` (i.e., `h` is the height of a block authenticating a change of BFT parameters), validator `v` creates a commit message for the block `b` at height `h` in its chain if validator `v` was active at height `h`.
- If `bftModule.existBFTParameters(h2+1)` returns `False` (i.e., `h2` is not the height of a block authenticating a change of BFT parameters), `v` creates a commit message for the block `b` at height `h2` in its chain if validator `v` was active at height `h`.

### On node restart,
In `generator.init()`,
- Create below function in consensus, and will be called by generator, `consensus.getMaxRemovalHeight()`
```
getMaxRemovalHeight():
    b = block at height maxHeightFinalized
    return b.header.aggregateCommit.height
```
Let `h1 = getMaxRemovalHeight() and h2 = bftModule.getBFTHeights().maxHeightPrecommitted`. Then the single commits are created as described in the above section.

### Acceptance Criteria

- Add unit test for create commits

mehmetegemen:
### What was the problem?

This PR resolves #6851 

### How was it solved?

- Added an event listener to generator to act on finalized height changes
- Added related interface changes for above change

### How was it tested?

- Not tested yet


framework/src/node/generator/generator.ts
```diff
@@ -241,6 +243,37 @@ export class Generator {
 				),
 			);
 		});
+
+		this._consensus.events.on(
+			CONSENSUS_EVENT_FINALIZED_HEIGHT_CHANGED,
+			// eslint-disable-next-line @typescript-eslint/no-misused-promises
+			async ({ from, to }: { from: number; to: number }) => {
```
shuse2, line 250: event second parameter shouldn't be async. it should wrap with Promise/then-error

framework/src/node/generator/generator.ts
```diff
@@ -241,6 +243,37 @@ export class Generator {
...
+				for (const [address, _] of this._keypairs.entries()) {
+					for (let height = from + 1; height < to + 1; height += 1) {
+						const blockHeader = await this._chain.dataAccess.getBlockHeaderByHeight(height);
+						const blsSecretKey = this._endpoint.blsKeys.get(address);
```
shuse2, line 254: similar to keyPair, blsKeys should be managed in generator as well. it should be initialized on `load` similar to keypair

framework/src/node/generator/generator.ts
```diff
@@ -241,6 +243,37 @@ export class Generator {
...
+							);
+						}
+
+						const validatorInfo = {
+							address,
+							blsPublicKey: getPublicKeyFromPrivateKey(blsSecretKey),
+							blsSecretKey,
+						};
+
+						const singleCommit = this._consensus.commitPool.createSingleCommit(
```
shuse2, line 268: where is this single commit used?

shuse2 (changes requested):
general approach LGTM

framework/src/node/generator/generator.ts
mehmetegemen: But `_loadGenerators` is used for debugging purposes unlike `updateForgingStatus` in the endpoint that is for prod. May you elaborate more on your comment?

framework/src/node/generator/generator.ts
mehmetegemen: I will add `addCommit` as we spoke.

framework/src/node/generator/generator.ts
mehmetegemen: You mean remove `async` from callback and then `Promise.all().then().catch()` inside the sync callback to use whatever values we need?

shuse2: im not sure about promise all, but yes. we shouldn't use `async` for the event

framework/src/node/generator/generator.ts
shuse2: i think it should behave the same as the generation key handling. it is used for testing in the network, and it is a feature we want to support as well

Incede (changes requested):
Looks good, minor suggestions.

Incede (approved):

shuse2 (approved):

PR #6977, opened 2022-01-13
Add rate limit reset job in generator and consensus - Closes #6940

### LINKED ISSUE #6940: Add limit reset functionality in consensus and generator network endpoint

ishantiw:
### What was the problem?

This PR resolves #6940 

### How was it solved?

♻️ Add start/stop job for rateLimit reset job 870f9d631b55cb84418287a1e97e0b0c2f492ace

### How was it tested?

`npm run test:unit` under `framework`

### DIFF
--- a/framework/src/node/consensus/consensus.ts
+++ b/framework/src/node/consensus/consensus.ts
@@ -308,7 +308,8 @@ export class Consensus {
 		});
 	}
 
-	public start(): void {
+	// eslint-disable-next-line @typescript-eslint/require-await
+	public async start(): Promise<void> {
 		this._endpoint.start();
 	}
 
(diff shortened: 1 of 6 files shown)


framework/src/node/consensus/consensus.ts
```diff
@@ -128,6 +128,7 @@ export class Consensus {
 			db: this._db,
 		} as EndpointArgs); // TODO: Remove casting in issue where commitPool is added here
 		const blockExecutor = this._createBlockExecutor();
+		this._endpoint.start();
```
shuse2, line 131: I think adding `start` in consensus and calling it from node would be more consistent

shuse2 (changes requested):

shuse2 (approved):

mitsuaki-u (approved):
Other than @shuse2 comment and unit test LGTM

PR #7073, opened 2022-03-14
Fix random module to use random hashOnion when config is not set - Closes #7072

### LINKED ISSUE #7072: Seed reveal should be generated even if config does not exist
### Expected behavior
When generator config doesn't exist for a particular generator, it should create random seed reveal although it will not receive reward.
This is the case already when generator runs out of hash onions.

### Actual behavior
It throws error if generator config does not exist.

### Steps to reproduce
Try to run a block without generator config

development-

shuse2:
### What was the problem?

This PR resolves #7072 

### How was it solved?

- Update to return undefined when hash onion config does not exist
- Update to use random seed reveal when hash onion is not set to the config

### How was it tested?

- Update unit test to check this case

### DIFF
--- a/framework/src/modules/random/module.ts
+++ b/framework/src/modules/random/module.ts
@@ -256,6 +256,12 @@ export class RandomModule extends BaseModule {
 			return prev;
 		}, undefined);
 		const hashOnionConfig = this._getHashOnionConfig(address);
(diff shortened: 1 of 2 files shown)


IkerAlus (approved):
LGTM

ishantiw (approved):

PR #7074, opened 2022-03-14
Update to use chain validate for static validation - Closes #7056

### LINKED ISSUE #7056: Fix block validation in consensus
### Expected behavior
Block size should be checked as part of static validation.
Instead of `block.validate()` it shoud call `chain.verifyAsset` function.
verifyAsset function should be renamed to `validateBlock`

### Actual behavior
https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/node/consensus/consensus.ts#L449 (There are several places in this file and synchronization)

### Steps to reproduce
Any block size will be accepted

development

shuse2:
### What was the problem?

This PR resolves #7056 

### How was it solved?

- Update to use `chain.validateBlock` instead of `Block.validate` in consensus
- Move `version` and `assets` validation to `validateBlock`
- Fix synchronization code to use `validate` when receiving block.
- Fix misuse of `verify` in synchronization. When validating detached block (block which is not consecutive of the current last block), `verify` cannot be used. Therefore, updated to use `validate`

Reference for synchronization: https://github.com/LiskHQ/lips/blob/main/proposals/lip-0014.md#moving-to-a-different-chain

### How was it tested?

- Updated unit tests

### DIFF
--- a/elements/lisk-chain/src/chain.ts
+++ b/elements/lisk-chain/src/chain.ts
@@ -15,7 +15,7 @@
 import { codec } from '@liskhq/lisk-codec';
 import { KVStore, NotFoundError } from '@liskhq/lisk-db';
 import * as createDebug from 'debug';
-import { MerkleTree } from '@liskhq/lisk-tree';
+import { regularMerkleTree } from '@liskhq/lisk-tree';
 import {
 	DEFAULT_MAX_BLOCK_HEADER_CACHE,
 	DEFAULT_MIN_BLOCK_HEADER_CACHE,
@@ -43,6 +43,11 @@ interface ChainInitArgs {
 	readonly db: KVStore;
 	readonly networkIdentifier: Buffer;
 	readonly genesisBlock: Block;
+}
+
+interface BlockValidationInput {
+	readonly version: number;
+	readonly acceptedModuleIDs: number[];
 }
 
 const debug = createDebug('lisk:chain');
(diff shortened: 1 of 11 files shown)


elements/lisk-chain/src/chain.ts
```diff
@@ -156,8 +161,16 @@ export class Chain {
...
 
-	public async verifyAssets(block: Block): Promise<void> {
+	public validateBlock(block: Block, inputs: BlockValidationInput): void {
 		block.validate();
+		if (block.header.version !== inputs.version) {
+			throw new Error(`Block version must be ${inputs.version}.`);
+		}
+		for (const asset of block.assets.getAll()) {
+			if (!inputs.acceptedModuleIDs.includes(asset.moduleID)) {
+				throw new Error(`Block asset from moduleID: ${asset.moduleID} is not accepted.`);
```
Incede, line 171:
```suggestion
				throw new Error(`Block asset with moduleID: ${asset.moduleID} is not accepted.`);
```

elements/lisk-chain/test/unit/chain.spec.ts
```diff
@@ -316,7 +317,7 @@ describe('chain', () => {
 		});
 	});
 
-	describe('verifyAssets', () => {
+	describe('validateBlock', () => {
```
Incede, line 320: Maybe also add an all valid test case

Incede (approved):
Changes look good, some unit tests need to be updated.

ishantiw (approved):

PR #7076, opened 2022-03-15
Implement `appendToOutboxTree` and `addToOutbox` - Closes #7042 

### LINKED ISSUE #7042: Implement addToOutbox and appendToOutboxTree internal functions
### Description

Add `addToOutbox` and `appendToOutboxTree` functionaility to interop store.

```
addToOutbox(chainID, ccm):
    serializedMessage = byte array corresponding to the serialized ccm according to the schema in LIP 0049
    appendToOutboxTree(chainID, serializedMessage)
    outboxRoot(chainID) = channel(chainID).outbox.root

    return True
```

### Acceptance Criteria

- Add unit test for `addToOutbox`
- Add unit test for `appendToOutboxTree`

mitsuaki-u:
### What was the problem?

This PR resolves #7042

### How was it solved?

- Added append to outbox and add to outbox functionality and tests

### How was it tested?

Added unit tests


framework/src/modules/interoperability/mainchain/store.ts
```diff
@@ -12,9 +12,123 @@
...
+		await channelSubstore.setWithSchema(
+			chainID,
+			{ ...channel, outbox: updatedOutbox },
+			channelSchema,
+		);
+
+		return true;
+	}
+
+	public async addToOutbox(chainID: Buffer, ccm: CCMsg) {
```
shuse2, line 45:
```
    # Set nonce in CCM only if sending from the chain (not if routing)
    if ccm.sendingChainID == ownChainAccount.ID:
        ccm.nonce = ownChainAccount.nonce
        ownChainAccount.nonce += 1
```
seems to be missing in the logic. any reason for this?

framework/test/unit/modules/interoperability/mainchain/store.spec.ts
```diff
@@ -0,0 +1,114 @@
...
+
+	describe('appendToOutboxTree', () => {
+		it('should update the channel store with the new outbox tree info', async () => {
+			// Act
+			await mainchainInteroperabilityStore.appendToOutboxTree(chainID, appendData);
+
+			// Assert
+			expect(channelSubstore.setWithSchema).toHaveBeenCalledWith(
+				chainID,
+				channelData,
```
shuse2, line 95: is this expectation correct? in L69, it's returning the same object

framework/src/modules/interoperability/mainchain/store.ts
mitsuaki-u: Where do you see that logic? That might be old and outdated.

shuse2: ah yes, i was looking at the old version. sorry!

framework/test/unit/modules/interoperability/mainchain/store.spec.ts
mitsuaki-u: It's correct because on L69 the updated merkleroot is returning the `channelData` so we can assume it is updated. For convenience it's only referencing the same data in the test.

shuse2: i think it's better not to mutate the test setup and expect it be equal because the expectation is ambiguous, and the test will not be atomic. (for example, add another test which uses the `channelData`, it will not work as expected)

mitsuaki-u: There is no mutation but I agree the expectation was not clear. Updated to return a different outbox value.

framework/src/modules/interoperability/base_interoperability_store.ts
```diff
@@ -40,8 +40,8 @@ export abstract class BaseInteroperabilityStore {
 	// To be implemented in base class
 	public abstract apply(ccu: CCUpdateParams, ccm: CCMsg): Promise<void>;
 	public abstract appendToInboxTree(chainID: number, appendData: Buffer): Promise<void>;
-	public abstract appendToOutboxTree(chainID: number, appendData: Buffer): Promise<void>;
-	public abstract addToOutbox(chainID: Buffer, ccm: CCMsg): Promise<void>;
+	public abstract appendToOutboxTree(chainID: Buffer, appendData: Buffer): Promise<boolean>;
+	public abstract addToOutbox(chainID: Buffer, ccm: CCMsg): Promise<boolean>;
```
ishantiw, line 44: It was not mentioned in the issue but if you can move the implementations of `appendToOutboxTree` and `addToOutbox` to `BaseInteroperabilityStore` class as the code will be the same for mainchain and sidechain.

ishantiw: Similar to https://github.com/LiskHQ/lisk-sdk/blob/72a8ede36f90bb334b386d91828a5c4766b5434e/framework/src/modules/interoperability/base_interoperability_store.ts#L48-L70

ishantiw (commented):
LGTM, only comment for moving these functions to `BaseInteroperabilityClass` as done recently for other functions.

framework/test/unit/modules/interoperability/store.spec.ts
```diff
@@ -0,0 +1,122 @@
...
+import { when } from 'jest-when';
+import {
+	MODULE_ID_INTEROPERABILITY,
+	STORE_PREFIX_CHANNEL_DATA,
+	STORE_PREFIX_OUTBOX_ROOT,
+} from '../../../../src/modules/interoperability/constants';
+import { MainchainInteroperabilityStore } from '../../../../src/modules/interoperability/mainchain/store';
+import { channelSchema, outboxRootSchema } from '../../../../src/modules/interoperability/schema';
+
+describe('Mainchain interoperability store', () => {
```
ishantiw, line 25:
```suggestion
describe('Base interoperability store', () => {
```

ishantiw (approved):
LGTM 🎉

framework/src/modules/interoperability/base_interoperability_store.ts
```diff
@@ -45,6 +57,36 @@ export abstract class BaseInteroperabilityStore {
...
+			appendPath: channel.outbox.appendPath,
+			size: channel.outbox.size,
+		});
+		await channelSubstore.setWithSchema(
+			chainID,
+			{ ...channel, outbox: updatedOutbox },
+			channelSchema,
+		);
+
+		return true;
```
shuse2, line 74: why is this returning `true`? same as below

mitsuaki-u: The LIP states to return true.

shuse2 (approved):

PR #7082, opened 2022-03-22
Implement Validators and DPoS updates - Closes #7004

### LINKED ISSUE #7004: Update Validators & Dpos module to match LIP updates
### Description

This [PR](https://github.com/LiskHQ/lips/pull/126) updates LIP44 & LIP57.
The implementation should be updated accordingly.

### Motivation

The implementation should be up to date with the specification.

### Additional Information
Logic & usage related to `getGeneratorsBetweenTimestamps` & `getValidatorAccount` needs to be updated.

mehmetegemen:
### What was the problem?

This PR resolves #7004 

### How was it solved?

- Updated `getValidatorsBetweenTimestamps` and `getValidatorAccount`
- DPoS `updateProductivity` now doesn't subtract 1 from beginning missed slot

### How was it tested?

- `getValidatorsBetweenTimestamps` unit tests redacted and added 2 new unit tests
- added 2 new `getValidatorAccount` unit tests
- DPoS `updateProductivity` tests are changed


framework/test/unit/modules/dpos_v2/module.spec.ts
```diff
@@ -1051,6 +1047,8 @@ describe('DPoS module', () => {
 					if (delegateAddress.equals(generatorAddress)) {
 						expect(currentDelegate.consecutiveMissedBlocks).toBe(0);
 						expect(currentDelegate.lastGeneratedHeight).toBe(nextForgedHeight);
+					} else if (delegateAddress.equals(delegateAddresses[0])) {
```
shuse2, line 1050: why do we need to handle this separately? the tests says all the delegates must have missedBlocks 1

framework/test/unit/modules/dpos_v2/module.spec.ts
```diff
@@ -1192,6 +1181,8 @@ describe('DPoS module', () => {
 						missedMoreThan1Block.some(missedForger => missedForger.equals(delegateAddress))
 					) {
 						expect(currentDelegate.consecutiveMissedBlocks).toBe(2);
+					} else if (delegateAddress.equals(lastForgerAddress)) {
```
shuse2, line 1184: same question as above

framework/test/unit/modules/dpos_v2/module.spec.ts
mehmetegemen: It was a misunderstanding, I omitted start and end slots thinking a round, but startSlot is previous round's last block, so corrected now. Please check last commit.

framework/test/unit/modules/dpos_v2/module.spec.ts
mehmetegemen: Logic of this test was broken since `lastForgerAddress` also misses 1 block, I omitted start and end slots of the range in my new commit([d8c1a8d](https://github.com/LiskHQ/lisk-sdk/pull/7082/commits/d8c1a8db69694350d1425353e60bc15d7c158a6d)), and fixed the logic, please see the commit.

framework/test/unit/modules/dpos_v2/module.spec.ts
```diff
@@ -1020,7 +1020,7 @@ describe('DPoS module', () => {
 
 				const missedBlocks: Record<string, number> = {};
 				// Make every delegate miss its block-slot except start and end slots
-				for (let i = 1; i < 102; i += 1) {
+				for (let i = 0; i < 102; i += 1) {
```
shuse2, line 1023: I think the comment and the logic doesn't match here?

framework/test/unit/modules/dpos_v2/module.spec.ts
```diff
@@ -1128,6 +1126,9 @@ describe('DPoS module', () => {
 				const currentTimestamp = 10290;
 				const lastForgedHeight = 926;
 				const nextForgedHeight = lastForgedHeight + 1;
+				// Last block slot is not included because it is currently generated
+				// and is not related to this test case
+				const delegatesExceptGenerator = delegateAddresses.slice(0, 102);
```
shuse2, line 1131: wouldn't it better if we check generator has 0 consecutiveMissedBlocks?

framework/test/unit/modules/dpos_v2/module.spec.ts
mehmetegemen: It actually matches, that confused me in the first place. The forging starts from last round's last block which is omitted as start slot. So if we start from current round's beginning block we omit start slot if we think context is bounded to `getValidatorsBetweenTimestamps`, it's a matter of context. I choosed this context because `missedBlocks` is related to `getValidatorsBetweenTimestamps`. Do you have another advice on that?

framework/test/unit/modules/dpos_v2/module.spec.ts
mehmetegemen: It was originally checking, I thought you didn't want it. Setting `consecutiveMissedBlocks` to 0 is a part of the unit since it's in the `_updateProductivity` as a simple assignment, so I think yes it would be good to check it to be sure about `_updateProductivity` unit's functioning as a whole. Do you approve it?

framework/test/unit/modules/dpos_v2/module.spec.ts
shuse2: I think 
> `missedBlocks` is related to `getValidatorsBetweenTimestamps`

this assumption is correct, but the loop executes 102 times where it should be 101 times if we dont include start and end?

mehmetegemen: Start: previous round's last block
End: current round's last block
= 104 blocks(1 + 103). Omit `Start` and `End`, you have 102 blocks. `Start` and `End` may point to same delegate since their positions are same, at least in this test case, so we set `consecutiveMissedBlocks` to 1 for other 102 delegates. This was my logic.

shuse2: Ah I think I understand. 
> Start: previous round's last block
End: current round's last block

I think it was originally first block of the current round to the  current block, but i think the result is the same, so it's ok

shuse2 (approved):

Incede (approved):

PR #7106, opened 2022-04-07
Implement cross chain registration command - Closes #7029

### LINKED ISSUE #7029: Implement CCRegistrationCMD cross chain command
### Description

Implement `CrossChainRegistrationCommand` under `cc_commands` for both `sidechain_interoperability` and `mainchain_interoperability` according to [LIP0049:Registration Message](https://github.com/LiskHQ/lips/blob/aced79e3a90c893f25c8f0735ce73732f17cced6/proposals/lip-0049.md#registration-message-1)
- Define cross chain ID as `CROSS_CHAIN_COMMAND_ID_REGISTRATION `
- Define schema as [LIP0049: registrationCCMParamsSchema](https://github.com/LiskHQ/lips/blob/aced79e3a90c893f25c8f0735ce73732f17cced6/proposals/lip-0049.md#parameters-2)
- Implement `exceute(crossChainCommandContext)` according to [LIP0049:Registration Message](https://github.com/LiskHQ/lips/blob/aced79e3a90c893f25c8f0735ce73732f17cced6/proposals/lip-0049.md#registration-message-1)
- Make sure `crossChainCommandContext` is defined

### Acceptance Criteria

- Should have all the unit tests covering `constructor` and `execute()` method

### LINKED ISSUE #7107: Cyclic dependency in interoperability module
### Expected behavior

There should be no cyclic dependency

### Actual behavior

`CCCommands` are part of modules, and while executing `CCCommands` we call internal functions like `terminateChainInternal` that calls `sendInternal`. `sendInternal` iterates over all the hooks (in this case beforeSendCCM) of interoperable modules for which it requires list of all the InteroperableModules's `CCCommands`. This introduces cyclic dependency as to execute a `CCCommands` we need access to Interoperability store to execute all the interoperability private functions and to create Interoperability store we need list of all the modules with its `CCCommands` and `CCAPI`.

ishantiw:
### What was the problem?

This PR resolves #7029 & #7107

### How was it solved?

- [♻️ Pass interoperable modules on calling the functions](https://github.com/LiskHQ/lisk-sdk/commit/fe6a6bd08c02b730e55524767f5873bdc1e595cd)
- [🌱 Add Registration cross chain command](https://github.com/LiskHQ/lisk-sdk/commit/9457d2f16d2068dbf72b4ecefda4a5fad4aa550b)
- [♻️ Only pass ccAPI into interoperabilityStore](https://github.com/LiskHQ/lisk-sdk/pull/7106/commits/06bffa1da157969d2c37d15ad61885a744593f31) Fixes #7107
  - To remove the cyclic dependency of ccCommands calling itself, only ccAPI is passed to interoperabilityStore and when calling apply we pass all the ccCommands as an argument.

### How was it tested?

`npm run test:unit test/unit/modules/interoperability`


framework/src/modules/interoperability/base_interoperability_module.ts
```diff
@@ -12,13 +12,16 @@
...
 
+import { BaseCCCommand } from './base_cross_chain_command';
+import { BaseInteroperableAPI } from './base_interoperable_api';
 import { BaseInteroperableModule } from './base_interoperable_module';
 import { MODULE_ID_INTEROPERABILITY, MODULE_NAME_INTEROPERABILITY } from './constants';
 
 export abstract class BaseInteroperabilityModule extends BaseInteroperableModule {
 	public id = MODULE_ID_INTEROPERABILITY; // Common id for mainchain/sidechain interoperability module
 	public name = MODULE_NAME_INTEROPERABILITY; // Common name for mainchain/sidechain interoperability module
 	// To manage interoperable modules
```
Incede, line 23: This comment probably needs to be updated

framework/src/modules/interoperability/base_interoperability_store.ts
```diff
@@ -330,21 +350,18 @@ export abstract class BaseInteroperabilityStore {
 			return;
 		}
 
-		const interoperableModule = this._interoperableModules.get(ccm.moduleID);
-
+		const ccCommands = interoperableCCCommands.get(ccm.moduleID);
 		// When moduleID is not supported
```
Incede, line 354: comment needs to be updated

framework/src/modules/interoperability/mainchain/cc_commands/registration.ts
```diff
@@ -0,0 +1,79 @@
...
+
+export class MainchainCCRegistrationCommand extends BaseCCCommand {
+	public ID = CROSS_CHAIN_COMMAND_ID_REGISTRATION;
+	public name = 'registration';
+	public schema = registrationCCMParamsSchema;
+	private readonly _interoperableCCAPIs: Map<number, BaseInteroperableAPI>;
+
+	public constructor(moduleID: number, interoperableCCAPIs: Map<number, BaseInteroperableAPI>) {
+		super(moduleID);
+		this._interoperableCCAPIs = interoperableCCAPIs;
```
Incede, line 39: Is this `interoperableCCAPIs` different for mainchain & sidechain command ?
If yes then maybe a distinct variable name might be better to understand.

Incede (changes requested):
Minor suggestions/questions.

framework/src/modules/interoperability/base_interoperability_store.ts
ishantiw: when ccCommands is undefined means that module entry doesn't exist, the next condition is to check ccCMDID

framework/src/modules/interoperability/mainchain/cc_commands/registration.ts
ishantiw: because CCCommands created in interoperable modules will never use or call `_interoperableCCAPIs `, they will only use API of interoperability module to interact

Incede: Then maybe distinct names for main/side chain would be better to understand. eg `mainchainInteroperableCCAPIs`

mitsuaki-u (changes requested):
A few missing checks but looks good

framework/src/modules/interoperability/mainchain/cc_commands/registration.ts
```diff
@@ -0,0 +1,79 @@
...
+		const decodedParams = codec.decode<CCMRegistrationParams>(
+			registrationCCMParamsSchema,
+			ccm.params,
+		);
+		const interoperabilityStore = this._getInteroperabilityStore(ctx.getStore);
+		const sendingChainChannelAccount = await interoperabilityStore.getChannel(ccm.sendingChainID);
+		const ownChainAccount = await interoperabilityStore.getOwnChainAccount();
+
+		if (
+			sendingChainChannelAccount.inbox.size !== 1 ||
```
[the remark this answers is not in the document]
ishantiw: Good catch 😉

Incede (approved):
There's one unresolved comment else LGTM

mitsuaki-u (approved):
Besides other comments LGTM

mehmetegemen (changes requested):
Magnificent work, just few comments

framework/src/modules/interoperability/base_interoperability_store.ts
```diff
@@ -96,6 +95,23 @@ export abstract class BaseInteroperabilityStore {
 		);
 	}
 
+	public async getChannel(chainID: number): Promise<ChannelData> {
+		const channelAccountStore = this.getStore(this._moduleID, STORE_PREFIX_CHANNEL_DATA);
+		return channelAccountStore.getWithSchema<ChannelData>(
```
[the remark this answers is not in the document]
ishantiw: its an async function so the caller function will use await for the returned async func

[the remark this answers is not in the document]
mehmetegemen: As I know `async` wraps the content of the function in a promise, so returning a simple value would needed to be awaited too. When you return a promise inside an async function, as I remember, you create a promise chain, that why it's working with a double promise. In my theoretical knowledge, this function should work without the `async` keyword if `getWithSchema` is an async function.

framework/src/modules/interoperability/base_interoperability_cc_commands.ts
```diff
@@ -14,12 +14,18 @@
...
+	protected readonly interoperableCCAPIs = new Map<number, BaseInteroperableAPI>();
 
 	public constructor(moduleID: number, interoperableCCAPIs: Map<number, BaseInteroperableAPI>) {
 		super(moduleID);
-		this._interoperableCCAPIs = interoperableCCAPIs;
+		this.interoperableCCAPIs = interoperableCCAPIs;
+	}
+
+	protected getInteroperabilityStore(getStore: getStoreFunction) {
+		return new SidechainInteroperabilityStore(this.moduleID, getStore, this.interoperableCCAPIs);
```
Incede, line 29: The name says `getInteroperabilityStore` but it's returning `SidechainInteroperabilityStore`. I think it's a bit misleading.

framework/src/modules/interoperability/mainchain/cc_commands/registration.ts
```diff
@@ -38,7 +36,7 @@ export class MainchainCCRegistrationCommand extends BaseInteroperabilityCCComman
 			registrationCCMParamsSchema,
 			ccm.params,
 		);
-		const interoperabilityStore = this._getInteroperabilityStore(ctx.getStore);
+		const interoperabilityStore = this.getInteroperabilityStore(ctx.getStore);
```
Incede, line 39: I am not sure if using an instance of `SidechainInteroperabilityStore` for mainchain commands is a good idea or maybe I am missing something ?

Incede (changes requested):

framework/src/modules/interoperability/types.ts
```diff
@@ -16,6 +16,7 @@ import { Logger } from '../../logger';
 import { APIContext, EventQueue } from '../../node/state_machine';
 import { SubStore } from '../../node/state_machine/types';
 
+export type getStoreFunction = (moduleID: number, storePrefix: number) => SubStore;
```
mehmetegemen, line 19: As per style guides type names are advised to be PascalCase, meaning `GetStoreFunction` is better here. Also using a verb in type name is a bit off I think, we can maybe say `StoreCallback`?

mehmetegemen: And I am sure you know these things so I think you should incorporate more rest into your working hours and don't work while you are sleepy dear Ishan 🙂  Sleep is so important 😴

framework/src/modules/interoperability/base_interoperability_cc_commands.ts
mehmetegemen: What do you think about making this abstract to inform to type system that there is a method here, but implementing store constructors in classes that inherit `BaseInteroperabilityCCCommand`? Because I didn't understand why we implement sidechain interoperability store constructor for every subclass, don't we need to create also mainchain interoperability store in some cases?

framework/src/modules/interoperability/mainchain/cc_commands/registration.ts
mehmetegemen: Ishan hurried to merge this PR late at night because it's blocking and mixed with tiredness he got confused 🙂But I agree with Rishi, with base class implementation this call here is using a sidechain interoperability store, just to clarify for world wide web.

framework/src/modules/interoperability/base_interoperability_cc_commands.ts
ishantiw: yes this was a mistake, while moving this function to BaseInteroperabilityStore I forgot they need specific Interoperability module's store. I will make this function abstract and to be implemented in their respective classes

framework/src/modules/interoperability/mainchain/cc_commands/registration.ts
ishantiw: same as above comment

framework/src/modules/interoperability/types.ts
```diff
@@ -159,6 +161,7 @@ export interface CCCommandExecuteContext {
 	networkIdentifier: Buffer;
 	eventQueue: EventQueue;
 	ccm: CCMsg;
-	getAPIContext: () => ImmutableAPIContext;
+	getAPIContext: () => APIContext;
```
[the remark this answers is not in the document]
ishantiw: only interoperability module will use CCAPIContext, and API will never use CCAPIContext or call it. The user when executes a ccm, then it can expect to have api to other modules to process something so its needed here

Incede (approved):

PR #7112, opened 2022-04-13
Bootstrap mainchain and sidechain registration command - Closes #7032 

### LINKED ISSUE #7032: Bootstrap mainchain/sidechain registration commands
### Description

**Under `/src/modules/mainchain_interoperability`**

- Create `commands/sidechain_registration.ts`
- Add missing constants and types
- Add `SidechainRegistrationCMD` class extending `BaseCommand`
- Add CMD id = `COMMAND_ID_SIDECHAIN_REG`,name=`sidechainRegistration` and [schema](https://github.com/LiskHQ/lips/blob/aced79e3a90c893f25c8f0735ce73732f17cced6/proposals/lip-0043.md#parameters) 

**Under `/src/modules/sidechain_interoperability`**

- Create `commands/sidechain_registration.ts`
- Add missing constants and types
- Add `MainchainRegistrationCMD` class extending `BaseCommand`
- Add CMD id = `COMMAND_ID_MAINCHAIN_REG`,name=`mainchainRegistration` and [schema](https://github.com/LiskHQ/lips/blob/aced79e3a90c893f25c8f0735ce73732f17cced6/proposals/lip-0043.md#parameters-1) 

### Acceptance Criteria

- All the files and constants should exist

mitsuaki-u:
### What was the problem?

This PR resolves #7032

### How was it solved?

- Add MainchainRegistrationCommand class
- Add SidechainRegistrationCommand class
- Add constants
- Add schemas

### How was it tested?

n/a

### DIFF
--- a/framework/src/modules/interoperability/constants.ts
+++ b/framework/src/modules/interoperability/constants.ts
@@ -17,10 +17,18 @@ export const MODULE_NAME_INTEROPERABILITY = 'interoperability';
 
 // General constants
 export const MAINCHAIN_ID = 1;
+export const MAINCHAIN_NAME = 'lisk-mainchain';
+export const TAG_CHAIN_REG_MESSAGE = 'LSK_CHAIN_REGISTRATION';
 export const LIVENESS_LIMIT = 2592000; // 30*24*3600
 export const MAX_CCM_SIZE = 10240;
 export const EMPTY_FEE_ADDRESS = Buffer.alloc(0);
 export const EMPTY_BYTES = Buffer.alloc(0);
+export const REGISTRATION_FEE = 1000000000;
+export const MAX_NUM_VALIDATORS = 199;
+export const MAX_LENGTH_NAME = 40;
+export const MAX_UINT32 = 4294967295;
+export const MAX_UINT64 = 18446744073709551615;
+export const THRESHOLD_MAINCHAIN = 68;
 
 // Store prefixes
 export const STORE_PREFIX_OUTBOX_ROOT = 0x0000;
@@ -48,3 +56,7 @@ export const CCM_STATUS_CROSS_CHAIN_COMMAND_NOT_SUPPORTED = 2;
 export const CCM_STATUS_CHANNEL_UNAVAILABLE = 3;
 export const CCM_STATUS_RECOVERED = 4;
 export const MIN_RETURN_FEE = BigInt(1000);
+
+// Commands
+export const COMMAND_ID_SIDECHAIN_REG = 0;
+export const COMMAND_ID_MAINCHAIN_REG = 1;
--- a/framework/src/modules/interoperability/sidechain/commands/mainchain_registration.ts
+++ b/framework/src/modules/interoperability/sidechain/commands/mainchain_registration.ts
@@ -1,0 +1,36 @@
+/*
+ * Copyright © 2022 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+import { BaseCommand } from '../../../base_command';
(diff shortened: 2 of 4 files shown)


framework/src/modules/interoperability/constants.ts
```diff
@@ -17,10 +17,18 @@ export const MODULE_NAME_INTEROPERABILITY = 'interoperability';
...
+export const TAG_CHAIN_REG_MESSAGE = 'LSK_CHAIN_REGISTRATION';
 export const LIVENESS_LIMIT = 2592000; // 30*24*3600
 export const MAX_CCM_SIZE = 10240;
 export const EMPTY_FEE_ADDRESS = Buffer.alloc(0);
 export const EMPTY_BYTES = Buffer.alloc(0);
+export const REGISTRATION_FEE = 1000000000;
+export const MAX_NUM_VALIDATORS = 199;
+export const MAX_LENGTH_NAME = 40;
+export const MAX_UINT32 = 4294967295;
+export const MAX_UINT64 = 18446744073709551615;
```
mehmetegemen, line 30: Can we have these values in a format like for example `2^64 – 1` for maxuint64? By the way this value is more than max number value of javascript which is 2^53-1 or 9007199254740991 so I think even though you assigned a value here, when we use this constant it will default it's value to 2^53-1 so maybe use a BigInt here or find another solution.

mehmetegemen (changes requested):

framework/src/modules/interoperability/sidechain/commands/mainchain_registration.ts
```diff
@@ -0,0 +1,36 @@
...
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+import { BaseCommand } from '../../../base_command';
+import { COMMAND_ID_MAINCHAIN_REG } from '../../constants';
+import { mainchainRegParams } from '../../schema';
+import { VerificationResult } from '../../../../node/state_machine';
+
+export class MainRegistrationCommand extends BaseCommand {
+	public id = COMMAND_ID_MAINCHAIN_REG;
```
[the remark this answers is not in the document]
mitsuaki-u: Prefer to keep it as defined in the LIP and I think in this case the meaning is discernible.

framework/src/modules/interoperability/constants.ts
```diff
@@ -27,7 +27,7 @@ export const REGISTRATION_FEE = 1000000000;
 export const MAX_NUM_VALIDATORS = 199;
 export const MAX_LENGTH_NAME = 40;
 export const MAX_UINT32 = 4294967295;
-export const MAX_UINT64 = 18446744073709551615;
+export const MAX_UINT64 = BigInt('18446744073709551615'); // BigInt((2 ** 64) - 1) - 1
```
mehmetegemen, line 30: Do you know where do we use this constant? Because `MAX_UINT32` is number and `MAX_UINT64` is bigint, they are not convertible to each other and this may be a problem for some interfaces depending on the usage.

mitsuaki-u: The constant will likely be used for validation of `bigint` in the code.  Of course the usage of `bigint` will be taken into account when it is used.

framework/src/modules/interoperability/constants.ts
```diff
@@ -17,10 +17,18 @@ export const MODULE_NAME_INTEROPERABILITY = 'interoperability';
 
 // General constants
 export const MAINCHAIN_ID = 1;
+export const MAINCHAIN_NAME = 'lisk-mainchain';
+export const TAG_CHAIN_REG_MESSAGE = 'LSK_CHAIN_REGISTRATION';
 export const LIVENESS_LIMIT = 2592000; // 30*24*3600
 export const MAX_CCM_SIZE = 10240;
 export const EMPTY_FEE_ADDRESS = Buffer.alloc(0);
 export const EMPTY_BYTES = Buffer.alloc(0);
+export const REGISTRATION_FEE = 1000000000;
```
Incede, line 26: In the LIP a `uint64` type is used for this variable although the current value fits in `uint32`. I think it'd be nice either to clarify the maximum allowed value of this constant or just use type `BigInt` since `JS number` supports maximum 53 bits.

Incede (approved):
Looks good, requested a small clarification.

mehmetegemen (approved):

Issue #7118, opened 2022-04-22
Update token transfer command


shuse2:
### Description
Update token transfer command to follow updated logic as in https://github.com/LiskHQ/lips/blob/main/proposals/lip-0051.md#token-transfer

### Acceptance criteria
- Token transfer unit test is updated
- All integration/functional tests are passing

Issue #7122, opened 2022-04-22
Update token API implementation


shuse2:
### Description
All API defined in [LIP](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0051.md#protocol-logic-for-other-modules-1) should be implemented except the API implemented in #7121 

### Acceptance Criteria
- All APIs are unit tested

Issue #7119, opened 2022-04-22
Update `initGenesisState` hook


shuse2:
### Description
Update `initGenesisState` hook according to https://github.com/LiskHQ/lips/blob/main/proposals/lip-0051.md#genesis-block-processing

### Acceptance criteria
- `initGenesisState` unit test is updated
- Example app is updated and working

PR #7138, opened 2022-04-25
Update token module and transfer command - Closes #7117 and #7118

### LINKED ISSUE #7117: Update token module definition
### Description
- Update token module config to
```
{
  minBalances: {
    tokenID: hex string,
    amount: number string,
  }[],
  // if supportedTokens is empty, all tokens are supported.
  supportedTokenIDs: string[],
}
```
- Add constants as described in https://github.com/LiskHQ/lips/blob/main/proposals/lip-0051.md#constants-and-notations
- Add store definition as described in https://github.com/LiskHQ/lips/blob/main/proposals/lip-0051.md#token-module-store

### Acceptance Criteria
- All above are defined

shuse2:
### What was the problem?

This PR resolves #7117 and #7118 

### How was it solved?

- Add schema and constants and transfer command as described in [LIP](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0051.md)

### How was it tested?

- All unit/integration tests should pass
- Update transfer command unit test


framework/src/modules/token/api.ts
```diff
@@ -231,4 +264,31 @@ export class TokenAPI extends BaseAPI {
 			return prev;
 		}, BigInt(0));
 	}
+
+	public async isNative(apiContext: ImmutableAPIContext, tokenID: TokenID): Promise<boolean> {
```
shuse2, line 268: Added `isNative` for fee module to check if chainID is native or not

framework/src/modules/token/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+import { CHAIN_ID_ALIAS_NATIVE, CHAIN_ID_LENGTH, TOKEN_ID_LENGTH } from './constants';
+import { TokenID } from './types';
+
```
shuse2, line 17: util functions for handling store key and token ID format

framework/src/modules/token/schemas.ts
```diff
@@ -61,39 +114,192 @@ export const getBalanceRequestSchema = {
...
+		tokenID: {
+			type: 'string',
+			format: 'hex',
+		},
 	},
-	required: ['address'],
+	required: ['address', 'tokenID'],
 };
 
 export const transferParamsSchema = {
```
shuse2, line 125: Updated schema, and add restriction on schema. This is used in the validation

framework/src/modules/token/module.ts
```diff
@@ -59,7 +67,11 @@ export class TokenModule extends BaseModule {
 		const genesisStore = codec.decode<GenesisTokenStore>(genesisTokenStoreSchema, assetBytes);
 		const userStore = context.getStore(this.id, STORE_PREFIX_USER);
 		for (const userData of genesisStore.userSubstore) {
-			await userStore.setWithSchema(userData.address, userData, userStoreSchema);
+			await userStore.setWithSchema(
+				getUserStoreKey(userData.address, userData.tokenID),
```
shuse2, line 71: Temporally update the genesis block handling. Proper implementation will be done in #7119

framework/src/modules/token/api.ts
```diff
@@ -52,26 +63,38 @@ export class TokenAPI extends BaseAPI {
 		apiContext: APIContext,
 		senderAddress: Buffer,
 		recipientAddress: Buffer,
-		_tokenID: TokenID,
+		tokenID: TokenID,
```
shuse2, line 66: this `transfer` function is the main change. Other API will be properly implemented in #7122

framework/src/modules/token/commands/transfer.ts
```diff
@@ -11,13 +11,20 @@
  *
  * Removal or modification of this copyright notice is prohibited.
  */
+import { LiskValidationError, validator } from '@liskhq/lisk-validator';
 import { BaseCommand } from '../../base_command';
-import { CommandExecuteContext } from '../../../node/state_machine';
+import {
```
shuse2, line 16: this command is the main file updated

framework/src/modules/token/api.ts
```diff
@@ -52,55 +63,87 @@ export class TokenAPI extends BaseAPI {
...
-					amount + this._minBalance
-				).toString()}`,
+				)} balance ${sender.availableBalance.toString()} is not sufficient for ${amount.toString()}`,
 			);
 		}
+
 		sender.availableBalance -= amount;
-		await userStore.setWithSchema(senderAddress, sender, userStoreSchema);
+
+		await userStore.setWithSchema(
```
MaximeGagnebin, line 85: Here do we risk subtracting the amount from the sender, and then the API failing for min balance ?

framework/src/modules/token/api.ts
```diff
@@ -52,55 +63,87 @@ export class TokenAPI extends BaseAPI {
...
 				lockedBalances: [],
 			};
 		}
-		if (recipient.availableBalance + amount < this._minBalance) {
-			throw new Error(
-				`Recipient ${recipientAddress.toString('hex')} balance ${(
-					recipient.availableBalance + amount
-				).toString()} is not sufficient for min balance ${this._minBalance.toString()}`,
-			);
+		const minBalance = this._getMinBalance(tokenID);
```
MaximeGagnebin, line 109: If i understand the function correctly, this would set minBalance =0 if the tokenID is not present in the minBalances config.
Here we would rather want the transfer to fail if `!recipientExist` as this transfer is not able to pay to initialize the account.

MaximeGagnebin (changes requested):

framework/src/modules/token/api.ts
shuse2: if the API fails, the store rollback, so it is ok to set the value

framework/src/modules/token/api.ts
```diff
@@ -52,55 +63,87 @@ export class TokenAPI extends BaseAPI {
...
 
 		let recipient: UserStoreData;
+		let recipientExist = false;
 		try {
-			recipient = await userStore.getWithSchema<UserStoreData>(recipientAddress, userStoreSchema);
+			recipient = await userStore.getWithSchema<UserStoreData>(
+				getUserStoreKey(recipientAddress, canonicalTokenID),
+				userStoreSchema,
+			);
+			recipientExist = true;
```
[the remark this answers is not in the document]
shuse2: I think I misunderstood the concept. Maybe it would be better to avoid the dependency to auth module, and rather only rely on `recipient holds any token`?
Also, user data is never deleted, so if there is a data corresponding to the address, I think it means `nonce > 0` indirectly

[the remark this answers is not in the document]
MaximeGagnebin: "User data is never deleted", ok, I guess that's fine. I thought that we would remove entries if the amount of tokens goes to 0. This way, we can clean up the store of all those empty accounts.

But if you tell me that 0 token entries are not really an issue (and probably they are not), we could leave them. We would need to update the LIP though.

How bad is it to rely on nonce?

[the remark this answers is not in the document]
shuse2: If we can resolve the issue i would try not to add dependency, but it's also not so bad i guess.

how about we still remove the entry if becomes empty, but if all becomes empty, user need to pay min balance to re-initialize the account?

mitsuaki-u (commented):
Main changes LGTM besides the open point with Maxime

framework/src/modules/token/api.ts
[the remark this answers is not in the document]
MaximeGagnebin: I think we settled on not removing empty entries, and the min balance logic is now quite different. Will resolve this comment.

MaximeGagnebin (approved):

PR #7145, opened 2022-04-29
Update token module initGenesisState hook - Closes #7119

### LINKED ISSUE #7119: Update `initGenesisState` hook

shuse2:
### What was the problem?

This PR resolves #7119 

### How was it solved?

- Update `initGenesisState` hook according to the LIP

### How was it tested?

- Added unit tests

### DIFF
--- a/framework/src/modules/token/module.ts
+++ b/framework/src/modules/token/module.ts
@@ -18,6 +18,7 @@ import {
 	CHAIN_ID_ALIAS_NATIVE,
 	defaultConfig,
 	EMPTY_BYTES,
+	LOCAL_ID_LENGTH,
 	MODULE_ID_TOKEN,
 	STORE_PREFIX_AVAILABLE_LOCAL_ID,
 	STORE_PREFIX_ESCROW,
@@ -95,7 +96,7 @@ export class TokenModule extends BaseModule {
 		const userStore = context.getStore(this.id, STORE_PREFIX_USER);
 		const copiedUserStore = [...genesisStore.userSubstore];
 		copiedUserStore.sort((a, b) => {
-			if (a.address.equals(b.address)) {
+			if (!a.address.equals(b.address)) {
 				return a.address.compare(b.address);
 			}
 			return a.tokenID.compare(b.tokenID);
@@ -118,7 +119,7 @@ export class TokenModule extends BaseModule {
 			// Validate sorting of userSubstore
 			if (
 				!userData.address.equals(copiedUserStore[i].address) ||
-				!userData.tokenID.equals(copiedUserStore[i].address)
+				!userData.tokenID.equals(copiedUserStore[i].tokenID)
 			) {
 				throw new Error('UserSubstore must be sorted by address and tokenID.');
 			}
@@ -204,7 +205,7 @@ export class TokenModule extends BaseModule {
 			escrowKeySet.add(key);
 			// validate terminated escrow chain ID/local ID order
 			if (
-				escrowData.escrowChainID.equals(copiedEscrowStore[i].escrowChainID) ||
+				!escrowData.escrowChainID.equals(copiedEscrowStore[i].escrowChainID) ||
 				!escrowData.localID.equals(copiedEscrowStore[i].localID)
 			) {
 				throw new Error('EscrowSubstore must be sorted by escrowChainID and localID.');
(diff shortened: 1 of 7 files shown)


framework/test/unit/modules/token/init_genesis_state_fixture.ts
```diff
@@ -0,0 +1,327 @@
...
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+// import { getRandomBytes } from '@liskhq/lisk-cryptography';
+
+const oneLisk = BigInt('100000000');
```
ishantiw, line 17: maybe `oneLiskToBeddows` is more accurate?

ishantiw (approved):
Excellent 🚀

framework/src/modules/token/module.ts
```diff
@@ -75,21 +88,235 @@ export class TokenModule extends BaseModule {
...
+			if (
+				!escrowData.escrowChainID.equals(copiedEscrowStore[i].escrowChainID) ||
+				!escrowData.localID.equals(copiedEscrowStore[i].localID)
+			) {
+				throw new Error('EscrowSubstore must be sorted by escrowChainID and localID.');
+			}
+			await escrowStore.setWithSchema(key, { amount: escrowData.amount }, escrowStoreSchema);
+		}
+
+		const nextAvailableLocakIDStore = context.getStore(this.id, STORE_PREFIX_AVAILABLE_LOCAL_ID);
```
mehmetegemen, line 216: I think `Locak` is a typo.

framework/src/modules/token/module.ts
```diff
@@ -75,21 +88,235 @@ export class TokenModule extends BaseModule {
...
+		// validate next available ID
+		const {
+			nextAvailableLocalID,
+		} = await nextAvailableLocakIDStore.getWithSchema<AvailableLocalIDStoreData>(
+			EMPTY_BYTES,
+			availableLocalIDStoreSchema,
+		);
+		// If maxLocalID is larger than nextAvailableLocalID, it is invalid
+		if (maxLocalID.compare(nextAvailableLocalID) > 0) {
+			throw new Error('Max token local ID is higher than next availableLocalID');
```
mehmetegemen, line 319: We reflect our variable naming in `availableLocalID` but don't reflect it in `Max token local ID`. `Max token local ID` is quite confusing while reading so I think either we should add hyphen to the words in it or reflect variable name in the error message to be more clear.

framework/src/modules/token/schemas.ts
```diff
@@ -317,10 +341,14 @@ export const genesisTokenStoreSchema = {
 					address: {
 						dataType: 'bytes',
 						fieldNumber: 1,
+						minLength: 20,
+						maxLength: 20,
```
mehmetegemen, line 345: Why don't we use `ADDRESS_LENGTH` here like other places in the schema using their own constants like `LOCAL_ID_LENGTH`?

framework/src/testing/fixtures/genesis-asset.ts
```diff
@@ -693,75 +695,16 @@ export const blockAssetsJSON = [
 			supplySubstore: [
 				{
 					localID: '00000000',
-					totalSupply: '1000000000000000',
+					totalSupply: '1120000000000000',
```
mehmetegemen, line 698: Is it possible to use `convertLSKToBeddows` here?

framework/test/unit/modules/token/init_genesis_state_fixture.ts
mehmetegemen: I think `oneLSKToBeddows` is more accurate, since `Lisk` a brand and `LSK` is the token's name. It's also aligned with our functions like `convertLSKToBeddows` and we can use this function here like `const oneLSKToBeddows = convertLSKToBeddows('1');`, I think this is more readable.

mehmetegemen (commented):
Just few comments, otherwise great job!

framework/src/modules/token/schemas.ts
shuse2: I updated this on later PR 🙏

framework/test/unit/modules/token/init_genesis_state_fixture.ts
shuse2: I think we could use `convertLSKToBeddows` but LSK or beddow is not the main point here, so ill just update the variable name

framework/src/modules/token/module.ts
shuse2: I think Max token local ID was not matching as you mentioned, so I updated to Max local ID

framework/src/testing/fixtures/genesis-asset.ts
shuse2: same as above, i think lsk is not relevant / correct here, so I would prefer to keep it this way

mehmetegemen (approved):

PR #7158, opened 2022-05-10
Integrate events into block execution - Closes #7127

### LINKED ISSUE #7127: Update block processing to handle events
### Description
- Add event root to the block header and update the schema
- Update state machine to handle reverting of the event based on result of command execution
- Create event root at the end of block generation and block execution
  - block generation should insert the event root at result calculation step
  - block execution should verify the event root at result verification step

### Acceptance Criteria
- Update state machine unit tests to check both success/fail command case
- Update consensus and generator tests for event root
- Update lisk-chain tests for new event root property
- All builds and tests must pass

shuse2:
### What was the problem?

This PR resolves #7127

### How was it solved?

- Add event root calculation to block generation and genesis block generation
- Add event root verification to block execution
- Add event queue handling to state machine

### How was it tested?

- Update unit and integration tests


framework/src/node/state_machine/constants.ts
```diff
@@ -0,0 +1,19 @@
...
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+export const EVENT_INDEX_INIT_GENESIS_STATE = Buffer.from([0]);
+export const EVENT_INDEX_FINALIZE_GENESIS_STATE = Buffer.from([1]);
+
+export const EVENT_INDEX_BEFORE_TRANSACTIONS = Buffer.from([2]);
+export const EVENT_INDEX_AFTER_TRANSACTIONS = Buffer.from([3]);
```
shuse2, line 19: @ricott1 I added the genesis prefix like this for now. Please let me know if other constants are better.

ricott1: these are fine, I will update the LIP when I open the PR on lips

framework/src/node/consensus/consensus.ts
```diff
@@ -742,6 +756,28 @@ export class Consensus {
 		}
 	}
 
+	private async _verifyEventRoot(block: Block, events: Event[]): Promise<void> {
+		if (events.length > MAX_EVENTS_PER_BLOCK) {
+			throw new Error(`Number of event cannot exceed ${MAX_EVENTS_PER_BLOCK} per block`);
```
ishantiw, line 761:
```suggestion
			throw new Error(`Number of events cannot exceed ${MAX_EVENTS_PER_BLOCK} per block`);
```

framework/src/node/state_machine/types.ts
```diff
@@ -15,7 +15,10 @@ import { Transaction } from '@liskhq/lisk-chain';
 import { IterateOptions } from '@liskhq/lisk-chain/dist-node/state_store';
 import { Schema } from '@liskhq/lisk-codec';
 import { Logger } from '../../logger';
-import { EventQueue } from './event_queue';
+
+export interface EventQueueAdder {
```
ishantiw, line 19: Maybe generic name? `EventQueueManager` or `EventQueueUtil`

framework/test/unit/node/state_machine/state_machine.spec.ts
```diff
@@ -124,13 +148,31 @@ describe('state_machine', () => {
...
-				eventQueue,
+				eventQueue: expect.any(Object),
 				getAPIContext: expect.any(Function),
 				getStore: expect.any(Function),
 			});
 			expect(systemMod.afterCommandExecute).toHaveBeenCalledTimes(1);
 			expect(mod.afterCommandExecute).toHaveBeenCalledTimes(1);
 		});
+
+		it('should add event with a topic with transaciton id', async () => {
```
ishantiw, line 159:
```suggestion
		it('should add event with a topic with transaction id', async () => {
```

ishantiw (approved):
Flawless 👏 , just a few minor suggestions

framework/src/node/state_machine/types.ts
shuse2: I'm not sure as an interface those name would be suited because it sounds like it has actual implementation.
The main goal here is to limit the scope to `add` function for the modules. (and adding extra logic in state machine)

framework/src/node/state_machine/api_context.ts
```diff
@@ -34,14 +34,29 @@ interface ImmutableSubStoreGetter {
...
 		immutableSubstoreGetter.getStore(moduleID, storePrefix),
 });
 
+export const wrapEventQueue = (eventQueue: EventQueue, topic: Buffer): EventQueueAdder => ({
+	add: (
+		moduleID: number,
+		typeID: Buffer,
+		data: Buffer,
+		topics?: Buffer[],
+		noRevert?: boolean,
```
mehmetegemen, line 47: What do you think about a default value here to have a certain false value instead of a falsy value? This can also enable you to change `noRevert` param name to `revert` and practice better naming.

shuse2: I think having default value is fine, but if we do, topics wont be optional or need default value as well.
It is `noRevert` because almost always it should be reverted, and in rare case it should intentionally put to `true`.
That's why it's not `revert` where default is `false`

mehmetegemen: 👍

Issue #6947, opened 2021-11-30
Provide application interface information


shuse2:
### Description

Expose an endpoint to return metadata describes the application interface including
- schema for block, transaction, params for commands, event and block assets
- endpoints
- registered events

### Motivation

Information for external system, such as lisk-service or lisk-desktop, to know what commands, endpoints, assets or events are available in the connected application should be exposed through the RPC

### Tasks
- [x] #7063
- [x] #6991
- [x] #7064
- [x] #7065

PR #7159, opened 2022-05-10
Add event endpoints - Closes #7126

### LINKED ISSUE #7126: Add `chain_getEvents` and `chain_proveEvents` endpoints
### Description
Add framework endpoints as below

##### Req
```
{
  height: number,
}
```
##### Res
```
{
  events: {
    moduleID: number,
    typeID: number,
    topics: string[],
    data: string,
    index: number,
  }[],
}
```
If height does not exist, it should return
```
{
  error: {
    message: `Events at height ${height} is not stored in the node.`,
  }
}
```
##### Req
```
{
  height: number,
  queries: string[],
}
```
##### Res
```
{
  siblingHashes: string[],
  queries: {
    key: string,
    value: string,
    bitmap: string,
  }[]
}
```
If height does not exist, it should return
```
{
  error: {
    message: `Events at height ${height} is not stored in the node.`,
  }
}
```

### Acceptance Criteria
- All endpoints are unit tested

shuse2:
### What was the problem?

This PR resolves #7126

### How was it solved?

- Add new endpoints in node endpoint

### How was it tested?

- Add unit tests

### Additional Information
- Namespace will be updated in #6947

### DIFF
--- a/elements/lisk-tree/src/index.ts
+++ b/elements/lisk-tree/src/index.ts
@@ -18,6 +18,7 @@ import { MerkleTree } from './merkle_tree/merkle_tree';
 import { calculateMerkleRoot, calculateMerkleRootWithLeaves } from './merkle_tree/utils';
 import { SparseMerkleTree } from './sparse_merkle_tree/sparse_merkle_tree';
 import { verify, calculateRoot } from './sparse_merkle_tree/utils';
+import { Proof as SMTProof, Query as SMTQuery } from './sparse_merkle_tree/types';
 
 export const regularMerkleTree = {
 	verifyProof,
@@ -33,4 +34,4 @@ export const sparseMerkleTree = {
 	SparseMerkleTree,
 };
 
-export { MerkleTree, SparseMerkleTree };
+export { MerkleTree, SparseMerkleTree, SMTProof, SMTQuery };
--- a/framework/src/node/endpoint.ts
+++ b/framework/src/node/endpoint.ts
@@ -11,11 +11,12 @@
  *
  * Removal or modification of this copyright notice is prohibited.
  */
-import { Chain } from '@liskhq/lisk-chain';
-import { NotFoundError } from '@liskhq/lisk-db';
-import { isHexString } from '@liskhq/lisk-validator';
+import { Chain, EventAttr, EVENT_KEY_LENGTH, SMTStore } from '@liskhq/lisk-chain';
+import { InMemoryKVStore, NotFoundError } from '@liskhq/lisk-db';
+import { isHexString, LiskValidationError, validator } from '@liskhq/lisk-validator';
+import { SparseMerkleTree, SMTProof } from '@liskhq/lisk-tree';
 import { BaseModule } from '../modules';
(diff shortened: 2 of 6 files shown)


mehmetegemen (approved):

elements/lisk-tree/src/index.ts
```diff
@@ -18,6 +18,7 @@ import { MerkleTree } from './merkle_tree/merkle_tree';
 import { calculateMerkleRoot, calculateMerkleRootWithLeaves } from './merkle_tree/utils';
 import { SparseMerkleTree } from './sparse_merkle_tree/sparse_merkle_tree';
 import { verify, calculateRoot } from './sparse_merkle_tree/utils';
+import { Proof as SMTProof, Query as SMTQuery } from './sparse_merkle_tree/types';
```
ishantiw, line 21: maybe in future we can already export SMTProof from lisk-tree as we rename it anyways

framework/src/node/endpoint.ts
```diff
@@ -32,6 +33,25 @@ interface InitArgs {
 	registeredModules: BaseModule[];
 }
 
+const proveEventsRequestSchema = {
+	$id: '/node/endpoint/proveEventsRequestSchema',
+	type: 'object',
+	required: ['height', 'queries'],
```
ishantiw, line 39: fieldNumber is missing

ishantiw, line 39: fieldNumber is missing

framework/test/unit/node/endpoint.spec.ts
```diff
@@ -0,0 +1,146 @@
...
+import { Event, StateStore } from '@liskhq/lisk-chain';
+import { getRandomBytes } from '@liskhq/lisk-cryptography';
+import { InMemoryKVStore } from '@liskhq/lisk-db';
+import { ModuleEndpointContext } from '../../../src';
+import { Endpoint } from '../../../src/node/endpoint';
+import { createImmutableAPIContext } from '../../../src/node/state_machine';
+import { nodeConfig } from '../../utils/configs';
+import { fakeLogger } from '../../utils/node';
+
+const createContext = (
```
ishantiw, line 24: can we use some existing createContext funcs in testing or add this one to make it reusable

ishantiw, line 24: can we use some existing createContext funcs in testing or add this one to make it reusable

ishantiw (commented):
LGTM, just a few comments

framework/src/node/endpoint.ts
shuse2: Endpoint req/res are not encoded(ie: standard JSON schema). Therefore, fieldNumber is not required and it shouldnt use dataType

elements/lisk-tree/src/index.ts
shuse2: yes, but the SMT types will be provided by lisk-db, so I will update there

ishantiw (approved):
🚀

PR #7162, opened 2022-05-12
Add metadata and endpoint - Closes #7063 and #7064

### LINKED ISSUE #7063: Update all modules to expose metadata
### Description
- Define `ModuleMetadata` interface as below
- All module should expose `metadata` as defined in below

```
public metadata(): ModuleMetadata {
    return {
        endpoints: [{
            name: 'getBalance',
            request: {
                // schema
                type: 'object',
            },
            response: {
                // schema
                type: 'object',
            },
        }],
       // This should be empty array for now
        events: [{
            typeID: 3,
            topics: ['transactionID'],
            data: {
                // schema
                type: 'object',
            },
        }],
        commands: [{
            commandID: 3,
            name: 'transfer',
            params: {
                // schema
                type: 'object',
            },
        }],
        assets: [{
            version: 0,
                // schema
            data: {
                type: 'object',
            },
        }],
    };
}
```

### Acceptance Criteria

All modules exposes metadata as defined in the description

### LINKED ISSUE #7064: Add system_getMetadata endpoint
### Description

Add `system_getMetadata` endpoint as below.

```
{
    modules: [{
        name: 'poa',
        id: 12,
        endpoints: [{
            name: 'token_getBalance',
            request: {
                type: 'object',
            },
            response: {
                type: 'object',
            },
        }],
        events: [{
            eventID: 3,
            name: 'token_transfer',
            data: {
                type: 'object',
            },
        }],
        commands: [{
            moduleID: 0,
            moduleName: 'token',
            commandID: 3,
            name: 'transfer',
            params: {
                type: 'object',
            },
        }],
        assets: [{
            version: 0,
            data: {
                type: 'object',
            },
        }],
    }],
}
```
Additionally, remove `registeredModule` parameters from `getNodeInfo` endpoint

### Acceptance Criteria

- All modules metadata should be aggregated and exposed as description above
- Unit test should be added

shuse2:
### What was the problem?

This PR resolves #7063 and resolves #7064

### How was it solved?

- Add metadata to all the modules
- Add system_getMetadata endpoint

### How was it tested?

- Add unit test
- Run dpos-mainchain example and call system_getMetadata

### DIFF
--- a/framework/src/modules/dpos_v2/module.ts
+++ b/framework/src/modules/dpos_v2/module.ts
@@ -17,7 +17,7 @@ import { objects as objectUtils, dataStructures, objects } from '@liskhq/lisk-ut
 import { isUInt64, LiskValidationError, validator } from '@liskhq/lisk-validator';
 import { codec } from '@liskhq/lisk-codec';
 import { GenesisBlockExecuteContext, BlockAfterExecuteContext } from '../../node/state_machine';
-import { BaseModule, ModuleInitArgs } from '../base_module';
+import { BaseModule, ModuleInitArgs, ModuleMetadata } from '../base_module';
 import { DPoSAPI } from './api';
 import { DelegateRegistrationCommand } from './commands/delegate_registration';
 import { ReportDelegateMisbehaviorCommand } from './commands/pom';
@@ -45,6 +45,11 @@ import {
 	delegateStoreSchema,
 	genesisDataStoreSchema,
 	genesisStoreSchema,
+	getAllDelegatesResponseSchema,
+	getDelegateRequestSchema,
+	getDelegateResponseSchema,
+	getVoterRequestSchema,
+	getVoterResponseSchema,
 	nameStoreSchema,
 	previousTimestampStoreSchema,
 	snapshotStoreSchema,
@@ -129,6 +134,39 @@ export class DPoSModule extends BaseModule {
 		this._voteCommand.addDependencies({
 			tokenAPI: this._tokenAPI,
 		});
+	}
+
+	public metadata(): ModuleMetadata {
+		return {
+			endpoints: [
+				{
+					name: this.endpoint.getAllDelegates.name,
+					response: getAllDelegatesResponseSchema,
+				},
+				{
+					name: this.endpoint.getDelegate.name,
+					request: getDelegateRequestSchema,
+					response: getDelegateResponseSchema,
+				},
+				{
+					name: this.endpoint.getVoter.name,
+					request: getVoterRequestSchema,
+					response: getVoterResponseSchema,
+				},
+			],
+			commands: this.commands.map(command => ({
+				id: command.id,
+				name: command.name,
+				params: command.schema,
+			})),
+			events: [],
+			assets: [
+				{
+					version: 0,
+					data: genesisStoreSchema,
+				},
+			],
+		};
 	}
 
 	// eslint-disable-next-line @typescript-eslint/require-await
--- a/framework/src/modules/dpos_v2/schemas.ts
+++ b/framework/src/modules/dpos_v2/schemas.ts
@@ -535,3 +535,123 @@ export const genesisStoreSchema = {
 		},
 	},
 };
+
+const delegateJSONSchema = {
+	type: 'object',
+	required: [
+		'name',
+		'totalVotesReceived',
+		'selfVotes',
+		'lastGeneratedHeight',
+		'isBanned',
+		'pomHeights',
+		'consecutiveMissedBlocks',
+	],
+	properties: {
+		name: {
+			type: 'string',
+		},
+		totalVotesReceived: {
+			type: 'string',
+			format: 'uint64',
+		},
+		selfVotes: {
+			type: 'string',
+			format: 'uint64',
+		},
+		lastGeneratedHeight: {
+			type: 'integer',
+			format: 'uint32',
+		},
+		isBanned: {
+			type: 'boolean',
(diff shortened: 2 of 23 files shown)


framework/src/modules/dpos_v2/module.ts
```diff
@@ -131,6 +136,39 @@ export class DPoSModule extends BaseModule {
...
+				{
+					name: this.endpoint.getDelegate.name,
+					request: getDelegateRequestSchema,
+					response: getDelegateResponseSchema,
+				},
+				{
+					name: this.endpoint.getVoter.name,
+					request: getVoterRequestSchema,
+					response: getVoterResponseSchema,
+				},
```
mehmetegemen, line 155: I think names can be automated. The thing you use is `Function.prototype.name` right? Then maybe something can iterate over the class properties, pick methods and save us from typing the assignment? Also schema names have a pattern too, if these patterns will be consistent in the future we can entirely create endpoints property from a function.

framework/src/modules/dpos_v2/schemas.ts
```diff
@@ -535,3 +535,123 @@ export const genesisStoreSchema = {
...
+		lastGeneratedHeight: {
+			type: 'integer',
+			format: 'uint32',
+		},
+		isBanned: {
+			type: 'boolean',
+		},
+		pomHeights: {
+			type: 'array',
+			items: { type: 'integer', format: 'uint32' },
```
mehmetegemen, line 571: type: "integer" is probably aligned with JSON Schema standard but is it aligned with our standards? Because as I remember in our codec LIP we stated that type can be either "object" or "array", for other things we need to use "dataType". Or is it different for array items?

framework/src/modules/dpos_v2/schemas.ts
```diff
@@ -535,3 +535,123 @@ export const genesisStoreSchema = {
...
+};
+
+export const getDelegateRequestSchema = {
+	$id: 'modules/dpos/endpoint/getDelegateRequest',
+	type: 'object',
+	required: ['address'],
+	properties: {
+		address: {
+			type: 'string',
+			format: 'hex',
```
mehmetegemen, line 587: Again, using this capabilities of AJV may not be aligned with our standards and may make "isHexString" function in validator library obsolote. Either let's introduce these capabilities in an informational LIP or add type predicate to "isHexString". If we go to the route of using AJV in a complex way soon everywhere can be "oneOf", "anyOf", we must have specs.

framework/src/modules/reward/schemas.ts
```diff
@@ -46,3 +46,27 @@ export const configSchema = {
 	},
 	required: ['tokenIDReward', 'offset', 'distance', 'brackets'],
 };
+
+export const getDefaultRewardAtHeightRequestSchema = {
+	$id: '/reward/endpoint/getDefaultRewardAtHeightRequest',
```
mehmetegemen, line 51: `$id` has leading `/` in here but this is not the case in other places in this PR.

framework/src/modules/dpos_v2/schemas.ts
shuse2: Endpoint request/response is not encoded using our codec. therefore, the schema is for JSON format

framework/src/modules/dpos_v2/module.ts
shuse2: name can be automated, but I couldn't think of any good way to handle request/response schema. Any suggestion?

framework/src/modules/dpos_v2/schemas.ts
shuse2: Same as above. For JSON validation, standard JSON schema is used

framework/src/modules/reward/schemas.ts
shuse2: I check the code, and it seems this is quite inconsistent everywhere. Ill create another issue for refactoring

shuse2: Created https://github.com/LiskHQ/lisk-sdk/issues/7173

framework/src/modules/dpos_v2/schemas.ts
mehmetegemen: I know, but confusing thing is both schemas are living in the same domain, and I am not sure if this is a good decision. Because by using JSON Schema standard over our standard in the same domain(resource path), you break the constraint of codec standard. This has one implication: in future if you want to parse or process all schemas, you set will be the JSON Schema superset for the logic of processing. There is nothing to distinguish a codec schema from endpoint schema here, and it will be a problem if we want distinguish it in the future. I don't know we will ever need it, but I am informing you.

framework/src/modules/dpos_v2/module.ts
mitsuaki-u: I think it would be better not to automate as it might make things more fragile later on.

framework/src/modules/dpos_v2/schemas.ts
shuse2: we are actually using both already (for example, for the config or other validation)
It's nice to have one schema definition, but currently we need both.
If you want, maybe you can create new issue?

mehmetegemen: I think it's something I should meditate on more, I'll act if there is a need for further action. Resolving for now.

mitsuaki-u (approved):
LGTM!

mehmetegemen (approved):

Issue #7039, opened 2022-02-22
Implement CrossChainUpdateCommand for SidechainInteroperabilityModule


ishantiw:
### Description

For `SidechainInteroperabilityModule`,
- Implement `verify(context: CommandVerifyContext): VerificationResult` of `CrossChainUpdateCMD` as follows,
  - [Liveness of Partner Chain](https://github.com/LiskHQ/lips/blob/aced79e3a90c893f25c8f0735ce73732f17cced6/proposals/lip-0053.md#liveness-of-partner-chain), make sure to use sidechain `isLive` function
  - [Liveness Requirement for the First CCU](https://github.com/LiskHQ/lips/blob/aced79e3a90c893f25c8f0735ce73732f17cced6/proposals/lip-0053.md#liveness-requirement-for-the-first-ccu)
  - [Certificate and Validators Update Validity](https://github.com/LiskHQ/lips/blob/aced79e3a90c893f25c8f0735ce73732f17cced6/proposals/lip-0053.md#certificate-and-validators-update-validity)
  - [InboxUpdate Validity](https://github.com/LiskHQ/lips/blob/aced79e3a90c893f25c8f0735ce73732f17cced6/proposals/lip-0053.md#inboxupdate-validity)

- Implement `execute(context: CommandExecuteContext): void` of `CrossChainUpdateCMD` as below,
 - Follow sidechain CCU execution logic [LIP0053:Sidechain CCU#execution](https://github.com/LiskHQ/lips/blob/aced79e3a90c893f25c8f0735ce73732f17cced6/proposals/lip-0053.md#cross-chain-updates-posted-on-sidechains)
 - Follow common processing logic as described [LIP0053:Common Processing CCU](https://github.com/LiskHQ/lips/blob/aced79e3a90c893f25c8f0735ce73732f17cced6/proposals/lip-0053.md#common-processing)

### Acceptance Criteria

- Unit tests for all the code paths for `verify()`
- Unit tests for all the code paths for `execute()`

PR #7165, opened 2022-05-12
Implement cross chain command for interoperability modules - Closes #7038 & #7039

### LINKED ISSUE #7038: Implement CrossChainUpdateCommand for MainchainInteroperabilityModule
### Description

For `MainchainInteroperabilityModule`,
- Implement `verify(context: CommandVerifyContext): VerificationResult` of `CrossChainUpdateCMD` as follows,
  - [Liveness of Partner Chain](https://github.com/LiskHQ/lips/blob/aced79e3a90c893f25c8f0735ce73732f17cced6/proposals/lip-0053.md#liveness-of-partner-chain), make sure to use mainchain `isLive` function
  - [Liveness Requirement for the First CCU](https://github.com/LiskHQ/lips/blob/aced79e3a90c893f25c8f0735ce73732f17cced6/proposals/lip-0053.md#liveness-requirement-for-the-first-ccu)
  - [Certificate and Validators Update Validity](https://github.com/LiskHQ/lips/blob/aced79e3a90c893f25c8f0735ce73732f17cced6/proposals/lip-0053.md#certificate-and-validators-update-validity)
  - [InboxUpdate Validity](https://github.com/LiskHQ/lips/blob/aced79e3a90c893f25c8f0735ce73732f17cced6/proposals/lip-0053.md#inboxupdate-validity)

- Implement `execute(context: CommandExecuteContext): void` of `CrossChainUpdateCMD` as below,
 - Follow mainchain CCU execution logic [LIP0053:Mainchain CCU#execution](https://github.com/LiskHQ/lips/blob/aced79e3a90c893f25c8f0735ce73732f17cced6/proposals/lip-0053.md#cross-chain-updates-posted-on-mainchain)
 - Follow common processing logic as described [LIP0053:Common Processing CCU](https://github.com/LiskHQ/lips/blob/aced79e3a90c893f25c8f0735ce73732f17cced6/proposals/lip-0053.md#common-processing)
### Acceptance Criteria

- Unit tests for all the code paths for `verify()`
- Unit tests for all the code paths for `execute()`

ishantiw:
### What was the problem?

This PR resolves #7038 & #7039

### How was it solved?

[♻️ Export calculateRootFromRightWitness from lisk-tree](https://github.com/LiskHQ/lisk-sdk/commit/8060c5de246e3bbad8faf3b8c3449afa014bd40c)

[🌱 Implement ccu on mainchain](https://github.com/LiskHQ/lisk-sdk/commit/9bc38478467c67949b51d7fc04b6e5997b57e80a)

[🌱 Implement CCU command on sidechain](https://github.com/LiskHQ/lisk-sdk/commit/c8e152a78e735d76fcee2c7288a91f2ed34c8b28)

### How was it tested?

`npm run test:unit interoperability`

### DIFF
--- a/framework/src/modules/interoperability/mainchain/commands/cc_update.ts
+++ b/framework/src/modules/interoperability/mainchain/commands/cc_update.ts
@@ -123,6 +123,17 @@ export class MainchainCCUpdateCommand extends BaseInteroperabilityCommand {
 			return certificateValidity;
 		}
 
+		const partnerValidatorStore = context.getStore(this.moduleID, STORE_PREFIX_CHAIN_VALIDATORS);
+		const partnerValidators = await partnerValidatorStore.getWithSchema<ChainValidators>(
+			partnerChainIDBuffer,
+			chainValidatorsSchema,
+		);
+		// If params contains a non-empty activeValidatorsUpdate
+		const validatorsHashValidity = checkValidatorsHashWithCertificate(txParams, partnerValidators);
+		if (validatorsHashValidity.error) {
+			return validatorsHashValidity;
+		}
+
 		// If params contains a non-empty activeValidatorsUpdate
 		const activeValidatorsValidity = checkActiveValidatorsUpdate(txParams);
 		if (activeValidatorsValidity.error) {
@@ -175,9 +186,6 @@ export class MainchainCCUpdateCommand extends BaseInteroperabilityCommand {
 			decodedCertificate,
 			header,
 		);
-
-		// If params contains a non-empty activeValidatorsUpdate
-		checkValidatorsHashWithCertificate(txParams, decodedCertificate, partnerValidators);
 
 		// CCM execution
 		const beforeSendContext = createCCMsgBeforeSendContext({
--- a/framework/src/modules/interoperability/mainchain/commands/cc_update.ts
+++ b/framework/src/modules/interoperability/mainchain/commands/cc_update.ts
@@ -13,7 +13,6 @@
  */
 
 import { codec } from '@liskhq/lisk-codec';
-import { verifyWeightedAggSig } from '@liskhq/lisk-cryptography';
 import { LiskValidationError, validator } from '@liskhq/lisk-validator';
 import { certificateSchema } from '../../../../node/consensus/certificate_generation/schema';
 import { Certificate } from '../../../../node/consensus/certificate_generation/types';
@@ -31,10 +30,7 @@ import {
 	CHAIN_TERMINATED,
 	COMMAND_ID_MAINCHAIN_CCU,
 	CROSS_CHAIN_COMMAND_ID_REGISTRATION,
-	EMPTY_BYTES,
-	LIVENESS_LIMIT,
 	MAINCHAIN_ID,
-	MESSAGE_TAG_CERTIFICATE,
 	STORE_PREFIX_CHAIN_DATA,
 	STORE_PREFIX_CHAIN_VALIDATORS,
 	STORE_PREFIX_CHANNEL_DATA,
@@ -57,12 +53,14 @@ import {
 } from '../../types';
 import {
 	checkActiveValidatorsUpdate,
+	checkCertificateTimestampAndSignature,
 	checkCertificateValidity,
 	checkInboxUpdateValidity,
 	checkLivenessRequirementFirstCCU,
-	computeValidatorsHash,
+	checkValidatorsHashWithCertificate,
+	checkValidCertificateLiveness,
+	commonCCUExecutelogic,
 	getIDAsKeyForStore,
-	updateActiveValidators,
 	validateFormat,
 } from '../../utils';
 import { MainchainInteroperabilityStore } from '../store';
@@ -148,7 +146,7 @@ export class MainchainCCUpdateCommand extends BaseInteroperabilityCommand {
 	public async execute(
 		context: CommandExecuteContext<CrossChainUpdateTransactionParams>,
 	): Promise<void> {
-		const { transaction, header, params: txParams } = context;
+		const { header, params: txParams } = context;
 		const chainIDBuffer = getIDAsKeyForStore(txParams.sendingChainID);
 		const partnerChainStore = context.getStore(this.moduleID, STORE_PREFIX_CHAIN_DATA);
 		const partnerChainAccount = await partnerChainStore.getWithSchema<ChainAccount>(
@@ -159,59 +157,25 @@ export class MainchainCCUpdateCommand extends BaseInteroperabilityCommand {
 		const decodedCertificate = codec.decode<Certificate>(certificateSchema, txParams.certificate);
 
 		// if the CCU also contains a non-empty inboxUpdate, check the validity of certificate with liveness check
-		if (
-			txParams.inboxUpdate &&
-			!(header.timestamp - decodedCertificate.timestamp < LIVENESS_LIMIT / 2)
-		) {
-			throw Error(
-				`Certificate is not valid as it passed Liveness limit of ${LIVENESS_LIMIT} seconds.`,
-			);
-		}
+		checkValidCertificateLiveness(txParams, header, decodedCertificate);
 
 		const partnerValidatorStore = context.getStore(this.moduleID, STORE_PREFIX_CHAIN_VALIDATORS);
 		const partnerValidators = await partnerValidatorStore.getWithSchema<ChainValidators>(
 			chainIDBuffer,
 			chainValidatorsSchema,
 		);
-		const { activeValidators, certificateThreshold } = partnerValidators;
 
 		// Certificate and Validators Update Validity
-		if (!txParams.certificate.equals(EMPTY_BYTES)) {
-			const verifySignature = verifyWeightedAggSig(
-				activeValidators.map(v => v.blsKey),
-				decodedCertificate.aggregationBits as Buffer,
-				decodedCertificate.signature as Buffer,
-				MESSAGE_TAG_CERTIFICATE.toString('hex'),
-				partnerChainAccount.networkID,
-				txParams.certificate,
-				activeValidators.map(v => v.bftWeight),
-				certificateThreshold,
-			);
-
-			if (!verifySignature || decodedCertificate.timestamp >= header.timestamp)
-				throw Error(
-					`Certificate is invalid due to invalid last certified height or timestamp or signature.`,
-				);
-		}
+		checkCertificateTimestampAndSignature(
+			txParams,
+			partnerValidators,
+			partnerChainAccount,
+			decodedCertificate,
+			header,
+		);
 
 		// If params contains a non-empty activeValidatorsUpdate
-		if (
-			txParams.activeValidatorsUpdate.length !== 0 ||
-			txParams.newCertificateThreshold > BigInt(0)
-		) {
-			const newActiveValidators = updateActiveValidators(
-				activeValidators,
-				txParams.activeValidatorsUpdate,
-			);
-			const validatorsHash = computeValidatorsHash(
-				newActiveValidators,
-				txParams.newCertificateThreshold || certificateThreshold,
-			);
-
-			if (!decodedCertificate.validatorsHash.equals(validatorsHash)) {
-				throw new Error('Validators hash is incorrect given in the certificate.');
-			}
-		}
+		checkValidatorsHashWithCertificate(txParams, decodedCertificate, partnerValidators);
 
 		// CCM execution
 		const interoperabilityStore = this.getInteroperabilityStore(context.getStore);
@@ -304,35 +268,16 @@ export class MainchainCCUpdateCommand extends BaseInteroperabilityCommand {
 				);
 			}
 		}
-		// Common ccm execution logic
-		const newActiveValidators = updateActiveValidators(
-			activeValidators,
-			txParams.activeValidatorsUpdate,
-		);
-		partnerValidators.activeValidators = newActiveValidators;
-		if (txParams.newCertificateThreshold !== BigInt(0)) {
-			partnerValidators.certificateThreshold = txParams.newCertificateThreshold;
-		}
-		await partnerValidatorStore.setWithSchema(
+		// Common ccu execution logic
+		await commonCCUExecutelogic({
+			certificate: decodedCertificate,
 			chainIDBuffer,
+			context,
+			partnerChainAccount,
+			partnerChainStore,
+			partnerValidatorStore,
 			partnerValidators,
-			chainValidatorsSchema,
-		);
-		if (!txParams.certificate.equals(EMPTY_BYTES)) {
-			partnerChainAccount.lastCertificate.stateRoot = decodedCertificate.stateRoot;
-			partnerChainAccount.lastCertificate.timestamp = decodedCertificate.timestamp;
-			partnerChainAccount.lastCertificate.height = decodedCertificate.height;
-			partnerChainAccount.lastCertificate.validatorsHash = decodedCertificate.validatorsHash;
-		}
-
-		await partnerChainStore.setWithSchema(chainIDBuffer, partnerChainAccount, chainAccountSchema);
-
-		const partnerChannelStore = context.getStore(transaction.moduleID, STORE_PREFIX_CHANNEL_DATA);
-		const partnerChannelData = await partnerChannelStore.getWithSchema<ChannelData>(
-			chainIDBuffer,
-			channelSchema,
-		);
-		await partnerChainStore.setWithSchema(chainIDBuffer, partnerChannelData, channelSchema);
+		});
 	}
 
 	protected getInteroperabilityStore(getStore: StoreCallback): MainchainInteroperabilityStore {
--- a/framework/src/modules/interoperability/mainchain/commands/cc_update.ts
+++ b/framework/src/modules/interoperability/mainchain/commands/cc_update.ts
@@ -52,7 +52,7 @@ import {
 } from '../../types';
 import {
 	checkActiveValidatorsUpdate,
-	checkCertificateTimestampAndSignature,
+	checkCertificateTimestamp,
 	checkCertificateValidity,
 	checkInboxUpdateValidity,
 	checkLivenessRequirementFirstCCU,
@@ -62,6 +62,7 @@ import {
 	getIDAsKeyForStore,
 	isInboxUpdateEmpty,
 	validateFormat,
+	verifyCertificateSignature,
 } from '../../utils';
 import { MainchainInteroperabilityStore } from '../store';
 
@@ -128,7 +129,7 @@ export class MainchainCCUpdateCommand extends BaseInteroperabilityCommand {
 			partnerChainIDBuffer,
 			chainValidatorsSchema,
 		);
-		// If params contains a non-empty activeValidatorsUpdate
+		// If params contains a non-empty activeValidatorsUpdate and non-empty certificate
 		const validatorsHashValidity = checkValidatorsHashWithCertificate(txParams, partnerValidators);
 		if (validatorsHashValidity.error) {
 			return validatorsHashValidity;
@@ -138,6 +139,16 @@ export class MainchainCCUpdateCommand extends BaseInteroperabilityCommand {
 		const activeValidatorsValidity = checkActiveValidatorsUpdate(txParams);
 		if (activeValidatorsValidity.error) {
 			return activeValidatorsValidity;
+		}
+
+		// When certificate is non-empty
+		const verifyCertificateSignatureResult = verifyCertificateSignature(
+			txParams,
+			partnerValidators,
+			partnerChainAccount,
+		);
+		if (verifyCertificateSignatureResult.error) {
+			return verifyCertificateSignatureResult;
 		}
 
 		const partnerChannelStore = context.getStore(transaction.moduleID, STORE_PREFIX_CHANNEL_DATA);
@@ -178,14 +189,8 @@ export class MainchainCCUpdateCommand extends BaseInteroperabilityCommand {
 			chainValidatorsSchema,
 		);
 
-		// Certificate and Validators Update Validity
-		checkCertificateTimestampAndSignature(
-			txParams,
-			partnerValidators,
-			partnerChainAccount,
-			decodedCertificate,
-			header,
-		);
+		// Certificate timestamp Validity
+		checkCertificateTimestamp(txParams, decodedCertificate, header);
 
 		// CCM execution
 		const beforeSendContext = createCCMsgBeforeSendContext({
--- a/framework/src/modules/interoperability/mainchain/commands/cc_update.ts
+++ b/framework/src/modules/interoperability/mainchain/commands/cc_update.ts
@@ -191,7 +191,7 @@ export class MainchainCCUpdateCommand extends BaseInteroperabilityCommand {
 		try {
 			decodedCCMs = txParams.inboxUpdate.crossChainMessages.map(ccm => ({
 				serialized: ccm,
-				deserilized: codec.decode<CCMsg>(ccmSchema, ccm),
+				deserialized: codec.decode<CCMsg>(ccmSchema, ccm),
 			}));
 		} catch (err) {
 			await interoperabilityStore.terminateChainInternal(
@@ -201,11 +201,14 @@ export class MainchainCCUpdateCommand extends BaseInteroperabilityCommand {
 
 			throw err;
 		}
-		if (partnerChainAccount.status === CHAIN_REGISTERED && !isInboxUpdateEmpty) {
+		if (
+			partnerChainAccount.status === CHAIN_REGISTERED &&
+			!isInboxUpdateEmpty(txParams.inboxUpdate)
+		) {
 			// If the first CCM in inboxUpdate is a registration CCM
 			if (
-				decodedCCMs[0].deserilized.crossChainCommandID === CROSS_CHAIN_COMMAND_ID_REGISTRATION &&
-				decodedCCMs[0].deserilized.receivingChainID === MAINCHAIN_ID
+				decodedCCMs[0].deserialized.crossChainCommandID === CROSS_CHAIN_COMMAND_ID_REGISTRATION &&
+				decodedCCMs[0].deserialized.receivingChainID === MAINCHAIN_ID
 			) {
 				partnerChainAccount.status = CHAIN_ACTIVE;
 			} else {
@@ -219,28 +222,31 @@ export class MainchainCCUpdateCommand extends BaseInteroperabilityCommand {
 		}
 
 		for (const ccm of decodedCCMs) {
-			if (txParams.sendingChainID !== ccm.deserilized.sendingChainID) {
+			if (txParams.sendingChainID !== ccm.deserialized.sendingChainID) {
 				await interoperabilityStore.terminateChainInternal(
 					txParams.sendingChainID,
 					beforeSendContext,
 				);
+
+				continue;
 			}
 			try {
-				validateFormat(ccm.deserilized);
+				validateFormat(ccm.deserialized);
 			} catch (error) {
 				await interoperabilityStore.terminateChainInternal(
 					txParams.sendingChainID,
 					beforeSendContext,
 				);
+
+				continue;
 			}
 			await interoperabilityStore.appendToInboxTree(
 				getIDAsKeyForStore(txParams.sendingChainID),
 				ccm.serialized,
 			);
-
-			if (ccm.deserilized.receivingChainID !== MAINCHAIN_ID) {
+			if (ccm.deserialized.receivingChainID !== MAINCHAIN_ID) {
 				await interoperabilityStore.forward({
-					ccm: ccm.deserilized,
+					ccm: ccm.deserialized,
 					ccu: txParams,
 					eventQueue: context.eventQueue,
 					feeAddress: context.transaction.senderAddress,
@@ -252,7 +258,7 @@ export class MainchainCCUpdateCommand extends BaseInteroperabilityCommand {
 			} else {
 				await interoperabilityStore.apply(
 					{
-						ccm: ccm.deserilized,
+						ccm: ccm.deserialized,
 						ccu: txParams,
 						eventQueue: context.eventQueue,
 						feeAddress: context.transaction.senderAddress,
--- a/framework/src/modules/interoperability/mainchain/commands/cc_update.ts
+++ b/framework/src/modules/interoperability/mainchain/commands/cc_update.ts
@@ -61,6 +61,7 @@ import {
 	checkValidCertificateLiveness,
 	commonCCUExecutelogic,
 	getIDAsKeyForStore,
+	isInboxUpdateEmpty,
 	validateFormat,
 } from '../../utils';
 import { MainchainInteroperabilityStore } from '../store';
@@ -178,32 +179,37 @@ export class MainchainCCUpdateCommand extends BaseInteroperabilityCommand {
 		checkValidatorsHashWithCertificate(txParams, decodedCertificate, partnerValidators);
 
 		// CCM execution
+		const beforeSendContext = createCCMsgBeforeSendContext({
+			feeAddress: context.transaction.senderAddress,
+			eventQueue: context.eventQueue,
+			getAPIContext: context.getAPIContext,
+			logger: context.logger,
+			networkIdentifier: context.networkIdentifier,
+			getStore: context.getStore,
+		});
 		const interoperabilityStore = this.getInteroperabilityStore(context.getStore);
-		const decodedCCMs = txParams.inboxUpdate.crossChainMessages.map(ccm => ({
-			serialized: ccm,
-			deserilized: codec.decode<CCMsg>(ccmSchema, ccm),
-		}));
-		if (
-			partnerChainAccount.status === CHAIN_REGISTERED &&
-			txParams.inboxUpdate.crossChainMessages.length !== 0 &&
-			txParams.inboxUpdate.messageWitness.siblingHashes.length !== 0 &&
-			txParams.inboxUpdate.outboxRootWitness.siblingHashes.length !== 0
-		) {
+		let decodedCCMs;
+		try {
+			decodedCCMs = txParams.inboxUpdate.crossChainMessages.map(ccm => ({
+				serialized: ccm,
+				deserilized: codec.decode<CCMsg>(ccmSchema, ccm),
+			}));
+		} catch (err) {
+			await interoperabilityStore.terminateChainInternal(
+				txParams.sendingChainID,
+				beforeSendContext,
+			);
+
+			throw err;
+		}
+		if (partnerChainAccount.status === CHAIN_REGISTERED && !isInboxUpdateEmpty) {
 			// If the first CCM in inboxUpdate is a registration CCM
 			if (
-				decodedCCMs.length > 0 &&
 				decodedCCMs[0].deserilized.crossChainCommandID === CROSS_CHAIN_COMMAND_ID_REGISTRATION &&
 				decodedCCMs[0].deserilized.receivingChainID === MAINCHAIN_ID
 			) {
 				partnerChainAccount.status = CHAIN_ACTIVE;
 			} else {
-				const beforeSendContext = createBeforeSendCCMsgAPIContext({
-					feeAddress: context.transaction.senderAddress,
-					eventQueue: context.eventQueue,
-					getAPIContext: context.getAPIContext,
-					logger: context.logger,
-					networkIdentifier: context.networkIdentifier,
-				});
 				await interoperabilityStore.terminateChainInternal(
 					txParams.sendingChainID,
 					beforeSendContext,
@@ -214,14 +220,6 @@ export class MainchainCCUpdateCommand extends BaseInteroperabilityCommand {
 		}
 
 		for (const ccm of decodedCCMs) {
-			const beforeSendContext = createCCMsgBeforeSendContext({
-				feeAddress: context.transaction.senderAddress,
-				eventQueue: context.eventQueue,
-				getAPIContext: context.getAPIContext,
-				logger: context.logger,
-				networkIdentifier: context.networkIdentifier,
-				getStore: context.getStore,
-			});
 			if (txParams.sendingChainID !== ccm.deserilized.sendingChainID) {
 				await interoperabilityStore.terminateChainInternal(
 					txParams.sendingChainID,
--- a/framework/src/modules/interoperability/mainchain/commands/cc_update.ts
+++ b/framework/src/modules/interoperability/mainchain/commands/cc_update.ts
@@ -73,7 +73,7 @@ export class MainchainCCUpdateCommand extends BaseInteroperabilityCommand {
 	public async verify(
 		context: CommandVerifyContext<CrossChainUpdateTransactionParams>,
 	): Promise<VerificationResult> {
-		const { params: txParams, transaction } = context;
+		const { params: txParams, transaction, getStore } = context;
 		const errors = validator.validate(crossChainUpdateTransactionParams, context.params);
 
 		if (errors.length > 0) {
@@ -84,7 +84,7 @@ export class MainchainCCUpdateCommand extends BaseInteroperabilityCommand {
 		}
 
 		const partnerChainIDBuffer = getIDAsKeyForStore(txParams.sendingChainID);
-		const partnerChainStore = context.getStore(transaction.moduleID, STORE_PREFIX_CHAIN_DATA);
+		const partnerChainStore = getStore(transaction.moduleID, STORE_PREFIX_CHAIN_DATA);
 		const partnerChainAccount = await partnerChainStore.getWithSchema<ChainAccount>(
 			partnerChainIDBuffer,
 			chainAccountSchema,
@@ -97,13 +97,15 @@ export class MainchainCCUpdateCommand extends BaseInteroperabilityCommand {
 				error: new Error(`Sending partner chain ${txParams.sendingChainID} is terminated.`),
 			};
 		}
-		const interoperabilityStore = this.getInteroperabilityStore(context.getStore as StoreCallback);
-		const isChainLive = await interoperabilityStore.isLive(partnerChainIDBuffer, Date.now());
-		if (partnerChainAccount.status === CHAIN_ACTIVE && !isChainLive) {
-			return {
-				status: VerifyStatus.FAIL,
-				error: new Error(`Sending partner chain ${txParams.sendingChainID} is not live.`),
-			};
+		const interoperabilityStore = this.getInteroperabilityStore(getStore);
+		if (partnerChainAccount.status === CHAIN_ACTIVE) {
+			const isChainLive = await interoperabilityStore.isLive(partnerChainIDBuffer, Date.now());
+			if (!isChainLive) {
+				return {
+					status: VerifyStatus.FAIL,
+					error: new Error(`Sending partner chain ${txParams.sendingChainID} is not live.`),
+				};
+			}
 		}
 
 		// Section: Liveness Requirement for the First CCU
--- a/framework/src/modules/interoperability/mainchain/commands/cc_update.ts
+++ b/framework/src/modules/interoperability/mainchain/commands/cc_update.ts
@@ -13,8 +13,7 @@
  */
 
 import { codec } from '@liskhq/lisk-codec';
-import { hash, verifyWeightedAggSig } from '@liskhq/lisk-cryptography';
-import { regularMerkleTree, sparseMerkleTree } from '@liskhq/lisk-tree';
+import { verifyWeightedAggSig } from '@liskhq/lisk-cryptography';
 import { LiskValidationError, validator } from '@liskhq/lisk-validator';
 import { certificateSchema } from '../../../../node/consensus/certificate_generation/schema';
 import { Certificate } from '../../../../node/consensus/certificate_generation/types';
@@ -36,12 +35,9 @@ import {
 	LIVENESS_LIMIT,
 	MAINCHAIN_ID,
 	MESSAGE_TAG_CERTIFICATE,
-	SMT_KEY_LENGTH,
 	STORE_PREFIX_CHAIN_DATA,
 	STORE_PREFIX_CHAIN_VALIDATORS,
 	STORE_PREFIX_CHANNEL_DATA,
-	STORE_PREFIX_OUTBOX_ROOT,
-	VALID_BLS_KEY_LENGTH,
 } from '../../constants';
 import { createCCMsgBeforeSendContext } from '../../context';
 import {
@@ -60,9 +56,12 @@ import {
 	StoreCallback,
 } from '../../types';
 import {
+	checkActiveValidatorsUpdate,
+	checkCertificateValidity,
+	checkInboxUpdateValidity,
+	checkLivenessRequirementFirstCCU,
 	computeValidatorsHash,
 	getIDAsKeyForStore,
-	rawStateStoreKey,
 	updateActiveValidators,
 	validateFormat,
 } from '../../utils';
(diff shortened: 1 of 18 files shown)


framework/src/modules/interoperability/mainchain/commands/cc_update.ts
```diff
@@ -12,18 +12,433 @@
...
+			}
+		}
+		const decodedCertificate = codec.decode<Certificate>(certificateSchema, txParams.certificate);
+		const invalidCertificateErrors = validator.validate(certificateSchema, decodedCertificate);
+		if (invalidCertificateErrors.length > 0) {
+			return {
+				status: VerifyStatus.FAIL,
+				error: new Error('Certificate does not follow valid certificate schema.'),
+			};
+		}
```
shuse2, line 138: missing non-empty certificate case here?

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
```diff
@@ -12,18 +12,433 @@
...
+					status: VerifyStatus.FAIL,
+					error: new Error('Certificate cannot be empty.'),
+				};
+			}
+
+			// params.activeValidatorsUpdate has the correct format
+			for (let i = 0; i < txParams.activeValidatorsUpdate.length; i += 1) {
+				const currentValidator = txParams.activeValidatorsUpdate[i];
+				const nextValidator = txParams.activeValidatorsUpdate[i + 1];
+				if (currentValidator.blsKey.byteLength !== VALID_BLS_KEY_LENGTH) {
```
shuse2, line 156: Not related to this PR, but we should align on the usage of `byteLength` vs `length`. (I think `byteLength` is correct, but `length` is used everywhere else)

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
```diff
@@ -12,18 +12,433 @@
...
+			);
+			if (!isSMTRootValid) {
+				return {
+					status: VerifyStatus.FAIL,
+					error: new Error(
+						'Failed at verifying state root when messageWitness and certificate are non-empty.',
+					),
+				};
+			}
+		} else if (!txParams.certificate) {
```
shuse2, line 226:
```suggestion
		} else {
```

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
```diff
@@ -12,18 +12,433 @@
...
+		const partnerChannelData = await partnerChannelStore.getWithSchema<ChannelData>(
+			partnerChainIDBuffer,
+			channelSchema,
+		);
+		// InboxUpdate Validity
+		const { crossChainMessages, messageWitness, outboxRootWitness } = txParams.inboxUpdate;
+		const ccmHashes = crossChainMessages.map(ccm => hash(ccm));
+
+		let newInboxRoot;
+		let newInboxAppendPath;
```
shuse2, line 181: to calculate the root recursively, we need to initialize with `partnerChannelData.inbox` data and use this variable when calculating the root

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
```diff
@@ -12,18 +12,433 @@
...
+		let newInboxSize;
+		for (const ccm of ccmHashes) {
+			const { appendPath, size } = regularMerkleTree.calculateMerkleRoot({
+				value: ccm,
+				appendPath: partnerChannelData.inbox.appendPath,
+				size: partnerChannelData.inbox.size,
+			});
+			newInboxAppendPath = appendPath;
+			newInboxSize = size;
```
shuse2, line 190: `newInboxRoot` also should be assigned here?

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
```diff
@@ -12,18 +12,433 @@
...
+		}
+		if (txParams.certificate) {
+			// If inboxUpdate contains a non-empty messageWitness, then update newInboxRoot to the output
```
shuse2, line 193: `// If inboxUpdate contains a non-empty messageWitness` check is missing?

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
```diff
@@ -12,18 +12,433 @@
...
+			if (!isSMTRootValid) {
+				return {
+					status: VerifyStatus.FAIL,
+					error: new Error(
+						'Failed at verifying state root when messageWitness and certificate are non-empty.',
+					),
+				};
+			}
+		} else if (!txParams.certificate) {
+			newInboxRoot = regularMerkleTree.calculateRootFromRightWitness(
```
shuse2, line 227: here also check of `// If inboxUpdate contains a non-empty messageWitness` is missing

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
```diff
@@ -12,18 +12,433 @@
...
+		if (
+			txParams.inboxUpdate &&
+			!(header.timestamp - decodedCertificate.timestamp < LIVENESS_LIMIT / 2)
+		) {
+			throw Error(
+				`Certificate is not valid as it passed Liveness limit of ${LIVENESS_LIMIT} seconds.`,
+			);
+		}
+		// Certificate and Validators Update Validity
+		const verifySignature = verifyWeightedAggSig(
```
shuse2, line 301: this should be checked only if certificate is non empty?

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
```diff
@@ -12,18 +12,433 @@
...
+		const newActiveValidators = updateActiveValidators(
+			activeValidators,
+			txParams.activeValidatorsUpdate,
+		);
+		const validatorsHash = computeValidatorsHash(
+			newActiveValidators,
+			txParams.newCertificateThreshold || certificateThreshold,
+		);
+
+		if (!decodedCertificate.validatorsHash.equals(validatorsHash)) {
```
shuse2, line 331: this should only be checked if non-empty `activeValidatorsUpdate`?

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
```diff
@@ -12,18 +12,433 @@
...
+		if (partnerChainAccount.status === CHAIN_REGISTERED) {
+			// If the first CCM in inboxUpdate is a registration CCM
+			if (
+				decodedCCMs[0].deserilized.crossChainCommandID === CROSS_CHAIN_COMMAND_ID_REGISTRATION &&
+				decodedCCMs[0].deserilized.receivingChainID === MAINCHAIN_ID
+			) {
+				partnerChainAccount.status = CHAIN_ACTIVE;
+			} else {
+				const beforeSendContext = createBeforeSendCCMsgAPIContext({
+					feeAddress: context.transaction.senderAddress,
```
shuse2, line 350: Out of this PR scope, but `feeAddress` was bit confusing during the token module implementation. maybe we could jest keep it as senderAddress.
Also, separate topic, token module needed ccm size, so we should add that as well

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
```diff
@@ -12,18 +12,433 @@
...
+		// CCM execution
+		const interoperabilityStore = this.getInteroperabilityStore(context.getStore);
+		const decodedCCMs = txParams.inboxUpdate.crossChainMessages.map(ccm => ({
+			serialized: ccm,
+			deserilized: codec.decode<CCMsg>(ccmSchema, ccm),
+		}));
+		if (partnerChainAccount.status === CHAIN_REGISTERED) {
+			// If the first CCM in inboxUpdate is a registration CCM
+			if (
+				decodedCCMs[0].deserilized.crossChainCommandID === CROSS_CHAIN_COMMAND_ID_REGISTRATION &&
```
shuse2, line 344: did we check somewhere that length of CCM is at least 1?

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
```diff
@@ -12,18 +12,433 @@
...
+						logger: context.logger,
+						networkIdentifier: context.networkIdentifier,
+					},
+					this.ccCommands,
+				);
+			}
+		}
+
+		// Common ccm execution logic
+		await this._bftAPI.setBFTParameters(
```
shuse2, line 420: I think interoperability shouldn't set the BFT params. it should call [updateActiveValidators](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0053.md#update-active-validators)

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
```diff
@@ -12,18 +12,433 @@
...
+			throw new Error('Validators hash is incorrect given in the certificate.');
+		}
+
+		// CCM execution
+		const interoperabilityStore = this.getInteroperabilityStore(context.getStore);
+		const decodedCCMs = txParams.inboxUpdate.crossChainMessages.map(ccm => ({
+			serialized: ccm,
+			deserilized: codec.decode<CCMsg>(ccmSchema, ccm),
+		}));
+		if (partnerChainAccount.status === CHAIN_REGISTERED) {
```
shuse2, line 341: this condition also should check if `inboxUpdate` is non empty?

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
```diff
@@ -12,18 +12,433 @@
...
+		}
+
+		// Common ccm execution logic
+		await this._bftAPI.setBFTParameters(
+			context.getAPIContext(),
+			precommitThreshold,
+			txParams.newCertificateThreshold ?? certificateThreshold,
+			updatedValidators,
+		);
+		partnerChainAccount.lastCertificate.stateRoot = decodedCertificate.stateRoot;
```
shuse2, line 426: there should be non-empty certificate check here?

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
```diff
@@ -12,18 +12,433 @@
...
+			}
+		}
+
+		// Common ccm execution logic
+		await this._bftAPI.setBFTParameters(
+			context.getAPIContext(),
+			precommitThreshold,
+			txParams.newCertificateThreshold ?? certificateThreshold,
+			updatedValidators,
+		);
```
shuse2, line 425: `newCertificateThreshold` check and update is missing

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
```diff
@@ -12,18 +12,433 @@
...
+		partnerChainAccount.lastCertificate.stateRoot = decodedCertificate.stateRoot;
+		partnerChainAccount.lastCertificate.timestamp = decodedCertificate.timestamp;
+		partnerChainAccount.lastCertificate.height = decodedCertificate.height;
+		partnerChainAccount.lastCertificate.validatorsHash = decodedCertificate.validatorsHash;
+
```
shuse2, line 430: `partnerChainOutboxRoot` check and set is missing

framework/src/modules/interoperability/sidechain/commands/cc_update.ts
```diff
@@ -12,18 +12,421 @@
...
+	private _validatorsAPI!: ValidatorsAPI;
 
-	// eslint-disable-next-line @typescript-eslint/require-await
-	public async execute(_context: CommandExecuteContext<unknown>): Promise<void> {
-		throw new Error('Method not implemented.');
+	public addDependencies(args: CrossChainCommandDependencies) {
+		this._bftAPI = args.bftAPI;
+		this._validatorsAPI = args.validatorsAPI;
+	}
+	public async verify(
```
shuse2, line 82: I think we can unify the verify code somehow with mainchain verify since the logic is exactly the same

framework/src/modules/interoperability/sidechain/commands/cc_update.ts
```diff
@@ -12,18 +12,421 @@
...
+			txParams.activeValidatorsUpdate,
+		);
+		const validatorsHash = computeValidatorsHash(
+			newActiveValidators,
+			txParams.newCertificateThreshold || certificateThreshold,
+		);
+
+		if (!decodedCertificate.validatorsHash.equals(validatorsHash)) {
+			throw new Error('Validators hash is incorrect given in the certificate.');
+		}
```
shuse2, line 333: also, common execution logic between mainchain/sidechain should be unified

framework/src/modules/interoperability/sidechain/commands/cc_update.ts
```diff
@@ -12,18 +12,421 @@
...
+
+		// CCM execution
+		const interoperabilityStore = this.getInteroperabilityStore(context.getStore);
+		const decodedCCMs = txParams.inboxUpdate.crossChainMessages.map(ccm => ({
+			serialized: ccm,
+			deserilized: codec.decode<CCMsg>(ccmSchema, ccm),
+		}));
+		if (partnerChainAccount.status === CHAIN_REGISTERED) {
```
shuse2, line 341: inboxUpdate non empty check is missing here as well?

framework/src/modules/interoperability/sidechain/commands/cc_update.ts
```diff
@@ -12,18 +12,421 @@
...
+			// If the first CCM in inboxUpdate is a registration CCM
+			if (
+				decodedCCMs[0].deserilized.crossChainCommandID === CROSS_CHAIN_COMMAND_ID_REGISTRATION &&
```
shuse2, line 344: same as mainchain, but CCM length check?

framework/src/modules/interoperability/sidechain/commands/cc_update.ts
```diff
@@ -12,18 +12,421 @@
...
+					feeAddress: context.transaction.senderAddress,
+					eventQueue: context.eventQueue,
+					getAPIContext: context.getAPIContext,
+					logger: context.logger,
+					networkIdentifier: context.networkIdentifier,
+				});
+				await interoperabilityStore.terminateChainInternal(
+					txParams.sendingChainID,
+					beforeSendContext,
+				);
```
shuse2, line 359: it should return here. On the mainchain side, it doesn't mention ` Exit the transaction processing`. Let's make sure if that's ok

framework/src/modules/interoperability/sidechain/commands/cc_update.ts
```diff
@@ -12,18 +12,421 @@
...
+					getStore: context.getStore,
+					logger: context.logger,
+					networkIdentifier: context.networkIdentifier,
+				},
+				this.ccCommands,
+			);
+		}
+
+		// Common ccm execution logic
+		await this._bftAPI.setBFTParameters(
```
shuse2, line 408: same comments as above to unify the code

shuse2 (changes requested):

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
ishantiw: The LIP says `partnerValidators.activeValidators = updateActiveValidators(partnerValidators.activeValidators, activeValidatorsUpdate)` I'm not sure where we set `partnerValidators.activeValidators`, [updateActiveValidators](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0053.md#update-active-validators) is already called above but it doesn't change anything, it simply returns a list but here it asks to set it in partners

framework/src/modules/interoperability/sidechain/commands/cc_update.ts
ishantiw: yes I think it should exit

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
ishantiw: If I assign then it will throw an error in deepscan coz its value is not used https://deepscan.io/dashboard/#view=project&tid=6759&pid=8869&bid=554464&subview=pull-request&prid=1233154 , we assign it later after which it is used

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
ishantiw: got confused, we have to use chain validator's store which was completely missing in the code

framework/src/modules/interoperability/sidechain/commands/cc_update.ts
ishantiw: I'll try to do it in Store class or as a utill func

framework/src/modules/interoperability/utils.ts
```diff
@@ -106,3 +146,295 @@ export const computeValidatorsHash = (
...
+			decodedCertificate.timestamp === 0
+		) {
+			return {
+				status: VerifyStatus.FAIL,
+				error: new Error('Certificate is missing required values.'),
+			};
+		}
+
+		// Last certificate height should be less than new certificate height
+		if (decodedCertificate.height <= partnerChainAccount.lastCertificate.height) {
```
shuse2, line 196: I think this should be strictly greater than

framework/src/modules/interoperability/utils.ts
```diff
@@ -106,3 +146,295 @@ export const computeValidatorsHash = (
...
+		const decodedCertificate = codec.decode<Certificate>(certificateSchema, encodedCertificate);
+		if (
+			decodedCertificate.blockID.equals(EMPTY_BYTES) ||
+			decodedCertificate.stateRoot.equals(EMPTY_BYTES) ||
+			decodedCertificate.validatorsHash.equals(EMPTY_BYTES) ||
+			decodedCertificate.timestamp === 0
+		) {
+			return {
+				status: VerifyStatus.FAIL,
+				error: new Error('Certificate is missing required values.'),
```
shuse2, line 191: [Optional] Maybe these function can return type `Error | undefined`

framework/src/modules/interoperability/utils.ts
```diff
@@ -106,3 +146,295 @@ export const computeValidatorsHash = (
...
+	return {
+		status: VerifyStatus.OK,
+	};
+};
+
+export const checkCertificateValidity = (
+	partnerChainAccount: ChainAccount,
+	encodedCertificate: Buffer,
+): VerificationResult => {
+	if (!encodedCertificate.equals(EMPTY_BYTES)) {
```
shuse2, line 181: prefer early return

framework/src/modules/interoperability/utils.ts
```diff
@@ -106,3 +146,295 @@ export const computeValidatorsHash = (
...
+};
+
+export const checkInboxUpdateValidity = (
+	txParams: CrossChainUpdateTransactionParams,
+	partnerChannelData: ChannelData,
+): VerificationResult => {
+	const partnerChainIDBuffer = getIDAsKeyForStore(txParams.sendingChainID);
+	const decodedCertificate = codec.decode<Certificate>(certificateSchema, txParams.certificate);
+	const { crossChainMessages, messageWitness, outboxRootWitness } = txParams.inboxUpdate;
+	const ccmHashes = crossChainMessages.map(ccm => hash(ccm));
```
shuse2, line 257: Just to be sure, this should be hashed here?

framework/src/modules/interoperability/utils.ts
```diff
@@ -106,3 +146,295 @@ export const computeValidatorsHash = (
...
+	let newInboxRoot;
+	let newInboxAppendPath = partnerChannelData.inbox.appendPath;
+	let newInboxSize = partnerChannelData.inbox.size;
+	for (const ccm of ccmHashes) {
+		const { appendPath, size } = regularMerkleTree.calculateMerkleRoot({
+			value: ccm,
+			appendPath: newInboxAppendPath,
+			size: newInboxSize,
+		});
+		newInboxAppendPath = appendPath;
```
shuse2, line 268: here I think newInboxRoot needs to be assigned as well

framework/src/modules/interoperability/utils.ts
```diff
@@ -106,3 +146,295 @@ export const computeValidatorsHash = (
...
+		txParams.inboxUpdate.messageWitness.siblingHashes.length !== 0 &&
+		txParams.inboxUpdate.messageWitness.partnerChainOutboxSize > 0
+	) {
+		newInboxRoot = regularMerkleTree.calculateRootFromRightWitness(
+			newInboxSize,
+			newInboxAppendPath,
+			messageWitness.siblingHashes,
+		);
+
+		const proof = {
```
shuse2, line 283: below here, I think it should be outside of the `messageWitness` if clause

framework/src/modules/interoperability/utils.ts
```diff
@@ -106,3 +146,295 @@ export const computeValidatorsHash = (
...
+	for (const ccm of ccmHashes) {
+		const { appendPath, size } = regularMerkleTree.calculateMerkleRoot({
+			value: ccm,
+			appendPath: newInboxAppendPath,
+			size: newInboxSize,
+		});
+		newInboxAppendPath = appendPath;
+		newInboxSize = size;
+	}
+	// If inboxUpdate contains a non-empty messageWitness, then update newInboxRoot to the output
```
shuse2, line 271: from the LIP, condition seems like
if cerficate is not empty && inboxUpdate is not empty
  if inboxUpdate contains non-empty message witness, update the root
check the root
elseif cerificate is empty && inboxUPdate is not empty
  if inboxUpdate contains non-empty message witness, update the root
check the root

so it could be like
```
if imboxUpdate is empty, return
if certrificate is nonempty
  if messageWitness is non-empty, update root
else
  if messageWitness is non-empty, update root

check the root
```
something like this?

framework/src/modules/interoperability/utils.ts
```diff
@@ -106,3 +146,295 @@ export const computeValidatorsHash = (
...
+		status: VerifyStatus.OK,
+	};
+};
+
+export const checkValidCertificateLiveness = (
+	txParams: CrossChainUpdateTransactionParams,
+	header: BlockHeader,
+	certificate: Certificate,
+) => {
+	if (txParams.inboxUpdate && !(header.timestamp - certificate.timestamp < LIVENESS_LIMIT / 2)) {
```
shuse2, line 340: maybe definition of non-empty `inboxUpdate` should be checked

framework/src/modules/interoperability/utils.ts
```diff
@@ -106,3 +146,295 @@ export const computeValidatorsHash = (
...
+	header: BlockHeader,
+) => {
+	// Certificate and Validators Update Validity
+	const { activeValidators, certificateThreshold } = partnerValidators;
+	if (!txParams.certificate.equals(EMPTY_BYTES)) {
+		const verifySignature = verifyWeightedAggSig(
+			activeValidators.map(v => v.blsKey),
+			certificate.aggregationBits as Buffer,
+			certificate.signature as Buffer,
+			MESSAGE_TAG_CERTIFICATE.toString('hex'),
```
shuse2, line 361: This is already `string`? also tag should be used as utf-8 Buffer?

framework/src/modules/interoperability/sidechain/commands/cc_update.ts
```diff
@@ -12,18 +12,261 @@
...
+			decodedCertificate,
+			header,
+		);
+
+		// If params contains a non-empty activeValidatorsUpdate
+		checkValidatorsHashWithCertificate(txParams, decodedCertificate, partnerValidators);
+
+		// CCM execution
+		const interoperabilityStore = this.getInteroperabilityStore(context.getStore);
+		const decodedCCMs = txParams.inboxUpdate.crossChainMessages.map(ccm => ({
```
shuse2, line 181: if CCM decode fails, it should terminate the chain?

framework/src/modules/interoperability/utils.ts
```diff
@@ -106,3 +146,295 @@ export const computeValidatorsHash = (
...
+		partnerChainAccount.lastCertificate.validatorsHash = certificate.validatorsHash;
+	}
+
+	await partnerChainStore.setWithSchema(chainIDBuffer, partnerChainAccount, chainAccountSchema);
+
+	const partnerChannelStore = context.getStore(
+		context.transaction.moduleID,
+		STORE_PREFIX_CHANNEL_DATA,
+	);
+	const partnerChannelData = await partnerChannelStore.getWithSchema<ChannelData>(
```
shuse2, line 435: Missing update logic for parnerChainOutboxRoot

framework/src/modules/interoperability/utils.ts
ishantiw: so here lastCertificateHeight >= newCertificateHeight, which should throw error, but I'll simplify to `partnerChainAccount.lastCertificate.height > decodedCertificate.height`

framework/src/modules/interoperability/utils.ts
ishantiw: I was doing that but then I saw it was looking more simplified in this form, also we need to capture different errors coming from these functions

framework/src/modules/interoperability/utils.ts
ishantiw: hmm not 100% sure but below is my understanding,
 
according to LIP0045 `appendPath: An array of hashes necessary to append new data to the tree efficiently. The default value of this property is an empty array.`

But I think when we receive it in InboxUpdate, we receive only serialized CCMs that are decoded and executed so we need to hash them to use it as appendPath to calculate the root

framework/src/modules/interoperability/utils.ts
ishantiw: since this assigned value is not used, and then reassigned later at `newInboxRoot = regularMerkleTree.calculateRootFromRightWitness(
			newInboxSize,
			newInboxAppendPath,
			messageWitness.siblingHashes,
		);`

which is eventually used it throws Deepscan error that assigned value is never used

framework/src/modules/interoperability/sidechain/commands/cc_update.ts
```diff
@@ -12,18 +12,261 @@
...
+		}));
+		if (
+			partnerChainAccount.status === CHAIN_REGISTERED &&
+			txParams.inboxUpdate.crossChainMessages.length !== 0 &&
+			txParams.inboxUpdate.messageWitness.siblingHashes.length !== 0 &&
+			txParams.inboxUpdate.outboxRootWitness.siblingHashes.length !== 0
+		) {
+			// If the first CCM in inboxUpdate is a registration CCM
+			if (
+				decodedCCMs.length > 0 &&
```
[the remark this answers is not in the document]
ishantiw: yeah I missed it too, updated

framework/src/modules/interoperability/sidechain/commands/cc_update.ts
ishantiw: yes, it should. Done

framework/src/modules/interoperability/utils.ts
ishantiw: ah i updated it but forgot to push from my local 😅

framework/src/modules/interoperability/utils.ts
ishantiw:
```
If params contains a non-empty certificate and an inboxUpdate, the validity of the latter is checked by:
```
where if inboxUpdate is non-empty then it should also have non-empty messageWitness as its a required property on schema. So I did a check for this whole block on top of this func `isInboxUpdateEmpty` where it checks for all the required properties of inboxUpdate.

so I assume, 
```
Then validate the newly updated root against the certificate state root using the provided outboxRootWitness. Using notation from [LIP 0039](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0039.md), this is done via the function verify(queryKeys, proof, certificate.stateRoot) with:
```
is also implying that inboxUpdate as well as messageWitness should exist.

framework/src/modules/interoperability/utils.ts
```diff
@@ -156,6 +157,13 @@ export const rawStateStoreKey = (storePrefix: number) => {
 	return Buffer.concat([DB_KEY_STATE_STORE, moduleIDBuffer, storePrefixBuffer]);
 };
 
+export const isInboxUpdateEmpty = (inboxUpdate: InboxUpdate) =>
```
shuse2, line 160: shouldn't this be `&&` condition?

framework/src/modules/interoperability/utils.ts
```diff
@@ -178,27 +186,31 @@ export const checkCertificateValidity = (
...
-				status: VerifyStatus.FAIL,
-				error: new Error('Certificate height should be greater than last certificate height.'),
-			};
-		}
+	const decodedCertificate = codec.decode<Certificate>(certificateSchema, encodedCertificate);
+	if (
+		decodedCertificate.blockID.equals(EMPTY_BYTES) ||
+		decodedCertificate.stateRoot.equals(EMPTY_BYTES) ||
+		decodedCertificate.validatorsHash.equals(EMPTY_BYTES) ||
+		decodedCertificate.timestamp === 0
```
shuse2, line 200: we might as well create emptyCertificate check

framework/src/modules/interoperability/utils.ts
shuse2: I think it's possible to have non-empty inbox update and empty message witness if I remember correctly.
I think similar check is done [here](https://github.com/LiskHQ/lisk-sdk/blob/9b722082fee3da23db5b514cc4da12555b6cf005/framework/src/modules/interoperability/utils.ts#L454-L457)

shuse2 (commented):
Looking good in general!
I think we should ensure empty check for messageWitness, certificate and inboxUpdate which is bit confusing.

framework/test/unit/modules/interoperability/mainchain/commands/cc_update.spec.ts
```diff
@@ -0,0 +1,277 @@
...
+	};
+
+	let partnerChainAccount: ChainAccount;
+	let partnerChannelAccount: ChannelData;
+	let context: CommandVerifyContext<CrossChainUpdateTransactionParams>;
+	let mainchainCCUUpdateCommand: MainchainCCUpdateCommand;
+	let params: CrossChainUpdateTransactionParams;
+	let activeValidatorsUpdate: ActiveValidator[];
+	let sortedActiveValidatorsUpdate: ActiveValidator[];
+	// let paramCertificate: Certificate;
```
Incede, line 90: This comment may be removed

Incede (changes requested):
Looks good in general, only minor suggestions.

framework/src/modules/interoperability/utils.ts
ishantiw: I thought if any one of them is true then we can say that the inboxUpdate is empty

ishantiw: So I have changed the condition to only flag inboxUpdate empty if all the properties are empty

framework/test/unit/modules/interoperability/utils.spec.ts
```diff
@@ -12,17 +12,53 @@
...
+} from '../../../../src/modules/interoperability/utils';
+import { certificateSchema } from '../../../../src/node/consensus/certificate_generation/schema';
+
+jest.mock('@liskhq/lisk-cryptography', () => ({
+	...jest.requireActual('@liskhq/lisk-cryptography'),
+}));
+
+jest.mock('@liskhq/lisk-tree', () => ({
+	...jest.requireActual('@liskhq/lisk-tree'),
+}));
```
shuse2, line 52: why is it mocking like this?

framework/test/unit/modules/interoperability/utils.spec.ts
```diff
@@ -68,4 +104,598 @@ describe('Utils', () => {
...
+			expect(partnerChannelStoreMock.getWithSchema).toHaveBeenCalledTimes(1);
+			expect(partnerChannelStoreMock.setWithSchema).toHaveBeenCalledTimes(1);
+			expect(context.getStore).toHaveBeenCalledTimes(1);
+			expect(partnerChainAccount.lastCertificate.height).toEqual(5); // original partnerValidator value unchange
+			expect(partnerChannelData.partnerChainOutboxRoot).toEqual(
+				defaultCalculatedRootFromRightWitness,
+			);
+		});
+
+		it('should run successfully and return undefined when messageWitness is empty', async () => {
```
shuse2, line 663: maybe also check if `calculateRootFromRightWitness` is used when message witness is not empty and vise versa

framework/test/unit/modules/interoperability/utils.spec.ts
```diff
@@ -68,4 +104,598 @@ describe('Utils', () => {
...
+			).resolves.toBeUndefined();
+			expect(partnerValidatorStore.setWithSchema).toHaveBeenCalledTimes(1);
+			expect(partnerChainStore.setWithSchema).toHaveBeenCalledTimes(1);
+			expect(partnerChannelStoreMock.getWithSchema).toHaveBeenCalledTimes(1);
+			expect(partnerChannelStoreMock.setWithSchema).toHaveBeenCalledTimes(1);
+			expect(context.getStore).toHaveBeenCalledTimes(1);
+			expect(partnerChainAccount.lastCertificate.height).toEqual(certificate.height);
+			expect(partnerChannelData.partnerChainOutboxRoot).toEqual(partnerChannelData.inbox.root);
+		});
+	});
```
shuse2, line 700: missing test for `checkInboxUpdateValidity`?

shuse2 (changes requested):

framework/test/unit/modules/interoperability/utils.spec.ts
ishantiw: we don't need it, removed

framework/src/modules/interoperability/utils.ts
```diff
@@ -345,61 +351,62 @@ export const checkInboxUpdateValidity = (
...
-		SMT_KEY_LENGTH,
-	);
-	if (!isSMTRootValid) {
-		return {
-			status: VerifyStatus.FAIL,
-			error: new Error(
-				'Failed at verifying state root when messageWitness and certificate are non-empty.',
-			),
+	// non-empty certificate and an inboxUpdate
+	if (!isCertificateEmpty(decodedCertificate) && !isInboxUpdateEmpty(txParams.inboxUpdate)) {
```
shuse2, line 355: inboxUpdate is never empty because it's checked at the beginning

framework/src/modules/interoperability/utils.ts
```diff
@@ -345,61 +351,62 @@ export const checkInboxUpdateValidity = (
...
+				newInboxSize,
+				newInboxAppendPath,
+				messageWitness.siblingHashes,
+			);
+		}
+		const proof = {
+			siblingHashes: outboxRootWitness.siblingHashes,
+			queries: [
+				{
+					key: partnerChainIDBuffer,
```
shuse2, line 368:
```
outboxKey = MODULE_ID_INTEROPERABILITY
            || STORE_PREFIX_OUTBOX
            || SHA-256(uint32be(partnerChainID)).
```
query key should be this? also for the `queryKeys` below

framework/src/modules/interoperability/utils.ts
```diff
@@ -345,61 +351,62 @@ export const checkInboxUpdateValidity = (
...
 
-	if (!(newInboxRoot as Buffer).equals(partnerChannelData.partnerChainOutboxRoot)) {
-		return {
-			status: VerifyStatus.FAIL,
-			error: new Error(
-				'Failed at verifying state root when messageWitness is non-empty and certificate is empty.',
-			),
-		};
+	// empty certificate and a non-empty inboxUpdate
+	if (isCertificateEmpty(decodedCertificate) && !isInboxUpdateEmpty(txParams.inboxUpdate)) {
```
shuse2, line 393: same as above. inboxUpdate is never empty here

shuse2 (changes requested):

framework/src/modules/interoperability/utils.ts
Incede: Just adding an reference to confirm this, the data is always hashed before being appended to the tree : https://github.com/LiskHQ/lips/blob/main/proposals/lip-0045.md#execution-1

framework/src/modules/interoperability/utils.ts
ishantiw: for queryKeys, its already using `const outboxKey = rawStateStoreKey(STORE_PREFIX_OUTBOX_ROOT);`

shuse2 (approved):
well done 🚀

framework/src/modules/interoperability/utils.ts
```diff
@@ -175,3 +211,327 @@ export const swapReceivingAndSendingChainIDs = (ccm: CCMsg) => ({
...
+			txParams.activeValidatorsUpdate,
+		);
+
+		const validatorsHash = computeValidatorsHash(
+			newActiveValidators,
+			txParams.newCertificateThreshold || partnerValidators.certificateThreshold,
+		);
+
+		if (!certificate.validatorsHash.equals(validatorsHash)) {
+			throw new Error('Validators hash is incorrect given in the certificate.');
```
Incede, line 479:
```suggestion
			throw new Error('Validators hash given in the certificate is incorrect.');
```

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
```diff
@@ -12,18 +12,278 @@
...
+		checkCertificateTimestampAndSignature(
+			txParams,
+			partnerValidators,
+			partnerChainAccount,
+			decodedCertificate,
+			header,
+		);
+
+		// If params contains a non-empty activeValidatorsUpdate
+		checkValidatorsHashWithCertificate(txParams, decodedCertificate, partnerValidators);
```
Incede, line 178: Any particular reason for moving this to `execute` from `verify` ?

framework/src/modules/interoperability/sidechain/commands/cc_update.ts
```diff
@@ -12,18 +12,265 @@
...
+			throw err;
+		}
+		if (
+			partnerChainAccount.status === CHAIN_REGISTERED &&
+			!isInboxUpdateEmpty(txParams.inboxUpdate)
+		) {
+			// If the first CCM in inboxUpdate is a registration CCM
+			if (
+				decodedCCMs[0].deserialized.crossChainCommandID === CROSS_CHAIN_COMMAND_ID_REGISTRATION &&
+				decodedCCMs[0].deserialized.receivingChainID === txParams.sendingChainID
```
Incede, line 210: It should be `ccm.sendingChainID == patnerChainID` for sidechain
```suggestion
				decodedCCMs[0].deserialized.sendingChainID === txParams.sendingChainID
```

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
```diff
@@ -12,18 +12,278 @@
...
+				decodedCCMs[0].deserialized.receivingChainID === MAINCHAIN_ID
+			) {
+				partnerChainAccount.status = CHAIN_ACTIVE;
+			} else {
+				await interoperabilityStore.terminateChainInternal(
+					txParams.sendingChainID,
+					beforeSendContext,
+				);
+
+				return; // Exit CCU processing
```
Incede, line 220: For the mainchain the LIP does not say to exit CCU processing.

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
```diff
@@ -12,18 +12,278 @@
...
+					txParams.sendingChainID,
+					beforeSendContext,
+				);
+
+				continue;
+			}
+			try {
+				validateFormat(ccm.deserialized);
+			} catch (error) {
+				await interoperabilityStore.terminateChainInternal(
```
Incede, line 236: Just to be sure - calling `terminateChainInternal` for the same chain id wouldn't have any unintended effects, right ? maybe `if else` could avoid it.

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
```diff
@@ -12,18 +12,278 @@
...
+		}
+
+		for (const ccm of decodedCCMs) {
+			if (txParams.sendingChainID !== ccm.deserialized.sendingChainID) {
+				await interoperabilityStore.terminateChainInternal(
+					txParams.sendingChainID,
+					beforeSendContext,
+				);
+
+				continue;
```
Incede, line 231: Why `continue`, wouldn't it automatically ?
same for `try catch` block below.

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
```diff
@@ -12,18 +12,278 @@
...
+			}
+			try {
+				validateFormat(ccm.deserialized);
```
Incede, line 234: I think the schema validation `validator.validate(ccmSchema, ccm)` is unnecessary here since it's already done at L194 `deserialized: codec.decode<CCMsg>(ccmSchema, ccm)`.

Same for sidechain.

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
```diff
@@ -12,18 +12,278 @@
...
+				serialized: ccm,
+				deserialized: codec.decode<CCMsg>(ccmSchema, ccm),
+			}));
+		} catch (err) {
+			await interoperabilityStore.terminateChainInternal(
+				txParams.sendingChainID,
+				beforeSendContext,
+			);
+
+			throw err;
```
Incede, line 202: Just to be sure, do we really want to throw and exit here for schema errors since we don't throw but continue for the same at L241 ?

framework/src/modules/interoperability/utils.ts
```diff
@@ -175,3 +211,327 @@ export const swapReceivingAndSendingChainIDs = (ccm: CCMsg) => ({
...
+	return {
+		status: VerifyStatus.OK,
+	};
+};
+
+export const checkCertificateValidity = (
+	partnerChainAccount: ChainAccount,
+	encodedCertificate: Buffer,
+): VerificationResult => {
+	if (encodedCertificate.equals(EMPTY_BYTES)) {
```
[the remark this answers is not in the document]
ishantiw: So the first condition checks whether a certificate is non-empty, empty buffer is like undefined, in that case we don't check the validity of certificate. whereas at L269  if certificate has non-empty buffer and we decode it then it should have all the required values in certificate

framework/src/modules/interoperability/utils.ts
```diff
@@ -175,3 +211,327 @@ export const swapReceivingAndSendingChainIDs = (ccm: CCMsg) => ({
...
+		MESSAGE_TAG_CERTIFICATE,
+		partnerChainAccount.networkID,
+		txParams.certificate,
+		activeValidators.map(v => v.bftWeight),
+		certificateThreshold,
+	);
+
+	if (!verifySignature || certificate.timestamp >= header.timestamp)
+		throw Error(
+			'Certificate is invalid due to invalid last certified height or timestamp or signature.',
```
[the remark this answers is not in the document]
ishantiw: earlier there was a height check that was removed but forgot to change the message. Changed it now

framework/src/modules/interoperability/utils.ts
```diff
@@ -175,3 +211,327 @@ export const swapReceivingAndSendingChainIDs = (ccm: CCMsg) => ({
...
+		status: VerifyStatus.OK,
+	};
+};
+
+export const checkInboxUpdateValidity = (
+	txParams: CrossChainUpdateTransactionParams,
+	partnerChannelData: ChannelData,
+): VerificationResult => {
+	// If inboxUpdate is empty then return success
+	if (isInboxUpdateEmpty(txParams.inboxUpdate)) {
```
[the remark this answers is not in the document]
ishantiw: That condition is checked later, for one condition is for empty certificate. so the required one is non-empty inboxUpdate

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
ishantiw: I think this needs to be clarified as @shuse2 pointed out earlier to exit.

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
ishantiw: you mean to check whether it is already terminated? I think we skip the loop when we call this function

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
ishantiw: we `continue` because we want to skip the processing of CCM any further and terminate the chain. If I don't continue then it will call `appendToInboxTree` or `apply` later on

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
ishantiw: its also checking the ccm size but it is kind of redundant that it checks for the schema again within this function. I think it is also the same in the LIP

framework/src/modules/interoperability/utils.ts
```diff
@@ -175,3 +211,327 @@ export const swapReceivingAndSendingChainIDs = (ccm: CCMsg) => ({
...
+	certificate: Certificate,
+	header: BlockHeader,
+) => {
+	// Certificate and Validators Update Validity
+	if (txParams.certificate.equals(EMPTY_BYTES)) {
+		return;
+	}
+
+	const { activeValidators, certificateThreshold } = partnerValidators;
+	const verifySignature = verifyWeightedAggSig(
```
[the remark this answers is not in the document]
ishantiw: actually this and the other comment, yes we should move it to verify(), earlier it was moved to some reason (getting validator from bft that required header height), but now its simplified so it should be moved back. Thanks for pointing this out

framework/src/modules/interoperability/utils.ts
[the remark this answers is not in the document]
Incede: I mean both are cases for empty certificate. For one the verify status `OK` doesn't make sense when the other is `FAIL`. I think it should the return the same status(with a diff error msg) in both the cases

framework/src/modules/interoperability/utils.ts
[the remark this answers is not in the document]
Incede: Side question maybe - Do you know if there is a preference b/w `verify` & `execute` hooks for computationaly heavy checks like `verifyWeightedAggSig` ? I mean do we want heavy computations to happen in verify hook or execute hook or it doesn't matter ?

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
Incede: hmm, I am not sure. In the LIP it doesn't mention to exit explicitly here but does in some other places so a bit hard to infer.

framework/src/modules/interoperability/utils.ts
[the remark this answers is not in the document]
ishantiw: According to https://github.com/LiskHQ/lips/blob/main/proposals/lip-0053.md#certificate-and-validators-update-validity,
```
If params contains a non-empty certificate, it is valid if:
```

from my understanding, it says if a certificate is present then we need to check its validity and missing values, else we skip this check and hence it will be status OK

framework/src/modules/interoperability/utils.ts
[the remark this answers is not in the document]
ishantiw: I think we want to catch as many cases as we can to avoid a transaction failing at execute where they have to pay the fee

framework/src/modules/interoperability/utils.ts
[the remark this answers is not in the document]
Incede: mm yea the LIP is not super clear about it but I think we need to clarify it since missing all values and empty buffer is the same case (i.e. empty certificate) in my understanding.

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
ishantiw: moved to verify

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
ishantiw: Clarifying with research

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
ishantiw: Clarifying with research

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
ishantiw: Related to above comments, Clarifying with research

framework/src/modules/interoperability/mainchain/commands/cc_update.ts
ishantiw: Related to above comments, Clarifying with research

Incede (approved):

framework/src/modules/interoperability/utils.ts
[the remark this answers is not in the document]
ishantiw: Checking with research team, will be tackled in https://github.com/LiskHQ/lisk-sdk/issues/7185

PR #7169, opened 2022-05-13
Remove validator dependency from BFT, consensus and generator - Closes #7168

### LINKED ISSUE #7168: Update Consensus not to dependent on validator module
### Description

- Move BFT module into consensus
- Update BFT module BFTParameters to store generator/BLS keys as below
```
{
	prevoteThreshold: bigint;
	precommitThreshold: bigint;
	certificateThreshold: bigint;
	validators: {
		address: Buffer,
		bftWeight: bigint,
		blsKey: Buffer,
		generatorKey: Buffer,
	}[];
	validatorsHash: Buffer;
}
```
- BFTParameters now can have bft weight 0, and order by shuffled address
- Add API ` getValidator(address: Buffer, height: number): Validator` which returns Validator information for specified height. (it is used to get a generator key to check signature)
- Update generator and consensus not to be dependent on validators module, but use bft module or consensus instead
- Fix impliesMaximalPrevotes

### Acceptance Criteria

- Fix all unit tests
- Check dpos example and ensure block execution / generation is working

shuse2:
### What was the problem?

This PR resolves #7168 

### How was it solved?

- Update BFT API
  - [Update BFT params](https://github.com/LiskHQ/lisk-sdk/pull/7169/files#diff-9f4b19ed8c1c4f6efbf16e8d32e01bcec9141ffe6df09616db2f94a22739e2e6R565-R627) to contain generatorKey and blsKey. Also, allow 0 BFT weight into BFT params and not sorting validators by address
  - Add [getValidator](https://github.com/LiskHQ/lisk-sdk/pull/7169/files#diff-80b27d4b3f3df66043297ad75ae7ec2134df5db03cd57263ce2ad9b36539866bR245) for getting generator/blskey for specific height
- Remove dependency of validator module from bft, consensus and generator
  - Use BFT Params for getting generator key to verify block signature
  - Use BFT Params for getting bls key to verify commit signature
- Move BFT into node (https://github.com/LiskHQ/lisk-sdk/pull/7169/commits/de43bb79968dbe588ce1a6921adc5e673946c6ae)
- Remove unused function from validators
  - getGeneratorAtTimestamp (Moved to consensus)
  - getSlotNumber  (Moved to consensus)
  - getSlotTime (Moved to consensus)
  - getGeneratorList (Use BFT Params)
  - setGeneratorList (Merged with BFT Params)

### How was it tested?

- Update unit tests
- Run DPoS mainchain example app


framework/src/node/bft/api.ts
```diff
@@ -106,15 +101,19 @@ export class BFTAPI extends BaseAPI {
 		};
 	}
 
+	// This function expects to be called after blockBFTInfos has been inserted for the executiong block.
```
shuse2, line 104: @janhack Let me know if this assumption is correct 🙏

framework/src/node/bft/api.ts
```diff
@@ -168,9 +167,6 @@ export class BFTAPI extends BaseAPI {
 		}
 		let aggregateBFTWeight = BigInt(0);
 		for (const validator of validators) {
-			if (validator.bftWeight <= 0) {
```
shuse2, line 171: here, when setting the bft params, it's accepting the 0 BFT weight

framework/src/node/bft/api.ts
```diff
@@ -244,16 +237,26 @@ export class BFTAPI extends BaseAPI {
 		return params.validators;
 	}
 
-	private async _computeValidatorsHash(
+	public async getValidator(
```
shuse2, line 240: Adding this function to get a validator info for particular height

framework/src/node/bft/api.ts
```diff
@@ -214,6 +204,9 @@ export class BFTAPI extends BaseAPI {
 
 		const nextActiveValidators: BFTVotesActiveValidatorsVoteInfo[] = [];
 		for (const validator of validators) {
+			if (validator.bftWeight === BigInt(0)) {
```
shuse2, line 207: ignoring 0 bft weight for active validator

framework/src/node/bft/schemas.ts
```diff
@@ -75,6 +72,14 @@ export const bftParametersSchema = {
...
 						fieldNumber: 2,
 					},
+					generatorKey: {
+						dataType: 'bytes',
+						fieldNumber: 3,
+					},
+					blsKey: {
+						dataType: 'bytes',
+						fieldNumber: 4,
+					},
```
shuse2, line 82: updating definition of BFT validator

framework/src/modules/dpos_v2/module.ts
```diff
@@ -564,27 +576,9 @@ export class DPoSModule extends BaseModule {
 
 		const apiContext = context.getAPIContext();
 
-		// get the last stored BFT parameters, and update them if needed
-		const currentBFTParams = await this._bftAPI.getBFTParameters(apiContext, height);
-		// snapshot.activeDelegates order should not be changed here to use it below
-		const bftWeight = [...snapshot.activeDelegates]
-			.sort((a, b) => a.compare(b))
```
shuse2, line 571: removed sorting to maintain shuffled list

framework/src/modules/dpos_v2/module.ts
```diff
@@ -596,18 +590,40 @@ export class DPoSModule extends BaseModule {
...
+			);
+			const bftWeight =
+				standbyDelegates.findIndex(addr => addr.equals(validatorAddress)) > -1
+					? BigInt(0)
+					: BigInt(1);
+			bftValidators.push({
+				address: validatorAddress,
+				generatorKey: validatorAccount.generatorKey,
+				blsKey: validatorAccount.blsKey,
+				bftWeight,
```
shuse2, line 618: here if validator is not active, setting the bftWeight to be zero. Also, maintaining shuffled order

framework/src/node/bft/api.ts
janhack: So compared to https://github.com/LiskHQ/lips/blob/main/proposals/lip-0058.md#impliesmaximalprevotes, you changed the function here assuming the block provided as input has already been added to `blockBFTInfos`?

This is fine, but then the input `header` in the function would not be needed and instead the header information from `blockBFTInfos` could be used.

shuse2: yes! Actually, I think this was not related to this PR. When we changed the BFT execution to be in `beforeTransactionsExecute` hook, this needed to be changed as well

framework/src/node/bft/api.ts
janhack: In the BFT LIP, validators with BFT weight 0 are not accepted in `setBFTParameters` and the API fails. If I understand it correctly here, now it is possible to set BFT weight to 0 and the validator is just skipped over. So the state information in the end would be the same. Is that correct?

shuse2: State information for the `bftVotes` will remain the same. However, BFT params will maintain 0 vote weight validators

framework/src/node/bft/api.ts
```diff
@@ -167,9 +166,10 @@ export class BFTAPI extends BaseAPI {
 			);
 		}
 		let aggregateBFTWeight = BigInt(0);
+		const activeValidators: Validator[] = [];
 		for (const validator of validators) {
-			if (validator.bftWeight <= 0) {
-				throw new Error('Invalid BFT weight. BFT weight must be a positive integer.');
+			if (validator.bftWeight > BigInt(0)) {
+				activeValidators.push(validator);
```
shuse2, line 172: ensure only validator with weight is included to compute validatorsHash

framework/src/node/consensus/consensus.ts
```diff
@@ -681,9 +686,10 @@ export class Consensus {
 	}
 
 	private async _verifyAssetsSignature(apiContext: APIContext, block: Block): Promise<void> {
-		const { generatorKey } = await this._validatorAPI.getValidatorAccount(
+		const { generatorKey } = await this._bftAPI.getValidator(
```
shuse2, line 689: Example of how BFTAPI getValidator / BFTParams is used for signature check

framework/src/modules/dpos_v2/module.ts
```diff
@@ -395,13 +394,26 @@ export class DPoSModule extends BaseModule {
...
 		const initDelegates = [...genesisStore.genesisData.initDelegates];
 		initDelegates.sort((a, b) => a.compare(b));
-		const bftWeights = initDelegates.map(d => ({
-			bftWeight: BigInt(1),
-			address: d,
-		}));
 		const initBFTThreshold = BigInt(Math.floor((2 * initDelegates.length) / 3) + 1);
-		await this._bftAPI.setBFTParameters(apiContext, initBFTThreshold, initBFTThreshold, bftWeights);
-		await this._validatorsAPI.setGeneratorList(apiContext, initDelegates);
+		const bftValidators = [];
```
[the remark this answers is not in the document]
janhack: The code regarding setting BFT parameters and generators here looks good. There is one difference though: before we only were setting new BFT parameters if actually the set of active validators changed. This will have the effect that a certificate has to be generated for every round and we do not skip over rounds without a change in active validators. This is a minor disadvantage though and I find it acceptable.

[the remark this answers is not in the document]
janhack: Here are the relevant lines in the certificate validation:
```
    heightNextBFTParameters = bftModule.getNextHeightBFTParameters(bftModule.getBFTHeights().maxHeightCertified+1)
    if previous function call returns valid height and aggregateCommit.height > heightNextBFTParameters-1:
        return False
```
see also https://github.com/LiskHQ/lips/blob/main/proposals/lip-0061.md#block-verification

framework/src/node/bft/api.ts
```diff
@@ -106,31 +101,27 @@ export class BFTAPI extends BaseAPI {
 		};
 	}
 
-	public async impliesMaximalPrevotes(
-		context: ImmutableAPIContext,
-		header: BlockHeader,
-	): Promise<boolean> {
+	// This function expects to be called after blockBFTInfos has been inserted for the executiong block.
+	// blockBFTInfos must be inserted first because for the block genesis block + 1, bftVotes.blockBFTInfos will be empty
+	public async currentHeaderImpliesMaximalPrevotes(context: ImmutableAPIContext): Promise<boolean> {
```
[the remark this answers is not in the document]
janhack: The function `currentHeaderImpliesMaximalPrevotes` looks all good.

framework/src/node/bft/endpoint.ts
```diff
@@ -35,4 +37,16 @@ export class BFTEndpoint extends BaseEndpoint {
 		}
 		return areDistinctHeadersContradicting(bftHeader1, bftHeader2);
 	}
+
+	public async getGeneratorList(ctx: ModuleEndpointContext): Promise<{ list: string[] }> {
+		const votesStore = ctx.getStore(this.moduleID, STORE_PREFIX_BFT_VOTES);
+		const bftVotes = await votesStore.getWithSchema<BFTVotes>(EMPTY_KEY, bftVotesSchema);
+		const { height: currentHeight } =
+			bftVotes.blockBFTInfos.length > 0 ? bftVotes.blockBFTInfos[0] : { height: 1 };
```
janhack, line 45: Should it maybe be `bftVotes.blockBFTInfos[0].height` instead of just `bftVotes.blockBFTInfos[0].height`?

janhack (commented):
Looks all good in general, just one line I was unsure about.

framework/src/node/bft/endpoint.ts
shuse2: `bftVotes.blockBFTInfos[0]` will return
```
{
	height: number;
	generatorAddress: Buffer;
	maxHeightGenerated: number;
	maxHeightPrevoted: number;
	prevoteWeight: bigint;
	precommitWeight: bigint;
}
```
and it is destructured on the left hand, so this is correct

janhack: Ah, I missed the destructing. Also not so familiar with that syntax, all good then!

framework/src/modules/dpos_v2/module.ts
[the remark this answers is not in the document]
shuse2: @janhack 
Updated: https://github.com/LiskHQ/lisk-sdk/pull/7169/commits/ff3be7ccdb9a38172be5438c2d6fc4a7073c9d46

- Updated to keep the BFTParams to only holds BLS key, and only updated when there is change
  - Also, the validators here is sorted lexicographically as before
- Created new generatorKeys store which holds address and generator key

framework/src/modules/dpos_v2/module.ts
```diff
@@ -602,20 +607,23 @@ export class DPoSModule extends BaseModule {
...
 				apiContext,
 				validatorAddress,
 			);
-			const bftWeight =
-				standbyDelegates.findIndex(addr => addr.equals(validatorAddress)) > -1
-					? BigInt(0)
-					: BigInt(1);
-			bftValidators.push({
+			// if validator is active
+			if (standbyDelegates.findIndex(addr => addr.equals(validatorAddress)) === -1) {
```
janhack, line 617: Would it maybe be better here to check if the delegate is active (instead of not standby). On Mainnet right now there are less standby delegates than active, but in general the set of standby delegates could be `>1000` whereas there are always only 101 active delegates.

framework/src/node/bft/api.ts
```diff
@@ -19,15 +19,18 @@ import { BaseAPI } from '../../modules/base_api';
 import { APIContext, BlockHeader, ImmutableAPIContext } from '../state_machine';
 import {
 	areDistinctHeadersContradicting,
+	getGeneratorKeys,
 	sortValidatorsByAddress,
 	sortValidatorsByBLSKey,
+	validtorsEqual,
```
janhack, line 25: Typo

framework/src/modules/dpos_v2/module.ts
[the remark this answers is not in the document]
janhack: Looks all good!

janhack (commented):
Just a minor typo and one suggestion, otherwise all good.

janhack (approved):

framework/src/node/bft/utils.ts
```diff
@@ -71,3 +75,55 @@ export const sortValidatorsByAddress = (validators: { address: Buffer }[]) =>
...
+
+	return true;
+};
+
+export const getGeneratorKeys = async (
+	keysStore: ImmutableSubStore,
+	height: number,
+): Promise<GeneratorKeys> => {
+	const start = intToBuffer(0, 4, BIG_ENDIAN);
+	const end = intToBuffer(height, 4, BIG_ENDIAN);
```
ishantiw, line 100: I think `BIG_ENDIAN` is default `intToBuffer(0, 4)`

framework/src/node/bft/utils.ts
```diff
@@ -71,3 +75,55 @@ export const sortValidatorsByAddress = (validators: { address: Buffer }[]) =>
...
+		throw new GeneratorKeysNotFoundError();
+	}
+	const [result] = results;
+
+	return codec.decode<GeneratorKeys>(generatorKeysSchema, result.value);
+};
+
+export const deleteGeneratorKeys = async (keysStore: SubStore, height: number): Promise<void> => {
+	const start = intToBuffer(0, 4, BIG_ENDIAN);
+	const end = intToBuffer(height, 4, BIG_ENDIAN);
```
ishantiw, line 117: same here

framework/src/node/bft/utils.ts
```diff
@@ -71,3 +75,55 @@ export const sortValidatorsByAddress = (validators: { address: Buffer }[]) =>
...
+	const results = await keysStore.iterate({
+		start,
+		end,
+	});
+	if (results.length <= 1) {
```
ishantiw, line 122: if there is only one result then we don't delete?

ishantiw (approved):
LGTM 🎉  Some minor comments

framework/src/node/bft/utils.ts
shuse2: yes, we need to maintain at least one generator keys

PR #7183, opened 2022-05-24
Update folder name to engine - Partially closes #7131

### LINKED ISSUE #7131: Create engine to run as a child process
### Description
- Create engine class which expects the application binary interface as defined in #7129 
- Engine should initialize consensus/generator/network and RPC endpoints
- Add engine starter to run the engine class as a child process

### Acceptance criteria
- Add unit tests for engine class
- All build and tests should be passing

shuse2:
### What was the problem?

This PR resolves #7131

### How was it solved?

Update folder name and class of `node` to `engine`

### How was it tested?

All test and build should pass


framework/src/engine/engine.ts
```diff
@@ -55,7 +55,7 @@ const isEmpty = (value: unknown): boolean => {
 
 const emptyOrDefault = <T>(value: T, defaultValue: T): T => (isEmpty(value) ? defaultValue : value);
 
-export class Node {
+export class Engine {
```
ishantiw, line 58: there are several places in Engine where we are referring to node in the logger like L87, L97, L109.

ishantiw (commented):
Apart from one comment, another thing is whether we are going to rename node.db->engine.db as well, if yes then whether in this PR or another one. Apart from these 2 points, LGTM

framework/src/engine/engine.ts
shuse2: Updating the logging message. 
For the database name, I think `node.db` still make sense because it's storing information of the node.
I think with this setup, (application + engine) creates a client which runs a node

ishantiw (approved):

Incede (approved):

PR #7199, opened 2022-06-03
 Update timestamp parameter from sendInternal to be optional - Closes #7154 

### LINKED ISSUE #7154: Update `timestamp` parameter from `sendInternal` to be optional
### Expected behavior

`sendInternal` function in interoperability store does not require `timestamp` in sidechains but is used in the mainchain. 

We make the `timestamp` parameter optional so we do not have to pass in `Date.now` arbitrarily in sidechain context.

### Actual behavior

It currently requires `timestamp`. 

### Steps to reproduce

Check function parameters

6.0

Incede:
### What was the problem?

This PR resolves #7154 

### How was it solved?

Update param & remove unnecessary usage

### How was it tested?

Updated test

### DIFF
--- a/framework/src/modules/interoperability/mainchain/store.ts
+++ b/framework/src/modules/interoperability/mainchain/store.ts
@@ -158,11 +158,14 @@ export class MainchainInteroperabilityStore extends BaseInteroperabilityStore {
 			return false;
 		}
 
+		if (!sendContext.timestamp) {
+			throw new Error('Timestamp is missing from the function parameters');
+		}
+
 		// Chain must be live; This check is always on the receivingChainID
 		const isReceivingChainLive = await this.isLive(
 			receivingChainIDAsStoreKey,
-			// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
-			sendContext.timestamp!,
+			sendContext.timestamp,
 		);
 		if (!isReceivingChainLive) {
 			return false;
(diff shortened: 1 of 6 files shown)


mitsuaki-u (approved):
LGTM!

framework/src/modules/interoperability/mainchain/store.ts
```diff
@@ -161,7 +161,8 @@ export class MainchainInteroperabilityStore extends BaseInteroperabilityStore {
 		// Chain must be live; This check is always on the receivingChainID
 		const isReceivingChainLive = await this.isLive(
 			receivingChainIDAsStoreKey,
-			sendContext.timestamp,
+			// eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+			sendContext.timestamp!,
 		);
```
ishantiw, line 166: maybe its better to check if timestamp is undefined and throw error in isLive function in mainchain as its needed to check for the liveness condition

ishantiw (changes requested):
LGTM, one comment related to timestamp check in isLive function

framework/src/modules/interoperability/mainchain/store.ts
ishantiw: isLive is being called from other places too, so maybe its better to throw it from one place

Incede: other places don't have optional param,  no ?

ishantiw: ah true, sorry ignore my comment

ishantiw (approved):
Thanks for the quick fix 👏

Issue #7195, opened 2022-06-02
Implement `childKeyDerivation` for Ed25519 key derivation


mitsuaki-u:
### Description

This function derives a child node from a parent node and the given child node index.

```
childKeyDerivation(node: extendedKey, index: uint32) -> extendedKey:
    indexBuffer = big-endian encoding of index using 4 bytes 
    hmac = PRF-HMAC-SHA-512(
        key=node.chainCode,
        data = 0x00 || node.key || indexBuffer
    )
    return {
        key: hmac[:32],
        chainCode: hmac[32:]
    }
```

Issue #7194, opened 2022-06-02
Implement `getMasterKeyFromSeed` for Ed25519 key derivation


mitsuaki-u:
### Description

This function derives an Ed25519 private key, as defined in RFC 8032, from seed

```
getMasterKeyFromSeed(seed: bytes) -> extendedKey:
    hmac = PRF-HMAC-SHA-512(
        key=0x656432353531392073656564 #utf8 encoding of 'ed25519 seed'
        data=seed
    )
    return {
        key: hmac[:32],
        chainCode: hmac[32:]
    }
```
To. create HMAC reference: https://nodejs.org/api/crypto.html#cryptocreatehmacalgorithm-key-options

Issue #7193, opened 2022-06-02
Implement `getKeyPairFromPassphraseAndPath` for Ed25519 key derivation


mitsuaki-u:
### Description

This function derives an Ed25519 private key, as defined in RFC 8032, from an input phrase and a derivation path.

```
getKeyPairFromPhraseAndPath(phrase: string, path: string) -> { publicKey: bytes, privateKey: bytes }:
    // calls private functions 
    pathArry = getPath(path)
    seed =  // logic to get private key from phrase and path 
    
    return sodiumNative/tweetnacl.fromSeed(seed)
```

### Acceptance Criteria

- Add unit tests

PR #7212, opened 2022-06-10
Implement ED25519 key derivation - Closes #7192, #7193, #7194, #7195

### LINKED ISSUE #7192: Implement `parseKeyDerivationPath` utility function for key derivation
### Description

This function validates and parses a path string to conform to the standard e.g. `m/44'/134'/0'` 

Validation rules:

e.g. `m/44'/134'/0'`
 - First char must be m
 - segment should be number or number with apostrophe
 - there should at least be one backslash
 - no special chars other than apostrophe 
 - check if number smaller than (2^31) / 2 

### Acceptance Criteria

- Test for various path inputs when using this function

mitsuaki-u:
### What was the problem?

This PR resolves #7192 #7193 #7194 #7195 

### How was it solved?

- Added path validation
- Added get childkey 
- Added get masterkey
- Added get keypair from path and phrase

### How was it tested?

- Added unit tests


elements/lisk-cryptography/src/encrypt.ts
```diff
@@ -166,3 +169,23 @@ const decryptAES256GCMWithPassword = (
...
+		throw new Error('Invalid path format');
+	}
+
+	const masterSeed = await Mnemonic.mnemonicToSeed(phrase);
+	let node = getMasterKeyFromSeed(masterSeed);
+
+	// slice first element which is `m`
+	for (const segment of path.split('/').slice(1)) {
+		// if segment includes apostrophe, we must add HARDENED_OFFSET
+		const segmentWithOffset = segment.includes("'")
```
shuse2, line 184: it seems like format logic is in both `isValidPath` and here.
Maybe better to have 
`getPathArray` or something and do validation and conversion to number?

elements/lisk-cryptography/src/utils.ts
```diff
@@ -33,3 +36,47 @@ export const writeBit = (buf: Buffer, bit: number, val: boolean): void => {
...
+
+			return segment.includes(`'`)
+				? parseInt(segment.slice(0, -1), 10) <= MAX_UINT32 / 2
+				: parseInt(segment, 10) <= MAX_UINT32 / 2;
+		});
+};
+
+export const getMasterKeyFromSeed = (seed: Buffer) => {
+	const hmac = crypto.createHmac('sha512', ED25519_CURVE);
+	const I = hmac.update(seed).digest();
```
shuse2, line 61: Prefer camel case. same for below

shuse2 (changes requested):

elements/lisk-cryptography/src/encrypt.ts
mitsuaki-u: I prefer to keep it semantically different, because validate function validates and a get function gets. But if you feel strongly I can update it :)

shuse2: how about `parseKeyDerivationPath` function and throw `parse error`?
Also, this logic will be used in the other BLS key derivation

shuse2 (approved):
Great work 🚀

PR #7228, opened 2022-06-20
Update README.md

sitetester:
### What was the problem?

This PR resolves #7228 

### How was it solved?

Currently our codebase is not compatible with latest yarn version (`yarn build` fails)

### How was it tested?

Downgrading latest version to 1.xx & simply running `yarn build` resulted in successful build.

### DIFF
--- a/sdk/README.md
+++ b/sdk/README.md
@@ -179,7 +179,7 @@ To contribute to `lisk-sdk`, `framework` or `elements`:
 
 1. Clone the repository: `git clone https://github.com/LiskHQ/lisk-sdk.git`
 
-2. Install yarn globally: [Install Yarn](https://classic.yarnpkg.com/en/docs/install)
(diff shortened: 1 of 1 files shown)


sdk/README.md
```diff
@@ -179,7 +179,7 @@ To contribute to `lisk-sdk`, `framework` or `elements`:
 
 1. Clone the repository: `git clone https://github.com/LiskHQ/lisk-sdk.git`
 
-2. Install yarn globally: [Install Yarn](https://classic.yarnpkg.com/en/docs/install)
+2. Install [Yarn Classic](https://classic.yarnpkg.com/en/docs/install) globally: (please note that we are still on Yarn 1.xx)
```
ishantiw, line 182: since Yarn classic is pointing to v1.xxx we don't need to add a note here 
```suggestion
2. Install [Yarn Classic](https://classic.yarnpkg.com/en/docs/install) globally
```

ishantiw (approved):

PR #7234, opened 2022-06-21
Fix getMetadata response and socket ipc data path- Closes #7172

### LINKED ISSUE #7172: Unable to post transactions against v6 node
### Expected behavior
It should be possible to successfully post the transaction to the node.

### Actual behavior
Throws timeout error:
```
Error: Response not received in 3000ms
```
### Steps to reproduce
Run node locally with IPC enabled and post any type of transaction

development

ishantiw:
### What was the problem?

This PR resolves #7172 

### How was it solved?

- [🐛 Fix getMetaData response](https://github.com/LiskHQ/lisk-sdk/commit/b1e5f4858033626338b67d09f195f2c82122acda)
- [🐛 Fix socketPath for IPC api](https://github.com/LiskHQ/lisk-sdk/commit/89c1c27160543a7ba744127de8dc0b47b744c658)

### How was it tested?

- Run `npm t` under `elements/lisk-api-client`
- Create an API client using createIPCClient and run a node with with `--api-ipc` enabled. check if it gets connected successfully and is able to call `system_getMetadata` endpoint.

### DIFF
--- a/elements/lisk-api-client/src/api_client.ts
+++ b/elements/lisk-api-client/src/api_client.ts
@@ -31,7 +31,10 @@ export class APIClient {
 	}
 
 	public async init(): Promise<void> {
-		this._metadata = await this._channel.invoke<ModuleMetadata[]>('system_getMetadata');
+		const { modules } = await this._channel.invoke<{ modules: ModuleMetadata[] }>(
+			'system_getMetadata',
+		);
+		this._metadata = modules;
 		this._schema = await this._channel.invoke<RegisteredSchemas>('system_getSchema');
 		this._node = new Node(this._channel);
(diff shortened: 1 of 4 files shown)


has5aan (approved):
- Passing tests.

- `APIClient` retrieves metadata (configured modules) at initialization upon manual testing with both `lisk-core-v4.0.0-alpha.0` and `examples\dpos-mainchain` which fixes IPC connection to `dpos-mainchain`.

sitetester (approved):

sitetester:
Somehow it would be better to run API client against a real node to get proper stats, currently tests are passing using mocking.

PR #7293, opened 2022-07-12
Fix endpoint store to use prefixed read writer - Closes #7239

### LINKED ISSUE #7239: dpos_getAllDelegates returns empty response
### Expected behavior
`dpos_getAllDelegates` should return the list of all the registered delegates.

### Actual behavior
It returns an empty object instead.

### Steps to reproduce
- Clone both `lisk-sdk` (development) and `lisk-core` (development) locally
- Within `lisk-sdk` run the following commands
  - `yarn install`
  - `npm run build`
  - `cd framework && yarn link`
  - `cd sdk && yarn link`
- And, then from within `lisk-core`
  - `npm install --registry https://npm.lisk.com`
  - `yarn link lisk-framework`
  - `yarn link lisk-sdk`
  - `npm run build`
  - Start Lisk Core: `rm -rf ~/.lisk && ./bin/run start --network devnet --api-ipc --console-log info`
- With the lisk-api-client, query the endpoint `dpos_getAllDelegates`

Lisk SDK `development` branch (commit hash `bb65777fe0bba6f230b1364be1ded94877d7a17b`)

shuse2:
### What was the problem?

This PR resolves #7239 

### How was it solved?

When invoking endpoint, `getStore` was using `StateStore`, but for the stateDB, it needed to use PrefixedReadWriter. 

### How was it tested?

- Run DPoS example with http RPC server enabled
- Call `dpos_getAllDelegates` and see if it returns all delegates
```
{
    "jsonrpc": "2.0",
    "id": "1",
    "method": "dpos_getAllDelegates",
    "params": {
    }
}
```

### DIFF
--- a/framework/src/controller/channels/in_memory_channel.ts
+++ b/framework/src/controller/channels/in_memory_channel.ts
@@ -13,7 +13,6 @@
  */
 
 import { ListenerFn } from 'eventemitter2';
-import { StateStore } from '@liskhq/lisk-chain';
 import { StateDB } from '@liskhq/lisk-db';
 import { Event, EventCallback } from '../event';
 import { Request } from '../request';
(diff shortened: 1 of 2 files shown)


ishantiw (approved):
Tested with IPC client and was able to retrieve all the delegates successfully 🎉

PR #7294, opened 2022-07-12
Update rpc config - Closes #7277

### LINKED ISSUE #7277: Update RPC config
### Description
Update RPC server to accept 
```
{
  modes: ('ipc' | 'ws' | 'http')[],
  port: number,
  host: string,
}
```

* IPC server socket should be placed in `${dataPath}/tmp/sockets/` with the same name as it is now. 
* Update ABI `init` definition to have above RPC config format.
* Update application config RPC to have the above config format. 
* Update RPC server to create http server when `http` or `ws` mode is enabled. If `ws` and `http` are both enabled, `http` server should be reused for `ws` server. 
* Path of the server should be `host:port/rpc` for http and `host:port/rpc-ws` for ws
* Update APIClient to expect the path in the `dataPath`
* Update commander `start` command as mentioned [here](https://github.com/LiskHQ/lisk-sdk/issues/7141#issuecomment-1173999816)

### Motivation
* HTTP and WS can be run on the same port with different path. Therefore, it is better to have single host and port to be specified.
* IPC file path can be fixed path in the `dataPath` so that APIClient doesn't need to know exact path other than the `dataPath` of the running node

### Acceptance Criteria

* Update all the test scenarios
* Check if IPC, HTTP and WS can be used from the client all at the same time if enabled using example app

### Additional Information
* [IPC Server](https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/engine/rpc/rpc_server.ts#L59-L63) defines the socket path
* [Application config](https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/schema/application_config_schema.ts#L258)
* [ABI interface](https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/abi/abi.ts#L129-L134) and abi handler and schema will be affected
* [RPC Server](https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/engine/rpc/rpc_server.ts#L66-L79), and HTTPServer and WSServer also needs to be updated to expose/use raw http server
* [IPC client](https://github.com/LiskHQ/lisk-sdk/blob/development/elements/lisk-api-client/src/ipc_channel.ts#L26-L33)

shuse2:
### What was the problem?

This PR resolves #7277

### How was it solved?

- Update RPC config
- Update commander start command to support updated format

### How was it tested?

```
./bin/run start --api-ws --api-http --api-ws
./bin/run block:get 3 --pretty

call http://127.0.0.1:7887/rpc with 

{
    "jsonrpc": "2.0",
    "id": "1",
    "method": "dpos_getAllDelegates",
    "params": {
        "height": 544
    }
}

call ws://127.0.0.1:7887/rpc-ws with the above request
```

### DIFF
--- a/commander/src/bootstrapping/commands/start.ts
+++ b/commander/src/bootstrapping/commands/start.ts
@@ -66,20 +66,28 @@ export abstract class StartCommand extends Command {
 				'Enable IPC communication. This will load plugins as a child process and communicate over IPC. Environment variable "LISK_API_IPC" can also be used.',
 			env: 'LISK_API_IPC',
 			default: false,
-			exclusive: ['api-ws'],
 		}),
 		'api-ws': flagParser.boolean({
 			description:
 				'Enable websocket communication for api-client. Environment variable "LISK_API_WS" can also be used.',
 			env: 'LISK_API_WS',
 			default: false,
-			exclusive: ['api-ipc'],
 		}),
-		'api-ws-port': flagParser.integer({
+		'api-http': flagParser.boolean({
 			description:
-				'Port to be used for api-client websocket. Environment variable "LISK_API_WS_PORT" can also be used.',
+				'Enable HTTP communication for api-client. Environment variable "LISK_API_HTTP" can also be used.',
+			env: 'LISK_API_HTTP',
+			default: false,
+		}),
+		'api-port': flagParser.integer({
+			description:
+				'Port to be used for api-client. Environment variable "LISK_API_PORT" can also be used.',
 			env: 'LISK_API_WS_PORT',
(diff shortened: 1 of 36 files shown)


has5aan (approved):
### Verification steps

- Executed test node, with IP, HTTP and WS enabled.

- Executed `block:get 3 --pretty`

- Executed http://127.0.0.1:7887/rpc and
- Executed ws://127.0.0.1:7887/rpc-ws with payload:
{
    "jsonrpc": "2.0",
    "id": "1",
    "method": "dpos_getAllDelegates",
    "params": {
        "height": 544
    }
}

PR #7296, opened 2022-07-13
Expose config values from dpos module - Closes #7213

### LINKED ISSUE #7213: Expose config values from dpos module
### Description

Expose numberOfActiveDelegate and numberOfStandbyDelegate from DPoS module and also check for other values to expose.

Expose
`dpos_getConstants` endpoint with response of 
```
{
	factorSelfVotes: number;
	maxLengthName: number;
	maxNumberSentVotes: number;
	maxNumberPendingUnlocks: number;
	failSafeMissedBlocks: number;
	failSafeInactiveWindow: number;
	punishmentWindow: number;
	roundLength: number;
	bftThreshold: number;
	minWeightStandby: bigint;
	numberActiveDelegates: number;
	numberStandbyDelegates: number;
	tokenIDDPoS: TokenIDDPoS;
}
```

`v6.0.0-alpha.0`

has5aan:
### What was the problem?

This PR resolves #7213

### How was it solved?

Under `framework/src/modules/dpos_v2` modified:

- `type.ts` - defines `ModuleConfigJSON` type.
- `endpoint.ts` - defines `getConstants` endpoint and `init` function to initialise, `ModuleConfig`.
- `module.ts` - calls `DPoSEndPoint.init` providing the module config and configures `getConstants` end-point.
- `utils.ts` - defines `getModuleConfig`, converts `ModuleConfigJSON` to `ModuleConfig`.

Under `framework/test/unit/modules/dpos_v2` modified / created:
- `endpoint.ts` - tests `getConstants` endpoint.
- `utils.ts` - tests `getModuleConfig` utility function.

### How was it tested?

Passing tests.

Testing the response of the getConstants by configuring the modified module with a test node. An approach would be to use the `APIClient`:

```javascript
require('@liskhq/lisk-api-client').createIPCClient('~/.lisk/dpos-mainchain').then(client => {
    client.invoke('dpos_getConstants').then(response => {
        console.log(response)
    })
})
```


framework/src/modules/dpos_v2/schemas.ts
```diff
@@ -655,3 +655,66 @@ export const getAllDelegatesResponseSchema = {
 		},
 	},
 };
+
+// export const getConstantsResponseSchema = {
```
shuse2, line 659: why comment out?

framework/src/modules/dpos_v2/types.ts
```diff
@@ -32,6 +32,22 @@ export interface ModuleConfig {
 	tokenIDDPoS: TokenIDDPoS;
 }
 
+export interface ModuleConfigJSON {
```
shuse2, line 35: you could use
```
export type ModuleConfigJSON = JSONObject<ModuleConfig>;
```

shuse2 (commented):
Generally LGTM, just minor point and the jenkins is failing

framework/src/modules/dpos_v2/types.ts
has5aan: Have changed the code based on the suggestions. Thanks.

ishantiw (approved):
Tested the endpoint and looks good 👏

PR #7298, opened 2022-07-13
Refactor to return last block bytes - Closes #7253

### LINKED ISSUE #7253: `NETWORK_RPC_GET_LAST_BLOCK` is not returning anything
### Expected behavior
`NETWORK_RPC_GET_LAST_BLOCK` network endpoint should return last block.

### Actual behavior
`NETWORK_RPC_GET_LAST_BLOCK` is not returning anything https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/engine/consensus/consensus.ts#L170-L172

### Steps to reproduce
Try to sync 2 nodes

### Additional information

- Add unit test to cover this RPC response

6.0.0-alpha.0

sitetester:
### What was the problem?

This PR resolves #7253

### How was it solved?

code refectored to return a value from an arrow function


shuse2 (approved):
LGTM.
Comment for @ishantiw , we discussed this should be tested in the functional test, but currently functional test is disabled by the zeroMQ issue.
We can recover the test once that problem is solved

ishantiw (approved):

PR #7300, opened 2022-07-14
 Update block execution based on LIP 0055 updates - Closes #7149 

### LINKED ISSUE #7149: Update block execution based on LIP 0055 updates
### Description

LIP 0055 is updated as the [PR](https://github.com/LiskHQ/lips/pull/133).

- Remove `sealBlock` application hook
- Rename `initBlock` application hook to `insertAssets`
- Update block header property from `assetsRoot` to `assetRoot`
- `aggregateCommit` should be inserted in block header initialization
  - Currently it happens [here](https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/engine/generator/generator.ts#L600), but it should happen [here](https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/engine/generator/generator.ts#L488-L492)
- Update state machine for failing transaction. If [command.execute](https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/state_machine/state_machine.ts#L159) fails, the state changed and events logged should be reverted but the block is valid / the transaction should be included. 
  - state and event queue needs to be restored in case of failure.
  - If execution fails, it should add standard event with `success: false` instead of success [here](https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/state_machine/state_machine.ts#L161-L166)
  - If command fails, it should return FAIL status, and if other part of transaction fails, INVALID status should be returned [here](https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/abi_handler/abi_handler.ts#L474)
  - Invalid transaction identified [here](https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/engine/generator/generator.ts#L536) or [here](https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/engine/generator/generator.ts#L711) in the generator should be removed from transaction pool.

### Additional Information
Original [LIP 0055](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0055.md)

Incede:
### What was the problem?

This PR resolves #7149 

### How was it solved?

Update as specified

### How was it tested?

Update tests


framework/src/state_machine/state_machine.ts
```diff
@@ -139,43 +140,73 @@ export class StateMachine {
...
+	public async executeTransaction(ctx: TransactionContext): Promise<number> {
+		let status = TransactionExecutionResult.OK;
 		const transactionContext = ctx.createTransactionExecuteContext();
 		for (const mod of this._systemModules) {
 			if (mod.beforeCommandExecute) {
-				await mod.beforeCommandExecute(transactionContext);
+				try {
+					await mod.beforeCommandExecute(transactionContext);
+				} catch (error) {
+					status = TransactionExecutionResult.INVALID;
```
shuse2, line 151: when status is `INVALID` we can abort the execution

shuse2 (approved):

ishantiw (approved):

PR #7304, opened 2022-07-15
 Transaction dry mode - Closes #5317 

### LINKED ISSUE #5317: Transaction dry mode
### Description

When receiving a transaction, if the optional `dry` flag is set to true, then the node will make sure the transaction is valid/receivable without processing it.

`POST /api/transactions?dry=true`

### Motivation

The idea is to simply to make sure a transaction is valid before submitting it. Application is mostly for developer who wants to develop clients for a running blockchain. One could argue there is most likely a `testnet` for that blockchain, but setting up conditions even on a `testnet` is time consuming whereas most of the time you just want to validate your schema/values & signature.
This would improve DX by removing friction.

Incede:
### What was the problem?

This PR resolves #5317 

### How was it solved?

Implement dry run endpoint for transactions

### How was it tested?

Added unit test


framework/src/engine/endpoint/txpool.ts
```diff
@@ -94,4 +108,44 @@ export class TxpoolEndpoint {
...
+		}
+
+		const stateStore = new StateStore(this._blockchainDB);
+		const consensus = await this._consensus.getConsensusParams(
+			stateStore,
+			this._chain.lastBlock.header,
+		);
+
+		const response = await this._abi.executeTransaction({
+			contextID: Buffer.alloc(1),
```
shuse2, line 138:
```suggestion
			contextID: Buffer.alloc(0),
```
dryRun doesn't require context, so it can use just 0 bytes here as well

shuse2 (changes requested):
only minor point, overall it looks good to me =)

shuse2 (approved):

ishantiw (approved):
🚀

PR #7307, opened 2022-07-15
Encode block while broadcasting to network - Closes #7254

### LINKED ISSUE #7254: When broadcasting a block, the block is not encoded
### Expected behavior
The block should be encoded when transporting to another node

### Actual behavior
The block is not encoded at https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/engine/consensus/consensus.ts#L535

### Steps to reproduce
Try to sync 2 nodes

### Additional information

- Add unit test to cover the broadcast of a [block from consensus](https://github.com/LiskHQ/lisk-sdk/blob/0be4e4ef6aada3c41e239fc7052296d1464d1d76/framework/src/engine/consensus/consensus.ts#L535)

6.0.0-alpha.0-

sitetester:
### What was the problem?

This PR resolves #7254

### How was it solved?

`block.getBytes()` is applied in call to `this._network.send(...)`

### DIFF
--- a/framework/src/engine/consensus/consensus.ts
+++ b/framework/src/engine/consensus/consensus.ts
@@ -535,7 +535,7 @@ export class Consensus {
 		const contextID = await this._verifyAssets(block);
 
 		if (!options.skipBroadcast) {
-			this._network.send({ event: NETWORK_EVENT_POST_BLOCK, data: block });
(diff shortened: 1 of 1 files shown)


ishantiw:
A new branch was created for the same PR and it closed this one automatically. The new related PR https://github.com/LiskHQ/lisk-sdk/pull/7308

PR #7327, opened 2022-07-20
Accept transaction params input from file - Closes #7276

### LINKED ISSUE #7276: Accept transaction params input from file
### Description

Enable users to provide transaction params input from file. 
Update `transaction:create` command to accept params input from file. 

- Add flag as `--file <FILE_PATH>` and when the flag is present then expect the params from the file.
- Only allow one flag either `--file` or `--params`

### Motivation

- When transaction params input length longer
  - User would find it difficult to type the input in CLI
  - When transaction params has inputs like [initValidators](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0043.md#parameters) its harder to type/copy paste into CLI with clarity
 - Easy of use
  
### Acceptance

- Add unit test for params passed as file

mosmartin:
### What was the problem?

This PR resolves [#7276](https://github.com/LiskHQ/lisk-sdk/issues/7276)

### How was it solved?

Added functions to parse the transaction parameters input file

### How was it tested?

Added unit tests.
Run `npm t` at the root of the `commander` directory

### DIFF
--- a/commander/src/bootstrapping/commands/transaction/create.ts
+++ b/commander/src/bootstrapping/commands/transaction/create.ts
@@ -32,7 +32,7 @@ import {
 import { PromiseResolvedType } from '../../../types';
 import { flagsWithParser } from '../../../utils/flags';
 import { getDefaultPath } from '../../../utils/path';
-import { getParamsFromPrompt, getPassphraseFromPrompt } from '../../../utils/reader';
+import { getParamsFromPrompt, getPassphraseFromPrompt, getFileParams } from '../../../utils/reader';
 import {
 	encodeTransaction,
 	getApiClient,
@@ -56,6 +56,7 @@ interface CreateFlags {
 	'no-signature': boolean;
 	'sender-public-key'?: string;
 	nonce?: string;
+	file?: string;
 }
 
 interface Transaction {
@@ -69,14 +70,23 @@ interface Transaction {
 }
 
 const getParamsObject = async (metadata: ModuleMetadataJSON[], flags: CreateFlags, args: Args) => {
+	let params: Record<string, unknown>;
+
 	const paramsSchema = getParamsSchema(
 		metadata,
 		cryptography.utils.intToBuffer(args.moduleID, 4).toString('hex'),
 		cryptography.utils.intToBuffer(args.commandID, 4).toString('hex'),
 	) as Schema;
-	const params = flags.params ? JSON.parse(flags.params) : await getParamsFromPrompt(paramsSchema);
 
-	return params as Record<string, unknown>;
+	if (flags.file) {
+		params = flags.file
+			? JSON.parse(getFileParams(flags.file))
+			: await getParamsFromPrompt(paramsSchema);
+	} else {
+		params = flags.params ? JSON.parse(flags.params) : await getParamsFromPrompt(paramsSchema);
+	}
+
+	return params;
 };
 
 const getPassphraseAddressAndPublicKey = async (flags: CreateFlags) => {
@@ -260,6 +270,7 @@ export abstract class CreateCommand extends Command {
 		}),
 		'data-path': flagsWithParser.dataPath,
 		pretty: flagsWithParser.pretty,
+		file: flagsWithParser.file,
 	};
 
 	protected _client!: PromiseResolvedType<ReturnType<typeof apiClient.createIPCClient>> | undefined;
--- a/commander/src/bootstrapping/commands/transaction/create.ts
+++ b/commander/src/bootstrapping/commands/transaction/create.ts
@@ -79,9 +79,7 @@ const getParamsObject = async (metadata: ModuleMetadataJSON[], flags: CreateFlag
 	) as Schema;
 
 	if (flags.file) {
-		params = flags.file
-			? JSON.parse(getFileParams(flags.file))
-			: await getParamsFromPrompt(paramsSchema);
+		params = JSON.parse(getFileParams(flags.file));
 	} else {
 		params = flags.params ? JSON.parse(flags.params) : await getParamsFromPrompt(paramsSchema);
 	}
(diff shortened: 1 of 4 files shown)


commander/src/utils/flags.ts
```diff
@@ -53,6 +53,11 @@ const prettyDescription = 'Prints JSON in pretty format rather than condensed.';
 
 const outputDescription = 'The output directory. Default will set to current working directory.';
 
+const transactionParametersInputFile = `Transaction parameters input file.
```
shuse2, line 56: I think this description is quite specific to the transaction creation.
Maybe we can just define it in commander/src/bootstrapping/commands/transaction/create.ts

shuse2 (commented):
minor point but overall LGTM, great job 🚀

commander/src/utils/flags.ts
mosmartin: Would making the description more generic work and keep it in the utils where it can be reused?

shuse2: Yes i think that is fine as well. but in that case i think the description should be purely about `file`

mosmartin: Yes, absolutely.

commander/src/utils/reader.ts
```diff
@@ -145,13 +146,17 @@ export const getPasswordFromPrompt = async (
...
+const getFileUnreadableError = (filePath: string): string =>
+	`File at ${filePath} could not be read.`;
 
-const getDataFromFile = (path: string) => fs.readFileSync(path, 'utf8');
+const getDataFromFile = (filePath: string) => fs.readFileSync(filePath, 'utf8');
 
 const ERROR_DATA_MISSING = 'No data was provided.';
 const ERROR_DATA_SOURCE = 'Unknown data source type.';
+const INVALID_JSON_FILE = 'Not a JSON file.';
+const FILE_NOT_FOUND = 'no such file or directory.';
```
Incede, line 159:
```suggestion
const FILE_NOT_FOUND = 'No such file or directory.';
```

commander/src/utils/reader.ts
```diff
@@ -352,3 +357,35 @@ export const getParamsFromPrompt = async (
...
+
+export const getFileParams = (filePath: string): string => {
+	let data = '';
+	const ext = getFileExtension(filePath);
+
+	if (ext === '.json') {
+		data = readParamsFile(filePath);
+	} else {
+		throw new ValidationError(INVALID_JSON_FILE);
+	}
```
Incede, line 388: I think this case is already checked in function `getFileExtension` at L382 and probably redundant.

commander/test/utils/reader.spec.ts
```diff
@@ -162,5 +164,50 @@ describe('reader', () => {
...
+				return expect(fileExtension).toEqual('.json');
+			});
+
+			it('should throw an error if no file extension is passed', () => {
+				const filePath = './some/path';
+				expect(() => getFileExtension(filePath)).toThrow('Not a JSON file.');
+			});
+		});
+
+		describe('readParamsFile', () => {
```
Incede, line 182: maybe also add a test case for failure i.e. `FILE_NOT_FOUND`

commander/test/utils/reader.spec.ts
```diff
@@ -162,5 +164,50 @@ describe('reader', () => {
...
+			beforeEach(() => {
+				jest.spyOn(fs, 'readFileSync').mockReturnValue(JSON.stringify(fileData));
+			});
+
+			it('should read a json file', () => {
+				const result = readParamsFile(filePath);
+				return expect(JSON.parse(result)).toEqual(fileData);
+			});
+
+			it('should have an amount', () => {
```
Incede, line 200: Optional: IMO this and the following expectation could be merged with the 1st case i.e `should read a json file and have an amount and a recipient address` also since they are not being handled separately (if-else or branch logic) in the source and do not impact the test coverage.

commander/src/bootstrapping/commands/transaction/create.ts
```diff
@@ -229,6 +237,8 @@ export abstract class CreateCommand extends Command {
 		'transaction:create 2 0 100000000 --params=\'{"amount":100000000,"recipientAddress":"ab0041a7d3f7b2c290b5b834d46bdc7b7eb85815","data":"send token"}\'',
 		'transaction:create 2 0 100000000 --params=\'{"amount":100000000,"recipientAddress":"ab0041a7d3f7b2c290b5b834d46bdc7b7eb85815","data":"send token"}\' --json',
 		'transaction:create 2 0 100000000 --offline --network mainnet --network-identifier 873da85a2cee70da631d90b0f17fada8c3ac9b83b2613f4ca5fddd374d1034b3 --nonce 1 --params=\'{"amount":100000000,"recipientAddress":"ab0041a7d3f7b2c290b5b834d46bdc7b7eb85815","data":"send token"}\'',
+		'transaction:create 2 0 100000000 --file=/txn_params.json',
```
Incede, line 240: Optional: would be nice to have a test case with this example in `create.spec`

Incede (changes requested):
Looks good in general

commander/src/utils/reader.ts
mosmartin: This is correct. The check is redundant.

commander/src/utils/flags.ts
mosmartin: Done!

commander/src/utils/reader.ts
mosmartin: Updated!

commander/src/utils/reader.ts
mosmartin: Updated!

commander/test/utils/reader.spec.ts
mosmartin: Done!

commander/test/utils/reader.spec.ts
mosmartin: Updated!

commander/src/utils/reader.ts
```diff
@@ -352,3 +357,27 @@ export const getParamsFromPrompt = async (
...
+	try {
+		const params = fs.readFileSync(filePath, 'utf8');
+		return params;
+	} catch (err) {
+		throw new ValidationError(FILE_NOT_FOUND);
+	}
+};
+
+export const getFileParams = (filePath: string): string => {
+	getFileExtension(filePath);
```
Incede, line 381: Minor point: since there is no assignment of the returned value the function & name could be changed to only check and throw as intended.

mosmartin: like instead of `getFileExtension` we have `checkFileExtension`?

Incede: `checkFileExtension` sounds good

mosmartin: Updated!

Incede (approved):

shuse2 (approved):

PR #7339, opened 2022-07-26
Invalid transactions should not make state changes - Closes #7299

### LINKED ISSUE #7299: executeTransaction abi handler should not make state changes when tx is invalid
### Description

When `executeTransaction` is called and the transaction is invalid, it should not make state changes to the context and should not return any event.

Currently, state db and event queue does not support multiple snapshot. Therefore, if transaction becomes invalid by `afterCommandExecute`, it only rolls back to the state before calling `command.execute`.

1. Update event queue to maintain multiple snapshots
2. Update lisk-db to the version which supports multiple snapshot

### Acceptance Criteria
- Add unit test to check if all state is rollbacked if `afterCommandExecute` fails
- Add unit test for event queue to check reverting to the older snapshot

### Additional Information
* Possible implementation of event queue interface
```
const snapshotID = eventQueue.createSnapshot();
eventQueue.revertSnapshot(snapshotID);
```
and all snapshot can be removed when using one of the snapshot

Related: https://github.com/LiskHQ/lisk-db/issues/21

mosmartin:
### What was the problem?

This PR resolves #7299

### How was it solved?

Added the ability to capture snapshots by ID so that it's easier to have multiple snapshots as well as the possibility of reverting to a particular snapshot denoted by an ID.

### How was it tested?

Added and updated existing unit tests
Run `npm test -t state_machine` at the root of the `framework` directory


framework/src/state_machine/state_machine.ts
```diff
@@ -194,6 +194,8 @@ export class StateMachine {
 				try {
 					await mod.afterCommandExecute(transactionContext);
 				} catch (error) {
+					ctx.eventQueue.restoreSnapshot(snapshotID);
+					ctx.stateStore.restoreSnapshot();
```
shuse2, line 198: if it fails on the `afterCommandExecute` it should revert to before calling `executeTransaction`

framework/src/state_machine/event_queue.ts
```diff
@@ -59,18 +59,19 @@ export class EventQueue {
...
-	public createSnapshot(): void {
+	public createSnapshot(): number {
 		this._snapshotIndex = this._events.length;
+		return this._snapshotIndex;
 	}
 
-	public restoreSnapshot(): void {
-		const newEvents = this._events.splice(this._snapshotIndex);
+	public restoreSnapshot(snapshotID: number): void {
+		const newEvents = this._events.splice(snapshotID);
```
shuse2, line 68: this way of generating snapshotID is quite smart. more efficient and straight forward than managing in map 🚀

framework/src/state_machine/state_machine.ts
```diff
@@ -165,7 +165,7 @@ export class StateMachine {
 		}
 		const command = this._getCommand(ctx.transaction.moduleID, ctx.transaction.commandID);
 		// Execute command
-		ctx.eventQueue.createSnapshot();
+		const snapshotID = ctx.eventQueue.createSnapshot();
 		ctx.stateStore.createSnapshot();
```
shuse2, line 169: inside of `createSnapshot` it should be calling stateReadWriter.createSnapshot and that should return `id` ad well similar to event queue

framework/test/unit/state_machine/state_machine.spec.ts
```diff
@@ -205,6 +205,49 @@ describe('state_machine', () => {
...
+				},
+				{
+					moduleID: utils.intToBuffer(2, 4),
+					typeID: Buffer.from([0, 0, 0, 0]),
+					data: utils.getRandomBytes(20),
+					topics: [utils.getRandomBytes(32)],
+				},
+			];
+
+			events.map(e => eventQueue.add(e.moduleID, e.typeID, e.data, e.topics));
```
shuse2, line 231: prefer to use `for...of`. `map` should always have left hand side to receive the result

framework/test/unit/state_machine/state_machine.spec.ts
```diff
@@ -205,6 +205,49 @@ describe('state_machine', () => {
...
+				utils.intToBuffer(3, 4),
+				Buffer.from([0, 0, 0, 1]),
+				utils.getRandomBytes(100),
+				[utils.getRandomBytes(32)],
+			);
+
+			mod.afterCommandExecute.mockImplementation(() => {
+				throw new Error('error');
+			});
+
```
shuse2, line 245: i think it should call `await stateMachine.executeTransaction(ctx);` to see if `mod.afterCommandExecute` fails, it rolls back to the original state/events

shuse2 (changes requested):

framework/package.json
```diff
@@ -45,7 +45,7 @@
 		"@liskhq/lisk-chain": "^0.4.0-alpha.0",
 		"@liskhq/lisk-codec": "^0.3.0-alpha.0",
 		"@liskhq/lisk-cryptography": "^4.0.0-alpha.0",
-		"@liskhq/lisk-db": "^0.3.0-debug.20-b44082d07ab158736d00f7fd1706ca72ca2c9d27",
+		"@liskhq/lisk-db": "^0.3.0-debug.20-5058ddc140c457740c275198deebe158b896f2d7",
```
shuse2, line 48: there are several places need the update. You can search with `"@liskhq/lisk-db"`

framework/test/unit/state_machine/state_machine.spec.ts
mosmartin: Ahh yes, map always returns!

framework/src/state_machine/event_queue.ts
mosmartin: Thanks!!

framework/src/state_machine/event_queue.ts
```diff
@@ -59,18 +59,19 @@ export class EventQueue {
 		});
 	}
 
-	public createSnapshot(): void {
+	public createSnapshot(): number {
 		this._snapshotIndex = this._events.length;
+		return this._snapshotIndex;
```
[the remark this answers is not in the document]
mosmartin: Actually it doesn't. Fixing it.

framework/package.json
mosmartin: Updated all places.

framework/src/state_machine/state_machine.ts
mosmartin: Should stateReadWriter be added to the ctx?

framework/test/unit/state_machine/state_machine.spec.ts
mosmartin: Updated the test.

framework/src/state_machine/state_machine.ts
shuse2: the `stateStore` is `PrefixedStateReadWriter`, so [here](https://github.com/LiskHQ/lisk-sdk/blob/b93207dbd1cd34b2fbe016f57c43c139873754fc/framework/src/state_machine/prefixed_state_read_writer.ts#L120-L127) i think you need to get the snapshotID and use them

mosmartin: Updated!

framework/src/state_machine/state_machine.ts
shuse2: I think `Snapshot` needs to be created and revert if before/afterCommandExecute fails to ensure Invalid transactions not to change the state.

we could do it in this function or the function using this function

mosmartin: Thanks for adding the beforeCommandEecute, I was focussing on the afterCommandExecute only. I have pushed the createSnapshot to before.

shuse2: Ah sorry i think the spec was not so clear. What I meant was
```
const rootSnapshot = xxx.createSnapshot();
try {
beforeCommandExecute()
} catch {
 //  back to root
}
const commandSnapshot = xxx.createSnapshot();
try {
  command.execute()
} catch {
 //  back to command snapshot
}
try {
afterCommandExecute()
} catch {
 //  back to root
}
```

framework/test/unit/state_machine/state_machine.spec.ts
```diff
@@ -205,6 +205,56 @@ describe('state_machine', () => {
...
+				header,
+				assets,
+				networkIdentifier,
+				transaction,
+				currentValidators: [],
+				impliesMaxPrevote: true,
+				maxHeightCertified: 0,
+				certificateThreshold: BigInt(0),
+			});
+			await stateMachine.executeTransaction(ctx);
```
shuse2, line 249: I think to test if it rolls backs to the original test, we need to mock `before/afterCommandExecute` to insert some event, and fail on `afterCommandExecute`?

mosmartin: Ahh yes, need to update the test.

mosmartin: Updated!

framework/test/unit/state_machine/state_machine.spec.ts
```diff
@@ -206,28 +227,52 @@ describe('state_machine', () => {
...
+				utils.getRandomBytes(100),
+				[utils.getRandomBytes(32)],
+			);
+
+			systemMod.beforeCommandExecute.mockImplementation(() => {
+				throw new Error('beforeCommandExecute failed');
+			});
+			mod.beforeCommandExecute.mockImplementation(() => {
+				throw new Error('beforeCommandExecute failed');
+			});
```
shuse2, line 264: If I understand the test scenario correctly
I think these functions are called within the `await stateMachine.executeTransaction(ctx);` so it should be mocked before calling them?
Also, instead of failing in `before` hook, i think we can add an event here, and fail on `after` then check if the end result is the same as original state?

shuse2 (approved):
Great work 🎉

framework/src/state_machine/state_machine.ts
```diff
@@ -143,11 +143,15 @@ export class StateMachine {
 	public async executeTransaction(ctx: TransactionContext): Promise<TransactionExecutionResult> {
 		let status = TransactionExecutionResult.OK;
```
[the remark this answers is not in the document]
mosmartin: Updated!

framework/src/state_machine/state_machine.ts
```diff
@@ -204,6 +212,8 @@ export class StateMachine {
 				try {
 					await mod.afterCommandExecute(transactionContext);
 				} catch (error) {
+					ctx.eventQueue.restoreSnapshot(eventQueueSnapshotID);
+					ctx.stateStore.restoreSnapshot(stateStoreSnapshotID);
 					status = TransactionExecutionResult.INVALID;
```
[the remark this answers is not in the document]
mosmartin: Updated!

framework/src/state_machine/state_machine.ts
```diff
@@ -177,15 +180,15 @@ export class StateMachine {
...
+			ctx.eventQueue.restoreSnapshot(commandEventQueueSnapshotID);
+			ctx.stateStore.restoreSnapshot(commandStateStoreSnapshotID);
 			ctx.eventQueue.add(
 				ctx.transaction.moduleID,
 				EVENT_STANDARD_TYPE_ID,
 				codec.encode(standardEventDataSchema, { success: false }),
 				[ctx.transaction.id],
 			);
-			status = TransactionExecutionResult.FAIL;
+			return TransactionExecutionResult.FAIL;
```
shuse2, line 191: here actually we cannot return `FAIL` because even if it fails, we want to execute `afterCommandExecute`. I think that's why `OK` status was defined in the beginning.

mosmartin: Makes sense if we want proceed with execution not to return 🙂

PR #7340, opened 2022-07-26
Implement and expose consensus endpoints - Closes #7326

### LINKED ISSUE #7326: Expose endpoints from consensus to support CCU creation
### Description

We need certain endpoints to provide data which will be used to create certificate and a CCU transaction within `framework/src/engine/endpoint/consensus.ts`

we need,

- **getBFTParameters**: `consensus_getBFTParameters`
```ts
getBFTParameters = async (
	height: number,
): Promise<{
        prevoteThreshold: string;
	precommitThreshold: string;
	certificateThreshold: string;
	validators: BFTValidator[];
	validatorsHash: string;
}>
```
where `BFTValidator` is,
```ts
BFTValidator {
	address: string;
	bftWeight: string;
	blsKey: string;
}
```

`getBFTHeights(_ctx: RequestContext): Promise<BFTHeights>` where `BFTHeights` is,
```ts
export interface BFTHeights {
	maxHeightPrevoted: number;
	maxHeightPrecommitted: number;
	maxHeightCertified: number;
}
```

### Acceptance Criteria

- Should have unit tests for the above endpoints

ishantiw:
### What was the problem?

This PR resolves #7326

### How was it solved?

[🌱 Add consensus endpoints](https://github.com/LiskHQ/lisk-sdk/commit/0257df682503b43308148289014259c98731ec5d)

[✅ Add unit test for consensus endpoints](https://github.com/LiskHQ/lisk-sdk/commit/f7c2e937445c37adda9f7837055a6ecffe6ba1d5)

### How was it tested?

Run `examples/dpos-mainchain/bin/run start --api-ipc` and use API client to invoke `consensus_getBFTParameters` and `consensus_getBFTHeights`

### DIFF
--- a/framework/src/engine/endpoint/consensus.ts
+++ b/framework/src/engine/endpoint/consensus.ts
@@ -1,0 +1,81 @@
+/*
+ * Copyright © 2022 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+import { StateStore } from '@liskhq/lisk-chain';
+import { Database } from '@liskhq/lisk-db';
+import { BFTAPI } from '../bft';
+import { BFTHeights } from '../bft/types';
+import { RequestContext } from '../rpc/rpc_server';
+
+interface EndpointArgs {
+	bftAPI: BFTAPI;
+	blockchainDB: Database;
+}
+
+export interface BFTValidatorJSON {
+	address: string;
+	bftWeight: string;
+	blsKey: string;
+}
+
+interface BFTParametersJSON {
+	prevoteThreshold: string;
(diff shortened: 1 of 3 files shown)


has5aan (approved):
Both endpoints work.

framework/src/engine/endpoint/consensus.ts
```diff
@@ -0,0 +1,81 @@
...
+	[key: string]: unknown;
+	private readonly _bftAPI: BFTAPI;
+	private readonly _blockchainDB: Database;
+
+	public constructor(args: EndpointArgs) {
+		this._bftAPI = args.bftAPI;
+		this._blockchainDB = args.blockchainDB;
+	}
+
+	public async getBFTParameters(ctx: RequestContext): Promise<BFTParametersJSON> {
```
Incede, line 50:
```suggestion
	public async getBFTParametersJSON(ctx: RequestContext): Promise<BFTParametersJSON> {
```
I think it's important to indicate the returned values are JSON since it's not common.

Incede (approved):
Looks good, minor suggestion.

framework/src/engine/endpoint/consensus.ts
ishantiw: ah actually all the endpoints will always be returning JSON format so if we do rename this then we will need to add suffix to all the endpoints so would be redundant

PR #7347, opened 2022-07-26
Update lisk-validators to return single error and compatible type check - Closes #7156

### LINKED ISSUE #7156: Update lisk-validators to return single error and compatible type check
### Description

Currently, when using `validate` function, 
```
const errors = validator.validate(schema, params)
if (errors.length > 0) {
}
```

`validator.validate<SomeType>(schema, params);`
where error thrown is `LiskValidationError`, and params should be `SomeType` after this function
https://github.dev/LiskHQ/lisk-sdk/blob/00595476a406f9413e77c7381803425d5d22cec9/elements/lisk-validator/src/lisk_validator.ts#L72-L87
```
public validate<T = Record<string, unknown>>(schema: object, data: unknown): asserts data is T {
	if (!this._validator.validate(schema, data)) {
		throw new LiskValidationError(convertErrorsToLegacyFormat(this._validator.errors as LiskErrorObject[]));
	}
}
public validateSchema(schema: object): asserts schema is SchemaObject {
	if (!this._validator.validateSchema(schema)) {
		throw new LiskValidationError(convertErrorsToLegacyFormat(this._validator.errors as LiskErrorObject[]));
	}
}
```

### Motivation

- We should have type guard on the validate function so after this function, type is ensured
- It should not be required to check error length

### Additional Information
- All usages needs to be updated

sitetester:
### What was the problem?

This PR resolves #7156 

### How was it solved?

Existing function is only returning errors and they are thrown at usages. By throwing validation errors  inside function, we are avoiding the duplication of throwing error at each usage.

### How was it tested?
Relevant test cases updated.

### DIFF
--- a/elements/lisk-chain/src/block_header.ts
+++ b/elements/lisk-chain/src/block_header.ts
@@ -12,15 +12,13 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
-import {ed, utils} from '@liskhq/lisk-cryptography';
-import {codec} from '@liskhq/lisk-codec';
-import {LiskValidationError, validator} from '@liskhq/lisk-validator';
+import { ed, utils } from '@liskhq/lisk-cryptography';
+import { codec } from '@liskhq/lisk-codec';
+import { validator, LiskValidationError } from '@liskhq/lisk-validator';
+import { EMPTY_BUFFER, EMPTY_HASH, SIGNATURE_LENGTH_BYTES, TAG_BLOCK_HEADER } from './constants';
+import { blockHeaderSchema, blockHeaderSchemaWithId, signingBlockHeaderSchema } from './schema';
+import { JSONObject } from './types';
 import {LiskErrorObject} from "@liskhq/lisk-validator/dist-node/types";
-import {JSONObject} from './types';
-import {EMPTY_BUFFER, EMPTY_HASH, SIGNATURE_LENGTH_BYTES, TAG_BLOCK_HEADER} from './constants';
-import {blockHeaderSchema, blockHeaderSchemaWithId, signingBlockHeaderSchema} from './schema';
-
-
 
 export interface BlockHeaderAttrs {
 	readonly version: number;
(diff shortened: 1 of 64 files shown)


elements/lisk-chain/src/block_header.ts
```diff
@@ -301,9 +302,7 @@ export class BlockHeader {
 	}
 
 	public getSigningBytes(): Buffer {
-		const blockHeaderBytes = codec.encode(signingBlockHeaderSchema, this._getSigningProps());
-
-		return blockHeaderBytes;
+		return codec.encode(signingBlockHeaderSchema, this._getSigningProps());
```
shuse2, line 305: I think this change was not necessary

elements/lisk-chain/src/transaction.ts
```diff
@@ -130,18 +130,14 @@ export class Transaction {
...
 	public getBytes(): Buffer {
-		const transactionBytes = codec.encode(transactionSchema, this as Record<string, unknown>);
-
-		return transactionBytes;
+		return codec.encode(transactionSchema, this as Record<string, unknown>);
 	}
 
 	public getSigningBytes(): Buffer {
-		const transactionBytes = codec.encode(transactionSchema, ({
+		return codec.encode(transactionSchema, ({
```
shuse2, line 137: same as above

framework/src/engine/generator/network_endpoint.ts
```diff
@@ -81,14 +81,24 @@ export class NetworkEndpoint extends BaseNetworkEndpoint {
...
+				throw error;
+			}
+
+			if (!objectUtils.bufferArrayUniqueItems(decodedData.transactionIds)) {
+				this._logger.warn({ peerId }, 'Received invalid getTransactions body');
 				this.network.applyPenaltyOnPeer({
 					peerId,
 					penalty: 100,
 				});
-				throw new LiskValidationError(errors);
```
shuse2, line 91: in this `if` it also should throw an error, otherwise it's change the behavior

shuse2 (changes requested):
Overall great job 🎉

shuse2 (approved):
Good job 🚀

ishantiw (approved):
Great work @sitetester  🎉

PR #7350, opened 2022-07-27
Fix reward module tokenID validation - Closes #7349 

### LINKED ISSUE #7349: Commander command blockchain:reset needs to be updated
### Expected behavior

Should clear all dbs and allow to re-rerun the node.

### Actual behavior

Error when re-rerunning the node:
```
Message: Invalid validator key.
 Trace: Error: Invalid validator key
```

### Steps to reproduce

Run node for few blocks. Run blockchain:reset. Then run the node again.

development

mitsuaki-u:
### What was the problem?

This PR resolves #7349

### How was it solved?

Added schema validation for config

### How was it tested?

Added unit test

### DIFF
--- a/framework/src/modules/reward/module.ts
+++ b/framework/src/modules/reward/module.ts
@@ -68,7 +68,7 @@ export class RewardModule extends BaseModule {
 			throw new LiskValidationError(errors);
 		}
 		this._moduleConfig = (config as unknown) as ModuleConfig;
-		this._tokenID = Buffer.from(this._moduleConfig.tokenID, 'utf8');
+		this._tokenID = Buffer.from(this._moduleConfig.tokenID, 'hex');
 
 		this.api.init({
 			config: {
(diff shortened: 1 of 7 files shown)


shuse2 (changes requested):
I think in reward module `init` it should set the default value if it's not given.
Also, we should expect config comes as `hex string`, and convert to Buffer inside of `init` function

framework/src/modules/reward/module.ts
```diff
@@ -68,7 +68,7 @@ export class RewardModule extends BaseModule {
 			throw new LiskValidationError(errors);
 		}
 		this._moduleConfig = (config as unknown) as ModuleConfig;
-		this._tokenIDReward = this._moduleConfig.tokenIDReward;
+		this._tokenID = Buffer.from(this._moduleConfig.tokenID, 'utf8');
```
shuse2, line 71:
```suggestion
		this._tokenID = Buffer.from(this._moduleConfig.tokenID, 'hex');
```

framework/src/modules/reward/module.ts
```diff
@@ -102,7 +102,7 @@ export class RewardModule extends BaseModule {
 		await this._tokenAPI.mint(
 			context.getAPIContext(),
 			context.header.generatorAddress,
-			this._tokenIDReward,
+			this._tokenID,
```
shuse2, line 105: I think the variable name was aligned with other modules

shuse2 (changes requested):

framework/src/modules/reward/module.ts
mitsuaki-u: After discussion will keep as is

shuse2 (approved):

PR #7358, opened 2022-08-01
Update the event queue wrapper - Closes #7301

### LINKED ISSUE #7301: Update eventQueue to give a wrapper to manage events at crossChainCommand level
### Description

We need an event wrapper that has the ability to let a CCU transaction to create/restore snapshot when executing `crossChainCommands`.

Currently we have,
https://github.com/LiskHQ/lisk-sdk/blob/bb7a6b2c3dc5bfded929e1209758298b0302f36e/framework/src/state_machine/transaction_context.ts#L94

where `wrappedEventQueue` doesn't have ability to create/restore snapshots.

### Motivation

We need an eventQueue where we can restore/create snapshot before a `crossChainCommand` is processed in a CCU Command.

### Additional Information

- [CCU Command](https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/modules/interoperability/mainchain/commands/cc_update.ts#L276)

mosmartin:
### What was the problem?

This PR resolves #7301

### How was it solved?

Added the create/restore snapshot functionality to the wrapper.

### How was it tested?

Added and updated existing unit tests.

### DIFF
--- a/framework/src/state_machine/api_context.ts
+++ b/framework/src/state_machine/api_context.ts
@@ -12,7 +12,6 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
-import { EVENT_STANDARD_TYPE_ID } from '@liskhq/lisk-chain';
 import { EventQueue } from './event_queue';
 import { PrefixedStateReadWriter, StateDBReadWriter } from './prefixed_state_read_writer';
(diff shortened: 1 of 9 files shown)


ishantiw (approved):
LGTM 🙂

PR #7367, opened 2022-08-04
Update random module - Closes #7341 

### LINKED ISSUE #7341: Update random module according to recent updates to LIP0046
### Description

Based on changes made in [PR:Update LIP0046](https://github.com/LiskHQ/lips/pull/149) we need to update Random module.

- Update [`seedRevealSchema`](https://github.com/LiskHQ/lips/blob/c595ce4a903318791921bf2fad9c8343a6eec98a/proposals/lip-0046.md#seed-reveal-schema)
- Update constant `SEED_REVEAL_HASH_SIZE` to `SEED_LENGTH = 16` as specified in the LIP
- Remove any byte length check for address or seed reveal if exist as it is not part of schema check

### Acceptance Criteria

- Should cover seed reveal schema update in the unit tests
- Add unit tests to check if schema covers length checks

mitsuaki-u:
### What was the problem?

This PR resolves #7341 

### How was it solved?

Updated random module schemas 

### How was it tested?

Updated unit tests

### DIFF
--- a/framework/src/modules/random/utils.ts
+++ b/framework/src/modules/random/utils.ts
@@ -14,7 +14,7 @@
 
 import * as cryptography from '@liskhq/lisk-cryptography';
 import { utils } from '@liskhq/lisk-cryptography';
-import { SEED_REVEAL_HASH_SIZE } from './constants';
+import { SEED_LENGTH } from './constants';
 import { ValidatorSeedReveal } from './types';
 
 export const isSeedValidInput = (
@@ -35,9 +35,7 @@ export const isSeedValidInput = (
 	if (!lastSeed) {
 		return false;
 	}
-	return lastSeed.seedReveal.equals(
-		cryptography.utils.hash(seedReveal).slice(0, SEED_REVEAL_HASH_SIZE),
-	);
+	return lastSeed.seedReveal.equals(cryptography.utils.hash(seedReveal).slice(0, SEED_LENGTH));
 };
 
 export const getSeedRevealValidity = (
@@ -60,7 +58,7 @@ export const getSeedRevealValidity = (
 
 	if (
 		!lastSeed ||
-		lastSeed.seedReveal.equals(cryptography.utils.hash(seedReveal).slice(0, SEED_REVEAL_HASH_SIZE))
+		lastSeed.seedReveal.equals(cryptography.utils.hash(seedReveal).slice(0, SEED_LENGTH))
 	) {
 		return true;
 	}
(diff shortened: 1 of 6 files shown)


framework/src/modules/random/utils.ts
```diff
@@ -60,7 +58,7 @@ export const getSeedRevealValidity = (
 
 	if (
 		!lastSeed ||
-		lastSeed.seedReveal.equals(cryptography.utils.hash(seedReveal).slice(0, SEED_REVEAL_HASH_SIZE))
+		lastSeed.seedReveal.equals(cryptography.utils.hash(seedReveal).slice(0, SEED_LENGTH))
```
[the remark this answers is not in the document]
mitsuaki-u: To your second point, break needs to remain otherwise there is a deepscan error: `Value assigned to variable 'maxheight' at this point is not used afterwards.`

framework/test/unit/modules/random/random_module.spec.ts
```diff
@@ -484,9 +484,12 @@ describe('RandomModule', () => {
...
 
-			await expect(
-				randomModule.verifyAssets(context.getBlockVerifyExecuteContext()),
-			).rejects.toThrow('Size of the seed reveal must be 16, but received 10.');
+			try {
+				await randomModule.verifyAssets(context.getBlockVerifyExecuteContext());
+			} catch (error: any) {
+				// eslint-disable-next-line jest/no-try-expect
+				expect(error?.message).toInclude(`Property '.seedReveal' minLength not satisfied`);
+			}
```
ishantiw, line 492: Why we need to change to try/catch?
```suggestion
				await randomModule.verifyAssets(context.getBlockVerifyExecuteContext());
			} catch (error: any) {
				// eslint-disable-next-line jest/no-try-expect
				expect(error?.message).toInclude(`Property '.seedReveal' minLength not satisfied`);
			}
```

ishantiw (commented):
LGTM, a minor question though

framework/test/unit/modules/random/random_module.spec.ts
mitsuaki-u: For convenience for the jest `toInclude` message since the format of the validator is long and unwieldy. If you have a better suggestion I can update it.

mitsuaki-u: This does not work: `string.includes is not a function`

shuse2: Actually this pattern should be avoided because if the function resolves, this test will pass as well

mitsuaki-u: If there is a better suggestion I can update it.

PR #7369, opened 2022-08-04
Add newBlockHandler for chain connector

### LINKED ISSUE #7221: Create new block handler for chain connector plugin
### Description

In order to function chain connector plugin, we create a handler `handleNewBlockEvent` for listening to `chain_newBlock` and `network_newBlock` event. On receiving a new block,
- Saves the block header to DB
- Checks for aggregate commit in the header, if yes, then save the `aggregateCommit` in the aggregateCommits store and also updates the last height certified from the `aggregateCommit`
- Checks for `ccm` and `liveness` frequency and based on that decides to create a CCU using another handler `createCCU` (if one of the conditions meet then where we check for number of blocks since last sent CCU height)
- `createCCU` calls `getNextCertificateFromAggregateCommits` which will compute the next CCU
- `sendCCU` calls the lisk client api to post the transaction on mainchain
- Now it will call the `cleanup()` where it will check for the last certified height and delete the block headers and `aggregateCommits` from the store.

### Acceptance Criteria

- Should have all the unit tests to handle each conditions within `handleNewBlockEvent` handler

### Additional Info

- Depends on #7219 #7224 #7222

ishantiw:
### What was the problem?

This PR resolves #7221

### How was it solved?

[🌱 Add newBlockHandler for chain connector](https://github.com/LiskHQ/lisk-sdk/commit/a0800a1749fc85bfcf279623ec5ab4cd76b6a870)

### DIFF
--- a/framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
+++ b/framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
@@ -22,16 +22,15 @@ import {
 	BFTHeights,
 	db as liskDB,
 	codec,
-	BFTValidator,
 	chain,
+	BFTParameters,
 } from 'lisk-sdk';
-import { CCM_BASED_CCU_FREQUENCY, LIVENESS_BASED_CCU_FREQUENCY, DB_KEY_CHAIN_CONNECTOR_INFO } from './constants';
+import { CCM_BASED_CCU_FREQUENCY, EMPTY_BYTES, LIVENESS_BASED_CCU_FREQUENCY } from './constants';
 import { getChainConnectorInfo, getDBInstance } from './db';
 import { Endpoint } from './endpoint';
 import { chainConnectorInfoSchema, configSchema } from './schemas';
-import { BFTParameters, ChainConnectorInfo, ChainConnectorPluginConfig, SentCCUs } from './types';
+import { ChainConnectorInfo, ChainConnectorPluginConfig, SentCCUs } from './types';
 
-type KVStore = liskDB.Database;
 interface Data {
 	readonly blockHeader: chain.BlockHeaderJSON;
 }
@@ -48,8 +47,7 @@ export class ChainConnectorPlugin extends BasePlugin<ChainConnectorPluginConfig>
 	private _lastCertifiedHeight!: number;
 	private _ccuFrequency!: CCUFrequencyConfig;
 	private _mainchainAPIClient!: apiClient.APIClient;
-	private _sidechainAPIClient?: apiClient.APIClient;
-	private _chainConnectorDB!: liskDB.Database;
+	private _sidechainAPIClient!: apiClient.APIClient;
 	private _chainConnectorState!: ChainConnectorInfo;
 	private readonly _sentCCUs: SentCCUs = [];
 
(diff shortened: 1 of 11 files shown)


framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
```diff
@@ -62,5 +101,82 @@ export class ChainConnectorPlugin extends BasePlugin<ChainConnectorPluginConfig>
 		if (this._sidechainAPIClient) {
 			await this._sidechainAPIClient.disconnect();
 		}
+		// Save all the state to the DB
+		if (this._chainConnectorState) {
+			await this._chainConnectorDB.set(
+				DB_KEY_CHAIN_CONNECTOR_INFO,
```
Incede, line 107: I think db is using `EMPTY_BYTES` as key

framework-plugins/lisk-framework-chain-connector-plugin/src/types.ts
```diff
@@ -33,4 +33,73 @@ export interface ChainConnectorInfo {
 	blockHeaders: chain.BlockHeader[];
 	aggregateCommits: AggregateCommit[];
 	validatorsHashPreimage: ValidatorsData[];
+	crossChainMessages: CCMsg[];
+}
+
+export interface BFTParameters {
```
Incede, line 39: I think some or more of the types can be exported from the framework instead of creating here.

Incede (changes requested):
Minor suggestions

has5aan (approved):

framework-plugins/lisk-framework-chain-connector-plugin/src/types.ts
```diff
@@ -34,6 +34,67 @@ export interface ChainConnectorInfo {
...
+}
+
+export interface MessageFeeTokenID {
+	chainID: Buffer;
+	localID: Buffer;
+}
+
+export interface MessageFeeTokenIDJSON {
+	chainID: string;
+	localID: string;
```
Incede, line 71: These types can also be imported directly from [here](https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/modules/interoperability/types.ts)

framework-plugins/lisk-framework-chain-connector-plugin/src/types.ts
```diff
@@ -34,6 +34,67 @@ export interface ChainConnectorInfo {
...
+export interface CCMsg {
+	readonly nonce: bigint;
+	readonly moduleID: Buffer;
+	readonly crossChainCommandID: Buffer;
+	readonly sendingChainID: Buffer;
+	readonly receivingChainID: Buffer;
+	readonly fee: bigint;
+	readonly status: number;
+	readonly params: Buffer;
+	readonly id: Buffer;
```
Incede, line 97: same, [here](https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/modules/interoperability/types.ts)

Incede (approved):
Looks good except a few more types can be directly exposed instead of redefining.

framework-plugins/lisk-framework-chain-connector-plugin/src/types.ts
ishantiw: exported

framework-plugins/lisk-framework-chain-connector-plugin/src/types.ts
ishantiw: exported

PR #7370, opened 2022-08-04
 Implement chain connector plugin Endpoint - Closes #7223 

### LINKED ISSUE #7223: Implement chain connector plugin Endpoint
### Description

Implement Chain Connector Plugin Endpoint
- `getSentCCUs` - Lists all the CCUs sent so far. 
- `getAggregateCommits` - Returns a list of all the aggregateCommits stored
- `getValidatorsInfoFromPreimage` - Returns a list of all the validatorsInfo stored in the 

**Note:** All the endpoints exposed should return result in JSON format.

### Acceptance Criteria

- Should have unit tests for all the endpoints

### Additional Info

- Depends on #7220

Incede:
### What was the problem?

This PR resolves #7223

### How was it solved?

Implement endpoints

### How was it tested?

Added unit tests

### DIFF
--- a/framework-plugins/lisk-framework-chain-connector-plugin/src/endpoint.ts
+++ b/framework-plugins/lisk-framework-chain-connector-plugin/src/endpoint.ts
@@ -1,0 +1,43 @@
+/*
+ * Copyright © 2022 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+import { BasePluginEndpoint, PluginEndpointContext, db as liskDB, AggregateCommit } from 'lisk-sdk';
+import { getChainConnectorInfo } from './db';
+import { SentCCUs, ValidatorsData } from './types';
+
+export class Endpoint extends BasePluginEndpoint {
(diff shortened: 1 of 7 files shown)


framework-plugins/lisk-framework-chain-connector-plugin/src/endpoint.ts
```diff
@@ -0,0 +1,43 @@
...
+export class Endpoint extends BasePluginEndpoint {
+	private _db!: liskDB.Database;
+	private _sentCCUs!: SentCCUs;
+
+	public init(db: liskDB.Database, sentCCUs: SentCCUs) {
+		this._db = db;
+		this._sentCCUs = sentCCUs;
+	}
+
+	public getSentCCUs(_context: PluginEndpointContext): SentCCUs {
```
ishantiw, line 28: even if its not async function we have to add it as async coz this is how we register and call all the endpoints exposed from modules/plugins.

framework-plugins/lisk-framework-chain-connector-plugin/src/endpoint.ts
```diff
@@ -0,0 +1,43 @@
...
+		return this._sentCCUs;
+	}
+
+	public async getAggregateCommits(_context: PluginEndpointContext): Promise<AggregateCommit[]> {
```
ishantiw, line 32: we expose only JSON format in endpoints now, same goes for other endpoints

ishantiw (commented):
Approach looks good in general. I think in the issue it was not explicitly mentioned but the endpoints should return JSON format. I have also updated the issue with this note. Will be nice if we can write a util for this, I have added some utils in my open PR which you can use as well https://github.com/LiskHQ/lisk-sdk/pull/7369

has5aan (approved):

PR #7371, opened 2022-08-05
Add new generator endpoints - Closes #7313

### LINKED ISSUE #7313: Add new generator endpoints
### Description

Add new endpoints below for [generator](https://github.dev/LiskHQ/lisk-sdk/blob/development/framework/src/engine/generator/endpoint.ts#L74).

The key should be stored in below format, where data should be defined as `bytes` and depending on the type, it should be encode/decode differently
```
const key = address
const value = encode({
  type: "encrypted" | "plain",
  data: {
    kdf: '',
    cipherText: '',
    iteration: '',
  } | {
    generatorKey: 0x0000,
    generatorPrivateKey: 0x0000,
    blsKey: 0x0000,
    blsPrivateKey: 0x0000,
  }
})
```

In general,
- the input needs to be validated using schema

#### setKey
Req
```
{
  address: 0x0000,
  type: "encrypted" | "plain",
  data: {
    kdf: '',
    cipherText: '',
    iteration: '',
  } | {
    generatorKey: 0x0000,
    generatorPrivateKey: 0x0000,
    blsKey: 0x0000,
    blsPrivateKey: 0x0000,
  }
}
```
Res
```
{}
```
#### getKeys
Req
```
{}
```
Res
```
{
  keys: [{
    address: 0x0000,
    type: "encrypted" | "plain",
    data: {
      kdf: '',
      cipherText: '',
      iteration: '',
    } | {
      generatorKey: 0x0000,
      generatorPrivateKey: 0x0000,
      blsKey: 0x0000,
      blsPrivateKey: 0x0000,
  }
  }],
}
```
#### hasKey
Req
```
{
  address: 0x00000
}
```
Res
```
{
  hasKey: true,
}
```

#### getStatus
Req
```
{
}
```
Res
```
{
  status: [
    address: 0x00000,
    height: 100,
    maxHeightGenerated: 100,
    maxHeightPrevoted: 200,
    enabled: true,
],
}
```
#### setStatus
`setStatus` does not need to check the contradiction

Req
```
{
  Address: 0x00000,
  height: 100,
  maxHeightGenerated: 100,
  maxHeightPrevoted: 200,
}
```
Res
```
{
}
```

#### updateStatus
update status should be the same logic as current `updateStatus` except there is no `overwrite` as `setStatus` will handle that.

Req
```
{
  address: 0x00000,
  enable: true;
  height: 100,
  maxHeightGenerated: 100,
  maxHeightPrevoted: 200,
  password: '';
}
```
Res
```
{
}
```

#### estimateSafeStatus
It should return status which can be used based on the shutdown time using this [algorithm](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0014.md#migrating-to-a-new-forging-node-without-data).

timeShutdown is unix timestamp

Req
```
{
  timeShutdown: 222222222222
}
```
Res
```
{
  height: 100,
  maxHeightGenerated: 100,
  maxHeightPrevoted: 100,
}
```
`maxHeightGenerated` and `maxHeightPrevoted` should be the same value as `height`

### Acceptance Criteria
- Add unit test for each cases
- DB prefix for storing the key can be `1`

shuse2:
### What was the problem?

This PR resolves #7313 

### How was it solved?

- Add endpoints described in the issue
- Fix RPC server response to show the error return from endpoint

### How was it tested?

- Add unit tests
- Run dpos-mainchain example and call RPC endpoints described in the issue
for example,
```
{
    "jsonrpc": "2.0",
    "id": "1",
    "method": "generator_setKeys",
    "params": {
        "address": "d04699e57c4a3846c988f3c15306796f8eae5c1c",
        "type": "plain",
        "data": {
            "generatorKey": "95df18ed6e79a6eaee2aaa44d712c3eedd46f9cf698adaf60eefcd5f54b5f772",
            "generatorPrivateKey": "bbb1f1aaf879efda97401e65ad2c1dbc811bcb035d5f9abef1b24c7bdc6cdcd195df18ed6e79a6eaee2aaa44d712c3eedd46f9cf698adaf60eefcd5f54b5f772",
            "blsKey": "82946cdbbfc5abac814232fa8240a1e1b6c033e6ea659a3d6ff57b771c1069029def23d4905f16a875371722668a7b1c",
            "blsPrivateKey": "25d34d5b1c4e3a373669ab8f00bf12fd4ab4ea8ff603ba6fa6b0ca34e85f8891"    
        }
    }
```

### DIFF
--- a/framework/src/engine/generator/endpoint.ts
+++ b/framework/src/engine/generator/endpoint.ts
@@ -12,32 +12,53 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
-import { encrypt, ed, bls, address as cryptoAddress } from '@liskhq/lisk-cryptography';
+import { encrypt } from '@liskhq/lisk-cryptography';
 import { Batch, Database } from '@liskhq/lisk-db';
 import { dataStructures } from '@liskhq/lisk-utils';
 import { validator } from '@liskhq/lisk-validator';
+import { codec } from '@liskhq/lisk-codec';
+import { Chain } from '@liskhq/lisk-chain';
 import { GeneratorStore } from './generator_store';
 import {
+	generatorKeysExist,
+	getAllGeneratorKeys,
+	getGeneratedInfo,
+	getGeneratorKeys,
 	getLastGeneratedInfo,
 	isEqualGeneratedInfo,
 	isZeroValueGeneratedInfo,
+	setGeneratorKey,
 	setLastGeneratedInfo,
 } from './generated_info';
 import {
+	encryptedMessageSchema,
+	EstimateSafeStatusRequest,
+	estimateSafeStatusRequestSchema,
+	EstimateSafeStatusResponse,
 	GeneratedInfo,
 	GetStatusResponse,
+	HasKeysRequest,
+	hasKeysRequestSchema,
+	plainGeneratorKeysSchema,
+	SetKeysRequest,
+	setKeysRequestSchema,
+	SetStatusRequest,
+	setStatusRequestSchema,
 	UpdateStatusRequest,
 	updateStatusRequestSchema,
 	UpdateStatusResponse,
 } from './schemas';
-import { Consensus, Keypair, Generator } from './types';
+import { Consensus, GeneratorKeys, PlainGeneratorKeyData } from './types';
 import { RequestContext } from '../rpc/rpc_server';
 import { ABI } from '../../abi';
+import { JSONObject } from '../../types';
+import { NotFoundError } from './errors';
 
 interface EndpointArgs {
-	keypair: dataStructures.BufferMap<Keypair>;
-	generators: Generator[];
+	keypair: dataStructures.BufferMap<PlainGeneratorKeyData>;
 	consensus: Consensus;
+	chain: Chain;
+	blockTime: number;
 	abi: ABI;
 }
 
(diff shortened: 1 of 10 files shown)


framework/src/engine/generator/endpoint.ts
```diff
@@ -48,76 +69,84 @@ interface EndpointInit {
...
+		this._chain = args.chain;
+		this._blockTime = args.blockTime;
 	}
 
 	public init(args: EndpointInit) {
 		this._generatorDB = args.generatorDB;
 	}
 
-	// eslint-disable-next-line @typescript-eslint/require-await
 	public async getStatus(_context: RequestContext): Promise<GetStatusResponse> {
```
mosmartin, line 91: I've noticed various `context` param definitions (`_context`, `ctx`, `_ctx`) in the same file/module. It's not a show stopper but should there be consistency in the naming?

shuse2: good point! `_ctx` was used when `ctx` is not used in the function so that typescript compiles, but `_context` shouldn't be used

framework/src/engine/generator/schemas.ts
```diff
@@ -76,12 +114,30 @@ export const updateStatusRequestSchema = {
 		maxHeightPrevoted: {
 			type: 'integer',
 		},
-		overwrite: {
-			type: 'boolean',
+	},
+};
+
+export interface EstimateSafeStatusRequest {
```
mosmartin, line 120: According to the estimateSafeStatus definition in the ticket, the request structure is

{ Address: 0x00000, timeShutdown: 222222222222 }

Address seems to me missing from this interface.

shuse2: ah yes, during the implementation, I realized `address` was actually not used in anything. I will update the issue

mosmartin (approved):
🚀

has5aan (approved):
All endpoints work! 🚀

PR #7373, opened 2022-08-05
Update module endpoint context - Closes #7312

### LINKED ISSUE #7312: Update module endpoint context to have access to module.db
### Description

- Update module endpoint to have access to `module.db`
When handling endpoint in [in-memory channel](https://github.dev/LiskHQ/lisk-sdk/blob/development/framework/src/controller/channels/in_memory_channel.ts#L104-L105), it should add access to `module.db` as `StateStore`. For [IPC channel](https://github.dev/LiskHQ/lisk-sdk/blob/development/framework/src/controller/channels/ipc_channel.ts#L209-L210), it should add interface, but throw error because other process cannot have access to this DB.
- Update function name for the endpoint and insertAsset context to get module.db to be `getOffchainStore` from `getGeneratorStore`

### Acceptance Criteria
- Add unit test for in-memory and ipc channels to make sure `getOffchainStore` is accessible

### Additional Information
- There might be better function name instead of `getOffchainStore`

mosmartin:
### What was the problem?

This PR resolves #7312

### How was it solved?

- Refactored the `getGeneratorStore` function.
-  Updated the `ModuleEndpointContext` interface
- Updated the endpoint handlers

### How was it tested?

Updated the unit tests and mocks.

### DIFF
--- a/framework/src/controller/channels/in_memory_channel.ts
+++ b/framework/src/controller/channels/in_memory_channel.ts
@@ -108,6 +108,10 @@ export class InMemoryChannel extends BaseChannel {
 					const stateStore = new PrefixedStateReadWriter(this._db.newReadWriter());
 					return stateStore.getStore(moduleID, storePrefix);
 				},
+				getOffchainStore: (moduleID: Buffer, storePrefix?: number) => {
+					const stateStore = new PrefixedStateReadWriter(this._db.newReadWriter());
+					return stateStore.getStore(moduleID, storePrefix ?? 0);
+				},
 				getImmutableAPIContext: () =>
 					createImmutableAPIContext(new PrefixedStateReadWriter(this._db.newReadWriter())),
 			}) as Promise<T>;
--- a/framework/src/controller/channels/ipc_channel.ts
+++ b/framework/src/controller/channels/ipc_channel.ts
@@ -214,6 +214,9 @@ export class IPCChannel extends BaseChannel {
 					throw new Error('getImmutableAPIContext cannot be called on IPC channel');
 				},
 				params: request.params ?? {},
+				getOffchainStore: () => {
+					throw new Error('getOffchainStore cannot be called on IPC channel');
+				},
 				logger: this._logger,
 			}) as Promise<T>;
 		}
--- a/framework/src/modules/random/module.ts
+++ b/framework/src/modules/random/module.ts
@@ -107,7 +107,7 @@ export class RandomModule extends BaseModule {
 	}
 
 	public async insertAssets(context: InsertAssetContext): Promise<void> {
-		const generatorSubStore = context.getGeneratorStore(this.id);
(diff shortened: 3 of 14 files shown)


framework/src/controller/channels/in_memory_channel.ts
```diff
@@ -108,6 +108,10 @@ export class InMemoryChannel extends BaseChannel {
 					const stateStore = new PrefixedStateReadWriter(this._db.newReadWriter());
 					return stateStore.getStore(moduleID, storePrefix);
 				},
+				getOffchainStore: (moduleID: Buffer, storePrefix?: number) => {
+					const stateStore = new PrefixedStateReadWriter(this._db.newReadWriter());
```
shuse2, line 112: `this_db` here refers to the `StateDB` but we need `moduleDB` passed to this class and used for the `offchain store`.
Also, if we do so, `PrefixedStateReadWriter` will not make sense so you can use the `StateStore` from the `lisk-chain`

shuse2 (changes requested):

framework/src/controller/channels/in_memory_channel.ts
mosmartin: How about `getStore`? Should remain as-is right? `this._db` refers to `StateDB`?

shuse2: yes, `getStore` should return the StateDB as it is. just offchainStore need to use the module id

mosmartin: Sorry was in the process of updating the comment 😃...thanks for the response

framework/src/controller/channels/in_memory_channel.ts
```diff
@@ -120,4 +145,14 @@ export class InMemoryChannel extends BaseChannel {
...
 		return resp.result;
 	}
+
+	private _getModuleDB(): Database {
+		if (this._moduleDB) {
+			return this._moduleDB;
+		}
+
+		const { data: dbFolder } = systemDirs(this.config.label, this.config.rootPath);
+		this._moduleDB = new Database(path.join(dbFolder, 'module.db'));
```
shuse2, line 155: the database cannot be opened multiple times. the DB instance needs to be passed down to this class through constructor similar to the `_db`

framework/src/controller/channels/in_memory_channel.ts
```diff
@@ -101,16 +105,20 @@ export class InMemoryChannel extends BaseChannel {
 				throw new Error('Handler does not exist.');
 			}
 
-			return handler({
+			return (await handler({
```
shuse2, line 108: Since it return promise, I think no need to `await` before returning?

shuse2 (commented):
Minor question about `await`, but otherwise LGTM 🚀

Incede (approved):

shuse2 (approved):

PR #7394, opened 2022-08-11
Update sdk to remove base fee - Closes #6627

### LINKED ISSUE #6627: Consider removing base fee reference from transaction pool
### Description
Remove base fee reference from transaction pool.

### Motivation

Fee would be one of the modules, and the parameters will should be defined in the module.
Because we currently require those 2 properties, they are part of the genesis config which all module shares.

### Additional Information
When this is achieved, we should remove `baseFees` properties from the genesis config

mitsuaki-u:
### What was the problem?

This PR resolves #6627

### How was it solved?

- Remove base fee from transaction pool
- Remove base fee from lisk-transactions
- Remove base fee from genesis block
- Remove base fee from generator
- Remove base fee from engine
- Remove base fee from modules

### How was it tested?

All tests updated

### DIFF
--- a/commander/src/utils/config.ts
+++ b/commander/src/utils/config.ts
@@ -17,7 +17,6 @@ export const defaultConfig = {
 		// eslint-disable-next-line @typescript-eslint/no-magic-numbers
 		maxTransactionsSize: 15 * 1024, // Kilo Bytes
 		minFeePerByte: 1000,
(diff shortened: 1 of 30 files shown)


ishantiw (approved):

ishantiw:
@sitetester Those are build files

PR #7411, opened 2022-08-16
Get all delegates endpoints should return address of the delegate - Closes #7387

### LINKED ISSUE #7387: Get all delegates endpoints should return address of the delegate
### Description

Add address to delegates record when calling endpoints `dpos_getDelegate` and `dpos_getAllDelegates` in hex string format.

### Acceptance Criteria

- Unit test for the above response should be present
- Should be tested with `examples/dpos-mainchain`

sitetester:
### What was the problem?

This PR resolves #7387 

`address` field is now visible in response JSON

### How was it tested?

test cases updated

### DIFF
--- a/framework/test/unit/modules/dpos_v2/endpoint.spec.ts
+++ b/framework/test/unit/modules/dpos_v2/endpoint.spec.ts
@@ -63,6 +63,7 @@ describe('DposModuleEndpoint', () => {
 		isBanned: false,
 		pomHeights: [0],
 		consecutiveMissedBlocks: 0,
+		address: address.toString('hex'),
 	};
 
 	const config: ModuleConfig = {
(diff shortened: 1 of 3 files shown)


framework/test/unit/modules/dpos_v2/endpoint.spec.ts
```diff
@@ -163,9 +165,20 @@ describe('DposModuleEndpoint', () => {
...
+			const address1Str = address1.toString('hex');
+			const address2Str = address2.toString('hex');
+
+			const addresses = [address1Str, address2Str];
+
+			const delegateData1 = Object.assign(delegateData, { address: address1Str });
+			const delegateData2 = Object.assign(delegateData, { address: address2Str });
+
+			// CAUTION!
+			// getAllDelegates() returns data in random order
```
ishantiw, line 177: I think it's not random, it's based on key, lexicographically. If it was giving different order every time we call then it maybe called random.

ishantiw (approved):

ishantiw:
@sitetester Great Job with the descriptive PR and to give more information on testing 👏

mitsuaki-u (approved):

PR #7415, opened 2022-08-17
Update moduleID usage - Closes #7384

### LINKED ISSUE #7384: Add base store and base event in the base module
### Description

- Add `BaseEvent` and `BaseStore`
  - Add validation for the name is `alphanumeric`
- Add `stores` and `events` map to base module
- Add `stores` and `events` to BaseEndpoint and BaseAPI
- Add validation for uniqueness on `store/moduleID+storePrefix` and `event/name` combination when [registering a module](https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/application.ts#L380)
- Add `id` calculation based on name for Base Store
  - module id is calculated by [this](https://github.com/LiskHQ/lips/pull/163/files#diff-d8333c129d4e3f2c7641b9d30b6a7b86789de6901a81ad0a7f8c2a0dfde1c747R100-R105)
  - Store prefix is calculated by `hash(name)[:2]`
- Update `getStore` function parameters to accept `(moduleID: Buffer, prefix: Buffer)`

### Acceptance Criteria

- All builds and tests are passing 
- Unit tests are added for basic functionalities of `BaseEvent` and `BaseStore`
- Unit tests for collision validations on `registerModule` are added

### Additional Information

- There is PoC in this [commit](https://github.com/LiskHQ/lisk-sdk/commit/34da8a46e542e985f5091e15817d54ca441eb401)

shuse2:
### What was the problem?

This PR resolves #7384

### How was it solved?

- Create`NamedRegistry` to store events and stores class
- Create BaseEvent and BaseStore class
- Update all modules to use new methods

### How was it tested?

- All tests and build should work
- Test dpos mainchain example and token transfer should work as expected

### Additional Information
- BFT module is kept it as it is since it's no longer part of modules


framework/src/modules/base_module.ts
```diff
@@ -44,10 +44,9 @@ export interface ModuleMetadata {
 	}[];
 	events: {
 		typeID: string;
```
ricott1, line 46: We use `name` for identifying events

framework/src/modules/base_event.ts
```diff
@@ -0,0 +1,52 @@
...
+
+	public abstract schema: Schema;
+
+	public get key(): Buffer {
+		return Buffer.from(this._moduleName + this.name, 'utf-8');
+	}
+
+	public get name(): string {
+		const name = this.constructor.name.replace('Store', '');
+		return name.charAt(0).toLowerCase() + name.substr(1);
```
ishantiw, line 32: should we extract this logic out to a function and use it for all the base classes

framework/src/modules/base_store.ts
```diff
@@ -0,0 +1,105 @@
...
+	public constructor(moduleName: string, version = 0) {
+		this._version = version;
+		this._storePrefix = utils.hash(Buffer.from(moduleName, 'utf-8')).slice(0, 4);
+		// eslint-disable-next-line no-bitwise
+		this._storePrefix[0] &= 0x7f;
+		const versionBuffer = Buffer.alloc(2);
+		versionBuffer.writeUInt16BE(this._version, 0);
+		this._subStorePrefix = utils
+			.hash(Buffer.concat([Buffer.from(this.name, 'utf-8'), versionBuffer]))
+			.slice(0, 2);
```
ishantiw, line 60: maybe we can extract all this logic to separate funcs so if we have to change anything we only have to change it in one file, same for other id, prefix, subStore prefix logics

ishantiw (approved):
I reviewed the baseStore, baseEvent, baseModule, namedRegistry logic, looks good, I have couple of comments but those can be fixed in another issue

framework/src/application.ts
```diff
@@ -377,13 +370,8 @@ export class Application {
 	// Private
 	// --------------------------------------
 
-	private _registerModule(mod: BaseModule, validateModuleID = false): void {
+	private _registerModule(mod: BaseModule): void {
 		assert(mod, 'Module implementation is required');
-		if (validateModuleID && mod.id.readInt32BE(0) < MINIMUM_EXTERNAL_MODULE_ID) {
```
mitsuaki-u, line 382: should we validate that name string is unique also?

mitsuaki-u (commented):
Looks good one question regarding validation

framework/src/application.ts
shuse2: https://github.com/LiskHQ/lisk-sdk/pull/7415/files/82465d035550e69f6993f539cbd562736537db4d#diff-cb2314f09fb36abaa9b56f7e516e6753a48441c7e8ac30478fd6ad2a9f998360R295

validation is checked here

framework/src/modules/base_store.ts
shuse2: since it's already too big, ill create another issue

mitsuaki-u (approved):

PR #7438, opened 2022-08-22
Create CCU - Closes #7224

### LINKED ISSUE #7224: Implement createCCU function for chain connector plugin
### Description

Implement `createCCU` function which the following params,

- `sendingChainId` - Config property of the plugin
- `Certificate` - returned by `getNextCertificateFromAggregateCommits`
- `activeValidatorsUpdate`: from the `getActiveValidatorsDiff` [Implementation](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0053.md#computing-the-validators-update)
- `newCertificateThreshold`: If the certificateThreshold used to create certificate.validatorsHash is different from the one stored in the chain account, include the new value in the newCertificateThreshold property. By checking the certificateThreshold with the `certificateThreshold` received from `getBFTParameters`
- `inboxUpdate`: The property inboxUpdate.crossChainMessages lists all CCMs that have been included in the sidechain outbox up to h2 (and which have not been included on the sidechain yet). 
  - In this case, the `messageWitness` will be empty. we can get CCMs from `getChannel` and check the outbox of the receiving chain which is mainchain here. 
  - Compute the inclusion proof for the outbox root of the sidechain account into the mainchain state root. This proof is then used to compute `inboxUpdate.outboxRootWitness`

### Acceptance Criteria

- Should have all the unit tests relevant for the CCU creation

### Additional Info

- It will be better to use all the validity conditions listed here in [LIP0053#parameters-validity](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0053.md#parameters-validity)

has5aan:
### What was the problem?

This PR resolves #7224 

### How was it solved?

Implementation of `_createCCUParams`, `_calculateInboxUpdate`, `_rawStateStoreKey`, `_validateCertificate` and `_verifyLiveness` functions.

### How was it tested?

Implemented unit tests.

### DIFF
--- a/framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
+++ b/framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
@@ -24,12 +24,32 @@ import {
 	codec,
 	chain,
 	BFTParameters,
+	CHAIN_ACTIVE,
+	CHAIN_TERMINATED,
+	LIVENESS_LIMIT,
+	ChainAccount,
+	BFTValidator,
+	OutboxRootWitness,
+	MESSAGE_TAG_CERTIFICATE,
+	ActiveValidator,
+	certificateSchema,
+	cryptography,
+	MODULE_ID_INTEROPERABILITY,
+	STORE_PREFIX_OUTBOX_ROOT,
 } from 'lisk-sdk';
 import { CCM_BASED_CCU_FREQUENCY, EMPTY_BYTES, LIVENESS_BASED_CCU_FREQUENCY } from './constants';
 import { getChainConnectorInfo, getDBInstance, setChainConnectorInfo } from './db';
 import { Endpoint } from './endpoint';
-import { chainConnectorInfoSchema, configSchema } from './schemas';
-import { ChainConnectorInfo, ChainConnectorPluginConfig, SentCCUs } from './types';
+import { chainConnectorInfoSchema, configSchema, ccmSchema } from './schemas';
+import {
+	ChainConnectorInfo,
+	ChainConnectorPluginConfig,
+	SentCCUs,
+	CrossChainUpdateTransactionParams,
+	InboxUpdate,
+	ProveResponse,
+} from './types';
+import { getActiveValidatorsDiff } from './utils';
 
 interface Data {
 	readonly blockHeader: chain.BlockHeaderJSON;
@@ -37,6 +57,24 @@ interface Data {
 interface CCUFrequencyConfig {
 	ccm: number;
 	liveness: number;
+}
+
+interface LivenessValidationResult {
+	status: boolean;
+	isLive: boolean;
+	chainID: Buffer;
+	certificateTimestamp: number;
+	blockTimestamp: number;
+}
+
+interface CertificateValidationResult {
+	status: boolean;
+	livenessValidationResult?: LivenessValidationResult;
+	chainStatus: number;
+	certificate: Certificate;
+	header: chain.BlockHeader;
+	hasValidBLSWeightedAggSig?: boolean;
+	message: string;
 }
 
 export class ChainConnectorPlugin extends BasePlugin<ChainConnectorPluginConfig> {
@@ -140,8 +178,7 @@ export class ChainConnectorPlugin extends BasePlugin<ChainConnectorPluginConfig>
 			height -= 1;
 		}
 
-		// eslint-disable-next-line no-useless-return, consistent-return
-		return;
+		return undefined;
 	}
 
 	// eslint-disable-next-line @typescript-eslint/no-empty-function
@@ -189,6 +226,79 @@ export class ChainConnectorPlugin extends BasePlugin<ChainConnectorPluginConfig>
 		);
 
 		await setChainConnectorInfo(this._chainConnectorPluginDB, chainConnectorInfo);
+	}
+
+	public async calculateCCUParams(
+		sendingChainID: Buffer,
+		certificate: Certificate,
+		newCertificateThreshold: bigint,
+	): Promise<CrossChainUpdateTransactionParams | undefined> {
+		let activeBFTValidatorsUpdate: BFTValidator[];
+		let activeValidatorsUpdate: ActiveValidator[] = [];
+		let certificateThreshold = newCertificateThreshold;
+
+		const blockHeader = this._chainConnectorState.blockHeaders.find(
+			header => header.height === certificate.height,
+		)!;
+
+		const chainAccount = await this._mainchainAPIClient.invoke<ChainAccount>(
+			'interoperability_getChainAccount',
+			{ chainID: sendingChainID },
+		);
+
+		const certificateBytes = codec.encode(certificateSchema, certificate);
+
+		const certificateValidationResult = await this._validateCertificate(
+			certificateBytes,
+			certificate,
+			blockHeader,
+			chainAccount,
+		);
+		if (!certificateValidationResult.status) {
+			this.logger.error(certificateValidationResult, 'Certificate validation failed');
+
+			return undefined;
+		}
+
+		if (!chainAccount.lastCertificate.validatorsHash.equals(certificate.validatorsHash)) {
+			const validatorDataAtCertificate = this._chainConnectorState.validatorsHashPreimage.find(
+				data => data.validatorsHash.equals(blockHeader.validatorsHash!),
+			)!;
+
+			const validatorDataAtLastCertificate = this._chainConnectorState.validatorsHashPreimage.find(
+				data => data.validatorsHash.equals(chainAccount.lastCertificate.validatorsHash),
+			)!;
+
+			if (
+				validatorDataAtCertificate.certificateThreshold ===
+				validatorDataAtLastCertificate.certificateThreshold
+			) {
+				certificateThreshold = BigInt(0);
+			}
+
+			activeBFTValidatorsUpdate = getActiveValidatorsDiff(
+				validatorDataAtLastCertificate.validators,
+				validatorDataAtCertificate.validators,
+			);
+
+			activeValidatorsUpdate = activeBFTValidatorsUpdate.map(
+				validator =>
+					({
+						blsKey: validator.blsKey,
+						bftWeight: validator.bftWeight,
+					} as ActiveValidator),
+			);
+		}
+
+		const inboxUpdate = await this._calculateInboxUpdate(sendingChainID);
+
+		return {
+			sendingChainID,
+			certificate: certificateBytes,
+			activeValidatorsUpdate,
+			certificateThreshold,
+			inboxUpdate,
+		};
 	}
 
 	private async _newBlockhandler(data?: Record<string, unknown>) {
@@ -296,7 +406,7 @@ export class ChainConnectorPlugin extends BasePlugin<ChainConnectorPluginConfig>
 		for (let i = 0; i < validatorData.validators.length; i += 1) {
 			if (aggregateCommit.aggregationBits[i] === 1) {
 				const blsKey = validatorData.validators[i].blsKey.toString('hex');
-				if (blsKeyToBFTWeight[blsKey] === undefined) {
+				if (!blsKeyToBFTWeight[blsKey]) {
 					return false;
 				}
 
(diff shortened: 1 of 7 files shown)


ishantiw (commented):
I think overall logic looks good to me, some general comments

framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
```diff
@@ -191,6 +226,82 @@ export class ChainConnectorPlugin extends BasePlugin<ChainConnectorPluginConfig>
...
+
+		const inboxUpdate = await this._calculateInboxUpdate(sendingChainID);
+
+		return {
+			sendingChainID,
+			certificate: certificateBytes,
+			activeValidatorsUpdate,
+			newCertificateThreshold: certificateThreshold,
+			inboxUpdate,
+		};
```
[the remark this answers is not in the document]
ishantiw: Issue created for the same https://github.com/LiskHQ/lisk-sdk/issues/7457

framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
```diff
@@ -311,8 +422,168 @@ export class ChainConnectorPlugin extends BasePlugin<ChainConnectorPluginConfig>
...
+				size: newInboxSize,
+			});
+			newInboxAppendPath = appendPath;
+			newInboxSize = size;
+			newInboxRoot = root;
+		}
+
+		const db = new InMemoryDatabase();
+
+		const smt = new SparseMerkleTree({
```
ricott1, line 453: I am very confused by this line. Why are we initializing the SMT with the inbox root as root hash?

framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
```diff
@@ -311,8 +422,168 @@ export class ChainConnectorPlugin extends BasePlugin<ChainConnectorPluginConfig>
...
+		await smt.update(outboxKey, newInboxRoot as Buffer);
+
+		const proofSMT = await smt.generateSingleProof(outboxKey);
+
+		const outboxRootWitness: OutboxRootWitness = {
+			bitmap: Buffer.from(proofSMT.binaryBitmap),
+			siblingHashes: proofSMT.siblingHashes,
+		};
+
+		const inboxUpdate: InboxUpdate = {
```
ricott1, line 469: We changed the params schema: https://github.com/LiskHQ/lips/blob/update-interoperability-LIPs/proposals/lip-0053.md#parameters

in particular, the inboxUpdate now does not contain the partnerChainOutboxSize

framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
```diff
@@ -311,8 +422,168 @@ export class ChainConnectorPlugin extends BasePlugin<ChainConnectorPluginConfig>
...
+			crossChainMessages: serialzedCCMs,
+			messageWitness: {
+				partnerChainOutboxSize: BigInt(0),
+				siblingHashes: [],
```
ricott1, line 473: this is only the case if you are posintg all pending CCMs in the CCU, which is not always possible (there is a max ccu size)

ricott1 (changes requested):
I reviewed only `_calculateInboxUpdate` for the moment.

framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
```diff
@@ -311,8 +422,168 @@ export class ChainConnectorPlugin extends BasePlugin<ChainConnectorPluginConfig>
...
+			rootHash: channel.inbox.root,
+		});
+
+		const outboxKey = this._rawStateStoreKey(sideChainID);
+
+		await smt.update(outboxKey, newInboxRoot as Buffer);
+
+		const proofSMT = await smt.generateSingleProof(outboxKey);
+
+		const outboxRootWitness: OutboxRootWitness = {
```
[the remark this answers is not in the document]
ishantiw: @ricott1 
If I understand correctly, empty certificate are only created when we send remaining partial list of ccms. Currently,  in chain-connector plugin, we have kept it simple without support for partial CCU updates. As of now, we let the user set a number of CCMs in config after which a CCU is triggered, ex, if a user sets 10 CCMs, then when we have 10 CCMs after lastCertifiedHeight we trigger a CCU.

I do like the idea to split the functions so to support partial update in future.
> even better by splitting this big function into several pieces: one for calculating the messageWitness, another for the outboxRootWitness

framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
ishantiw: coz of this reason https://github.com/LiskHQ/lisk-sdk/pull/7438#discussion_r957484065 we don't have partial CCU updates to keep it simple right now thats why this `messageWitness` will be empty

framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
ishantiw: @has5aan Update the params schema as described in the LIP update https://github.com/LiskHQ/lips/blob/update-interoperability-LIPs/proposals/lip-0053.md#parameters

framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
ricott1: But before we release we definitely need to have this feature, otherwise a chain could be stuck by just sending enough CCMs

framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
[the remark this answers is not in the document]
ricott1: I see. Maybe I would set the max CCU size rather than the number of CCMs, as that is the real problem

framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
ricott1: Actually I'm just confused about what are you trying to do here. You create a new SMT from an empty DB, passing a non-empty root? Then you update the tree, which should overwrite the root anyway. Then how can you use that SMT to generate a proof? isn't the only key you inserted just the root?

ishantiw: oh instead of in-memory db it should've been stateDB, need to find a way to access the stateDB through plugin or we can create an endpoint to generate proof on application

framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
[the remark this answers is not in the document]
ishantiw: True, I have created an issue to tackle it https://github.com/LiskHQ/lisk-sdk/issues/7458

framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
ishantiw: Issue created to tackle it for `v6.0.0` https://github.com/LiskHQ/lisk-sdk/issues/7458

framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
```diff
@@ -191,6 +226,82 @@ export class ChainConnectorPlugin extends BasePlugin<ChainConnectorPluginConfig>
...
+
+		const chainAccount = await this._mainchainAPIClient.invoke<ChainAccount>(
+			'interoperability_getChainAccount',
+			{ chainID: sendingChainID },
+		);
+
+		const certificateBytes = codec.encode(certificateSchema, certificate);
+
+		const certificateValidationResult = await this._validateCertificate(
+			certificateBytes,
```
ricott1, line 250: Why don't we encode the certificate in the `_validateCertificate` function? Also, can't the `codec.encode` function fail (for instance if the message does not follow the schema)?

ricott1: indeed, it can fail. Than it makes sense to me that we encode as part of the certificate validation, catch the encoding error and return it as a false result (I agree we should use more specific error types)

ishantiw: @has5aan you can move encoding part inside `private async _validateCertificate(` and also return specific error message for different fail conditions for logging to atleast know the reason why certificate validation failed. https://github.com/LiskHQ/lisk-sdk/blob/d656b033dbebbaa61a5711148af9fc044c418ecf/framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts#L247

ishantiw (commented):
I think overall implementation looks good to me, there is a comment and also there is a recent change: 
1. CCU Params schema change => newCertificateThreshold->certificateThreshold
2. `networkIdentifier` is removed and `chainID` is used instead

framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
```diff
@@ -311,8 +423,140 @@ export class ChainConnectorPlugin extends BasePlugin<ChainConnectorPluginConfig>
...
+	): Promise<InboxUpdate> {
+		// TODO: Fetch as many CCMs as configured in _ccuFrequency.ccm, between lastCertificateHeight and certificateHeight - After #7569
+		const serialzedCCMs = this._chainConnectorState.crossChainMessages
+			.slice(0, this._ccuFrequency.ccm)
+			.map(ccm => codec.encode(ccmSchema, ccm));
+
+		const outboxKey = this._rawStateStoreKey(sendingChainID);
+
+		const stateProveResponse = await this._sidechainAPIClient.invoke<ProveResponse>('state_prove', {
+			stateRoot: chainAccount.lastCertificate.stateRoot,
```
shuse2, line 438: is this asking for proof particularly for the `stateRoot` from the last certificate or should it be against current state root?

framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
```diff
@@ -311,8 +423,140 @@ export class ChainConnectorPlugin extends BasePlugin<ChainConnectorPluginConfig>
...
+			queries: [outboxKey],
+		});
+		const inclusionProofOutboxRoot: OutboxRootWitness = {
+			bitmap: stateProveResponse.proof.queries[0].bitmap,
+			siblingHashes: stateProveResponse.proof.siblingHashes,
+		};
+
+		const inboxUpdate: InboxUpdate = {
+			crossChainMessages: serialzedCCMs,
+			messageWitnessHashes: [],
```
shuse2, line 448:
```suggestion
			messageWitness: [],
```
shouldn't it be `messageWitness` like above? also when should this be filled?

framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
```diff
@@ -191,6 +227,82 @@ export class ChainConnectorPlugin extends BasePlugin<ChainConnectorPluginConfig>
...
+			);
+		}
+
+		const inboxUpdate = await this._calculateInboxUpdate(sendingChainID, chainAccount);
+
+		return {
+			sendingChainID,
+			certificate: certificateBytes,
+			activeValidatorsUpdate,
+			newCertificateThreshold: { certificateThreshold },
```
shuse2, line 301: shouldn't this be a value instead of object?

framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
```diff
@@ -311,8 +423,140 @@ export class ChainConnectorPlugin extends BasePlugin<ChainConnectorPluginConfig>
...
+			return result;
+		}
+
+		if (chainAccount.status === CHAIN_ACTIVE) {
+			result.status = true;
+
+			return result;
+		}
+
+		const validatorData = this._chainConnectorState.validatorsHashPreimage.find(
```
shuse2, line 502: does it need undefined check in case there is no preimage found?

framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
```diff
@@ -191,6 +227,82 @@ export class ChainConnectorPlugin extends BasePlugin<ChainConnectorPluginConfig>
 		await setChainConnectorInfo(this._chainConnectorPluginDB, chainConnectorInfo);
 	}
 
+	public async calculateCCUParams(
```
shuse2, line 230: is this mentioned in any LIP as appendix?

shuse2 (changes requested):

framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
shuse2: This comment is invalid since prop name is updated in the LIP

framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
ishantiw: yes it should be current state root

framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
ishantiw: yes its better to have this check here

framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
ishantiw: no its not mentioned anywhere how to calculate CCU Params, just overview of the process to create CCU here https://github.com/LiskHQ/lips/blob/main/proposals/lip-0053.md#creating-and-posting-cross-chain-update-transactions

elements/lisk-tree/src/index.ts
```diff
@@ -17,7 +17,7 @@ import { calculateRootFromUpdateData } from './merkle_tree/calculate';
 import { calculateRootFromRightWitness, verifyRightWitness } from './merkle_tree/right_witness';
 import { MerkleTree } from './merkle_tree/merkle_tree';
 import { calculateMerkleRoot, calculateMerkleRootWithLeaves } from './merkle_tree/utils';
-import { SparseMerkleTree } from './sparse_merkle_tree/sparse_merkle_tree';
+import { SparseMerkleTree, SingleProof } from './sparse_merkle_tree/sparse_merkle_tree';
```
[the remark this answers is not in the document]
ishantiw: Maybe follow the changes made in [this PR](https://github.com/LiskHQ/lisk-sdk/pull/7509) to see the response

framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
```diff
@@ -191,6 +229,79 @@ export class ChainConnectorPlugin extends BasePlugin<ChainConnectorPluginConfig>
...
+		const certificateValidationResult = await this._validateCertificate(
+			certificateBytes,
+			certificate,
+			blockHeader,
+			chainAccount,
+		);
+		if (!certificateValidationResult.status) {
+			this.logger.error(certificateValidationResult, 'Certificate validation failed');
+
+			return undefined;
```
ishantiw, line 261: I think only `return` will suffice
```suggestion
			return;
```

framework-plugins/lisk-framework-chain-connector-plugin/src/utils.ts
```diff
@@ -102,3 +103,38 @@ export const channelDataJSONToObj = (channelData: ChannelDataJSON): ChannelData
...
+	const activeValidatorsUpdate: BFTValidator[] = [];
+
+	for (const newValidator of newValidators) {
+		const existingValidator = currentValidators.find(
+			validator =>
+				Buffer.compare(validator.blsKey, newValidator.blsKey) === 0 &&
+				validator.bftWeight === newValidator.bftWeight,
+		);
+
+		if (existingValidator === undefined) {
```
ishantiw, line 120:
```suggestion
		if (!existingValidator) {
```

framework-plugins/lisk-framework-chain-connector-plugin/src/utils.ts
```diff
@@ -102,3 +103,38 @@ export const channelDataJSONToObj = (channelData: ChannelDataJSON): ChannelData
...
+			activeValidatorsUpdate.push(newValidator);
+		}
+	}
+
+	for (const currentValidator of currentValidators) {
+		const newValidator = newValidators.find(
+			validator => Buffer.compare(validator.blsKey, currentValidator.blsKey) === 0,
+		);
+
+		if (newValidator === undefined) {
```
ishantiw, line 130:
```suggestion
		if (!newValidator) {
```

framework-plugins/lisk-framework-chain-connector-plugin/test/unit/plugin.spec.ts
```diff
@@ -619,15 +627,586 @@ describe('ChainConnectorPlugin', () => {
...
+				certificateBytes,
+				certificate,
+				blockHeader,
+				chainAccount,
+			);
+
+			expect(result.status).toBe(false);
+		});
+
+		it('should not validate if ValidatorsData  for block header is undefined', async () => {
```
ishantiw, line 834:
```suggestion
		it('should not validate if ValidatorsData for block header is undefined', async () => {
```

ishantiw (changes requested):
Some comments but still reviewing the logic and will add more comments if needed

framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
```diff
@@ -191,6 +229,80 @@ export class ChainConnectorPlugin extends BasePlugin<ChainConnectorPluginConfig>
...
+		const certificateValidationResult = await this._validateCertificate(
+			certificateBytes,
+			certificate,
+			blockHeader,
+			chainAccount,
+		);
+		if (!certificateValidationResult.status) {
+			this.logger.error(certificateValidationResult, 'Certificate validation failed');
+
+			return;
```
shuse2, line 261:
```suggestion
			return undefined;
```

framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
```diff
@@ -191,6 +229,80 @@ export class ChainConnectorPlugin extends BasePlugin<ChainConnectorPluginConfig>
...
+					({
+						blsKey: validator.blsKey,
+						bftWeight: validator.bftWeight,
+					} as ActiveValidator),
+			);
+		}
+
+		const inboxUpdate = await this._calculateInboxUpdate(sendingChainID, certificate.stateRoot);
+
+		// eslint-disable-next-line no-useless-return, consistent-return
```
shuse2, line 296: with the above change, I think we can remove this eslint disable

framework-plugins/lisk-framework-chain-connector-plugin/src/chain_connector_plugin.ts
```diff
@@ -311,8 +423,153 @@ export class ChainConnectorPlugin extends BasePlugin<ChainConnectorPluginConfig>
...
+	): Promise<InboxUpdate> {
+		// TODO: Fetch as many CCMs as configured in _ccuFrequency.ccm, between lastCertificateHeight and certificateHeight - After #7569
+		const serialzedCCMs = this._chainConnectorState.crossChainMessages
+			.slice(0, this._ccuFrequency.ccm)
+			.map(ccm => codec.encode(ccmSchema, ccm));
+
+		const outboxKey = this._rawStateStoreKey(sendingChainID);
+
+		const stateProveResponse = await this._sidechainAPIClient.invoke<ProveResponse>('state_prove', {
+			stateRoot: certificateStateRoot,
```
shuse2, line 438: `stateRoot` is not required in the request, it's only possible to generate the proof against the current state

shuse2 (approved):

PR #7450, opened 2022-08-24
 Improve logs on framework - Closes #7376 

### LINKED ISSUE #7376: Improve logs on framework
### Description
In order to track and debug better on the client, logs needs to be improved.
Logs should be added to the below with the specified level

#### Info
- Registered modules, commands, and endpoints for application (when starting the application)
- Available endpoints from engine (when starting the application)
- Detected a fork (height, generator, id, maxHeightPrevoted, maxHeightGenerated)
- Executed a block (height, generator, id, # of transactions, # of assets, # of events)
- Creating new certificate (height, generator)
- Starting/Finishing block synchronization (height, generator, sync mode)
- RPC server log (error | success, method)

#### Debug
- on each error for the state machine functions. (`executeTransaction`, `verifyTransaction` etc)
- Propagating a block (height, generator, id, # of transactions, # of assets)

### Additional Informaition
- Try running dpos-mainchain application, and see if the above logs are sufficient to track the standard behavior

Incede:
### What was the problem?

This PR resolves #7376 

### How was it solved?

Add logs as specified

### How was it tested?

Observe logs after running sample application


framework/src/engine/rpc/rpc_server.ts
```diff
@@ -113,9 +117,11 @@ export class RPCServer {
 				(socket, message) => {
 					this._handleRequest(message)
 						.then(data => {
+							this._logger.info({ status: 'success' }, 'WSServer.start');
```
shuse2, line 120: This is not really about start, but it's for handling ws request. Same for http and IPC

framework/src/engine/rpc/rpc_server.ts
```diff
@@ -113,9 +117,11 @@ export class RPCServer {
...
 							socket.send(JSON.stringify(data));
 						})
 						.catch((error: JSONRPC.JSONRPCError) => {
+							this._logger.info({ status: 'error' }, 'WSServer.start');
```
shuse2, line 124: It would be nice to have error message here. Same for http and IPC

framework/src/application.ts
```diff
@@ -304,6 +305,7 @@ export class Application {
 					.then(() => {
 						this.logger.debug(this._controller.getEvents(), 'Application listening to events');
 						this.logger.debug(this._controller.getEndpoints(), 'Application ready for actions');
+						this.logger.info(this._controller.getEndpoints(), 'Application ready for actions');
```
shuse2, line 308:
```suggestion
						this.logger.info(this._controller.getEndpoints(), 'Application ready for endpoints');
```
also we can remove the `debug` level above

framework/src/engine/consensus/consensus.ts
```diff
@@ -536,11 +576,32 @@ export class Consensus {
...
+					number_of_assets: block.assets.getAll().length,
+				},
+				'Block broadcasted',
+			);
 			this.events.emit(CONSENSUS_EVENT_BLOCK_BROADCAST, {
 				block,
 			});
 		}
 		const events = await this._executeBlock(contextID, stateStore, block);
+		this._logger.info(
```
shuse2, line 594: I think this log should be at the end of this function since it can still fail at this point

shuse2: also property name should always be camel case

shuse2 (changes requested):

shuse2 (approved):

framework/src/engine/consensus/consensus.ts
```diff
@@ -452,6 +462,16 @@ export class Consensus {
 					},
 					'Discarding block due to double forging',
 				);
+				this._logger.info(
```
Phanco, line 465: (Optional) That would look nicer if this obj can be represented by a function

framework/src/engine/rpc/rpc_server.ts
```diff
@@ -151,31 +157,40 @@ export class RPCServer {
...
 			throw new Error(`Method ${method} in ${namespace} is already registered.`);
 		}
 		this._rpcHandler[namespace][method] = handler;
+		this._logger.info({ status: 'success' }, 'RegisterEndpoint');
+		this._logger.info(`Registered ${namespace} endpoint ${method}`);
 	}
 
 	public registerNotFoundEndpoint(handler: NotFoundHandler): void {
 		if (this._notFoundHandler) {
+			this._logger.info({ status: 'error' }, 'RegisterNotFoundEndpoint');
```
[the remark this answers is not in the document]
Incede: As per the issue it is intended to be under `info`
```
RPC server log (error | success, method)
```
The exact reason is unknown to me. Probably the OP(@shuse2 ) can answer.

framework/src/engine/consensus/consensus.ts
Incede: Defining types for logs isn't very common probably since they can change iteratively so I'd leave it as it is unless you've a strong opinion.

Phanco (approved):

framework/src/engine/rpc/rpc_server.ts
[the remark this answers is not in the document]
shuse2: I suggested not to use `error` because it's not the error node operator can handle

PR #7468, opened 2022-09-01
Replace asset.moduleID with asset.module

### LINKED ISSUE #7465: Lisk API client fails on getting blocks
### Expected behavior

Lisk API Client should be able to get blocks and transactions

### Actual behavior

When we get block by ID then it throws error because getAssetSchema is getting schema using `moduleID` instead of `module`
https://github.com/LiskHQ/lisk-sdk/blob/c7d572296ccb9f89f8bf21af64230ecaaedf8a84/elements/lisk-api-client/src/codec.ts#L58

### Steps to reproduce

- Run `/examples/dpos-mainchain/bin/run start --api-ipc`
- In another terminal run `/examples/dpos-mainchain/bin/run block:get ba690c03d637b20a309d0d80281b3dbda5b55680e95894d72c59eb0f3e899d60`

`6.0.0-alpha`

sitetester:
### What was the problem?

This PR resolves #7465 

### How was it solved?

asset.moduleID replaced with asset.module

### How was it tested?
- `cd /lisk-sdk/examples/dpos-mainchain`
-  `nano ~/.lisk/dpos-mainchain/config/default/config.json`
- change to `ipc` like this 
```
"rpc": {
                "modes": ["ipc"]
        },
```
- `./bin/run start`
- `./bin/run block:get bef09f78d1c7e14780046762b630431759919f5255866fae418cea14b9c962d3 --pretty` (in separate terminal)

### DIFF
--- a/elements/lisk-api-client/src/codec.ts
+++ b/elements/lisk-api-client/src/codec.ts
@@ -187,11 +187,7 @@ export const decodeAssets = (
 	);
 	const decodedAssets: DecodedBlockAsset[] = [];
 	for (const asset of assets) {
-		const assetSchema = getAssetDataSchema(
-			blockVersion,
-			{ module: asset.module.toString('hex') },
-			metadata,
-		);
+		const assetSchema = getAssetDataSchema(blockVersion, { module: asset.module }, metadata);
 		const decodedData = codec.decode<Record<string, unknown>>(assetSchema, asset.data);
 		decodedAssets.push({
 			...asset,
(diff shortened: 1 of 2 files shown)


ishantiw (commented):
Looks good, Maybe also rename these error messages `moduleID`->`module`

https://github.com/LiskHQ/lisk-sdk/blob/c7d572296ccb9f89f8bf21af64230ecaaedf8a84/elements/lisk-api-client/src/codec.ts#L45

https://github.com/LiskHQ/lisk-sdk/blob/c7d572296ccb9f89f8bf21af64230ecaaedf8a84/elements/lisk-api-client/src/codec.ts#L52

Phanco (approved):

elements/lisk-api-client/src/codec.ts
```diff
@@ -189,7 +189,7 @@ export const decodeAssets = (
 	for (const asset of assets) {
 		const assetSchema = getAssetDataSchema(
 			blockVersion,
-			{ moduleID: asset.moduleID.toString('hex') },
+			{ module: asset.module.toString('hex') },
```
shuse2, line 192: `asset.module` is string, so it shouldnt need `toString`

ishantiw: @sitetester You need to update BlockAsset and Event interfaces in `lisk-api-client` (module type from Buffer to string)

ishantiw (approved):

shuse2 (approved):

PR #7471, opened 2022-09-01
Remove interoperability_getInteroperabilityStore endpoint

### LINKED ISSUE #7466: Remove interoperability_getInteroperabilityStore endpoint
### Expected behavior

We should not expose protected or private functions from endpoints

### Actual behavior

We expose protected function `getInteroperabilityStore` in the interoperability endpoint which we should modify to be private function

### Steps to reproduce

Run `examples/dpos-mainchain` and see logs for exposed endpoints

`v6.0.0-alpha`

sitetester:
### What was the problem?

This PR resolves #7466

### How was it solved?
`getInteroperabilityStore` method from concrete & abstract class is replaced with a private method in `abstract` class

### How was it tested?
- `cd /lisk-sdk/examples/dpos-mainchain/`
-  `./bin/run start --api-ipc` (it shouldn't show registered endpoints in logs like this)

```bash
18:00:20 INFO lisk-framework: Starting controller (module=lisk:app)
18:00:20 INFO lisk-framework: [
  'app_getRegisteredActions',
  'app_getRegisteredEvents',
  'auth_getAuthAccount',
  'auth_isValidSignature',
  'auth_isValidNonce',
  'validators_validateBLSKey',
  'validators_getValidator',
  'token_getBalances',
  'token_getBalance',
  'token_getTotalSupply',
  'token_getSupportedTokens',
  'token_getEscrowedAmounts',
  'reward_getDefaultRewardAtHeight',
  'random_isSeedRevealValid',
  'dpos_getVoter',
  'dpos_getDelegate',
  'dpos_getAllDelegates',
  'dpos_getConstants'
] Application ready for endpoints (module=lisk:app)
```

### DIFF
--- a/framework/src/modules/interoperability/base_interoperability_endpoint.ts
+++ b/framework/src/modules/interoperability/base_interoperability_endpoint.ts
@@ -36,6 +36,7 @@ export abstract class BaseInteroperabilityEndpoint<
 	T extends BaseInteroperabilityStore
 > extends BaseEndpoint {
 	protected readonly interoperableCCMethods = new Map<string, BaseInteroperableMethod>();
+	protected abstract getInteroperabilityStore: (context: StoreGetter | ImmutableStoreGetter) => T;
 
 	public constructor(
 		protected stores: NamedRegistry,
@@ -153,6 +154,4 @@ export abstract class BaseInteroperabilityEndpoint<
 			size: box.size,
 		};
 	}
-
-	protected abstract getInteroperabilityStore(context: StoreGetter | ImmutableStoreGetter): T;
 }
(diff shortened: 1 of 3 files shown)


framework/src/modules/interoperability/base_interoperability_endpoint.ts
```diff
@@ -31,19 +31,22 @@ import { TerminatedOutboxAccountJSON } from './stores/terminated_outbox';
 import { BaseInteroperabilityStore } from './base_interoperability_store';
 import { chainAccountToJSON } from './utils';
 import { ImmutableStoreGetter, StoreGetter } from '../base_store';
+import { MainchainInteroperabilityStore } from './mainchain/store';
+import { SidechainInteroperabilityStore } from './sidechain/store';
```
ishantiw, line 35: hmm its better if we don't reference to Mainchain or sidechain as they are derived from these base classes, else the base class will depend on these classes that derived.
Maybe we can switch to instance member property instead of member?
Something like,
```ts
protected abstract getInteroperabilityStore: (context: StoreGetter | ImmutableStoreGetter) => T;
```

ishantiw (changes requested):
Make sure base classes don't depend on mainchain/sidechain classes

ishantiw (approved):

mitsuaki-u (approved):

PR #7478, opened 2022-09-05
Store usedHashOnions by address for Random Module - Closes #7437

### LINKED ISSUE #7437: Update random module to store usedHashOnion by address
### Description

Currently all hash onions are stored together for historical reason.
Now we can iterate if needed, we should store the `usedHashOnion` per address.
```
key = address
value = {
  usedHashOnions: {
	readonly count: number;
	readonly height: number;
  }[]
}
```

### Acceptance Criteria

### Additional Information
- This should be done after #7384

has5aan:
### What was the problem?

This PR resolves #7437 

### How was it solved?

Modified `UsedHashOnion` and `UsedHashOnionStoreObject` removing `address`, introducing interface.

### How was it tested?

Implemented unit tests for `UsedHashOnionStoreObject`.

### DIFF
--- a/framework/src/modules/random/module.ts
+++ b/framework/src/modules/random/module.ts
@@ -14,7 +14,7 @@
 
 import { utils } from '@liskhq/lisk-cryptography';
 import { codec } from '@liskhq/lisk-codec';
-import { dataStructures, objects } from '@liskhq/lisk-utils';
+import { objects } from '@liskhq/lisk-utils';
 import { validator } from '@liskhq/lisk-validator';
 import {
 	BlockAfterExecuteContext,
@@ -25,7 +25,7 @@ import {
 } from '../../state_machine';
 import { BaseModule, ModuleInitArgs, ModuleMetadata } from '../base_module';
 import { RandomMethod } from './method';
-import { defaultConfig, EMPTY_KEY, STORE_PREFIX_USED_HASH_ONION } from './constants';
+import { defaultConfig, EMPTY_KEY } from './constants';
 import { RandomEndpoint } from './endpoint';
 import {
 	blockHeaderAssetRandomModule,
@@ -33,11 +33,11 @@ import {
 	isSeedRevealValidResponseSchema,
 	randomModuleConfig,
 } from './schemas';
-import { BlockHeaderAssetRandomModule, UsedHashOnion } from './types';
+import { BlockHeaderAssetRandomModule } from './types';
+import { UsedHashOnion, UsedHashOnionsStore } from './stores/used_hash_onions';
 import { Logger } from '../../logger';
 import { isSeedValidInput } from './utils';
 import { ValidatorRevealsStore } from './stores/validator_reveals';
-import { UsedHashOnionsStore } from './stores/used_hash_onions';
 import { HashOnion, HashOnionStore } from './stores/hash_onion';
 
 export class RandomModule extends BaseModule {
@@ -90,14 +90,16 @@ export class RandomModule extends BaseModule {
 		// Get used hash onions
 		let usedHashOnions: UsedHashOnion[] = [];
 		try {
-			const usedHashOnionsData = await generatorSubStore.get(context, STORE_PREFIX_USED_HASH_ONION);
+			const usedHashOnionsData = await generatorSubStore.get(
+				context,
+				context.header.generatorAddress,
+			);
 			usedHashOnions = usedHashOnionsData.usedHashOnions;
 		} catch (error) {
 			if (!(error instanceof NotFoundError)) {
 				throw error;
 			}
 		}
-
 		// Get next hash onion
 		const nextHashOnion = await this._getNextHashOnion(
 			usedHashOnions,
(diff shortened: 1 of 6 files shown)


framework/src/modules/random/types.ts
```diff
@@ -16,11 +16,13 @@ export interface ModuleConfig {
 	maxLengthRevealsMainchain?: number;
 }
 
-export interface UsedHashOnion {
+export interface UsedHashOnionInfo {
```
has5aan, line 19: Is the name for the interface acceptable, any suggestions for changes? Preferred to create a new interface for two fields as they are being referred enough times to merit for an interface definition.

framework/src/modules/random/stores/used_hash_onions.ts
```diff
@@ -90,4 +56,16 @@ export class UsedHashOnionsStore extends BaseOffchainStore<UsedHashOnionStoreObj
 			throw error;
 		}
 	}
+
+	public async setLatest(
```
has5aan, line 60: Could you please confirm or correct the implementation for `setLatest`, is it supposed to be simple update?

shuse2: I think how it's currently done is bit more complex. the code is around [here](https://github.com/LiskHQ/lisk-sdk/blob/18d9320b5d0179f258b94f454a228f4f134a2546/framework/src/modules/random/module.ts#L112-L137).

Assuming we have correct `nextHashOnion`, it's checking if the usedHashOnion with the same count exist in the list, overwrite otherwise push to the end.
This is because in case of reverting, it's possible that the used hash onion was also reverted; thus, it needs to use the same hash onion.

framework/src/modules/random/stores/used_hash_onions.ts
```diff
@@ -62,10 +58,18 @@ export class UsedHashOnionsStore extends BaseOffchainStore<UsedHashOnionStoreObj
 		address: Buffer,
 		usedHashOnionInfo: UsedHashOnionInfo,
 	): Promise<void> {
-		const UsedHashOnionStoreObject = await this.get(ctx, address);
+		const usedHashOnionStoreObject = await this.get(ctx, address);
```
has5aan, line 61: The changes now update if the hashOnion with same count exists, could you please confirm or correct the understanding?

framework/src/modules/random/stores/used_hash_onions.ts
```diff
@@ -62,10 +58,18 @@ export class UsedHashOnionsStore extends BaseOffchainStore<UsedHashOnionStoreObj
...
 
-		UsedHashOnionStoreObject.usedHashOnions.usedHashOnions.push(usedHashOnionInfo);
+		const index = usedHashOnionStoreObject.usedHashOnions.usedHashOnions.findIndex(
+			ho => ho.count === usedHashOnionInfo.count,
+		);
+
+		if (index === -1) {
```
has5aan, line 67: The changes now update if the usedHashOnion with same count exists, could you please confirm or correct the understanding?

framework/src/modules/random/module.ts
```diff
@@ -25,15 +25,16 @@ import {
...
 import { RandomEndpoint } from './endpoint';
 import {
 	blockHeaderAssetRandomModule,
 	isSeedRevealValidRequestSchema,
 	isSeedRevealValidResponseSchema,
 	randomModuleConfig,
 } from './schemas';
-import { BlockHeaderAssetRandomModule, UsedHashOnion } from './types';
+import { BlockHeaderAssetRandomModule } from './types';
+import { UsedHashOnion } from './stores/used_hash_onions';
```
mosmartin, line 37: I think you can combine this import and the one on line 41 together as they import from the same file. e.g. 
`import { UsedHashOnion , UsedHashOnionsStore } from './stores/used_hash_onions';`

framework/src/modules/random/module.ts
```diff
@@ -122,7 +124,7 @@ export class RandomModule extends BaseModule {
 			usedHashOnions.push(nextUsedHashOnion);
 		}
 
-		const updatedUsedHashOnion = this._filterUsedHashOnions(
+		const updatedUsedHashOnion = generatorSubStore.filterUsedHashOnions(
```
shuse2, line 127: filter happens in the `set` function, so no need to do it here?

framework/src/modules/random/module.ts
```diff
@@ -122,7 +124,7 @@ export class RandomModule extends BaseModule {
 			usedHashOnions.push(nextUsedHashOnion);
```
shuse2, line 124: this index logic is also done inside of `setLatest` function?

framework/src/modules/random/stores/used_hash_onions.ts
```diff
@@ -90,4 +86,71 @@ export class UsedHashOnionsStore extends BaseOffchainStore<UsedHashOnionStoreObj
...
+		}
+
+		await this.set(
+			ctx,
+			address,
+			this.filterUsedHashOnions(usedHashOnionStoreObject.usedHashOnions, finalizedHeight),
+		);
+	}
+
+	public filterUsedHashOnions(
```
shuse2, line 115: probably this can be now private function with above change

framework/test/unit/modules/random/stores/used_hash_onion_store.spec.ts
```diff
@@ -0,0 +1,99 @@
+import { StateStore } from '@liskhq/lisk-chain';
```
shuse2, line 1: license header is missing

framework/test/unit/modules/random/stores/used_hash_onion_store.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+					height: 13,
+				},
+			],
+		};
+
+		beforeEach(async () => {
+			await usedHashOnionStore.set(context, address, defaultUsedHashOnionStoreObject);
+		});
+
+		it('Upsert provided usedHashInfo, corresponding to count and filters usedHashOnions having height greater than provided height', async () => {
```
shuse2, line 90: maybe add cases which tests below?
* adding to the last of the array
* adding to the middle of the array
* filter by finalizedHeight

framework/test/unit/modules/random/stores/used_hash_onion_store.spec.ts
```diff
@@ -0,0 +1,99 @@
...
+		offchainStore = new StateStore(db);
+		usedHashOnionStore = new UsedHashOnionsStore('random');
+		context = {
+			getOffchainStore: (storePrefix: Buffer, subStorePrefix: Buffer) =>
+				offchainStore.getStore(storePrefix, subStorePrefix),
+		};
+	});
+
+	describe('filterUsedHashOnion', () => {
+		it('if highest usedHashOnion has height less than finalizedHeight it is included in the result, rest are removed', async () => {
```
shuse2, line 27: prefer to have the test like
```
it('should xxx
```

shuse2 (changes requested):

framework/test/unit/modules/random/stores/used_hash_onion_store.spec.ts
has5aan: The test case actually covers, update and insert case, however, the filter by finalized heights is missing, I can have the filter by height test case covered in the same test case, but probably dividing them into three different test cases is a better idea.

framework/src/modules/random/module.ts
shuse2: I think then intention was to replace `generatorSubStore.set` and use `setLatest` and remove the duplicate code?

shuse2 (approved):
🚀

mosmartin (approved):
🥳

PR #7497, opened 2022-09-07
Remove hex address - Closes #7469

### LINKED ISSUE #7469: Remove usage of `hex` address
### Description

- For `dataType: bytes`, when using codec functions `toJSON`, `fromJSON`, `decodeJSON`, `encodeJSON`, it should look at `format` property and see how it should be decoded from `string` => `Buffer`. Default should be `hex` but `address` can be used.
- Add `address` format validation
- Update endpoints which uses address in Lisk32 format
- Update configs which uses `hex` address to Lisk32 format
- Update commands which uses `hex` address

### Acceptance Criteria

- There is no `hex` address used anywhere in the system

shuse2:
### What was the problem?

This PR resolves #7469 

### How was it solved?

- Remove all hex address usage from SDK
  - commander
  - config
  - logs
  - endpoints

### How was it tested?

- All test should pass
- Config does not include hex address format
- Added tests to codec and validators


has5aan (approved):
:coffin: `address.toString('hex')`

mosmartin (approved):
🚀

PR #7530, opened 2022-09-14
Add legacy block validation  - Closes #7504

### LINKED ISSUE #7504: Add validation logic for blocks and transaction root
### Description

Add static validation logic for incoming blocks as described in [LIP0063](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0063.md#getting-the-lisk-core-v3-block-history)

This function is called from `sync` function and it can be a util function like below,

`validateBlock(receivedBlock: Buffer, deserializedNextBlock)`

#### Validation:
- To calculate blockID of the incoming block and check its blockID with previousBlockID of the block at height `h + 1`
  - Calculate the block ID from the received blockBytes and compare with previousBlockID
- Should pass schema check as given for that range of blocks
 - Deserialize the block
- Payload should be checked against transactionRoot by calculating transactionRoot from the payload using MerkleTree

### Acceptance Criteria

- Should have unit tests covering the validation steps

shuse2:
### What was the problem?

This PR resolves #7504

### How was it solved?

- Implemented decode logics
- Implemented validation logics

### How was it tested?

- Add tests using blocks from mainnet

### DIFF
--- a/framework/src/engine/legacy/codec.ts
+++ b/framework/src/engine/legacy/codec.ts
@@ -12,14 +12,71 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
-import { blockSchemaV2 } from './schemas';
+import { codec, Schema } from '@liskhq/lisk-codec';
+import { utils } from '@liskhq/lisk-cryptography';
+import { blockHeaderSchemaV2, blockSchemaV2 } from './schemas';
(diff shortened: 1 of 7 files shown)


mosmartin (approved):
🚀 LGTM!

ishantiw (approved):

PR #7535, opened 2022-09-16
Expose endpoints for legacy under endpoints in engine

### LINKED ISSUE #7488: Expose endpoints for legacy under endpoints in engine
### Description

Implement endpoint for legacy under engine/endpoints/legacy and call methods of [`storage.ts`](https://github.com/LiskHQ/lisk-sdk/blob/feature/7136-support-legacy-blocks-and-transactions/framework/src/engine/legacy/storage.ts) class which will be called api clients to get legacy blocks and transactions in JSON format

```ts
public getBlockByID: (blockID, snapshotBlockID) => BlockJSON
public getBlockByHeight: (height, snapshotBlockID) => BlockJSON
```

### Acceptance Criteria

- Should have unit tests for all the endpoints listed

sitetester:
### What was the problem?

This PR resolves #7488

### How was it solved?
Legacy endpoints are now exposed through engine & can be seen on server startup.

### How was it tested?

1. test cases added in framework/test/unit/engine/legacy/endpoint.spec.ts

2. Logs output
Screenshot 2022-09-16 at 02 01 49 PM

### DIFF
--- a/framework/src/engine/engine.ts
+++ b/framework/src/engine/engine.ts
@@ -208,7 +208,6 @@ export class Engine {
 		const legacyEndpoint = new LegacyEndpoint({
 			db: this._legacyDB,
 		});
-		legacyEndpoint.init(this._blockchainDB);
 
 		const chainEndpoint = new ChainEndpoint({
 			chain: this._chain,
--- a/framework/src/engine/engine.ts
+++ b/framework/src/engine/engine.ts
@@ -46,6 +46,7 @@ import { ConsensusEndpoint } from './endpoint/consensus';
 import { EngineConfig } from '../types';
 import { readGenesisBlock } from '../utils/genesis_block';
 import { LegacyChainHandler } from './legacy/legacy_chain_handler';
+import { LegacyEndpoint } from './legacy/endpoint';
 
 const isEmpty = (value: unknown): boolean => {
 	switch (typeof value) {
@@ -203,6 +204,12 @@ export class Engine {
 			logger: this._logger,
 			chainID: this._chain.chainID,
 		});
+
+		const legacyEndpoint = new LegacyEndpoint({
+			db: this._legacyDB,
+		});
+		legacyEndpoint.init(this._blockchainDB);
+
 		const chainEndpoint = new ChainEndpoint({
 			chain: this._chain,
 			genesisBlockTimestamp: genesis.header.timestamp,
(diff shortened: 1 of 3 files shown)


framework/src/engine/engine.ts
```diff
@@ -203,6 +204,12 @@ export class Engine {
 			logger: this._logger,
 			chainID: this._chain.chainID,
 		});
+
+		const legacyEndpoint = new LegacyEndpoint({
+			db: this._legacyDB,
+		});
+		legacyEndpoint.init(this._blockchainDB);
```
shuse2, line 211: I think `legacyEndpoint` won't need blockchain.db. maybe it was my mistake to add `init` function on bootstrap, so feel free to remove this

framework/test/unit/engine/legacy/endpoint.spec.ts
```diff
@@ -13,12 +13,62 @@
...
+	const legacyEndpoint = new LegacyEndpoint({ db: new InMemoryDatabase() as any });
+	const encodedBlock = codec.encode(blockSchemaV2, {
+		header: codec.encode(blockHeaderSchemaV2, blockFixtures[0].header),
+		transactions: blockFixtures[0].transactions,
 	});
 
-	describe('getBlockByHeight', () => {
-		it.todo('test behaviors');
+	legacyEndpoint.storage.getBlockByID = jest.fn().mockResolvedValue(encodedBlock);
+	legacyEndpoint.storage.getBlockByHeight = jest.fn().mockResolvedValue(encodedBlock);
```
shuse2, line 33: stub should be done in `beforeEach` hook, and the syntax is

```
just.spyOn(legacyEndpoint.storage, 'getBlockByID').mockResolvedValue(encodedBlock);
```

The reason why we mock in the beforeEach is to reduce the possibility of non-atomic test, so that stub is always created in the beforeEach hook

shuse2 (changes requested):

framework/test/unit/engine/legacy/endpoint.spec.ts
```diff
@@ -13,12 +13,62 @@
...
+
+	describe('LegacyEndpoint', () => {
+		const matchExpectations = (block: LegacyBlockJSON) => {
```
ishantiw, line 36: a suggestion on naming
```suggestion
		const matchBlockExpectations = (block: LegacyBlockJSON) => {
```

ishantiw (commented):
Apart from Shu's comment, just another minor comment

PR #7536, opened 2022-09-16
Add legacy property to nodeInfo - Closes #7498

### LINKED ISSUE #7498: Add legacy property on nodeInfo to be shared to peers
### Description

Update NodeInfo to have `legacy:  Buffer [ ]` property where it will be an array of `snapshotBlockID` in Buffer format, communicating to the network that they have legacy blocks between 2 hard forks.

- Update customNodeInfoSchema with `legacy:  Buffer [ ]`
- Update customNodeInfo interface

### Acceptance Criteria

- Should have all the unit tests updated for nodeInfo

ishantiw:
### What was the problem?

This PR resolves #7498

### How was it solved?

[🌱 Add legacy property to nodeInfo](https://github.com/LiskHQ/lisk-sdk/commit/30cf910e9a0ca21e92ea70ca8f246e181560d7fb)

### How was it tested?

- Run two nodes with configuration that connects them and observe that there is `legacy` property is being shared

### DIFF
--- a/framework/src/engine/consensus/consensus.ts
+++ b/framework/src/engine/consensus/consensus.ts
@@ -549,11 +549,13 @@ export class Consensus {
 			});
 			await this._executeValidated(block);
 
+			// Since legacy property is optional we don't need to send it here
 			this._network.applyNodeInfo({
 				height: block.header.height,
 				lastBlockID: block.header.id,
 				maxHeightPrevoted: block.header.maxHeightPrevoted,
 				blockVersion: block.header.version,
+				legacy: [], // TODO: call legacyChainHandler to get the updated value https://github.com/LiskHQ/lisk-sdk/issues/7503
 			});
 		});
 	}
--- a/framework/src/engine/consensus/consensus.ts
+++ b/framework/src/engine/consensus/consensus.ts
@@ -555,7 +555,6 @@ export class Consensus {
 				lastBlockID: block.header.id,
 				maxHeightPrevoted: block.header.maxHeightPrevoted,
 				blockVersion: block.header.version,
-				legacy: [], // TODO: call legacyChainHandler to get the updated value https://github.com/LiskHQ/lisk-sdk/issues/7503
 			});
 		});
 	}
--- a/framework/src/engine/network/schema.ts
+++ b/framework/src/engine/network/schema.ts
@@ -15,6 +15,7 @@
 export const customNodeInfoSchema = {
 	$id: '/nodeInfo/custom',
 	type: 'object',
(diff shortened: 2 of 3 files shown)


framework/src/engine/consensus/consensus.ts
```diff
@@ -549,11 +549,13 @@ export class Consensus {
 			});
 			await this._executeValidated(block);
 
+			// Since legacy property is optional we don't need to send it here
 			this._network.applyNodeInfo({
 				height: block.header.height,
 				lastBlockID: block.header.id,
 				maxHeightPrevoted: block.header.maxHeightPrevoted,
 				blockVersion: block.header.version,
+				legacy: [], // TODO: call legacyChainHandler to get the updated value https://github.com/LiskHQ/lisk-sdk/issues/7503
```
shuse2, line 558: It would be nice if consensus don't need to know about `legacyChainHandler`

framework/src/engine/network/schema.ts
```diff
@@ -15,6 +15,7 @@
 export const customNodeInfoSchema = {
 	$id: '/nodeInfo/custom',
 	type: 'object',
+	required: ['height', 'blockVersion', 'lastBlockID', 'maxHeightPrevoted'],
```
shuse2, line 18: `legacy` is missing here? although all of the property in codec is required

framework/src/engine/consensus/consensus.ts
ishantiw: yeah I'm just thinking how to manage this, maybe within the network we store `nodeInfo` and every time we call `applyNodeInfo(<Partial<NodeInfo>>)` we update the passed properties and send the whole nodeInfo to the network

framework/src/engine/network/schema.ts
ishantiw: I had thought in my mind whether we should always send legacy prop everytime we apply nodeInfo coz legacy field might not update once its synced and its unnecessary to send it again and again

framework/src/engine/consensus/consensus.ts
shuse2: yea maybe that would be better approach than adding dependency

framework/src/engine/network/schema.ts
shuse2: for the lisk-codec, all the properties must exist, so you need to send the value anyway although in this case, it's array, so technically it's possible to skip, for consistency we can just add empty array if it doesnt exist

shuse2 (approved):

PR #7541, opened 2022-09-20
Implement sync function of legacy chain handler - Closes  #7547 & #7486

### LINKED ISSUE #7486: Implement syncBlocks function of LegacyChainHandler
### Description

Implement `syncBlocks` function of `LegacyChainHandler`
1. Get a list of connectedPeers from the network
2.  Find peers with legacy info corresponding to `snapshotBlockID` (if it contains `legacy` property with the `snapshotBlockID` as requested then it means it has the legacy blocks history)
3. If no peer is found then throw `PeerNotFound` error
4. Randomly select a peer from the filtered peers list (something like `Math.trunc(Math.random() * filteredPeers.length - 1)`)
5. Request the peer with `getLegacyBlocksFromId` RPC and receive response with request param `{ id: Buffer }`
4.1. if response is invalid or `validateLegacyBlock(blockBytes, nextBlock)` fails then ban the peer with 100 penalty by calling `network.applyPenalty(peerID, banScore)`
6.  Save the blocks and repeat step 1 to 4 until we reach startHeight defined in the config.
7.  Once we reach the startHeight of a bracket we save the bracketInfo and return `setLegacyChainBracketInfo(bracketInfo)`

### Acceptance Criteria

- Should have all the cases covered with the unit tests 

### Addtional Information

- Depends on #7504
- Depends on https://github.com/LiskHQ/lisk-sdk/issues/7498
- Define `PeerNotFound` error with argument as message 
- Relates to https://github.com/LiskHQ/lisk-sdk/issues/7547

### LINKED ISSUE #7547: Implement sync method on LegacyChainHandler
### Description

- Implement `sync` method
  1. Iterate through `brackets` and call `bracketInfo = getLegacyChainBracketInfo(bracket.snapshotBlockID)`
  2. If `bracket.startHeight === bracketInfo.lastHeight` then skip to next bracket and call step 1
  3. If `bracket.startHeight !== bracketInfo.lastHeight` then call `lastBlock = getBlockByHeight(bracketInfo.lastHeight)`
  4. Call `syncBlocks(bracket, lastBlock)`, if `syncBlocks` function throws `PeerNotFound` error then wait for 2 minutes (to let enough shuffling to happen to find peer with legacyBlocks) and call this function again
  5. If syncBlocks function returns then continue with next bracket and call step 1 until all the brackets are synced and then call `applyNodeInfo({ legacy: brackets.map(b => b.snapshotBlockID )})` with all the bracket's snapshotBlockID

### Acceptance Criteria

- Should have unit test for sync method and cover all the cases

### Additional Information

- Depends on #7545
- Relates to https://github.com/LiskHQ/lisk-sdk/issues/7486
- https://github.com/LiskHQ/lisk-sdk/issues/7674

sitetester:
### What was the problem?

This PR resolves #7547 and resolves #7486

### How was it solved?

`syncBlocks()` function is commented with flow of syncing legacy blocks

### DIFF
--- a/framework/src/engine/legacy/legacy_chain_handler.ts
+++ b/framework/src/engine/legacy/legacy_chain_handler.ts
@@ -15,6 +15,7 @@
 import { codec } from '@liskhq/lisk-codec';
 import { P2PRequestPacket } from '@liskhq/lisk-p2p/dist-node/types';
 import { Database, NotFoundError } from '@liskhq/lisk-db';
+import { INVALID_PEER_INFO_PENALTY } from '@liskhq/lisk-p2p/dist-node/constants';
 import { LegacyConfig } from '../../types';
 import { Network } from '../network';
 import { getBlocksFromIdResponseSchema } from '../consensus/schema';
@@ -66,10 +67,6 @@ export class LegacyChainHandler {
 	}
 
 	public async sync() {
-		if (!this._legacyConfig.sync) {
-			return;
-		}
-
 		for (const bracket of this._legacyConfig.brackets) {
 			const encodedBracketInfo = await this._storage.getLegacyChainBracketInfo(
 				Buffer.from(bracket.snapshotBlockID, 'hex'),
@@ -78,6 +75,8 @@ export class LegacyChainHandler {
 				legacyChainBracketInfoSchema,
 				encodedBracketInfo,
 			);
+
+			// means this bracket is already synced/parsed (in next `syncBlocks` step)
 			if (bracket.startHeight === bracketInfo.lastBlockHeight) {
 				continue;
 			}
@@ -85,9 +84,11 @@ export class LegacyChainHandler {
 				await this._storage.getBlockByHeight(bracketInfo.lastBlockHeight),
 			).block;
 
+			// start parsing bracket from `lastBlock` height`
 			await this._trySyncBlocks(bracket, lastBlock);
 		}
 
+		// when ALL brackets are synced/parsed, finally update node with it's `legacy` property
 		this._network.applyNodeInfo({
 			legacy: this._legacyConfig.brackets.map(bracket =>
 				Buffer.from(bracket.snapshotBlockID, 'hex'),
@@ -115,16 +116,17 @@ export class LegacyChainHandler {
 	/**
 	 * Flow of syncing legacy blocks
 	 *
-	 * Check if we have `sync` property `true` in configuration
-	 * getConnectedPeers from network
-	 * Filter peers having their `legacy` buffer array property contains `snapshotBlockID`
+	 * Check if we have `sync` property `true` in configuration (already done in engine.ts)
+	 * call getConnectedPeers from network
+	 * Filter peers having their `legacy` buffer array contains `snapshotBlockID`
 	 * If there is no peer, throw error, this error will be used in outside function to retry calling `syncBlocks` after x amount of time
 	 * Get a random peer from list of filtered peers with legacy info
 	 * Make a request to that random peer by calling its `getLegacyBlocksFromId` method with `data` property set to `legacyBlock.header.id`
-	 * Try to decode response data buffer into Block, apply penalty in case of error
-	 * Start saving parsed blocks one by one
-	 * repeat, if last parsed block height is still higher than bracket.startHeight
-	 * finally, update node info
+	 * Try to decode response data buffer using getBlocksFromIdResponseSchema, apply penalty in case of error & retry syncBlocks
+	 * Validate `blocks: Buffer[]`, apply penalty in case of error & retry syncBlocks
+	 * Decode `blocks: Buffer[]` into LegacyBlock[] & start saving one by one
+	 * If last block height is still higher than bracket.startHeight, save bracket with `lastBlockHeight: lastBlock?.header.height` & repeat syncBlocks
+	 * If last block height equals bracket.startHeight, simply save bracket with `lastBlockHeight: lastBlock?.header.height`
 	 */
 	// eslint-disable-next-line @typescript-eslint/member-ordering
 	public async syncBlocks(bracket: LegacyBlockBracket, legacyBlock: LegacyBlock): Promise<void> {
@@ -148,26 +150,37 @@ export class LegacyChainHandler {
 		};
 
 		const response = await this._network.requestFromPeer({ ...p2PRequestPacket, peerId });
-		if (!(response.data as []).length) {
-			return;
-		}
 
 		// `data` is expected to hold blocks in DESC order
 		const { data } = response;
 		let legacyBlocks: LegacyBlock[];
+
+		const applyPenaltyAndRepeat = async () => {
+			this._network.applyPenaltyOnPeer({ peerId, penalty: INVALID_PEER_INFO_PENALTY });
+			await this.syncBlocks(bracket, legacyBlock);
+		};
+
 		try {
 			// this part is needed to make sure `data` returns ONLY `{ blocks: Buffer[] }` & not any extra field(s)
 			const { blocks } = codec.decode<{ blocks: Buffer[] }>(
 				getBlocksFromIdResponseSchema,
 				data as Buffer,
 			);
+			if (!(blocks.length > 0)) {
+				await applyPenaltyAndRepeat();
+			}
+
 			this._applyValidation(blocks);
+
 			legacyBlocks = blocks.map(block => decodeBlock(block).block);
+			if (!(legacyBlocks.length > 0)) {
+				await applyPenaltyAndRepeat();
+			}
 		} catch (err) {
-			this._network.applyPenaltyOnPeer({ peerId, penalty: 100 });
-			return;
+			await applyPenaltyAndRepeat(); // catch validation error
 		}
 
+		// @ts-expect-error Variable 'legacyBlocks' is used before being assigned.
 		for (const block of legacyBlocks) {
 			if (block.header.height > bracket.startHeight) {
 				await this._storage.saveBlock(
@@ -178,6 +191,7 @@ export class LegacyChainHandler {
 			}
 		}
 
+		// @ts-expect-error Variable 'legacyBlocks' is used before being assigned.
 		const lastBlock = legacyBlocks[legacyBlocks.length - 1];
 		if (lastBlock && lastBlock.header.height > bracket.startHeight) {
 			await this._updateBracketInfo(lastBlock, bracket);
--- a/framework/src/engine/legacy/legacy_chain_handler.ts
+++ b/framework/src/engine/legacy/legacy_chain_handler.ts
@@ -12,56 +12,202 @@
  * Removal or modification of this copyright notice is prohibited.
  */
 
-import { Database, InMemoryDatabase } from '@liskhq/lisk-db';
+import { codec } from '@liskhq/lisk-codec';
+import { P2PRequestPacket } from '@liskhq/lisk-p2p/dist-node/types';
+import { Database, NotFoundError } from '@liskhq/lisk-db';
 import { LegacyConfig } from '../../types';
+import { Network } from '../network';
+import { getBlocksFromIdResponseSchema } from '../consensus/schema';
 import { Storage } from './storage';
+import { LegacyBlock, LegacyBlockBracket, Peer, LegacyChainBracketInfo } from './types';
+import { decodeBlock, encodeBlock } from './codec';
+import { PeerNotFoundWithLegacyInfo } from './errors';
+import { validateLegacyBlock } from './validate';
+import { legacyChainBracketInfoSchema } from './schemas';
 
 interface LegacyChainHandlerArgs {
 	legacyConfig: LegacyConfig;
+	network: Network;
 }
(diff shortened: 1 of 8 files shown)


framework/src/engine/legacy/legacy_chain_handler.ts
```diff
@@ -12,41 +12,170 @@
...
+
+		for (const bracket of this._legacyConfig.brackets) {
+			const bracketInfo = await this._storage.getLegacyChainBracketInfo(bracket.snapshotBlockID);
+			if (bracket.startHeight === bracketInfo.lastHeight) {
+				continue;
+			}
+			const lastBlock = await this._storage.getBlockByHeight(bracketInfo.lastHeight);
+			const { id } = decodeBlock(lastBlock).block.header;
+			await this._trySyncBlocks(bracket, Buffer.from(id).toString('hex'));
+			await this._storage.saveBracketInfo(bracketInfo);
```
ishantiw, line 64: here we don't set updated bracketInfo so we need to save the updated one after the sync

framework/src/engine/legacy/legacy_chain_handler.ts
```diff
@@ -12,41 +12,170 @@
...
+		});
+	}
+
+	private async _trySyncBlocks(bracket: LegacyBlockBracket, lastBlock: string) {
+		try {
+			await this.syncBlocks(bracket, lastBlock);
+		} catch (err) {
+			if (err instanceof PeerNotFoundWithLegacyInfo) {
+				// eslint-disable-next-line @typescript-eslint/no-misused-promises
+				setTimeout(async () => {
```
ishantiw, line 78: create a private variable in this class to capture setTimeout return value and use it to clearTimeout after we finish sync.

ishantiw (changes requested):
Overall approach looks good to me, few comments

framework/test/unit/engine/legacy/legacy_chain_handler.spec.ts
```diff
@@ -48,7 +122,17 @@ describe('Legacy Chain Handler', () => {
 		});
 	});
 
-	describe('syncBlocks', () => {
-		it.todo('test behaviors');
+	describe('sync', () => {
+		it('test behaviors', async () => {
```
mosmartin, line 126: I've noticed you've not used the convention `it(should...)` as well as you've not defined the behavior that is being tested.

framework/src/engine/legacy/legacy_chain_handler.ts
```diff
@@ -12,56 +12,193 @@
...
-	db: Database | InMemoryDatabase;
+	db: Database;
 }
 
 export class LegacyChainHandler {
+	private readonly _network: Network;
+	private _storage!: Storage;
 	private readonly _legacyConfig: LegacyConfig;
-	private _db!: Database | InMemoryDatabase;
+	private _timeout!: NodeJS.Timeout;
```
ishantiw, line 41: maybe give more meaning name here

framework/src/engine/legacy/legacy_chain_handler.ts
```diff
@@ -12,56 +12,193 @@
...
-				Buffer.from(bracket.snapshotBlockID, 'hex'),
-			);
-
-			if (!bracketInfo) {
-				await storage.setLegacyChainBracketInfo(Buffer.from(bracket.snapshotBlockID), {
+			try {
+				await this._storage.getLegacyChainBracketInfo(Buffer.from(bracket.snapshotBlockID, 'hex'));
+			} catch (err) {
+				// Save config brackets in advance, these will be used in next step (`sync`)
+				await this._storage.setLegacyChainBracketInfo(Buffer.from(bracket.snapshotBlockID), {
```
ishantiw, line 56: If there is any other error apart from key not found we should handle it 
```ts
if (!(error instanceof NotFoundError)) {
				throw error;
			}
```

framework/src/engine/legacy/legacy_chain_handler.ts
```diff
@@ -12,56 +12,193 @@
...
+				continue;
+			}
+			const lastBlock = decodeBlock(
+				await this._storage.getBlockByHeight(bracketInfo.lastBlockHeight),
+			).block;
+
+			await this._trySyncBlocks(bracket, lastBlock);
+			await this._storage.setLegacyChainBracketInfo(
+				Buffer.from(bracket.snapshotBlockID, 'hex'),
+				bracketInfo,
```
ishantiw, line 88: you need to update the bracket info with the last block after syncing which will be the block at the startHeight. Its better to set bracket everytime we finish one call with the peer, so once we process 100 blocks received, we should set bracket and then call another peer for next 100 blocks.

framework/src/engine/legacy/legacy_chain_handler.ts
```diff
@@ -12,56 +12,193 @@
...
+					block.header.id as Buffer,
+					block.header.height,
+					encodeBlock(block),
+				);
+				lastBlock = block;
+			}
+		}
+
+		if (lastBlock && lastBlock.header.height > bracket.startHeight) {
+			await this.syncBlocks(bracket, lastBlock);
```
ishantiw, line 185: we can add setbracket with updated bracketInfo before calling syncBlocks

framework/src/engine/legacy/legacy_chain_handler.ts
```diff
@@ -12,56 +12,193 @@
...
+		}
```
ishantiw, line 186: before we exit the sync , we update bracketInfo

framework/test/unit/engine/legacy/legacy_chain_handler.spec.ts
```diff
@@ -14,32 +14,144 @@
...
+			},
+			{
+				peerId: 'peerId-2',
+				options: {
+					legacy: [randomSnapshotBlockID, expectedSnapshotBlockID],
 				},
+			},
+		];
+		// CAUTION: This is copy of blockFixtures[0](19583715), ONLY height has changed
+		legacyBlock19583716 = {
```
ishantiw, line 73: best would be to move this to fixture file

ishantiw (changes requested):

mosmartin (commented):
Besides the open comments, LGTM 🚀

mosmartin (approved):
Having another look.

framework/src/engine/legacy/legacy_chain_handler.ts
```diff
@@ -12,56 +12,202 @@
...
-		// TODO: implement the logic, general flow is as below
-		// 1. check config if sync legacy flag is true
-		// 2. Checks to what height blocks are present for a snapshotBlockID backwards
-		// 3. checks for all the peers node info if any peer has legacy blocks for the snapshot block ID, if not wait for X second
-		// 4. sync the blocks
-		// 5. Update node info
+	public async sync() {
+		if (!this._legacyConfig.sync) {
+			return;
+		}
```
ishantiw, line 71: this is already checked in engine so we don't need to check here again

framework/src/engine/legacy/legacy_chain_handler.ts
```diff
@@ -12,56 +12,202 @@
...
+	 * Check if we have `sync` property `true` in configuration
+	 * getConnectedPeers from network
+	 * Filter peers having their `legacy` buffer array property contains `snapshotBlockID`
+	 * If there is no peer, throw error, this error will be used in outside function to retry calling `syncBlocks` after x amount of time
+	 * Get a random peer from list of filtered peers with legacy info
+	 * Make a request to that random peer by calling its `getLegacyBlocksFromId` method with `data` property set to `legacyBlock.header.id`
+	 * Try to decode response data buffer into Block, apply penalty in case of error
+	 * Start saving parsed blocks one by one
+	 * repeat, if last parsed block height is still higher than bracket.startHeight
+	 * finally, update node info
```
ishantiw, line 127: I think the description needs to be updated according to the latest code

framework/src/engine/legacy/legacy_chain_handler.ts
```diff
@@ -12,56 +12,202 @@
...
+			// this part is needed to make sure `data` returns ONLY `{ blocks: Buffer[] }` & not any extra field(s)
+			const { blocks } = codec.decode<{ blocks: Buffer[] }>(
+				getBlocksFromIdResponseSchema,
+				data as Buffer,
+			);
+			this._applyValidation(blocks);
+			legacyBlocks = blocks.map(block => decodeBlock(block).block);
+		} catch (err) {
+			this._network.applyPenaltyOnPeer({ peerId, penalty: 100 });
+			return;
```
ishantiw, line 168: I think here it should throw error instead of return otherwise , it will continue to the syncing next bracket without finishing this bracket

framework/src/engine/legacy/legacy_chain_handler.ts
```diff
@@ -12,56 +12,202 @@
...
+		const randomPeerIndex = Math.trunc(Math.random() * peersWithLegacyInfo.length - 1);
+		const { peerId } = peersWithLegacyInfo[randomPeerIndex];
+
+		const p2PRequestPacket: P2PRequestPacket = {
+			procedure: 'getLegacyBlocksFromId',
+			data: legacyBlock.header.id,
+		};
+
+		const response = await this._network.requestFromPeer({ ...p2PRequestPacket, peerId });
+		if (!(response.data as []).length) {
```
ishantiw, line 151: I think this condition seems irrelevant, as we are checking response.data which is a buffer that is decoded at line 165

framework/src/engine/legacy/legacy_chain_handler.ts
```diff
@@ -12,56 +12,202 @@
...
+		const { data } = response;
+		let legacyBlocks: LegacyBlock[];
+		try {
+			// this part is needed to make sure `data` returns ONLY `{ blocks: Buffer[] }` & not any extra field(s)
+			const { blocks } = codec.decode<{ blocks: Buffer[] }>(
+				getBlocksFromIdResponseSchema,
+				data as Buffer,
+			);
+			this._applyValidation(blocks);
+			legacyBlocks = blocks.map(block => decodeBlock(block).block);
```
ishantiw, line 165: i think here you  should check whether blocks.length >0 , else you should applyPenalty 100 and this._network.applyPenaltyOnPeer({ peerId, penalty: 100 and call syncBlocks instead of return

framework/src/engine/legacy/legacy_chain_handler.ts
ishantiw: so in general this code block should be like,
prefer this being called and applied penalty separately,

- check blocks.length > 0, apply penalty, log that penalty was applied to a peer and call sync() again
- if the above check passes, call _applyValidation, log that penalty was applied to a peer, if fails then apply penalty and call sync() again

ishantiw (changes requested):

framework/src/engine/legacy/legacy_chain_handler.ts
```diff
@@ -15,6 +15,7 @@
 import { codec } from '@liskhq/lisk-codec';
 import { P2PRequestPacket } from '@liskhq/lisk-p2p/dist-node/types';
 import { Database, NotFoundError } from '@liskhq/lisk-db';
+import { INVALID_PEER_INFO_PENALTY } from '@liskhq/lisk-p2p/dist-node/constants';
```
ishantiw, line 18: wrong import and maybe we should simply use 100 penalty or declare a constant within legacy to use

ishantiw (changes requested):
Just one point otherwise lgtm

ishantiw (approved):
Great job @sitetester 👏

framework/src/engine/legacy/legacy_chain_handler.ts
```diff
@@ -12,56 +12,215 @@
...
+			await this.syncBlocks(bracket, legacyBlock);
+		};
+
+		try {
+			// this part is needed to make sure `data` returns ONLY `{ blocks: Buffer[] }` & not any extra field(s)
+			const { blocks } = codec.decode<{ blocks: Buffer[] }>(
+				getBlocksFromIdResponseSchema,
+				data as Buffer,
+			);
+			if (!(blocks.length > 0)) {
```
mosmartin, line 168: For this ` if (!(blocks.length > 0)) ` piece of logic, would it be better to use the opposite operator (<=) `if (blocks.length <= 0)` instead? Not a show stopper by any means, just a suggestion.

framework/src/engine/legacy/legacy_chain_handler.ts
```diff
@@ -12,56 +12,215 @@
...
+				await applyPenaltyAndRepeat();
+			}
+
+			this._applyValidation(blocks);
+
+			legacyBlocks = blocks.map(block => decodeBlock(block).block);
+			if (!(legacyBlocks.length > 0)) {
```
mosmartin, line 175: For this `if (!(legacyBlocks.length > 0))` piece of logic, would it be better to use the opposite operator (<=) `if (legacyBlocks.length <= 0)` instead? Not a show stopper by any means, just a suggestion.

PR #7544, opened 2022-09-21
 Manage node to sync legacy blocks in the start of the node - Closes #7503 

### LINKED ISSUE #7503: Manage node to sync legacy blocks in the start of the node
### Description

Manage node to sync legacy blocks in the start of the node (can be disabled with a flag within legacy). Disable block processing (maybe a flag on receiving a block from the network to ignore when we are syncing legacy blocks)

- In case the `legacy.sync` flag from the config is false then simply skip the below step
- We can call `legacyChainHandler.sync` and await after network.start in the `engine.start` so we don't have to manage sync flag and emitting event for syncing legacy blocks
- ~~Pass `legacyChainHandler` to consensus so that when consensus calls `network.applyNodeInfo` it can pass current legacy blocks~~ Since we can apply partial nodeInfo we don't have this requirement

### Acceptance Criteria

- Should have unit tests covering the solution

Incede:
### What was the problem?

This PR resolves #7503 

### How was it solved?

Implement as specified

### How was it tested?

Added unit test

### DIFF
--- a/framework/src/engine/consensus/consensus.ts
+++ b/framework/src/engine/consensus/consensus.ts
@@ -37,6 +37,7 @@ import { AbortError, ApplyPenaltyAndRestartError, RestartError } from './synchro
 import { BlockExecutor } from './synchronizer/type';
 import { Network } from '../network';
 import { NetworkEndpoint, EndpointArgs } from './network_endpoint';
+import { LegacyNetworkEndpoint } from '../legacy/network_endpoint';
 import { EventPostBlockData, postBlockEventSchema } from './schema';
 import {
 	CONSENSUS_EVENT_BLOCK_BROADCAST,
@@ -52,6 +53,7 @@ import {
 	NETWORK_RPC_GET_BLOCKS_FROM_ID,
 	NETWORK_RPC_GET_HIGHEST_COMMON_BLOCK,
 	NETWORK_RPC_GET_LAST_BLOCK,
+	NETWORK_LEGACY_GET_BLOCKS_FROM_ID,
 } from './constants';
 import { GenesisConfig } from '../../types';
 import { AggregateCommit } from './types';
@@ -74,6 +76,7 @@ interface InitArgs {
 	logger: Logger;
 	genesisBlock: Block;
 	db: Database;
+	legacyDB: Database;
 }
 
 interface ExecuteOptions {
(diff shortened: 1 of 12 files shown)


framework/src/engine/engine.ts
```diff
@@ -104,6 +104,9 @@ export class Engine {
 	public async start() {
 		await this._init();
 		await this._network.start();
+		if (this._config.legacy.sync) {
+			await this._legacyChainHandler.syncBlocks();
```
ishantiw, line 108: maybe add a log.info statement before after this call as the node will be busy with syncing for a while.

framework/src/engine/consensus/consensus.ts
```diff
@@ -61,13 +61,15 @@ import { ValidatorInfo } from './certificate_generation/types';
 import { BFTModule } from '../bft';
 import { ABI, TransactionExecutionResult, TransactionVerifyResult } from '../../abi';
 import { isEmptyConsensusUpdate } from './utils';
+import { LegacyChainHandler } from '../legacy/legacy_chain_handler';
```
ishantiw, line 64: i think now we don't need LegacyChainHandler for anything in consensus

framework/test/unit/engine/engine.spec.ts
```diff
@@ -102,6 +105,35 @@ describe('engine', () => {
 			expect(Network.prototype.start).toHaveBeenCalledTimes(1);
 		});
 
+		it('should initialize legacy chain handler but not start syncing blocks if config is set to false', () => {
+			expect(LegacyChainHandler.prototype.init).toHaveBeenCalledTimes(1);
+			expect(LegacyChainHandler.prototype.syncBlocks).toHaveBeenCalledTimes(0);
+		});
+
+		it('should initialize legacy chain handler an start syncing blocks if config is set to true', async () => {
```
ishantiw, line 113:
```suggestion
		it('should initialize legacy chain handler and start syncing blocks if config is set to true', async () => {
```

ishantiw (commented):
LGTM, a few comments

framework/src/engine/consensus/consensus.ts
Incede: I think it was mentioned in the isse `Pass legacyChainHandler to consensus so that when consensus calls network.applyNodeInfo it can pass current legacy blocks`

ishantiw: I just updated the issue by removing this requirement - Since we can apply partial nodeInfo we don't have this requirement

Incede: ok, I already updated the PR to remove it.

ishantiw (approved):
🚀

Phanco (approved):

PR #7549, opened 2022-09-22
Update base event - Closes #7546

### LINKED ISSUE #7546: Update BaseEvent to allow empty data event
### Description

Currently the `BaseEvent` class needs a mandatory data & schema field and emitting event with empty data is not possible. It should be updated to allow to emit events with empty data field. 

### Motivation

With some of the new events in interoperability empty data is emitted to tackle cases where it crosses the maximum 1024 bytes limit. 

### Additional Information
Under `mainchain_registration.ts` function `execute` should be updated to use `InvalidRegistrationSignatureEvent` extended from `BaseEvent` instead of directly calling `eventQueue.add`. Corresponding test(s) should also be updated.

shuse2:
### What was the problem?

This PR resolves #7546

### How was it solved?

- Update base event to allow empty schema
- Update mainchain registration event to use event class

### How was it tested?

- No test is touched since it's refactoring

### DIFF
--- a/framework/src/modules/base_event.ts
+++ b/framework/src/modules/base_event.ts
@@ -19,9 +19,9 @@ export interface EventQueuer {
 }
 
 export abstract class BaseEvent<T> {
-	private readonly _moduleName: string;
+	public schema?: Schema;
 
-	public abstract schema: Schema;
+	private readonly _moduleName: string;
 
 	public get key(): Buffer {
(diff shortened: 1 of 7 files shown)


Incede (commented):
Looks good but I think test also needs to be updated.

Incede (approved):

Phanco (approved):

PR #7593, opened 2022-09-28
Update token transfer command - Closes #7573

### LINKED ISSUE #7573: Update token transfer command
### Description

Update token transfer command according to the [PR](https://github.com/LiskHQ/lips/blob/08935fbff959dd14f1fc0c2b1b64890f904d5eae/proposals/lip-0051.md#token-transfer)

- Config schema needs to be updated, and the default value should be as specified in the LIP

### Acceptance Criteria

- All test should pass
- New unit test to cover the added logic is required
- Unit test must be updated accordingly

### Additional Information
- This change requires #7571 and #7572
- It requires integration and functional tests update

has5aan:
### What was the problem?

This PR resolves #7573

### How was it solved?

Made changes to `execute` method to `initializeUserAccount` for `tokenID` and `recipientAddress` if not defined and `validate` method to fail if the sender's `availableBalance` is less than `amount` or USER_SUBSTORE_INITIALIZATION_FEE.

### How was it tested?

Implemented unit tests.

### DIFF
--- a/framework/src/modules/token/commands/transfer.ts
+++ b/framework/src/modules/token/commands/transfer.ts
@@ -11,38 +11,92 @@
  *
  * Removal or modification of this copyright notice is prohibited.
  */
+
+import * as cryptography from '@liskhq/lisk-cryptography';
 import { validator } from '@liskhq/lisk-validator';
+import { dataStructures } from '@liskhq/lisk-utils';
 import { BaseCommand } from '../../base_command';
 import {
 	CommandExecuteContext,
 	CommandVerifyContext,
+	MethodContext,
 	VerificationResult,
 	VerifyStatus,
 } from '../../../state_machine';
 import { TokenMethod } from '../method';
 import { transferParamsSchema } from '../schemas';
+import { UserStore } from '../stores/user';
+import { InitializeUserAccountEvent } from '../events/initialize_user_account';
+import { TokenID } from '../types';
+import { TransferEvent } from '../events/transfer';
 
 interface Params {
-	tokenID: Buffer;
+	tokenID: TokenID;
 	amount: bigint;
 	recipientAddress: Buffer;
 	data: string;
+	accountInitializationFee: bigint;
 }
 
 export class TransferCommand extends BaseCommand {
 	public schema = transferParamsSchema;
 	private _method!: TokenMethod;
+	private _accountInitializationFee!: bigint;
+	private _feeTokenID!: TokenID;
 
-	public init(args: { method: TokenMethod }) {
+	public init(args: {
+		method: TokenMethod;
+		feeTokenID: TokenID;
+		accountInitializationFee: bigint;
+	}) {
 		this._method = args.method;
+		this._feeTokenID = args.feeTokenID;
+		this._accountInitializationFee = args.accountInitializationFee;
 	}
 
-	// eslint-disable-next-line @typescript-eslint/require-await
 	public async verify(context: CommandVerifyContext<Params>): Promise<VerificationResult> {
 		const { params } = context;
 
 		try {
 			validator.validate(transferParamsSchema, params);
+
+			const userStore = this.stores.get(UserStore);
+
+			const balanceCheck = new dataStructures.BufferMap<bigint>();
+
+			const userAccountExists = await userStore.has(
+				context,
+				userStore.getKey(params.recipientAddress, params.tokenID),
+			);
+
+			if (!userAccountExists) {
+				if (params.accountInitializationFee !== this._accountInitializationFee) {
+					throw new Error('Invalid account initialization fee.');
+				}
+
+				balanceCheck.set(this._feeTokenID, this._accountInitializationFee);
+			}
+
+			balanceCheck.set(
+				params.tokenID,
+				(balanceCheck.get(params.tokenID) ?? BigInt(0)) + params.amount,
+			);
+
+			for (const [tokenID, amount] of balanceCheck.entries()) {
+				const availableBalance = await this._method.getAvailableBalance(
+					context.getMethodContext(),
+					context.transaction.senderAddress,
+					tokenID,
+				);
+
+				if (availableBalance < amount) {
+					throw new Error(
+						`${cryptography.address.getLisk32AddressFromAddress(
+							context.transaction.senderAddress,
+						)} balance ${availableBalance.toString()} is not sufficient for ${amount.toString()}.`,
+					);
+				}
+			}
 		} catch (err) {
 			return {
 				status: VerifyStatus.FAIL,
@@ -56,20 +110,70 @@ export class TransferCommand extends BaseCommand {
 
 	public async execute(context: CommandExecuteContext<Params>): Promise<void> {
 		const { params } = context;
-		// TODO: Fix in https://github.com/LiskHQ/lisk-sdk/issues/7573
-		await this._method.initializeUserAccount(
-			context.getMethodContext(),
-			params.recipientAddress,
-			params.tokenID,
-			context.transaction.senderAddress,
-			BigInt(50000000),
+
+		const userStore = this.stores.get(UserStore);
+
+		const recipientAccountKey = userStore.getKey(params.recipientAddress, params.tokenID);
+
+		const recipientAccountExists = await userStore.has(context, recipientAccountKey);
+
+		if (!recipientAccountExists) {
+			await this._intializeUserStore(
+				context,
+				params.recipientAddress,
+				params.tokenID,
+				context.transaction.senderAddress,
+			);
+		}
+
+		const senderAccountKey = userStore.getKey(context.transaction.senderAddress, params.tokenID);
+		const senderAccount = await userStore.get(context, senderAccountKey);
+		if (senderAccount.availableBalance < params.amount) {
+			throw new Error(
+				`${cryptography.address.getLisk32AddressFromAddress(
+					context.transaction.senderAddress,
+				)} balance ${senderAccount.availableBalance.toString()} is not sufficient for ${params.amount.toString()}.`,
+			);
+		}
+
+		senderAccount.availableBalance -= params.amount;
+		await userStore.save(context, context.transaction.senderAddress, params.tokenID, senderAccount);
+
+		const recipientAccount = await userStore.get(context, recipientAccountKey);
+		recipientAccount.availableBalance += params.amount;
+		await userStore.save(context, params.recipientAddress, params.tokenID, recipientAccount);
+
+		this.events.get(TransferEvent).log(context, {
+			senderAddress: context.transaction.senderAddress,
+			recipientAddress: params.recipientAddress,
+			tokenID: params.tokenID,
+			amount: params.amount,
+		});
+	}
+
+	private async _intializeUserStore(
+		methodContext: MethodContext,
+		address: Buffer,
+		tokenID: TokenID,
+		initPayingAddress: Buffer,
+	) {
+		const userStore = this.stores.get(UserStore);
+
+		await userStore.addAvailableBalanceWithCreate(methodContext, address, tokenID, BigInt(0));
+
+		await this._method.burn(
+			methodContext,
+			initPayingAddress,
+			this._feeTokenID,
+			this._accountInitializationFee,
 		);
-		await this._method.transfer(
-			context.getMethodContext(),
-			context.transaction.senderAddress,
-			params.recipientAddress,
-			params.tokenID,
-			params.amount,
-		);
+
+		const initializeUserAccountEvent = this.events.get(InitializeUserAccountEvent);
+		initializeUserAccountEvent.log(methodContext, {
+			address,
+			tokenID,
+			initPayingAddress,
+			initializationFee: this._accountInitializationFee,
+		});
 	}
 }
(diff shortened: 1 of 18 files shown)


framework/src/modules/token/commands/transfer.ts
```diff
@@ -43,6 +48,42 @@ export class TransferCommand extends BaseCommand {
...
+
+			const balanceChecks: { [key: string]: bigint } = {};
+
+			if (userStore.getKey(params.recipientAddress, tokenID).length === 0) {
+				if (params.accountInitializationFee !== USER_SUBSTORE_INITIALIZATION_FEE) {
+					throw new Error('Invalid account initialization fee');
+				}
+
+				balanceChecks[
+					TOKEN_ID_STORE_INITIALIZATION.toString('hex')
```
shuse2, line 64: We need to compute the mainchain token ID from it's own chain ID here because in LIP it mentiones the LSK token ID, but for testnet mainchain token ID is different from mainnet token ID.
 to do that, think we should create an util function
```
getMainchainTokenID(ownChainID: Buffer) =>
  Buffer.concat([ownChainID.slice(0, 1), Buffer.alloc(3 + 4, 0)])
```

framework/src/modules/token/commands/transfer.ts
```diff
@@ -43,6 +48,42 @@ export class TransferCommand extends BaseCommand {
...
+				] = USER_SUBSTORE_INITIALIZATION_FEE;
+			}
+
+			const tokenIDSerialized = tokenID.toString('hex');
```
shuse2, line 68:
```suggestion
			const hexTokenID = tokenID.toString('hex');
```
I think serialize means it's bytes, but in this case it's hex string

framework/src/modules/token/commands/transfer.ts
```diff
@@ -43,6 +48,42 @@ export class TransferCommand extends BaseCommand {
...
+				balanceChecks[tokenIDSerialized] += params.amount;
+			} else {
+				balanceChecks[tokenIDSerialized] = params.amount;
+			}
+
+			for (const tokenIDKey of Object.keys(balanceChecks)) {
+				const availableBalance = await this._method.getAvailableBalance(
+					context.getMethodContext(),
+					senderAddress,
+					Buffer.from(tokenIDKey),
```
shuse2, line 80: When you call `Buffer.from` default is `utf-8`, so you need to specify `hex` if you inserted as `hex`

framework/src/modules/token/commands/transfer.ts
```diff
@@ -56,6 +97,14 @@ export class TransferCommand extends BaseCommand {
...
 	public async execute(context: CommandExecuteContext<Params>): Promise<void> {
 		const { params } = context;
+
+		await this._intializeUserStore(
+			context,
+			params.recipientAddress,
+			params.tokenID,
+			context.transaction.senderAddress,
+		);
+
```
shuse2, line 107: Now `transferInternal` has bit different logic than `transfer`, maybe you can create an empty function `transferInternal` on `method.ts` with the same interface and call the function for now

shuse2 (changes requested):

mosmartin (approved):
🚀

framework/src/modules/token/commands/transfer.ts
```diff
@@ -12,37 +12,86 @@
...
+				balanceCheck.set(TOKEN_ID_STORE_INITIALIZATION, this._accountInitializationFee);
+			}
+
+			if (balanceCheck.has(params.tokenID)) {
+				const balance = balanceCheck.get(TOKEN_ID_STORE_INITIALIZATION)! + params.amount;
+
+				balanceCheck.set(TOKEN_ID_STORE_INITIALIZATION, balance);
+			} else {
+				balanceCheck.set(params.tokenID, params.amount);
+			}
```
shuse2, line 80:
```suggestion
			balanceCheck.set(params.tokenID, balanceCheck.get(params.tokenID) ?? BigInt(0) + params.amount);
```

framework/src/modules/token/commands/transfer.ts
```diff
@@ -12,37 +12,86 @@
...
+				const availableBalance = await this._method.getAvailableBalance(
+					context.getMethodContext(),
+					context.transaction.senderAddress,
+					this._method.getMainchainTokenID(),
+				);
+
+				if (availableBalance < tokenBalance) {
+					throw new Error('Insufficient balance.');
+				}
+			}
```
shuse2, line 94:
```suggestion
			for (const [tokenID, amount ] of balanceCheck.entries()) {
				const availableBalance = await this._method.getAvailableBalance(
					context.getMethodContext(),
					context.transaction.senderAddress,
					tokenID,
				);

				if (availableBalance < amount) {
					throw new Error(`${address.getLisk32AddressFromAddress(context.transaction.senderAddress)} balance ${availableBalance.toString()} is not sufficient for ${amount.toString()}`);
				}
			}
```

framework/src/modules/token/commands/transfer.ts
```diff
@@ -56,20 +105,63 @@ export class TransferCommand extends BaseCommand {
...
+
+		const senderAccountKey = userStore.getKey(context.transaction.senderAddress, params.tokenID);
+		const senderAccount = await userStore.get(context, senderAccountKey);
+		const recipientAccount = await userStore.get(context, recipientAccountKey);
+
+		senderAccount.availableBalance -= params.amount;
+		recipientAccount.availableBalance += params.amount;
+
+		await userStore.save(context, context.transaction.senderAddress, params.tokenID, senderAccount);
+		await userStore.save(context, params.recipientAddress, params.tokenID, recipientAccount);
```
shuse2, line 132: 1. Sender available balance should be checked against params.amount because it's possible that before this `execute` function, sender paid the fee and there is no remaining balance
2. sender should be updated and saved, then recipient should be fetched and deduct the balance because there is a case where sender and recipient are the same

framework/src/modules/token/constants.ts
```diff
@@ -37,6 +37,8 @@ export const TOKEN_ID_LENGTH = CHAIN_ID_LENGTH + LOCAL_ID_LENGTH;
 export const LOCAL_ID_LSK = Buffer.alloc(LOCAL_ID_LENGTH, 0);
 export const CHAIN_ID_LSK = Buffer.from([0, 0, 0, 0]);
 export const TOKEN_ID_LSK = Buffer.from([0, 0, 0, 0, 0, 0, 0, 0]);
+export const USER_SUBSTORE_INITIALIZATION_FEE = '5000000';
+export const TOKEN_ID_STORE_INITIALIZATION = TOKEN_ID_LSK;
```
shuse2, line 41: instead of this constant `config.feeTokenID` should be used

framework/src/modules/token/schemas.ts
```diff
@@ -46,6 +46,10 @@ export const configSchema = {
 				format: 'hex',
 			},
 		},
+		accountInitializationFee: {
+			type: 'string',
+			format: 'uint64',
+		},
```
shuse2, line 52:
```suggestion
		accountInitializationFee: {
			type: 'string',
			format: 'uint64',
		},
		feeTokenID: {
			type: 'string',
			format: 'hex',
			minLength: TOKEN_ID_LENGTH,
			maxLength: TOKEN_ID_LENGTH,
		},
```
I think i forgot to add the feeTokenID 🙏

framework/test/unit/modules/token/commands/transfer.spec.ts
```diff
@@ -160,277 +201,268 @@ describe.skip('Transfer command', () => {
...
-			await expect(
-				command.execute(context.createCommandExecuteContext(transferParamsSchema)),
-			).rejects.toThrow('balance 200000000 is not sufficient');
+			const result = await command.verify(context.createCommandVerifyContext(transferParamsSchema));
+			expect(result.status).toEqual(VerifyStatus.FAIL);
+			expect(result.error?.message).toInclude('Invalid account initialization fee.');
 		});
 
-		it('should resolve when recipient exist for different tokenID but does not have enough balance for minBalance', async () => {
+		it('should fail if token balance for tokenID TOKEN_ID_STORE_INITIALIZATION is less than the sum of configured initialzation fee and transaction amount', async () => {
```
shuse2, line 228:
```suggestion
		it('should fail if token balance for feeTokenID is less than the sum of configured initialzation fee and transaction amount', async () => {
```

framework/test/unit/modules/token/commands/transfer.spec.ts
```diff
@@ -160,277 +201,268 @@ describe.skip('Transfer command', () => {
...
+			const amount = BigInt(100000000);
+
+			command['_method'].getAvailableBalance = jest
+				.fn()
+				.mockResolvedValue(amount + BigInt(USER_SUBSTORE_INITIALIZATION_FEE) - BigInt(1));
```
shuse2, line 233:
```suggestion
			jest
				.spyOn(command['_method'], 'getAvailableBalance')
				.mockResolvedValue(amount + BigInt(USER_SUBSTORE_INITIALIZATION_FEE) - BigInt(1));
```
same for other cases

framework/src/modules/token/commands/transfer.ts
```diff
@@ -56,20 +105,63 @@ export class TransferCommand extends BaseCommand {
...
+		initPayingAddress: Buffer,
+	) {
+		const userStore = this.stores.get(UserStore);
+
+		await userStore.addAvailableBalanceWithCreate(methodContext, address, tokenID, BigInt(0));
+
+		await this._method.burn(
+			methodContext,
+			initPayingAddress,
+			TOKEN_ID_STORE_INITIALIZATION,
```
shuse2, line 155: this should be `feeTokenID`

framework/test/unit/modules/token/commands/transfer.spec.ts
```diff
@@ -160,277 +201,268 @@ describe.skip('Transfer command', () => {
...
-				localTokenID.slice(4),
-			);
-			expect(supply.totalSupply).toEqual(totalSupply - MIN_BALANCE);
+			const result = await command.verify(context.createCommandVerifyContext(transferParamsSchema));
+			expect(result.status).toEqual(VerifyStatus.OK);
 		});
+	});
 
-		it('should resolve and not burn supply when tokenID is not native and recipient does not exist', async () => {
+	describe('execute', () => {
```
shuse2, line 339: Please add cases
1. sender has enough balance for initialization fee but not for amount
2. sender and recipient are the same

shuse2 (changes requested):

framework/src/modules/token/commands/transfer.ts
```diff
@@ -12,37 +12,86 @@
...
+				context,
+				userStore.getKey(params.recipientAddress, params.tokenID),
+			);
+
+			if (!userAccountExists) {
+				if (params.accountInitializationFee !== this._accountInitializationFee) {
+					throw new Error('Invalid account initialization fee.');
+				}
+
+				balanceCheck.set(TOKEN_ID_STORE_INITIALIZATION, this._accountInitializationFee);
```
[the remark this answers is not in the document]
has5aan: We are also adding `feeTokenID` as part of `transferParamsSchema` and if I have understood your comment above correctly, `feeTokenID` is also injected by the module to the command, so should we also check if the `feeTokenID` received in transaction parameters is also the same as the `feeTokenID` configured by the module? 🤔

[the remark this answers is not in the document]
shuse2: maybe there is misunderstanding, we are not adding `feeTokenID` as part of transferParamsSchema 🤔

framework/src/modules/token/commands/transfer.ts
```diff
@@ -11,38 +11,96 @@
...
 	}
 
-	// eslint-disable-next-line @typescript-eslint/require-await
 	public async verify(context: CommandVerifyContext<Params>): Promise<VerificationResult> {
 		const { params } = context;
 
 		try {
 			validator.validate(transferParamsSchema, params);
+
+			if (context.transaction.senderAddress.equals(params.recipientAddress)) {
```
shuse2, line 63: It should be allowed to send transfer to  itself, also i think this is not defined in LIP?

framework/src/modules/token/commands/transfer.ts
```diff
@@ -56,20 +114,71 @@ export class TransferCommand extends BaseCommand {
...
+			);
+		}
+
+		const recipientAccount = await userStore.get(context, recipientAccountKey);
+
+		senderAccount.availableBalance -= params.amount;
+		recipientAccount.availableBalance += params.amount;
+
+		await userStore.save(context, context.transaction.senderAddress, params.tokenID, senderAccount);
+		await userStore.save(context, params.recipientAddress, params.tokenID, recipientAccount);
```
shuse2, line 149: with the verification fix, this needs to be handled in order
```
senderAccount.availableBalance -= params.amount;
await userStore.save(context, context.transaction.senderAddress, params.tokenID, senderAccount);
```
Specifically, above code needs to happen before getting the recipientAccount

framework/test/unit/modules/token/commands/transfer.spec.ts
```diff
@@ -93,6 +131,35 @@ describe.skip('Transfer command', () => {
 			expect(result.error?.message).toInclude(".tokenID' minLength not satisfied");
 		});
 
+		it('should fail if sender and recipient are the same', async () => {
```
shuse2, line 134: this should not fail, and handled as a use case

shuse2 (changes requested):

shuse2 (approved):

mosmartin (approved):
🥳

PR #7604, opened 2022-10-03
Fix to return empty array when calling getEvents - Closes #7598

### LINKED ISSUE #7598: RPC endpoint chain_getEvents fails to get events by height
### Expected behavior

`chain_getEvents` should return all the events emitted during block execution for a block height

### Actual behavior

Returns error when getting by height even if the event was emitted 

```ts
 "error": {
        "message": "Specified key 0900000076 does not exist",
        "code": -32600
    }
```

### Steps to reproduce

- Run dpos-mainchain example app
- Call chain_getEvents RPC for a specific height

- `6.0.0-alpha.3`

shuse2:
### What was the problem?

This PR resolves #7598

### How was it solved?

Update to return empty array when the event does not exist for the height

### How was it tested?
send transaction to dpos-mainchain example
```
./bin/run transaction:send 0a05746f6b656e12087472616e7366657218002080c2d72f2a20a3f96c50d0446220ef2f98240898515cbba8155730679ca35326d98dcfb680f032290a0810000000000000001080c2d72f1a143152da8f438d4591068553438c1a62edc2a161d4220227273a40cbefbdbeda4aa4d41e0a5f2887378ddec84ecf41f1227f029e442a82b3c29b56423c087354378368f719e55b1d8cc5e5ccc6aade8517d854c72799f2ca7cbd01
```
- call ` ./bin/run endpoint:invoke chain_getEvents --pretty '{"height": X }' ` for both height with event and height without event

### DIFF
--- a/elements/lisk-chain/src/data_access/storage.ts
+++ b/elements/lisk-chain/src/data_access/storage.ts
@@ -244,10 +244,17 @@ export class Storage {
 	}
 
 	public async getEvents(height: number): Promise<Event[]> {
-		const eventsByte = await this._db.get(concatDBKeys(DB_KEY_BLOCK_EVENTS, uint32BE(height)));
-		const events = decodeByteArray(eventsByte);
+		try {
+			const eventsByte = await this._db.get(concatDBKeys(DB_KEY_BLOCK_EVENTS, uint32BE(height)));
+			const events = decodeByteArray(eventsByte);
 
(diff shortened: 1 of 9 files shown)


has5aan (approved):
Non-existent block height results in `[ ]`.
Transaction for the concerned block height that included the transaction.

mitsuaki-u (approved):

mitsuaki-u (approved):

PR #7606, opened 2022-10-03
Update transaction and command verify step to have partial header - Closes #7603

### LINKED ISSUE #7603: Add header timestamp and height to `CommandVerifyContext`
### Description

In order to reject some transaction, add
```
header: {
  height: number;
  timestamp: number;
}
```
to the `CommandVerifyContext`

### Acceptance Criteria

- Check type and values are added to the context
- Add unit test

shuse2:
### What was the problem?

This PR resolves #7603

### How was it solved?

- Add header to `VerifyTransactionRequest` ABI call
- Update ABI handler to use header from context if exist, otherwise use from request similar to execute transaction
- Update caller of `abiClient.verifyTransaction` to have `header`. while calling with block, use the block otherwise last block header

### How was it tested?

- Check verifyTransaction and command.verify context has partial block header
- Update unit tests

### DIFF
--- a/framework/src/abi/schema.ts
+++ b/framework/src/abi/schema.ts
@@ -367,7 +367,7 @@ export const afterTransactionsExecuteResponseSchema = {
 export const verifyTransactionRequestSchema = {
 	$id: '/abi/verifyTransactionRequest',
 	type: 'object',
-	required: ['contextID', 'transaction'],
+	required: ['contextID', 'transaction', 'header'],
 	properties: {
 		contextID: {
 			fieldNumber: 1,
@@ -376,6 +376,10 @@ export const verifyTransactionRequestSchema = {
 		transaction: {
 			fieldNumber: 2,
 			...transactionSchema,
+		},
+		header: {
+			fieldNumber: 3,
+			...blockHeaderSchema,
 		},
 	},
 };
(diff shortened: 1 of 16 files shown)


framework/src/engine/consensus/consensus.ts
```diff
@@ -966,6 +966,7 @@ export class Consensus {
 				const { result: verifyResult } = await this._abi.verifyTransaction({
 					contextID,
 					transaction: transaction.toObject(),
+					header: block.header.toObject(),
```
ishantiw, line 969: I thought we wanted to pass partial header object,
```ts
{
  height: number;
  timestamp: number
}
```

framework/src/abi/schema.ts
```diff
@@ -373,6 +373,10 @@ export const verifyTransactionRequestSchema = {
 			fieldNumber: 2,
 			...transactionSchema,
 		},
+		header: {
```
ishantiw, line 376: with this the request will accept a full header object which we don't want to pass

framework/src/engine/consensus/consensus.ts
shuse2: https://github.com/LiskHQ/lisk-sdk/pull/7606/files#diff-0bac9c6e42245fd8bdb1b7378788978dc188e9c225a0d91c5b5f08e28dab2e9bR76

here it's passing partial block header to the module. For the ABI level, it has `context` as well, so it will be inconsistent to pass only those 2 properties.
Probably it's better to revisit the ABI schema, ill create an issue for it

ishantiw: yeah we can think about it when we revisit ABI schema 👍

ishantiw (approved):

mitsuaki-u (approved):

PR #7613, opened 2022-10-05
 Update token module interoperability methods - Closes #7577

### LINKED ISSUE #7577: Update token module interoperability methods
### Description

Update interoperability methods below as defined in the [PR](https://github.com/LiskHQ/lips/blob/08935fbff959dd14f1fc0c2b1b64890f904d5eae/proposals/lip-0051.md#beforecrosschaincommandexecution-1) in `cc_methods.ts`
- beforeCrossChainCommandExecution
- beforeCrossChainMessageForwarding
- verifyCrossChainMessage
- recover

Remove existing
- beforeSendCCM
- beforeRecoverCCM
- beforeApplyCCM

### Acceptance Criteria

- Tests for existing functions are removed
- New logics are covered by unit tests

### Additional Information
- This change requires #7571 and #7572

Incede:
### What was the problem?

This PR resolves #7577

### How was it solved?

Update as specified [here](https://github.com/LiskHQ/lips/blob/08935fbff959dd14f1fc0c2b1b64890f904d5eae/proposals/lip-0051.md#beforecrosschaincommandexecution-1) and based on recent updates in previous PR's

### How was it tested?

Added/updated unit tests


framework/src/modules/token/cc_method.ts
```diff
@@ -22,182 +22,287 @@ import {
...
-		// await userStore.addAvailableBalanceWithCreate(
-		// 	methodContext,
-		// 	ctx.trsSender,
-		// 	canonicalTokenID,
-		// 	ccm.fee,
-		// );
+	public async beforeCrossChainCommandExecution(ctx: BeforeApplyCCMsgMethodContext): Promise<void> {
+		const { trsSender, ccm } = ctx;
+		const methodContext = ctx.getMethodContext();
+		const relayerAddress = trsSender;
```
shuse2, line 67: `trsSender` is public key, so it needs to be converted to address

framework/src/modules/token/cc_method.ts
```diff
@@ -22,182 +22,287 @@ import {
...
+		const { id: ownChainID } = await this._interopMethod.getOwnChainAccount(methodContext);
+		const { messageFeeTokenID } = await this._interopMethod.getChannel(
+			methodContext,
+			ccm.sendingChainID,
+		);
```
shuse2, line 72: this should use `getMessageFeeTokenID`

framework/src/modules/token/cc_method.ts
```diff
@@ -22,182 +22,287 @@ import {
...
-		// 	methodContext,
-		// 	ctx.trsSender,
-		// 	canonicalTokenID,
-		// 	ccm.fee,
-		// );
+	public async beforeCrossChainCommandExecution(ctx: BeforeApplyCCMsgMethodContext): Promise<void> {
+		const { trsSender, ccm } = ctx;
+		const methodContext = ctx.getMethodContext();
+		const relayerAddress = trsSender;
+		const { id: ownChainID } = await this._interopMethod.getOwnChainAccount(methodContext);
```
shuse2, line 68: `ownChainID` can be passed from module, so you can inject from the constructor

framework/src/modules/token/cc_method.ts
```diff
@@ -22,182 +22,287 @@ import {
...
+		const { messageFeeTokenID } = await this._interopMethod.getChannel(
+			methodContext,
+			ccm.sendingChainID,
+		);
+		const { chainID: feeTokenChainID, localID: feeTokenLocalID } = messageFeeTokenID;
```
shuse2, line 73: you can use `const [chainID] = splitTokenID(tokenID);` after updating to `getMessageFeeTokenID`

framework/src/modules/token/cc_method.ts
```diff
@@ -22,182 +22,287 @@ import {
...
+				ccm.sendingChainID,
+				feeTokenLocalID,
+			);
+			if (escrowedAmount < ccm.fee) {
+				this.events.get(BeforeCCCExecutionEvent).error(
+					methodContext,
+					{
+						sendingChainID: ccm.sendingChainID,
+						receivingChainID: ccm.receivingChainID,
+						messageFeeTokenID: feeTokenLocalID,
```
shuse2, line 88: this should pass the whole `messageFeeTokenID` from above

framework/src/modules/token/cc_method.ts
```diff
@@ -22,182 +22,287 @@ import {
...
+				ccm.sendingChainID,
+				feeTokenLocalID,
+				ccm.fee,
+			);
+		}
+
+		await userStore.addAvailableBalanceWithCreate(
+			methodContext,
+			ctx.trsSender,
+			feeTokenLocalID,
```
shuse2, line 111:
```suggestion
			messageFeeTokenID,
```

framework/src/modules/token/cc_method.ts
```diff
@@ -22,182 +22,287 @@ import {
...
+				this._interopMethod,
+				ccm.sendingChainID,
+				feeTokenLocalID,
+				ccm.fee,
+			);
+		}
+
+		await userStore.addAvailableBalanceWithCreate(
+			methodContext,
+			ctx.trsSender,
```
shuse2, line 110: here also, it should be address

framework/src/modules/token/cc_method.ts
```diff
@@ -22,182 +22,287 @@ import {
...
+						relayerAddress,
+					},
+					TokenEventResult.FAIL_INSUFFICIENT_BALANCE,
+				);
+
+				throw new Error('Insufficient balance in the sending chain for the message fee.');
+			}
+
+			const escrowStore = this.stores.get(EscrowStore);
+			await escrowStore.deductEscrowAmountWithTerminate(
```
shuse2, line 99: here no need to terminate, i think it's only deducting the balance.
if we fetch the escrow account above when checking, you mutate the fetched value and set here. In this way, there will be one less encode/decode

shuse2: you could add additional function in escrow store to handle this case FYI

framework/src/modules/token/cc_method.ts
```diff
@@ -22,182 +22,287 @@ import {
...
-		const account = codec.decode<UserStoreData>(userStoreSchema, ctx.storeValue);
-		const address = ctx.storeKey.slice(0, ADDRESS_LENGTH);
+
 		const chainID = ctx.storeKey.slice(ADDRESS_LENGTH, ADDRESS_LENGTH + CHAIN_ID_LENGTH);
+		const tokenID = ctx.storeKey.slice(ADDRESS_LENGTH, ADDRESS_LENGTH + TOKEN_ID_LENGTH);
 		const localID = ctx.storeKey.slice(ADDRESS_LENGTH + CHAIN_ID_LENGTH);
 		const totalAmount =
-			account.availableBalance +
-			account.lockedBalances.reduce((prev, curr) => prev + curr.amount, BigInt(0));
+			account!.availableBalance +
```
shuse2, line 270: I think we sould avoid `!`

framework/src/modules/token/cc_method.ts
```diff
@@ -22,182 +22,287 @@ import {
...
+
+const CHAIN_ID_ALIAS_NATIVE = Buffer.alloc(0); // To be removed
 
 export class TokenInteroperableMethod extends BaseInteroperableMethod {
+	private readonly _tokenMethod: TokenMethod;
+
 	private _interopMethod!: InteroperabilityMethod;
 
-	public constructor(stores: NamedRegistry, events: NamedRegistry, _tokenMethod: TokenMethod) {
+	public constructor(stores: NamedRegistry, events: NamedRegistry, tokenMethod: TokenMethod) {
```
shuse2, line 55: It seems like most of the tokenMethod usage below can be replaced by stores

framework/src/modules/token/cc_method.ts
```diff
@@ -13,191 +13,270 @@
...
+				);
+
+				throw new Error('Insufficient balance in the sending chain for the message fee.');
+			}
+
+			escrowAccount.amount -= ccm.fee;
+			await escrowStore.set(methodContext, escrowKey, escrowAccount);
+		}
+
+		await userStore.addAvailableBalanceWithCreate(methodContext, relayerAddress, tokenID, ccm.fee);
```
shuse2, line 97: I think with the update, you can use `addAvailableBalance` in general because account always needs to be initialized separately now, so it's ok to fail if the account doesn't exist. Same for below

shuse2 (commented):
Some minor comments but overall looks good 🚀

framework/src/modules/token/cc_method.ts
```diff
@@ -13,191 +13,270 @@
...
+						messageFeeTokenID: tokenID,
+						messageFee: ccm.fee,
+					},
+					TokenEventResult.INSUFFICIENT_ESCROW_BALANCE,
+				);
+
+				throw new Error('Insufficient balance in the sending chain for the transfer.');
+			}
+
+			const updatedEscrowAccount = await escrowStore.get(methodContext, escrowKey);
```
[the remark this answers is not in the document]
Incede: yep

framework/src/modules/token/cc_method.ts
```diff
@@ -13,191 +13,266 @@
...
+						messageFeeTokenID: tokenID,
+						messageFee: ccm.fee,
+					},
+					TokenEventResult.INSUFFICIENT_ESCROW_BALANCE,
+				);
+
+				throw new Error('Insufficient balance in the sending chain for the transfer.');
+			}
+
+			const updatedEscrowAccount = await escrowStore.getOrDefault(methodContext, escrowKey);
```
shuse2, line 166: i think if you are getting the updated version here, you could use `get` and on the above comparison, you need to use the updated version as well?

framework/test/unit/modules/token/cc_method.spec.ts
```diff
@@ -69,104 +80,69 @@ describe('CrossChain Forward command', () => {
...
+			Buffer.concat([Buffer.from([3, 0, 0, 0]), defaultTokenID]),
 			{ amount: defaultEscrowAmount },
 		);
 	});
 
-	// TODO: Update with https://github.com/LiskHQ/lisk-sdk/issues/7577
-	describe.skip('beforeApplyCCM', () => {
-		it('should reject if fee is negative', async () => {
+	describe('beforeCrossChainCommandExecution', () => {
+		it('should credit fee to transaction sender if token id is not native', async () => {
```
shuse2, line 140: even if token ID is native, it is crediting to transaction sender / relayer?

framework/test/unit/modules/token/cc_method.spec.ts
```diff
@@ -271,15 +266,19 @@ describe('CrossChain Forward command', () => {
...
+			checkEventResult(
+				methodContext.eventQueue,
+				BeforeCCMForwardingEvent,
+				TokenEventResult.FAIL_INSUFFICIENT_BALANCE,
+			);
 		});
 
-		it('should credit fee to transaction sender if message fee token id is not native', async () => {
-			interopMethod.getChannel.mockResolvedValue({ messageFeeTokenID: defaultForeignTokenID });
+		it('should deduct escrow account for fee and credit to sender if ccm command is forward', async () => {
```
shuse2, line 279: it should credit to escrow of the receiving chain?

framework/test/unit/modules/token/cc_method.spec.ts
```diff
@@ -288,90 +287,134 @@ describe('CrossChain Forward command', () => {
...
+			).rejects.toThrow('Insufficient balance in the sending chain for the transfer.');
+			checkEventResult(
+				methodContext.eventQueue,
+				BeforeCCMForwardingEvent,
+				TokenEventResult.INSUFFICIENT_ESCROW_BALANCE,
+			);
 		});
 
-		it('should deduct escrow account for fee and credit to sender if token id is native', async () => {
+		it('should deduct escrow account for fee+ccm.params.amount and credit to sender if ccm command is forward', async () => {
```
shuse2, line 359: ccm command is transfer?

framework/test/unit/modules/token/cc_method.spec.ts
```diff
@@ -382,19 +425,19 @@ describe('CrossChain Forward command', () => {
 					logger: fakeLogger,
 					chainID: utils.getRandomBytes(32),
 				}),
-			).rejects.toThrow('Fee must be greater or equal to zero');
+			).resolves.toBeUndefined();
 		});
 
-		it('should credit receiving chain escrow account for fee if message token id is native', async () => {
+		it('should reject if token id is native and fee payer does not have sufficient balance', async () => {
```
shuse2, line 431:
```suggestion
		it('should reject if token id is native and sending chain escrow account does not have sufficient balance', async () => {
```

shuse2 (changes requested):

framework/test/unit/modules/token/cc_method.spec.ts
Incede: yes, I added this condition as a seperate test (number 3 in this scope) `should deduct escrow account for fee and credit to sender if token id is native` but I can also combine both if you prefer.

shuse2 (approved):

has5aan (approved):

PR #7616, opened 2022-10-06
Move Chainsafe BLST from optional peer dependency to required dependency

bobanm:
### What was the problem?

I could not complete SDK bootstrap using `yarn`. I was getting error:

```
Retrieving BLST native bindings...
Error: Error importing BLST native bindings: 404 Not Found
```

### How was it solved?

Moved `@chainsafe/blst` to be a regular dependency of `lisk-cryptography` package, instead of an optional dependecy.

I also bumped it to the latest version 0.2.6.

### How was it tested?

I tried to run tests with `yarn test`, but the process hung. Probably I still have to configure something on my side to run tests? 🤔

Submitting this PR will trigger tests to run in the CI pipeline, right?

### DIFF
--- a/elements/lisk-cryptography/package.json
+++ b/elements/lisk-cryptography/package.json
@@ -35,6 +35,7 @@
 		"build:check": "node -e \"require('./dist-node')\""
 	},
 	"dependencies": {
+		"@chainsafe/blst": "0.2.6",
 		"@liskhq/lisk-passphrase": "4.0.0-alpha.0",
 		"buffer-reverse": "1.0.1",
 		"ed2curve": "0.3.0",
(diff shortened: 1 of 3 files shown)


bobanm:
Hmm, seems like bump to BLST wrapper v0.2.6 did mess one test up. Investigating and will provide a solution...

shuse2:
@bobanm Actually, the blst library was intentionally added as peer dependency. This cryptography library is used for both browser and in node environment, and to support this, it is important that it's not always installed

ishantiw:
There is a duplicate PR https://github.com/LiskHQ/lisk-sdk/pull/7621 tackling the issue. So closing this one.

PR #7639, opened 2022-10-14
Search for common block should look until finalized height - Closes #6990 

### LINKED ISSUE #6990: Search for common block should look until finalized height
### Description

While searching for the common block while block synchronization, it look 10 rounds for 3 times, but it should look until finalized height.

### Motivation

While synchronization, we look into 10 rounds of blocks for 3 times but it should be capped at the finalized height.

### Additional Information
- it is implemented here https://github.com/LiskHQ/lisk-sdk/blob/development/framework/src/engine/consensus/synchronizer/block_synchronization_mechanism.ts#L350

Incede:
### What was the problem?

This PR resolves #6990

### How was it solved?

Update common block search logic

### How was it tested?

Update unit test


framework/src/engine/consensus/synchronizer/block_synchronization_mechanism.ts
```diff
@@ -381,14 +375,12 @@ export class BlockSynchronizationMechanism extends BaseSynchronizer {
 					blockHeaders.map(block => block.id),
 				);
 			} catch (e) {
-				numberOfRequests += 1;
+				currentRound -= blocksPerRequestLimit;
+				currentHeight = currentRound * validators.length;
```
shuse2, line 379: why is it updating the current height in the error case?
Also what happen if `_getHighestCommonBlockFromNetwork` fails with network reason?

shuse2 (approved):
🚀

has5aan (approved):

PR #7640, opened 2022-10-14
Add an auth module endpoint which sorts a multisignature group - Closes #7611

### LINKED ISSUE #7611: Expose auth endpoint to sort multisignature registration signatures array
### Description

Allow Lisk wallets and community tools in general to sort the signatures of multisignature registration according to Lisk Protocol.

Endpoint should be `auth_sortMultisignatureGroup`

req
```
{
  mandatory: { publicKey: string, signature: string }[],
  optional: { publicKey: string, signature: string }[],
}
```
res
```
{
  mandatoryKeys: string[],
  optionalKeys: string[],
  signatures: string[],
}
```

- public key needs to be sorted lexicographically and signature must match the public key
- If signature is empty, we need to insert empty string in the result `signatures`

### Acceptance Criteria

- Should expose an endpoint from `auth` module to sort signatures inside params to sort according to Lisk Protocol
- Should position the signatures of members at the correct position

### Additional Information
- Sorting is verified in `framework/src/modules/auth/commands/register_multisignature.ts` verify step
- Sorting is also implemented in `elements/lisk-transactions/src/sign.ts`

bobanm:
### What was the problem?

This PR resolves #7611

### How was it solved?

Added a new endpoint in `auth` module which takes keys and signatures as request and returns an object with sorted keys and signatures.

### How was it tested?

Unit test provided. I haven't done any integration testing.


framework/test/unit/modules/auth/endpoint.spec.ts
```diff
@@ -487,4 +487,52 @@ describe('AuthEndpoint', () => {
...
+					},
+					{
+						publicKey: '0000000000000000000000000000000000000000000000000000000000000000',
+						signature:
+							'11111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111111',
+					},
+					{
+						publicKey: 'bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb',
+						signature:
+							'00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000',
```
shuse2, line 509: I think it would be nice to have empty signature case here as well even though it will be handled in the implementation correctly

mosmartin (commented):
Aside from the open comments, it looks 👍.

framework/test/unit/modules/auth/endpoint.spec.ts
bobanm: Scroll only 6 lines down to the line 515. There is the empty signature test data already 🙂

It doesn't make any difference at all if signature is empty or not. The sorting logic uses only `publicKey` property. It will correctly assign the relevant value of `signature` property to the response array, regardless of if the signature is empty or not.

shuse2 (approved):

framework/test/unit/modules/auth/endpoint.spec.ts
shuse2: ah yea true the test was there! my bad =)

mosmartin (approved):
👏

PR #7645, opened 2022-10-17
Align Commander `node` topic name to `system` namespace - Closes #7635

### LINKED ISSUE #7635: Align Commander `node` topic name to `system` namespace
### Expected behavior

Commands in `system` namespace should be accessed using `system` topic.

### Actual behavior

`node` namespace was renamed to `system`. However, Lisk Commander still uses `node` topic to access endpoints in `system` namespace.

### Steps to reproduce

`./bin/run` will show `node` topic instead of `system`

bobanm:
### What was the problem?

This PR resolves #7635 

### How was it solved?

Renamed folders and files

### How was it tested?

The following commands run correctly when invoked from `dpos-mainchain` example:

```bash
./bin/run system:node-info --pretty
./bin/run system:metadata --pretty
```

Also, simply running `./bin/run` without any arguments shows correct topic name.


shuse2 (approved):

mosmartin (approved):
LGTM 🚀

PR #7653, opened 2022-10-18
 Fix blockchain:import / blockchain:export - Closes #7570 

### LINKED ISSUE #7570: Fix blockchain:import / blockchain:export
### Description

blockchain:import and blockchain:export should pack all the necessary on-chain information.
```
- blockchain.db
- state.db
```

- compress/decompress are already done in the existing command, so it should follow the same way

### Motivation

All on-chain information must be able to backup and restore with the command

### Additional Information
Currently, export is compressing the `blockchain.db` from data path and outputs tar ball and import is decompressing the tar ball and outputting to the data path.

Incede:
### What was the problem?

This PR resolves #7570 

### How was it solved?

Update as specified

### How was it tested?

- Updated unit tests
- Tested manually

### DIFF
--- a/commander/src/bootstrapping/commands/blockchain/export.ts
+++ b/commander/src/bootstrapping/commands/blockchain/export.ts
@@ -17,7 +17,7 @@ import { Command } from '@oclif/core';
 import { join } from 'path';
 import * as tar from 'tar';
 import { flagsWithParser } from '../../../utils/flags';
-import { getBlockchainDBPath, getDefaultPath, getFullPath } from '../../../utils/path';
+import { getDefaultPath, getFullPath } from '../../../utils/path';
 
 export class ExportCommand extends Command {
 	static description = 'Export to <FILE>.';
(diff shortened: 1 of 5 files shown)


commander/src/bootstrapping/commands/blockchain/export.ts
```diff
@@ -37,19 +37,19 @@ export class ExportCommand extends Command {
...
 			? flags['data-path']
 			: getDefaultPath(this.config.pjson.name);
-		const blockchainPath = getBlockchainDBPath(dataPath);
+		const inputPath = join(dataPath, 'data');
 		const exportPath = flags.output ? flags.output : process.cwd();
 
 		this.log('Exporting blockchain:');
-		this.log(`   ${getFullPath(blockchainPath)}`);
+		this.log(`   ${getFullPath(inputPath)}`);
 		const filePath = join(exportPath, 'blockchain.db.tar.gz');
```
shuse2, line 45: Since now it contains multiple db, maybe it's better to name it `backup.tar.gz` or something similar.
Also, if the filePath doesn't exist it fails, so better to create folder if it doesn't exist

shuse2 (changes requested):

commander/src/bootstrapping/commands/blockchain/export.ts
Incede: I think the name of the file should be aligned with the logs `Exporting blockchain` and the command name itself so I think it's okay to not change or otherwise rename to `blockchain.tar.gz` whichever you prefer.

What do you mean it fails ? If no `--output-path` flag is provided it outputs to `process.cwd()`.

shuse2: Make sense, then I think `blockchain.tar.gz` as you mentioned would be better

it fails when `--output-path` is specified and the target folder doesn't exist.

commander/test/bootstrapping/commands/blockchain/import.spec.ts
```diff
@@ -25,6 +25,8 @@ import { Awaited } from '../../../types';
 describe('blockchain:import', () => {
 	const defaultDataPath = path.join(homedir(), '.lisk', 'lisk-core');
 	const defaultBlockchainDBPath = getBlockchainDBPath(defaultDataPath);
+	const defaultStateDBPath = getStateDBPath(defaultDataPath);
+	const defaultOutputPath = path.join(defaultDataPath, 'data');
 	const pathToBlockchainGzip = '/path/to/blockchain.db.tar.gz';
```
bobanm, line 30:
```suggestion
	const pathToBlockchainGzip = '/path/to/blockchain.tar.gz';
```

bobanm: Also update all the asserts to check for `blockchain.tar.gz`

shuse2 (approved):

bobanm (approved):

PR #7654, opened 2022-10-18
Update schema and validateFormat function

Phanco:
### What was the problem?

This PR resolves #INSERT_ISSUE_NUMBER


Phanco:
This PR is replaced by #7655

PR #7675, opened 2022-10-20
Update `createTerminatedState` in interoperability module - Closes #7559 

### LINKED ISSUE #7559: Update `createTerminatedStateAccount` of baseInteroperabilityStore
### Description

The `createTerminatedStateAccount` function creates an entry in the terminated state substore.

```python
def createTerminatedStateAccount(chainID: ChainID, stateRoot: bytes = EMPTY_BYTES) -> None:    
    if chainAccount(chainID) exists:
        chainAccount(chainID).status = CHAIN_STATUS_TERMINATED
        remove the entry with storeKey = chainID from the outbox root substore

        # If no stateRoot is given as input, get it from the state
        if len(stateRoot) == 0:
            stateRoot = chainAccount(chainID).lastCertificate.stateRoot

        terminatedState = {
            "stateRoot": stateRoot,
            "mainchainStateRoot": EMPTY_BYTES,
            "initialized": True
        }

        # Emit chain status updated event.
        emitEvent(
            module = MODULE_NAME_INTEROPERABILITY,
            name = EVENT_NAME_CHAIN_ACCOUNT_UPDATED,
            data = chainAccount(chainID),
            topics = [chainID]
        )

    # State root is not available, set it to empty bytes temporarily.
    # This should only happen on a sidechain.
    else:
        # Processing on the mainchain
        if ownChainAccount.chainID == CHAIN_ID_MAINCHAIN:
            # If the account does not exist on the mainchain, the input chainID is invalid.
            raise Exception('Chain to be terminated is not valid.')

        terminatedState = {
            "stateRoot": EMPTY_BYTES,
            "mainchainStateRoot": chainAccount(CHAIN_ID_MAINCHAIN).lastCertificate.stateRoot,
            "initialized": False
        }

    create an entry in the terminated state substore with
        storeKey = chainID
        storeValue = encode(terminatedStateAccountSchema, terminatedState)

    emitEvent(
        module = MODULE_NAME_INTEROPERABILITY,
        name = EVENT_NAME_TERMINATED_STATE_CREATED,
        data = terminatedState,
        topics = [chainID]
    )
``` 

### Acceptance Criteria

- Should pass its unit test

### Additional Information

- Related https://github.com/LiskHQ/lips/pull/148

mitsuaki-u:
### What was the problem?

This PR resolves #7559 

### How was it solved?

- Updated `createTerminatedState`

### How was it tested?

Added unit tests

### DIFF
--- a/framework/src/modules/interoperability/base_interoperability_store.ts
+++ b/framework/src/modules/interoperability/base_interoperability_store.ts
@@ -38,6 +38,7 @@ import {
 	OwnChainAccount,
 	CCMApplyContext,
 	TerminateChainContext,
+	CreateTerminatedStateAccountContext,
 } from './types';
 import { getCCMSize, getIDAsKeyForStore } from './utils';
 import {
@@ -55,6 +56,8 @@ import { OutboxRootStore } from './stores/outbox_root';
 import { TerminatedStateAccount, TerminatedStateStore } from './stores/terminated_state';
 import { ChainAccountStore } from './stores/chain_account';
 import { TerminatedOutboxAccount, TerminatedOutboxStore } from './stores/terminated_outbox';
+import { ChainAccountUpdatedEvent } from './events/chain_account_updated';
+import { TerminatedStateCreatedEvent } from './events/terminated_state_created';
 
 export abstract class BaseInteroperabilityStore {
 	public readonly events: NamedRegistry;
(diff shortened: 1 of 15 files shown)


framework/src/modules/interoperability/base_interoperability_store.ts
```diff
@@ -241,65 +243,69 @@ export abstract class BaseInteroperabilityStore {
...
 			terminatedState = {
 				stateRoot: EMPTY_BYTES,
-				mainchainStateRoot: chainAccount.lastCertificate.stateRoot,
+				mainchainStateRoot: mainchainAccount.lastCertificate.stateRoot,
 				initialized: false,
 			};
 		}
 
 		const terminatedStateSubstore = this.stores.get(TerminatedStateStore);
 		await terminatedStateSubstore.set(this.context as StoreGetter, chainID, terminatedState);
```
ishantiw, line 294: I think you forgot to emit event here as mentioned in the LIP

framework/test/unit/modules/interoperability/store.spec.ts
```diff
@@ -246,24 +258,30 @@ describe('Base interoperability store', () => {
...
 				.mockResolvedValue(ownChainAccount1 as never);
 
 			await expect(
-				mainchainInteroperabilityStore.createTerminatedStateAccount(chainIdNew),
-			).resolves.toEqual(false);
+				mainchainInteroperabilityStore.createTerminatedStateAccount(
+					createTerminatedStateAccountContext,
+					chainIdNew,
+				),
+			).rejects.toThrow();
```
ishantiw, line 272: can we also add the message which is thrown here?

framework/test/unit/modules/interoperability/store.spec.ts
```diff
@@ -224,9 +225,17 @@ describe('Base interoperability store', () => {
 			nonce: BigInt('0'),
 		};
 
+		const createTerminatedStateAccountContext = {
+			eventQueue: new EventQueue(0),
+		};
+
 		it('should set appropriate terminated state for chain id in the terminatedState sub store if chain account exists for the id and state root is provided', async () => {
```
ishantiw, line 232: I think we need to also assert the emitted event `EVENT_NAME_CHAIN_ACCOUNT_UPDATED ` and also for `await terminatedStateSubstore.set` and event  `EVENT_NAME_TERMINATED_STATE_CREATED`

framework/test/unit/modules/interoperability/store.spec.ts
```diff
@@ -237,7 +246,10 @@ describe('Base interoperability store', () => {
 
 		it('should set appropriate terminated state for chain id in the terminatedState sub store if chain account exists for the id but state root is not provided', async () => {
```
ishantiw, line 247: missing assertion for emitted event and `await terminatedStateSubstore.set` similar to above comment and also applies for the following test cases

ishantiw (changes requested):

framework/src/modules/interoperability/base_interoperability_store.ts
```diff
@@ -241,65 +244,72 @@ export abstract class BaseInteroperabilityStore {
...
-				await outboxRootSubstore.del(this.context as StoreGetter, chainID);
-			}
-			terminatedState = {
-				stateRoot,
-				mainchainStateRoot: EMPTY_BYTES,
-				initialized: true,
-			};
-		} else if (isExist) {
+		const chainSubstore = this.stores.get(ChainAccountStore);
+		const chainAccountExists = await this.chainAccountExist(chainID);
```
ishantiw, line 255: you can use `chainSubstore.has` instead of using store's chainAccountExist function as we want to remove it in the coming PR

framework/src/modules/interoperability/base_interoperability_store.ts
```diff
@@ -241,65 +244,72 @@ export abstract class BaseInteroperabilityStore {
 		return terminatedOutboxSubstore.get(this.context, chainID);
 	}
 
-	public async createTerminatedStateAccount(chainID: Buffer, stateRoot?: Buffer): Promise<boolean> {
-		const chainSubstore = this.stores.get(ChainAccountStore);
-		const isExist = await this.chainAccountExist(chainID);
+	public async createTerminatedStateAccount(
+		context: CreateTerminatedStateAccountContext,
```
ishantiw, line 248: why adding context here? we can use `this.context` instead within the methods of store class

framework/src/modules/interoperability/base_interoperability_store.ts
```diff
@@ -241,65 +244,72 @@ export abstract class BaseInteroperabilityStore {
...
+			this.events
+				.get(ChainAccountUpdatedEvent)
+				.log({ eventQueue: context.eventQueue }, chainID, chainAccount);
+		} else {
 			// Processing on the mainchain
 			const ownChainAccount = await this.getOwnChainAccount();
 			if (ownChainAccount.chainID.equals(getIDAsKeyForStore(MAINCHAIN_ID))) {
 				// If the account does not exist on the mainchain, the input chainID is invalid.
-				return false;
+				throw new Error('Chain to be terminated is not valid');
```
ishantiw, line 278:
```suggestion
			throw new Error('Chain to be terminated is not valid.');
```

Phanco:
Looks good, and I also agree with the part the `context` may not be necessary

Phanco (approved):

framework/src/modules/interoperability/base_interoperability_store.ts
mitsuaki-u: Because we need to pass in `eventQueue` to the events

ishantiw (approved):

PR #7677, opened 2022-10-20
Update ABI not to pass the consensus information

### LINKED ISSUE #7567: Update validators module to detach engine and application
### Description

- Update validators module to store generator list with schema with `ValidatorsParamsStore` where key is `EMPTY_BYTES`
Schema
```
export const validatorSchema = {
	type: 'object',
	required: ['address', 'bftWeight', 'generatorKey', 'blsKey'],
	properties: {
		address: {
			fieldNumber: 1,
			dataType: 'bytes',
			format: 'lisk32',
		},
		bftWeight: {
			fieldNumber: 2,
			dataType: 'uint64',
		},
		generatorKey: {
			fieldNumber: 3,
			dataType: 'bytes',
		},
		blsKey: {
			fieldNumber: 4,
			dataType: 'bytes',
		},
	},
};

export const validatorsParamsSchema = {
	type: 'object',
	required: ['validators', 'preCommitThreshold', 'certificateThreshold'],
	properties: {
		validators: {
			fieldNumber: 1,
			type: 'array',
			items: {
				...validatorSchema,
			},
		},
		preCommitThreshold: {
			fieldNumber: 2,
			dataType: 'boolean',
		},
		certificateThreshold: {
			fieldNumber: 4,
			dataType: 'uint64',
		},
	},
};
```
- Add `setValidatorsParams` method to validators module, and this method should insert the updated next validators to the context
```
interface NextValidatorsSetter {
setNextValidators: (
		preCommitThreshold: bigint,
		certificateThreshold: bigint,
		validators: Validator[],
	) => void;
}
setValidatorsParams(
        methodContext: MethodContext,
        nextValidatorsSetter: NextValidatorsSetter,
	preCommitThreshold: bigint,
	certificateThreshold: bigint,
	validators: Validator[],
) => void
```
- Update block header to include `impliesMaxPrevotes` as fieldNumber `12` and increment other properties later
- Update DPoS module to call `setValidatorsParams` from validators module instead of calling `context.setNextValidators`
- Remove `Consensus` type and schema from ABI
- Update to use `b.header. aggregateCommit.height` instead of `maxHeightCertified` in the application
- `impliesMaxPrevotes` usage from the context should be replaced by `b.header.impliesMaxPrevotes`

### Acceptance Criteria

- All tests including functional test should pass

### Additional Information
- Block fixtures needs to be updated to include `impliesMaxPrevotes`
- Genesis block for dpos-mainchain example needs to be regenerated

shuse2:
### What was the problem?

This PR resolves #7567 

### How was it solved?
Update engine and application based on https://github.com/LiskHQ/lips/pull/177

- Update validators module to store consensus information before passing to ABI
- Move `getSlotTime`, `getSlotNumber` and `getGeneratorAtTimestamp` to BFT method, and update its usages
- Update DPoS module to use validators module instead of replying directly through context
- Update block header to add impliesMaxPrevote
- Update consensus to verify impliesMaxPrevote
- Update generator to add impliesMaxPrevote
- Add genesis block validation for impliesMaxPrevote
- Update dpos module unlock to use aggregateCommit.height for checking certified

### How was it tested?

- Add more test case impliesMaxPrevote method
- Update the usage and test for the consensus params

### Additional Information
Currently genesis timestamp is not stored in BFT module https://github.com/LiskHQ/lips/pull/177#discussion_r1008477483


framework/src/engine/bft/method.ts
```diff
@@ -98,22 +98,42 @@ export class BFTMethod {
 		};
 	}
 
-	// This function expects to be called after blockBFTInfos has been inserted for the executiong block.
-	// blockBFTInfos must be inserted first because for the block genesis block + 1, bftVotes.blockBFTInfos will be empty
-	public async currentHeaderImpliesMaximalPrevotes(stateStore: StateStore): Promise<boolean> {
+	public async impliesMaximalPrevotes(
```
shuse2, line 101: @janhack does this function looks ok?
I needed to update the function so that it can be called before/after the bftVotes has been updated

janhack: I don't think this works. For instance, assume:
- header.height == 310
- header.maxHeightGenerated == 1
- MAX_LENGTH_BLOCK_BFT_INFOS == 309
Then, if you call `impliesMaximalPrevotes` before updating `bftVotes`, then you call check that bftVotes.blockBFTInfos[308].generatorAddress == header.generatorAddress` (this check could return True or False). However, if you do this check after updating `bftVotes`, then the previous block from the generator is not contained in `bftVotes.blockBFTInfos` any more.

janhack: We could change the protocol however so that `impliesMaximalPrevotes` only uses the last `MAX_LENGTH_BLOCK_BFT_INFOS-1` instead of `MAX_LENGTH_BLOCK_BFT_INFOS` blocks.

shuse2: ah that's a good point, blockBFTInfos will be truncated at the edge.
since we add the property to block header, we can control when to call this function precisely, i will remove the logic to handle both and ensure it will be called before updating the bftVotes

framework/src/engine/bft/method.ts
```diff
@@ -98,27 +98,40 @@ export class BFTMethod {
...
+		stateStore: StateStore,
+		header: { height: number; generatorAddress: Buffer; maxHeightGenerated: number },
+	): Promise<boolean> {
 		const votesStore = stateStore.getStore(MODULE_STORE_PREFIX_BFT, STORE_PREFIX_BFT_VOTES);
 		const bftVotes = await votesStore.getWithSchema<BFTVotes>(EMPTY_KEY, bftVotesSchema);
-		const [currentHeader] = bftVotes.blockBFTInfos;
-		const previousHeight = currentHeader.maxHeightGenerated;
+		// if blockBFTInfos is empty (ie: right after executing genesis block), next block never implies prevotes
+		if (bftVotes.blockBFTInfos.length === 0) {
```
janhack, line 108: This case is differently in the specifications https://github.com/LiskHQ/lips/pull/177/files. Note that the first block after the genesis block can imply prevotes, namely for the block itself.

framework/src/engine/bft/method.ts
janhack: Okay, that sounds good. I resolve this comment then.

framework/src/engine/bft/method.ts
```diff
@@ -210,6 +195,14 @@ export class BFTMethod {
 			validatorsHash,
 		};
 
+		const votesStore = stateStore.getStore(MODULE_STORE_PREFIX_BFT, STORE_PREFIX_BFT_VOTES);
+		const bftVotes = await votesStore.getWithSchema<BFTVotes>(EMPTY_KEY, bftVotesSchema);
+		// This assumes bftVotes.blockBFTInfos will contain currently executing block
+		const nextHeight =
+			(bftVotes.blockBFTInfos.length > 0
+				? bftVotes.blockBFTInfos[0].height
```
Incede, line 203: I think this is different in specification, as per the LIP it should be `bftVotes.blockBFTInfos[0].height + 1`

framework/src/engine/bft/method.ts
```diff
@@ -235,39 +228,48 @@ export class BFTMethod {
...
+		return this.getSlotNumber(timestamp) === slot;
 	}
 
-	private _computeValidatorsHash(validators: BFTValidator[], certificateThreshold: bigint) {
-		const activeValidators: ValidatorsHashInfo[] = [];
+	private _computeValidatorsHash(validators: Validator[], certificateThreshold: bigint) {
+		const validatorsHashInfo: ValidatorsHashInfo[] = [];
 		for (const validator of validators) {
-			activeValidators.push({
+			if (validator.bftWeight <= BigInt(0)) {
```
Incede, line 262: I am not sure if this is correct/needed. From the LIP PR `The BFT weight of the validators is allowed to be 0.` and if we already check for negative weights before as per previous comment.

framework/src/engine/bft/utils.ts
```diff
@@ -82,59 +80,7 @@ export const sortValidatorsByAddress = (validators: { address: Buffer }[]) =>
...
-		return;
-	}
-	// Delete all BFT Parameters except the one of largest height which is at most the input height
-	for (let i = 0; i < results.length - 1; i += 1) {
-		await keysStore.del(results[i].key);
-	}
-};
-
-export const computeValidatorsHash = (validators: BFTValidator[], certificateThreshold: bigint) => {
+export const computeValidatorsHash = (validators: Validator[], certificateThreshold: bigint) => {
```
Incede, line 83: probably duplicate of `_computeValidatorsHash` in `method.ts`

framework/src/engine/consensus/fork_choice/fork_choice_rule.ts
```diff
@@ -24,13 +24,23 @@ export enum ForkStatus {
...
 	slots.getSlotNumber(block.timestamp);
 
 export const isBlockReceivedWithinForgingSlot = (
 	slots: Slots,
 	{ timestamp, receivedAt }: BFTHeader,
-): boolean => slots.isWithinTimeslot(slots.getSlotNumber(timestamp), receivedAt);
+): boolean =>
+	slots.isWithinTimeslot(
+		slots.getSlotNumber(timestamp),
+		receivedAt ?? Math.floor(Date.now() / 1000),
```
Incede, line 42: `Math.floor(Date.now() / 1000)`
Does the default value come from a LIP ?

Incede (changes requested):
Did not review the tests yet but maybe move the bft tests from module to engine

Incede (changes requested):
As per the current logic of 'setBFTParameters' the test 'should store BFT parameters with height latest blockBFTInfo + 1' should fail but it's passing, I am not sure why ?

framework/src/engine/consensus/fork_choice/fork_choice_rule.ts
shuse2: I think this was default value in the implementation from the past.

receivedAt is defined only when received from the P2P, and if not defined, it just treat as it is just received now

framework/src/engine/bft/method.ts
Incede: I am not sure how. I see the existing implementation (dev branch) also uses `bftVotes.blockBFTInfos[0].height + 1` for `nextHeight` similar to the LIP.

framework/src/engine/consensus/fork_choice/fork_choice_rule.ts
Incede: ok, probably the confusion was since there is no default value for `receivedAt` in the current implementation, do you know why ?

shuse2: it was internally handled in the `isWithinTimeslot` / Slot before, but in the LIP method, there is no optional so it's handled outside

framework/src/engine/bft/method.ts
shuse2: ah that's true, the current height is `bftVotes.blockBFTInfos[0].height` and `nextHeight` is `+1`

shuse2: Re-checking the code, the next height is
```
const nextHeight =
			(bftVotes.blockBFTInfos.length > 0
				? bftVotes.blockBFTInfos[0].height
				: bftVotes.maxHeightPrevoted) + 1;
```
it is adding `1` at the end?

Incede: ah, I see now there is a parenthesis in the end. my bad, please disregard this and other comment about the test.

framework/src/engine/bft/method.ts
```diff
@@ -153,9 +171,6 @@ export class BFTMethod {
 		}
 		let aggregateBFTWeight = BigInt(0);
 		for (const validator of validators) {
-			if (validator.bftWeight <= 0) {
```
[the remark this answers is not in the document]
Incede: would be nice to also have a unit test covering the updated logic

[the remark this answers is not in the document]
shuse2: added =)

Incede (approved):

has5aan (approved):

janhack (commented):
The function `impliesMaximalPrevotes` looks all good.

PR #7690, opened 2022-10-24
Adds events for dpos module - Closes #7662

### LINKED ISSUE #7662: Add new events to DPoS module
### Description

Implement all the new events for the DPoS module as in [LIP-0057](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#events)

### Acceptance Criteria
- All events are defined and name/properties are matching

### Additional Information
- use `error` for persistent event and else `log` for event class

has5aan:
### What was the problem?

This PR resolves #7662 

### How was it solved?

Defined events and their schema for;
- DelegateRegistered
- DelegateVoted
- DelegatePunished
- DelegateBanned

### How was it tested?

Successful build.

### DIFF
--- a/framework/src/modules/dpos_v2/events/delegate_banned.ts
+++ b/framework/src/modules/dpos_v2/events/delegate_banned.ts
@@ -1,0 +1,45 @@
+/*
+ * Copyright © 2022 Lisk Foundation
+ *
+ * See the LICENSE file at the top-level directory of this distribution
+ * for licensing information.
+ *
+ * Unless otherwise agreed in a custom licensing agreement with the Lisk Foundation,
+ * no part of this software, including this file, may be copied, modified,
+ * propagated, or distributed except according to the terms contained in the
+ * LICENSE file.
+ *
+ * Removal or modification of this copyright notice is prohibited.
+ */
+
+import { BaseEvent, EventQueuer } from '../../base_event';
+
+export interface DelegateBannedEventData {
+	address: Buffer;
+	height: bigint;
+}
+
+export const delegateBannedDataSchema = {
+	$id: '/pos/events/delegateBannedData',
+	type: 'object',
+	required: ['address', 'height'],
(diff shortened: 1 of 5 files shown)


framework/src/modules/dpos_v2/events/delegate_banned.ts
```diff
@@ -0,0 +1,47 @@
...
+export const delegateBannedDataSchema = {
+	$id: '/pos/events/delegateBannedData',
+	type: 'object',
+	required: ['address', 'height'],
+	properties: {
+		address: {
+			dataType: 'bytes',
+			fieldNumber: 1,
+			minLength: ADDRESS_LENGTH,
+			maxLength: ADDRESS_LENGTH,
```
shuse2, line 32: for address, please use `format: 'lisk32'`

framework/src/modules/dpos_v2/events/delegate_voted.ts
```diff
@@ -0,0 +1,65 @@
...
+			fieldNumber: 3,
+		},
+		result: {
+			dataType: 'uint32',
+			fieldNumber: 4,
+		},
+	},
+};
+
+export class DelegateVotedEvent extends BaseEvent<DelegateVotedEventData> {
```
shuse2, line 59: For the event which has `result` which can be used as persistent and non-persistent, it will be better to differentiate `log` and `error` similar to the token events

shuse2 (changes requested):

framework/src/modules/dpos_v2/events/delegate_banned.ts
shuse2: for the lisk32 format, the length is checked for the format, so no need to add `min/maxLength`

shuse2 (approved):

Incede (approved):

PR #7693, opened 2022-10-25
Update supported token store - Closes #7646

### LINKED ISSUE #7646: Update supported token store
### Description

Update supported token method/store to follow the logic defined in [LIP](https://github.com/LiskHQ/lips/blob/ddb478aaa0d2ad8d12857498e5890cdbc40d0fab/proposals/lip-0051.md#supportalltokens)

### Acceptance Criteria

Unit tests are updated for supported token store

mosmartin:
### What was the problem?

This PR resolves #7646

### How was it solved?

- Refactored the SupportedTokensStore class

### How was it tested?

- Updated the unit tests


framework/src/modules/token/stores/supported_tokens.ts
```diff
@@ -135,11 +149,17 @@ export class SupportedTokensStore extends BaseStore<SupportedTokensStoreData> {
 		}
 		const [chainID] = splitTokenID(tokenID);
 		let supported: SupportedTokensStoreData = { supportedTokenIDs: [] };
-		try {
-			supported = await this.get(context, chainID);
-		} catch (error) {
-			if (!(error instanceof NotFoundError)) {
-				throw error;
+		const supportExist = await this.has(context, chainID);
+		if (supportExist) {
```
shuse2, line 153: i think `has` is not necessary here because below `NotFoundError` is handled and if the data does not exist, it will use the default value above

framework/src/modules/token/stores/supported_tokens.ts
```diff
@@ -154,20 +174,25 @@ export class SupportedTokensStore extends BaseStore<SupportedTokensStoreData> {
...
-			return;
+			throw new Error('All tokens are supported.');
 		}
-		const [chainID] = splitTokenID(tokenID);
-		const supported = await this.get(context, chainID);
-		const index = supported.supportedTokenIDs.findIndex(id => id.equals(tokenID));
-		if (index < 0) {
-			throw new Error(`TokenID ${tokenID.toString('hex')} is not supported.`);
+		if (tokenID.equals(TOKEN_ID_LSK)) {
+			throw new Error('Cannot remove support for LSK token.');
```
shuse2, line 180: This is checked in `isMainchainOrNative`, so it should be thrown above

framework/src/modules/token/stores/supported_tokens.ts
```diff
@@ -154,20 +174,25 @@ export class SupportedTokensStore extends BaseStore<SupportedTokensStoreData> {
...
 		}
-		supported.supportedTokenIDs.splice(index, 1);
-
-		if (supported.supportedTokenIDs.length > 0) {
-			await this.set(context, chainID, supported);
-		} else {
-			await this.del(context, chainID);
+		const [chainID] = splitTokenID(tokenID);
+		const supportExist = await this.has(context, chainID);
+		if (supportExist) {
```
shuse2, line 184: Prefer early return

framework/src/modules/token/stores/supported_tokens.ts
```diff
@@ -154,20 +174,25 @@ export class SupportedTokensStore extends BaseStore<SupportedTokensStoreData> {
...
+			if (supported.supportedTokenIDs.length === 0) {
+				throw new Error('All tokens from the specified chain are supported.');
+			}
+			const index = supported.supportedTokenIDs.findIndex(id => id.equals(tokenID));
+			if (index >= 0) {
+				supported.supportedTokenIDs.splice(index, 1);
+			}
+			if (supported.supportedTokenIDs.length === 0) {
+				await this.del(context, chainID);
+			}
```
shuse2, line 195: i think test for these logic need to be added

shuse2 (changes requested):

framework/src/modules/token/stores/supported_tokens.ts
shuse2: since it's thrown above, safe to remove this code as well

framework/test/unit/modules/token/stores/supported_tokens.spec.ts
```diff
@@ -276,12 +276,40 @@ describe('SupportedTokensStore', () => {
...
 			await expect(store.has(context, Buffer.from([1, 1, 1, 1]))).resolves.toBeTrue();
+			// Checking that the store has the correct data.
 			await expect(store.get(context, Buffer.from([1, 1, 1, 1]))).resolves.toEqual({
 				supportedTokenIDs: [
 					Buffer.from([1, 1, 1, 1, 1, 0, 1, 1]),
+					tokenID,
 					Buffer.from([1, 1, 1, 1, 1, 0, 0, 1]),
 				],
 			});
 		});
```
shuse2, line 287: it would be nice to have a case where
3 tokens are supported and become 2 tokens with same order after removing

shuse2 (approved):

framework/test/unit/modules/token/stores/supported_tokens.spec.ts
mosmartin: Update from L279

framework/src/modules/token/stores/supported_tokens.ts
```diff
@@ -150,24 +161,30 @@ export class SupportedTokensStore extends BaseStore<SupportedTokensStoreData> {
 
 	public async removeSupportForToken(context: StoreGetter, tokenID: Buffer): Promise<void> {
 		if (this._isMainchainOrNative(tokenID)) {
-			return;
+			throw new Error('All tokens are supported.');
```
bobanm, line 164: I think this error should go to the section below, where `allSupported` is being checked.

The LIP only covers the case for trying to remove LSK token, but it makes sense to handle LSK and native tokens as 1 case and throw error: _Can not remove support for LSK or a native token_.

bobanm (changes requested):

framework/src/modules/token/stores/supported_tokens.ts
```diff
@@ -150,24 +161,30 @@ export class SupportedTokensStore extends BaseStore<SupportedTokensStoreData> {
...
 		}
 		const allSupported = await this.allSupported(context);
 		if (allSupported) {
 			return;
```
bobanm, line 168: Here it needs to throw 'All tokens are supported', which you actually throw in the case above, where user tries to remove LSK or native token.

framework/src/modules/token/stores/supported_tokens.ts
mosmartin: It was there initially until it was rightly pointed out that the private method `_isMainchainOrNative()` is already doing this check. Hence it's made more sense to throw the error where this private method is being invoked. WDYT?

bobanm: What I'm saying is that the error message here is wrong. We have 2 different cases which require 2 different error messages:

1. a user tries to remove LSK or a native token [which is what you're handling here]
1. a user tries to remove a token of a foreign chain, while the local chain supports all tokens [which is what this error message suggests]

When a user tries to remove LSK or a native token, I think it's incorrect to give error message that _"All tokens are supported."_. Because we don't know that before we check `allSupported()` method.

Also, LIP specifies error message _"Can not remove support for LSK token."_, which is not handled in your code. I propose that attempt to remove LSK or native token is handled with the same error message, because it is checked by the same function, but we can also separate that in 2 cases.

bobanm (approved):

PR #7695, opened 2022-10-25
Remove redundant Token module config interface - Closes #7694

### LINKED ISSUE #7694: Remove redundant Module Config interface from Token module
### Description

For some reason, there are 2 defined `ModuleConfig` interfaces used by `TokenModule`:

In `types.ts` there is:

```typescript
export interface ModuleConfig {
    minBalances: {
        tokenID: string;
        amount: string;
    }[];
    feeTokenID: string;
    accountInitializationFee: string;
    escrowInitializationFee: string;
}
```

In `methods.ts`, there is:

```typescript
interface TokenMethodConfig {
    minBalances: MinBalance[];
    ownChainID: Buffer;
    userAccountInitializationFee: bigint;
    escrowAccountInitializationFee: bigint;
    feeTokenID: Buffer;
}
```

Not only they use different data types, but the property names are different too.

Clean code design ensures that there are no duplicated versions of the same interface with different shape.

The implementation in `DPoSModule` look superior, as it defines only 1 config interface, and uses the internal data types [numbers, bigints and Buffers] rather than handling them as strings.

### Motivation

In order to prevent errors from maintain two version of config interfaces, we should stick to one, as implemented in `DPoSModule`.

bobanm:
### What was the problem?

This PR resolves #7694

### How was it solved?

1. Removed redundant `ModuleConfig` from `types.ts` and moved and renamed `TokenMethodConfig` from `method.ts` to `types.ts`.
1. Cleaned up schema and constants

### How was it tested?

Unit and integration tests are passing.

### DIFF
--- a/framework/src/modules/token/constants.ts
+++ b/framework/src/modules/token/constants.ts
@@ -38,8 +38,6 @@ export const TOKEN_ID_LSK = Buffer.from([0, 0, 0, 0, 0, 0, 0, 0]);
 export const USER_SUBSTORE_INITIALIZATION_FEE = BigInt(5000000);
 export const ESCROW_SUBSTORE_INITIALIZATION_FEE = BigInt(5000000);
 
-export const TOKEN_ID_STORE_INITIALIZATION = TOKEN_ID_LSK;
-
 export const defaultConfig = {
 	userAccountInitializationFee: USER_SUBSTORE_INITIALIZATION_FEE.toString(),
 	escrowAccountInitializationFee: ESCROW_SUBSTORE_INITIALIZATION_FEE.toString(),
--- a/framework/src/modules/token/constants.ts
+++ b/framework/src/modules/token/constants.ts
@@ -21,7 +21,7 @@ export const MIN_MODULE_NAME_LENGTH = 1;
 export const MAX_MODULE_NAME_LENGTH = 32;
 
 export const MAX_TRANSACTION_AMOUNT = '9223372036854775807';
-export const DEFAULT_MIN_REMAINING_BALANCE = '5000000';
+export const DEFAULT_MIN_REMAINING_BALANCE = BigInt(5000000);
 
 export const CCM_STATUS_OK = 0;
 export const CCM_STATUS_TOKEN_NOT_SUPPORTED = 64;
@@ -38,19 +38,20 @@ export const TOKEN_ID_LENGTH = CHAIN_ID_LENGTH + LOCAL_ID_LENGTH;
 export const LOCAL_ID_LSK = Buffer.alloc(LOCAL_ID_LENGTH, 0);
 export const CHAIN_ID_LSK = Buffer.from([0, 0, 0, 0]);
 export const TOKEN_ID_LSK = Buffer.from([0, 0, 0, 0, 0, 0, 0, 0]);
-export const USER_SUBSTORE_INITIALIZATION_FEE = '5000000';
-export const TOKEN_ID_STORE_INITIALIZATION = TOKEN_ID_LSK;
-
+export const USER_SUBSTORE_INITIALIZATION_FEE = BigInt(5000000);
 export const ESCROW_SUBSTORE_INITIALIZATION_FEE = BigInt(5000000);
+
+export const TOKEN_ID_STORE_INITIALIZATION = TOKEN_ID_LSK;
 
 export const defaultConfig = {
 	minBalances: [
 		{
 			tokenID: Buffer.alloc(TOKEN_ID_LENGTH, 0).toString('hex'),
-			amount: DEFAULT_MIN_REMAINING_BALANCE,
+			amount: String(DEFAULT_MIN_REMAINING_BALANCE),
 		},
 	],
-	accountInitializationFee: USER_SUBSTORE_INITIALIZATION_FEE,
+	userAccountInitializationFee: String(USER_SUBSTORE_INITIALIZATION_FEE),
+	escrowAccountInitializationFee: String(ESCROW_SUBSTORE_INITIALIZATION_FEE),
 };
 
 export const EMPTY_BYTES = Buffer.alloc(0);
--- a/framework/src/modules/token/constants.ts
+++ b/framework/src/modules/token/constants.ts
@@ -21,14 +21,11 @@ export const MIN_MODULE_NAME_LENGTH = 1;
 export const MAX_MODULE_NAME_LENGTH = 32;
 
 export const MAX_TRANSACTION_AMOUNT = '9223372036854775807';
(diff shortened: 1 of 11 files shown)


framework/src/modules/token/constants.ts
```diff
@@ -38,19 +38,20 @@ export const TOKEN_ID_LENGTH = CHAIN_ID_LENGTH + LOCAL_ID_LENGTH;
...
 export const ESCROW_SUBSTORE_INITIALIZATION_FEE = BigInt(5000000);
 
+export const TOKEN_ID_STORE_INITIALIZATION = TOKEN_ID_LSK;
+
 export const defaultConfig = {
 	minBalances: [
 		{
 			tokenID: Buffer.alloc(TOKEN_ID_LENGTH, 0).toString('hex'),
-			amount: DEFAULT_MIN_REMAINING_BALANCE,
+			amount: String(DEFAULT_MIN_REMAINING_BALANCE),
```
shuse2, line 50: To be consistent, i think it's better to use `toString` for the conversion

framework/src/modules/token/constants.ts
```diff
@@ -38,19 +38,20 @@ export const TOKEN_ID_LENGTH = CHAIN_ID_LENGTH + LOCAL_ID_LENGTH;
 export const LOCAL_ID_LSK = Buffer.alloc(LOCAL_ID_LENGTH, 0);
 export const CHAIN_ID_LSK = Buffer.from([0, 0, 0, 0]);
 export const TOKEN_ID_LSK = Buffer.from([0, 0, 0, 0, 0, 0, 0, 0]);
-export const USER_SUBSTORE_INITIALIZATION_FEE = '5000000';
-export const TOKEN_ID_STORE_INITIALIZATION = TOKEN_ID_LSK;
-
+export const USER_SUBSTORE_INITIALIZATION_FEE = BigInt(5000000);
 export const ESCROW_SUBSTORE_INITIALIZATION_FEE = BigInt(5000000);
 
+export const TOKEN_ID_STORE_INITIALIZATION = TOKEN_ID_LSK;
```
shuse2, line 44: maybe check if this value is used, if not we can remove them

framework/src/modules/token/types.ts
```diff
@@ -14,24 +14,25 @@
...
-}
-
 export interface MinBalance {
 	tokenID: Buffer;
 	amount: bigint;
 }
 
+export interface ModuleConfig {
+	minBalances: MinBalance[];
+	ownChainID: Buffer;
```
shuse2, line 28: ownChainID does not exist in the module config (it comes from the genesis config)

framework/src/modules/token/types.ts
```diff
@@ -14,24 +14,25 @@
...
-	escrowInitializationFee: string;
-}
-
 export interface MinBalance {
 	tokenID: Buffer;
 	amount: bigint;
 }
 
+export interface ModuleConfig {
+	minBalances: MinBalance[];
```
shuse2, line 27: could you check if minBalances is still used? if not we can just remove

shuse2 (changes requested):

framework/src/modules/token/constants.ts
bobanm: It seems it is used only in 1 single test case in `commands/transfer.spec.ts`, but nowhere in the app code:

_should pass if token balance for token ID TOKEN_ID_STORE_INITIALIZATION is at least the sum of configured initialization fee and transaction amount_

Should I move that constant from `constants.ts` to the test? Or, if it's not used anywhere in the code, should I just remove both the constant and the test case?

framework/src/modules/token/types.ts
bobanm: Yes, `ownChainID` is added to the module config during module init. The module needs to pass it to method, which uses it. Would it be better to remove it from `ModuleConfig` interface, and then merge `moduleConfig` object with an object that contains only `ownChainID` property, and pass that to `method.init()`?

framework/src/modules/token/types.ts
bobanm: I found it only to be used in tests. So I presume it is safe to remove it?

framework/src/modules/token/constants.ts
bobanm: Removed `TOKEN_ID_STORE_INITIALIZATION` from constants, removed the test case and changed value for `feeTokenID` id `Transfer` test suite to be `TOKEN_ID_LSK`.

framework/src/modules/token/types.ts
bobanm: These are my favorite kind of changes. I love removing unused code and making the codebase leaner 😀

shuse2: yes safe to remove as long as test pass =)

framework/src/modules/token/types.ts
shuse2: i think it would be confusing to have it in this interface because it looks like it's changeable from config, but it's not.
probably better to just have different interface for method

shuse2: but actually we have another issue to remove ownChainID, so it's fine to keep it as it is

bobanm: Moved `ownChainID` to `MethodConfig` interface.

framework/src/modules/token/constants.ts
bobanm: Yes, but casting using `String(value)` is more consistent with how we cast bigints with `BigInt(value)` 😉 

Anyhow, I changed it all to `toString()`

framework/test/unit/modules/token/commands/transfer.spec.ts
```diff
@@ -264,34 +257,6 @@ describe('Transfer command', () => {
 			);
 		});
 
-		it('should pass if token balance for token ID TOKEN_ID_STORE_INITIALIZATION is at least the sum of configured initialization fee and transaction amount', async () => {
```
shuse2, line 267: this test should still be valid

bobanm: It should use `TOKEN_ID_LSK` as token ID, instead of `TOKEN_ID_STORE_INITIALIZATION`, right?

shuse2: the test name should be
`should pass if token balance is at least the sum of configured initialization fee and transaction amount when feeTokenID and sending tokenID are the same`

bobanm: Updated

shuse2 (approved):

mosmartin (approved):

PR #7699, opened 2022-10-26
Updates PoS stores - Closes #7664

### LINKED ISSUE #7664: Update stores in DPoS module
### Description

- Update all the DPoS module stores as defined in the [LIP]().

### Acceptance Criteria

- All the build and tests must be passing

### Additional Information
- Further changes are coming to this issue

has5aan:
### What was the problem?

This PR resolves #7664 

### How was it solved?

Updated store definition and schema based on LIP-0057.

### How was it tested?

Updated existing tests around new schema.


shuse2 (approved):

framework/src/modules/dpos_v2/module.ts
```diff
@@ -336,6 +340,10 @@ export class DPoSModule extends BaseModule {
 				isBanned: dposValidator.isBanned,
 				pomHeights: dposValidator.pomHeights,
 				consecutiveMissedBlocks: dposValidator.consecutiveMissedBlocks,
+				// TODO: Issue: #7669
+				commission: 0,
+				lastCommissionIncreaseHeight: 0,
+				sharingCoefficients: [{ tokenID: Buffer.alloc(8), coefficient: Buffer.alloc(24) }],
```
bobanm, line 346: For the length of the buffers, instead of using number literals, it is much better practice to use constants `TOKEN_ID_LENGTH` and `MAX_NUM_BYTES_Q96` which you have already defined in constants anyway.

Not only here, but in all other parts of the code where you assign values to `sharingCoefficients` array.

Incede (approved):

framework/src/modules/dpos_v2/module.ts
has5aan: This is temporary as the issues referred will define intentional values.

Issue #7660, opened 2022-10-19
Token module metadata references request schema for response endpoint


bobanm:
Response should reference the response schema, not request schema. Probably a copy-paste error.

https://github.com/LiskHQ/lisk-sdk/blob/d7e2214a3a2f134e0f38d2e72401aacb5e70d95d/framework/src/modules/token/module.ts#L97

PR #7705, opened 2022-10-27
Update Token module endpoints - Closes #7579 #7660

### LINKED ISSUE #7579: Update token module endpoints
### Description

1. Add token module endpoint `getInitializationFees` which returns below using config
```
{
  "userAccount": "50000000",
  "escrowAccount": "50000000",
}
```

2. Update token module endpoints to use updated stores referring to [LIP stores](https://github.com/LiskHQ/lips/blob/98eafb761a4792cc0918996a13f72703629195ff/proposals/lip-0051.md#token-module-store)
3. Add token module endpoint `isSupported`
req
```
{
  tokenID: '',
}
```
res
```
{
  supported: true
}
```
4. Add token module endpoint `getSupportedTokens`
res
```
  supportedTokens: []
```
where array should include `['*']` when all tokens are supported. It should also include mainchain / native tokens.
If all native token from a chain is supported, it should return `['00000020********']`

### Acceptance Criteria

- It should be reflected in the metadata
- Unit test for the endpoint is added

### Additional Information
- This change requires #7573 and #7574 for the config values

bobanm:
### What was the problem?

This PR resolves #7579 #7660

### How was it solved?

* Fixed `TokenModule.init()` to get config from a `ModuleConfig` object
* Implemented `isSupported` endpoint
* Implemented `getInitializationFees` endpoint
* Partially implemented `getSupportedTokens` endpoint

### How was it tested?

Added unit tests. Skipped the one that should return false when checking if a not-supported token is supported.

### DIFF
--- a/framework/src/modules/token/module.ts
+++ b/framework/src/modules/token/module.ts
@@ -188,7 +188,7 @@ export class TokenModule extends BaseInteroperableModule {
 		this.stores.get(SupportedTokensStore).registerOwnChainID(this._ownChainID);
 		this.crossChainTransferCommand.init({ ownChainID: this._ownChainID });
 
-		this.method.init(Object.assign(config, { ownChainID: this._ownChainID }));
+		this.method.init({ ...config, ownChainID: this._ownChainID });
 		this.crossChainMethod.init(this._ownChainID);
 		this.endpoint.init(config);
 		this._transferCommand.init({
--- a/framework/src/modules/token/module.ts
+++ b/framework/src/modules/token/module.ts
@@ -25,8 +25,11 @@ import {
 	getBalanceRequestSchema,
 	getBalanceResponseSchema,
 	getBalancesRequestSchema,
+	getBalancesResponseSchema,
 	getEscrowedAmountsResponseSchema,
 	getSupportedTokensResponseSchema,
+	isSupportedRequestSchema,
+	isSupportedResponseSchema,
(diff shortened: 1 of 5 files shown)


framework/src/modules/token/module.ts
```diff
@@ -163,29 +171,30 @@ export class TokenModule extends BaseInteroperableModule {
...
 		this.stores.get(SupportedTokensStore).registerOwnChainID(this._ownChainID);
 		this.crossChainTransferCommand.init({ ownChainID: this._ownChainID });
 
-		this.method.init({
-			ownChainID: this._ownChainID,
-			escrowAccountInitializationFee: BigInt('50000000'),
-			feeTokenID: Buffer.from(config.feeTokenID, 'hex'),
-			userAccountInitializationFee: BigInt('50000000'),
-		});
+		this.method.init(Object.assign(config, { ownChainID: this._ownChainID }));
```
shuse2, line 191: you can use spread operator instead of `Object.assign`

shuse2 (commented):
LGTM for the existing implementation

mosmartin (approved):
🚀

framework/src/modules/token/module.ts
bobanm: Good point. Updated.

shuse2 (approved):

PR #7706, opened 2022-10-27
Implement getChainValidators and isChainIDAvailable - Closes #7596

### LINKED ISSUE #7596: Implement isChainIDAvailable and getChainValidators interoperability endpoints
### Description

**getChainValidators**
- Implement `getChainValidators` in BaseInteroperabilityStore and call it in the endpoint
- getChainValidators in BaseInteroperabilityStore
```python
def getChainValidators(chainID: ChainID) -> Validators:
    if chainAccount(chainID) does not exist:
        raise Exception('Chain account does not exist.')
    return validators(chainID)
```
- In endpoint return in JSON format

**isChainIDAvailable**
- Implement `isChainIDAvailable` in BaseInteroperabilityStore and call it in the endpoint
- isChainIDAvailable in BaseInteroperabilityStore
```python
def isChainIDAvailable(chainID: ChainID) -> bool:
    if chainAccount(chainID) does not exist:
        return True
    return False
```
- In endpoint return in JSON format

### Acceptance Criteria

- Should have all the related unit tests

### Additional Information

- Related LIP PR https://github.com/LiskHQ/lips/pull/148

mitsuaki-u:
### What was the problem?

This PR resolves #7596 

### How was it solved?

- getChainValidators
- isChainIDAvailable

### How was it tested?

Added unit test


Phanco (approved):

framework/src/modules/interoperability/base_interoperability_endpoint.ts
```diff
@@ -148,6 +149,26 @@ export abstract class BaseInteroperabilityEndpoint<
 		};
 	}
 
+	public async getChainValidators(
+		context: ModuleEndpointContext,
+		chainID: Buffer,
+	): Promise<ChainValidators> {
+		const interoperabilityStore = this.getInteroperabilityStore(context);
+		const chainValidators = await interoperabilityStore.getChainValidators(chainID);
```
ishantiw, line 157: Please use `ChainValidatorsStore` class instead of interoperabilityStore as we are removing it from store and will not be valid, so you will need to move the logic from store to here

framework/src/modules/interoperability/base_interoperability_endpoint.ts
```diff
@@ -148,6 +149,26 @@ export abstract class BaseInteroperabilityEndpoint<
...
+
+		return chainValidators;
+	}
+
+	public async isChainIDAvailable(
```
ishantiw, line 162: same here, use `ChainAccountStore` store directly instead of using `interoperabilityStore`

ishantiw (changes requested):

framework/src/modules/interoperability/base_interoperability_endpoint.ts
mitsuaki-u: Ok then the issue was not written correctly as it says: `Implement getChainValidators in BaseInteroperabilityStore and call it in the endpoint`.

mitsuaki-u: @ishantiw @Phanco As per our discussion, we have decided to remove both functions from base store class in lieu of the upcoming PR: https://github.com/LiskHQ/lisk-sdk/pull/7700.  The functions will remain in base endpoint.

Furthermore, the tests for the endpoints were removed in this PR as they will be implemented in PR: https://github.com/LiskHQ/lisk-sdk/pull/7700

ishantiw (approved):

PR #7709, opened 2022-10-30
Update PoS Delegate Registration command

### LINKED ISSUE #7665: Update delegate registration command
### Description

- Update delegate registration command to include the events and new property according to the [LIP](https://github.com/LiskHQ/lips/blob/main/proposals/lip-0057.md#delegate-registration).
- Check current verify/execute logic is aligned with the LIP

### Acceptance Criteria

- Add tests for new property handling and the events

bobanm:
### What was the problem?

This PR resolves #7665

### How was it solved?

1. Added `delegateRegistrationFee` property to `DelegateRegistrationCommand`
1. Added missing verification steps to `DelegateRegistrationCommand.verify()`
1. Added burning of fees to `DelegateRegistrationCommand.execute()`
1. `DelegateRegistrationCommand` now emits `DelegateRegisteredEvent`

### How was it tested?

* Added a test case which validates that event `DelegateRegisteredEvent` was emitted. It checks both that the emit request was correctly made, and that the event is in the event queue.
* Added a test case which throws an error if provided delegate registration fee is not the same amount as defined in config.
* Added a test case which throws an error if user does not have enough balance to pay delegate registration fee.


framework/src/modules/dpos_v2/commands/delegate_registration.ts
```diff
@@ -52,30 +56,47 @@ export class DelegateRegistrationCommand extends BaseCommand {
...
+			return {
+				status: VerifyStatus.FAIL,
+				error: new Error('Invalid delegate registration fee.'),
+			};
+		}
+
+		const balance = await this._tokenMethod.getAvailableBalance(
+			context,
+			transaction.senderAddress,
+			TOKEN_ID_FEE,
```
shuse2, line 69: similar to token module, this should come from the config

framework/src/modules/dpos_v2/schemas.ts
```diff
@@ -12,36 +12,35 @@
...
 		name: {
 			dataType: 'string',
 			fieldNumber: 1,
-			minLength: 1,
-			maxLength: MAX_LENGTH_NAME,
 		},
-		generatorKey: {
+		blsKey: {
 			dataType: 'bytes',
+			length: BLS_PUBLIC_KEY_LENGTH,
```
shuse2, line 28: `length` is not valid JSON schema. Please keep `minLength`, `maxLength`

bobanm: I copied that from the LIP:

https://github.com/LiskHQ/lips-staging/blob/main/proposals/lip-0057.md#parameters

The LIP is then incorrect and should be updated?

shuse2: LIP is just saying the length should be as defined. How it's validated is up to the implementation.
the implementation wise, it needs to use min/max because the Ajv we use don't have `length` property

framework/src/modules/dpos_v2/commands/delegate_registration.ts
```diff
@@ -52,6 +61,25 @@ export class DelegateRegistrationCommand extends BaseCommand {
...
+		);
+		if (balance < DELEGATE_REGISTRATION_FEE) {
```
shuse2, line 76: `DELEGATE_REGISTRATION_FEE` is also configurable constants, so it should come from config

framework/src/modules/dpos_v2/commands/delegate_registration.ts
```diff
@@ -113,15 +146,19 @@ export class DelegateRegistrationCommand extends BaseCommand {
 			isBanned: false,
 			pomHeights: [],
 			consecutiveMissedBlocks: 0,
-			// TODO: Issue #7665
-			commission: 0,
-			lastCommissionIncreaseHeight: 0,
-			sharingCoefficients: [{ tokenID: Buffer.alloc(8), coefficient: Buffer.alloc(24) }],
+			commission: 10000,
```
shuse2, line 149: maybe define constant?

bobanm: A local constant, or a constant in default config, and then get the value from config, like for `DELEGATE_REGISTRATION_FEE`?

shuse2: Probably both are valid for this, but i think constant like `DELEGATE_REGISTRATION_FEE ` is fine for now

shuse2 (approved):

has5aan (approved):

